PR #1989, opened 2022-03-01
Support pre-existing Gen-ZAM executables

### LINKED ISSUE #1963: Cross-compiling zeek  && Gen-ZAM
When I was cross-compiling, I encountered the following problem:
[ 84%] [sh] Generating ZAM operations
/bin/sh: 1: /home/ubuntu/tmp/zeek-4.1.1/build/src/Gen-ZAM: Exec format error
make[3]: *** [src/CMakeFiles/zeek_objs.dir/build.make:128：src/ZAM-AssignFlavorsDefs.h] 错误 2
make[3]: 离开目录“/home/ubuntu/tmp/zeek-4.1.1/build”
make[2]: *** [CMakeFiles/Makefile2:4627：src/CMakeFiles/zeek_objs.dir/all] 错误 2
make[2]: 离开目录“/home/ubuntu/tmp/zeek-4.1.1/build”
make[1]: *** [Makefile:163：all] 错误 2
make[1]: 离开目录“/home/ubuntu/tmp/zeek-4.1.1/build”

Briefly describe our compilation process：
1、./configure --builddir=../zeek-buildtools
2、cd ../zeek-buildtools
3、make binpac bifcl
4、cd zeek
5、./configure --toolchain=~/tmp/zeek-4.1.1/toolchain.cmake --with-binpac=$(pwd)/../zeek-buildtools/auxil/binpac/src/binpac --with-bifcl=$(pwd)/../zeek-buildtools/auxil/bifcl/bifcl
6、make

**toolchain.cmake:**

set(CMAKE_SYSTEM_NAME Linux)

set(toolchain /home/ubuntu/compiler/loongarch64-linux-gnu-2022-01-26-vector)
set(CMAKE_C_COMPILER   ${toolchain}/bin/loongarch64-linux-gnu-gcc)
set(CMAKE_CXX_COMPILER ${toolchain}/bin/loongarch64-linux-gnu-g++)

set(CMAKE_FIND_ROOT_PATH
    /home/ubuntu/openwrt-21.02/staging_dir/target-loongarch64-linux-gnu_glibc
    ${toolchain}/loongarch64-linux-gnu/sysroot
)

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)

The Gen-ZAM executable is in the format of my target platform, but when cross-compiling it executes.
Can you help me analyze the problem?

ckreibich:
This allows reuse of a previously built `Gen-ZAM` via `--with-gen-zam=...`, per the commits and NEWS blurb. It does not currently install `Gen-ZAM` in the distribution, which would be more consistent with `binpac` and `bifcl`, but I'm not sure it's warranted just for the cross-compilation use case. Let me know if you'd prefer this — it's a minor delta on top of this changeset.

Resolves #1963.


timwoj (commented):
How would this work then? The user would run a build to get the gen-zam executable and then run a second build using it?

ckreibich:
Yes that's right. The approach sketched in #1963 is basically just that: a separate builddir configuration in which you selectively build the tools you need locally on the build system. You then configure the cross-compilation by pointing `configure` at those pre-built tools.

My thinking re. `Gen-ZAM` was that it's currently only of use to the Zeek build, and hence is not worth installing alongside `binpac`/`bifcl`. We could easily do that, it's just that most people would never need it. Another option would be to investigate how to make the build system smart enough to transparently build `Gen-ZAM` for the build, not the target, but doing that alone would be inconsistent with `binpac`/`bifcl` too. It also sounds hairy to me because there's just no awareness of two compiler setups anywhere in the configuration.

rsmmr:
I think this should work, I'm just wondering if for consistency we should relocate `gen-zam` into it's own little CMake project in `zeek-aux`, just as we do for `bifcl`. Everything else would stay the same. That way people would have full control over the `gen-zam` build if they need to, and it'd be all similar to how `bifcl` and `binpac` are structured (except that the latter got its own submodule).

ckreibich:
I like that, yeah. I just noticed that `bifcl` already has its own submodule ... so then it'd be most consistent to also have one for `Gen-ZAM`, yes?

rsmmr:
> I like that, yeah. I just noticed that `bifcl` already has its own submodule ... so then it'd be most consistent to also have one for `Gen-ZAM`, yes?

Ah, missed that. Then I guess yes, though not strictly necessary I'd say.

ckreibich:
I've started putting this together after double-checking with @vpax. I'll ready the submodule in https://github.com/ckreibich/gen-zam and will move it over when ready. I've preserved the commit history on Gen-ZAM. Looks pretty straightforward.

ckreibich:
(Moved this back to WIP stage as I assemble the pieces)

timwoj (approved):

timwoj:
It looks like merging this broke the package builds on OBS. Here's just one example, but the others are similar: https://build.opensuse.org/package/live_build_log/security:zeek/zeek-nightly/openSUSE_Leap_15.2/x86_64

0xxon:
Yeah, that's my fault - was a bit busy for a few days :)

I will fix this on... Monday.

0xxon:
This should be fixed with the next build (which will complete in a couple of hours).

PR #1993, opened 2022-03-04
Improve error message when receiving unexpected record content via Broker

rsmmr:
Broker sends record values as a vector of the corresponding fields. If
the received values doesn't match the actual record fields, Zeek has
so far been reporting this not-so-helpful error message:

    warning: failed to convert remote event 'got_result' arg #1,  got vector, expected record

This change turns the latter part into a list of the field types we
actually received, which makes debugging much easier.

(No test, seems overkill to add a spawn a Broker client to exercise
this.)

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -1391,17 +1391,35 @@ void Manager::ProcessEvent(const broker::topic& topic, broker::zeek::Event ev)
 		{
 		auto got_type = args[i].get_type_name();
 		const auto& expected_type = arg_types[i];
-		auto val = detail::data_to_val(std::move(args[i]), expected_type.get());
+		auto val = detail::data_to_val(args[i], expected_type.get());
 
 		if ( val )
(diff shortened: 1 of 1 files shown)


src/broker/Manager.cc
```diff
@@ -1391,17 +1391,35 @@ void Manager::ProcessEvent(const broker::topic& topic, broker::zeek::Event ev)
...
-			                  name.data(), i, got_type, expected_name);
+			if ( strcmp(expected_name, "record") == 0 && strcmp("vector", got_type) == 0 )
+				{
+				// This means the vector elements didn't align with the record
+				// fields. Produce an error message that shows what we
+				// received.
+				std::string elements;
+				for ( const auto& e : broker::get<broker::vector>(args[i]) )
+					{
+					if ( elements.size() )
```
timwoj, line 1411: Nit: use `! elements.empty()`

timwoj (approved):
Looks good to me, just one nit.

Issue #1752, opened 2021-08-26
ASAN error for table initialization in global scope with &optional in keys


awelzel:
Hello,

while trying to reproduce a different issue, using `&optional` in records used as table keys and global initialization triggers an ASAN error with Zeek 4.1:

```
module Crash;

type Key: record {
  k0: string;
  k1: string &optional;
};

global init_key = [$k0="x"];

global state: table[Key] of count = {
  [init_key] = 5,
};

event zeek_init() {
  print(fmt("init_key in state: %d", init_key in state));
}
```

* It does not crash when `k1` is initialized.
* It does not crash when `state` is moved into `zeek_init()` as `local`.

```
$ ZEEKPATH=$(./build/zeek-path-dev ) ./build/src/zeek --version
./build/src/zeek version 4.1.0-debug
$ ZEEKPATH=$(./build/zeek-path-dev ) ./build/src/zeek ./optional-fields-table-key.zeek 
=================================================================

==17089==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603001964f18 at pc 0x56349c961e4c bp 0x7ffdb6343a70 sp 0x7ffdb6343a68
READ of size 1 at 0x603001964f18 thread T0
    #0 0x56349c961e4b in std::_Optional_base<zeek::ZVal, true, true>::_M_is_engaged() const /usr/include/c++/8/optional:897
    #1 0x56349c961e4b in std::optional<zeek::ZVal>::operator bool() const /usr/include/c++/8/optional:1205
    #2 0x56349c961e4b in zeek::RecordVal::HasField(int) const /home/awelzel/corelight-oss/zeek/src/zeek/Val.h:1201
    #3 0x56349c961e4b in zeek::RecordVal::GetField(int) const /home/awelzel/corelight-oss/zeek/src/zeek/Val.h:1223
    #4 0x56349c961e4b in zeek::detail::CompositeHash::SingleTypeKeySize(zeek::Type*, zeek::Val const*, bool, int, bool, bool) const /home/awelzel/corelight-oss/zeek/src/CompHash.cc:524
    #5 0x56349c9632d4 in zeek::detail::CompositeHash::ComputeKeySize(zeek::Val const*, bool, bool) const /home/awelzel/corelight-oss/zeek/src/CompHash.cc:648
[...]

PR #1999, opened 2022-03-12
Integrating = / += / -= expressions with initializers

vpax:
For historical reasons Zeek has had two completely separate code paths for initializing global variables using `= { .... }` as well as `redef`'ing using `+=` and `-=`, versus the implementation of the `=`/`+=`/`-=` expressions.  This disconnect has led to diverging semantics - such as the ability to use "cross-product" lists in initializers to expand out complex initializations - as well as varying support for things like coercions, per the discussion in https://github.com/zeek/zeek/issues/1752 (which the changes here can close).

This PR unifies the two approaches.  It extends the semantics of `=`/`+=`/`-=` expressions to allow for `{ .... }` initializer lists on the RHS, and treats them as arguments to a constructor whose type matches that of the LHS.  For full compatibility, it also (1) supports cross-product initializer lists, (2) enables direct `+=` for `table`, `set`, `vector` and `pattern` values when the RHS's type matches that of the LHS, (3) in a similar fashion enables direct `-=` for `table` and `set` values, and (4) supports listing multiple sets for `+=` initialization.  Doing so enables us to rip out nearly as much code as required for supporting this change, simplifying a number of interfaces.

Finally, I discovered to my amusement/horror that Zeek supports global initialization without specifying _any_ operator.  For example:
> global x "foobar";

infers a type of `string` for `x` and initializes it to `"foobar"`.  This PR still supports that behavior but flags it as deprecated.

There will be a companion PR shortly for updating the documentation.  I'll link to it here when it's ready.


timwoj:
I have a couple of nits that I'll just fix during the merge. I'm waiting to fix CI builds before I merge this though just to make sure everything is green there.

PR #2003, opened 2022-03-19
Quick check to see whether the ZEEK_BOT_TOKEN exists

ckreibich:
Ignore — this won't get merged. I'm checking the behavior of ZEEK_BOT_TOKEN which still seems to have problems.


ckreibich:
I updated the token in `zeek/zeek` once more (I had kept its value locally after yesterday's refresh) and a PR-driven CI run can access it just fine, per this PR's run. If docs generation still has a problem tomorrow, it's not about the token's existence but about passing it into `action/checkout`.

Closing this now.

Issue #1976, opened 2022-02-21
Integrate Spicy into Zeek


rsmmr:
We'll add both the Spicy toolchain and the Spicy plugin as submodules to Zeek. They'll be built by default, with the plugin compiled into Zeek statically. We'll add a `configure` option to alternatively use external versions of Spicy and/or plugin instead, but I don't think we want an option to disable Spicy altogether. The plan is to ensure that from 5.0 onwards, Spicy will always be available in Zeek.

On the Spicy side, we'll start maintaining an LTS branch that the Zeek LTS version can rely on. For Zeek feature releases, we'll always pull in the most recent Spicy feature release.

Note that the Zeek binary distributions will need to be updated as well to show Spicy.

PR #2006, opened 2022-03-23
Provide Spicy with Zeek by default

bbannier:


ci/centos-7/Dockerfile
```diff
@@ -38,11 +44,21 @@ RUN yum -y install \
...
 
+# Install a recent flex for Spicy.
+# ADD https://github.com/westes/flex/releases/download/v2.6.4/flex-2.6.4.tar.gz /tmp/flex
+# RUN curl -sSL "https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.tar.gz" | tar xzf - -C "${CMAKE_DIR}" --strip-components 1 \
+RUN curl -sSL "https://github.com/westes/flex/releases/download/v${FLEX_VERSION}/flex-${FLEX_VERSION}.tar.gz" | tar xzf - -C /tmp \
+ && (cd /tmp/flex-${FLEX_VERSION} \
+     && ./configure --prefix=${FLEX_DIR} \
+     && make -j`nproc` install) \
+ && rm -rf /tmp/flex-${FLEX_VERSION}
+
```
bbannier, line 55: Spicy heavily uses object libraries (both to propagate flags as well as link dependencies). Support for that is incomplete with the CMake older than 3.15; Zeek OTOH currently requires CMake from the range 3.5...3.18.

.cirrus.yml
```diff
@@ -272,15 +280,15 @@ sanitizer_task:
...
     # AddressSanitizer uses a lot more memory than a typical config.
-    memory: 12GB
+    memory: 16GB
   << : *CI_TEMPLATE
   test_fuzzers_script: ./ci/test-fuzzers.sh
   coverage_script: ./ci/upload-coverage.sh
   env:
     CXXFLAGS: -DZEEK_DICT_DEBUG
-    ZEEK_CI_CONFIGURE_FLAGS: *SANITIZER_CONFIG
+    ZEEK_CI_CONFIGURE_FLAGS: *SANITIZER_CONFIG *WITHOUT_SPICY_CONFIG
```
bbannier, line 291: Currently `ZEEK_INCLUDE_PLUGINS` does not seem to correctly set up link dependencies for `zeek_fuzzer_shared` and with Spicy enabled we'd see this error:

```
/usr/bin/ld: libzeek_fuzzer_shared.so: undefined reference to `plugin::_Zeek_Spicy::__bif_functions_init(zeek::plugin::Plugin*)'
/usr/bin/ld: libzeek_fuzzer_shared.so: undefined reference to `plugin::_Zeek_Spicy::__bif_consts_init(zeek::plugin::Plugin*)'
/usr/bin/ld: libzeek_fuzzer_shared.so: undefined reference to `plugin::_Zeek_Spicy::__bif_events_init(zeek::plugin::Plugin*)'
```

.gitmodules
```diff
@@ -52,3 +52,9 @@
 [submodule "auxil/gen-zam"]
 	path = auxil/gen-zam
 	url = https://github.com/zeek/gen-zam
+[submodule "auxil/spicy-plugin"]
+	path = auxil/spicy-plugin
+	url = https://github.com/zeek/spicy-plugin
+[submodule "auxil/spicy"]
+	path = auxil/spicy/spicy
+	url = https://github.com/zeek/spicy
```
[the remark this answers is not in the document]
bbannier: We will also need zeek/spicy#1157.

bbannier:
Closes #1976.

CMakeLists.txt
```diff
@@ -374,6 +374,52 @@ else ()
...
+        # as part of Zeek so spicy-plugin cannot use `spicy-config` at configure
+        # time to set these.
+        set(SPICY_CONFIG "<builtin>")
+        set(HAVE_SPICY "YES")
+        set(SPICY_HAVE_TOOLCHAIN "YES")
+        set(SPICY_INCLUDE_DIRS_RUNTIME
+            ${CMAKE_SOURCE_DIR}/auxil/spicy/spicy/hilti/runtime/include
+            ${CMAKE_SOURCE_DIR}/auxil/spicy/spicy/spicy/runtime/include
+            ${CMAKE_BINARY_DIR}/auxil/spicy/spicy/include)
+        set(SPICY_INCLUDE_DIRS_TOOLCHAIN
```
bbannier, line 391: On macos with Homebrew one currently needs to unlink both the `zeek` and `spicy` packages if they are installed as otherwise the spicy-plugin build picks up incorrect include paths. Let's check whether we can have our own include paths earlier in the path so that issue does not come up anymore.

testing/btest/spicy/spicyc.spicy
```diff
@@ -0,0 +1,11 @@
+# @TEST-DOC: Smoke test for a bundled Spicy.
+#
+# @TEST-REQUIRES: test -x ${BUILD}/auxil/spicy/spicy/bin/spicy-dump
```
rsmmr, line 3: I'm wondering if we should skip the `requires` to catch of spicy didn't build for some reason. I believe generally the Zeek tests assume having a default built, without specific `configure` options, so Spicy should now always be available.

testing/btest/spicy/spicyz.test
```diff
@@ -0,0 +1,26 @@
+# @TEST-DOC: Smoke test for a custom Spicy analyzer hooked into Zeek.
+#
+# @TEST-REQUIRES: test -x ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz
```
rsmmr, line 3: Same questions about `requires`.

testing/btest/btest.cfg
```diff
@@ -32,6 +32,8 @@ ZEEK_DISABLE_ZEEKYGEN=1
 ZEEK_ALLOW_INIT_ERRORS=1
 ZEEK_SUPERVISOR_NO_SIGKILL=1
 UBSAN_OPTIONS=print_stacktrace=1
+SPICY_PATH=%(testbase)s/../../auxil/spicy/spicy/spicy/lib:%(testbase)s/../../auxil/spicy/spicy/hilti/lib
+HILTI_CXX_INCLUDE_DIRS=%(testbase)s/../../auxil/spicy/spicy/hilti/runtime/include:%(testbase)s/../../auxil/spicy/spicy/spicy/runtime/include:%(testbase)s/../../build/auxil/spicy/spicy/include:%(testbase)s/../../build:%(testbase)s/../../build/src:%(testbase)s/../../build/src/include:%(testbase)s/../../src:%(testbase)s/../../auxil/broker/include
```
rsmmr, line 36: Any chance to do without these? That seems not only fragile but may also hide problems. Generally, the Spicy tools should be able to figure this out on their own (though maybe that doesn't work for the Zeek-internal build)?

auxil/spicy/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+if ( NOT BISON_EXECUTABLE )
+   find_package(BISON REQUIRED)
+endif ()
+
+get_filename_component(dir ${BISON_EXECUTABLE} DIRECTORY ABSOLUTE)
+set(BISON_ROOT ${dir}/..)
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-missing-braces")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-vla")
```
rsmmr, line 21: This is portable across compilers?

auxil/spicy/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Spicy needs the full prefix flex and bison while Zeek captures only the paths
+# to the executables. Derive the prefixes from the binary paths if given under
+# the assumption that their bindir is under their prefix (which also implies
+# that one such prefix even exists).
```
rsmmr, line 4: Could this move either into Spicy's cmake, or into Zeek higher-level CMake? I think it would be preferable to avoid this additional layer of CMake tweaking the Spicy build here, and then just have `auxil/spicy` be the submodule directly, just to remove complexity and follow the layout/structure for other submodules.

testing/btest/broker/remote_event.zeek
```diff
@@ -1,3 +1,5 @@
+# @TEST-REQUIRES: test -x ${BUILD}/auxil/spicy/spicy/bin/spicy-config
+#
```
rsmmr, line 2: Why do the Broker tests care about `spicy-config`?

testing/btest/spicy/spicyz.test
```diff
@@ -0,0 +1,26 @@
...
+
+public type Dummy = unit {
+    # Consume all data. We split data into lines and log the number of lines and the lines when done.
+    data: bytes &eod &convert=$$.split(b"\r\n");
+
+    on %done { print |self.data|, self; }
+};
+# @TEST-END-FILE
+
+# @TEST-START-FILE test.evt
```
rsmmr, line 22: Let's add an event here, too, plus then a Zeek script handling the event. Just to make sure the whole pipeline works.

rsmmr (commented):
A first set of comments:


- Running from the build directory has trouble finding paths for me:

```
(. zeek-path-dev.sh && spicyz -o test.hlto test.spicy test.evt)
[error] /Users/robin/work/zeek/spicy/build/test.spicy:1:1-10:3: cannot import module 'spicy_rt': cannot find file
[error] /Users/robin/work/zeek/spicy/build/test.spicy:1:1-10:3: cannot import module 'hilti': cannot find file
[error] <Spicy Plugin for Zeek>: errors encountered during resolving
```

(Wondering if this is related to why the `btest.cfg` need all the paths too).

- Have you tried installing a `zkg` analyzer package using this branch? (Just asking, I haven't tried it yet.)

- On installation, the Spicy cmake files land in `${PREFIX}/lib/zeek-spicy/cmake` for me, vs. Zeek's cmake files going into `${PREFIX}/share/zeek/cmake`, vs. `spicy-config --cmake-path` showing `${PREFIX}/share/spicy/cmake` (the latter is probably the correct place).

- I know we have been going back and forth on static vs dynamic linking. Looking at Broker and Binpac, I'm now wondering if we should link dynamically by default and provide a `configure` option `--enable-static-spicy` to switch to a static build; that's how those two do it.

CMakeLists.txt
bbannier: Will add a patch addressing this.

auxil/spicy/CMakeLists.txt
bbannier: This mirrors how the flag is enabled, https://github.com/zeek/cmake/blob/b1c9e55d1e837d46fc36312b567245013cf9646a/SetDefaultCompileFlags.cmake#L10:

```.cmake
set(EXTRA_COMPILE_FLAGS_CXX "-Wno-register -Werror=vla")
```

testing/btest/broker/remote_event.zeek
bbannier: The output of Broker's `num_logs_outgoing` depends on whether Spicy is loaded or not (with Spicy loaded it is one larger). This appears to be deterministic, but looking over the implementation for that counter it is unclear to me why it would depend on Spicy, so I opted to disable the tests instead of digging further.

testing/btest/spicy/spicyc.spicy
bbannier: We currently disable Spicy for sanitizer builds (there are three of them now). Spicy doesn't handle sanitizers other than `address` well since currently it assumes that if a sanitizer build is enabled as `address` being active.

Let's keep this conditional for now. Like discussed offline I'll introduce a dedicated script to detect whether Spicy was enabled during configure time.

testing/btest/btest.cfg
bbannier: See above.

testing/btest/spicy/spicyz.test
bbannier: See above.

0xxon:
I just realized that we should probably try building this branch on OBS before we merge it, to see if this will work with binary packages. Doing that is  not really super complicated - just let me know when it is in a state that you think is ready-ish for testing, and I will trigger a build in a branch on OBS.

rsmmr:
I tried it both from an install path and from inside the build directory, and either way works well for me now. 

Couple small things I noticed:

- The `zeek-path-dev.*` wrappers ends up with odd quotation marks for `PATH`: ```setenv PATH "/Users/robin/work/zeek/master/build/src":"/Users/robin/work/zeek/master/build/auxil/spicy/spicy/bin":"/Users/robin/work/zeek/master/build/src/builtin-plugins/spicy-plugin/bin/":$PATH```
- I see that `spicy-path.in` and `hilti-cxx-include-dirs.in` live in `ci/`, which doesn't feel right. We have other `.in` at the top-level (not great, but would be consistent); or maybe move them right into the Spicy plugin?

rsmmr:
I'm good otherwise, but I think it'd be helpful for somebody else to do a review for the merge here. @timwoj up for it?

CMakeLists.txt
```diff
@@ -616,6 +668,53 @@ CheckOptionalBuildSources(auxil/zeek-aux  Zeek-Aux  INSTALL_AUX_TOOLS)
...
+
+        # Use the debug runtime libraries if we are building a debug Zeek.
+        if ( SPICY_BUILD_MODE STREQUAL "Debug" )
+            list(TRANSFORM _spicy_rt_libs APPEND "-debug")
+        endif ()
+
+        # If we build spicy-plugin into Zeek we also need to build the Spicy
+        # runtime dependencies into Zeek. Since no matter how Spicy itself was
+        # linked this is always a static library, link the object files so we
+        # get all sybols and can can resolve all potential dependencies of HLTO
```
timwoj, line 683:
```suggestion
        # get all symbols and can can resolve all potential dependencies of HLTO
```

.gitmodules
```diff
@@ -58,3 +58,9 @@
 [submodule "auxil/out_ptr"]
 	path = auxil/out_ptr
 	url = https://github.com/soasis/out_ptr.git
+[submodule "auxil/spicy-plugin"]
+	path = auxil/spicy-plugin
+	url = https://github.com/zeek/spicy-plugin
```
timwoj, line 63: Have the commits used for these been given a tag? I think we said we were going to try to pin to specific release tags, right?

ci/centos-7/Dockerfile
```diff
@@ -17,13 +23,13 @@ RUN yum -y install \
   && yum clean all && rm -rf /var/cache/yum
 
 RUN yum -y install \
-    devtoolset-7 \
+    devtoolset-8 \
```
timwoj, line 26: We need to make sure to update the OS/Compiler support matrix for this change. Similar questions with all of the compiler version changes later.

ci/centos-7/Dockerfile
```diff
@@ -38,11 +44,19 @@ RUN yum -y install \
...
 
 RUN echo 'unset BASH_ENV PROMPT_COMMAND ENV' > /usr/bin/zeek-ci-env && \
-    echo 'source /opt/rh/devtoolset-7/enable' >> /usr/bin/zeek-ci-env
+    echo 'source /opt/rh/devtoolset-8/enable' >> /usr/bin/zeek-ci-env && \
+    echo 'export PATH=${PATH}:${FLEX_DIR}/bin' >> /usr/bin/zeek-ci-env
 
 ENV BASH_ENV="/usr/bin/zeek-ci-env" \
     ENV="/usr/bin/zeek-ci-env" \
-    PROMPT_COMMAND=". /usr/bin/zeek-ci-env"
+    PROMPT_COMMAND=". /usr/bin/zeek-ci-env" \
```
timwoj, line 62: Errant backslash, or is that actually needed?

ci/centos-7/Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM centos:7
 
+ENV FLEX_VERSION=2.6.4 \
+    FLEX_DIR=/opt/flex
```
timwoj, line 4: Consistency with the other files says both of the lines should start with `ENV`.

Nitpicky, I know.

ci/debian-10/Dockerfile
```diff
@@ -2,9 +2,11 @@ FROM debian:10
 
 ENV DEBIAN_FRONTEND="noninteractive" TZ="America/Los_Angeles"
 
+ENV CMAKE_DIR "/opt/cmake"
+ENV CMAKE_VERSION "3.19.1"
```
timwoj, line 6: I vaguely remember conversations about this, but are we bumping the minimum cmake required to build Zeek in general?

ci/macos/prepare.sh
```diff
@@ -7,4 +7,4 @@ set -x
 
 brew update
 brew upgrade cmake openssl@1.1
-brew install swig bison
+brew install swig bison flex
```
timwoj, line 10: https://docs.zeek.org/en/current/install.html#prerequisites should be updated to mention installing flex via brew.

ci/centos-7/Dockerfile
bbannier: Not yet. Depending on feedback to this PR more changes may need to go into Spicy or spicy-plugin. Before the next Zeek release we should have cut releases of Spicy and spicy-plugin though, so I am leaving this open for now.

ci/debian-10/Dockerfile
bbannier: Since the decision was to enable Spicy support by default we will need to bump the required CMake version. I'll update `cmake_minimum_required` in `CMakeLists.txt` for that.

ci/macos/prepare.sh
bbannier: There are a number of docs issues outstanding, I'll collect them here:

- [ ] update prerequistes (CMake, Bison, Flex version, ...)
- [ ] add a minimal (trampoline?) section on Spicy to the Zeek docs which forwards to the Spicy docs

.gitmodules
bbannier: Not yet. Depending on feedback to this PR more changes may need to go into Spicy or spicy-plugin. Before the next Zeek release we should have cut releases of Spicy and spicy-plugin though, so I am leaving this open for now.

ci/centos-7/Dockerfile
bbannier: This PR still requires docs changes, see [below](https://github.com/zeek/zeek/pull/2006#discussion_r865184038)

timwoj:
Docs updates withstanding, this looks good to me.

timwoj:
Also, this probably could do with a big entry in the NEWS file.

ci/macos/prepare.sh
bbannier: Docs update is zeek/zeek-docs#117.

timwoj (approved):

timwoj:
Checking if @rsmmr wants to make one more pass at this. If not I'll go ahead and merge it this week.

NEWS
```diff
@@ -6,12 +6,25 @@ release. For an exhaustive list of changes, see the ``CHANGES`` file
...
 Breaking Changes
 ----------------
 
 New Functionality
 -----------------
 
+- Zeek now includes support for building Spicy and spicy-plugin as part of
+  Zeek. This feature is enabled by default, and can be turned off by passing
+  the ``--disable-spicy`` flag to ``./configure``.
+
```
rsmmr, line 27:
```suggestion
- Zeek now includes Spicy and its corresponding Zeek plugin. Spicy is a parser generator 
  that makes it easy to create robust parsers for network protocols and file formats, 
  without writing any C++ code. For more information, see: 
  https://docs.zeek.org/projects/spicy. The Spicy integration is enabled by default, and 
  can be turned off by passing the ``--disable-spicy`` flag to ``./configure``.
```

rsmmr (approved):

PR #2012, opened 2022-03-26
Fix incorrect infinite loop instructions added by lower versions of GCC

AmazingPP:
After running zeek for nearly 20 days, the process is not responding. I used GDB to attach the zeek process and found that zeek entered an infinite loop.

`DNS_Interpreter::ParseRR_EDNS`
```asm
    6cd4:	c7 45 00 01 00 00 00 	mov    DWORD PTR [rbp+0x0],0x1
    6cdb:	c7 45 00 01 00 00 00 	mov    DWORD PTR [rbp+0x0],0x1
    6ce2:	eb f0                	jmp    6cd4 <_ZN4zeek8analyzer3dns6detail15DNS_Interpreter12ParseRR_EDNSEPNS2_11DNS_MsgInfoERPKhRiiS7_+0xf4>
```
But when I look at the C++ source code, I can't find where these three lines of instructions come from.

So, I use IDA to decompile DNS.cc.o. Found that GCC 7.3.1 added a strange infinite loop instruction before calling `ExtractShort`
image

Pseudocode:
image

infinite loop will triggered when `len == 1`

I guess it is caused by the compilation optimization bug of GCC 7.3.1

## Solution
Use a higher version of GCC, such as GCC 9.3.1.

Or

Tell the compiler do not to optimize `len`

## Sample
OS: CentOS 7
GCC Version: 7.3.1
Zeek Version: 4.2.0
Build Type: RelWithDebInfo

[GCC7.3.1_RelWithDebInfo_DNS.cc.o.zip](https://github.com/zeek/zeek/files/8354583/GCC7.3.1_RelWithDebInfo_DNS.cc.o.zip)

### DIFF
--- a/src/analyzer/protocol/dns/DNS.cc
+++ b/src/analyzer/protocol/dns/DNS.cc
@@ -693,6 +693,10 @@ bool DNS_Interpreter::ParseRR_EDNS(detail::DNS_MsgInfo* msg, const u_char*& data
 	// parse EDNS options
 	while ( len > 0 )
 		{
+#if (defined(__GNUC__) && __GNUC__ < 8)
+		// Older versions of GCC will add strange infinite loop instructions here
+		util::do_not_optimize(len);
+#endif
 		uint16_t option_code = ExtractShort(data, len);
 		int option_len = ExtractShort(data, len);
 		// check for invalid option length
--- a/src/util.h
+++ b/src/util.h
@@ -568,5 +568,25 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
 std::string json_escape_utf8(const char* val, size_t val_size,
                              bool escape_printable_controls = true);
 
+// The do_not_optimize(...) function can be used to prevent a value or
+// expression from being optimized away by the compiler. This function is
+// intended to add little to no overhead.
+// See: https://youtu.be/nXaxk27zwlk?t=2441
+template <typename T>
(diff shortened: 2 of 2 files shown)


0xxon:
First - thank you very much for digging into this. This looks interesting - and strange. 

That being said - I don't feel super great about adding a specific exception like this for an old compiler. This is especially true since we will probably up our compiler requirements to GCC 9 with the next release (to support spicy) - which is the same time at which this patch would be included.

AmazingPP:
> First - thank you very much for digging into this. This looks interesting - and strange.
> 
> That being said - I don't feel super great about adding a specific exception like this for an old compiler. This is especially true since we will probably up our compiler requirements to GCC 9 with the next release (to support spicy) - which is the same time at which this patch would be included.

It's my pleasure.

I agree that this code would not be necessary if we were sure to up the compiler requirements to GCC 9

rsmmr:
Actually it looks like the new requirement will be GCC 8.3. Is that sufficiently new to not exhibit this issue?

AmazingPP:
> Actually it looks like the new requirement will be GCC 8.3. Is that sufficiently new to not exhibit this issue?

that's sufficiently new.

By the way, should I close this PR?

timwoj:
Yep, I think we can close this PR since it looks like we have a way forward without making any changes.

Issue #2009, opened 2022-03-25
error: invalid conversion from 'const unsigned char*' to 'Bytef*' (when compile ZIP.cc in docker)


cxdinter:
Hi,
  I am new to Zeek.
  I downloaded Zeek master tree and compile it with NXP iMX8 cross compile in Ubuntu, it is ok and working fine on iMX8 board.
  But, when I create a docker image for the cross compile tool chain, then use docker image to build Zeek, always receive below error message. I don't know why. Any suggestion ? Thank you in advance.

use below command to login docker shell and compile Zeek 
docker run -it --rm -v /home/newdisk/Projects/zeek/zeek:/zeek -w /zeek --user=1000:1000 imx8-tiger /bin/bash 
 
Received error:
[ 71%] Building CXX object src/analyzer/protocol/zip/CMakeFiles/plugin-Zeek-ZIP.dir/ZIP.cc.o
/zeek/src/analyzer/protocol/zip/ZIP.cc: In member function 'virtual void zeek::analyzer::zip::ZIP_Analyzer::DeliverStream(int, const u_char*, bool)':
/zeek/src/analyzer/protocol/zip/ZIP.cc:62:29: error: invalid conversion from 'const unsigned char*' to 'Bytef*' {aka 'unsigned char*'} [-fpermissive]
  Bytef* orig_next_in = zip->next_in;
                        ~~~~~^~~~~~~

timwoj:
I generally don't recommend `const_cast` but this might be an ok place for it. See if this patch fixes it for you:

```
diff --git a/src/analyzer/protocol/zip/ZIP.cc b/src/analyzer/protocol/zip/ZIP.cc
index 5226fbd4f..970034e56 100644
--- a/src/analyzer/protocol/zip/ZIP.cc
+++ b/src/analyzer/protocol/zip/ZIP.cc
@@ -59,7 +59,7 @@ void ZIP_Analyzer::DeliverStream(int len, const u_char* data, bool orig)
        zip->next_in = (Bytef*)data;
        zip->avail_in = len;
 
-       Bytef* orig_next_in = zip->next_in;
+       Bytef* orig_next_in = const_cast<unsigned char*>(zip->next_in);
        size_t orig_avail_in = zip->avail_in;
 
        while ( true )
```

cxdinter:
> I generally don't recommend `const_cast` but this might be an ok place for it. See if this patch fixes it for you:
> 
> ```
[...]

PR #2014, opened 2022-03-29
GH-2009: Add const_cast to fix ZIP analyzer failure on some platforms

timwoj:
Fixes a simple bug with the zip analyzer on some rare platforms. This PR is mostly just for CI coverage.


0xxon:
To prevent needing the const_cast - could we not just change `Bytef* orig_next_in = zip->next_in;` to `auto orig_next_in = zip->next_in;`, so that it has the type that the platform expects?

That looks less scary and does not need a big comment.

0xxon (approved):

Issue #1500, opened 2021-04-09
BasicThread thread sanitizer issue


JustinAzoff:
I built zeek using ```./configure --sanitizers=thread```

https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual

Zeek seems to run ok listening on a live interface, but running on a single packet pcap produces the attached during termination:

[thread_sanitizer_out.txt](https://github.com/zeek/zeek/files/6286793/thread_sanitizer_out.txt)

The important bit seems to just be the 

```
WARNING: ThreadSanitizer: data race (pid=3895)
  Write of size 1 at 0x7b7c00015e19 by thread T12:
    #0 zeek::threading::BasicThread::launcher(void*) BasicThread.cc:165 (zeek:x86_64+0x10044cc01)
    #1 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(void*), zeek::threading::BasicThread*> >(void*) type_traits:4339 (zeek:x86_64+0x10044d00a)

  Previous read of size 1 at 0x7b7c00015e19 by main thread:
    #0 zeek::threading::BasicThread::SignalStop() BasicThread.cc:109 (zeek:x86_64+0x10044cc69)
    #1 zeek::threading::Manager::Terminate() Manager.cc:52 (zeek:x86_64+0x10044e8d1)
    #2 zeek::detail::terminate_bro() zeek-setup.cc:324 (zeek:x86_64+0x10021e8c6)
    #3 zeek::detail::cleanup(bool) zeek-setup.cc:930 (zeek:x86_64+0x10021e193)
    #4 main main.cc:81 (zeek:x86_64+0x100a3abad)
```

which is flagging the

```
void BasicThread::Done()
    {
    DBG_LOG(DBG_THREADING, "Thread %s has finished", name);

    terminating = true;
    killed = true;
    }
```

setting the bools to true.

You can do this to ignore all of BasicThread,

```
justin@mbp:/tmp$ echo 'race:BasicThread.cc' > suppression_file 
justin@mbp:/tmp$ TSAN_OPTIONS="suppressions=suppression_file"  zeek -r foo.pcap
```

but I don't see a way to make it more granular than that.

0xxon:
Hm. I wonder if there is a way to get threadsanitizer to just ignore these.

[...]

PR #2036, opened 2022-04-08
Mask our signal handlers' triggering signals around thread creation

ckreibich:
This addresses the Supervisor-related thread sanitizer complaint in #1500.

I've verified that this produces clean tsan runs on top of @timwoj's `topic/timw/bump-libkqueue` branch when building with clang, including when running parallelized btests.

Minor note: there's an indentation bug after the perftools disabler block starting [here](https://github.com/zeek/zeek/blob/master/src/zeek-setup.cc#L657-L660) — it never unindents through the end of the `setup()` function. Happy to fix that while we're at it. One way to do so is by always using the extra block (which seems safer anyway) and ifdef-protecting only the `HeapLeakChecker::Disabler disabler;` line in it.

### DIFF
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -398,6 +398,22 @@ static std::vector<std::string> get_script_signature_files()
 	return rval;
 	}
 
+// Helper for masking/unmasking the set of signals that apply to our signal
+// handlers: sig_handler() in this file, as well as stem_signal_handler() and
+// supervisor_signal_handler() in the Supervisor.
+static void set_signal_mask(bool do_block)
+	{
+	sigset_t mask_set;
+
+	sigemptyset(&mask_set);
+	sigaddset(&mask_set, SIGCHLD);
+	sigaddset(&mask_set, SIGTERM);
+	sigaddset(&mask_set, SIGINT);
+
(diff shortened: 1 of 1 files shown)


timwoj (commented):
The signal changes look good to me.

Regarding the indentation bug, it looks like clang-format isn't handling that bare brace set properly. It shouldn't be indenting it, and then it doesn't know that it should unindent the block following it. Try wrapping the opening/closing braces with `// clang-format {off,on}` and see if that fixes it.

ckreibich:
I tried a few things but am not having much luck. Looks like it indeed doesn't format the parts surrounded by the comment instruction, but it also doesn't take any cues from the disabled part. However, I noticed that when I remove the `#ifdef`s, it unindents as it should, so I think it's getting tripped up by the combo of the `#ifdef` and the block.

When I only remove the first `#ifdef` block (which admittedly is very mean), it gets very huffy and runs for around 15 seconds before indenting the lead-in comment. :-)

ckreibich:
I tweaked the function naming a bit to be more correct.

0xxon (approved):

timwoj:
Glad to see this landed. Now to figure out what's still broken in a TSan build.

Issue #2032, opened 2022-04-08
Use-after-free using tables with &create_expire when only inserting (possibly due to dictionaryresizing?)


awelzel:
Hey,

this is likely just the same or very close to what #2017 is about, but I'm opening a separate issue as #2017 focused on concurrent deletes, while it seems theres also an issue when inserting new elements and triggering resize of a table. This is a bit of a wild guess, but we observe segfaults with the same DoExpire()->ExpireAccessTime() call frame for tables for which there are no explicitly deletes happening for.

The following reproducer crashes with a custom Zeek 4.1.1 and `zeek version 5.0.0-dev.210-debug` build. There's no real reason for the randomness / parameters other than I fuzzed around until it triggered crashes. I suspect this can also be triggered when inserting elements outside of the expire_func if one gets the timing just right.

It crashes most of the time here, but if it doesn't after a few seconds best to restart.

```
redef table_expire_delay = 0.01sec;
redef table_incremental_step = 1;
redef table_expire_interval = 0.01sec;

redef exit_only_after_terminate = T;

function expire(t: table[count] of count, k: count): interval {
  if ( rand(100) < 80 ) {

    # Force insert elements
    local i = 0;
    while (++i < 100)
      t[i + 20000] = i + 20000;

    return 0.01sec;
  }

  print(fmt("Expiring %s", k));
  return 0.0sec;
}

global tbl: table[count] of count &create_expire=0.05sec &expire_func=expire;

const populates_num = 3;
const populates_per_second = 10;

event populate(round: count) {

  local i = round * populates_num;
  local iend = (round + 1) * populates_num + 1;
  while (++i < iend) {
    tbl[i] = i;
  }
  print(fmt("round %s size=%s", round, |tbl|));
  schedule double_to_interval(1.0/populates_per_second) { populate(++round) };
}

event zeek_init() {
  event populate(1);
}
```

It doesn't always crash, so may need to run a few times.
```
$ zeek ./table_expire__insert.zeek
round 1 size=3
...
Expiring 15
Expiring 17
round 27 size=159
=================================================================
==1641876==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000b92598 at pc 0x564c5064f59d bp 0x7ffdd7f5e0f0 sp 0x7ffdd7f5e0e8
READ of size 4 at 0x602000b92598 thread T0
[...]

Issue #2017, opened 2022-03-31
Use-after-free errors during table expiration when concurrent deletes happen


awelzel:
Hello,

running the following script with an ASAN enabled build triggers a heap-use-after-free violation. Tested on a custom Zeek 4.1.1 build as well as on the most recent master (zeek version 5.0.0-dev.210-debug on Debian bullseye with --sanitizers=address).

It looks as if the `expire_iterator` isn't invalidated whenever there are explicit deletes on a table - though not sure that's the right fix.
https://github.com/zeek/zeek/blob/master/src/Val.cc#L2437

## Backgroud

We have a Zeek package that schedules explicit events for doing `delete` on a table which also has a &read_expire set and stores connection uids. We observe higher likelyhood of segfaults happening on systems with this package enabled compared to other systems. Given the default table_* settings (step=5000, delay=0.01), the chance of triggering the issue is low and heavily depending on the environment.
 
The following reproducer sets those values much lower to stress-test the expiration and concurrent deletes and reproduces the crash fairly quickly, but not sure it's a good test for the suite due to minimum runtime required.

## Reproducer
```
redef table_expire_interval = 0.1sec;
redef table_incremental_step = 100;
redef table_expire_delay = 0.5sec;

redef exit_only_after_terminate = T;

global tbl: table[string] of vector of count &default = vector() &create_expire=1sec;

const populates_per_second = 100;
const populates_num = 100;

event cleanup(idx: string) {
  delete tbl[idx];
}

event populate(round: count) {

  local i = 0;
  while (++i < populates_num) {
    local val = rand(1000000);
    local val_str = cat(val);
    # print(fmt("round %s %s val=%s", round, i, val));
    tbl[val_str] = vector(val);

    # Schedule an explicit delete at most a second away.
    local random_cleanup_delay = double_to_interval(rand(100) / 100.0);
    schedule random_cleanup_delay { cleanup(val_str) };
  }

  print(fmt("round %s size=%s", round, |tbl|));
  schedule 1sec/populates_per_second { populate(++round) };
}

event zeek_init() {
  event populate(1);
}
```

## Output
```
$ zeek table_expire.zeek 
=================================================================
==3214020==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000b56098 at pc 0x55e03d33b59d bp 0x7ffffe6666d0 sp 0x7ffffe6666c8
READ of size 4 at 0x602000b56098 thread T0
[...]

PR #2040, opened 2022-04-11
Fix robust iterators when modifying dictionary during iteration.

rsmmr:
Addresses #2032.


rsmmr:
@awelzel I believe I understand now what's going on, and I'll actually stick with my prior fixes, they seem the best way forward after all. I have only reorganized the changes into two commits, added more comments, and also added one new `assert` to check an assumption I'm making. But in a release build, there's no functional change to my prior patch.

I'll make this ready to review once CI turns green.

timwoj (approved):

rsmmr:
I'm moving this back to draft. The two commits so fine are fine, but we have more trouble:  #2017 is triggering at least 2 further independent errors. For one of them, I got a fix already as well; the 2nd I'm still looking into.

 I'm also going to add some of the reproducers scripts as tests.

rsmmr:
Turning this back to ready for review. The last two commits are new, the first two commits remain unchanged.

timwoj (approved):

PR #2047, opened 2022-04-13
Management framework: add support for running actions on cluster nodes, provide `get_id_value` API

ckreibich:
At a high level, this implements `zeek-client get-id-value <ID> [nodes]`, a way to retrieve values out of the running cluster nodes in JSON. Example:

```
$ zeek-client get-id-value Log::active_streams logger-01 | jq
{
  "logger-01": {
    "Broker::LOG": {
      "columns": null,
      "path": "broker",
      "policy": "Broker::log_policy"
    },
    "Cluster::LOG": {
      "columns": null,
      "path": "cluster",
      "policy": "Cluster::log_policy"
    },
    ...
}
```
Under the hood this implements a generic mechanism for the controller to allow it to run something on any subset of cluster nodes, aggregate the results, and respond back to the client.

I'll bump the external testsuite tomorrow to include new tests — I ran out of time today.


rsmmr (approved):
The generic machinery for handling requests/reply look good. Having that, adding the `get_id_value` command seem pretty straight forward then.

ckreibich:
@rsmmr, fyi, the changes I pushed on top of your approval are twofold:
- Results coming into agents from the cluster nodes used the `$instance` field to identify themselves, instead of `$node`. Both fields are now filled correctly.
- For commands where zeek-client receives multiple results, it didn't report errors in these results in deterministic order.

PR #2049, opened 2022-04-13
Update libkqueue for Coverity fixes

timwoj:
Coverity had some findings in libkqueue when we updated our fork of it recently (https://github.com/zeek/zeek/pull/2037). Those findings were reported upstream and fixed. This PR bumps us to a newer version that includes those fixes. It's not a tagged release this time, but it's about 10 commits ahead of where we were before (right at the 2.6.0 tag) and all of those commits seem useful to have.

Note that they're currently getting their own Coverity setup going so there may be more fixes coming down the pipe, but we can likely hold off on bringing those in for a bit. This is just clearing out the ones that were found by our builds.


rsmmr (approved):
(Approving without any opinion on the libkqueue-side changes).

PR #2051, opened 2022-04-14
Update gen-zam submodule

0xxon:
Let CI run for the following PR: https://github.com/zeek/gen-zam/pull/1


0xxon:
This might also be a candidate for a point-release - due to the fact that it can break compiles on a platform that will become supported.

Note - gcc-12 is available on Ubuntu 22.04, but not the default compiler, so it does not break compilation using the default compiler.

ckreibich:
I've bumped the submodule directly (acaa9ec01e00447b0deca37385f08797895b00b1) so I'm going to go ahead and close this PR. I'll put it on the backport list. Seems we'll likely just want to patch the in-tree Gen-ZAM.cc there, since it's not split out into a submodule in the release branches.

PR #2058, opened 2022-04-19
Management framework: consistency fixes around event() vs Broker::publish

ckreibich:
Minor changeset, see commit.

I'm tempted to switch the generation of events that are exclusively sent to another component to `Broker::publish()` altogether (i.e., drop auto-publishing) since it's more explicit, but it is also a fair bit more verbose ... so holding off for now.

### DIFF
--- a/scripts/policy/frameworks/management/agent/main.zeek
+++ b/scripts/policy/frameworks/management/agent/main.zeek
@@ -57,6 +57,12 @@ global g_nodes: table[string] of Management::Node;
 global g_cluster: table[string] of Supervisor::ClusterEndpoint;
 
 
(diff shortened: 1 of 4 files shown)


rsmmr (approved):
I support switching to `Broker::publish()`, agree that autopublishing hurts readability in cases where events are used for explicit messaging across Zeek instances (vs. events where it's primarily publish/subscribe bus pattern where receivers may be local or remote).

ckreibich:
Ah, great. Let me move them over then. I'll update the PR.

ckreibich:
Okay Robin, it's all `Broker::publish()` where applicable now.

PR #2064, opened 2022-04-21
Add more `weird.log` entries for malformed TCP options

fox-ds:
When looking at some anomalous traffic and diving into the `TCPSessionAdapter.cc` file, it appears that Zeek just assumes that the TCP options are always according to the RFC. However, in some cases, TCP options can even be misused for covert C2 channels. In this example, we were looking at a `ls -la` command that was embedded into a TCP option.

After replaying the pcap in Zeek and looking at the `tcp_option` and `tcp_options` events, there wasn't any sign of something extraordinary about the TCP options. The comments in the `TCPSessionAdapter.cc` file hint at possible future logging entries for these anomalous options, which are implemented in this commit.

In summary, this commit adds a few new entries in the `weird.log`:

- tcp_nonstandard_option
- tcp_option_no_room_for_length
- tcp_option_trashed_length
- tcp_option_length_exceeds_options_space
- tcp_remaining_options_after_TCPOPT_EOL

Note that this change will only add these new entries in the `weird.log` when either the `tcp_option` or `tcp_options` event is overridden.


0xxon (commented):
This generally looks good to me...

However, this also changes some low-level behavior in our TCP stack. As such, I think this is something that optimally should see quite a bit of testing. I know that in principle this should just work - but I am always worried that there is something that will break because of this.

One slightly concerning change to me in this is the fact that we continue parsing after `TCPOPT_EOL`. I think it is a good idea to do that - but I am not sure what that might cause in practice (even if it is just excessive warnings because something adds random padding afterwards).

The second thing that I don't quite get is the option length patching. Why is this desirable, and are we sure that it is ok to cast away the const-ness of the tcphdr (I only tracked it up one function call where it also seems to be const).

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -1670,29 +1670,66 @@ int TCPSessionAdapter::ParseTCPOptions(const struct tcphdr* tcp, bool is_orig)
 
 		unsigned int opt_len;
 
+		// Non-standard TCP option kinds should not be common
+		// https://www.iana[.]org/assignments/tcp-parameters/tcp-parameters.xhtml
```
grigorescu, line 1674: Minor nit, but I don't understand why this URL is sanitized. Just makes it harder to actually reference.

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -1670,29 +1670,66 @@ int TCPSessionAdapter::ParseTCPOptions(const struct tcphdr* tcp, bool is_orig)
 
 		unsigned int opt_len;
 
+		// Non-standard TCP option kinds should not be common
```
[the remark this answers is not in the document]
fox-ds: Good suggestion, I'll change this!

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
fox-ds: It's a force of habit, I guess. I'll remove it here.

fox-ds:
@0xxon Thanks for your feedback! I'll have a look at the option length patching again, and check if this is really needed or can be left out of this PR. I primarily did this to get the command data in the `tcp_option` event, but this solution may not be desirable. 

Regarding the additional data after the `TCPOPT_EOL`; maybe we could only send out a Weird log if the next byte is a valid TCP option kind. I'll have a look at the feedback in the coming weeks, thanks again!

0xxon:
That sounds great. Let us know if you need anything in the meantime :)

rsmmr:
@fox-ds fyi, we'll start wrapping things up for 5.0 soon if you'd like this to still go in.

rsmmr:
I'm hearing that more changes are coming here but won't make it for 5.0. I'll close this for now, please reopen when updated.

PR #2068, opened 2022-04-24
Fix a handful of DNS_Mgr issues

timwoj:
This PR resolves all of the following that cropped up from automated testing post-merge of the c-ares work:

- Fixes a rare crash involving async DNS lookups. It was detected by oss-fuzz initially. The `asyncs_timeouts` priority queue was used by the old `DNS_Mgr` code to handle timeouts coming from DNS lookups. Now that c-ares handles the timeouts internally for us, there's no reason to continue to track them ourselves.
- Modifies `DNS_Mgr::GetNextTimeout()` to request the next timeout value from c-ares instead of using a fixed value. This shouldn't make a huge difference in the long run about when timeouts happen, but it's possible for those two values to be different and the one from c-ares should be more accurate.
- Pre-initialize the c-ares channel object in DNS_Mgr, which fixes Coverity finding 1488318.
- Wrap the call to doctest's `MESSAGE()` method in `Reporter` in a try/catch block for `doctest::detail::TestFailureException`. This fixes a huge list of findings from Coverity about an uncaught exception. The catch block will simply output the message to stdout like normal, just so the output is captured somewhere.

### DIFF
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -1403,7 +1403,21 @@ double DNS_Mgr::GetNextTimeout()
 	if ( asyncs_pending == 0 )
 		return -1;
 
-	return run_state::network_time + DNS_TIMEOUT;
+	fd_set read_fds, write_fds;
+
+	FD_ZERO(&read_fds);
+	FD_ZERO(&write_fds);
+	int nfds = ares_fds(channel, &read_fds, &write_fds);
+	if ( nfds == 0 )
+		return -1;
+
+	struct timeval tv;
+	tv.tv_sec = DNS_TIMEOUT;
+	tv.tv_usec = 0;
+
+	struct timeval* tvp = ares_timeout(channel, &tv, &tv);
+
+	return run_state::network_time + static_cast<double>(tvp->tv_sec) + (static_cast<double>(tvp->tv_usec) / 1e6);
 	}
 
 void DNS_Mgr::ProcessFd(int fd, int flags)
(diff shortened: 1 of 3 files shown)


src/DNS_Mgr.cc
```diff
@@ -1403,7 +1403,22 @@ double DNS_Mgr::GetNextTimeout()
 	if ( asyncs_pending == 0 )
 		return -1;
 
-	return run_state::network_time + DNS_TIMEOUT;
+	int nfds = 0;
+	fd_set read_fds, write_fds;
+
+	struct timeval tv;
+	tv.tv_sec = DNS_TIMEOUT;
+	tv.tv_usec = 0;
```
rsmmr, line 1411: Suggested readability tweak: move this block to just before `struct timeval* tvp = ares_timeout(channel, &tv, &tv);`.

src/DNS_Mgr.cc
```diff
@@ -1403,7 +1403,22 @@ double DNS_Mgr::GetNextTimeout()
...
+
+	FD_ZERO(&read_fds);
+	FD_ZERO(&write_fds);
+	nfds = ares_fds(channel, &read_fds, &write_fds);
+	if ( nfds == 0 )
+		return -1;
+
+	struct timeval* tvp = ares_timeout(channel, &tv, &tv);
+
+	return run_state::network_time + tvp->tv_sec + tvp->tv_usec;
```
rsmmr, line 1421: Doesn't `tv_usec` need to be scaled to seconds before adding?

rsmmr (approved):
Couple small things, otherwise ready to merge.

src/DNS_Mgr.cc
timwoj: Good point. That'll avoid allocating it if we don't need it too.

PR #2071, opened 2022-04-25
Rework FindCAres.cmake to not use ExternalProject, fixing OBS builds

timwoj:
FindCAres.cmake uses `ExternalProject_Add()` to manually call `cmake` on the c-ares submodule and then later to build it. This was done mainly so that we could capture the output from cmake for it, but it causes a bunch of problems with library paths on a number of platforms. I previously fixed the issues with GitHub workflows, but that fix apparently broke OBS builds.

This PR changes how that file works to just use `add_subdirectory` like broker, etc, which simplifies things greatly.

A branch with the same name in the cmake repo should be merged at the same time.


0xxon:
Nice branch name :)

An OBS build with this branch is running - you can check the result at https://build.opensuse.org/package/show/home:0xxon:branches:security:zeek-c-cares/zeek-nightly

PR #2078, opened 2022-04-29
Type descriptions using names if available

vpax:
Currently, `Type::Describe` (and children) generate an explicit description of the type.  For cases where the type has a name, it's a lot handier if the description is instead the name.  This (1) compresses some huge messages down to things like "`connection`", (2) sidesteps the problem that type descriptions are almost-but-not-quite parseable (a headache for Event Tracing), (3) in general makes error messages more clear.

My motivation for this PR is (2) above, because otherwise I'm going to have the replicate the functionality in Event Tracing anyway.

If you're wondering about the `plugin.hooks` baseline change, it's because lambdas compute their hashes based on a description of themselves, which includes their type, and that has now changed.


0xxon (commented):
I do like this - the new output is much more helpful in my opinion.

Small question - could we use this as an opportunity to document the Describe/DoDescribe functions in Type.h?

The only other question that I had: might it make sense to still include what type the name has? As in - I noticed that this changed the test output from "enum color" to "color", which is less useful. And I could also see that "record Outer" might be more informative than just "Outer".

vpax:
> Small question - could we use this as an opportunity to document the Describe/DoDescribe functions in Type.h?

`Describe` is a general `Obj` method, so I don't think it makes sense to document it locally.  `DoDescribe` is internal, not meant for external calling.  I could make it `protected` to emphasize that?

> The only other question that I had: might it make sense to still include what type the name has?

The trick here is that for event tracing, the description needs to be parseable, so we'd need to extend the interface to `Desc` to support expressing that notion. There's currently a form called `PORTABLE` that isn't documented and isn't used anywhere in the base code.  We could consider changing it to `PARSEABLE` (instead of adding `PARSEABLE` as a new mode, which would lead expand the already somewhat busy control logic used by `Describe`).

That said, I'd like if possible to decouple this from the current change (which fixes a significant bug in event tracing).  So maybe for now the action item is mark `PORTABLE` as deprecated?

0xxon:
> > Small question - could we use this as an opportunity to document the Describe/DoDescribe functions in Type.h?
> 
> `Describe` is a general `Obj` method, so I don't think it makes sense to document it locally. `DoDescribe` is internal, not meant for external calling. I could make it `protected` to emphasize that?

Fair. If it is internal only, making it `protected` seems like it makes sense.

> > The only other question that I had: might it make sense to still include what type the name has?
> 
> The trick here is that for event tracing, the description needs to be parseable, so we'd need to extend the interface to `Desc` to support expressing that notion. There's currently a form called `PORTABLE` that isn't documented and isn't used anywhere in the base code. We could consider changing it to `PARSEABLE` (instead of adding `PARSEABLE` as a new mode, which would lead expand the already somewhat busy control logic used by `Describe`).
> 
> That said, I'd like if possible to decouple this from the current change (which fixes a significant bug in event tracing). So maybe for now the action item is mark `PORTABLE` as deprecated?

Also fair.

vpax:
I've pushed changes to make `DoDescribe` non-public, and to flag `DESC_PORTABLE` as deprecated.

vpax:
> Suggested formatting:

Thanks, Tim, now pushed.

> It'd also be good to recommend something for the user to use instead of DESC_PORTABLE, if such a thing exists

The trick here is that it's not even clear what DESC_PORTABLE is supposed to do!  I have a vague memory that I introduced it eons ago with the intent that it would be an ASCII format that (1) could be readily moved across architectures, and (2) could be edited (with appropriate care).  But I'm not sure that that's right.  I could suggest using DESC_READABLE instead, but there's no reader for it.  (There _almost_ is, namely Zeek's own parser, but it needs a tweak or two to get there.)

0xxon (approved):

PR #2079, opened 2022-04-30
fix for sets containing recursive records

vpax:
In working on some event tracing issues, I discovered that the following one-liner:

> global crash_me: set[Conn::RemovalHook];

will crash Zeek 4.x.  Amusingly, the problem is the same as the one that @rsmmr flagged in https://github.com/zeek/zeek/pull/2077, namely the presence of recursive records.  This simple PR fixes the issue and updates the test suite to check for it.

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -1367,7 +1367,10 @@ static void table_entry_val_delete_func(void* val)
 	delete tv;
 	}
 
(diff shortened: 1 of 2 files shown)


vpax:
FYI, per the discussion for https://github.com/zeek/zeek/pull/2077, I've updated this PR to use a stack-based set rather than a static one to prevent infinite recursion.

timwoj (approved):

PR #2080, opened 2022-05-01
ZAM maintenance for recent changes and some newly exercised corner cases

vpax:
At the beginning of each month I try to run ZAM on the test suite to see if any new problems arise. This PR fixes several problems (and defers one).


timwoj:
The Alpine CI build has some weird errors in a couple of the tests. See https://cirrus-ci.com/task/4690327469555712. The build eventually succeeds, but it's done the exact same thing two builds in a row so it doesn't look truly transient.

PR #2086, opened 2022-05-03
Fixes Coverity issues reported from DNS_Mgr merge

timwoj:
Reports fixed: 1488295, 1488292, 1488281, 1488279

### DIFF
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -246,7 +246,9 @@ DNS_Request::~DNS_Request() { }
 void DNS_Request::MakeRequest(ares_channel channel, DNS_Mgr* mgr)
 	{
 	// This needs to get deleted at the end of the callback method.
-	auto req_data = new CallbackArgs{this, mgr};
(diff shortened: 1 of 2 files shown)


src/DNS_Mgr.cc
```diff
@@ -246,7 +246,9 @@ DNS_Request::~DNS_Request() { }
 void DNS_Request::MakeRequest(ares_channel channel, DNS_Mgr* mgr)
 	{
 	// This needs to get deleted at the end of the callback method.
-	auto req_data = new CallbackArgs{this, mgr};
+	auto req_data = std::make_unique<CallbackArgs>();
+	req_data->req = this;
+	req_data->mgr = mgr;
```
rsmmr, line 251: Just curious: why can't we pass these through `make_unique` arguments?

rsmmr (approved):

src/DNS_Mgr.cc
rsmmr: Ack, thanks, that's fine, I was just curious.

PR #2087, opened 2022-05-03
Fine-grained profiling of scripts

vpax:
This PR is mainly about producing per-function-body profiles of script execution, in terms of CPU & memory impact.  The new `--profile-scripts[=file]` flag instructs Zeek to write (upon termination) a profile of every function body to the given file (default: _stdout_), along with some aggregate profiles.  A function body is associated with a `function` or a BiF (which can only have one), or a `hook` or `event` handler (can have multiple).

The profile is in TSV, quasi-Zeek-log format.  Here's an example:

> #fields	function	location	type	ncall	tot_CPU	child_CPU	tot_Mem	child_Mem
#types	string	string	string	count	interval	interval	count	count
getenv	.../base/init-bare.zeek:5489	BiF	2	0.000020	0.000000	0	0
Version::parse	.../base/misc/version.zeek:57	function	1	0.000657	0.000253	53248	28672
...
dns_end	.../base/protocols/dns/./main.zeek:398	event	19868	0.104888	0.000000	0	0
dns_end	.../base/protocols/dns/./main.zeek:409	event	19868	0.467579	0.279640	94208	94208
...
dns_end	2-locations	event	39736	0.572467	0.000000	94208	0
...
all-BiFs	46-locations	BiF	147057	1.226884	0.000000	4599808	0
total	244-locations	TOTAL	488972	13.146864	0.000000	15212544	0
non-scripts	<no-location>	TOTAL	215335	4.815573	0.000000	94638080	0

(where to keep things readable I've elided the common location prefix to `...`)

Each line gives the name of the function, the location of the body, its type (function/BiF/hook/event), the total number of calls, the total CPU time accumulated across those calls, the subset of that time that's due to children (other functions) called by the body, and then the same sort of states for memory usage in bytes.  Memory usage is approximate and mostly meaningful for relative comparisons, or for spotting state management issues.

For any function/hook/event with multiple bodies, the profile includes a summary across all of those (see `dns_end` above).  The profile also includes statistics aggregated over all BiFs, over all scripts, and due to _non_-script execution (i.e., the rest of Zeek's processing).  For the example above, we can see that scripts accounted for more than 2/3s of all CPU time (but only ~12% of memory); and that for scripts, most of that was not due to calls to BiFs.

Previous discussion flagged that there's a profiling file format from Google that perhaps we should use.  I looked into it and it wasn't an apt fit: it included notions that don't have relevance for scripts, and didn't appear to be extensible to include some of the information we'd like.

Profiling is expensive (roughly doubled execution time on a longer run that I tried), so not something that you'd likely want to run in live deployment.  Because of that, I haven't added a BiF to dump out the profile and then reset it to collect for another interval, though that might be a natural (easy) extension at some point.

In the process of developing the capability, I first looked at some of the existing logging and found some bugs in it, so this PR includes fixes for those.  Profiling startup then quickly pointed up an expensive BiF that startup always calls, which from its usage could be replaced by a new, much lighter-weight BiF (`global_options()`), which this PR also includes.  Finally, running the BTests found an unrelated bug in the printing of `when` statements, so there's a fix for that, too.


src/Func.cc
```diff
@@ -795,6 +790,9 @@ ValPtr BuiltinFunc::Invoke(Args* args, Frame* parent) const
 		g_trace_state.LogTrace("\tFunction return: %s\n", d.Description());
 		}
 
+	if ( spm )
```
awelzel, line 793: Just putting this out as food-for-thought if there's an opportunity to expose StartInvocation/EndInvocation to plugins.

There is at least one plugin that tracks function runtime via HookCallFunction and exposes it through a prometheus interface:
https://github.com/esnet/zeek-exporter/blob/master/src/Plugin.cc#L106

A recent MR around OpenTelemetry also hooks into Invoke():
https://github.com/zeek/zeek/pull/1998/files#diff-bc6dffe506a4add540a9b1df23ca34794aed8eec4ace91eeae9a9c7c12193e6aR396

Generally, seems if external plugins could hook into start and end invocation of functions, it could foster some interesting experiments around performance collection outside of the code base.

I was initially wondering if two new Hooks would be reasonable, but maybe the `spm` could represent a collection of `ScriptProfileMgr` instances (interface being just the `StartInvocation` `EndInvocation` methods) that Plugins can implement and register/deregister at will (possibly even during runtime).

Just thinking out loud. Maybe better suited for the discussions section.

src/ScriptProfile.cc
```diff
@@ -0,0 +1,185 @@
...
+
+	} // namespace zeek::detail
+
+void activate_script_profiling(const char* fn)
+	{
+	FILE* f;
+
+	if ( fn )
+		{
+		f = fopen(fn, "w");
```
awelzel, line 175: I only skimmed through. Is this missing a corresponding `fclose()`?

src/util.cc
```diff
@@ -2058,6 +2058,13 @@ int time_compare(struct timeval* tv_a, struct timeval* tv_b)
 		return tv_a->tv_sec - tv_b->tv_sec;
 	}
 
+double curr_CPU_time()
+	{
+	struct timespec ts;
+	clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &ts);
```
awelzel, line 2064: For the purpose of script profiling I wonder this should be using `CLOCK_THREAD_CPUTIME_ID` to get the cpu time solely for the main thread rather than the the whole process.

Unclear how much it matters, but there could be busy caf and logger threads running in parallel next to script execution that may skew the resulting times around.

src/Func.cc
vpax: Yeah I can see that being handy ... would prefer though to get this simpler approach integrated for now since I'd like to have the functionality in 5.0.

src/ScriptProfile.cc
vpax: Good catch, thanks - addressed in a new push just now.

src/util.cc
vpax: I can see that going either way.  Over time, those busy additional threads might correlate with the script's own activity, in which case it's a plus.  OTOH, if they don't, and are bursty, then yeah it's a minus.

How portable is `CLOCK_THREAD_CPUTIME_ID` ?  Assuming it's readily available, another approach would be to provide both CPU times ... though that starts making the profile a bit clunky.

awelzel: > Over time, those busy additional threads might correlate with the script's own activity, in which case it's a plus. OTOH, if they don't, and are bursty, then yeah it's a minus.

Scenario I had in mind would be some function doing a `Log::write()`. The actual JSON rendering/gzip/.. happening in a parallel running  logger thread while the main thread moved on executing a different/unrelated handler/function.

But given PROCESS_CPUTIME_ID provides useful results, maybe just something to keep in mind at this point as it might not matter in practice.

> How portable is CLOCK_THREAD_CPUTIME_ID ? Assuming it's readily available, another approach would be to provide both CPU times ... though that starts making the profile a bit clunky.

Mostly not sure about OSX, probably okay on Linux/BSD I guess. Sticking with PROCESS_CPUTIME_ID is probably better than clunky.

timwoj (approved):
Looks good to me. You might write a secondary github issue for exposing the invocation stuff to plugins as suggested by @awelzel.

PR #2090, opened 2022-05-05
Robustly finding unused ("orphan") event handlers / hooks / functions

vpax:
Per https://github.com/zeek/zeek/discussions/1970, this PR adds functionality to statically analyze scripts at start-up to find any event handlers, hooks, or functions (which I'll just call "functions" in the remainder) that can't be invoked.  These errors are suppressed for functions annotated with `&is_used`.  They're also ignored for `&deprecated` functions, and the functionality as a whole can be turned off by specifying `--no-usage-warnings` on the command line.

The addition required extending the AST traversal framework to work for attributes.  This leads to some minor BTest baseline changes.  The change also includes new `&is_used` attributes for functions in the base scripts that are in fact used (sometimes only due to conditional code, or only if a corresponding `policy/` script is loaded).  Finally, some of the unused base script functions appear to be vestigial, so I've marked those with `&deprecated`.


vpax:
> I have a generic question on this - which I maybe should have raised earlier, but I did not think about it - does this work with Spicy evt files?

I _think_ it should work fine.  It considers an event as made-available-by-the-event-engine if `event_registry` knows about it.  I'd assume that Spicy registers its events in a fashion that does this ... but if it doesn't, then yeah this would require a refinement.

If the Spicy integration is ready to go, then I can test against it.  If not, then I think it makes sense to merge this and then we'll catch the issue during final integration testing for Spicy.  If there's an issue, it's likely easy to work around (since Spicy has to _somehow_ register its event handlers), or, at worse, for now make this feature opt-in rather than opt-out.

timwoj:
> If the Spicy integration is ready to go, then I can test against it.

The Spicy integration was merged into master yesterday. Have at it!

src/script_opt/UsageAnalyzer.cc
```diff
@@ -0,0 +1,224 @@
...
+		if ( t->AsFuncType()->Flavor() == FUNC_FLAVOR_FUNCTION )
+			continue;
+
+		if ( reachables.count(id) > 0 )
+			continue;
+
+		auto flavor = t->AsFuncType()->FlavorString();
+		auto loc = id->GetLocationInfo();
+
+		reporter->Warning("%s %s (%s:%d): cannot be invoked", flavor.c_str(), id->Name(),
```
bbannier, line 55: When I first encountered this, the emitted warning wasn't clear to me. Maybe this could be stated in a way which formulates what a user be do about this and maybe also more clearly call out why we think this is an issue. It would also be nice if the location format of this diagnostic would follow what Zeek already does for source parse errors so that downstream tooling (e.g., editors) can use a single error format to parse out location and error.

Maybe something like:

> warning in /tmp/bla.zeek, line 2: found event handle for non-existing event "does_not_exist" which will never be invoked

src/script_opt/UsageAnalyzer.cc
```diff
@@ -0,0 +1,224 @@
...
+		if ( reachables.count(id.get()) > 0 )
+			continue;
+
+		auto f = GetFuncIfAny(id);
+		if ( ! f )
+			continue;
+
+		auto loc = id->GetLocationInfo();
+
+		reporter->Warning("function %s (%s:%d): cannot be called", id->Name(), loc->filename,
```
bbannier, line 80: See comment on diagnostic style above.

bbannier (commented):
I cannot comment on the implementation side, but left some comments from a user's perspective.

src/Options.cc
```diff
@@ -106,6 +106,8 @@ void usage(const char* prog, int code)
 	fprintf(stderr, "    -s|--rulefile <rulefile>        | read rules from given file\n");
 	fprintf(stderr, "    -t|--tracefile <tracefile>      | activate execution tracing\n");
 	fprintf(stderr, "    -u|--usage-issues               | find variable usage issues and exit\n");
```
awelzel, line 108: Mostly wondering if the new option (and possibly the old `--usage-issues`) could have another word added so they don't look so related and easy to confuse.
```
--usage-issues
--no-usage-warnings
```
I do not have great suggestions. Those come to mind: `--variable-usage-issues` or `--variable-usage-warnings` and `--no-func-usage-warnings`, or `--no-invoke-usage-warnings`, or `--Wno-invoke-warnings`.

src/Options.cc
```diff
@@ -106,6 +106,8 @@ void usage(const char* prog, int code)
...
+	fprintf(stderr, "       --no-usage-warnings          | suppress warnings of unused "
```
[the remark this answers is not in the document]
0xxon: Yeah, FWIW, I really like that idea.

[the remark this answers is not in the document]
vpax: Meaning I should add that with this PR ... or that folks like it as something to add to the to-do list for the future?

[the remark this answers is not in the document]
timwoj: Given that the option is new in this PR, I'd say yes. On top of that, since 5.0 is about to come out including all of these changes, it'd be good to settle on an argument format now before people are using it instead of changing it between 5.0 and 5.1.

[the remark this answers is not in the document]
vpax: I went to do this and then noticed that `-W` is already used (for the "watchdog" - gotta wonder if anyone still actually needs that).  Suggestions?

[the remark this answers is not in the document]
vpax: Given that we'd have to deprecate `-W` (which I'm all in favor of!), I propose decoupling these issues, so my latest push does not address that

src/script_opt/UsageAnalyzer.cc
vpax: Thanks for the thought.  I've gone ahead and regularized these

src/script_opt/UsageAnalyzer.cc
vpax: (same)

src/Options.cc
vpax: I realized that a better name is `--no-unused-warnings`, which also helps with the name conflict.  Latest push includes that

scripts/base/frameworks/notice/main.zeek
```diff
@@ -585,7 +585,7 @@ function is_being_suppressed(n: Notice::Info): bool
 
 # Executes a script with all of the notice fields put into the
 # new process' environment as "ZEEK_ARG_<field>" variables.
-function execute_with_notice(cmd: string, n: Notice::Info)
+function execute_with_notice(cmd: string, n: Notice::Info) &deprecated
```
timwoj, line 588: Should there be warning text for all of these deprecated script functions? Something like "Remove in v6.1. Usage testing indicates this function is unused" would be good enough.

timwoj (commented):
One question about the script deprecations. Everything else looks good to me.

scripts/base/frameworks/notice/main.zeek
vpax: Yeah, that's a good thought.  Want to go ahead and add that, or should I?

vpax:
> The test failure in the last build looks like this: #2134

It also shows that there should be some `&is_used`'s added to `scripts/base/frameworks/intel//cluster.zeek`.  If you do the deprecation text per the above, I'll send you a patch for those; if I do it, I'll add this directly.

scripts/base/frameworks/notice/main.zeek
timwoj: Have at it. I'm tracking down some other stuff today.

vpax: Just pushed

timwoj (approved):

timwoj:
Anyone else have comments on this? Otherwise I'll merge it on Monday.

PR #2094, opened 2022-05-11
Add exact name of the Top-k algorithm.

0xxon:
I needed to figure out which exact algorithm we use for our
probabilistic top-k measurements. It turns out that we do not mention
this in our source tree at all so far.


scripts/base/frameworks/sumstats/plugins/topk.zeek
```diff
@@ -1,4 +1,9 @@
 ##! Keep the top-k (i.e., most frequently occurring) observations.
+##!
+##! This plugin uses a probabilistic algorithm to count the top-k elements.
+##! The algorithm (calles Space-Saving) is described in the paper Efficient
```
timwoj, line 4:
```suggestion
##! The algorithm (called Space-Saving) is described in the paper "Efficient
```

PR #2100, opened 2022-05-13
Add more information to warning messages from input framework

### LINKED ISSUE #1857: Some input related errors do not include the file/line information
We currently have a test that shows this issue: btest/scripts/base/frameworks/input/invalidset.zeek

The baseline is this (repeated twice, just showing the first 4)

```
EventErrorEvent, Could not convert line 'name\x09127.0.0.1' of ../input.log to Val. Ignoring line., Reporter::WARNING
EventErrorEvent, Error while reading set or vector, Reporter::WARNING
EventErrorEvent, Invalid value for subnet: 127.0.0.1, Reporter::WARNING
EventErrorEvent, Skipping input with missing non-optional value, Reporter::WARNING
```

The first error is good.. mentions the invalid line and the file it comes from.  However, the other errors don't do this.  "Skipping input with missing non-optional value" and "Error while reading set or vector" can be very difficult to track down the root cause.

it is easier to see this if the test is modified to only have the first line in the input file:

```
@TEST-START-FILE input.log
#separator \x09
#fields i       s
name    -
@TEST-END-FILE
```

in that case, the btest outputs only this:

```
warning: Skipping input with missing non-optional value
```

timwoj:
Fixes #1857


timwoj:
@JustinAzoff Does this cover what you were wanting? I added line information to some of the other lines in the `stdout` output (notably the "Could not convert" messages), but that information ends up getting duplicated in the `stderr` output too.

JustinAzoff:
I believe so!  Those new error messages look really good.

src/input/readers/ascii/Ascii.h
```diff
@@ -62,6 +59,8 @@ class Ascii : public ReaderBackend
 	bool DoUpdate() override;
 	bool DoHeartbeat(double network_time, double current_time) override;
 
+	zeek::detail::Location* GetLocationInfo() const override { return read_location.get(); }
```
rsmmr, line 62: Return `const` pointer to make clear caller shouldn't modify this?

src/threading/MsgThread.cc
```diff
@@ -341,46 +343,122 @@ void MsgThread::Finished()
...
-	SendOut(new detail::ReporterMessage(detail::ReporterMessage::WARNING, this, msg));
+	ODesc desc;
+
+	if ( auto* location = GetLocationInfo() )
+		{
+		location->Describe(&desc);
+		desc.Add(": ");
+		}
+
+	desc.Add(msg);
```
rsmmr, line 368: This code is repeated a  number  of times across these methods, suggest to factor it out.

rsmmr (approved):

PR #2108, opened 2022-05-14
speed up ZAM compilation by capping function size when inlining

vpax:
It turns out that ZAM compilation time is dominated by some really huge functions created by recursively inlining.  This PR limits the size to which a function can grow during inlining (set to <= 1000 AST nodes).  On some problematic sets of scripts this cuts compilation time in half, which makes it possible to remove some ZAM-related limitations in the test suite.  (The PR also includes a couple of minor `-a zam` Baseline alternative updates due to recent separate changes.)


src/script_opt/Inline.cc
```diff
@@ -139,8 +140,15 @@ void Inliner::InlineFunction(FuncInfo* f)
 	// particular body.
 	curr_frame_size = f->Scope()->Length();
 
+	auto oi = f->Body()->GetOptInfo();
```
timwoj, line 143: Out of curiosity to help me understand the changes, why does this save off the values before calling `Inline()`? Is it because they might change further inside the call to inline that statement, and we want to reset them back to what they were before we continue with later statements?

timwoj: Perfect, that was the explanation I was looking for 👍

timwoj:
Merged with one minor change in the merge commit to make the constant a file-level value and to make it `constexpr`.

Issue #2115, opened 2022-05-18
Btest failures with OpenSSL 3.0.1 on CentOS Stream 9


timwoj:
There's apparently some bug in the OpenSSL 3.0.1 package on CentOS Stream 9 that causes some of our btests to fail with a "certificate signature failure" warning. It's only on that one platform because 3.0.0+ work just fine on other platforms (@0xxon notes she tested up to 3.0.3 on another machine).

See https://cirrus-ci.com/task/5837691458682880 as an example of the failures.

0xxon:
I just set up a CentOS 9 stream VM - and can confirm that this happens there every time.

I may not have the time to dig deeply into this this week - but I can try at the beginning of next week.

PR #2116, opened 2022-05-18
Enable ccache for CI builds

timwoj:
This also includes two other fixes:
- Enabling 'greedy' CPU mode for builds that it wasn't enabled for before
- A temporary fix for builds on CentOS Stream 9 until we can sort out the openssl 3.0.1 issue on that platform (see #2115)


rsmmr (approved):
Not too familiar with all this but I didn't spot anything surprising at least. :-)

PR #2122, opened 2022-05-23
Add DOCKERFILE_VERSION variable to all CI dockerfiles

timwoj:
We ran into a bug with the CentOS 9 CI build recently that effectively stemmed from the fact that we hadn't rebuilt that docker instance in a while, and a package update caused something to break in the Zeek tests. This PR adds a new variable to every one of our dockerfiles that allows us to force rebuilds of them. Preferably we'd do this around each release cycle so we can catch these kinds of changes more easily.

### DIFF
--- a/ci/alpine/Dockerfile
+++ b/ci/alpine/Dockerfile
@@ -1,5 +1,9 @@
 FROM alpine:latest
 
+# A version field to invalide Cirrus's build cache when needed, as suggested in
+# https://github.com/cirruslabs/cirrus-ci-docs/issues/544#issuecomment-566066822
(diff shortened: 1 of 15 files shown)


ckreibich (approved):
Thank you! Afaik the existing cache invalidation stuff on Cirrus doesn't cover the Docker builds, so yet more reason to add this. 

There's a "clear all repository caches" button here ...
https://cirrus-ci.com/settings/repository/5681261976223744

... but Fedor never followed up here:
https://github.com/cirruslabs/cirrus-ci-docs/issues/544

PR #2124, opened 2022-05-23
Use ZEEK_DICT_DEBUG more in Dict.cc instead of DEBUG

timwoj:
This changes all of the uses of `DEBUG` to `ZEEK_DICT_DEBUG` in the dictionary code to avoid a big slow-down when using `DEBUG` in general.


timwoj:
@vpax I could have sworn this got merged in with some other stuff you were doing but it apparently didn't.

vpax:
I thought so too!  But it's possible I dropped it going from my "work" branch to my PR branch assuming you'd do it separately.

rsmmr (approved):

Issue #1995, opened 2022-03-07
Better error message for parsing pcap with non-Ethernet encapsulation


ynadji:
When Zeek is run on pcaps it does not understand due to encapsulation, it throws the following error:

```
$ capinfos ../udp.pcap| grep Encap
                     Encapsulation = IEEE 802.15.4 Wireless PAN (104 - wpan)
$ zeek -Cr ../udp.pcap
fatal error in /usr/local/share/zeek/base/frameworks/packet-filter/./main.zeek, line 285: Bad pcap filter 'ip or not ip'
```

For someone new to Zeek, this is a fairly cryptic message as evidenced by [this thread](https://zeekorg.slack.com/archives/CSZBXF6TH/p1646180487407939) on the Zeek Slack. We were able to address the issue by changing the encapsulation type, but if the error suggested this as a possible root cause I believe newer users could figure out how to fix the problem themselves without having to ask in Slack.

I was going to open a PR that provided a hint, but I realized I don't quite understand the implications here. I see the error is primarily structured to show [a BPF filter compilation error](https://github.com/zeek/zeek/blob/52b3e159a415fb3acd584f95aa64e2d2a9c0b282/scripts/base/frameworks/packet-filter/main.zeek#L285), but that doesn't seem to be the case here. Is there a way to differentiate these two? Is there a list of encapsulation types that Zeek supports somewhere? Is the right way™ to fix this to simply add new encapsulation types? Happy to try to address this issue, but I'd appreciate some feedback from the Zeek pros on the smart way to do this.

ynadji:
Another example of this biting users:
* https://zeekorg.slack.com/archives/CSZBXF6TH/p1652827889017099
* https://zeekorg.slack.com/archives/CSZBXF6TH/p1652828033482849

timwoj:
The error message set in `PktSrc::Error` is the same error though:

```
(const std::string &) msg = "cannot compile BPF filter \"ip or not ip\""
```

which continues to make it not terribly useful. Even if it got past the call to`FatalError` it would just result in the same error message later. The PktSrc opens just fine here because the pcap PktSrc doesn't really check for anything other than the file being a valid pcap. It doesn't care what's actually in it.

There's a couple issues here:

1. Yes, that message does get eaten and it shouldn't. At the very least, it should get propagated up to script land so it can be output instead of just dumping out something generic.
[...]

PR #2126, opened 2022-05-24
Propagate BPF_Program error message to script land

timwoj:
This fixes the messaging part of #1995. There's two things fixed here:

1. The error messages from `BPF_Program::Compile()` aren't propagated into script-land, and instead a generic error message is printed on failures with just the filter string that failed. The error string is now propagated for pcap packet sources.
2. There are two versions of `BPF_Program::Compile()`, one of which takes a `pcap_t*` pointer and uses a different method that allows for returning more informative error messages. We were only calling the non-error message version, even via the pcap packet source. This adds a new method that packet sources can override to call different methods for compiling the filters, and the pcap source uses this to call the correct version of `Compile()`.


rsmmr (approved):

timwoj:
Re: the above test output, would it be enough to fix the test to just search for `syntax error` and return success if it's found?

PR #2129, opened 2022-05-24
Management framework: cluster node placement and log management

ckreibich:
This is a series of smallish changes to make defaults "just work" for operational use of a Zeek cluster in a single-machine deployment. When using the management framework, the following now hold:

- Each Zeek node (including agents and controller) run out of their own directory in `ZEEK_SPOOL/nodes/<node_name>`
- Each node has its own `stdout` / `stderr` file
- Log rotation is active, with rotated logs ending up in `ZEEK_SPOOL/log-queue` for later finalization by `zeek-archiver`


scripts/policy/frameworks/management/persistence.zeek
```diff
@@ -0,0 +1,52 @@
...
+	local open_str = strftime(Log::default_rotation_date_format, ri$open);
+	local close_str = strftime(Log::default_rotation_date_format, ri$close);
+	local base = fmt("%s__%s__%s__", ri$path, open_str, close_str);
+	local rval = Log::RotationPath($file_basename=base);
+	return rval;
+	}
+
+redef Log::default_rotation_interval = 1 hrs;
+redef Log::enable_local_logging = T;
+redef Log::enable_remote_logging = F;
```
rsmmr, line 45: I'm wondering if they should be logging remotely as well (or even only? not sure).

scripts/policy/frameworks/management/persistence.zeek
```diff
@@ -0,0 +1,52 @@
...
+
+# For any kind of Zeek process we steer rotated logs awaiting archival into a
+# queue directory in the spool.
+redef Log::default_rotation_dir = build_path(Management::get_spool_dir(), "log-queue");
+
+# Additional default storage locations:
+redef Broker::table_store_db_directory = build_path(Management::get_state_dir(), "broker");
+@if ( getenv("CLUSTER_NODE") != "" )
+redef Cluster::default_store_dir = build_path(Management::get_state_dir(), "cluster");
+@endif
```
rsmmr, line 27: This is a central directory for all nodes, right? I'm wondering whether the storage backing a table should stay node-specific. In principle one can create Broker storage masters on any node. If that's useful, might depend; but if one does, I don't think they their storage should be visible to each other (because you'd be using clones if you want state synced across).

rsmmr (approved):

scripts/policy/frameworks/management/persistence.zeek
ckreibich: I think "as well" is a good call, yeah. There's currently nothing that subscribes to logs from agent/controller, so I figured I'd keep it off, but that might change soon when we expand the client's ability to receive logs. For now I don't think we can get by without local logging for these. I'll push an update.

scripts/policy/frameworks/management/persistence.zeek
ckreibich: Yeah, it is. Here too, I was wondering too. :-) This is another concept from zeekctl, per its `BrokerDBDir` and `DefaultStoreDir` options, which are also global. If we can do without these, I'd prefer it.

It does bring up one question: where should each node's directory live? Right now that's under `spool/nodes`, with the assumption that there's only log content in there. That's already a bit of a stretch given that the `stdout`/`stderr` files are in there too. My take on spools is that one thing dumps stuff there, and another thing cleans it out, so for non-logs that doesn't fit well. With this change to the stores, I'm thinking the node directories should instead be `var/lib/nodes/<name>`. The spool would then really only be for rotated but not yet archived logs — perhaps in `spool/logs/`, to leave room for other types of spool content in the future.

Does that sound good?

rsmmr: Yes, that sounds good. That indeed expresses the difference in semantics as well: this is important data that's to be persisted, rather a "spool" where things just move through.

scripts/policy/frameworks/management/persistence.zeek
rsmmr: Would be neat if these logs came out through the logger as well, though that's a chicken-and-egg I guess ...

PR #2130, opened 2022-05-25
Bump Spicy and spicy-plugin

bbannier:


bbannier:
This seems to fix most of the compiler warnings reported in zeek/spicy#1188 @ckreibich @timwoj, but I still see some warnings due to Spicy's `IntrusivePtr`. If you are fine with another bump of Spicy later, I think this one could be ready for review.

timwoj (approved):

PR #2139, opened 2022-05-31
Management framework: monitor outcome of cluster deployment and report failures

ckreibich:
Deploying a cluster config so far was best-effort: you'd trigger it, and then infer via `get-nodes` whether the cluster launched as you desired. This PR fixes this as follows:
- It verifies that launched nodes actually get off the ground by waiting for them to check in with the agent via a `Management::Node::API::notify_node_hello`. This is quite handy and brings us half-way to the "cluster is ready" event we've discussed in the past: all that's needed now is an extra roundtrip via the controller to determine that all agents have had their nodes come up, and then broadcast that fact.
- The `set_configuration` response event now normally contains one record per deployed node, indicating its outcome. In the corner-case of deploying no nodes, a single result record still reports success. 
- For failed nodes this output now includes the stdout and stderr of the failed node. This was pretty tricky and relies on a new set of scripts the framework now deploys into the Supervisor, for added state-keeping. This relies on #2138 to build up the output state while still allowing Zeek nodes to write dedicated stdout/stderr files.

Example of a launch with a failing node:

```
$ zeek-client set-config etc/cluster.cfg
{
  "logger-01": {
    "instance": "agent-hogthrob4",
    "success": true
  },
  "logger-02": {
    "instance": "agent-hogthrob4",
    "success": true
  },
  "manager": {
    "instance": "agent-hogthrob4",
    "stderr": "error in /home/christian/inst/opt/zeek/share/zeek/base/frameworks/cluster/./nodes/manager.zeek, line 26: unknown identifier EEEK, at or near \"EEEK\"\ninternal error in /home/christian/inst/opt/zeek/share/zeek/base/frameworks/cluster/./nodes/manager.zeek, line 26: Failed to find variable named: Spicy::codegen_debug",
    "stdout": "",
    "success": false
  },
  "worker-01": {
    "instance": "agent-hogthrob4",
    "success": true
  },
  "worker-02": {
    "instance": "agent-hogthrob4",
    "success": true
  }
}
```


ckreibich:
This is a stacked PR, currently sitting on top of the #2138 branch.

rsmmr (approved):

Issue #2101, opened 2022-05-13
Build failure if older version of c-ares already exists on the system


timwoj:
Thanks to @leres for tracking this down a bit further.

If an older version of c-ares is already installed on the system in /usr/local, Zeek builds will fail with the following error:

```
[ 29%] Building C object auxil/c-ares/src/lib/CMakeFiles/c-ares.dir/ares__addrinfo2hostent.c.o
/home/leres/TEST/zeek-master.git/auxil/c-ares/src/lib/ares__addrinfo2hostent.c:132:41: error: no member named 'name' in 'struct ares_addrinfo'
      (*host)->h_name = ares_strdup(ai->name);
                                    ~~  ^
/home/leres/TEST/zeek-master.git/auxil/c-ares/src/lib/ares__addrinfo2hostent.c:133:42: error: no member named 'name' in 'struct ares_addrinfo'
      if ((*host)->h_name == NULL && ai->name)
                                     ~~  ^
2 errors generated.
*** Error code 1
```

From https://github.com/zeek/zeek/pull/2097#issuecomment-1125535533:

> 
> What I've found is that auxil/c-ares/src/lib/CMakeFiles/c-ares.dir/flags.make defines C_INCLUDES and puts -I/usr/local/include before -I/home/leres/TEST/leres-zeek.git/auxil/c-ares/include:
> 
> ```
> -I/home/leres/TEST/leres-zeek.git/auxil/rapidjson/include
> -I/home/leres/TEST/leres-zeek.git/auxil/paraglob/include
> -I/home/leres/TEST/leres-zeek.git/auxil/highwayhash
> -I/home/leres/TEST/leres-zeek.git/build/auxil/broker/include
> -I/home/leres/TEST/leres-zeek.git/auxil/broker/include
> -I/usr/local/include
> -I/home/leres/TEST/leres-zeek.git/build/auxil/binpac/lib
> -I/home/leres/TEST/leres-zeek.git/auxil/binpac/lib
> -I/home/leres/TEST/leres-zeek.git/build/auxil/c-ares
> -I/home/leres/TEST/leres-zeek.git/auxil/c-ares
> -I/home/leres/TEST/leres-zeek.git/auxil/c-ares/include
> -I/home/leres/TEST/leres-zeek.git/auxil/c-ares/src/lib/include
> ```
> 
> But I don't understand what determines the order so I don't know how to change it.

This is probably a blocker for 5.0.

leres:
[...]

PR #2143, opened 2022-05-31
GH-2101: Update cmake submodule to fix c-ares include path problem

timwoj:
We still haven't managed to reproduce this anywhere but on @leres's system, but this change should fix the problem there. I have some concerns about whatever causes c-ares to insert /usr/local/include into the path breaking something else, but again I haven't been able to nail down why it actually does that.

This requires merging the branch with the same name on the cmake submdoule.


rsmmr (approved):
Also looked at the CMake change, +1 there too.

PR #2147, opened 2022-06-02
Add lookup_autonomous_system() BIF that returns AS number and org

### LINKED ISSUE #2141: BIF to obtain autonomous_system_organization
I help maintain a Zeek package https://github.com/brimdata/geoip-conn that adds geolocation data to Zeek `conn` records. Up until now it's only added fields for the data returned by the `lookup_location()` BIF. A community user recently filed a [feature request](https://github.com/brimdata/geoip-conn/issues/39) to have autonomous system data added. I found I was able to leverage the `lookup_asn()` BIF to get the `autonomous_system_number` from the MaxMind ASN DB, but it looks like there's not currently a BIF to acquire the `autonomous_system_organization`. Any chance a BIF could be added to include this also?

philrz:
The existing BIF `lookup_asn()` only returns the autonomous system number. However, the ASN database from MaxMind includes the autonomous system organization as well. This PR adds a new BIF `lookup_autonomous_system()`  that returns a record containing both the ASN and the organization.

I took the approach of adding a new BIF rather than changing `lookup_asn()` since I figured users that might be depending on the existing BIF would not appreciate having to make changes to accommodate the different return type. With the approach taken here, users could switch over to the new BIF if that suits them and 
`lookup_asn()` could perhaps be deprecated in the future if desired.

I confess I'm not much of a C++ developer but was able to write this by following the pattern of `lookup_asn()` and `lookup_asn()`. To prove that it's working, I have a [branch of the geoip-conn Zeek package](https://github.com/philrz/geoip-conn/tree/as-org) that leverages this new BIF. Here's an example of the new AS org field being populated when Zeek processes the test pcap  [dns/loc-29-trunc.pcap](https://github.com/zeek/zeek/raw/master/testing/btest/Traces/dns/loc-29-trunc.pcap) from the Zeek repo.

```
$ zkg install https://github.com/philrz/geoip-conn --version fe0655225ed8b0f9eef1674fa21a25b14fc65fb9
The following packages will be INSTALLED:
  https://github.com/philrz/geoip-conn (fe0655225ed8b0f9eef1674fa21a25b14fc65fb9)

Proceed? [Y/n] y
Installing "https://github.com/philrz/geoip-conn".............................................................................................................................................................................................
Installed "https://github.com/philrz/geoip-conn" (fe0655225ed8b0f9eef1674fa21a25b14fc65fb9)
Loaded "https://github.com/philrz/geoip-conn"

$ zeek -C -r loc-29-trunc.pcap local
WARNING: No Site::local_nets have been defined.  It's usually a good idea to define your local networks.

$ zq -Z conn.log 
{
    _path: "conn",
    ts: 2020-11-09T00:02:02.851655Z,
    uid: "CHNGUcFixuswCrxsi",
    id: {
        orig_h: 79.141.82.250,
        orig_p: 57483 (port=uint16),
        resp_h: 192.188.22.52,
        resp_p: 53 (port)
    },
    proto: "udp" (=zenum),
    service: "dns",
    duration: 195us,
    orig_bytes: 72 (uint64),
    resp_bytes: 234 (uint64),
    conn_state: "SF",
    local_orig: null (bool),
    local_resp: null (bool),
    missed_bytes: 0 (uint64),
    history: "Dd",
    orig_pkts: 1 (uint64),
    orig_ip_bytes: 100 (uint64),
    resp_pkts: 1 (uint64),
    resp_ip_bytes: 262 (uint64),
    tunnel_parents: null (|[string]|),
    geo: {
        orig: {
            country_code: "CH",
            region: "GE",
            city: "Bernex",
            latitude: 46.1727,
            longitude: 6.0732,
            as_number: 44166 (uint64),
            as_org: "NTD SA"
        },
        resp: {
            country_code: "US",
            region: null (string),
            city: null (string),
            latitude: 37.751,
            longitude: -97.822,
            as_number: 3443 (uint64),
            as_org: "ESNET-AS"
        }
    }
}
```

Fixes #2141

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -984,6 +984,14 @@ type geo_location: record {
 	longitude: double &optional;	##< Longitude.
 } &log;
 
+## GeoIP autonomous system information.
+##
+## .. zeek:see:: lookup_autonomous_system
+type geo_autonomous_system: record {
+	number: count &optional;	##< The autonomous system number.
+	organization: string &optional;	##< Associated organization.
+} &log;
+
 ## The directory containing MaxMind DB (.mmdb) files to use for GeoIP support.
 const mmdb_dir: string = "" &redef;
 
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -4319,7 +4319,7 @@ function lookup_location%(a: addr%) : geo_location
 ## Returns: The number of the ASN that contains *a*.
 ##
 ## .. zeek:see:: lookup_location lookup_autonomous_system
-function lookup_asn%(a: addr%) : count
+function lookup_asn%(a: addr%) : count &deprecated="Remove in v6.1.  Functionality is now handled by lookup_autonomous_system()."
 	%{
 #ifdef USE_GEOIP
 	mmdb_check_asn();
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -4198,7 +4198,7 @@ static bool mmdb_try_open_asn ()
 ##
 ## Returns: A boolean indicating whether the db was successfully opened.
 ##
-## .. zeek:see:: lookup_asn
+## .. zeek:see:: lookup_asn lookup_autonomous_system
 function mmdb_open_location_db%(f: string%) : bool
 	%{
 #ifdef USE_GEOIP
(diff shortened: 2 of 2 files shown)


timwoj:
> Should we consider deprecating `lookup_asn` since this is retrieving more data in the same way?

Whoops, didn't realize you mentioned this in the original PR description. My opinion is that we should go ahead and deprecate it, but mark it for removal in 6.1 instead of 5.1 since the 5.0 release is so close at this point.

PR #2148, opened 2022-06-02
Change table format so it better embeds in other sphynx docs

stevesmoot:
When using zeek metadata (comments) in making other docs the list style embeds much better than the table.  this also matches the format in conn for conn_state

### DIFF
--- a/scripts/base/protocols/conn/main.zeek
+++ b/scripts/base/protocols/conn/main.zeek
@@ -99,24 +99,19 @@ export {
 
 		## Records the state history of connections as a string of
 		## letters.  The meaning of those letters is:
-		##
-		## ======  ====================================================
-		## Letter  Meaning
-		## ======  ====================================================
-		## s       a SYN w/o the ACK bit set
-		## h       a SYN+ACK ("handshake")
-		## a       a pure ACK
-		## d       packet with payload ("data")
-		## f       packet with FIN bit set
-		## r       packet with RST bit set
-		## c       packet with a bad checksum (applies to UDP too)
-		## g       a content gap
-		## t       packet with retransmitted payload
-		## w       packet with a zero window advertisement
-		## i       inconsistent packet (e.g. FIN+RST bits set)
-		## q       multi-flag packet (SYN+FIN or SYN+RST bits set)
-		## ^       connection direction was flipped by Zeek's heuristic
-		## ======  ====================================================
+		## * s       a SYN w/o the ACK bit set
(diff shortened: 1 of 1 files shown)


ckreibich:
Sure Smoot, we can figure something out ... but let me at least ask — could you fix this with table styling or some such? In our docs the tables look better than lists, so it'd be nicer to move `conn_state` to a table than `history` to a list, an there aren't a lot of people who turn Zeek code comments into other documents. :-)

Btw, there's another table, in [`SSL::Info$history`](https://docs.zeek.org/en/master/scripts/base/protocols/ssl/main.zeek.html#type-SSL::Info) that would likely need to change as well.

rsmmr:
Yeah, I tend to agree with Christian here, would be nice if we could address this in some other way than removing the table layout. Could we add some additional markup that maybe Sphinx ignores, but helps externally? (Just brainstorming here).

0xxon:
Just adding a third agree here - tables look nicer; I actually just recently reformatted something in the SSL docs to use tables because of that.

rsmmr:
So the underlying issue here is that a table makes it difficult to include that markup into *another* table: Sphinx doesn't support nested tables apparently. It's a not a problem for the Zeek docs, but can be elsewhere.

rsmmr:
A definition list I can see I think, that keeps the structure at least.

rsmmr:
So I tried the definition list idea, but it doesn't look very nice either I think honestly:
Screen Shot 2022-09-28 at 11 11 36

awelzel:
> So I tried the definition list idea, but it doesn't look very nice either I think honestly:

>> The docs use zeekygen comments to auto-gen the script reference, but rather than rendering record types into a table, it renders as a definition list that is then styled table'ish (?). Maybe that's an alternative to rendering record types into tables

Oh oh :-/ I had meant that @stevesmoot could consider rendering the outer record types as definition lists (as is done on docs.zeek.org) and keep the table within that.

stevesmoot:
I agree that doesnt look great.  Over the weekend I started on my ZW talk and wrote some code to detect def-list tables in comments and reformat them in the RST output.  So I guess we can close as WONTFIX.

rsmmr:
> Oh oh :-/ I had meant that @stevesmoot could consider rendering the outer record types as definition lists (as is done on docs.zeek.org) and keep the table for the history within that as a workaround.

Yeah, but I had been wondering if we could generally switch to that in the code as well.

PR #2154, opened 2022-06-02
Merge PQ_Timer into base TimerMgr class

timwoj:
This is a post-5.0-release PR.

We removed `CQ_TimerMgr` back in the 3.1 release, in favor of `PQ_TimerMgr`. I'm not sure of the details why we didn't merge these together back then, but this PR merges the `PQ_Timer` code into the base `TimerMgr` class. We've had 2.5 years to implement another timer manager, and no one's done it. It's under the `zeek::detail` namespace so `PQ_TimerMgr` doesn't need to be deprecated here.

### DIFF
--- a/src/Timer.cc
+++ b/src/Timer.cc
@@ -72,11 +72,16 @@ TimerMgr::TimerMgr()
 	num_expired = 0;
 	last_advance = last_timestamp = 0;
 
+	q = std::make_unique<PriorityQueue>();
+
 	if ( iosource_mgr )
 		iosource_mgr->Register(this, true);
 	}
 
-TimerMgr::~TimerMgr() { }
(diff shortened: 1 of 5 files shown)


src/Timer.cc
```diff
@@ -72,11 +72,16 @@ TimerMgr::TimerMgr()
...
+	q = std::make_unique<PriorityQueue>();
+
 	if ( iosource_mgr )
 		iosource_mgr->Register(this, true);
 	}
 
-TimerMgr::~TimerMgr() { }
+TimerMgr::~TimerMgr()
+	{
+	q.reset();
```
rsmmr, line 83: Isn't this happening automatically anyways?

PR #2163, opened 2022-06-04
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron: '0 0 * * 1,3,5'
 
+permissions:
+  contents: read
+
 jobs:
   scan:
     if: github.repository == 'zeek/zeek'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,6 +18,9 @@ env:
   IMAGE_FILE: /tmp/zeek-image.tar
   IMAGE_PATH: /tmp
 
+permissions:
+  contents: read
+
 jobs:
   docker-build:
     runs-on: ubuntu-latest
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -9,8 +9,13 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  contents: read
+
 jobs:
   generate:
+    permissions:
+      contents: write  # for Git to git push
     if: github.repository == 'zeek/zeek'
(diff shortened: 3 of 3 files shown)


timwoj:
Can you give some detail about what exactly this is fixing? For example, we already know the Docker workflow already doesn't have access to secrets if running from forks. We had it fail on forked builds, prior to adding the check to only allow it to run from the main repo.

0xxon:
@naveensrinivasan: small ping on the question that Tim raised.

naveensrinivasan:
> @naveensrinivasan: small ping on the question that Tim raised.

It is just not about tokens. 

Without right permissions like read a dangerous workflow can write to the repository

0xxon:
Ok, but isn't contents: read the default permission, which means that this changes nothing?

https://docs.github.com/en/actions/security-guides/automatic-token-authentication seems to indicate so.

naveensrinivasan:
> Ok, but isn't `contents: read` the default permission, which means that this changes nothing?
> 
> https://docs.github.com/en/actions/security-guides/automatic-token-authentication seems to indicate so.

It is based on how the permission is set https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I'll let you decide. Thanks

timwoj:
I set the default permissions in the Zeek org to the more restrictive setting. I'm also going to merge this PR, since it'll be needed for the docs push to work, but I'm curious to know what else breaks by me setting the org permissions. We'll have to keep an eye on things for a bit.

bbannier:
> ... I'm curious to know what else breaks by me setting the org permissions. We'll have to keep an eye on things for a bit.

This broke the Spicy release action which uses `softprops/action-gh-release` to create a release and upload artifacts. Explicitly requesting

```yaml
permissions:
    contents: write
```

made this work again.

PR #2168, opened 2022-06-08
Management framework: fix agent, controller, and supervisor network reachability

ckreibich:
Agent and controller by default listened only on loopback, while the Supervisor (when running) listened globally. This flips it around for saner defaults.


scripts/policy/frameworks/management/controller/config.zeek
```diff
@@ -86,7 +86,7 @@ function network_info(): Broker::NetworkInfo
 	else if ( Management::default_address != "" )
 		ni$address = Management::default_address;
 	else
-		ni$address = "127.0.0.1";
+		ni$address = "0.0.0.0";
```
rsmmr, line 89: Torn if the same applies here: when on a single-system cluster, make the default local? Controller is different of course because the client may run elsewhere, but still. Either way, would be helpful to show people how to change this setting, like with a short example in the docs.

rsmmr (approved):

scripts/policy/frameworks/management/agent/config.zeek
```diff
@@ -106,7 +106,7 @@ function endpoint_info(): Broker::EndpointInfo
 	else if ( Management::default_address != "" )
 		network$address = Management::default_address;
 	else
-		network$address = "127.0.0.1";
+		network$address = "0.0.0.0";
```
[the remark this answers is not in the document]
ckreibich: Good thought, yeah. I hit the same question in the context of config validation yesterday: I wondered whether to complain about configs in which a cluster node requests the same listening port as the controller — for which you need to know whether the controller runs alongside that agent. Turns out that at moment this is relatively tricky to answer. An easy solution would be for the agents' `notify_agent_hello` event to controllers to include the hostname they run on — controllers can check that against their own. There's enough here for a follow-up PR.

scripts/policy/frameworks/management/controller/config.zeek
ckreibich: Yes, agree here too. I'll cover it.

PR #2170, opened 2022-06-10
Management framework: zeek-archiver support

ckreibich:
By default, agents now run `zeek-archiver` automatically, with archived logs ending up in the installation's logs directory (and a datestamp subdirectory in it, provided by `zeek-archiver`). The user can disable this, configure a different cmd, configure a different destination, and adjust the archival interval.


scripts/policy/frameworks/management/agent/main.zeek
```diff
@@ -142,6 +153,52 @@ function send_set_configuration_response(req: Management::Request::Request)
...
+	# this will often be expected, since in larger clusters many instances
+	# may not run loggers.
+	if ( run_archival && file_size(Log::default_rotation_dir) > 0 )
+		{
+		cmd = fmt("%s -1 %s %s",
+		    cmd, Log::default_rotation_dir,
+		    Management::Agent::archive_dir);
+
+		Management::Log::info(fmt("triggering log archival via '%s'", cmd));
+		system(cmd);
```
rsmmr, line 196: I'm going back and forth on using `system()` here, but I think it's ok: it'll background the job and if Zeek terminates while it's still in progress, the archive should just retry next time. I'd still say it'd be a nicer to have either the supervisor control the process, or the archiver daemonize itself; but both would be quite a bit of work, so seems ok for now (until we maybe teach the supervisor to run external processes later anyways).

rsmmr (approved):

PR #2172, opened 2022-06-14
Management framework: config validation

ckreibich:
This adds a validation stage to the controller that looks for possible misconfigurations when the client uploads a configuration. 

Also I noticed that I overlooked possible collisions in node port auto-assignment, for which this includes a fix, as well as for the corner case of entirely empty configurations.

Finally, this adds two on/off knobs for allowing the stdout/stderr of output coming from nodes to appear on the Supervisor console despite redirection to files, since this allows "docker logs" to show all such output, simplifying test troubleshooting.

### DIFF
--- a/scripts/policy/frameworks/management/controller/main.zeek
+++ b/scripts/policy/frameworks/management/controller/main.zeek
@@ -103,6 +103,13 @@ global config_nodes_lacking_ports: function(config: Management::Configuration): 
 # port. This assumes those ports are actually available on the instances.
 global config_assign_ports: function(config: Management::Configuration);
 
+# Validate the given configuration in terms of missing, incorrect, or
+# conflicting content. Returns T if validation succeeds, F otherwise. The
+# function updates the given request's success state and adds any error
+# result(s) to it.
+global config_validate: function(config: Management::Configuration,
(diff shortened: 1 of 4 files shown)


scripts/policy/frameworks/management/controller/main.zeek
```diff
@@ -668,8 +668,25 @@ event Management::Controller::API::set_configuration_request(reqid: string, conf
...
+		g_config_reqid_pending = "";
+
+		Management::Log::info(fmt("tx Management::Controller::API::set_configuration_response %s",
+		    Management::Request::to_string(req)));
+		Broker::publish(Management::Controller::topic,
+		    Management::Controller::API::set_configuration_response, req$id, req$results);
+		Management::Request::finish(req$id);
+		}
+
+	# Otherwise, check if we're able to send the config to all agents
```
timwoj, line 866: This "otherwise" makes me think this this call should either be in an `else` or the above block should have returned. Is that right?

timwoj (changes requested):

rsmmr (approved):

timwoj (approved):

PR #2174, opened 2022-06-14
CI: Add Ubuntu 22.04 (LTS Release, EOL 04/2027)

0xxon:


timwoj (approved):
Can you make sure to make the same changes in the broker configuration while you're at it?

PR #2182, opened 2022-06-14
Fix pattern matching in Cirrus dependabot check

timwoj:


timwoj:
I pull this change in via one of the dependabot PRs.

PR #2184, opened 2022-06-15
CI: Add Ubuntu 22.04 (LTS Release, EOL 04/2027) (For release/5.0)

0xxon:
#2174 equivalent for release/5

### DIFF
--- a/ci/ubuntu-22.04/Dockerfile
+++ b/ci/ubuntu-22.04/Dockerfile
@@ -1,0 +1,39 @@
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND="noninteractive" TZ="America/Los_Angeles"
+
+# A version field to invalide Cirrus's build cache when needed, as suggested in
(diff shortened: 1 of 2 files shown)


ci/ubuntu-22.04/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+    sqlite3 \
+    swig \
+    unzip \
+    wget \
+    zlib1g-dev \
+  && apt autoclean \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install websockets junit2html
+RUN gem install coveralls-lcov
```
timwoj, line 39: The lcov gem only needs to be installed on whatever ubuntu version we're using for the sanitizer build, since that's the only place the coverage stuff runs. That said, I guess it's fine to install it on all of the ubuntu ones just in case we change which VM that uses.

timwoj (approved):

PR #2194, opened 2022-06-17
Convert Dictionary types to be templated classes

timwoj:
This is something I've had sitting on a branch and have been rebasing for over a year now. I had intended to do more to it before doing a PR, but instead decided to take one leap and get this part in first. The rest can come later, and is described below. This isn't ready for merging yet, since it depends on a couple of commits in the deprecation-removal work that I'll open a PR for after 5.0 is officially out. Only the "Convert" commit below is core to this work.

This PR converts the `Dictionary`, `DictIterator`, `RobustDictIterator`, and `DictEntry` types to be templates. This has the fortunate side effect of removing the need to store the value in each entry as a `void*`, since we can now know the actual type stored in the dictionary. The only changes outside of the Dict files are the need to remove uses of `DictEntry::GetValue<>` (now a deprecated method) and the need to specify template types for iterators. My only real concern with these changes is the potential for a bit slower build times, since most of the code was moved out of `Dict.cc` and into `Dict.h`, which means the whole thing needs to be read every time we include it (which is roughly a third of the build).

Also included is a commit that turns `PDict` into a simple type alias. This requires adding a few more functions to the base `Dictionary` class, but it leaves the possibility to just remove `PDict` altogether (after deprecation) and rely just on the base class for all dictionary objects.

Future things I'd like to tackle:
- Making the outward-facing dictionary API more like `std::map`'s, just like we did for `List` a couple years ago.
- Support [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding) for iteration. I started working on this originally and it became quite complicated with the keying stuff in `DictEntry`. I haven't revisited it in a while and I might have a better idea now.
- Better commenting in the `Dict` header for the outward-facing API.

This PR is dependent on some of the deprecation cleanup in https://github.com/zeek/zeek/pull/2199 and that should be merged first.


rsmmr (approved):
I'm assuming that most of the changes in `Dict.{h,cc}` is just moving code from `cc` to `h`, so didn't look much at the implementation code. 

> the need to specify template types for iterators

That's the one breaking change, where we can't deprecate first, right? Feels ok to me though, given how this makes usage easier and the fix is simple.

> My only real concern with these changes is the potential for a bit slower build times

Would it be helpful to move some code that's independent of T into a base class, where it could then stay in `cc`? However, not sure how much code that would actually apply to, so may not be worth the trouble.

timwoj:
> I'm assuming that most of the changes in Dict.{h,cc} is just moving code from cc to h, so didn't look much at the implementation code.

Yep. It's mostly just moving code from the cc file to the h file. There's some minor changes here and there because we weren't terribly careful about keep argument names the same.

> That's the one breaking change, where we can't deprecate first, right? Feels ok to me though, given how this makes usage easier and the fix is simple.

Correct, we can't deprecate it. That's part of why I'm doing this so early in the development cycle so that we can catch problems associated with it well early. I'll add a note to NEWS for it as well.

> Would it be helpful to move some code that's independent of T into a base class, where it could then stay in cc? However, not sure how much code that would actually apply to, so may not be worth the trouble.

I'll look into that in a separate PR, in the interest of getting this one in.

PR #2195, opened 2022-06-17
Removing deprecated capture-by-reference closures

vpax:
This PR removes the previously deprecated form of lambda closures that captured values by reference instead of by copy.  Say goodbye to a lot of cruft!


timwoj:
I'm going to hold off on merging this until 5.0 is fully released to avoid any weird merge conflicts that might arise if we need to cherry pick something over to the release branch.

src/Expr.cc
```diff
@@ -4747,32 +4751,28 @@ LambdaExpr::LambdaExpr(std::unique_ptr<function_ingredients> arg_ing, IDPList ar
...
 	{
 	auto ft = type->AsFuncType();
 	const auto& captures = ft->GetCaptures();
 
-	capture_by_ref = false;
+	auto desc = when_parent ? "\"when\" statement" : "lambda";
 
 	if ( ! captures )
 		{
 		if ( outer_ids.size() > 0 )
```
timwoj, line 4763:
```suggestion
		if ( ! outer_ids.empty() )
```

timwoj (approved):
One minor nit. I'll fix it during the merge.

src/Expr.cc
vpax: Oops yeah, thx.  I try to canonicalize those but sometimes miss some!

PR #2199, opened 2022-06-20
Remove deprecations marked for v5.1

timwoj:
This is a counterpart to https://github.com/zeek/zeek/pull/2195, and also shouldn't be merged until 5.0 is fully released to avoid possible merge conflicts with things we might need to cherry-pick from master to release-5.0.

This is the remainder of the deprecations that are marked for removal in 5.0. It should probably be reviewed commit-by-commit since each commit should be fairly self-contained and not touch other areas.


scripts/base/files/x509/__load__.zeek
```diff
@@ -1,4 +1,4 @@
 @load ./main
 @load ./certificate-event-cache
 
-@load ./log-ocsp
+@load ./log-ocsp
```
0xxon, line 4: This one is a mistake, right? The only difference is now that the EOL at the end of the file is missing - which should be there.

src/Dict.cc
```diff
@@ -530,80 +328,82 @@ TEST_CASE("dict robust iteration replacement")
 	delete val4;
 	}
 
+// TODO: should this test be rewritten for the new iteration style?
+/*
 TEST_CASE("dict iterator invalidation")
-	{
-	PDict<uint32_t> dict;
+    {
+    PDict<uint32_t> dict;
```
0xxon, line 335: These and the following changes all look extremely weird - I know that this is now commented out, but it simultaneously changes all indentation tabs to spaces.

That is not right, is it?

0xxon (commented):
Looks good to me, minus the indentation changes that I commented on.

That being said - I did not look at every single line in excessive detail :)

scripts/base/files/x509/__load__.zeek
timwoj: Fixed.

src/Dict.cc
timwoj: I went back and reimplemented that test using the new iterators, like I should have done in the first place.

Issue #2198, opened 2022-06-20
Misidentification of certain TLS protocol values in Zeek, especially in DTLS


0xxon:
Aashish shared an interesting trace with me, in which the `ssl.log` shows a random `unknown-` value as the SSL protocol version.

Closer examination shows that this is a DTLSv1.2 connection, which is established after STUN is used in the same connection. In this case, the originators of the STUN session and of the DTLS sessions are reversed. This means that for the UDP connection, the side that sends the client hello is the receiver, and the side that sends the server hello is the originator.

It turns out that both our SSL parsing and our SSL scripts currently use the originator/responder of the underlying connection to identify a few packet types. In this specific case, the client sent a `supported_versions` extension. The parser and the scripts identified this as a server_side extension (since is_orig = F) and used the bogus value in there as the version of the TLS connection.

This is a bit of an annoying problem. The best easily workable solution is probably to have a separate definition of origin/responder for TLS/DTLS, which checks which side of the connection sends the client_hello. That way also would require next to no script changes.

0xxon:
It turns out that we have a test in our testsuite that shows this behavior (`dtls-stun-dpd.test`). We just don't use any values from the extensions in this case - which is why it was never noticed.

PR #2211, opened 2022-06-24
SSL Analyzer: track connection direction by messages

0xxon:
This PR changes the way in which the SSL analyzer tracks the direction
of connections. So far, the SSL analyzer assumed that the originator of
a connection would send the client hello (and other associated
client-side events), and that the responder would be the SSL servers.

In some circumstances this is not true, and the initiator of a
connection is the server, with the responder being the client. So far
this confused some of the internal statekeeping logic and could lead to
mis-parsing of extensions.

This reversal of roles can happen in DTLS, if a connection uses STUN -
and potentially in some StartTLS protocols.

This PR tracks the direction of a TLS connection using the hello
request, client hello and server hello handshake messages. Furthermore,
it changes the SSL events from providing is_orig to providing is_client,
where is_client is true for the client_side of a connection. Since the
argument positioning in the event has not changed, old scripts will
continue to work seamlessly - the new semantics are what everyone
writing SSL scripts will have expected in any case.

There is a new event that is raised when a connection is flipped. A
weird is raised if a flip happens repeatedly.

Addresses GH-2198.


src/analyzer/protocol/ssl/proc-certificate.pac
```diff
@@ -3,10 +3,18 @@ function proc_certificate(is_orig: bool, certificates : bytestring[]) : bool
...
+
+#ifndef USE_FLIPPED
+	bool flipped_ = zeek_analyzer()->GetFlipped();
+#endif
+
 	zeek::ODesc common;
 	common.AddRaw("Analyzer::ANALYZER_SSL");
 	common.Add(zeek_analyzer()->Conn()->StartTime());
-	common.AddRaw(is_orig ? "T" : "F", 1);
+	common.AddRaw(is_orig ^ flipped_ ? "T" : "F", 1);
```
rsmmr, line 17: Fancy. :-) I'm going back and forth if I would find method/function `IsClient()` more readable, as this `^` construct is used quite a bit. Can't decide, so I'll just note it.

src/analyzer/protocol/ssl/proc-certificate.pac
```diff
@@ -3,10 +3,18 @@ function proc_certificate(is_orig: bool, certificates : bytestring[]) : bool
 	if ( certificates->size() == 0 )
 		return true;
 
+	// this has to execute in both contexts, ssl and tls-handshake. In one we have flipped_,
+	// in the other we have ssl_analyzer()->GetFlipped(). And in both cases the other case
+	// does not work (and cannot be made to work easily).
+
+#ifndef USE_FLIPPED
```
rsmmr, line 10: Can we avoid this?

src/analyzer/protocol/ssl/tls-handshake.pac
```diff
@@ -8,6 +8,7 @@
 
 #include "zeek/analyzer/protocol/ssl/types.bif.h"
 #include "zeek/analyzer/protocol/ssl/events.bif.h"
+#define USE_FLIPPED
```
rsmmr, line 11: Why do we need this define?

rsmmr (approved):

src/analyzer/protocol/ssl/tls-handshake.pac
0xxon: So....
`proc-certificate.pac` is included by both the ssl analyzer (for SSLv2) and the SSL handshake analyzer (for SSLv3 up). The tracking takes place in the SSL handshake analyzer. So - in case this is called in the context of the handshake analyzer, you can just access the local variable. In case of the SSL analyzer, you have to call the function that calls into the SSL handshake analyzer to get the variable.

I was not able to come up with a better way that does not have tons of code duplication. I at first tried to do it with c++ metaprogramming - but I couldn't get that to work - and it started looking like a horrible mess.

I am open to better ideas though :)

0xxon: Oh, yes - and the define here is to indicate that this is the TLS handshake analyzer that has the `flipped_` variable available. Proc-certificate uses that to just use it, if the define is present (which is it, because it is included indirectly from this file).

On the plus side - this is not easy to mess up. If you remove this, the build breaks.

rsmmr: Ah, ok, now I also get the comment. :) 

This is in ` proc_certificate`, could you pass `flipped_` in as a parameter? If not, then I don't have a better idea either, fine with me.

0xxon: Yup, that works. I... for some reason did not think about that. I am going to update this PR, let the testsuite run, and then merge.

PR #2212, opened 2022-06-24
GH-2183: Rework Packet checksummed variable naming

### LINKED ISSUE #2183: Packet::{l2,l3}_checksummed variables are poorly named
While looking at finally merging https://github.com/J-Gras/zeek-af_packet-plugin/pull/30, @J-Gras noticed that the `l2_checksummed` and `l3_checksummed` variables in `Packet` are poorly named. `l2` should be the network layer (meaning IP/Layer 3) and `l3_checksummed` should be the transport layer (meaning TCP,etc/Layer 4). We should rename these variables for correctness.

We can't just outright rename them because they have to go through the deprecation cycle. We could either just extend the names and make them `layer{3,4}_checksummed` or we can name them something like `mac_checksummed` and `trans_checksummed`, but neither of those names feel really great.

timwoj:
This fixes the issues with the `Packet::{l2,l3}_checksummed` variables as laid out in #2183. It moves the two existing variables to match the layers they actually apply to, and adds a new `l4_checksummed` variable to cover the transport layer.

Fixes #2183


timwoj:
@grigorescu @J-Gras does this look good to you guys?

You'll have to add an ifdef block around the code in the DPDK plugin to check for the Zeek version, since the variables aren't being deprecated here even though the semantics are changing. I think you're the only user of them in their current state.

J-Gras:
I suggest to add a comment that documents `Packet::l{2,3,4}_checksummed` flags only apply to the outermost encapsulation layer in the presence of tunnels. For example, Linux seems to support offloading further down the stack (see [skb->csum_level](https://elixir.bootlin.com/linux/latest/source/include/linux/skbuff.h#L116)).

timwoj:
> I suggest to add a comment that documents Packet::l{2,3,4}_checksummed flags only apply to the outermost encapsulation layer in the presence of tunnels. For example, Linux seems to support offloading further down the stack (see [skb->csum_level](https://elixir.bootlin.com/linux/latest/source/include/linux/skbuff.h#L116)).

Oh whoops. I had done that, and then forgot to force-push the branch. It's there now.

grigorescu:
Looks good to me, thanks @timwoj !

rsmmr (approved):

Issue #405, opened 2019-06-08
Zeek renaming: internals (C++ / CMake)


jsiwek:
The internal C++ source-code and CMake infrastructure have many uses of "bro" still, but not critical to change those for 3.0.  A good time to consider updating them is if/when we revamp the API we expose for plugins to use.

timwoj:
I think with all of the changes from the namespacing work and PR #1144, all that's left here is bro_inet_ntop.h and all of the CMake stuff.

rsmmr:
This could also come with a separate v2 plugin CMake API.

timwoj:
After the above PR, the only things remaining are in the CMake modules. There's a plan to basically redo our CMake setup, so I'm going to close this out and we can revisit those namings in a separate issue/PR.

PR #2227, opened 2022-06-29
Rename almost all of the remaining ocurrances of "bro" in C++ land

timwoj:
Just because I felt like the code base shouldn't exist without some deprecations in it, this PR covers most of the remaining changes for https://github.com/zeek/zeek/issues/405. The only things left are the CMake bits and some variables in the SMB analyzer that I wasn't sure what to do with:

```
src/analyzer/protocol/smb/smb-mailslot.pac:39:	bro_major_ver : uint8;
src/analyzer/protocol/smb/smb-mailslot.pac:40:	bro_minor_ver : uint8;
src/analyzer/protocol/smb/smb-mailslot.pac:76:	bro_major_ver : uint8;
src/analyzer/protocol/smb/smb-mailslot.pac:77:	bro_minor_ver : uint8;
src/analyzer/protocol/smb/smb-mailslot.pac:89:	bro_major_ver : uint8;
src/analyzer/protocol/smb/smb-mailslot.pac:90:	bro_minor_ver : uint8;
```

There's also this line in the MIME analyzer that might be able to be changed to `DEBUG`:

```
src/analyzer/protocol/mime/MIME.cc:1278:#ifdef DEBUG_BRO
```

Lastly, it includes some changes to bifcl for the same kind of bro-renaming plus a bunch of changes to binpac that will be described in a separate PR.


testing/btest/Baseline/spicy.spicyz-jit/output
```diff
@@ -1,4 +1,49 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+In file included from /Users/tim/Desktop/projects/zeek/testing/btest/.tmp/test_f8f889238522e372-a3a474a9fddd8515.cc:8:
+/Users/tim/Desktop/projects/zeek/build/src/builtin-plugins/spicy-plugin/include/zeek-spicy/runtime-support.h:712:5: warning: 'bro_int_t' is deprecated: Remove in v6.1. Use zeek_int_t. [-Wdeprecated-declarations]
+    bro_int_t bt = (it >= 0 ? it : std::numeric_limits<::bro_int_t>::max());
```
rsmmr, line 4: Looks like we need to update the plugin. Either way, this should probably not be part of the baseline?

src/Obj.cc
```diff
@@ -60,7 +60,13 @@ int Obj::suppress_errors = 0;
 Obj::~Obj()
 	{
 	if ( notify_plugins )
-		PLUGIN_HOOK_VOID(HOOK_BRO_OBJ_DTOR, HookBroObjDtor(this));
+		{
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+		PLUGIN_HOOK_VOID(HOOK_BRO_OBJ_DTOR, HookObjDtor(this));
+#pragma GCC diagnostic pop
+		PLUGIN_HOOK_VOID(HOOK_OBJ_DTOR, HookObjDtor(this));
```
rsmmr, line 68: Isn't this running `HookObjDtor(this)` twice if there are both old-style and new-style plugins?

src/plugin/Manager.cc
```diff
@@ -883,26 +888,31 @@ void Manager::HookUpdateNetworkTime(double network_time) const
...
 
-	hook_list* l = hooks[HOOK_BRO_OBJ_DTOR];
+	hook_list* l = hooks[HOOK_OBJ_DTOR];
 
 	if ( l )
 		for ( hook_list::iterator i = l->begin(); i != l->end(); ++i )
 			{
 			Plugin* p = (*i).second;
-			p->HookBroObjDtor(obj);
+			p->HookObjDtor(obj);
```
rsmmr, line 911: I believe we need to call both here, otherwise the old one will not execute at all any more.

Actually, this together with the note above lets me wondering if we need to do this differently and treat the two as completely separate hooks that are just both triggered at the same time (inside the Obj destructor), even if that means code duplication.

src/plugin/Plugin.h
```diff
@@ -113,13 +116,82 @@ class Configuration
...
+	 */
+	[[deprecated("Remove in v6.1. Use zeek_version.")]] std::function<const char*()> bro_version;
 
 	/**
-	 * One can assign BRO_PLUGIN_BRO_VERSION to this to catch
+	 * One can assign ZEEK_PLUGIN_ZEEK_VERSION to this to catch
 	 * version mismatches at link(!) time.
 	 */
-	const char* (*bro_version)();
+	std::function<const char*()> zeek_version;
```
rsmmr, line 194: Do we need the new methods above because of the change to `std::function`? If so, I'm wondering if we should just stay with the raw function pointer just for brevity.

rsmmr (commented):
Nice job. My main question is around the plugin hook changes if they consistently trigger old and new hooks.

src/plugin/Plugin.h
timwoj: No, the same thing happens even without `std::function`. You get an error from the compiler about ambiguity with the `move` constructor specifically, but adding just that causes other errors to pop up.

src/plugin/Manager.cc
timwoj: That'd be easy enough to do, and to mark the old version of the request methods as deprecated.

testing/btest/Baseline/spicy.spicyz-jit/output
timwoj: I think the intention was to catch compilation warnings on every platform that zeek runs for CI, but I'd also think those should be caught during the spicy builds instead of the zeek builds.

timwoj: I don't think I can fix this in the context of the btest. The problem is that the output gets reordered depending on each run of the test so the various warnings show up in different places in the file, and the test fails. @bbannier Is there someway I can tell spicy via zeek to not output the compiler output during the run unless it fails?

src/plugin/Manager.cc
timwoj: I implemented this, and rebased the whole thing on `master` again.

testing/btest/Baseline/spicy.spicyz-jit/output
timwoj: > They are not emitted anymore if one drops e.g., HILTI_DEBUG=compiler

I tried that and it's still emitting the warnings.

src/Obj.cc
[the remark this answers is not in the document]
timwoj: Ah, you're right. I missed one 😅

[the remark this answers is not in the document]
rsmmr: > I don't see anywhere that `HookObjDtor` is implemented that `HookBroObjDtor` is missing

Ack, that was me misreading the diff.

rsmmr (approved):

PR #2228, opened 2022-06-30
Add gulrak/filesystem to auxil

awelzel:
This is adding https://github.com/gulrak/filesystem as a submodule
into auxil for use as a compiler-independent std::filesystem drop-in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -458,6 +458,17 @@ include_directories(BEFORE
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/auxil/rapidjson/include/rapidjson
         DESTINATION include/zeek/3rdparty/rapidjson/include)
 
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/auxil/filesystem/include/ghc
+        DESTINATION include/zeek/3rdparty/)
+
+# Create 3rdparty/ghc within the build directory so that the include for
+# "zeek/3rdparty/ghc/filesystem.hpp" works within the build tree.
+execute_process(COMMAND "${CMAKE_COMMAND}" -E make_directory
+                "${CMAKE_CURRENT_BINARY_DIR}/3rdparty/")
+execute_process(COMMAND "${CMAKE_COMMAND}" -E create_symlink
+                "${CMAKE_CURRENT_SOURCE_DIR}/auxil/filesystem/include/ghc"
+                "${CMAKE_CURRENT_BINARY_DIR}/3rdparty/ghc")
+
 # Optional Dependencies
 
 set(USE_GEOIP false)
--- a/.gitmodules
+++ b/.gitmodules
@@ -64,3 +64,6 @@
 [submodule "auxil/spicy"]
 	path = auxil/spicy/spicy
 	url = https://github.com/zeek/spicy
+[submodule "auxil/filesystem"]
+	path = auxil/filesystem
+	url = https://github.com/gulrak/filesystem.git
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -4,6 +4,7 @@
 path_classifiers:
   library:
     - "auxil/broker/caf/"
+    - "auxil/filesystem/"
     - "auxil/libkqueue/"
     - "auxil/highwayhash/"
     - "auxil/rapidjson/"
--- a/COPYING.3rdparty
+++ b/COPYING.3rdparty
@@ -533,6 +533,32 @@ POSSIBILITY OF SUCH DAMAGE.
 
 ==============================================================================
 
+%%% auxil/filesystem
+
+==============================================================================
+
+Copyright (c) 2018, Steffen Schümann <s.schuemann@pobox.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+==============================================================================
+
 %%% auxil/highwayhash
 
 ==============================================================================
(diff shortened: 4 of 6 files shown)


CMakeLists.txt
```diff
@@ -520,6 +520,7 @@ endif ()
 # to have priority over any pre-existing/system-wide headers need to appear early in
 # compiler search path.
 include_directories(BEFORE ${broker_includes})
+include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR}/auxil/filesystem/include)
```
bbannier, line 523: Just wondering, there's certainly a lot of precedence here to use `include_directories`, but that's really the most global hammer one could use. Since [this library provides a target one could depend on](https://github.com/gulrak/filesystem/blob/cd6805e94dd5d6346be1b75a54cdc27787319dd2/CMakeLists.txt#L39) a better approach might be to just add it to targets which depend on it instead.

awelzel: Thanks for these pointers. I'm not super familiar with cmake (and yes, just copied the approach from the other submodules).

I looked a bit and my high-level understanding is to use `target_link_libraries()`. Outside of plugins, I don't think there are separate targets, and for static/builtin plugins target_link_libraries() doesn't seem to be used and I failed to make it work right.

Though for static/builtin plugins, may also be okay to do the `include_directories` within their local CMakeLists.txt instead of having a global one. Then, for util.h/util.cc (as a candidate using ghc::filesystem), that could be transitioned to a library via `add_library()`.

So for this PR maybe just remove the top-level include_directories and figure out the users when needed? Other thoughts?

rsmmr: Actually for consistency I'd just go with the same style as for the other libraries for now. We'll eventually make this all nicer / more modern with a larger CMake overhaul.

rsmmr (commented):
Let's pull this into the Zeek namespace as well so that people can use it through `zeek::filesystem` (instead of `ghc::...`). That will make the transition easier once we switch to using `std`. Can put that `into src/util.h.`

And then I would suggest we add some simple unit test as well, just confirming some basic usage of the library to make sure it all builds ok.

awelzel:
> Let's pull this into the Zeek namespace as well so that people can use it through `zeek::filesystem` (instead of `ghc::...`). That will make the transition easier once we switch to using `std`. Can put that `into src/util.h.`
> 
> And then I would suggest we add some simple unit test as well, just confirming some basic usage of the library to make sure it all builds ok.

I've force-pushed this out. I hope I found the right places where to add things.

bbannier:
> /opt/zeek-release/include/zeek/util.h:78:10: fatal error: ghc/filesystem.hpp: No such file or directory
>      78 | #include <ghc/filesystem.hpp>
>         |          ^~~~~~~~~~~~~~~~~~~~

For some reason I cannot reproduce this locally, but for running tests from a build tree we add some special setup in the from-build-tree env script, https://github.com/zeek/zeek/blob/fe5f5bca57236a1eb0ac4b457d221f749c9479e2/CMakeLists.txt#L94-L107. The important variable here is `HILTI_CXX_INCLUDE_DIRS`, you could try adding the absolute path containing `ghc/` to that list. But then again, for some reason it seems that the test runner script doesn't even source these files; I am confused how any of this actually manages to run successfully, https://github.com/zeek/zeek/blob/fe5f5bca57236a1eb0ac4b457d221f749c9479e2/ci/test.sh#L53-L70

awelzel:
> For some reason I cannot reproduce this locally, but for running tests from a build tree we add some special setup in the from-build-tree env script, 
...
> . The important variable here is `HILTI_CXX_INCLUDE_DIRS`, you could try adding the absolute path containing `ghc/` to that list.

It's failing when using an installed tree, too.

Example, based on one of the tests that fail showing the include paths used by spicyz doesn't reference/include anything around `3rdparty`.

```
$ spicyz -D jit ./test.spicy  -o test.hlto
...
[debug/jit] [job 2] /usr/bin/c++ -c -I/opt/zeek-release/include -fPIC -std=c++17 -g -O3 -DNDEBUG -fvisibility=hidden -I/opt/zeek-release/include -Wall -Wno-unused -Wno-register -Werror=vla -Wno-vla -I /home/awelzel/corelight-oss/zeek/build/src/builtin-plugins/spicy-plugin/include -I /opt/zeek-release/include -o /tmp/user/1000/test_33c027bb49bcf936-e96d8a86fa7219.o /tmp/user/1000/test_33c027bb49bcf936-e96d8a86fa7219.cc
...
In file included from /opt/zeek-release/include/zeek/Tag.h:11,
                 from /opt/zeek-release/include/zeek/packet_analysis/Analyzer.h:6,
                 from /home/awelzel/corelight-oss/zeek/build/src/builtin-plugins/spicy-plugin/include/zeek-spicy/zeek-compat.h:39,
                 from /home/awelzel/corelight-oss/zeek/build/src/builtin-plugins/spicy-plugin/include/zeek-spicy/cookie.h:19,
                 from /home/awelzel/corelight-oss/zeek/build/src/builtin-plugins/spicy-plugin/include/zeek-spicy/runtime-support.h:22,
                 from /tmp/user/1000/test_33c027bb49bcf936-e96d8a86fa7219.cc:8:
/opt/zeek-release/include/zeek/util.h:78:10: fatal error: ghc/filesystem.hpp: No such file or directory
   78 | #include <ghc/filesystem.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~
...
```

It can be fixed by explicity setting `HILTI_CXX_INCLUDE_DIRS`, but don't think we should ask that of users.
```
$ HILTI_CXX_INCLUDE_DIRS=/opt/zeek-release/include/zeek/3rdparty/filesystem/include/ spicyz -D jit ./test.spicy  -o test.hlto
...
[debug/jit] [job 4] exited with code 0
```

awelzel:
I have pushed a commit that brute-force puts symlinks for`ghc` into `<install-prefix>/include` and `build/` so that external plugins and spicyz work from the installation and build tree. Doesn't feel great, not sure it's horrible. Very open to feedback.

> But then again, for some reason it seems that the test runner script doesn't even source these files; I am confused how any of this actually manages to run successfully, 

@bbannier  - think it is this in btest.cfg that makes it work:
https://github.com/zeek/zeek/blob/ee88aa3acbe541f679b202687f00f0aa242e77a3/testing/btest/btest.cfg#L35-L37

Maybe worth ditching the zeek-path-dev.sh pieces if not used?

bbannier:
> I have pushed a commit that brute-force puts symlinks forghc into <install-prefix>/include and build/ so that external plugins and spicyz work from the installation and build tree. Doesn't feel great, not sure it's horrible. Very open to feedback.o

Since you have figured out how the test used to pass with `hilti-cxx-include-dirs` (thanks for that, I actually forgot that I added that 🤦) you could alternatively update both the env scripts and `hilti-cxx-include-dirs` to be consistent with how other external headers are handled. If you decide to use the symlink approach, make sure to validate that the symlink does not get installed (via naive copying) as it would point to the source tree and would e.g., break binary packages.

> Maybe worth ditching the zeek-path-dev.sh pieces if not used?

I'd say to keep them both. `hilti-cxx-include-dirs` makes sure that BTests pass, even if the env scripts are not used; the env scripts on the other hand are used by devs who work in the build tree and might e.g., expect to be able to launch `zeek` with Spicy sources like they can with an installed Zeek.

awelzel:
> If you decide to use the symlink approach, make sure to validate that the symlink does not get installed (via naive copying) as it would point to the source tree and would e.g., break binary packages.

I think I got this part right, but I'm still very unhappy about it all. The other thing that comes to mind is to make the following work within the build tree and installation tree:
```
#include "zeek/3rdparty/ghc/filesystem.hpp"
```

That should be better as it avoids the symlinks in the installation tree and doesn't put ghc into the top-level include-dir.

awelzel:
@rsmmr / @bbannier - I pushed a version that I'm now happier with and removed all of the intermediate commits. Maybe this can move forward now?

bbannier (approved):
Neat! This doesn't feel like a hack anymore, matches existing code elsewhere, and should be not too hard to maintain going forward. This looks nice to me.

rsmmr:
Agree, this is a nice solution (and something we could move RapidJSON over too).

rsmmr (approved):

PR #2230, opened 2022-06-30
Bump Spicy and spicy-plugin to their latest release.

bbannier:


timwoj:
Should this be pulled over into release/5.0 as well?

bbannier:
> Should this be pulled over into release/5.0 as well?

@timwoj, yes please, this needs to be in `release/5.0` and part of zeek-5.0.

PR #2232, opened 2022-06-30
Fix RAND_MAX check on FreeBSD 12

timwoj:


timwoj:
@0xxon I merged https://github.com/zeek/zeek/pull/2224 earlier not realizing that FreeBSD 12 builds are broken by it. Does this look reasonable as a fix?

bbannier (commented):
I was wondering why this assertion was here in the first place, and believe it was probably added to ensure that both `uint32_t` halves of the `uint64_t` calculated here in below loop are distributed uniformly; if `RAND_MAX` is less than the maximum `uint32_t` this is not true anymore.

It looks like the value of `RAND_MAX` is also used implicitly [here](https://github.com/zeek/zeek/blob/cf952200287dc673424202faccb17b1a5c651eaa/src/util.cc#L500=) which then also causes `prng` and `random_number` to return values `<=RAND_MAX`.

Another issue is that differing `RAND_MAX` values would not make `prng` and its users platform-independent anymore like advertised.

TBH, I think this might need a better fix which requires more thinking. You might still want to merge this to unblock CI.

0xxon:
Looks reasonable for now - but yes, we should probably fix this differently :/

0xxon:
Thinking about it - this function is in `zeek::detail` and not used.

I move to just delete it completely.

timwoj:
@ckreibich confirmed this method isn't used by any packages in the package repo, so I'll go ahead and push this.

PR #2234, opened 2022-07-01
Handle trailing delimeters for sets/vectors in config framework

### LINKED ISSUE #636: Config framework doesn't handle trailing commas in sets
A trailing comma in the config file prevents the entire line from being loaded:

    Site::local_nets	192.168.0.0/16,10.0.0.0/8,172.16.0.0/16,

triggers

    warning: bro.cfg/Input::READER_CONFIG: Could not convert line 'Site::local_nets	192.168.0.0/16,10.0.0.0/8,172.16.0.0/16,' to value. Ignoring line.

timwoj:
Fixes #636 

This PR also includes a rework of the config framework to get rid of one more use of regex.h, and some string utilities that we're using in zeek-agent-v2 that were useful for the primary change.


src/util.h
```diff
@@ -565,5 +565,92 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
...
+	return split(std::string(s), std::string(delim));
+	}
+
+inline std::vector<std::wstring> split(const wchar_t* s, const wchar_t* delim)
+	{
+	return split(std::wstring(s), std::wstring(delim));
+	}
+
+/**
+ * Returns a string view with all trailing characters of a given set removed.
```
awelzel, line 620: The usage of "string view" in the comment made me look for `std::string_view`, but seems a string copy is actually returned. There may be nothing to fix, just wanted to point it out.

timwoj: Agreed that that's a bit confusing. There's nothing that says those functions can't both take and return `string_view`s, but the comment should probably be changed regardless.

testing/btest/Baseline/scripts.base.frameworks.input.sqlite.types/out
```diff
@@ -12,5 +12,5 @@ BB
 
 }, vc=[10, 20, 30], vs=[], vn=<uninitialized>]
 0
-1
+0
```
0xxon, line 15: This is incorrect now. In the input data, `p$vs` is a vector of string that contains one zero-length element. The vector thus has length 1.

(Sorry - I did not check what causes this change, I was just curious about the baseline change).

Some of the changes in setspecialcases also look suspicious - e.g. where s=[, , , , ,] is replaced with an empty set. I did not verify that - but I assume it is the same problem.

Note that you probably have to handle the case of a string set like "a," differently in the config framework than in the general ascii reader. In the config reader you probably just want to handle that like a trailing comma. In the ascii reader you probably want to take the strict interpretation, which is "two string elements, one being an "a", one being an empty string".

0xxon (changes requested):
See inline comment - this changes some test outputs in ways that, in my opinion, are not correct.

testing/btest/Baseline/scripts.base.frameworks.input.sqlite.types/out
timwoj: Ok, completely reworked this to just change the regex in the config framework (which I swear I tried earlier). The rest of the changes were reverted.

0xxon: One question - looking at the PR, does this now accept something like

````
a,b,c,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```

as meaning `a,b,c`?

timwoj: That's correct.

0xxon: Ok, next question - do we really want to allow that? I mean - I get one comma at the end, but that seems a bit excessive.
Though, to be fair, I am not sure that I really see a lot of drawbacks...

0xxon: Fair - thinking about it, I think I am fine with both alternatives. I am honestly not sure which I prefer anymore.

PR #2244, opened 2022-07-07
Avoid calling safe_realloc unnecessarily from ODesc::Grow()

timwoj:
This is a silly bug in `ODesc::Grow()`, where we call `safe_realloc` even if the buffer size doesn't change. This theoretically shouldn't cause a `realloc`, but on some extremely pathological cases I've seen significant performance improvements by avoiding calling it (think ~50x reductions in time spent).

On more normal-looking traffic, it's looking like up to 2% improvement in CPU used (tcpreplay of data-center-like traffic at 500mbps for 5 minutes) and up to 2.5% improvement in pcap read times (pcap file with a very large number of http sessions).

### DIFF
--- a/src/Desc.cc
+++ b/src/Desc.cc
@@ -373,10 +373,15 @@ void ODesc::AddBytesRaw(const void* bytes, unsigned int n)
 
 void ODesc::Grow(unsigned int n)
 	{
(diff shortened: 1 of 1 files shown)


vpax:
O + M + G.

PR #2247, opened 2022-07-08
Fix usage of broker::convert(double,timespan)

timwoj:
I'm pretty sure this is correct. `broker::convert(double,timespan)` simply calls [`std::chrono::duration_cast`](https://en.cppreference.com/w/cpp/chrono/duration/duration_cast) which doesn't have a failure case. In that case, we don't need to check anything anyways, hence why that version of `convert` returns `void` now.

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -2050,9 +2050,9 @@ void Manager::PrepareForwarding(const std::string& name)
 
 void Manager::SetMetricsExportInterval(double value)
 	{
-	broker::timespan ts;
-	if ( broker::convert(value, ts) )
-		bstate->endpoint.metrics_exporter().set_interval(ts);
+	using dbl_seconds = std::chrono::duration<double>;
+	auto ts = std::chrono::duration_cast<broker::timespan>(dbl_seconds{value});
+	bstate->endpoint.metrics_exporter().set_interval(ts);
 	}
 
 void Manager::SetMetricsExportTopic(std::string value)


src/broker/Manager.cc
```diff
@@ -2051,8 +2051,8 @@ void Manager::PrepareForwarding(const std::string& name)
 void Manager::SetMetricsExportInterval(double value)
 	{
 	broker::timespan ts;
-	if ( broker::convert(value, ts) )
-		bstate->endpoint.metrics_exporter().set_interval(ts);
+	broker::convert(value, ts);
+	bstate->endpoint.metrics_exporter().set_interval(ts);
```
Neverlord, line 2055: This really shouldn't use the `convert`-API at all. It's a simple conversion between chrono types:

```
using dbl_seconds = std::chrono::duration<double>;
auto ts = std::chrono::duration_cast<broker::timespan>(dbl_seconds{value});
```

rsmmr: > It's a simple conversion between chrono types:

(Seeing "simple" and "chrono types" in the same sentence makes me chuckle. 😄  )

rsmmr: @Neverlord one question though: understood that it should be considered internal, but I thought you had said that both signatures would continue to work in case somebody is using it?

Neverlord: The upstream change was altering the result type from `bool` to `void` since this conversion cannot fail. C++ can't overload on return types, so this change cannot be done 'backwards compatible'.

However, the `convert` functions aren't meant to be called directly. I'm probably the one that had put that `convert` here, because I'm used to working with "Broker internals" and sometimes use the wrong level of abstraction as a result. Instead, this could should've called `broker::to<broker::timespan>(value)` (the `convert` functions basically exist as implementation detail for `broker::to`). This would have continued to work since `broker::to` does properly support both signatures.

In this particular case, though, there's no reason to ask Broker at all for a conversion since we're dealing with standard types. Again, probably my fault for just using Broker internals as convenience out of habit.

timwoj: Alright, pushed @Neverlord's suggestion

ckreibich (approved):

PR #2248, opened 2022-07-08
Publish container images to ECR in addition to docker.io.

bbannier:


bbannier:
This needs more work, at least
- [ ] if we prefer ECR going forward update docs
- [ ] mirror existing images from docker.io to ECR

ckreibich:
I've never locally dealt with the ECR so I'll need to do some twiddling to see how I enable access for myself. I'll follow up when I've done that. I agree with your suggestions re docs and syncing, Benjamin.

It would also be good to tweak the CI setup so that it actually pushes to ECR even for the PR, so we're sure the current secrets work before we merge it.

ckreibich:
Sorry to let this sit for so long @bbannier. Let's just get it in and see how it works out in practice. I can confirm the secrets are in place. We can mirror the older images over separately. I think it'll be more important to see whether we can really use the ECR as a drop-in replacement.

ckreibich (approved):

awelzel:
Hmm - Spicy is pusing into `zeek/spicy` and `zeek/spicy-dev` rather than zeekurity. Should we align the Zeek actions with this even if it's inconsistent/non-transparent when switching between the registries?

https://github.com/zeek/spicy/blob/765924e34f1ecf939d240d9c52f928f4656ea489/.github/workflows/docker.yml#L82-L85

Personally like `zeek/zeek` more than `zeekurity`.

bbannier:
> Hmm - Spicy is pusing into `zeek/spicy` and `zeek/spicy-dev` rather than zeekurity. Should we align the Zeek actions with this even if it's inconsistent/non-transparent when switching between the registries?
> 
> https://github.com/zeek/spicy/blob/765924e34f1ecf939d240d9c52f928f4656ea489/.github/workflows/docker.yml#L82-L85
> 
> Personally like `zeek/zeek` more than `zeekurity`.

Actually it looks like CI is failing since the repo doesn't exist. We _need to_ push to `zeek` instead of `zeekurity`.

Issue #2217, opened 2022-06-27
Add table_values function


ynadji:
Given a table `t` of type `table[X] of T`, `table_values(t)` returns a `vector of T` of all the values in `t`. I would not expect the results to have a guaranteed order.

Issue #2216, opened 2022-06-27
Add table_keys function


ynadji:
Given a table `t` of type `table[T] of any`, `table_keys(t)` returns a `set[T]` of the keys of `t`.

Issue #2208, opened 2022-06-23
Ordered sets/tables


vpax:
We keep running into situations where it would be handy to have two different notions of ordering associated with sets and tables.

One type of ordering is "insertion order", meaning that the container remembers the order in which items were added.  This would then be the order in which elements are provided for `for` loops or when printing/logging (other than per the next item).

The other type of ordering is "canonical ordering", meaning that if you print/log the container, the elements always come out in a given order.  One could imagine doing this by allowing the script writer to specify an associated comparison function (for sorting); or an associated canonicalizer (which might sort but might do something else); and/or having standard canonical formats for all the easy cases (strings are lexical, numerics are by the usual numerical ordering, IP addresses by conversion-to-integer-equivalent).

For canonical ordering, it's not clear to me whether that needs to be the `for` loop ordering - I think it's probably okay that it isn't, and that would also allow the use of both "insertion ordering" (which _would_ affect loop ordering) and "canonical ordering" (which would override it when printing/logging).

timwoj:
We currently have `ORDERED` and `UNORDERED` modes for Dictionary objects. `ORDERED` mode is the "insertion order" version that you're describing above. We don't modify the table elements for it, but we keep the keys in a separate vector in the order they were inserted. When we do retrievals based on indexes (and for looping) we actually do the lookup into the ordered list and go off that ordering instead.

It wouldn't be too much effort to add a canonical ordering mode that does something similar, but keeps the ordered vector sorted based on a predicate method provided to the Dictionary object when it's created.

timwoj:
Another question here that I just thought of (as I'm looking at another Dictionary issue) is how this would be controlled via script-land. Most likely the easiest way is to add an attribute that you can pass when creating a `table` or `set`. That attribute would get passed down to the `TableVal` constructor to set the ordering on the internal `Dictionary`.

PR #2252, opened 2022-07-08
Add table_keys and table_values functions

AmazingPP:
Implement #2216 #2217


vpax:
Looks on track, but seems some thought is required for `table_keys` when a table has multiple indexes rather than just one.  Also, looks like `table_keys` won't work on `set` arguments?

AmazingPP:
> Looks on track, but seems some thought is required for `table_keys` when a table has multiple indexes rather than just one. Also, looks like `table_keys` won't work on `set` arguments?

Do you mean I need to add test cases like `table[port, string, bool] of string` for `table_keys` ?

I think `table_keys` should only for `table` type, put a `set` into `table_keys` is meaningless. because `table_keys` returns `set[T]`

vpax:
> Do you mean I need to add test cases like `table[port, string, bool] of string` for `table_keys` ?

Yes.

> I think `table_keys` should only for `table` type, put a `set` into `table_keys` is meaningless. because `table_keys` returns `set[T]`

I see your point.  I was thinking it would return `vector of T`, in which case it'd be meaningful.  OTOH, that works poorly for tables/sets with multiple indices.  (That said, see my next general comment.)

vpax:
I'm concerned about the lack of type-checking for BiFs returning `any` types.  At a minimum, `table_values` could return a `vector of any` rather than `any`, though that only helps a little.  Unfortunately, I don't believe `table_keys` can return `set[any]` as IIRC the latter isn't allowed by the type system.

In more general terms, Zeek needs a clean way of dealing with type variables (or something like C++ templates).  Then these functions could have signatures like `table_values(tbl: table[type T1] of type T2): vector of T2` and `table_keys(tbl: table[type T1] of type T2): set[T2]` so that we'd get proper type-checking.  Designing such a general mechanism is however a significant undertaking (and note that in these examples, `T1` might be a `list` of types, not just a single type), so not clear that we should hold up these BiFs waiting for that.

AmazingPP:
I see, Zeek lack of generic function feature is really troublesome. Is anyone working on this feature? 

Can we force the user to use `as` to specify the type when `any` is the function returns and return value is used  in the context? Just like `local foo = table_keys(t) as set[port]` and `bar(table_keys(t) as set[string, addr])`. I think this is a way to check when function returning `any` types. We can implement it at the Zeek script syntax parsing.

I noticed a BIF function `filter_subnet_table`, it uses `any` to return a `table`.

vpax (approved):
LGTM.  I played with it a bit to see whether the dynamic `any` typing could be abused to cause a crash, but the interpreter's dynamic typing is too pervasive (a good thing here).

timwoj:
@AmazingPP can you squash down at least that fixup commit? I'll merge it after.

src/zeek.bif
```diff
@@ -1185,6 +1186,58 @@ function clear_table%(v: any%): any
...
+		}
+
+	return vv;
+	%}
+
+## Gets all keys from a table.
+##
+## t: The :zeek:type:`table`
+##
+## Returns: A ``set of T`` of all the keys in t.
```
awelzel, line 1218: Is there a requirement or something preventing us from using a vector here as well? While keys don't have a deterministic order currently, in combination with #2208 the returned vector would have a defined order.

Returning a vector might be more efficient as well it skips building up a new Dict and all the underlying composite hash / index structures. And at the same time, it seems the operations against the returned set of `table_keys()` (`| |`, `in`, `!in`, ?) could be done directly against the table?

@ynadji - your tickets asked for `table_keys()` to return `set[T]` - was there a specific motivation?

AmazingPP: > Is there a requirement or something preventing us from using a vector here as well? 

I think that if tables/sets with multiple indices, using `vector` will make the situation worse. 

- if returning a `vector of list`. The `list` type does not seem to be allowed to be used in Zeek scripts.
- if returning a `vector of vector of any`. I'm not sure whether the `vector` is allowed to store different types of data, But if I am the user of this function, I will be confused.

awelzel: Thanks. I guess vector elements can't have "index type".

Taking a step back, I'm still wondering what operations/use-cases the `set` returned by `table_keys()` allows that the table itself does not. But there are utilities/bifs taking a `set` and that `set` may now (easily) be the keys of a table.

ynadji: Yeah, it's mostly a helper/syntactic sugar. My immediate use-case was precisely what you're describing: I had a function that took a `set` and the `set` I wanted was of the keys of a `table`. Any use case that uses a `table` for data aggregation/bookkeeping but at the end of the day only needs the keys or the values is likely to use these functions. Furthermore, this is provided by a lot of languages that provide and associated array data structure, so programmers new to Zeek are likely to expect it to be present.

PR #2253, opened 2022-07-09
Tweak to signal-masking during Zeek startup

### LINKED ISSUE #2239: Can't exit zeek with ctrl+c when reading scripts from stdin
On Debian 11, with Zeek 5.0.0, running as follows doesn't exit Zeek via `ctrl+c`:
```
$ zeek
^C
^C^C^C^C^C^C^C^C^C
^C
^\
Quit (core dumped)
```

This worked fine with 4.2.1:
```
$ zeek --version
zeek version 4.2.2
$ zeek
^C
$
```

@bbannier  says it doesn't happen for him on osx, but does happen via docker, too.
```
$ docker run --rm -it zeekurity/zeek:5.0.0 
root@9c30f7ce9dbe:/# zeek
^C^C^C^C

^C^C^C^C^C^C^C
^\Quit (core dumped)
```

The good part, `ctrl+c` does work when actually listening on an interface.
```
root@9c30f7ce9dbe:/# zeek -i lo
listening on lo

^Creceived termination signal
0 packets received on interface lo, 0 (0.00%) dropped, 0 (0.00%) not processed
```

ckreibich:
The signal-masking in place to prevent unintended signal handling in threads interfered with script parsing, which includes reading script content from stdin, which includes interactive Zeek sessions. This simpy pauses the masking during parsing, see comments.

I'm also adding three btests:
- `core.sigterm-regular` and `core.sigterm-stdin` check basic signal-handling during regular processing and parsing. The latter test fails without the masking fix included.
- `core.load-stdin` simply tests script-loading from stdin, in isolation and in combination with other scripts. Maybe I missed it, but I couldn't find any tests for this.

Resolves #2239.


awelzel:
Could we unblock the signals before the unittests run as well? E.g. Ctrl+c'ing out of a `--test` run should work. That was actually the initial "usability" issue I stumbled across rather than the read-script-from-stdin.

ckreibich:
Now there Arne, thanks for the reminder. I first wondered why ctrl-c'ing during the unit tests is even an issue because they used to run so quickly. Then I saw that the tests now take just under a minute on my system. I created #2258 for that just now.

testing/btest/core/sigterm-regular.sh
```diff
@@ -0,0 +1,49 @@
...
+# See the sigterm-stdin.sh test for additional explanation of what's happening.
+#
+# Use a separate output file since btest-bg-wait replaces .stdout/.stderr:
+# @TEST-EXEC: bash %INPUT >output 2>&1
+
+# Helper to return the PID of the Zeek process launched in the background.
+zeek_pid() {
+    # The btest-bg-run .pid file contains the parent of the Zeek process
+    local ppid=$(cat zeek/.pid)
+    ps -xo pid,ppid,comm | awk "\$2 == \"$ppid\" && \$3 == \"zeek\" { print \$1 }"
```
rsmmr, line 13: Is this portable?

src/zeek-setup.cc
```diff
@@ -778,9 +779,15 @@ SetupResult setup(int argc, char** argv, Options* zopts)
 		if ( options.event_trace_file )
 			etm = make_unique<EventTraceMgr>(*options.event_trace_file);
 
+		// Parsing involves reading input files, including any input
+		// interactively provided by the user at the console. Temporarily
+		// undo the signal mask to allow ctrl-c. Ideally we'd do this only
+		// when we actually end up reading interactively from stdin.
```
rsmmr, line 786: Could we check `isatty(0)`?

testing/btest/core/sigterm-regular.sh
ckreibich: Portable ... enough? :-) I admit that my bar was CI. `awk` seems safe here, and for `ps` I googled various OS manpages.

src/zeek-setup.cc
ckreibich: Ah — yes, it's certainly possible to check we're reading from stdin when that happens. I was more concerned about whether we want to start adjusting the signal mask in more source files because the startup is already pretty complex. The determination that we're reading from stdin lives over in `load_files()` in scan.l.

testing/btest/core/sigterm-regular.sh
rsmmr: "googled various OS manpages" sounds good enough. :-)

rsmmr (approved):

PR #2254, opened 2022-07-09
Management framework: fix for an agent boot-time race condition

ckreibich:
During agent restarts it could happen that instructions arrive from the controller that lead to interaction with the Supervisor prior to established peering, causing events to get lost. The agents now ensure that the Supervisor peering is established before saying hello to controllers, closing the race.

I ran the testsuite in CI in my fork and it has passed a few hundred times in a row now, so hopefully it's solid.


scripts/policy/frameworks/management/agent/main.zeek
```diff
@@ -1049,6 +1054,14 @@ event Broker::peer_added(peer: Broker::EndpointInfo, msg: string)
...
 	if ( peer$network$address == sni$address && peer$network$bound_port == sni$bound_port )
+		g_supervisor_peered = T;
+
+	# If the Supervisor hasn't yet peered with us, don't broadcast
+	# notify_agent_hello. Doing so would exposes a race: we might receive
+	# commands from the controller that lead to requests to the Supervisor
+	# that it won't yet receive. It's easier to handle this here than to
+	# push the wait into all types of received commands.
+	if ( g_supervisor_peered == F )
 		return;
```
rsmmr, line 1069: I don't have overall flow well in my head, but just looking at this change in isolation I'm wondering about the change in behavior if I'm reading the diff right. Previously it would return immediately if it's the supervisor; now it always proceeds to rest of the handler in that case. That intentional?

rsmmr (approved):

ckreibich:
I hit another error in my fork (in `tests.deploy-error.sh`) that I'll investigate prior to merging this.

ckreibich:
Okay ... there's a roughly 1-in-300 change that the `deploy-error` test fails, and I haven't so far been able to make sense of it. I'm hoping that should this happen again we'll have more output, given the changes in this PR. Karma suggests it'll hit soon after this merge. :-P

PR #2255, opened 2022-07-09
Use the seed for Broker's ID in deterministic mode

Neverlord:
From @J-Gras (see https://github.com/zeek/broker/issues/253):

> For seeded Zeek runs (`zeek -G random.seed`), Node IDs will still vary as they seem to be based on process IDs. If this isn't too complicated or breaks other things, having deterministic node IDs in seeded runs would be handy for testing.

This PR uses the seed for generating Broker endpoint ID. Broker internally uses utilities from `<random>`, so as long as they behave the same on all platforms, the ID should be always the same when running with the same seed.

One note: when running a cluster, the broker IDs **must** differ for all nodes. So passing the same seed to two or more nodes in the cluster would break Broker.

@J-Gras is this change what you have been looking for? Mind testing it?

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -163,8 +163,21 @@ struct opt_mapping
 class BrokerState
 	{
 public:
+	static broker::endpoint_id make_id()
+		{
+		if ( ! util::detail::have_random_seed() )
+			{
+			return broker::endpoint_id::random();
+			}
+		else
+			{
+			auto useed = static_cast<unsigned>(util::detail::initial_seed());
+			return broker::endpoint_id::random(useed);
+			}
+		}
+
 	BrokerState(broker::configuration config, size_t congestion_queue_size)
-		: endpoint(std::move(config)),
+		: endpoint(std::move(config), make_id()),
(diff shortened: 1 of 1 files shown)


Neverlord:
@J-Gras seems like this is a no-show. All btests fail, because the Broker instances now have the same IDs and the handshakes won't succeed. So it seems like all Zeek test instances run in deterministic mode. Unless there's a straightforward way to provide a different seed to the Zeek instances that `btest` spins up, it's probably best to just close this PR as a dead end.

J-Gras:
My current solution just filters out the node IDs from the logs. So I am fine to close this as a dead end. One final thought: If the node name would be available at this point (e.g. in `config`) and we would assume the name to be unique, would it work to combine the Zeek seed and the node name to get a unique ID?

Neverlord:
As long as the name is unique, this should work. In this case, I would probably just hash the name and use that as seed for generating the endpoint ID.

Neverlord:
> would it work to combine the Zeek seed and the node name to get a unique ID?

Since they all seem to use the same seed (otherwise they wouldn't produce the same UUID), concatenating the seed to something probably doesn't add much. Whether assuming that the configured ID is unique - I cannot tell. :)

Issue #2192, opened 2022-06-15
`broker.ssl-mismatch` test is unstable on macOS


timwoj:
I haven't seen this fail too much on Cirrus, but it fails pretty consistently on my local machine. It hits a timeout because the zeek process never exits.

0xxon:
Same here; I think it just doesn't pass at all for me anymore.

timwoj:
This is fixed in the above merge.

PR #2257, opened 2022-07-10
Avoid race conditions in broker.ssl-mismatch test

Neverlord:
This PR adds a new `__peer_no_retry` BIF for Broker that's meant only for testing. Without this, some tests enter a retry loop that wastes time while running it and also introduces a race in tests (what happens first, the btest timeout of the loop completion?). I'm using it for a single test as of this PR, but there are more tests that could (or should) use the new function instead.

The `broker.ssl-mismatch` test itself was re-organized to make sure that the clients only try to connect after the listener is up and running to avoid a race where a client could observe an error if it tries to connect before the port has been opened.

Relates #2192. With this patch, I could no longer get the `broker.ssl-mismatch` test to fail locally (ran 100 times with no issues).

### DIFF
--- a/src/broker/Manager.h
+++ b/src/broker/Manager.h
@@ -144,6 +144,13 @@ public:
 	void Peer(const std::string& addr, uint16_t port, double retry = 10.0);
 
 	/**
+	 * Initiate a peering with a remote endpoint but tries only once.
+	 * @param addr an address to connect to, e.g. "localhost" or "127.0.0.1".
+	 * @param port the TCP port on which the remote side is listening.
+	 */
+	void PeerNoRetry(const std::string& addr, uint16_t port);
+
+	/**
 	 * Remove a remote peering.
 	 * @param addr the address used in bro_broker::Manager::Peer().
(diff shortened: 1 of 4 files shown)


src/broker/Manager.h
```diff
@@ -143,6 +143,13 @@ class Manager : public iosource::IOSource
 	 */
 	void Peer(const std::string& addr, uint16_t port, double retry = 10.0);
 
+	/**
+	 * Initiate a peering with a remote endpoint but tries only once.
```
timwoj, line 147: Is the intention for this method to only be used by tests? We probably should mention that if so.

timwoj (approved):
Looks good to me. I just had one question about a comment.

src/broker/Manager.h
0xxon: I am going to assume that the answer is "yes", and merge it with a comment added. If this is incorrect, we can remove the comment later.

(Getting annoyed about the test failures on my machine)

PR #2263, opened 2022-07-11
Run clang-format on all of our plugin test c++ files

timwoj:
This is a general cleanup thing I've had sitting around for a while now.


rsmmr (approved):

bbannier:
This patch seems to only reformat source, but not header files. We originally disabled formatting by using a per-directory override https://github.com/zeek/zeek/blob/master/testing/btest/plugins/.clang-format#L1.

With that file removed the files in its folder should be formatted as usual, i.e., `pre-commit run -a clang-format` should change all differently formatted files.

timwoj:
Good catch, thanks.

PR #2266, opened 2022-07-13
Fix tail -F semantics when want_record=F and add tests for the new features

leres:
While writing a test for the new "tail -F semantics" I found that the $want_record=F case was broken (errno 25). So instead of opening /dev/null when the input file is missing change READER_RAW to avoid I/O until it can be opened.

Add two tests, one for when the event handler is called with a record and one for when it's called with a string.


leres:
Honestly, opening `/dev/null` was the cowards way out back when I was not super familiar with the Raw.cc module. But when I found the case where the `fread()` in `Raw::GetLine()` would fail with errno 25 (Inappropriate ioctl for device) I decided it would be better to figure how to make it possible for the file to not be open. In the end the changes were minimal.

timwoj:
I added this to the list for 5.0.1.

leres:
Do you have a rough ETA on 5.0.1?

timwoj:
Not yet.

0xxon:
Hm. Are we sure we want the tail -F stuff still in 5.0.x? To me this feels like a pretty major change of functionality for a bugfix release, and larger than I would expect.

leres:
Is there really changed functionality? I would describe it as added functionality especially considering that updating the documentation was as simple as inserting this line:

> Zeek also checks to see if the file being followed has been renamed or rotated.  The file is closed and reopened when tail detects that the filename being read from has a new inode number.

I suppose there are error cases that are changed by the code:
1.  A typo in the input file name or missing input file will wait for the correct file instead of giving an error
2. Code that depends on zeek not seeing changes after a file is rotated (unlikely)

I've viewed this a bug since last May when I discovered two of my boxes had stopped blocking attackers because zeek had stopped watching a logfile because it had been rotated.

PR #2273, opened 2022-07-14
Add telemetry framework

awelzel:
* Add a telemetry framework
* Support collecting and working with metrics in script land
* Add `telemetry.log` and `telemetry_histogram.log` logs

External baseline update PRs:
https://github.com/zeek/zeek-testing/pull/8
https://github.com/zeek/zeek-testing-private/pull/6

PR for the documentation bits:
https://github.com/zeek/zeek-docs/pull/128


src/telemetry/telemetry.bif
```diff
@@ -548,3 +549,19 @@ function Telemetry::__dbl_histogram_sum%(val: opaque of dbl_histogram_metric%):
 		return zeek::make_intrusive<DoubleVal>(0.0);
 		}
 	%}
+
+# Fix types?
```
rsmmr, line 553: I don't fully recall, but I believe we indeed can't do this better than an `any_vec`.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
 	}
 
+// -- collect metric stuff -----------------------------------------------------
+
+const std::vector<std::string_view> Manager::Metric::label_names() const
+	{
+	std::vector<std::string_view> r;
+
+	// Labels are backed by strings, so string_view should be safe.
+	for ( auto const& l : broker::telemetry::label_names(family_) )
```
rsmmr, line 53:
```suggestion
	for ( const auto& l : broker::telemetry::label_names(family_) )
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
 	pimpl.swap(ptr);
 	}
 
+// -- collect metric stuff -----------------------------------------------------
+
+const std::vector<std::string_view> Manager::Metric::label_names() const
```
rsmmr, line 48:
```suggestion
 std::vector<std::string_view> Manager::Metric::label_names() const
```

No need for `const` restriction here, it's copied/moved anyways.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	for ( auto const& l : broker::telemetry::label_names(family_) )
+		r.push_back(l);
+
+	return r;
+	}
+
+Manager::Metric::Metric(Manager::Metric::MetricType metric_type,
+                        const broker::telemetry::metric_family_hdl* family,
+                        std::optional<double> value, std::optional<int64_t> count_value,
+                        std::vector<std::string_view> labels)
```
rsmmr, line 62: Any reason not to use the same order for ctor arguments as in class and initialization? Generally, less confusing if all is the same order.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	: metric_type_(metric_type), family_(family), labels_(labels), value_(value),
```
rsmmr, line 63:
```suggestion
	: metric_type_(metric_type), family_(family), labels_(std::move(labels)), value_(value),
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	r->Assign(rt_off_prefix, make_intrusive<zeek::StringVal>(prefix()));
+	r->Assign(rt_off_name, make_intrusive<zeek::StringVal>(name()));
+	r->Assign(rt_off_helptext, make_intrusive<zeek::StringVal>(helptext()));
+	r->Assign(rt_off_unit, make_intrusive<zeek::StringVal>(unit()));
+	r->Assign(rt_off_is_total, val_mgr->Bool(is_sum()));
+
+	auto label_names_vec = make_intrusive<zeek::VectorVal>(svt);
+	int i = 0;
+	for ( auto const& l : label_names() )
+		label_names_vec->Assign(i++, make_intrusive<StringVal>(l));
```
rsmmr, line 92:
```suggestion
	for ( const auto& l : label_names() )
		label_names_vec->Append(make_intrusive<StringVal>(l));
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	r->Assign(rt_off_labels, label_names_vec);
+
+	auto et = zeek::id::find_type<zeek::EnumType>("Telemetry::MetricType");
+	zeek_int_t metric_type_int = -1;
+
+	// This is mapping internal enum values for the types to the
+	// right EnumVal. TODO: There might be a better way.
+	switch ( metric_type_ )
```
rsmmr, line 100: I don't think you can get around the switch, but you can declare an enum type in a bif and then use those constants instead of the `Lookups`, then this should compress quite a bit.  See, e.g., `BrokerProtocol` in `src/broker/comm.bif`

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	{
+	static auto svt = zeek::id::find_type<zeek::VectorType>("string_vec");
+	static auto rt = zeek::id::find_type<zeek::RecordType>("Telemetry::Metric");
+	static auto rt_off_opts = rt->FieldOffset("opts");
+	static auto rt_off_label_values = rt->FieldOffset("label_values");
+	static auto rt_off_value = rt->FieldOffset("value");
+	static auto rt_off_count_value = rt->FieldOffset("count_value");
+
+	auto r = make_intrusive<zeek::RecordVal>(rt);
+	// TODO: Cache this somewhere. Maybe do not set it here at all.
```
rsmmr, line 144: Seems the manager can maintain a cache of the opts records. Either create them here on demand on 1st use, or pre-compute elsewhere.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	auto opts_record = get_metric_opts_record();
+	r->Assign(rt_off_opts, opts_record);
+
+	if ( value_ )
+		r->Assign(rt_off_value, make_intrusive<zeek::DoubleVal>(*value_));
+	if ( count_value_ )
+		r->Assign(rt_off_count_value, val_mgr->Count(*count_value_));
```
rsmmr, line 151:
```suggestion
	if ( value_ )
		r->Assign(rt_off_value, make_intrusive<zeek::DoubleVal>(*value_));
		
	if ( count_value_ )
		r->Assign(rt_off_count_value, val_mgr->Count(*count_value_));
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+
+	auto label_values_vec = make_intrusive<zeek::VectorVal>(svt);
+	int i = 0;
+	for ( auto const& l : labels_ )
+		label_values_vec->Assign(i++, make_intrusive<StringVal>(l));
```
rsmmr, line 156: Per above.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	r->Assign(rt_off_label_values, label_values_vec);
+
+	return r;
+	}
+
+namespace bt = broker::telemetry;
```
rsmmr, line 162: Better for readability to avoid the shortcut.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+			if ( fnmatch(name_pattern_.c_str(), name.c_str(), 0) == FNM_NOMATCH )
+				return false;
+			}
+
+		return true;
+		}
+
+	std::vector<Manager::Metric> metrics_;
+	std::string prefix_pattern_;
+	std::string name_pattern_;
```
rsmmr, line 259: Don't use trailing underscores.

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		label_values_vec->Assign(i++, make_intrusive<StringVal>(l));
+	r->Assign(rt_off_label_values, label_values_vec);
+
+	return r;
+	}
+
+namespace bt = broker::telemetry;
+
+/**
+ * A collector implementation fetching info and values (for counters and guages);
```
rsmmr, line 165:
```suggestion
 * A collector implementation fetching info and values (for counters and gauges);
```

(Saw this typo a few more times.)

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		: prefix_pattern_(prefix), name_pattern_(name)
+		{
+		}
+
+	void operator()(const bt::dbl_counter_family_hdl* family, const bt::dbl_counter_hdl* counter,
+	                const std::vector<std::string_view>& label_values)
+		{
+		if ( matches(upcast(family)) )
+			metrics_.push_back({MT::DOUBLE_COUNTER, bt::upcast(family), (double)bt::value(counter),
+			                    std::nullopt, label_values});
```
rsmmr, line 182: Style nit, I  *believe* this would work too:

```suggestion
			metrics_.emplace_back(MT::DOUBLE_COUNTER, bt::upcast(family), (double)bt::value(counter),
			                    std::nullopt, label_values);
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+public:
+	MetricsCollector(std::string_view prefix, std::string_view name)
+		: prefix_pattern_(prefix), name_pattern_(name)
+		{
+		}
+
+	void operator()(const bt::dbl_counter_family_hdl* family, const bt::dbl_counter_hdl* counter,
+	                const std::vector<std::string_view>& label_values)
+		{
+		if ( matches(upcast(family)) )
```
rsmmr, line 180: (I don't quite know what the `upcast` does, but I imagine it's part of the collector's API?)

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+			metrics_.push_back({MT::DOUBLE_COUNTER, bt::upcast(family), (double)bt::value(counter),
+			                    std::nullopt, label_values});
+		}
+
+	void operator()(const bt::int_counter_family_hdl* family, const bt::int_counter_hdl* counter,
+	                const std::vector<std::string_view>& label_values)
+		{
+		if ( matches(upcast(family)) )
+			metrics_.push_back({MT::INT_COUNTER, bt::upcast(family), (double)bt::value(counter),
```
rsmmr, line 189: Prefer new-style casts:

```suggestion
			metrics_.push_back({MT::INT_COUNTER, bt::upcast(family), static_cast<double>(bt::value(counter)),
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+private:
+	/**
+	 * Does the given family match the prefix, name patterns?
+	 */
+	bool matches(const bt::metric_family_hdl* family)
+		{
+		// Complications due to bt::prefix and bt::name returning
+		// a string_view which isn't guaranteed to be null-terminated.
+		bool have_prefix = ! prefix_pattern_.empty();
+		bool have_name = ! name_pattern_.empty();
```
rsmmr, line 238: Just move into the if-conditions?

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+
+		if ( have_prefix )
```
rsmmr, line 240: What supposed to the result of prefix/name are empty? Should the matching fail then (i.e. do we need `else` branches?)

src/telemetry/Manager.cc
```diff
@@ -38,6 +43,232 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		}
+
+	std::vector<Manager::Metric> metrics_;
+	std::string prefix_pattern_;
+	std::string name_pattern_;
+	};
+
+std::vector<Manager::Metric> Manager::CollectMetrics(std::string_view prefix, std::string_view name)
+	{
+	std::vector<Manager::Metric> metrics;
```
rsmmr, line 264: Not needed it looks like.

rsmmr (commented):
I've looked only at the C++ for now, left some comments there. Feedback is It's mostly small, nit-picky stuff on style and some local code improvements. But overall this is good, nice job finding your way around here. 

We should eventually add tests for the metrics collection and the globbing, but that can happen  in script-land too, wherever it's easiest.

src/telemetry/Manager.h
```diff
@@ -50,6 +57,73 @@ class Manager
...
 	virtual void InitPostScript();
 
+	/**
+	 * Instances for use in __list_metrics, returned by by CollectMetrics().
+	 *
+	 * This is using the broker::telemetry low-level types rather
+	 * than IntCounter, DblCounter from zeek::telemetry due to the
+	 * latter expecting non-const pointers and I didn't want to
+	 * cast them away. The former provide a powerful enough API, too.
+	 */
```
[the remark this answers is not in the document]
awelzel: The `telemetry::IntGauge` (as an example) can be constructed with a non-const `broker::telemetry::int_gauge_hdl*` (the class offers Inc/Dec that mutates the underlying guage). The collector API only provides the const versions of these pointers (in the current form at least). I'm not sure what could be changed with the classes, but given that all we'd be using is `Value() on those, seemed okay to directly use the `broker::telemetry::value()` helper directly.

https://github.com/zeek/zeek/blob/318c50f0802958d8da81b114b1ddc220acbdb448/src/telemetry/Gauge.h#L78-L84

I think with your suggestion of  compressing the "CollectedMetric" it may look different enough to see where and whether it's a concern.

src/telemetry/Manager.h
```diff
@@ -50,6 +57,73 @@ class Manager
 	 */
 	virtual void InitPostScript();
 
+	/**
+	 * Instances for use in __list_metrics, returned by by CollectMetrics().
```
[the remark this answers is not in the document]
awelzel: Thanks for both points. I like CollectedMetric much better. I think this is where I'll start iterating.

src/telemetry/Manager.h
```diff
@@ -50,6 +57,73 @@ class Manager
...
+		std::string_view prefix() const { return broker::telemetry::prefix(family_); }
+		std::string_view name() const { return broker::telemetry::name(family_); }
+		const std::vector<std::string_view> label_names() const;
+		std::string_view helptext() const { return broker::telemetry::helptext(family_); }
+		std::string_view unit() const { return broker::telemetry::unit(family_); }
+		bool is_sum() const { return broker::telemetry::is_sum(family_); }
+		};
+
+	/**
+	 * Collect all Counter and Guage metrics, glob filtered by prefix and name.
```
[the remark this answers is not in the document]
awelzel: There'll be a separate method that expands Histograms into individual Counters for use by script-land `Telemetry::list_histogram_metrics`. Skipped that in the first round.

src/telemetry/Manager.cc
awelzel: Uhm, certainly not intentional. Thanks.

src/telemetry/Manager.cc
awelzel: The intention was to have an empty string match everything. If that feels wrong can certainly change and use `"*"` defaults in script-land instead.

awelzel: (would also remove  the awkward `empty()` test).

src/telemetry/Manager.cc
awelzel: The whole method is gone after removing the accessors.

src/telemetry/Manager.cc
awelzel: I attempted this. While it didn't compress the Lookup calls are gone. But I may have missed something.

src/telemetry/Manager.cc
awelzel: It's part of broker::telemetry outside of the collector logic to go from a specific family handle to the metric_family_hdl base class:

https://github.com/zeek/broker/blob/master/src/telemetry/metric_family.cc#L25-L33

But I'm sure Dominik has other thoughts if this is actually "right".

src/telemetry/Manager.h
```diff
@@ -50,6 +57,73 @@ class Manager
...
+	 *
+	 * @param name
+	 * @param prefix
+	 *
+	 * @return std::vector<Metric>
```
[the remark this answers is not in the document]
awelzel: Thanks, these were ide generated and didn't remove them. I'll take inspiration from the existing comments.

src/telemetry/Manager.h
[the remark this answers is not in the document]
awelzel: I'm resolving this one for now after renaming and removing the accessors.

We could probably hide those helper structs completely if the `CollectMetrics()` interface returned RecordValPtrs directly, but I didn't do this as I'm not sure it'd be better.

src/telemetry/Manager.h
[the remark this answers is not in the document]
awelzel: I'm resolving this for now, but also pinging @Neverlord in the hope he can look over what's there currently.

awelzel:
> We should eventually add tests for the metrics collection and the globbing, but that can happen in script-land too, wherever it's easiest.

I've added a few classic btest style exercising parts of the API.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,505 @@
...
+##! somewhat lower-level telemetry.bif functions.
+
+module Telemetry;
+
+export {
+	type labels_table: table[string] of string;
+	type labels_vector: vector of string;
+
+	## Default bounds/buckets for histograms if no bounds are explicitly provided.
+	const default_histogram_bounds = vector(0.01, 0.1, 1.0, 10.0, 30.0, 60.0) &redef;
```
rsmmr, line 11: Wondering if a default is useful given that semantics of bounds are very metric-specific?

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,505 @@
...
+	                                  label_values: labels_vector &default=vector(),
+	                                  amount: double &default=1.0): bool;
+	## Decrement a Gauge by the given amount through a GaugeFamily.
+	global gauge_family_dec: function(g: GaugeFamily,
+	                                  label_values: labels_vector &default=vector(),
+	                                  amount: double &default=1.0): bool;
+	## Set a Gauge by the given amount through a GaugeFamily.
+	global gauge_family_set: function(g: GaugeFamily,
+	                                  value: double,
+	                                  label_values: labels_vector &default=vector()): bool;
```
rsmmr, line 127: Some API functions receive the labels before the value, some after. Would be good to make the order consistent.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,505 @@
...
+
+	# API for listing/iterating metrics follows.
+
+	## Type of elements returned by the :zeek:see:`Telemetry::collect_metrics` function.
+	type Metric: record {
+		## The MetricOpts describing this histogram
+		opts: MetricOpts;
+
+		## The labels associated with this Metric.
+		label_values: vector of string;
```
rsmmr, line 197: Just name `labels`? Should be clear the it's the values here.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,505 @@
...
+	global gauge_family_dec: function(g: GaugeFamily,
+	                                  label_values: labels_vector &default=vector(),
+	                                  amount: double &default=1.0): bool;
+	## Set a Gauge by the given amount through a GaugeFamily.
+	global gauge_family_set: function(g: GaugeFamily,
+	                                  value: double,
+	                                  label_values: labels_vector &default=vector()): bool;
+
+	## Increment a Gauge by the given amount through a GaugeFamily.
+	global gauge_family_inc_t: function(g: GaugeFamily,
```
rsmmr, line 130: As discussed offline, let's remove the table versions.

rsmmr (commented):
I did a quick high-level pass over the Zeek scripts, and didn't notice much beyond what we already discussed; it's looking good. So go ahead and tie things up, and then I'll review in more detail once it's out of draft.

scripts/base/frameworks/telemetry/main.zeek
awelzel: These were purposefully "inconsistent" so that `&default`'s work out.

The `_inc` versions have `label_values` as the second argument and `amount` as third because it's more likely that `label_values` is to be varied/set. In the `gauge_family_set()` case, the `value` field is required as there's no reasonable default, while the `label_values` may be left default (empty), that's why I moved the required one forward.

That said, consistent order is probably better. IMO `_inc()` should continue to be what it is and default to `amount=1.0` with labels being more likely to vary. For the `_set()` case that means `label_values` will become a required second parameter.

I'll make the change.

scripts/base/frameworks/telemetry/main.zeek
awelzel: Removed now. Other libraries have defaults, but also agree may be too metric specific.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
+##! Module for recording and querying metrics. This modules wraps
+##! the lower-level telemetry.bif functions.
```
rsmmr, line 2: Suggest to add a note here about these being accessible via Prometheus as well, easy to miss that link.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+
+module Telemetry;
+
+export {
+	type labels_table: table[string] of string;
+	type labels_vector: vector of string;
```
rsmmr, line 8: Add short comments here as well, rule of thumb is that everything in the export section should be documented.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+		prefix: string;
+
+		## The human-readable name of the metric.
+		name: string;
+
+		## The unit of the metric. Use the pseudo-unit "1" if this is a unit-less metric.
+		unit: string;
+
+		## Documentation for this metric.
+		helptext: string &default="Zeek Script Metric";
```
rsmmr, line 22:
```suggestion
		help_text: string &default="Zeek Script Metric";
```

Or maybe just `doc`?

rsmmr: ... and I think I'd leave this optional instead of a having a default (because the default doesn't seem very helpful on its own).

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+
+		## The labels names (also called dimensions) of the metric. When
```
rsmmr, line 24:
```suggestion
		## The label names (also called dimensions) of the metric. When
```

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+		## When creating a :zeek:see:`Telemetry::HistogramFamily`,
+		## describes the number and bounds of the individual buckets.
+		bounds: vector of double &optional;
+
+		## The same meaning as *bounds*, but as :zeek:type:`count`.
+		## Only set in the return value of
+		## :zeek:see:`Telemetry::collect_histogram_metrics`.
+		## for histograms when the underlying type is ``int64_t``,
+		## otherwise ignored.
+		count_bounds: vector of count &optional;
```
rsmmr, line 43: I'm wondering if this is worth is worth having  at all (not very strong either way, just wanted to note).

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+		__family: opaque of dbl_counter_metric_family;
+		__labels: vector of string;
+	};
+
+	## Type representing a counter metric with initialized label values.
+	##
+	## Counter metrics only ever go up and reset when the process
+	## restarts. Use :zeek:see:`Telemetry::counter_inc` or
+	## :zeek:see:`Telemetry::counter_set` to modify counters.
+	## An example for a counter is the number number of log writes
```
rsmmr, line 68:
```suggestion
	## An example for a counter is the number of log writes
```

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+
+	## Register a counter family.
+	global register_counter_family: function(opts: MetricOpts): CounterFamily;
+
+	## Get a :zeek:see:`Telemetry::Counter` instance given family and label values.
+	global counter_with: function(cf: CounterFamily,
+	                              label_values: labels_vector &default=vector()): Counter;
+
+	## Increment a :zeek:see:`Telemetry::Counter` by `amount`.
+	## Using a negative `amount` is an error.
```
rsmmr, line 83: Document the return value (also below).

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+	## Helper to set a :zeek:see:`Telemetry::Gauge` to the given `value`.
+	global gauge_set: function(g: Gauge, value: double): bool;
+
+	## Increment a :zeek:see:`Telemetry::Gauge` by the given `amount` through
+	## the :zeek:see:`Telemetry::GaugeFamily`.
+	## This is a short-cut for :zeek:see:`Telemetry::gauge_inc`.
+	## Using a negative amount is an error.
+	global gauge_family_inc: function(g: GaugeFamily,
+	                                  label_values: labels_vector &default=vector(),
+	                                  amount: double &default=1.0): bool;
```
rsmmr, line 153:
```suggestion
	                                  amount: double &default=1.0): bool;
	                                  
```

(also below, just for consistency).

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+
+	## Telemetry sync hook.
+	##
+	## This hook is invoked every :zeek:see:`Telemetry::sync_interval`
+	## for script writers to synchronize or mirror metrics with the
+	## telemetry subsystem. For example, when tracking table or value
+	## sizes with gauges, the value in question can be set on an actual
+	## :zeek:see:`Telemetry::Gauge` instance during execution of this hook.
+	##
+	## Implementations should be light-weight, this hook may be called
```
rsmmr, line 207:
```suggestion
	## Implementations should be lightweight, this hook may be called
```

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+	                                           name: string &default="*"): vector of HistogramMetric;
+}
+
+
+## Internal helper to create the labels table.
+function make_labels(keys: vector of string, values: labels_vector): labels_table
+	{
+	local labels: labels_table;
+	for ( i in keys )
+		labels[keys[i]] = values[i];
```
rsmmr, line 300:
```suggestion
		labels[keys[i]] = values[i];
		
```

also below (insert empty line after if-block).

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+		local f = Telemetry::__dbl_histogram_family(
+			opts$prefix,
+			opts$name,
+			opts$labels,
+			opts$bounds,
+			opts$helptext,
+			opts$unit,
+			opts?$is_total ? opts$is_total : F
+		);
+		return HistogramFamily($__family=f, $__labels=opts$labels);
```
rsmmr, line 444: Indentation is one level off here.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,498 @@
...
+
+# Fallback Histogram when there are issues with the labels.
+global error_histogram_hf = register_histogram_family([
+	$prefix="zeek",
+	$name="telemetry_histogram_usage_error",
+	$unit="1",
+	$helptext="This histogram is returned when label usage for histograms is wrong. Check reporter.log if non-zero.",
+	$bounds=vector(1.0)
+]);
+#
```
rsmmr, line 455:
```suggestion

```

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+	redef enum Log::ID += { LOG, LOG_HISTOGRAM };
+
+	type Info: record {
+		## Timestamp of reporting.
+		ts: time &log;
+
+		## Peer that generated this log.
+		peer: string &log;
+
+		## Contains the values "counter" or "gauge" depending on
```
rsmmr, line 18:
```suggestion
		## Contains the value "counter" or "gauge" depending on
```

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+##! using metrics accessible via the Telemetry module.
+
+@load base/frameworks/telemetry
+
+module Telemetry;
+
+export {
+	redef enum Log::ID += { LOG, LOG_HISTOGRAM };
+
+	type Info: record {
```
rsmmr, line 11: Add a short comment to the type as well.

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+		## Timestamp of reporting.
+		ts: time &log;
+
+		## Peer that generated this log.
+		peer: string &log;
+
+		## Contains the values "counter" or "gauge" depending on
+		## the underlying metric type.
```
rsmmr, line 19:
```suggestion
		## The underlying metric type.
```

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+	global log_policy_histogram: Log::PolicyHook;
+
+	## Event triggered for every record in the stream.
+	global log_telemetry: event(rec: Info);
+
+	## Event triggered for every record in the histogram stream.
+	global log_telemetry_histogram: event(rec: HistogramInfo);
+
+	## How often metrics are reported.
+	option log_interval = 60sec;
```
rsmmr, line 89: Suggest to move this to the top, that makes it's easier to spot option that can be changed.

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+
+	## Only metrics with prefixes in this set will be included in the
+	## telemetry.log and telemetry_histogram.log files by default.
```
rsmmr, line 92:
```suggestion
	## `telemetry.log` and `telemetry_histogram.log` files by default.
```

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+
+		if ( |log_prefixes| > 0 && m$opts$prefix !in log_prefixes )
+			next;
+
+		local rec: Info;
+		rec$ts = ts;
+		rec$peer = peer_description;
+		# Convert 'Telemetry::DOUBLE_COUNTER' to 'counter' by
+		# trimming everything before '_'
+		rec$metric_type = to_lower(gsub(cat(m$opts$metric_type), /^[^_]+_/, ""));
```
rsmmr, line 123: This looks like a pretty expensive operation. I'd do a switch instead just mapping labels directly to what we want them to be.

scripts/policy/frameworks/telemetry/log.zeek
```diff
@@ -0,0 +1,190 @@
...
+		rec$prefix = m$opts$prefix;
+		rec$name = m$opts$name;
+		if ( m$opts$unit != "1" )
+			rec$unit = m$opts$unit;
+		else
+			delete rec$unit;
```
rsmmr, line 129: Do we need this? Wouldn't the field be unset by default?

rsmmr (commented):
I made a pass over everything in `scripts/` and `testing/`. That's all great, left mostly tiny nit-picking comments again. One overall style note for the Zeekygen comments: We usually prefer listing parameters explicitly (see https://github.com/zeek/zeek/blob/master/scripts/zeekygen/example.zeek#L155). You are pretty good at referring to them inside the text already, so don't think it's worth changing it everywhere, but just keep in mind for the future.

rsmmr:
Should we include that `zeek_version_info` metric from the docs by default?

scripts/base/frameworks/telemetry/main.zeek
awelzel: The bifs use  `helptext`, but those are positional and better to have `_` it here.

> Or maybe just doc?
I'll stick with "help" due to the C++ API and broker using that throughout.

Hmm, I'll go for making it required (ping if you object). There's some prom tooling that will point out missing help text and it's not too cumbersome to add it and provides documentation, too. It' not necessarily saving on the verbosity, too.

scripts/base/frameworks/telemetry/main.zeek
awelzel: :thinking:

scripts/policy/frameworks/telemetry/log.zeek
awelzel: > Do we need this? Wouldn't the field be unset by default?

I switched to `Info()` construction rather than `local rec: Info;`. Not needed anymore and less confusing.

scripts/policy/frameworks/telemetry/log.zeek
awelzel: I'm not quite following this one as it's part of the line above, so leaving it.

awelzel:
> Should we include that `zeek_version_info` metric from the docs by default?

I think that could be nice to have as a default. I've added it into framework/main.zeek. It broke the external tests due to that being included in telemetry.log then and being non-deterministic... It's filtered out now via `Telemetry::log_policy`.

One thing that stands-out is that either broker or caf sorts the labels alphabetically somewhere. Specifically for the `zeek_version_info` metric it may read more natural if the labels would be kept in the order as was used to create the family. Maybe @Neverlord has thoughts about that.

```
# HELP zeek_version_info The Zeek version
# TYPE zeek_version_info gauge
zeek_version_info{beta="false",commit="289",debug="true",endpoint="zeek",major="5",minor="1",patch="0",version_number="50100",version_string="5.1.0-dev.289-debug"} 1.000000 1658936589580
```

src/telemetry/Manager.h
```diff
@@ -50,6 +58,101 @@ class Manager
...
+
+	struct CollectedHistogramMetric
+		{
+		CollectedHistogramMetric(MetricType metric_type,
+		                         const broker::telemetry::metric_family_hdl* family,
+		                         std::vector<std::string_view> label_values, int64_t count,
+		                         std::vector<double> values, double sum,
+		                         std::optional<std::vector<int64_t>> count_values,
+		                         std::optional<int64_t> count_sum, std::vector<double> bounds,
+		                         std::optional<std::vector<int64_t>> count_bounds)
```
Neverlord, line 107: I found it very hard to follow this code, as it's flattening a data structure and duplicates arguments for each type at the same time. This also makes it rather easy to construct a logically inconsistent object by mixing types.

Logically, a histogram is an ordered list of buckets. Why not preserve this structure?

Something along those lines:

```
template <class T>
struct Bucket {
    T upper_bound;
    int64_t count;
};

template<class T>
struct Histogram {
  T sum;
  std::vector<Bucket<T>> buckets;
};

using IntHistogram = std::vector<Histogram<int64_t>>;

using DblHistogram = std::vector<Histogram<int64_t>>;

CollectedHistogramMetric(const broker::telemetry::metric_family_hdl* family, std::vector<std::string_view> label_values, std::variant<IntHistogram, DblHistogram> data)
// ...
```

Instead of the `variant` argument, this could also have two constructors instead if you prefer that style. In any case, the `metric_type` argument should not be necessary. This is always a histogram and we should be able to figure out whether it's a `double` or `int64_t` histogram based on the arguments.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -0,0 +1,614 @@
...
+@load base/misc/version
+
+module Telemetry;
+
+export {
+	## Alias for a vector of label values.
+	type labels_vector: vector of string;
+
+	## Type that captures options used to create metrics.
+	type MetricOpts: record {
```
Neverlord, line 17: Maybe it's just me, but this name had me confused and I needed to look up the definition to figure out what it's actually about. So this is basically the definition of a metric, i.e., all the meta information without the actual values. So maybe `MetricDecl` or `MetricDef` would give a better intuition about the purpose of this type?

src/telemetry/Manager.h
```diff
@@ -50,6 +58,101 @@ class Manager
...
+		double sum;
+		std::optional<std::vector<int64_t>> count_values;
+		std::optional<int64_t> count_sum;
+
+		// The following two should be the same for every metric
+		// within a family, but the broker::telemetry API doesn't
+		// allow us to query bounds/buckets using the family, so
+		// for now get those from the instances.
+		std::vector<double> bounds;
+		std::optional<std::vector<int64_t>> count_bounds;
```
Neverlord, line 131: When using the `variant`-based approach from above, the members would boil down to:

```
		const broker::telemetry::metric_family_hdl* family;
		std::vector<std::string_view> label_values;
		std::variant<IntHistogram, DblHistogram> data;
```

Storing a `MetricType` is unnecessary (and even counterproductive) since we can look at the type of `data` to figure out which `MetricType` this correlates to.

src/telemetry/telemetry.bif
```diff
@@ -3,7 +3,17 @@
 
 module Telemetry;
 
+enum MetricType %{
```
Neverlord, line 6: Why is this duplicated here?

src/telemetry/telemetry.bif
```diff
@@ -548,3 +558,36 @@ function Telemetry::__dbl_histogram_sum%(val: opaque of dbl_histogram_metric%):
 		return zeek::make_intrusive<DoubleVal>(0.0);
 		}
 	%}
+
+
+function Telemetry::__collect_metrics%(prefix: string, name: string%): any_vec
+	%{
+	auto metrics = telemetry_mgr->CollectMetrics(sv(prefix), sv(name));
+
+	static auto metrics_vector_type = zeek::id::find_type<VectorType>("any_vec");
```
Neverlord, line 567: Just wondering, is this lookup really that slow that we need to cache it? Is this a common pattern in Zeek code? Putting a static in here basically forces the compiler to put a mutex and something like a `pthread_once` here, so it's definitely not free.

src/telemetry/telemetry.bif
```diff
@@ -548,3 +558,36 @@ function Telemetry::__dbl_histogram_sum%(val: opaque of dbl_histogram_metric%):
...
+	auto vec = zeek::make_intrusive<VectorVal>(metrics_vector_type);
+
+	for ( const auto& m : metrics )
+		{
+		auto rec = m.GetMetricRecord();
+		vec->Append(rec);
```
Neverlord, line 573: Is there a particular reason to stretch this out over multiple lines? Could just be:

```suggestion
		vec->Append(m.GetMetricRecord());
```

src/telemetry/telemetry.bif
```diff
@@ -548,3 +558,36 @@ function Telemetry::__dbl_histogram_sum%(val: opaque of dbl_histogram_metric%):
...
+		}
+
+	return vec;
+	%}
+
+function Telemetry::__collect_histogram_metrics%(prefix: string, name: string%): any_vec
+	%{
+	auto metrics = telemetry_mgr->CollectHistogramMetrics(sv(prefix), sv(name));
+
+	static auto metrics_vector_type = zeek::id::find_type<VectorType>("any_vec");
```
Neverlord, line 583: Same question here. By having `static` variables in two different functions, we ask the compiler for two complete sets of synchronization primitives to cache the same thing.

src/telemetry/telemetry.bif
```diff
@@ -548,3 +558,36 @@ function Telemetry::__dbl_histogram_sum%(val: opaque of dbl_histogram_metric%):
...
+	auto vec = zeek::make_intrusive<VectorVal>(metrics_vector_type);
+
+	for ( const auto& m : metrics )
+		{
+		auto rec = m.GetHistogramMetricRecord();
+		vec->Append(rec);
```
Neverlord, line 589: Same question as before, why give a name to a temporary?

```suggestion
		vec->Append(m.GetHistogramMetricRecord());
```

src/telemetry/Manager.h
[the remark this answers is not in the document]
Neverlord: I couldn't find that comment in the current version. Is this still an open point?

awelzel:
> > Should we include that `zeek_version_info` metric from the docs by default?
> 
> I think that could be nice to have as a default. I've added it into framework/main.zeek. It broke the external tests due to that being included in telemetry.log then and being non-deterministic... It's filtered out now via `Telemetry::log_policy`.

So, this makes the `plugins/hooks.zeek` test fail and contain a lot of version specific output in the baseline:
```
  +0.000000 | HookCallFunction Telemetry::__dbl_gauge_family(zeek, version_info, [version_number, major, minor, patch, commit, beta, debug, version_string], The Zeek version, 1, F)
  +0.000000 | HookCallFunction Telemetry::__dbl_gauge_inc(<no value description>, 1.0)
  +0.000000 | HookCallFunction Telemetry::__dbl_gauge_metric_get_or_add(<no value description>, {[version_number] = 50100,[debug] = true,[beta] = false,[patch] = 0,[commit] = 289,[major] = 5,[minor] = 1,[version_string] = 5.1.0-dev.289-debug})
  +0.000000 | HookCallFunction Telemetry::__dbl_gauge_value(<no value description>)
   0.000000 | HookCallFunction Telemetry::__dbl_histogram_family(zeek, telemetry_histogram_usage_error, [], [1.0], This histogram is returned when label usage for histograms is wrong. Check reporter.log if non-zero., 1, F)
  +0.000000 | HookCallFunction Telemetry::gauge_family_set([__family=<no value description>, __labels=[version_number, major, minor, patch, commit, beta, debug, version_string]], [50100, 5, 1, 0, 289, false, true, 5.1.0-dev.289-debug], 1.0)
  +0.000000 | HookCallFunction Telemetry::gauge_set([__metric=<no value description>], 1.0)
  +0.000000 | HookCallFunction Telemetry::gauge_with([__family=<no value description>, __labels=[version_number, major, minor, patch, commit, beta, debug, version_string]], [50100, 5, 1, 0, 289, false, true, 5.1.0-dev.289-debug])
  +0.000000 | HookCallFunction Telemetry::make_labels([version_number, major, minor, patch, commit, beta, debug, version_string], [50100, 5, 1, 0, 289, false, true, 5.1.0-dev.289-debug])
```

@rsmmr - if you have a hint how to avoid that? I guess we could create it from within C++, too, but :man_shrugging:

src/telemetry/Manager.cc
Neverlord: Yep, looks all right to me.

Usually in C++, the compiler performs upcasts for us. For example:

```
struct Base { virtual ~Base() { } };

struct Derived : Base { };

void foo(Base*);

int main() {
  Derived obj;
  foo(&obj);
}
```

Here, the compiler does the upcast from `Derived` to `Base` automatically when passing `obj` to `foo`.

However, the telemetry types are opaque. So to the compiler, a `dbl_counter_family_hdl` has no relation to a `metric_family_hdl` whatsoever. The *actual* types are of course related. The `upcast` functions allow us to go from the more derived type to a "base type" on the type-erased telemetry types. It's just an extra hoop to go through since the compiler no longer sees the type relations.

src/telemetry/Manager.h
```diff
@@ -50,6 +58,101 @@ class Manager
 	 */
 	virtual void InitPostScript();
 
+	/**
+	 * Instances returned by CollectMetrics().
+	 */
+	enum MetricType
```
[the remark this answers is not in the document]
timwoj: I'm always ok with going towards the modern-C++ way of doing things. We can definitely update the style convention with some guidance as well.

Neverlord:
> One thing that stands-out is that either broker or caf sorts the labels alphabetically somewhere. Specifically for the `zeek_version_info` metric it may read more natural if the labels would be kept in the order as was used to create the family. Maybe @Neverlord has thoughts about that.
> 
> ```
> # HELP zeek_version_info The Zeek version
> # TYPE zeek_version_info gauge
> zeek_version_info{beta="false",commit="289",debug="true",endpoint="zeek",major="5",minor="1",patch="0",version_number="50100",version_string="5.1.0-dev.289-debug"} 1.000000 1658936589580
> ```

They get sorted internally to allow comparing labels so that you get the same metric instance when passing the same labels (regardless of order). Prometheus/Grafana may also sort them internally, but I imagine you'd probably pull that version out for a Dashboard anyways.

src/telemetry/Manager.h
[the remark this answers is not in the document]
awelzel: Maybe: For gathering the data, this still uses the new `collect` and the `broker::telemetry` free functions rather than `zeek::telemetry::DblGauge` or `zeek::telemetry::DblCounter` types. These are what you'd use primarily in C++ for actual metrics tracking/updating, so they may not be the best fit.

Haven't seen all the comments, but if that didn't stand out to you, too much it may be okay to keep it resolved.

src/telemetry/telemetry.bif
awelzel: This is for making the enum `MetricType` enum available to Zeek script land, too.

I'll follow the example of the enum class you provided elsewhere (thanks!), but seem that has the same pattern of being replicated in a bif file:
https://github.com/zeek/zeek/blob/master/src/broker/comm.bif#L66

rsmmr:
> @rsmmr - if you have a hint how to avoid that? I guess we could create it from within C++, too, but 🤷‍♂️

No great idea, but I wouldn't object to hardcoding some sanitization into the hook test, like just replacing the telemetry function arguments with `(...)`,

src/telemetry/telemetry.bif
Neverlord: Duplicating code is rarely a good idea. I can't tell why https://github.com/zeek/zeek/blob/master/src/broker/comm.bif#L66 does this, but it looks very suspicious to me (I'll look into the Broker BIF and probably file a PR for that). As long as the enum type is in a header (and it is), the BIF should simply include the proper header and use it.

Neverlord: @rsmmr actually, https://github.com/zeek/zeek/blob/master/src/broker/comm.bif#L66 looks like it's from the recent WebSocket additions. Do you remember details why the code duplicates an enum?

src/telemetry/telemetry.bif
awelzel: Thanks for pointing this out. I wasn't aware of the thread safety / `-fno-threadsafe-statics` aspect here.

There are (older?) examples where a separate initialization function and static variables are used:
https://github.com/zeek/zeek/blob/master/src/file_analysis/File.cc#L41-L79

Newer code seems to use `static auto` directly in functions quite a bit:
https://github.com/zeek/zeek/blob/master/src/IP.cc#L682-L689
https://github.com/zeek/zeek/blob/master/src/zeek.bif#L2285-L2290

Looking at some generated code for a specific gcc compiler, it doesn't look like the mutex is taken everytime, just for the actual initialization. There's likely still some overhead to check whether initialization happened everytime, but not taking a lock heavy overhead: https://godbolt.org/z/Yzshch6eq

I'll keep it for now, but also up changing it to something else if there are more/other opinions.

src/telemetry/Manager.h
```diff
@@ -50,6 +58,101 @@ class Manager
...
+		{
+		DOUBLE_COUNTER,
+		INT_COUNTER,
+		DOUBLE_GAUGE,
+		INT_GAUGE,
+		DOUBLE_HISTOGRAM,
+		INT_HISTOGRAM,
+		};
+
+	struct CollectedMetric
```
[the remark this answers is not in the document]
awelzel: The twist is that `CollectedMetric` currently does hold histograms, they just don't have a value. The motivation was that script land can list metrics regardless of their type with a single call.

However, given we have a separate `collect_histogram_metrics` already and that we even create a separate log, maybe that was too ambitious. We can ask them to invoke `collect_metrics()` and `collect_histogram_metrics()` instead and have things simpler here, too.

(@rsmmr - ack / nack?)

src/telemetry/telemetry.bif
awelzel: See above.

scripts/base/frameworks/telemetry/main.zeek
awelzel: This opts part comes from here: https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Opts

In the few usage examples and tests constructed, the name doesn't actually appear as it's an anonymous record, so leaving it for now.

src/telemetry/Manager.h
```diff
@@ -50,6 +58,101 @@ class Manager
...
+		{
+
+		CollectedMetric(Manager::MetricType metric_type,
+		                const broker::telemetry::metric_family_hdl* family,
+		                std::vector<std::string_view> label_values, std::optional<double> value,
```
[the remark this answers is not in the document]
awelzel: For script land, we're treating everything as double, but also expose the `int` values as `count` if they exist. Further, histograms don't have a single value and were also included here, too.

But I think your suggestion makes it all better. Let me do the easy ones first and then I'll try a `std::variant` version that also skips histograms as mentioned here: https://github.com/zeek/zeek/pull/2273#discussion_r933053586

src/telemetry/telemetry.bif
rsmmr: Not sure what's duplicated here, the `enum` in the bif should be making this available as a script-level type. Is that happening elsewhere already?

src/telemetry/Manager.h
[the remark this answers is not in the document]
awelzel: @Neverlord - I've worked on your suggestions and switched to std::variant for both the `CollectedValueMetric` and the `HistogramMetric` (including adding explicit buckets).

I think it looks better, but there's are some awkward `std::holds_alternative` in the `AsMetricRecord` and `AsHistogramMetricRecord` render functions. This is mostly because a separate ambition to make the `int` values available to script land if we have them.

Please take another look. I'm going to resolve the other comments that went into the same general direction. Thanks!

src/telemetry/Manager.h
awelzel: Thanks for that!

src/telemetry/Manager.h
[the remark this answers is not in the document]
awelzel: > C-style enumerations are very brittle due to implicit conversions to int and they pollute their scope.
There was in fact a bug where the C-style enum was passed forward :face_exhaling:

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		{
+		auto opts_record = telemetry_mgr->GetMetricOptsRecord<int64_t>(metric_type, family);
+		r->Assign(opts_idx, opts_record);
+
+		// For int64, convert the value to double and also assign
+		// the count_value field in the Zeek record.
+		int64_t vvalue = std::get<int64_t>(value);
+		r->Assign(value_idx, make_intrusive<zeek::DoubleVal>(static_cast<double>(vvalue)));
+		r->Assign(count_value_idx, val_mgr->Count(vvalue));
+		}
```
Neverlord, line 156: Might be a good use case for `visit`, since you duplicate some of the logic:

```cpp
auto fn = [&](auto val) {
  using val_t = decltype(val);
  auto opts_record = telemetry_mgr->GetMetricOptsRecord<val_t>(metric_type, family);
  r->Assign(opts_idx, opts_record);
  if constexpr (std::is_same_v<val_t, double>) {
    r->Assign(value_idx, make_intrusive<zeek::DoubleVal>(val));
  } else {
    r->Assign(value_idx, make_intrusive<zeek::DoubleVal>(static_cast<double>(val)));
    r->Assign(count_value_idx, val_mgr->Count(val));
  }
};
std::visit(fn, value);
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+
+		r->Assign(values_idx, values_vec);
+		r->Assign(count_values_idx, count_values_vec);
+
+		r->Assign(sum_idx, zeek::make_intrusive<zeek::DoubleVal>(histogram_data.sum));
+		r->Assign(observations_idx, zeek::make_intrusive<zeek::DoubleVal>(observations));
+
+		r->Assign(count_sum_idx, val_mgr->Count(histogram_data.sum));
+		r->Assign(count_observations_idx, val_mgr->Count(observations));
+		}
```
Neverlord, line 231: Could also use `visit` (see comment above) to avoid code duplication.

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+			[&opts_record](const auto& histogram_data)
+			{
+				auto bounds_vec = make_intrusive<zeek::VectorVal>(double_vec_type);
+				for ( const auto& b : histogram_data.buckets )
+					bounds_vec->Append(zeek::make_intrusive<zeek::DoubleVal>(b.upper_bound));
+
+				opts_record->Assign(opts_rt_off_bounds, bounds_vec);
+
+				// If this is an int64_t histogram, also fill count_bounds
+				using T = std::decay_t<decltype(histogram_data)>;
```
Neverlord, line 261: It's very unusual to have a typedef named `T` since that's generally a template parameter. I think a more speaking name would avoid such confusion:

```suggestion
				using HistogramType = std::decay_t<decltype(histogram_data)>;
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		{
+		}
+
+	/**
+	 * @return true if the given family's prefix and name match, else false;
+	 */
+	bool operator()(const broker::telemetry::metric_family_hdl* family)
+		{
+		std::string prefix(broker::telemetry::prefix(family));
+		std::string name(broker::telemetry::name(family));
```
Neverlord, line 294: Using regular parens sometimes don't do what you might expect ("most vexing parse"). It's generally a good rule to avoid them:

```suggestion
		auto prefix = std::string{broker::telemetry::prefix(family)};
		auto name = std::string{broker::telemetry::name(family)};
```

(You also just change the paren types, but using `auto` consistently also makes sure variables are always initialized.)

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+
+		if ( fnmatch(prefix_pattern.c_str(), prefix.c_str(), 0 == FNM_NOMATCH) )
+			return false;
+
+		if ( fnmatch(name_pattern.c_str(), name.c_str(), 0) == FNM_NOMATCH )
+			return false;
+
+		return true;
```
Neverlord, line 302: 1. I believe you have misplaced `== FNM_NOMATCH` in the first `if`-block (should move out of the parens)
2. I think writing it like this would be easier to read (and less code):

```suggestion
		return fnmatch(prefix_pattern.c_str(), prefix.c_str(), 0) != FNM_NOMATCH
		        && fnmatch(name_pattern.c_str(), name.c_str(), 0) != FNM_NOMATCH;
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +53,449 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+			metrics.emplace_back(Manager::CollectedValueMetric::NewCounter(
+				broker::telemetry::upcast(family), extract_label_values(labels),
+				broker::telemetry::value(counter)));
+		}
+
+	void operator()(const broker::telemetry::int_counter_family_hdl* family,
+	                const broker::telemetry::int_counter_hdl* counter,
+	                broker::telemetry::label_list labels)
+		{
+		if ( matches(upcast(family)) )
```
Neverlord, line 334: Since all overloads `upcast` the family object, why not consider just passing `metric_family_hdl*` in from Broker in the first place? (see: https://github.com/zeek/broker/pull/258#discussion_r934782170).

src/telemetry/telemetry.bif
Neverlord: The `enum` itself is duplicated.

Once as a C-style enum in https://github.com/zeek/zeek/blob/master/src/broker/comm.bif#L66 and once as `enum class` in https://github.com/zeek/zeek/blob/master/src/broker/Manager.h#L83. Wouldn't it work to use the "proper" definition from `broker/Manager.h` and make that available to script land directly?

Neverlord (commented):
Just some suggestions / questions.

src/telemetry/Manager.cc
awelzel: > I believe you have misplaced == FNM_NOMATCH in the first if-block (should move out of the parens)

Thanks, yes. There's a bit of testing, but the misplacing happened to still do the right thing (passing 0 as flags and checking for any truthy value as no-match. Yikes.

src/telemetry/Manager.cc
awelzel: Thanks!

awelzel:
Following PRs for external baseline updates:
https://github.com/zeek/zeek-testing/pull/8
https://github.com/zeek/zeek-testing-private/pull/6

src/telemetry/Manager.h
```diff
@@ -50,6 +59,112 @@ class Manager
 	 */
 	virtual void InitPostScript();
 
+	enum class MetricType
```
rsmmr, line 62: Here and below: everything that's public in a header, should be having Doxygen comments (types, methods, public instance variables, etc.), including documenting parameters and return types.

src/telemetry/Manager.h
```diff
@@ -50,6 +59,112 @@ class Manager
...
+			return CollectedValueMetric(MetricType::Counter, family, std::move(label_values),
+			                            value);
+			}
+
+		static CollectedValueMetric NewGauge(const broker::telemetry::metric_family_hdl* family,
+		                                     std::vector<std::string_view> label_values,
+		                                     std::variant<double, int64_t> value)
+			{
+			return CollectedValueMetric(MetricType::Gauge, family, std::move(label_values), value);
+			}
```
rsmmr, line 88: Do these wrappers add much over users of the class calling the constructors themselves with the appropriate `MetricType` enum? Feels like it's just making the class API more complex for little gain, unless I'm missing something?

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	static auto name_idx = metric_opts_type->FieldOffset("name");
+	static auto help_text_idx = metric_opts_type->FieldOffset("help_text");
+	static auto unit_idx = metric_opts_type->FieldOffset("unit");
+	static auto is_total_idx = metric_opts_type->FieldOffset("is_total");
+	static auto labels_idx = metric_opts_type->FieldOffset("labels");
+	static auto bounds_idx = metric_opts_type->FieldOffset("bounds");
+	static auto metric_type_idx = metric_opts_type->FieldOffset("metric_type");
+
+	if ( metric_opts_cache.count(family) != 0 )
+		return metric_opts_cache[family];
```
rsmmr, line 91: Nit: can save a lookup here:
```suggestion
	if ( auto x = metric_opts_cache.find(family); x != metric_opts_cache.end() )
		return x->second;
```

(or something like that; not tested :-) )

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	r->Assign(prefix_idx, make_intrusive<zeek::StringVal>(broker::telemetry::prefix(family)));
+	r->Assign(name_idx, make_intrusive<zeek::StringVal>(broker::telemetry::name(family)));
+	r->Assign(help_text_idx, make_intrusive<zeek::StringVal>(broker::telemetry::helptext(family)));
+	r->Assign(unit_idx, make_intrusive<zeek::StringVal>(broker::telemetry::unit(family)));
+	r->Assign(is_total_idx, val_mgr->Bool(broker::telemetry::is_sum(family)));
+
+	auto label_names_vec = make_intrusive<zeek::VectorVal>(string_vec_type);
+	for ( const auto& l : broker::telemetry::label_names(family) )
+		label_names_vec->Append(make_intrusive<StringVal>(l));
+	r->Assign(labels_idx, label_names_vec);
```
rsmmr, line 103: For readbility, add empty line after block

```suggestion

	r->Assign(labels_idx, label_names_vec);
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	// histogram.
+	zeek_int_t metric_type_int = -1;
+	if constexpr ( std::is_same<T, double>::value )
+		{
+		if ( metric_type == MetricType::Counter )
+			metric_type_int = BifEnum::Telemetry::MetricType::DOUBLE_COUNTER;
+		else if ( metric_type == MetricType::Gauge )
+			metric_type_int = BifEnum::Telemetry::MetricType::DOUBLE_GAUGE;
+		else if ( metric_type == MetricType::Histogram )
+			metric_type_int = BifEnum::Telemetry::MetricType::DOUBLE_HISTOGRAM;
```
rsmmr, line 116: Any reason not to use a `switch` here?

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+		}
+	else
+		{
+		if ( metric_type == MetricType::Counter )
+			metric_type_int = BifEnum::Telemetry::MetricType::INT_COUNTER;
+		else if ( metric_type == MetricType::Gauge )
+			metric_type_int = BifEnum::Telemetry::MetricType::INT_GAUGE;
+		else if ( metric_type == MetricType::Histogram )
+			metric_type_int = BifEnum::Telemetry::MetricType::INT_HISTOGRAM;
```
rsmmr, line 125: Same.

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	static auto opts_idx = metric_record_type->FieldOffset("opts");
+	static auto labels_idx = metric_record_type->FieldOffset("labels");
+	static auto value_idx = metric_record_type->FieldOffset("value");
+	static auto count_value_idx = metric_record_type->FieldOffset("count_value");
+
+	auto r = make_intrusive<zeek::RecordVal>(metric_record_type);
+
+	auto label_values_vec = make_intrusive<zeek::VectorVal>(string_vec_type);
+	for ( const auto& l : label_values )
+		label_values_vec->Append(make_intrusive<StringVal>(l));
```
rsmmr, line 152:
```suggestion
		label_values_vec->Append(make_intrusive<StringVal>(l));
		
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	static auto count_observations_idx = histogram_metric_type->FieldOffset("count_observations");
+	static auto count_sum_idx = histogram_metric_type->FieldOffset("count_sum");
+
+	zeek::RecordValPtr opts_record;
+
+	auto r = make_intrusive<zeek::RecordVal>(histogram_metric_type);
+
+	auto label_values_vec = make_intrusive<zeek::VectorVal>(string_vec_type);
+	for ( const auto& l : label_values )
+		label_values_vec->Append(make_intrusive<StringVal>(l));
```
rsmmr, line 191:
```suggestion
		label_values_vec->Append(make_intrusive<StringVal>(l));

```

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+
+	auto fn = [&](auto val)
+	{
+		using val_t = decltype(val);
+		auto opts_record = telemetry_mgr->GetMetricOptsRecord<val_t>(metric_type, family);
+		r->Assign(opts_idx, opts_record);
+		r->Assign(value_idx, as_double_val(val));
+		if constexpr ( std::is_same_v<val_t, int64_t> )
+			r->Assign(count_value_idx, val_mgr->Count(val));
+	};
```
rsmmr, line 163:
```suggestion
	};
	
```

src/telemetry/Manager.cc
```diff
@@ -38,6 +69,415 @@ void Manager::InitPostBrokerSetup(broker::endpoint& ep)
...
+	// This is a bit annoying. The bounds / upper bounds of the individual
+	// buckets are not accessible through the family hdl. We set the
+	// opts$bounds field based on bounds collected from an individual
+	// metric instance.
+	//
+	// As the opts_record is cached by the manager, so below should only
+	// happen once.
+	//
+	// This is probably working around something that should be fixed
+	// in the broker::telemetry level.
```
rsmmr, line 236: Seems worth creating a ticket for on the Broker-side to provide this, and then clean this up here. Shouldn't be too difficult I assume. (But ok to leave as is for now).

src/telemetry/telemetry.bif
rsmmr: The first `enum` declaration is part of the BiF language, not a C construct itself. The manager could `#include` that bif-generated header with the resulting enum, but that's not so nice to use, which is why, iirc, I added a separate one instead (maybe there was a technical issue as well, don't remember).

rsmmr (commented):
One final set of comments.

src/telemetry/Manager.h
awelzel: I think I covered these now.

src/telemetry/Manager.h
awelzel: Motivation was to protect a CollectValueMetric with MetricType::Histogram, but removed them in favor of less code now.

src/telemetry/Manager.cc
awelzel: There was one more that matched the extra line after lamda. That's fixed, too.

rsmmr (approved):
Nice job pulling it all together!

PR #2275, opened 2022-07-14
Fix some warnings from bro_version deprecation

timwoj:
This fixes this warnings that didn't pop up on macOS when I was doing the deprecations, but appear on Ubuntu.


awelzel (approved):
I don't quite get the details why the push needs to be around the whole constructor (and a bit of googling didn't turn up anything for me). It does fix the warnings I saw on Debian with GCC 10, too, so I'm approving.

awelzel:
> I don't quite get the details why the push needs to be around the whole constructor.

Oh! It's warning on the default initialization of the bro_version member.

PR #2277, opened 2022-07-16
Add bitshift operators

### LINKED ISSUE #2226: Add bitshift operators
Primarily interested in logical shifts on `count`s, but wouldn't complain if arithmetic shifting were supported as well (given the type/sign). Seems reasonable to use `>>` and `<<` as they is common in many programming languages.

AmazingPP:
Fixes #2226 

Support logical and arithmetic shifts


vpax:
There's a bunch more needed for script optimization, but I think that's instead on me.  I have a monthly reminder to deal with script optimization bitrot.  On the plus side, as long as this compiles, it won't break any script optimization other than scripts that use the new operators.

timwoj:
@AmazingPP If you want to squash this down, I'll merge it.

AmazingPP:
> @AmazingPP If you want to squash this down, I'll merge it.

Done

PR #2278, opened 2022-07-16
smb2: Raise smb2_file_delete for CREATE with FILE_DELETE_ON_CLOSE

### LINKED ISSUE #2276: SMB: FILE_DELETE not always shown in logs
SMB is not always logging FILE_DELETE in the `smb_files.log`.

This pcap helps demonstrate the problem -- there should be a `FILE_DELETE` for `opcreatep\roattr\rttaor.dat` but it's missing. Other `FILE_DELETE` mentions are logged:
    https://wiki.wireshark.org/uploads/__moin_import__/attachments/SMB2/ifstest.cap.gz

See packets 309 and 310 from the pcap

awelzel:
When a CREATE request contains the FILE_DELETE_ON_CLOSE option and
the subsequent CREATE response indicates success, we now raise the
smb2_file_delete event to log a delete action in smb_files.log and
also give users a way to handle this scenario.

The provided pcap was generated locally by recording a smbtorture run
of the smb2.delete-on-close-perms test case.

Placed the create_options into the CmdInfo record for potential
exposure in smb_cmd.log (wasn't sure how that would look so left it
for the future).

Fixes #2276.


timwoj (approved):
Looks reasonable to me.

PR #2279, opened 2022-07-18
DebugLogger: "auto const" to "const auto"

awelzel:
This was pointed out in a separate PR so fixing the occurrence I
had managed to sneak in already.


vpax:
Just curious - what difference does the ordering make?

awelzel:
> Just curious - what difference does the ordering make?

My understanding is none. Robin had pointed out the convention in another PR. Based on grepping `for.*const auto` and `for.*auto const` the one in this PR is the odd one out and I had only added it a few days back, so figured it's worth the follow-up.

timwoj (approved):

PR #2280, opened 2022-07-18
Fix local `const` variables can be modified via loops

### LINKED ISSUE #1971: Local `const` variables can be modified via loops
If I use a local `const` variable as a loop index it's value gets modified.

```zeek
# file l.zeek
module x;

event zeek_init() {
	const i = 1000;

	# Set `i=2`.
	for (i in vector(1, 2, 3)) { }
	print i;

	# Set `i=999`.
	for (i in table([999]="")) { }
	print i;

	# Rejected: const is not a modifiable lvalue (x::i)
	# i = 1001;
}
```

```.console
$ zeek -b l.zeek
2
999
```

If I comment out the last line I get an error. I would expect a similar error or e.g., `const variable used in for loop` which exists in similar form for globals here.

AmazingPP:
Fixes #1971

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -1911,6 +1911,9 @@ for_head:
 				{
 				if ( loop_var->IsGlobal() )
 					loop_var->Error("global variable used in for loop");
+
+				if ( loop_var->IsConst() )
+					loop_var->Error("constant used in for loop");
 				}
 
 			else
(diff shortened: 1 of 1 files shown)


vpax:
Minor comment: this might be worth factoring into a help function, given it's repeated in a bunch of places.

AmazingPP:
> Minor comment: this might be worth factoring into a help function, given it's repeated in a bunch of places.

Done

rsmmr:
Slight tweaks in #2287, will merge after CI passes there.

Issue #2188, opened 2022-06-15
Modbus logs not corresponding to wireshark export


AaronDeRybelHowest:
When I create a standard modbus log from a pcapng file i receive **12361** lines.
When I create an export and filter on modbus/tcp recognised in wireshark **12790** lines.

There seems to have been some packets lost in converting the network capture to this log. 
The way wireshark recognizes is to check for protocol=TCP and port=502 (also see: https://ask.wireshark.org/question/16386/decoding-modbus-tcp-on-different-port-number/)
But how does zeek recognizes modbus packets? This could explain the difference in lines.

AaronDeRybelHowest:
Hello @0xxon , 
Did you have time to look into this? 
I also made a comparison based on the timestamps of zeek logs and wireshark export of modbus protocol. I can share this with you

0xxon:
Hi,
sorry - no, I did not have time to look into this - and might not have time to look into this myself. So it would be great if you could share your data publicly so that others can take a look at it too.

The comparison would be interesting, especially the differences.

AaronDeRybelHowest:
I've made the github public: https://github.com/AaronDeRybelHowest/pcap-share
I also included a couple of files in different formats

Anyone that can look into this different processing of modbus packets?

micrictor:
The disparity is because the default script only logs if `is_orig` is False. 

https://github.com/zeek/zeek/blob/1d2c12e9801018575f462a496b7da87e2cbd21e1/scripts/base/protocols/modbus/main.zeek#L58

Removing that condition from the script results in the expected number of entries in the log.

I think that this is because your PCAP is mid-stream - there's no TCP handshake indicating the start of the TCP session.

0xxon:
Oh, that is great - thank you very much for examining this.

micrictor:
I need to look into it more - the is_orig check seems like it might have
unintended consequences even if the session initiation is observed.

AaronDeRybelHowest:
Hi @micrictor ,
Thank you so much for looking into this, If I can help in any way in this analysis please let me know.

PR #2281, opened 2022-07-18
Improve default Modbus analyzer logging

micrictor:
Improve upon the default Modbus over TCP/IP analyzer logging.

* Log both modbus requests and responses
* Add Modbus/TCP transaction and unit identifier fields to the log
* In the case of Modbus response PDUs, invert the originator and responder in the log. This is in order to reflect the fact that the Modbus/TCP client is the responder to the Modbus response PDU, and the TCP server is the originator of the response PDU.

Addresses #2188


scripts/base/protocols/modbus/main.zeek
```diff
@@ -48,14 +52,28 @@ event modbus_message(c: connection, headers: ModbusHeaders, is_orig: bool) &prio
...
+	## doesn't follow the direction of the TCP session
+	if ( ! is_orig )
+		{
+				const orig_c = copy(c);
+
+				c$id$orig_h = orig_c$id$resp_h;
+				c$id$orig_p = orig_c$id$resp_p;
+
+				c$id$resp_h = orig_c$id$orig_h;
+				c$id$resp_p = orig_c$id$orig_p;
```
0xxon, line 69: I don't feel great about this. I am not entirely sure - but I think this might cause issues, since this means that the underlying internal connection record, and the script-level connection are now out of sync.

Perhaps we should just expose `FlipRoles` to scriptland, so that a script can set it? Or just mark the fact that the order was reversed in some other way in the modbus log.

micrictor:
I, also, don’t feel too great about it - I have some concerns about users
that may rely on the way orig/resp currently is. No better way to start a
conversation than kick it out there!

It’s already implicitly included by way of `is_orig`, but I think that
having it only represented there masks the truth of the underlying
protocol. Is PDU a ubiquitous enough term that adding a `pdu_type` field to
the log, and setting it to “REQ” or “RESP” makes it more clear?

scripts/base/protocols/modbus/main.zeek
micrictor: Re-did this - instead of swapping conn record values, added a field to indicate the PDU direction

0xxon (approved):
Looks good to me - thank you. Unless anyone else has any comments, I am going to merge this tomorrow-ish.

PR #2286, opened 2022-07-19
Add btest that runs the doctest-based unit tests.

### LINKED ISSUE #2267: Run unit test as part of `make test`
The tests in the `testing` folder currently cover our btests and any external testsuites, but we don't invoke `zeek --test`. We do run the unit tests in CI, [here](https://github.com/zeek/zeek/blob/master/ci/test.sh#L38-L45). Since we now build the unit tests by default, it'd be good to also run those from the `testing` folder.

(@rsmmr, fyi)

rsmmr:
Closes #2267.

### DIFF
--- a/testing/btest/core/doctest.zeek
+++ b/testing/btest/core/doctest.zeek
@@ -1,0 +1,7 @@
+# @TEST-REQUIRES: zeek --test -h >/dev/null
+# @TEST-EXEC: ZEEK_SEED_FILE= zeek --test >&2
+#
+# @TEST-DOC: Run the doctest-based unit tests.
+#
+# Note that we need to clear the hash seed as the units tests seem to make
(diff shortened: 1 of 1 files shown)


0xxon (approved):
Looks good to me. I am just curious about why the custom seed breaks ordering - especially since ordering is not guaranteed without a seed.

rsmmr:
> Looks good to me. I am just curious about why the custom seed breaks ordering - especially since ordering is not guaranteed without a seed.

I'm not sure either actually, a couple of the dictionary iteration tests break.

0xxon:
> I'm not sure either actually, a couple of the dictionary iteration tests break.

Weeeeeeeeeeeird. I wonder if they are completely stable then.

PR #2290, opened 2022-07-19
Return an error for duplicate record field names

### LINKED ISSUE #1628: Prevent duplicate definition in record
I was surprised to just find out that it is possible to have two members with the same name in a record as long as they have differing types.

The following code:

```
type test: record {
	a: count;
	a: string &optional;
};

event zeek_init()
	{
	local a = test($a=5);
	print a;
	}
```

Yields the output:

```
[a=5, a=<uninitialized>]
```

We should probably prevent this :)

timwoj:
This one turned into a bit of a rabbit hole, and fixes more than just the original bug reported in the issue.

Fixes #1628


src/List.h
```diff
@@ -131,6 +131,9 @@ template <typename T, ListOrder Order = ListOrder::ORDERED> class List
 	// Return nth ent of list (do not remove).
 	T& operator[](int i) const { return entries[i]; }
 
+	T& at(int i) { return entries[i]; }
+	const T& at(int i) const { return entries[i]; }
```
0xxon, line 135: At least in the STL, the `at` operation typically does a bound-check. This does not seem to do this. Is there a reason to also have this in addition to the `[]` operator?

I might just be missing something here.

src/util.h
```diff
@@ -571,5 +571,56 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
 std::string json_escape_utf8(const char* val, size_t val_size,
                              bool escape_printable_controls = true);
 
+/**
+ * Splits a string at all occurrences of a delimiter. Successive occurrences
+ * of the delimiter will be split into multiple pieces.
+ *
+ * \note This function is not UTF8-aware.
+ */
```
0xxon, line 579: Since this is a new utility function - should we add unit tests for it?

src/util.h
```diff
@@ -571,5 +571,56 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
...
+template <typename T> std::vector<T> split(T s, const T& delim)
```
0xxon, line 580: Could this be `template <typename T> std::vector<T> split(const T s, const T& delim)`, to allow for const string arguments?

src/util.h
```diff
@@ -571,5 +571,56 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
...
+	}
+
+template <typename T, typename U = typename T::value_type*> std::vector<T> split(T s, U delim)
+	{
+	return split(s, T{delim});
+	}
+
+inline std::vector<std::string> split(const char* s, const char* delim)
+	{
+	return split(std::string(s), std::string(delim));
```
0xxon, line 617: Could this work with a string_view, instead of constructing full strings?

testing/btest/scripts/base/frameworks/input/setspecialcases.zeek
```diff
@@ -26,7 +26,7 @@ type Idx: record {
 
 type Val: record {
 	s: set[string];
-	s: vector of string;
+	ss: vector of string;
```
0xxon, line 29: heh.  I never noticed this one.

0xxon (commented):
I added a couple of small questions.

src/util.h
timwoj: It should, yes. I'll copy over the ones from zeek-agent-v2.

src/util.h
timwoj: It'd require some changes since `s` gets modified by the loop internally, but certainly doable.

src/util.h
timwoj: That depends. The `vector` would also have to contain `std::string_view` as well. Would that be surprising to users? I suppose as long as it's well documented it shouldn't matter much.

testing/btest/scripts/base/frameworks/input/setspecialcases.zeek
timwoj: Yep. The surprising things that popup when you fix a bug :smiley:

src/List.h
timwoj: It just makes it possible to do t->at(i) instead of having to do (*t)[i]. It results in the same code either way, so maybe I'm just being anal about how it looks.

src/util.h
0xxon: Ah, I did not think about that. I was honestly only thinking about probably unnecessary efficiency here.

src/util.h
0xxon: Ah, I missed the substring operation assigning to the same variable. That's fine then.

src/List.h
0xxon: FWIW -as an unsuspecting user, I think I might expect the operator to do bounds-checking, mostly based on the fact that the stl does it for all at operators.

That being said - I don't have super strong feelings about it.

0xxon (approved):

src/List.h
timwoj: Yep, I was just looking at `std::vector`. `at()` does bounds checking, where as `operator[]` doesn't. I'll just take it out 😃

src/util.h
timwoj: I completely agree that constructing a string here feels really bad. I could convert the output vector to `std::string` in the `const char*` version of the method, but I feel like that just pushes the time sink to later in the process.

PR #2291, opened 2022-07-19
GH-1344: Give better warning when using a type that doesn't exist

### LINKED ISSUE #1344: Core dump: cloning illegal base Type
$ zeek --version
zeek version 4.0.0-rc1

An accidental typo revealed a core-dump. You can replicate with the code snippet here:

```
$ cat crash.zeek

type mac: stringA;

event zeek_init()
{
        print fmt ("hello");
}

```

`Core was generated by 'zeek -i ixl3 ../host-profiling.zeek'.`
```
Program terminated with signal SIGABRT, Aborted.
#0  0x0000000801b171ba in thr_kill () from /lib/libc.so.7
(gdb) bt
#0  0x0000000801b171ba in thr_kill () from /lib/libc.so.7
#1  0x0000000801b155e4 in raise () from /lib/libc.so.7
#2  0x0000000801a897e9 in abort () from /lib/libc.so.7
#3  0x00000000007ce997 in zeek::Reporter::InternalError (this=<optimized out>, fmt=<optimized out>)
    at /home/bro/install/zeek-4.0.0-rc1/src/Reporter.cc:205
#4  0x0000000000815039 in zeek::Type::ShallowClone (this=<optimized out>) at /home/bro/install/zeek-4.0.0-rc1/src/Type.cc:219
#5  0x000000000083de8f in zeek::detail::add_type (id=0x44a1d50, t=...,
    attr=std::unique_ptr<std::__1::vector<zeek::IntrusivePtr<zeek::detail::Attr>, std::__1::allocator<zeek::IntrusivePtr<zeek::detail::Attr> > >> = 0x0)
    at /home/bro/install/zeek-4.0.0-rc1/src/Var.cc:383
#6  0x00000000005e0f7f in yyparse () at parse.y:1183
#7  0x00000000006fc57d in zeek::detail::setup (argc=2, argv=<optimized out>, zopts=0x2331960) at /home/bro/install/zeek-4.0.0-rc1/src/zeek-setup.cc:650
#8  0x00000000005d8315 in main (argc=6, argv=0x0) at /home/bro/install/zeek-4.0.0-rc1/src/main.cc:13
(gdb) up
#1  0x0000000801b155e4 in raise () from /lib/libc.so.7
(gdb) up
#2  0x0000000801a897e9 in abort () from /lib/libc.so.7
(gdb) up
#3  0x00000000007ce997 in zeek::Reporter::InternalError (this=<optimized out>, fmt=<optimized out>)
    at /home/bro/install/zeek-4.0.0-rc1/src/Reporter.cc:205
205             abort();
(gdb) up
#4  0x0000000000815039 in zeek::Type::ShallowClone (this=<optimized out>) at /home/bro/install/zeek-4.0.0-rc1/src/Type.cc:219
219                             reporter->InternalError("cloning illegal base Type");
(gdb) p
The history is empty.
(gdb) l
214                     case TYPE_SUBNET:
215                     case TYPE_ANY:
216                             return make_intrusive<Type>(tag, base_type);
217
218                     default:
219                             reporter->InternalError("cloning illegal base Type");
220             }
221             return nullptr;
222             }
223
(gdb) quit`

```

timwoj:
Fixes #1344


src/Var.cc
```diff
@@ -429,14 +429,25 @@ void add_type(ID* id, TypePtr t, std::unique_ptr<std::vector<AttrPtr>> attr)
...
 	TypePtr tnew;
 
 	if ( (t->Tag() == TYPE_RECORD || t->Tag() == TYPE_ENUM) && old_type_name.empty() )
 		// An extensible type (record/enum) being declared for first time.
 		tnew = std::move(t);
 	else
+		{
+		// If the old type is an error or the old type doesn't exist, then return
+		// an error instead of trying to clone it.
+		if ( t->Tag() == TYPE_ERROR && t->InternalType() == TYPE_INTERNAL_ERROR )
```
rsmmr, line 442: The comment's "or" doesn't seem to match the condition's `&&`? Maybe it doesn't need to, but then I'd be wondering if we need to check both tag and internal type?

rsmmr (approved):

src/Var.cc
timwoj: We can't just check the tag, since there are cases where tag is set to error but the internal type is set to something valid. Would it be sufficient to just check the internal type for an error?

rsmmr: Ah, then I'd think checking just the tag (not the internal type) would give us what we want (because `SetError()` updates the tag, but not the internal type). But sounds like you're saying that's not actually the right thing here?

Either way, fine to leave as you have it if that's working as desired, not worth the trouble.

timwoj: One of the enum btests was failing if I just used the tag.

Issue #1890, opened 2021-12-10
"mixing vector and scalar operands is deprecated" doesn't hold for string operands


vpax:
The following:

> global g = vector(3, 5, 7);
> print g == 7;

produces the warning `mixing vector and scalar operands is deprecated`, and prints `[F, F, T]`.  However, the following:

> global g = vector("a", "b", "c");
> print g == "b";

doesn't produce any warning (and prints `[F, T, F]`).  This seems inconsistent.

ezekiiel:
Is the desired behavior here that

```zeek
const nv = vector(1, 2, 3);
print nv * 2;
```

also fails or should this only apply to equality expressions?

vpax:
The desired behavior is that (1) yes, the example you give also generates a warning, and (2) mixed string operations (equality or otherwise) generate warnings, just like they do for arithmetics.

Basically, we've decided that the functionality of mixing vector and scalar operands was a mistake (much as I like it in some contexts!) and thus shouldn't be allowed, period.

PR #2294, opened 2022-07-21
Some additional changes for the bit-shift operators

timwoj:
This is a follow-on to #2277 

The primary change is to add some error handling for some undefined behavior (reported by ubsan) with bit-shift operations in C++:
- Both left and right shifts return `runtime error: shift exponent -1 is negative` when shifting by a negative number.
- Performing a left shift on a negative number returns `runtime error: left shift of negative value`.

It also adds a btest covering left and right shift of vectors of integral values. This feature is a little questionable to me in general. https://github.com/zeek/zeek/issues/1890 added better messaging for deprecated mixing of scalar and vector operations, but the bit-shift operators don't seem to trigger the warning for some reason.

I should have caught these issues in my review, but the review of https://github.com/zeek/zeek-docs/pull/127 made me realize that there weren't enough tests added for these changes.


vpax:
@timwoj regarding not triggering a deprecation warning for using these on vectors, I'm game to look into that when I do the maintenance for script optimization early next month.

src/Expr.cc
```diff
@@ -960,11 +960,47 @@ ValPtr BinaryExpr::Fold(Val* v1, Val* v2) const
 			DO_UINT_FOLD(^);
 			break;
 		case EXPR_LSHIFT:
-			DO_INT_FOLD(<<);
+			{
+			if ( is_integral )
+				{
+				if ( i2 < 0 )
+					RuntimeError("left shift by negative value");
```
rsmmr, line 967: Any reason not to use `RuntimeErrorWithCallStack` for the new error cases?

rsmmr (approved):

src/Expr.cc
vpax: @rsmmr Isn't that for internal issues?  The surrounding code only uses it when there's confusion over a type tag or such.

rsmmr: Hmm, yeah, I think you're right. I guess I would have expected an internal error then for such type confusion if the assumption is that it can't happen other than for bugs.

PR #2295, opened 2022-07-21
Return an error if redef'ing a variable without redef attribute

### LINKED ISSUE #2131: Segfault when redef'ing global function without &redef
```
¡ zeek --version
zeek version 4.2.0-dev.406
```

I have not tried this with a newer version of Zeek.

```zeek
¡ cat ~/tmp/foo.zeek
export {
        global func1: function();
        global func2: function();
}

function func1()
        {
        print "func1()";
        }

function func2()
        {
        print "func2()";
        }

export {
        global func = func1;
}

event zeek_init()
        {
        func();
        }

¡ zeek ~/tmp/foo.zeek
func1()

¡ zeek ~/tmp/foo.zeek -e 'redef func = func2;'
zsh: segmentation fault  zeek ~/tmp/foo.zeek -e 'redef func = func2;'
```

This works as expected if the second export block has `global func = func1 &redef;`:

```zeek
¡ zeek ~/tmp/foo.zeek -e 'redef func = func2;'
func2()
```

I expect an error like `func does not have &redef` or something similar.

timwoj:
Fixes #2131


src/Var.cc
```diff
@@ -223,6 +223,12 @@ static void make_var(const IDPtr& id, TypePtr t, InitClass c, ExprPtr init,
 				id->Warn("redefinition requires \"redef\"", redef_obj, true);
 			}
 
+		else if ( dt == VAR_REDEF && ! id->IsRedefinable() )
```
awelzel, line 226: Not that I know my way around here, but wondering if the `if ( dt == VAR_REDEF )` condition of the block below could be re-used?

https://github.com/zeek/zeek/pull/2295/files#diff-7b1d693024302c91261b7aae6d2a1d658971067a46b912c3965b44f0fd4f1663R243

timwoj: I had added it there originally but it falls into the block directly below the one I added. Since we're redef'ing a function in this instance, it ends up calling `add_prototype` and gets into the call chain that eventually results in the crash. Adding the extra if block avoids that.

awelzel (approved):

PR #2298, opened 2022-07-27
Support /s pattern modifier

timwoj:
I'm opening this as a draft to look at the approach, since the pattern matching code is a bit weird in places. I left some TODOs in the code in places that I wasn't entirely sure how to handle, mostly around how the CCL object is created.


src/RE.cc
```diff
@@ -394,14 +408,22 @@ static RE_Matcher* matcher_merge(const RE_Matcher* re1, const RE_Matcher* re2, c
...
 	snprintf(merge_text, n, "(%s)%s(%s)", text1, merge_op, text2);
 
 	RE_Matcher* merge = new RE_Matcher(merge_text);
 	delete[] merge_text;
 
+	// TODO: I'm not sure this is correct. If either of the joined matchers
+	// have the single-line flag set, then the conjuction will have it set.
+	// But that's not really right, since only the part that has the flag
+	// should be single-line. For example if you have /a.c/s | /d.f/, the
+	// dot should only match \n for the first part.
```
vpax, line 423: I agree that this is a bit weird, and conceivably could be buggy.  I haven't yet looked at the rest of the changes, but I'd think a general approach would be for `/s` to result in translating (initially, not at compilation time) any `.`'s in the pattern to a new meta-character that means anything-including-newlines.  If that were done, then there's no need to explicitly propagate the SingleLine status.

src/RE.h
```diff
@@ -59,12 +59,13 @@ enum match_type
...
-	explicit Specific_RE_Matcher(match_type mt, int multiline = 0);
+	explicit Specific_RE_Matcher(match_type mt, bool multiline = false);
 	~Specific_RE_Matcher();
 
 	void AddPat(const char* pat);
 
 	void MakeCaseInsensitive();
+	void SetSingleLineMode();
 
 	void SetPat(const char* pat) { pattern_text = util::copy_string(pat); }
```
vpax, line 71: This shouldn't be a `copy_string` any more, just a direct assignment.

src/re-parse.y
```diff
@@ -212,6 +215,9 @@ ccl		:  ccl TOK_CHAR '-' TOK_CHAR
 				}
 			else
 				$1->Add($2);
+
+			if (zeek::detail::re_single_line && $2 == '.' )
```
vpax, line 219: I don't believe this is correct.  `.` inside a character class is a literal, not a meta-character.

vpax: Also, I'd expect line 115: 

> 			$$ = new zeek::detail::NFA_Machine(new zeek::detail::NFA_State(zeek::detail::rem->AnyCCL()));

to pass a flag to `AnyCCL` indicating whether it wants the include-newline version of `.` or the don't-include-newline version, with a matching change for `AnyCCL` to create each of those on demand.

src/scan.l
```diff
@@ -562,13 +562,28 @@ F	RET_CONST(zeek::val_mgr->False()->Ref())
...
+	yylval.re_modes.single_line = false;
 	return TOK_PATTERN_END;
 	}
 
-<RE>"/i"	{
+<RE>(\/[is]{0,2})	{
 	BEGIN(INITIAL);
-	yylval.b = true;
+
+	if (strlen(yytext) == 2)
```
vpax, line 573: (whitespace around outer parens doesn't match clang style)

vpax (commented):
It's possible I'm missing some considerations, let's chat further

timwoj:
The wild inconsistency of what the flags mean is sort of annoying with this when it comes to naming things. For example, what we call `multiline` only really covers part of what most other languages call that:

> (`?m`/`/m`) for “multi-line mode” makes the [caret and dollar](https://www.regular-expressions.info/anchors.html) match at the start and end of each line in the subject string. In Ruby, (?m) makes the dot match all characters, without affecting the caret and dollar which always match at the start and end of each line in Ruby. In Tcl, (?m) also prevents the dot from matching line breaks.

src/re-parse.y
timwoj: How should the `EquivClass` in the `Specific_RE_Matcher` be updated in this case then? Do the same thing, using a separate EC for the two modes?

vpax: I think it should be the same EC for both branches.  ECs span the entirety of the regular expression, so if an RE mixes both types of "any", then the EC needs to be informed of that via `CCL_Use()`.

timwoj: Alright, I think 2d86aa39e630dcbe8ed1b10b4111653ab9cf0cbd covers all of that correctly now.

src/RE.cc
```diff
@@ -47,10 +48,22 @@ Specific_RE_Matcher::~Specific_RE_Matcher()
...
+			}
+
+		return single_line_ccl;
+		}
+
 	if ( ! any_ccl )
-		{ // Create the '.' character class.
+		{
 		any_ccl = new CCL();
 		if ( ! multiline )
```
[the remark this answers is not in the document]
timwoj: Hope about this. I'll go ahead and merge this, and then write a second issue for removing the multiline argument. That would let this bake for a little bit and make sure we didn't miss anything.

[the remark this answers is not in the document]
vpax: sounds good

PR #2302, opened 2022-07-28
Prevent large dhcp log entries

pbcullen:
A flood of DHCP traffic can result in very large log entries consisting
of many uids and/or msg_types. Such large log entries can disrupt a SIEM
ingestion pipeline. This change forces a log entry to be written when
the number of uids or the number of msg_types exceed a certain value.
The values are treated as options for easy configuration.


rsmmr (approved):
I see CI is still running, so waiting for that, but looks good tme.

PR #2305, opened 2022-07-29
TCP Options-27,28,29,34 invalid length check

fatemabw:
Recently saw some malicious traffic with invalid TFO (TCP Fast Open) cookie length - 2 bytes. This traffic was picked up by a suri alert for "TCP options invalid length". I realized that Zeek didn't flag it, bcoz TFO - TCP option no. 34, isn't checked against the invalid length. Hence, thought to write some quick length check for 4 popularly seen TCP options.

I also realized that the weirds associated with TCP options aren't triggered and logged by default, until "tcp_option/tcp_options" events are explicitly called in a script. We might want to change that, as it would be nice to log these weirds by default w/o explicitly calling tcp_option/tcp_options events. 
Test: I was testing this code against a pcap with bad length TFO option, and it didn't log any weirds, but when I tested the same code against the same pcap with an explicit script having tcp_options event called, then I saw the weirds getting logged in weird.log. That's the current behavior.


scripts/base/init-bare.zeek
```diff
@@ -380,6 +380,10 @@ export {
 		send_timestamp: count &optional;
 		## Kind 8: 4-byte echo reply timestamp value.
 		echo_timestamp: count &optional;
+		## Kind 27: TCP Quick Start Response value.
+                rate: count &optional;
+                ttl_diff: count &optional;
+                qs_nonce: count &optional;
```
rsmmr, line 386: Looks like the whitespace is a bit off here.

testing/btest/core/tcp/options.zeek
```diff
@@ -38,6 +38,9 @@ event tcp_options(c: connection, is_orig: bool, options: TCP::OptionList)
 				print fmt("    send ts: %s", o$send_timestamp);
 				print fmt("    echo ts: %s", o$echo_timestamp);
 				break;
+			case 27:
+                                print fmt("     TTL Diff: %s", o$ttl_diff);
+                                break;
```
rsmmr, line 43: I suppose the trace the test is using doesn't contain option 27 because otherwise the baseline would change. Do you have a trace with option 27 that you could add? And then log all the new fields. (Ok to skip traces for the new weirds I think, not worth the effort it seems.)

Some whitespace tweaking needed here too.

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -1803,6 +1803,50 @@ int TCPSessionAdapter::ParseTCPOptions(const struct tcphdr* tcp, bool is_orig)
...
+						auto ttl_diff = o[3];
+						auto qs_nonce = ntohl(*reinterpret_cast<const uint32_t*>(o + 4));
+						option_record->Assign(8, rate);
+						option_record->Assign(9, ttl_diff);
+						option_record->Assign(10, qs_nonce);
+						}
+					else
+						{
+						add_option_data(option_record, o, length);
+						Weird("tcp_option_QSResponse_invalid_len", util::fmt("%d", length));
```
rsmmr, line 1820: Seems the existing weirds use all-lower-case, so I suggest to do that here too for consistency (same for the other new weirds below).

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -1803,6 +1803,50 @@ int TCPSessionAdapter::ParseTCPOptions(const struct tcphdr* tcp, bool is_orig)
...
+					break;
+
+				case 34:
+					// TCP Fast open TFO
+					if ( (length != 2) && (length < 6 || length > 18) )
+						{
+						add_option_data(option_record, o, length);
+						Weird("tcp_option_TFO_invalid_len", util::fmt("%d", length));
+						}
+					break;
```
rsmmr, line 1849:
```suggestion
					break;
					
```

testing/btest/core/tcp/options.zeek
fatemabw: Unfortunately, the pcap I tested the code with don't have option 27, I can try getting a scapy created pcap for that option and test with it, if we want to test it against that option too.
The only option I was able to check with one of our pcaps was TFO - 34:

1658331756.796497	CDWjhQ1NKSgdvEJwva	109.238.x.x	49380	x.x.x.x	80	tcp_option_TFO_invalid_len	4	F	zeek	TCP
1658331756.796497	C6brlK1euEJ8Z3j7A7	109.238.y.y	7396	y.y.y.y	80	tcp_option_TFO_invalid_len	4	F	zeek	TCP

fatemabw (commented):
Resolved most of the comments except the test pcap one.

scripts/base/init-bare.zeek
fatemabw: Sorry about that, copy paste from the server to the github sometime messes with the whitespace. Fixed. :)

testing/btest/core/tcp/options.zeek
rsmmr: yeah, I think finding/creating a trace with option 27 would be great if you could do that, otherwise we'd be adding a code path here without exercising it.

rsmmr:
Do you think you can still create a pcap? Otherwise I'll go ahead and merge as is.

fatemabw:
> 

Give me one more day, and I will let you know by tomorrow. Sorry for the delay on this.

timwoj:
@rsmmr Are you happy with this one now?

rsmmr:
Yes, I'll merge, thanks @fatemabw!

PR #2307, opened 2022-08-01
convert a loop to a Syntactic sugar way

KSroido:
BTW I just konw and start to use zeek and willing to contribute some code.
I've went through doc one or two times, but still got a problem that is what are  ```.l``` and ```.y``` files?
it looks just like combination of C style include file and zeek code. But what exact purpose are them?
I thought zeek is something like python, and this repo is a combination of zeek's basic scripts (with file extension ```.zeek```) and zeek's interpreter(implemented by C++). Am I right
I appreciate your time

### DIFF
--- a/src/Attr.cc
+++ b/src/Attr.cc
@@ -279,9 +279,9 @@ const AttrPtr& Attributes::Find(AttrTag t) const
 
 void Attributes::RemoveAttr(AttrTag t)
 	{
-	for ( auto it = attrs.begin(); it != attrs.end(); )
+	for ( auto& it : attrs )
 		{
-		if ( (*it)->Tag() == t )
+		if ( it->Tag() == t )
 			it = attrs.erase(it);
 		else
 			++it;
--- a/src/Attr.cc
+++ b/src/Attr.cc
@@ -279,7 +279,7 @@ const AttrPtr& Attributes::Find(AttrTag t) const
 
 void Attributes::RemoveAttr(AttrTag t)
 	{
-	for ( auto& it : attrs )
+	for ( auto it : attrs )
 		{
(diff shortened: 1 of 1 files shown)


KSroido:
I am sorry but maybe I should dig into it more before PR

AmazingPP:
> I've went through doc one or two times, but still got a problem that is what are `.l` and `.y` files?
> it looks just like combination of C style include file and zeek code. But what exact purpose are them?

`.l` file is the input for the [Flex](https://github.com/westes/flex). `.y` file is the input for the [Bison](https://www.gnu.org/software/bison/). You can simply think that the `.l` file provides lexical analysis and the `.y` file provides syntax analysis. They work together to parse the Zeek scripts.

MySQL's SQL parser is also implemented by Flex and Bison[^Parsing-SQL].

> I thought zeek is something like python, and this repo is a combination of zeek's basic scripts (with file extension `.zeek`) and zeek's interpreter(implemented by C++). Am I right

Not completely right, These two parts are not the focus of this repository. Zeek is actually a network traffic analyzer[^what-is-zeek].  Zeek scripts is one part of Zeek,  as an event-driven scripting language. it provides great flexibility to analyze network traffic. Zeek scripts can generate real-time alerts and also execute arbitrary external programs on demand. One might use this functionality to trigger an active response to an attack.

[^Parsing-SQL]: https://www.oreilly.com/library/view/flex-bison/9780596805418/ch04.html

[^what-is-zeek]: https://docs.zeek.org/en/master/about.html#what-is-zeek

KSroido:
Thanks a lot @AmazingPP

KSroido:
BTW ur blog is so good. learnt a lot @AmazingPP

PR #2308, opened 2022-08-01
GH-2034: Store module names and use them in lookups for ifdef

### LINKED ISSUE #2034: Allow @ifdef to work on module namespaces
Exported variables in modules can be used in `@ifdef`, but the module namespaces themselves cannot. For example:

```zeek
$ cat bar.zeek
module Bar;

export {
  option exists = T;
}
$ cat foo.zeek
event zeek_init()
{
  @ifdef ( Bar::exists )
    print "Bar is defined";
  @endif
}
$ zeek foo.zeek
$ zeek bar.zeek foo.zeek
Bar is defined
$ cat foo2.zeek
event zeek_init()
{
  @ifdef ( Bar )
    print "Bar is defined";
  @endif
}
$ zeek foo2.zeek
$ zeek bar.zeek foo2.zeek
```

It would be cleaner, imo, if `@ifdef` also worked on the module namespace itself.

timwoj:
Fixes #2034

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -156,6 +156,8 @@ static int func_hdr_cond_epoch = 0;
 EnumType* cur_enum_type = nullptr;
 static ID* cur_decl_type_id = nullptr;
 
+std::set<std::string> module_names;
+
 static void parse_new_enum(void)
 	{
 	// Starting a new enum definition.
@@ -1269,6 +1271,7 @@ decl:
 		TOK_MODULE TOK_ID ';'
 			{
 			current_module = $2;
+			module_names.insert($2);
 			zeekygen_mgr->ModuleUsage(::filename, current_module);
 			}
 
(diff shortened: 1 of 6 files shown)


rsmmr (commented):
This look good in terms of mechanics, but I'm wondering if that parser should be the one storing this state (i.e., either in `scan.l` or `parse.y`), not `Scope`. That way, it would be located where it's used and updated, which seems more natural to me.

timwoj:
> That way, it would be located where it's used and updated, which seems more natural to me.

That seems reasonable. I was mostly trying to avoid making yet another bunch of globals. I guess if it's just the one `set` it's not a big deal.

timwoj:
Reworked to just store it inside parse/scan.

rsmmr (approved):

PR #2315, opened 2022-08-03
Squelch the zeekygen warnings for command line

### LINKED ISSUE #2310: Zeekygen warning for @loads via -e command line
Loading a script via -e command line (e.g. `zeek -e "@load test"`) generates the following warning:

    warning in ./test.zeek, line 1: Failed to add Zeekygen script doc dependency test.zeek for <command line>

Maybe this way the script is never added to `scripts` in the [zeekygen manager](https://github.com/zeek/zeek/blob/master/src/zeekygen/Manager.cc).

AmazingPP:
Fixes #2310


awelzel:
Hey, thanks for your PR.

Would a side-effect of adding `"<command line>"` to the manager as a script be that the following special case isn't needed anymore:

https://github.com/zeek/zeek/blob/master/src/zeekygen/Manager.cc#L36-L37

However, after looking a bit more, it doesn't seem that scripts listed directly on the command line directly (i.e. when using `zeek ./script1 ./script2`) would go through `Manager::ScriptDependency` either. It seems `@load` from `-e` should be treated similarly: Maybe short-cut `<command line>` within `Manager::ScriptDependency` instead?

AmazingPP:
> Would a side-effect of adding `<command line>` to the manager as a script be that the following special case isn't needed anymore:
> 
> https://github.com/zeek/zeek/blob/85e8d1788708328e09756a0c5006d0c29833c042/src/zeekygen/Manager.cc#L36-L37

Yes, When using the `-e` parameter, Zeek is essentially executing a script, So I think add `<command line>` to the manager is correct, And `WarnMissingScript` should be removed.

> However, after looking a bit more, it doesn't seem that scripts listed directly on the command line (i.e. when using `zeek ./script1 ./script2`) would go through `Manager::ScriptDependency` either. 

`Manager::ScriptDependency` Called only when using `@load`, Indicates that the current script depends on the script that needs to be loaded.
https://github.com/zeek/zeek/blob/85e8d1788708328e09756a0c5006d0c29833c042/src/scan.l#L372-L378

using `zeek ./script1 ./script2`. it's just loading scripts without dependencies.
https://github.com/zeek/zeek/blob/85e8d1788708328e09756a0c5006d0c29833c042/src/zeek-setup.cc#L680-L681

awelzel:
> 
> `Manager::ScriptDependency` Called only when using `@load`, Indicates that the current script depends on the script that needs to be loaded.
> 
> using `zeek ./script1 ./script2`. it's just loading scripts without dependencies. So it won't go through `Manager::ScriptDependency`.
> 

Understood, yes. My thought is that a `zeek -e '@load ./script1'` could be treated just the same as `zeek ./script1` from a zeekygen dependency perspective. Ignoring the dependency relationship on the synthetic `<command line>` filename in Manager also reasonable. I'm leaning on the ignoring side of things, maybe others have stronger opinions. @rsmmr?

AmazingPP:
I thought about it again. I think I may have ignored the meaning of zeekygen itself(to generating documents). It seems that generating documents should not include `command line` scripts.

awelzel:
> I thought about it again. I think maybe I have ignored the meaning of zeekygen itself(to generating documents). It seems that generating documents should not include `command line` scripts.

The scripts loaded from the command line should be registered with zeekygen and that should happen already through `load_files()`, but we don't need to track the dependency on the synthetic `<command line>`.

If you agree, would you be up for submitting a change to replicate the following if statement in `Manager::ScriptDependency()`. That should be enough to squelch the warning from #2310.
https://github.com/zeek/zeek/blob/85e8d1788708328e09756a0c5006d0c29833c042/src/zeekygen/Manager.cc#L36-L37

AmazingPP:
> If you agree, would you be up for submitting a change to replicate the following if statement in `Manager::ScriptDependency()`. That should be enough to squelch the warning from #2310.

Completed, please review.

awelzel (approved):
Seems good to me.

PR #2320, opened 2022-08-04
Expand fuzzer coverage

timwoj:
- Expands the fuzzer coverage to http, ftp, imap, and smtp
- Documents the process for generating fuzzer corpora from pcap files
- Fixes a bug in the fuzzer setup so that event handlers exist, which should allow a bit more coverage on the existing fuzzers
- Adds a couple of extra files to corpus for packet-fuzzer to exercise ICMP, which we weren't covering at all previously.


JustinAzoff:
Interesting.. the asan_sanitizer crashed in the fuzz buffer code.. Might be a bug in that triggered by an invalid corpus file?

timwoj:
> Interesting.. the asan_sanitizer crashed in the fuzz buffer code.. Might be a bug in that triggered by an invalid corpus file?

Ah, looks like the ICMP files I added at the last minute.

timwoj:
Actually it looks like a file that was already in the corpus when I added the ICMP ones (`ipv6-hbh-routing0.trace.cap`). I grabbed the zip file from the master branch and it happens with those files too. I wonder how the heck this is passing on other builds, because it's clearly failing here.

JustinAzoff:
Interesting.. is that change working fine but the increased coverage actually causing an issue to be found?

timwoj:
> Interesting.. is that change working fine but the increased coverage actually causing an issue to be found?

Right. The problem looks like because the overall length is smaller (91 vs 99 bytes) when it gets into `IPv6_Hdr_Chain::Init()`, it bails out early and then completely misses out on one of the headers. At least that's where I am with it so far. I'm looking at how we handle the original pcap with `testing/btest/core/raw_packet.zeek` loaded vs how we handle the converted version via the fuzzer.

JustinAzoff:
The reason might have something to do with how we deliberately setup the fuzzbuffer stuff:

```
                // The point of allocating a new buffer here is to better detect
                // analyzers that may over-read within a chunk  -- ASan wouldn't
                // complain if that happens to land within the full input buffer
                // provided by the fuzzing engine, but will if we allocate a new buffer
                // for each chunk.
                rval.data = std::make_unique<unsigned char[]>(rval.size);
                memcpy(rval.data.get(), chunk_begin, rval.size);
                return {std::move(rval)};
```

there's a flag `copy` in iosource Packet::init that does a similar thing, but iosource/pcap/Source.cc doesn't set it.
If the icmp code is over-reading by 8 bytes but those 8 bytes were already allocated asan won't complain.

timwoj:
> there's a flag `copy` in iosource Packet::init that does a similar thing, but iosource/pcap/Source.cc doesn't set it. If the icmp code is over-reading by 8 bytes but those 8 bytes were already allocated asan won't complain.

Yah but this isn't even the ICMP code. This is `ipv6-hbh-routing0.trace` that's causing problems. Here's a strange data point. I added a bunch of printf debugging to `IPv6_Hdr_Chain::Init()` to print out bytes, lengths, and the header types being processed. In a normal run, you get this:

```
60 00 00 00
41 0 40
2b 00 01 04
0 43 8
11 04 00 02
43 17 40
```

`60 00 00 00` is the start of the IPv6 header. `2b 00 01 04` is 40 bytes forward of that, which is the first IPv6 option (hop-by-hop). `11 04 00 02` is 8 bytes forward of that, and is the next IPv6 option (routing header). On the other hand, the fuzzer does this:

```
60 00 00 00
41 0 40
11 04 00 02
0 17 40
```

For some reason, it's advancing an extra 8 bytes in the first loop and skipping over the hop-by-hop header, which is likely causing problems later. It's not something about the IPv6 code because we're clearly doing the right thing in the normal case. 

We're not running any event handlers or anything here, so it's not that change. We're only triggering the ASAN reporting because of the event handler trying to call `ToPktHdrVal()` on an ipv6 chain that was built incorrectly for some reason.

JustinAzoff:
Ah!  It's entirely possible that the pcap_simplify did the wrong thing, maybe this part is grabbing the wrong bytes:

https://github.com/JustinAzoff/pcap_simplify/blob/master/pcap-to-pkt-with-headers/main.go#L25-L30

and that could result in the 8 byte discrepancy. Are the bytes just wrong in the simplified file?

If the gopacket bits grabbed the wrong bytes I could see that breaking something.  Possibly still means there's a bug in zeek though :-)

timwoj:
Huh, looks like the go code did eat one of the headers:

```
tim@nuc corpus% hexdump -C -s 49 $TRACES/ipv6-hbh-routing0.trace              [topic/timw/2142-more-fuzzers*]
00000031  00 00 00 86 dd 60 00 00  00 00 3b 00 40 20 01 04  |.....`....;.@ ..|
00000041  f8 00 04 00 07 02 e0 81  ff fe 52 ff ff 20 01 04  |..........R.. ..|
00000051  f8 00 04 00 07 02 e0 81  ff fe 52 9a 6b 2b 00 01  |..........R.k+..|
00000061  04 00 00 00 00 11 04 00  02 00 00 00 00 20 01 00  |............. ..|
00000071  78 00 01 00 32 00 00 00  00 00 00 00 01 20 01 00  |x...2........ ..|
00000081  78 00 01 00 32 00 00 00  00 00 00 00 02 00 35 00  |x...2.........5.|
00000091  35 00 0b 86 31 58 58 58                           |5...1XXX|
00000099
tim@nuc corpus% hexdump -C ipv6-hbh-routing0.trace.cap                        [topic/timw/2142-more-fuzzers*]
00000000  01 50 4b 54 02 60 00 00  00 00 3b 00 40 20 01 04  |.PKT.`....;.@ ..|
00000010  f8 00 04 00 07 02 e0 81  ff fe 52 ff ff 20 01 04  |..........R.. ..|
00000020  f8 00 04 00 07 02 e0 81  ff fe 52 9a 6b 11 04 00  |..........R.k...|
00000030  02 00 00 00 00 20 01 00  78 00 01 00 32 00 00 00  |..... ..x...2...|
00000040  00 00 00 00 01 20 01 00  78 00 01 00 32 00 00 00  |..... ..x...2...|
00000050  00 00 00 00 02 00 35 00  35 00 0b 86 31 58 58 58  |......5.5...1XXX|
00000060

```

Ignore that `-s` there on the first command, it's just to make the corresponding start of the IP header line up correctly. I agree though that the Zeek code is definitely not handling this correctly.

timwoj:
I think the issue that Zeek's running into here is that even though the Go code ate that header, the IPv6 header before it still thinks the type of the next header is the one that go eaten. That leads it to try to process something it can't find, and then screws everything up. It obviously shouldn't crash though so I'll fix that.

timwoj:
Alright, looks like that commit fixed the ipv6 asan failure. This is ready for review now.

src/EventRegistry.h
```diff
@@ -60,6 +60,8 @@ class EventRegistry
 
 	void PrintDebug();
 
+	void GenerateAllHandlers();
```
rsmmr, line 63: Please document what this does.

rsmmr: Also, maybe rename to `ActivateAllHandlers`? (Because it's not the handlers being generated, it's the events).

rsmmr (approved):
(Haven't looked very closely at the new protocol fuzzers.)

PR #2321, opened 2022-08-05
GH-2319: Add change handlers to Site

### LINKED ISSUE #2319: Update regexes/table in `Site` when `Site::local_zones` or `Site::local_nets` are updated
`Site::local_zones` and `Site::local_nets` are both options, meaning they are expected to possibly be modified at runtime. However, the regexes and table that are used for detecting if a zone is local or retrieving the CIDR a local IP address belongs to are only [computed once at `zeek_init`](https://github.com/zeek/zeek/blob/6cd37247476181a6f79ed69dc5771cfdde28a1f3/scripts/base/utils/site.zeek#L237). I'd like to change this, as I have code that will depend on these being updated. I could think of two ways to do this and would like feedback if one or both are preferred, or there's a third way that would be even better.

1. Add a [change handler](https://docs.zeek.org/en/master/frameworks/configuration.html#change-handlers) for `Site::local_zones` and `Site::local_nets` and update the regexes and table, respectively, or
1. Expose functions that do the update that the code performing the update can call at their leisure.

My intended use case won't update either frequently, so I'd prefer (1) but there's nothing preventing someone from repeatedly updating either, which could cause problems that would be alleviated by (2). Thoughts? Either would also require either (a) changing the regexes/tables to be `option`s/have `&backend=Broker::MEMORY`, respectively, or (b) using `Broker` to update the values across the cluster. My gut feeling is `&backend=Broker::MEMORY` for the table and using `Broker` to update the regexes, but would appreciate advice here.

Feel free to assign this to me.

ynadji:
Adds a change handler to `Site::local_nets`, `Site::local_zones`, and `Site::neighbor_zones` so when `Config::set_value` is used to change the sets `Site::local_nets_table`, `Site::local_dns_suffix_regex`, and `Site::local_dns_neighbor_suffix_regex` are updated. This ensures that changes of these values at runtime are properly reflected when calling `Site::is_local_name`, `Site::is_neighbor_name`, and when using `Site::local_nets_table` to find the CIDR a local address is a member of.

Closes #2319


awelzel (changes requested):
Thanks! Two comments that hopefully are reasonable.

Mind updating the `plugins.hooks` baseline, too?

awelzel (approved):
Seems good to me.

Issue #2000, opened 2022-03-14
notice.log unconditionally contains email_dest


awelzel:
This came up previously in internal discussions, but creating an issue for tracking.

Since Zeek 4.1, the notice.log unconditionally contains an `email_dest` field. Using the AsciiWriter with JSON, this is rendered as `email_dest: []` into every entry.

We're deploying Zeek on an appliance where email is not supported. The `mail_dest` field is always empty and the `email_dest` field in notice.log has no meaning in this scenario.

Some reasons are unnecessary extra bytes in the output, extra documentation needed for this new field and educating users what the `email_dest` field is for - with the unsatisfactory answer of "nothing/unused".

We'll work around it by removing it with `$exclude`.

Mostly wanted to raise that the field could've been opt-in rather than unconditionally added to the notice.log, particularly if it's unused / has no meaning / will always be empty in a default configuration. Is it too late now?

Thoughts?

0xxon:
Making the field &optional would remove it from the json output too, right?

Might that be a good enough solution?

Apart from that - are the extra bytes in a low-traffic log (which notice.log should be) really a huge problem? There always will be cases where a field that we include in a log won't do anything in a specific configuration/environment. We typically err on the side of including more than including less information.

awelzel:
`&optional` would help. There is now the new `LogAscii::json_include_unset_fields`, so that would bring it back, but not for everyone.

I believe it'll still show up with the batch logs as a column (?). As an aside, that caused a bit of churn to update btests of packages baseline notice.log - we have a lot of those. We've added `zeek-cut` for pre-processing - and that was probably a good step.
 
> Apart from that - are the extra bytes in a low-traffic log (which notice.log should be) really a huge problem?

I have no numbers and it's probably not in practice.

Issue #1742, opened 2021-08-23
Provide a Zeek log schema


ckreibich:
We regularly encounter the need for a given Zeek version/installation to answer the question "What do the logs look like?". Several folks have built their own takes on this in the past, addressing their own requirements, but we have nothing official. It seems that at a minimum...

- It should capture the full set of log types available, the type, name and meaning of each field in that log, as well as their Zeek-level attributes
- It should get reported in a way that's easy to post-process (making JSON a likely contender)
- It needs to be able to capture a specific Zeek installation so that package-specific additions, redefs, etc are accounted for
- It should enable diffing so for example the differences in logs from one version to the next are captured

There's related work:
- Jon's https://github.com/jsiwek/zeek-print-log-info
- [JSON Schema](https://json-schema.org/) provides a way to describe datasets. It won't capture the nuances of Zeek log semantics but offers a starting point we could build on
- Several solutions exist for diffing JSON

This is just a starting point. It'd be great to hear from folks who've wanted this in the past, to understand their requirements.

stevesmoot:
Not immediately addressed by the JSON request - would like to filter and create in markup a la the docs e.g.,
`conn
----
.. list-table::
   :header-rows: 1
   :class: longtable
   :widths: 1 3

   * - Field (Type)
     - Description

   * - ``ts`` (time)
     - This is the time of the first packet.`
(putting markup in markup, may not come through right, but looks ok here in editor).

anthonykasza:
The idea of Zeek generating its own log schemas came from an internal conversation at Corelight.  We thought it would be neat and useful to have Zeek generate its own schema at runtime since log formats and names are applied in a layering fashion and plugins/packages may alter formats.

As all log stream activation happens within `zeek_init` events, one could set a very low priority `zeek_init` event which prints all streams it knows about. Props to @sethhall for the below script.

```
function describe_record(outer_field: string, rf: record_field_table): string_vec
	{
	local lines = string_vec();
	for ( field, meta in rf )
		{
		if ( meta$log )
			{
			if ( /^record / in meta$type_name )
				{
				local record_type = split_string(meta$type_name, / /)[1];
				local new_lines = describe_record(field, record_fields(record_type));
				for ( i in new_lines )
					{
					lines += new_lines[i];
					}
				}
			else
				{
					{
					local field_prefix = outer_field == "" ? "" : cat(outer_field, ".");
[...]

PR #2322, opened 2022-08-05
RFC: Support redef'ing of record field attributes

awelzel:
I'm not sure this is very useful outside of the &log attribute, but for
that it would allow users to remove &log from `Notice::Info$email_dest`
 without impacting any of the existing logic. Or, `&log` could be removed
from the base scripts and users can opt-in logging by redef'ing the field
with the `&log` attribute.

Relates to #2000.


0xxon:
Sounds great to me - the ability to add/remove attributes is something that has been missing for too long; the syntax looks fine to me.

awelzel:
>  Sounds great to me - the ability to add/remove attributes is something that has been missing for too long; the syntax looks fine to me.

Thanks for looking! I had chatted a bit with @rsmmr and except for the `&log` attribute the others may be eye-brow raising questionable: Adding and more so removing `&default` or `&optional`? Changing the default value from a  `&default` would also be quite special.

I'll respin this with just supporting `&log` and erring for the others at this point.  There was mentioning that `&log` was always special in that it's not part of the language, but belongs more with the logging framework. Allowing more attributes later on is easier than removing it again.

The use case of adding `&log` to existing fields in `Info` records isn't as prevalent (most fields in Info are `&log` anyway), but there could be a new model of adding columns that aren't `&log` by default to Info record types. These can then be opted-in in via `redef += { &log }` rather than adding a `&log` into a base scripts for maybe a minority of users.

Possible other scenario might be 3rd-party packages where the author added tons of columns to a log, but one would like to only select a handful without patching the 3rd-party package.

Long story short: `&log` only for now :-)

rsmmr:
I like the use cases that this enables (the `+=` even more than than the `-=`: as Arne wrote, that would allow us leaving new fields out of the logs by default if value is small, but still enable people to get them in there if they wanted). 

I somewhat torn on the mechanism because we're extending the language feature just for these cases. At the same time, I can't think of a better way to achieve this either, so I'm fine with it. On the plus side, we could extend this later to similar syntax for modifying attributes in other contexts than just record fields (e.g., modify expire timeout on tables).

So:

> Long story short: &log only for now :-) 

👍 from me

J-Gras:
> I somewhat torn on the mechanism because we're extending the language here just for these cases. At the same time, I can't think of a better way to achieve this either, so I'm fine with it.

If it is just about logging, maybe [`Log::Filter`](https://docs.zeek.org/en/master/scripts/base/frameworks/logging/main.zeek.html#type-Log::Filter) are an option. They already support to exclude fields. Slightly adapting how `include` works or introducing something like `additional_fields` might work to support the discussed use cases.

awelzel:
> If it is just about logging, maybe [`Log::Filter`](https://docs.zeek.org/en/master/scripts/base/frameworks/logging/main.zeek.html#type-Log::Filter) are an option. They already support to exclude fields. Slightly adapting how `include` works or introducing something like `additional_fields` might work to support the discussed use cases.

Yep. `$exclude` is in fact the workaround for #2000 on two independent Zeek distributions. It's a bit unfortunate, however, because filters can be created dynamically and each have individual `$exclude` / `$include` settings. Depending on situation (think user provided scripts creating filters) this needs to be considered individually for each filter instantiation.

Somewhat related, `ssl/files.zeek` supports redef's to "include" certificate subjects back into the ssl.log. It's implemented by adding those fields to the `$exclude` set of the default filter:

https://github.com/zeek/zeek/blob/master/scripts/base/protocols/ssl/files.zeek#L108-L123

There is a limitation here that those options are only in effect for the default filter. If you create filters manually, the options are not meaningful unless the same logic is implemented again (or the default filter is cloned). Adding and removing `&log`  for the fields in question gives an alternative approach. E.g. could remove &log from the base scripts and if someone really wants them, ask them to redef them with `&log` again. Whether that's better, not sure, but it's an alternative.

There's also the topic of log schema generation #1742 - if implemented by traversing records looking for `&log` attributes, the downstream filtering on the filters wouldn't be visible and the produced schema may have fields that aren't actually part of the log.

Thanks for your thoughts. But think I'll go ahead.

rsmmr (approved):

PR #2325, opened 2022-08-08
telemetry: Switch to histogram family bucket/bound accessors

awelzel:
Instead of determining the bucket bounds from a concrete metric instance,
broker now provides accessors via the concrete histogram family handles.

Broker side:
https://github.com/zeek/broker/pull/263

@Neverlord cc - let me know if you have any suggestions to improving the code. ~As in DM, this currently excludes the implicit Inf bucket. I'm torn if that's better or worse. It may be better in that for creating the histogram metrics, the Inf bucket isn't specified so including it may be surprising.~ (The Inf bucket is now included as before).

### DIFF
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -76,6 +76,8 @@ zeek::RecordValPtr Manager::GetMetricOptsRecord(Manager::MetricType metric_type,
                                                 const broker::telemetry::metric_family_hdl* family)
 	{
 	static auto string_vec_type = zeek::id::find_type<zeek::VectorType>("string_vec");
+	static auto double_vec_type = zeek::id::find_type<zeek::VectorType>("double_vec");
+	static auto count_vec_type = zeek::id::find_type<zeek::VectorType>("index_vec");
 	static auto metric_opts_type = zeek::id::find_type<zeek::RecordType>("Telemetry::MetricOpts");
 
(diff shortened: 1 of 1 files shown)


Neverlord:
@awelzel thanks for putting this together so quickly! I was about to ask you to try the new functions. :)

> this currently excludes the implicit Inf bucket. I'm torn if that's better or worse.

That's an implementation detail that shines through - and that I just haven't considered yet. The family basically stores the configuration, which hasn't the implicit 'infinite' bucket. That's a detail of the CAF implementation, though. Broker has a similar interface for accessing the buckets. I'd say tweak the Broker side so that they behave the same, ack?

awelzel:
> Broker has a similar interface for accessing the buckets. I'd say tweak the Broker side so that they behave the same, ack?
I'm not fully clear if that means Inf will be included or not, but I don't have a strong opinion either way, so whatever you feel makes more sense.

Neverlord:
@awelzel I've updated the Broker PR. Families and instances now both include the implicit 'infinite' bucket at the end. I think the API makes more sense this way. Let me know if you're ok with the changes and whether everything works on the Zeek end. I'll merge the Broker PR if you give me green light to do so. 🙂

awelzel:
>  I'll merge the Broker PR if you give me green light to do so. slightly_smiling_face
@Neverlord - good to go. The Inf bucket now shows up again.

rsmmr (approved):

PR #2327, opened 2022-08-08
Fix module-scoped type definitions that conflict with existing global ones

### LINKED ISSUE #1443: Scope qualifiers fail for types with the same base name
The following code snippet:
```
type r: record { a: count; };

module test;

type r: record { b: string; };

event zeek_init()
        {
        local x: GLOBAL::r;
        x$a = 5;
        }
```
produces `line 14: no such field in record (test2::x$a)`.  The same thing happens if instead of global scope, the first `r` has scope in another module.

The core problem appears to be that the logic for the parser's `global_or_event_id` production reuses an existing global lookup without considering its associated module scoping.

timwoj:
This makes handling of type definitions at the global scope the same as for constants, global variables, options, etc. This fixes a type clash that can occur when a type is defined in the global module, but then a type is defined at global scope within a module with the same name.

Fixes #1443


0xxon (approved):
Looks good to me - but I am not the greatest expert at the parsing code.

Issue #1678, opened 2021-07-15
Coordinating enable/disable analyzers between multiple scripts


anthonykasza:
Zeek's default base scripts currently disable analyzers, for protocols which support encryption, after the protocol's handshake and once a connection begins using encryption. Module namespaces which do this include `SSL`, `SSH`, and `RDP`. These namespaces each export a boolean option named `disable_analyzer_after_detection` which controls some logic that wraps a call to the [disable_analyzer](https://docs.zeek.org/en/master/scripts/base/frameworks/analyzer/main.zeek.html#id-Analyzer::disable_analyzer) bif. These options are essentially a global setting for _all_ connections of that protocol type. Either all encrypted messages of a connection are processed or none of the encrypted messages of a connection are processed.

More granular control of analyzers would be useful. One use case is scripts which rely on events generated from encrypted messages, such as the [ssl_encrypted_data](https://docs.zeek.org/en/master/scripts/base/bif/plugins/Zeek_SSL.events.bif.zeek.html#id-ssl_encrypted_data) event. One can imagine a script which doesn't want the SSL analyzer to be disabled until some number of `ssl_encrypted_data` events are raised.

One way to enable more granular control would be to change the `disable_analyzer_after_detection` options from `bool` types to `count` types. The name of the variable should likely be renamed to something like `registered_analyzer_uses` or similar. The default value of the counter could be zero. New scripts could then "register" themselves by incrementing that counter's value. Upon satisfying some condition, new scripts would then need to decrement the counter's value to "unregister" themselves. Existing logic in the base scripts would need to be changed to ensure a namespace's counter equals zero before calling `disable_analyzer` for a particular connection.

ckreibich:
It's interesting to compare this with the file analysis framework's analyzers — there, you have to explicitly add analyzers, specifying which event handler you want invoked. You can later remove such analyzers, and the involved state allows you to do this selectively:

```
event file_stream_event_1(f: fa_file, data: string)
        {
        print fmt("file_stream_event_1: File %s, content '%s'", f$id, data[:10]);
        Files::remove_analyzer(f, Files::ANALYZER_DATA_EVENT, [$stream_event=file_stream_event_1]);
        }

event file_stream_event_2(f: fa_file, data: string)
        {
        print fmt("file_stream_event_2: File %s, content '%s'", f$id, data[:10]);
        }

event file_new(f: fa_file)
        {
[...]

Issue #1593, opened 2021-06-10
Composability issues with disable_analyzer()


vpax:
Background: the `disable_analyzer()` BiF allows a script to turn off a particular analyzer associated with a given connection identifier. This is useful for situations like when a script sets `SSL::disable_analyzer_after_detection` to `F` so it can process further into an SSL connection, and then does enough of that to determine it has no more work and wants to skip the load of processing the rest of the connection.

Problem: if two separate scripts both want this functionality, they don't compose well together.  Each will set `SSL::disable_analyzer_after_detection=F`, but then when the first one calls `disable_analyzer()` it will deactivate the analyzer even though the second still wants to continue processing.

Possible fix: add an API for scripts registering interest in a particular analyzer continuing, and disinterest once they're done.  So the above sequence might instead look like: `SSL::register_for_analysis_after_detection()` (a new function that internally just bumps a counter of how many interested parties their are, and sets `SSL::disable_analyzer_after_detection=F`), and then, for a given connection, something like `SSL::release_analysis_interest()` (which internally decrements the current usage counter, and calls `disable_analyzer()` if it has gone to zero).

It may be that this problem exists for other offloading scenarios, too.

rsmmr:
Yeah, I see how this would be useful for SSL. Not sure it generalizes beyond that as the `disable_analyzer_after_detection` is pretty specific to SSL. So sounds like something to just build into the SSL scripts.

awelzel:
Seems #1593 is a duplicate of this. Should we combine them somehow?

PR #2328, opened 2022-08-09
Introduce global disabling_analyzer() hook to veto disable_analyzer()

awelzel:
This hook can be used to coordinate disabling an analyzer for a given
connection. The contract is simple: Any script can veto a disable_analyzer()
call by breaking from this hook. The decision is local to the script taking
into account any state attached to the connection object or script specific
state stored elsewhere.
A script breaking from the hook takes over the responsibility to call
disable_analyzer() at a later point when it finds the condition due to which
it vetoed fulfilled (which may be never).

Signature:

    disabling_analyzer: hook(c: connection, atype: AllAnalyzers::Tag, aid: count);

Example use-cases are keeping the SSL analyzer enabled for finger-printing
until a certain amount of bytes or packets have been transferred or
similarly the connection duration exceed a certain threshold.

Other example use-cases might be keeping analyzers for SSH, RDP or SSL
enabled for connections from specific subnets.

It's a bit quirky as it makes disable_analyzer() a maybe operation. While log
policy hooks and/or the notice hook have similar semantics, they are not as
stateful. It still seems like a quite powerful primitive.

The disable_analyzer() call in dpd/main.zeek may motivate the addition of a
force flag as a follow-up for situations where the caller "knows better" or
absolutely wants to override.

Closes https://github.com/zeek/zeek/issues/1678 https://github.com/zeek/zeek/issues/1593.

---
Looking for comments on the approach, naming and semantics.


awelzel:
@vpax - you had opened #1593 originally and @anthonykasza #1678. I believe they are very related. Would you (or someone working on the involved scripts) mind looking through the proposal and see if that could work out?

awelzel:
> @awelzel Yes, this looks handy for the use cases we have. One thing a bit peculiar will be that if the analyzer is responsible for later calling `disable_analyzer()`, its hook will presumably _again_ be called, which means it may need to squirrel away some state just so it knows to decline that second call. I'm not sure there's a clean fix for that, though.

Thanks!

Yes, a script calling disable_analyzer() itself will have it's own hook invoked again. It then should allow (not break) rather than decline the second call though. But either way, think the pattern would be to use state on `connection` or in a table for declining and if there's none or state indicates it's okay to disable, don't break.

scripts/base/init-bare.zeek
```diff
@@ -584,6 +584,22 @@ type fa_metadata: record {
 	inferred: bool &default=T;
 };
 
+## A hook taking a connection, analyzer tag and analyzer id that can be
+## used to veto disabling analyzers. This hook is invoked synchronously
+## during a :zeek:see:`disable_analyzer` call.
+##
```
rsmmr, line 590: Worth adding an explicit note here on *how* to veto: by breaking.

rsmmr (approved):
Looks great. Just one comment nit, and please add a `NEWS` entry.

rsmmr (approved):

PR #2329, opened 2022-08-09
Use ID name (including module name) to create EventExpr when possible

### LINKED ISSUE #163: Forward declarations of events don't work inside a module namespace
Moved from https://bro-tracker.atlassian.net/browse/BIT-71

Created by Robin Sommer at 2009-02-20T13:43:57.000-0600:
> Forward declarations of events aren't correctly resolved when used inside a module namespace. The worst thing about this is that they fail silently: there's no error message, the handler is just not executed. 

The example below never prints anything. Once the module statement is removed, everything works fine though.

```
module Foo;

global bar: event();

event bar()
    {
    print "bar";
    }

    {
    event bar();
    }
```

Comment by Jon Siwek at 2015-03-16T14:38:28.976-0500:
> Other examples of the problem referenced in [BIT-984](https://bro-tracker.atlassian.net/browse/BIT-984).

Comment by Robin Sommer at 2016-07-08T18:46:57.988-0500:
> I don't see an easy fix for this. Unfortunately event IDs are handled somewhat differently than other IDs, but the global prototype cannot adjust to that because the parser doesn't know that it's an event before it's too late.
> 
> Closing, as there's a work-around by using fully scoped event names (see [BIT-984](https://bro-tracker.atlassian.net/browse/BIT-984)).

timwoj:
I'm not entirely sure about this one, but it fixes the problem reported in #163. The problem was that we're passing just unscoped event name into the `EventExpr` constructor, so if the event doesn't exist with just that name, it fails to find it. If you were to change the call to the event to include the module name (so `Foo::Bar` instead of `bar`) it worked fine. This does essentially the same thing. If we were able to find the event ID with the module name attached, use that. Otherwise, fall back to just using the event name by itself.

Fixes #163

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -1936,9 +1936,13 @@ event:
 
 				if ( id->IsDeprecated() )
 					reporter->Warning("%s", id->GetDeprecationWarning().c_str());
-				}
 
-			$$ = new EventExpr($1, {AdoptRef{}, $3});
+				$$ = new EventExpr(id->Name(), {AdoptRef{}, $3});
+				}
+			else
+				{
+				$$ = new EventExpr($1, {AdoptRef{}, $3});
+				}
 			}
 	;
 
--- a/testing/btest/Baseline/scripts.policy.misc.capture-loss/capture_loss.log
+++ b/testing/btest/Baseline/scripts.policy.misc.capture-loss/capture_loss.log
@@ -8,4 +8,5 @@
 #fields	ts	ts_delta	peer	gaps	acks	percent_lost
 #types	time	interval	string	count	count	double
 XXXXXXXXXX.XXXXXX	0.000000	zeek	0	0	0.0
+XXXXXXXXXX.XXXXXX	0.000000	zeek	0	0	0.0
 #close XXXX-XX-XX-XX-XX-XX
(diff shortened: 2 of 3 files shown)


timwoj:
Hm, btests were all passing on my local machine. I must have missed something.

rsmmr (approved):
Seems reasonable to me (assuming tests pass). One change I'd suggest is do the scope lookup only if the ID isn't scoped to begin with. Probably won't make a real difference, but seems cleaner.

timwoj:
> One change I'd suggest is do the scope lookup only if the ID isn't scoped to begin with.

The lookup has to happen regardless because it needs the deprecation information and such.

timwoj:
Interestingly, the same test failure occurs if I make this change to the test script:

```
diff --git a/testing/btest/scripts/policy/misc/capture-loss.zeek b/testing/btest/scripts/policy/misc/capture-loss.zeek
index d51576a5b..a6ce0e571 100644
--- a/testing/btest/scripts/policy/misc/capture-loss.zeek
+++ b/testing/btest/scripts/policy/misc/capture-loss.zeek
@@ -8,5 +8,5 @@ module CaptureLoss;
 
 event zeek_init()
        {
-       event take_measurement(network_time(), 0, 0);
+       event CaptureLoss::take_measurement(network_time(), 0, 0);
        }
```

I'm not entirely sure why yet, but it ends up calling `take_measurement` twice , with the second one probably coming from the `schedule` statement at the end of `take_measurement`.

PR #2335, opened 2022-08-11
files.log: Unroll and introduce uid and id fields

awelzel:
This is a script-only change that unrolls File::Info records into
multiple files.log entries if the same file was seen over different
connections by single worker. Consequently, the File::Info record
gets the commonly used uid and id fields added. These fields are
optional for File::Info - a file may be analyzed without relation
to a network connection (e.g by using Input::add_analysis()).

The existing tx_hosts, rx_hosts and conn_uids fields of Files::Info
are not meaningful after this change and removed by default. Therefore,
files.log will have them removed, too.

The tx_hosts, rx_hosts and conn_uids fields can be revived by using the
policy script frameworks/files/deprecated-txhosts-rxhosts-connuids.zeek
included in the distribution. However, with v6.1 this script will be
removed.

---

The following PRs from the external repos go with this one:

https://github.com/zeek/zeek-testing/pull/9
https://github.com/zeek/zeek-testing-private/pull/7

---

References to public postings of this step:

https://zeekorg.slack.com/archives/CTGDG83B8/p1659963508344319

https://community.zeek.org/t/changing-files-log-unrolling-and-removal-of-rx-hosts-tx-hosts-and-conn-uids/6644

https://docs.google.com/document/d/1M2QNbofW__MeDNHbFTqpy8CdpWXZrjB4D4OiUB-fJ3U/edit#


scripts/base/frameworks/files/main.zeek
```diff
@@ -572,5 +563,28 @@ event file_state_remove(f: fa_file) &priority=10
...
+	for ( [cid], c in f$conns )
+		{
+		# Make a copy of the record when there's more than one
+		# connection so that the log_files event doesn't see
+		# the same record multiple times due to it being queued
+		# by reference in Log::write() rather than by copy.
+		local info = |f$conns| > 1 ? copy(f$info) : f$info;
+		info$uid = c$uid;
+		info$id = cid;
+		Log::write(Files::LOG, info);
```
rsmmr, line 588: A semantic question: by just copying the other fields, seems we can end up with information inside a single entry that comes from multiple connections. For example, for the range requests, we still see the aggregate size information. I don't think there's a way around that with a script-only implementation, it would require C++ side changes to fix that. I'm wondering if that's something to just accept for now until we make internal changes, or if it's too misleading/confusing.

rsmmr (commented):
Looks good to me, with one caveat about semantics of the unrolling. Nice tests, too. Let's wait a couple more days if anybody has further feedback on this change.

scripts/base/frameworks/files/main.zeek
awelzel: > I don't think there's a way around that with a script-only implementation, it would require C++ side changes to fix that. I'm wondering if that's something to just accept for now until we make internal changes, or if it's too misleading/confusing.

That is a good point I hadn't quite considered. The existing documentation for the `_bytes` fields are in relation to file rather than the connection, so could frame it as "redundant / replicated" information and accept it for now.

After a bit of pondering: I believe it's mostly `seen_bytes` and `missing_bytes` that should be tracked per connection. `total_bytes` should be the same anyway and `overflow_bytes` is coming from the `FileReassembler` where there's only a single one (per worker).

I can give tracking seen/missed bytes per connection a spin, I'm not sure what it would look like, but doesn't seem it should be too hard. Yes/no?

rsmmr: Well, the thing is that we'd still remain inconsistent as long as *some* pieces are coming from multiple connections: at the very least, that won't yield consistent results across multiple workers; which is one thing we were hoping to fix here. To do that, I think we'd need to completely separate file state by connection. 

But not saying we should go there, nor that this is a blocker; I think we can indeed do a lot with documentation here.  

Turning just some fields into per-connection values seems confusing, too, though. And if, say, `seen_bytes` now is per-connection, we don't have the total seen value anymore either.

I think I'm in favor on going with this as is for now, just making sure the documentation is clear what's going on. That'll keep current semantics while still addressing the original problem of `files.log` being hard to parse externally.

awelzel:
@rsmmr - I've rebased on top of master and removed the per-connection tracking. External test suite PRs in description. Ready for another look. Thanks!

rsmmr (approved):

PR #2336, opened 2022-08-12
Make result of IP::ParsePacket easier to understand

timwoj:
This also removes the error message from ParsePacket if the proto value isn't one of IPPROTO_IPV4 or IPPROTO_IPV6. This can happen in cases where the data inside of a tunnel is encrypted, such as with Aruba tunnels containing IEEE 802.11 QoS CCMP packets. The log message in these cases becomes very spammy and fills up reporter.log since it isn't rate-limited at all.

The problem with the log messages has been reported by a couple of different users now. The changes for the result value just make reasoning about them easier. Unfortunately, there's no way to overload based on the return value alone, but `ParsePacket` is only used by tunnel analyzers and when we did the tunnel rework we determined that it was unlikely used by anyone else in the wild. I added a note to the breaking changes section that the method signature changed since it causes the build to fail.

Alternatively, I could add a second method with a different name that the first one calls. That would let me deprecate the first version. I could go either way on this.

### DIFF
--- a/src/packet_analysis/protocol/teredo/Teredo.cc
+++ b/src/packet_analysis/protocol/teredo/Teredo.cc
@@ -192,8 +192,8 @@ bool TeredoAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pack
 	// but on the other hand we duplicate this work here. maybe this header could just be stored
 	// and reused in the IP analyzer somehow?
 	std::shared_ptr<IP_Hdr> inner = nullptr;
-	int rslt = packet_analysis::IP::ParsePacket(len, te.InnerIP(), IPPROTO_IPV6, inner);
-	if ( rslt > 0 )
+	auto result = packet_analysis::IP::ParsePacket(len, te.InnerIP(), IPPROTO_IPV6, inner);
+	if ( result == packet_analysis::IP::ParseResult::CaplenTooLarge )
 		{
 		if ( inner->NextProto() == IPPROTO_NONE && inner->PayloadLen() == 0 )
 			// Teredo bubbles having data after IPv6 header isn't strictly a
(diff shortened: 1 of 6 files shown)


timwoj:
This might be a candidate for 5.0.1, as I've now heard from three different users that are running into this.

src/packet_analysis/protocol/teredo/Teredo.cc
```diff
@@ -192,8 +192,8 @@ bool TeredoAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pack
 	// but on the other hand we duplicate this work here. maybe this header could just be stored
 	// and reused in the IP analyzer somehow?
 	std::shared_ptr<IP_Hdr> inner = nullptr;
-	int rslt = packet_analysis::IP::ParsePacket(len, te.InnerIP(), IPPROTO_IPV6, inner);
-	if ( rslt > 0 )
+	auto rslt = packet_analysis::IP::ParsePacket(len, te.InnerIP(), IPPROTO_IPV6, inner);
```
awelzel, line 195: Unrelated: The other two users use `result` instead of `rslt` - worth expanding?

There are some later variables `encap_index` and `inner_packet` that aren't missing characters, so would als fit better here I guess?

awelzel (commented):
Not sure I'm qualified to approve, but lgtm. One unrelated nit.

Zeek 5.0.1 sounds good, too.

rsmmr:
I don't like breaking API with a patch release, so if this goes into 5.0.1 I'd prefer some other way. A second method would work, but I'm wondering: could we just use integer constants instead of enums to stay with the new code but without changing the return type? From reading the comments in the code, I believe they were already used like constants anyways, so it wouldn't make a semantic difference; but not sure if I am missing something.

timwoj:
I split this change into two commits. Only the first one should make it into 5.0.1.

rsmmr (approved):

Issue #2339, opened 2022-08-15
PCAP Linux Cooked Capture version 2 unsupported


simeonmiteff:
Zeek supports the `DLT_LINUX_SLL` PCAP link-type but not `DLT_LINUX_SLL2` which has a different header (see https://www.tcpdump.org/linktypes/LINKTYPE_LINUX_SLL2.html).

Here is the result of reading [linux_dlt_sll2.pcap.gz](https://github.com/zeek/zeek/files/9335249/linux_dlt_sll2.pcap.gz):

```console
$ docker run --rm -ti -v /tmp:/tmp zeekurity/zeek:latest /bin/bash
root@4c46efe95c19:/# zeek -r /tmp/linux_dlt_sll2.pcap 
root@4c46efe95c19:/# ls *.log
packet_filter.log  weird.log
root@4c46efe95c19:/# cat weird.log 
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	weird
#open	2022-08-15-05-17-21
#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	name	addl	notice	peer	source
#types	time	string	addr	port	addr	port	string	string	bool	string	string
1660534249.872259	-	-	-	-	-	unknown_ip_version	-	F	zeek	IP
1660535793.578961	-	-	-	-	-	unknown_ip_version	-	F	zeek	IP
#close	2022-08-15-05-17-21
```
tshark is happy to read the file:
```
$ tshark -r /tmp/linux_dlt_sll2.pcap
    1   0.000000    192.0.2.1 → 192.0.2.1    ICMP 104 Echo (ping) request  id=0x0008, seq=1/256, ttl=64
    2   0.000029    192.0.2.1 → 192.0.2.1    ICMP 104 Echo (ping) reply    id=0x0008, seq=1/256, ttl=64 (request in 1)
    3  14.216305 fe80::8c36:6ff:fe44:acaf → fe80::8c36:6ff:fe44:acaf ICMPv6 124 Echo (ping) request id=0x0009, seq=1, hop limit=64
    4  14.216335 fe80::8c36:6ff:fe44:acaf → fe80::8c36:6ff:fe44:acaf ICMPv6 124 Echo (ping) reply id=0x0009, seq=1, hop limit=64 (request in 3)
    5 1543.706702 8e:36:06:44:ac:af →              ARP 48 Who has 192.0.2.2? Tell 192.0.2.1
[...]

Issue #2132, opened 2022-05-25
BPF Filtering fails for certain link types


timwoj:
This is a follow-up to #1995. When loading a pcap with certain link types, libpcap will return an error that looks like this:

```
fatal error in /Users/tim/Desktop/projects/zeek/scripts/base/frameworks/packet-filter/./main.zeek, line 292: cannot compile BPF filter "ip or not ip": pcap_compile(ip or not ip): IEEE 802.15.4 link-layer type filtering not implemented
```

This isn't necessarily an error though, it's just that libpcap only implements link-layer filtering for certain link types. See the `gen_linktype()` method in libpcap at https://github.com/the-tcpdump-group/libpcap/blob/64ed5bbd93520ff63ac735d07ef3514ca0dcdd68/gencode.c#L3190 for a list of what is supported and what isn't. Libpcap doesn't return any sort of error codes or anything about why a filter failed to compile, only an error string. Thankfully that string seems to be fairly stable as of the last 20 years or so (I looked), so we can use the presence of "filtering not implemented" in the error string.

Unfortunately, that's the easy part. We still need to propagate that "failed but not failed" state upwards through the call stack to the original call to `Pcap::precompile_pcap_filter` in `PacketFilter::install` in script-land so it can exit cleanly instead of just assuming that every return value of `false` means a hard failure. On top of that, we need to investigate whether this is something that happens on live traffic as well.

timwoj:
For the second part of this, I'm going to opt not to break the existing API for `PktSrc::PrecompileBPFFilter`. Instead I'm planning to add a method that returns either an enum value or a string for the reason for the failure (I haven't decided which yet). I'm leaning towards a string since that would let third-party packet source plugins implement their own error strings. It'll probably actually need to run a `pair<bool,string>` that indicates whether the error is catastrophic and the reason. Callers of `PrecompileBPFFilter` and friends can check for the bool result, and then call that new method if the result is false and handle it appropriately.

PR #2340, opened 2022-08-15
Add support for DLT_LINUX_SLL2 PCAP link-type

simeonmiteff:
Herewith a fix for #2339 by adding a new packet analyzer that handles the `DLT_LINUX_SLL2` PCAP link-type.

This is basically a copy of the `linux_sll` analyzer. I changed the link layer header struct, the link-type number and the name. I couldn't see an obvious way to pass the link-type to the existing analyzer to make it support both types. Please let me know if I missed something - I'm happy to rework the MR to avoid having two near-identical analyzers if that is possible.

For testing I created a small PCAP file with coverage for the four ethertypes listed in `scripts/base/packet-protocols/linux_sll2/main.zeek` (IPv4, IPv6, ARP and RARP). There is a new `btest` case in `testing/btest/core/linuxsll2.zeek` (is this the correct location?).

There is one deficiency in the test: it doesn't look like zeek's ARP protocol analyzer handles the RARP opcodes, so I don't have an event to test against.

---

Required external PRs:
* CMake for checking SLL2 support in libpcap: https://github.com/zeek/cmake/pull/48


awelzel:
Thanks @simeonmiteff - mind updating the `plugins/hooks` baseline, too?

simeonmiteff:
@awelzel sorry I should have run the whole btest suite first. I've fixed the `plugins/hooks` baseline and then found two others fixed in [fff93d2](https://github.com/zeek/zeek/pull/2340/commits/fff93d23a5a812d238bb95ccb5ec72de35470da7) (`coverage/{default, bare}-load`), both of which seem obviously related to my changes.

However, I'm unable to get all the tests to pass. The following baselines seem out of date:
  - `bifs/x509_verify` (changed certificates)
  - `scripts/base/files/x509/files` (the order of the `analyzers` field changes from `X509,SHA256,SHA1,MD5` to `SHA256,X509,SHA1,MD5`)
  - `scripts/base/frameworks/logging/hooks` (has an extra, perhaps spurious `Log::log_stream_policy` line at the end)

Perhaps this is due to something about the way I'm configuring/building zeek locally. The `scripts/base/files/x509/files` issue above is the one failing in CI (example here: https://github.com/zeek/zeek/pull/2340/checks?check_run_id=7850452244). Do you think this is an expected change in the baseline due to the new analyzer?

awelzel:
> bifs/x509_verify (changed certificates)
> scripts/base/frameworks/logging/hooks 

These two "fail" for me locally, too, but only when run with `btest -u`, as then btest-diff prompts (I was very confused the other day, too). When run without `-u`, they appear to pass - they have a `|| btest-diff` and `|| true` in them. The former I get, the latter I don't quite :-)

> scripts/base/files/x509/files (the order of the analyzers field changes from X509,SHA256,SHA1,MD5 to SHA256,X509,SHA1,MD5)
> Do you think this is an expected change in the baseline due to the new analyzer?

So - it's very unfortunate, but here's my explanation. The new SLL2 analyzer changes the enum/tag value for the X509 analyzer:

Your branch:
```
$ zeek -e 'print enum_to_int(Files::ANALYZER_MD5); print enum_to_int(Files::ANALYZER_SHA1); print enum_to_int(Files::ANALYZER_SHA256); print enum_to_int(Files::ANALYZER_X509)'
24
25
26
113
```

Previous Zeek master:
```
$ /opt/zeek-5.1.0-dev.351/bin/zeek -e 'print enum_to_int(Files::ANALYZER_MD5); print enum_to_int(Files::ANALYZER_SHA1); print enum_to_int(Files::ANALYZER_SHA256); print enum_to_int(Files::ANALYZER_X509)'
24
25
26
112
```

And that make iteration order around `Files::mime_type_by_analyzer` different than before:

x.zeek:
```
@load base/frameworks/files
module Files;

event zeek_init() {
        for ( tag in mime_type_to_analyzers["application/x-x509-ca-cert"] )
                print tag;
}

```

Your branch:
```
$ zeek -G ./random.seed x.zeek 
Files::ANALYZER_SHA1
Files::ANALYZER_X509
Files::ANALYZER_MD5
Files::ANALYZER_SHA256
```

Previous Zeek master:
```
$ /opt/zeek-5.1.0-dev.351/bin/zeek -G ./random.seed x.zeek 
Files::ANALYZER_SHA1
Files::ANALYZER_SHA256
Files::ANALYZER_X509
Files::ANALYZER_MD5
```

And that different order inserts into `set[string]` which then likely perturbs the final order of the log a bit more:

https://github.com/zeek/zeek/blob/f631551ffb8eb581521fa02b78075441c3401013/scripts/base/frameworks/files/main.zeek#L556-L558

https://github.com/zeek/zeek/blob/f631551ffb8eb581521fa02b78075441c3401013/scripts/base/frameworks/files/main.zeek#L421

Because we run with -G it's deterministic, but not very predictable.
So while IMO it's unfortunate the order of values in the log changes due to introduction of a new (packet) analyzer, I think it can be explained. Given `Files::Info$analyzers` is a `set[string]` the order isn't really guaranteed anyway. #2208 may help here (or using an explicit vector + sort for the logged  `analyzers` field would have as well).

If you update the baseline, I think this can go in. Sorry for the wall of text.

ckreibich:
> The former I get, the latter I don't quite :-)

:-)

For `scripts/base/frameworks/logging/hooks` the idea was that the test uses `@TEST-START-NEXT` quite a bit, and not each of the resulting tests actually produces `other.log` or `output`, so the file existence test shouldn't trigger a test failure. If a test that's expected to produce these would fail to produce them altogether, that'd be a bug that these tests won't detect correctly.

simeonmiteff:
> If you update the baseline, I think this can go in. Sorry for the wall of text.

@awelzel no... I love it. Thanks for the detailed response!

I've updated the `scripts/base/files/x509/files` baseline in 0904374

Remaining test failures in the CI:
  - `core/init-error.zeek`: again I'm not sure why this is failing in the asan and ubsan tasks and passing for me :grin: 
  - `core/linuxsll2` (the new test introduced in this PR) failed on centos 7, ubuntu 18 and debian 10. This is because SLL2 support was added in libpcap version 1.10, and all of these distros have older libpcap.

The libpcap issue manifests when zeek calls `pcap_compile()`: `fatal error: problem with trace file /zeek/testing/btest/Traces/linux_dlt_sll2.pcap (cannot compile BPF filter "": pcap_compile(): unknown data link type 276)`

AFAIK this should only happen when zeek built against libpcap < 1.10 reads a `DLT_LINUX_SLL2` PCAP file. The error message is not ideal, but the alternative is that zeek silently ignores the packets in the file. Personally I prefer the latter because (albeit obscure) the message does actually describe the problem and I prefer knowing if zeek couldn't read the PCAP file.

In commit ac850ff I have added a `@TEST-REQUIRES` line to the top of `core/linuxsll2` that checks for the _absence_ of this error in stderr, which causes btest to skip the test on builds against old libpcap.

ckreibich:
Following up on my earlier comment: oh dear — the `|| true` of course masks the whole thing either way. That's a bug. I'll make a pass over the `scripts.base.frameworks.logging.hooks` test.

awelzel:
> core/linuxsll2 (the new test introduced in this PR) failed on centos 7, ubuntu 18 and debian 10. This is because SLL2 support was added in libpcap version 1.10, and all of these distros have older libpcap.

Short of fingerprinting those distro's seems reasonable. You could probably shorten the TEST-REQUIRES to something that exits with success on the first raw_packet event (indicating a packet from that pcap made it up into the pipelien) rather than grepping, but might not be obvious on first sight either, so probably worth a comment. 
```
zeek -b -C -r Traces/linux_dlt_sll2.pcap  -e 'event raw_packet(p: raw_pkt_hdr) { exit(0); }' 2>&1 >/dev/null
```

timwoj:
You could always add another flag to zeek-config.h that denotes whether a system has a libpcap with support for it. There's a couple other tests that check for existence of flags in that file as well.

simeonmiteff:
> I clicked retry for those and they failed again. After looking, I think it's fair game to do the following. "1st event" and "2nd event" have the same `&priority` and so it's not guaranteed which one executes first. Give the 2nd one the implicit `&priority=0`. Does that fix it?

Yes it does, thanks (added with 43fa9e6)

simeonmiteff:
> You could always add another flag to zeek-config.h that denotes whether a system has a libpcap with support for it. There's a couple other tests that check for existence of flags in that file as well.

Ah this seems more elegant than hacking the test not to run. I'm going to give it a shot...

simeonmiteff:
@timwoj success! I've added the flag and reworked the test to use it. I also use the flag to compile in a test against the link type at PCAP open time. This improves the error message:

Before:

```console
$ zeek -b -C -r Traces/linux_dlt_sll2.pcap core/linuxsll2.zeek                                                        
fatal error: problem with trace file Traces/linux_dlt_sll2.pcap (cannot compile BPF filter "": pcap_compile(): unknown data link type 276)
```
After:
```console
$ zeek -b -C -r Traces/linux_dlt_sll2.pcap core/linuxsll2.zeek 
fatal error: problem with trace file Traces/linux_dlt_sll2.pcap (PCAP file is using DLT_LINUX_SLL2 which is unsupported by your libpcap version)
```

simeonmiteff:
@awelzel I've opened https://github.com/zeek/cmake/pull/48 to for the cmake change that sets the `DONT_HAVE_LIBPCAP_DLT_LINUX_SLL2` flag that is now used by the enhanced error handling and for skipping the btest case on older libpcap (@timwoj's suggestion above).

TBH I'm not clued up about the process for PRs that depend on changes in submodules. I'm happy to wait for https://github.com/zeek/cmake/pull/48 to be merged and then revert the submodule URL so that this PR can be merged without futzing.

All the CI tests are passing now. Please let me know if you have any concerns or other changes you'd like to see.

src/iosource/pcap/Source.cc
```diff
@@ -186,6 +186,13 @@ void PcapSource::OpenOffline()
 		InternalError("OS does not support selectable pcap fd");
 
 	props.link_type = pcap_datalink(pd);
+#ifdef DONT_HAVE_LIBPCAP_DLT_LINUX_SLL2
+        if ( props.link_type == 276 )
```
awelzel, line 190: There seems to be space vs tab confusion here and below.

src/iosource/pcap/Source.cc
```diff
@@ -186,6 +186,13 @@ void PcapSource::OpenOffline()
...
+            {
+		Error("PCAP file is using DLT_LINUX_SLL2 which is unsupported by your libpcap version");
```
awelzel, line 192: Worth an InternalError that also aborts execution?

src/iosource/pcap/Source.cc
bbannier: This is caught by the pre-commit checks which I just manually triggered in CI, see https://github.com/zeek/zeek/wiki/Coding-Style-and-Conventions#user-content-clang-format.

src/iosource/pcap/Source.cc
simeonmiteff: @awelzel yeah `InternalError` was the first thing I did but then I grepped through the codebase and it seems this is used for errors that are truly internal (e.g., consequence of a bug), while this feels more like an "external error" :grin: (i.e., user is asking us to do something unsupported on this system). Please advise if I have this right...

src/iosource/pcap/Source.cc
simeonmiteff: Thanks @bbannier. Fixed in 7724b8e.

BTW, is there a guide somewhere for running the github actions and pre-commit hooks locally?

bbannier: As for the pre-commit GH action, we have the wiki paragraph I linked above (TLDR; `pre-commit run -a` to run all pre-commit hooks on all files in the repo). This automatically downloads e.g., a `clang-format` binary.

For the other GH actions see their definitions in `.github/workflows`.

ckreibich:
> TBH I'm not clued up about the process for PRs that depend on changes in submodules.

@simeonmiteff That's a good point, yeah — I just added a paragraph about this to the contributor docs [here](https://github.com/zeek/zeek/wiki/Contribution-Guide#submitting-contributions). The biggies are (1) mention the related PRs in the comments so we understand they relate, (2) try to have the submodules pointed correctly so the entirety of your contribution can run as one in our CI. The latter can be tricky and we're happy to help. We also make sure that the submodules are set right once we merge the contribution(s).

simeonmiteff:
> > TBH I'm not clued up about the process for PRs that depend on changes in submodules.
> 
> @simeonmiteff That's a good point, yeah — I just added a paragraph about this to the contributor docs [here](https://github.com/zeek/zeek/wiki/Contribution-Guide#submitting-contributions). The biggies are (1) mention the related PRs in the comments so we understand they relate, (2) try to have the submodules pointed correctly so the entirety of your contribution can run as one in our CI. The latter can be tricky and we're happy to help.

Ok thanks, that makes sense and I think I've done it right up to this point :point_up: 

> We also make sure that the submodules are set right once we merge the contribution(s).

Do I have this right?:
1. I wait for https://github.com/zeek/cmake/pull/48 to be merged
2. I push a last commit to my zeek branch, to restore the submodule on from my `zeek/cmake` fork back to the parent, and add a note to this PR to say I've done so
3. You (zeek maintainers) do a final review and merge if you're happy

ckreibich:
> Do I have this right?:

You can do that, but you don't have to. As we merge your PRs, we can track and patch up the submodule pointers. We'll merge the cmake PR first. When we then merge this PR (with the technically invalid submodule pointer) we can straighten out the pointer in the merge commit. There are lots of examples of this in our merge history. I'm generally neutral on merging vs rebasing, but the fact that the merge commit can add tweaks on top of the "raw" PR is quite handy.

simeonmiteff:
> > Do I have this right?:
> 
> You can do that, but you don't have to. As we merge your PRs, we can track and patch up the submodule pointers. We'll merge the cmake PR first. When we then merge this PR (with the technically invalid submodule pointer) we can straighten out the pointer in the merge commit. There are lots of examples of this in our merge history. I'm generally neutral on merging vs rebasing, but the fact that the merge commit can add tweaks on top of the "raw" PR is quite handy.

Ok cool, thanks for explaining. I'll leave it to you (or whoever merges this PR) then - less synchronisation seems better :+1:

awelzel (approved):
I think that's good to go now.

src/iosource/pcap/Source.cc
```diff
@@ -186,6 +186,13 @@ void PcapSource::OpenOffline()
 		InternalError("OS does not support selectable pcap fd");
 
 	props.link_type = pcap_datalink(pd);
+#ifdef DONT_HAVE_LIBPCAP_DLT_LINUX_SLL2
+	if ( props.link_type == 276 )
+		{
+		Error("PCAP file is using DLT_LINUX_SLL2 which is unsupported by your libpcap version");
```
timwoj, line 192: I'm a little concerned about reporting an error here sinc reporter messages aren't rate-limited it at all. If you're getting a lot of these packets on an unsupported system your reporter log will explode. I don't think the previous weird of `unknown_ip_version` is correct here though. I wonder if we should just silently return instead, but that doesn't feel right either.

What does libpcap do if it gets one of these and it doesn't support the link type? Is the packet object not well-formed? Could we not just let the processing flow happen and let the SLL2 analyzer check it?

timwoj (changes requested):

src/iosource/pcap/Source.cc
awelzel: I haven't tested, but this is in `PcapSource::OpenOffline()` rather than on a per-packet basis and we skip the `Opened()` callback. Wouldn't we just run into this and exit then?

https://github.com/zeek/zeek/blob/f5df2a132a0cf6e6601404968a9977b50a9685e1/src/RunState.cc#L157-L159

simeonmiteff: @awelzel correct, it just exits while opening the PCAP source. My comment above: https://github.com/zeek/zeek/pull/2340#issuecomment-1219440742 shows the before/after behaviour.

@timwoj interestingly libpcap < 1.10.0 is blissfully ignorant about the link type, and if all you do is this, it works:
```c
char errbuf[PCAP_ERRBUF_SIZE];
struct pcap_pkthdr *header;
const u_char *pkt_data;

fp = pcap_open_offline(filename, errbuf);

while ( pcap_next_ex( fp, &header, &pkt_data) >= 0 ) {
    // do something with the packet
}
``` 

Specifically, `pkt_data` contains the full packet including link layer (which zeek would be able to handle with the new analyzer).

However if you try this:
```c
struct bpf_program program;

int res == pcap_compile(fp, &program, "ip or not ip", 0, PCAP_NETMASK_UNKNOWN);
if (res != 0) {
    printf("Error compiling filter [%d]: %s\n", res, pcap_geterr(fp));
}
```
...you get `Error compiling filter [-2]: unknown data link type 276`, which is exactly what happens in zeek. I thought about conditionally disabling the BPF code but that seems like a case of diminishing returns for the extra code complexity.

Since zeek would exit anyway due to the BPF error, adding this check improves the error message and handles the condition in the most logical place (perhaps preventing future surprises if the BPF code changes).

timwoj: I'm hoping to fix the BPF error in a more generic fashion via https://github.com/zeek/zeek/issues/2132, probably before 5.1 comes out.

simeonmiteff: @timwoj ok, I see. Please let me know if you'd like me to remove this then.

timwoj: I'd say remove it. If it turns out to be a big problem, we can always add it back in. I just don't want to end up special-casing a bunch of link types in `Packet`, if possible.

timwoj (approved):

PR #2341, opened 2022-08-15
Fix a swath of AUTO_CAUSES_COPY issues reported by Coverity

timwoj:


awelzel (approved):
Without much knowledge if it makes a huge difference, approving.

PR #2345, opened 2022-08-16
Support other byte lengths in bytestring_to_count

### LINKED ISSUE #695: bytestring_to_count handle different byte lens
https://github.com/zeek/zeek/blob/e5db1f085c034451d8b1068472554146ab103619/src/zeek.bif#L2667

Please implement inputs for bytestring with length of 3, 5, 6, 7.  This is useful for embedded devices that optimizes for space as use as minimum as needed.

timwoj:
I realize this code is a little repetitive, but it uses different value sizes in each block to make things clearer.

Fixes #695


src/zeek.bif
```diff
@@ -2961,26 +2961,32 @@ function bytestring_to_count%(s: string, is_le: bool &default=F%): count
...
+		}
+	if ( s->Len() == sizeof(uint8_t) )
 		{
 		uint8_t value = 0;
 		memcpy(&value, p, sizeof(uint8_t));
 		return zeek::val_mgr->Count(value);
 		}
-
-	case sizeof(uint16_t):
+	else if ( s->Len() <= sizeof(uint16_t) )
```
ckreibich, line 2976: This is always going to be a `sizeof(uint16_t)`, no? (Because otherwise the above if-clause will kick in?)

src/zeek.bif
```diff
@@ -2961,26 +2961,32 @@ function bytestring_to_count%(s: string, is_le: bool &default=F%): count
...
 		uint16_t value = 0;
 
 		if ( (host_bigendian && is_le) || (! host_bigendian && ! is_le) )
 			{
 			char buf[sizeof(uint16_t)];
+			memset(buf, 0, sizeof(uint16_t));
 			char *d = &buf[sizeof(uint16_t)-1];
 
-			for ( i = 0; i < sizeof(uint16_t); i++ )
+			if ( is_le )
```
ckreibich, line 2986: I'm struggling with these `is_le` checks here and below — is it intentional that here it's `if ( is_le )` and below it's `if ( ! is_le )`?

src/zeek.bif
timwoj: Good point. I'll just make that `==`.

timwoj: That also means that block doesn't need to change at all. I'll remove all of those changes.

src/zeek.bif
timwoj: See above about removing the changes on this block 😃

ckreibich (approved):

Issue #2289, opened 2022-07-19
Conditional parsing leads to subtle bugs


vpax:
`@conditional`'s are treated as statements in Zeek's grammar.  (I imagine they're parsed, rather than dealt with by the lexer, because they can include expressions that require parsing.)  This leads to violations of the Principle Of Least Surprise, such as:

> event zeek_init()
        {
@if ( T )
        if ( F )
@else
        if ( F )
@endif
                print "There's no way this should happen";
        }

What happens here is that the `@else` gets treated as the conditional statement of the first `if`, and so the `print` statement comes after the entire `if`, and thus always executes.

_Maybe_ this can be fixed by introducing a grammar nonterminal that's allows for a conditional _followed_ by a statement (as a unit).  But I haven't thought that fully through.

PR #2347, opened 2022-08-17
parse.y/directives: Reject directives as statements 

awelzel:
Avoid the issue outlined in #2289 where the @if or @else is taken
as the statement for an `if` or `for`by changing the grammar such
that @if, @ifdef, @else, @endif can only occur within stmt_list.

Effectively this means it's not possible to do the following anymore
and instead now cause syntax errors:

    # Print's "cond true" with Zeek 5.0 even though the `if ( F )`
    # should be in effect.

    if ( F )
        @if ( cond )
            print "cond true";
        @else
            print "cond false";
        @endif

or

    # Print's "hello" once with Zeek 5.0
    local v = vector( 1, 2, 3 );

    for ( i in v )
        @if ( T )
        print("hello")
        @endif

To make these work, additional braces are now required.

    if ( T )
        {
    @if ( cond )
            print "cond true";
    @else
            print "cond false";
    @endif
        }

    for ( i in v )
        {
    @if ( T )
        print("hello")
    @endif
        }

That's just one idea to maybe circumvent the problem? Minimally this
commit contains a few test-cases.


awelzel:
@vpax - this PR wouldn't allow the construct from #2289 anymore. I'll stare more at the `.y` file around actually making them work. But wondering whether "not allowing" could be an option or you'd rather see them work.

Curiously, it doesn't trigger any fallout in the existing test suite (which is also good because there might have been subtle bugs then).

vpax:
@awelzel Cool that looks to me like a good solution.

awelzel:
>  @awelzel Cool that looks to me like a good solution.

Okay, thanks. I'm moving this into "ready for review", maybe someone else has thoughts. If someone comes up to support this without explicit braces and expected semantics, we can enable it going forward, but for now plug the hole you showed in #2289 .

awelzel:
Chatted with @rsmmr and he'd prefer to explicitly check reject "NullStatements" representing conditionals for `stmt` rather than changing the where `conditional` can appear. Will rework.

awelzel:
> Chatted with @rsmmr and he'd prefer to explicitly check reject "NullStatements" representing conditionals for stmt rather than changing the where conditional can appear. Will rework.

@rsmmr  - this is ready for another look. I hope I hit close to what we discussed ;-)

src/Stmt.h
```diff
@@ -525,7 +525,7 @@ class InitStmt final : public Stmt
 class NullStmt final : public Stmt
 	{
 public:
-	NullStmt() : Stmt(STMT_NULL) { }
+	NullStmt(StmtTag arg_tag = STMT_NULL) : Stmt(arg_tag) { }
```
rsmmr, line 528: I'd rather not overwrite the tag. The tags are meant to identify the type of the statement, with "type" here corresponding to the class. By (optionally) setting this to something else, we're breaking the association (which could also cause trouble elsewhere, not sure).

 Two alternative suggestions:

1. Use a boolean or enum instead that indicates that this `NullStmt` reflects a directive, without changing its tag.
2. Create a new statement class `Directive` that then can return a corresponding tag (needs a check if other parts of the code now need to know about the new statement).

rsmmr (approved):

PR #2352, opened 2022-08-18
Truncate Docker image artifact at end of Docker CI workflow

ckreibich:
In the (private) zeek-security repo we currently hit the 2GB storage limit
whenever CI runs happen multiple times in one day, because our Docker Github
workflow preserves the output of the first job (the image build) for the second
job (the cluster tests).

For successful runs, there's no use for the Docker image to survive as a CI
artifact. (For failures, it might help to grab the image for local testing.)

There's no Github-official way to delete an attachment within a workflow, so
this simply truncates the artifact if the workflow succeeds.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -168,3 +168,16 @@ jobs:
           name: cluster-btest
           path: testing/external/zeek-testing-cluster/.tmp
           if-no-files-found: ignore
+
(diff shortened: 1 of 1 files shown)


timwoj (approved):
Looks fine to me. It's kinda dumb that there's no standard way to do this.

ckreibich:
Yeah. There's no shortage of tickets complaining about it. Thanks!

PR #2358, opened 2022-08-19
parse.y/conditionals: Attempt to give @if/@ifdef/... better symantics...

awelzel:
when used as statements below if or for. Related to #2289, alternative to #2347.

---

I've looked a bit more at the @if / @ifdef "statements" and this PR would allow to use @if without braces with for loops or if statements:

```
if ( T )
    @if ( T )
        print "This will print";
```
This is in contrast to #2347 which would syntax-error these constructs.

There's something I'm a bit stuck, however, and that's `@if` at the global scope currently need to be stuffed with a `;` and I haven't figured out how to skip over that. Any pointers welcome..

See for example the following two tests that require it:

https://github.com/zeek/zeek/compare/topic/awelzel/2289-experiment-to-allow-atif-as-statements#diff-a30184051abd50dae1dbda728a27212b7d93903a9d0ddc6c8becd022dc947dbcR4

https://github.com/zeek/zeek/compare/topic/awelzel/2289-experiment-to-allow-atif-as-statements#diff-89cfbeecc17a72aa387421c5f2d106ebb04a3bd2c2875e4a089e16dc732cddfeR38


awelzel:
Closing after discussions with Robin. Will go with #2347

PR #2359, opened 2022-08-19
Zeekygen: add Sphinx labels to each plugin's generated documentation

ckreibich:
For packet/protocol/file analyzers, Zeekygen runs similar content generation code driven by the plugin manager's registered plugins. It's currently not possible to link to a specific plugin's documentation from anywhere within the documentation — Sphinx is picky about linking directly to (sub)sections elsewhere.

This tweak adds a Sphinx label derived from the plugin name, allowing simple references via `:ref:`. For example, the HTTP analyzer at

https://docs.zeek.org/en/master/script-reference/proto-analyzers.html#zeek-http

becomes linkable from anywhere via:

    :ref:`plugin-zeek-http`

I considered adding a btest but I'm not sure how — if I add a test for the `proto_analyzer` index, for example, the output is the full protocol analyzer documentation, and too sensitive for baselining. I'm about to use this in our docs, so it'll at least be covered that way, but suggestions are welcome.

Also includes removal of the outdated `BRO_DISABLE_BROXYGEN` in the btests.

### DIFF
--- a/src/zeekygen/Target.cc
+++ b/src/zeekygen/Target.cc
@@ -6,6 +6,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <regex>
 
 #include "zeek/Reporter.h"
 #include "zeek/analyzer/Component.h"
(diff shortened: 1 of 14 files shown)


ckreibich:
Our CI setup sets the docs submodule branch to `master`, but this PR would need the branch that's part of the PR. It'll pass when        zeek/zeek-docs#136 is merged.

rsmmr (approved):

PR #2360, opened 2022-08-20
Add Zeek script option for metric import topics

### LINKED ISSUE #2343: Support Broker::metrics_import_topics as option
Add support `Broker::metrics_import_topics` so that the metrics export/import can be configured via Zeek scripts only. Currently the import topics can only be set via `BROKER_METRICS_IMPORT_TOPICS`.

Some boiler-plate in Zeek scripts is in the following branch:
https://github.com/zeek/zeek/compare/topic/awelzel/try-broker-metrics-import-topics

@Neverlord - this is what we chatted about, but creating the ticket for tracking now.

Neverlord:
Closes #2343.

---

Requires: https://github.com/zeek/broker/pull/270

### DIFF
--- a/scripts/base/frameworks/broker/main.zeek
+++ b/scripts/base/frameworks/broker/main.zeek
@@ -152,6 +152,13 @@ export {
 	## BROKER_METRICS_EXPORT_TOPIC is defined.
 	option metrics_export_topic = "";
 
+	## Topics imported for metrics. This should be set on the node that
+	## is importing metrics from other nodes.
+	## Zeek overrides any value provided in zeek_init or earlier at
+	## startup if the environment variable
+	## BROKER_METRICS_IMPORT_TOPICS is defined.
+	option metrics_import_topics: vector of string = vector();
+
 	## ID for the metrics exporter. When setting a target topic for the
 	## exporter, Broker sets this option to the suffix of the new topic *unless*
 	## the ID is a non-empty string. Since setting a topic starts the periodic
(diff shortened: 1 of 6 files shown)


awelzel:
@Neverlord - the baseline for the `plugins.hooks` btest has changed due to the new bif:

https://cirrus-ci.com/task/6100190124834816?logs=test#L136

Would you be up for updating that?

awelzel:
I've tested these changes to enable metrics and centralized import to the manager by default in a cluster setup. So :+1: from me.

Neverlord:
> Would you be up for updating that?

I've updated the baseline. I've also had to update the Baseline for `testing/btest/Baseline/scripts.base.frameworks.telemetry.internal-metrics/out` since it's checking an internal CAF metric. If possible, I think it'd be best to not have baseline for such metrics. How many messages have been processed at that point depends on what CAF and Broker do internally on startup, so it's by no means a stable quantity.

awelzel:
> If possible, I think it'd be best to not have baseline for such metrics. How many messages have been processed at that point depends on what CAF and Broker do internally on startup, so it's by no means a stable quantity.

Agree. I had already skipped the values for histograms representing timings, so could drop the values for the gauge/counters, too. Even just listing them might not be super stable, but that at least provide some visibility into what metrics are being exposed or added.

0xxon (approved):

ckreibich:
Should this be covered in https://docs.zeek.org/en/master/frameworks/telemetry.html, or https://github.com/zeek/zeek/wiki/Telemetry? (I'm wondering whether the wiki page should at this point get folded into the docs altogether — thoughts welcome.)

PR #2361, opened 2022-08-21
stmt: Support iterating over vector values

awelzel:
I ran into this a few times wanting to iterate over just the values
of a vector and wondering whether that could just work.

This enables the following, where `v` will be value of `vec[i]`.

    local vec = vector("zero", "one", "two");

    for ( i, v in vec )
        print i, v;

---

Related PRs:
zeek-docs: https://github.com/zeek/zeek-docs/pull/137


src/Stmt.cc
```diff
@@ -1290,23 +1290,33 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr, IDPtr val_var)
 	{
 	value_var = std::move(val_var);
 
-	if ( e->GetType()->IsTable() )
+	auto t = e->GetType();
+	zeek::TypePtr yield_type;
+
+	if ( t->IsTable() )
+		{
```
rsmmr, line 1297: Formatting nit: Can skip braces around single-line bodies (below too).

rsmmr (commented):
Nice, looks like a natural extension to me. (I'd prefer having a way of iterating *only* over values for vectors  but we don't have syntax for that, and this fits in well with what we have.)

It probably needs a documentation update somewhere (and a NEWS entry).

src/Stmt.cc
awelzel: I force-pushed, I'm not sure it's better, feel free to fixup during the merge.

awelzel:
zeek-docs: https://github.com/zeek/zeek-docs/pull/137

rsmmr (approved):
LGTM, I'll merge later unless somebody beats me to it.

PR #2369, opened 2022-08-24
Switch macOS Monterey build to new m1-based VM

timwoj:
Fedor from Cirrus recently mentioned that they have new M1-based macOS VMs for use, which are apparently quite a bit more efficient than the existing macOS images. They're only available for Monterey and Ventura (due to being M1-based). This PR switches over our Monty build to use that image.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -242,7 +242,7 @@ alpine_task:
 # We aim to support both the current and previous macOS release.
 macos_monterey_task:
   macos_instance:
-    image: monterey-xcode-13.1
+    image: ghcr.io/cirruslabs/macos-monterey-base:latest
   prepare_script: ./ci/macos/prepare.sh
   << : *CI_TEMPLATE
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -242,7 +242,7 @@ alpine_task:
 # We aim to support both the current and previous macOS release.
 macos_monterey_task:
   macos_instance:
-    image: monterey-xcode-13.1
+    image: ghcr.io/cirruslabs/macos-monterey-base:latest
```
ckreibich, line 245: Interesting! Details here: https://cirrus-ci.org/guide/macOS/#available-images

ckreibich (approved):

.cirrus.yml
timwoj: Definitely. I've been meaning to look at that Tart tool mentioned there since I got my M1 machine but I never get around to it.

PR #2370, opened 2022-08-24
telemetry: In a cluster, open port 9911 for Prometheus by default

awelzel:
Port 9911 has been allocated here:

https://github.com/prometheus/prometheus/wiki/Default-port-allocations

Logic is put into telemetry/cluster.zeek. Doing it in `cluster/nodes/`
seemed like quite some extra splitting for just those few settings.


awelzel:
I'll check for writing a test that'd curl the manager:9911/metrics URL, but if someone feels like merging it as is, I've tested locally :-)
```
$ curl -sf localhost:9911/metrics | grep zeek_version
# HELP zeek_version_info The Zeek version
# TYPE zeek_version_info gauge
zeek_version_info{beta="false",commit="452",debug="true",endpoint="manager",major="5",minor="1",patch="0",version_number="50100",version_string="5.1.0-dev.452-debug"} 1.000000 1661361926451
zeek_version_info{beta="false",commit="452",debug="true",endpoint="logger",major="5",minor="1",patch="0",version_number="50100",version_string="5.1.0-dev.452-debug"} 1.000000 1661361926451
...
```

awelzel:
The FreeBSD 13 task failed once failing to listen on 9911, but succeeded after restarting it. Hope that's okay.

NEWS
```diff
@@ -131,6 +131,10 @@ Changed Functionality
   This prevents callbacks into script-land through change handlers when parts
   of the environment have already been torn down.
 
+- When running in cluster mode, the manager by default now imports metrics
+  from all other cluster nodes via broker topics and opens port 9911/tcp for
```
ckreibich, line 135: I've been catching up on the telemetry work and I find it'd be helpful to explain somewhere with examples how Broker topics and their naming / structure relate to metrics collection, most likely over in `zeek-docs`.

testing/btest/scripts/base/frameworks/telemetry/cluster.zeek
```diff
@@ -0,0 +1,81 @@
...
+# @TEST-EXEC: btest-bg-run proxy-1 ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=proxy-1 zeek -b %INPUT
+# @TEST-EXEC: btest-bg-run worker-1  ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=worker-1 zeek -b %INPUT
+# @TEST-EXEC: btest-bg-wait 10
+# @TEST-EXEC: btest-diff manager-1/.stdout
+
+@load base/frameworks/cluster
+@load base/frameworks/telemetry
+@load base/utils/active-http
+
+# Query the Prometheus endpoint using ActiveHTTP for testing, oh my.
```
ckreibich, line 21: An aside: when these test setups start to feel hairy, we could instead put them into the cluster management testsuite — `docker-compose` and `curl` are arguably more straightforward than `btest-bg-run` and `ActiveHTTP`, and there's no need to keep the testsuite strictly focused on the Management framework. (Seems fine for me here, but something to keep in mind for the future — we could no doubt benefit from more cluster framework testing.)

ckreibich (approved):

ckreibich:
Oops! I just noticed that the testcase fails reliably on my machine when I run it parallelized:
```
$ while true; do make; [ -f /tmp/done ] && break; done
[ 93%] scripts.base.frameworks.telemetry.cluster ... failed
1 of 1306 tests failed, 4 skipped
make: *** [Makefile:10: btest-verbose] Error 1
[ 92%] scripts.base.frameworks.telemetry.cluster ... failed
1 of 1306 tests failed, 4 skipped
make: *** [Makefile:10: btest-verbose] Error 1
[ 93%] scripts.base.frameworks.telemetry.cluster ... failed
1 of 1306 tests failed, 4 skipped
make: *** [Makefile:10: btest-verbose] Error 1
[ 91%] scripts.base.frameworks.telemetry.cluster ... failed
1 of 1306 tests failed, 4 skipped
make: *** [Makefile:10: btest-verbose] Error 1
```
It runs perfectly when run in isolation. I noticed this in the parallelized setting:
```
error: curl -s -g -o "/tmp/zeek-activehttp-rFj3eGxkRR5_body" -D "/tmp/zeek-activehttp-rFj3eGxkRR5_headers" -X "GET" -m 60 "http://localhost:9911/metrics" && touch "/tmp/zeek-activehttp-rFj3eGxkRR5_body" |/Input::READER_RAW: Child process exited with non-zero return code 7
```
Probably just a timeout problem ... happy to dig further tomorrow but for now I undid the approval.

NEWS
awelzel: Agree on updating the docs to point out that metrics importing/centralization is now happening by default (the docs currently have some rambling around setting environment variables to enable this) and possibly how to turn it off.

As of how broker topics and naming relate to metrics collection, not sure it's `zeek-docs` material: It's an implementation detail that may or may not change down the road, too.

awelzel:
> Probably just a timeout problem ... happy to dig further tomorrow but for now I undid the approval.

Hmm... The "Child process exited with non-zero return code 7" from curl is "Failed to connect to host.", so maybe the Prometheus server wasn't up yet, or it had been opened by another cluster test that then stopped.

I've pushed a fixup to use a dynamic metrics port, because using the fixed 9911/tcp across all test is likely prone to errors when running in parallel, I might have just gotten lucky so far.

Thanks for the pointers at the zeek-testing-cluster, I'm looking a bit.

NEWS
ckreibich: Good point about the implementation side of things, that wasn't clear to me. Developer docs are a bit tricky — there is https://docs.zeek.org/en/master/devel/index.html, but we've also been using the wiki for such things. Please feel free to handle as you see fit.

awelzel:
> Oooh Arne I misled you with the timeout comment. It looks like it's a port collision. Further digging revealed

Yeah, I suspected as much - I guess my previous comment went under:

>> Hmm... The "Child process exited with non-zero return code 7" from curl is "Failed to connect to host.", so maybe the
>> Prometheus server wasn't up yet, or it had been opened by another cluster test that then stopped.
>>
>> I've pushed a fixup to use a dynamic metrics port, because using the fixed 9911/tcp across all test is likely
>> prone to errors when running in parallel, I might have just gotten lucky so far.

Had you tried the current state of this branch? I had pushed a commit very similar to yours changing the port to a dynamically allocate one instead of using 9911/tcp across all the tests.

So, think this is good to go?

ckreibich:
Huh, I'm confused ... I went through the merge routine for your updated branch to verify, and `update-changes` picked up the fact that there was a second commit. But I didn't see the dynamic port allocation it in the code.

Anyway — surely something on my side. I'll resolve this one now.

PR #2372, opened 2022-08-24
Avoid race conditions in broker.ssl_auth_failure btest

### LINKED ISSUE #2367: broker.ssl_auth_failure consistently fails on Mac Os (monterey) after broker update
After updating broker to zeek/broker@92b4287c3a744923116e9c34af2e8b656f393b29, the broker.ssl_auth_failure test consistently fails on my mac os installation with:

```
$ btest -d broker.ssl_auth_failure
[  0%] broker.ssl_auth_failure ... failed
  % '$SCRIPTS/wait-for-file send/failed 20 || (btest-bg-wait -k 1 && false)' failed unexpectedly (exit code 1)
  % cat .stderr
  <<< [90842] zeek -b ../recv.zeek >recv.out
  received termination signal
  >>>
  <<< [90852] zeek -b ../send-check.zeek >send.out
  received termination signal
  >>>
  <<< [90870] zeek -b ../send.zeek >send.out
  received termination signal
  received signal while waiting for thread reporter/Log::WRITER_ASCII, aborting all ...
  >>>
```

Changing the broker submodule pointer is the only required change. The previous version (which Zeek master points to as of now), works. That version is zeek/broker@8d03476847411b0ab7b9c47ea9eaf7b1535d08c2 of broker. 

Convenience diff link: https://github.com/zeek/broker/compare/8d03476847411b0ab7b9c47ea9eaf7b1535d08c2..92b4287c3a744923116e9c34af2e8b656f393b29

This test was done with a clean Zeek checkout at 882949004591a1af6ac2dbb112c20a06e7465bba on Mac Os 12.5.1; my system uses OpenSSL 3.0.5 (macports).

timwoj:
I think this should fix #2367. It adds one more `touch` to ensure that the listener is up before trying to connect to it, as well as uses `__peer_no_retry` from d3ad3c618a62c6b415a47c54fdd679f26c9ed34f to avoid connecting peers getting stuck retrying a connection that will never succeed.

### DIFF
--- a/testing/btest/broker/ssl_auth_failure.zeek
+++ b/testing/btest/broker/ssl_auth_failure.zeek
@@ -3,8 +3,10 @@
 # @TEST-PORT: BROKER_PORT
 #
 # @TEST-EXEC: btest-bg-run recv "zeek -b ../recv.zeek >recv.out"
-# @TEST-EXEC: btest-bg-run send-check1 "zeek -b ../send-check.zeek >send.out"
 # @TEST-EXEC: $SCRIPTS/wait-for-file recv/listen-ready 20 || (btest-bg-wait -k 1 && false)
+
+# @TEST-EXEC: btest-bg-run send-check1 "zeek -b ../send-check.zeek >send.out"
+# @TEST-EXEC: $SCRIPTS/wait-for-file recv/connected 20 || (btest-bg-wait -k 1 && false)
 #
 # @TEST-EXEC: btest-bg-run send "zeek -b ../send.zeek >send.out"
 # @TEST-EXEC: $SCRIPTS/wait-for-file send/failed 20 || (btest-bg-wait -k 1 && false)
@@ -98,7 +100,7 @@ BTdqMbieumB/zL97iK5baHUFEJ4VRtLQhh/SOXgew/BF8ccpilI=
 
 event zeek_init()
 	{
-	Broker::peer("127.0.0.1", to_port(getenv("BROKER_PORT")));
+	Broker::__peer_no_retry("127.0.0.1", to_port(getenv("BROKER_PORT")));
 	}
 
 event Broker::peer_added(endpoint: Broker::EndpointInfo, msg: string)
@@ -119,7 +121,7 @@ redef Broker::ssl_certificate = "../cert.1.pem";
 
 event zeek_init()
 	{
-	Broker::peer("127.0.0.1", to_port(getenv("BROKER_PORT")));
+	Broker::__peer_no_retry("127.0.0.1", to_port(getenv("BROKER_PORT")));
 	}
 
 event Broker::peer_added(endpoint: Broker::EndpointInfo, msg: string)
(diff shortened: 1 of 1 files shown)


timwoj:
This should get included in the 5.0.1 release since it's broken on the release branch.

Neverlord (approved):
LGTM. 👍

rsmmr:
I am reading this right then that the test relied on the previous (erroneous) behavior of aborting after 10 attempts?

Neverlord:
> I am reading this right then that the test relied on the previous (erroneous) behavior of aborting after 10 attempts?

I think there's more to the story. The last functional change to this test was back in 2020 with c8545c85d898517ac7112e1509500394c50f2332. There are two newer commits on the file, but they don't touch the test code itself. This was before we've changed to the new Broker version, i.e., before the change in behavior and with indefinite retries.

It looks like there's another slight change in Behavior in Broker and it no longer sends error messages unless it aborts completely. This was after 10 attempts or after the first try when setting the retry time to 0.

However, I think the test shouldn't look for an `error` event in the first place. It should receive `peer_unavailable`, which is what we actually want to test.

I didn't dig into the old sources to figure out why Broker used to send an `error` here although it was still trying to connect. We could also bring back the "double reporting" of `peer_unavailable` as both `peer_unavailable` event *and* `error`, even tough it's quite redundant.

Neverlord:
> I didn't dig into the old sources to figure out why Broker used to send an error here although it was still trying to connect. We could also bring back the "double reporting" of peer_unavailable as both peer_unavailable event and error, even tough it's quite redundant.

Sorry, I've confused that with a status code. `peer_unavailable` actually *is* an error. So the test should use that IMHO, since it's more specific that just "any error". I would need to dig into the old sources (Zeek 4.x) to understand why it used to work previously with indefinite retries.

timwoj:
So the use of `Broker::error` in the test should be `Broker::peer_unavailable`? Doing that doesn't fix the test failure, unfortunately.

timwoj:
Now that I take a minute and look more deeply, I don't see where `peer_unavailable` is defined. Am I missing something? I see `peer_{added,removed,lost}` but not `unavailable`.

timwoj:
On top of the above, there's a couple of other tests that handle `Broker::error` in the same way, including `broker.ssl-mismatch` which is the one this PR is based on. Given all of that, should I just merge this fix as-is so I can get `master` and the 5.0 release branch back into a working state, and we can fix which event (and/or add the missing one) later?

Neverlord:
Well, if there's no handler for `peer_unavailable`, that explains why it hasn't been used before. I don't know (or remember) the history of why the BIFs have handlers for some of the peer-related error codes but not all.

> should I just merge this fix as-is so I can get master and the 5.0 release branch back into a working state, and we can fix which event (and/or add the missing one) later?

If `master` is broken, then I'd go with that. I'll also look into this in depth to see why it stopped working without the changes (and work on a patch ASAP). A first check I did with `broker-benchmark` confirmed that Broker does emit `peer_unavailable` messages for each failed attempt, which means the test should've continued to work without the changes. That being said, I think the changes should go in regardless since they make the test more deterministic.

PR #2368, opened 2022-08-24
Bump Spicy dependencies to spicy-1.5.1 and spicy-plugin-1.3.17.

bbannier:

PR #2373, opened 2022-08-24
Bump Spicy dependencies to spicy-1.5.1 and spicy-plugin-1.3.17.

timwoj:
This is the `master` version of https://github.com/zeek/zeek/pull/2368. I'll merge it once CI is done with it.


bbannier:
Not sure why this fails when its base branch seems to pass CI. I retriggered CI, but it failed again. I then retriggered the `ubuntu22` job with a terminal, and it failed both with this patch and on its parent commit.

PR #2375, opened 2022-08-25
Expand Supervisor to support loading additional scripts before user scripts

ckreibich:
I hit another scenario in which the fact that the Supervisor's script customization adds to the very end of the script pile tripped me up. We've touched this on occasion in the past and I've so far worked around it, but it always feels clumsy, so here's an additional vector for loading before the user-provided ones.

I'm unsure about naming, thoughts welcome. Something like `$pre_scripts` and `$post_scripts` seems nicer, but that would break existing scripts relying on the existing naming. We could add `$post_scripts` and deprecate `$scripts`.


ckreibich:
I rather like `addl_base_scripts` and `addl_user_scripts` for the two vectors.

ckreibich:
> I rather like addl_base_scripts and addl_user_scripts for the two vectors.

I went ahead and added these.

src/supervisor/Supervisor.h
```diff
@@ -198,7 +198,19 @@ class Supervisor : public iosource::IOSource
...
+		std::vector<std::string> addl_base_scripts;
+		/**
+		 * Additional script filename/paths that the node should load
+		 * after any user-specified scripts.
+		 */
+		std::vector<std::string> addl_user_scripts;
+		/**
+		 * The former name for addl_user_scripts, now deprecated.  Use
+		 * addl_user_scripts instead. The scriptland equivalent has a
+		 * proper deprecation tag.
```
timwoj, line 213: Any reason to not mark the C++ side deprecated too?

testing/btest/supervisor/config-scripts.zeek
```diff
@@ -36,7 +38,15 @@ event zeek_init()
...
-		print node_output_file, "supervised node zeek_init()";
+
+@ifdef ( Pre::counter )
+		# Even though this else-block runs only in the supervised node,
+		# it still needs to parse correctly in the Supervisor, so we
+		# need to condition this with availability of the counter (which
+		# only exists in the supervised node).
+		++Pre::counter;
+		print node_output_file, fmt("supervised node zeek_init(), counter at %s", Pre::counter);
+@endif
```
timwoj, line 49: Is this block related to this PR or something else?

timwoj (changes requested):

src/supervisor/Supervisor.h
ckreibich: I wondered about that. It adds a bunch of warnings to our build that seemed excessive to me given the script-layer deprecation, but it would technically be correct to deprecate there as well. Happy to do as you prefer.

testing/btest/supervisor/config-scripts.zeek
ckreibich: This PR — the test verifies the load order by adding the counter in the post-base scripts, using it in the test script itself, and again in the post-user ones.

src/supervisor/Supervisor.h
timwoj: It's not part of a `detail` namespace so we should probably follow the process. At worst, you can add pragmas to the Zeek code to silence the warnings for our builds.

ckreibich: Added, Tim. You can see the noise here: https://cirrus-ci.com/task/6385802665000960?logs=build#L3071

ckreibich: I've pushed an update with pragma use as per our discussion.

timwoj (approved):
One nit. LGTM otherwise.

src/supervisor/Supervisor.h
```diff
@@ -143,6 +143,16 @@ class Supervisor : public iosource::IOSource
 	 */
 	struct NodeConfig
 		{
+#pragma GCC diagnostic push
```
timwoj, line 146: Maybe just add a comment here about why these are defined and that they can possibly be removed later.

ckreibich: Good call Tim, update pushed.

PR #2378, opened 2022-08-25
Add a generic fuzzer instead of new files for each new fuzzer

timwoj:
This avoids us having to create a new `.cc` file for every new fuzzer we add if the only thing that changes is the name of the analyzer.

It also moves the corpora data into a separate directory to collect it all more neatly. This change will require a change to the oss-fuzz repo as well.


src/fuzzers/CMakeLists.txt
```diff
@@ -49,6 +45,19 @@ macro(ADD_FUZZ_TARGET _name)
...
     endif ()
+endmacro()
+
+macro(ADD_FUZZ_TARGET _name)
+    set(_fuzz_target zeek-${_name}-fuzzer)
+    set(_fuzz_source ${_name}-fuzzer.cc)
+    setup_fuzz_target(${_fuzz_target} ${_fuzz_source})
+endmacro ()
+
+macro(ADD_GENERIC_FUZZ_TARGET _name)
```
awelzel, line 56: Mostly thinking the name here and the `.cc` file should be a little less generic such as "generic-analyzer-fuzzer" or "generic-protocol-fuzzer" in case there's ever another generic fuzzer showing up?

Had also wondered if a single executable receiving the analyzer name as `argv[]` or via `environment` could work, but that might not how it's done with libfuzzer.

timwoj: > Mostly thinking the name here and the .cc file should be a little less generic such as "generic-analyzer-fuzzer" or "generic-protocol-fuzzer" in case there's ever another generic fuzzer showing up?

Sure, that sounds like a good idea.

> Had also wondered if a single executable receiving the analyzer name as argv[] or via environment could work, but that might not how it's done with libfuzzer (and searching didn't show anything obvious).

I went down that route at first but as far as I can tell there's no easy way to pass an argument to the fuzzer via the scripts we provide to oss-fuzz. They pass a set of arguments for running things for both libFuzzer and hongfuzz and I don't think we can hook into that to pass our own. Generating a bunch of extra binaries feels a little bad but at least now we don't need to add source files for each one.

awelzel (approved):

PR #2379, opened 2022-08-25
frameworks/dpd: Move to frameworks/analyzer/dpd, load in bare-mode

awelzel:
Point for feedback/discussions:

* Because frameworks/analyzer is loaded via init-frameworks-and-bifs the
  dpd functionality (really just dpd.log and disabling of analyzers) is
  now enabled even in bare mode.
* ~There are some baseline changes that I haven't wrapped my head around.
  E.g., logging.field-extension: It seems it's all the same connections,
  but the conn uids have changed around.~
* Not sure we need to keep frameworks/base/dpd/__load__.zeek around
  or can just remove it right away.


awelzel:
> There are some baseline changes that I haven't wrapped my head around.
E.g., logging.field-extension: It seems it's all the same connections,
but the conn uids have changed around.

I ran into this for a different change and I believe I understand now: Before this change, there are no handlers for the `analyzer_confirmation()` or `analyzer_violation()` events when running in bare mode. After this change, there are and they are likely some of the first events that trigger a `Connection::GetVal()` call which in turn will initialize the `uid` values of connections. And most likely in a different order now..

scripts/base/frameworks/dpd/__load__.zeek
```diff
@@ -1 +1,3 @@
-@load ./main
+@deprecated "Remove in v6.1 - use @load base/frameworks/analyzer instead";
```
rsmmr, line 1: Given that bare mode loads this anyways now, do we want to just say "Remove in v6.1 - now loaded automatically"?

scripts/policy/frameworks/dpd/packet-segment-logging.zeek
```diff
@@ -4,7 +4,7 @@
 ##! A caveat to logging packet data is that in some cases, the packet may
 ##! not be the packet that actually caused the protocol violation.
 
-@load base/frameworks/dpd
+@load base/frameworks/analyzer
```
rsmmr, line 7: This needed?

scripts/base/frameworks/dpd/__load__.zeek
rsmmr, line 1: As this is now loaded in bare mode anyways, do want to just say "Remove in v6.1 - now loaded automatically"?

rsmmr: And, yes, I'd keep this for now.

rsmmr:
> Because frameworks/analyzer is loaded via init-frameworks-and-bifs the
dpd functionality (really just dpd.log and disabling of analyzers) is
now enabled even in bare mode.

Sounds fine to me.

scripts/policy/frameworks/dpd/packet-segment-logging.zeek
awelzel: Dropped. The policy files are suspiciously in `policy/frameworks/dpd` still, but there's not `policy/frameworks/analyzers`, so keeping it as is.

rsmmr (approved):
Looks good. What's up with that Montery CI error though? Seems unrelated?

awelzel:
> Looks good. What's up with that Montery CI error though? Seems unrelated?

Yep, likely. I'm seeing these canceled/timing out regularly. I pressed re-run.

Would this be worth a separate ticket?

rsmmr:
The Monterey one actually did't look like a timeout but a compiler error, which is what confused me. Maybe I read it wrong though. I'll merge this.

PR #2380, opened 2022-08-26
gtpv1: Do not register for protocol detection

awelzel:
While reviewing/understanding the analyzer setup, it didn't seem like
GTPv1 implements packet_analysis::Analyzer::DetectProtocol(), so
should not register it for protocol detection either.

Alternatively, maybe DetectProtocol() should've been implemented in
which case maybe this should be an issue?


timwoj (approved):
I went back to 4.0 and there isn't a dpd file for gtpv1, so this was probably just a copy/paste bug from when I was reimplementing all of the tunnel analyzers.

PR #2383, opened 2022-08-26
Add type checking to string_cat arguments

### LINKED ISSUE #2382: string_cat doesn't type-check its arguments
The following:

> event zeek_init()
	{
	print string_cat(network_time());
	}

results in a crash, because `string_cat` simply assumes that its arguments are all Zeek strings.

timwoj:
Fixes #2382


src/strings.bif
```diff
@@ -64,7 +64,15 @@ function string_cat%(...%): string
 	%{
 	int n = 0;
 	for ( const auto& a : @ARG@ )
+		{
+		if ( a->GetType()->Tag() != TYPE_STRING )
+			{
+			zeek::reporter->Error("string_cat() requires string arguments");
+			return zeek::make_intrusive<zeek::StringVal>("");
```
awelzel, line 71:
```suggestion
			return val_mgr->EmptyString();
```

Otherwise might be helpful to return the violating pos/type, but no strong opinion.

awelzel (approved):

src/strings.bif
timwoj: Some of the error messages in BIFs return just a string (via a call to `Reporter::Error`) and some of them report a location (via `emit_builtin_error`). I have no idea why one or the other is used in each case.

Issue #2031, opened 2022-04-08
Missing `Analyzer{Confirmation,Violation}` methods for file analyzers


rsmmr:
Both protocol and packet analyzers have them, but file analyzers don't. Seems for consistency we should add them otherwise it's unclear how to report those conditions for files.

Also, should those feed into dpd.log as well then? Probably not, because `base/frameworks/dpd/main.zeek` is tailored to protocols. But what else to do with them?

And actually: what do we do with packet analyzer violations currently?

awelzel:
We have the more generic `_info` versions of the confirmation/violation events now and the file analyzer class has respective C++ methods now, too.

> Also, should those feed into dpd.log as well then? Probably not, because base/frameworks/dpd/main.zeek is tailored to protocols. But what else to do with them?


Related comment from @rsmmr over at spicy-plugin: https://github.com/zeek/spicy-plugin/issues/101#issuecomment-1081759145

> We record parse errors in a separate log file, say spicy.log or analyzers.log. We'd probably want to make creation of that optional too in case people aren't interested.

I like the `analyzer.log` that logs all violations regardless of analyzer type and confirmation state. It would likely end-up with a few optional fields or a generic string field. Imagining that this is more on the debug side of things. Optionally logging confirmations might be interesting, too. 

> And actually: what do we do with packet analyzer violations currently?

Depends of whether they have an associated connection: If they have one and it was confirmed, it should get logged via dpd.zeek. If there's only a confirmation, c$service / c$conn$service will contain it).

If there's no connection associated with a packet analyzers, confirmations/violations are not propagated to script land.

---

I'm proposing to close this issue and opening a new one for the `analyzer.log` creation. Other thoughts?

PR #2390, opened 2022-08-31
Introduce generic analyzer_confirmation_info and analyzer_violation_info

awelzel:
Introduce two new events for analyzer confirmation and analyzer violation
reporting. The current analyzer_confirmation and analyzer_violation
events assume connection objects and analyzer ids are available which
is not always the case. We're already passing aid=0 for packet analyzers
and there's not currently a way to report violations from file analyzers
using analyzer_violation, for example.

These new events use an extensible Info record approach so that additional
(optional) information can be added later without changing the signature.
It would allow for per analyzer extensions to the info records to pass
analyzer specific info to script land. It's not clear that this would be
a good idea, however.

The previous analyzer_confirmation and analyzer_violation events
continue to exist, they are invoked from script-land using the new
analyzer_confirmation_info and analyzer_violation_info events.

The last point yields to a subtle change of behavior visible in some of
the disable-analyzer tests: As the old events are now scheduled from
script-land now, they may be raised after protocol specific events have
been processed already. I presume if that matters to someone we can ask
them to use the new event.


awelzel:
Related to #2031

awelzel:
Adding this here for thoughts: @ckreibich mentioned that an `analyzer_not_implemented` or `analyzer_todo` event came up in discussions with @vpax (?) to raise information from an analyzers to script land for parts of a protocol that just hasn't been done (rather than using weird or debug logging).

I'm wondering a bit if that would be worth considering a very generic `analyzer_event_info` and funneling confirm/violation/todo through that. But I don't think it is - having discrete events seems more sensible and common.

vpax:
> Adding this here for thoughts: @ckreibich mentioned that an `analyzer_not_implemented` or `analyzer_todo` event came up in discussions with @vpax (?) to raise information from an analyzers to script land for parts of a protocol that just hasn't been done (rather than using weird or debug logging).

I don't recall that.  That said, it makes sense to me, and as you then comment, I think it also makes sense to keep the events discrete, rather than bundled.

rsmmr:
I looked over this high-level: I like the approach with the two new events and passing them records with fields set as a available. I'll go over in more detail but one question: What's the plan for the existing events? This code keeps triggering them, but doesn't deprecate. I think I'd do deprecate and then phase them out over time.

Also wondering if we should keep triggering the old ones from C++ so that we don't change timing for them. While for most events this wouldn't be an issue, I can see it for `analyzer_violation` that the order may impact something.

awelzel:
> I looked over this high-level: I like the approach with the two new events and passing them records with fields set as a available. I'll go over in more detail but one question: What's the plan for the existing events? This code keeps triggering them, but doesn't deprecate. I think I'd do deprecate and then phase them out over time.

Okay. I had intended to keep them and even thought of introducing a `file_analyzer_violation(...)` based off the generic events, but seems you're leaning in the other direction. One other motivating factor for not deprecating them was that we just  deprecated `protocol_violation` and introduced `analyzer_violation` as replacement, and now we're deprecating them again. But, oh well I guess ;-)

> 
> Also wondering if we should keep triggering the old ones from C++ so that we don't change timing for them. While for most events this wouldn't be an issue, I can see it for `analyzer_violation` that the order may impact something.

Yep, okay. I actually ran into the timing issue for the "dpd segement logging" script which gets the current packet, but that already had a comment that it might not be doing the right thing. If we go for deprecation of those event, I wouldn't mind continuing to have the C++ layer raise the events. If we weren't deprecating, that would feel a bit duplicated..

I'll address both as follow-up commits, but will probably rebase once #2379 is in. Maybe better to wait before you look closer?

rsmmr:
> Okay. I had intended to keep them and even thought of introducing a `file_analyzer_violation(...)` based off the generic events, but seems you're leaning in the other direction.

Yeah, that would get confusing I think.

> One other motivating factor for not deprecating them was that we just deprecated `protocol_violation` and introduced >`analyzer_violation` as replacement, and now we're deprecating them again. 

Yeah, I know, but I'd rather work towards where we want to be eventually, even if that means changing again.

I'd actually prefer to call the new events `analyzer_confirmation/violation` as well, just with different arguments. But I don't see how we'd get there without breaking existing scripts.

> I'll address both as follow-up commits, but will probably rebase once #2379 is in. Maybe better to wait before you look closer?

Ack.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -75,21 +82,37 @@ event analyzer_violation(c: connection, atype: AllAnalyzers::Tag, aid: count,
...
+	c$dpd = dpd;
 	}
 
-event analyzer_violation(c: connection, atype: AllAnalyzers::Tag, aid: count, reason: string) &priority=5
+event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo ) &priority=5
 	{
 	if ( atype in ignore_violations )
 		return;
 
+	if ( ! info?$c || ! info?$aid )
```
rsmmr, line 111: I'm wondering if we should indeed want to key this on having `c`? If we didn't, then a file analyzer could still set a connection as well. In your earlier version you had new bifs to check what type of analyzer it is, which I liked. How about going back to that?

rsmmr: Same below for other tests for the type of analyzer.

scripts/base/init-bare.zeek
```diff
@@ -2268,6 +2268,52 @@ const RPC_status = {
...
+##
+## .. zeek:see:: analyzer_confirmation_info
+type AnalyzerConfirmationInfo: record {
+	## The connection related to this confirmation, if any.
+	c: connection &optional;
+
+	## Specific analyzer instance that can be used to
+	## reference the analyzer when using builtin functions like
+	## :zeek:id:`disable_analyzer`.
+	aid: count &optional;
```
rsmmr, line 2281: Just for easier reading, I'd move this either to the top or to the end, so that `c` and `f` stay together.

scripts/base/init-bare.zeek
```diff
@@ -2268,6 +2268,52 @@ const RPC_status = {
 	[RPC_UNKNOWN_ERROR] = "unknown"
 };
 
+## Generic analyzer confirmation info record.
+##
+## .. zeek:see:: analyzer_confirmation_info
+type AnalyzerConfirmationInfo: record {
+	## The connection related to this confirmation, if any.
```
rsmmr, line 2275: Related to my comment above I would suggest to explicitly say when this is set: only for protocol analyzers, or generally if there a connection semantically associated to this activity?

scripts/base/init-bare.zeek
```diff
@@ -2268,6 +2268,52 @@ const RPC_status = {
...
+	## The connection related to this violation, if any.
+	c: connection &optional;
+
+	## Specific analyzer instance that can be used to
+	## reference the analyzer when using builtin functions like
+	## :zeek:id:`disable_analyzer`.
+	aid: count &optional;
+
+	## The file object related to this violation, if any.
+	f: fa_file &optional;
```
rsmmr, line 2303: Similar as above: move this to after `c` for readability..

scripts/base/init-bare.zeek
```diff
@@ -2268,6 +2268,52 @@ const RPC_status = {
...
+
+	## Piece of binary data that was parsed and caused the violation.
+	data: string &optional;
+
+	# TBD: Possibly more of the field that was parsed and caused
+	# the violation? Or some opaque table[string] of string that's
+	# analyzer specific?
+	# Or maybe the packet that trigged the violation?
+	# pkt: pcap_packet &optional;
```
rsmmr, line 2312: I'd skip this until we find a use case; can always extend the record.

src/file_analysis/analyzer/pe/PE.cc
```diff
@@ -30,6 +30,7 @@ bool PE::DeliverStream(const u_char* data, uint64_t len)
 		}
 	catch ( const binpac::Exception& e )
 		{
+		AnalyzerViolation(util::fmt("Binpac exception: %s", e.c_msg()));
```
rsmmr, line 33: This will be user-visible so would skip mentioning binpac (e.g., "analyzer exception", or directly "PE exception").

src/packet_analysis/Analyzer.cc
```diff
@@ -166,30 +166,67 @@ void Analyzer::Weird(const char* name, Packet* packet, const char* addl) const
...
+
+	const auto& tval = arg_tag ? arg_tag.AsVal() : tag.AsVal();
+	event_mgr.Enqueue(analyzer_confirmation_info, tval, info);
+	}
+
+void Analyzer::EnqueueAnalyzerConfirmation(session::Session* session, zeek::Tag arg_tag)
+	{
+	// TBD: Should we just skip raising this event if session is NULL?
+	//      But also wondering what can be done with a confirmation that
+	//      doesn't have a session? Metrics tracking?
```
rsmmr, line 186: Not immediately sure, but I'm realizing another issue here: raising a confirmation event per packet is pretty expensive. I'm wondering if confirmations just generally don't make sense for packet analyzers?

src/packet_analysis/Analyzer.cc
```diff
@@ -204,4 +241,16 @@ void Analyzer::AnalyzerViolation(const char* reason, session::Session* session,
...
+void Analyzer::AnalyzerViolation(const char* reason, session::Session* session, const char* data,
+                                 int len, zeek::Tag arg_tag)
+	{
+	session->SetAnalyzerState(GetAnalyzerTag(), session::AnalyzerConfirmationState::VIOLATED);
+
+	if ( analyzer_violation_info )
+		EnqueueAnalyzerViolationInfo(session, reason, data, len, tag);
+
+	if ( analyzer_violation )
+		EnqueueAnalyzerViolation(session, reason, data, len, tag);
```
rsmmr, line 253: This could likewise get expensive if the same violation keeps triggering for many packets. Not sure what to do about that ... (could do some thresholding/limiting, but not straight-forward how).

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -63,9 +67,12 @@ event analyzer_confirmation(c: connection, atype: AllAnalyzers::Tag, aid: count)
 	add c$service[analyzer];
 	}
 
-event analyzer_violation(c: connection, atype: AllAnalyzers::Tag, aid: count,
-                         reason: string) &priority=10
+event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo) &priority=10
```
rsmmr, line 70: This keeps limiting DPD to connection analyzers, right? Is the plan to extend this to files later?

src/packet_analysis/Analyzer.cc
awelzel: Within Zeek the confirmation only happens once for a session using "AnalyzerState" (and from looking at the code only for tunnel analyzers that actually have an outer session), so that's not per-packet unless in worst-case tunneling scenario.

For violations as you pointed out below there's no cap per session, however, and these are likely more interesting.

> I'm wondering if confirmations just generally don't make sense for packet analyzers?

There is some basic DetectProtocol() logic, so it might make sense, but could also start tracking occurrences as members or metrics and make those available to script-land.

(...but I'm then tempted to question the generic events again if they are just for protocol and file analyzers...)

src/packet_analysis/Analyzer.cc
awelzel: Yes. See reply above.

scripts/base/frameworks/analyzer/dpd.zeek
awelzel: With DPD you're thinking only dpd.log, or the disable_analyzer() logic?

* I could see a less connection-oriented log for tracking analyzer_violations where the analyzer would be the primary entity of interest (or extend the existing dpd.log with fuid and make connection fields optional).

* About `disable_analyzer()`: If a FileAnalyzer's currently returns `false` from it's `DeliverStream()` method, it's disabled - didn't plan to change something around here right now, though let me know if that's another topic for consideration (motivation might have been the hashes where that's the right thing to do).

https://github.com/zeek/zeek/blob/b6512e82c57f2b632b6b328a3a413d77c55d79c8/src/file_analysis/File.cc#L420-L424

src/packet_analysis/Analyzer.cc
awelzel: > I'm wondering if confirmations just generally don't make sense for packet analyzers?

Minimally one effect would be that the `c$service` fields don't contain `ayiya` or `vxlan` for tunnel connections and we'd need to find a separate way to propagate those up.

scripts/base/frameworks/analyzer/dpd.zeek
awelzel: I've added the bifs back and wrappers respective modules and guarded the event handlers with those. Could also see them all in the global namespace as they become quite long to type and may fit there, too.

rsmmr: Did you push those updates? Not seeing them. But yes, I can see them in the global namespace.

scripts/base/frameworks/analyzer/dpd.zeek
rsmmr: I could see the same try-to-parse-and-see-if-it-works logic for files that we do for protocols. That way, we could rely less on signature for MIME types, maybe not at all anymore (not sure). But ok to leave that for later.

scripts/base/frameworks/analyzer/dpd.zeek
awelzel: Sorry about that. I just pushed that commit..

awelzel: They are now global as `is_protocol_analyzer()`, `is_packet_analyzer()`.

awelzel:
@rsmmr - I've pushed a few more fixups. I'd rebase on top of master and we could possibly chat with this is Zeek 5.1 or 5.2 material. It might be nice to have it in 5.1 as a stepping-stone.

rsmmr:
Sorry, too late now for 5.1, but yeah, go ahead and rebase on master and then I'll take another look and we can wrap it up.

awelzel:
>  Sorry, too late now for 5.1, but yeah, go ahead and rebase on master and then I'll take another look and we can wrap it up.

@rsmmr - I have rebased, squashed and added the "violation only once for packet analyzers" change we talked about.

Let me know if something looks off to you.

awelzel:
I had looked at analyzer_confirmation/violation a bit more.

* Found #2437 while there. This will conflict with this branch so can also fold it in here if you'd prefer, but could also see it back-port worthy.

* The analyzer/dpd.zeek script within master can currently be tricked into calling `disable_analyzer()` with a packet analyzers tag and tunnel connections that will have no effect. 

* In the same area one can also trigger a dpd.log for every other vxlan packet in a connection by interleaving good and corrupted packets. It's mostly unexpected.

The latter two should be fixed by raising confirmation/violations only once and also ignoring packet analyzers when attempting to disable analyzers.

src/packet_analysis/Analyzer.cc
```diff
@@ -166,30 +166,64 @@ void Analyzer::Weird(const char* name, Packet* packet, const char* addl) const
 	session_mgr->Weird(name, packet, addl, GetAnalyzerName());
 	}
 
+void Analyzer::EnqueueAnalyzerConfirmationInfo(session::Session* session, zeek::Tag arg_tag)
```
rsmmr, line 169: Where is the check if we have session at all, and only generate an event if we do? Is that up to the caller to check? I'd add a check into these methods if session is null, easy to break otherwise.

Same for the methods below.

rsmmr (commented):
Looks good, just one question.

src/packet_analysis/Analyzer.cc
awelzel: I was forth and back on that and minimally the crash indicated that one is doing something wrong. Calling AnalyzerConfirmation/Violation for packet analyzers without sessions currently isn't valid. Protecting from it makes things safer, but also swallows the confirmation/violation that someone may expect.

I'll add it, but if we have some sort of WarnOnce() I'd probably like to use that here.

PR #2396, opened 2022-09-02
Allow BPF filter compiler to determine severity of errors

timwoj:
I'm opening this as a draft because I'm not the hugest fan of the approach I took, but I'm struggling to find another one. This PR adds support for the `BPF_Filter` class to determine the severity of errors that it gets back from libpcap and to pass that error state upwards to a caller. This allows, for example, the pcap setup to not fail out on a pcap file that contains unsupported link types (such as in https://github.com/zeek/zeek/issues/1995).

The reason for the current approach is that I'm trying not to modify the signatures of the methods in PktSrc since that class is widely used by external plugins.


awelzel (commented):
Uhm, I forgot to submit this this morning :man_facepalming: 

Submitting as "comments"

awelzel (commented):
Uhm, I forgot to submit this this morning :man_facepalming: 

Submitting as "comments"

src/iosource/PktSrc.h
```diff
@@ -336,8 +352,20 @@ class PktSrc : public IOSource
...
+	 * PrecompileBPFFilter(). This is primarily used by the pcap source
+	 * use a different version of BPF_Filter::Compile;
+	 *
+	 * @param filter the filtering string being compiled.
+	 *
+	 * @return The compiled filter or nullptr if compilation failed.
+	 */
 	virtual detail::BPF_Program* CompileFilter(const std::string& filter);
 
+	std::pair<bool, std::string> filter_error;
```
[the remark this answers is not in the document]
timwoj: It has to be `protected` so that it can be set via `pcap/Source.cc`. I could add a `Set` method since we already have a `Get` method, I suppose.

src/iosource/BPF_Program.cc
```diff
@@ -120,12 +122,17 @@ bool BPF_Program::Compile(int snaplen, int linktype, const char* filter, uint32_
 	int err = pcap_compile_nopcap(snaplen, linktype, &m_program, (char*)filter, optimize, netmask,
 	                              my_error);
 	if ( err < 0 )
-		errbuf = std::string(my_error);
+		auto errstr = std::string(pcap_geterr(pcap));
+	error = {IsCatastophicError(errstr), util::fmt("pcap_compile(%s): %s", filter, errstr.c_str())};
```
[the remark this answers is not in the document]
timwoj: > Is this somehow missing braces?

What do you mean?

> IMO the caller can format it nicer if they want to and pre-formatting makes a callers live more difficult

Leaving it with `pcap_compile` there indicates what function it was that was called that made it fail. I could go either way on removing it, but I feel like that makes it more descriptive. At most I could see removing the filter string that was passed.

src/iosource/PktSrc.h
```diff
@@ -98,11 +98,23 @@ class PktSrc : public IOSource
...
+	 * Returns a reason why BPF compilation failed. This can be called
+	 * if PrecompileBPFFilter or PrecompileFilter returns false to
+	 * determine a reason.
+	 *
+	 * @return a pair containing a bool indicating whether the error is
+	 * considered catastrophic and an error string from the compilation.
+	 * If the bool is true, this indicates that setup using the filter
+	 * should exit with an error.
+	 */
+	std::pair<bool, std::string> FilterCompilationError() const { return filter_error; }
```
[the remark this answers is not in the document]
timwoj: > Maybe GetFilerCompilationError() ?

We have lots and lots of places in the code base where there is no `Get` on methods that simply return a value. I don't know why it's like that, but this just follows that.

> Also wonder if a struct with .fatal or .catastrophic (or even enum status code OK, FATAL, ... rather than bool) would be worth it?

That sounds like a good idea. `bool` is never terribly descriptive.

src/iosource/PktSrc.cc
```diff
@@ -31,6 +31,7 @@ PktSrc::PktSrc()
 	{
 	have_packet = false;
 	errbuf = "";
+	filter_error = std::make_pair(false, "");
```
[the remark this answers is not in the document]
timwoj: Yep.

src/iosource/BPF_Program.cc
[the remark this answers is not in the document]
timwoj: > What do you mean?

Never mind, I see it now.

src/iosource/pcap/Source.cc
```diff
@@ -263,17 +263,12 @@ void PcapSource::DoneWithPacket()
...
 	}
 
-bool PcapSource::PrecompileFilter(int index, const std::string& filter)
-	{
-	return PktSrc::PrecompileBPFFilter(index, filter);
-	}
-
 detail::BPF_Program* PcapSource::CompileFilter(const std::string& filter)
 	{
 	std::string errbuf;
```
awelzel, line 268: This errbuf is now unused within this function?

Should the following lines from the PktSrc class be replicated within here as well?

https://github.com/zeek/zeek/pull/2396/files#diff-a7e13a32da0ee2af046a24b86a43b3325625943ee076789c70bc728575dde1d0R214-R215

I'm having a bit of a hard-time piecing this all together, but without the lines, the actual filter_error won't be propagated to the sources's `errbuf` that is then later retrieved via `Pcap::error()`, correct?

https://github.com/zeek/zeek/blob/f8bc23d3e1a6542b8e2212537200b805e49992a6/scripts/base/frameworks/packet-filter/main.zeek#L287-L292

src/iosource/PktSrc.h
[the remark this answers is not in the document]
awelzel: Gotcha, I think I was just surprised as it was the first one. I think I very slowly get how this is all fitting together..

src/iosource/pcap/Source.cc
timwoj: Yep, fixed. I missed updating that in one of the prior commits.

timwoj:
This is getting pretty close. One thing I noticed while doing a bit more testing with it this morning was the log entries in notice.log:

```
Failure Installing packet filter failed ip or not ip    -       -       -       -       -       Notice::ACTION_LOG    (empty) 3600.000000     -       -       -       -       -
```

and in packet_filter.log:

```
#fields ts      node    filter  init    success
#types  time    string  string  bool    bool
1662655749.172863       zeek    ip or not ip    T       F
```

For the notice, we aren't really failing here. That log just needs to be updated to say something more accurate. For the packet_filter entry, I'm debating if that last field should be an `F`. It didn't technically fail, but was it really a success?

src/iosource/pcap/Source.cc
timwoj: > [zeek/scripts/base/frameworks/packet-filter/main.zeek](https://github.com/zeek/zeek/blob/f8bc23d3e1a6542b8e2212537200b805e49992a6/scripts/base/frameworks/packet-filter/main.zeek#L287-L292)

This part's definitely a bit of a problem. Basically this is the cause of the notice I mentioned below. `Pcap::precompile_filter` doesn't actually fail anymore except on `FATAL` problems, so it doesn't fall into this block. `install_pcap_filter` *does* fail though, so it reports the notice later on. I can confirm that does the right thing on `FATAL` errors though:

```
tim@coregeek2 build% zeek -f "kaput" -r ~/Downloads/pop3.pcap
error: Error returned during BPF configuration: pcap_compile(kaput): can't parse filter expression: syntax error
fatal error in /Users/tim/Desktop/projects/zeek/scripts/base/frameworks/packet-filter/./main.zeek, line 292: cannot compile BPF filter "kaput": pcap_compile(kaput): can't parse filter expression: syntax error
```

I'll have to look into why `install_pcap_filter` doesn't like what's happening.

timwoj: Alright, `install_pcap_filter` is failing because although `precompile_filter` didn't return a failure it also didn't store a `BPF_Filter` object into the `DefaultPcapFilter` index. That's because `BPF_Filter` didn't really succeed at compiling the filter. I wonder if it would be better for the `BPF_Filter` object to exist but store the compilation state locally instead of returning it back to the PktSrc. I'll have to consider that one a bit more.

This is definitely a want-to-have change, so it's ok if it doesn't make it into 5.1. I'd rather get it right first than rush it through.

timwoj:
I'm finally happy with the state of this so I'm moving it from a draft to a regular PR.

timwoj:
This PR includes changes to the zeek-testing, zeek-testing-private, and zeek-docs repos. There are branches with the same name in those repos that need to be merged when this one is merged.

timwoj:
> I think the only thing missing is a btest that triggers warnings, but I'm waiting for an OK from the person who gave me the pcap that originated this. If they don't get back to me I'll see about building one with scapy or something.

I ended up building a single-packet pcap with scapy that exercises this code, and added a test in https://github.com/zeek/zeek/pull/2396/commits/74d9325cfe78dec5db0d328a367880420c7e45af

src/iosource/BPF_Program.cc
```diff
@@ -120,12 +120,19 @@ bool BPF_Program::Compile(int snaplen, int linktype, const char* filter, uint32_
 	int err = pcap_compile_nopcap(snaplen, linktype, &m_program, (char*)filter, optimize, netmask,
 	                              my_error);
 	if ( err < 0 )
-		errbuf = std::string(my_error);
+		{
+		state = GetStateFromMessage(errstr);
```
awelzel, line 124: `errstr` isn't defined here. Should that be std::string(my_error) instead?

But then, my Debian box, a Centos-7 container and FreeBSD don't seems to have that error parameter and all CI is green as well, is this actually live code anywhere? Custom libpcap?

Maybe we can replace this code with`#error Please report to use what you're running` or reject configs with `LIBPCAP_PCAP_COMPILE_NOPCAP_HAS_ERROR_PARAMETER` further up rather than have it untested?

scripts/base/init-bare.zeek
```diff
@@ -5496,6 +5496,17 @@ export {
 	const flowbuffer_contract_threshold = 2 * 1024 * 1024 &redef;
 }
 
+module Pcap;
```
awelzel, line 5499: There's already a module Pcap on line 5121 further up. Any reason it couldn't be there?

src/iosource/BPF_Program.cc
```diff
@@ -120,12 +120,19 @@ bool BPF_Program::Compile(int snaplen, int linktype, const char* filter, uint32_
...
+		state_message = util::fmt("pcap_compile(%s): %s", filter, pcap_geterr(pcap);
+		}
 #else
-	int err = pcap_compile_nopcap(snaplen, linktype, &m_program, (char*)filter, optimize, netmask);
+	int err = pcap_compile_nopcap(static_cast<int>(snaplen), linktype, &m_program, (char*)filter,
+	                              optimize, netmask);
 
+	// We have no way of knowing what the error actually was because pcap_compile_nocap doesn't
+	// return an error string nor any other information, so just assume every failure is
+	// fatal.
```
awelzel, line 133: It seems all that  `pcap_compile_nopcap()` does is to open a dead `pd` with the given snaplen and linktype and then call pcap_compile() on it.

https://github.com/the-tcpdump-group/libpcap/blob/d09102e61a868f96fb68dea8fcf08b06388bb450/gencode.c#L872-L885

As an alternative to the comment, we could probably duplicate those few lines and instantiate the pcap handle and then pass it on to
`BPF_Program::Compile(pcap_t* pcap, ...)` and capture the error there? It would also allow capturing/differentiate the `pcap_open_dead()` error.

scripts/base/frameworks/packet-filter/main.zeek
```diff
@@ -48,6 +48,9 @@ export {
 
 		## Indicate if the filter was applied successfully.
 		success: bool  &log &default=T;
+
+		## A string reason why the filter failed to be created/installed.
+		reason: string &log &optional;
```
awelzel, line 53: One comparable field is in dpd.log saying `failure_reason` indicating it should hold error/failure information. Would that be worth adding to make it more descriptive?

The naming we use for this field varies across layers (pcap_geterr/errstr/my_error, state_message and reason up here), but that might just me getting confused. I understand we specifically don't treat the message as error everywhere.

awelzel (commented):
I've left some more comments around dead-code and the nopcap version which I think could be open-coded to determine the error.

I think otherwise I'm happy to approve this though I don't feel super familiar with all this.

scripts/base/init-bare.zeek
timwoj: Nope, I just didn't look to see if it was already there.

src/iosource/BPF_Program.cc
timwoj: You also got me wondering whether `pcap_compile_nopcap` sets `pcap_error`, and since it just calls `pcap_compile` underneath it should. That means we do actually get an error string back from it similar to the `pcap_compile` version of the method.

src/iosource/BPF_Program.cc
timwoj: The manpage for pcap_compile_nopcap actually says this:

> A return of -1 indicates an error; the error text is unavailable. (pcap_compile_nopcap() is a wrapper around pcap_open_dead(), pcap_compile(), and pcap_close(); the latter three routines can be used directly in order to get the error text for a compilation error.) 

So yes, we could just duplicate that code and call `pcap_error` if one of those methods fails.

timwoj: Also, I think making that change would remove all of the concerns about the other version `pcap_compile_nopcap`, since we'd stop calling it anyways.

scripts/base/frameworks/packet-filter/main.zeek
timwoj: > One comparable field is in dpd.log saying failure_reason indicating it should hold error/failure information. Would that be worth adding to make it more descriptive?

Sure.

> The naming we use for this field varies across layers (pcap_geterr/errstr/my_error, state_message and reason up here), but that might just me getting confused. I understand we specifically don't treat the message as error everywhere.

I agree, but it's mostly just because those are the names of the things that are stored in the variables, such as `errstr` because we're calling `pcap_geterr`. Using `state` made sense to me because it's not really an error but more just the current state of the filter compilation process. `failure_reason` would make sense because it's only printed out on a failure (or well, a failure or a warning).

timwoj:
There's something busted about this branch on CI now. It keeps not being able to find the commit hash, even though if I re-run the job with terminal access I can clone it just fine.

src/iosource/BPF_Program.cc
awelzel: > Also, I think making that change would remove all of the concerns about the other version pcap_compile_nopcap, since we'd stop calling it anyways.

Nice. Two birds one stone. The generic error makes sense to me, too.

src/iosource/BPF_Program.cc
awelzel: > We'll find out fairly quickly if someone is using it. In the meantime we could leave the code in place #ifdef'd like it is, and make a note to remove it at a later date.

Resolving this thread as it's gone now. The `LIBPCAP_PCAP_COMPILE_NOPCAP_HAS_ERROR_PARAMETER` vestigial are still in cmake, but doesn't need to hold up this one.

timwoj:
I opened https://github.com/cirruslabs/cirrus-ci-web/issues/454 to figure out why this build is failing.

awelzel (approved):

PR #2398, opened 2022-09-03
Fix deprecation not flagged and incorrect line number in record

### LINKED ISSUE #2384: Deprecation not flagged in record construction
In the following, the three uses of a deprecated field throw three warnings:
```
type Foo: record {
        a: int;
        b: string &optional &deprecated="that's a no no";
};

event zeek_init()
        {
        local f1: Foo = [$a=1, $b="mhmm"];
        local f2: Foo = record($a=1, $b="mhmm");
        local f3 = Foo($a=1);
        f3$b = "mhmm";
        }
```
```
$ zeek ./test.zeek
warning in ././test.zeek, line 9: deprecated (Foo$b): that's a no no
warning in ././test.zeek, line 10: deprecated (Foo$b): that's a no no
warning in ././test.zeek, line 11: deprecated (Foo$b): that's a no no
```
But when moving the field access into the record constructor, Zeek remains silent:
```
event zeek_init()
        {
        local f4 = Foo($a=1, $b="mhmm");
        }
```
```
$ zeek ./test.zeek
$
```
Also, the line numbers reported in the first example seem curious. Here's the file with line numbers:
```
01 type Foo: record {
02      a: int;
03      b: string &optional &deprecated="that's a no no";
04 };
05
06 event zeek_init()
07      {
08      local f1: Foo = [$a=1, $b="mhmm"];
09      local f2: Foo = record($a=1, $b="mhmm");
10      local f3 = Foo($a=1);
11      f3$b = "mhmm";
12      }
```
I'd have expected the line numbers to be 8, 9, and 11 (or one less if counting from 0). If you remove line 11, we're down to two warnings:
```
$ zeek ./test.zeek
warning in ././test.zeek, line 9: deprecated (Foo$b): that's a no no
warning in ././test.zeek, line 10: deprecated (Foo$b): that's a no no
```

AmazingPP:
Fixes #2384


src/Expr.cc
```diff
@@ -3443,6 +3443,8 @@ RecordConstructorExpr::RecordConstructorExpr(RecordTypePtr known_rt, ListExprPtr
 				ExprError(err.c_str());
 				}
 			}
+		else if ( known_rt->IsFieldDeprecated(i) )
+			reporter->Warning("%s", known_rt->GetFieldDeprecationWarning(i, false).c_str());
```
ckreibich, line 3447: Thank you! How about we switch all the `reporter->Warning()` calls after `IsFieldDeprecated()` over to `Warn()`, as you do below? The various combos of manual location push/pop and reporter invocations in this file are pretty subtle, and that would make it a little more consistent. Plus, the extra detail in the deprection output seems useful.

AmazingPP: It's fine!

AmazingPP:
The `reporter::Warning()` uses `line_number` to print the line numbers.
https://github.com/zeek/zeek/blob/0cce28308697b9a5398f4651a5cfeb262543fc3a/src/input.h#L34

In general, The `parse.y` will build `Expr` and `Stmt` and other things first. Then `scan.l` will increase `line_number` each time when it scans a `\n`.

https://github.com/zeek/zeek/blob/0cce28308697b9a5398f4651a5cfeb262543fc3a/src/scan.l#L211-L215

But here, `line_number` will be increased before `build_local` called. This causes `reporter::Warning()` to output the wrong line number (Always 1 greater than the correct line number).

https://github.com/zeek/zeek/blob/0cce28308697b9a5398f4651a5cfeb262543fc3a/src/parse.y#L1863-L1869

awelzel (approved):
Just one nit trying to move the new warning generation in the other loop, otherwise lgtm.

Issue #2033, opened 2022-04-08
PktSrc support for batch/ring interfaces


emillynge:
Zeek Plugins exist for  a multitude of packet sources that are capable of reading packets in batches:
 * https://github.com/J-Gras/zeek-af_packet-plugin
 * https://github.com/zeek/zeek-netmap
 * https://github.com/ntop/bro-pf_ring

The main appeal of these drivers, are that they allow to poll once, and the process a whole batch of packets.

Having looked into `PktSrc.cc` ,`Manager.cc` and `RunState.cc`, it seems that the main loop is written in a way that does not allow the above plugins to take advantage of batch reading.

1) a poll is forced *and* can only happen every 10 iterations due to `poll_interval`.  Maybe the `pkt_src` should be allowed to "own" this parameter?
2) upon `PktSrc.Process()` only a single packet is read. Is there a reason for not just reading as many packets as possible until exhausted?

A common characteristic of these plugins are, that they can generally tell whether the source is "ready" (has more packets to read), without doing a poll. It seems it should be possible to utilize this, in  `Manager.FindReadySources(std::vector<IOSource*>* ready)

If I were to contribute some PR that address some of these issues, what would be your preferred solution?

timwoj:
I like this idea so I put it into the 5.1 milestone.

emillynge:
Awesome! 

When are you aiming at a 5.1 release?

What sort of approach would you support in terms of API changes to `PktSrc` base class?

timwoj:
I honestly can't say. We try to keep to roughly a 4 month cadence on releases but things have a tendency to slip over time. As for the API changes, we won't know until someone digs into it. Also, things being the milestone aren't a guarantee of them getting done by that release, but more just planning.

PR #2401, opened 2022-09-06
Include in Jan's AF_PACKET plugin as builtin plugin

awelzel:
This has come up a few times and the motivation is mainly better "first timer"
experience with Zeek. Concretely, if one wants to run a Zeek cluster with
multiple workers and reasonable load balancing on Linux, AF_PACKET is a decent
start. Without AF_PACKET support being built into Zeek, however, a new user's
next experience is that of setting up a development environment in order
to compile an external plugin (think compiler, kernel headers, zkg, ...).
Only to get what could be termed basic functionality.

This is using the ZEEK_INCLUDE_PLUGINS infrastructure. I've used the all
upper case spelling of AF_PACKET in the help output because it seems everyone
else references/writes it like that. I think we should also write it
like that in the docs.


awelzel:
I had chatted with @rsmmr about this and he could also see shipping the plugin's code within the Zeek source. I'm leaning towards preferring that too, but building it in as a submodule was so easy. I might open a separate PR / branch that pulls in the source code, just to have that as a contrast.

The main reason why including the source code with Zeek is that it would a) provide an example of a more advanced packet source within the Zeek tree. Examples of other advanced packet sources need special hardware (e.g., myricom, napatech) and more difficult to experiment with.

As of experimentation, having the AF_PACKET source code within the Zeek tree may make iterating on changes to the PktSrc interface easier (thinking #2033).

ckreibich:
@J-Gras, fyi!

ckreibich:
The main argument I'm aware of for going the submodule route is that resulting improvements to the plugin become available also to users who pull in the plugin via a `zkg` package. I'm not sure if that's decisive in the end. We could at first use the submodule, and later (say in the 6.x cycle) fully pull it in when people should no longer be using a non-AF_PACKET-enabled Zeek — but that might be overthinking it.

If we stick with the submodule, we should "graduate" the plugin from Jan's space into Zeek proper (assuming Jan you're okay with that!).

ckreibich:
Relates to zeek/zeek-docs#87

J-Gras:
I like the submodule approach! Given that there are some todos, likely to be addressed in this process (e.g., https://github.com/J-Gras/zeek-af_packet-plugin/issues/29), and it might take some time for people to update to newer Zeek versions, the submodule approach seems to be a good choice. I am also happy to migrate the repo to the Zeek namespace on Github. This should make experimentation with changes to the Packet Source interface seamless.

awelzel:
> Given that there are some todos, likely to be addressed in this process (e.g., https://github.com/J-Gras/zeek-af_packet-plugin/issues/29),

See my last commit - I think using just the latest is fine for now, but adding kernel headers to all the CI things is annoying and also made me wonder whether that dependency is "real":

* Suricata provides AF_PACKET support without requiring kernel headers (and they support it as first-class citizen)
* I've commented out the kernel headers related things in zeek-af_packet-plugin and it still compiles and de-installed the kernel-headers package locally and everything is happy. The linux related headers used by zeek-af_packet-plugin on my system are coming from `linux-libc-dev`:
  ```
  $ sudo dpkg -S /usr/include/linux/if_packet.h
  linux-libc-dev:amd64: /usr/include/linux/if_packet.h
  ```

I'll check tomorrow, but I think the linux-headers dependency can just go unless there's a good reason. It's believe it's needed for building out-of-tree modules that interact with the kernel API, but AF_PACKET is on a different level and just user-space headers should be fine. I might be missing some historic background, but maybe that can be history?

awelzel:
This PR is currently pointing at a branch in Jan's repo where the kernel-headers dependency has been removed and seems that all distros we have in `ci/` are happy with this:

https://github.com/J-Gras/zeek-af_packet-plugin/pull/36

The btests got some more normalization tweaks to ignore `Zeek_AF_Packet` related things. The bigger effort will likely be docs :-)

J-Gras:
I guess you are right: The dependency isn't needed anymore. My understanding is that kernel headers are also required by user space programs that interact with the OS, but Linux distributions are expected to provide kernel headers (https://kernelnewbies.org/KernelHeaders). However, these headers might not be up to date/match the running kernel. If I remember correctly, the plugin had to support CentOS 6 (2.6 kernels), when it was created. So I think the dependency was necessary back then. Nowadays, AF_Packet with tpacket v3 should be widely available, so that the defaults are just fine to use. The dependency was just never challenged. So, thanks a lot!

testing/btest/coverage/bare-load-baseline.test
```diff
@@ -12,5 +13,6 @@
 # @TEST-EXEC: test -e loaded_scripts.log
 # @TEST-EXEC: cat loaded_scripts.log | egrep -v '#' | awk 'NR>0{print $1}' | sed -e ':a' -e '$!N' -e 's/^\(.*\).*\n\1.*/\1/' -e 'ta' >prefix
 # @TEST-EXEC: (test -L $BUILD && basename $(readlink $BUILD) || basename $BUILD) >buildprefix
-# @TEST-EXEC: cat loaded_scripts.log | sed "s#`cat buildprefix`#build#g" | sed "s#`cat prefix`##g" >canonified_loaded_scripts.log
+# @TEST-EXEC: cat loaded_scripts.log | sed "s#`cat buildprefix`#build#g" | sed "s#`cat prefix`##g" >prefix_canonified_loaded_scripts.log
+# @TEST-EXEC: grep -v 'Zeek_AF_Packet' prefix_canonified_loaded_scripts.log > canonified_loaded_scripts.log
```
ckreibich, line 17: Arne what motivates this filtering, also below in the plugin? It's fine with me, I'm just trying to understand it in relation to the Spicy plugin, where we don't filter either. Is it the conceptual fact that the scripts live outside of Zeek proper?

ckreibich:
Sorry for the delay here, folks. The PR looks good to me. My only suggestion is that we migrate the package prior to merging. I just forked the plugin repo into the Zeek project before I realized that there's proper way to transfer ownership, in the "Danger" settings of the original repo. If that sounds okay to you @J-Gras, please feel free to transfer any time and we'll look out for it over here. (I've deleted the fork.)

Fwiw, I feel pretty good about the submodule route. We keep it functional and maintained for folks running older Zeeks, I like the symmetry with the Spicy plugin, and we can always change it down the road. Also, wouldn't problems like #2403 also come up if it's included directly, not via a submodule? We'd still be building a plugin.

testing/btest/coverage/bare-load-baseline.test
awelzel: In short: Avoiding baseline differences between Linux and OSX developer systems for those tests in the fairly straight-forward way. On OSX there'll never be AF_Packet scripts loaded while on Linux by default there will.

Spicy is platform-agnostic and build by default everywhere, that's the relation/difference I see.

Though, I've run multiple times into the situation where I used a non-spicy-included Zeek build locally for dev just to be told by CI later the plugin/hooks.zeek baseline is outdated. That test does not run locally if one has spicy disabled and I had wanted to avoid introducing more of that (or not running certain tests on OSX just because AF_Packet shows in the baselines).

awelzel:
> Also, wouldn't problems like #2403 also come up if it's included directly, not via a submodule? We'd still be building a plugin.

Yes. Technically the problem will continue to exist.

On the non-technical side, though, if AF_Packet is part of the Zeek source tree proper, building it as an external plugin effort and can be stamped unsupported or "you know what you're doing". With keeping the submodule, we kind of imply keeping the external repo and supporting installing it as an external plugin.

ckreibich:
Fyi the migration of Jan's package is now complete: https://github.com/zeek/zeek-af_packet-plugin

awelzel:
>  Fyi the migration of Jan's package is now complete: https://github.com/zeek/zeek-af_packet-plugin

Rebased, updated, and moved the NEWS entry from the 5.1 section up into 5.2 now.

awelzel:
CI failed due to centos-7 being old enough to not have the `TP_STATUS_CSUM_VALID`.

Proposed fix to the plugin here: https://github.com/zeek/zeek-af_packet-plugin/pull/43

ckreibich:
Sorry for the delay Arne — it all looks good to me. If there's nothing more from you and Jan, this'll go in tomorrow. Thank you both!

ckreibich (approved):

ckreibich:
Fyi I merged this with a submodule tweak to point it at Jan's new 4.0.0 release commit, i.e. the latest in its `master`.

PR #2404, opened 2022-09-08
Add complex construction support for table constructors

### LINKED ISSUE #2385: Table constructors allow creating values with non-Zeek-accessible types
https://github.com/zeek/zeek/blob/686eb54f9590fb491582f1bab10b754c9bd419fc/testing/btest/bifs/table_values.zeek#L8 constructs a complex table using:
> table(
		["web"] = { [80/tcp, "http"], [443/tcp, "https"] },
		["login"] = { [21/tcp, "ftp"], [23/tcp, "telnet"] });

If you print the type of this table, you get: `table[string] of list of list of port,string,list of port,string`.  In general, I don't think we (currently) want to expose `list` types to script-land when we can avoid it, so I think this constructor should either produce an error, or result in a more reasonable type.  Regarding what's more reasonable, the following works fine:
> table(
                ["web"] = set( [80/tcp, "http"], [443/tcp, "https"] ),
                ["login"] = set( [21/tcp, "ftp"], [23/tcp, "telnet"] ));

with the unsurprising type `table[string] of set[port,string]`.

AmazingPP:
Fixes #2385 

Hi @vpax -  I'm not sure if this is correct. So, Please review it.

Originally, I also wanted to implement complex construction for `set` and all other things that use `ListEpxr` to construct. But Zeek scripts syntax parsing can only parse the nesting of `ListEpxr` for `table` constructors.


vpax:
Apologies for taking so long to circle back to this.  My main question is whether we event _want_ to support such constructors.  I find the alternatives given in https://github.com/zeek/zeek/issues/2385 of using explicit `set()` constructors etc. simpler and cleaner.  @timwoj you put some facet of that on the list for 5.1 - do you want to chime in?

AmazingPP:
> My main question is whether we event _want_ to support such constructors. I find the alternatives given in #2385 of using explicit `set()` constructors etc. simpler and cleaner. 

Yes. I think we need to discuss and decide if complex implicit constructs are acceptable.
- If it is acceptable, let's make it produce the correct type.
- if not, I can change this PR to "complex implicit constructs should throw errors", then force the user to construct it explicitly.

timwoj:
> @timwoj you put some facet of that on the list for 5.1 - do you want to chime in?

I added this to the 5.1 milestone but it definitely doesn't have to make it in if it's complicated. At this point I'd actually prefer to leave it out. I don't have any real opinion on whether we want it or not, it just looked like a bug that we should fix and added it to the milestone.

timwoj:
@vpax Are your feelings still the same on this? Should I just close it and have some more discussion in the issue instead of leaving this PR open indefinitely?

vpax:
> @vpax Are your feelings still the same on this? Should I just close it and have some more discussion in the issue instead of leaving this PR open indefinitely?

Sorry for - again - taking so long to reply.  Yes, I think we should just go with explicit constructors, rather than complex-and-a-bit-obscure implicit ones based on braces etc.  So I'd advocate for closing this (with many thanks to @AmazingPP for exploring it).

Issue #2387, opened 2022-08-29
Runtime of copy() for tables of different sizes varies hugely


awelzel:
Splitting this out from #2386: Runtime for copying a table of a simple string to count table varies wildly and unexpectedly. Initial population time and pure iteration time appear to increase reasonably, while `copy()` time increases 100 fold.

Running below script shows that the time required for copying() increases from ~1second for 1mio entries to ~50 seconds and 100seconds for 1.1mio and 1.2mio entries, respectively. At 1.7mio entries, the time is back at below 2 seconds.

```
$ time zeek ./large-copy.zeek 
Size, 1000000, fill, 0.713404, iterate, 0.453252, copy, 0.91069
Size, 1100000, fill, 1.137778, iterate, 0.492146, copy, 50.985543
Size, 1200000, fill, 1.3587, iterate, 0.514669, copy, 101.756562
Size, 1300000, fill, 1.444776, iterate, 0.577019, copy, 105.51632
Size, 1400000, fill, 1.581012, iterate, 0.608411, copy, 68.34778
Size, 1500000, fill, 1.683537, iterate, 0.670468, copy, 18.79153
Size, 1600000, fill, 1.990621, iterate, 0.709804, copy, 9.908489
Size, 1700000, fill, 2.000903, iterate, 0.736333, copy, 1.367809
Size, 1800000, fill, 2.257185, iterate, 0.804593, copy, 1.481103
Size, 1900000, fill, 2.200775, iterate, 1.085882, copy, 1.5915
Size, 2000000, fill, 2.303482, iterate, 0.908776, copy, 1.784987
Size, 2100000, fill, 2.320749, iterate, 1.087852, copy, 10.341843
```

```
module Test;

global tbl: table[string] of count;

function fill_table(limit: count)
        {
        local i = 0;
        while ( i < limit )
                {
                ++i;
                tbl[cat(i)] = i;
                }
        }

function iterate_table(t: table[string] of count)
        {
        local c = 0;
        for ( [k], v in t )
                c += v;
        }

event zeek_init()
        {
        local limit = 100000000;

        local one_million = 1000000;
        local i = one_million;
        local tbl2: table[string] of count;

        while ( i < limit )
                {
                tbl = table();
                local st = current_time();
                fill_table(i);
                local etf = current_time();
                iterate_table(tbl);
                local itf = current_time();
                tbl2 = copy(tbl);
                local etc = current_time();
[...]

PR #2410, opened 2022-09-09
Minor cleanups in Dict code

timwoj:
This is just a few minor changes I found while investigating #2387

### DIFF
--- a/src/Dict.h
+++ b/src/Dict.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <algorithm>
+#include <cmath>
 #include <cstdint>
 #include <fstream>
 #include <memory>
@@ -456,7 +457,7 @@ public:
 			{
 			// If an initial size is speicified, init the table right away. Otherwise wait until the
 			// first insertion to init.
-			log2_buckets = Log2(initial_size);
+			log2_buckets = static_cast<unsigned char>(std::log2(initial_size));
 			Init();
 			}
 
@@ -941,14 +942,6 @@ private:
 		}
 
 	// bucket math
-	int Log2(int num) const
-		{
-		int i = 0;
-		while ( num >>= 1 )
-			i++;
-		return i;
-		}
-
 	int ThresholdEntries() const
 		{
 		// Increase the size of the dictionary when it is 75% full. However, when the dictionary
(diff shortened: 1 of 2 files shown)


src/Dict.h
```diff
@@ -728,10 +729,9 @@ template <typename T> class Dictionary
 	/// The capacity of the table, Buckets + Overflow Size.
 	int Capacity(bool expected = false) const
```
awelzel, line 730: Given you're doing cleanups and I stumbled over that: Would you be up for removing the expected arg here? And either open-code the `bucket_count` access at the only caller that uses `expected=true` or introduce an `ExpectedCapacity()` helper? The callsite with `Capacity(true)` is hardly descriptive.

src/Dict.h
```diff
@@ -931,25 +931,23 @@ template <typename T> class Dictionary
...
 
+	void SetLog2Buckets(int value)
+		{
+		log2_buckets = value;
+		bucket_count = 1 << log2_buckets;
+		bucket_capacity = (1 << log2_buckets) + log2_buckets;
+		}
+
 	/// Buckets of the table, not including overflow size.
 	int Buckets(bool expected = false) const
```
awelzel, line 942: I didn't see a `Buckets(true)` caller, so maybe the expected parameter can be removed?

src/Dict.h
```diff
@@ -1057,10 +1057,11 @@ template <typename T> class Dictionary
 		{
 		ASSERT(table && -1 <= position && position < Capacity());
 
+		auto current_cap = Capacity();
```
awelzel, line 1060: Wonder if for consistency with the other lines this should use `int` instead of `auto`, but not sure.

https://github.com/zeek/zeek/pull/2410/files#diff-38cf25ea2152135d0b4bbd2f78b1e6cd3357b3497f2fd9d53275ecabc2c3169dR1034

awelzel: May also apply for the other one.

timwoj:
All of the above sounded good to me, and are fixed in the fixup commit. I also added one more commit that marked the move operations as `noexcept`, like they should have been originally (clang-tidy complains about this).

awelzel (approved):

PR #2412, opened 2022-09-09
Fix iteration over ordered Dict objects

### LINKED ISSUE #2392: `DictIterator` does not honor `DictOrder::ORDERED` mode
I added this unit test to Dict.cc:

```
TEST_CASE("dict ordered iteration")
	{
	PDict<uint32_t> dict(DictOrder::ORDERED);

	// These key values are specifically contrived to be inserted
	// into the dictionary in a different order by default.
	uint32_t val = 15;
	uint32_t key_val = 5;
	auto key = std::make_unique<detail::HashKey>(key_val);

	uint32_t val2 = 10;
	uint32_t key_val2 = 25;
	auto key2 = std::make_unique<detail::HashKey>(key_val2);

	uint32_t val3 = 30;
	uint32_t key_val3 = 45;
	auto key3 = std::make_unique<detail::HashKey>(key_val3);

	dict.Insert(key.get(), &val);
	dict.Insert(key2.get(), &val2);
	dict.Insert(key3.get(), &val3);

	int count = 0;

	for ( const auto& entry : dict )
		{
		auto* v = static_cast<uint32_t*>(entry.value);
		uint32_t k = *(uint32_t*)entry.GetKey();

		// The keys should be returned in the same order we inserted
		// them, which is 5, 25, 45.
		if ( count == 0 )
			CHECK(k == 5);
		else if ( count == 1 )
			CHECK(k == 25);
		else if ( count == 2 )
			CHECK(k == 45);

		count++;
		}
	}
```

It returns the keys in the order that would be expected if the dictionary was in `UNORDERED` mode and the test fails. This affects being able to loop using ranged-for loops in C++ as well as `for` statements in script-land.

timwoj:
This PR fixes dictionary iterators when the dictionary is in ordered mode. This was discovered while looking at culprits for #2387. Previously, iteration on an ordered dictionary didn’t take the order into account at all and just used the normal iteration. The commits in this can be reviewed separately since they’re all self contained. For the iteration changes, the major difference is that we’re now storing the order as a vector of `HashKey`. This adds a layer of stability to the ordering without needing to keep the entire `DictEntry` in the vector. There’s an additional `LookupEntry` method that was added as well to facilitate this.

Also added are copy/move operators and constructors plus equality/inequality operators to `HashKey` to support using it in a vector, plus unit tests to exercise them.

Fixes #2392

### DIFF
--- a/src/Dict.h
+++ b/src/Dict.h
@@ -163,6 +163,8 @@ public:
 	bool operator!=(const DictEntry& r) const { return ! Equal(r.GetKey(), r.key_size, r.hash); }
 	};
 
+using DictEntryVec = std::vector<detail::HashKey>;
+
 	} // namespace detail
 
 template <typename T> class DictIterator
@@ -198,6 +200,8 @@ public:
 		dict = that.dict;
 		curr = that.curr;
 		end = that.end;
+		ordered_iter = that.ordered_iter;
+
 		dict->IncrIters();
 		}
 
@@ -215,6 +219,8 @@ public:
 		dict = that.dict;
 		curr = that.curr;
 		end = that.end;
+		ordered_iter = that.ordered_iter;
+
 		dict->IncrIters();
 
 		return *this;
@@ -234,6 +240,7 @@ public:
 		dict = that.dict;
 		curr = that.curr;
 		end = that.end;
+		ordered_iter = that.ordered_iter;
 
 		that.dict = nullptr;
 		}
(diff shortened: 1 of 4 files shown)


src/Dict.h
```diff
@@ -629,7 +691,10 @@ template <typename T> class Dictionary
 		ASSERT(num_entries >= 0);
 		// e is about to be invalid. remove it from all references.
 		if ( order )
-			order->erase(std::remove(order->begin(), order->end(), entry), order->end());
+			{
+			auto hk = entry.GetHashKey();
```
rsmmr, line 695: This seems unnecessarily expensive, can we do this without copying the key into a new `unique_ptr`?

src/Hash.cc
```diff
@@ -184,6 +185,29 @@ HashKey::HashKey(const void* arg_key, size_t arg_size, hash_t arg_hash, bool /*
 	key = (char*)arg_key;
 	}
 
+HashKey::HashKey(const HashKey& other) : HashKey(other.key, other.size, other.hash) { }
+
+HashKey::HashKey(HashKey&& other) noexcept
```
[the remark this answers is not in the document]
timwoj: The move constructor can't be because it explicitly sets certain fields in the other object to certain values to make sure it's in a valid-but-invalid state.

src/Hash.cc
```diff
@@ -184,6 +185,29 @@ HashKey::HashKey(const void* arg_key, size_t arg_size, hash_t arg_hash, bool /*
...
+
+	other.size = 0;
+	other.is_our_dynamic = false;
+	other.key = nullptr;
+	}
+
+HashKey::~HashKey()
+	{
+	if ( is_our_dynamic )
+		delete[](char*) key;
```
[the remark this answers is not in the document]
timwoj: I'm actually not entirely sure why this is typecasted actually. We're storing it as a char* already, `reinterpret_cast`ed from a double array in `HashKey::Allocate`.

src/Dict.h
timwoj: I changed it to be a simple for-loop looking for the key data. Alternatively I could make it a remove_if and a lambda, but it results in the same code in the end.

timwoj: After running into a memory leak with the last change (because move operators are overwriting things as the element is removed), I'm kinda wondering if I need to store the whole key in the order vector after all. Is the hash generated of the key guaranteed to be unique to the point that I could just store the hash value? Then I could add another lookup method to Dictionary (maybe make it protected) that could take the hash value and do the lookup from that?

rsmmr: No, (very) strictly speaking the hash won't be sufficient. But now I'm confused, I thought you are already storing the full key in the ordering vector. Need to take another look later.

timwoj: > But now I'm confused, I thought you are already storing the full key in the ordering vector. Need to take another look later.

It is. I was just pondering whether it'd be possible to not have to copy the whole key over there. Not a big deal.

rsmmr: Got it, misread. So one idea: could the ordering vector just store a raw pointer to the data from the actual entry? As long as we can manage life-times consistently, that should still be safe.

timwoj: There was a reason I didn't do that, but it's currently escaping me. I want to say it was something about the way we store the HashKey objects in the DictEntry, and the pointers getting screwed up when elements had to move around the Dictionary during remaps.

rsmmr: Makes sense.

rsmmr:
I believe this is good then now, unless I"m missing any pending issue?

rsmmr (approved):

PR #2418, opened 2022-09-17
-O C++ maintenance

vpax:
As basically a companion to https://github.com/zeek/zeek/pull/2417, here's a similar set of changes for compiling scripts to C++. They're all conceptually straightforward - happy to explain any of them in more detail if helpful.

Note, one file I modified - `src/script_opt/CPP/RuntimeVec.cc` - consistently gets clang-formatted on my system differently than what CI evidently expects.  I tried to back out the low-level differences so hopefully it'll pass, but if not it'll need some sort of manual fixup (that's hard for me to do!).


src/script_opt/CPP/RuntimeVec.cc
```diff
@@ -103,20 +104,25 @@ VEC_OP1(comp, ~, )
 
 // A kernel for applying a binary operation element-by-element to two
 // vectors of a given low-level type.
-#define VEC_OP2_KERNEL(accessor, type, op)                                                         \
+#define VEC_OP2_KERNEL(accessor, type, op, zero_check)                                             \
```
timwoj, line 107: I know this isn't related to this PR, but I'd really love to see if we can turn these into lambdas at some point instead of macros.

timwoj (approved):

src/script_opt/CPP/RuntimeVec.cc
vpax: Ah yeah maybe we can.  I'm not used to thinking in terms of lambdas for C++ stuff so went the primitive route here, but point taken.

PR #2424, opened 2022-09-21
Skip darwin builds on zeek-security repo

timwoj:
The private zeek-security repo uses compute credits from Cirrus in order to run builds, since it's not a public repository. According to Fedor from Cirrus, the darwin builds eat a large number of the credits leading us to have to ask for more quite frequently. This PR adds some extra conditions to the `only_if` in the cirrus configuration to skip darwin builds on that repository but not others so that we still have normal coverage on the public repos.

I have some concerns about Darwin builds not being handled automatically for security PRs, but I think there's enough of us doing development on macOS that we'd encounter any problems early. Otherwise we would catch the failures once the issues are merged back into the public Zeek repo.

I tested this on a separate repo called `zeek-build-skip`. You can see the build that skipped the macOS builds at https://cirrus-ci.com/build/4697397258354688 (ignore the fact that they all failed, just note that the macOS builds are missing).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,14 +52,19 @@ sanitizers_resource_template: &SANITIZERS_RESOURCE_TEMPLATE
   greedy: true
 
 ci_template: &CI_TEMPLATE
+  # Rules for skipping builds:
+  # - Don't do darwin builds on zeek-security repo because they use up a ton of compute credits.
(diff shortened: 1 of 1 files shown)


0xxon (approved):
Looks good to me - I perhaps just would add a comment to the .cirrus.yml that it is skipped for compute credits. For future readers of that file :)

PR #2429, opened 2022-09-23
cmake: Exclude empty directories due to headers install

### LINKED ISSUE #2422: Default installation creates empty directory `include/zeek/script_opt/CPP/maint`
In zeek/spicy#1267 @leres reported a number of empty directories being created on install, among them `include/zeek/script_opt/CPP/maint`. We should check whether we can omit that directory from installation.

awelzel:
This followed some pointers from here and opted for explicit exclusion:

https://stackoverflow.com/questions/55451084/cmake-files-matching-pattern-copies-empty-directories

Fixes #2422

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -584,6 +584,12 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
         PATTERN "3rdparty/*" EXCLUDE
         # The "zeek -> ." symlink isn't needed in the install-tree
         REGEX "^${escaped_path}$" EXCLUDE
+
+	# FILES_MATCHING creates empty directories:
+	# https://gitlab.kitware.com/cmake/cmake/-/issues/17122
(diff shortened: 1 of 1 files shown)


timwoj (approved):

src/CMakeLists.txt
```diff
@@ -584,6 +584,12 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
         PATTERN "3rdparty/*" EXCLUDE
         # The "zeek -> ." symlink isn't needed in the install-tree
         REGEX "^${escaped_path}$" EXCLUDE
+
+	# FILES_MATCHING creates empty directories:
+	# https://gitlab.kitware.com/cmake/cmake/-/issues/17122https://gitlab.kitware.com/cmake/cmake/-/issues/17122
```
timwoj, line 589: The URL got duplicated here. Looks fine otherwise.

Issue #2377, opened 2022-08-25
when statement with missing capture triggering heap-buffer-overflow


awelzel:
The following script with Zeek 5.0 and latest master ASAN build causes heap-buffer-overflow reports.

Adding `[url]` to `when` solves the warning and appears to fix the crash, but maybe there's more lurking?

(found while writing a test using ActiveHTTP and a when statement, this is a condensed down version)
```
# @TEST-EXEC: zeek -b %INPUT

redef exit_only_after_terminate = T;

@load base/utils/active-http

global COUNT = 0;

function async_fun(url: string): string
        {
        return when ( COUNT > 0 )
                {
                return "string result";
                }
        }

event run_test()
        {
        local url = "http://localhost:-1/metrics";
        when ( local string_result = async_fun(url) )
                {
                local endpoints: vector of string;
                local x = find_all("axaaxaa}", /a/);
                terminate();
                }
        timeout 0.1sec
                {
                for ( info in set("a", "b") )
                        print info;
                terminate();
                }
        }

event zeek_init()
        {
        schedule 0.1sec { run_test() };
        }
```

```
$ zeek -b when-capture-crash.zeek
warning in ./when-capture-crash.zeek, lines 25-31: "when" statement referring to locals without an explicit [] capture is deprecated: url (when (string_result = async_fun(url)) { x = find_all(axaaxaa}, /^?(a)$?/)terminate()} timeout 100.0 msecs { for ([info] in set(a, b)) print infoterminate()})
=================================================================
==2874716==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x604001363908 at pc 0x55791f8497a0 bp 0x7ffd66a6eef0 sp 0x7ffd66a6eee8
WRITE of size 8 at 0x604001363908 thread T0
    #0 0x55791f84979f in zeek::Val* std::__exchange<zeek::Val*, decltype(nullptr)>(zeek::Val*&, decltype(nullptr)&&) /usr/include/c++/10/bits/move.h:155
    #1 0x55791f84979f in zeek::Val* std::exchange<zeek::Val*, decltype(nullptr)>(zeek::Val*&, decltype(nullptr)&&) /usr/include/c++/10/utility:292
    #2 0x55791f84979f in zeek::IntrusivePtr<zeek::Val>::release() ../src/zeek/zeek/IntrusivePtr.h:130
    #3 0x55791f84979f in zeek::detail::Frame::ClearElement(int) ../src/Frame.cc:639
[...]

PR #2434, opened 2022-09-26
fix for deprecated when's where the inner frame is larger than the outer frame

vpax:
This fixes https://github.com/zeek/zeek/issues/2377, which arises from recent `when` changes incorrectly introducing (and then discarding) an inner scope for the `when` body & timeout.  The fix is to merge that scope with its surrounding scope.

### DIFF
--- a/src/Scope.cc
+++ b/src/Scope.cc
@@ -190,6 +190,20 @@ ScopePtr pop_scope()
 	return old_top;
 	}
 
+void merge_top_scope()
+	{
+	if ( scopes.size() < 2 )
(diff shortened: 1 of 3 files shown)


awelzel (approved):
I've tested the reproducer from #2377 and it works as expected.

It would be a test for deprecated functionality going away, so okay to not include it.

timwoj:
Is this a candidate to pull into a 5.0.3 release and add to 5.1?

vpax:
Yes, I think that'd be good. It fixes a potential write-to-unallocated memory.

PR #2437, opened 2022-09-26
packet_analysis: Do not raise analyzer_confirmation per-packet for tunnels

awelzel:
There's a logic error in the packet analyzer's AnalyzerConfirmation() method that causes analyzer_confirmation() events to be raised for every packet rather than stopping after the first confirmation which appears to have been the intention. This affects, for example, VXLAN and Geneve tunnels.

The optional arg_tag parameter was used for short-circuit'ing, but the return value of GetAnalyzerTag() used for setting the session state causing the disconnect.

In scenarios where Zeek receives purely tunneled monitoring traffic, this may result in a non-negligible performance impact.

Somewhat related, ensure the session state is set to violated before short-circuiting if no analyzer_violations are installed.

~Suggesting this as a 5.0.3 candidate.~


awelzel:
> Suggesting this as a 5.0.3 candidate.

Hmm - this must have been around since Zeek 4.1, actually. So maybe not that important. I'd expect the impact to be fairly visible with vxlan-only-traffic scenarios, however.

JustinAzoff:
This change seems to have a massive positive impact.

I benchmarked it using

```
hyperfine --export-markdown ~/tmp/vxlan_results.md -r 5 \
    -L zeek zeek,zeek.orig \
    -L pcap https_big_upload,https_big_upload_vxlan \
    '{zeek} -Cr ~/pcaps/{pcap}.pcap'
```

where 
 * `zeek.orig` is current master and `zeek` is this branch
 * `https_big_upload` is a single large tls connection, and the `_vxlan` one is the pcap wrapped in a vxlan tunnel.

which output

|:---|---:|---:|---:|---:|
| `zeek -Cr https_big_upload.pcap` | 3.866 ± 0.074 | 3.800 | 3.981 | 1.02 ± 0.02 |
| `zeek.orig -Cr https_big_upload.pcap` | 3.787 ± 0.041 | 3.719 | 3.822 | 1.00 |
| `zeek -Cr https_big_upload_vxlan.pcap` | 6.056 ± 0.079 | 5.955 | 6.169 | 1.60 ± 0.03 |
| `zeek.orig -Cr https_big_upload_vxlan.pcap` | 15.709 ± 0.150 | 15.563 | 15.904 | 4.15 ± 0.06 |

src/packet_analysis/Analyzer.cc
```diff
@@ -168,26 +168,30 @@ void Analyzer::Weird(const char* name, Packet* packet, const char* addl) const
 
 void Analyzer::AnalyzerConfirmation(session::Session* session, zeek::Tag arg_tag)
 	{
-	if ( session->AnalyzerState(arg_tag) == session::AnalyzerConfirmationState::CONFIRMED )
+	auto const& effective_tag = arg_tag ? arg_tag : GetAnalyzerTag();
```
rsmmr, line 171: Style nit:

```suggestion
	const auto& effective_tag = arg_tag ? arg_tag : GetAnalyzerTag();
```

src/packet_analysis/Analyzer.cc
```diff
@@ -168,26 +168,30 @@ void Analyzer::Weird(const char* name, Packet* packet, const char* addl) const
...
-	const auto& tval = arg_tag ? arg_tag.AsVal() : tag.AsVal();
-	event_mgr.Enqueue(analyzer_confirmation, session->GetVal(), tval, val_mgr->Count(0));
+	event_mgr.Enqueue(analyzer_confirmation, session->GetVal(), effective_tag.AsVal(),
+	                  val_mgr->Count(0));
 	}
 
 void Analyzer::AnalyzerViolation(const char* reason, session::Session* session, const char* data,
                                  int len, zeek::Tag arg_tag)
 	{
+	auto const& effective_tag = arg_tag ? arg_tag : GetAnalyzerTag();
```
rsmmr, line 188:
```suggestion
 const auto& effective_tag = arg_tag ? arg_tag : GetAnalyzerTag();
```

rsmmr (approved):
This is great. I do think it's one for 5.0.3.

src/packet_analysis/Analyzer.cc
awelzel: Again.. :-(

Seems with clang-format 14 there's now a QualifierAlignment option to help me ..

awelzel:
I have squashed in the `auto const` to const auto` change.

Issue #2399, opened 2022-09-06
Support disabling packet analyzers


awelzel:
This ticket is about tracking ideas and implementing functionality to disable packet analyzers.

Existing analyzers for tunnels have special `enable` flags.

https://github.com/zeek/zeek/blob/875e81883c666e2058a976bd84e5ebd734e9faf2/src/const.bif#L18-L22

These are not runtime changeable (though there may not be a good reason for that limitation).

Further, it's currently *not* possible to disable packet analyzers using, for example, the `AllAnalyzers::Tag` allocated to them. Let alone using the data passed to `analyzer_violation()` from packet analyzers (calling `disable_analyzer()` for a packet analyzer tag is not sensible as `disable_analyzer()` looks for a protocol analyzer attached to the connection).

Ideas:
* Remove analyzers out of the analyzer tree using their tags (and optionally identifier)
  * Need to know at which point in the tree they are located (or support recursive/wild-card removal)
  * This approach modifies the analyzer tree, so to re-enable analyzers, the caller requires knowledge of the "right" locations in the tree and identifiers

* Add a "enabled or disable" flag into the dispatcher table entries
  * The `AnalyzerPtr` entries become structs/classes of `{AnalyzerPtr, enabled}`
  * Can disable analyzers but keep them in the tree (again, recursively or at individual points)
  * Enabled/disabled lookup is fast (no indirection via `AnalyzerPtr` needed, within the memory of the table)
  * Easier to re-enable without needing to know the locations of analyzers/identifiers (recursively enable in the whole tree, for example)
  * Possible signature: `disable_packet_analyzer(parent, child, identifier: int =-1 &otional, recursive: bool &default=true)`
  * Analyzer tree structure is immutable

* Use the Component mechansim as done for protocol and file analyzers
  * May require a `->Lookup()` to get the component on a per-packet basis may be prohibitive

I'm thinking of sketching/implementing the second approach and see how it fares. Thoughts.

cc @J-Gras

ckreibich:
The main driver for this is symmetry with protocol and file analyzers, right? If so, the third option sounds appealing to me — it looks like the lookup in the packet analyzers' dispatcher is just a vector index. I like the second option too, but it might be worth checking if the third is in fact problematic before going a new route for the packet analyzers.

rsmmr:
[...]

PR #2443, opened 2022-09-29
packet_analysis: Introduce PacketAnalyzer::__disable_analyzer()

awelzel:
@rsmmr - mind skimming if this goes into a direction that you think is okay for #2399? It exploits the fact that the manager is holding AnalyzerPtr's for the individual analyzers.

Will hookup Analyzer::disable_analyzer() and add a test or two, but in case something stands out better to know early.


awelzel:
> Yes, that looks good to me. (I keep forgetting that packet analyzers are instantiated only once, so yes, putting the state right in the analyzer makes sense).

I'm marking this ready. I'm uncertain about the `Analyzer::disable_analyzer()` extension: It seems nice to have a single entry point and be able to reuse the `Analyzer::disabled_analyzers` mechanism, but could also see it being separated/duplicated in the `PacketAnalyzer` module over at `scripts/base/packet-protocols/main.zeek` to not mix up the different analyzers.

@rsmmr - mainly looking for opinion/guidance here. Over in the issue you mentioned extending the existing one and the current state is my interpretation of that :-)

> And I think we could then also extend the script-side disable_analyzer to support packet analyzers as well.

rsmmr (approved):
I like it this way. So ready to merge? 

And ... what about file analyzers? Want to extend `disable_analyzer` to that next?

PR #2445, opened 2022-09-30
Management framework: support Broker's WebSocket transport in the controller

ckreibich:
This adds an additional listening port, defaulting to 2149, to the controller, where it now listens via Broker's WebSocket transport for client connections. All of the heavy lifting is in the `zeek-client` update, pulled in via submodule pointer. It also updates the cluster testsuite.


ckreibich:
Yeah, we could. I only added that port for symmetry with how the controller also has its own classic-Broker port in `Management::Controller::default_port`, for a total of:

* Controller, WebSocket Broker: 2149
* Controller, classic Broker: 2150
* Agent, classic Broker: 2151

We can keep the Websocket port at the default if we don't want to keep this adjacency. I did originally consider using the default ports throughout (also for classic Broker), but then realized it complicates support for all-in-one deployments, where agent and controller would then conflict — the agent always needs to listen so the local cluster nodes can peer with it, and the controller still needs a classic-Broker port for agents peering with it (assuming agents connect to it; the default).

rsmmr (approved):

PR #2448, opened 2022-09-30
Bring back extended &group attribute functionality

awelzel:
This is a rough draft/sketch showing the general idea and adding a few tests for demo purposes.

This introduces an EventGroup entity. Event groups may be of different kinds (currently attribute and module).

Semantics: Any event handler part of a disabled group will not be executed.

---

Related branches/PRs:
zeek-docs:
https://github.com/zeek/zeek-docs/tree/topic/awelzel/event-group-attribute
https://github.com/zeek/zeek-docs/pull/152


0xxon:
Same comment as to #2449 applies here - I think it makes sense to also ask for feedback in https://community.zeek.org.

src/Func.cc
```diff
@@ -128,6 +128,22 @@ void Func::AddBody(detail::StmtPtr /* new_body */,
...
+
+void Func::UpdateBodies()
+	{
+	static auto group_disabled = [](const auto& g)
+	{
+		return g->IsDisabled();
+	};
+
+	for ( auto& b : bodies )
+		b.disabled_groups = std::count_if(b.groups.cbegin(), b.groups.cend(), group_disabled);
```
rsmmr, line 144: Matter of taste but I'd just compress this inline:
```suggestion
		b.disabled_groups = std::count_if(b.groups.cbegin(), b.groups.cend(), [](const auto& g) { return g->IsDisabled(); });
```

src/Func.cc
```diff
@@ -817,6 +864,39 @@ static int get_func_priority(const std::vector<AttrPtr>& attrs)
 	return priority;
 	}
 
+// Get a function's groups from its Scope's attributes. Errors if it sees any
+// problems. Less strict than get_func_priority()
```
rsmmr, line 868: What does "less strict" mean here?

src/Func.cc
```diff
@@ -831,6 +911,9 @@ function_ingredients::function_ingredients(ScopePtr scope, StmtPtr body)
 		{
 		priority = get_func_priority(*attrs);
 
+		for ( auto& group : get_func_groups(*attrs) )
+			groups.insert(group);
+
```
rsmmr, line 916: Can this just be `groups = get_func_groups(*attrs)`?

src/Func.cc
```diff
@@ -845,6 +928,10 @@ function_ingredients::function_ingredients(ScopePtr scope, StmtPtr body)
 		priority = 0;
 
 	this->body = std::move(body);
+
+	// Add this function to whatever module group it was defined in.
+	auto module_group = event_registry->RegisterGroup(EventGroupKind::Module, current_module);
```
rsmmr, line 933: I'd avoid accessing the global here, and instead pass the current module as an argument into the constructor. Otherwise, this is easy to use this wrongly by constructing a function ingredient when `current_module` isn't set right.

src/Func.h
```diff
@@ -100,8 +110,15 @@ class Func : public Obj
...
 
 	// Add a new event handler to an existing function (event).
-	virtual void AddBody(detail::StmtPtr new_body, const std::vector<detail::IDPtr>& new_inits,
-	                     size_t new_frame_size, int priority = 0);
+	[[deprecated("Remove in v6.1 - Use function_ingredients version")]] virtual void
+	AddBody(detail::StmtPtr new_body, const std::vector<detail::IDPtr>& new_inits,
+	        size_t new_frame_size, int priority = 0);
+
+	// Add a new event handler to an existing function (event).
+	virtual void AddBody(const zeek::detail::function_ingredients* ingredients);
```
rsmmr, line 118: Use a smart pointer for the argument here to make ownership/memory management clear.

src/Func.h
```diff
@@ -100,8 +110,15 @@ class Func : public Obj
...
+
+	// Update body disabled_count according to groups.
+	void UpdateBodies();
```
rsmmr, line 121: This is just called from the registry, right? Feels a bit confusing to have it part of the normal, public API then, because it's not clear if/when anybody would call it. Couple ideas: make it non-public and the registry a friend, or put the logic to update the counter directly into the registry (and probably likewise make it a friend). 

(I'm not completely sure but if we did the latter and also added a mapping from event group to body to the registry (as I'm suggesting elsewhere), we might not need `groups` set inside the body anymore.)

src/EventRegistry.cc
```diff
@@ -146,4 +147,61 @@ void EventRegistry::ActivateAllHandlers()
...
+// shouldn't take ages either.
+void UpdateFuncBodies()
+	{
+	for ( const auto& handler_name : event_registry->AllHandlers() )
+		{
+		EventHandler* eh = event_registry->Lookup(handler_name);
+		eh->GetFunc()->UpdateBodies();
+		}
+	}
+	}
```
rsmmr, line 187: Agree it wouldn't take ages :-), but it still feels like doing too much here given that we're just changing a single group. I think I'd add the additional bookkeeping to make the registry track which events handlers are part of each group. Then it can address them directly.

rsmmr (commented):
I looked over it and left some notes where I noticed things. 

My main point is where we track/update the state on which handlers are active; I find the flow there a bit hard to follow right now because it divvies it up between the bodies and registry, and the API contract doesn't feel quite clear.  

One other thing: Internally it makes sense to have the different kinds of groups to separate the namespaces. However, I'm wondering about the user-facing script API: The module event groups are just implicitly defined from the user's perspective, and hence I'm not sure that doing `disable_event_group("MyModule", EVENT_GROUP_KIND_MODULE);` is very intuitive. I'm wondering if on the script-side we should split this into different functions: `disable_event_group` would refer only to what defined through `&group`; and then we could add `disable_module()` or maybe `disable_module_events`() to toggle those ones. What do you think?

src/Func.h
awelzel: Thanks. I'm heavily considering removing this part and adding an overload/optional parameter for the additional groups instead. I'm not entirely happy with either one, but leaning towards less happy with the new `function_ingredients` parameter due to changing the additional signature and deprecation (at the cost of not being happy adding yet another parameter...).

awelzel:
I looked over it and left some notes where I noticed things.

> My main point is where we track/update the state on which handlers are active; I find the flow there a bit hard to follow right now because it divvies it up between the bodies and registry, and the API contract doesn't feel quite clear.

I can't say I'm happy with it. As a background: My main issue / constraint was that I did not want to change `std::vector<Func::Body> bodies` in `class Func` to `std::vector<Func::Body*>` (or a smart ptr variant). It is not in `zeek::detail` and not having the pointer indirection might be better for performance (nothing measured though).

That in turn means there are currently no stable pointers to `Func::Body` instances. The bodies vector may be resized/reallocated when adding more bodies - potentially at by plugins doing `AddBody()` at runtime (which may or may not be super valid, but it does seem to work).

That mostly as background: I can attempt hiding/friend the existing UpdateBodies() and possibly track FuncPtr objects in the EventGroup as a small optimization (but it wouldn't get rid of the groups within the Body itself :-( ).

If you think the above assumptions/constraints don't make sense, maybe it's easier to talk it through.


> One other thing: Internally it makes sense to have the different kinds of groups to separate the namespaces. However, I'm wondering about the user-facing script API: [...] What do you think?

I like your proposal of not exposing the kind and instead use different APIs.

src/Func.h
rsmmr: Agree, I actually almost suggested that as well.

src/Func.cc
awelzel: I've updated the comment. get_func_priority() checks for non-applicable attributes for functions as a side-effect and I didn't want to duplicate that.

src/Func.cc
awelzel: I don't know why, but seems clang-format gets very confused around tabs and spaces if I do that. I might stick with the longer version for now.

Screenshot-20221021-182458

src/Func.h
awelzel: I moved the logic to EventRegistry and made it a private friend of Func. Being part of the public API was a mishap.

> (I'm not completely sure but if we did the latter and also added a mapping from event group to body to the registry (as I'm suggesting elsewhere), we might not need groups set inside the body anymore.)

Still as is with the motivation given elsewhere around not being able to have stable pointers to body's.

src/EventRegistry.cc
awelzel: I've kept this for now. We could track FuncPtr's in groups, but that's essentially "just" an optimization (likely an effective one though). But it doesn't change the fact that we still need to keep track of groups within each body.

awelzel:
I've squashed everything down into a single commit, added a NEWS entry. Un-drafting the PR for now.

@rsmmr - this still has the naive "iterate all handlers and their bodies" approach. I'm not sure the FuncPtr tracking is worth it, but feel free to state your opinion here. There's one more TBD around the ScriptFunc but I'm somewhat happier with it all now.

Concrete users would come as a follow-up. Potentially we can tweak things within 5.2 if things don't quite align.

awelzel:
After chatting with @rsmmr , for `disable_module_events()` and `enable_module_events()` `hook`'s should be tracked, too. Will be added here.

awelzel:
> After chatting with @rsmmr , for disable_module_events() and enable_module_events() hooks should be tracked, too. Will be added here.

Think you can look now. The naming "event groups" may be a topic now, but can also argue hooks and events are related enough..

src/zeek.bif
```diff
@@ -5575,3 +5575,88 @@ function is_packet_analyzer%(atype: AllAnalyzers::Tag%) : bool
...
+## group: The group to disable.
+##
+## .. zeek:see:: enable_event_group
+function disable_event_group%(group: string%) : bool
+	%{
+	return zeek::val_mgr->Bool(disable_event_group(zeek::EventGroupKind::Attribute,
+	                                               group->CheckString()));
+	%}
+
+## Enable all event handlers in the given module.
```
rsmmr, line 5637:
```suggestion
## Enable all event handlers and hooks in the given module.
```

src/zeek.bif
```diff
@@ -5575,3 +5575,88 @@ function is_packet_analyzer%(atype: AllAnalyzers::Tag%) : bool
...
+##
+## All event handlers defined in the given module will be enabled if not
+## disable through an event group.
```
rsmmr, line 5640:
```suggestion
## All event handlers and hooks defined in the given module will be enabled if not
## disabled otherwise through an event group.
```

src/zeek.bif
```diff
@@ -5575,3 +5575,88 @@ function is_packet_analyzer%(atype: AllAnalyzers::Tag%) : bool
...
+## module_name: The module to enable.
+##
+## .. zeek:see:: disable_module_events enable_event_group disable_event_group
+function enable_module_events%(module_name: string%) : bool
+	%{
+	return zeek::val_mgr->Bool(enable_event_group(zeek::EventGroupKind::Module,
+	                                              module_name->CheckString()));
+	%}
+
+## Disable all event handlers in the given module.
```
rsmmr, line 5651:
```suggestion
## Disable all event handlers and hooks in the given module.
```

src/zeek.bif
```diff
@@ -5575,3 +5575,88 @@ function is_packet_analyzer%(atype: AllAnalyzers::Tag%) : bool
...
+##
+## All event handlers defined in the given module will be disabled.
```
rsmmr, line 5653:
```suggestion
## All event handlers and hooks defined in the given module will be disabled.
```

NEWS
```diff
@@ -39,6 +39,26 @@ New Functionality
   been extended to work for packet and file analyzers. This now allows to
   leverage ``Analyzer::disabled_analyzers`` for these kinds of analyzers.
 
+- Re-introduce event groups. Allow the &group attribute on event and hook
+  handlers for annotating them with one or more event groups. These groups
+  can be disabled and enable during runtime. Disabling an event group implies
+  disabling all event and hook handlers that are part of it.
+
+  The main difference to the previous implementation is its granularity: It
```
rsmmr, line 47:
```suggestion
  The main difference to a previous implementation in (very) old Zeek versions is its granularity: It
```

src/Func.h
```diff
@@ -70,10 +73,15 @@ class Func : public Obj
 		{
 		detail::StmtPtr stmts;
 		int priority;
+		std::set<EventGroupPtr> groups;
+		int disabled_groups = 0;
```
rsmmr, line 77: We don't really need the count, do we? Then how about just doing `bool disabled = false` here now?

rsmmr: And I'd bring back a comment that this is maintained through the event groups mgmt.

rsmmr (commented):
Ok, I like this, seems like the best possible implementation given the constraints.

 Final set of comments, I'll merge once you update.

awelzel:
> Final set of comments, I'll merge once you update.

Sorry for the delay - should now be good.

There's the following zeek-docs branch that as a minimum makes the :zeek:attr:`&group` available and as a very small section:
https://github.com/zeek/zeek-docs/tree/topic/awelzel/event-group-attribute

I'll open a PR, but it'll fail because it depends on `enable/disable_event_group()` - I hope you can straighten this out. Let me know if that could be done better.

PR #2449, opened 2022-09-30
Introduce special treatment for the blank identifier _

awelzel:
Mostly: Do not instantiate variables within for loops and allow reusing differently typed blanks which previously wasn't possible.

This may be missing some corner-cases, but the added tests seem to work as expected and nothing else fell apart it seems.


src/Stmt.cc
```diff
@@ -1275,9 +1284,14 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr)
...
 			}
 
-		const auto& t = (*loop_vars)[0]->GetType();
+		const auto& lv = (*loop_vars)[0];
+		const auto& t = lv->GetType();
 
-		if ( ! t )
+		if ( lv->IsBlank() )
+			{
+			lv->Error("can not use blank identifier iterating over string");
```
bbannier, line 1292: Just an observation, the following file somehow leads to wrong diagnostic locations:

```console
$ cat foo.zeek
event zeek_init()
	{
	local _ = "";
	for ( _ in "foo")
		;
	}
$ zeek --parse-only foo.zeek
error in ./foo.zeek, line 3: blank identifier used as local (_)
error in ./foo.zeek, line 3: can not use blank identifier iterating over string (_)
```

I would have expected diagnostics from 2 different lines.

testing/btest/language/for-blank.zeek
```diff
@@ -0,0 +1,29 @@
...
+		print v;
+	print "canaries", a, b;
+
+	print "== t1";
+	for ( _, v in t1 )
+		print v;
+	print "canaries", a, b;
+
+	print "== t2";
+	for ( [_,c,_], v in t2 )
```
bbannier, line 25: It would be nice if I could add a placeholder for the key without destructuring it, but this currently produces an error:

```console
$ cat foo.zeek
event zeek_init()
	{
	local t2 = table(["a", 1, T] = "a1a", ["b", 2, F] = "b2b", ["c", 3, T] = "c3c");
	for ( _, v in t2 )
		; 
	}
$ zeek foo.zeek
error in ./foo.zeek, line 4: wrong index size (t2)
```

testing/btest/language/for-blank.zeek
```diff
@@ -0,0 +1,29 @@
...
+	local a = "a canary";
+	local b = "b canary";
+
+	local vec = vector("a", "b", "c");
+	local t1 = table(["keya"] = "a", ["keyb"] = "b", ["keyc"] = "c");
+	local t2 = table(["a",1,T] = "a1a", ["b",2,F] = "b2b", ["c",3,T] = "c3c");
+
+	print "== vec";
+	for ( _, v in vec )
+		print v;
```
bbannier, line 16: Might be worth capturing that using `_` both for index and value works,

```zeek
for ( _, _ in vec )
...
```

testing/btest/language/local-blank.zeek
```diff
@@ -0,0 +1,20 @@
...
+event zeek_init()
+	{
+	local _:string = "1";
+	}
+
+#@TEST-START-NEXT
+event zeek_init()
+	{
+	local _:count= "1";
+	}
```
bbannier, line 20: One can currently declare `_` constants (which I think doesn't hurt). Could you add a test for that as well?

```zeek
function foo(): count
	{
	print "foo!";
	return 10;
	}

const _ = foo();
```

Such constants probably shouldn't show up in zeekygen output.

src/parse.y
```diff
@@ -1870,13 +1870,21 @@ stmt:
 			set_location(@1, @7);
 			if ( ! locals_at_this_scope.empty() )
 			       locals_at_this_scope.back().insert($2);
-			$$ = build_local($2, $3, $4, $5, $6, VAR_REGULAR, ! $8).release();
+
+			if ( $2->IsBlank() )
+				$2->Error("blank identifier used as local");
```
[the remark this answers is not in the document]
awelzel: I've pushed a version and test that should allow for this now.

awelzel:
I've pushed a few updates to address some of Benjamin's feedback (thanks!) and a few more test.

I still like the idea of `_` being somewhat special (particularly when it allows small optimizations). The current approach does feel a bit ad-hoc (I might be missing some fundamentals), but the tests show how `_` could be leveraged specifically around the `for` loop.

src/Stmt.cc
```diff
@@ -1233,7 +1243,11 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr)
 			const auto& lv = (*loop_vars)[i];
 			const auto& lvt = lv->GetType();
 
-			if ( lvt )
+			if ( lv->IsBlank() )
+				{
+				// nop
+				}
```
rsmmr, line 1249: Just use `continue`.

src/Stmt.cc
```diff
@@ -1312,7 +1336,11 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr, IDPtr val_var)
 		}
 
 	// Verify value_vars type if it's already been defined
-	if ( value_var->GetType() )
+	if ( value_var->IsBlank() )
+		{
+		value_var = ID::nil;
+		}
```
rsmmr, line 1342: Braces not needed.

rsmmr (approved):
This looks good to me. Do we want to discuss this more still, or is it ready to merge?

testing/btest/language/local-blank.zeek
awelzel: I've added a test for using `const`. No zeekygen extras, however.

As long as there's no case for `_` suddenly showing in the docs (you couldn't reference and there's no real reason to use one) I'd postpone.

Issue #2446, opened 2022-09-30
Zeek variadic function crash


awelzel:
Ran into this while looking at #2425 - has been around forever and probably somewhere in the "don't do this" corner.

It segfauls with bro 2.5.5 on  try.zeek.org as well as on latest master.

```
$ cat any-crash.zeek                                                                                                                                                                                                                                                                                             
function f(x: any)                                                                                                                                                                                                                                                                                                                                          
        {                                                                                                                                                                     
        print x;                                                                                                                                                              
        }                                                                                                                                                                     
                                                                                       
event zeek_init()                                                                                                                                                             
        {                                                                              
        f(1);                                                                                                                                                                                                                                                                                                                                               
        f(1, 2);                                                                       
        f(1, 2, 3);                                                                                                                                                           
        }

$ zeek any-crash.zeek                                                                                                                                                                                                                                                                                            
1                                                                                                                                                                             
=================================================================                                                                                                             
==1170275==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000a51a60 at pc 0x561f7f85f749 bp 0x7ffc2bbe6a60 sp 0x7ffc2bbe6a58                                                                                                                                                                                                                 
READ of size 8 at 0x602000a51a60 thread T0                                                                                                                                    
    #0 0x561f7f85f748 in zeek::IntrusivePtr<zeek::Val>::get() const ../src/zeek/zeek/zeek/IntrusivePtr.h:155                                                                  
    #1 0x561f7f85f748 in decltype ((({parm#1}.get)())!=(({parm#2}.get)())) zeek::operator!=<zeek::Val, zeek::Val>(zeek::IntrusivePtr<zeek::Val> const&, zeek::IntrusivePtr<zeek::Val> const&) ../src/zeek/zeek/zeek/IntrusivePtr.h:284
    #2 0x561f7f85f748 in zeek::detail::ScriptFunc::Invoke(std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >*, zeek::detail::Frame*) const ../src/Func.cc:394                                                                                                                                                      
    #3 0x561f7f7657da in zeek::detail::CallExpr::Eval(zeek::detail::Frame*) const ../src/Expr.cc:4711                                                                         
[...]

Issue #2425, opened 2022-09-21
Unhelpful error when `sep` or `def` are not strings with `cat_sep`


ynadji:
```
$ zeek --version
zeek version 5.1.0-dev.497

$ zeek -e 'print cat_sep(";", set(0.0.0.0/24), "foo");'
fatal error in <no location>: Val::CONVERTER (table/string) ({0.0.0.0/24})
zsh: abort      zeek -e 'print cat_sep(";", set(0.0.0.0/24), "foo");'

$ zeek -e 'print cat_sep(set(0.0.0.0/24), ";", "foo");'
fatal error in <no location>: Val::CONVERTER (table/string) ({0.0.0.0/24})
zsh: abort      zeek -e 'print cat_sep(set(0.0.0.0/24), ";", "foo");'
```

I would've expected something more helpful, like a type error as we see with:

```
$ zeek -e 'print bytestring_to_count(set(0.0.0.0/24));'
error in string and <command line>, line 1: type clash (string and set(0.0.0.0/24))
error in <command line>, line 1 and string: type mismatch (set(0.0.0.0/24) and string)
error in <command line>, line 1: argument type mismatch in function call (bytestring_to_count(set(0.0.0.0/24), F))
```

awelzel:
This appears more complicated than I thought. Type checking `cat_sep` is not done as it ends-up as follows in zeek.bif.zeek:
```
global cat_sep: function(va_args: any): string ;
```
Adding runtime type checking into the `cat_sep` BIF implementation by hand is "too" late as the error is triggered in the "preamble" generated from `bifcl`.

```
...
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff6cdd537 in __GI_abort () at abort.c:79
#2  0x000055555901c3e1 in zeek::Reporter::FatalErrorWithCore (this=this@entry=0x614000000040, fmt=fmt@entry=0x55555781d0a0 "%s") at ../src/Reporter.cc:152
#3  0x0000555559049639 in zeek::Obj::BadTag (this=this@entry=0x604000002710, msg=msg@entry=0x555557830940 "Val::CONVERTER", t1=0x55555782faa0 "count", 
    t2=t2@entry=0x55555782fba0 "string") at ../src/Obj.cc:108
#4  0x000055555919d617 in zeek::Val::AsStringVal (this=0x604000002710) at ../src/Val.cc:80
[...]

PR #2450, opened 2022-10-02
Add variadic params support

AmazingPP:
Inspired by #2446 #2425, I feel that we need to provide language-level variadic params features for Zeek scripts. So I implemented it.

Unfinished tasks:
- [x] Implement the Zeek scripts  to C++ optimization part.
- [ ] Updating the `bifcl`.
- [ ] Optional: Maybe we need a syntactic sugar to make variadic paramsmore intuitive and usable, like this: `function cat(...va_args: vector of any);`


AmazingPP:
Hmm, updating bifs takes more work than I thought, so I decided to open a new PR to do it.

ckreibich:
Fupeng, thanks! For this level of contribution it's helpful to sync up with us before coding. We'd normally start a thread over in Discussions, hash out the details, and then move on to PRs. It's not clear here that we really need variadic parameters in Zeek scripts (`bifcl` is another matter), and if there's agreement that we do, then we can figure out technical details and think about when to work on this. I'm going to close this, but we can certainly revisit this.

PR #2453, opened 2022-10-04
Bump spicy-plugin to v1.3.19.

bbannier:


bbannier:
@timwoj, the bump contains a minor code update where we could trigger an runtime error with debug logging enabled. I think the bump could be pulled into the next path release w/o much risk (we can get rid of the small cleanup patches for that if you want, but I think they shouldn't be risky either).

PR #2456, opened 2022-10-05
Topic/jasonlu/fix dict clone performance

jasonlue:
Fix Dict clone performance issue when Dict is 1M+. Please refer https://github.com/zeek/zeek/issues/2387 for more details
- Stores Fibonacci hash directly in the dictionary entry so that SizeUp() doesn't have to recalculate the final hash.
- Detects performance deterioration by monitoring average distance to find the next space to insert the new item. When the space is hard to find, it means the load factor is just too high at the moment. SizeUp the dictionary (double the size) when this happens.
- Fine tunes the threshold for performance deterioration:  The condition is:
       (1) The current load factor is high ( >  25%)
       (2) Average distance to find a space for insertion is long (> 32)
       (3) Enough samples are collected: (> 128 times.)
- Adds a unit test to detect this performance deterioration for the future. run ./zeek --test -ts="dict verify clone performance" and verify it passes. This unit tests compare times spent between initial insertion and cloning for 3 rounds for 1M 1.5M and 2M dictionary sizes. The test fails if the average time spent on cloning is 3 times greater than initial insertion.
- Adds another unit to measure time spent in detail for different situations. this unit test is defined out and can be used when necessary in the future.
       

@timwoj @JustinAzoff @awelzel @pbcullen


src/Dict.h
```diff
@@ -1162,11 +1189,13 @@ template <typename T> class Dictionary
 	void InsertRelocateAndAdjust(detail::DictEntry<T>& entry, int insert_position)
 		{
 /// e.distance is adjusted to be the one at insert_position.
-#ifdef ZEEK_DICT_DEBUG
+#ifdef DEBUG
```
timwoj, line 1192: Is this change intentional? All of the other uses of ZEEK_DICT_DEBUG stayed the same.

src/Dict.h
```diff
@@ -1504,6 +1533,9 @@ template <typename T> class Dictionary
 	uint32_t num_entries = 0;
 	uint32_t max_entries = 0;
 	uint64_t cum_entries = 0;
+	int64_t space_distance_samples = 0;
+	// how far the space is
```
timwoj, line 1537: Is this is the space between two elements in the table? I'm not entirely sure what you mean by "space" in all of the changes.

jasonlue: When a new item is to be inserted, it is inserted as the last element in its cluster (elements of the same bucket). However, if that spot is not empty (not a space), it has to relocate the existing element that occupies the spot before inserting itself.  Now the new element needs to find its new location. And the new location is the end of its own cluster. The chain effect continues until a space (empty spot) is found so that no new element needs to be relocated.

The space distance is the distance between this empty spot and the spot of the externally inserted element. 

This chain effect, (like recursion) causes a linear search until the next space is found in the table from bucket of the element.  When space is hard to find, it takes long time.

The reason the space if faraway is the load factor is just too large for the current hash distribution.

timwoj: Ah, so it's basically a factor of how long it takes for the table to rebalance after inserting an element that would initially be considered a collision?

src/Dict.h
timwoj: Noted, thanks.

src/Dict.cc
```diff
@@ -12,6 +12,196 @@ namespace zeek
...
 TEST_SUITE_BEGIN("Dict");
 
+//to show outputs of unit tests, define PRINTF as fprintf by uncommenting fprintf.
+//and run ./zeek --test -tc="dict verify clone performance"
+#define PRINTF(fmt, args...) //fprintf(stderr, fmt, args)
+
+struct ReportDuration
+	{
+	std::chrono::_V2::system_clock::time_point start;
+	std::chrono::_V2::system_clock::time_point stop;
```
timwoj, line 22: These two lines are causing build failures on a number of the CI hosts. On a local build, changing it to `std::chrono::high_resolution_clock::time_point` appears to fix it.

timwoj (changes requested):

src/Dict.h
jasonlue: That's correct.

src/Dict.cc
jasonlue: changed accordingly

src/Dict.cc
```diff
@@ -18,8 +18,8 @@ TEST_SUITE_BEGIN("Dict");
 
 struct ReportDuration
 	{
-	std::chrono::_V2::system_clock::time_point start;
-	std::chrono::_V2::system_clock::time_point stop;
+	std::chrono::system_clock::time_point start;
```
timwoj, line 21: This still isn't it unfortunately. macOS complains as below.

```
[ 93%] Building CXX object src/CMakeFiles/zeek_objs.dir/EventTrace.cc.o
../../src/Dict.cc:26:9: error: no viable overloaded '='
                start = std::chrono::high_resolution_clock::now();
                ~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/__chrono/time_point.h:31:28: note: candidate function (the implicit copy assignment operator) not viable: no known conversion from 'time_point<std::chrono::steady_clock, duration<[...], ratio<[...], 1000000000>>>' to 'const time_point<std::chrono::system_clock, duration<[...], ratio<[...], 1000000>>>' for 1st argument
class _LIBCPP_TEMPLATE_VIS time_point
                           ^
```

src/Dict.cc
```diff
@@ -124,10 +124,10 @@ TEST_CASE("dict verify clone performance")
 #if 0 //open this test to measure performance in detail
 TEST_CASE("dict measure clone")
 	{
-	constexpr int iMax = 10;
+	constexpr int iMax = ;
```
timwoj, line 127: Was this an intentional change? I know it's in an #ifdef so it doesn't build.

timwoj (changes requested):
I'm trying to build and test on macOS to try the script from #2387 locally. I figured I'd run the tests and such while I was at it. My build is configured with `--sanitizers=address`. While your new unit test passes (after about 30 seconds), I'm seeing some other tests fail including one failing with a `heap-use-after-free`. A bunch of the btests fail as well but I didn't investigate why.

src/Dict.cc
jasonlue: Hi, Tim,

Let me back track the PR to just include the Dict.h modifications and exclude the unit tests. This make it easy and we need to verify it.

I'll send another PR for unit test

jasonlue:
@timwoj How does this sound?

timwoj:
Overridden by #2478

PR #2461, opened 2022-10-06
Update CT log list.

0xxon:
This uses the v3 json as a source for the first time. The test needed some updating because Google removed a couple more logs - in the future this should hopefully not be neccessary anymore because I think v3 should retain all logs.


0xxon:
Big diff because the URL scheme changed slightly.

timwoj (approved):

timwoj:
Regarding pulling this into 5.1, should the zeek-aux changes go in as well?

0xxon:
I honestly don't think it matters. And - if it is extra work, we can also skip this for 5.1 - I don't think it is important enough.

PR #2464, opened 2022-10-07
Move API for disabling packet analyzers into component.

rsmmr:
Move API for disabling packet analyzers into component.

Before, that API was part of the analyzers themselves, which meant we couldn't disable a packet analyzer before it had been instantiated.  That's different from protocol/file analyzers, where we disable them through the corresponding component. The lack of the component-side API prevented Spicy from replacing packet analyzers at startup.

The reason we had put this into analyzer originally was performance so that we don't need a component lookup for every packet. This change keeps that optimization by caching the on/off state in the analyzer itself as well, but now with the component being the one controlling it.

We also add the enabled/disabled state for packet analyzers to the output of `-NN`.

Additionally, this sneaks in a little warning change, too.


rsmmr:
@awelzel, I needed this for Spicy, what do you think?

src/packet_analysis/Manager.cc
```diff
@@ -30,7 +30,10 @@ void Manager::InitPostScript(const std::string& unprocessed_output_file)
 	for ( const auto& analyzerComponent : GetComponents() )
 		{
 		if ( AnalyzerPtr newAnalyzer = InstantiateAnalyzer(analyzerComponent->Tag()) )
+			{
+			newAnalyzer->SetEnabled(analyzerComponent->Enabled());
```
awelzel, line 34: Only thought was if that could be moved into `InstantiateAnalyzer()` as the contract seems that one currently is required to explicitly call `SetEnabled()` on the result.

After looking, that method is private for the packet_analysis/Manager, so that seems mood currently.

awelzel (approved):
Thanks for cleaning and aligning this, makes sense to me.

src/packet_analysis/Manager.cc
rsmmr: Yeah, what's why I moved that out of public.

PR #2469, opened 2022-10-10
Surface a better CMake error if the user passes an invalid sanitizer name

timwoj:
This PR adds a little bit of CMake to double-check that the names of the sanitizers passed to configure are valid before moving on with the configuration. Previously, you would eventually get an error while trying to configure one of the dependency libraries and have to dig in from there. Instead now you get something like:

```
% ./configure --sanitizers=bad
...
-- Performing Test COMPILER_SUPPORTS_SANITIZERS
-- Performing Test COMPILER_SUPPORTS_SANITIZERS - Failed
CMake Error at CMakeLists.txt:246 (message):
  Invalid sanitizer compiler flags: -fsanitize=bad
```


CMakeLists.txt
```diff
@@ -235,6 +235,18 @@ if ( ZEEK_SANITIZERS )
 
     set(_sanitizer_flags "-fsanitize=${ZEEK_SANITIZERS}")
 
+    # The linker command used by check_cxx_compiler_flag requires you to also pass the sanitizer to
+    # it or it fails. The best way to do this is to set CMAKE_REQUIRED_LINK_OPTIONS, but save off a
+    # copy of it so it can be reset back to what it was previously afterwards.
+    set(_temp_link_options ${CMAKE_REQUIRED_LINK_OPTIONS})
+    set(CMAKE_REQUIRED_LINK_OPTIONS ${_sanitizer_flags})
```
rsmmr, line 242: I'm wondering if we should append this to the current value, in case that comes with other options needed?

rsmmr (approved):

CMakeLists.txt
timwoj: I'm trying to think whether that variable would be set at all before CMake gets to this point, and I don't think it would be, but you're probably right that it's best to assume it would.

PR #2484, opened 2022-10-14
-O gen-C++ October maintenance

vpax:
This PR provides updates to the compile-to-C++ script compiler: the main things covered are (1) fixes for including attributes correctly when initializing globals, and (2) only generating code to initialize globals when compiling for "standalone" execution.


timwoj (approved):

testing/btest/Baseline.cpp/coverage.basic-no-errors/.stderr
```diff
@@ -0,0 +1,2 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+<...>/pp-alarms.zeek, line 100: Hello, Zeek!
```
awelzel, line 2: That location actually looks a bit off (same in the corresponding reporter.log). Not that it's super important probably, but seemed surprising.

vpax: > That filename actually looks a bit off (same in the corresponding reporter.log). Not that it's super important probably, but seemed surprising.

(Thanks for the careful look-over!) Yeah, I thought about just skipping that test, but figured it's better to test the functionality.  The wonky location associated with the report is because compiled code (deliberately) doesn't try to maintain detailed location information.

PR #2489, opened 2022-10-19
Add reverse SSH setup event & logging

fox-ds:
A follow up after our talk last week at ZeekWeek2022, where we showed the example of QUIETEXIT; an implant that will act like an SSH server once it has set up a connection to the attacker. Zeek currently determines the client & server side of an SSH connection based on the TCP handshake. This is mostly true, however as showed in the included PCAP, in some situations Zeek can't correctly determine the client & server roles. By adding an extra event, and optionally logging whether the key exchange was set up by the server, users may be able to check their network for (anomalies in) reverse SSH connections. On purpose, it does not switch the `client` and `server` log fields; it's up to the user to check for this reversed setup and possibly reverse these, to not change behavior with previous Zeek versions.

Any feedback is appreciated!

Joost


src/analyzer/protocol/ssh/ssh-analyzer.pac
```diff
@@ -120,6 +120,16 @@ refine flow SSH_Flow += {
 		return true;
 		%}
 
+	function proc_ssh2_keys_setup_direction(is_orig: bool): bool
```
awelzel, line 123: Thanks for the PR!

(I hope someone else from the team with more experience chimes in) I see the convenience of the generic "setup direction" event, but do wonder if introducing 4 new (low-level) events for the individual `_INIT` and `_SECRET` messages would be the more usual approach and provide interesting additional information for other scenarios (and also make it easier to debug/localize in case things don't quite work as expected).

Minimally they all could provide `c: connection` and `is_orig: bool` and possibly additional parameters if available in the message (seeing e, output_token, encrypted_payload). But adding more parameter is always possible without breaking backwards compat.

It certainly does add more boilerplate in the .pac files and also in scripts they'd all need to be handled rather than having one specific event everything is funneled through though.

scripts/base/protocols/ssh/main.zeek
```diff
@@ -54,6 +54,9 @@ export {
 		host_key_alg:    string       &log &optional;
 		## The server's key fingerprint
 		host_key:        string       &log &optional;
+		## Indicate whether there is a discrepancy between the host that
+		## set up the TCP connection, and sets up the SSH key exchange.
+		reversed_conn:   bool         &log &optional;
```
awelzel, line 59: Somewhat delayed, but the other question that comes up with this change is whether this field is something we want to add to ssh.log by default.

It seems for most deployments this field will always be blank. Once it's `T`, it's of course interesting and very valuable, but generally it's entropy is extremely low.

To me it sounds more appropriate to make this a notice (or weird with action notice). This may also match better with existing workflows monitoring notices/weirds, rather than having "features" in the logs.

Adding it to ssh.log can of course be done by additional scripts by extending the log, it's just that the default scripts wouldn't include this field.

Other thoughts?

vpax: > To me it sounds more appropriate to make this a notice (or weird with action notice). This may also match better with existing workflows monitoring notices/weirds, rather than having "features" in the logs.

My thoughts too.

fox-ds: Thanks for the feedback! I was also thinking whether the correct place for this would be in the `base` scripts, or the `policy` ones, didn't think of a notice/weird log yet as we use it in the `ssh.log` internally. The current commit does not sound like the best solution indeed, so I think that leaves us with a few possible solutions:

1. Change the `base/protocols/ssh/main.zeek` script and let it create a `notice.log` entry
2. Create a `policy/protocols/ssh/detect-reverse-ssh.zeek` script and let it create an optional feature in the `ssl.log`
3. Create a `policy/protocols/ssh/detect-reverse-ssh.zeek` script and let it create a feature in the `notice.log`

My personal suggestion is to do both 1 and 2; enable the detection in the notice log by default, but also optionally write the field to an `ssl.log`. What do you think?

awelzel: > Create a policy/protocols/ssh/detect-reverse-ssh.zeek script and let it create a feature in the notice.log
> My personal suggestion is to do both 1 and 2; enable the detection in the notice log by default, but also optionally write the field to an ssl.log. What do you think?

I'd actually steer towards just doing 3) and loading the new script by default via `site/local.zeek`. There aren't notices raised from `base/protocols/*` currently so that would be new about 1). It's unclear if 2) really needs to be in the base distribution.

There's certainly also the argument that the notice could be added through an external `zkg` "detection package". However, given how closely it's connected with the new event(s) and pretty self-contained, adding it directly to `policy` seems reasonable to me for this one.

fox-ds: Thanks again! Going to have a look at it today

src/analyzer/protocol/ssh/ssh-analyzer.pac
ckreibich: I like this suggestion as well — it shifts the eventing more toward mechanism, gaining flexibility down the road.

scripts/base/protocols/ssh/main.zeek
ckreibich: Just to add a thought that's implicit in the above: we could do none of these here, and instead add the field in a Zeek package for the folks eager to have it. We're reluctant to add anything to the policy folder at this point (says the guy who dumped the whole management framework in there — forgive the irony!).

vpax: I like @ckreibich 's suggestion.

fox-ds: Yes, then I'll move the script to a separate package.

awelzel: Binpac/analyzer changes would stay in this repo. It'd also be great to keep the pcap and the test. The latter would needs some changes - maybe handling the new events directly and just printing.

fox-ds: Cool. Closing this PR for now and going to submit a new PR with the feedback applied. Thanks for all the feedback!

PR #2496, opened 2022-10-21
enum_names: Support naming types with a string

awelzel:
In his ZeekWeek 2022 presentation, @stevesmoot mentioned that he had a difficult time looking up enum names when all he had was a string naming the type.

Add support to enum_names() to transparently lookup the type if a string is provided. This is similar in how record_fields() behaves when being passed a string.


rsmmr (approved):

vpax:
I thought Smoot wanted something different: a BiF that given an enum type would return all of the elements of that enum.

awelzel:
I've pinged Smoot out-of-band - he'll chime in when he finds the time. I *think* this was one of the things, but there were a few more that he has queued.

PR #2497, opened 2022-10-21
Switch to bundled CAF for Coverity build

ckreibich:
zeek/broker#285 prevents use of an external CAF without also using an external incubator, which isn't currently tweakable at the `configure` level. This fails the Coverity build: https://github.com/zeek/zeek/actions/runs/3293864275/jobs/5442294447

The reason for using an external CAF are historic and may no longer hold, so this is a check to see whether we can now use a normal build.

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -42,11 +42,8 @@ jobs:
             curl \
             wget
 
-      - name: Install CAF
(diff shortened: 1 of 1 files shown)


ckreibich:
This all looks fine to me: https://github.com/zeek/zeek/actions/runs/3300453933/jobs/5444968655

Coverity has also processed the build correctly. So I think this is good to go.

timwoj (approved):

PR #2504, opened 2022-10-26
parse.y: Traverse AST post parsing to detect break/next usage issues

### LINKED ISSUE #2440: "next" and "break" lack context-checking
The following program compiles without complaint:

> event zeek_init()
> {
> next;
> break;
> }

It should generate an error during script compilation flagging that those don't make sense in the current context.

awelzel:
Seemed easiest to do it via the traversal infrastructure as we do not otherwise track enough context/scope when instantiating break or next statements.

Might be worth moving this out of src/parse.y, but didn't exactly know where. Or maybe we wait until there's more such trivial validations popping up

Fixes #2440


awelzel:
> Seemed easiest to do it via the traversal infrastructure as we do not otherwise track enough context/scope when instantiating break or next statements.

An alternative would be to track the same state during parsing instead of relying on the `Traverse()` functionality. I can see that being preferred over what I did in the current version..

rsmmr (commented):
I like using traversal, it keeps the validation localized inside a small, self-contained piece of code.

I'm wondering if the code could be a compressed a bit. Do we need to track for/while and switch separately? The code for each is very similar, would just a single stack work? Also, do we need a stack at all, or could we just use a counter that we increment each time we get into a loop/switch, and decrement on exit?

I'd prefer to move this somewhere else than `parse.y` but you'r right that we don't have a good place right now. Ideally, we'd have a separate `Validator` that would collect such tests, but yeah, probably best done later once we have more checks.

awelzel:
> I'm wondering if the code could be a compressed a bit. Do we need to track for/while and switch separately? The code for each is very similar, would just a single stack work? Also, do we need a stack at all, or could we just use a counter that we increment each time we get into a loop/switch, and decrement on exit?

I separated them out as one can't use `next` in `switch`. You're obviously right, a counter is enough given we only do `stack.size()`.

I can do 3 separate counters and introduce helpers for `next` / `break` - that may make it less confusing than having the for/while one combined.

awelzel:
@rsmmr - could you take another look? I've moved the class into `src/ScriptValidation.h` and also now just track all statement depths in a map. It didn't compress overly well, but maybe you have further ideas.

src/ScriptValidation.h
```diff
@@ -0,0 +1,77 @@
...
+namespace zeek::detail
+	{
+
+// Track statements and hooks and validate break and next usage.
+class BreakNextScriptValidation : public TraversalCallback
+	{
+public:
+	TraversalCode PreStmt(const Stmt* stmt)
+		{
+		stmt_depths[stmt->Tag()] += 1;
```
rsmmr, line 16: Now this is tracking all statements, which semantically feels a bit odd. How about on more helper: `IsRelevantStatement()` (or such) testing whether it's a any of our stmts? And then just early-exit in all the methods if not relevant.

src/ScriptValidation.h
```diff
@@ -0,0 +1,77 @@
...
+
+#pragma once
+
+#include "zeek/Traverse.h"
+
+namespace zeek::detail
+	{
+
+// Track statements and hooks and validate break and next usage.
+class BreakNextScriptValidation : public TraversalCallback
```
rsmmr, line 11: I don't think this needs to be in a header file. If we add more validators, that would get a bit awkward. I suggest to move all the code into `ScriptValidation.cc`, and move the function `script_validation()` there as well. Then just add a prototype for `script_validation()` to `ScriptValidation.h` (or maybe some existing header somewhere if a new one feels too much; or even just into `parse.y`).

rsmmr (commented):
Nice, quite a bit more readable. Just a couple more small comments.

rsmmr (approved):

PR #2506, opened 2022-10-26
Bump spicy-plugin to v1.3.21.

bbannier:


bbannier:
@timwoj, this bump needs to be backported to both `release/5.0` and `release/5.1`.

timwoj:
> EDIT: this should also go into release/5.2

It will, in a few months when we do 5.2.

PR #2507, opened 2022-10-26
scripts/analyzer: Introduce requested_analyzers

awelzel:
In certain deployment scenarios, all or some analyzers are disabled by default. However, conditionally/optionally loaded scripts may require them.

This change adds a set in the Analyzer module where where external scripts can record the requirement/request for a certain analyzer to subsequently act on these.


awelzel:
@sethhall @sowmyaramapatruni - I took a first stab around the required/requested analyzer topic.

IIUC, the main driver to have this in Zeek base is for external scripts that use this sort of mechanism to not rely/require a custom implementation and continue to work easily with a vanilla open-source distribution whenever released to the public.

Some points that come up:
1) Does this work for you?
2) Naming suggestions?
3) Would the `redef` be sufficient, or is a dynamic `Analyzer::request_analyzer()` still useful?

Could you maybe outline why using `Analyzer::enable_analyzer()` within `zeek_init()` in those external scripts wouldn't be enough? The declarative approach is nice, but currently lost why the `enable_analyzer()` wasn't enough.

awelzel:
@sowmyaramapatruni / @sethhall - did you have a chance to take a look?

sowmyaramapatruni:
Hey Arne,

Sorry for the delayed response. Thankyou for writing this. I have couple of questions.

1. Does this work only for enabling analyzers? why not for disabling. I see an option "redef Analyzer::disable_all = T;" for disabling all analyzers, but not for individual analyzers.
2. Will the function work for Spicy analyzers too?
3. Naming looks good and for your last question on redef, I am not sure if redef works for sensor-core. @sethhall could answer better.

awelzel:
> 
> 1. Does this work only for enabling analyzers? why not for disabling. I see an option "redef Analyzer::disable_all = T;" for disabling all analyzers, but not for individual analyzers.

For disabling analyzer there already exists a set `Analyzer::disabled_analyzers` which can be used to disable individual analyzers. These are disabled at `zeek_init()` time, something that for `requested_analyzers` is conditional.

Could certainly see calling it `Analyzer::enabled_analyzers` as well, but because the request was to have a place to stash analyzer tags and have external scripts enable/disable scripts, the two options seemed more reasonable.

> 2. Will the function work for Spicy analyzers too?

Yes.

> 3. Naming looks good and for your last question on redef, I am not sure if redef works for sensor-core. @sethhall could answer better.

It should, but happy to hear also Seth's thoughts.

NEWS
```diff
@@ -144,6 +144,10 @@ New Functionality
   ``ssh2_gh_gex_init``, ``ssh2_gss_init``, ssh2_rsa_secret``) to detect
   when SSH client and server roles are reversed.
 
+- Analyzer tags added to ``Analyzer::requested_analyzers`` will be enabled
+  at ``zeek_init()`` time. This is primarily useful in combination with
+  ``Analyzer::disable_all`` and external scripts requiring analyzers.
```
ckreibich, line 149: Might be worth stating here that it's a no-op when `disable_all` remains at its default `F`. I could also see more complex interactions if other code uses `Analyzer::disable_analyzer()` at startup time, but that's likely better covered in the docs — there's still zeek/zeek-docs#151.

scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -149,6 +149,24 @@ export {
...
+	## This set can be added to via ``redef`` or calling :zeek:see:`Analyzer::request_analyzers`.
+	global requested_analyzers: set[AllAnalyzers::Tag] = {} &redef;
+
+	## Request the given analyzer tag to be enabled.
+	##
+	## Calling this method is only valid in :zeek:see:`zeek_init` handlers
+	## with a priority greater than -5.
+	##
+	## tags: The analyzer tags to request.
+	global request_analyzers: function(tags: set[AllAnalyzers::Tag]);
```
ckreibich, line 169: Arne should we leave out this function? If requesting analyzers can only happen at startup, redef'ing `requested_analyzers` is actually the better mechanism, no?

awelzel: I'll drop in a fixup. The implementation that is being copied uses a function though. I pinged @sethhall for some input and then we squash or ditch the fixup.

awelzel: I just removed it without the extra commit. If there's follow-up feedback we can at the function back.

ckreibich (approved):
Thanks Arne!

PR #2511, opened 2022-10-27
egrep reported as obsolete by opensuse-tumbleweed builds

timwoj:
One of the CI builds from the Microsoft work started reporting the following error:

```
  egrep: warning: egrep is obsolescent; using grep -E
```

This only happens on the opensuse-tumbleweed build, but the `-E` flag for grep should exist on every supported platform at this point. There's some other usages in submodules but this fixes the btest failures inside of the main repo, so those are less important.


ckreibich:
Please ignore the failed cluster test in this one. It's a good one — we caught a race condition in that test that I'll need to resolve.

ckreibich (approved):

PR #2513, opened 2022-10-28
zeekygen: Add BIFs to access declaring scripts

awelzel:
Two new BIFs get_identifier_declaring_script() and get_record_field_declaring_script() mirroring existing functions for comments. This allows to query the declaring script information from Zeek scripts and further determine if a redef operations were involved for record fields or enum names by comparing the declaring script with the one of the involved type. See the tests.

Yet another one of @stevesmoot's requests.


src/zeekygen/zeekygen.bif
```diff
@@ -33,6 +33,26 @@ function get_identifier_comments%(name: string%): string
 	return comments_to_val(d->GetComments());
 	%}
 
+## Retrieve the declaring script associated with an identifier
+## (e.g. a variable or type).
+##
+## name: a script-level identifier
+##
+## Returns: declaring script associated with *name*. If *name* is not a known
+##          identifier, an empty string is returned.
```
bbannier, line 42: I get an empty string for builtins, e.g, `get_identifier_declaring_script("count")`, so either the documentation or the implementation should be tweaked (maybe more interesting with language evolution in mind, same result for `get_identifier_declaring_script("print")`).

Naively I would assume that being able to get results with builtins as well would be useful. I wonder whether it wouldn't make sense e.g., return a richer type, e.g., an instance of

```zeek
type DeclaringScript: record {
	path: string &optional;
};
```

I didn't check, but I suspect `get_record_field_declaring_script` could show the same issue for types without matching scriptland record.

awelzel: > I get an empty string for builtins, e.g, get_identifier_declaring_script("count"), so either the documentation or the implementation should be tweaked (maybe more interesting with language evolution in mind, same result for get_identifier_declaring_script("print")).
> Naively I would assume that being able to get results with builtins as well would be useful.

Hmm, so that was mostly copied from the `_comments` bif. Maybe the fix is to include the `script-level` part before the `identifier` in the `Returns:` documentation?

One could probably explicitly populate zeekygen info for builtins if that was interesting, but it's not clear what the declaring script would be.

awelzel:
@stevesmoot - would you have cycles to skim the added tests if that would work for your use-case? Thanks!

stevesmoot:
will do

stevesmoot:
Reading the tests they seem fine.  Tried using the functions and mostly worked but not 100%, so may have missed a case.  Out of time tonight though, will try Thurs/Fri. Likely on my side, but trying to be thorough.

stevesmoot:
confirmed my problem was in fact my problem.  Was a little surprised at all the base/init-bare.zeek origins (vs the file where they are added to the record).

src/zeekygen/zeekygen.bif
```diff
@@ -73,15 +89,14 @@ function get_package_readme%(name: string%): string
 
 ## Retrieve the Zeekygen-style comments (``##``) associated with a record field.
 ##
-## name: the name of a record type and a field within it formatted like
-##       a typical record field access: "<record_type>$<field>".
+## name: the name of a script-level record type and a field within it formatted
+##       lke a typical record field access: "<record_type>$<field>".
```
ckreibich, line 93:
```suggestion
##       like a typical record field access: "<record_type>$<field>".
```

src/zeekygen/zeekygen.bif
```diff
@@ -98,3 +113,35 @@ function get_record_field_comments%(name: string%): string
...
+##          not point to a known script-level record type or a known field within
+##          a record type, an empty string is returned.
+##
+## .. zeek:see:: get_identifier_declaring_script
+function get_record_field_declaring_script%(name: string%): string
+	%{
+	string accessor = name->CheckString();
+	size_t i = accessor.find('$');
+
+	if ( i > accessor.size() - 2 )
```
ckreibich, line 135: The idea here is to cover both "`$` is the last character" and `i == string::npos`, right?

ckreibich (approved):
Very handy, thanks. My only suggestion is to cover failing name lookups in the tests as well.

src/zeekygen/zeekygen.bif
awelzel: It's the same code as employed by `get_record_field_comments()`. Your explanation is what I'm reading, too.

awelzel:
> mirroring existing functions for comments. This allows to query the declaring
> script information from Zeek scripts and further determine if a redef operations

After reading the commit message again: Technically a redef of a record can happen within the same file and we have cases, but for documentation purposes of "which script adds this field" the proposed ones seem reasonable (and is what the ReST output uses).

Issue #2514, opened 2022-10-28
Allow to process all pending timers


J-Gras:
The documentation for `max_timer_expires` still states that setting this to 0 will cause Zeek to process all timers:
https://github.com/zeek/zeek/blob/2ed42ef7719f078724fd11f00811aca9847eac99/scripts/base/init-bare.zeek#L2099-L2103
As part of #1689 the 0 special case was removed. It would be nice to reintroduce a possibility to expire all timers. Probably the sole use case is reading PCAPs: Processing the same PCAP under different configurations given a fixed limit on expires can cause unwanted side-effects.

PR #2516, opened 2022-10-29
Allow to process all pending timers

J-Gras:
This introduces a new flag to expire all pending timers (see #2514). If a magic/large value approach is preferred, the btest might still be useful.


src/Timer.cc
```diff
@@ -142,7 +142,8 @@ void TimerMgr::Expire()
 int TimerMgr::DoAdvance(double new_t, int max_expire)
 	{
 	Timer* timer = Top();
-	for ( num_expired = 0; (num_expired < max_expire) && timer && timer->Time() <= new_t;
+	for ( num_expired = 0; (num_expired < max_expire || zeek::detail::expire_all_timers) && timer &&
```
awelzel, line 145: Thanks for the report in #2514. Now we know what relied on the `max_expire == 0` magic/assumption. Oh my.

>  If a magic/large value approach is preferred, 

If the fix is to reference a global variable down in `DoAdvance()`, IMO `zeek::detail::expire_all_timers` should be replaced with `zeek::detail::max_timer_expires == 0` instead to re-instantiate the documented behavior. Don't think the magic value is bad per-se, it was just not obvious from the code that it had that meaning.

It might be nicer to pass an explicit "bool expire_all" to `Advance()` / `DoAdvance()`, but then there's already global variable references in `TimerMgr->Process()` so may just not be worth it for now (and would need to deal with `Process()` anyway somehow). Generally wonder if current_dispatched and max_timer_expires should be mostly confined into `TimerMgr` solely, but that's a bit of a different topic.

ckreibich: Could there ever be a reason why one would legitimately set `max_timer_expires` to zero, say for some troubleshooting scenario? Otherwise I'd agree with bringing back zero as special meaning, which is relatively common for our `init-bare` tunables. If we don't do that, then I also vote for explicit arguments to `Advance()` / `DoAdvance()` because it keeps the decision making somewhat more centralized (in `expire_timers()`).

awelzel: #2616 is an alternative for this PR. It doesn't add the explicit parameters, but tries to make it explicit the `== 0` behavior exists.

J-Gras: > Could there ever be a reason why one would legitimately set `max_timer_expires` to zero, say for some troubleshooting scenario?

Given that this would prevent any timer form being dispatched, I doubt there is a use case outside of dev environments.

awelzel: > Given that this would prevent any timer form being dispatched, I doubt there is a use case outside of dev environments.

Also, previously max_timer_expires=0 has had the behavior of dispatching all timers, so we're fixing a regression. I think also we can assume no one has started to rely on `max_timer_expires=0` not expiring anything :crossed_fingers:

Issue #2476, opened 2022-10-13
RFC: add --disable-broker option for removing broker support


timwoj:
This was brought up in the zeek slack recently and again during the Microsoft talk at ZeekWeek. Does it make sense to add an option to configure that would disable all Broker integration? This would disable building CAF/Broker, as well as disable building any features in Zeek that rely on it for communication. In the end, you would effectively get a pure single-process version of Zeek from the build.

timwoj:
From Elad at Microsoft:
> A --disable-broker would be perfect for us.

awelzel:
> This would disable building CAF/Broker, as well as disable building any features in Zeek that rely on it for communication. In the end, you would effectively get a pure single-process version of Zeek from the build.

It also would disable the Telemetry subsystem as that relies on metrics storage in broker/caf. But maybe that's a hint that a non-broker/caf based telemetry/Manager could have a place.

0xxon:
I am not entirely sure how I feel about this. Just being able to assume that we have communication available makes things easier in a couple of ways - there are a bunch of scripts where, without broker, we would have to gate functionality behind an "if broker is available".

The feature matrix of stuff that does/doesn't work with this enabled also gets a bit complex, as the comment of Arne shows.

So - my gut feeling is no. I also don't see why this would be a problem for the typical user - you can just not use it, even if it is built in. Could we instead just have Broker buildable on Windows, assuming that's the issue?

vijaypas:
> From Elad at Microsoft:
> 
> > A --disable-broker would be perfect for us.
I'll second Elad. Yes, a --disable-broker option will be excellent!

awelzel:
This topic has come up in the discussions previously: https://github.com/zeek/zeek/discussions/2063

(The original issue was solved differently though).

ckreibich:
> But maybe that's a hint that a non-broker/caf based telemetry/Manager could have a place.

That was my first reaction, too. Broker has become a risk factor for our releases, so it'd be helpful to flesh out that feature matrix, especially if we can't immediately state it ourselves.

> Could we instead just have Broker buildable on Windows, assuming that's the issue?

Afaik the use-case for both Microsoft and Acumera (@vijaypas) is build minimization.

timwoj:
With all of the work to pick out the pieces of Broker in to separate components (namely Telemetry mentioned above), would this be more possible now?

awelzel:
[...]

PR #2518, opened 2022-10-31
Microsoft's Port of Zeek for Windows 🎉🪟

voidbar:
This is Microsoft's Port and Adaptations of Zeek for Windows 🪟.
Using these changes, one can compile Zeek natively on a windows machine using MSVC and then run the resulting executable.

After brewing this for over a year, spending countless developer hours, and cooperating with the Corelight team, we are excited to share with the community our fruits of labor. 
Also, big thanks to Brim as some of their initial contributions helped us bootstrap our own.

Accompanied to this PR, there are numerous submodules that have already been adated to Windows and already pushed to upstream:
- CMake: https://github.com/zeek/cmake/pull/49
- Paraglob: https://github.com/zeek/paraglob/pull/21
- Binpac: https://github.com/zeek/binpac/pull/20
- Bifcl: https://github.com/zeek/bifcl/pull/18
- Broker: https://github.com/zeek/broker/pull/286
- Gen-Zam: https://github.com/zeek/gen-zam/pull/5
- Libunistd (Part of this PR)


eladsolomon-ms:
Amazing milestone! Can't wait to see what the community would do with Zeek for Windows. 🤩🎉🪟

Kudos to @voidbar and @timwoj for driving this complex contribution process.

CMakeLists.txt
```diff
@@ -653,11 +806,13 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cmake DESTINATION share/zeek
         USE_SOURCE_PERMISSIONS
         PATTERN ".git" EXCLUDE)
 
-# Install wrapper script for Bro-to-Zeek renaming.
-include(InstallShellScript)
-include(InstallSymlink)
-InstallShellScript("bin" "zeek-wrapper.in" "zeek-wrapper")
-InstallSymlink("${CMAKE_INSTALL_PREFIX}/bin/zeek-wrapper" "${CMAKE_INSTALL_PREFIX}/bin/bro-config")
+if ( NOT WIN32 )
```
timwoj, line 809:
```suggestion
if ( NOT MSVC )
```

CMakeLists.txt
```diff
@@ -599,7 +747,12 @@ if ( ${CMAKE_SYSTEM_NAME} MATCHES Linux )
     endif ()
 endif ()
 
-set(DEFAULT_ZEEKPATH .:${ZEEK_SCRIPT_INSTALL_PATH}:${ZEEK_SCRIPT_INSTALL_PATH}/policy:${ZEEK_SCRIPT_INSTALL_PATH}/site:${ZEEK_SCRIPT_INSTALL_PATH}/builtin-plugins)
+set(DEFAULT_ZEEKPATH_PATHS . ${ZEEK_SCRIPT_INSTALL_PATH} ${ZEEK_SCRIPT_INSTALL_PATH}/policy ${ZEEK_SCRIPT_INSTALL_PATH}/site ${ZEEK_SCRIPT_INSTALL_PATH}/builtin-plugins)
+if ( WIN32 )
```
timwoj, line 751:
```suggestion
if ( MSVC )
```

CMakeLists.txt
```diff
@@ -300,7 +435,9 @@ FindRequiredPackage(FLEX)
 FindRequiredPackage(BISON)
 FindRequiredPackage(PCAP)
 FindRequiredPackage(OpenSSL)
-FindRequiredPackage(BIND)
+if ( NOT WIN32 )
```
timwoj, line 438:
```suggestion
if ( NOT MSVC )
```

CMakeLists.txt
```diff
@@ -2,9 +2,125 @@
 # auxil/zeek-aux/plugin-support/skeleton/CMakeLists.txt
 cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
 
+if ( WIN32 )
```
timwoj, line 5:
```suggestion
if ( MSVC )
```

src/Val.cc
```diff
@@ -3963,28 +3963,26 @@ ValManager::ValManager()
...
 	if ( port_num >= 65536 )
 		{
 		reporter->Warning("bad port number %d", port_num);
 		port_num = 0;
 		}
 
-	return ports[port_type][port_num];
+	std::pair key{port_num, port_type};
+	if ( ports.find(key) == ports.end() )
+		{
```
timwoj, line 3978: These braces can be removed (or run clang-format, which will likely remove them).

src/fuzzers/FuzzBuffer.cc
```diff
@@ -2,6 +2,10 @@
 #define _GNU_SOURCE
 #endif
 
+#ifdef _MSC_VER
+#include <mem.h>
+#endif
```
timwoj, line 7: This block should go down underneath the #include below it, per the style guide.

src/plugin/Manager.cc
```diff
@@ -160,6 +162,9 @@ void Manager::SearchDynamicPlugins(const std::string& dir)
 bool Manager::ActivateDynamicPluginInternal(const std::string& name, bool ok_if_not_found,
                                             std::vector<std::string>* errors)
 	{
+#if defined(_MSC_VER)
+	return false;
```
timwoj, line 166: Add a comment here about why this function doesn't work on Windows.

src/util.h
```diff
@@ -95,9 +110,7 @@ namespace zeek
 class ODesc;
 class RecordVal;
 
-// Expose ghc::filesystem as zeek::filesystem until we can
-// switch to std::filesystem.
-namespace filesystem = ghc::filesystem;
+
```
timwoj, line 113: Extra blank line

timwoj (changes requested):
I made a quick pass at it for a first round of comments. 

One general comment is that the code flips back and forth between `#if defined()` and `#ifdef` quite a bit. I know it's sort of superfluous, but it'd be nice to standardize on one of them. That might be something to consider for after this lands though, since it's most likely a problem across the entire code base.

src/Val.h
```diff
@@ -327,13 +333,13 @@ class ValManager
...
-	const PortValPtr& Port(uint32_t port_num, TransportProto port_type) const;
+	const PortValPtr& Port(uint32_t port_num, TransportProto port_type);
 
 	// Host-order port number already masked with port space protocol mask.
-	const PortValPtr& Port(uint32_t port_num) const;
+	const PortValPtr& Port(uint32_t port_num);
 
 private:
-	std::array<std::array<PortValPtr, 65536>, NUM_PORT_SPACES> ports;
+	std::map<std::pair<uint32_t, TransportProto>, PortValPtr> ports;
```
awelzel, line 342: Should we consider switch this to 1) `std::unordered_map` and 2) possibly to just `<uint32_t, PortValPtr>` where the `uint32_t` value is already the port space masked value?

A naive benchmark indicates `unordered_map` being considerably faster once there are enough elements in the map, albeit slower for few elements.
```
-------------------------------------------------------------------------------
Benchmark                                     Time             CPU   Iterations
-------------------------------------------------------------------------------
std_map_std_pair/32                        7.90 ns         7.90 ns     91855437
std_map_std_pair/64                        8.13 ns         8.13 ns     85607659
std_map_std_pair/512                       31.0 ns         31.0 ns     22837893
std_map_std_pair/1024                      48.0 ns         48.0 ns     14880275
std_unordered_map_std_pair/32              14.4 ns         14.4 ns     50066974
std_unordered_map_std_pair/64              14.2 ns         14.2 ns     49192062
std_unordered_map_std_pair/512             14.7 ns         14.7 ns     47303798
std_unordered_map_std_pair/1024            15.6 ns         15.6 ns     44525559
std_unordered_map_just_uint32_t/32         11.1 ns         11.1 ns     62767385
std_unordered_map_just_uint32_t/64         9.26 ns         9.26 ns     75598475
std_unordered_map_just_uint32_t/512        12.0 ns         12.0 ns     58354134
std_unordered_map_just_uint32_t/1024       12.7 ns         12.7 ns     54651908
```

```
#include <benchmark/benchmark.h>

#include <cstdint>
#include <map>
#include <string>
#include <unordered_map>

struct pair_hash
{
    template <class T1, class T2>
    std::size_t operator() (const std::pair<T1, T2> &pair) const {
        return std::hash<T1>()(pair.first) ^ std::hash<T2>()(pair.second);
    }
};

#define TCP_PORT_MASK 0x10000
#define UDP_PORT_MASK 0x20000
#define ICMP_PORT_MASK 0x30000

enum TransportProto
        {
        TRANSPORT_UNKNOWN,
        TRANSPORT_TCP,
        TRANSPORT_UDP,
        TRANSPORT_ICMP,
        };


static void std_map_std_pair(benchmark::State& state) {
  std::map<std::pair<uint32_t, TransportProto>, uint32_t> map;
  const int map_sz = state.range(0);

  for (int i = 0; i < map_sz; i++) {
    const auto k = std::pair<uint32_t, TransportProto>(i, TransportProto(i % 3 + 1));
    map[k] = i;
  }

  int i = 0;
  for (auto _ : state) {
    i = (i + 13) % map_sz;
    const auto k = std::pair<uint32_t, TransportProto>(i, TransportProto(i % 3 + 1));
    benchmark::DoNotOptimize(map[k]);
    benchmark::ClobberMemory();
  }
}

static void std_unordered_map_std_pair(benchmark::State& state) {
  std::unordered_map<std::pair<uint32_t, TransportProto>, uint32_t, pair_hash> map;
  const int map_sz = state.range(0);

  for (int i = 0; i < map_sz; i++) {
    const auto k = std::pair<uint32_t, TransportProto>(i, TransportProto(i % 3 + 1));
    map[k] = i;
  }

  int i = 0;
  for (auto _ : state) {
    i = (i + 13) % map_sz;
    const auto k = std::pair<uint32_t, TransportProto>(i, TransportProto(i % 3 + 1));
    benchmark::DoNotOptimize(map[k]);
    benchmark::ClobberMemory();
  }
}

static void std_unordered_map_just_uint32_t(benchmark::State& state) {
  std::unordered_map<uint32_t, int32_t> map;
  const int map_sz = state.range(0);

  for (int i = 0; i < map_sz; i++) {
    const uint32_t k = i | (i % 2 == 0 ? TCP_PORT_MASK : UDP_PORT_MASK);
    map[k] = i;
  }

  int i = 0;
  for (auto _ : state) {
    i = (i + 13) % map_sz;
    const uint32_t k = i | (i % 2 == 0 ? TCP_PORT_MASK : UDP_PORT_MASK);
    benchmark::DoNotOptimize(map[k]);
    benchmark::ClobberMemory();
  }
}

BENCHMARK(std_map_std_pair)->Range(32, 8<<7);
BENCHMARK(std_unordered_map_std_pair)->Range(32, 8<<7);
BENCHMARK(std_unordered_map_just_uint32_t)->Range(32, 8<<7);

BENCHMARK_MAIN();
```

awelzel: The original std::array version appears to run twice as fast as the std::unordered_map/uint32_t version still. It is a microbenchmark, but PortVal lookups aren't exactly uncommon.

src/IntrusivePtr.h
```diff
@@ -25,6 +27,12 @@ struct NewRef
 	{
 	};
 
+/**
+ * This has to be forward decalred and known here in order for us to be able
```
awelzel, line 31:
```suggestion
 * This has to be forward declared and known here in order for us to be able
```

src/statistics/Plugin.cc
```diff
@@ -0,0 +1,54 @@
+
```
awelzel, line 1: Mostly a high-level comment. I think this is very useful functionality. I have a very similar external plugin stashed away somewhere.

Given this is part of Zeek base, IMO, we can also track the queued times directly in the individual [EventHandler](https://github.com/zeek/zeek/blob/6dec7d6027eb52ad55699b5064238a5eefc9b317/src/EventHandler.h#L18-L20) instances rather than an external data structure.

The bif implementation would then enumerate all event_handler's and return the counts for each (potentially only the `!=0` ones).

I'd also suggest to avoid the "reset" functionality and move to a only-ever-incrementing counter model at least within the base and what the bif exposes. If a consumers are interested in relatives/deltas, they can add logic to compute them, but increasing counters are easier to work with.

.gitmodules
```diff
@@ -70,3 +70,6 @@
 [submodule "auxil/zeek-af_packet-plugin"]
 	path = auxil/zeek-af_packet-plugin
 	url = https://github.com/zeek/zeek-af_packet-plugin.git
+[submodule "auxil/libunistd"]
+	path = auxil/libunistd
+	url = https://github.com/zeek/libunistd
```
timwoj, line 75: The license for libunistd needs to be added to COPYING.3rdparty.

ci/build.ps1
```diff
@@ -0,0 +1,33 @@
...
+"@.Replace("`r`n", "")
+
+$commands += @"
+cmake.exe --build $WorkingDirectory --config $BuildType
+"@
+
+$commands += @"
+cmake.exe --install $WorkingDirectory
+"@
+cmd /c ($commands -join " && ")
```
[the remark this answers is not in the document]
voidbar: Don't we use it as part of the CI process?

src/CMakeLists.txt
```diff
@@ -468,10 +492,18 @@ elseif (${COMPILER_ARCHITECTURE} STREQUAL "power")
        ../auxil/highwayhash/highwayhash/hh_vsx.cc
   )
 elseif(${COMPILER_ARCHITECTURE} STREQUAL "x86_64")
+  if (MSVC)
+    set(_avx_flag /arch:AVX2)
+    set(_sse_flag /d2archSSE42)
```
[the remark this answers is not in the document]
voidbar: done

.cirrus.yml
```diff
@@ -59,7 +59,7 @@ ci_template: &CI_TEMPLATE
   only_if: >
     ( $CIRRUS_REPO_NAME != 'zeek-security' || $CIRRUS_OS != "darwin" ) &&
     ( ( $CIRRUS_PR != '' && $CIRRUS_BRANCH !=~ 'dependabot/.*' ) ||
-    ( $CIRRUS_REPO_NAME == 'zeek' &&
+    ( $CIRRUS_REPO_NAME == 'ms-zeek' &&
```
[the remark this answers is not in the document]
voidbar: done

CMakeLists.txt
```diff
@@ -2,9 +2,125 @@
...
+endif ()
+
+if ( NOT CMAKE_INSTALL_LIBDIR )
+    # Currently, some sub-projects may use GNUInstallDirs.cmake to choose the
+    # library install dir, while others just default to "lib".  For sake of
+    # consistency, this just overrides the former to always use "lib" in case
+    # it would have chosen something else, like "lib64", but a thing for the
+    # future may be to standardize all sub-projects to use GNUInstallDirs.
+    set(CMAKE_INSTALL_LIBDIR lib)
+endif ()
```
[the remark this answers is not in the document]
voidbar: done

CMakeLists.txt
voidbar: Can't do, MSVC being registered only after invocation to the `project()` function

src/fuzzers/FuzzBuffer.cc
voidbar: done

ci/build.ps1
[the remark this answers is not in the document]
timwoj: Not on Cirrus. Cirrus uses the {prepare,build,test}.sh files. That said, we could definitely convert over. I have a note here to convert those files to powershell scripts once everything's been merged.

src/plugin/Manager.cc
voidbar: done

src/util.h
voidbar: done

src/IntrusivePtr.h
voidbar: done

.gitmodules
voidbar: done

CMakeLists.txt
```diff
@@ -2,9 +2,125 @@
...
+    # Set CMAKE flags for supported windows build.
+    set(DISABLE_PYTHON_BINDINGS true)
+    set(BROKER_DISABLE_TESTS true)
+    set(BROKER_DISABLE_DOC_EXAMPLES true)
+    add_definitions(-DDOCTEST_CONFIG_NO_MULTITHREADING)
+
+    # Disable Spicy as it is not yet supported in Windows.
+    set(DISABLE_SPICY true)
+
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
```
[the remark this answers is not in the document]
voidbar: done

src/Val.cc
voidbar: ran clang format.

ci/build.ps1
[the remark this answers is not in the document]
voidbar: > I made a quick pass at it for a first round of comments.
> 
> One general comment is that the code flips back and forth between `#if defined()` and `#ifdef` quite a bit. I know it's sort of superfluous, but it'd be nice to standardize on one of them. That might be something to consider for after this lands though, since it's most likely a problem across the entire code base.

Went over the calls to `#defined (_MSC_VER)` and changed them to `#ifdef _MSC_VER`

voidbar:
> I made a quick pass at it for a first round of comments.
> 
> One general comment is that the code flips back and forth between `#if defined()` and `#ifdef` quite a bit. I know it's sort of superfluous, but it'd be nice to standardize on one of them. That might be something to consider for after this lands though, since it's most likely a problem across the entire code base.

Went over the calls to `#defined (_MSC_VER)` and changed them to `#ifdef _MSC_VER`

ci/build.ps1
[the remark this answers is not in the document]
timwoj: Oh wait, this file was deleted at some point. I must have been looking at a specific commit when i added the comment. Ignore me here. 😅

src/statistics/Plugin.cc
timwoj: Can you think of any reason we can't just use the telemetry framework for this?

awelzel: > Can you think of any reason we can't just use the telemetry framework for this?

He, that's the right question I guess.. Mostly depends on which level we want to use it at: If telemetry framework in script-land, we'd still need to have a bif to get to the actual counters (or implement `new_event` and count the events, but that's probably not going to fly).

Or implement previously mentioned per-EventHandler counter with the C++ telemetry API. That should just work and will make them then available through script-land telemetry.

The two downsides I can think of with the latter:

* I believe C++ API generated metrics are currently unconditionally exposed via Prometheus and I'm unaware of knobs here. Not all setups may want to expose such a large number of metrics if they aren't going to look closely at them.

*   Using either, C++ API or script-land telemetry will make (with what we have today) Broker/CAF a hard dependency.

src/Val.h
timwoj: > The original std::array version appears to run twice as fast as the std::unordered_map/uint32_t version still.

That's a pretty stark difference. I'm not sure that I feel comfortable losing that performance difference for the majority of users.

src/statistics/Plugin.cc
awelzel: >  Both approaches will make Broker/CAF a hard-dependency.

Well - we must have this today already because the SessionManager tracks connections using the C++ telemetry API (those stats are available through bifs, too, so it's maybe softer). I was mainly thinking about #2476 when stating this.

timwoj: Part of #2476 would be figuring out how to do a Broker-free telemetry manager. I just feel like if we have the framework there to track telemetry-type data, adding something new that tracks it in a different way seems sorta backwards. I guess it depends on if Microsoft would be able to use the Broker version of the telemetry data for what they're doing here.

timwoj: Alright I finally had a chance to look at the code around the comments here. I think I agree that it would be better for the individual event handlers to handle this themselves, but with one question. Would leaving it as a plugin make us able to disable the functionality (and thus the cost of keeping track) easier? If no, then it's definitely cleaner to let the event handlers take care of it as it also removes all of the plugin code and the extra hook.

src/Val.h
timwoj: Let's go ahead and change this over to the `std_unordered_map_just_uint32_t` recommended version here, and we can write an issue to revisit the performance difference between that and `std::array` and how to handle it.

src/statistics/Plugin.cc
awelzel: Opinionated - we can probably consider incrementing the metric (or just a C++ counter) per `QueueEvent()` on the handler cheap enough and circumvent the disable functionality question. I can't imagine a `++handler->times_queued` would show up in a profile. If the telemetry/metrics add too much overhead due to being fancier/thread-safe/... we'll probably need to look for improvements there.

timwoj: Makes sense to me. Let's go ahead and make this change as part of this PR.

src/packet_analysis/protocol/ip/IP.cc
```diff
@@ -62,7 +62,8 @@ bool IPAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
 			return false;
 			}
 
-		ip_hdr = std::make_shared<IP_Hdr>((const struct ip6_hdr*)data, false, len);
+		packet->ip_hdr = std::make_shared<IP_Hdr>((const struct ip6_hdr*)data, false,
```
timwoj, line 65: This appears to be the change that broke the btests on the builds. Probably just an issue with the rebase.

```suggestion
		ip_hdr = std::make_shared<IP_Hdr>((const struct ip6_hdr*)data, false,
```

timwoj:
`zeek::util::current_time()` doesn't work because the `gettimeofday()` method in libunistd overflows the 32-bit integers it's trying to insert the values into. The patch below fixes `current_time`, if you want to include it in this PR. I'd just use chrono for everything but I need to make sure it doesn't add some weird performance issues on POSIX systems for (similar to the other chrono change I made). I'll do that after we get this all merged.

```
index 79da74bf9..5a56b3d0c 100644
--- a/src/util.cc
+++ b/src/util.cc
@@ -2039,8 +2039,17 @@ RETSIGTYPE sig_handler(int signo);
 double current_time(bool real)
        {
        struct timeval tv;
+
+#ifdef _MSC_VER
+       auto now = std::chrono::system_clock::now();
+       auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+         now.time_since_epoch());
+       tv.tv_sec = ms.count() / 1000;
+       tv.tv_usec = (ms.count() % 1000) * 1000;
+#else
        if ( gettimeofday(&tv, 0) < 0 )
                reporter->InternalError("gettimeofday failed in current_time()");
+#endif

        double t = double(tv.tv_sec) + double(tv.tv_usec) / 1e6;

```

timwoj:
`zeek::File::SetBuf` has a bug related to the usage of `setvbuf`. The Windows version of that method doesn't follow the same semantics as the POSIX version. The POSIX version allows passing a NULL for the buffer and a zero for the length of the buffer, and `setvbuf` will simply set the buffering time and make the buffer underneath the next time `read`/`write` are called on the file descriptor. On Windows, unless you pass `_IONBF` to the method, you must pass a valid buffer and the length must be between 2 and INT_MAX. This is fine, except that `_IONBF` turns off file buffering altogether, so we'd need to investigate the performance ramifications of that.

Unless, of course, you all know some other way to turn file buffering on and off on a file descriptor other than calling `setvbuf`. Either way, we should probably do the `_IONBF` version for Windows for now in order for things not to crash at startup. The patch below fixes it for now:

```
diff --git a/src/File.cc b/src/File.cc
index b2a167740..2ed354830 100644
--- a/src/File.cc
+++ b/src/File.cc
@@ -202,7 +202,14 @@ void File::SetBuf(bool arg_buffered)
        if ( ! f )
                return;

+#ifndef _MSC_VER
        if ( setvbuf(f, NULL, arg_buffered ? _IOFBF : _IOLBF, 0) != 0 )
+#else
+       // TODO: this turns off buffering altogether because Windows wants us to pass vaild
+       // buffer and length if we're going to pass one of the other modes. We need to
+       // investigate the performance ramifications of this.
+       if ( setvbuf(f, NULL, _IONBF, 0) != 0 )
+#endif
                reporter->Error("setvbuf failed");

        buffered = arg_buffered;
```

src/Val.h
voidbar: Done!

src/Val.cc
```diff
@@ -3962,29 +3962,23 @@ ValManager::ValManager()
...
+const PortValPtr& ValManager::Port(uint32_t port_num, TransportProto port_type)
 	{
 	if ( port_num >= 65536 )
 		{
 		reporter->Warning("bad port number %d", port_num);
 		port_num = 0;
 		}
 
-	return ports[port_type][port_num];
+	if ( ports.find(port_num) == ports.end() )
```
awelzel, line 3975: I think `port_num` needs to be masked here and below based on:
```
uint32_t port_masked = PortVal::Mask(port_num, port_type)
...
return ports[port_masked];
```

Otherwise, for example, TCP port 53 and UDP port 53 will be confused and take the same spot in the map.

timwoj: @voidbar I fixed this on my fork, if you want to merge that PR in here.

timwoj (approved):
I'm happy with where this ended up. LGTM. :shipit: 

@awelzel Did you have any other comments, or can we merge this and fix anything else in separate issues?

src/stats.bif
```diff
@@ -484,3 +484,33 @@ function get_reporter_stats%(%): ReporterStats
...
+		{
+		auto handler = event_registry->Lookup(name);
+		auto call_count = handler->CallCount();
+
+		if ( call_count > 0 )
+			{
+			auto eventStatRecord = zeek::make_intrusive<zeek::RecordVal>(recordType);
+			eventStatRecord->Assign(0, zeek::make_intrusive<zeek::StringVal>(name));
+			eventStatRecord->Assign(1, zeek::val_mgr->Count(handler->CallCount()));
+			rval->Assign(i, std::move(eventStatRecord));
```
awelzel, line 510: For the future: VectorVal has an `->Append()` which looks a bit nicer.

awelzel:
> [@awelzel](https://github.com/awelzel) Did you have any other comments, or can we merge this and fix anything else in separate issues?

I looked at the portval things and the statistics replacement. The latter is now using a plain uint64_t instead of telemetry/metrics infrastructure. For me that's fine and we can re-expose these as metrics through script-land via the bif, too.

I'd probably throw a few name ideas/changes in as comments, but maybe we can do that afterwards. Main one is `times_queued` vs `times_called` in `EventNameCounter` and whether `get_event_handler_call_counts()` could be named more generic given the implementation is using an extensible record type which would allow to return more info. Maybe `get_event_handler_stats()` or `_info()` or something less specific.

src/stats.bif
timwoj: Fixed in c0bd2e43b120bfff490424d4ac54c669c7024d96, along with your other comments below.

kmahyyg:
is there any prebuilt binaries available here? I found microsoft/ms-zeek already 404-ed.

voidbar:
@kmahyyg, there are no pre-built binaries, but you can find the compilation command and docker used in the CI here: https://github.com/zeek/zeek/tree/master/ci/windows

timwoj:
> I found microsoft/ms-zeek already 404-ed.

That repo was the one we were using for the merge. All of the changes that were in there are in the main zeek repository now.

Like Tomer said above, we don't provide binary builds for Windows.

PR #2520, opened 2022-11-01
include-plugins: Fixes for include_directories and library linking

awelzel:
Related PRs:
https://github.com/zeek/cmake/pull/56

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -195,11 +195,6 @@ foreach (plugin_dir ${BUILTIN_PLUGIN_LIST})
         set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${plugin_dir}/cmake")
     endif()
 
-    # Setup the include path for built source artifacts.
-    include_directories(AFTER
-                        ${plugin_dir}/src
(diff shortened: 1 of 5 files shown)


awelzel:
I think this is now okay for someone to look. Building zeek with  `--include-plugins=bro-http2` is/was broken due to

1) spicy-plugin picking up events.bif.h from bro-http2 and complaining there's no Spicy namespace/module.
2) The libnghttp2 and libbrotli .so libraries from bro-http2 are passed upwards via bro_SUBDIR_LIBS and then used with add_executable() for zeek, but that does not result in proper linking them. Add a new bro_PLUGIN_LINK_LIBS and use that instead of bro_SUBDIR_LIBS.
3) `builtin-plugins/_preload__.zeek` was loaded too late and so types declared in bro-http2's types.zeek not visible to bif initialization.

I'm not sure that's all right and nice, but it does allow to builtin bro-http2 and hopefully remove a few more workarounds from @sethhall 's setup.

rsmmr (approved):
It's a bit hard to say for sure if this could break something somewhere, but from looking over the commits, the changes look all reasonable to me and at least I can't think of anything right now where the changes would cause a problem. I'll approve, but let's see if anybody else chimes in.

PR #2527, opened 2022-11-02
Configure script: drop --with-caf, add -D option

### LINKED ISSUE #2524: Drop `--with-caf=` configure flag
Since specifying external CAF no longer works (zeek/broker#285) we should remove that flag altogether. I believe the same applies to Broker, though I haven't verified that specifying one there actually throws an error.

Neverlord:
Drops the broken `--with-caf` option and adds a generic way to pass arguments directly to CMake (basically the same changes as in https://github.com/zeek/broker/pull/287).

With the new `-D` option, developers can plug in a custom CAF like this:

```sh
./configure --build-type=debug -D CAF_ROOT=$PATH1 -D CAFIncubator_ROOT=$PATH2
```

(The CAF incubator pulls in a new networking module for CAF. That has been merged to the main repository in the meantime, so we won't need this anymore after migrating to the next version of CAF.)

Closes #2524.

### DIFF
--- a/configure
+++ b/configure
@@ -83,8 +83,6 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --with-bison=PATH      path to bison executable
     --with-broker=PATH     path to Broker install root
                            (Zeek uses an embedded version by default)
-    --with-caf=PATH        path to C++ Actor Framework install root
-                           (a Broker dependency that is embedded by default)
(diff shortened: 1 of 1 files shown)


awelzel (approved):

Neverlord:
Force-push to resolve conflicts with recent merge of `--display-cmake`.

timwoj (approved):

PR #2531, opened 2022-11-02
Update DNS type strings to match correct mappings

timwoj:
This PR fixes a couple of incorrect mappings from code to strings in the DNS scripts, and adds a few that we were missing in comparison to Wireshark's implementation (mostly from other RFCs).

The fixed mappings are: 
- 30 should be NXT not EID
- 31 should be EID not NIMLOC
- 32 should be NIMLOC not NB. NB was originally defined in RFC 1002, but it was later made obsolete and replaced by NIMLOC. Similarly, NBSTAT was originally defined at 33, but was replaced by SRV.


rsmmr (approved):

awelzel:
I saw a flurry of updates in the external repos for baselines and was wondering if this should make it into `NEWS` so it's visible. Maybe that's already queued somewhere?

timwoj:
It is :)

PR #2532, opened 2022-11-02
LGTM Fixes

timwoj:
Given that LGTM is shutting down as of Dec 1, I figured I'd go through and knock out the easy stuff before we lose access to the findings.


rsmmr (approved):
Approve except for one note.

PR #2543, opened 2022-11-05
Do not merge: integration test for Broker-side CMake changes

Neverlord:
Please don't merge.

This PR is a sanity check for building Zeek with a Broker integration branch that pulls in https://github.com/zeek/broker/pull/288 and https://github.com/zeek/broker/pull/289. I don't see how they could break something and all is well locally, but you never know with CMake. 😉


Neverlord:
Ok, seems safe to pull in the changes.

Issue #2530, opened 2022-11-02
NTLM feature + potential bug?


nadavkluger:
Hello!
I made [these](https://github.com/zeek/zeek/compare/master...nadavkluger:zeek:topic/nadavk/ntlm?expand=1) changes and when I run it on [example.pcapng](https://github.com/zeek/zeek/files/9923870/mb.txt)
(Change file name extension after download) I get different result than I should
![image](https://user-images.githubusercontent.com/27895842/199592559-6851bbad-bbf0-4cc1-8e5c-643f59d63d21.png)
Here is a good script to find the wanted outcome https://github.com/mlgualtieri/NTLMRawUnHide

Besides this problem, I noticed `proc_ntlm_challenge` in `ntlm-analyzer.pac` is called only 3 times for this pcap (added printf) but it should be called way more times. There are many more type II messages in this pcap.

For anyone less fluent in the NTLM protocol the [MS-NLMP].pdf on Microsoft's website is the ultimate reference for details

timwoj:
Did you push the changes to your fork? Github is showing that it's the same as our current master branch.

nadavkluger:
Hmm.. Yes I did. So what is the correct format to show the diff? Upload the files here? (I do not have permissions to open a branch)

nadavkluger:
Update: It turns out it does not work for specific protocols. For instance on [this pcap](https://wiki.wireshark.org/uploads/__moin_import__/attachments/SMB2/smb-on-windows-10.pcapng) it works :)

It seems like the problem is in calling `forward_gssapi` (in smb-gssapi.pac) from `SMB1_session_setup_andx_response_ntlm` ( in smb1-com-session-setup-andx.pac)

Also probably `DCE_RPC_Bind_Ack` misses the security blob :(

timwoj:
Fixed via the above PR.

PR #2551, opened 2022-11-08
NTLM challenge & response

nadavkluger:
Exposing challenge and response from NTLM protocol to script-land.
related to:
https://github.com/zeek/zeek/issues/2530
but does not address the suspected bug. Will open a bug ticket in the near future separately.

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -2871,6 +2871,8 @@ export {
 	type NTLM::Challenge: record {
 		## The negotiate flags
 		flags       : NTLM::NegotiateFlags;
+		## A 64-bit value that contains the NTLM challenge.
+		challenge   : count;
 		## The server authentication realm. If the server is
 		## domain-joined, the name of the domain. Otherwise
 		## the server name. See flags.target_type_domain
@@ -2895,6 +2897,8 @@ export {
 		session_key : string &optional;
 		## The Windows version information, if supplied
 		version     : NTLM::Version &optional;
+		## The client's response for the challenge
+		response    : string &optional;
 	};
 }
 
--- a/src/analyzer/protocol/ntlm/ntlm-analyzer.pac
+++ b/src/analyzer/protocol/ntlm/ntlm-analyzer.pac
@@ -143,15 +143,16 @@ refine connection NTLM_Conn += {
 
 		auto result = zeek::make_intrusive<zeek::RecordVal>(zeek::BifType::Record::NTLM::Challenge);
 		result->Assign(0, build_negotiate_flag_record(${val.flags}));
+		result->Assign(1, ${val.challenge});
 
 		if ( ${val}->has_target_name() )
-			result->Assign(1, utf16_to_utf8_val(zeek_analyzer()->Conn(), ${val.target_name.string.data}));
+			result->Assign(2, utf16_to_utf8_val(zeek_analyzer()->Conn(), ${val.target_name.string.data}));
 
 		if ( ${val}->has_version() )
-			result->Assign(2, build_version_record(${val.version}));
+			result->Assign(3, build_version_record(${val.version}));
 
 		if ( ${val}->has_target_info() )
-			result->Assign(3, {zeek::AdoptRef{}, build_av_record(${val.target_info},  ${val.target_info_fields.length})});
+			result->Assign(4, {zeek::AdoptRef{}, build_av_record(${val.target_info},  ${val.target_info_fields.length})});
 
 		zeek::BifEvent::enqueue_ntlm_challenge(zeek_analyzer(),
 		                                 zeek_analyzer()->Conn(),
@@ -182,6 +183,9 @@ refine connection NTLM_Conn += {
 
 		if ( ${val}->has_version() )
 			result->Assign(5, build_version_record(${val.version}));
+
+		if ( ${val}->has_response() )
+			result->Assign(6, to_stringval(${val.response.string.data}));
 
 		zeek::BifEvent::enqueue_ntlm_authenticate(zeek_analyzer(),
 		                                    zeek_analyzer()->Conn(),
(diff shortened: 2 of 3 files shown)


awelzel:
@nadavkluger - sorry for the late response and thanks for opening a PR. Are you certain you committed the most recent state of your changes to the branch?

Running your branch (and the CI pipeline indicates this too) something is amiss:
```
$ zeek
error in <...>/zeek/scripts/base/init-bare.zeek, line 2877: identifier not defined: NTLM::NTLMv2_Response
```

Is it possible the `ntlmv2response` field you added in init-bare.zeek should be of type string rather than `NTLMv2_Response`?

As of testing. Have any of the pcaps you've shared in #2530 been created by you can could be added to the Zeek code base for testing?

scripts/base/init-bare.zeek
```diff
@@ -2856,6 +2856,8 @@ export {
 		version     : NTLM::Version &optional;
 		## Attribute-value pairs specified by the server
 		target_info : NTLM::AVs &optional;
+
+		challenge   : string;
```
awelzel, line 2860: IMO this should be placed after the `flags` field and the `Assign()` offsets for the challenge record updated within `ntlm-analyzer.pac`. This would allow the layout to stay the same as it appears on the wire and docs

 Also the documentation string would be useful. This is from the NTLM docs: "A 64-bit value that contains the NTLM challenge.".

I haven't fully understood why you prefer to have it as (byte)string rather than a count. I believe you stated that somewhere, but I'm failing to find the reference. Mind providing that as input?

nadavkluger: Will do :) 
I used byte(string) instead of count because I had to convert it from big endian (network order) to little endian/

scripts/base/init-bare.zeek
```diff
@@ -2895,6 +2897,8 @@ export {
 		session_key : string &optional;
 		## The Windows version information, if supplied
 		version     : NTLM::Version &optional;
+
+		ntlmv2response    : string &optional;
```
awelzel, line 2901: Could this be named just `response` instead?

scripts/base/protocols/ntlm/main.zeek
```diff
@@ -35,6 +35,9 @@ export {
 		## Internally used field to indicate if the login attempt
 		## has already been logged.
 		done: bool  &default=F;
+
+		server_challenge: string &log &optional;
+		ntlmv2hash: string &log &optional;
```
awelzel, line 40: This will add `server_challenge` and `ntlmv2hash` to the default `ntlm.log`. I'm tempted to say they aren't needed/interesting in all deployments. `ntlmv2hash` ends-up containing redundant info (username, domain and the challenge).

With your changes adding the `challenge` and `response` fields to the respective NTLM types, the log can be extended via an external package or just additional script. It does not need to be part of Zeek base. Adding them by default requires deployments that are not interesting in them to opt-out. I'd go for opt-in for these.

src/analyzer/protocol/ntlm/ntlm-analyzer.pac
```diff
@@ -138,21 +138,22 @@ refine connection NTLM_Conn += {
 
 	function proc_ntlm_challenge(val: NTLM_Challenge): bool
 		%{
-		if ( ! ntlm_challenge )
+        	if ( ! ntlm_challenge )
```
awelzel, line 141: Here and elsewhere in this file is some space to tabs conversion and spurious whitespace insertion. Mind reverting/aligning?

awelzel (changes requested):

scripts/base/protocols/ntlm/main.zeek
nadavkluger: Hmm, how about having it as an optional script like in kerberos ticket-logging ?

scripts/base/init-bare.zeek
nadavkluger: Done :)

scripts/base/protocols/ntlm/main.zeek
nadavkluger: For now I removed my changes from the ntlm script

scripts/base/init-bare.zeek
awelzel: Hmm, I'm a bit confused. The challenge should've been parsed as little-endian and converted to `count` - did something unexpected happen?

The glossary for NTLM describes the `nonce` as a number and the `challenge` as a nonce, so might be nice if we keep the `uint64_t` / `count`.

awelzel:
@nadavkluger - in your original report #2530 you mentioned:


> smb1
> It seems like the problem is in calling forward_gssapi (in smb-gssapi.pac) from SMB1_session_setup_andx_response_ntlm ( in smb1-com-session-setup-andx.pac)

> dce-rpc
> Also probably DCE_RPC_Bind_Ack (dce_rpc-protocol.pac) misses the security blob :( when NTLM is used inside DCERPC

It doesn't seem like this PR fixes anything around this part. If that's still relevant maybe re-open as a new ticket and possibly add a bit to the description what is done here (mainly exposing challenge/response to script-land?).

src/analyzer/protocol/ntlm/ntlm-analyzer.pac
nadavkluger: Done

scripts/base/init-bare.zeek
nadavkluger: When I use `uint64` and `count` in `ntlm-protocol.pac` and `init-bare.zeek` respectively instead of byte(string) the challenge is in reverse endianess:
image

jsoref:
Could I encourage someone to update the PR summary to use `challenge` instead of `challange`? 🙂

scripts/base/init-bare.zeek
awelzel: That seems mostly a "display" problem, but if bytestring is easier to work with can leave that.

scripts/base/init-bare.zeek
```diff
@@ -2895,6 +2897,8 @@ export {
 		session_key : string &optional;
 		## The Windows version information, if supplied
 		version     : NTLM::Version &optional;
+
```
awelzel, line 2900: Mind a short string describing this field.

src/analyzer/protocol/ntlm/ntlm-analyzer.pac
```diff
@@ -182,7 +183,10 @@ refine connection NTLM_Conn += {
 
 		if ( ${val}->has_version() )
 			result->Assign(5, build_version_record(${val.version}));
-
+        
+        if ( ${val}->has_response() )
+            result->Assign(6, to_stringval(${val.response.string.data}));
+
```
awelzel, line 189: These new lines are indented with spaces rather than tabs. Could you check your editor settings (e.g. noexpandtab, tabstop=8) and re-align.

awelzel:
Ready for mere.

Issue #2426, opened 2022-09-21
Zeek flipped connections doesn’t trigger correct analyzers


fatemabw:
Recently we saw some legit connections from a client using source port 6669 connecting to a web server on port 80, Zeek missed the initial SYN from the client and hence miss-analyzed the connection as “IRC” rather than “HTTP” without flipping the connection (i.e. according to zeek it was an outbound IRC connection from 80 -> 6669, and not the inbound HTTP connection). This is the case with Zeek v5.0.2 and older.

While discussing it with Justin A, he mentioned that the recent[ PR#2191](https://github.com/zeek/zeek/pull/2191) by Vern flips the connection when the initial SYN is lost and if subsequent data or acks are seen in the traffic. When tested a pcap with the above mentioned connection (6669 -> 80) against Zeekv5.1.0-rc1, it actually does flip the connection (i.e. an inbound 6669 -> 80 connection), which is good to see, but again the traffic is classified as IRC and not HTTP. Hence the connection is not fully flipped in terms of what analyzers are triggered after the connection is fliped. Also while testing with misc/dump-events script, the DPD isn’t even invoked at all, which should have caught that the connection is web and not irc, which makes us think that the DPD might also not be functioning correctly when the connections are flipped.

Thanks to Justin, we were able to test out some usecases and we came to a conclusion that it’s a bug in Zeek when connections are flipped without triggering correct analyzers.

Some tested cases:

**Behavior with Zeek v5.1.0-rc1**

Case1: when zeek is run with default analyzers loaded, it flips the connection with IRC analyzer confirmation:

$ zeek -r get-rewrite.pcap
$ cat conn.log
1362692526.939084	CJ8MKQBmTisFsKF3i	141.142.228.5	6669	192.150.187.43	80	tcp	**irc**	0.141744	136	5007	SF	-	-	0	**^hADadFf**	456	7	5371	-

Case2: When zeek is run in bare mode with only HTTP analyzer and conn analyzers loaded, it correctly tags HTTP analyzer to the connection:

$ zeek -r get-rewrite.pcap -b  base/protocols/http  base/protocols/conn
$ cat conn.log
1362692526.939084	CzXECf1cmkQBsxRAE7	141.142.228.5	6669	192.150.187.43	80	tcp	**http**	0.141744	136	5007	SF	-	-	0	**^hADadFf**	448	7	5379	-

[...]

PR #2552, opened 2022-11-09
Fixing productive connections with missing SYN still considered partial after flipping direction.

rsmmr:
In #2191, we added endpoint flipping for cases where a connection starts with a SYN/ACK followed by ACK or data. The goal was to treat the connection as productive and go ahead and parse it. But the TCP analyzer could continue to consider it partial after flipping, meaning that app layers would bail out. #2426 shows such a case: HTTP gets correctly activated after flipping through content inspection, but it won't process anything because `IsPartial()` returns true. As the is-partial state reflects whether we saw the first packets each in direction, this patch now overrides that state for the originally missing SYN after flipping.

We actually had the same problem at another location already as well, where it only happened to work because of another state issue with flipping direction.  The 1st commit in this PR fixes that, but the corresponding unit test then broke after that change. This PR updates that logic as well to override the state.

This fix is a bit of a hack, but the best solution I could think of without introducing larger changes.


awelzel:
Can't say I follow the state completely, but looking that `is_partial` flag and comments, but is the following hunk candidate for a `SetFirstPacketSeen()` invocation?

https://github.com/zeek/zeek/pull/2552/files#diff-c3d1ce1be2b00008ee17808c9bd6c630ee16e7201e2b6006d8cc5d40368bbad3R920-R921

```
		else if ( flags.ACK() && peer->state == analyzer::tcp::TCP_ENDPOINT_ESTABLISHED )
			{
			// No SYN packet from originator but SYN/ACK from
			// responder, and now a pure ACK. Problably means we
			// just missed that initial SYN. Let's not treat it
			// as partial and instead establish the connection.
			endpoint->SetState(analyzer::tcp::TCP_ENDPOINT_ESTABLISHED);
			is_partial = 0;
			}
```

rsmmr:
Yeah, I think you're right, good catch. I've inserted that as well.

awelzel (approved):

timwoj:
Can you throw a note in the NEWS about this before I merge it?

rsmmr:
I'll update NEWS and merge.

PR #2568, opened 2022-11-11
Fast follow-up changes to Windows support

timwoj:
These are a couple of minor things that we had hanging around (thanks @awelzel!) prior to merging Windows support, but didn't want to hold up the PR for.


awelzel (approved):
LGTM

PR #2582, opened 2022-11-15
zeekygen/normalize_script_path: Special case plugin dirnames without _

### LINKED ISSUE #2577: Multiple plugins in ZEEK_PLUGIN_PATH lead to "Duplicate Zeekygen script documentation" warning
If you build (but don't install) 2 plugins, say `t/P1` and `t/P2`, and set up ZEEK_PLUGIN_PATH to include their `build` directories, then when running Zeek you get the message:

> warning: Duplicate Zeekygen script documentation: build/scripts/__preload__.zeek
warning in /Users/vern/zeek/t/P2/build/scripts/__preload__.zeek, line 10: Duplicate Zeekygen script documentation: build/scripts/types.zeek
warning: Duplicate Zeekygen script documentation: build/lib/bif/__load__.zeek
warning: Duplicate Zeekygen script documentation: build/scripts/__load__.zeek

awelzel:
In normal installations, a plugin's basename has an underscore in it to separate the namespace from the plugin name. E.g Zeek_Spicy. When there is no underscore, this is most likely due to ./build being picked up when using ZEEK_PLUGIN_PATH. The basename ends-up "build" and is susceptible to collisions.

Prepend one parent directory as a heuristic to make this scenario less likely, assuming ./build is usually below a repository checkout that uniquely identifies the plugin.

Fixes #2577

### DIFF
--- a/src/zeekygen/utils.cc
+++ b/src/zeekygen/utils.cc
@@ -146,6 +146,16 @@ std::string normalize_script_path(std::string_view path)
 		{
 		auto rval = util::detail::normalize_path(path);
 		auto prefix = util::SafeBasename(p->PluginDirectory()).result;
+
+		// Collision avoidance when there's no _ in the plugin basename such
+		// as when using ./build within a plugin checkout for testing. Include
+		// the parent in the normalized path assuming it's unique.
+		if ( prefix.find('_') == std::string::npos )
+			{
+			auto parent = util::SafeBasename(util::SafeDirname(p->PluginDirectory()).result).result;
+			prefix = parent + "/" + prefix;
+			}
+
 		return prefix + "/" + rval.substr(p->PluginDirectory().size() + 1);
 		}
 


rsmmr (commented):
I'm not very familiar with Zeekygen internals, is this change to the prefix visible in the output?

awelzel:
> I'm not very familiar with Zeekygen internals, is this change to the prefix visible in the output?

Not for the docs.zeek.org documentation: This change is only relevant for script paths of dynamic plugins and "our" documentation is generated without any dynamic plugins loaded.

awelzel:
If this is too ad-hoc I'm all ears for other approaches. As mentioned elsewhere, the scenario in #2577 is a bit special and was looking for a non-intrusive way to squelch it.

rsmmr (approved):

awelzel:
@ckreibich has commented in #2577 two approaches (changing the `build` directory, or being more clever):

> Zeek's plugin manager finds plugins by recursive descent through folders in ZEEK_PLUGIN_PATH. It finds a plugin directory when it sees a __bro_plugin__ file. So in Vern's case ZEEK_PLUGIN_PATH could be something like path/to/t, not path/to/t/P1/build:path/to/P2/build. This could be a guide to making paths unique: if the plugin basename isn't unique, try to make it unique by expanding it up to the relevant ZEEK_PLUGIN_PATH component: here that'd yield P1/build and P2/build. There could still be the current warning once that doesn't work out.

The PR here makes the assumption that if there's not a `_` in the basename, we go up in the path unconditionally rather than detecting a collision. It's somewhat related to above proposal.

So I'm going to merge this and if #2577 comes up in some other form we can be clever then.

PR #2585, opened 2022-11-15
Update Broker for improved shutdown behavior

Neverlord:
Zeek-side sanity checking for ~~https://github.com/zeek/broker/pull/294~~ https://github.com/zeek/broker/pull/296.

This bug showed up during testing at Corelight where Zeek would stall for no apparent reason. We might also have seen this in the testing group before, but we need to confirm this.


timwoj:
Can this be moved out of draft status? This includes all of the changes from the last few days, right?

Neverlord:
> Can this be moved out of draft status? This includes all of the changes from the last few days, right?

Yep, I've just pushed an update to the submodule ref to the right commit (force-pushed the clang-tidy fixes upstream). But other than that, this should be good to go.

awelzel:
Is this still relevant? It seems master has now all the submodule bumps?

Neverlord:
> Is this still relevant? It seems master has now all the submodule bumps?

The original bump has made it's way to the main branch. I've re-purposed the branch for sanity checking of https://github.com/zeek/broker/pull/300. This should also fix the the spurious fails in the `broker.disconnect` btest that we've seen recently.

ckreibich:
So this should not get merged until zeek/broker#300 is resolved, right? The submodule bump currently points at the tip of that PR's branch.

timwoj:
@Neverlord Do you want to rebase this and update it one more time now that the above PR landed?

Neverlord:
@timwoj done, I'll mark this ready for review if CI is happy.

PR #2499, opened 2022-10-23
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/zeek/commit/594a75aa0c34b8b9e5f8162efa821966e7d035b7#commitcomment-87673290

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/zeek/commit/8886aa33edf77be20795c9b5ae14a35d7344f0ca

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

PR #2588, opened 2022-11-15
Spelling testing

jsoref:
Subset of #2499


testing/scripts/snmp-test.zeek
```diff
@@ -87,7 +87,7 @@ function print_snmp_bulk_pdu(pdu: SNMP::BulkPDU)
 	{
 	print fmt("    request_id:      %s", pdu$request_id);
 	print fmt("    non_repeaters:   %s", pdu$non_repeaters);
-	print fmt("    max_repititions: %s", pdu$max_repititions);
+	print fmt("    max_repetitions: %s", pdu$max_repetitions);
```
awelzel, line 90: The second change referencing the field we either need to revert, or adapt `init-bare.zeek`, too. It's typo in the record definition in init-bare.zeek.

https://github.com/zeek/zeek/blob/ec3eca05492031d59f121046f377d101d090a06b/scripts/base/init-bare.zeek#L4661-L4666

Do we want to make a breaking change that fixes the typo in the field, or be nice and not change it, or introduce the fixed name and a `&deprecated` attribute for the old one? Maybe open an issue and drop the `repetitions` fix from this PR.

The CI pipeline output shows:
```
  error in /zeek/testing/btest/../scripts/snmp-test.zeek, line 90: no such field in record (pdu$max_repetitions)
```

Here, for example:
https://cirrus-ci.com/task/4854512750428160?logs=test#L258

awelzel:
Except for the repetitions fix the rest is good. The changes require baseline updates as CI points out - I guess we can do that on our end? @timwoj ? Would suggest to squash all commits down before merge.

https://cirrus-ci.com/task/4854512750428160?logs=test#L0

testing/scripts/snmp-test.zeek
timwoj: I'd leave this change out for now.

timwoj (approved):
The rest of the changes look fine to me. I agree that it should be squashed. I think I sent you instructions before for how to do Baseline updates. If not, can you let me know and I'll show you how. Those should go in as part of this PR since it's what's causing them.

jsoref:
Sorry, lemme update the test baselines...

testing/scripts/snmp-test.zeek
jsoref: Dropped

awelzel (approved):

PR #2597, opened 2022-11-17
Log streams and event groups

awelzel:
This extends the Log::Stream type with a set event_groups for
implicitly disabling events when a stream is disabled.


awelzel:
I think the `&group=` annotation is rather nosiy/cumbersome in the dns.zeek file.

One idea is in one of the commit messages around having a global variable as a convention and referencing that one throughout instead of the strings on each handler. Another would be to leverage module event groups instead.

ckreibich:
> One idea is in one of the commit messages around having a global variable as a convention and referencing that one throughout instead of the strings on each handler. Another would be to leverage module event groups instead.

The module level seems more appealing to me, simply because you wouldn't need to touch as many points in the code. I'm curious whether we'll encounter cases where module-level on/off would be too crude ...  we could potentially expand the module-level enable/disable calls with pattern support if that's the case?

awelzel:
> The module level seems more appealing to me, simply because you wouldn't need to touch as many points in the code. I'm curious whether we'll encounter cases where module-level on/off would be too crude ... we could potentially expand the module-level enable/disable calls with pattern support if that's the case?

Yes, we could open a module called `DNS::Logging` within main.zeek (or create a new file) and then pass the "module group" to the log stream. I yesterday added that possiblity.

I'm not quite sure if and how this all will turn out and if it's (easily) applicable to all log streams.
* dns seemed reasonable
* ssl was a bit more work, but should be okay (mainly disable_analyzer_after_detection muddled a bit with logging)
* c$http state tracking is relied on by many external and some internal scripts. Also, `CONNECT` requests trigger `Tunnel::register_tunnel()` which isn't logging related. So this would either need splitting or not opt-into groups for now.

I'll go through some more and keep notes.

awelzel:
After chatting with @rsmmr / @ckreibich and pinging in community.zeek.org:

At this point we'll not modify the existing base scripts, but only provide the logging infrastructure/changes to associate event attribute/module groups with with logging streams. This will allow external scripts to use this feature/mechanism. If it proves useful we may start using it with some of the log streams in base.

Instead for the `EventGroup` type we'll introduce a convention around what kind of group a string represents. Conceivable a certain pattern like `<module-of-log-id-enum>::<group-name>` would denote a module event group while a bare string would be a attribute event group.

Closing this to re-open a new branch/PR.

PR #2600, opened 2022-11-17
Fix logic error in vfmt() when growing the buffer

dopheide-esnet:
This was changed in a8fc63e (the merge of microsoft/master) introducing a logic error and not growing the buffer when n==buf_len.  We hit this when compiling very large regexs, but vfmt() is used in a couple other places too I think.


dopheide-esnet:
I guess I should point out that vfmt() is called by fmt() so this could break a lot of big strings.

timwoj (approved):
Looks fine to me. Can you fix the commit message?

PR #2603, opened 2022-11-20
Additional script optimization fixes

vpax:
This is a collection of tweaks for script optimization.  The fixes are mainly for `-O gen-C++`: a new initialization model for `-O gen-standalone-C++`, a fix for converting `count` values to `int`, and disambiguation for empty constructors that return different types.  There's also a ZAM fix for C++ optimizers that discard the entire contents of `ASSERT`s.


0xxon (approved):
I admittedly only gave this a quite casual glance - but looks good to me.

PR #2604, opened 2022-11-21
tighten type equivalence for vector constructors

vpax:
The impetus behind this PR was discovering that using recursive types in vector constructors resulted in crashes due to infinite recursion while trying to "merge" the types to come up with the final type to associate with the constructed vector.  (FYI, this comes up in a practical context when constructing a vector of functions that take `connection` records, as those are recursive unless using `-b` bare mode due to connection `RemovalHook`s.)

Rather than changing the merge code to catch such recursion - which is conceptually tricky, as the merge needs to create a new type, rather than just returning a verdict as in the somewhat similar situation for `same_type()` - this change only allows type variation in vector constructors for arithmetic types directly present.  Use of any other type must match directly, which seems like a reasonable enough restriction.

The PR includes a new BTest test case that crashes earlier versions of Zeek.

### DIFF
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -2673,6 +2673,31 @@ TypePtr merge_types(const TypePtr& arg_t1, const TypePtr& arg_t2)
 		}
 	}
 
+TypePtr merge_type_list(detail::ListExpr* elements)
+	{
+	TypeList* tl_type = elements->GetType()->AsTypeList();
+	const auto& tl = tl_type->GetTypes();
+
+	if ( tl.size() < 1 )
+		{
+		reporter->Error("no type can be inferred for empty list");
+		return nullptr;
+		}
+
+	auto t = tl[0];
+
+	if ( tl.size() == 1 )
+		return t;
+
+	for ( size_t i = 1; t && i < tl.size(); ++i )
+		t = merge_types(t, tl[i]);
+
+	if ( ! t )
+		reporter->Error("inconsistent types in list");
+
+	return t;
+	}
+
 TypePtr maximal_type(detail::ListExpr* elements)
 	{
 	TypeList* tl_type = elements->GetType()->AsTypeList();
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -63,11 +63,6 @@ const char* type_name(TypeTag t)
 	return type_names[int(t)];
 	}
 
-// Given two types, returns the "merge", in which promotable types
-// are promoted to the maximum of the two.  Returns nil (and generates
-// an error message) if the types are incompatible.
-static TypePtr merge_types(const TypePtr& t1, const TypePtr& t2);
-
 Type::Type(TypeTag t, bool arg_base_type)
 	: tag(t), internal_tag(to_internal_type_tag(tag)), is_network_order(zeek::is_network_order(t)),
 	  base_type(arg_base_type)
@@ -2696,7 +2691,7 @@ TypePtr maximal_type(detail::ListExpr* elements)
 
 	for ( size_t i = 1; t && i < tl.size(); ++i )
 		{
-		auto& tl_i = tl[i];
+		const auto& tl_i = tl[i];
 
 		if ( t == tl_i )
 			continue;
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -63,6 +63,11 @@ const char* type_name(TypeTag t)
 	return type_names[int(t)];
 	}
 
+// Given two types, returns the "merge", in which promotable types
+// are promoted to the maximum of the two.  Returns nil (and generates
+// an error message) if the types are incompatible.
+static TypePtr merge_types(const TypePtr& t1, const TypePtr& t2);
+
 Type::Type(TypeTag t, bool arg_base_type)
 	: tag(t), internal_tag(to_internal_type_tag(tag)), is_network_order(zeek::is_network_order(t)),
 	  base_type(arg_base_type)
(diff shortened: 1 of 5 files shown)


src/Type.h
```diff
@@ -923,15 +923,10 @@ extern Type* flatten_type(Type* t);
 // Returns the "maximum" of two type tags, in a type-promotion sense.
 extern TypeTag max_type(TypeTag t1, TypeTag t2);
 
-// Given two types, returns the "merge", in which promotable types
-// are promoted to the maximum of the two.  Returns nil (and generates
-// an error message) if the types are incompatible.
-TypePtr merge_types(const TypePtr& t1, const TypePtr& t2);
```
awelzel, line 929: Unless vscode shows it wrong, this and `merge_type_list` below aren't in the `zeek::detail` namespace, so I believe technically they require a deprecation cycle and can be removed earliest in v6.1.

Practically, I'm unsure there are users.

src/Type.h
```diff
@@ -923,15 +923,10 @@ extern Type* flatten_type(Type* t);
...
-
-// Given a list of expressions, returns a (ref'd) type reflecting
-// a merged type consistent across all of them, or nil if this
-// cannot be done.
-TypePtr merge_type_list(detail::ListExpr* elements);
+// Given a list of expressions, returns the maximal type consistent across
+// all of them, or nil if this cannot be done.  "Maximal" incorporates
+// notions of arithmetic coercion, but otherwise requires type-equivalence.
+TypePtr maximal_type(detail::ListExpr* elements);
```
awelzel, line 929: Out of my depth and probably just me, but this new name is rather close to the existing `max_type()`.

Ignore if there's no real other candidate or the distinction is clear enough.

awelzel: shallow_merge_type_list is probably too naive?

src/Type.h
vpax: Yeah, I really doubt there are such users.  I only made `merge_types()` private to `Type.cc` to avoid the confusion of it being exposed but no one apparently using it - we could just keep it public if we want.

For `merge_type_list()`, I do think we should remove it (rather than fix its crash, which as I sketch would be a pain).  I can abide a deprecation cycle for that if the view is we should do so.

vpax: Hmmmm `merge_type_list` takes an argument in the `detail::` namespace.  That seems contradictory ... ?

src/Type.h
vpax: It returns a single type, not a list of types (so I find the original name pretty strange!).  They're not a merge in the sense that one would expect, e.g., it does not coalesce compatible-but-not-identical records.  I'm open to other suggestions, but I think `shallow_merge_type_list` isn't as apt as `maximal_type`.

src/Type.h
awelzel: > Hmmmm merge_type_list takes an argument in the detail:: namespace. That seems contradictory ... ?

That exists in other places as well (`Func::AddBody` / `Func::Invoke`). I would guess the "detail" object being passed/received from non-detail functions should mostly be treated as an opaque if possible.

I'm going to resolve this discussion assuming there are no out-of-tree users.

awelzel: > I only made merge_types() private to Type.cc to avoid the confusion of it being exposed but no one apparently using it - we could just keep it public if we want.I only made merge_types() private to Type.cc to avoid the confusion of it being exposed but no one apparently using it - we could just keep it public if we want.

> I'm going to resolve this discussion assuming there are no out-of-tree users.

Hmm, after more pondering. In the off-chance there being users of `merge_types()`, we don't actually provide a replacement that they could convert to. So I'm paddling back to keeping `merge_types()` available.

src/Type.cc
```diff
@@ -2690,7 +2695,21 @@ TypePtr merge_type_list(detail::ListExpr* elements)
 		return t;
 
 	for ( size_t i = 1; t && i < tl.size(); ++i )
-		t = merge_types(t, tl[i]);
+		{
+		auto& tl_i = tl[i];
```
awelzel, line 2699:
```suggestion
		const auto& tl_i = tl[i];
```

vpax: Okay I've added that tweak and made `merge_types()` externally available again.

awelzel (approved):
Approving. See review comments for consideration that we're removing something from `zeek::`.

src/Type.h
ckreibich: I know it's tempting to call this unlikely and I indeed don't see any users of the function in our package collection, but our contract does actually say "Any changes to this code [i.e., the non-`detail` namespace] will strictly follow the Zeek team's deprecation process" (per the wiki). So I agree — we should either keep the function or put in a deprecation.

vpax: Note, per the below, I've reverted the change to `merge_types()`.

ckreibich: Ah — I read over the latest diff and was referring to `merge_type_list`. If I read the above right, we're in agreement that we should add deprecate it while introducing `maximal_type`, right?

vpax: We can deprecate it, but we don't have a direct replacement.  `maximal_type` will work for simple (and IMHO "sane") cases, but it's not the full semantics.  Given that, what should the `[[deprecated("...")]]` message say?

ckreibich: How about `[[deprecated("Remove in v6.1. Use maximal_type() if possible. See GH-2604.")]]`? That's similar to when we deprecated `MemoryAllocation()`, for which we also didn't provide a direct equivalent.

vpax: Sounds good.  I've pushed a change to do this.

ckreibich (approved):

PR #2617, opened 2022-11-24
parse.y: Allow trailing commas for table, set, vector and record cons…

awelzel:
…tructions

Python, Ruby, Javascript, Go, ..., allow use of trailing commas and is even recommended in some style-guides as it keeps diffs smaller. The black formatter for Python even goes as far to take a trailing comma as an indication to format a list one-item on a line.

It has been a bit unusual to not be able to put trailing commas in Zeek scripts, so this change allows for it.

It probably allows for commas in more places that I haven't fully grasped. Maybe we should tighten those down again if we find them surprising.


awelzel:
> This is okay for invocations in Python too, though I don't believe I've encountered or used it there in practice. It feels a bit strange to me to have this in Zeek, but I could see going either way. Thoughts?

He - that one was not obvious to me, but yes `expr ( opt_expr_list )` gets converted into the callexpr and `opt_expr_list` now accepts the trailing comma.

A single parameter with a trailing comma. If you split parameters one-per-line, the missing trailing `,` looks better IMO.
```
         Analyzer::schedule_analyzer(                                            
                 chan$orig_h,                                                    
                 chan$resp_h,                                                    
                 chan$resp_p,                                                    
                 Analyzer::ANALYZER_FTP_DATA,                                    
                 5mins,                                                          
         );  
```

awelzel:
I added a test around the function declarations and calls. It's definitely used within Python (and black as this magic comma option, too).

The one outcome that annoys me currently a bit is that one can now add trailing commas in table keys, too, while you can't in index expressions.
```
global tab: table[string] of string = {
  ["a",] = "b",
  ["c",] = "d",
}
```
Not sure how easy it is to disallow that, but can give it a whirl.

awelzel:
> The one outcome that annoys me currently a bit is that one can now add trailing commas in table keys, too, while you can't in index expressions.
> Not sure how easy it is to disallow that, but can give it a whirl.

I've force-pushed and this is now disallowed. Also added a few more tests around this.

ckreibich:
I think it's pretty telling that they made this optional even in the oh-so-opinionated black. :-) But I think this now a great sweet-spot for Zeek. I'm also pretty optimistic that we can make `zeek-format` smart enough to add these commas during formatting only when it has already decided to break an expression across multiple lines.

ckreibich (approved):

PR #2618, opened 2022-11-28
CI: add Fedora 37

ckreibich:


awelzel (approved):
Unrelated: Do you know if the FreeBSD 14 task is spuriously failing?

ckreibich:
Yeah it's strange — I just retried it with Cirrus's console feature to dig in, and it looks like it's running through. So it indeed looks spurious. I wonder whether it's just not as stable given that 14 isn't released yet.

awelzel:
I had looked yesterday at the following freshports page: https://www.freshports.org/devel/py-pip

And it did show "pkg-fallout" for the py39-pip package for `FreeBSD:14:amd64`, so probably just intermittent.

PR #2623, opened 2022-11-30
testing/external: Bump cluster testsuite

awelzel:
...also add/propose aliases for sync-commits / sync-repos, because I could not remember for the past 5 times or so in which direction they would sync and would find myself reading the README, Makefile, or comments in the scripts.


awelzel:
@0xxon  - this one fixes the get-id-value.sh test in the external cluster suite.

The related PR is here: https://github.com/zeek/zeek-testing-cluster/pull/21

Let me know how you want to handle this? I *think* if this one passes, we can merge the above and update the commit has with the merged result?

0xxon:
Sounds good to me - I shall do that.

PR #2625, opened 2022-12-01
ci/alpine: Use musl-fts-dev rather than fts-dev

awelzel:
With the release of alpine 3.17 (2022-11-22), the fts-dev package was renamed to musl-fts-dev. Change it accordingly.

This is currently breaking CI in a fork of zeek/zeek. Suspect some caching prevents it from breaking here currently.

Alpine 3.16 is still supported till 2024-05-23, but we're essentially not doing CI testing for it anymore. Appears to be in line with the comment in .cirrus.yml

### DIFF
--- a/ci/alpine/Dockerfile
+++ b/ci/alpine/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:latest
 
 # A version field to invalidate Cirrus's build cache when needed, as suggested in
 # https://github.com/cirruslabs/cirrus-ci-docs/issues/544#issuecomment-566066822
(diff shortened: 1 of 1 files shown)


0xxon (approved):

timwoj:
> Suspect some caching prevents it from breaking here currently.

For reference, unless our dockerfile changes for a build, it won't get rebuilt. That means we were probably still using some installation of Alpine prior to this change. It's possible to manually force a rebuild of our docker images by updating the `DOCKERFILE_VERSION` value in the individual dockerfiles.

PR #2627, opened 2022-12-01
CMakeLists: Disable gold detection for bundled spicy

awelzel:
Two reasons. First, it is surprising that built-in Spicy is using a different linker than the Zeek build. Second, when the gold linker is installed, Spicy will ignore any -fuse-ld= option provided in LDFLAGS due to always appending -fuse-ld=gold to its linker flags. This avoids both scenarios.

See also zeek/spicy#1308

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -548,6 +548,7 @@ endif ()
 
 if ( NOT DISABLE_SPICY )
     if ( NOT SPICY_ROOT_DIR )
+        set(USE_GOLD OFF)
         add_subdirectory(auxil/spicy)
 
         # Set variables used by the spicy-plugin build since we are building Spicy


bbannier (changes requested):
I tested a stock Zeek build with your changes from zeek/spicy#1308, and unless I missed a subtlety I do not think the change here is needed.

awelzel:
> I tested a stock Zeek build with your changes from https://github.com/zeek/spicy/pull/1308, and unless I missed a subtlety I do not think the change here is needed.

If USE_GOLD defaulting to OFF makes it into Spicy and once  bumped in Zeek it's not needed. And for the release branches likely not important enough as no one has complained previously.

Closing.

bbannier:
I'll backport zeek/spicy#1308 to Spicy's 1.5 release branch so it can eventually appear in zeek-5.0 maintenance releases. Until then the workaround would be to manually disable gold in Zeek builds with bundled Spicy with `cmake -DUSE_GOLD=OFF`.

PR #2631, opened 2022-12-02
Added several events for detailed info on the SSH2 key init directions

fox-ds:
Re-submission of PR #2489 after applying the feedback and removing the detection logic. The detection logic can be enabled just like in the test, or by installing a `zkg` package which will be created later on.

Any feedback is, again, appreciated!

Original PR text:

A follow up after our talk last week at ZeekWeek2022, where we showed the example of QUIETEXIT; an implant that will act like an SSH server once it has set up a connection to the attacker. Zeek currently determines the client & server side of an SSH connection based on the TCP handshake. This is mostly true, however as showed in the included PCAP, in some situations Zeek can't correctly determine the client & server roles. By adding an extra event, and optionally logging whether the key exchange was set up by the server, users may be able to check their network for (anomalies in) reverse SSH connections. On purpose, it does not switch the client and server log fields; it's up to the user to check for this reversed setup and possibly reverse these, to not change behavior with previous Zeek versions.

Joost


awelzel:
UBSAN test failure appears unrelated.
```
[#6] broker.disconnect ... failed on retry #2
  % 'btest-diff recv2/recv2.out' failed unexpectedly (exit code 1)
  % cat .diag
  endpoint discovered, found a new peer in the network
  receiver got event, 2
  peer added, handshake successful
  --- /tmp/test-diff.181291.recv2.recv2.out.baseline.tmp	2022-12-02 13:04:03.164525254 +0000
  +++ /tmp/test-diff.181291.recv2.recv2.out.tmp	2022-12-02 13:04:03.160525254 +0000
  @@ -1,3 +1,3 @@
   endpoint discovered, found a new peer in the network
  -peer added, handshake successful
   receiver got event, 2
  +peer added, handshake successful

  % cat .stderr
  <<< [181016] zeek -b ../recv.zeek >recv.out
  warning in ./../recv.zeek, line 6: handler for non-existing event cannot be invoked (my_event)
  received termination signal
  >>>
  <<< [181048] zeek -b ../send.zeek >send.out
  error in /zeek/scripts/base/frameworks/broker/./log.zeek, line 83: Broker error (Broker::PEER_UNAVAILABLE): (00000000-0000-0000-0000-000000000000, *127.0.0.1:1213, "unable to connect to remote peer")
  error in /zeek/scripts/base/frameworks/broker/./log.zeek, line 83: Broker error (Broker::PEER_UNAVAILABLE): (00000000-0000-0000-0000-000000000000, *127.0.0.1:1213, "unable to connect to remote peer")
  received termination signal
  >>>
  <<< [181153] zeek -b ../recv.zeek >recv2.out
  warning in ./../recv.zeek, line 6: handler for non-existing event cannot be invoked (my_event)
  received termination signal
  >>>
```

awelzel (approved):
This looks good to me, but maybe @rsmmr could sign-off as well?

@joost-fox - mind squashing the follow-up commit? I'd soon merge this in then.

On the previous PR #2489 there had been discussion around removing the ssh.log format changes and limit to the mechanism required.

Thanks @fox-ds !

fox-ds:
@awelzel squashing is fine! Thanks again for reviewing!

fox-ds:
I had a look at some of the available PCAPs, and I think only the `ssh2_ecc_init` and `ssh2_dh_gex_init` events can be tested currently, there are no other key exchange PCAP's available. I'll have a look if I can generate two pcaps for both the RSA and GSSAPI key exchange methods, would be a nice addition in the test suite.

fox-ds:
After Googling a bit, I think those two other options are really not frequently used and/or legacy. I'll add in an extra test for the `ssh2_dh_gex_init` and edit some of the `zeek:see::` entries to point to the new events.

fox-ds:
@rsmmr I changed some of the descriptions and references in `events.bif` and also added the extra test.

PR #2636, opened 2022-12-07
logging: Add event_groups field to Stream

awelzel:
Integrate event groups with the logging framework and add a test for demo purposes.

~TODO: cluster-testing get-id-value is failing due to change of `Stream` type. Should probably use our own type over in the cluster-testing repo rather than using something that is prone to be changed.~


scripts/base/frameworks/logging/main.zeek
```diff
@@ -373,6 +373,26 @@ export {
...
+		## This field allows for both, attribute event groups and module
+		## event groups. If an attribute event group exists with a
+		## given name, it is always selected. Further, if the group
+		## name contains a ``::`` module separator and a corresponding
+		## module event group exist, is is used. This allows to toggle
+		## ``&group="DNS::Logging"`` and ``module DNS::Logging`` handlers
+		## by providing just ``DNS::Logging`` as group name.
+		##
+		## It is not possible to disable top-level module event groups
+		## like ``DNS``.
```
rsmmr, line 394: I'm wondering if this is too inconsistent: seems from a user perspective, there's really no reason why this shouldn't work for top-level module names. Instead of looking for `::`, we could instead just check if a module event group exists of that name?

rsmmr (commented):
Looks good to me, except for that one question.

scripts/base/frameworks/logging/main.zeek
awelzel: I've removed the special case. If there's a module of a given name, it's now toggled.

rsmmr (approved):

PR #2638, opened 2022-12-08
[RFC] Move dump-events to base and make runtime configurable only

awelzel:
This is mostly and idea/demo/experiment to measure overhead, but could possibly be used for some other policy scripts.

The hope is that the `disable_module_events("DumpEvents")` short-circuits things enough so there's not actual performance impact from loading this script.

There's a second commit on top that helps this by disabling EventHandler when all the function bodies are disabled to shape off a few more cycles. That's fairly ad-hoc and could be more integrated, but I think it's "right" here.

This make the `DumpEvents` module  fully runtime configurable by disabling its module events when `DumpEvents::enabled` is `false`.

I'll gather some `zeek -r` timings, but posting this here for general thumbs-up/thumbs-down thoughts.


rsmmr:
Assuming performance isn't hurt, I like it. I'm also generally a fan of turning things into `options` with runtime configurability. 

(Btw, best to set PR to draft while still in progress, helps understanding if author considers it finished and ready for final review..)

awelzel:
> Assuming performance isn't hurt, I like it. I'm also generally a fan of turning things into options with runtime configurability.
> (Btw, best to set PR to draft while still in progress, helps understanding if author considers it finished and ready for final review..)

Thanks, yes, definitely was more something as an idea for discussion (and dump-events being the worst case example performance wise). I'll gather sum numbers for discussion.

PR #2639, opened 2022-12-08
btest/dump-events: Do not skip everywhere and update baselines

awelzel:


awelzel:
Hmm - I thought that would be easier.

The `run_sync_hook` event  and the `ChecksumOffload::check` event are both scheduled at a 10second interval. It seems they are scheduled in different order on Linux vs MacOSX/BSD, but I can't quite fathom how outside of some timestamp/double sorting acting comparison being different when the timestamps are equal.

Has something like this come up before?

From CI (https://github.com/zeek/zeek/pull/2639/checks?check_run_id=9971238967)
```
   XXXXXXXXXX.XXXXXX Broker::log_flush
  -XXXXXXXXXX.XXXXXX run_sync_hook
   XXXXXXXXXX.XXXXXX ChecksumOffloading::check
  +XXXXXXXXXX.XXXXXX run_sync_hook
   XXXXXXXXXX.XXXXXX filter_change_tracking
   XXXXXXXXXX.XXXXXX new_connection
```

rsmmr:
Is the sort not stable maybe? 

Either way, would be fine to just filter out I'd say. 

Nice catch, btw.

awelzel:
> Either way, would be fine to just filter out I'd say.

I've just spun up FreeBSD to see if that is visible there. I'll resort to filter if nothing obvious comes up. Thanks.

awelzel:
> Is the sort not stable maybe?

I've added stable sorting to the Function bodies. The side effect should be that handlers of the same priority are now guaranteed to be executed in load order, which I think is nice :-)

awelzel:
> > guaranteed to be executed in load order, which I think is nice :-)
> 
> Actually, that lets me shudder. :-) I'd leave the official stance at "execution order is undefined". Loading order feels to fragile to rely on for my taste.

He, okay ;-) - can we leave the `std::stable_sort` as proposed to have the tests stable, though?

I can remove the note from the commit message about the load order and we don't need to otherwise mention it.

rsmmr (approved):

rsmmr:
Yes, using it for making the tests stable is fine.

PR #2649, opened 2022-12-13
GH-2637: Re-remove the bro symlink wrapper

### LINKED ISSUE #2637: Windows merge undid bro symlink removal
It looks like the windows merge undid 8829292c0c441e0a8bd88d675b0d8611ccfcddca. We should probably pull that in again.

timwoj:
Fixes #2637


awelzel (commented):
Could we remove `zeek-wrapper.in` now, too, or may some external scripting expect that to be found in the source tree?

timwoj:
> Could we remove `zeek-wrapper.in` now, too, or may some external scripting expect that to be found in the source tree?

I don't see why not. The entire point of that file is simply to report errors for the old executables, which this change is disabling anyways.

awelzel (approved):
I guess you'll squash.

PR #2651, opened 2022-12-14
Provide infrastructure to migrate legacy analyzers to Spicy inside the Zeek tree.

rsmmr:
As initial examples, this branch ports the Syslog and Finger analyzers
over. We leave the old analyzers in place for now and activate them
iff we compile without any Spicy.

Needs `zeek-spicy-infra` branches in https://github.com/zeek/spicy/pull/1333, https://github.com/zeek/spicy-plugin/pull/155, https://github.com/zeek/cmake/pull/57, and https://github.com/zeek/zeek-testing-private/pull/11.


src/analyzer/protocol/finger/finger.spicy
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2022 by the Zeek Project. See LICENSE for details.
+
+module Finger;
+
+import zeek;
```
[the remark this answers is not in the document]
rsmmr: Valid point. I went for a single file because these analyzers will be shipping inside the Zeek tree, but I can see going the other way too.

Another thought: the most common use-case for Zeek-specific logic is calling the DPD confirm/reject functions. We could do that differently, so that it doesn't need any Zeek-specific code. For one, calls to `zeek::reject_protocol` might often be unnecessary on the first place because the plugin does them anyways when an error bubbles up; so could just remove. I'll just do that for both syslog and finger.

For `confirm_protocol` we could either extend EVT files to allow doing it from there (say, `on Foo::%done -> confirm`); or we could give Spicy some generic function (`spicy::confirm_protocol`) that host applications can hook into. Then the Zeek plugin would tie that to DPD's confirmation, and other host applications could just leave it a no-op if they don't have anything comparable.

awelzel:
Hey - I've tried to build branch locally and am running into an issue:
```
[error] /home/awelzel/corelight-oss/zeek/src/analyzer/protocol/finger/finger.spicy:42:22-46:2: call does not match any function: spicy_rt::initializeParsedUnit(<value_ref<spicy_rt::ParsedUnit>>, <value_ref<Finger::Reply>>, <const __library_type("::hilti::rt::TypeInfo const*")>)
```

This seems like it should be something local, but I did the recursive submodule update and otherwise usual build.l
```
set -x
PREFIX=${PREFIX:-/opt/zeek-dev}

export CFLAGS="-ggdb -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
export CXXFLAGS="-ggdb -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
export LDFLAGS="-fuse-ld=lld"

./configure \
    --ccache \
    --generator=Ninja \
    --build-dir=./build \
    --prefix=$PREFIX \
    --enable-debug \
    --build-type=debug \
    --disable-zkg \
    --disable-btest \
    --disable-python \
    --disable-zeekctl \
    --disable-zeek-client \
    --disable-broker-tests \
    --enable-static-binpac  \
    --enable-static-broker \
    --binary-package \
    --sanitizers=address \
    --enable-fuzzers \
    $*
```

awelzel:
> This seems like it should be something local, but I did the recursive submodule update and otherwise usual build.

Okay: I've had an existing installation in `/opt/zeek-dev` and appears that `spicyz` picks the installed version of `spicy_rt.hlt` over the one that can be found through the provided `-L <...>/zeek/auxil/spicy/spicy/spicy/lib` from the source tree.

That seems slightly surprising (when compared to `-I` from cc which will preceed system directories) and likely sub-optimal for the builtin case here. After removing `/opt/zeek-dev` locally the error from above is gone.

rsmmr:
> That seems slightly surprising (when compared to `-I` from cc which will preceed system directories) and likely sub-optimal for the builtin case here. After removing `/opt/zeek-dev` locally the error from above is gone.

Ack, I'll look into that.

src/analyzer/protocol/syslog/syslog.evt
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2022 by the Zeek Project. See LICENSE for details.
+
+protocol analyzer Syslog over UDP:
```
awelzel, line 3: It took me a bit to understand why the service in `conn.log` doesn't have `spicy_syslog` and instead just `syslog` that's been showing up for other Spicy analyzers in quite a few places. It seems that is "just" because this line isn't `protocol analyzer spicy::Syslog`. 

Rather unrelated to this branch, but I wonder if we should update the package-template for protocol/packet/file analyzers to remove the usage of the `spicy::` prefix in the generated .evt files, too. The appearance of `spicy_` in the conn.log has come up elsewhere as not useful/appropriate or exposing an implementation detail.

I can start/create an issue over there as well.

src/analyzer/protocol/finger/finger.spicy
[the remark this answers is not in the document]
rsmmr: > For one, calls to `zeek::reject_protocol` might often be unnecessary on the first place because the plugin does them anyways when an error bubbles up; so could just remove. I'll just do that for both syslog and finger.

Turns out the plugin already raises violations automatically for unhandled parse errors, so safe to just skip.

> we could give Spicy some generic function (`spicy::confirm_protocol`) that host applications can hook into. Then the Zeek plugin would tie that to DPD's confirmation, and other host applications could just leave it a no-op if they don't have anything comparable.

This seems generally useful, so I'm going that route.

rsmmr:
This is ready, but I'll start the merge process with the subprojects and will file a few individual PRs there to get their pieces in place first. Once that's all in, I'll turn the main Zeek PR to ready for review.

rsmmr:
This is ready for review.

bbannier:
Before going into the details I wanted to check whether it supports all the use cases we wanted (not all tested in CI). We need to be able to build against an externally installed Spicy (shared or static) and/or against a different version of the plugin, configured with `--with-spicy` and `--with-spicy-plugin`. It looks like this is currently broken even at configure time.

The migration of the Finger analyzer looks okay to me, but CI for that currently hinges on us never making the different sanitizer CI jobs use Spicy (they are currently `--disable-spicy`). Maybe we need a dedicated CI for testing legacy analyzers in a job explicitly without Spicy (maybe as a nightly job @awelzel started putting infra in?).

awelzel:
> and/or against a different version of the plugin, configured with --with-spicy and --with-spicy-plugin. It looks like this is currently broken already when only using one of external Spicy/spicy-plugin, even at configure time.

Mostly FWIW: I could see us removing `--with-spicy-plugin` and ask developers for manual bumping/pointing the submodule instead. For AF_Packet we do not have a similar option and I'd like to avoid the proliferation.

The `--with-spicy` should continue to work, it's such a time-saver when doing one-off container builds that one can just install Spicy's binary packages directly.

> Maybe we need a dedicated CI for testing legacy analyzers in a job explicitly without Spicy (maybe as a nightly job @awelzel started putting infra in?).

I don't think there's actual infra, but I might be missing a bit what you're referring to.

rsmmr:
Arne:

> Mostly FWIW: I could see us removing `--with-spicy-plugin`

Agree with that.

> The `--with-spicy` should continue to work, it's such a time-saver when doing one-off container builds that one can just install Spicy's binary packages directly.

Actually, I'd prefer for that to go, too. We have such a proliferation of  Zeek/Spicy/Plugin versions and ways to install them, it's hard to keep that all working in all combinations. In particular, having one extern but the other not, feels like we could skip.

We do have `--disable-spicy`, and that's actually the only one I use. When I don't want the internal Spicy, I use that and then install Spicy (source or binary) and the plugin separately. Would that work for you (and your container builds), too?

 




> 
> > Maybe we need a dedicated CI for testing legacy analyzers in a job explicitly without Spicy (maybe as a nightly job @awelzel started putting infra in?).
> 
> I don't think there's actual infra, but I might be missing a bit what you're referring to.

rsmmr:
Benjamin:

> The migration of the Finger analyzer looks okay to me, but CI for that currently hinges on us never making the different sanitizer CI jobs use Spicy (they are currently `--disable-spicy`). Maybe we need a dedicated CI for testing legacy analyzers in a job explicitly without Spicy (maybe as a nightly job @awelzel started putting infra in?).

Yep, I noticed that too: looks like we should have a dedicated (non-sanitizer) CI run with `--disable-spicy`.

awelzel:
> We do have --disable-spicy, and that's actually the only one I use. When I don't want the internal Spicy, I use that and then install Spicy (source or binary) and the plugin separately. Would that work for you (and your container builds), too?

Hmm, cringing a bit, but if it pains you too much I'll adapt or use more ccache magic.

Other thought is that bundling dependencies isn't great for distributions (some packaging guidelines suggest to submit upstream patches to allow for external dependencies), but we're probably far from that and Zeek/Spicy are rather closely coupled so that whatever a distribution bundles might not work with the latest Zeek version anyway (or analyzers).

As of spicy-plugin: There's another thought to go even further: In the scenario that one passes `--disable-spicy` and builds everything externally, do we provide a way to leverage Spicy analyzers maintained within Zeek's source tree? If not (or if that would just complicate things), I could also see us merging the plugin completely for Zeek 6.0 with the sentiment that in-tree Spicy analyzers can only be leveraged with the builtin spicy plugin If we want to support in-tree analyzers with an external plugin, that's a different story, but might be a scenario to avoid.

Sorry for all the text, I'll try to look, run this tomorrow :crossed_fingers:

timwoj:
I gave it a cursory glance and it all looks ok to me. My only question is whether we have a plan to deprecate/eventually remove the "legacy" version of the analyzer. Given that we're always building with spicy by default, how long should we expect the old version to live on?

Along those same lines, should there be some sort of warning during configure if there are in-tree spicy plugins, but you've disabled spicy?

rsmmr:
Ok, let's do this: For now, I'll try to get the options back to working so that we don't need to make this call right now; I think I see what broken them. But longer term, we probably still want to think about what combinations we want to keep supporting.

rsmmr:
> My only question is whether we have a plan to deprecate/eventually remove the "legacy" version of the analyzer. Given that we're always building with spicy by default, how long should we expect the old version to live on?

Something to talk about, but my take is that at some point we should stop supporting running without Spicy at all (i.e., remove `—disable-spicy`), and then it becomes an individual per-analyzer decision as we port them over. Then we either immediately remove legacy analyzers once we have Spicy version, or we allow each a grace period of, say, one major release where we keep a knob to turn them back on.

> Along those same lines, should there be some sort of warning during configure if there are in-tree spicy plugins, but you've disabled spicy?

Sure, I'll add something.

rsmmr:
Updated this with (1) supporting external Spicy/plugin versions again (tested in all combinations on macOS), and (2) add warnings about legacy/missing analyzers without Spicy.

CMakeLists.txt
```diff
@@ -586,12 +593,20 @@ if ( NOT DISABLE_SPICY )
     endif ()
 
     if ( NOT SPICY_PLUGIN_PATH )
+        set(_spicy_plugin "included")
```
bbannier, line 596: This value is never used.
```suggestion
```

CMakeLists.txt
```diff
@@ -547,14 +547,17 @@ else ()
 endif ()
 
 if ( NOT DISABLE_SPICY )
+    set(USE_SPICY_ANALYZERS yes)
+
     if ( NOT SPICY_ROOT_DIR )
+        set(HAVE_SPICY yes) # evaluated by Spicy plugin build
```
bbannier, line 553: We never set a value for this for the `else` branch.

This makes me wonder, do we actually need `HAVE_SPICY` on top of `NOT DISABLE_SPICY`? Outside of `NOT DISABLE_SPICY AND SPICY_ROOT_DIR` it seems to mirror `NOT DISABLE_SPICY` perfectly; if the value for the missing branch is `OFF` they would be equivalent.

testing/scripts/diff-remove-spicy-abspath
```diff
@@ -0,0 +1,12 @@
+#! /usr/bin/env bash
```
bbannier, line 1:
```suggestion
#!/usr/bin/env bash
```

bbannier (approved):
From the Spicy and CMake side this looks okay to me, only left trivial issues.

scripts/base/protocols/finger/spicy-events.zeek
```diff
@@ -0,0 +1,33 @@
+##! Events generated by the Syslog analyzer.
+
+@ifdef ( Spicy::available ) # must not be used with legancy analyzer
```
awelzel, line 3:
```suggestion
@ifdef ( Spicy::available ) # must not be used with legacy analyzer
```

scripts/base/protocols/finger/spicy-events.zeek
```diff
@@ -0,0 +1,33 @@
+##! Events generated by the Syslog analyzer.
```
awelzel, line 1:
```suggestion
##! Events generated by the Finger analyzer.
```

scripts/base/protocols/syslog/spicy-events.zeek
```diff
@@ -0,0 +1,21 @@
+##! Events generated by the Syslog analyzer.
+
+@ifdef ( Spicy::available ) # must not be used with legancy analyzer
```
awelzel, line 3:
```suggestion
@ifdef ( Spicy::available ) # must not be used with legacy analyzer
```

src/analyzer/protocol/finger/finger.spicy
```diff
@@ -0,0 +1,54 @@
...
+        # We require valid UTF-8 to weed out binary data.
+        self.user = self.arg[0].decode();
+
+        if ( |self.arg[1]| > 0 )
+            self.host = self.arg[1].decode();
+    }
+
+    on %done  {
+        if ( |self.arg[0]| > 0 || self.whois )
+        spicy::accept_input();
```
awelzel, line 33:
```suggestion
            spicy::accept_input();
```

testing/scripts/have-spicy
```diff
@@ -1,6 +1,2 @@
 #!/bin/sh
-if grep -q "DISABLE_SPICY:BOOL=true" "${BUILD}"/CMakeCache.txt; then
-    return 1
-else
-    true
-fi
+"${BUILD}/zeek-config" --have_spicy >/dev/null
```
awelzel, line 2: I *think* this needs to be `--have-spicy-analyzers` instead - those tests are skipped for me currently and probably also in CI.
After fixing, the test fails with an ASAN build on my end due to the following: https://github.com/zeek/spicy/issues/1310

I think we `--disable-spicy` for the Zeek ASAN build in CI (not sure why that was, actually), so it's not blocking for this PR. It is generally unfortunate that due to zeek/spicy#1310 spicy-dump, spicy-driver, etc from a Zeek ASAN-enabled are not usable.

testing/btest/coverage/default-load-baseline.test
```diff
@@ -7,7 +7,7 @@
 # prefix to make the test work everywhere. That's what the sed magic
 # below does. Don't ask. :-)
 
-# @TEST-REQUIRES: test -x ${BUILD}/auxil/spicy/spicy/bin/spicy-config
+# @TEST-REQUIRES: have-spicy
```
awelzel, line 10:
```suggestion
# @TEST-REQUIRES: ${SCRIPTS}/have-spicy
```
Or, does that happen to be in the path? Elsewhere it's used with the `${SCRIPTS}` prefix so minimally would be nice to be consistent.

I've found `TEST-REQUIRES` lines to be a mine-field, because command not found, or command-failure is hard to distinguish from a failure result. There might be a btest feature lingering :-)

testing/btest/scripts/base/frameworks/logging/field-extension-invalid.zeek
```diff
@@ -1,4 +1,4 @@
-# @TEST-REQUIRES: test -x ${BUILD}/auxil/spicy/spicy/bin/spicy-config
+# @TEST-REQUIRES: have-spicy
```
awelzel, line 1: See above.

CMakeLists.txt
rsmmr: This is how it used to be, and actually it's exactly what I broke originally when the external builds weren't working anymore: I had changed it to get set it on all branches, so that it could be used like `USE_SPICY_ANALYZERS` is used now.  The problem with that is that the spicy-plugin picks up on `HAVE_SPICY` being set; if it is, it doesn't look for an external Spicy installation anymore. So the trick is to leave it undefined if we want it to do that. 

Yeah, it's messy, but nothing I wanted to touch further right now ...

testing/scripts/have-spicy
rsmmr: > I _think_ this needs to be `--have-spicy-analyzers` instead - 

Good catch, I renamed the option. Will update.

> I think we `--disable-spicy` for the Zeek ASAN build in CI (not sure why that was, actually), so it's not blocking for this PR.

Agree, it's an existing problem, something to look into, but I'll leave that for later. FWIW, we do ASAN builds on the Spicy side.

testing/btest/coverage/default-load-baseline.test
rsmmr: Yeah, it's in `PATH`, but doesn't hurt for consistency.

rsmmr: > I've found `TEST-REQUIRES` lines to be a mine-field, because command not found, or command-failure is hard to distinguish from a failure result. There might be a btest feature lingering :-)

Yeah, indeed. I have been thinking about switching to using a particular exit code for "skip".

awelzel (approved):

timwoj:
> Something to talk about, but my take is that at some point we should stop supporting running without Spicy at all (i.e., remove `—disable-spicy`), and then it becomes an individual per-analyzer decision as we port them over. Then we either immediately remove legacy analyzers once we have Spicy version, or we allow each a grace period of, say, one major release where we keep a knob to turn them back on.

I'd agree with removing `--disable-spicy`, if we can figure out how to make the builds work on lower-memory machines. For example, my NUC with 8GB of RAM cannot complete a build unless I turn off spicy. The build eventually hangs, and then the compiler crashes due to being OOM.

rsmmr:
> For example, my NUC with 8GB of RAM cannot complete a build unless I turn off spicy. The build eventually hangs, and then the compiler crashes due to being OOM.

Is that even with `-j1`?

PR #2652, opened 2022-12-14
intel/seen/file-names: Use file_over_new_connection()

### LINKED ISSUE #2647: Zeek Intelligence Framework - Intel::FILE_NAME not matching
I am using Zeek v5.0.2 and have experimented with the Intel framework. I got all Intel type to match and produce an entry in intel.log and notice.log except for the Intel::FILE_NAME. 
I send a file test.exe via HTTP to a HTTP server for testing. In my http.log everything is fine and the orig_filenames field contains the value test.exe. Still, the file name does not result in an Intel match. I also tested the Intel::FILE_HASH by providing the hash of my test file and in this case a match is produced for the file hash.
Are there any ideas on how tho get Intel::FILE_NAME to match with HTTP or could this be a bug?

awelzel:
The seen/file-names script relies on f$info$filename to be populated. For HTTP and other network protocols, however, this field is only populated during file_over_new_connection() that's running after file_new().

Use the file_new() event only for files without connections and file_over_new_connection() implies that f$conns is populated, anyway.

Special case SMB to avoid finding files twice, because there's a custom implementation in seen/smb-filenames.zeek.

Fixes #2647


scripts/policy/frameworks/intel/seen/file-names.zeek
```diff
@@ -3,9 +3,28 @@
 
 event file_new(f: fa_file)
 	{
+	# If there are connections attached, we'll be using
+	# file_over_new_connection() for reporting the
+	# filename instead as it's more likely to be populate.
```
ckreibich, line 8:
```suggestion
	# filename instead as it's more likely to be populated.
```

ckreibich (approved):
LGTM, Arne.

scripts/policy/frameworks/intel/seen/file-names.zeek
awelzel: Fixed and also moved the tests into scripts/policy.

Issue #2521, opened 2022-11-01
GitHub is not displaying our license correctly


timwoj:
Both @Neverlord and @jsoref recently brought up that GitHub isn’t properly detecting our license, as shown below.

![Screen Shot 2022-11-01 at 11 49 01 AM](https://user-images.githubusercontent.com/2653616/199314089-be18b7e2-444e-4f00-8c6f-200acb6b5016.png)

If you click on that license line, you get this window:

![Screen Shot 2022-11-01 at 7 56 13 AM](https://user-images.githubusercontent.com/2653616/199314260-b3fde630-b742-4dd5-86fc-b43544a67cb1.png)

I spent a little time tinkering with this in a personal repo, and it looks like the following issues exist:

- The extra line at the bottom of COPYING is confusing it, so it doesn’t see that it’s 3-clause BSD. If I add that line in my personal repo, the same thing happens. If the line isn't there, it's properly detected.
- Having the Copyright line separated onto multiple lines confuses it. If it's all on one line, it's fine.
- COPYING.3rdparty is confusing it, and it’s trying to use all of the licenses in there as well.

I'm not sure what the right answer here is. I like having the two files named closely to the same thing, mostly so they appear next to each other in the file listing. As for the extra line at the bottom of COPYING, does it need to exist? Is it ok to make the copyright line all one line?

ckreibich:
Jeeeez! I futzed too and for me even the difference between enumerating the clause with "1.", "2.", "3." vs "(1)", "(2)", "(3)" breaks detection.

I noticed that renaming `COPYING.3rdparty` to `COPYING-3rdparty` makes it ignore that file.

Github doesn't actually mention that `COPYING` should work, but [Licensee](https://github.com/licensee/licensee), the Gem [apparently](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license) driving detection, lists it in its guts as something to check. I couldn't find anything in Licensee that would let the repo control detection. Presence of `LICENSE` also doesn't override the others. All of this seems pretty ... rudimentary.

I don't think we have strong rules re. what `COPYING` needs to contain. We can update it to appease Github/Licensee. I'm more concerned that we have that formatting in many of our repos.

jsoref:
[...]

PR #2653, opened 2022-12-14
Some fixes for our license files

timwoj:
This does two things:
- Renames COPYING.3rdparty so that GitHub will stop confusing it for a license file, and hopefully provide proper linkage to the actual COPYING file containing our license.
- Updates our license to 2023, meaning this probably shouldn't be merged until after Jan 1.

Part of #2521, but not a complete fix yet.


ckreibich (approved):
Thank you Tim :-)

PR #2656, opened 2022-12-15
EventRegistry/Func: Disable events when all bodies are disabled

awelzel:
This is just a small optimization on top of what is there.

Add state to Func for tracking if any enabled bodies exist which allows us to propagate it up to the EventHandler::bool() operator. In turn, when all bodies of an event's Func have been runtime disabled, the event itself will not be invoked anymore.

Experiments have shown that this allows runtime toggling of new_event() without performance impact when disabled. This could enable use-cases where new_packet() handlers are enabled for a split second once in a while to either dump or sample raw packet data at runtime.


awelzel:
Pulling out the interesting changes from #2638.

src/EventRegistry.cc
```diff
@@ -171,19 +173,32 @@ EventGroup::~EventGroup() noexcept { }
...
-	static auto is_disabled = [](const auto& g)
+	static auto is_group_disabled = [](const auto& g)
 	{
 		return g->IsDisabled();
 	};
 
-	for ( const auto& func : funcs )
+	for ( auto& func : funcs )
+		{
+		const auto& bodies = func->GetBodies();
```
timwoj, line 191: This variable isn't used anywhere.

timwoj (approved):
One minor comment, looks good otherwise.

PR #2657, opened 2022-12-15
analyzer: Add analyzer.log for logging violations/confirmations

### LINKED ISSUE #2031: Missing `Analyzer{Confirmation,Violation}` methods for file analyzers

awelzel:
By default this only logs all the violations, regardless of confirmation state (for which there's still dpd.log). It includes packet, protocol and file analyzers.

This uses options, change handlers and event groups for toggling the functionality at runtime.

Closes #2031


awelzel:
@rsmmr - any thoughts on that log (format and defaults). You had previously mentioned it here:
https://github.com/zeek/spicy-plugin/issues/101#issuecomment-1081759145

(And I also thought it provides an interesting demo for event groups, too).

scripts/base/frameworks/analyzer/logging.zeek
```diff
@@ -0,0 +1,142 @@
...
+
+	## The record type defining the columns to log in the analyzer logging stream.
+	type Info: record {
+		## Timestamp of confirmation or violation.
+		ts:             time              &log;
+		## "violation" or "confirmation"
+		cause:          string            &log;
+		## "packet", "file" or "protocol"
+		analyzer_kind:  string            &log;
+		analyzer_name:  string            &log;
```
rsmmr, line 25: Doc string missing for this field.

scripts/base/frameworks/analyzer/logging.zeek
```diff
@@ -0,0 +1,142 @@
...
+		## Connection UID if available.
+		uid:            string            &log &optional;
+		## Connection identifier if available
+		id:             conn_id           &log &optional;
+		## File UID if available.
+		fuid:           string            &log &optional;
```
rsmmr, line 31: Maybe regroup so that `uid` and `fuid` are side-by-side?

scripts/base/frameworks/analyzer/logging.zeek
```diff
@@ -0,0 +1,142 @@
...
+		## Data causing failure or violation if available.
+		## Truncated to 40 bytes.
+		failure_data:   string            &log &optional;
+	};
+
+	## Enable analyzer.log
+	option enable = T;
+
+	## Enable analyzer_confirmation, too. They are usually less
+	## interesting outside of
```
rsmmr, line 45: outside of what? :-)

rsmmr (approved):
Nice, I like this.

scripts/base/frameworks/analyzer/logging.zeek
awelzel: Fixed and added a few more things to the whole file..

testing/btest/Baseline/coverage.bare-load-baseline/canonified_loaded_scripts.log
```diff
@@ -116,6 +116,17 @@ scripts/base/init-frameworks-and-bifs.zeek
     scripts/base/frameworks/input/readers/sqlite.zeek
   scripts/base/frameworks/analyzer/__load__.zeek
     scripts/base/frameworks/analyzer/dpd.zeek
+    scripts/base/frameworks/analyzer/logging.zeek
+      scripts/base/frameworks/config/__load__.zeek
```
awelzel, line 120: If I read this diff correctly, we'll now be loading `frameworks/config` and `frameworks/cluster` in bare mode. This might be okay, but wanted to point it out explicitly.

awelzel:
I've reworked this enough so probably should have another look even if approved by Robin.

testing/btest/Baseline/coverage.bare-load-baseline/canonified_loaded_scripts.log
rsmmr: I think that's fine, but maybe should add them them to `base/init-frameworks-and-bifs.zeek` then to make that explicit.

rsmmr (commented):
Looks good to me (again :-)). 

The only thing I'm wondering is if we should add some rate-limiting as a backstop in case an analyzer ends up triggering a ton of violations in some environment.

awelzel:
> 
> The only thing I'm wondering is if we should add some rate-limiting as a backstop in case an analyzer ends up triggering a ton of violations in some environment.

I'm leaning towards not dynamically rate-limiting as the chosen limits might still end-up to high and hurt, or too low and go unnoticed.

I had added an `ignore_analyzers` redef'ble set for excluding analyzers when something is totally off the rails. It's also always possible to completely disable the log, but hopefully either action promotes feedback back to us.

https://github.com/zeek/zeek/pull/2657/files#diff-8ac5e916597115fda9e10c31ca2d9e125c3fd088a3e9eb2f44d863103bfdf58dR61

PR #2663, opened 2022-12-16
Fix sprintf warnings on macOS

timwoj:
The most recent compiler update for macOS marked sprintf as deprecated, so we started getting warnings from all of the places that use it. @vpax recently fixed a few of these in binpac (see https://github.com/zeek/binpac/pull/24). This PR fixes the ones in Zeek itself.

### DIFF
--- a/src/DFA.cc
+++ b/src/DFA.cc
@@ -231,16 +231,17 @@ void DFA_State::Dump(FILE* f, DFA_Machine* m)
 			if ( xtions[i] != s )
 				break;
 
-		char xbuf[512];
+		constexpr int xbuf_size = 512;
+		char* xbuf = new char[xbuf_size];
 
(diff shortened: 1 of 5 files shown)


ckreibich (commented):
Thankfully not too many! I did some smartipants grepping and see two `sprintf()` in Dict.h — we should probably fix these too?

timwoj:
> Thankfully not too many! I did some smartipants grepping and see two `sprintf()` in Dict.h — we should probably fix these too?

Fixed. They were in debug code that doesn't get used normally so I had just skipped over them. I also checked the `vasprintf` usage just to be sure, and it works by allocating a buffer large enough to hold the output string so it's not a concern.

ckreibich (approved):

Issue #2575, opened 2022-11-12
Windows: Live packet capture failures


timwoj:
I ran a script with zeek that simply calls the `Pcap::findalldevs()` bif during `zeek_init`. It returned an empty set of devices. We might need to switch from libpcap to npcap or winpcap instead.

timwoj:
It looks like libpcap/winpcap are pretty non-functional on Windows other than for reading files. https://npcap.com/ appears to be the library nmap and Wireshark are using, but it's a non-free license.

Neverlord:
> The free version of Npcap may be used (but not externally redistributed) on up to 5 systems ([free license details](https://github.com/nmap/npcap/blob/master/LICENSE)). It may also be used on unlimited systems where it is only used with [Nmap](https://nmap.org/), [Wireshark](https://www.wireshark.org/), and/or [Microsoft Defender for Identity](https://docs.microsoft.com/en-us/defender-for-identity/what-is).

How does that even work for Wireshark? I though they're using the GPL. How can they rely on a non-free library?

Anyways, since npcap seems to grant license exceptions to OSS projects, maybe we could just ask them?

timwoj:
> How does that even work for Wireshark? I though they're using the GPL. How can they rely on a non-free library?

I have no idea. You have the option to not install it while installing Wireshark, but that just means that you can't read live traffic. Maybe that's the workaround?

> Anyways, since npcap seems to grant license exceptions to OSS projects, maybe we could just ask them?

Worth a shot, I suppose. As of right now we're not intending to distribute a binary Windows installer, so we might not have to worry about the licensing at all? I legitimately don't know how that works.

ckreibich:
npcap still [implements the regular pcap API](https://npcap.com/vs-winpcap.html), right? If so, we wouldn't really have to switch any code, but perhaps warn if we can tell that on Windows we're not dealing with npcap. Is that what you mean Tim?

timwoj:
I believe it depends on how you install it. I think you can install it in a way that isn't compatible, or at least there was an option during installation for that when I last did it. I'll look into that this week.

timwoj:
We made some strides in Slack today with getting this work, to the point of it reading from a network interface and creating logs. There's work to do to make it more generic so everything will still build and work on other platforms, but the work starts at https://zeekorg.slack.com/archives/CSZBXF6TH/p1671718031812489.

Thanks to Gavin Li for helping to troubleshoot this.

timwoj:
[...]

PR #2671, opened 2023-01-04
Fixes to support the Npcap library on Windows

timwoj:
This is part of the fixes for #2575. This adds support for linking against Npcap, which lets Zeek actually find network interfaces and capture packets briefly. There are some outstanding problems starting from that point, but we'll fix them separately.

- Ignore conan libpcap if PCAP_ROOT_DIR is passed
- Update the cmake submodule to pick up changes for finding the right paths to npcap
- Add lazy-loading of npcap so the library path gets set correctly at startup

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -11,6 +11,37 @@
 
 #ifdef _MSC_VER
 #include <fcntl.h> // For _O_BINARY.
+
+// By default, Windows only looks in the System32 directory for dlls. Npcap installs
+// into System32\Npcap, so we have to add that path to the search path for DLLs or
+// the process won't find it. This is annoying, but it's how the Npcap project
+// recommends we do it. See https://npcap.com/guide/npcap-devguide.html#npcap-feature-native
+// for further info.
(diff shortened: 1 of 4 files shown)


src/main.cc
```diff
@@ -11,14 +11,48 @@
 
 #ifdef _MSC_VER
 #include <fcntl.h> // For _O_BINARY.
+
+// By default, Windows only looks in the System32 directory for dlls. Npcap installs
+// into System32\Npcap, so we have to add that path to the search path for DLLs or
+// the process won't find it. This is annoying, but it's how the Npcap project
+// recommends we do it. See https://npcap.com/guide/npcap-devguide.html#npcap-feature-native
+// for further info.
+static void init_npcap_dll_path()
```
awelzel, line 20: Only thought was whether this could live closer to iosource/pcap and called from an early plugin hook, but seems BPF_Program and maybe others would need it, too. And may be too early for a win.cc file.

awelzel (approved):

PR #2676, opened 2023-01-08
fix for crash when record definitions repeat a field name

vpax:
The following input crashes on Master:
```
type r: record {
        a: count;
        a: count;
};

global b: r;
```
This PR fixes the crash and also normalizes the associated error message.


testing/btest/Baseline/language.record-duplicate-fields/.stderr
```diff
@@ -1,3 +1,2 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-error in <...>/record-duplicate-fields.zeek, line 7: Duplicate field 'a' found in record definition
-
+error in <...>/record-duplicate-fields.zeek, line 7: duplicate field 'a' found in record definition
```
awelzel, line 2: This test (record-duplicate-fields) wasn't crashing on master. Would you be up for adding a new one for the scenario you found? Seems the following does it:
```
$ cat language/record-duplicate-fields-global.zeek 
# @TEST-EXEC-FAIL: zeek -b %INPUT
# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff .stderr

type test: record {
        a: count;
        a: count;
};

global a: test;
```

src/Type.cc
```diff
@@ -1337,6 +1337,10 @@ void RecordType::Create(std::vector<std::optional<ZVal>>& r) const
 		{
 		auto& init = field_inits[i];
 
+		if ( ! init )
```
awelzel, line 1340: This doesn't quite seem the right fix: IIUC, elements of `field_inits` are never NULL. `fields_inits` in the test case has size 1 and the index used is `1` which is overreading the vector. It just happens to yield NULL. Changing `[i]` to `at(i)` raises `std::out_of_range`:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
```

Further, in the same realm, the following test case triggers an assertion:
```
# @TEST-EXEC-FAIL: zeek -b %INPUT
# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff .stderr

type test: record {
        a: count;
        a: count;
        b: count;
};

global a: test;
```
error in ./language/record-duplicate-fields-global.zeek, line 8: duplicate field 'a' found in record definition
zeek: ../src/Type.cc:1062: void zeek::RecordType::AddField(unsigned int, const zeek::TypeDecl*): Assertion `field == field_inits.size()' failed.
```

File: src/Type.cc [hunk 3]
```diff
@@ -1337,6 +1337,10 @@ void RecordType::Create(std::vector<std::optional<ZVal>>& r) const
 		{
 		auto& init = field_inits[i];
```
awelzel, line 1338: Could we drop the `&` here and possibly make it `auto*` although that seems to be deduced. Minimally with the null check below that was a bit confusing.

vpax:
Closing this as it's easier to do a fresh branch/PR rather than tweak this one into shape.

PR #2679, opened 2023-01-09
Fix ZAM memory leak

awelzel:
Patch from @vpax and a zam/basic.test so that the current CI setup
has a bit of a chance of catching leaks/crashes.


awelzel:
Added as a Zeek 5.1.2 candidate.

0xxon (approved):

PR #2681, opened 2023-01-10
fix for crashes when record definitions repeat a field name

vpax:
This is a reworking of https://github.com/zeek/zeek/pull/2676 to incorporate @awelzel's suggestions: (1) address crash (with a different cause than the one originally fixed) that occurs when there are additional record fields beyond the one with the repeated name, (2) btest coverage for the issues, (3) low-level code tidiness.


src/Type.cc
```diff
@@ -1335,7 +1338,11 @@ void RecordType::Create(std::vector<std::optional<ZVal>>& r) const
 
 	for ( int i = 0; i < n; ++i )
 		{
-		auto& init = field_inits[i];
+		auto* init = field_inits[i];
+
+		if ( ! init )
```
awelzel, line 1343: That `if` is never true. With or without this change there are never `nullptr`s in the vector. With this hunk dropped, the test continues fine.

awelzel: I can fix this up tomorrow and merge if you give a thumbs-up, or otherwise you can push.

vpax: > That `if` is never true. With or without this change there are never `nullptr`s in the vector. With this hunk dropped, the test continues fine.

That's weird, because I added the test because the debugger showed it was crashing with a null `init`.  Must be that that was a side-effect of the inits-getting-out-of-sync, now fixed elsewhere.

> I can fix this up tomorrow and merge if you give a thumbs-up, or otherwise you can push.

Sure, please go ahead - thanks!

PR #2692, opened 2023-01-17
github-actions/docker: Build arm64 container images for releases and pushes to master

awelzel:
Building the arm64 images is really slow, so don't do it for PRs at this point.

I've pushed a tag in awelzel/zeek and it got picked up building for two platforms: https://github.com/awelzel/zeek/actions/runs/3942844845

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,6 +32,22 @@ jobs:
         with:
           submodules: "recursive"
 
+      - name: Compute target platforms for container images
+        id: target_platforms
+        run: |
+          # Only build arm64 for tagged releases or pushes to master.
+          ALL_PLATFORMS="platform=linux/amd64,linux/arm64"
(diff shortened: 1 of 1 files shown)


awelzel:
As background, @sethhall tried the "zeek-demo" project on an ARM based OSX and while the `zeek/zeek-dev` image started for him, it was emulated using Qemu and neither AF_PACKET nor libpcap functioned.

awelzel:
Hmm - the build over in awelzel/zeek hit the 6 hours GitHub action timeout. The amd64 build completed after ~1.6h running in parallel with arm64. The arm64 build never completed, so this won't fly :-/

Maybe using the Cirrus arm64/amd64 docker_builder for image building images on the native architecture separately, then docker manifest wrangling once that is done within a Github Action.

It seems time is spent mostly on caf / Spicy, so switching to external dependencies might help, too. But they don't exist for arm64, so that's not an option currently.

Issue #2674, opened 2023-01-06
Container images for arm64


awelzel:
@sethhall brought up it would be useful to him to have container images for an architecture other than x86_64

Tutorials around this involve docker/setup-qemu-action and buildx in GithubActions.

awelzel:
For reference, those are the errors  an a M1 mac with colima and x86 emulation available when attempting to start `zeek -i ...`:
```
$ docker run --rm -it zeek/zeek-dev
Unable to find image 'zeek/zeek-dev:latest' locally
latest: Pulling from zeek/zeek-dev
8740c948ffd4: Pull complete
ca7f5aab30b7: Pull complete
fccc4664a9e6: Pull complete
Digest: sha256:9741ab90e2e66a6791192291acd7de4d761e062c1083a1a012bb0acfc8873915
Status: Downloaded newer image for zeek/zeek-dev:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
root@58e0f0f7d552:/# zeek -i eth0
fatal error: problem with interface eth0 (pcap_error: not open (pcap_create))

root@58e0f0f7d552:/# zeek -i af_packet::eth0
fatal error: problem with interface af_packet::eth0 (Protocol not available)
```

PR #2694, opened 2023-01-18
Build arm64 container images and move to Cirrus CI

awelzel:
### Motivation

Modern ARM based M1 OSX systems can not use the container images we publish, see #2674.

After trying to build arm64 images within a GitHub action using buildx (#2692), it was found that the job times out after 6 hours. Cirrus offers docker_builder tasks for amd64 and arm64 natively and builds are faster (Seem to complete after 20-30 minutes, no caching). However, when using separate builder, we can't use buildx and resort to using `docker manifest` by hand to create the manifest list.

### Images and tags created

For builds with `$CIRRUS_TAG` on branch `release/*`, two arch specific tags are pushed into `zeek/zeek` and both included into the `zeek/zeek:<tag>` manifest:
* `zeek/zeek:<tag>-arm64`
* `zeek/zeek:<tag>-amd64`

For builds on `master`, the tags included in the `zeek/zeek-dev:latest` manifest list are:
 * `zeek/zeek-dev:latest-arm64`
 * `zeek/zeek-dev:latest-amd64`

This branch also includes pushing to `zeek/zeek-next:latest` for testing, with the tags being `zeek/zeek-next/latest-<arch>`. This is to be removed or commented out upon merge.

### TODO

 * ~Add AWS ECR pushing?~
 * ~Drop the GitHub action container image build completely? Currently the cluster testing piggy backs on the container image that is built there. It could wait for the cirrus action to build the image and use that. For now, would keep it.~

Relates to #2674


ckreibich:
Just for context: we so far thought it was relatively convenient to have the image builds on Github because we're already slow to complete a full run of builds on Cirrus, plus we tend to max out our credits there. So while Github's been slow for individual builds, when considering it in parallel to Cirrus's larger workload it's been a practical split of the jobs.

timwoj:
> Just for context: we so far thought it was relatively convenient to have the image builds on Github because we're already slow to complete a full run of builds on Cirrus, plus we tend to max out our credits there. So while Github's been slow for individual builds, when considering it in parallel to Cirrus's larger workload it's been a practical split of the jobs.

Public repositories don't use compute credits, so you're fine there. I agree that adding another build just makes everything take that much longer but @0xxon also has a PR open that removes two of them. We do need to disable this for private repos though, like we're already doing for some of the other tasks.

awelzel:
Okay, the pushing do Docker Hub has worked, the ones to AWS ECR fail because there's no `zeek/zeek-next` to push to (which has been used for testing on Docker Hub).
```
[16:21:37.848] The push refers to repository [public.ecr.aws/zeek/zeek-next]
[16:21:37.977] 3a9b6fb6c300: Preparing
[16:21:37.977] a871b97de18e: Preparing
[16:21:37.977] afd7e44a4e08: Preparing
[16:21:38.389] name unknown: The repository with name 'zeek-next' does not exist in the registry with id 'zeek'
```

The multiarch image was pushed to `zeek/zeek-next:latest`, inspecting it:
```
$ docker manifest inspect zeek/zeek-next:latest | jq -c  '.manifests|.[]|.platform'
{"architecture":"amd64","os":"linux"}
{"architecture":"arm64","os":"linux","variant":"v8"}
```

A M1 Mac user, also reports that with the `zeek/zeek-next:latest`, running `zeek -i af_packet::eth0` starts and listens without an error.

awelzel:
@timwoj / @ckreibich - I've marked this as ready for review after removing the testing bits. Things should work out. Minimally we should see if `zeek/zeek-dev` is updated from `master` soon after merging this.

If #2719 makes it in we could consider skipping the image builds for PRs, too. The publishing step should not trigger for PRs already and only for `zeek/zeek`, too.

awelzel:
Okay - Moving this into read-for-review again.

The cluster testing has now also been moved over to Cirrus. ccache was hooked up for the container builds (mostly inspired by the zeek-testing-cluster approach).

As a comparison, the GitHub action to build the amd64 image takes ~50-60 minutes, while in this branch using ccache with a populated cache it's down to ~3mins. Not a fair comparison, of course, but maybe a bit of a motivation to move this forward.

I'm not super happy about all the scripting and how it looks, but it would be great to move this forward :-)

.cirrus.yml
```diff
@@ -505,3 +505,30 @@ container_image_manifest_docker_builder:
   depends_on:
     - arm64_container_image
     - amd64_container_image
+
+cluster_testing_docker_builder:
+  cpu: *CPUS
+  memory: *MEMORY
+  only_if: $CIRRUS_REPO_FULL_NAME == 'zeek/zeek'
```
timwoj, line 521: Should this be limited to not run on every PR?

timwoj (commented):
Just a couple of questions.

.cirrus.yml
awelzel: I'm fine either way - the GitHub action currently runs on PRs and those tests are not otherwise covered by the existing tasks.

.cirrus.yml
```diff
@@ -382,3 +382,95 @@ windows_task:
...
+# more than 6 hours and timed out. Using separate builders on Cirrus allows
+# us build natively and much faster at the expense of the docker manifest
+# wrangling (and not being able to use the nice GitHub actions).
+docker_build_template: &DOCKER_BUILD_TEMPLATE
+  set_image_tag_script: echo "IMAGE_TAG=zeek/zeek-multiarch:${CIRRUS_ARCH}" >> $CIRRUS_ENV
+  build_script:
+    - git submodule update --init --recursive
+    - set -x; cd docker && docker build -f Dockerfile --tag ${IMAGE_TAG} ..
+    - set -x; docker save $IMAGE_TAG | zstd > image.zst
+    - curl -v -X POST --data-binary @image.zst http://$CIRRUS_HTTP_CACHE_HOST/${CIRRUS_BUILD_ID}-image-${CIRRUS_ARCH}
```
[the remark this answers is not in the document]
awelzel: I can give it a shot - the manifest task fetches those images again, so the `curl` invocations seemed nice and explicit, but let me try.

.cirrus.yml
awelzel: I'd leave it for now.

timwoj (approved):

PR #2701, opened 2023-01-19
"Best Guess" unknown keyboard / language variants

jeff-bb:
If the lookup table does not have an entry, it will just log as the raw decimal language/keyboard code. With this change, if we do not have an entry in the lookup table, we'll look at the low order / 4 least significant bits to see if we have a match. The high order / 4 most significant bits are flags/modifiers to the base language/keyboard code. We'll append that it is a "Best Guess"

(This is my first attempt at Zeek scripting, apologies upfront if I'm missing obvious language features. I feel like the const language lookup should return a success/fail return code that we would key off of, but unsure how to accomplish that so instead went for string matching on value in == value out).

### DIFF
--- a/scripts/base/protocols/rdp/main.zeek
+++ b/scripts/base/protocols/rdp/main.zeek
@@ -195,7 +195,7 @@ event rdp_client_core_data(c: connection, data: RDP::ClientCoreData) &priority=5
 		{
 		if (data$keyboard_layout & 0xffff in RDP::languages)
 			{
-			c$rdp$keyboard_layout = fmt("%s (Best Guess)", RDP::languages[data$keyboard_layout & 0xffff]);
+			c$rdp$keyboard_layout = fmt("%s (Best Guess of %d)", RDP::languages[data$keyboard_layout & 0xffff], data$keyboard_layout);
 			}
 		else
 			{
--- a/scripts/base/protocols/rdp/main.zeek
+++ b/scripts/base/protocols/rdp/main.zeek
@@ -188,6 +188,13 @@ event rdp_client_core_data(c: connection, data: RDP::ClientCoreData) &priority=5
 	set_session(c);
 
 	c$rdp$keyboard_layout       = RDP::languages[data$keyboard_layout];
+	
+	if (c$rdp$keyboard_layout == fmt("keyboard-%d", data$keyboard_layout))
+		{
+		c$rdp$keyboard_layout = RDP::languages[data$keyboard_layout & 0xffff];
+		c$rdp$keyboard_layout = c$rdp$keyboard_layout + " (Best Guess)";
+		}
+	
 	c$rdp$client_build          = RDP::builds[data$client_build];
 	c$rdp$client_name           = data$client_name;
 	c$rdp$client_dig_product_id = data$dig_product_id;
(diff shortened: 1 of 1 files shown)


scripts/base/protocols/rdp/main.zeek
```diff
@@ -188,6 +188,13 @@ event rdp_client_core_data(c: connection, data: RDP::ClientCoreData) &priority=5
 	set_session(c);
 
 	c$rdp$keyboard_layout       = RDP::languages[data$keyboard_layout];
```
timwoj, line 190: This could actually use the `in` operator to reduce some of the calls to `fmt`.

Something like:

```
if ( data$keyboard_layout in RDP::languages )
	{
	c$rdp$keyboard_layout = RDP::languages[data$keyboard_layout];
	}
else
	{
	c$rdp$keyboard_layout = fmt("%s (Best Guess)", RDP::languages[data$keyboard_layout] & 0xffff]);
	}
````

See if that does what you're expecting from it.

**Review by jeff-bb — COMMENTED** [review]
I've implemented suggested changes.

File: scripts/base/protocols/rdp/main.zeek [hunk 2]
```diff
@@ -187,7 +187,22 @@ event rdp_client_core_data(c: connection, data: RDP::ClientCoreData) &priority=5
...
-	c$rdp$keyboard_layout       = RDP::languages[data$keyboard_layout];
+	if (data$keyboard_layout in RDP::languages)
+		{
+		c$rdp$keyboard_layout = RDP::languages[data$keyboard_layout];
+		}
+	else
+		{
+		if (data$keyboard_layout & 0xffff in RDP::languages)
+			{
+			c$rdp$keyboard_layout = fmt("%s (Best Guess)", RDP::languages[data$keyboard_layout & 0xffff]);
```
awelzel, line 198: Should we include the original value E.g. `"(Best Guess of %d)"` or `%08x`? We otherwise don't have the actual value in the log anymore.

awelzel:
Do you happen to have a pcap you can share and that can be added as a test where the "best guess" logic is triggered?

scripts/base/protocols/rdp/main.zeek
jeff-bb: In the old code, if there was a match, just the human readable value was listed and the decimal numerical if not. The change here expands the matching capability, but in the event of no match, it still logs in the previous form.

I agree with this though, if we aren't a 100% match, why not include the raw in the output? I'll make the change shortly.

scripts/base/protocols/rdp/main.zeek
```diff
@@ -187,7 +187,22 @@ event rdp_client_core_data(c: connection, data: RDP::ClientCoreData) &priority=5
...
+		{
+		if (data$keyboard_layout & 0xffff in RDP::languages)
+			{
+			c$rdp$keyboard_layout = fmt("%s (Best Guess of %d)", RDP::languages[data$keyboard_layout & 0xffff], data$keyboard_layout);
+			}
+		else
+			{
+			c$rdp$keyboard_layout = fmt("keyboard-%d", data$keyboard_layout);
+			}
+	}
```
timwoj, line 204: Small indentation issue here, but I'll fix it in the merge.

timwoj (approved):

PR #2702, opened 2023-01-20
Rework TeredoAnalyzer::DetectProtocol to not catch as many packets

timwoj:
The current version of this code catches all packets that start with `0x00 0x00` and `0x00 0x01`, which leads to a large number of analyzer violations from Teredo. This PR reworks this code to check for the start of an IPv6 header after the Teredo headers. This should help avoid a large number of false positives.

It also includes a number of btest updates. Most of these are because `AnalyzerViolation` ends up calling `UID::Set`, which bloats the number of UIDs assigned. Removing all of these additional calls causes all of the conn IDs in the test baselines to be different for certain tests. The other change is that the test for the new analyzer.log functionality needs to use a different pcap since the wikipedia trace no longer triggers Teredo violations.


src/packet_analysis/protocol/teredo/Teredo.cc
```diff
@@ -270,10 +270,62 @@ bool TeredoAnalyzer::DetectProtocol(size_t len, const uint8_t* data, Packet* pac
...
+		return true;
+
+	// If the above regex failed, a Teredo packet *must* start with a zero in the first
+	// byte. Dump anything otherwise.
+	if ( data[0] != 0x00 )
+		return false;
+
+	uint16_t val = data[1];
+
+	if ( val == 1 )
```
awelzel, line 286: Looks different and more commented, but the basis for this was the Wireshark source you mentioned I guess? Should we put a link in the commit message or even inline as a reference/credit? Any rules around that?

https://gitlab.com/wireshark/wireshark/-/blob/master/epan/dissectors/packet-teredo.c#L237-302

awelzel: Hmm, taking a step back, `Teredo::DoParse()` seems to do what `DetectProtocol()` now does as well including weird logging.

So we have the parsing logic twice now which makes me think the naive initial patch to trigger the violation only after confirmation just as good/bad? I can see either.

What plays into this all is that for packet analyzer we don't have an easy way to disable / remove the analyzer from a session like we can do for protocol analyzer. So `DetectProtocol()` will run for any following packets of a session, too.

JustinAzoff:
ah, neat, teredo is one of the reasons we initially added the DPD::max_violations, since that kept happening.

src/packet_analysis/protocol/teredo/Teredo.cc
timwoj: > So we have the parsing logic twice now which makes me think the naive initial patch to trigger the violation only after confirmation just as good/bad? I can see either.

I can as well. My concern with just adding the confirmation checks is that we're still going to be calling `AnalyzePacket` for a bunch of packets that should have been ignored. Do you think it'd be enough to add a very-naive check for a packet starting with either `0x00` or in the range of `0x60`-`0x6f`? That should kick out a lot of packets and not duplicate all of the work in `DoParse`.

Maybe just add both checks?

awelzel: > Do you think it'd be enough to add a very-naive check for a packet starting with either 0x00 or in the range of 0x60-0x6f?

You mean in DetectProtocol() and removing the regex then? I have unfortunately not a good intuition :-(

I also realize that in AnalzyerPacket() there's the short-circuiting for when a violation happened and that will stop being effective if we never call `AnalyzerViolation()`
```
	else if ( AnalyzerViolated(packet->session) )
		return false;
```

I'd be also fine with the duplication as is if that filters out most. Hopefully the other tunnels aren't as trigger-happy :-/

timwoj: > You mean in DetectProtocol() and removing the regex then? 

I think if I check for those two bytes, even before doing the regex, would cut out most things faster than waiting for the regex first.

> I'd be also fine with the duplication as is if that filters out most. Hopefully the other tunnels aren't as trigger-happy

Teredo and AYIYA are the only analyzers (tunnels or not) that implements DetectProtocol, because they were the only ones doing DPD before the tunnel analyzers got moved to be packet analyzers. AYIYA's uses some magic numbers in the first three bytes, but it's not nearly as broad as Teredo.

awelzel: Let me resolve that thread, I'm also okay without the reference and think I understood what you intend with the fast check.

Only last thought would to extend the pre-check with `(data[0] == 0x00) && (data[1] == 0x00 || data[1] = 0x01)` open-coding the other regex part.

awelzel (approved):

awelzel:
> It also includes a number of btest updates. Most of these are because AnalyzerViolation ends up calling UID::Set, which bloats the number of UIDs assigned. Removing all of these additional calls causes all of the conn IDs in the test baselines to be different for certain tests.

I once ran into that, too, and was mighty confused. Could maybe assign UIDs unconditionally to have that more stable. IIUC that needs to happen anyway with the first event that includes a ConnVal record.

timwoj:
I'm going to include this and the follow-up in the next 5.1 release.

PR #2707, opened 2023-01-23
store: Extend SQLiteOptions around data safety and performance

awelzel:
Add configurability of synchronous and journal_mode for SQLite backed
Broker data stores. Setting these to synchronous=normal and journal_mode=wal
can significantly improve throughput at the cost of some durability in
the presence of power loss or OS crash. In the context of Zeek, this is
likely more than acceptable.

Additionally, add integrity_check and failure_mode options to support deleting
and re-opening a corrupted SQLite database at store creation.


scripts/base/frameworks/broker/store.zeek
```diff
@@ -57,12 +57,62 @@ export {
...
+		SQLITE_JOURNAL_MODE_WAL,
+	};
+
 	## Options to tune the SQLite storage backend.
 	type SQLiteOptions: record {
 		## File system path of the database.
 		## If left empty, will be derived from the name of the store,
 		## and use the '.sqlite' file suffix.
 		path: string &default = "";
+
```
ckreibich, line 87: We probably want `##` on these comments so they get processed by Zeekygen?

scripts/base/frameworks/broker/store.zeek
```diff
@@ -57,12 +57,62 @@ export {
...
+		# database left in place and the data store in a
+		# non-functional state.
+		failure_mode: FailureMode &default=FAIL;
+
+		# Run PRAGMA integrity_check after opening the database
+		# and fail according to ``failure_mode``. Running PRAGMA
+		# integrity_check may take a non-negligible amount of time,
+		# so you are advised to experiment with the expected sizes
+		# of your databases if that is acceptable. Corrupted databases
+		# should be reliably detected when this setting this ``T``.
```
ckreibich, line 114:
```suggestion
		# should be reliably detected when this setting is ``T``.
```

testing/btest/broker/store/brokerstore-backend-sqlite-corrupt-delete.zeek
```diff
@@ -0,0 +1,42 @@
+# @TEST-DOC: Populate a database, corrupt it then observe Zeek's behavior deleting the database and reopening it.
+# @TEST-REQUIRES: dd --version > dd-version.txt
+# @TEST-REQUIRES: test -e /dev/zero
+# @TEST-EXEC: zeek -b %INPUT >> out
+
+# Evil
```
ckreibich, line 6: :-)

testing/btest/broker/store/brokerstore-backend-sqlite-corrupt-delete.zeek
```diff
@@ -0,0 +1,42 @@
+# @TEST-DOC: Populate a database, corrupt it then observe Zeek's behavior deleting the database and reopening it.
+# @TEST-REQUIRES: dd --version > dd-version.txt
```
ckreibich, line 2: Just curious here whether we need to keep the version. Fine with me if you feel it's useful.

scripts/base/frameworks/broker/store.zeek
```diff
@@ -57,12 +57,62 @@ export {
 		SQLITE,
 	};
 
+	# Behavior when the backend is found to be corrupt or
+	# failing and there's a possibility to get back to
+	# working by wiping/deleting state. For example, a
+	# corrupted SQLite database can be deleted.
+	type FailureMode: enum {
```
ckreibich, line 64: Since Zeek's enums lose their type name when referring them outside of the module (e.g. `Broker::FAIL`) I'm wondering whether these should be `SQLITE_FAIL` and `SQLITE_DELETE`? There's already `SUCCESS` and `FAILURE` via `QueryStatus`, which is a bit close in naming. These are specific to SQLite, right?

testing/btest/Baseline/broker.store.brokerstore-backend-sqlite-corrupt-delete/out
```diff
@@ -0,0 +1,5 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+store is open
+populated 100 rows
+store is open
+populated 100 rows
```
ckreibich, line 5: Do you think it's worth verifying in the test that these post-delete writes actually made it into the database?

ckreibich (changes requested):
This is great! Very nice tests, too. Just one request/question re. enum naming.

scripts/base/frameworks/broker/store.zeek
awelzel: Yes, I was in the same corner. Loosing the prefix is quite unfortunate. It doesn't need to be completely specific to SQLite, but today it is, so I'm happy to add the prefix, but will go full `SQLITE_FAILURE_MODE_FAIL`, because `DELETE` and `FAIL` are just too generic short.

Thanks!

testing/btest/Baseline/broker.store.brokerstore-backend-sqlite-corrupt-delete/out
awelzel: Sure.

ckreibich (approved):

timwoj:
I merged another broker PR that was sitting ahead of this one, so you'll need to fix that conflict in the submodule.

ckreibich:
Docs in zeek/docs#171.

PR #2708, opened 2023-01-23
analyzer: Move disabling_analyzer() hook into Analyzer module

awelzel:
When disabling_analyzer() was introduced, it was added to the GLOBAL module. The awkward side-effect is that implementing a hook handler in another module requires to prefix it with GLOBAL. Alternatively, one can re-open the GLOBAL module and implement the handler in that scope.

Both are not great, and prefixing with GLOBAL is ugly, so move the identifier to the Analyzer module and ask users to prefix with Analyzer.


awelzel:
@ckreibich / @rsmmr - is there a general guideline that hooks just should't live in `GLOBAL`? Or could we deduce one from this, assuming the sentiment of the change is reasonable.

rsmmr:
That sounds like a bug to me, one shouldn't never be required to use `GLOBAL`, that's really just an internal artifact. Lookups without scope should look there automatically.

Generally, I don't think there's a policy one way or the other. Generally having a module scope is preferable for hooks I'd say, but it should be a semantic decision, not a technical.

awelzel:
> That sounds like a bug to me, one shouldn't never be required to use `GLOBAL`, that's really just an internal artifact. Lookups without scope should look there automatically.

Hmm, I seem to remember that being discussed elsewhere. I'll open an issue, a hello-world example indicates it doesn't work.

> Generally, I don't think there's a policy one way or the other. Generally having a module scope is preferable for hooks I'd say, but it should be a semantic decision, not a technical.

Hmm, I'll keep this one open for now, I like the new one style a bit better, even if the motivation is a bug..

awelzel:
> Hmm, I'll keep this one open for now, I like the new one style a bit better, even if the motivation is a bug..

Tracking in #2709 - suggest we still move this one forward if solely for pragmatism. I don't think there's any other global hook in the API currently fwiw.

awelzel:
>> Generally, I don't think there's a policy one way or the other. Generally having a module scope is preferable for hooks I'd say, but it should be a semantic decision, not a technical.

> Hmm, I'll keep this one open for now, I like the new one style a bit better, even if the motivation is a bug..

@rsmmr / @ckreibich  - would you be up for approving still? AFAICT, we currently do not have "global hooks" as a script-level interface for users (compare with `Log::log_stream_policy`, `Notice::policy`) and this would align to what we have (and steer around the namespacing issue).

rsmmr (approved):
Looks reasonable to me.

PR #1661, opened 2021-07-08
Support for doctest in plugins

ckreibich:
I put this together in the end — let's see how it fares in CI. No functional change here over what I mentioned in #1477. It just adds a testcase.

PR #2710, opened 2023-01-23
Modernize and streamline CMake scaffolding

Neverlord:
This PR redesigns the CMake scaffolding to move away from modifying global CMake state like `CMAKE_CXX_FLAGS` that has caused some headache in the past with integrating other CMake-based projects to a cleaner (and "modern") target-based setup. All still compatible with CMake 3.15 (tested with CMake 3.15.0 on macOS).

Flags that should affect all parts of Zeek are now bundled in `zeek_internal` to propagate compiler flags, defines, etc. to all builtin parts of Zeek.

All things Zeek eventually feed into the two new top-level targets `zeek_exe` (standalone binary) and/or `zeek_lib` (when building Zeek as static lib).

Plugins can now be defined in a single function call instead of calling a sequence of functions all polluting the parent scope. The legacy functions still exist for backwards-compatibility, though.

Just to pick one example, currently a plugin is defined like this:

```cmake
include(ZeekPlugin)

include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})

zeek_plugin_begin(Zeek DNP3)
zeek_plugin_cc(DNP3.cc Plugin.cc)
zeek_plugin_bif(events.bif)
zeek_plugin_pac(dnp3.pac dnp3-analyzer.pac dnp3-protocol.pac dnp3-objects.pac)
zeek_plugin_end()
```

The "new way" is a single call that defines all things declaratively:

```cmake
zeek_add_plugin(
    Zeek
    DNP3
    SOURCES
        DNP3.cc
        Plugin.cc
    BIFS
        events.bif
    PAC
        dnp3.pac
        dnp3-analyzer.pac
        dnp3-protocol.pac
        dnp3-objects.pac
)
```

The `PAC` block can be repeated, whereas each block defines one parser (some plugins have multiple parsers).

So far, I've got rid of function calls that change global state such as `add_compile_options`. Instead, flags are propagated through the new `zeek_internal` target and there are a couple of convenience functions to add internal flags/defines/includes as a "drop-in replacement" for the global CMake functions.

There are still some TODOs left (like generating a proper CMake package for Zeek), but I think this is a good stage to get in feedback.


CMakeLists.txt
```diff
@@ -9,12 +10,177 @@ endif()
...
+    endif ()
+    if ( FN_ARGS_INTERNAL_DEPENDENCIES )
+         add_dependencies(${target_name} ${FN_ARGS_INTERNAL_DEPENDENCIES})
+    endif ()
+
+    # Feed into the main Zeek target(s).
+    zeek_add_obj_lib(${target_name})
+endfunction()
+
+# Convenience macros to safe ourselves some typing.
```
timwoj, line 177:
```suggestion
# Convenience macros to save ourselves some typing.
```

CMakeLists.txt
```diff
@@ -886,11 +1067,11 @@ if ( NOT DISABLE_SPICY )
         # seems this does not work reliable (due incomplete support for object
         # libraries in CMake?).
         list(TRANSFORM _spicy_rt_libs APPEND "-objects")
-        target_link_libraries(zeek ${_spicy_rt_libs})
+        target_link_libraries(${spicyZeekTarget} PUBLIC ${_spicy_rt_libs})
 
         # Make sure we build targets of spicy-plugin after the `spicy` target.
-        add_dependencies(plugin-Zeek-Spicy spicy)
-        add_dependencies(spicyz spicy)
+        #add_dependencies(plugin-Zeek-Spicy spicy)
```
timwoj, line 1073: Are these needed anymore?

src/analyzer/protocol/rdp/CMakeLists.txt
```diff
@@ -1,11 +1,20 @@
...
-	zeek_plugin_bif(events.bif)
-	zeek_plugin_bif(types.bif)
-	zeek_plugin_pac(rdp.pac rdp-analyzer.pac rdp-protocol.pac ../asn1/asn1.pac)
-	zeek_plugin_pac(rdpeudp.pac rdpeudp-analyzer.pac rdpeudp-protocol.pac)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
+    RDP
+    SOURCES
+	      RDPEUDP.cc
```
timwoj, line 5: Indentation weirdness

src/broker/CMakeLists.txt
```diff
@@ -1,20 +1,14 @@
-include(ZeekSubdir)
-
-include_directories(BEFORE
-                    ${CMAKE_CURRENT_SOURCE_DIR}
-                    ${CMAKE_CURRENT_BINARY_DIR}
+zeek_add_subdir_library(
+    comm
```
timwoj, line 2: Should we rename this library to `broker` at this point?

src/file_analysis/CMakeLists.txt
```diff
@@ -1,23 +1,17 @@
-include(ZeekSubdir)
-
-include_directories(BEFORE
-                    ${CMAKE_CURRENT_SOURCE_DIR}
-                    ${CMAKE_CURRENT_BINARY_DIR}
+zeek_add_subdir_library(
+    file_analysis
+    INTERNAL_DEPENDENCIES ${BIF_BUILD_TARGET}
```
timwoj, line 3: This library depends on `BIF_BUILD_TARGET` but the broker one depended on `bif_deps`. What's the difference there?

src/packet_analysis/protocol/linux_sll/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
-
-include(ZeekPlugin)
-
-include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-
-zeek_plugin_begin(PacketAnalyzer LinuxSLL)
-zeek_plugin_cc(LinuxSLL.cc Plugin.cc)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: Should this be `PacketAnalyzer` instead of `Zeek`?

src/packet_analysis/protocol/ppp_serial/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
-
-include(ZeekPlugin)
-
-include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-
-zeek_plugin_begin(PacketAnalyzer PPPSerial)
-zeek_plugin_cc(PPPSerial.cc Plugin.cc)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: Should this be `PacketAnalyzer` instead of `Zeek`?

src/packet_analysis/protocol/pppoe/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
-
-include(ZeekPlugin)
-
-include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-
-zeek_plugin_begin(PacketAnalyzer PPPoE)
-zeek_plugin_cc(PPPoE.cc Plugin.cc)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: Should this be `PacketAnalyzer` instead of `Zeek`?

src/packet_analysis/protocol/teredo/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
-include(ZeekPlugin)
-
-zeek_plugin_begin(Zeek Teredo)
-zeek_plugin_cc(Teredo.cc Plugin.cc)
-zeek_plugin_bif(events.bif)
-zeek_plugin_bif(functions.bif)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: This is `Zeek` in the existing code, but probably should be made `PacketAnalyzer` like the others.

src/packet_analysis/protocol/udp/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
...
-include(ZeekPlugin)
-
-include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-
-zeek_plugin_begin(Zeek UDP)
-zeek_plugin_cc(UDP.cc UDPSessionAdapter.cc Plugin.cc)
-zeek_plugin_bif(events.bif)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: I think this one has to stay as `Zeek` or other things break, but we should probably test that. Same with the TCP one as well.

src/packet_analysis/protocol/wrapper/CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
-
-include(ZeekPlugin)
-
-include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
-
-zeek_plugin_begin(PacketAnalyzer Wrapper)
-zeek_plugin_cc(Wrapper.cc Plugin.cc)
-zeek_plugin_end()
+zeek_add_plugin(
+    Zeek
```
timwoj, line 2: Should this be `PacketAnalyzer` instead of `Zeek`?

timwoj (changes requested):
I made a first pass at it. In general, I like it a lot. I had some minor formatting comments, and some questions about the structure/location of some of the added code.

src/CMakeLists.txt
```diff
@@ -525,34 +547,45 @@ set(zeek_SRCS
...
+        zeek_exe
+        PRIVATE
+            main.cc
+            ${zeek_HEADERS}
+            ${bro_SUBDIR_LIBS}
+            ${bro_PLUGIN_LIBS})
+    target_link_libraries(
+        zeek_exe
+        PRIVATE
+            ${bro_PLUGIN_LINK_LIBS}
```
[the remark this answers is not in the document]
timwoj: Maybe something like this https://github.com/robotology/yarp/blob/ea6ed180ddb102679d8b8492fc40329719d7b8e4/cmake/YarpRenamedOption.cmake ?

CMakeLists.txt
```diff
@@ -9,12 +10,177 @@ endif()
...
+# Usage:
+# zeek_add_subdir_library(
+#   <name>
+#   SOURCES ...
+#   [INCLUDE_DIRS ...]
+#   [DEPENDENCIES ...]
+#   [INTERNAL_DEPENDENCIES ...]
+#   [BIFS ...]
+# )
+function(zeek_add_subdir_library name)
```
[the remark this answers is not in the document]
Neverlord: Certainly a matter of preference. I would keep them there, because they modify the target defined above and moving the functions someplace else IMHO makes it more difficult to follow the CMake source.

CMakeLists.txt
Neverlord: Probably not. Once I got it all working again, I'll make a clean-up pass.

src/CMakeLists.txt
```diff
@@ -497,15 +524,10 @@ elseif(${COMPILER_ARCHITECTURE} STREQUAL "x86_64")
...
                               ${_avx_flag})
-  set_source_files_properties(../auxil/highwayhash/highwayhash/hh_sse41.cc PROPERTIES COMPILE_FLAGS
+  set_source_files_properties(${hhash_dir}/hh_sse41.cc PROPERTIES COMPILE_FLAGS
                               ${_sse_flag})
-
-  list(APPEND HH_SRCS
-      ../auxil/highwayhash/highwayhash/hh_avx2.cc
-      ../auxil/highwayhash/highwayhash/hh_sse41.cc
-  )
 endif ()
```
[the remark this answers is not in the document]
Neverlord: Since `highwayhash` is a fork of a 3rd-party repository that doesn't use CMake, I think it's fine doing it that way. That being said, maybe they would be open to a contribution for CMake in their repository as a better mid-term solution?

src/CMakeLists.txt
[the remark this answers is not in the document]
Neverlord: I'll make a pass to see what `bro` variables we can safely get rid off. Some of the variables are used by the dynamic plugins, so changing them would break existing modules.

src/analyzer/protocol/ssl/CMakeLists.txt
```diff
@@ -1,20 +1,35 @@
...
+    SOURCES
+        SSL.cc
+        DTLS.cc
+        Plugin.cc
+    BIFS
+        types.bif
+        events.bif
+        functions.bif
+        consts.bif
+    PAC
```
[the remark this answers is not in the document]
Neverlord: Each `PAC` block is one parser. This parser just happens to have a ton of files. I've kept the comments that explain what each set of files does in the resulting parser.

The documentation for this function is in `ZeekPlugin.cmake`:

```
# Usage:
# zeek_add_plugin(
#   <namespace>
#   <name>
#   [INCLUDE_DIRS ...]
#   [DEPENDENCIES ...]
#   [DIST_FILES ...]
#   [SOURCES ...]
#   [BIFS ...]
#   [[PAC ...] ... ]
# )
# * INCLUDE_DIRS:
#   Adds additional include directories for building the plugin. By default, the
#   function adds `CMAKE_CURRENT_BINARY_DIR` as additional include directory.
# * DEPENDENCIES:
#   Adds additional CMake targets as extra dependencies.
# * DIST_FILES:
#   Adds additional files to install alongside the plugin when building a
#   dynamic plugin. Ignored when building static plugins.
# * SOURCES:
#   List of C++ files for compiling the plugin.
# * BIFS:
#   List of BIF files (*.bif) for compiling the plugin.
# * PAC:
#   Adds a BinPAC parser to the plugin. The function accepts multiple `PAC`
#   blocks. Each block defines a single BinPAC parser.
```

src/broker/CMakeLists.txt
Neverlord: Will do.

src/file_analysis/CMakeLists.txt
Neverlord: Thanks for pointing those out. Seems like they've been a leftover and I just didn't clean them up.

src/packet_analysis/protocol/linux_sll/CMakeLists.txt
Neverlord: Yep. Fixed.

src/packet_analysis/protocol/pppoe/CMakeLists.txt
Neverlord: Fixed.

src/packet_analysis/protocol/ppp_serial/CMakeLists.txt
Neverlord: Fixed.

src/packet_analysis/protocol/teredo/CMakeLists.txt
Neverlord: Done.

Neverlord: Actually, things break when moving this to `Zeek`. It's in the Zeek namespace on current master: https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/teredo/CMakeLists.txt#L3.

There's also some weirdness with some BIFs being baked directly into Zeek. The Broker BIFs for example have to use the "alternative mode" to get built directly into Zeek or else the `btests` complain. Maybe the tests are just overly strict? If we do clean that up, I'd prefer doing that as a followup, though, to focus this PR on modernizing the CMake without changing the behavior (wherever possible).

timwoj: Yeah, this is weird. For some reason it doesn't like the definitions in functions.bif in both the gtpv1 and teredo analyzers if you move them to the other namespace.

timwoj: Ah, I had to nuke build/scripts/base/bif/plugins/Zeek_Teredo.functions.bif.zeek and then fix scripts/base/packet-protocols/teredo/main.zeek to point at the right file. I agree this is out of scope for these changes. I'm fine with making a note somewhere and coming back to them later.

Neverlord:
@timwoj I've extracted a few TODOs from your comments for later (either later commits or separate tickets):

- [ ] Rename `comm` library to `broker`
- [ ] Make a pass to see whether more `bro` variables can be safely renamed / dropped
- [ ] Discuss whether we can (and want to) move plugins to better fitting namespaces (UDP/TCP/teredo/wrapper are currently all in `Zeek`)

Aside from that, I currently have a few btests failing locally. One is `doctest-disabled.zeek`. This test checks that even if Zeek was compiled with `doctest` enabled, Plugins can disable `doctest` independently. The flag for `doctest` is propagated through the new `internal` target. I'd really like to keep it that way, since it's the cleanest way and I think that this test ultimately checks for something very, very niche. So unless someone has a strong opinion on allowing plugins to have a different `doctest` setting than Zeek, I'd go ahead fix the btest by removing it. 🙂

The other failing `btest` instances are from `spicy`. Zeek crashes with a segfault at runtime with a very unhelpful stacktrace pointing so some fiber. Hopefully @rsmmr or @bbannier can tell me what I'm doing wrong when linking to Spicy / creating target dependencies.

Other than that, I think this is nearing completion now. There are some small improvements we could still make later, but overall this has shifted to a target-based setup now. I've also got rid of the very complex setup under `bifcl.cmake` that played ping-pong with `src/CMakeLists.txt` over some cache variables. Instead of doing that, we now create all files that we generate for BIFs in `CMakeLists.txt` once and then `bifcl.cmake` simply appends to those files as we go along. This makes it much more obvious (and straightforward) what's happening.

I've also ripped out that weird logic where Plugins would read back the CMake cache from main Zeek to re-construct where it can find things. Instead of that, Zeek is writing a CMake package file now. For the plugins, it's also writing a `ZeekPluginConfig.cmake` file that imports targets for `BifCl` and `BinPAC`. This file only exists in the build directory for the plugins and won't be installed. This setup is quick & dirty at the moment and has a TODO in it to support multi-configuration CMake generators. So right now it might not work on Windows because of that.

Speaking of package file: Zeek has a CMake package file now. 😉

We still need to triple-check that it's still putting all things into the "historically expected location" to not break existing tooling for bundling Zeek. Once I have everything else running, I'll check this on my end, but that's just one OS we support.

Neverlord:
And I've rebased this on the weekend, but it's already running into a conflict again. Sigh. Rebasing again soon.

src/analyzer/protocol/ssl/CMakeLists.txt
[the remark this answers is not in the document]
timwoj: Ah, right on. I kinda like that you can break it up like that actually.

timwoj:
I added the `fullci` tag to this so that Cirrus builds all of the platforms. It should work on your next push.

ckreibich:
Dominik a bit of context on the doctest situation after digging up the old ticket in zeek/zeek#1661: the rationale back then was simply that in some contexts doctests in a plugin will constitute bloat, and so it's nice to be able to exclude them. I don't see any requirements that made the ability to disable doctests in a plugin a must.

So if we simply adopt Zeek's doctests status for plugin builds that should be just fine, and what you're hoping for, right? That is, if Zeek has doctest support enabled, plugins should also build with it, and if Zeek does not (configure with `--skip-cpp-tests`) then the plugin builds should also skip them.

awelzel:
> The other failing btest instances are from spicy. Zeek crashes with a segfault at runtime with a very unhelpful stacktrace pointing so some fiber. Hopefully @rsmmr or @bbannier can tell me what I'm doing wrong when linking to Spicy / creating target dependencies.

Linux, GCC and ASAN enabled? Maybe it's the following one?

https://github.com/zeek/spicy/issues/1310

It isn't in zeek/master yet, but :crossed_fingers: soon.

bbannier:
> > The other failing btest instances are from spicy. Zeek crashes with a segfault at runtime with a very unhelpful stacktrace pointing so some fiber. Hopefully @rsmmr or @bbannier can tell me what I'm doing wrong when linking to Spicy / creating target dependencies.
> 
> Linux, GCC and ASAN enabled? Maybe it's the following one?
> 
> [zeek/spicy#1310](https://github.com/zeek/spicy/issues/1310)

I talked with Dominik this week, and I think this was something else, related to inconsistent linking which Dominik was able to pin down. Not he has a fix yet.

Neverlord:
> So if we simply adopt Zeek's doctests status for plugin builds that should be just fine, and what you're hoping for, right? That is, if Zeek has doctest support enabled, plugins should also build with it, and if Zeek does not (configure with --skip-cpp-tests) then the plugin builds should also skip them.

@ckreibich cool, thanks! When building with Zeek, we propagate the doctest flags implicitly to the plugins now. Btw, Plugins can still opt to enable doctest even if Zeek has been compiled without unit tests. Adding a flag is always easier than taking flags away. 🙂

@awelzel just a plain `debug` build. The issue was mixing debug and release symbols. That's all been sorted out after @bbannier helped me understand the Spicy setup a bit more.

Seems like I got it working locally, so I'll test it on Linux next and then check whether we still install the same things into the same locations.

Neverlord:
Closing here. I've pulled out a large chunk of the changes here for #2799. I'll probably try re-integrating the target-based flag management from here once the other PR is merged, but I think making a new PR instead of re-purposing this one is cleaner.

PR #2712, opened 2023-01-23
Add test for new handling of unknown RDP keyboards

timwoj:
This adds a test for the changes merged in from #2701. The pcap here is a modified version of `$TRACES/rdp/rdp-x509.pcap` with a couple packets dropped and the keyboard_layout value modified in the client data packet. The TCP checksum is wrong, but we can just use `-C` to read it.


awelzel (approved):
Thanks.

PR #2715, opened 2023-01-24
MySQL: Fix endianness, introduce mysql_eof() event

awelzel:
We were parsing MySQL using bigendian even though the protocol is specified as with "least significant byte first" [1]. This is most problematic when parsing length encoded strings with 2 byte length fields...

Further, I *think*, the EOF_Packet parsing was borked, either due to testing the CLIENT_DEPRECATE_EOF with the wrong endianness, or due to the workaround in Resultset processing raising mysql_ok(). Introduce a new mysql_eof() that triggers for EOF_Packet's and remove the fake mysql_ok() Resultset invocation to fix. Adapt the mysql script and tests to account for the new event.

This is a quite backwards incompatible change on the event level, but due to being quite buggy in general, doubt this matters to many.

I think there is more buried, but this fixes the violation of the simple "SHOW ENGINE INNODB STATUS" and the existing tests continue to succeed...

[1] https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_integers.html


timwoj (approved):
I double-checked against what Wireshark was saying about the pcap used in the `scripts.base.protocols.mysql.wireshark` btest, and the output looks like it matches with regards to the EOFs. I had one minor spelling comment.

awelzel:
Thanks @timwoj - I think I'll add the other plaintext test from #2642 and if that looks good merge it and then :crossed_fingers:

PR #2718, opened 2023-01-24
GH-2684: Stop violating VXLAN for forwarding failures

### LINKED ISSUE #2684: vxlan encapsulated igmp v2 packets triggering analyzer violations
When encapsulating local traffic using VXLAN and forwarding to Zeek, ran into the VXLAN analyzer raising an AnalyzerViolation when the inner protocol is IGMP.

It doesn't seem we support IGMP and that in turn is causing the VXLAN analyzer to report "invalid inner packet" violations as the inner packet can not be parsed. Not because it's invalid, but because we don't know how to.

Here's a pcap:
[vxlan-encapsulated-igmp-v2.pcap.zip](https://github.com/zeek/zeek/files/10403703/vxlan-encapsulated-igmp-v2.pcap.zip)

This also causes the vxlan part in in conn.log.service to be removed, which is a bit unfortunate. The  inner IPs do not show as connections at all. Wireshark (and plain tcpdump) are doing quite a bit nicer here.

```
conn.log:
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       proto   service duration        orig_bytes      resp_bytes      conn_state      local_orig      local_resp      missed_bytes history orig_pkts       orig_ip_bytes   resp_pkts       resp_ip_bytes   tunnel_parents
#types  time    string  addr    port    addr    port    enum    string  interval        count   count   string  bool    bool    count   string  count   count   count   count   set[string]
1673538029.809899       CV0d964jPH8Y7h2UFl      172.30.0.1      48036   172.30.0.2      4789    udp     -       -       -       -       S0      -       -       0       D       1       82      0       0   -
1673538054.797831       CYhATE26GDTK2AlgXc      172.30.0.1      45303   172.30.0.2      4789    udp     -       -       -       -       S0      -       -       0       D       1       82      0       0   -
1673538167.375490       Cmdvdk1rM4jFTl1VE7      172.30.0.1      36030   172.30.0.2      4789    udp     -       -       -       -       S0      -       -       0       D       1       82      0       0   -
#close  2023-01-12-17-24-38


$ tcpdump -v -r ./vxlan-encapsulated-igmp-v2.pcap
reading from file ./vxlan-encapsulated-igmp-v2.pcap, link-type EN10MB (Ethernet), snapshot length 262144
16:40:29.809899 IP (tos 0x0, ttl 64, id 47941, offset 0, flags [DF], proto UDP (17), length 82)
    tinkyx390.48036 > 172.30.0.2.4789: VXLAN, flags [I] (0x08), vni 4242
IP (tos 0x0, ttl 1, id 62566, offset 0, flags [none], proto IGMP (2), length 32, options (RA))
    192.168.9.80 > 224.0.0.251: igmp v2 report 224.0.0.251
16:40:54.797831 IP (tos 0x0, ttl 64, id 52732, offset 0, flags [DF], proto UDP (17), length 82)
    tinkyx390.45303 > 172.30.0.2.4789: VXLAN, flags [I] (0x08), vni 4242
IP (tos 0x0, ttl 1, id 52471, offset 0, flags [none], proto IGMP (2), length 32, options (RA))
    192.168.9.123 > 224.0.0.251: igmp v2 report 224.0.0.251
16:42:47.375490 IP (tos 0x0, ttl 64, id 5749, offset 0, flags [DF], proto UDP (17), length 82)
    tinkyx390.36030 > 172.30.0.2.4789: VXLAN, flags [I] (0x08), vni 4242
IP (tos 0x0, ttl 1, id 44659, offset 0, flags [none], proto IGMP (2), length 32, options (RA))
    192.168.9.80 > 224.0.0.251: igmp v2 report 224.0.0.251
```
tcpdump/wireshark show the inner 192.168.9.80 / 224.0.0.251.

Part of the answer is probably "Add IGMP" support (https://github.com/zeek-plugins/igmp), but maybe we should squelch the violations somehow as there's nothing wrong with the VXLAN part.

timwoj:
Failure to forward out of the VXLAN analyzer shouldn't be considered a violation of that analyzer. This removes that call.

Fixes #2684


awelzel (approved):
Ah, that looks easy enough :-)

I'm approving, but the pcap from the ticket could be added as a test for conn.log / analyzer.log checking.

timwoj:
One thing that I noticed while adding the test is that the code doesn't call `AnalyzerConfirmation` unless it's able to forward the packet. That means conn.log doesn't mark the packet as VXLAN in this case. I checked the other tunnel analyzers, and it goes back and forth. Both Teredo and AYIYA confirm it before calling `ForwardPacket` once they have enough information. Geneve and VXLAN call it only after a successful call to `ForwardPacket`. GRE doesn't call it at all.

PR #2719, opened 2023-01-25
cirrus: Introduce SKIP_TASK_ON_PR

awelzel:
Limit the number of tasks we run on PRs to only a subset instead of
the full set to reduce the amount of resources we use for PRs. Labeling
a PR with "fullci" can be used enable running all tasks.

This removes a total of 14 task execution from PRs by default at the cost of
needing to monitor master builds more closely for distro specific breakage.

This could probably be even be more advanced:

Leveraging changesIncludeOnly('scripts/*', 'testing/*'), we might be able to
limit to running a single Linux distro rather than a whole set of different
for simple script changes. For now I've left that out, however.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -66,6 +66,12 @@ branch_whitelist: &BRANCH_WHITELIST
       )
     ) )
 
+skip_task_on_pr: &SKIP_TASK_ON_PR
+  # Skip this task on PRs if it does not have the fullci label,
+  # it continues to run for direct pushes to master/release.
+  skip: >
+    ($CIRRUS_PR != '' && $CIRRUS_PR_LABELS !=~ '.*fullci.*')
+
 ci_template: &CI_TEMPLATE
   << : *BRANCH_WHITELIST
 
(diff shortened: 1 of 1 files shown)


ckreibich:
I like this. My personal preference would be to spell out that one wants full CI in the PR comment rather than a Github label.

awelzel:
> I like this. My personal preference would be to spell out that one wants full CI in the PR comment rather than a Github label.

TLDR: I can ping Cirrus CI if they could make the PR description available reliably or I'm missing something, but for the time being labels seem fairly pragmatic.

There are the following env variables available that can be used in `only_if` / `skip` conditions:
```
CIRRUS_CHANGE_MESSAGE	Commit message or PR title and description, depending on trigger event (Non-PRs or PRs respectively).
CIRRUS_CHANGE_TITLE	First line of CIRRUS_CHANGE_MESSAGE
```
Testing in an external repo, when creating the PR, these contain the PR title and description that allow for that. However, for subsequent pushes to the PR's branch,  they are then populated with the commit title / message instead (making an opt-in `[fullci]` tag in the PR description not functional. You'd put that into each commit. I don't think we can query the GitHub API for evaluating `skip:` or `only_if:` conditions - that seems to happen very early. We could do that within the job and early exit, but that would schedule and allocate resources just to then decide to early exist.

It's visually also nicely standing out in the list of pending PRs. So you have some early indication which PRs use that (or which don't when you think they should), without needing to dig into the description whether and where a `[fullci]` is somewhere to be found.


For the command-line, can do `gh pr create -l fullci` or  `gh pr edit --add-label fullci`.

Any big gripe with the labels otherwise? I'm open for better suggestions to the `fullci` one, but was already thinking `winci` could be interesting, too.

ckreibich:
> Any big gripe with the labels otherwise?

Nope ... thanks for the explanation. Another benefit of labels is that the PR history nicely shows their addition/removal if one changes them, which wouldn't be the case for PR descriptions. So yeah, count me in.

ckreibich (approved):

Issue #2669, opened 2023-01-03
Error messages for type clashes in iteration loops lack line numbers


vpax:
For the following input:
```
global a: table[count] of count;
global b: table[int] of int;

event zeek_init()
        {
        for ( i, x in a ) print i;
        for ( i, x in b ) print i;
        }
```
Zeek correctly generates errors about the type clashes, but they lack line numbers:
```
error in count and int: type clash in iteration (count and int)
error in count and int: type clash in iteration (count and int)
```
making them difficult to find in large scripts.

timwoj:
Looks like the type object generated for the loop var here has no location data in it:

https://github.com/zeek/zeek/blob/master/src/Stmt.cc#L1243-L1253

Just passing down `lv`'s location info into `Error` doesn't help because `obj2` also has a location value in it that's not useful. This is probably going to have to be fixed down in the parser to set proper locations on these types.

timwoj:
Staring at this a bit more, shouldn't `i` and `x` be out of scope for the second for-loop, thus negating the error message altogether?

vpax:
> Staring at this a bit more, shouldn't `i` and `x` be out of scope for the second for-loop, thus negating the error message altogether?

Perhaps, but the same problem occurs without the reuse, e.g.:
```
global a: table[count] of count;

event zeek_init()
        {
        local d: double;
        for ( d in a )
                print "inside the loop";
        }
```

timwoj:
Yeah, I dug a bit further into it and it gets mired in how we build that error message in `Obj::Error`. `ind_type` there is the global `int` type. It has a location attached to it, but it's all zeroes other than the name. Cloning the type and modifying the location causes output that doesn't look any better (it replaces the first `int` with the location info). Basically we're tricking `Obj::DoMsg` into using the filename as the type name in the output, which isn't great.

vpax:
FYI, the following makes the messages less-sucky and basically ok IMHO:

```
diff --git a/src/Stmt.cc b/src/Stmt.cc
index 0d5306cf9..e3448272b 100644
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -1249,7 +1249,7 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr)
 			else if ( lvt )
 				{
 				if ( ! same_type(lvt, ind_type) )
-					lvt->Error("type clash in iteration", ind_type.get());
+					e->Error("type clash in iteration", lvt.get());
[...]

PR #2721, opened 2023-01-26
GH-2669: Give script location for type clash in for-loop construction

timwoj:
This changes the error reported by the code in #2669 to the following error messages, which are a bit more descriptive and include a location:

```
error in /Users/tim/Desktop/2669.zeek, line 7 and count: type clash in iteration (b and count)
error in /Users/tim/Desktop/2669.zeek, line 7 and count: type clash in iteration (b and count)
```


awelzel (approved):
Approving.

The test update seems a bit of a side-effect where we report more error after deciding that the loopvar is already not usable due to being global.

PR #2722, opened 2023-01-26
Initial set of static analysis fixes from PVS-Studio

timwoj:
I recently got a license for [PVS-Studio](https://pvs-studio.com/en/pvs-studio/?utm_source=github&utm_medium=organic&utm_campaign=open_source) for the Zeek project. This PR is the first of a set of changes fixing the findings from their analyzer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,3 +109,14 @@ License
 
 Zeek comes with a BSD license, allowing for free use with virtually no
 restrictions. You can find it [here](https://github.com/zeek/zeek/blob/master/COPYING).
+
+
+Tooling
+-------
+
+We us the following tooling to help discover issues to fix, amongst a number of
+others.
+
+- [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/)
+- [Coverity](https://scan.coverity.com/projects/bro)
+- [PVS-Studio](https://pvs-studio.com/en/pvs-studio/?utm_source=github&utm_medium=organic&utm_campaign=open_source) - static analyzer for C, C++, C#, and Java code.
--- a/src/Dict.h
+++ b/src/Dict.h
@@ -645,7 +645,7 @@ public:
 			auto loc = detail::GetCurrentLocation();
 			reporter->RuntimeError(&loc,
 			                       "Attempted to create DictEntry with excessively large key, "
-			                       "truncating key (%" PRIu64 " > %d)",
+			                       "truncating key (%" PRIu64 " > %u)",
 			                       key_size, detail::DictEntry<T>::MAX_KEY_SIZE);
 			}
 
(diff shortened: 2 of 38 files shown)


README.md
```diff
@@ -109,3 +109,13 @@ License
 
 Zeek comes with a BSD license, allowing for free use with virtually no
 restrictions. You can find it [here](https://github.com/zeek/zeek/blob/master/COPYING).
+
+
+Tooling
+-------
+
+Zeek uses the following tooling to help discover issues to fix, amongst a number
```
timwoj, line 117: I could see adding more to this section (clang-tidy? anything in pre-commit?).

src/Dict.h
```diff
@@ -417,9 +417,43 @@ template <typename T> class RobustDictIterator
...
+
+	RobustDictIterator& operator=(const RobustDictIterator& other)
+		{
+		if ( this == &other )
+			return *this;
+
+		delete inserted;
+		delete visited;
 		dict = nullptr;
+		next = -1;
```
ckreibich, line 456: Should this do something with `curr` also if there's no `other.dict`?

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -436,7 +436,7 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig)
 				// Remove nick name.
 				parts.erase(parts.begin());
 
-				if ( parts.size() > 0 && parts[0][0] == ':' )
+				if ( parts[0][0] == ':' )
```
ckreibich, line 439: I'm not immediately following why this is safe to do — perhaps context makes it a given.

src/plugin/Manager.cc
```diff
@@ -630,7 +630,7 @@ Plugin* Manager::LookupPluginByPath(std::string_view _path)
 static bool hook_cmp(std::pair<int, Plugin*> a, std::pair<int, Plugin*> b)
 	{
 	if ( a.first == b.first )
-		return util::strtolower(a.second->Name()) < util::strtolower(a.second->Name());
+		return util::strtolower(a.second->Name()) < util::strtolower(b.second->Name());
```
ckreibich, line 633: Ha!

src/analyzer/protocol/irc/IRC.cc
timwoj: We already checked to see if `parts.size()` is < 2, and bailed out. We then removed an element which guarantees that the size is at least `1` here. I removed the check because it doesn't actually do anything,

src/plugin/Manager.cc
timwoj: Right? This one and the shadowing one in `Span` made me chuckle.

ckreibich (approved):

PR #2725, opened 2023-01-26
NEWS updates for 5.2

timwoj:
The meat of this is in the first commit. The second commit is just some formatting fixes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,6 +25,22 @@ Breaking Changes
   adapted accordingly. Users of ``mysql_ok()`` likely need to switch to
   ``mysql_eof()``.
 
+- Zeek will now exit at startup if an external plugin (e.g. from a package) is
+  discovered to have the same name as a built-in plugin. See below for the
+  change regarding the AF_PACKET plugin now being built-in for an example of
+  this potentially being triggered.
+
+- DNS query type strings were updated to match the current standardized list of
+  strings. This changes the string reported for a small subset of query types:
+
+	30: Changed from "EID" to "NXT"
+	31: Changed from "NIMLOC" to "EID"
+	32: Changed from "NB" to "NIMLOC"
+
+- The ``--with-caf`` option for the ``configure`` script was removed. Broker now
+  requires specific versions of CAF per Zeek release, and passing an
+  externally-built version of CAF often lead to build failures.
+
 New Functionality
 -----------------
 
@@ -35,6 +51,10 @@ New Functionality
   builds done with Visual Studio 2019. Information on how to build on Windows is
   available in the Zeek documentation. Note also that Spicy is currently
   unsupported and will be fixed in the future.
+
+  The feature as checked into the repository is not considered production-ready.
+  There are many bugs to squash and features to improve, and we will be steadily
+  fixing things over the next few months.
 
   The Zeek team wants to give a huge thank you to the team at Microsoft for all
   of their effort in completing this port.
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -107,84 +127,114 @@ New Functionality
...
   environments.
 
+- A new ``join_string_set`` BIF was added, replacing the existing script-level
+  version from utils/strings.zeek.
+
+- A new ``&ordered`` attribute for tables and sets was added. This attribute
+  causes iteration over a table/set to return elements in the order of their
+  insertion.
+
+- A new ``-D`` argument was added to the ``configure`` script to allow passing
```
awelzel, line 193: Is this Zeek NEWS worthy?

NEWS
```diff
@@ -107,84 +127,114 @@ New Functionality
...
+- The ``type_name`` of enum types produced by ``record_fields()`` now includes
+  the actual type name rather than just ``"enum"``.
 
 - Passing non-string ``sep`` and ``def`` arguments to ``cat_sep()`` isn't a
   fatal error anymore. More descriptive error messages are produced, too.
 
 - The number of analyzer violation events that can be raised by protocol
-  analyzer instances is now capped by the const ``max_analyzer_violation_events``.
+  analyzer instances is now capped by the const
+  ``max_analyzer_violation_events``.
```
awelzel, line 230: Uhm, there might have been a conflict resolution hickup on my side. This is the same as below, but without mentioning file analyzers which came as a follow up.

Can you drop this hunk in favor of the one below?

NEWS
```diff
@@ -200,22 +250,35 @@ Changed Functionality
...
-- The MQTT scripts registering the analyzer and DPD signatures have
-  been moved from the policy folder to base and are loaded by default.
+- The MQTT scripts registering the analyzer and DPD signatures have been moved
+  from the policy folder to base and are loaded by default.
 
 - Notices created for files transferred over multiple connections will now be
   associated with one of the connections rather than none.
 
+- Non-fatal errors when setting up BPF filtering will no longer cause Zeek to
+  exit, but instead will log the error in reporter.log and continue processing.
```
awelzel, line 260: There was also something added to `packet_filter.log` ? Maybe I'm misremembering. Maybe not important.

NEWS
```diff
@@ -31,27 +47,31 @@ New Functionality
   available in the Zeek documentation. Note also that Spicy is currently
   unsupported and will be fixed in the future.
 
+  The feature as checked into the repository is not considered production-ready.
```
awelzel, line 50: When reading "the feature" I was wondering about something specific/small, but that's about Windows support.

Maybe: "Windows is not considered a production-ready platform at this point."

awelzel (changes requested):

NEWS
timwoj: I could go either way. It's notable since we removed the option for CAF, and that we're already calling out a number of other additions to that script, but I could also see it being considered a more-technical detail that most users won't care about.

NEWS
timwoj: You're right, there was. It was just that the error propogated from libpcap appears in packet_filter.log instead of something generic. I'll add that in.

NEWS
timwoj: Done

NEWS
awelzel: Up to you then :-)

timwoj:
If no one has any other comments, I'll merge this on Monday.

PR #2732, opened 2023-01-27
Add a few extra length checks to Teredo::DetectProtocol

timwoj:
This PR prevents some integer wrapping when detecting Teredo packets.


awelzel (approved):
:+1:

PR #2737, opened 2023-01-30
Update binpac submodule for modernization fixes

timwoj:
This is the zeek-side counterpart to https://github.com/zeek/binpac/pull/26. That PR should be merged first.


timwoj:
Note that I'm completely fine holding onto this PR until after 5.2 is out. It's making changes to something critical and it should probably have time baking in `master`.

PR #2739, opened 2023-01-31
Fix two btest-diff checks that couldn't fail

ckreibich:
Interview question — are `if A; then B; else C; fi` and `A && B || C` equivalent?

:-)

### DIFF
--- a/testing/btest/Baseline/scripts.base.frameworks.logging.hooks-10/output
+++ b/testing/btest/Baseline/scripts.base.frameworks.logging.hooks-10/output
@@ -5,4 +5,3 @@ foo
 Log::log_stream_policy
 bar
 bar
-Log::log_stream_policy
--- a/testing/btest/scripts/base/frameworks/logging/hooks.zeek
+++ b/testing/btest/scripts/base/frameworks/logging/hooks.zeek
@@ -237,11 +237,11 @@ global output = open("output");
 
 hook Log::log_stream_policy(rec: any, id: Log::ID)
 	{
-	print output, "Log::log_stream_policy";
-
 	if ( id == Test::LOG )
 		{
 		local r: Test::Info = rec;
+
+		print output, "Log::log_stream_policy";
 
 		if ( r$status == "foo" )
 			break;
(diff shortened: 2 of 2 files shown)


0xxon:
Interesting - it fails only for the asan builds :)

ckreibich:
And really strangely too — I'm digging into these.

ckreibich:
Okay, riveting stuff. The output in one of the baselines does indeed deviate in the ASAN build, which is interesting — it's a reporter log entry not sneaking into the codepath where it otherwise does. I'm not going to chase that further since this is sloppiness in the test that should be suppressed. That explains this part:
```
  --- /tmp/test-diff.140187.output.baseline.tmp	2023-01-31 02:55:44.953997940 +0000
  +++ /tmp/test-diff.140187.output.tmp	2023-01-31 02:55:44.948997945 +0000
  @@ -4,4 +4,3 @@
   Log::log_stream_policy
   bar
   bar
  -Log::log_stream_policy
```
More interesting is what happens next: the test in question uses a bunch of `@TEST-START-NEXT`, and it turns out that retries for multpart tests are just broken — they run on mismatched baselines. For the test in question, `bar` never enters the picture, yet we have:
```
  --- /tmp/test-diff.201352.test.log.baseline.tmp	2023-01-31 02:57:41.671878307 +0000
  +++ /tmp/test-diff.201352.test.log.tmp	2023-01-31 02:57:41.664878314 +0000
  @@ -6,5 +6,5 @@
   #open XXXX-XX-XX-XX-XX-XX
   #fields	t	status
   #types	time	string
  -XXXXXXXXXX.XXXXXX	foo
  +XXXXXXXXXX.XXXXXX	bar
   #close XXXX-XX-XX-XX-XX-XX
```
The fix is simple but the test cases are fiddly and I'm out of steam. Will finish tomorrow.

timwoj (approved):

timwoj:
Did you mean to change the btest submodule with this?

ckreibich:
Yeah, to pull in zeek/btest#87 in case it would have still mattered — but it didn't.

ckreibich:
(Forgot to push the rebase here before merging it locally — all good.)

PR #2740, opened 2023-02-01
ci: arm64 container follow-up

awelzel:


awelzel:
I've ditched the arm_debian11_task due to actually causing test failures. Going to look at those separately.

docker/README
```diff
@@ -1,19 +1,26 @@
 Container image for Zeek
 ========================
 
-This directory contains a minimal container image for Zeek. This image is
-published automatically to [DockerHub](https://hub.docker.com/u/zeekurity) for
+This directory allows building a minimal container image for Zeek. This image is
+published automatically to [DockerHub](https://hub.docker.com/u/zeek) for
```
timwoj, line 5: Aren't we also publishing these to the AWS registry?

timwoj (approved):
Looks good. Just one question about the README.

docker/README
awelzel: Yes, I've added a small addition there.

PR #2741, opened 2023-02-01
Cirrus CI: Run arm64 Debian 11 task

awelzel:
This currently fails, using Cirrus for logging into the instance.


awelzel:
Possibly fix merged into #2741 .

PR #2746, opened 2023-02-01
Fix issues due to signed/unsigned char discrepancies

awelzel:
On arm64, char is unsigned, while on x86 it's signed. This PR deals with that discrepancy.

In summary:
* on arm64, the HTTP analyzer would mistakenly allow non-ascii characters in the request line
* on x86, the netbios name decoding would mistakenly return non-ascii characters 
* logging of an invalid base64 character with the high-bit set on x86 would show as a negative number, on arm64 as a positive one.


timwoj (approved):

ckreibich:
Arne what did you do to find these — look compiler warnings? I'm wondering what prevents these things from slipping in in the future. Thx.

awelzel:
> Arne what did you do to find these — look compiler warnings? I'm wondering what prevents these things from slipping in in the future. Thx.

Test failures / baseline difference between x86 and arm64. There's possibly more things in areas where we have less coverage (or no tests that would show it). I don't believe there were compiler warnings, and skimming coverity also nothing there.

PR #2757, opened 2023-02-03
Fix git tag handling in the Docker image uploads workflows, for 5.0.x and 5.1.x.

ckreibich:
This brings the docker.yml version in the `release/5.0` branch up to where it was before we nuked it from master, and modernizes the git tag handling to `lts` and `latest`, away from `release`, which we no longer use.

This is untested, and we'd need to do the same for `release/5.1`. This is all transient until the new Cirrus setup trickles into release branches (like `release/5.2` now has, though I'm not sure it'll handle these tags correctly yet).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,6 +19,9 @@ env:
   IMAGE_FILE: /tmp/zeek-image.tar
   IMAGE_PATH: /tmp
 
+permissions:
+  contents: read
+
 jobs:
   docker-build:
     runs-on: ubuntu-latest
@@ -26,16 +29,16 @@ jobs:
       TEST_TAG: zeek:latest
       CONFFLAGS: --generator=Ninja --build-type=Release
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           submodules: "recursive"
 
       # Create and boot a loader. This will e.g., provide caching
       # so we avoid rebuilds of the same image after this step.
-      - uses: docker/setup-buildx-action@v1
+      - uses: docker/setup-buildx-action@v2
 
       - name: Build image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v3
         with:
           context: ./
           file: docker/Dockerfile
@@ -52,14 +55,14 @@ jobs:
 
       - name: Get version
         id: version
-        run: echo "::set-output name=RELEASE_VERSION::$(cat VERSION)"
+        run: echo "RELEASE_VERSION=$(cat VERSION)" >> $GITHUB_OUTPUT
 
       - name: Compute target tag
         id: target
         env:
           RELEASE_VERSION: ${{ steps.version.outputs.RELEASE_VERSION }}
         run: |
-          # Translate the Github reference into a tag name.
+          # Translate the GitHub reference into a tag name.
           #
           # - `release` tag maps to `zeek:latest`
           # - `v*` tag (excluding `v*-dev` tags) maps to `zeek:RELEASE_VERSION`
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -52,30 +55,43 @@ jobs:
...
-            echo "::set-output name=tag::zeek:${RELEASE_VERSION}"
+            echo "tag=zeek:${RELEASE_VERSION}" >> $GITHUB_OUTPUT
           fi
 
-      - name: Login to DockerHub
+      - name: Login to ECR
+        # Don't publish on forks. Also note that secrets for the login are not
+        # available for pull requests, so trigger on pushes only.
+        if: github.repository == 'zeek/zeek' && github.event_name == 'push'
         uses: docker/login-action@v1
```
timwoj, line 87: Should this be `login-action@v2` as well?

timwoj (changes requested):

.github/workflows/docker.yml
ckreibich: Indeed. Thank you for catching that. Fixed.

awelzel:
> This is all transient until the new Cirrus setup trickles into release branches (like release/5.2 now has, though I'm not sure it'll handle these tags correctly yet).

It doesn't yet. It only pushes "real" versioned images or `zeek/zeek-dev:latest`. I have vague memories we may have discussed this become one of the release steps on Slack. We could automate possibly with Cirrus manual tasks. 

The GitHub actions today (unless I missed something) use different image builds for the latest/lts tags compared to the versioned tags. E.g., `v5.0.6` and `latest` would be two different builds when instead they could be additional tags (or manifests) for the same set of images.

ckreibich:
> The GitHub actions today (unless I missed something) use different image builds for the latest/lts tags compared to the versioned tags. E.g., v5.0.6 and latest would be two different builds when instead they could be additional tags (or manifests) for the same set of images.

Yep. In the Github action the build kicks/kicked off per pushed tag.

It would be very nice to shrink the number of release steps and automate this, because it's conceptually nice to drive a lot from git tagging, and iirc we've been inconsistent about the Docker pushes in most recent releases. (This is not a complaint!)

This strikes me as doable — a push of git tag `lts` could check if a git version tag at the same commit already exists, and vice versa, and then shortcut the operation. Same for the git `latest` tag.

ckreibich:
Actually, I'm wondering whether we could stick to just pushing release version tags, and have CI do everything else: a new release on the most recent `release/x.0` branch automatically becomes `lts`, and on the most recent release branch it becomes `latest`, in both Docker and git? Might be too much magic.

awelzel:
> Might be too much magic.

I think this is great. It does seem racy when `release/5.2` has been branched and we do another `5.1.x` (the situation right now / same for lts when branching release/6.0 and doing a last 5.0.x). That might be okay though, or would need some more magic/cleverness that the latest `release/` branch chosen actually has a tag since branching off.

The Cirrus CI container setup doesn't have lts/latest handling, partly because what you're pointing out: We haven't actually pushed these tags consistently and I didn't want to replicate doing a whole new build - as we'd store twice as many images than needed (and with aarch64 being added another 2x). But I like your idea just doing it from the tags, and with `docker manifest create` we can just reference the same images for the latest/lts tags without the rebuild.

timwoj:
Is this being overridden by #2766 then?

timwoj (approved):

ckreibich:
Yeah — it'd be nice to have all setups consistent again and I hadn't anticipated having a working version of the Cirrus setup to try out. I'll review zeek/zeek#2766 now and close this out — can always revisit if needed.

PR #2758, opened 2023-02-06
Fix missing get_file_handle and new_event overhead

awelzel:
Prevent overhead when fuzzing due to:
* non-implemented `get_file_handle()` method for POP3 calling `get_file_handle()` repeatedly for every `DataIn()` call.
* `new_event` invocation requiring expensive argument construction (fueled by the previous issue)


timwoj (approved):
:shipit:

PR #2760, opened 2023-02-07
logging: Telemetry for streams and writers

awelzel:
This adds one metric per log stream and one metric per log writer (path based)
to track the number of writes on a stream level as well as on a writer level.

    $ curl -sSf localhost:8181/metrics | grep Conn
    zeek_log_writer_writes_total{endpoint="",filter-name="default",module="HTTP",path="http",stream="HTTP::LOG",writer="Log::WRITER_SQLITE"} 1 1677497572770
    zeek_log_stream_writes_total{endpoint="",module="HTTP",stream="HTTP::LOG"} 1 1677497572770

The initial version of this change also included metrics around log
write vetoes, but given no log policies exist in the default configuration
and they are mostly interesting for a few streams/writers only, skip this
for now. These can always be added by the script writer, too.

The difference between the stream level writes and concrete writers can
be used to deduce the number of vetoes (or errors) as a starting point.

I've been forth and back with normalizing the labels or prettifying the Log::ID enums, but in the end I think it's better to push that down to the consumer of the data as that's really a display problem.


awelzel:
@Neverlord - mind checking the usage of the telemetry family's and counters?

I've been wondering if allowing default construction (as a null counter) would make it a bit more easy to add telemetry to some existing code without diverging too much from existing patterns (see the struct Filter / struct Stream, but maybe that was just an unfortunate place to start).

src/logging/Manager.cc
```diff
@@ -750,12 +804,22 @@ bool Manager::Write(EnumVal* id, RecordVal* columns_arg)
 			auto v = filter->policy->Invoke(columns, IntrusivePtr{NewRef{}, id},
 			                                IntrusivePtr{NewRef{}, filter->fval});
 			if ( v && ! v->AsBool() )
+				{
+				// Hrml... not sure what's right. Seems not counting
+				// vetoes that are redundant is somewhat reasonable.
```
Neverlord, line 809: No intuition either, but the comment sounds like there should be some discussion happening on this at some point. Is there a ticket or discussion attached to it? Than a link to that would help. Otherwise, I'd drop the comment or reword it to just explain the rationale (e.g. "Don't count redundant vetoes to have the metric reflect unique vetoes.").

Neverlord (commented):
Looks good, just a couple nitpicks on naming convention and (lack of) smart pointer usage.

> I've been wondering if allowing default construction (as a null counter)

Passing them to the constructors seems fine to me - and IMHO the right thing to do in general. If we would allow null-pointers, then we would need to add null-checks. Either at the the handle types internally (causing extra branches and runtime overhead for all uses) or at call-side, which is easy to forget and can result in segfaults. If at all reasonable, I'd really like to keep the non-null guarantee to avoid the runtime and code overhead. If we do have places where it's unreasonable to enforce the non-null guarantee, then we need to decide which way we want to go (enforcing null-checks always or putting the burden on developers).

src/logging/Manager.cc
awelzel: I've dropped the conditional counting and removed the comment. I think the unusual metric behavior can be explained by the not-quite-intuitive behavior that if a veto happened anyway on a stream level, we still ask the filter whether it would veto, too.

src/logging/Manager.cc
```diff
@@ -132,7 +153,23 @@ Manager::Stream::~Stream()
 		delete *f;
 	}
 
-Manager::Manager() : plugin::ComponentManager<logging::Component>("Log", "Writer")
+Manager::Manager()
+	: plugin::ComponentManager<logging::Component>("Log", "Writer"),
+	  total_log_stream_writes_family(telemetry_mgr->CounterFamily(
+		  "zeek", "log_stream_writes", {"module", "stream_id"},
```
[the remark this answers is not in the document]
awelzel: Ah, yes. If we never want a user to pass `_`, maybe some runtime checking would be useful. At least to me there's such a strong connection to the Prometheus metrics, that I'll probably do that very often I'm afraid.

awelzel:
@ckreibich - is this something you'd sign off on? I've added those metrics over in zeek/zeek-demo as scripts, but I think they are valuable enough in base.

We could consider not including the `veto` ones as they're probably not very valuable without any policy handlers.

ckreibich:
I think this is great and pretty much a must if we want to consider our out-of-the-box metrics reasonably exhaustive. Cool! A few thoughts that came to mind while reading:
- There are a few possibilities for a log write to be aborted further down in `Manager::Write()`, after your increment of `total_writes_passed`. We could move that further down, or introduce a third metric — something like `total_writes_done` — to count the ones really happening.  It would be even cooler to get some sort of assurance that the write actually succeeded, but I think that info is too tucked away in the `WriterBackend` to be generally accessible.
- Sometimes packages sneak in filters without a particularly helpful module or name, and it would be helpful to understand what script or package the write/filter came from. But even if we had that, I don't see that information lending itself well to metric naming, so probably best to skip that.
- Metrics around log rotation might also be helpful. (Could certainly be a follow-up PR.)
- A very minor detail I got hung up on is the asymmetry between "writes" / "writes_vetoed" and "writes_passed" / "writes_vetoed".

awelzel:
> There are a few possibilities for a log write to be aborted further down in Manager::Write(), after your increment of total_writes_passed. We could move that further down, or introduce a third metric — something like total_writes_done — to count the ones really happening. It would be even cooler to get some sort of assurance that the write actually succeeded, but I think that info is too tucked away in the WriterBackend to be generally accessible.

I'm going to move it down to just before sending it to the backend. I thought most were "should never happen cases" (and some seem to better have a `reporter->Error/Warning` rather than just returning), but minimally the plugin hook skipping log writes seems benign.

> A very minor detail I got hung up on is the asymmetry between "writes" / "writes_vetoed" and "writes_passed" / "writes_vetoed".

Hmm, yes.

So, I'll actually go and drop the veto metrics from this PR and only have "stream writes total" and "filter writes total", where the latter represents anything that actually made it all the way to the filter as you suggest.

For any more detail between those two points (which log policy or filter policy vetoed or which plugin hook took over a write): Users can install their own metrics in script or plugin land if they are interested. In a default installation without stream/filter policy hooks, providing metrics that will always be 0 seems little useful.

awelzel:
I've repushed it removing the veto metrics and adding more labels for the lower log writer level. Also added lazy initialization of the stream metrics to avoid exposing zero metrics (I could see either way being better, but not logging streams that are never logged to seems somewhat reasonable).

This has now become fairly small and better suited as a start IMO here.

awelzel:
I'm going to pull this in with @Neverlord approval. Had chatted with @ckreibich a bit offline and it's now a rather condensed and hopefully useful set of metrics to have by default.

Issue #2765, opened 2023-02-08
Container image build follow-ups


awelzel:
We've switched to building container images on Cirrus CI. A few topics have come up around this:

- `docker` invocations talking to Docker Hub or the AWS registry can timeout, retry them once or twice.
- The `release/lts` and `release/latest` (?) tags functionality has not been moved over to Cirrus CI. @ckreibich [proposed an approach solely based on Git branches/tags](https://github.com/zeek/zeek/pull/2757#issuecomment-1418647000) that would avoid managing these extra tags.
- Consider pushing `zeek/zeek:{major}:{minor}` image tags (in addition to the `{major}:{minor}:{patch}` ones). This can allow pinning to `zeek/zeek:5.0`, pulling in updates for any patch versions, but without surprisingly jumping a minor version that may introduce new warnings, change log formats, etc.
- Do not promote the `lts` tag too much (maybe rename to lts-ci?). It's a good/nice tag to have for CI package testing, but for production use-cases it's unsuitable. Referencing `zeek/zeek:lts`, there's a risk that the tag jumps a full major version upon a new `docker pull`, so it's a long-term-surprise rather than long-term-stable.

awelzel:
Closing after merging https://github.com/zeek/zeek/pull/2766

PR #2766, opened 2023-02-08
cirrus/container: Automatically create and push lts and latest manifests

awelzel:
Should cover most of #2765 

@ckreibich - I'm still testing this over in my fork. Opening as a draft PR for now. It adds more shell scripting, so if you'd rather see this done differently, feel free to point that out.

https://github.com/awelzel/zeek/pull/4

### DIFF
--- a/ci/container-images-tag-and-push.sh
+++ b/ci/container-images-tag-and-push.sh
@@ -1,22 +1,24 @@
 #!/bin/bash
 #
-# This script expects two local images in the local container registry:
+# This script expects two images in the local container registry:
 #
 #   zeek/zeek-multiarch:arm64
 #   zeek/zeek-multiarch:amd64
 #
-# It retags these according to the environment ARCH_IMAGE_NAME and
-# ARCH_IMAGE_TAG as zeek/${ARCH_IMAGE_NAME}:${ARCH_IMAGE_TAG}-{arm64,amd64},
-# pushes them to the registry, then creates a manifest based on MANIFEST_NAME
-# and MANIFEST_TAG environment variables as zeek/${MANIFEST_NAME}:${MANIFEST_TAG}
-# including the two tags.
+# It retags these according to the environment variables IMAGE_NAME and
+# IMAGE_TAG as zeek/${IMAGE_NAME}:${IMAGE_TAG}-{arm64,amd64}, pushes them
+# to the registry, then creates a manifest as zeek/${IMAGE_NAME}:${IMAGE_TAG}
+# containing the arch specific tags and pushes it.
 #
 # REGISTRY_PREFIX can be used to prefix images with a registry. Needs
 # to end with a slash.
+#
 set -eux
 
 REGISTRY_PREFIX=${REGISTRY_PREFIX:-}
 ZEEK_IMAGE_REPO=${ZEEK_IMAGE_REPO:-zeek}
+
+ADDITIONAL_MANIFEST_TAGS=${ADDITIONAL_MANIFEST_TAGS:-}
 
 # Check for ending slash in registry prefix
 if [ -n "${REGISTRY_PREFIX}" ]; then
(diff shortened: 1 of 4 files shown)


awelzel:
Over in a fork this is running on tag 6.0.0 and 6.1.5:

https://cirrus-ci.com/build/5087138480390144
https://cirrus-ci.com/build/6671673047384064

Theoretically, these should automatically push respective lts and latest tags and 6.0 and 6.1 tags into `arnewelzel/zeek`.

awelzel:
The builds from the above referenced did what I had expected. Marking this ready.

ci/find-current-version.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+# Iterate through all candidate branches, determine if a corresponding
+# v{x}.{y}.0 tag exists for that branch. If so, that'll be the most recent
+# (highest sorting) branch where we had a release.
+for ref in $(git show-ref | grep -E "${PATTERN}" | awk '{ print $2 }' | sort -rn); do
+    version=$(echo $ref | sed -E 's,^.*/(.+)$,\1,g')
+    tag_ref="refs/tags/v${version}.0"
+
+    # Find the tag using *objname as Zeek is using annotated tags.
```
ckreibich, line 41: Do you mean this as a filter to rule out unannotated tags? I stumbled over this while playing with the script, because I didn't annotate my test tags. If you don't mean to insist on annotated tags, I suggest `git rev-list -n 1 ${tag_ref} 2>/dev/null`.

ci/find-current-version.sh
```diff
@@ -0,0 +1,57 @@
...
+}
+
+if [ $# -ne 1 ]; then
+    usage
+fi
+
+if [ "${1}" = "lts" ]; then
+    PATTERN=".* refs/remotes/${REMOTE}/release/[0-9]+\.0\$"
+elif [ "${1}" = "feature" ]; then
+    PATTERN=".* refs/remotes/${REMOTE}/release/[0-9]+\.[1-9][0-9]*\$"
```
ckreibich, line 29: (Same lts vs feature release concern here as in my first comment.)

ci/find-current-version.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo "Usage $0 <lts|feature>" >&2
+    exit 1
+}
+
+if [ $# -ne 1 ]; then
+    usage
+fi
+
+if [ "${1}" = "lts" ]; then
+    PATTERN=".* refs/remotes/${REMOTE}/release/[0-9]+\.0\$"
```
ckreibich, line 27: Just curious why you match on remotes here — because we wouldn't use this on repos where the release branch only exists locally, not pushed yet? I think that's okay, just wanted to flag.

ci/container-images-tag-and-push.sh
```diff
@@ -26,13 +28,32 @@ if [ -n "${REGISTRY_PREFIX}" ]; then
...
 fi
 
-docker tag ${ZEEK_IMAGE_REPO}/zeek-multiarch:arm64 ${REGISTRY_PREFIX}${ZEEK_IMAGE_REPO}/${ARCH_IMAGE_NAME}:${ARCH_IMAGE_TAG}-arm64
-docker tag ${ZEEK_IMAGE_REPO}/zeek-multiarch:amd64 ${REGISTRY_PREFIX}${ZEEK_IMAGE_REPO}/${ARCH_IMAGE_NAME}:${ARCH_IMAGE_TAG}-amd64
-docker push ${REGISTRY_PREFIX}${ZEEK_IMAGE_REPO}/${ARCH_IMAGE_NAME}:${ARCH_IMAGE_TAG}-arm64
-docker push ${REGISTRY_PREFIX}${ZEEK_IMAGE_REPO}/${ARCH_IMAGE_NAME}:${ARCH_IMAGE_TAG}-amd64
+# Forward arguments to docker and retry the command once if failing (e.g network issues).
+function do_docker {
+    if ! docker "$@"; then
+        docker "$@"
```
ckreibich, line 34: I'm sure you considered this, but should this sleep before the retry?

ckreibich (commented):
Great scripts! Just a few questions around the git logic.

With this in place, we could even automate `lts` and `latest` tag placement in git itself. If we can compute which releases should bear these tags, then there's technically no need for the release manager to place them manually.

ci/find-current-version.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+#
+# Helper script to return the current lts or feature version based on
+# the branch and tags of Zeek's repository.
+#
+# * lts is the highest sorting release/x.0 branch which contains a tag
+#   of the form form '^v{x}.0.0$'.
+#
+# * feature is the highest sorting release/x.[^0]+ branch that does
```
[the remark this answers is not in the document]
awelzel: Clearly - for some reason the `.0` was never feature in my mind. Let me rethink/redo this.

ci/find-current-version.sh
awelzel: update-changes -R makes it an annotated tag, so I thought we could enforce it. I'm happy to relax - makes testing easier, too.

ci/container-images-tag-and-push.sh
awelzel: Ah yes, was thinking more of the upload stops after a few seconds and stalls and immediate retry will be success. I'll add 5 seconds, but otherwise no advanced retry mechanisms - there's task retry on a higher level, too.

ci/find-current-version.sh
awelzel: Mainly because Cirrus clones the repo with single-branch and we then fetch the remote release heads as a follow-up:
https://github.com/zeek/zeek/pull/2766/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR520

We could fetch them directly into `refs/heads` locally (or otherwise check out all release branches locally), but it seemed more explicit to indicate determination of feature/lts depends on the remote state of the repo (minimally in CI).

awelzel:
> With this in place, we could even automate lts and latest tag placement in git itself.

What are these used for? Were they only used to trigger the GitHub docker action? Or are they referenced by outside projects?

I'd do that as a second step, because this would involve write access to the Git repository from within CI which seems separate enough.

awelzel:
Okay:

Tag v7.0.1 when there has not yet been a v7.1.0 tag: This pushed the tags "7.0.1 lts latest 7.0" to `arnewelzel/zeek`:
https://cirrus-ci.com/task/6155043108290560

Creating release/7.1.0 and tag v7.1.0 as a follow-up: This pushed "7.1.0 latest 7.1" to `arnewelzel/zeek`
https://cirrus-ci.com/build/6277863872135168

Creating tag 7.0.2: This pushed "7.0.2 lts and 7.0", but not the "latest" tag, because of the above 7.1.0 tag.
https://cirrus-ci.com/build/6025187959767040

@ckreibich , one more round, it may now actually do the right thing.

ckreibich (approved):
Nice! Fingers crossed. :-)

awelzel:
Okay. I've squashed this down and merged into master :crossed_fingers: 

@timwoj , 357bdc71ada1372fa690194de8dde097cf6d7b6b, would be the one for `release/5.2` and I guess we'll see if things actually work out once you tag `v5.2.0`.

timwoj:
Ok, I'll go ahead and pull it over.

PR #2767, opened 2023-02-09
Add zeek -V/--build-info

### LINKED ISSUE #1405: Store information about the build during 'make install'
We ran into some issues with the 4.0-RC testing where we were seeing a bug in an install but couldn't actually know what commits the build was based on. Can we add something to the build output, either a text file or an extended `--version` argument that outputs more information about the source the build was based on? Something like the commit hash of the zeek tree and the submodules?

awelzel:
This renders a JSON string within the CMake infrastructure and puts it into the existing version.c.in file. -V then just prints it out.

Closes #1405


src/CMakeLists.txt
```diff
@@ -30,6 +30,9 @@ set(bro_REGISTER_BIFS CACHE INTERNAL "BIFs for automatic registering" FORCE)
 set(bro_BASE_BIF_SCRIPTS CACHE INTERNAL "Zeek script stubs for BIFs in base distribution of Zeek" FORCE)
 set(bro_PLUGIN_BIF_SCRIPTS CACHE INTERNAL "Zeek script stubs for BIFs in Zeek plugins" FORCE)
 
+# Poor man's JSON escaping as this is renderd into a C string.
```
timwoj, line 33:
```suggestion
# Poor man's JSON escaping as this is rendered into a C string.
```

timwoj (approved):
This looks pretty good to me. I can't think of anything else useful to add to it.

awelzel:
> This looks pretty good to me. I can't think of anything else useful to add to it.

One thing I realized: The information will only be collected at configure time. So a commit after `./configure ...` won't be reflected in the JSON, unless reconfigured. That's likely okay for now and always okay in CI currently.

PR #2768, opened 2023-02-09
Monthly ZAM maintenance: improved error messages + a fix for "in" expressions

vpax:
This PR changes ZAM to associated `CallExpr` nodes with ZAM low-level call instructions, to generate better error messages (particularly when calling BiFs).  It also includes a fix for compiling `in` expressions where the RHS operand is compound (like a function call), rather than just a variable name.

Needs this tweak to `gen-zam`: https://github.com/zeek/gen-zam/pull/6


timwoj (approved):
Should this go into 5.2?

vpax:
> Should this go into 5.2?

I was hoping it could, both of the fixes are non-trivial user-visible improvements.

PR #2769, opened 2023-02-09
Don't specify the version of openssl in ci/macos/prepare.sh

timwoj:


ci/macos/prepare.sh
```diff
@@ -6,14 +6,14 @@ set -e
...
 python3 -m pip install --user websockets
 
 # Brew doesn't create the /opt/homebrew/opt/openssl symlink if you install
 # openssl@1.1, only with 3.0. Create the symlink if it doesn't exist.
-if [ ! -e /opt/homebrew/opt/openssl ]; then
-    if [ -d /opt/homebrew/opt/openssl@1.1 ]; then
-        ln -s /opt/homebrew/opt/openssl@1.1 /opt/homebrew/opt/openssl
-    fi
-fi
+#if [ ! -e /opt/homebrew/opt/openssl ]; then
```
awelzel, line 15: If this is not necessary, maybe it can be removed as a follow-up?

awelzel (approved):
I'm going to merge this as it's in all branch runs.

PR #2770, opened 2023-02-10
Add PBB (802.1ah) decapsulation

ekoyle:
Decapsulate 802.1ah (provider backbone bridging) traffic.

I used the pcap from https://redmine.openinfosecfoundation.org/attachments/1733 for testing, but I'm not sure what the license is on that file.


ekoyle:
Would it make more sense to use the customer MACs for l2_src and l2_dst instead of those from the ethernet header?  The customer MACs should be the original values from the packet that was encapsulated, which might be more interesting than what is in the ethernet header.

ekoyle:
Also, since PBB is encapsulating an ethernet header, should the pbb analyzer mimic the logic in the ethernet analyzer?  I think it can technically encapsulate any ethernet frame.

timwoj:
> Also, since PBB is encapsulating an ethernet header, should the pbb analyzer mimic the logic in the ethernet analyzer? I think it can technically encapsulate any ethernet frame.

It could technically parse out just the bits that it needs and then pass the rest down to the Ethernet packet analyzer, but I'm not sure what else is beyond the PBB header and the Ethernet analyer would need to know how to forward to it. Given that it's not that much data it's probably easier to just reimplement the parts you need.

src/packet_analysis/protocol/pbb/PBB.h
```diff
@@ -0,0 +1,31 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#pragma once
+
+#include "zeek/packet_analysis/Analyzer.h"
+#include "zeek/packet_analysis/Component.h"
+
+#define PBB_LEN 18u
```
timwoj, line 8: These should be defined like `constexpr int PBB_LEN = 18` instead of `#defines`. They should also be defined in the cc file since they're only used there.

timwoj (changes requested):
Seems pretty straight-forward. It'd be good to add a btest that exercises it for the code coverage.

ekoyle:
PBB allows encapsulation of an ethernet frame.  The header encapsulates all of the data from the original ethernet frame across a service provider network (so that the original ethernet frame can be re-created on the far end), along with some additional data that is useful for the provider.

Practically, you end up with something like:
```
Ethernet w/provider SRC/DST MAC addresses
possible provider VLAN tags
possible provider MPLS tags
PBB header (including customer SRC/DST MAC addresses & ethertype)
...
```

What I was wondering is if we need to put in handling for the inner ethertype being less than 1500/1536 (since this is encapsulating an ethernet frame, that seems like the "correct" approach).  I don't know if zeek does anything with llc traffic, so I didn't implement that.

timwoj:
Ah, ok. I think I get it. It's basically a second ethernet-based packet encapsulated inside another ethernet-based packet. The 802.1ah header is just an Ethernet header with some extra fields (namely the I-Tag and/or Footer). I could see building a new packet out of the encapsulated data after stripping out the extra fields and just forwarding it down to the Ethernet analyzer instead of duplicating the work.

scripts/base/packet-protocols/pbb/main.zeek
```diff
@@ -2,13 +2,5 @@ module PacketAnalyzer::PBB;
...
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x0800, PacketAnalyzer::ANALYZER_IP);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x86DD, PacketAnalyzer::ANALYZER_IP);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x0806, PacketAnalyzer::ANALYZER_ARP);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x8035, PacketAnalyzer::ANALYZER_ARP);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x8100, PacketAnalyzer::ANALYZER_VLAN);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x8100, PacketAnalyzer::ANALYZER_VLAN);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x88A8, PacketAnalyzer::ANALYZER_VLAN);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x9100, PacketAnalyzer::ANALYZER_VLAN);
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x8864, PacketAnalyzer::ANALYZER_PPPOE);
+	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_PBB, 0x6558, PacketAnalyzer::ANALYZER_ETHERNET);
```
timwoj, line 5: If the only place this is ever going to forward is into the ethernet analyzer, you can remove this `zeek_init` handler altogether and add something like this instead:

```
export {
        ## Default analyzer
        const default_analyzer: PacketAnalyzer::Tag = PacketAnalyzer::ANALYZER_ETHERNET &redef;
}
```

With that in place, the `PROTO_TEB` constant can go away and you don't have to pass a protocol value into `ForwardPacket`.

timwoj (changes requested):
Not sure why, but the `scripts.base.files.x509.files` btest is failing because of this. The `analyzers` is ending up in a different order. It probably just needs a baseline update.

ekoyle:
I can't figure out any reason for the btest results to have changed.

 * `scripts.base.files.x509.files` looks pretty straightforward (order changed from `X509,SHA256` to `SHA256,X509`, which is really strange)
 * `scripts.policy.misc.dump-events` is a bit more concerning, it has the same order change but also there are some where `md5` is changing from `<uninitialized>` to an MD5 hash.

I think both of these are side-effects of analyzers running in a different order, but I can't figure out why that would be happening (maybe it's just a weird optimizer quirk?)

It also looks like `scripts.policy.misc.dump-events` has some additional `file_hash` lines in the `all-events-no-args.log` that weren't there before.  I've tried everything I can think of (commenting out PBB in zeek scripts, moving the PBB stuff to the end of the cmake file), but nothing seems to make any difference.

awelzel:
> I can't figure out any reason for the btest results to have changed.
> 
> * `scripts.base.files.x509.files` looks pretty straightforward (order changed from `X509,SHA256` to `SHA256,X509`, which is really strange)

There's a long-winded explanation here: https://github.com/zeek/zeek/pull/2340#issuecomment-1216306937

The same happened when support for the LinuxSLL2 link type was added because underlying enum values are changing which causes `set/tables` to be ordered differently.

> * `scripts.policy.misc.dump-events` is a bit more concerning, it has the same order change but also there are some where `md5` is changing from `<uninitialized>` to an MD5 hash.

It seems your change somehow provokes `file_hash` being raised to script land before `x509_certificate`, `x509_extension`, ...

I have not yet grasped why, but it's definitely the `AddComponent()` call in the new Plugin.c file that provokes it.

It's likely harmless (the populated md5 fields in the output are because `file_hash` populates it). Next to plugin.hooks, dump-events is one of those unspecific tests that trigger on most anything. And then there's surprising behavior changes when underlying enum values change.

awelzel:
>
> It's likely harmless (the populated md5 fields in the output are because `file_hash` populates it). Next to plugin.hooks, dump-events is one of those unspecific tests that trigger on most anything. And then there's surprising behavior changes when underlying enum values change.

Ah, yes, it's the same cause: The set here is likely iterated in a different order and so analyzer are registered (and later processed) in different orders, too:

https://github.com/zeek/zeek/blob/3de785114b5b1d632e96231bb41abce1723c2598/scripts/base/frameworks/files/main.zeek#L555-L561

So, I think just update them and possibly reference some of this discussion.

awelzel:
FWIW, I could see pre-sorting registered analyzer tags to make the `Files::add_analyzer()` more predictable. It won't fix the situation where things change/shuffle when a new file analyzer is added, but that would at least be expected. This is the second time adding a "packet analyzer" perturbs something seemingly unrelated and leaves a new contributor confused.

```
--- a/scripts/base/frameworks/files/main.zeek
+++ b/scripts/base/frameworks/files/main.zeek
@@ -335,7 +335,7 @@ global registered_protocols: table[Analyzer::Tag] of ProtoRegistration = table()
 
 # Store the MIME type to analyzer mappings.
 global mime_types: table[Files::Tag] of set[string];
-global mime_type_to_analyzers: table[string] of set[Files::Tag];
+global mime_type_to_analyzers: table[string] of vector of Files::Tag;
 
 global analyzer_add_callbacks: table[Files::Tag] of function(f: fa_file, args: AnalyzerArgs) = table();
 
@@ -480,9 +480,16 @@ function register_for_mime_type(tag: Files::Tag, mt: string) : bool
 
        if ( mt !in mime_type_to_analyzers )
                {
-               mime_type_to_analyzers[mt] = set();
+               mime_type_to_analyzers[mt] = vector();
                }
-       add mime_type_to_analyzers[mt][tag];
+
+       mime_type_to_analyzers[mt] += tag;
+
+       # Make analyzer order predictable
+       local files_tag_cmp = function(t1: Files::Tag, t2: Files::Tag): int {
+               return strcmp(cat(t1), cat(t2));
+       };
+       sort(mime_type_to_analyzers[mt], files_tag_cmp);
 
        return T;
        }
@@ -556,7 +563,7 @@ event file_sniff(f: fa_file, meta: fa_metadata) &priority=10
             meta$mime_type in mime_type_to_analyzers )
                {
                local analyzers = mime_type_to_analyzers[meta$mime_type];
-               for ( a in analyzers )
+               for ( _, a in analyzers )
                        Files::add_analyzer(f, a);
                }
        }
```

ekoyle:
Looks like I misread the output of the `all-events-no-args.log` diff -- it is also just changing order, no added lines.

I went ahead and updated the btests.

timwoj (approved):

PR #2771, opened 2023-02-11
Treat private address space as site-local by default

### LINKED ISSUE #2761: 100.64.0.0/10 is not tagged as local net
In eve log, `local_orig` / `local_resp` are set to `false` with the subnet 100.64.0.0/10.

```
{
  "ts": 1675724365.526336,
  "uid": "CCS5ib1CTbiP9XDijf",
  "id.orig_h": "xxxx",
  "id.orig_p": 56937,
  "id.resp_h": "100.102.248.9",
  "id.resp_p": xxxx,
  "proto": "tcp",
  "duration": 9.5367431640625e-07,
  "orig_bytes": 0,
  "resp_bytes": 0,
  "conn_state": "S0",
  "local_orig": true,
  "local_resp": false,
  "missed_bytes": 0,
  "history": "S",
  "orig_pkts": 2,
  "orig_ip_bytes": 96,
  "resp_pkts": 0,
  "resp_ip_bytes": 0,
  "orig_l2_addr": "xxxx",
  "resp_l2_addr": "xxxx"
}
```

However this subnet is present in share/zeek/base/utils/site.zeek, so it should be identified as a local net

```
# Shared Address Space (also known as Carrier-grade NAT), see :rfc:`6598`
100.64.0.0/10,
```

ckreibich:
We currently define `Site::private_address_space` to an exhaustive list of IANA-supplied private address ranges, and leave `Site::local_nets` empty, for the user to fill. Newer users are often surprised by this, wondering why e.g. 192.168/16 addresses don't get reported as local in our logs. Then there's the fact that zeekctl defaults `Site::local_nets` to 10/8, 172.16/12, and 192.168/16, a subset of `Site::private_address_space`.

If we default to including private address space in local nets, while allowing (power) users to adjust these things as before, this should get a smoother.  We've touched this several times in recent past and I believe were in agreement that this would be good to have. I've looked at at couple of ways of doing this:

* An on/off switch that controls `Site::is_local_addr()` so it potentially also looks at `Site::is_private_addr()`. This doesn't help much: `Site::local_nets` is public and we have several places that condition on `|Site::local_nets| > 0`, so this'd be too inconsistent.
* Simply start off `Site::local_nets` from the same default as `Site::private_address_space`, i.e., `local_nets = copy(private_address_space)`. This nails the beginning but then becomes unintuitive: changes to `Site::private_address_space`, including redefs, don't propagate.
* Treat `Site::private_address_space` like a subset of `Site::local_nets`. Since both are config options, they can actually be updated at runtime, so changes to the former can be copied over into the latter.

This PR implements the last option. It's really just a change handler application, with a bit of subtlety in the fact that one change handler triggers another (see commits), and the fact that this default change handler activity finds its way into config.log, which polluted a bunch of tests, so I added a small mechanism to suppress its logging — seems fine to me, since "real" updates to either of the sets will still log as expected.

A new `Site::private_space_is_local` toggle controls the whole thing, and defaults to true. There's a new set of tests that shows the behavior. I've not yet updated the external testsuite. I've eyeballed the changes (a whole bunch, as you'd imagine...) and it's what you'd expect: a lot of field updates from `-` to `T` or `F`, but also a handful of new log writes since code that required local nets now engages.

Resolves #2761.


scripts/base/frameworks/config/main.zeek
```diff
@@ -147,6 +147,9 @@ function format_value(value: any) : string
 
 function config_option_changed(ID: string, new_value: any, location: string): any &is_used
 	{
+	# Some option updates reflect Zeek-internal activity and shouldn't be logged.
+	if ( location == "<internal>" )
```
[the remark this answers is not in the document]
ckreibich: Heh, I was there too! I settled for `<internal>` because it's a location, not so much a purpose, but this is perfectly fine with me — I'll go for `<skip-config-log>`.

scripts/base/utils/site.zeek
```diff
@@ -255,6 +283,35 @@ function update_neighbor_zones_regex(id: string, new_value: set[string]): set[st
...
+	# Subtle: calling Option::set() on Site::local_nets will cause its
+	# change handler update_local_nets_table() to trigger directly. The latter
+	# will access Site::private_address_space, which will still have its old
+	# value since this change handler hasn't returned yet. Scheduling an
+	# event to take care of it doesn't quite work either since it'll run too
+	# late. Instead we coordinate the change handlers directly: an extra set
+	# stores the pending update, and a state bit indicates whether it's to be
+	# used since update_local_nets_table() can also trigger via updates to
+	# Site::local_nets itself.
+	pending_pas = new_value;
```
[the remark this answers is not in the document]
ckreibich: This is pretty subtle — we can get by without `pending_pas` if we tweak the `new_local_nets` computation a little bit to ensure `private_address_space` is really in there. It's better that way, so I like your suggestion a lot. Coming up.

scripts/base/utils/site.zeek
```diff
@@ -255,6 +283,35 @@ function update_neighbor_zones_regex(id: string, new_value: set[string]): set[st
...
 	}
 
+function update_private_address_space(id: string, new_value: set[subnet]): set[subnet]
+	{
+	# This change handler mirrors the changes to private ranges into
+	# Site::local_nets. It does not use clusterization: the update to the
+	# private address space already propagates, so we just apply the change
+	# locally.
+	local new_privates = new_value - private_address_space;
+	local old_privates = private_address_space - new_value;
```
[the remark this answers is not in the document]
ckreibich: Yes, this is true. I think this is fine, but for sure the relationship of the two sets needs detailed documentation.

awelzel:
The other more general thought around this changeset is that it will cause "havoc" / "churn" downstream for any package, plugin or analyzer maintainers using btest-diff on a full conn.log. If they adhered to best practices and added tests we're now forcing them to update their packages. Nothing really broke (or got deprecated) - which is what tests should signify - it's just churn.

Specifically supporting multiple Zeek versions with the same set of tests would either involve zeek-cut or somehow disabling the option that's compatible with older versions. One could argue that with Zeek 6.0, there's theoretically not multiple version that are subject to be supported anyway, but reality is most likely different.

Of course, `local_orig` and `local_resp` shouldn't be in the baselines for packages if they are not relevant, but clearly Zeek's base provides enough examples that it's fine to include them into baselines.

I'm not sure there's a good way around this outside of making this opt-in and enabling it via local.zeek / zeekctl, or providing a reasonable guideline to opt-out of this for tests in a way that's compatible with older version as well.

Maybe that's just all a topic in itself, because #2772 will trigger the same situation with `dns.log`, albeit hopefully with less impact.

awelzel (approved):
There seems to be still a baseline for dump-events not being happy. I'm otherwise approving because I guess we want to move this forward.

ckreibich:
All good points, yep. I was indeed wondering whether to leave it off by default — it's still much easier to switch one Boolean than to have to navigate `Site::private_address_space` vs `Site::local_nets` management yourself. On the other hand, that Boolean is also easily toggled the other way. I don't really like the idea of us avoiding (rare) major changes because it will break existing tests that are based on the notion of baselining full logs — imo a risky proposition to begin with.

Happy to hear more opinions here. For now I'm going to take a look at that CI test failure on a subset of systems, which looks funky.

scripts/base/utils/site.zeek
```diff
@@ -103,10 +103,15 @@ export {
 		[fe80::]/10,
 	};
 
-	## Networks that are considered "local".  Note that ZeekControl sets
-	## this automatically.
```
rsmmr, line 107: I'd keep the ZeekControl comment: That's not only about private address space. ZeekControl users normally configure all of their local nets over there.

scripts/base/frameworks/config/main.zeek
```diff
@@ -147,6 +147,9 @@ function format_value(value: any) : string
 
 function config_option_changed(ID: string, new_value: any, location: string): any &is_used
 	{
+	# Some option updates reflect Zeek-internal activity and shouldn't be logged.
+	if ( location == "<skip-config-log>" )
```
rsmmr, line 151: Is there a place where we could document this?

rsmmr (approved):
Seems we should update ZeekControl as well now to remove the private space from local nets.

rsmmr:
>  I don't really like the idea of us avoiding (rare) major changes because it will break existing tests that are based on the notion of baselining full logs — imo a risky proposition to begin with.

Agree with that if we don't do it too often. 6.0 seems like a good time for changes like this.

scripts/base/utils/site.zeek
ckreibich: Yes that's true. I removed it because of the upcoming zeekctl change that pairs with this, but you're right — it's still worth flagging the connection. Will do.

scripts/base/frameworks/config/main.zeek
ckreibich: Yeah, I'd say in the config framework docs since they cover change handlers in some detail.

I just noticed that the logs chapter in the manual doesn't cover config.log at all yet.

ckreibich:
I've rebased the whole thing on top of master, brought back the zeekctl comment, and added a NEWS update. I'll merge tomorrow if folks don't have further comments, and will follow it up with the corresponding zeekctl tweak and a docs update.

ckreibich:
At last ... I'll follow up with the discussed zeekctl and docs changes tomorrow, but for now I just wanted to get this merged. The rebases are pretty tedious.

PR #2772, opened 2023-02-11
Parse DNSSEC AD and CD bits

### LINKED ISSUE #2672: The DNS analyzer should break the 3-bit header Z field into Z / DC / AD bits
Zeek's DNS analyzer treats the Z field in a message header as three bits (see [here](https://github.com/zeek/zeek/blob/master/src/analyzer/protocol/dns/DNS.h#L324) and [here](https://github.com/zeek/zeek/blob/master/src/analyzer/protocol/dns/DNS.cc#L1793)), as originally defined. DNSSEC broke that field up into a single Z bit, an "authentic data" AD bit, and a "checking disabled" CD bit, see [RFC 2535](https://www.rfc-editor.org/rfc/rfc2535#section-6.1). The semantics of the fields then needed several iterations in subsequent RFCs so it's all a bit messy, but essentially CD may be set by the resolver, and AD may be [set by resolver or server](https://www.rfc-editor.org/rfc/rfc6840#section-5.7).

Wireshark and tcpdump both parse the fields: Wireshark's query strings are `dns.flags.authenticated` and `dns.flags.checkdisable`, tcpdump shows CD in queries as `%`, and AD in responses as `$` after the query ID.

Zeek technically provides all values in the script layer since the 3-bit Z field is rendered as a `count`, but it'd be nicer to separate out the bits as well.

micrictor:
Parse authentic data (AD) and checking disabled (CD) bits according to RFC 2535. Leaves the Z field as-is, in case users are already handling this elsewhere and depend on the value being the integer for all 3 bits.

https://www.rfc-editor.org/rfc/rfc2535#section-6.1

Fixes #2672


scripts/base/protocols/dns/main.zeek
```diff
@@ -429,6 +443,14 @@ event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qcla
 	c$dns$qtype_name  = query_types[qtype];
 	c$dns$Z           = msg$Z;
```
timwoj, line 445: Does AD get parsed in a request? I only did a quick skim of the RF, but it sounds like it should only be valid in a response.

micrictor: In the canonical RFC for AD, it was left undefined; RFC 6840 clarified that the AD bit can be set by the request or response
https://www.rfc-editor.org/rfc/rfc6840#section-5.7

timwoj (approved):

timwoj:
Looks good to me. Can you update `testing/external/commit-hash.zeek-testing` to the commit from your PR on that repo, and re-push this for the CI coverage? I have to imagine that it'll still fail because zeek-testing-private fails but I'll have to update that part.

scripts/base/protocols/dns/main.zeek
```diff
@@ -359,6 +367,13 @@ hook DNS::do_reply(c: connection, msg: dns_msg, ans: dns_answer, reply: string)
 		# Currently only standard queries are tracked.
 		return;
 
+	# Decode AD bit according to RFC 2535 if Z is set.
+	# Leave Z untouched in case existing users post-process the number value.
+	if ( msg$Z != 0 )
```
awelzel, line 372: IMO we should adapt the analyzer to decode the bits and then extend `dns_msg` with AD and CD fields (possibly always set to whatever Z would evaluate to rather than making them optional). This would give script writers immediate access to the fields without relying on logic in `dns/main.zeek`.

Extending dns.log with these new fields could be discussed separately (not sure the original ticket assumed that would be done). These fields might be useful enough in every deployment, but could also see not adding them (or not marking them `&log`) by default. But suggest to add them for now.

awelzel (changes requested):

scripts/base/protocols/dns/main.zeek
micrictor: Decoding the bits directly in the parser SGTM, and makes sense as these flags should be parsed out the same as RA/RD/etc - will make that change. Unless someone thinks otherwise, I'll still leave `Z` untouched, on the off chance someone downwstream depends on it holding all three bits. 

I'll still add them to the log - IMO, we log every other standard part of the DNS header, so we should log these as well.

ckreibich: I agree with all here but this last bit is unusual — we don't normally leave a field in a message header and then parse out its bits in addition. I would normally vote for deprecation of the old `Z` field and introduction of a new one. On the other hand this case seems a bit special because even popular DNS resources often don't mention AD & CD.

 So ... in the end I agree — let's keep the Z field as-is, but let's also be very clear in the docstrings about how Z, AD, and CD relate in our implementation.

awelzel:
@micrictor - are you still interested and planning in addressing the feedback?

micrictor:
Yes! Sorry about the delay - life is currently happening to me at an above
average rate, it seems.

scripts/base/protocols/dns/main.zeek
micrictor: Done - newly pushed changes moved the bit-parsing to the protocol parser. Also pushed a change to docs that explicitly calls out that Z is the 3-bit Z specified in the original DNS RFC.

scripts/base/protocols/dns/main.zeek
```diff
@@ -428,6 +443,8 @@ event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qcla
 	c$dns$qtype       = qtype;
 	c$dns$qtype_name  = query_types[qtype];
 	c$dns$Z           = msg$Z;
+	c$dns$AD          = msg$AD;
```
awelzel, line 446: I'm unsure if we should set `AD` when processing `dns_request`: In requests it should be 0 IIUC.

We could instead consider to flag AD being set in requests (or CD being set in responses) as a weird, but can do as a follow up.

scripts/base/protocols/dns/main.zeek
```diff
@@ -60,9 +60,17 @@ export {
 		## The Recursion Available bit in a response message indicates
 		## that the name server supports recursive queries.
 		RA:            bool               &log &default=F;
-		## A reserved field that is usually zero in
-		## queries and responses.
+		## A reserved field that is usually zero in queries and
```
awelzel, line 63: Should we drop the comment as it is not applicable today anymore?

testing/btest/btest-results.xml
```diff
@@ -0,0 +1,1503 @@
+<?xml version="1.0" ?>
```
awelzel, line 1: Mind dropping that file and the tmp.tar.gz file below again?

awelzel (commented):
Hey @micrictor - I had a few more comments and it seems you added some unrelated files (and a merge commit is in the history, too). There's something off with the external testing baselines, too.

If you agree with the comments and would rather hand this off, I'm happy to pick it up and see it through. I'd squash the history a bit, but keep your authorship, of course.

Let us know.

scripts/base/protocols/dns/main.zeek
micrictor: RFC 6840 states that the AD bit can be set by the request or response
https://www.rfc-editor.org/rfc/rfc6840#section-5.7

micrictor: More info on why - The value set here will be overriden by the response value if both the request and reply are seen by Zeek. Definitely worth adding weird logs in the future.

scripts/base/protocols/dns/main.zeek
micrictor: Updated comment to match the zeek-docs, let me know if that addresses your concern

scripts/base/protocols/dns/main.zeek
awelzel: Thanks for re-linking that RFC.

> request and reply are seen by Zeek

Which should be the common case. What's there now I think is okay. As you mention, the `AD` column in the log will be overwritten by the response's AD flag and that's semi great.

I'm paging @ckreibich if we should solve this somehow differently in a follow-up (new column, or let script writers extend the log themselves), but I'll look into merging your this changeset as is.

awelzel (approved):

PR #2774, opened 2023-02-13
Bump spicy-plugin to v1.3.23.

bbannier:


timwoj:
Does this need to go into 5.2 as well? It looks like @rsmmr still has a few things in flight too, should I wait for that?

bbannier:
> Does this need to go into 5.2 as well? It looks like @rsmmr still has a few things in flight too, should I wait for that?

No, this is only for `release/5.0`. Zeek's `release/5.0` uses spicy-plugin 1.3.x while 5.2 uses spicy-plugin 1.4.x. I'll cut a release and will bump it later this week.

PR #2778, opened 2023-02-13
Call AnalyzerConfirmation earlier in the process for VXLAN and Geneve

### LINKED ISSUE #2720: Analyzer confirmations for tunnel analyzers are not consistent
I ran into this while adding a test to https://github.com/zeek/zeek/pull/2718. The way that `AnalyzerConfirmation` in the tunnel analyzers isn't consistent. Teredo and AYIYA call `AnalyzerConfirmation` once the protocol is confirmed good, prior to calling `ForwardPacket`. VXLAN and Geneve call `AnalyzerConfirmation` only after a successful call to `ForwardPacket`, meaning it might not call it at all if the internal packet is something we can't process. GRE doesn't call `AnalyzerConfirmation` at all.

timwoj:
This also adds some additional validation of the sessions passed around during processing of packets that contain Geneve, VXLAN, and AYIYA tunnels, similar to how we were already doing it for Teredo.

Fixes #2720

### DIFF
--- a/src/packet_analysis/protocol/ayiya/AYIYA.cc
+++ b/src/packet_analysis/protocol/ayiya/AYIYA.cc
@@ -13,6 +13,17 @@ bool AYIYAAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packe
 	if ( ! BifConst::Tunnel::enable_ayiya )
 		return false;
 
+	// AYIYA always comes from a TCP or UDP connection, which means that session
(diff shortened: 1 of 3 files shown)


awelzel (approved):
I think that's fine.

I was a bit concerned about AnalyzerViolated and AnalyzerConfirmation are going to a `std::map` for every packet, but that isn't showing up in profiles running against a 1GB genve+vxlan encapsulated pcap. I presume that's because the map only has one or two entries.

PR #2779, opened 2023-02-14
ci/collect-repo-info.py: Recognize git worktrees, too

awelzel:
Benjamin reported ci/collect-repo-info.py failing for him in Git worktrees, relax the .git check.

### DIFF
--- a/ci/collect-repo-info.py
+++ b/ci/collect-repo-info.py
@@ -21,8 +21,16 @@ GIT = "git"
 logger = logging.getLogger(__name__)
 
 
-def git(*args):
-    return subprocess.check_output([GIT, *args]).decode("utf-8")
+def git(*args, **kwargs):
+    return subprocess.check_output([GIT, *args], **kwargs).decode("utf-8")
+
+
+def git_is_repo(d: pathlib.Path):
+    try:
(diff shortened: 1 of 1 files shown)


ci/collect-repo-info.py
```diff
@@ -151,7 +151,7 @@ def included_plugin_dir_conv(v):
     # Attempt to collect info from git first and alternatively
     # fall back to a repo-info.json file within what is assumed
     # to be a tarball.
-    if (zeek_dir / ".git").is_dir():
+    if (zeek_dir / ".git").exists():
```
bbannier, line 154: This fixes the issue I saw.

There is a similar check below, maybe it would make sense to put this into some function `is_git_checkout`, https://github.com/zeek/zeek/blob/252cd33283eac6bbe0721cc158faf302c5468868/ci/collect-repo-info.py#L114

That could also allow us change this the more verbose "Git API" instead of relying on implementation details (which could be easily fooled), e.g., check `git rev-parse --is-inside-work-tree` instead of simply checking for existence of `.git`.

bbannier (approved):

ci/collect-repo-info.py
awelzel: Applied your suggestion.

bbannier (approved):

PR #2781, opened 2023-02-14
Monthly maintenance for compiling scripts to C++

vpax:
This PR has two main changes: (1) fixes a bad memory read when initializing attributes in compiled code, and (2) fixes an order-of-initialization race condition when generating C++ code.  As part of this second, I also added annotations in the initializations to show what Zeek script values/types/etc. in the original scripts they correspond to.

### DIFF
--- a/src/script_opt/CPP/InitsInfo.h
+++ b/src/script_opt/CPP/InitsInfo.h
@@ -116,7 +116,7 @@ public:
 	// The largest initialization cohort of any item in this collection.
 	int MaxCohort() const { return static_cast<int>(instances.size()) - 1; }
 
-	// Returns the number of initializations in this collection that below
+	// Returns the number of initializations in this collection that belong
 	// to the given cohort c.
 	int CohortSize(int c) const { return c > MaxCohort() ? 0 : instances[c].size(); }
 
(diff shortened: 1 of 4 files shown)


src/script_opt/CPP/InitsInfo.h
```diff
@@ -258,9 +258,8 @@ class CPP_CompoundInitsInfo : public CPP_InitsInfo
 class CPP_InitInfo
 	{
 public:
-	// No constructor - basic initialization happens when the object is
-	// added via AddInstance() to a CPP_InitsInfo object, which in turn
-	// will lead to invocation of this object's SetOffset() method.
+	CPP_InitInfo(const IntrusivePtr<Obj>& _o) : o(_o.get()) { }
```
awelzel, line 261: This is without context at all, but mostly wondering if the `o` member could/should be a `zeek::IntrusivePtr<Obj>` rather than a raw pointer. Otherwise the pointer may be invalidated. It is possible to null an intrusive ptr, but maybe I'm missing something else.

awelzel: Hmm, I may have understood. This may also receive `const ID*` rather than a `IDPtr` and that's difficult to convert to an IntrusivePtr.

awelzel (approved):
Should this go into Zeek 5.2? Or should we generally recommend using `master` anyway for the C++ optimization?

vpax:
> Should this go into Zeek 5.2? Or should we generally recommend using `master` anyway for the C++ optimization?

It would be nice if it can go into 5.2.  The compiler is now sufficiently mature that bugs are less common and it would be helpful to have a quasi-stable version so that those using it can do so incidentally rather than tracking `master`.

timwoj:
I forgot to grab this for 5.2.0-rc2 but I just cherry-picked it over so it'll be in the final release.

vpax:
@timwoj Thanks!

PR #2789, opened 2023-02-16
Bump Spicy and spicy-plugin to latest releases.

bbannier:


bbannier:
@timwoj, this should go into 5.2. We can also bump it into `master`.

The `arm_debian11` task kept running into OOM issues when building Spicy. I had to into the CI container, stop the CI build script and manually invoked `make` with lower concurrency to get the build to finish. I then resumed the CI build script which then passed. This also updated ccache so this should hopefully pass more easily in the future unless we do another Spicy bump.

timwoj:
Yeah, merging this into both `master` and `release/5.2` broke the builds on Cirrus because of what you noted above. 

master: https://cirrus-ci.com/build/4938134253731840
release/5.2: https://cirrus-ci.com/build/5847113733505024

awelzel:
> This also updated ccache so this should hopefully pass more easily in the future unless we do another Spicy bump.

PRs don't use the same caching namespace. From the Cirrus docs:

> Tasks for PRs upload caches to a separate caching namespace to not interfere with caches used by other tasks. But such PR tasks can read all caches even from the main caching namespace for a repository.

timwoj:
I bumped the default config on Cirrus to 6 CPUs and 24GB of RAM for the release/5.2 build and all of the tasks were able to complete. We're still using greedy mode, but maybe we should consider disabling that and just going to this as the standard configuration instead. We're clearly hitting some sort of limit in the spicy build where the number of available CPUs (because of greedy mode) outstrips the amount of available RAM for the tasks.

PR #2790, opened 2023-02-16
Move get_relative_seq and get_segment_len to public static methods

### LINKED ISSUE #2788: Exposing TCP utility functions
Somewhere along the line the change introduced in 85acdc14e43a9f69b04e60f74bc10515afb3af1b to expose the utility functions `get_relative_seq` and `get_segment_len` got lost, and the functions are again static/private (now to `packet_analysis/protocol/tcp/TCPSessionAdapter.cc` here: https://github.com/zeek/zeek/blob/5ec72a76989a912cbd967700907b55031af8a51f/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc#L80 and here: https://github.com/zeek/zeek/blob/5ec72a76989a912cbd967700907b55031af8a51f/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc#L178).  For writing plugins that manipulate low-level TCP info, it's useful to have these publicly accessible.

timwoj:
These functions got moved to be private file-level static methods during the work to make the TCP and UDP analyzers as packet analyzers. This PR restores them as public class-level static methods in the TCP session adapter.

Fixes #2788

### DIFF
--- a/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
+++ b/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
@@ -77,7 +77,7 @@ void TCPSessionAdapter::Done()
 	finished = 1;
 	}
 
-static int get_segment_len(int payload_len, analyzer::tcp::TCP_Flags flags)
+int TCPSessionAdapter::get_segment_len(int payload_len, analyzer::tcp::TCP_Flags flags)
 	{
 	int seg_len = payload_len;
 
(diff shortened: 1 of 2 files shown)


awelzel (approved):
Only thought was whether they should be camel-case for consistency, but we're reverting to a prior state where they hadn't been either, so :+1:

timwoj:
@vpax Would it be helpful to backport this to 5.0.x?

PR #2794, opened 2023-02-17
TableVal: Propagate &ordered through copy()

### LINKED ISSUE #2793: &ordered attribute is not propagated through copy()
The following btest reproduces the issue. The `copy(sorted_sset)` should have the same iteration order than the `sorted_sset`, but seems the ordered flag is not propagated.

Ran into this attempting to put an `&ordered` field on a Log record (that is later copied) and was quite confused trying to figure out what's going on.

```
# language/set-iteration.zeek
# @TEST-EXEC: zeek -b %INPUT >out
# @TEST-EXEC: btest-diff out
# @TEST-EXEC: btest-diff .stderr

global sset: set[string];
global sorted_sset: set[string] &ordered;

event zeek_init()
        {
        local i = 0;
        while ( ++i <= 3 )
                {
                add sset[cat(i)];
                add sorted_sset[cat(i)];
                }

        print "sset";
        for ( s in sset )
                print s;

        print "sorted_sset";
        for ( s in sorted_sset )
                print s;

        print "copy(sorted_sset)";
        for ( s in copy(sorted_sset) )
                print s;
        }
```

out:
```
$ cat Baseline/language.set-iteration/out 
sset
2
3
1
sorted_sset
1
2
3
copy(sorted_sset)
2
3
1
```

awelzel:
Copying an &ordered table or set would result in a copy that is not ordered. This seems rather surprising behavior, so propagate the &ordered attribute.

~No other attributes are propagated, but we can't change their behavior easily. &ordered was added with Zeek 5.2, so we can make it behave more intuitively. It wouldn't be a breaking change, as the order had been unspecified before anyway.~

They are (mostly), but not passed through the constructor in `DoClone()`.

Closes #2793


awelzel:
@timwoj , IMO that could be a 5.2 candidate (as &ordered is coming only with 5.2) if there's general agreement that this is the more intuitive behavior.

PR #2798, opened 2023-02-19
MSVC build warning fixes

### LINKED ISSUE #2574: Windows: review warnings from build
The build on Windows results in a rather large number of compiler warnings. We should review these and see if there's any that we can easily fix.

timwoj:
This fixes most of the warnings in the MSVC build. There are two related PRs in the bifcl and binpac repos as well (see below for the linked PRs). The remaining warnings are in the paraglob submodule, in the Supervisor which is unsupported currently on Windows, and in a complicated macro expansion in script_opt which is also completely untested on Windows.

Fixes #2574

### DIFF
--- a/src/IP.cc
+++ b/src/IP.cc
@@ -536,7 +536,8 @@ IPv6_Hdr_Chain::~IPv6_Hdr_Chain()
 	delete finalDst;
 	}
 
-void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, int total_len, bool set_next, uint16_t next)
+void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, uint64_t total_len, bool set_next,
+                          uint16_t next)
 	{
 	length = 0;
 	uint8_t current_type, next_type;
--- a/src/EventTrace.cc
+++ b/src/EventTrace.cc
@@ -722,7 +722,7 @@ std::string DeltaVectorCreate::Generate(ValTraceMgr* vtm) const
 	return std::string(" = vector(") + vec + ")";
 	}
 
-EventTrace::EventTrace(const ScriptFunc* _ev, double _nt, int event_num) : ev(_ev), nt(_nt)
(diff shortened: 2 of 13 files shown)


timwoj:
Now that the binpac and bifcl branches are merged, and I fixed the freebsd 14 failure on CI, this is ready for review.

src/IP.cc
```diff
@@ -536,7 +536,8 @@ IPv6_Hdr_Chain::~IPv6_Hdr_Chain()
 	delete finalDst;
 	}
 
-void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, int total_len, bool set_next, uint16_t next)
+void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, uint64_t total_len, bool set_next,
```
awelzel, line 539: Only wonder if a smaller type or just an `unsigned int` would do? We hopefully never actually need the 64bit here and having it reads a bit like it could be.

But this isn't a well founded argument to change it.

awelzel (approved):

src/IP.cc
timwoj: The warning being fixed here is this one:

```
..\src\Frag.cc(306): note: see reference to function template instantiation 'std::shared_ptr<zeek::IP_Hdr> std::make_shared<zeek::IP_Hdr,ip6_hdr*&,bool,uint64_t&,const zeek::IPv6_Hdr_Chain*&,bool>(ip6_hdr *&,bool &&,uint64_t &,const zeek::IPv6_Hdr_Chain *&,bool &&)' being compiled
```

The length value being passed here is built by Frag.cc as:

```
	// We have it all.  Compute the expected size of the fragment.
	uint64_t n = proto_hdr_len + frag_size;
```

`frag_size` is a `uint64_t`. Basically, it'd end up being a smaller change to make the argument take a `uint64_t`, even if it was a bit oversized, than to try to unwind changing the size of the values passed around in the fragment handler.

PR #2802, opened 2023-02-21
Avoid ccache busting from version bumps

awelzel:
This PR splits the version information from `zeek-config.h` into `zeek-version.h` to avoid ccache busting for everything that includes `zeek/zeek-config.h` just because the version changed. It further avoids pulling `zeek-version.h` into `Plugin.h` unless  we're building an external dynamic plugin. Builtin and in-tree plugins do not need the link time version enforcement.

When building an external plugin, `zeek-config.h` continues to include `zeek-version.h` for compatibility. External dynamic plugins should be unaffected by this change.

This change does break external plugins being built statically into Zeek that rely on `ZEEK_VERSION` or `ZEEK_VERSION_NUMBER`. Chatting with @rsmmr, this is probably okay as it affects plugins we control, or otherwise the users of `--include-plugins` can be considered advanced. Happy to hear other thoughts.

For me locally, this improves ccached build times after a VERSION bump from ~12mins down to 1min realtime. E.g., that is after merging a feature branch and running update-changes. For me, that's a significant workflow improvement. Especially when the feature branch only contained Zeek script changes, the recompilation hit due to the VERSION bump change was/is very annoying.

It will also have a positive effect on CI builds as most merges to master should run faster as the cache is much more effective. We'll spend less CPU cycles which means we can run more CI builds or save energy :-)

Testing was done by very roughly doing the following after modifying `VERSION` by hand.
```
time (rm -rf build && ./my-build.sh  && ninja -j 6 -C build )
```
This is a debug + asan build on Debian 10, gcc10, i7-8565U, ccache on.

Current master gives me this:
```
real    12m10.639s
user    62m52.225s
sys     4m12.432s
```

With this branch, I'm getting:
```
real    1m5.925s
user    3m17.369s
sys     0m34.338s

real    1m0.670s
user    3m7.056s
sys     0m31.742s
```

With these changes, there's still parts of the spicy-plugin being rebuilt and it appears all builtin spicy analyzers are recompiled. I believe there's a similar cache busting issue for the builtin spicy analyzers which we should probably look at when there's more than two analyzers (even now spicyz invocations for me are extra slow, so this may happen sooner rather than later).

Ack/nack?


timwoj (approved):

timwoj:
This looks good to me. I left it unmerged because I saw there's something similar open against spicy. Does that need to be merged ahead of this?

PR #2804, opened 2023-02-21
Update benchmarker Cirrus secrets to use org-level values

timwoj:
I'll need to cherry-pick this into the release branches as well as update broker with the same values, but this PR is checking to make sure this works first.


timwoj:
Ok this is working now. I'm going to go ahead and merge it since there isn't really anything to review.

PR #2811, opened 2023-02-22
Use ensurepip to install pip on FreeBSD

timwoj:
The py39-pip package is sometimes not available on FreeBSD, and using ensurepip is the recommended way to install pip according to the pip documentation. ensurepip should be available in every python installation.


timwoj:
Looks like this needs to be ported over to Broker at least too.

ckreibich (approved):
LGTM. Whenever we have distro-level pip packages I'd stick with using those over `ensurepip` — not sure what's going on with FreeBSD here.

Issue #2730, opened 2023-01-27
File extract when file open on "Windows" platform  should use binary mode to ignore CRLF translate


chuanyi:
https://github.com/zeek/zeek/blob/e9caea9694c586a69efac1772396993642e35f50/src/file_analysis/analyzer/extract/Extract.cc#L21

Correct:
`file_stream = fopen(filename.data(), "wb"); `

Refers:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170

timwoj:
I grepped through the code and there's a few other places that might benefit from this:

- src/iosource/pcap/Dumper.cc, which opens a file in append mode but not as binary
- `write_random_seeds` in util.cc, which opens a file in `w+` but not binary. I don't think that one matters at all.
- `rotate_file` in util.cc, which opens a file in write but not binary. I'm not sure if it would matter though, as this should only be used by log rotation and those files would have already been written in whatever the local line endings are I think.

timwoj:
Fixed via https://github.com/zeek/zeek/pull/2815

PR #2815, opened 2023-02-22
GH-2730: Use binary mode when opening files in Extract file analyzer

timwoj:


awelzel (approved):
Assuming `testing/btest/scripts/base/files/extract/limit.zeek` would've caught this?

timwoj:
> Assuming `testing/btest/scripts/base/files/extract/limit.zeek` would've caught this?

Possibly. I'll see if that was one of the tests currently failing on Windows, but I'd bet it's also failing for other reasons too (paths).

timwoj:
> > Assuming `testing/btest/scripts/base/files/extract/limit.zeek` would've caught this?
> 
> Possibly. I'll see if that was one of the tests currently failing on Windows, but I'd bet it's also failing for other reasons too (paths).

Surprisingly, this change actually fixes that test on Windows. Hooray!

PR #2817, opened 2023-02-23
cirrus: Lower ccache max size and max files settings

awelzel:
After #2802, depending on the task/platform, a mostly cached build on Cirrus CI
is taking 30-60seconds. Eye balling a few tasks, it is now taking longer to
download and unpack as well as check and re-upload the accumulated cache in
the beginning and end of a task.

For Debian 11, this was ~1:20 and 1:10 with a cache size of ~4.2GB. The
default size limit for ccache is 5GB, there's no limit to the number of files.

Running a fresh build on Debian 11, ccache -s indicates the actual
required cache size is ~100MB and the number of files in cache is ~2.7k.
Lower cache size and number of files, such that we don't unnecessarily
accumulate the cache and spend resources on downloading, checking
and re-uploading the cache.

    root@cirrus-ci-task-4907974120964096# ccache -s | grep -E 'files|size'
    files in cache                      2736
    cache size                          96.7 MB
    max cache size                      5.0 GB

PRs have their own cache namespace, so they won't thrash the main's
branch cache. I think main and release share their branch, so we should
not be super aggressive. Though we're probably okay with cache misses on
release once in a while.

Update: After finding mtime based cleanup not working as expected, bumped to 1000M and
introduced an epoch so we can flush the cache if too much data is accumulated and the
random eviction hurts more than just flushing once.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -139,6 +139,10 @@ env:
   CCACHE_DIR: /tmp/ccache
   CCACHE_COMPRESS: 1
 
+  # Ensure reasonable ccache pruning
+  CCACHE_MAXSIZE: 500M
+  CCACHE_MAXFILES: 20000
+
 # Linux EOL timelines: https://linuxlifecycle.com/
 # Fedora (~13 months): https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -435,7 +435,7 @@ docker_build_template: &DOCKER_BUILD_TEMPLATE
   always:
     ccache_cache:
       folder: /tmp/ccache
-      fingerprint_script: echo ccache-$CIRRUS_TASK_NAME-$CIRRUS_OS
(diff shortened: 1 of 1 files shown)


awelzel:
Locally, setting CCACHE_MAXSIZE to something lower than the current cache size will prune the cache whenever new entries are written to the cache. So over the duration of a few runs, that should happen.

Going to up it to 500M to accomodate for release branches.

awelzel:
Hmm, after staring a bit more, I've opened the following on the Cirrus CI side. ccache cleanup is based mtime and mtime is not preserved through the cache instruction, making the cleanup strategy random eviction instead. This, for example, affects the freebsd12 task that has accumulated 5G of cached object files.

https://github.com/cirruslabs/cirrus-ci-agent/issues/277

We could wait and see if the above can be fixed on the Cirrus side, but think we could also introduce an epoch into the cache fingerprint and start with fresh/empty caches now that we make more effective use of it.

awelzel:
I've pushed a follow up commit to flush the caches once and also add a bit of documentation why pruning might not work as expected.

I do believe this is useful as it'll shape off a minute or two on all tasks that have accumulated immense caches.

timwoj (approved):

PR #2818, opened 2023-02-23
ci/collect-repo-info: No hard dependency on git, take two

awelzel:
Collecting information about builtin external plugins also tried to use `git` unconditionally.

### DIFF
--- a/ci/collect-repo-info.py
+++ b/ci/collect-repo-info.py
@@ -129,7 +129,7 @@ def collect_plugin_info(plugin_dir: pathlib.Path):
     except FileNotFoundError:
         logger.warning("No VERSION found in %s", plugin_dir)
 
-    if git_is_repo(plugin_dir):
+    if git_available() and git_is_repo(plugin_dir):
         result.update(git_generic_info(plugin_dir))
 
     return result


0xxon:
I just checked if there is a repo-info.json present in the archive, just to make sure that's not the problem. And - it is present there.

The `ls` output inside the zeek folder looks like this:

```bash
CHANGES		COPYING-3rdparty  NEWS	     VERSION  cmake	 docker		 scripts  zeek-config.h.in  zeek-version.h.in
CMakeLists.txt	INSTALL		  README     auxil    configure  man		 src	  zeek-config.in    zkg-config.in
COPYING		Makefile	  README.md  ci       doc	 repo-info.json  testing  zeek-path-dev.in
```

And the contents of the json file look reasonable.

awelzel:
> I just checked if there is a repo-info.json present in the archive, just to make sure that's not the problem. And - it is present there.

Yes, the problem is that there is another code path in collect-repo-info.py that assumed `git` is available. This PR fixes that.

0xxon:
Ah, I am blind :)

0xxon (approved):

PR #2821, opened 2023-02-24
Fix re-instantiation of analyzers that were just removed.

### LINKED ISSUE #2801: Cannot add child analyzer if same analyzer previously removed
It looks like one cannot add a child analyzer if such an analyzer was previously added and removed again. This seems to be due to `AddChildAnalyzer` just checking whether a child with that tag exists ([here](https://github.com/zeek/zeek/blob/f6872b588ec9db15baf645b2afac2ca027842dbf/src/analyzer/Analyzer.cc#L417-L418) and [here](https://github.com/zeek/zeek/blob/f6872b588ec9db15baf645b2afac2ca027842dbf/src/analyzer/Analyzer.cc#L391-L396)), but not whether it already is in `removing` state. What the function probably should do instead is to only check for existing _active_ children.

We ran into this as part of zeek/spicy-plugin#189, also see the small reproducer there, https://github.com/zeek/spicy-plugin/issues/189#issuecomment-1437176017.

rsmmr:
When an analyzer is being removed from a connection's analyzer tree,
it's at first only marked as removed, but not yet actually deleted.
The problem was that until destruction happens, the analyzer methods
checking presence of an analyzer would continue to indicate that it's
there, even though semantically one wouldn't expect to find it anymore
after that removal operation. That in turn then prevented one from
adding a new analyzer of the same type during that time until
destruction (because we don't allow more than one of each kind at the
same time).

This fix changes the logic searching for analyzers to ignore any
existing ones scheduled for removal. I believe this shouldn't have any
further side effects.

Closes #2801.

### DIFF
--- a/src/analyzer/Analyzer.cc
+++ b/src/analyzer/Analyzer.cc
@@ -5,6 +5,7 @@
 #include <binpac.h>
 #include <algorithm>
 
+#include "zeek/3rdparty/doctest.h"
 #include "zeek/Event.h"
 #include "zeek/ZeekString.h"
 #include "zeek/analyzer/Manager.h"
@@ -480,11 +481,11 @@ void Analyzer::PreventChildren(zeek::Tag tag)
 bool Analyzer::HasChildAnalyzer(zeek::Tag tag)
 	{
 	LOOP_OVER_CHILDREN(i)
-	if ( (*i)->tag == tag )
+	if ( (*i)->tag == tag && ! ((*i)->removing || (*i)->finished) )
 		return true;
 
 	LOOP_OVER_GIVEN_CHILDREN(i, new_children)
-	if ( (*i)->tag == tag )
+	if ( (*i)->tag == tag && ! ((*i)->removing || (*i)->finished) )
 		return true;
 
 	return false;
@@ -492,7 +493,7 @@ bool Analyzer::HasChildAnalyzer(zeek::Tag tag)
 
 Analyzer* Analyzer::FindChild(ID arg_id)
 	{
-	if ( id == arg_id )
+	if ( id == arg_id && ! (removing || finished) )
 		return this;
 
 	LOOP_OVER_CHILDREN(i)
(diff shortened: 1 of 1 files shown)


ckreibich (approved):
Nice unit test :-)

src/analyzer/Analyzer.cc
```diff
@@ -909,3 +910,41 @@ void SupportAnalyzer::ForwardUndelivered(uint64_t seq, int len, bool is_orig)
...
+		a->RemoveChildAnalyzer(b1);
+
+		CHECK(! conn->FindAnalyzer("IMAP"));
+
+		auto b2 = zeek::analyzer_mgr->InstantiateAnalyzer("IMAP", a->Conn());
+		REQUIRE(b2);
+
+		REQUIRE(a->AddChildAnalyzer(b2));
+		CHECK(conn->FindAnalyzer("IMAP"));
+		conn->Done();
```
bbannier, line 948: Ah, this is the magic bit I was missing with the sketch(y) unit test in the original report.

Maybe more as a comment, having to call a `Done` method or else there would be a random segfault isn't ideal in a unit test since e.g., any `REQUIRE` failure above would skip the rest of the test, which would then lead to weird issues elsewhere. A less fragile approach would be to use RAII to clean up the connection automatically, e.g.,

```cpp
		auto conn = std::unique_ptr<zeek::Connection, void (*)(zeek::Connection*)>(
			new zeek::Connection(zeek::detail::ConnKey(t), 0, &t, 0, &p),
			[](auto* c)
			{
				c->Done();
				delete c;
			});
```

bbannier (approved):

src/analyzer/Analyzer.cc
rsmmr: Yeah, was a bit of trial and error until the unit test stopped crashing. Agree in general on the comment, but not sure it's worth it: I'm sure there are more cleanup issues in all of this artificial setting up. it's a broader issue: Zeek's classes & state isn't built for operating outside the full system. And at least it'll cause problem on failures only, so not too worried here.

bbannier:
@rsmmr, do you think this can be backported to 5.0 and maybe also 5.2 which is currently being prepared?

awelzel:
Not sure we should/can do something, but wanted to leave a note that the unit triggered 3 coverity issues:

Screenshot-20230302-111116

Issue #2480, opened 2022-10-13
Cluster startup event


J-Gras:
In a clustered Zeek, there are various use cases for things to wait until the cluster has been successfully started (i.e. all nodes are available). For example, several btests like [notice/suppression-cluster.zeek](https://github.com/zeek/zeek/blob/16601eafa2a11b7a8741e6369050afdec175445f/testing/btest/scripts/base/frameworks/notice/suppression-cluster.zeek) just count nodes/workers before starting their work. It would be extremely handy to introduce a startup event as part of either the cluster framework or the management framework.

With respect to #2479 it might make sense to introduce events for different startup stages as well.

bbannier:
Does [`Management::Controller::API::notify_agents_ready`](https://docs.zeek.org/en/master/scripts/policy/frameworks/management/controller/api.zeek.html#id-Management::Controller::API::notify_agents_ready) provide what you are looking for?

Note that in general one cannot assume any cluster state after receiving such an event; agent could e.g., become available right after the event was triggered -- they might e.g., have become unreachable or be shut down right after you processed the event.  You should allow for that in anything which relies on a certain state by hooking into additional events to update your picture of the cluster.

J-Gras:
> Does [Management::Controller::API::notify_agents_ready](https://docs.zeek.org/en/master/scripts/policy/frameworks/management/controller/api.zeek.html#id-Management::Controller::API::notify_agents_ready) provide what you are looking for?

Thanks for the pointer! If I got it right, this event is used on the agent level of the management framework. I am looking for an event that refers to the cluster node level. These are two separate things, right?

ckreibich:
Thanks for the ticket, Jan! :-) Yes that's right, it's a different level. Let me cover some of the recent discussion: for orderly cluster start/stop, the Management framework is the right level, because it drives those steps, just like `zeekctl` already does it in its world.

A `cluster_init`, on the other hand, looks at home in the cluster framework, at least as long as we keep the model of a fully predefined, "rigid" cluster. This model implies that every node has full understanding of the topology and can therefore declare when all of its peerings are established. With pure Broker that's tricky (because `Broker::peer_added` provides no notion of what the new peer does/is), but the Cluster framework already puts `Cluster::hello` on top of that, pairing node names with Broker IDs.

[...]

PR #2823, opened 2023-02-27
Add `cluster_connected` event

J-Gras:
I gave it a try and implemented a `cluster_init` event to address #2480. It's not particularly pretty, but I think it's okay for the current state of the cluster framework.


testing/btest/Baseline/scripts.base.frameworks.cluster.cluster_init/manager-1..stdout
```diff
@@ -0,0 +1,8 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+### NOTE: This file has been sorted with diff-sort.
+Got cluster_init event
```
timwoj, line 3: I'm confused. The way I read the docs for the events, `node_connected` should be sent once a node has connected to all of its peers and `cluster_init` should be sent after all of the nodes report that they're connected to all of their peers. Shouldn't that means `cluster_init` happens at the end here?

J-Gras: You're right. However, the output is sorted as the sequence of `node_connected` might be arbitrary. Probably better to skip the node names in the output so we don't need the sorting.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -241,6 +254,29 @@ export {
...
+	## broadcasted from the manager once all nodes reported that they have
+	## established all peerings to fellow cluster nodes based on the given
+	## cluster layout.
+	##
+	## .. warning::
+	##
+	##     There is no tracking of cluster node connectivity. Thus, there is
+	##     no guarantee that all peerings still exist at the time of this event
+	##     being raised.
+	global cluster_init: event();
```
awelzel, line 278: `cluster_init()` is very generic and possibly "over-promising": One proposal would to call it: `cluster_connected()` with the same semantics as in this PR (all nodes connected to their peers and broadcasted).

As for the "over-promising", there are other cluster related functionality that may not yet be done, but just seeing `cluster_init()` may well be considered completed, too:
* Intel file distribution
* Config framework distribution
* Anything in third-party scripts related to clusters

Could you see `cluster_connected()`, too?

I'll look at the rest later. Had chatted with @rsmmr about this there's definitely a need for such an event as it has been requested enough in non-testing contexts before.

J-Gras: Actually I started with the pair of `cluster_connected` and `cluster_disconnected`. However, the latter would have added some more complexity. As I couldn't come up with a pressing use case for `cluster_disconnected` I dropped it and switched from "connected" to "init". So I am happy to go back to `cluster_connected`.

awelzel: Sounds good. `cluster_connected` should only happen once and be documented as such, that's easy to understand and implement "correctly". Otherwise a restarting node may confuse script logic that should've been simple an assumed it only happens once.

If a node crashes comes back again, can have a separate `cluster_reconnected` event for that in the future if it's needed (or can also impelement node_down's by hand.

(all Robin's suggestion - none mine :-) )

I'd forego the `cluster_disconnected` event for now.

bbannier: > `cluster_connected` should only happen once and be documented as such, that's easy to understand and implement "correctly". Otherwise a restarting node may confuse script logic that should've been simple an assumed it only happens once.

I am not sure I follow this reasoning. I would have assumed that any non-test code which cares about a global *cluster ready* event would also care about the cluster not being ready and take some action on that. Leaving users in the illusion that cluster state is static after startup might actually cause more confusion in the end, not less.

My gut feeling would have been that we'd want to explicitly document how this event can be meaningfully used and what caveats it has (ideally with workarounds for e.g., nodes coming and going).

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -262,35 +298,52 @@ export {
 	global nodeid_topic: function(id: string): string;
 }
 
-global active_worker_ids: set[string] = set();
+# Track the names of fellow nodes each node should connect to
+global nodes_awaited: set[string] = set();
```
awelzel, line 302: That name is pretty confusing: It's the set of nodes we're expecting to see a `node_up` for, correct?

Some suggestions would be `nodes_expected`, `nodes_expected_up`, maybe `nodes_awaiting_up`.

"nodes awaited" to me means  the set of "nodes we have waited for" which is the opposite of what it's used for, but may come from a previous implementation.

> Track the names of fellow nodes each node should connect to

Track the names of peers we're expecting to connect/peer ?

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -341,11 +396,25 @@ event Cluster::hello(name: string, id: string) &priority=10
...
-		add active_worker_ids[id];
-		worker_count = |active_worker_ids|;
-		}
+		worker_count = |active_node_ids[WORKER]|;
+	}
+
+event Cluster::node_up(name: string, id: string) &priority=-10
+	{
+	local mgr = nodes[node]?$manager ? nodes[node]$manager : "";
+	if ( name !in nodes_awaited && name != mgr )
```
awelzel, line 410: That guard is a bit confusing here. Is it actually needed? A `delete` for a non-existing element is a noop.

It may need a guard `|nodes_awaited| == 0` and early return to avoid triggering the events multiple times.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -366,11 +435,12 @@ event Broker::peer_lost(endpoint: Broker::EndpointInfo, msg: string) &priority=1
 			Cluster::log(fmt("node down: %s", node_name));
 			delete n$id;
 
+			local active_ids = active_node_ids[n$node_type];
+			delete active_ids[endpoint$id];
+			active_node_ids[n$node_type] = active_ids;
```
awelzel, line 440: At the risk of missing something intricate: Is this "just" to make shorter lines, or did one line not somehow do what it should?
```
delete active_node_ids[n$node_type][endpoint$id]
```

awelzel (commented):
Sorry for being verbose, but hope that helps a bit.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -44,6 +44,10 @@ export {
 	## time machine nodes in a cluster.  Used with broker-enabled cluster communication.
 	const time_machine_topic = "zeek/cluster/time_machine" &redef;
 
+	## The topic name used for exchanging messages that are relevant to
+	## all nodes in a cluster.  Used with broker-enabled cluster communication.
+	const all_nodes_topic = "zeek/cluster/all-nodes" &redef;
+
```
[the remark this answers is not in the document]
J-Gras: That is still correct as connections in a cluster do not form a tree topology. My assumption was that forwarding won't be used in a cluster. Actually, the forwarding feature even vanished from the broker documentation ([old version](https://docs.zeek.org/projects/broker/en/v2.0.2/comm.html#forwarding)).
I am okay with the config framework's approach but I guess it is hard to maintain. Actually the config framework is missing time machine nodes. How about introducing a set `Cluster::broadcast_topics` which can be looped to add all relevant topics?

Another fix would be to make broker not forward every topic or allow to exclude topics from forwarding.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -212,6 +216,15 @@ export {
...
 
+	## Returns the number of nodes defined in the cluster layout for a given
+	## node type.
+	global get_node_count: function(node_type: NodeType): count;
+
+	## Returns the number of nodes per type, the calling node is currently
+	## connected to. This is primarily intended for use by the manager to find
+	## out how many nodes should be responding to requests.
+	global get_active_node_count: function(node_type: NodeType): count;
+
```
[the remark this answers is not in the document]
J-Gras: > Could we do without them, or do without the former?

Nope, that's a feature needed elsewhere ;)

> "active" is a new term that may be very wide in interpretation.

Actually this is just a generalization of the original `active_worker_ids` below. Exposing this as public API will be useful. As `active_worker_ids` wasn't exposed, it should be okay to rename.

> Also wondering whether we could return more than just the count [...]

In that case we should probably just expose `active_node_ids` or `ids_of_nodes_that_are_in_state_up` ;) Active nodes are tracked based on their name. If more infos are required, the name could be used to index `Cluster::nodes`.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -241,6 +254,29 @@ export {
 	## locally whenever a connected cluster node becomes disconnected.
 	global node_down: event(name: string, id: string);
 
+	## When using broker-enabled cluster framework, this event will be sent to
+	## the manager whenever a cluster node has established all peerings to
+	## its fellow cluster nodes based on the given cluster layout.
```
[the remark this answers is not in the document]
J-Gras: There are two reasons I came up with a new term here:
1. The term "peer" is already used by broker. A broker peer and a "cluster peer" are two different things and shouldn't be confused.
2. The term "peer" implies equality ("peer: one that is of equal standing with another"). This is actually not the case in a cluster as fellows are of different types.

> "a cluster node has established peerings with all its peers"

I would argue that this is highly confusing by making it easy to mix up different levels of abstraction. A "peering" at the cluster level will be considered "established" once the hello messages are exchanged.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -241,6 +254,29 @@ export {
...
+	##
+	## .. warning::
+	##
+	##     There is no tracking of cluster node connectivity. Thus, there is
+	##     no guarantee that all peerings still exist at the time of this event
+	##     being raised.
+	global node_connected: event(name: string, id: string);
```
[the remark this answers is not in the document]
J-Gras: How about `all_fellows_connected`?

scripts/base/frameworks/cluster/main.zeek
J-Gras: > That name is pretty confusing: It's the set of nodes we're expecting to see a node_up for, correct?

No, eventually we should see a `node_up` for nodes that we connect to as well as nodes that connect to us. `nodes_awaited` are nodes that we await, i.e. the local node connects to (see `setup-connections.zeek`).

scripts/base/frameworks/cluster/main.zeek
J-Gras: There will be more `node_up`s than nodes awaited (see comment above) and `node_connected` should be triggered only once (the documentation wasn't updated, sorry). Furthermore, my thought was that a node could be connected to all its fellows before the manager is up.

scripts/base/frameworks/cluster/main.zeek
J-Gras: > delete active_node_ids[n$node_type][endpoint$id]

This won't do what you expect it to do. Maybe this creates a temp copy of `active_node_ids[n$node_type]`?

scripts/base/frameworks/cluster/nodes/manager.zeek
```diff
@@ -23,3 +23,40 @@ redef Log::default_rotation_interval = 24 hrs;
 ## Use the cluster's delete-log script.
 redef Log::default_rotation_postprocessor_cmd = "delete-log";
 @endif
+
+module Cluster;
+
+global nodes_started: set[string] = set();
```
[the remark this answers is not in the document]
J-Gras: Long but descriptive :)

scripts/base/frameworks/cluster/nodes/manager.zeek
```diff
@@ -23,3 +23,40 @@ redef Log::default_rotation_interval = 24 hrs;
...
+
+event zeek_init()
+	{
+	Broker::auto_publish(all_nodes_topic, Cluster::cluster_connected);
+	}
+
+event Cluster::node_connected(name: string, id: string)
+	{
+	add nodes_started[name];
+	if ( |nodes_started| == |Cluster::nodes| )
```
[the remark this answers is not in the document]
J-Gras: Actually `node_connected` will be emitted only once per node (the documentation wasn't updated) so no need for additional guarding here.

> If a node crashed and re-starts to join a previously fully-connected cluster, it (and only it) should probably receive a new Cluster::cluster_connected event once it connected to all its peers.

Let's take this discussion to another PR :)

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -262,35 +298,52 @@ export {
...
-		local n = Cluster::nodes[name];
-
 		if ( n$node_type != node_type )
 			next;
 
 		rval += NamedNode($name=name, $node=n);
 		}
 
-	return rval;
+	return sort(rval, nncmp);
```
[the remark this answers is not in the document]
J-Gras: For me that turned out to be more LOC and harder to read. Maybe I got the syntax wrong?

scripts/base/frameworks/cluster/nodes/manager.zeek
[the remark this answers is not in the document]
vpax: (Nit, no need to initialize to an empty `set()`, that happens by default.)

scripts/base/frameworks/cluster/main.zeek
awelzel: As in DM, the `delete` here will actually work, it's the `add` part for the same table that caused confusion to the &default usage not assigning (https://github.com/zeek/zeek/discussions/2668)

```
	local active_ids = active_node_ids[n$node_type];
	add active_ids[id];
	active_node_ids[n$node_type] = active_ids;
```

IMO (but IMO only), dropping `&default` and explicitly initializing would be less surprising
```
if ( n$node_type !in active_node_ids )
    active_node_ids[n$node_type] = set();

add active_node_ids[n$node_type][id]
```

scripts/base/frameworks/cluster/main.zeek
awelzel: I should've left the part in my comment about not being a native speaker ;-) I still find it confusing, but this might be me.

scripts/base/frameworks/cluster/main.zeek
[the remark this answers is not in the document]
awelzel:
```
sort(rval, function(l: NamedNode, r: NamedNode): int { return strcmp(l$name, r$name); });
```

ckreibich:
I've worked my way through the code and would like to think this one over for a bit still, so please don't merge just yet even if you're happy with everything. I promise to be quick.

Up-leveling from the detail, a few comments:
- I like `cluster_connected` as the name, because it's true that symmetry with `zeek_init` (the original motivator for `cluster_init`) does not apply — `zeek_init` sets up things _before_ Zeek gets going, so `cluster_init` would suggest that we do things prior to the cluster bootstrapping, just the opposite of what the event is to convey.
- Yes, this event is technically imperfect in the sense that the cluster might be failing at the time it's still in the progress of communicating `cluster_connected`, etc. But, a normal Zeek cluster is stable: nodes come up, the cluster topology converges, and the topology does not change going forward (if we deviate from this model, we will need to rethink a lot more than just this event). Sure, nodes will on occasion OOM, crash, etc. But the benefit of having this event outweighs the conceivable situation of constant flapping etc, since most of the time it will provide reliable execution of an activity that currently is _always_ messy.
- I do not think we should trigger `cluster_connected` only once. The name implies that the cluster can enter a connected state repeatedly, and learning that the cluster has re-connected is valuable. It's not hard to write code in a handler that indeed only runs once if that's desired.
- I agree that we should not attempt to implement disconnect events at this time.
- There was discussion (alas I don't recall where) of a corresponding `cluster_done` event. The difference to a `cluster_disconnect` would be that it's final — you'd learn that the cluster's presence can no longer be assumed and won't return. I don't immediately see applications for this that can't also be implemented via `zeek_done`. It seems more important to coordinate clean cluster shutdown, which (mirroring clean cluster bootstrap) is something for zeekctl and the management framework to sort out, not the cluster framework. Still, I wanted to flag this in case somebody has thoughts on this.

scripts/base/frameworks/cluster/main.zeek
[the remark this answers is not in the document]
awelzel: Maybe `node_fully_connected` is differentiating and meaningful enough? Bit of a compromise.

I see your argument with "fellows" vs "peers", but unsure the name should show up in event names just yet.

scripts/base/frameworks/cluster/main.zeek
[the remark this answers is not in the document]
awelzel: The suggestion did confuse things and thanks for the clarification. Still not sure about fellows, but resolving this.

awelzel:
> I do not think we should trigger cluster_connected only once. The name implies that the cluster can enter a connected state repeatedly, and learning that the cluster has re-connected is valuable. It's not hard to write code in a handler that indeed only runs once if that's desired.

There was and will be a bit of offline communication about this. Raising the event just once make things easier (and follow up work could introduce a `cluster_reconnected`).

Even with raise once semantics, the following should be given consideration:
* a non-manager node crashes and comes back: It should likely raise a `cluster_connected()` event so we can guarantee it's always happening some time after startup for an individual node within a cluster.
* The manager crashes: in the best case it somehow learns from other nodes that it was the one that crashed and raises `cluster_connected` for itself to make progress.

scripts/base/frameworks/cluster/main.zeek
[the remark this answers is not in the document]
J-Gras: To be honest, I am not a huge fan of "fellows" as well. However, I am keen to make sure the difference becomes apparent.

scripts/base/frameworks/cluster/main.zeek
J-Gras: Yep, I got confused here. I like the explicit initialization. Seems indeed easier to understand.

scripts/base/frameworks/cluster/nodes/manager.zeek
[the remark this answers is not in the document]
J-Gras: I removed the initialization and, given the above discussions, went with `nodes_fully_connected` here. Hope that's okay.

ckreibich:
A few follow-ups to today's discussion:
- I think instead of "awaited" we settled on "pending".
- Re. the "fellow" naming matter, a few more suggestions — we mean the things `Broker::peer()` establishes a connection to, so connectees, listening peers, or out-peers (from in/out edges in a directed graph) might work.
- It'd be good to have some traceability via `Cluster::log()` of a node becoming fully connected and the manager triggering `cluster_connected`, for troubleshooting.

J-Gras:
I've done some cleanup and added new test cases. In this context I have one follow-up to make sure I am on the right track: The consesus, as I understood, is that we do not yet want to implement a `cluster_reconnect` event (which would also need to track disconnects). Although one has to assume that cluster nodes restart from time to time, this would move us towards a world of elastic clusters, which most likely need more fundamental re-thinking of the cluster framework.

This leaves us with the question how to handle restarting nodes wrt. `cluster_connected`:
1) Resend `cluster_connected` to the restarted nodes as Arne suggested. To distinguish between "the cluster just started" and "the local node has restarted" I would add a `redispatched` flag to the event signature. However, receiving a redispatched `cluster_connected` could mean "everything". The cluster could be far from a connected state as other nodes might have failed as well and did not reconnect.
2) Send the event only once. As Arne pointed out, if nodes would rely on `cluster_connected` to do initalization, restarting nodes would never reinitialize. That seems unfortunate. One could argue this is by design and the event must not be used for these types of initialization tasks (`node_up` might be a better choice in these cases anyway).

If we decide to go with option 2, I would agree that moving the event into a new namespace like `Cluster::Testing/Experimental/Static` might be a good idea. In both cases I was wondering whether it would make sense to rename the event to `cluster_started`.

awelzel:
> I would add a redispatched flag to the event signature. However, receiving a redispatched cluster_connected could mean "everything". However, receiving a redispatched cluster_connected could mean "everything". The cluster could be far from a connected state as other nodes might have failed as well and did not reconnect.

The re-connecting case was mostly for the node that crashed to proceed in the same way as if it had seen the initial `cluster_connected`. If we don't send `cluster_connected`, but a script uses it to, e.g., call `continue_processing()`, a restarting node would be stuck.

I would not add a parameter indicating the difference though: If someone acts/cares for it, they likely should not be using `cluster_connected` to begin with. If they start to wonder what happens if a nodes restart arbitrarily, they likely realize they should not be using `cluster_connected` either.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -241,6 +257,36 @@ export {
...
+	##
+	## redispatched: If true, the node has previously signaled that it is fully
+	##               connected. This may happen in case the manager restarts.
+	##
+	## .. warning::
+	##
+	##     There is no tracking of cluster node connectivity. Thus, there is
+	##     no guarantee that all peerings still exist at the time of this event
+	##     being raised.
+	global node_fully_connected: event(name: string, id: string, redispatched: bool);
```
awelzel, line 276: > 	redispatched: If true, the node has previously signaled that it is fully
>	               connected. This may happen in case the manager restarts.

The implementation looks like it's currently passing `seen_cluster_connected` and that seems needed, but the docstring disagrees. I like "seen_cluster_connected" over "dispatched", too, as the current `redispatched` is closer to "sending a second time" which isn't the same (and indeed isn't good enough for what it's being used for).


I'm also forth and back if we should expose `node_fully_connected` as is. It seems a nice primitive to have, but  the `redispatched / seen_cluster_connected` might be confusing to the "local user". I think it's also raised multiple times locally for a restarting manager.

This could be fixed by separating the local event and the one we send to the manager and letting go of auto_publish.

Essentially, the same motivation as for `cluster_connected`: Ensure it is raised once as a local event and no extra info that would allow advanced logic. But I might be too biased here.

scripts/base/frameworks/cluster/nodes/manager.zeek
```diff
@@ -23,3 +23,53 @@ redef Log::default_rotation_interval = 24 hrs;
...
+
+event zeek_init()
+	{
+	fully_connected_nodes_pending = table_keys(nodes);
+	}
+
+event Cluster::node_fully_connected(name: string, id: string, redispatched: bool)
+	{
+	# If a node redispatched the event, the cluster has already been seen connected.
+	seen_cluster_connected = redispatched ? T : seen_cluster_connected;
```
awelzel, line 39:
```suggestion
	seen_cluster_connected = redispatched || seen_cluster_connected;
```

scripts/base/frameworks/cluster/nodes/manager.zeek
```diff
@@ -23,3 +23,53 @@ redef Log::default_rotation_interval = 24 hrs;
...
+
+	delete fully_connected_nodes_pending[name];
+	if ( !seen_cluster_connected && |fully_connected_nodes_pending| == 0 )
+		{
+		event Cluster::cluster_connected();
+
+		for ( topic in broadcast_topics )
+			Broker::publish(topic, Cluster::cluster_connected);
+		}
```
awelzel, line 48: Is this missing the targeted `Cluster::cluster_connected` publish to the node_topic because it's not yet clear whether/how to do this?

E.g, something like this for an individual node that comes after cluster_connected happened?
```
if ( seen_cluster_connected && !redispatched)
   {
   Broker::publish(Cluster::node_topic(name), Cluster::cluster_connected)
   }
```
And separately, a manager raising `Cluster::cluster_connected()` for itself once if it finds someone else saw `cluster_connected` before. Think the `||` suggestion above could be changed to an `if` where this happens then.


Separately wondering a bit if an explicit `did_publish_cluster_connected` global known to the manager here would make things easier to read.

scripts/base/frameworks/cluster/main.zeek
J-Gras: So this is essentially a vote to rename `redispatched` to `seen_cluster_connected`, correct?

Regarding splitting the events: I am not a fan of wrapping users in cotton. Exposing it all provides greater flexibility. If needed, one could just check `name` or `seen_cluster_connected`. Writing Zeek scripts I never found myself grumbling that some information I am not interested in was exposed. Unfortunately, the inverse has happened here and there.

scripts/base/frameworks/cluster/nodes/manager.zeek
J-Gras: The current state just implements variant 2, because I had a slight preference for that approach ;)

> Separately wondering a bit if an explicit did_publish_cluster_connected global known to the manager here would make things easier to read.

I think if we go with `cluster_started`, `seen_cluster_connected` becomes `is_cluster_started` and things should be readable without introducing more variables.

J-Gras:
> If someone acts/cares for [whether a node has restarted], they likely should not be using `cluster_connected` to begin with. If they start to wonder what happens if a nodes restart arbitrarily, they likely realize they should not be using `cluster_connected` either.

I am not sure I can follow here. The above would be an argument for variant 2: Send the event only once. If restarts matter `cluster_connected` is the wrong choice.

In contrast, variant 1 would allow to use the event for initialization tasks but may surprise users that do not consider restarts. One way or the other, you allow users to shoot themselves in the foot.

That being said, while both implementations would make sense to me, the name `cluster_connected` doesn't fit any of them. From my perspective, non of the options yields behaviour one would intuitively expect from a `cluster_connected` event anymore. I guess `cluster_started` or something similar would be a better fit. In particular, if it makes people think "What means started?" ;)

ckreibich:
Sorry for the silence from my side, I had a bunch of distractions to deal with last week. My understanding after our call was that we'd go for option 2 — send `cluster_connected` only once, with sufficient robustness in the manager to understand when it has restarted, so it doesn't re-broadcast the event.

My original thought was that it would be _easier_ to send the event repeatedly when the cluster fully connects again, but our discussion revealed that that's not the case. I also like the idea of leaving more sophisticated scenarios up to the nodes and `node_up`, because I could easily see use-cases of `cluster_connected` focusing on the manager. We can iterate further as use-cases emerge.

Sound good?

scripts/base/frameworks/cluster/main.zeek
J-Gras: Turned out that `redispatch` just didn't click so we settled on `resent`.

J-Gras:
Sounds good! I've just incorporated Arne's latest style feedback.

For completeness (and maybe documentation) let me try to summarize where we landed (@rsmmr, fyi):
* When freshly started, each cluster node tracks connection establishments it is responsible to initiate to other cluster nodes.
* As soon as a node has established all its connections and the manager is available, the node signals `node_fully_connected` to the manager.
* Once all nodes signalled fully connected, the manager raises `cluster_connected` locally and broadcasts the event.
* If the manager restarts, `cluster_connected` will neither be broadcasted again nor raised locally on the manager.
* Likewise, if any other node restarts, `cluster_connected` will neither be broadcasted nor raised locally on the restarted node.

Finally, there are only two questions left:
* Should we move the event to a new namespace like `Cluster::Testing/Experimental/Static`?
* Should we rename the even to `cluster_started` leaving `cluster_connected` for a more elaborate implementation?

timwoj:
> my vote is on putting it into some Cluster::Testing module

`Cluster::Testing` sounds more like a module for things that are used to test clusters. Reading back that doesn't sound like the intention. If this is something that we're adding but we're not really sure is stable yet, I agree with @J-Gras's suggestion of `Cluster::Experimental`.

timwoj:
@J-Gras Is this just waiting for reviews at this point?

J-Gras:
> @J-Gras Is this just waiting for reviews at this point?

Yes, I think we just need a decision on the naming aspects.

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -0,0 +1,85 @@
+##! Experimental features of the Cluster framework.
+
+@load base/frameworks/cluster
+
+module Cluster::Experimental;
+
+export {
+	## When using broker-enabled cluster framework, this event will be sent to
+	## the manager once a cluster node has set up all its outgoing connections
```
ckreibich, line 9: A nit-picky suggestion here — it's not just that the node has set up outgoing connections (that to me would be `Broker::peer()`, but these have also checked in via `Cluster::node_up()`.

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+hook Cluster::connect_node_hook(connectee: Cluster::NamedNode)
+	{
+	add connectees_pending[connectee$name];
+	}
+
+event Cluster::node_up(name: string, id: string) &priority=-10
+	{
+	local mgr = Cluster::nodes[Cluster::node]?$manager ? Cluster::nodes[Cluster::node]$manager : "";
```
ckreibich, line 62: I find this subtle ... could you add a comment that explains the use of `$manager`?

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -0,0 +1,55 @@
+##! This script provides experimental features in the cluster manager.
```
ckreibich, line 1: Could clarify here that it is only loaded on the manager node, so one doesn't look for `@if ( Cluster::local_node_type() == Cluster::MANAGER )`.

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -0,0 +1,85 @@
...
+	##
+	## resent: If true, the node has previously signaled that it is fully
+	##         connected. This may happen in case the manager restarts.
+	##
+	## .. warning::
+	##
+	##     There is no tracking of cluster node connectivity. Thus, there is
+	##     no guarantee that all peerings still exist at the time of this event
+	##     being raised.
+	global node_fully_connected: event(name: string, id: string, resent: bool);
```
ckreibich, line 24: This is suuuper picky, but "resent" suggests something else to me, that happened in the past, as opposed to flagging that it is this event that is being re-sent. It also means to despise. :-) How about "resending", "repeat" or "retransmit"?

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -0,0 +1,55 @@
...
+global fully_connected_nodes_pending: set[string];
+
+event zeek_init()
+	{
+	fully_connected_nodes_pending = table_keys(Cluster::nodes);
+	}
+
+event node_fully_connected(name: string, id: string, resent: bool)
+	{
+	# If a node has already seen the cluster connected, the manager most likely
```
ckreibich, line 17: Just a suggestion to help make the leap from "if a node has seen" to `resent` being `T`:
```suggestion
	# If a node resends this event, it has already seen the cluster connected. The manager most likely
```

ckreibich (changes requested):
This looks great. I only have two real requests:
- A NEWS entry. I really like the explicit experimental namespacing, so we could introduce a new section in the file for each release, for experimental stuff.
- On the btests, a quick comment at the beginning of each test what it's about. It's just hard to piece this back together a few months in, or if unfamiliar with the tests.

The rest are nits that you're welcome to ignore or that would have helped me understand the code a bit more quickly.

Thank you very much for all this!

J-Gras:
I incorporated the latest feedback, added documentation to the tests, updated the NEWS, rebased everything on master and  squashed/rewrote the history. Hopefully, this should be ready to merge now.

ckreibich (approved):
Thank you very much, Jan!

PR #2824, opened 2023-02-27
Simplify btests using `cluster_init` event

J-Gras:
This is just to showcase the benefits of #2823.


J-Gras:
Yes, I'll update this PR :)

testing/btest/scripts/base/frameworks/sumstats/basic-cluster.zeek
```diff
@@ -10,7 +10,7 @@
 # @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-sort btest-diff manager-1/.stdout
 
 @load base/frameworks/sumstats
-@load base/frameworks/cluster
+@load policy/frameworks/cluster/experimental
```
awelzel, line 13: I believe the more common style would be to add experimental as additional `@load` and not rely/imply transitive @loads, but cc'ing @ckreibich / @0xxon .

Sumstats also loads base/frameworks/cluster, so it was already redundant, but possibly on purpose? :man_shrugging:

J-Gras: Yeah, I thought in this case it's appropriate given that experimental is a "submodule" of Cluster.

J-Gras:
> Seems reasonable to have them outside of relying on tests for functionality in an experimental namespace. Can revert/skip the commit during merge.

Anticipated and that's why I made this a separate commit... 😄 

> As an aside, the "Testing" namespace for this use-case would've been fitting, "Experimental" looks quite odd in the tests.

"Testing" will look odd in other use cases 😉

testing/btest/scripts/base/frameworks/sumstats/basic-cluster.zeek
awelzel: Mostly the base vs policy difference made me stop. I've merged the version as you had it now and if there are other opinions we'll adapt.

PR #2826, opened 2023-02-27
Bump zeekctl and add NEWS entries

awelzel:
These are added to the 5.2 section of NEWS. This is for backporting into release/5.2.


awelzel:
@timwoj - not sure how to do these things? Is it okay if this goes into master and then being cherry-picked into release/5.2 afterwards? Other preferences?

timwoj (approved):

PR #2829, opened 2023-02-27
Force usage of a different debian mirror for Docker image builds

timwoj:
Debian by default uses the deb.debian.org host to automatically determine the best mirror to connect to for packages. Recently our CI setup has been fighting a ton of timeouts from that host which building the images for and executing the cluster tests. This PR switches the host for apt to use the mirror at OSUOSL. I opted for that one specifically because the Zeek project already has a relationship with that group in case anything else comes up.

### DIFF
--- a/docker/builder.Dockerfile
+++ b/docker/builder.Dockerfile
@@ -3,6 +3,8 @@
 # Layer to build Zeek.
 FROM debian:bullseye-slim
 
+RUN echo 'Acquire::Retries "3";' > /etc/apt/apt.conf.d/80-retries
+
 # Configure system for build.
 RUN apt-get -q update \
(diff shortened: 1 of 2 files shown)


awelzel (changes requested):
As in DM, we're distributing these images to Docker Hub, not sure we want to change the mirror there.

Maybe the following helps enough?
```
echo 'Acquire::Retries "3";' > /etc/apt/apt.conf.d/80-retries
```

https://askubuntu.com/a/1107071

timwoj:
I don't see any retries in the build that ran but it likely doesn't hurt to just merge the retry change in. If anything we can keep an eye on it.

awelzel (approved):

PR #2831, opened 2023-02-27
Default known stores to false to avoid Broker stores

### LINKED ISSUE #2803: Change defaults in Known scripts to not use broker stores
We have seen a number of clusters recently having various performance issues that are fixed by changing these settings:

```
redef Known::use_host_store = F;
redef Known::use_service_store = F;
redef Known::use_cert_store = F ;
```

I'm not sure if it's large clusters only, or if a small cluster will also show issues.  Recently on one system with 30 workers I noticed a 20-30% CPU reduction on each worker as well as a drop of over 100% across the manager and proxies.

timwoj:
A naive test with 2009-M57-day11-18 says that this is about a 3% time savings at the cost of about 0.2% extra memory usage, assuming you enable all three of hosts, services, and certs.

Tested with a release build, loading the three policy files, averaged over 10 runs after a warmup run to put everything in cache.

master:
Time: 6.70s
Memory: 118364.4 MB

This branch:
Time: 6.47s
Memory: 118594.4

I can run a longer test on os-perf-2 but this seems fairly promising.

Fixes #2803


JustinAzoff:
Is that using pcap replay on a cluster or `zeek -r` on a pcap?  The performance difference would potentially be much larger on a cluster.

awelzel:
Mostly after skimming the scripts and as I mentioned it somewhere in Slack:

When in broker store mode, these `known-` scripts do not make any attempt at rate-limiting calling `Broker::put_unique` for unproductive outcomes (essentially duplicated information)

Given there are event handlers for `connection_established`, `ssl_established`, `connection_state_remove`, I could see a `Broker::put_unique()` for every individual "connection" to a local/tracked DNS server twice, and for every connection to a local/tracked ssl service twice as well.

Without broker stores enabled, the `publish_hrw()` calls are rate limited to avoid unproductive event forwarding to proxies.

---

I think the change is good, but possibly needs follow-up to improve these scripts, or generally discuss about patterns for rate-limiting in scripts.

timwoj:
> Is that using pcap replay on a cluster or `zeek -r` on a pcap? The performance difference would potentially be much larger on a cluster.

Just with `-r`. I agree that it will be better with a cluster test, but this was an easy way to test that it was doing anything at all.

awelzel (approved):
I think this is okay with a big entry to NEWS? Maybe ask for a second approval?

The one caveat here is that in a cluster configuration without proxies, I believe these scripts become non-functional, while they would've worked previously. If we can't accept that, then this needs some more thinking.

PR #2832, opened 2023-02-28
X.509: expose the signature type inside the tbs certificate

0xxon:
This change exposes the signature tyope inside the signed portion of an X.509 certificate. In the past, we only exposed the signature type that is contained inside the signature, which is outside the signed portion of the X.509 certificate.

In theory, both signature fields should have the same value; it is, however, possible to encode differing values in both fields. The new field is not logged by default.


awelzel (approved):
Wonder if this might be worth a `fullci` labeled run in case SSL is acting differently on older distros, but should be seen after merge to master, too.

PR #2836, opened 2023-03-02
Bump spicy-plugin to v1.3.24.

bbannier:


timwoj:
We're done with the 5.1 cycle now that 5.2 is out. Do we still need this?

bbannier:
> We're done with the 5.1 cycle now that 5.2 is out. Do we still need this?

if nothing we never cut a release from in the 5.1 series again this is not needed.

timwoj:
Closing this because we're unlikely to do another 5.1 release.

PR #2839, opened 2023-03-02
ScriptValidation: Make break/next a warning until Zeek 6.1

awelzel:
The ja3 package uses next instead of return and triggers the new errors with Zeek 5.2. That seems somewhat bad.

In case we want to do a 5.2.1 that makes this a warning, this would be change to do so.

---

This is mostly an RFC, I'd be much more happy we just keep the error and move on.

The most recently merged PR in the ja3 repo is from March 2021, so I'm not so sure giving more time will necessarily help.
https://github.com/salesforce/ja3/pulls?q=is%3Apr+is%3Aclosed

@ckreibich, thoughts?


timwoj (approved):
I agree with this. It definitely is a breaking change, and if we have the opportunity for it to go through a deprecation cycle of sorts then we should do that.

vpax:
> It definitely is a breaking change, and ...

I disagree that it's actually a "breaking change".  It's flagging code that has always been incorrect (and undefined), not changing something that _used to be_ defined as valid to no longer work.  If we view that sort of change as breaking, then it will apply any time we tighten up error-checking.

That said, for this particular change I don't have a problem with downgrading it to a warning for now.

ckreibich:
My initial take on this was harsher — that this package should get fixed, not least because the change is straightforward. It would be one thing if we broke it in a patch release, but nobody should update to 5.2 without some testing, and if the JA3 package is indeed unmaintained, we'll just kick the can down the road.

If it indeed turns out to be unmaintained, we should have a discussion about how/whether to maintain JA3 functionality ourselves — I'd lean toward our own package since the Salesforce one could be improved in a few ways. (The alternative implementation I'm aware of, https://github.com/hosom/bro-ja3, is also outdated.)

For now, I'd wait a bit longer to see if somebody picks this up on the Salesforce side.

timwoj:
@awelzel Did you have anything else to do here or can this be merged (it's still a draft)?

awelzel:
Mainly waiting for @ckreibich to make a call. I'll mark it ready, but there's activity on the salesforce/ja3 to figure out how to fix it, just doesn't seem easy.

vpax:
> ... but there's activity on the salesforce/ja3 to figure out how to fix it, just doesn't seem easy.

Weird - I believe the fix is to just change it to `return`, IIRC.

timwoj:
I still think we should merge this here, even if the ja3 package is fixed, as there may be other packages out there that we're not aware of that are hitting this.

ckreibich:
I got an update from Salesforce — they're looking at possibly taking until the end of March to turn the JA3 PR around. Therefore my take is that we merge this, and start exploring options for a more speedily maintained version of JA3.

ckreibich (approved):

PR #2841, opened 2023-03-03
Add pcap_file option to supervised nodes.

awelzel:
This allows to start Supervised nodes with a pcap_file argument rather than interface.

This is based on changes from @J-Gras.


awelzel:
I've left out extending `Cluster::Node` with `pcap_file`, but now wonder if it should be there for consistency (but at the same time am not convinced that `$interface` should've ever made it into `Cluster::nodes`). Thoughts welcome.

J-Gras:
Just for the record, I used `$interfaces` to implement https://github.com/J-Gras/add-interfaces. Not sure whether people somehow rely on this. All I can tell is that the repo has currently 106 unique cloners.

0xxon (approved):
I am not an expert on this part of the codebase - but it does look good to me.

0xxon:
Is this something that requires a documentation change/addition? It looks a bit like that might be worthwhile here.

ckreibich (commented):
This all looks good, thank you. If somebody provides both interface and pcap, is that covered the same way as if somebody would do so on the command line? If not, it would be good to protect against it. `Supervisor::Create()` already does some sanity-checking.

ckreibich:
@0xxon Yeah it will, though Jan has a bunch more work in the pipeline for pcap support. I'm thinking it makes sense to tackle that later as more pieces fall into place, but I'll tag it now.

awelzel:
> If somebody provides both interface and pcap, is that covered the same way as if somebody would do so on the command line? If not, it would be good to protect against it. Supervisor::Create() already does some sanity-checking.

Thanks. I've added this check now and also updated the baselines for the localnets change If CI seems happy with it I'll move ahead and merge.

PR #2842, opened 2023-03-03
Allow offline packet sources to register FDs.

J-Gras:
This just removes a check that prevents any offline packet source from registering a file descriptor. If provided, Zeek could use the descriptor to determine whether the source is ready.

```
// We only register the file descriptor if we're in live
// mode because libpcap's file descriptor for trace files
// isn't a reliable way to know whether we actually have
// data to read.
```
Given the above comment, the check was introduced for a good reason. However, it seems strange that the PCAP source sets the FD in the first place:

https://github.com/zeek/zeek/blob/8cf9e5b3741956901b9bc0ce5ad249570c39c549/src/iosource/pcap/Source.cc#L185-L188

If the original comment was about `fileno` failing we should already be good. If there are cases where `fileno` returns a valid descriptor but using the descriptor does not work as expected, it might make sense to set `selectable_fd = -1` in the PCAP source itself. @timwoj, can you remember?

So far, I have done some basic performance tests on my local machine and did not spot any issues. Nevertheless, I guess the change should probably be tested on all supported platforms.

### DIFF
--- a/src/iosource/PktSrc.cc
+++ b/src/iosource/PktSrc.cc
@@ -83,17 +83,11 @@ void PktSrc::Opened(const Properties& arg_props)
 		}
 
 	if ( props.is_live )
-		{
 		Info(util::fmt("listening on %s\n", props.path.c_str()));
 
-		// We only register the file descriptor if we're in live
-		// mode because libpcap's file descriptor for trace files
-		// isn't a reliable way to know whether we actually have
-		// data to read.
-		if ( props.selectable_fd != -1 )
-			if ( ! iosource_mgr->RegisterFd(props.selectable_fd, this) )
-				reporter->FatalError("Failed to register pktsrc fd with iosource_mgr");
-		}
+	if ( props.selectable_fd != -1 )
+		if ( ! iosource_mgr->RegisterFd(props.selectable_fd, this) )
+			reporter->FatalError("Failed to register pktsrc fd with iosource_mgr");
 
 	DBG_LOG(DBG_PKTIO, "Opened source %s", props.path.c_str());
 	}
--- a/src/iosource/pcap/Source.cc
+++ b/src/iosource/pcap/Source.cc
@@ -182,10 +182,11 @@ void PcapSource::OpenOffline()
 		return;
 		}
 
-	props.selectable_fd = fileno(pcap_file(pd));
-
-	if ( props.selectable_fd < 0 )
-		InternalError("OS does not support selectable pcap fd");
+	// We don't register the file descriptor if we're in offline mode,
+	// because libpcap's file descriptor for trace files isn't a reliable
+	// way to know whether we actually have data to read.
+	// See https://github.com/the-tcpdump-group/libpcap/issues/870
+	props.selectable_fd = -1;
 
 	props.link_type = pcap_datalink(pd);
 	props.is_live = false;


timwoj:
> If there are cases where fileno returns a valid descriptor but using the descriptor does not work as expected, it might make sense to set selectable_fd = -1 in the PCAP source itself. @timwoj, can you remember?

libpcap will return a proper file descriptor, but it's isn't really usable for reading data. See https://github.com/the-tcpdump-group/libpcap/issues/870 for discussion about that. You can read from it, but it it doesn't understand EOF correctly and so you never know if you're actually out of data to read. You're probably ok to just set it to `-1` in the offline case and let the normal handling of that value kick in. I don't recall why I didn't just do that in the first place.

src/iosource/PktSrc.cc
```diff
@@ -82,18 +82,11 @@ void PktSrc::Opened(const Properties& arg_props)
...
-
-		// We only register the file descriptor if we're in live
-		// mode because libpcap's file descriptor for trace files
-		// isn't a reliable way to know whether we actually have
-		// data to read.
-		if ( props.selectable_fd != -1 )
-			if ( ! iosource_mgr->RegisterFd(props.selectable_fd, this) )
-				reporter->FatalError("Failed to register pktsrc fd with iosource_mgr");
-		}
+	Info(util::fmt("listening on %s\n", props.path.c_str()));
```
awelzel, line 85: Would keep the "Listening on" message under `props.is_live`. Wouldn't that change the output of `zeek -r` testing (which currently is silent) to a single line saying "listening on /path/to/pcap"?

J-Gras: Correct, I thought about adding a "reading from /path/to/pcap" message depending on `props.is_live`. However, this would require adding the path diff-canonifier to some tests and might break external tests as well. So probably not worth it.

timwoj (approved):
I have a slight concern that something like npcap does actually support reading from pcap files like this, but that code path wouldn't have been handled anyways as it is.

awelzel:
Hmm - I think this subtly broke something.

After the merge, the capture-loss log for m57-short shows diffs on alpine, centos7, and also ubuntu22. This might be something racy, as at least ubuntu22 ran on the PR (though it seems it has failed there once, too). I'm going to trigger a re-run, but for the time being will revert the merge and we can retry with a "fullci" label.

https://cirrus-ci.com/task/6659512367579136
https://cirrus-ci.com/task/5322506228203520

---

This is the output of the alpine job in CI and below is the actual content/snippet from the artifact indicating the timestamp used for that line is "realtime" and not pcap time.
```
  #### btest-diff capture_loss.log
  <Content not shown>
  --- /tmp/test-diff.195178.capture_loss.log.baseline.tmp	2023-03-08 10:31:22.447289552 +0000
  +++ /tmp/test-diff.195178.capture_loss.log.tmp	2023-03-08 10:31:22.439289559 +0000
  @@ -8,7 +8,6 @@
   #fields	ts	ts_delta	peer	gaps	acks	percent_lost
   #types	time	interval	string	count	count	double
   #close XXXX-XX-XX-XX-XX-XX
  -XXXXXXXXXX.XXXXXX	481.X	zeek	0	11	0.X
   XXXXXXXXXX.XXXXXX	60.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	900.X	zeek	0	23	0.X
   XXXXXXXXXX.XXXXXX	900.X	zeek	0	9	0.X
  @@ -105,3 +104,4 @@
   XXXXXXXXXX.XXXXXX	910.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	917.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	922.X	zeek	0	0	0.X
  +XXXXXXXXXX.XXXXXX	XXXXXXXXXX.XXXXXX	zeek	0	11	0.X
  #### btest-diff capture_loss.log
  <Content not shown>
  --- /tmp/test-diff.196158.capture_loss.log.baseline.tmp	2023-03-08 10:31:24.175287979 +0000
  +++ /tmp/test-diff.196158.capture_loss.log.tmp	2023-03-08 10:31:24.168287986 +0000
  @@ -8,7 +8,6 @@
   #fields	ts	ts_delta	peer	gaps	acks	percent_lost
   #types	time	interval	string	count	count	double
   #close XXXX-XX-XX-XX-XX-XX
  -XXXXXXXXXX.XXXXXX	481.X	zeek	0	11	0.X
   XXXXXXXXXX.XXXXXX	60.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	900.X	zeek	0	23	0.X
   XXXXXXXXXX.XXXXXX	900.X	zeek	0	9	0.X
  @@ -105,3 +104,4 @@
   XXXXXXXXXX.XXXXXX	910.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	917.X	zeek	0	0	0.X
   XXXXXXXXXX.XXXXXX	922.X	zeek	0	0	0.X
  +XXXXXXXXXX.XXXXXX	XXXXXXXXXX.XXXXXX	zeek	0	11	0.X
```

capture_loss.log from the archive:
```
#separator \x09
#set_separator  ,
#empty_field    (empty)
#unset_field    -
#path   capture_loss
#open   2023-03-08-10-31-08
#fields ts      ts_delta        peer    gaps    acks    percent_lost
#types  time    interval        string  count   count   double
1258790472.799632       60.000012       zeek    0       0       0.0
1258791372.799641       900.000009      zeek    0       0       0.0
...
1258875412.953009       910.153441      zeek    0       0       0.0
1258876316.410820       903.457811      zeek    0       0       0.0
1678271482.020664       419395165.609844        zeek    0       11      0.0
#close  2023-03-08-10-31-22
```

PR #2844, opened 2023-03-05
Script optimization maintenance for March 2023

vpax:
Here's my more-or-less monthly set of updates for script optimization.

First, running down a subtle test suite disagreement for scripts compiled to C++ wound up surfacing a general bug in `EnumVal`s, namely that their internal representation is of type `zeek_int_t` but their `Get()` method returned an `int`.  This can cause problems when `int`s are 32-bit (which I guess they now default to?) because when `Analyzer`s use `Tag`s with subtypes, those subtypes are embedded in the corresponding `EnumVal` starting at bit position 32.  This of course wraps to negative for 32-bit `int`, and will lead to ambiguity if there's more than 2 subtypes.

In addition to fixing that problem, maintenance for `-O gen-C++` covers:
* better (than nothing) run-time errors for compiled scripts, which is done by faking up a `CallExpr` AST node associated with a given function/hook/event invocation
* fixes for dealing with recursive types
* a fix for values in compiled scripts that contain embedded C++ comment markers


testing/btest/Baseline.cpp/bifs.to_addr/error
```diff
@@ -1,2 +1,2 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-error in <no location>: failed converting string to IP address (not an IP)
+error in <...>/to_addr.zeek (C++), line 12: failed converting string to IP address (Users_vern_warehouse_zeek_zeek_master_02Mar23_testing_btest__tmp_bifs_to_addr_to_addr_zeek__global_stmts__zf() and not an IP)
```
awelzel, line 2: This looks a bit unusual:
* The original baseline only has "not an IP" in parenthesis, this one has some path encoded form of a function, too (for the other error reports as well). If that's valuable information, maybe it should be somewhere close to the `(C++)` part?
* There seems to be user/directory specific output in these baselines (judging by /Users/vern/warehouse). Might need to add a canonifier (or suppress that variable part during generation) such that running the tests in other directories, in CI, or by other people won't fail.

testing/btest/Baseline.cpp/coverage.basic-no-errors/.stderr
```diff
@@ -1,2 +1,2 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-Hello, Zeek!
+<...>/basic-no-errors.test (C++), line 11: Hello, Zeek!
```
awelzel, line 2: The original baseline has "line 13".  Is it possible that the C++ variation does not account for empty lines? Skimming some of the other logs, seems to be the case there, too.

src/Type.h
```diff
@@ -398,6 +398,16 @@ class TableType : public IndexType
 public:
 	TableType(TypeListPtr ind, TypePtr yield);
 
+	// Used by script compilation to update a "stub" table type
+	// (which is specified by using a nil "ind" value in the constructor)
+	// with its actual index & yield - necessary for dealing with
+	// recursive types.
+	void SetIndexAndYield(TypeListPtr ind, TypePtr yield)
```
awelzel, line 405: I suspect this is not safe to be used more than once and/or only in specific circumstances otherwise current table content and index/yield types are inconsistent. Could we runtime protect from wrong usage?


Pondering a bit, we're adding this to the public API, and even though it says "used by script compilation", it is provided to anyone for use until we deprecate it. I could also see friending the CPP script_opt implementation to avoid adding it for anyone. I might be overthinking this...

awelzel: An alternative could be to have a separate CPPTableType version inherited from TableType (or IndexType). The member variables themselves are already protected, so could fiddle with them that way.

testing/btest/Baseline.cpp/bifs.to_addr/error
vpax: Regarding the first: yeah, this is a side effect of using a slip-shod method of tracking calls in scripts compiled to C++.  Previously, this would have reported `no location` (as shown in the diff), and in some cases actually caused crashes.  Now, it provides the root of the C++ code that eventually led to the problematic call.  Not great, of course, though significantly better than before ... and reflecting a conscious design decision that `-O gen-C++` emphasizes performance over error pinpointing, with the model being that in general you should only compile scripts to C++ after being satisfied with their correctness.

Regarding the second point, thanks for flagging.  I've implemented a canonifier.  The question is now how to integrate it into the BTest framework.  It seems a pity to alter the BTests itself to use it, since in general there's no obvious reason which ones need it versus something else (for example, for `bifs.to_addr` above).  I can simply extend `diff-remove-abspath` to include this as part of its removal, with a comment to that effect - does that sound like the best way to go?  (If we had `$SCRIPTS` search paths then we could add a new directory for auxiliary scripts, e.g. `testing/CPP-scripts`; but without search paths that doesn't work since it won't have the other stuff in it that's needed by other BTests.)

BTW, the addition looks like:
```
 $sed 's#/+#/#g' |
+    $sed 's#[^( ]*testing_btest#testing_btest#' |
     $sed 's#([a-zA-Z]:)?/([^	 :/]{1,}/){1,}([^	 :/]{1,})#<...>/\3#g'
```
which IMHO is modest enough that adding that to the current `scripts/diff-remove-abspath` seems ok.

testing/btest/Baseline.cpp/coverage.basic-no-errors/.stderr
vpax: Per the above, this is reflecting that the error message tracks the beginning of compiled-to-C++ execution, rather than the direct location of the error.

src/Type.h
vpax: I like the `CPPTableType` approach, and have implemented that.  Will push once we resolve the above re error message canonicalization.

testing/btest/Baseline.cpp/bifs.to_addr/error
awelzel: No good intuition :-(

If canonifiers were used solely from the PATH, things would be easier, but that's not the layout, so, hmm.

Could possibly set SCRIPTS to a directory specifically for the CPP environment, but it would require replicating the old dir with a bunch of symlinks or platform specific overlay tricks to make this work quite right. Neither seems attractive.

The other thought would be to normalize the paths to the `cwd` or some other anchor within Zeek. I haven't really looked where it's coming from or if that's sensible, but with std::filesystem (as zeek::fileystem) being available there should be helpers so it's not too painful.

But, not opposed to the canonifier either. Suggest to add a `<...>` to indicate the replacement better for when it triggers unexpectedly?  Or maybe `<--->` or to differentiate from the existing one.

I'm resolving this assuming canonifier. If someone else has stronger opinions can reopen.

src/Val.h
```diff
@@ -155,7 +155,7 @@ class Val : public Obj
 
 	UNDERLYING_ACCESSOR_DECL(detail::IntValImplementation, zeek_int_t, AsInt)
 	UNDERLYING_ACCESSOR_DECL(BoolVal, bool, AsBool)
-	UNDERLYING_ACCESSOR_DECL(EnumVal, int, AsEnum)
+	UNDERLYING_ACCESSOR_DECL(EnumVal, zeek_int_t, AsEnum)
```
awelzel, line 158: That's making it at least consistent with EnumType->Lookup(). Maybe it'll start warning/errors on `int y = x->AsEnum()` constructs, but that would be for the better.

@timwoj - pinging you explicitly. Mark as resolved if there are no concerns.

testing/btest/Baseline.cpp/bifs.to_addr/error
vpax: Okay I've pushed a change to the canonifier.  Good thought re a marker, I went with `<___>` because what's removed is generally a bunch of tokens separated by `_`'s.

src/Type.h
vpax: Now pushed.

src/Val.h
vpax: @timwoj I think the pending issues are resolved with the push I just did, other than double-checking on this issue.

timwoj: Sorry, I must have missed the original ping on this one. I'm fine with this. There's so many places in this code where we implicitly cast from larger types to smaller ones.

awelzel (approved):
@vpax - I'll go ahead and merge this.

I'm going to try splitting d8c5cb84ca78bb47d0b3906310182f814f7b3a64 into two separate commits such that the AsEnum() change is isolated from the CPP error tracking/reporting.

The error reporting doesn't immediately strike me as a 5.2.x patch release feature, though the zeek_int_t part might be.

src/broker/Manager.cc
```diff
@@ -729,15 +729,15 @@ bool Manager::PublishLogCreate(EnumVal* stream, EnumVal* writer,
 
 	if ( ! stream_id )
 		{
-		reporter->Error("Failed to remotely log: stream %d doesn't have name", stream->AsEnum());
+		reporter->Error("Failed to remotely log: stream %lld doesn't have name", stream->AsEnum());
```
awelzel, line 732: Going to change these to "..%" PRI64 "..." as they trigger compile warnings on my end:
```
../src/broker/Manager.cc:795:54: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 3 has type ‘zeek_int_t’ {aka ‘long int’} [-Wformat=]
  795 |   reporter->Error("Failed to remotely log: writer %lld doesn't have name", writer->AsEnum());
      |                                                   ~~~^                     ~~~~~~~~~~~~~~~~
      |                                                      |                                   |
      |                                                      long long int                       zeek_int_t {aka long int}
      |                                                   %ld
```

Issue #2806, opened 2023-02-22
Builtin external plugins and zeek/zeek-version.h backwards compat


awelzel:
With #2802, ZEEK_VERSION* macros were split from zeek-config.h into zeek-version.h. For backwards compatibility, when building an external plugin, zeek-config.h does include zeek-version.h. This mechanism does not extend to "builtin external" plugins - those configured with `--include-plugins`. If they reference ZEEK_VERSION_NUMBER, zeek/zeek-version.h must be included explicitly. For AF_Packet and spicy-plugin that was done in zeek/zeek-af_packet-plugin#45, zeek/spicy-plugin#48.

If we could string through a `-DZEEK_BUILTIN_EXTERNAL_PLUGIN=true` or so, that would allow extending the `zeek/zeek-version.h` include condition also to builtin plugins. I don't see how we otherwise know we're building an external plugin.

For most users, there's no impact, but minimally @bbannier was wondering if that isn't too fragile. Other opinions?

Maybe it's not worth the trouble, in which case we can close this again.

This is where builtin external plugins are added, I'd guess we could push and pop to the compile options, but @Neverlord may have better ideas: https://github.com/zeek/zeek/blob/ebc956324358d194c77dd2fe67572fd10ca35b9f/src/CMakeLists.txt#L214-L222

Neverlord:
I'll fix it as a followup to https://github.com/zeek/zeek/pull/2799. 🙂

Issue #2482, opened 2022-10-14
include-plugins: binpac based plugins can not be used (include path issue)


awelzel:
This has been tested with the release/5.0 branch but should be the same on master.

When using `--include-plugins` to include a plugin that leverages binpac, the binpac_target fails to setup the `<zeek_base>/src` path as an include directory, causing binpac errors as follows:

For example, using the icsnpp-bacnet repo (https://github.com/cisagov/icsnpp-bacnet) and attempting to building it fails:
```
$ ./configure [...] --include-plugins=/path/to/icsnpp-bacnet
$ make

# The binpac invocations produces the following errors, but doesn't actually fail...
/home/awelzel/corelight-oss/icsnpp-bacnet/src/bacnet.pac:1: error: cannot include file "zeek/binpac.pac"                                                                                                                                                                                                                                                    
/home/awelzel/corelight-oss/icsnpp-bacnet/src/bacnet.pac:2: error: cannot include file "zeek/zeek.pac"

# following are more errors due to the missing includes
src/builtin-plugins/bacnet_pac.h:156:26: error: expected ‘)’ before ‘*’ token                                                                                                                                                                                                                                                                                 156 |  BACNET_Conn(ZeekAnalyzer * zeek_analyzer);                                                                                                                                                                                                                                                                                                               |             ~            ^~                                                                                                                                                                                                                                                                                                                               |                          )                                                                                                                                                                                                                                                                                                                          src/builtin-plugins/bacnet_pac.h:162:2: error: ‘ZeekAnalyzer’ does not name a type; did you mean ‘FlowAnalyzer’?                                                                                                                                                                                                                                              162 |  ZeekAnalyzer * zeek_analyzer() const { return zeek_analyzer_; }                                                                                                                                                                                                                                                                                          |  ^~~~~~~~~~~~                                                                                                                                                                                                                                                                                                                                             |  FlowAnalyzer                                                                                                                                                                                                                                                                                                                                       
src/builtin-plugins/bacnet_pac.h:171:2: error: ‘ZeekAnalyzer’ does not name a type; did you mean ‘FlowAnalyzer’?                                                                                                                                                                                                                                              171 |  ZeekAnalyzer * zeek_analyzer_;                                                                                                                                                                                                                                                                                                                           |  ^~~~~~~~~~~~                                                                                                                                                                                                                                                                                                                                             |  FlowAnalyzer                                                        
```

awelzel:
[...]

PR #2847, opened 2023-03-06
Add --include-plugins CI job

awelzel:
This is similar to the zeekctl setup, but does reference a few external projects. As can be seen by the involved commits as well as the fact that it's using forks of all these plugins, currently things are in a dire shape.

* The reason for using plugin forks is the `zeek/zeek-version.h` issue I caused. That is tracked by #2806 going forward (builtin plugins would receive ZEEK_VERSION_NUMBER again through the old `zeek-config.h`.

* #2482 was re-introduced when the `zeek -> .` symlink was removed, so I'm naively trying to add it back (fixed now in master)

* Multiple include-plugins may or may have not worked previously, but currently because of using eval and `;` creates truncated commands within the `configure` script.

* collect repo info learned how to ignore comments and empty lines in VERSION files to be able to parse zeek-kafka's (https://github.com/SeisoLLC/zeek-kafka/blob/main/VERSION)

I do think it would be nice to sneak this in before some of @Neverlord 's cmake restructuring, or shortly afterwards for some assurance this feature keeps working.


ckreibich (changes requested):
This isn't so much a "request changes" than "let's discuss for a sec". :-)

I like the early warning heads-up but I worry about placing the test at the same level as the rest of our PR workflow, and the sort-of-reliance on external projects.

If external developments break this test and we notice as part of our regular PR work, that's confusing. If the CI notification mechanism were clearly separate, or not an outright error, I'd find that information great to have, though. Cirrus has [some support](https://cirrus-ci.org/guide/writing-tasks/#failure-toleration) for this. If we run this via chron, we'd still get timely notifications, it'd stay out of our PR work, and we should still be able to use ccache state to speed up the build, right? (One could also argue that for a nightly the build time wouldn't matter too much, if this is tricky.) Could you see any of these?

Regarding the plugins ... we don't normally use personal forks in the project CI setup. A middle ground could be to fork those repos needing tweaks into the https://github.com/zeek-packages/ space, perhaps with a name that indicates their testing context so people don't confuse them. We've had that space for a while but not really figured out what to do with it. But — if we need to fork them, doesn't that defeat the purpose of using an external plugin?

For the repos you didn't fork, we're using master/main for the test, right? Would it make sense to use their latest release tags instead? It'd isolate us from their development hygiene and shift focus more to breakage we'd introduce.

Ideally we'd use reliable, well maintained external projects for the reasons you flag, but only ones to which we have a sufficiently direct line to coordinate quickly. That's a given for Community ID, and I think also for zeek-kafka and icsnpp-bacnet. I'm less sure about bro-http2 (but I admit that's mainly because I've not looked at that one in a while!).

awelzel:
> This isn't so much a "request changes" than "let's discuss for a sec". :-)

You're right on all fronts. Lets term this a shot at a fairly optimistic smoke test with lots of entropy and opportunities for things to break without us doing something. Not sure how bad it would've turned out to be, but thanks for blocking this :-)

With what you bring up, seems most sensible to just shelve what's here and re-start with hello-world plugins for the cases tested that we can include directly within Zeek (like testing/btest/plugins). Think this steers clear of most concerns/questions that would come up.

PR #2849, opened 2023-03-07
Manually prune CI ccache.

bbannier:
Since ccache in Cirrus currently does not preserve timestamps the timestamp-based cache pruning implemented by ccache likely would just remove random files. To work around this, this patch implements a manual pruning step. This step runs after the build so that at least the files used in the build should have their timestamps updated. We can then force eviction of some of the unused files from the cache by cleaning the cache with a size less than the maximum cache size.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -113,6 +113,9 @@ ci_template: &CI_TEMPLATE
       type: text/html
     cache_statistics_script:
       ccache --show-stats
+    ccache_prune_script:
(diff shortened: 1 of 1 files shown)


awelzel (approved):
We can try. We're still far away from 700MB ccache sizes on most Linux distros though, so probably doesn't matter for the next few weeks/months and possibly we could go as low as 500MB moving forward.

Thanks.

bbannier:
Could you take on merging this @awelzel? I cannot do that.

PR #2851, opened 2023-03-07
Sidestep handling HTTP 0.9 replies, avoiding analyzer violation

timwoj:
The HTTP analyzer assumes that the response looks like HTTP/1.0 and above, having a status code and everything. This doesn't work with HTTP/0.9 since those kinds of fields don't exist. HTTP/0.9 is just a request followed immediately by the data requested. This PR handles 0.9 separately, skipping all of the additional processing and just assuming that the following data is what was requested.


src/analyzer/protocol/http/HTTP.cc
```diff
@@ -1489,6 +1485,13 @@ const String* HTTP_Analyzer::UnansweredRequestMethod()
 
 int HTTP_Analyzer::HTTP_ReplyLine(const char* line, const char* end_of_line)
 	{
+	// There isn't any validation for reply data from an HTTP/0.9 request.
+	if ( request_version.major == 0 && request_version.minor == 9 )
```
awelzel, line 1489: I think this can work as a fix. I do wonder now a bit about the following HTTP message consumption because
1) the first line of the actual content was consumed as a reply line 
2) it looks like we'd parse the rest of the response as a MIME message which isn't true for HTTP/0.9

So, I'm almost tempted to suggest turning off the HTTP analyzer in addition to the added early return. Possibly also moving the 0.9 version check below into the `PrefixMatch` condition?

I think as a follow-up it may be worthwhile to discuss if HTTP/0.9 should be a separate analyzer altogether to remove the intertwining :-(

timwoj: > it looks like we'd parse the rest of the response as a MIME message which isn't true for HTTP/0.9

Yeah, this is definitely not good. Maybe just implement a deliver method for HTTP_Message that sidesteps a bunch of the processing if it's 0.9?

> So, I'm almost tempted to suggest turning off the HTTP analyzer in addition to the added early return. Possibly also moving the 0.9 version check below into the PrefixMatch condition?

I assume you mean in `HTTP_ReplyLine`? Falling into that if statement was what got us here. I think a fast exit just setting the version is the cleanest way out.

awelzel: > Maybe just implement a deliver method for HTTP_Message that sidesteps a bunch of the processing if it's 0.9?

I could see that as a follow-up, too, depending on how hairy it gets?

> I assume you mean in HTTP_ReplyLine? Falling into that if statement was what got us here. I think a fast exit just setting the version is the cleanest way out.

Right, I thought within the "does not start with "HTTP/" check to do the HTTP 0.9 request test and allow success. Mostly as a heuristic if there's a 0.9 request with a 1.0/1.1 response, but given we have not seen that, probably better to skip.

timwoj: > I could see that as a follow-up, too, depending on how hairy it gets?

Added in the latest commit.

0xxon:
Just as a note - this change seems to make several log lines in the tests less useful than they were before. E.g. in `scripts.base.protocols.http.http-methods`.

timwoj:
> Just as a note - this change seems to make several log lines in the tests less useful than they were before. E.g. in `scripts.base.protocols.http.http-methods`.

Yeah, @awelzel and I talked about this on the side the other day. The reason is because those requests are passing `GET /HTTP/1.1`. Note the lack of a space between the requested path and the HTTP version. The previous code would look at that, say it's HTTP/1.0 and then process it as a bad request. The patch says instead that it's a HTTP/0.9 request since 0.9 only requires a path after the GET. That said, HTTP/0.9 also only supports GET requests so any of the other methods have errored out in some other way.

timwoj:
Alright, merged in your change, squashed everything down, and added a weird for unknown HTTP/0.9 request methods.

src/analyzer/protocol/http/HTTP.cc
```diff
@@ -947,6 +979,18 @@ void HTTP_Analyzer::DeliverStream(int len, const u_char* data, bool is_orig)
...
 					HTTP_Request();
 					InitHTTPMessage(content_line, request_message, is_orig, HTTP_BODY_MAYBE, len);
+
+					// For HTTP/0.9, turn off the content_line analyzer for the
+					// responder because we expect raw data.
+					if ( request_version == HTTP_VersionNumber{0, 9} )
+						{
+						if ( request_method->ToStdString() != "GET" )
+							Weird("invalid_http_09_request_method", request_method->CheckString());
+						;
```
awelzel, line 989: Stray ; ?

src/telemetry/telemetry.bif
```diff
@@ -78,6 +78,7 @@ bool is_valid(zeek::Span<const zeek::telemetry::LabelView> labels,
 		{
 		return std::find(keys.begin(), keys.end(), x.first) != keys.end();
 		};
+
```
awelzel, line 81: Nit, that change seems spurious.

awelzel (approved):
@timwoj , you did review of the change suggested? :crossed_fingers: Would be nice to have a second review, but leaving an approval.

Less information in the http-methods log is unfortunate, but the requests are somewhat all over the place (and may or may not be considered HTTP).

EDIT: Seems a baseline update is missing.

PR #2852, opened 2023-03-08
Bump Spicy to latest main

awelzel:
To fix compile warnings.


bbannier (approved):
Looks good to me. Currently CI isn't done, but I wouldn't expect anything silly.

PR #2857, opened 2023-03-09
Allow offline packet sources to register FDs.

J-Gras:
Unfortunately, #2842 turned out to have some unexpected side-effects and had to be reverted. This is just a re-open to keep it on the radar. The logic related to `selectable_fd` in `GetNextTimeout()` might be part of the puzzle:
https://github.com/zeek/zeek/blob/b73dda5cff7a939444e646508cc796584982017e/src/iosource/PktSrc.cc#L277-L312

### DIFF
--- a/src/iosource/PktSrc.cc
+++ b/src/iosource/PktSrc.cc
@@ -83,17 +83,11 @@ void PktSrc::Opened(const Properties& arg_props)
 		}
 
 	if ( props.is_live )
-		{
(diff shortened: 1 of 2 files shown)


awelzel:
@J-Gras - I'll close this. I had let the CI tasks run a few more times and some tasks failed. Your commit is included in #2859 with corresponding GetNextTimeout() changes that seem to fare better.

PR #2859, opened 2023-03-10
PktSrc: Avoid calling ExtractNextPacketInternal() in GetNextTimeout()

awelzel:
This reworks 2aec7640dd60c3ea3cffd82461588567e406db34 (zeek/zeek#2039) to
avoid calling ExtractNextPacketInternal() within GetNextTimeout() for
the non-pseudo-realtime case. Also relates to zeek/zeek#2842.

The intention of the referenced change was to avoid a 0.00002 timeout when
a non-selectable packet source has more packets queued. This was implemented
by checking for a new packet within GetNextTimeout().

The proposed change switches to an predictive approach: Use the result of
the previous ExtractNextPacket() call (stored as had_packet) as an indication
whether more packets are to be expected.

Calling ExtractNextPacketInternal() within GetNextTimeout() may cause
surprising behavior as some packet source may block [1] or spent a significant
amount of time (e.g. applying BPF filters [2]) within ExtractNextPacket().
The result of GetNextTimeout() should be available immediately as guidance
for the main-loop and the actual work should happen within the ->Process()
method.

This change also attempts to separate the pseudo-realtime logic from the
non-pseudo-realtime.

[1] https://github.com/hosom/bro-napatech/blob/00c4d657e034927301d5f8d9bc03eca81e619699/src/Napatech.cc#L116
[2] https://github.com/sethhall/bro-myricom/blob/58b25c8ebac6d184ff8ff0c27f8da2b603694dab/src/Myricom.cc#L250

---

Addition: This change needs @J-Gras proposal to set libpcap's unreliable FD to -1 when in offline mode, because GetNextTimeout() now assumes that a set selectable_fd is properly usable and doesn't distingush between the Live/Offline case anymore. I think this is for the better, actually.

### DIFF
--- a/src/iosource/PktSrc.cc
+++ b/src/iosource/PktSrc.cc
@@ -85,17 +85,11 @@ void PktSrc::Opened(const Properties& arg_props)
 		}
 
 	if ( props.is_live )
-		{
 		Info(util::fmt("listening on %s\n", props.path.c_str()));
 
-		// We only register the file descriptor if we're in live
-		// mode because libpcap's file descriptor for trace files
-		// isn't a reliable way to know whether we actually have
-		// data to read.
-		if ( props.selectable_fd != -1 )
-			if ( ! iosource_mgr->RegisterFd(props.selectable_fd, this) )
-				reporter->FatalError("Failed to register pktsrc fd with iosource_mgr");
-		}
+	if ( props.selectable_fd != -1 )
+		if ( ! iosource_mgr->RegisterFd(props.selectable_fd, this) )
+			reporter->FatalError("Failed to register pktsrc fd with iosource_mgr");
 
 	DBG_LOG(DBG_PKTIO, "Opened source %s", props.path.c_str());
 	}
@@ -104,7 +98,7 @@ void PktSrc::Closed()
 	{
 	SetClosed(true);
 
-	if ( props.is_live && props.selectable_fd != -1 )
+	if ( props.selectable_fd != -1 )
 		iosource_mgr->UnregisterFd(props.selectable_fd, this);
 
 	DBG_LOG(DBG_PKTIO, "Closed source %s", props.path.c_str());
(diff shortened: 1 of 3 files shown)


awelzel:
@antcodd - any chance you could take a peek if that change might work out, too?

src/iosource/PktSrc.cc
```diff
@@ -276,39 +279,43 @@ bool PktSrc::GetCurrentPacket(const Packet** pkt)
...
+	// between calls to select() so just use that.
+
+	// A heuristic to avoid short sleeps when a non-selectable live packet source has
+	// more packets queued is to return 0.0 if the source has yielded a packet on the
+	// last call to ExtractNextPacket() or has a packet ready.
+	//
+	// A non-selectable offline packet source is assumed to always be ready.
+	if ( props.selectable_fd == -1 )
+		{
+		if ( have_packet || had_packet || ! IsLive() )
```
J-Gras, line 305: With respect to the assumption about offline packet sources being always ready, I am wondering whether the `! IsLive()` check can be omitted. If the assumption holds, I would have expected `had_packet` to be true at this point anyway.

Another idea would be to make this more explicit by introducing a `IsReady/Idle()` method so that the packet source can influence the decision. The base class could provide the `had_packet` logic as a default. However, I might be overthinking this, as Packet Sources without FDs should probably just override `GetNextTimeout()` and replicate the suspended- and pseudo-realtime logic if necessary.

awelzel: > If the assumption holds, I would have expected had_packet to be true at this point anyway.

Good callout. At least for libpcap that should hold and I agree this adds/keeps a special case while one motivation was to streamline things. A non-selectable and non-ready offline packet source being subject to the same small sleep actually sounds more reasonable than the special case. Thanks!

I'm a bit worried that something else topples over because the very first timeout will now not be 0.0, but we'll see what CI says.

timwoj (approved):

PR #2861, opened 2023-03-12
Install git as part of the centos-7 CI image

timwoj:
I have no idea if this was part of the base image that Cirrus uses and was removed recently, but it's missing right now and builds are failing because of it.


timwoj:
We were already installing git236 but it got lost on the image somehow. I forced the docker image to rebuild by updating the epoch in the Dockerfile. Everything seems happy now 🤷🏼

Issue #2856, opened 2023-03-09
Delayed/buffered/aged packets cause timing artifacts / premature connection expiration


awelzel:
It has been reported that NICs with large packet buffers cause timing artifacts when queued packets are too far in the past, exceeding syn timeouts, session timeouts or other inacitivity timeouts.

A reproducer is below script. It busy-spins in script land to delay following packets by a configurable amount. When setting a delay of 7.0 seconds and monitoring curl requests to localhost, the conn.log ends-up as follows: A single connection is split into two: S and ^haD...

```
1678370118.531234       CQn4ev3z8XHcCy2YSl      ::1     44060   ::1     80      tcp     -       -       -       -       S0      -       -       0       S       1       80      0       0       -
1678370118.531246       CFGEQm169HXtB8Yml7      ::1     44060   ::1     80      tcp     http    0.000296        73      853     SF      -       -       0       ^hADadFf        5       433     5       1221-
1678370119.538894       CJOlLt2M5Ay3NKYuWa      ::1     44062   ::1     80      tcp     -       -       -       -       S0      -       -       0       S       1       80      0       0       -
1678370119.538907       C7JLrC3fG5xULiKXfc      ::1     44062   ::1     80      tcp     http    0.000320        73      853     SF      -       -       0       ^hADadFf        5       433     5       1221-
```

Lowering the udp_inacitivity_timeout to 5secs using the same script, artifacts with UDP/DNS "connections" can be provoked (could also delay packets by 1min+).

```
1678371136.705769       ChnpGk303HpQduR34c      172.17.0.2      51362   1.1.1.1 53      udp     dns     -       -       -       S0      -       -       0       D       1       77      0       0       -
1678371136.726418       CPcZfF3pveTn9bNPOc      172.17.0.2      51362   1.1.1.1 53      udp     dns     -       -       -       SHR     -       -       0       ^d      0       0       1       97      -
1678371136.744709       CGamXJTIaAVSkIG3h       172.17.0.2      32817   1.1.1.1 53      udp     dns     -       -       -       S0      -       -       0       D       1       77      0       0       -
1678371136.764838       CxNmCj2hOX7fk4zj06      172.17.0.2      32817   1.1.1.1 53      udp     dns     -       -       -       SHR     -       -       0       ^d      0       0       1       97      -
```

Current understanding is that there are various places within Zeek where network time is forwarded to wallclock time fairly unconditionally (when in live mode at least) whether or not the packet source is dry or empty:

[...]

PR #2866, opened 2023-03-14
iomanager: Collect all sources with zero timeouts as ready

awelzel:
Previously, if two iosources returned 0.0 as their timeout, only one of them would be considered ready. An always ready source therefore may starve other ready ones due to this and minimally this behavior seems surprising.

Offline pcap sources are always ready and return 0.0 for GetNextTimeout() (unless in pseudo-realtime), so we can also remove the offline source special case.

This also makes the hard-coded poll_intervals of 100 and 10 configurable via `const &redef` vars. I have kept the comment, about 100 being too high for fast live networks, though I do not know how helpful that really is.


awelzel:
@timwoj , I hope this is not too small a PR. I'm looking at #2856 and there's a few intricate pieces bubbling up.

If you'd rather have them queue up in a branch, let me know.

src/iosource/Manager.cc
```diff
@@ -175,6 +175,14 @@ void Manager::FindReadySources(ReadySources* ready)
 					}
```
timwoj, line 175: If we're going to just automatically insert zero-timeout sources, I think this whole if statement could change as well to something like:

```
if (timeout == -1 || (next > 0 && next < timeout))
  {
  timeout = next;
  timeout = iosource;
  }
```

There's no reason to check for the zero-timeout case because we're covering it already later. I'm slightly concerned that a zero-timeout source could lead to starvation of any other timeout sources, but I guess they'd eventually fire when they hit the case where their timeout is zero as well.

timwoj (changes requested):

src/iosource/Manager.cc
awelzel: As in DM, tried to adapt with your suggestion. Not sure it's better now.

I think the IsLive() special case could be stream-lined a bit, but not wanting to change too much at once...

scripts/base/init-bare.zeek
```diff
@@ -5581,6 +5581,23 @@ const digest_salt = "Please change this value." &redef;
 ## :zeek:see:`find_all_ordered` BIFs.
 const max_find_all_string_length: int = 10000 &redef;
 
+## How many rounds to go without checking IO sources with file descriptors
+## for readiness by default. This is used when reading from traces.
+##
+## Very roughly, when reading from a pcap, setting this to 100 results in
+## 100 packets being processed without checking FD based IO sources.
+const io_poll_interval_default = 100 &redef;
```
timwoj, line 5589: You should add some comments to these about how they generally shouldn't be changed in practice.

timwoj (approved):
Looks generally ok to me. Just one nit.

scripts/base/init-bare.zeek
awelzel: I added a `.. note::` about that, thanks.

PR #2868, opened 2023-03-14
Add configure option for preallocating PortVal objects

### LINKED ISSUE #2571: Windows: Investigate performance differences with using `unordered_map` for `PortVal` caching
From https://github.com/zeek/zeek/pull/2518#pullrequestreview-1163245371:

> The original std::array version appears to run twice as fast as the std::unordered_map/uint32_t version still. It is a microbenchmark, but PortVal lookups aren't exactly uncommon, so we may at least keep that in mind.
>
> Edit: Could consider making this runtime/compiletime configurable if it matters as a knob for memory vs performance.

timwoj:
This PR restores the old model of pre-allocating all of the `PortVal` objects at startup, but adds an option to `configure` that allows usage of the new model of not pre-allocating them.

I considered making this a script-level option but ValManager gets initialized before script parsing starts and the PortVals should exist prior to that as well. I also considered making it a runtime option, but the code ends up cleaner by making it compile-time. There's a question of whether a runtime option would be more useful for things like btests though.

Fixes #2571

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -80,6 +80,17 @@ Changed Functionality
 
 - The ``&on_change`` attribute of set and tables is propagated through ``copy()``.
 
+- Revert back to old method of preallocting ``PortVal`` objects for all valid
+  port numbers, as it was implemented prior to the Windows port. Not
+  preallocating these objects saves a minor amount of memory for short runs of
+  Zeek, but comes at a performance cost for having to allocate the objects every
+  time a new port is seen plus do map lookups for each port. This memory savings
+  is mostly lost for long runs of Zeek, since all of the ports will likely end
+  up allocated in time.
+
+  If the version from the Windows port is desired, a new configure option
+  ``--disable-port-prealloc`` will disable the preallocation and enable the map
+  lookup version.
 
 Zeek 5.2.0
 ==========
--- a/configure
+++ b/configure
@@ -68,6 +68,7 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --disable-btest        don't install BTest
     --disable-btest-pcaps  don't install Zeek's BTest input pcaps
     --disable-cpp-tests    don't build Zeek's C++ unit tests
+    --disable-port-prealloc disable pre-allocating the PortVal array in ValManager
     --disable-python       don't try to build python bindings for Broker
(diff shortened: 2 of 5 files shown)


src/Val.cc
```diff
@@ -3988,10 +3999,14 @@ const PortValPtr& ValManager::Port(uint32_t port_num, TransportProto port_type)
 		}
 
 	auto port_masked = PortVal::Mask(port_num, port_type);
```
awelzel, line 4001: port_masked can go into `#else` otherwise it's unused.

configure
```diff
@@ -68,6 +68,7 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --disable-btest        don't install BTest
     --disable-btest-pcaps  don't install Zeek's BTest input pcaps
     --disable-cpp-tests    don't build Zeek's C++ unit tests
+    --disable-port-prealloc disable pre-allocating the PortVal array in ValManagerx
```
awelzel, line 71:
```suggestion
    --disable-port-prealloc disable pre-allocating the PortVal array in ValManager
```

I could see us not even hooking it up with `./configure` and assuming the advanced `-D` cmake settings if someone wants to flip it.

NEWS
```diff
@@ -66,6 +66,17 @@ New Functionality
...
   unless there wasn't a reply.
 
+- Add new configure option ``--disable-port-prealloc``. Prior to the merging of
+  the Windows port, Zeek preallocated PortVal objects for every valid port
+  number. This came at a cost of a minor amount of memory, but provided better
+  performance. With the Windows port, we removed this preallocation in favor of
+  the smaller memory footprint at the cost of having to allocate a new object
+  each time a new port number is seen in network traffic plus a potentially
+  slower lookup in a map vs an array. Over the span of a long run of Zeek, this
+  memory savings is lost but the performance cost still exists. This change
```
awelzel, line 76: Nit pickery. Maybe start the paragraph with: Revert to pre-allocated PortVals (or additionally put under "Changed Functionality"). Seems more relevant than the new config option and is a bit hidden at the end of the paragraph.

I'm also wondering if this is 5.2.1 material, or if we should suggest to distributors to pick the patch for better performance.

awelzel (approved):

Issue #2869, opened 2023-03-14
`scripts.base.protocols.ssl.x509-invalid-extension` btest fails on macOS Ventura


timwoj:
```
  % 'btest-diff .stdout' failed unexpectedly (exit code 1)
  % cat .diag
  == File ===============================
  == Diff ===============================
  --- /tmp/test-diff.94359..stdout.baseline.tmp	2023-03-14 20:46:52
  +++ /tmp/test-diff.94359..stdout.tmp	2023-03-14 20:46:52
  @@ -1,4 +0,0 @@
  -UNDEF
  -\x04a0_\xa1]\xa0[0Y0W0U\x16\x09image/gif0!0\x1f0\x07\x06\x05+\x0e\x03\x02\x1a\x04\x14\x8f\xe5\xd3\x1a\x86\xac\x8d\x8ek\xc3\xcf\x80j\xd4H\x18,{\x19.0%\x16#http://logo.verisign.com/vslogo.gif
  -UNDEF
  -\x04a0_\xa1]\xa0[0Y0W0U\x16\x09image/gif0!0\x1f0\x07\x06\x05+\x0e\x03\x02\x1a\x04\x14\x8f\xe5\xd3\x1a\x86\xac\x8d\x8ek\xc3\xcf\x80j\xd4H\x18,{\x19.0%\x16#http://logo.verisign.com/vslogo.gif
  =======================================

  % cat .stderr
  1398367809.961683 expression error in /Users/tim/Desktop/projects/zeek-master/testing/btest/.tmp/scripts.base.protocols.ssl.x509-invalid-extension/x509-invalid-extension.test, line 11: field value missing (ext$short_name)
  1398367809.961683 expression error in /Users/tim/Desktop/projects/zeek-master/testing/btest/.tmp/scripts.base.protocols.ssl.x509-invalid-extension/x509-invalid-extension.test, line 11: field value missing (ext$short_name)
```

This looks like an issue with OpenSSL 3.1.0. The last master (https://cirrus-ci.com/build/5440554075422720) build that passed had OpenSSL 3.0.8 installed by brew during the prepare step. The first CI build that failed was https://cirrus-ci.com/build/6717823930597376, which had OpenSSL 3.1.0 installed by brew. My local machine also has 3.1.0 and is failing here as well. We can pin CI to a specific version to get builds going again, if needed.

cc @0xxon

0xxon:
oh, fun. Thank you for flagging this.

PR #2870, opened 2023-03-15
Fix the x509-invalid-extension test on OpenSSL 3.1

0xxon:
OpenSSL 3.1 switched from outputting UNDEF to not giving a short name in this case. Luckily this only requires a tiny test change.

We might consider pulling this into older versions, for ease of CI testing.

Fixes GH-2869


awelzel (approved):

awelzel:
@0xxon - ack for just merging this in?

PR #2873, opened 2023-03-15
dns: Remove AD and CD from log

awelzel:
There was a misunderstanding whether to include them by default in the dns.log, so remove them again.

There had also been a discussion and quirk that AD of a request would always be overwritten by reply in the dns.log unless the reply is missing. For now, let users extend dns.log themselves for what best fits their requirements, rather than adding these flags by default.

Add a btest to print AD and CD flags for smoke testing still.


ckreibich (approved):

NEWS
```diff
@@ -66,10 +66,8 @@ New Functionality
 - Add logging metrics for streams (``zeek-log-stream-writes``) and writers
   (``zeek-log-writer-writes-total``).
 
-- The AD and CD flags are now parsed from DNS requests and responses and available
-  on the ``dns_msg`` record. The dns.log has been extended by AD and CD columns
-  with CD being populated from requests and AD being populated from replies,
-  unless there wasn't a reply.
+- The AD and CD flags are now parsed from DNS requests and responses and
+  available on the ``dns_msg`` record provided many of the ``dns_*`` events.
```
ckreibich, line 70: Let me suggest a bit more detail here:
```suggestion
- The DNS analyzer now parses RFC 2535's AD ("authentic data") and CD ("checking
  disabled") flags from DNS requests and responses,  making them available in
  the ``dns_msg`` record provided by many of the ``dns_*`` events. The existing
  ``Z`` field remains unchanged and continues to subsume the two flags, for
  backward compatibility.
```

Issue #2825, opened 2023-02-27
Allow to suppress `&deprecated` warnings


J-Gras:
It would be nice to have the ability to suppress deprecation warnings for a code fragment or a complete file. For example, in #2823 it would make sense to deprecate `Cluster::worker_count`. However, as the script itself needs to maintain the count as long as it is present, it would generate warnings.

awelzel:
I stared a bit because that might be helpful if we do #2877 (or the Node$ip deprecation). Think zeekctl populating cluster-layout and suppressing the warnings, but external users using `Node$ip` or `Node$interface` receiving them.

One idea would be new directives of the following form:
```
@start-ignore-deprecations
...
@end-ignore-deprecations


@start-ignore-deprecations
@load ./some-deprecated-script.zeek
@end-ignore-deprecations
```

Could also have `@ignore-deprecations` for the whole file, but tempted to say that's could be start/end over the whole file ;-)

It looks a bit verbose, but at the same time somewhat fits into the existing directives, and generally being verbose/noisy for things that should be used sparingly standing is probably a good thing.

Unfortunately there's not a single deprecation "channel" where messages go through, and there are a few ad-hoc messages here and there.

vpax:
A minor thought would be to consider doing this via a new `@pragma` "conditional", which would let us add other stuff in the future using a common framework.  E.g. `@pragma no-ZAM` to work around a ZAM bug.

timwoj:
> A minor thought would be to consider doing this via a new `@pragma` "conditional", which would let us add other stuff in the future using a common framework. E.g. `@pragma no-ZAM` to work around a ZAM bug.

I like this quite a bit. Are you envisioning this being like the existing @if directive where there's an opening/closing statement for a block of sorts, or something more like `#pragma` in C++ where how it's handled is dependent on the pragma itself? Pragmas in C++ are handled as individual statements. For examples, `#pragma once` only has to be placed once at the top of the file, but GCC has their diagnostic pragmas that allow pushing/popping different directives for various blocks.

Something similar to @if would definitely be easier to implement, but the C++ one is quite a bit more powerful.

awelzel:
> I like this quite a bit. Are you envisioning this being like the existing @if directive where there's an opening/closing statement for a block of sorts, or something more like #pragma in C++ where how it's handled is dependent on the pragma itself? 

[...]

PR #2881, opened 2023-03-17
Promote some deprecation warnings to errors

awelzel:
* out-of-scope has been warned about since Zeek 4.1
* Non captured locals have been warned about since Zeek 4.1
* vector scalar operations have been warned about since 2018

None of these said which version they'll be removed in, but seems it has been long enough.

This removes some of the ad-hoc deprecations so there's less to deal with for #2825


vpax:
FYI, back in the day @timwoj promised me (well sort of) that I'd get to yank out the old code associated with the `when` stuff. I see that That Day Has Come.

timwoj:
> Hmm, I didn't quite yank all that much except for making an error. I'd appreciate you reviewing (didn't yet assign because CI not green). Maybe there's a lot more to remove. I did rip out the scalar-vector things 

If they've been deprecated that long, we're good to remove them per the deprecation policy.

src/Expr.cc
```diff
@@ -1634,14 +1584,18 @@ AddExpr::AddExpr(ExprPtr arg_op1, ExprPtr arg_op2)
...
 
-	if ( IsVector(bt2) )
+	if ( IsVector(bt1) && IsVector(bt2) )
+		{
+		bt1 = op1->GetType()->AsVectorType()->Yield()->Tag();
 		bt2 = op2->GetType()->AsVectorType()->Yield()->Tag();
+		}
+	else if ( IsVector(bt1) || IsVector(bt2) )
+		{
+		ExprError("cannot mix vector and scalar operands");
```
vpax, line 1596: There's an opportunity to factor this out in a number of places, might be worth considering.

src/parse.y
```diff
@@ -984,13 +983,8 @@ expr:
...
 					{
-					if ( out_of_scope_locals.count(id.get()) > 0 &&
-					     warned_about_locals.count(id.get()) == 0 )
-						{
-						// Remove in v5.1
-						reporter->Warning("use of out-of-scope local %s deprecated; move declaration to outer scope", id->Name());
-						warned_about_locals.insert(id.get());
-						}
+					if ( out_of_scope_locals.count(id.get()) > 0 )
+						id->Error("use of out-of-scope local; move declaration to outer scope");
```
vpax, line 987: Just a comment: I think this is fine for now.  One day we might want to switch to truly removing the local from the scope, once its scripting scope terminates (which would allow redefinition of the local in a subsequent scope, as is supported by most common languages).

vpax:
@awelzel looks good, added some very minor comments, all are ignorable.

Regarding ripping out code, there's more that can be done for `when`s, happy to do so after this gets merged.

src/Expr.cc
awelzel: Yes. Added a static helper rather than adding a member that would be part of the interface. It's not pretty, but it's less lines.

src/Stmt.h
```diff
@@ -606,9 +606,8 @@ class WhenInfo
 
 private:
 	// True if the "when" statement corresponds to old-style deprecated
-	// semantics (no captures, but needing captures).  Also generates
-	// the corresponding deprecation warnings, which are associated
-	// with "ws".
+	// semantics (no captures, but needing captures).  Also triggers
+	// an error associated with "ws".
 	bool IsDeprecatedSemantics(StmtPtr ws);
```
timwoj, line 611: Does the naming of this method make sense since it's not deprecated semantics anymore? Maybe it could be `IsInvalidSemantics`?

src/Expr.cc
```diff
@@ -1577,26 +1577,41 @@ ValPtr SizeExpr::Fold(Val* v) const
...
 	}
 
-AddExpr::AddExpr(ExprPtr arg_op1, ExprPtr arg_op2)
-	: BinaryExpr(EXPR_ADD, std::move(arg_op1), std::move(arg_op2))
+// Fill op1 and op2 type tags into bt1 and bt2.
+//
+// If both operands are vectors, use their yield type tag. If
+// either, but not both operands, is a vector, cause an expression
+// error and return false.
+static bool types_from_scalars_or_vectors(Expr* e, TypeTag& bt1, TypeTag& bt2)
```
timwoj, line 1585: How about prefixing the name of this with `get_`? I don't think it's obvious what it does just from the name as-is.

timwoj (changes requested):
Just a couple of naming questions.

src/Stmt.h
awelzel: Possibly, yes. There was this comment from @vpax :

> Regarding ripping out code, there's more that can be done for whens, happy to do so after this gets merged.

Vern might have thoughts what to call it down the road, too, or remove the helper from the class altogether. I'll leave it for now as is as it seems there's more to drop. Leaving it as is for now.

src/Expr.cc
awelzel: Thanks, I've added the `get_`.

timwoj (approved):

Issue #2853, opened 2023-03-08
Current Prometheus output is malformed due to dashes in label names


awelzel:
@J-Gras reported telemetry/metrics being non functional when trying the zeek/zeek-demo project.

The reason is that one of the metrics is using a `-` in the label names as added during #2760 .
```
zeek_log_writer_writes_total{endpoint="",filter-name="default",module="PacketFilter",path="packet_filter",stream="PacketFilter::LOG",writer="Log::WRITER_ASCII"} 1 1678289396841
zeek_log_writer_writes_total{endpoint="",filter-name="default",module="LoadedScripts",path="loaded_scripts",stream="LoadedScripts::LOG",writer="Log::WRITER_ASCII"} 458 1678289396841
```

This is coming from here:
https://github.com/zeek/zeek/blob/b73dda5cff7a939444e646508cc796584982017e/src/logging/Manager.cc#L148-L149

@Neverlord - what's your take? I used a `-` in the label name to stay consistent with `-` in the metric names. It seems broker/caf should sanitize label names to underscores for broker exposition like it does with metrics names, or should a "user" do that?

awelzel:
This is from promtool:
```
root@ca8109769446:/var/zeek/logs# curl -Ssf localhost:9911/metrics | grep '^zeek_log' | promtool check metrics
error while linting: text format parsing error in line 3: expected '=' after label name, found '-'
```
![Screenshot-20230308-163811](https://user-images.githubusercontent.com/2100509/223758248-cf3bffd4-40a0-4c87-a450-56e5a58bcc2a.png)

This is from the Prometheus Server page:

Neverlord:
> what's your take? I used a - in the label name to stay consistent with - in the metric names. It seems broker/caf should sanitize label names to underscores for broker exposition like it does with metrics names, or should a "user" do that?

I'll fix it upstream.

timwoj:
Oh nice. I saw this the other day on os-perf-2 and made a note, but hadn't had a chance to dig into it yet.

timwoj:
Fixed via the linked PRs above

PR #2882, opened 2023-03-17
Update Broker for fixing #2853

Neverlord:
@timwoj this also picks up other recently merged Broker PR. Hope that's ok. Let me know if I should rebase the upstream so it only picks up the CAF patch.


awelzel (approved):
There'll be one more submodule bump after merging https://github.com/zeek/broker/pull/317 ?

Otherwise, seems good and would be nice to get the Prometheus issue fixed.

Neverlord:
Not sure if it's related or not: I had to restart one of the CI builds. It's green after re-triggering the failed build, but did this happen before?

> There'll be one more submodule bump after merging https://github.com/zeek/broker/pull/317?

Technically, we wouldn't need that. The upstream Broker would simply contain an extra merge commit. I don't know how Tim usually pulls these in. 🙂

timwoj:
Generally I prefer to merge the commit into master on broker and then update the zeek pointer to that. Otherwise we end up pointing at a weird spot in the the commit tree.

PR #2883, opened 2023-03-18
Add JavaScript support when libnode is available

awelzel:
zeek.on('zeek_init', () => {
        console.log('Hello, Zeek!');
    });

For interaction with external systems and HTTP APIs, JavaScript and the
Node.js ecosystem beat Zeek script. Make it more easily accessible by
including ZeekJS with Zeek directly.

When a recent enough libnode version is found on the build system, ZeekJS is
added as a builtin plugin. This behavior can be disabled via
``--disable-javascript``. Linux distributions providing such a package are
Ubuntu (22.10) and Debian (testing/bookworm) as libnode-dev.
Fedora provides it as nodejs-devel.

This plugin takes over loading of .js or .cjs files. When no such files
are provided to Zeek, Node and the V8 engine are not initialized and
should not get into the way.

This should be considered experimental.


0xxon (approved):
This looks good to me. However, this needs a documentation update too, right?

CMakeLists.txt
```diff
@@ -747,6 +747,26 @@ if ( ${CMAKE_SYSTEM_NAME} MATCHES Linux )
...
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR}/auxil/zeekjs/cmake)
+    find_package(Nodejs)
+
+    if ( NODEJS_FOUND )
+        if ( ${NODEJS_VERSION} VERSION_LESS "16.13.0" )
+            message(STATUS "Node.js version ${NODEJS_VERSION} is too old, need 16.13 or later. Not enabling JavaScript support.")
+            set(ZEEK_HAVE_JAVASCRIPT no)
+        else ()
+            set(ZEEKJS_PLUGIN_PATH ${CMAKE_SOURCE_DIR}/auxil/zeekjs)
+            string(APPEND ZEEK_INCLUDE_PLUGINS ";${ZEEKJS_PLUGIN_PATH}")
```
rsmmr, line 760: `LIST APPEND` is a bit more readable, and safer too.

rsmmr (approved):
I haven't further looked at the actual submodule in a while, but this integration at least looks good to me too, pretty straightforward.

testing/scripts/have-javascript
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if grep -q "ZEEK_HAVE_JAVASCRIPT:INTERNAL=no" "${BUILD}"/CMakeCache.txt; then
```
ckreibich, line 3: I've noticed a small hiccup with this — with libnode available, but configuring with `--disable-javascript`, CMakeCache.txt contains this:
```
//Zeek has JavaScript support
ZEEK_HAVE_JAVASCRIPT:INTERNAL=
```
That is, the grep doesn't catch and so the javascript tests don't get skipped but fail. (When libnode isn't actually on my system, it says `=no` as expected.)

Simply turn the test around to look specifically for `=yes` and I think all is well.

ckreibich (approved):
I've tinkered with this on my system and it's all working great (except for that trivial `--disable-javascript` hiccup) — so awesome.

When nodejs-devel isn't on my system the configure output looks a bit funky (version ".."?):
```
-- Could NOT find Nodejs (missing: NODEJS_INCLUDE_DIR V8_CONFIG_INCLUDE_DIR) (found version "..")
--      version: ..
--    libraries: /usr/lib64/libnode.so.108
--         uv.h: /usr/include
--   v8config.h: V8_CONFIG_INCLUDE_DIR-NOTFOUND
```
I think that's not our doing since we're just using `find_package(Nodejs)`, but I wanted to flag in case I'm missing something.

ckreibich:
One more thought: we could intentionally put libnode in one of the older-distro CI Dockerfiles so we test the version check in CI.

ckreibich:
A NEWS entry woule be good too, per recent discussion. I doubt we'll forget this one, though. :-)

testing/scripts/have-javascript
awelzel: Thanks for catching that.

awelzel:
> I think that's not our doing since we're just using find_package(Nodejs), but I wanted to flag in case I'm missing something.

Eh - that `find_package(Nodejs)` is minimally my doing ;-) It's looking a bit better now by not setting NODEJS_VERSION if the header/lib wasn't found.

> One more thought: we could intentionally put libnode in one of the older-distro CI Dockerfiles so we test the version check in CI.

Done for debian-11, looks as follows:
```
...
-- Found LibUV: /usr/lib/x86_64-linux-gnu/libuv.so (found version "1.40.0") 
-- Found Nodejs: /usr/include (found version "12.22.12") 
--      version: 12.22.12
--    libraries: /usr/lib/x86_64-linux-gnu/libnode.so;/usr/lib/x86_64-linux-gnu/libuv.so
--         uv.h: /usr/include
--   v8config.h: /usr/include/node
-- Node.js version 12.22.12 is too old, need 16.13 or later. Not enabling JavaScript support.
--   Building in plugin: spicy-plugin (/zeek/auxil/spicy-plugin)
--   Building in plugin: zeek-af_packet-plugin (/zeek/auxil/zeek-af_packet-plugin)
...
```

PR #2886, opened 2023-03-20
cirrus: No container images on PRs

awelzel:
Skip building container images (and skip cluster testing)
unless running with fullci.

They don't provide a lot of additional testing coverage, so
should be fine to just run them after merges to master.


timwoj (approved):
Yeah, that seems reasonable to me. We don't upload those containers anywhere, right? Only the ones off master and the release branches?

PR #2888, opened 2023-03-22
Support suppressing deprecated warning

### LINKED ISSUE #2825: Allow to suppress `&deprecated` warnings

awelzel:
* Add `reporter->Deprecation()` to channel all deprecation messages though (all that I've found so far).
* Introduce `@pragma push ignore-deprecations` handling and `@pragma pop`

Closes #2825 .


awelzel:
This went the `@pragma` route rather than magic comments to suppress deprecations. There's one test showing the optics of it a bit.

It doesn't look super pretty, but then it should almost never be used :man_shrugging:

timwoj:
I like the pragma approach because it lets us add similar directives more easily. It also matches how other languages do it, which makes the mental load on users easier.

awelzel:
I'm going to mark this as ready. I don't overly like how disabling looks in the scripts, but then it also does not need to look very pretty. The `reporter->Deprecation()`.

testing/btest/language/ignore-deprecations.zeek
```diff
@@ -0,0 +1,71 @@
...
+}
+event zeek_init()
+	{
+	print c;
+@pragma start-ignore-deprecations
+	print c;
+@pragma end-ignore-deprecations
+	}
+
+@TEST-START-NEXT
```
bbannier, line 17: Mixing BTest's `@` with Zeek's `@` becomes pretty hard to follow, could you put these behind comments like you already did in the test header?

bbannier (commented):
Since this PR lays out the expected semantics for Zeek `@pragma` directives, I wonder whether the semantics are like we want them.

Right now these `@pragma`s behave very much like C's `#pragma`s in that simply push a pragma on a single global stack which is e.g., shared across all loaded files. The following prints no deprecation warning:

```console
$ zeek x.zeek
1
```

```zeek
# file: x.zeek

@load ./weird_deprecation

global c: count = 1 &deprecated="Oh noes!";

event zeek_init()
	{
	print c; # Should we warn on this access?
	}
```


```zeek
# file: weird_deprecation.zeek

@pragma start-ignore-deprecations # YOLO.
```

I personally would prefer more fine-grained semantics, i.e., in this case a `@pragma` should at most have an effect until the end of the file. Otherwise it becomes very easy for code to be infected by whatever loaded files did.

awelzel:
>  in that simply push a pragma on a single global stack which is e.g., shared across all loaded files. The following prints no deprecation warning:
[...]
> I personally would prefer more fine-grained semantics, i.e., in this case a `@pragma` should at most have an effect until the end of the file.

Nice catch, thanks! I think your example should just err due to not being balanced within the file. But the balancing/stack pushing aspect would be specific to what "command" follows after `@pragma`. For start-stop the current one clearly took a short-cut :-)

awelzel:
> You could also consider dropping the tag from the end pragma completely to make this harder to use in too confusing ways, e.g.,

I'll allow the space between start/end and rename to push/pop and make a tiny bit generic. Popping without a value seems fine, but optionally adding it nicer for documentation.
```
@pragma pop
```

I'll stop at that point though assuming people are happy with the syntax. I personally don't see a ton of use-cases right now to build something overly generic. And the suppression of deprecation warnings is almost exclusively a Zeek internal use-case.

src/scan.l
```diff
@@ -135,6 +137,63 @@ static void start_conditional()
 	current_file_has_conditionals = true;
 	}
 
+static void do_pragma(const std::string& pragma)
+	{
+	static std::unordered_set <std::string> known_stack_pragmas = {
+		"ignore-deprecations",
+	};
+
+	if ( pragma.size() == 0 )
```
timwoj, line 146: Nit

```suggestion
	if ( pragma.empty() )
```

More of the same below.

src/scan.l
```diff
@@ -135,6 +137,63 @@ static void start_conditional()
...
+			if ( known_stack_pragmas.count(parts[1]) == 0 )
+				zeek::reporter->Warning("Pushing unknown pragma value '%s'", parts[1].c_str());
+
+			pragma_stack.push_back(parts[1]);
+		}
+	else if ( parts[0] == "pop" )
+		{
+		if ( pragma_stack.size() == 0 || pragma_stack.size() == entry_pragma_stack_depth.back() )
+			{
+			zeek::reporter->FatalError("@pragma pop underflow");
```
timwoj, line 171:
```suggestion
			zeek::reporter->FatalError("@pragma pop: no active pragmas");
```

timwoj (approved):
I like where this ended up. It's consistent with how other common languages do it, so it should be more familiar to users. I just had a few nits, but nothing major.

rsmmr (approved):
Looks good to me as well. I would have been fine wo/ that whole pragma stack logic as well, given our one limited use case, but I guess it doesn't hurt ...

PR #2893, opened 2023-03-24
Broker: Remove network time initialization

awelzel:
Remove the special case of initializing network time if it hasn't happened yet. The argument about broker.log containing 0.0 timestamps is more a problem of the log, not something that would justify modifying network time globally. For broker.log and possibly cluster.log, it might be more reasonable to use current time, anyway.

I was a bit wary about tables backed by broker stores being populated with network_time set to 0.0, but there seems to exist logic and assumptions that this is okay: It should be the same as if one populates a table with expirations set within zeek_init().

In fact, staring a bit more, *not setting* network time might be more correct as workers that don't see packets would never set zeek_start_network_time which is used within the expiration computation.


awelzel:
@rsmmr - going to request a review from you. Does the text make a bit of sense to you?

timwoj (approved):
Yeah, this makes sense to me. I'm not sure why broker would be responsible for initializing network time in the first place, and having the logs show zero times for a bit is ok.

PR #2895, opened 2023-03-24
Fix a bunch of -Wunqualified-std-cast-call warnings from clang 15

timwoj:
I started using clang 15 on my local Ubuntu installation and it's spitting out a bunch of warnings from unqualified calls to `std::move()`. This PR fixes all of the finding in Zeek itself. There's some other findings in CAF that trickle down to reports via the Broker build, as well as a few in Gen-Zam.


timwoj:
GitHub fixed whatever was broken about their certs on Friday, so this is ok to review now.

awelzel (approved):

PR #2898, opened 2023-03-27
Replace deprecated Conan generators with new ones

timwoj:
The `conan` and `conan_find_package` generators were deprecated in Conan 1.58. This PR replaces them with the recommended `CMakeDeps` and `CMakeToolchain` generators that are fully-released in Conan 2.0. I tested these with Conan 1.54 on my local machine and they work there as well.


timwoj:
I have to tinker with this one a bit more. It's not as easy as just changing the generators, because CMake doesn't find libpcap during the tests. It doesn't even look like it's passing it to the build commands in CMakeError.log.

PR #2901, opened 2023-03-29
Add GetChildAnalyzer() and IsPreventedChildAnalyzer()

### LINKED ISSUE #2899: There seems to be a non-recursive version of `Analyzer::FindChild` missing
Most of the `Analyzer` functions related to the analyzer tree seem to work on direct children (`HasChildAnalyzer`, `AddChildAnalyzer`, `RemoveChildAnalyzer`). The exception is `Analyzer::FindChild` which recurses over the analyzer tree. This is unfortunate, but probably too late to fix.

I came across this while working on zeek/spicy-plugin#198 which is about explicitly manipulating the Zeek analyzer tree in a safe way, e.g., via their `ID`s.

Assume we have built the following analyzer tree where e.g., _Analyzer B_ lives in a separate package or in Zeek proper and we have no knowledge of its internals:

```mermaid
graph TD;
A[Analyzer A]-->B;
B[Analyzer B]--> |2| ASN1[ASN.1];
A--> |1| ASN2[ASN.1];
```

We can detect edge _1_ with `HasChildAnalyzer` since that function does not recurse. If we however naively call `FindChild` we will likely get the analyzer at the end of edge _2_ since this function recurses and performs a depth-first search. The only way to get the analyzer at the end of edge _1_ is to iterate `GetChildren` ourself.

If there was a prominent non-recursive version of `FindChild` we could avoid this pitfall. Alternatively and more cumbersome, if `FindChild` would operate breath-first we could at least use it together with `HasChildAnalyzer` to get the wanted child analyzer.

awelzel:
Currently, it's not possible to understand whether adding a child analyzer would succeed. There might

1) be an existing child analyzer of the type or
2) the child analyzer's type might have been prevented from adding

The only real way to do this currently is via "Leap before you look" calling AddChildAnalyzer() with a new analyzer instance, which makes implementing `get_or_add()` a bit difficult.

Closes #2899


src/analyzer/Analyzer.cc
```diff
@@ -470,25 +466,33 @@ bool Analyzer::Remove()
...
 
+bool Analyzer::IsPreventedChildAnalyzer(const zeek::Tag& tag) const
+	{
+	return std::find(prevented.begin(), prevented.end(), tag) != prevented.end();
+	}
+
 bool Analyzer::HasChildAnalyzer(zeek::Tag tag)
+	{
+	return GetChildAnalyzer(tag) != nullptr;
+	}
```
bbannier, line 483: Since you are already touching this function, WDYT about fixing the parameter copy (not in line with other functions here), and fixing const correctness (but that's a bigger issue in this file)?

```suggestion
bool Analyzer::HasChildAnalyzer(const zeek::Tag& tag) const
	{
	return GetChildAnalyzer(tag) != nullptr;
	}
```

src/analyzer/Analyzer.h
```diff
@@ -436,13 +436,35 @@ class Analyzer
...
 	 * Returns true if analyzer has a direct child of a given type.
 	 *
 	 * @param tag The type of analyzer to check for.
 	 */
 	bool HasChildAnalyzer(zeek::Tag tag);
 
+	/**
+	 * Returns a pointer to a direct child analyzer of the given type.
+	 *
+	 * Note, the returned pointer is owned by the analyzer and may
```
bbannier, line 459:
```suggestion
	 * Note that the returned pointer is owned by the analyzer and may
```

bbannier (approved):

src/analyzer/Analyzer.cc
awelzel: I'm happy to do that (and hat it previously) but was only 99% sure about that being API compatible. So if you give a thumbs-up on that I'll do that :-)

awelzel: I did that, it seems it shouldn't matter for a normal user.

PR #2903, opened 2023-03-29
Bump Spicy and spicy-plugin to their latest development snapshots.

bbannier:


bbannier:
I am marking this as ready for review, even though there still were two OOM failures duing the CI build (retriggered). Since this PR bumps dev snapshots this patch should only go into `master`.

timwoj (approved):

PR #2909, opened 2023-03-30
Type: Add TypeManager->TypeList() and use for ListVal()

awelzel:
It turns out that for every ListVal() we construct, we also allocate and
construct a new TypeList instance, even though they are all the same.
Pre-create and cache the type instances.

The following script runs ~10% faster for me after this change.

    global tbl: table[string] of string;
    global i = 0;
    while ( ++i < 10000000 )
        tbl["a"] = "a";

### DIFF
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -1981,6 +1981,22 @@ detail::TraversalCode VectorType::Traverse(detail::TraversalCallback* cb) const
 	HANDLE_TC_TYPE_POST(tc);
 	}
 
+TypeManager::TypeManager()
+	{
+	for ( auto i = 0u; i < base_list_types.size(); ++i )
+		{
+		TypeTag tag = static_cast<TypeTag>(i);
+		TypePtr pure_type = tag == TYPE_ANY ? nullptr : base_type(tag);
+		base_list_types[tag] = make_intrusive<zeek::TypeList>(pure_type);
+		}
+	}
+
+const TypeListPtr& TypeManager::TypeList(TypeTag t) const
+	{
+	assert(t >= 0 && t < NUM_TYPES);
+	return base_list_types[t];
+	}
+
 // Returns true if t1 is initialization-compatible with t2 (i.e., if an
 // initializer with type t1 can be used to initialize a value with type t2),
 // false otherwise.  Assumes that t1's tag is different from t2's.  Note
--- a/src/Type.h
+++ b/src/Type.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <array>
 #include <list>
 #include <map>
 #include <optional>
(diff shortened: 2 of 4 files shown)


src/Val.h
```diff
@@ -332,6 +332,9 @@ class ValManager
 
 	inline const StringValPtr& EmptyString() const { return empty_string; }
 
+	// Get a ListVal instance with the given type tag.
```
timwoj, line 335: I'm not a huge fan of adding a method for this. Can we not add a new constructor to ListVal that just takes a TypeTag, add a `friend` to ListManager, and have the constructor do a lookup itself? You'd have to remove the default argument from the version that takes a `TypeList` but that would at least give you something explicit to deprecate.

awelzel: Thanks. Let me give it a spin. If that avoids the shotgun updates everywhere :+1:

timwoj: Yeah that's kinda what I was hoping too. This is a really large changeset for something that seems like it should be fairly trivial.

src/Type.cc
```diff
@@ -1981,6 +1981,22 @@ detail::TraversalCode VectorType::Traverse(detail::TraversalCallback* cb) const
...
+	{
+	for ( auto i = 0u; i < base_list_types.size(); ++i )
+		{
+		TypeTag tag = TypeTag(i);
+		TypePtr pure_type = tag == TYPE_ANY ? nullptr : base_type(tag);
+		base_list_types[tag] = make_intrusive<zeek::TypeList>(pure_type);
+		}
+	}
+
+const TypeListPtr& TypeManager::TypeList(TypeTag t)
```
timwoj, line 1994:
```suggestion
const TypeListPtr& TypeManager::TypeList(TypeTag t) const
```

awelzel:
Script runtime before the change ~7.1 seconds, after ~6.5 seconds.

src/Type.cc
```diff
@@ -1981,6 +1981,22 @@ detail::TraversalCode VectorType::Traverse(detail::TraversalCallback* cb) const
 	HANDLE_TC_TYPE_POST(tc);
 	}
 
+TypeManager::TypeManager()
+	{
+	for ( auto i = 0u; i < base_list_types.size(); ++i )
+		{
+		TypeTag tag = TypeTag(i);
```
timwoj, line 1988:
```suggestion
		TypeTag tag{i};
```

awelzel: That didn't work out:
```
../src/Type.cc:1988:25: error: invalid conversion from ‘unsigned int’ to ‘zeek::TypeTag’ [-fpermissive]
 1988 |   TypeTag tag = TypeTag{i};
```

I'm going for a static_cast<TypeTag>(i), or do you have strong opinions for something else?

timwoj: Oh yeah, typetag's an enum isn't it? Ignore that suggestion then.

PR #2912, opened 2023-03-30
Fix forwarding of tunnelled packets

J-Gras:
There was a small bug in the AYIYA, Geneve and VXLAN analyzers that caused them to only forward encapsulated packets if they are longer than the corresponding protocol headers. This also adds two test cases for Geneve that should cover the fix.


timwoj (approved):

src/packet_analysis/protocol/geneve/Geneve.cc
```diff
@@ -70,27 +70,30 @@ bool GeneveAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pack
...
 	data += hdr_size;
 
-	// We've successfully parsed everything, so we might as well confirm this.
+	// We've successfully parsed the Geneve part, so we might as well confirm this.
 	AnalyzerConfirmation(packet->session);
 
+	if ( len == 0 )
+		{
+		// A Geneve header that isn't followed by a tunnelled packet seems weird.
+		Weird("geneve_empty_packet", packet);
```
awelzel, line 79: I wonder if they do exist, there's minimally a "control packet" bit and I could see that being just headers, no payload. But until they pop up, seems a weird is fine.

>    O (1 bit):  Control packet.  This packet contains a control message.
      Control messages are sent between tunnel endpoints.  Tunnel
      endpoints MUST NOT forward the payload, and transit devices MUST
      NOT attempt to interpret it.

https://datatracker.ietf.org/doc/rfc8926/

awelzel (approved):
I've put this on the list of backports, at least for Geneve that is more than just a small bug assuming header usage is somewhat prevalent.

@J-Gras , do you happen to have a vxlan pcap, too? Otherwise I can also build one.

awelzel:
> @J-Gras , do you happen to have a vxlan pcap, too? Otherwise I can also build one.

Ah, ignore. For VXLAN that can't really happen in practice (otherwise we'd hopefully have heard about it) due to Ethernet or IP header being already > 8 bytes.

src/packet_analysis/protocol/geneve/Geneve.cc
J-Gras: I was almost certain that there are empty control packets but I wasn't able to find any. Looking at https://datatracker.ietf.org/doc/html/draft-ietf-nvo3-geneve-oam (although this doesn't mention the O flag) and https://www.mail-archive.com/nvo3@ietf.org/msg04692.html, I started to believe that empty payloads would be weird for control packets as well.

PR #2913, opened 2023-03-31
PktSrc: Introduce Pcap::non_fd_timeout

### LINKED ISSUE #2296: High System Load switching form v3.0.16 to v4.0.7 or v5.0.0
Hi,

So I've been slowly switch to v. 4 and observed a strange increase on the same machine with same traffic (~ 500 Mbit/s) and the same settings between versions
This screenshot represents v3.0.14
image
and this v.4.0.7
image
Same exact behavior on v5.0.0

Tested on two machines with ubuntu 18.04 and 20.04 same results. The only thing that comes to my mind that might be helpful is that we do use myricom cards. But I'm at a loss, any ideas ?

awelzel:
Increasing this value 10x has lowered CPU usage on a Myricom based deployment significantly with reportedly no adverse side-effects.

After reviewing the Zeek 3 IO loop, my hunch is that previously when no packets were available, we'd sleep 20usec every loop iteration after calling ->Process() on the packet source. With current master ->Process() is called 10 times on a packet source before going to sleep just once for 20 usec. Likely this explains the increased CPU usage reported.

It's probably too risky to increase the current value, so introduce a const &redef value for it for advanced users to tweak it. A middle ground might be to lower ``io_poll_interval_live`` to 5 and increase the new ``Pcap::non_fd_timeout`` setting to 100usec.

While this doesn't really fix #2296, we now have enough knobs for tweaking.

Closes #2296.

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -5207,6 +5207,32 @@ export {
 	## interfaces.
 	const bufsize = 128 &redef;
 
+	## Default timeout for packet sources without file descriptors.
+	##
+	## For libpcap based packet sources that do not provide a usable
+	## file descriptor for select(), the timeout provided to the IO
+	## loop is either zero if a packet was most recently available
+	## or else this value.
+	##
+	## Depending on the expected packet rate per-worker and the amount of
+	## available packet buffer, raising this value can significantly reduce
+	## Zeek's CPU usage at the cost of a small delay before processing
+	## packets. Setting this value too high may cause packet drops due
+	## to running out of available buffer space.
+	##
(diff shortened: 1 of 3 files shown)


scripts/base/init-bare.zeek
```diff
@@ -5203,6 +5203,32 @@ export {
 	## Number of bytes per packet to capture from live interfaces.
 	const snaplen = 9216 &redef;
 
+	## Default timeout for packet sources without file descriptors.
```
awelzel, line 5206: I'm not overly happy this being in the Pcap module (as it's more generic to PktSrc), but given packet sources may just override `GetNextTimeout()` and ignore it, seems fair game, too.

ckreibich (approved):

Issue #2075, opened 2022-04-28
@if block parsing problem when located within lambda return value


0xxon:
I tried making some script-code backwards compatible and notices that this code...

```zeek
function test()
	{
	local make_epoch_result = function(pass_name: string): function(ts: time)
		{
@if ( Version::at_least("4.1") )
		return function [pass_name] (ts: time)
@else
		return function (ts: time)
@endif
			{
			print pass_name;
			};
		};

	}
```

Throws the following error when run with a current master version of Zeek:

```
error in ./test2.zeek, line 7: syntax error, at or near "@else"
```

I _think_ that this is supposed to work fine - but someone please correct me if I am incorrect. One can work around this problem by changing the placement of the `{` after the return statement.

awelzel:
This appears to be fixable via the following hunk, but not sure #2289 is that easy, too.

```
 anonymous_function:
-               TOK_FUNCTION begin_lambda lambda_body
-                       { $$ = $3; }
+               TOK_FUNCTION begin_lambda conditional_list lambda_body
+                       { $$ = $4; }
        ;
```

awelzel:
> @awelzel will that change really work? The { for the body that precedes the @if in the example is part of the lambda_body, so that conditional will be prior to any elements of the body.

I will take a closer look tomorrow. I added a btest with above example and it seemed to work. Will double check and get back to you. Thanks for looking.

awelzel:
@vpax - the example of this ticket involves a lamda returning a lambda. It's the `@else` following `function [pass_name] (ts: time)` (`begin_lambda`) what upsets the parsing. The first `@if` within the lambda_body seems to work out as it's part of lambda_body's stmt_list.

The proposed fix is similar to what we do for func_hdr / func_body:

https://github.com/zeek/zeek/blob/fad18cb352d245e1076e7b2a23d229b3c4af463f/src/parse.y#L1401-L1406

I've opened a PR, please chime in if I misunderstood or you have further concerns/thoughts. Thanks!

vpax:
@awelzel I see.  Given that change, does the snippet at the top now work correctly?

awelzel:
> @awelzel I see. Given that change, does the snippet at the top now work correctly?

Yep. I've added it here as a test-case:

https://github.com/zeek/zeek/pull/2346/files#diff-c7537e2ab0b02836a74f51fd55edd9a3fee02a971df499131030bd24b12f7dc1R8

vpax:
[...]

PR #2917, opened 2023-04-02
Compile-to-C++ monthly maintenance - April 2023

vpax:
The only maintenance for `-O gen-C++` this month concerns BTests with conditional code.  For one of these, the fix is add a `# @TEST-REQUIRES` to skip it for C++ compilation.  The other have conditional code that can now be removed, as it's testing for old Zeek versions.

UPDATE: I'd forgotten about a static analysis concern regarding a possible null pointer de-ref.  I don't believe it can happen in practice, but I've now added an `ASSERT` which should let the analyzer determine that it's not an issue.

### DIFF
--- a/testing/btest/language/at-if-lambda.zeek
+++ b/testing/btest/language/at-if-lambda.zeek
@@ -9,11 +9,7 @@ event zeek_init()
 	{
 	local make_epoch_result = function(pass_name: string): function(ts: time): time
 		{
-@if ( Version::at_least("4.1") )
 		return function [pass_name] (ts: time): time
-@else
-		return function (ts: time)
-@endif
 			{
 			print pass_name;
 			return ts;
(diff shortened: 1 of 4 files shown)


testing/btest/language/at-if-lambda.zeek
```diff
@@ -9,11 +9,7 @@ event zeek_init()
 	{
 	local make_epoch_result = function(pass_name: string): function(ts: time): time
 		{
-@if ( Version::at_least("4.1") )
```
awelzel, line 12: Hmm, this is the main piece of the test case and reproducer from the referenced #2075 :-)

So either we remove the two test parts that use `Version::at_least()`, or place a `TEST-REQUIRES` to skip the test in CPP mode. Slight preference for the latter. Or split it in two and just skip one, but not sure it's worth it.

awelzel (changes requested):

testing/btest/language/at-if-lambda.zeek
vpax: Ah, I missed that the whole point of this test is to have an `@if` inside a lambda! :-P  In that case, yes, a `TEST-REQUIRES` to skip it makes sense.  I thought it was just a now-vestigial test.

Do you want to make the tweak or should I go ahead and do so?

awelzel: I'll pick it up (and if I get it wrong you'll see next time around).

awelzel (approved):

PR #2919, opened 2023-04-03
iosource/Manager: Make packet sources readiness heuristic configurable

awelzel:
...and relax the current logic by default.

In a zero-packet scenario, we're currently polling non-selectable packet
sources 10 times in a row, then relax for 20usec. Roughly, this may result
in 500000  pkt_src->Process() calls per second (possibly somewhat less)
when there are no packets available. This is a lot.

Further, in low packet rate scenarios, e.g. 500kpps over 8 workers, an
individual workers should only see 62 kpps. Asking the packet source
much more often isn't very useful (and in fact queueing up packets may
be more efficient).

Add a heuristic to stop calling Process() poll interval times in a row
when a packet source has not yielded packets over the past
packet_source_inactivity_count times. This setting, together with
Pcap::non_fd_timeout, should allow to fine tune for better idle CPU
usage, but also behave better out-of-the box.

Relates to #2296.

### DIFF
--- a/src/iosource/PktSrc.cc
+++ b/src/iosource/PktSrc.cc
@@ -31,7 +31,7 @@ PktSrc::PktSrc()
 	{
 	have_packet = false;
 	// Small lie to make a new PktSrc look like the previous ExtractNextPacket() was successful.
-	had_packet = true;
+	idle_counter = 0;
 	errbuf = "";
 	SetClosed(true);
 	}
@@ -141,7 +141,7 @@ void PktSrc::Done()
 
 bool PktSrc::HasBeenIdleFor(double interval) const
 	{
-	if ( have_packet || had_packet )
+	if ( have_packet || idle_counter == 0 )
 		return false;
 
 	// Take the hit of a current_time() call now.
@@ -185,7 +185,8 @@ bool PktSrc::ExtractNextPacketInternal()
 
 	if ( ExtractNextPacket(&current_packet) )
 		{
-		had_packet = true;
+		if ( idle_counter != 0 )
+			idle_counter = 0;
 
 		if ( current_packet.time < 0 )
 			{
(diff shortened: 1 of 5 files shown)


awelzel:
@timwoj - wondering if you could see that.

This is from Slack conversations with @mgagliardo who reported much improved CPU usage after increasing the PktSrc timeout. Something like the proposed change should allow to keep a lowerish packet source timeout (configurable with #2913 anyway), but also back-off from asking the packet source for a packet over and over again.

> With your reported packet rates (~100kpps over 18 workers, correct?), an individual worker sees only ~5555 packet/second, but Zeek's IO loop logic is asking the packet source ~500000/s for a new packet (I hope my math is correct: The current logic calls Process() 10x in a row, then relaxes for 0.000020 seconds). The previous IO loop was much less aggressive IIUC and I'm debating a change to back-off.
It is much more efficient to ask the NIC less frequently (so it can queue up a few packets), but then extract all available packets at once (that's what the cherry picked PR improves) ad then "removing the 0" prolongs the queueing/batching time. This introduces a bit of processing latency for individual packets, but Zeek very much acceptable. We're not doing high-frequency trading or so. With the increased the sleep time, a Zeek worker  should now be only" 50000/s which is still a whole lot and why I think you might be okay even with longer sleep times so long you're not running out of buffer space.


https://zeekorg.slack.com/archives/CSZBXF6TH/p1680511770109069?thread_ts=1680178291.086299&cid=CSZBXF6TH

src/iosource/PktSrc.cc
```diff
@@ -207,13 +208,13 @@ bool PktSrc::ExtractNextPacketInternal()
...
 		// and detect it as idle.
-		if ( had_packet )
+		if ( idle_counter == 0 )
 			{
 			DBG_LOG(DBG_PKTIO, "source %s is idle now", props.path.c_str());
 			idle_at_wallclock = zeek::util::current_time(true);
 			}
 
-		had_packet = false;
+		idle_counter = std::max(idle_counter, idle_counter + 1);
```
timwoj, line 217: I assume this is to prevent overflow?

timwoj (approved):
I think this looks ok. How much has it been tested on links with a lot of traffic? I just want to make sure that while we're optimizing for the low-traffic cases, we're not causing bottlenecks for the high-traffic ones.

awelzel:
I'm going to close this for the time being. Once #2926 is out with Zeek 6.0 and reports indicate that non-selectable packet sources continue to show high CPU usage at low packet rates can revisit.

With 6.0 there'll be some low-level IO loop tweaking possible, too, just not an adaptive approach as sketched here.

PR #2924, opened 2023-04-05
ci: Add debian-12 / bookworm

awelzel:
The next version of Debian (bookworm) had a hard-freeze on
2023-03-16. Seems reasonable to have it in CI now.

https://lists.debian.org/debian-devel-announce/2023/03/msg00004.html

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -201,6 +201,14 @@ centos7_task:
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
 
+debian12_task:
+  container:
+    # Debian 12 (bookworm) EOL: (not yet released)
+    dockerfile: ci/debian-12/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
+  << : *SKIP_TASK_ON_PR
+
 debian11_task:
   container:
     # Debian 11 EOL: June 2026
--- a/ci/debian-12/Dockerfile
+++ b/ci/debian-12/Dockerfile
@@ -1,0 +1,37 @@
+FROM debian:bookworm
+
+ENV DEBIAN_FRONTEND="noninteractive" TZ="America/Los_Angeles"
+
+# A version field to invalidate Cirrus's build cache when needed, as suggested in
+# https://github.com/cirruslabs/cirrus-ci-docs/issues/544#issuecomment-566066822
+ENV DOCKERFILE_VERSION 20230405
+
+RUN apt-get update && apt-get -y install \
+    bison \
+    bsdmainutils \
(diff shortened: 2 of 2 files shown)


timwoj (approved):

ckreibich:
Just a reminder here that if we add a platform in Zeek proper, we also should do it in the submodules ...

- zeek-archiver
- zeek-aux
- broker
- binpac

... and ideally cherry-pick the introduction back into the relevant release branches. It's tedious stuff — suggestions for streamlining are very welcome. I don't think we can fully centralize any of this into a CI-dedicated submodule since the above submodules are all _just_ a bit different. But perhaps there's scope for scripting the introduction of a new platform, or some such.

I'm happy to write a CI maintenance wiki page that at least walks through the manual steps if folks think it'd be useful.

timwoj:
> I don't think we can fully centralize any of this into a CI-dedicated submodule since the above submodules are all just a bit different.

We can certainly make the Dockerfiles common. Some VMs would include more in them than they need for individual repos, but at least we would only have to manage a single set of them. The build scripts and .cirrus.yml are where everything starts to suck.

bbannier:
Submodules missing from that list are also Spicy and spicy-plugin. Since both of these modules generate C++ code adding a new platform could also potentially mean adding support for a different compiler (version). Most of the time this just works, but not validating this with the project test suites always makes me slightly nervous. 

Another approach I tried to push for in the past was to add more integration testing of submodules into this repo, i.e., instead of blindly pulling in say broker here also add a job which runs broker's full test suite e.g., weekly/nightly/… (the component tests in this repo only do smoke tests, at least for Spicy). With that zeek/zeek would somehow validate that the components it bundles and claims to support actually work.

awelzel:
Uff...

> I don't think we can fully centralize any of this into a CI-dedicated submodule since the above submodules are all just a bit different. But perhaps there's scope for scripting the introduction of a new platform, or some such.
> [...]
> We can certainly make the Dockerfiles common. Some VMs would include more in them than they need for individual repos, but at least we would only have to manage a single set of them. The build scripts and .cirrus.yml are where everything starts to suck.

Before we script the addition of a new platform into a .cirrus.yaml and across projects, maybe would be good to look more seriously at starlak. It seems powerful enough to load a centralized list of platforms from an external repo and then produce tasks based off of that.

ckreibich:
> Submodules missing from that list are also Spicy and spicy-plugin

I'm glad you bring that up, Benjamin — I always considered these out of scope due to the magnitude of the Spicy project. It's not been an oversight that I never submitted CI PRs to these. It's good to know that you'd like to see these treated the same way (which also makes sense due to Spicy's inclusion).

PR #2927, opened 2023-04-06
Fix a long-standing bug in the math around continue_processing

timwoj:
@initconf managed to track down an error in one of his scripts that was reading an input file. It called `continue_processing()` once the script finished being read, even if processing wasn't suspended. Unfortunately a very long-standing bug (all the way back to SVN at least!) in `continue_processing()` caused the stack counter for suspension to roll over to negative. On top of that, we only checked whether the counter was zero to determine whether we were suspended, so negative values counted as being suspended.

This also adds a new BIF for retrieving the current status of suspension, since I needed it for the added test.


timwoj:
This is a candidate to backport into the 5.0/5.2 releases we're about to do.

awelzel (approved):
As in Slack, I'm not sure about the stacking/queueing of suspends being a good idea. In fact, someone calling `continue_processing()` twice may be reason to warn them, but also approving because this affects almost no one out there :-)

Putting it on the 5.0/5.2 list.

PR #2929, opened 2023-04-06
Add timestamps to events

J-Gras:
This adds timestamps to events. Each event is stamped to the time it is intended to be handled. A new BiF, `current_event_time()`, can be used to obtain the current event's timestamp. The original motivation was to use this for precise manual control of network time in cluster setups. However, this might also be handy to analyze overload situations (i.e. `network_time() > current_event_time()`).

The PR requires the corresponding broker update.


src/broker/Manager.h
```diff
@@ -182,18 +182,20 @@ class Manager : public iosource::IOSource
 	 * Peers advertise interest by registering a subscription to some prefix
 	 * of this topic name.
 	 * @param name the name of the event
+	 * @param ts the timestamp the event is intended to be executed
 	 * @param args the event's arguments
 	 * @return true if the message is sent successfully.
 	 */
-	bool PublishEvent(std::string topic, std::string name, broker::vector args);
+	bool PublishEvent(std::string topic, std::string name, double ts, broker::vector args);
```
awelzel, line 189: This is in the `zeek::` namespace, so we can't insert that ts parameter here. Same arguments as for `Call()`.

I'm also stumbling over putting `ts` right between `name` and `args`. `name` is invoked with `args`, so they have a close relationship, while `ts` seems to add some additional info about when. Might just be me reading too much into it :-)

src/zeek.bif
```diff
@@ -354,6 +354,19 @@ function set_network_time%(nt: time%): bool
 	return zeek::val_mgr->True();
 	%}
 
+## Returns the timestamp of the last raised event. The timestamp reflects the
+## network time the event was intended to be executed. For scheduled events,
+## this is the time the event was scheduled to. For any other event, this is
```
awelzel, line 359: I think this should be "scheduled for", but non-native :sweat_smile:

testing/btest/language/event-ts-scheduled.zeek
```diff
@@ -0,0 +1,18 @@
+# @TEST-EXEC: zeek -b -r $TRACES/ticks-dns-1hr.pcap %INPUT > out
+# @TEST-EXEC: btest-diff out
+
+global runs = 0;
+
+event test()
+	{
+	print fmt("[%D] Test was scheduled at %D", network_time(), current_event_time());
```
awelzel, line 8: Because this is a scheduled event, I *think* this should say "scheduled for" instead of "scheduled at" ? It could be interesting to add the scheduled_at value by passing `network_time()` as an event parameter to `test()` and outputting that, too.  Passing the interval as event parameter and including it in the output could be helpful when trying to make sense of the baseline, as well :-)

testing/btest/broker/remote_event_auto_ts.zeek
```diff
@@ -0,0 +1,82 @@
...
+
+
+# @TEST-START-FILE recv.zeek
+
+redef exit_only_after_terminate = T;
+
+global msg_count = 0;
+
+event zeek_init()
+		{
```
awelzel, line 60: Nit, this is double tab indented for a few lines.

testing/btest/broker/remote_event_ts.zeek
```diff
@@ -0,0 +1,77 @@
...
+
+
+# @TEST-START-FILE recv.zeek
+
+redef exit_only_after_terminate = T;
+
+global msg_count = 0;
+
+event zeek_init()
+		{
```
awelzel, line 58: Same double indentation nit.

src/broker/Manager.cc
```diff
@@ -681,7 +696,10 @@ bool Manager::PublishEvent(string topic, RecordVal* args)
 		xs.emplace_back(data_val->data);
 		}
 
-	return PublishEvent(std::move(topic), event_name, std::move(xs));
+	// At this point we come from script-land. This means that publishing of the event was
+	// explicitly triggered. Hence, the timestamp is set to the current network time. This also
+	// means that timestamping cannot be manipulated from script-land for now.
```
awelzel, line 701: Is that comment anymore special here in Broker than it is for `event test()` or `schedule 1min { test() };`? The timestamps of the above are just based on network_time and can't be manipulated either, otherwise I missed something somewhere.

src/broker/Manager.cc
```diff
@@ -652,7 +667,7 @@ bool Manager::PublishEvent(string topic, std::string name, broker::vector args)
 		return true;
 
 	DBG_LOG(DBG_BROKER, "Publishing event: %s", RenderEvent(topic, name, args).c_str());
-	broker::zeek::Event ev(std::move(name), std::move(args));
+	broker::zeek::Event ev(std::move(name), ts, std::move(args));
```
awelzel, line 670: Similarly as for the other API changes: Is this breaking backward compat of `broker::zeek::Event`?

Probably better for zeek/broker#331

awelzel (changes requested):
This looks great. Most comments about avoiding API breakage in the `zeek` namespace and some naming suggestions, but otherwise seems fairly straightforward :+1:

testing/btest/language/event-ts-scheduled.zeek
J-Gras: I like that idea. Makes it indeed easier to read the baseline.

src/EventHandler.h
```diff
@@ -22,42 +23,75 @@ class EventHandler
 public:
 	explicit EventHandler(std::string name);
 
-	const char* Name() const { return name.data(); }
+	const char* Name() const
```
awelzel, line 26: Is there a chance you're using a more recent clang-format version than the one in `.pre-commit-config.yaml` (13.0.0) or you didn't run `pre-commit run` before/after the commit?

Running `pre-commit run -a` within Zeek reverts most of the spurious newline changes. Unfortunately it doesn't seem Zeek's CI setup currently runs the pre-commit check, but I haven't understood why.

src/Event.h
```diff
@@ -7,6 +7,7 @@
 
 #include "zeek/Flare.h"
 #include "zeek/IntrusivePtr.h"
+#include "zeek/RunState.h"
```
awelzel, line 10: We should avoid that RunState.h here and in EventHandler.h. It pull in all the runstate "globals" and are pretty high up headers included/used in lots of places.

Can do instead what Val.h does, declaring network_time as extern and limiting the exposure that way:
https://github.com/zeek/zeek/blob/75245bd3656fdaf7db229d8e476677bb13fdefd4/src/Val.h#L58-L64

I'm happy to fix that up during the merge, whatever you prefer.

awelzel (commented):
LGTM. Two more comments, but these can be straightened out during the merge, too, as you prefer.

If you do another round, would recommend a bit of rebase/squash because with the parameter reordering the history could be confusing to look at.

This is now otherwise waiting/blocked on zeek/broker#331, so not yet selecting "Approve".

src/EventHandler.h
timwoj: Cirrus doesn't, but there should be a Github Action that runs it. It doesn't appear to have run on this PR though.

awelzel: It ran previously, actually: https://github.com/zeek/zeek/pull/2929/checks?check_run_id=12577202183

Not sure what's going on. Maybe on a re-push it'll :crossed_fingers:

J-Gras:
```
$ clang-format --version
clang-format version 16.0.1
```
I switched to using `pre-commit` :)

src/Event.h
```diff
@@ -31,6 +36,7 @@ class Event final : public Obj
 	analyzer::ID Analyzer() const { return aid; }
 	EventHandlerPtr Handler() const { return handler; }
 	const zeek::Args& Args() const { return args; }
+	double Time() const { return timestamp; }
```
ckreibich, line 39: I'm not as deeply in the time management code as you, but I'm noticing that we use "time", "timestamp", and "ts" all to convey timestamps. Might help to try to settle on one.

J-Gras: Please address this during merge. As you might guess, my vote is on "timestamp" and "ts" where an abbreviation seems appropriate ;)

awelzel:
@J-Gras , this is the diff I have locally for the compat test case after changing the Python interface

```
+import datetime
 import os
 import sys
 
@@ -79,7 +80,11 @@ with broker.Endpoint() as ep, \
        ep.publish(broker_topic, my_event)
 
        print("send event with timestamp")
-       my_event = broker.zeek.Event("my_event", "with ts", ts=23.0)
+       ts = datetime.datetime.fromtimestamp(23.0, broker.utc)
+       metadata = {
+               broker.zeek.Metadata.NETWORK_TIMESTAMP: ts,
+       }
+       my_event = broker.zeek.Event("my_event", "with ts", metadata=metadata)
        ep.publish(broker_topic, my_event)
 
        ep.shutdown()
```

Mostly there's no convenience `ts` parameter on the Python layer and also need typing as datetime / broker.Timestamp now (consistency with timestamps being converted to `datetime.datetime` in Python).

J-Gras:
> this is the diff I have locally for the compat test case after changing the Python interface

Thanks! I've updated the PR.

awelzel:
Hey - the following interface commit has a test using websockets for sending/receiving network timestamps in case you'd want to pick that up:
https://github.com/zeek/zeek/commit/d4b8dd3e028bd21f64039ae7d69893453702c8e4

PR #2936, opened 2023-04-08
ListVal::Append: Fix memory leak after re-using TypeList instances

awelzel:
Commit 24c606b4df92f9871964c5bcb2fe90e43a177b1f added a TypeManager to re-use TypeList instances of the same type (particularly TYPE_ANY) and thereby caused a pretty bad memory leak. It missed that for non-pure ListVals, appending a Val would also append that element's type to the type list of the Val. This resulted in the pre-allocated TYPE_ANY to bloat whenever a ListExpr was eval'ed.

Due to the above error not causing any other apparent issues in the test suite (other than the memory leak) and neither causing any test failures with the type removed, I'm unclear why the logic was there to begin with.

### DIFF
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -1988,12 +1988,16 @@ TypeManager::TypeManager()
 		TypeTag tag = static_cast<TypeTag>(i);
 		TypePtr pure_type = tag == TYPE_ANY ? nullptr : base_type(tag);
 		base_list_types[tag] = make_intrusive<zeek::TypeList>(pure_type);
+		base_list_types[tag]->Append(pure_type);
 		}
 	}
 
 const TypeListPtr& TypeManager::TypeList(TypeTag t) const
 	{
 	assert(t >= 0 && t < NUM_TYPES);
+	// This assert is to catch type lists mutation due to
+	// AsTypeList() removal of const being a dangerous trap.
+	assert(base_list_types[t]->GetTypes().size() == 1);
 	return base_list_types[t];
 	}
 
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -1220,9 +1220,7 @@ void ListVal::Append(ValPtr v)
 			Internal("heterogeneous list in ListVal::Append");
 		}
 
-	const auto& vt = v->GetType();
 	vals.emplace_back(std::move(v));
-	type->AsTypeList()->Append(vt);
 	}
 
 TableValPtr ListVal::ToSetVal() const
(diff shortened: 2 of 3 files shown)


awelzel:
I'm not super happy with the fix as because I don't fully understand what might have been the intention. As written in the commit message, seems having way too many types in a singleton non-pure list or just having the any/nullptr makes no difference. @vpax - any reaction from your side?

I had been testing something with JavaScript and garbage collection and was dumb folded why memory usage would constantly grow even with just Zeek. Test setup is replaying ~50kpps of recorded SSL traffic to 100 Alexa Top Sites pages over and over again.

Without this fix, memory grows fairly quickly to 500MB, then 700MB and eventually 1GB plus. After the fix, memory stays at a pretty constant 280MB.

For the record, mostly running as follows, then `jeprof` directly.
```
$ export MALLOC_CONF="prof:true,prof_final:true,prof_prefix:jez.out,lg_prof_interval:30"
$ /opt/zeek-dev-prod/bin/zeek -b misc/stats base/misc/version base/protocols/ssl  -C -i fuzzout LogAscii::use_json=T Stats::report_interval=1sec    -f 'port 443'
jeprof /opt/zeek-dev-prod/bin/zeek jez.out.182221.*.heap 
```

Screenshot-20230408-232943

vpax:
@awelzel are you flagging me because of my general interest in AST things, or for a particular reason (in which case I need more context).  I'm not able to fully tease apart just what you're flagging above, in particular "seems having way too many types in a singleton non-pure list or just having the any/nullptr makes no difference" - there are enough negatives, and ambiguity about pre- https://github.com/zeek/zeek/commit/24c606b4df92f9871964c5bcb2fe90e43a177b1f vs. pre- https://github.com/zeek/zeek/pull/2936/commits/2601d6f85669adfaff1f3dd5d4fa52686e30ab03 that I'm not able to follow ...

awelzel:
> @awelzel are you flagging me because of my general interest in AST things, or for a particular reason (in which case I need more context).

Mostly the former. Maybe the re-use of TypeLists by ListVals with 2601d6f . It did reduce allocations significantly though so it would be nice to keep.

>  there are enough negatives, and ambiguity about pre- 24c606b vs. pre- 2601d6f that I'm not able to follow ...

...thanks, if the proposed change here doesn't strike you as completely off, I'll keep it as is.

vpax:
> ...thanks, if the proposed change here doesn't strike you as completely off, I'll keep it as is.

It's the opposite - I'm saying that unfortunately I can't understand just what the change is from your description ...

A general observation: it's fine to manage pure `TypeList`s from a manager, like done for base types etc.  For impure lists, however, it should be the `TypeList` class that handles appending to them, not a manager, since there are all sorts of combinations.  Looking at the current code, this doesn't seem to be the case, so I'm pretty confused, and related to that I wouldn't presume that the manager was added to particularly handle the `TYPE_ANY`/impure case, I'd guess it was added to address the common case of single-element table indexing, which needs a pure `TypeList`.

One other observation: I believe impure `TypeList`s are mainly associated with indexing multi-index tables (and sets).  If so, then the types for those containers could contain a pre-constructed `TypeList` that's just used for each of those, saving a lot of allocations.

awelzel:
> One other observation: I believe impure TypeLists are mainly associated with indexing multi-index tables (and sets). If so, then the types for those containers could contain a pre-constructed TypeList that's just used for each of those, saving a lot of allocations.

This was the goal of 24c606b, but I've clearly not fully understood implications. #2949 reverts the original change. I'm closing this PR.

PR #2937, opened 2023-04-09
Record access optimizations

vpax:
The bulk of this PR provides optimizations for hot spots in logging and compiled-to-C++ scripts by switching them from regular `RecordVal` field accesses to lower-level `ZAM`-oriented accesses that avoid creating lots of ephemeral `Val` objects.

The PR also optimizes construction of `connection` records by leveraging the fact that when a `RecordVal` is created, its aggregate fields are already initialized.  Populating those fields rather than replacing them with separately constructed ones saves some unnecessary creation/deletion of unused aggregates.  (I have in mind a more generalized version of how to better leverage this, but will save it for a separate effort.)

Finally, in the process of analyzing hotspots I noticed that under some circumstances the AST construction can produce redundant table/set coercions, so I added a tweak to avoid those.


vpax:
I'm closing this to prepare another version that incorporates the fix mentioned above regarding avoiding redundant construction of aggregates.  I also need to run down the ASAN issue flagged by CI.

PR #2938, opened 2023-04-10
Optimizations for "record" creation / access

vpax:
I've been using a large operational PCAP to do some execution profiling, looking for ways to improve execution times.  The changes here together yield about a 2.5-3% improvement in my timings.  They cover four areas:

- Converting a hot-spot in the logger to avoid constructing ephemeral `Val` objects when all that's needed is access to the underlying values of record fields. 
- A similar change for scripts-compiled-to-C++ to directly access record fields rather than going through `Val` objects.
- Changing the creation of `RecordVal`s to defer initialization of aggregate fields until they are accessed.  This turns out to be the biggest win of the changes, because there's a common pattern of creating a record and then assigning separately constructed aggregates to its aggregate fields.  In those instances, the original initialization (which isn't super cheap) is immediately discarded unused, so it's beneficial to not create them in the first place.
- A minor tweak to avoid constructing redundant record coercions.


src/Type.cc
```diff
@@ -992,39 +992,102 @@ void TypeDecl::DescribeReST(ODesc* d, bool roles_only) const
 		}
 	}
 
-// The following tracks how to initialize a given field, for fast execution
-// of Create().
+class DirectFieldInit final : public FieldInit
```
timwoj, line 995: Can you add some comments about what all of these individual init types do? It's kind of hard to track.

src/ZVal.h
```diff
@@ -10,6 +10,7 @@ namespace zeek
 	{
 
 class AddrVal;
+class EnumVal;
```
timwoj, line 13: How the heck did that build before?

vpax: HAH yeah I wondered the same thing.  It's presumably because previously the only times `ZVal.h` was included, `Val.h` had already been included.  The new include for `Type.h` doesn't have that property, however (to avoid circularities).

src/Type.cc
vpax: Sounds good.

src/Type.h
```diff
@@ -683,12 +690,8 @@ class RecordType final : public Type
 
 	void AddFieldsDirectly(const type_decl_list& types, bool add_log_attr = false);
 
-	/**
-	 *
-	 * Populates a new instance of the record with its initial values.
-	 * @param r  The record's underlying value vector.
-	 */
-	void Create(std::vector<std::optional<ZVal>>& r) const;
+	const auto& FieldInits() const { return field_inits; }
```
0xxon, line 693: Could you perhaps add some documentation for the two getter functions here? I know there is a comment below that explains this a bit - but it is easy to miss this, when just looking at the functions.

vpax: Likewise, sounds good.

Are there more reviewing comments potentially coming (a question also for @timwoj), or should I go ahead with the commenting updates?

vpax:
Just pushed a set of changes for both commenting per the above requests and some tightening in terms of both naming and making it clear that the record field initializers are only meant for use by `RecordVal`s.

awelzel (commented):
Mostly left "random" comments. The one biggie is RecordType::Create() being an API break. I let someone with authority decide whether that's okay. I doubt there are users, but according to our API stability guarantees it's not okay.

awelzel:
The compiler on windows doesn't like something here:
```
C:\zeek\src\include\zeek/IntrusivePtr.h(128): error C2664: 'void zeek::Unref(zeek::Obj *)': cannot convert argument 1 from 'zeek::TypeVal *' to 'zeek::Obj *'
C:\zeek\src\include\zeek/IntrusivePtr.h(128): note: Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast
C:\zeek\src\include\zeek\Obj.h(185): note: see declaration of 'zeek::Unref'
C:\zeek\src\include\zeek/IntrusivePtr.h(120): note: while compiling class template member function 'zeek::IntrusivePtr<zeek::TypeVal>::~IntrusivePtr(void)'
C:\zeek\src\include\zeek/ZVal.h(94): note: see reference to function template instantiation 'zeek::IntrusivePtr<zeek::TypeVal>::~IntrusivePtr(void)' being compiled
C:\zeek\src\include\zeek/ZVal.h(94): note: see reference to class template instantiation 'zeek::IntrusivePtr<zeek::TypeVal>' being compiled
```

src/Type.h
```diff
@@ -717,9 +717,22 @@ class RecordType final : public Type
 
 	void DoDescribe(ODesc* d) const override;
 
-	// Maps each field to how to initialize it.  Uses pointers due to
-	// keeping the FieldInit definition private to Type.cc (see above).
-	std::vector<FieldInit*> field_inits;
+	// Field initializations that can be deferred to first access,
+	// beneficial for fields that are separately iniitialized prior
+	// to first access.
+	std::vector<std::optional<std::unique_ptr<FieldInit>>> deferred_inits;
```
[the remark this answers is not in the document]
ckreibich: My read is that it's just a construct to meaningfully represent a deferred-init in the case of duplicate record fields until we error out after parsing. But a clarifying comment would be helpful.

(Total nit-pick: "iniitialized" in the comment, btw)

src/script_opt/CPP/RuntimeOps.h
```diff
@@ -109,6 +115,42 @@ inline ValPtr field_access__CPP(const RecordValPtr& rec, int field)
...
+#define VP_FIELD_ACCESS(type, zaccessor)                                                           \
+	inline type##Ptr field_access_##type##__CPP(const RecordValPtr& r, int field)                  \
+		{                                                                                          \
+		auto rv = CPPRuntime::RawOptField(r, field);                                               \
+		if ( rv )                                                                                  \
+			return {NewRef{}, rv->zaccessor()};                                                    \
+		return cast_intrusive<type>(field_access__CPP(r, field));                                  \
+		}
+
+VP_FIELD_ACCESS(StringVal, AsString)
```
ckreibich, line 142: "VP" for "ValPointer", yes? :-)

ckreibich (changes requested):
Cool find! It all looks good to me except for the deprecation aspect, which I feel we need to become more serious about.

I find the new code somewhat harder to follow, but that's the nature of such optimizations.

src/Type.h
```diff
@@ -683,13 +690,6 @@ class RecordType final : public Type
 
 	void AddFieldsDirectly(const type_decl_list& types, bool add_log_attr = false);
 
-	/**
-	 *
-	 * Populates a new instance of the record with its initial values.
-	 * @param r  The record's underlying value vector.
-	 */
-	void Create(std::vector<std::optional<ZVal>>& r) const;
```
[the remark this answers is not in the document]
vpax: Yeah, it shouldn't have been public in the first place, but given that it was (my bad), I've reinstated it (with deprecation) for the upcoming push.  I also switched the `protected` to `private` given the `final`.

src/Type.h
```diff
@@ -600,10 +601,16 @@ class TypeDecl final
 
 using type_decl_list = PList<TypeDecl>;
 
-// The following tracks how to initialize a given field.  We don't define
-// it here because it requires pulling in a bunch of low-level headers that
-// would be nice to avoid.
-class FieldInit;
+// The following tracks how to initialize a given field.
+
+class FieldInit
```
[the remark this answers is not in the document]
vpax: Good call-out, addressed in the pending push.

src/Type.h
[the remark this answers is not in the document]
vpax: Actually that was vestigial (and not for error-handling).  I've removed the `std::optional` and switched things to use nil pointers as nothing-to-do. (Also fixed the typo!)

src/Val.h
```diff
@@ -1403,8 +1423,10 @@ class RecordVal final : public Val, public notifier::detail::Modifiable
 	static void DoneParsing();
 
 protected:
+	friend class zeek::logging::Manager;
```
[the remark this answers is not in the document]
vpax: I'm a bit hesitant here, because users of `RawOptField` are responsible for memory management, which is easy to get wrong.  So I'd rather there be a high bar for access to it.

`VectorVal::RawVec` is public just because there are numerous points where ZAM needs access to it, and also potentially `-O C++` code (not yet implemented).

src/script_opt/CPP/RuntimeOps.h
vpax: Yep!

ckreibich:
Thanks Vern. The Tumbleweed failure will go away if your rebase your branch onto master. Not sure about the Windows one...

ckreibich (approved):

vpax:
> The Tumbleweed failure will go away if you rebase your branch onto master

FYI, I tried, but wound up with a mess of a tree that wouldn't build :-(.  Chatting with Tim, it wasn't clear just what had gone wrong, so I decided to punt on that issue.  After Tim confirmed that the MSVC problem was deeper than my attempted fixes, I came up with the simpler approach that I've now pushed (and reverted the old one), which seems to do the trick - phew!

src/Val.h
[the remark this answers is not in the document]
awelzel: That is fair and I agree. I now remember there's also the GetFieldAs() API (and for vectors the type specific At() methods), so I can try to leverage these first.

awelzel:
Hmm - I merged without updating NEWS: We made something faster, it might be NEWS worthy. From a user perspective going into too many details might not be worth it though and saying "it should be faster" seems also a bit soft :man_shrugging: 

cc @timwoj  - thoughts?

PR #2910, opened 2023-03-30
Fixes for building on GCC 13

timwoj:
While running test builds on OBS for https://github.com/zeek/zeek/pull/2799, we ran into a problem where the OpenSUSE Tumbleweed VM updated to GCC 13. This causes a few different build failures in Zeek:

- The [recently-merged flow scopes](https://github.com/zeek/broker/pull/307) feature in Broker had a use-after-free bug that didn't show up in testing. This PR updates the broker submodule to pull in that plus a few other changes.
- Spicy fails to build due to a new warning in one of their dependencies. This PR adds a new flag to the Zeek build to avoid that error.
- `WeirdState.h` needs `<cstdint>` included to find uint64_t correctly.

PR #2939, opened 2023-04-11
Only use -Wno-changes-meaning if it exists

timwoj:
This is a follow-up to https://github.com/zeek/zeek/pull/2910 that only uses the `-Wno-changes-meaning` flag if it exists, since some other compilers complain very loudly (such as on macOS) if you use a flag that doesn't exist.


ckreibich (approved):
Sigh. I was nervous about this but assumed GCC's behavior (of ignoring unsupported `-Wno-*`) is the norm. Fingers crossed.

Issue #2920, opened 2023-04-03
Removing initialization requirement for container-valued options


vpax:
Currently, you can't write:
```
option foo: set[count] &redef;
```
where `foo` is an option that defaults to an empty set.  Instead, you need to use:
```
option foo: set[count] = set() &redef;
```
or else you get an `option variable must be initialized` error.

That error makes sense for non-containers.  However, containers are defined to start off initialized-but-empty, rather than uninitialized.  So I'd like to drop the requirement of including explicit initializations for them, since I wind up needing do-nothing initializers like in the example above fairly often.

PR #2940, opened 2023-04-11
GH-2920: Don't warn on uninitialized container options

timwoj:
This seems like a really simple fix for this, but it's passing the additional test and the variables appear initialized correctly.


0xxon (changes requested):
This now succeeds on the following zeek script - where it should warn:

```
@load misc/stats

type TestRecord: record {
	a: count;
	b: Stats::Info &optional;
};

option foo: TestRecord &redef;

event zeek_init()
	{
	print foo;
	}
```

vpax:
In my original issue, I should have clarified that it's for table/set/vector containers.  I agree with @0xxon that it's not viable for records.  But the others all have default notions of "empty" which should work find.

0xxon (approved):
Looks good to me. Would it perhaps make sense to include a test - like the script above - just to make sure that this keeps warning when one tries doing it with records?

PR #2950, opened 2023-04-13
broker/messaging: Runtime type checks for pool and key

### LINKED ISSUE #2935: Segfault in 5.2.0:  zeek::Obj::BadTag - wrong use of Cluster::publish_hrw causes proxies to crash
So I stumbled into another bug which causes a Segfault - Attached  script (see below) replicates the bug and also its pretty clear what goofup caused it.

In this specific instance - I can crash proxies to generate a coredump. I haven't fully checked implications on the workers. 

Here is the coredump:

```
Core was generated by `/home/bro/zeek-5.2.0/bin/zeek -U .status -p zeekctl -p zeekctl-live -p local -p'.
Program terminated with signal SIGABRT, Aborted.
Sent by thr_kill() from pid 42746 and user 20201.
#0  0x00000008020de69a in thr_kill () from /lib/libc.so.7
[Current thread is 1 (LWP 102425)]
(gdb) bt
#0  0x00000008020de69a in thr_kill () from /lib/libc.so.7
#1  0x00000008020dcaf4 in raise () from /lib/libc.so.7
#2  0x0000000802052719 in abort () from /lib/libc.so.7
#3  0x0000000000a5ead7 in zeek::Reporter::FatalErrorWithCore (this=<optimized out>, fmt=<optimized out>) at /home/bro/install/zeek-5.2.0/src/Reporter.cc:152
#4  0x0000000000a6b38f in zeek::Obj::BadTag (this=<optimized out>, msg=<optimized out>, t1=<optimized out>, t2=<optimized out>)
    at /home/bro/install/zeek-5.2.0/src/Obj.cc:112
#5  0x0000000000ac79cb in zeek::Val::AsRecordVal (this=0x803061a90) at /home/bro/install/zeek-5.2.0/src/Val.cc:78
#6  0x00000000009de019 in zeek::detail::FieldExpr::Fold (this=0x803101e80, v=0x19019) at /home/bro/install/zeek-5.2.0/src/Expr.cc:3253
#7  0x00000000009ce535 in zeek::detail::UnaryExpr::Eval (this=0x803101e80, f=<optimized out>) at /home/bro/install/zeek-5.2.0/src/Expr.cc:651
#8  0x00000000009ce4f6 in zeek::detail::UnaryExpr::Eval (this=0x8030aa360, f=0x0) at /home/bro/install/zeek-5.2.0/src/Expr.cc:615
#9  0x00000000009d3757 in zeek::detail::SizeExpr::Eval (this=<optimized out>, f=0x0) at /home/bro/install/zeek-5.2.0/src/Expr.cc:1617
#10 0x00000000009ceb65 in zeek::detail::BinaryExpr::Eval (this=0x8030ad640, f=0x811873200) at /home/bro/install/zeek-5.2.0/src/Expr.cc:704
#11 0x0000000000a9e09e in zeek::detail::ExprStmt::Exec (this=0x8030aa3c0, f=0x811873200, flow=@0x7fffffffddec: zeek::detail::FLOW_NEXT)
    at /home/bro/install/zeek-5.2.0/src/Stmt.cc:433
#12 0x0000000000aa406a in zeek::detail::StmtList::Exec (this=<optimized out>, f=0x811873200, flow=@0x7fffffffddec: zeek::detail::FLOW_NEXT)
    at /home/bro/install/zeek-5.2.0/src/Stmt.cc:1697
#13 0x0000000000a392b9 in zeek::detail::ScriptFunc::Invoke (this=<optimized out>, args=0x7fffffffde48, parent=<optimized out>)
    at /home/bro/install/zeek-5.2.0/src/Func.cc:416
#14 0x0000000000a16c51 in zeek::BifFunc::Cluster::publish_hrw_bif (frame=0x811873500, BiF_ARGS=0x7fffffffe038)
    at /home/bro/install/zeek-5.2.0/src/broker/messaging.bif:227
#15 0x0000000000a3b297 in zeek::detail::BuiltinFunc::Invoke (this=0x8034f6700, args=<optimized out>, parent=<optimized out>)
    at /home/bro/install/zeek-5.2.0/src/Func.cc:717
#16 0x00000000009e9257 in zeek::detail::CallExpr::Eval (this=0x8057831a0, f=0x811873500) at /home/bro/install/zeek-5.2.0/src/Expr.cc:4716
#17 0x0000000000a9e09e in zeek::detail::ExprStmt::Exec (this=0x8057831f0, f=0x811873500, flow=@0x7fffffffe27c: zeek::detail::FLOW_NEXT)
    at /home/bro/install/zeek-5.2.0/src/Stmt.cc:433
#18 0x0000000000aa406a in zeek::detail::StmtList::Exec (this=<optimized out>, f=0x811873500, flow=@0x7fffffffe27c: zeek::detail::FLOW_NEXT)
    at /home/bro/install/zeek-5.2.0/src/Stmt.cc:1697
#19 0x0000000000a392b9 in zeek::detail::ScriptFunc::Invoke (this=<optimized out>, args=0x807086a60, parent=<optimized out>)
    at /home/bro/install/zeek-5.2.0/src/Func.cc:416
#20 0x00000000009a644d in zeek::EventHandler::Call (this=0x8057949a0, vl=<optimized out>, no_remote=<optimized out>)
    at /home/bro/install/zeek-5.2.0/src/EventHandler.cc:106
#21 0x00000000009a4da9 in zeek::Event::Dispatch (this=0x807086a40, no_remote=false) at /home/bro/install/zeek-5.2.0/src/Event.cc:56
#22 0x00000000009a53d6 in zeek::EventMgr::Drain (this=0x10d4528 <zeek::event_mgr>) at /home/bro/install/zeek-5.2.0/src/Event.cc:157
#23 0x0000000000a8c1ae in zeek::run_state::detail::run_loop () at /home/bro/install/zeek-5.2.0/src/RunState.cc:338
#24 0x00000000007ec45d in main (argc=<optimized out>, argv=<optimized out>) at /home/bro/install/zeek-5.2.0/src/main.cc:102
```

proxies crash:
```
Name         Type    Host             Pid     VSize  Rss  Cpu   Cmd
logger       logger  localhost        42453   465M   112M  33%  zeek
manager      manager localhost        42519   414M    89M   6%  zeek
proxy-1      proxy   localhost        <not running>
proxy-2      proxy   localhost        <not running>
proxy-3      proxy   localhost        <not running>
proxy-4      proxy   localhost        <not running>
worker-1     worker  localhost        43009     1G     1G 178%  zeek
worker-2     worker  localhost        43047     1G     1G 179%  zeek
worker-3     worker  localhost        43088     1G     1G 177%  zeek
worker-4     worker  localhost        43061     1G     1G 175%  zeek
```

Attached script can replicate the crash and coredumps for you.

```
$ cat bug.zeek
module Test;

redef Log::print_to_log = Log::REDIRECT_STDOUT;

export {
        global send_to_proxy: event(cid: conn_id);
        global send_to_manager: event(cid: conn_id);
}

function get_subnet(ip: addr): subnet
{
        local scanner_subnet: subnet;

        if ( is_v6_addr(ip) )
                scanner_subnet = mask_addr(ip, 64);
        else
                scanner_subnet = mask_addr(ip, 24);

        return scanner_subnet;
}

event connection_state_remove(c: connection)
{
        local cid = c$id;

        # just to make log less noisy
        if ( cid$orig_h !in 1.0.0.0 / 8 )
                return;

@if ( Cluster::is_enabled() )
        local u = get_subnet(c$id$orig_h);
        Cluster::publish_hrw(Cluster::proxy_pool, u, Test::send_to_proxy, c$id);
@else
        event Test::send_to_proxy(c$id);
@endif
}

event Test::send_to_proxy(cid: conn_id)
{
        print fmt("running on: %s %s", Cluster::node, cid);

@if ( Cluster::is_enabled() )
        local u = cid$resp_p;

        # I know this should be:
        # Broker::publish(Cluster::manager_topic, Test::send_to_manager, cid);
        # but somehow I managed to goof up - and below causes the segfault

        Cluster::publish_hrw(Cluster::manager_topic, u, Test::send_to_manager, cid);
@else
        event Test::send_to_manager(c$id);
@endif
}

event Test::send_to_manager(cid: conn_id)
{
        print fmt("running on %s %s", Cluster::node, cid);
}
```

awelzel:
publish_hrw() and publish_rr() are excluded from type checking due to their variadic nature. Passing wrongly typed arguments previously triggered an abort, now the result is runtime errors. This isn't great, but it's better than crashing Zeek.

This is mostly "crash avoidance" as typing for variadic functions doesn't exist.

Addition: After extending bifcl a bit, this also reverts `cat_sep()` to a variadic function with two required parameters which looks much better for documentation purposes, though the actual behavior hasn't improved here.

Closes #2935

Related bifcl PR: https://github.com/zeek/bifcl/pull/26


src/broker/messaging.bif
```diff
@@ -219,10 +241,17 @@ function Cluster::publish_rr%(pool: Pool, key: string, ...%): bool
 function Cluster::publish_hrw%(pool: Pool, key: any, ...%): bool
 	%{
 	static zeek::Func* topic_func = nullptr;
+	static zeek::RecordTypePtr pool_type = nullptr;
```
timwoj, line 244: Is this used?

src/broker/messaging.bif
```diff
@@ -179,13 +189,25 @@ type Cluster::Pool: record;
 ##       to it.
 ##
 ## Returns: true if the message is sent.
-function Cluster::publish_rr%(pool: Pool, key: string, ...%): bool
+function Cluster::publish_rr%(pool: Pool, key: any, ...%): bool
```
timwoj, line 192: If you pass a non-string to a string argument, the function is still called, correct? Just the type is still wrong on the object? The reason I'm asking is that leaving this as a `string` gives more indication to a user what should be passed without having to read the documentation. If it has to be `any` for everything to work, that's fine too though.

timwoj (changes requested):

src/broker/messaging.bif
awelzel: There's two things at play:
* bifcl makes this a `global Cluster::publish_rr: function(va_args: any): bool`, which is excluded from static type checking
* bifcl *does* generate the following preamble with an unconditional `->AsStringVal()`:
```
        zeek::Val* pool = (zeek::Val*) ((*BiF_ARGS)[0].get());
        zeek::StringVal* key = (zeek::StringVal*) ((*BiF_ARGS)[1].get()->AsStringVal());
```

So, today, in 5.2.1, you can trigger a crash as follows, which this patch makes a runtime error.
```
$ zeek -e 'Cluster::publish_rr(Cluster::Pool(), 1/tcp)'
fatal error in <no location>: Val::CONVERTER (port/string) (1/tcp)
Aborted (core dumped)
```

I agree that dropping the `: string` part is subpar. An alternative that I hadn't thought of before would be to teach `bifcl` not to do/assume the type conversions when dealing with those vargs bifs. Thoughts?

This all is treading around the bigger Zeek script vaargs issue though.

timwoj: > I agree that dropping the : string part is subpar. An alternative that I hadn't thought of before would be to teach bifcl not to do/assume the type conversions when dealing with those vargs bifs. Thoughts?

I wonder how much code would break if you did this though. Ugh.

I guess this is a good enough fix for now, but we should definitely flag it in the other issue so that it can be properly fixed if/when someone gets to that.

timwoj (approved):
Leaving a good-enough-for-now approval.

awelzel:
> Leaving a good-enough-for-now approval.

Thanks. I've started at bifcl and proposed to include the type checks there. This also allows to revert the previous `cat_sep()` changes.

Would you mind looking this over again (and the bifcl PR) as well?

timwoj (approved):
I like where this ended up a lot more with the bifcl change. Having that argument be defined as a string makes a lot more sense, and the bifcl change should hopefully catch some of these issues before they turn into crash reports.

PR #2955, opened 2023-04-15
Add irc_dcc_send_ack event and fix missing fields

AmazingPP:
1. Fix `dcc_mime_type` and `fuid` were never assigned a value in `irc.log`.
2. Fix `IRC_DATA` wrongly extracting the receiver's acknowledgment message (received length) in IRC DCC transfers as a file.


src/analyzer/protocol/file/File.h
```diff
@@ -31,12 +38,13 @@ class File_Analyzer : public analyzer::tcp::TCP_ApplicationAnalyzer
...
 	std::string file_id_orig;
 	std::string file_id_resp;
+	Direction direction;
 	};
 
 class IRC_Data : public File_Analyzer
 	{
 public:
-	explicit IRC_Data(Connection* conn) : File_Analyzer("IRC_Data", conn) { }
+	explicit IRC_Data(Connection* conn) : File_Analyzer("IRC_Data", conn, RESP) { }
```
awelzel, line 47: @AmazingPP - as always thanks for finding this and proposing a fix. Do you monitor a lot of IRC? Extracting acknowledgements to files isn't great :-)

Do you think there's more use-cases for the generalized RESP / ORIG filtering? Do you have something external in mind? Given we only have two File_Analyzer implementations and this scenario is so specific to IRC, seems okay to put it into the analyzer directly.

In fact, it's so specific/curious, I could even see a `irc_dcc_ack(c: connection, bytes_received: count)` event raised from it if someone was really interested instead of discarding it outright.

What would you think about this sketch (untested and of course feel free to skip the `irc_dcc_ack` part, but it would be nice to have a comment somewhere in the code why we're explicitly ignoring resp for IRC.
```
 
+       void DeliverStream(int len, const u_char* data, bool orig) override
+               {
+               // Bytes from originator are acknowledgements
+               if ( ! orig )
+                       File_Analyzer::DeliverStream(len, data, orig);
+               else
+                       {
+                       auto offset = ntohl(data);  /// ??
+                       if ( irc_dcc_ack )
+                               EnqueueConnEvent(irc_dcc_ack, offset); /// ??
+                       }
+               }
+
+       void Undelivered(uint64_t seq, int len, bool orig) override
+               {
+               if ( ! orig )
+                       File_Analyzer::Undelivered(seq, len, orig);
+               }
+

```

AmazingPP: > @AmazingPP - as always thanks for finding this and proposing a fix. Do you monitor a lot of IRC? Extracting acknowledgements to files clearly isn't great :-)

Yes. I confirm that the IRC DCC client will send with the total number of bytes received.
- [An IRC client implementation](https://github.com/gawel/irc3/blob/76d6849d5e7a531d649aca766f623f9f30a55545/irc3/dcc/client.py#L133-L134)
- [DCC SEND](https://en.wikipedia.org/wiki/Direct_Client-to-Client#DCC_SEND)

> Do you think there's more use-cases for the generalized RESP / ORIG filtering? Do you have something external in mind? Given we only have two File_Analyzer implementations and this scenario is so specific to IRC, seems okay to put it into the analyzer directly.

It seems that there aren't many more use-cases for the generalized RESP/ORIG filtering, as the separation of control and data streams in file transfer protocols is already relatively uncommon. Therefore, it should be okay to include this functionality directly into the analyzer, especially since there are only two File_Analyzer implementations and this scenario is specific to IRC.

> What would you think about this sketch ......

Very good guidance, I will try to implement it.

awelzel (approved):
I'm going to merge. I'll revert the C++ namespace change of the IRC_Data analyzer.

We do move the declaration of IRC_Data from protocol/file/File.h to protocol/irc/IRC.h. But, if someone actually customized IRC_Data and didn't include protocol/irc/IRC.h for other reasons, I'll be surprised (and/or suggest they change the header).

PR #2957, opened 2023-04-17
Add better Aruba support, expand IEEE 802.11 QoS support

### LINKED ISSUE #2167: Aruba and IEEE 802.11 QoS CCMP + Aggregates
We recently ran into an issue where we found some different data inside of an Aruba tunnel than expected. This manifested itself on the system as a ton of reporter warnings saying "Bad IP protocol version in IP::ParsePacket". After some investigation of the data we determined that our assumptions about Aruba data were a bit incorrect:

> // ARUBA. Following headers seem like they're always a 26-byte 802.11 QoS header, then
> // an 8-byte LLC header, then IPv4. There's very little in the way of documentation
> // for ARUBA's header format. This is all based on the one sample file we have that
> // contains it.

The header following the QoS header isn't necessarily always an LLC header. It depends a bit on what flags are set in the QoS header. There were two deviations in the data we were seeing:

1) The Protected bit in the Frame Control Field Flags byte was set. This indicates that the data contained after the QoS header is encrypted in some way (usually however the 802.11 network is encrypted?), and so will look like random data and not an LLC header.
2) The Payload Type bit in the QoS Control field was set. This indicates that the payload is an aggregate MDSU, containing multiple packets inside of a single QoS packet. Aggregation is an optimization in the 802.11 format wherein it will batch smaller packets together and send them as one unit instead of as separate packets.

We don't handle either of these cases well at all. With the assumption above, we just skip over what we think is an LLC header and send the rest to be parsed as an IP packet. That leads to the repeated warning message for all of these packets. Unfortunately we don't have a pcap from the system that was having this problem. Cloudshark's public repository has one containing the offending QoS headers, but they're not wrapped in Aruba like we're currently having trouble with. That pcap is at https://www.cloudshark.org/captures/ca7828d13464.

There's a few fixes here that could be implemented:
1) Remove the warning from `ParsePacket` and return a different value that would lead to the caller reporting a weird instead. At the least, this would get rate-limited since the weirds are based on the source and destination IP from the outer packet.
2) Fix the GRE analyzer to handle the problem bits in the IEEE header and do different things with them. I could see reporting a weird for CCMP and just outright skipping any of those packets.

(2) leads into some other tasks though:
2a) Should we just implement an IEEE 802.11 QoS packet analyzer that handles the header correctly and let Aruba forward into that analyzer? How does it get back to GRE to be processed as encapsulation in that case?
2b) How should aggregates be handled in the encapsulation case? We currently don't have any sort of re-entrance in the tunnel code where you can process down an encapsulation stack, and then potentially back up a few steps and go back down another stack.

timwoj:
This PR is two parts:

1. Expand the protocol types that are used to detect Aruba in the GRE analyzer, and fix the handling of those tunnels a bit to allow passing them down to the IEEE 802.11 analyzer.
2. Expand support in the IEEE 802.11 analyzer to properly handle CCMP-encrypted data and A-MSDU aggregate subframes

Each commit is fairly self-contained, so it's probably easiest to review commit-by-commit instead of as a whole.

Fixes #2167


src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
```diff
@@ -106,5 +106,17 @@ bool IEEE802_11Analyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet*
 	uint32_t protocol = (data[0] << 8) + data[1];
 	data += 2;
 
-	return ForwardPacket(len - len_80211, data, packet, protocol);
+	if ( packet->tunnel_type == BifEnum::Tunnel::NONE )
```
awelzel, line 178: I can't say I've grasped this. It's somehow relying on outer analyzer setting tunneling so we trigger some special logic below? That probably makes sense for the GRE/ARUBA setup?

awelzel (commented):
@timwoj - not sure the comments are valuable. I'm not enough into this to actually properly grasp the changes :-) better.

src/iosource/Packet.cc
```diff
@@ -67,7 +67,7 @@ void Packet::Init(int arg_link_type, pkt_timeval* arg_ts, uint32_t arg_caplen, u
 	ip_hdr.reset();
 
 	proto = -1;
-	tunnel_type = BifEnum::Tunnel::IP;
+	tunnel_type = BifEnum::Tunnel::NONE;
```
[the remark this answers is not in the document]
timwoj: Good point. It definitely feels bad that it defaults to `IP`, but I can't say that someone out there isn't expecting it to be like that. Let me see if I can work around it.

src/TunnelEncapsulation.h
```diff
@@ -249,6 +249,11 @@ class EncapsulationStack
 		return nullptr;
 		}
 
+	/**
+	 * Pops the last element off the encapsulation stack.
+	 */
+	void Pop() { conns->pop_back(); }
```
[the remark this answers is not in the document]
timwoj: Yeah, probably. I thought that std::vector::pop_back was a no-op, but according to the docs it can be UB.

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
```diff
@@ -26,19 +26,25 @@ bool IEEE802_11Analyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet*
...
-		// control field.
+		// Skip in case of A-MSDU subframes indicated by QoS control field.
 		if ( data[len_80211] & 0x80 )
 			return false;
 
+		// Check for the protected bit. This means the data is encrypted and we can't
+		// do anything with it.
+		if ( data[1] & 0x40 )
+			{
+			Weird("encrypted_802_11_qos_data", packet);
```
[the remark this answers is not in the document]
timwoj: That said, we were previously just dropping all A-MSDU packets without any warnings so maybe that's fine.

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
timwoj: This is actually the impetus for that change in the default `tunnel_type`. We check the tunnel type here because we need to reset the protocol being passed down to `packet_analysis::IP::ParsePacket` (called by the `IPTunnel` analyzer). Otherwise it reports a bad protocol identifier.

In reverse, if the default `tunnel_type` is set to `IP` and not `NONE`, it will erroneously fall into the second if block here and try to call the tunnel analyzer for something that isn't a tunnel.

src/iosource/Packet.cc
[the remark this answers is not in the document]
awelzel: Resolving - think NONE is more reasonable.

src/packet_analysis/protocol/gre/GRE.cc
```diff
@@ -150,29 +152,39 @@ bool GREAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
...
+				Weird("truncated_GRE", packet);
+				return false;
+				}
+
+			// TODO: it might make more sense to add an aruba analyzer here. it would duplicate a
+			// bit of the code later in this analyzer but it could handle everything more cleanly
+			// and not need a bunch of special cases in this code.
+			is_aruba = true;
+			gre_link_type = DLT_EN10MB;
+			gre_version = 1;
```
awelzel, line 177: Is this actually reasonable? The gre_version is parsed from the header, so we should not override it. For the added pcap, wireshark says gre_version=0. We have other places in the IPTunnel analyzer where testing for (gre_version != 0) means there's IP fields (that's probably also buggy because gre_version is initialized with -1).

I realize this construct might work out currently, but it's very surprising to change a parsed field to something else that's not actually in the packet.

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
awelzel: TLDR: This special case here didn't click at all. It does a bit now, but I think it shouldn't be there, but that probably requires surgery.

This is sending the IP packet within 802.11 through the IPTunnel analyzer because there's an outer tunnel. That seems fairly surprising: Even if there's an outer tunnel, an IP packet within 802.11 should be sent to just the IP analyzer. The outer tunnel should not be relevant (and forwarding based on just the LLC protocol be "the right thing" outside of crazy things (and this doesn't seem too crazy).

So.. staring for a long time: I'm seeing that with this change, a tunnel.log entry for the outer GRE connection is created and that seems nice. If I patch to just dispatch on llc.protocol, that tunnel.log entry is gone and that's sad, I agree. Though that's pretty confusing and actually seems an issue on the level of the GRE analyzer implemenation: Conceptually, any IP/GRE/something tunnel should likely produce a tunnel.log entry as long as the IP and GRE parts are valid, even if the "something part" isn't something that Zeek can deal with.

I realize the current code is currently structured much differently and by default any inner protocols within GRE are sent through IPTUNNEL by default and that then happens to create the tunnel.log. This also made me understand where the special case here really originates from, but at the same time appears pretty rough.

I think the cleaner solution would be to push tunnel instantiation and resulting tunnel.log generation into GRE itself rather than doing it indirectly through the IPTUNNEL analyzer.

awelzel (commented):
@timwoj - I've spent time digging and found things that I'm not finding pretty. See the comments.

I'm okay approving as you might want to move it forward. Just seems the GRE/IP tunnel analyzer has some sharp and rough edges and the interplay is little obvious.

src/packet_analysis/protocol/gre/GRE.cc
```diff
@@ -150,29 +152,39 @@ bool GREAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
 				}
 			}
 
-		else if ( proto_typ == 0x8200 )
+		else if ( ((proto_typ & 0x8200) == 0x8200 && (proto_typ & 0x0F) == 0) ||
```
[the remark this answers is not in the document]
timwoj: > This is essentially what wireshark has this big list for? They call it "ARUBA WLAN"

Yep.

> Other approaches here would be to register 802.11 directly with GRE for all these ids in script land?

The problem with registering it with IEEE 802.11 is that you then have to duplicate some of the tunnel/header handling in that analyzer, when it really should be in the GRE analyzer. We're already handling other protocol types by number here as well.

src/packet_analysis/protocol/gre/GRE.cc
```diff
@@ -225,15 +237,18 @@ bool GREAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
...
+	// and only payload packet remains.  The only thing different is the tunnel type enum value to
+	// use.
 	packet->tunnel_type = BifEnum::Tunnel::GRE;
 	packet->gre_version = gre_version;
 	packet->gre_link_type = gre_link_type;
 	packet->proto = proto;
 
-	ForwardPacket(len, data, packet);
+	if ( forward_by_proto )
+		ForwardPacket(len, data, packet, proto);
```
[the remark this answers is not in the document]
timwoj: Ah, good point. The `forward_by_proto` bit can be removed since the default is via IP tunnel.

src/packet_analysis/protocol/gre/GRE.cc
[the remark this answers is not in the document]
timwoj: Never mind, I see what you meant. Yeah, it would be more correct to add the forwarding rules based on the types there. I'll have to see what setting `packet->proto` to one of the aruba types does downstream though.

src/packet_analysis/protocol/gre/GRE.cc
timwoj: Removed.

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
timwoj: The question is whether that should happen as part of this PR.

The current IPTunnelAnalyzer was born out of a desire to not repeat a bunch of code around handling the encapsulation and whatnot in each of the tunnel analyzers. Maybe it could be reworked to become a base class for the other tunnel analyzers, but it also needs to handle regular IP-in-IP tunnels as well.

J-Gras: FWIW, I agree that the current approach is hard to follow and I like the suggested solutions. Moving this into another PR makes sense to me.

scripts/base/packet-protocols/gre/main.zeek
```diff
@@ -6,5 +6,30 @@ export {
 
 event zeek_init() &priority=20
 	{
-	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_GRE, 0x0069, PacketAnalyzer::ANALYZER_IEEE802_11);
+	# ARUBA
```
J-Gras, line 9: Conceptually, I would have expected the other GRE 0 types (ERSPAN etc.) to appear here as well.

J-Gras (commented):
@awelzel did a thorough job. I don't really have anything to add.

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
awelzel: > The question is whether that should happen as part of this PR.

No. I understand this was meant as a quick fix, but you asked for review :-)

I have one more idea/patch below that removes the IEEE802.11 special case. That makes it a bit more graspable for me and possibly clearer how to re-structure in the future. If you don't like it, I'll sign-off on whatever otherwise, you know this stuff much better :-)

> The current IPTunnelAnalyzer was born out of a desire to not repeat a bunch of code around handling the encapsulation and whatnot in each of the tunnel analyzers. Maybe it could be reworked to become a base class for the other tunnel analyzers, but it also needs to handle regular IP-in-IP tunnels as well.

Yep, after staring even longer a few things came together. The IPTUNNEL analyzer seems rather GRE specific, so that was initially very confusing (and it still is).

I've tried one more thing: Doing the register_packet_analyzer() of the ARUBA types for IPTUNNEL instead of GRE. That works out well and allows to remove the IEEE802.11 tunnel special case there (which IMO really stood out as bad). Now the setup is GRE->IPTUNNEL->IEEE802.11 and it seems to work for the new test cases added :crossed_fingers: 

This is the change.
https://github.com/zeek/zeek/commit/78bc32c97c0e9525b63ec460dceecad3d3c8d322

The one quirky part that is left is that we need to keep setting gre_link_type properly in GRE, otherwise IPTunnel uses DLT_RAW for the 802.11 packet within and that's not valid.

Maybe there's actually room for an extension to `register_packet_analyzer()` to pass the link_type during registration to enable that use-case and doesn't rely on whatever the analyzer before ended up setting in gre_link_type? Definitely future though.

```
PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_IPTUNNEL, 0x8200, PacketAnalyzer::ANALYZER_IEEE802_11, PacketAnalyzer::LINK_TYPE_IEEE802_11);
```

timwoj: > What confuses me here is that the GRE types for Aruba will still end up in the configuration of the IPTunnel analyzer, right?

Yeah, this is the part that I'm not really happy with either. That said, not having the weird work around in the 802.11 analyzer is nicer. I committed Arne's version since it removes the work around, but this should probably all be reinvestigated.

testing/btest/Baseline/plugins.hooks/output
```diff
@@ -656,10 +656,13 @@
 0.000000   MetaHookPost  CallFunction(PacketAnalyzer::register_packet_analyzer, <frame>, (PacketAnalyzer::ANALYZER_GENEVE, 2054, PacketAnalyzer::ANALYZER_ARP)) -> <no result>
 0.000000   MetaHookPost  CallFunction(PacketAnalyzer::register_packet_analyzer, <frame>, (PacketAnalyzer::ANALYZER_GENEVE, 2269, PacketAnalyzer::ANALYZER_IP)) -> <no result>
 0.000000   MetaHookPost  CallFunction(PacketAnalyzer::register_packet_analyzer, <frame>, (PacketAnalyzer::ANALYZER_GENEVE, 25944, PacketAnalyzer::ANALYZER_ETHERNET)) -> <no result>
+0.000000   MetaHookPost  CallFunction(PacketAnalyzer::register_packet_analyzer, <frame>, (PacketAnalyzer::ANALYZER_GRE, 105, PacketAnalyzer::ANALYZER_IEEE802_11)) -> <no result>
```
awelzel, line 659: Suspect this and the below are out-dated.

testing/btest/Baseline/core.tunnels.gre-aruba-ccmp/weird.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	weird
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	name	addl	notice	peer	source
+#types	time	string	addr	port	addr	port	string	string	bool	string	string
+XXXXXXXXXX.XXXXXX	-	10.226.22.9	0	10.102.16.111	0	encrypted_802_11_qos_data	-	F	zeek	IEEE802_11
```
awelzel, line 10: That weird.log baseline isn't needed anymore in the latest version.

awelzel (approved):

src/packet_analysis/protocol/ieee802_11/IEEE802_11.cc
awelzel: > > What confuses me here is that the GRE types for Aruba will still end up in the configuration of the IPTunnel analyzer, right?
>
> Yeah, this is the part that I'm not really happy with either. [...] but this should probably all be reinvestigated.

(Re-)moving the IP analyzer pieces into GRE directly so it becomes closer to VXLAN/Geneve (and closer to what Jan posted) would be reasonable. I'm not sure Aruba or e.g., Tranparent Ethernet Bridging (0x6558) need to be analyzer: there's nothing to be parsed - they'd be coded mappers from GRE to IEEE802.11 or Ethernet analyzer. Didn't seem there's anything fancy about them other than not fitting the GRE / iptunnel analyzer setup and that makes it all complicated.

I've approved, @timwoj , assuming you can figure out test failures etc :-)

Issue #2107, opened 2022-05-13
Perhaps adding hooks for intercepting script execution


vpax:
Per the discussion in https://github.com/zeek/zeek/pull/2087, it might make sense to generalize the `StartInvocation()`/`EndInvocation()` approach used for script profiling to allow other hooking into script execution.  For example, you could imagine enabling monkey-patching whereby the hook could supersede execution of a given function in a given context.

awelzel:
Stumbling over the `spm` "ScriptProfileMgr" and `etm` "EventTraceMgr" "hooks" within `Func.cc` while doing #3485 and remembering this ticket.

It seems `etm` could be removed from `Func.cc` if we had a `HookPostCallFunction` which is invoked after executing all bodies of a function (essentially the counterpart to `HookCallFunction` without changing execution).

The `spm` calls could replaced with two `HookExecBody()` and `HookPostExecBody()` hooks. The first hook would be similar to `HookCallFunction`, but on the level of bodies.

I think this would overall be worth it, as it would open up implementing powerful profiling tools outside of core Zeek as plugins.

`ScriptProfile.cc` and `EventTrace.cc` would need to be switched to be more plugin'ish and leverage the new hooks instead of being globals.

---

```
$ git grep -n -C2  -e 'etm->' -e 'spm->' src/Func.cc
src/Func.cc-346-
src/Func.cc-347-    if ( etm && Flavor() == FUNC_FLAVOR_EVENT )
src/Func.cc:348:        etm->StartEvent(this, args);
src/Func.cc-349-
src/Func.cc-350-    if ( g_trace_state.DoTrace() ) {
--
src/Func.cc-372-
src/Func.cc-373-        if ( spm )
src/Func.cc:374:            spm->StartInvocation(this, body.stmts);
src/Func.cc-375-
src/Func.cc-376-        f->Reset(args->size());
--
src/Func.cc-394-
src/Func.cc-395-        if ( spm )
src/Func.cc:396:            spm->EndInvocation();
src/Func.cc-397-
src/Func.cc-398-        if ( f->HasDelayed() ) {
--
src/Func.cc-424-
src/Func.cc-425-    else if ( etm && Flavor() == FUNC_FLAVOR_EVENT )
src/Func.cc:426:        etm->EndEvent(this, args);
src/Func.cc-427-
src/Func.cc-428-    // Warn if the function returns something, but we returned from
--
[...]

PR #2958, opened 2023-04-17
Add call stacks to script profiler output

### LINKED ISSUE #2485: Enhance output `--profile-scripts` to include trace of function calls on the stack
Currently, `--profile-scripts` shows time and memory usage for individual events and function calls, however, it does not include the stack trace to know _where_ the function was called from. By altering the output to include the stack traces, it would be easy to generate [flamegraphs](https://github.com/brendangregg/FlameGraph#flame-graphs-visualize-profiled-code) of the CPU execution time to better understand the performance profile of large Zeek packages on large PCAPs.

As an example, these [comments in the flamegraph repo](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-perf.pl#L14-L26) show the standard output from the Linux `perf` tool, as well as the output of the linked Perl script that is used as input to generate the flamegraph. I think the list of different profiler programs in that repo helps justify altering the Zeek script profiling output to either (1) contain the stack traces to allow generating flamegraphs or (2) output directly in a format that can be used to generate flamegraphs.

timwoj:
This adds callstacks to the output from --profile-scripts so they can be used by tools like flamegraph. See the linked issue for sample output. This also adds a basic btest for that argument since we weren't testing this code at all.

Fixes #2485

### DIFF
--- a/src/ScriptProfile.cc
+++ b/src/ScriptProfile.cc
@@ -17,7 +17,7 @@ void ScriptProfile::StartActivation()
 	start_stats.SetStats(util::curr_CPU_time(), start_memory);
 	}
 
-void ScriptProfile::EndActivation()
+void ScriptProfile::EndActivation(const std::string& stack)
 	{
 	uint64_t end_memory;
 	util::get_memory_usage(&end_memory, nullptr);
@@ -25,7 +25,7 @@ void ScriptProfile::EndActivation()
 	delta_stats.SetStats(util::curr_CPU_time() - start_stats.CPUTime(),
 	                     end_memory - start_stats.Memory());
 
-	AddIn(&delta_stats, false);
+	AddIn(&delta_stats, false, stack);
 	}
 
 void ScriptProfile::ChildFinished(const ScriptProfile* child)
@@ -33,7 +33,7 @@ void ScriptProfile::ChildFinished(const ScriptProfile* child)
 	child_stats.AddIn(child->DeltaCPUTime(), child->DeltaMemory());
 	}
 
-void ScriptProfile::Report(FILE* f) const
+void ScriptProfile::Report(FILE* f, bool with_traces) const
 	{
 	std::string l;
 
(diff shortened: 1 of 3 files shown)


src/ScriptProfile.cc
```diff
@@ -47,9 +47,14 @@ void ScriptProfile::Report(FILE* f) const
 
 	std::string ftype = is_BiF ? "BiF" : func->GetType()->FlavorString();
 
-	fprintf(f, "%s\t%s\t%s\t%d\t%.06f\t%.06f\t%" PRIu64 "\t%" PRIu64 "\n", Name().c_str(),
+	std::string call_stacks;
+	for ( const auto& [s, count] : Stacks() )
```
awelzel, line 51: Only thought is whether `Stacks()` should be `RenderedCallStack()` and the string formatting a bit hidden.

The other one: This is producing pretty long lines when testing real workload? Does it make the existing output which might have been meant to be consumable as a table on the terminal difficult to use?

awelzel (approved):
Approving with the caveat of possibly making it difficult to use the output without post-processing due to overly long lines, but that might be a call for @ynadji or @vpax to weigh in on.

src/ScriptProfile.cc
vpax: Yeah, I wonder about that too.  I also chatted with Tim about how conceptually it seems to me there should be three additional columns, one for number-of-calls-per-tree (which is what this branch currently implements, and what Flame Graphs use), one for CPU-per-tree, and one for memory-per-tree.  So in my ideal world, we'd (1) have all three of those, but also (2) have a different/additional flag for turning them on, so it's still easy to get the simpler works-well-in-a-terminal-window version that's the current behavior.

timwoj: If that's the desire, then I'd propose 4 columns. One with the call stacks, and then one for each of the calls, cpu time, and memory. It'd add a little bit of time to post-processing, but it would reduce the amount of duplicate data in the output.

timwoj: The latest commit added a `--profile-scripts-traces` argument to turn the trace output on and off. It also splits out the call counts and adds CPU/memory data. The new output looks like this from `$TRACES/wikipedia.trace`:

```
#fields function        location        type    ncall   tot_CPU child_CPU       tot_Mem child_Mem       stacks        stack_calls     stack_CPU       stack_memory
#types  string  string  string  count   interval        interval        count   count   string  string  string        string
Log::write      /Users/tim/Desktop/projects/zeek/scripts/base/frameworks/logging/./main.zeek:839        function      83      0.001765        0.001309        163840  163840          Log::write;zeek_init;PacketFilter::install|Log::write;http_message_done|Log::write;dns_end|Log::write;connection_state_remove;DNS::finalize_dns;DNS::log_unmatched_msgs;DNS::log_unmatched_msgs_queue|Log::write;connection_state_remove;DNS::finalize_dns|Log::write;connection_state_remove     1|14|14|12|8|34 0.000152|0.000257|0.000260|0.000293|0.000128|0.000675   65536|16384|16384|0|0|65536
```

src/ScriptProfile.h
```diff
@@ -46,16 +46,28 @@ class ScriptProfileStats
...
 	// whether the folding should include increasing the number of calls.
-	void AddIn(const ScriptProfileStats* eps, bool bump_num_calls = true)
+	void AddIn(const ScriptProfileStats* eps, bool bump_num_calls = true,
+	           const std::string& stack = "")
 		{
 		if ( bump_num_calls )
 			ncalls += eps->NumCalls();
 
+		if ( ! stack.empty() )
+			{
```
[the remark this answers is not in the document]
timwoj: Cleaned this up a bit in the latest commit.

awelzel (commented):
Mostly naming / nit, my previous approval still stands.

Generally, I think the profiling functionality would've been a good candidate to explore/iterate in an external plugin first. The current output isn't something that's immediately pipe-able into Flamegraph.pl IIUC and that would be cool to do. I'm also unsure what guarantees about format stableness are supposed to give for this.

Having this in-tree also means we're not providing hooks for external plugins to do this or explore other approaches to script profiling. The esnet prometheus exporter bends-over-backwards to do the same for direct Prometheus exporting. We do have that part tracked in #2107 though.

PR #2961, opened 2023-04-19
file_analysis/Manager: Remove RunState.h include

awelzel:
The file_analysis/Manager.h header only needs run_state::terminating, so declare just that similar to what Val.h does. This breaks ZBody compilation due to OP_READING_LIVE_TRAFFIC_V and OP_READING_TRACES_V accessing run-state::reading_live and run_state::reading_traces. Add the include to ZBody.cc as a fix.

This may break external plugins, too, but then they didn't have the right includes setup to begin with.

### DIFF
--- a/src/file_analysis/Manager.h
+++ b/src/file_analysis/Manager.h
@@ -7,7 +7,6 @@
 #include <string>
 
 #include "zeek/RuleMatcher.h"
-#include "zeek/RunState.h"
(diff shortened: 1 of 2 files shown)


awelzel:
This came up during discussion of #2929 and seemed reasonable to propose.

ckreibich (approved):
Seems fine to me, though I'd tread lightly with these.

Issue #2324, opened 2022-08-08
Recognize LLC/SNAP after VLAN


J-Gras:
So far, Zeek assumes to deal with Ethernet II frames when handling VLAN tags. Actually, VLAN tags can be inserted in all Ethernet frame types including LLC and SNAP. While the Ethernet analyzer [considers other frame types](https://github.com/zeek/zeek/blob/857b21ae73b967485e499f84f5cb474ab22a3dfe/src/packet_analysis/protocol/ethernet/Ethernet.cc#L56), VLAN tags followed by LLC/SNAP will cause entries in the unknown protocols log.

This is triggered by some STP variant. See [packetlife.net](https://packetlife.net/captures/protocol/stp/) for an example ([MSTP_Intra-Region_BPDUs.cap](https://www.cloudshark.org/captures/9b9797bd79f6)):
```
zeek -e '@load policy/misc/unknown-protocols redef UnknownProtocol::sampling_rate=1;' -r MSTP_Intra-Region_BPDUs.cap
```

From the top of my head I can imagine three solutions:

1. Replicate the Ethernet analyzer's distinction code in the VLAN analyzer.
2. Remove the VLAN analyzer and integrate it into the Ethernet analyzer.
3. Come up with dedicated Sub-Analyzers that consume some data and then "jump back" into the parent packet analyzer.

The first seems to be the easiest but would replicate code and (maybe) configuration. The second doesn't seem too complicated. The drawback would be that we circumvent the packet analysis composition mechanism. The third approach would need some thinking and most likely change the existing interfaces. If there are other applications (e.g. tunneling) this might still be worth to consider.

J-Gras:
Forgot to include mention a [discussion](https://forums.cabling-design.com/ethernet/ethernet-vlan-using-llc-snap-encapsulation-6862-.htm) that provides some pointers to the potentially relevant standards.

timwoj:
> Replicate the Ethernet analyzer's distinction code in the VLAN analyzer.

I'll be honest, I'm not entirely sure why were did it this way in the first place. Why didn't we just implement SNAP/Novell/LLC analyzers (or leave them unimplemented) and forward into them like any other packet analyzer? Those analyzers wouldn't need to be too complicated, and then the Ethernet and VLAN analyzers could just forward into them. It would break the one packet-protocol-plugin test, since it relies on that mechanism, but it seems more correct to me.

J-Gras:
I am not sure I can follow.
[...]

PR #2962, opened 2023-04-19
Add LLC, SNAP, and Novell-style 802.3 (IPX) packet analyzers

timwoj:
This PR adds three new packet analyzers: LLC, SNAP, and Novell-style 802.3 (IPX). Previously we had a janky setup for custom definitions for these as a way to call them from the Ethernet analyzer, but #2324 brought up that we'd have to duplicate a bunch of that setup to support calling them from other analyzers.

Instead, this PR removes that setup and adds packet analyzers for those protocols, including some basic parsing for LLC and SNAP. It also adds a test for ARP contained in a SNAP header. Tests for the other two are missing due to the lack of good packet captures that go from one of those headers to a header we already process. For example I have packet captures containing regular LLC headers, but none that go from LLC to something we’re already parsing to test events.

Lastly, it adds in support for going from the VLAN analyzer into these new analyzers, since that was the original reason why I started looking at this.


timwoj:
@J-Gras I requested a review from you on this since we talked about it in a call the other day, so you have most of the context of what is happening here.

src/packet_analysis/protocol/llc/LLC.cc
```diff
@@ -0,0 +1,32 @@
...
+
+#include "zeek/packet_analysis/protocol/llc/LLC.h"
+
+using namespace zeek::packet_analysis::LLC;
+
+LLCAnalyzer::LLCAnalyzer() : zeek::packet_analysis::Analyzer("LLC") { }
+
+bool LLCAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
+	{
+	// An LLC header is at least 3 bytes, check for that first.
```
J-Gras, line 11: With the suggested changes to the Ethernet analyzer, I think the overall packet length could be validated as well.

src/packet_analysis/protocol/llc/LLC.cc
```diff
@@ -0,0 +1,32 @@
...
+	if ( (data[2] & 0x03) != 0x03 )
+		llc_header_len++;
+
+	if ( len < llc_header_len )
+		{
+		Weird("truncated_llc_header", packet);
+		return false;
+		}
+
+	return true;
```
J-Gras, line 31: I am not sure how we would continue here to add further protocols like the Spanning Tree Protocol. Any ideas?

src/packet_analysis/protocol/snap/SNAP.cc
```diff
@@ -0,0 +1,50 @@
...
+	int oui = (data[0] << 16) | (data[1] << 8) | data[2];
+	int protocol = (data[3] << 8) | data[4];
+
+	data += 5;
+	len -= 5;
+
+	if ( oui == 0 )
+		{
+		// If the OUI is zero, the protocol is a standard ethertype and can be
+		// forwarded as such.
```
J-Gras, line 45: My first thought was to multiplex based on OUI to a "SNAP-OUI-Analyzer", which could then forward based on the protocol. Probably overkill...

J-Gras (changes requested):
Looks good to me! I think the VLAN analyzer needs a second look. Everything else shouldn't be a dealbreaker. One thing I came across is that the [documentation](https://docs.zeek.org/en/master/frameworks/packet-analysis.html) needs an update as well. Maybe also a good spot to document the magic values outside of the code itself.

scripts/base/packet-protocols/ethernet/main.zeek
```diff
@@ -22,4 +13,8 @@ event zeek_init() &priority=20
 	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x9100, PacketAnalyzer::ANALYZER_VLAN);
 	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x8864, PacketAnalyzer::ANALYZER_PPPOE);
 	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x8926, PacketAnalyzer::ANALYZER_VNTAG);
+
+	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x0001, PacketAnalyzer::ANALYZER_SNAP);
+	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x0002, PacketAnalyzer::ANALYZER_NOVELL_802_3);
+	PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_ETHERNET, 0x0003, PacketAnalyzer::ANALYZER_LLC);
```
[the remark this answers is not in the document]
timwoj: > Maybe it makes sense to introduce consts here and reuse them in the VLAN analyzer.

That sounds good.

> Also, how about using 1501-1503

I was going as low as I could just in case we have to add more of these later.

src/packet_analysis/protocol/snap/SNAP.cc
timwoj: I think that's overkill as well. The SNAP analyzer isn't so complicated that we would need to break it up to handle the different OUI modes.

scripts/base/packet-protocols/ethernet/main.zeek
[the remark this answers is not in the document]
timwoj: > That sounds good.

Actually now that I think about this one more, it doesn't really make sense to re-use consts from the ethernet analyzer into the vlan analyzer. The two analyzers are distinct, and the values they use for these don't really mean anything between them. I could see defining a set of consts in both though.

src/packet_analysis/protocol/ethernet/Ethernet.cc
```diff
@@ -62,25 +53,21 @@ bool EthernetAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pa
 			return false;
 			}
 
-		// Let specialized analyzers take over for non Ethernet II frames.
-		// Note that pdata remains at the start of the ethernet frame.
+		len -= 14;
+		data += 14;
```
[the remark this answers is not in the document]
timwoj: This is skipping the source address, destination address, and the length field to get to the start of the LLC header. That's 14 bytes, unless I'm missing something?

src/packet_analysis/protocol/llc/LLC.cc
timwoj: I think the length check in the ethernet analyzer is wrong as it is, to be honest:

```
		if ( 16 >= len )
			{
			Weird("truncated_ethernet_frame", packet);
			return false;
			}
```

We already made that comparison earlier. This should be checking whether the remaining data fits, which is the value of `protocol` that was pulled out earlier.

src/packet_analysis/protocol/ethernet/Ethernet.cc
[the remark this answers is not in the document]
J-Gras: Nevermind, I got confused by an [overview](https://www.ionos.de/digitalguide/server/knowhow/ethernet-frame/) that made it look like in case of SNAP the last two bytes describe a type instead of the length.

scripts/base/packet-protocols/ethernet/main.zeek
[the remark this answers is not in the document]
J-Gras: As 802.1Q uses the same values, i.e. Ethertypes, I think it would be okay to share. However, defining them in both analyzers is fine as well.

src/packet_analysis/protocol/llc/LLC.cc
timwoj: I went ahead and added forwarding based on the DSAP and remove the default rule for LLC.

J-Gras: > I assume it should just forward based on the DSAP?

I have no idea but that sounds reasonable :)

J-Gras (approved):
Looks good to me. At some point we might be able to make the VLAN analyzer become a "child-analyzer" and get rid of the duplicated code.

src/packet_analysis/protocol/ethernet/Ethernet.cc
```diff
@@ -56,31 +53,25 @@ bool EthernetAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pa
...
 			return false;
 			}
 
 		// Let specialized analyzers take over for non Ethernet II frames.
-		// Note that pdata remains at the start of the ethernet frame.
-
-		AnalyzerPtr eth_analyzer = nullptr;
-
-		if ( data[14] == 0xAA && data[15] == 0xAA )
+		if ( data[0] == 0xAA && data[1] == 0xAA )
```
awelzel, line 66: Couldn't this have leveraged the existing `DetectProtocol()` infrastructure reasonably well without introducing artificial  protocol identifiers? That would've been a bit more concise? The code here and the repeat in VLAN.cc is `DetectProtocol()` for SNAP/Novell in disguise, no? :-)

Not sure about the LLC default analyzer part, but that maybe just works (?) or may need some massaging to default to it within the packet-analysis infrastructure.

J-Gras: 🙈 indeed `DetectProtocol()` looks like it was made for exactly that use case...

timwoj: @awelzel and I talked about this a bit. I like the idea of using `DetectProtocol()` except the function has to return true or false if it matches some heuristic. There isn't a heuristic for LLC. It could always just return true, but that means the check in `ForwardPacket` is always going to hit that case and try to call `AnalyzePacket` on the LLC analyzer.

A workaround for calling it from the ethernet analyzer would be to make `DetectProtocol` check `packet->eth_type`. Unfortunately, that value isn't set from the VLAN analyzer. In that case it'd be set to 0x8100 since that's what the ethernet analyzer set it to before the VLAN analyzer got called.

I could pull the protocol detection bits out of `ForwardPacket`, call that from the analyzers, and then call the LLC analyzer if it doesn't match, but then the Ethernet/VLAN analyzers have to have a copy of the LLC analyzer directly. That violates how we're trying to keep to the analyzers from really knowing about each other beyond the registrations (and basically adds back in one of the hacks that this PR originally removed).

J-Gras: > ... and there isn't a heuristic for LLC

Wouldn't that just be the inverse of SNAP and Novell, i.e. payload neither starts with 0xFFFF nor 0xAAAA? The problem here is that the heuristics might trigger for unknown Ethernet II types as well. One could argue that we are dealing with two protocols in one and should split the analyzer again. In that case LLC could become the default analyzer for non-Ethernert II.

timwoj: > The problem here is that the heuristics might trigger for unknown Ethernet II types as well.

Yes, exactly.

> One could argue that we are dealing with two protocols in one and should split the analyzer again. In that case LLC could become the default analyzer for non-Ethernert II.

So what, you'd have an analyzer that gets called for Ethernet II and one that gets called for everything else? Would that also split the VLAN analyzer into two?

J-Gras: > So what, you'd have an analyzer that gets called for Ethernet II and one that gets called for everything else?

Yes, that's what I had in mind. I guess we would have a generic Ethernet analyzer that just extracts src and dst MACs, one analyzer for Ethernet II and one Ethernet multiplexer.  The Ethernet II and Ethernet multiplexer analyzers would use `DetectProtocol` to attach based on the Ethertype/length value. The Ethernet multiplexer could then use the first two payload bytes for classic dispatching (0xFFFF, 0xAAAA, and default for LLC):

```
          [DetectProtocol]        [Identifier]

                 ┌► EthernetII         ┌─► Novell
GenericEthernet ─┤                     │
                 └► EthernetMultiplex ─┼─► SNAP
                                       │
                                       └─► LLC (default)
```

> Would that also split the VLAN analyzer into two?

Forwarding from the VLAN analyzer would be done the same way it's done in the generic Ethernet analyzer: Use `DetectProtocol` to distinguish between Ethernet II and the rest.

All in all, these would be a lot of indirections. In particular, this would add an additional dispatching step to the probably most common Ethernet II path. My feeling is still that introducing child packet analyzers would be a bit cleaner. When called, a child analyzer would consume a number of bytes and just return how many bytes it processed to the calling analyzer. The calling analyzer could then continue processing at this point. VLAN would be the child in this case. I think I vaguely remember a draft for packet analyzers in which `ForwardPacket` returned a pair indicating parsing success and bytes left (that might be a textbook example of a monad ;). This way the "child-analyzer-concept" would just blend in.

PR #2965, opened 2023-04-20
logging/cluster/supervisor: Multi-logger awareness

awelzel:
When multiple loggers are configured in a Supervisor controlled cluster
configuration, encode extra information into the rotated filename to
identify which logger produced the log.

This is similar to the approach taken for ZeekControl, re-using the
log_suffix terminology, but as there's only a single zeek-archiver
process and no postprocessors and no other side-channel for additional
information, we encode extra metadata into the filename. zeek-archiver
is extended to recognize the special metadata part of the filename.

This also solves the issue that multiple loggers in a supervisor setup
overwrite each others log files within a single log-queue directory.

---

zeek-archiver side PR: https://github.com/zeek/zeek-archiver/pull/17


awelzel:
I've rebased on master to leverage the new `Cluster::get_node_count()` API so this is now a bit smaller.

NEWS
```diff
@@ -116,6 +116,10 @@ New Functionality
   ``make-archive-name`` script and using the new ``ZEEK_ARG_LOG_SUFFIX``
   environment variable.
 
+- A supervisor controlled Zeek cluster is now multi-logger aware. This avoids
+  loggers overwriting each others log files within a single log-queue directory.
```
ckreibich, line 120: Total nit: 
```suggestion
  loggers overwriting each others' log files within a single log-queue directory.
```

scripts/base/frameworks/cluster/nodes/logger.zeek
```diff
@@ -22,6 +22,32 @@ redef Log::default_rotation_interval = 1 hrs;
...
+		if ( |v| == 0 )  # Assume concious decision to skip this entry.
+			next;
+
+		if ( /[,=]/ in k || /[,=]/ in v )
+			{
+			Reporter::warning(fmt("Invalid log_metadata: k='%s' v='%s'", k, v));
+			next;
+			}
+
+		metadata_vec += fmt("%s=%s", k, v);
```
ckreibich, line 45: Should we run `k` and `v` through `strip()`?

ckreibich (approved):
Thanks! When this and zeek/zeek-archiver#17 are in, we should add a multi-logger testcase to the `zeek-testing-cluster` testsuite that ties it all together. (Happy to do that, but you will undoubtedly beat me to it. :-)

NEWS
awelzel: Hmm, I added a `'`. After a bit of googling I've put it before the "s" feel free to fixup.

PR #2968, opened 2023-04-21
Add community_id_v1() based on corelight/zeek-community-id

awelzel:
"Community ID" has become an established flow hash for connection correlation
across different monitoring and storage systems. Other NSMs have had native
and built-in support for Community ID since late 2018. And even though the
roots of "Community ID" are very close to Zeek, Zeek itself has never provided
out-of-the-box support and instead required users to install an external plugin.

While we try to make that installation as easy as possible, an external plugin
always sets the bar higher for an initial setup and can be intimidating.
It also requires a rebuild operation of the plugin during upgrades. Nothing
overly complicated, but somewhat unnecessary for such popular functionality.

This isn't a 1:1 import. The options are parameters and the "verbose"
functionality  has been removed. Further, instead of a `connection`
record, the new bif works with `conn_id`, allowing computation of the
hash with little effort on the command line:

    $ zeek -e 'print community_id_v1([$orig_h=1.2.3.4, $orig_p=1024/tcp, $resp_h=5.6.7.8, $resp_p=80/tcp])'
    1:RcCrCS5fwYUeIzgDDx64EN3+okU

Reference: https://github.com/corelight/zeek-community-id/

PRs in for external test suites:

https://github.com/zeek/zeek-testing-private/pull/13
https://github.com/zeek/zeek-testing/pull/12


JustinAzoff:
> Further, instead of a connection record, the new bif works with conn_id

yay, that should make some other odd things easier.

I also see you changed it to use `new_connection` which I had been meaning to do a PR for.

ckreibich (approved):
Thanks Arne. We should cover this in the docs, and say somewhere that you cannot run both this and the package, which should be obvious. I don't see a particularly natural place for this right now — I could see a new section "Popular customizations", with Community ID as a first section.

PR #2973, opened 2023-04-23
extending script profiling to include summaries for functions/hooks/event handlers

vpax:
The current Zeek script coverage machinery produces per-`Stmt` execution counts. It's handy to also have these for per-function (including hooks and event handlers), so one can tell at a glance if certain functions are untouched by a given set of runs.  This PR adds that, along with some minor low-level tidying.  The output includes the word `TOTAL` at the end of the line to make it easy to search for just those summaries (this last tweak was done as a `fixup!` which um I've never used before, so I may have botched it :-P).


src/ScriptCoverageManager.cc
```diff
@@ -20,19 +19,17 @@ namespace zeek::detail
...
 		return;
 
-	Ref(s);
-	stmts.push_back(s);
+	stmts.push_back({NewRef{}, s});
+	}
+
+void ScriptCoverageManager::AddFunction(IDPtr func_id, StmtPtr body)
+	{
+	func_instances.push_back(pair{func_id, body});
```
timwoj, line 32: You should be able to omit the `pair` here.

src/ScriptCoverageManager.cc
```diff
@@ -127,31 +124,48 @@ bool ScriptCoverageManager::WriteStats()
...
 		}
 
-	for ( auto s : stmts )
+	for ( auto& s : stmts )
 		{
-		ODesc location_info;
-		s->GetLocationInfo()->Describe(&location_info);
 		ODesc desc_info;
 		s->Describe(&desc_info);
 		string desc(desc_info.Description());
```
timwoj, line 131: `desc` isn't used anymore.

src/ScriptCoverageManager.cc
```diff
@@ -127,31 +124,48 @@ bool ScriptCoverageManager::WriteStats()
...
-		if ( usage_map.find(location_desc) != usage_map.end() )
-			usage_map[location_desc] += s->GetAccessCount();
-		else
-			usage_map[location_desc] = s->GetAccessCount();
+		TrackUsage(s, desc_info.Description(), s->GetAccessCount());
 		}
 
-	map<pair<string, string>, uint64_t>::const_iterator it;
-	for ( auto& um : usage_map )
+	for ( auto& fi : func_instances )
```
timwoj, line 135:
```suggestion
	for ( auto& [func, body] : func_instances )
```

And remove the two variables from the body.

src/ScriptCoverageManager.cc
```diff
@@ -127,31 +124,48 @@ bool ScriptCoverageManager::WriteStats()
...
+
 	fclose(f);
 	return true;
 	}
 
+void ScriptCoverageManager::TrackUsage(const ObjPtr& obj, std::string desc, uint64_t cnt)
+	{
+	ODesc location_info;
+	obj->GetLocationInfo()->Describe(&location_info);
+	canonicalize_desc cd{delim};
```
timwoj, line 157: Does it make sense to either make this static or save it off somewhere? It looks like `delim` doesn't ever change, so it doesn't need to be remade every call.

src/ScriptCoverageManager.cc
```diff
@@ -127,31 +124,48 @@ bool ScriptCoverageManager::WriteStats()
...
+		auto& func = fi.first;
+		auto& body = fi.second;
+
+		auto ft = func->GetType<FuncType>();
+		auto desc = ft->FlavorString() + " " + func->Name() + " TOTAL";
+
+		TrackUsage(body, desc, body->GetAccessCount());
 		}
 
+	for ( auto& um : usage_map )
```
timwoj, line 146: This could do the same as above with a structured binding to reduce the `um.first` and `um.second` usage, plus be a bit more descriptive. Something like:

```suggestion
	for ( auto& [location_info, count] : usage_map )
```

timwoj (changes requested):
The content looks good. I like the change to remove the uses of bare pointers quite a bit. I left some comments about C++ usage.

vpax:
@timwoj thanks for the style pointers!  (I especially like `for ( auto& [x, y] : ... )`.) I've integrated all of them in the latest push.

timwoj:
> I especially like `for ( auto& [x, y] : ... )`.

I started working on support for that construct with Zeek's `Dictionary` type but I hit a bunch of snags around hash keys. I should go back to that and see if I can figure it out.

src/ScriptCoverageManager.cc
```diff
@@ -127,31 +124,47 @@ bool ScriptCoverageManager::WriteStats()
...
 		}
 
-	map<pair<string, string>, uint64_t>::const_iterator it;
-	for ( auto& um : usage_map )
+	for ( auto& [func, body] : func_instances )
 		{
-		fprintf(f, "%" PRIu64 "%c%s%c%s\n", um.second, delim, um.first.first.c_str(), delim,
-		        um.first.second.c_str());
+		auto ft = func->GetType<FuncType>();
+		auto desc = ft->FlavorString() + " " + func->Name() + " TOTAL";
```
timwoj, line 137: I'm not sure about "TOTAL" here. That makes me think that it's a "total usage" count of some sort and then wonder where the actual value is. I don't have any better ideas though. Maybe "FULL"?

vpax: Yeah I wondered about that too. I wanted _some_ simple way of pulling out the function lines (it also works to grep on _`<tab>`_`(event|hook|function)` but that's certainly ugly).

How about `BODY`, given it's the coverage count for exactly one body? (An event handler, for example, might have several of these - each with different location information of course.)

timwoj: Yeah, that sounds better 👍🏼

timwoj (approved):
Looks good to me now. If you squash down at least that fixup commit, I'll merge it.

PR #2977, opened 2023-04-25
Update AF-Packet submodule

awelzel:
* Mask VLAN ID from tp_vlan_tci field to fix vlan > 4095 reported by Zeek when PCP and/or DEI bits are set.
* Descriptive error message when interface is down. Instead of "Invalid argument", Zeek now reports "interface is down".

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -277,6 +277,10 @@ Changed Functionality
 - The IRC base script now use ``file_sniff()`` instead of ``file_new()`` for
   DCC file transfers to capture ``fuid`` and inferred MIME type in irc.log.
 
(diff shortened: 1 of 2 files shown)


0xxon (approved):
Looks good.

Just one comment - if this enables additional tests (which is indicated by the changed TestDirs directive), perhaps a fullci run would make sense.

awelzel:
> Looks good.
> 
> Just one comment - if this enables additional tests (which is indicated by the changed TestDirs directive), perhaps a fullci run would make sense.

Good call. Re-pushed after labeling "fullci".

PR #2978, opened 2023-04-25
smb2: Limit per-connection read/ioctl/tree state

awelzel:
Users on Slack observed memory growth in an environment with a lot of SMB traffic. jeprof memory profiling pointed at the offset and fid maps kept per-connection for smb2 read requests.

These maps can grow unbounded if responses are seen before requests, there's packet drops, just one side of the connection is visible, or we fail to parse responses properly.

Forcefully wipe out these maps when they grow too large and raise smb2_discarded_messages_state() to notify script land about this.

Additionally, this implements an `smb2_discarded_messages_state()` handle in the base scripts similar to what the external [zeek-smb-clear-state](https://github.com/corelight/zeek-smb-clear-state) package is doing, discarding script-layer state when the parser state was discarded.

Screenshot-20230414-113226

References:
https://zeekorg.slack.com/archives/CSZBXF6TH/p1679551396832949


awelzel:
@JustinAzoff - because we spoke on Slack. This would be the first idea for a "discarded state event" for SMB that also wipes the script-layer state. Any feedback on name or the general idea is welcome if you have the time.

I'll try to get a test-case to trigger this. One idea would be bpf filtering one side of a connection and overwriting the limit to something much smaller.

awelzel:
Request for testing from the original reporters :crossed_fingers: 
https://zeekorg.slack.com/archives/CSZBXF6TH/p1682515595288909?thread_ts=1679551396.832949&cid=CSZBXF6TH

awelzel:
> Request for testing from the original reporters crossed_fingers

They may only be able to test mid May. I'm relatively confident with the changes and it would be great to have them in 6.0 modulo thoughts on the current approach. The previous iteration of the patch showed promising results, and the current version is mostly a bit generalized.

This is the before-after graph in the user's environment:

image (1)

scripts/base/init-bare.zeek
```diff
@@ -2983,6 +2983,14 @@ export {
 	##
 	## .. zeek:see:: smb_pipe_connect_heuristic
 	const SMB::pipe_filenames: set[string] &redef;
+
+	## The maximum number of messages for which to retain state
+	## about offsets, fids, or tree ids before discarding these
+	## to free up internal parser state. When state is discarded,
+	## :zeek:see:`smb2_discarded_messages_state` is raised.
```
timwoj, line 2990: > before discarding these to free up internal parser state

Is this true if `enable_clear_state` is set to false? If that's the case, `smb2_discarded_messages_state` just reports the weird and exits.

src/analyzer/protocol/smb/smb2-com-ioctl.pac
```diff
@@ -17,6 +17,17 @@ refine connection SMB_Conn += {
 
 	function proc_smb2_ioctl_request(val: SMB2_ioctl_request) : bool
 		%{
+		if ( zeek::BifConst::SMB::max_pending_messages &&
```
timwoj, line 20:
```suggestion
		if ( zeek::BifConst::SMB::max_pending_messages > 0 &&
```

src/analyzer/protocol/smb/smb2-com-read.pac
```diff
@@ -34,6 +34,17 @@ refine connection SMB_Conn += {
 			                                    ${val.read_len});
 			}
 
+		if ( zeek::BifConst::SMB::max_pending_messages &&
```
timwoj, line 37:
```suggestion
		if ( zeek::BifConst::SMB::max_pending_messages > 0 &&
```

src/analyzer/protocol/smb/smb2-protocol.pac
```diff
@@ -230,6 +230,16 @@ refine connection SMB_Conn += {
 		%{
 		if ( is_orig )
 			{
+			if ( zeek::BifConst::SMB::max_pending_messages &&
```
timwoj, line 233:
```suggestion
			if ( zeek::BifConst::SMB::max_pending_messages > 0 &&
```

timwoj (changes requested):
One question and some nits.

scripts/base/init-bare.zeek
awelzel: > Is this true if enable_clear_state is set to false? If that's the case, smb2_discarded_messages_state just reports the weird and exits.

The "internal parser state" (those std::maps in binpac) is cleared even when enable_clear_state=F. That setting only toggles script-layer state resetting. I see now that this is confusing. Didn't have `enable_clear_state` initially - I'll stuff a `_script_state` into the name, though not sure that's immediately better.

timwoj (approved):

PR #2979, opened 2023-04-25
Allow plugins to call find_package(Zeek)

Neverlord:
Fixes an issue where plugins would skip the bootstrapping phase when calling `find_package(Zeek)` prior to including `ZeekPlugin.cmake`. This affects the spicy-plugin when building it as dynamic plugin outside of Zeek.


timwoj:
@Neverlord Is this PR still needed after the flurry of other PRs for cmake that have landed?

Neverlord:
This originally came up in the Spicy plugin. That's moving to the main repo now, so it probably won't need to go find Zeek anymore. I don't know if there are other plugins that do `find_package(Zeek)`, but I think adding these changes is still worth it to avoid similar problems with other plugins in the future.

timwoj:
I grepped through all of the packages on the package repo and don't see any other users. I'll go ahead and pull this in.

PR #2982, opened 2023-04-25
intel: support reading files directly on workers

JustinAzoff:
Add Intel::read_files_on_workers option that allows workers to read intel files directly and skip any intel cluster communication based distribution.


JustinAzoff:
I think this is mostly right.. only issue is you can't call `Intel::insert` on a worker anymore to add a new item to the cluster.

I think it might be possible to fix that though by changing some things around.  The intel framework doesn't have a concept of where the intel item came from. If doesn't track it was from the files - which would no longer need to be synced across, and if they came from someone calling `Intel::insert` in a script. an extra param to insert could maybe fix that, or just by reworking the internal stuff to use a different method and having `Intel::insert` ONLY be used for custom scripts dynamically adding intel data.

JustinAzoff:
Oh, this also needs to handle remove_indicator locally. grr.

JustinAzoff:
handling remove_indicator has the same problem.. something like this would work

```
 @if ( Cluster::local_node_type() == Cluster::MANAGER )
 event zeek_init()
        {
-       Broker::auto_publish(Cluster::worker_topic, remove_indicator);
+       if ( !Intel::read_files_on_workers)
+               Broker::auto_publish(Cluster::worker_topic, remove_indicator);
        }
```

but then the manual remove wouldn't work, just like the manual add no longer works.

awelzel:
@JustinAzoff - are you going to move this further along or switching approaches altogether in which case we could close it.

JustinAzoff:
I think after our chat the other day the best path forward is just to ensure that broker performance is good in the general case.  While fixing this would improve things, it wouldn't help other, non-intel broker use cases.

Issue #2394, opened 2022-09-02
Manager runs out of memory distributing large input file to proxies


pbcullen:
1. Intel is supposed to be distributed through the proxies. There is some code in the intel framework for this:

```
# On the manager, the new_item event indicates a new indicator that
# has to be distributed.
event Intel::new_item(item: Item) &priority=5
    {
    local pt = Cluster::rr_topic(Cluster::proxy_pool, "intel_insert_rr_key");

    if ( pt == "" )
        # No proxies alive, publish to all workers ourself instead of
        # relaying via a proxy.
        pt = Cluster::worker_topic;

    Broker::publish(pt, Intel::insert_indicator, item);
    }
```

However, at startup the input file is loaded from zeek_init and that code runs before the proxies have fully connected to the manager. This results in the manager attempting to directly distribute the data to all the workers concurrently.

Moving the zeek_init event in scripts/base/frameworks/intel/input.zeek to something that gets triggered 10s after startup appears to mostly fix the problem. That gives the proxies 10s to connect to the manager before the manager starts reading the input file.

2. Even the manager distributing the data should not OOM
Despite the data file only being 50MB, and there only being 16-30 workers, this very quickly blows up a 4G manager limit.

3. Alternatively, this could be solved by syncing the file to all hosts and having all cluster processes read the input file directly off the disk. Then, an intel "single box cluster" mode could be added that avoids using most of the code in scripts/base/frameworks/intel/cluster.zeek.

awelzel:
Hey @pbcullen  - thanks. Which Zeek version has this been observed with? Is this something new with 5.0, or has been around with Zeek 4.0/4.2 already just never reported?

Tangentially, looking at some lines above the ones you pointed out, wonder if these can make a manager go OOM if once restart all workers at once? The lines you pointed out would be somewhat surprising to me unless publishing to the worker_topic buffers for each worker individually, but haven't actually tested anything.

https://github.com/zeek/zeek/blob/61c001a57e9b0575b7fb8c878872707459a70c7c/scripts/base/frameworks/intel/cluster.zeek#L31-L40

pbcullen:
This issue was found on Zeek 4.2.2.

JustinAzoff:
> unless publishing to the worker_topic buffers for each worker individually

As I understand it publishing the same message to multiple connected peers copies the message once for each peer.

initconf:
So I have faced this issue many times under different heuristics and conditions
over years.

[...]

PR #2986, opened 2023-04-27
Broker: Make GetNextTimeout() readiness configurable

awelzel:
Introduce a new const &redef Broker::use_zero_timeout_iosource to allow Broker's IO source be processed before the poll interval would pick it. This should reduce message processing latency significantly at the cost of less batching up of Broker messages. Not enabled by default due to potential subtleness of the change.


awelzel:
I'm closing this. It was mostly as an idea/improvement for #2394 but seems other fixes will be tried first. @J-Gras, if you're interested in this let me know separately, but you should be able to set `io_poll_interval_default=1` to mostly get same effect with current master (at the cost of more polling).

PR #2989, opened 2023-04-27
CI Platform updates

### LINKED ISSUE #2980: Add 'canary' build for some rolling Linux distro
We've run into a couple of cases lately where a Linux distro updates a package and we miss the update because our CI Docker images are cached and not updated unless we manually do it. We should add a build that updates to the latest version of all of the packages outside of the image caching (fedora-rawhide, debian-unstable, etc). This would let us catch these things earlier.

timwoj:
- Remove EOL Fedora 36
- Add new Fedora 38
- Add Debian-Unstable (sid) build that runs apt-get upgrade during every build to force us to always see new packages

I marked the sid build to only run on master builds or with the fullci tag, but I could easily be swayed to make it run all the time.

Fixes #2980


.cirrus.yml
```diff
@@ -201,6 +201,14 @@ centos7_task:
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
 
+debian12_task:
+  container:
+    # Debian Unstable (sid)
+    dockerfile: ci/debian-12/Dockerfile
```
awelzel, line 207:
```suggestion
    dockerfile: ci/debian-unstable/Dockerfile
```

.cirrus.yml
```diff
@@ -201,6 +201,14 @@ centos7_task:
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
 
+debian12_task:
```
awelzel, line 204: Needs unstable in the task name?

.cirrus.yml
```diff
@@ -201,6 +201,15 @@ centos7_task:
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
 
+debian_unstable_task:
```
awelzel, line 204: In retrospect, may seem curious we run unstable but not testing, but good for now I think :-)

awelzel (approved):
One nit for adding `nodejs-devel` to Fedora 38.

The debian unstable distro doesn't currently ship with gcc-13. Fedora rawhide may be more bleeding-edge in general. Any reason you had preferred debian unstable?

I'm approving, but maybe we could add rawhide, too?

timwoj:
> I'm approving, but maybe we could add rawhide, too?

To be honest I'd rather just add one of the rolling distros. We already have opensuse-tumbleweed too that seems to be ahead of debian-unstable as well, so maybe that one should just be automatically updated?

timwoj:
> > I'm approving, but maybe we could add rawhide, too?
> 
> To be honest I'd rather just add one of the rolling distros. We already have opensuse-tumbleweed too that seems to be ahead of debian-unstable as well, so maybe that one should just be automatically updated?

Except for some reason, running `zypper patch` on tumbleweed didn't patch to python 3.11 even though the package site says it should be available. That's not terribly ideal.

PR #2999, opened 2023-04-29
Move build defaults from configure to CMake

### LINKED ISSUE #2711: CMake should define the same defaults as configure
I've run into this a few times on Windows. We don't define the same set of defaults inside the CMake files that we do inside of `configure`. Since we're not running `configure` on Windows to set everything up, and instead are running CMake directly, we're consistently missing features and such that should be enabled (the last one was zeek-aux not building because INSTALL_AUX_TOOLS wasn't set to true).

Here are the defaults as defined in `configure` currently:

```
append_cache_entry ENABLE_DEBUG BOOL false
append_cache_entry ENABLE_PERFTOOLS BOOL false
append_cache_entry ENABLE_JEMALLOC BOOL false
append_cache_entry ENABLE_ZEEK_UNIT_TESTS BOOL true
append_cache_entry BUILD_SHARED_LIBS BOOL true
append_cache_entry INSTALL_AUX_TOOLS BOOL true
append_cache_entry INSTALL_BTEST BOOL true
append_cache_entry INSTALL_BTEST_PCAPS BOOL true
append_cache_entry INSTALL_ZEEK_ARCHIVER BOOL true
append_cache_entry INSTALL_ZEEK_CLIENT BOOL true
append_cache_entry INSTALL_ZEEKCTL BOOL true
append_cache_entry INSTALL_ZKG BOOL true
append_cache_entry CPACK_SOURCE_IGNORE_FILES STRING
append_cache_entry ZEEK_SANITIZERS STRING ""
append_cache_entry ZEEK_INCLUDE_PLUGINS STRING ""
```

Neverlord:
Moving the defaults for build variables from the `configure` script to `CMakeLists.txt` gives the same default behavior on platforms where the `configure` script is not available (Windows) and also allows a pure CMake-based work flow (e.g., the standard `cmake -S . -B build`) without having to manually adjust the defaults.

The `configure` script also becomes much simpler as a result.

@timwoj the original ticket was about adding the defaults to CMake in addition to having them in the `configure` script, but I think we make life much easier if there's a single point in the build process where the defaults come from. And the natural place for this is the CMake file.

I've also added a new option to the configure script: `--show-config`. This simply greps the CMake cache for the most relevant entries. Browsing the CMake cache manually can be quite overwhelming. And even `cmake -LAH` contains a lot of entries that most people using the `configure` script just don't care (or know) about. I've used this `grep` statement to compare CMake caches before and after the changes and though it might be useful to others. Let me know if you think otherwise.

Closes #2711.


configure
```diff
@@ -198,6 +169,14 @@ while [ $# -ne 0 ]; do
             echo "${usage}" 1>&2
             exit 1
             ;;
+        --show-config)
+            if [ ! -f "$builddir/CMakeCache.txt" ]; then
+                echo "Error: no CMake build found under '$builddir'." 1>&2
+                exit 1
+            fi
+            grep -E "^ENABLE_|^ZEEK_|^INSTALL_|^CMAKE_INSTALL_PRE|^CMAKE_C.*_FLAGS|^CMAKE_BUILD" "$builddir/CMakeCache.txt" | grep -v ':INTERNAL'
```
bbannier, line 177: I wonder whether it would be useful to include `CMAKE_CXX_COMPILER`/`CMAKE_C_COMPILER` here. Right now this is injected through `CXX=... ./configure` and it not showing up in `build/config.status` always lead me to question installations.

Neverlord:
Ah, one thing I forgot to mention regarding `ZEEK_ROOT_DIR`: we force this to be equal to CMAKE_INSTALL_PREFIX during the CMake run. The configure script does not allow to set this individually (it’s always set to `CMAKE_INSTALL_PREFIX` as well), so we don’t need this as a cache variable.

configure
Neverlord: Sure, if we agree to keep this option, I'll update the regex string.

CMakeLists.txt
```diff
@@ -9,12 +9,62 @@ endif()
...
+  unset(build_type_lower)
+endif ()
+
+# CMake options (Boolean flags).
+option(BUILD_SHARED_LIBS "Build targets as shared libraries." ON)
+option(ENABLE_DEBUG "Build Zeek with additional debugging support." ${ENABLE_DEBUG_DEFAULT})
+option(ENABLE_JEMALLOC "Link against jemalloc." OFF)
+option(ENABLE_PERFTOOLS "Build with support for Google perftools." OFF)
 option(ENABLE_ZEEK_UNIT_TESTS "Build the C++ (doctest) unit tests?" ON)
+option(ENABLE_ZEEK_UNIT_TESTS "Build the C++ unit tests." ON)
```
timwoj, line 35: `ENABLE_ZEEK_UNIT_TESTS` got duplicated here.

CMakeLists.txt
```diff
@@ -9,12 +9,62 @@ endif()
...
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON
-    CACHE INTERNAL "Write JSON compile commands database")
+    CACHE INTERNAL "Whether to write a JSON compile commands database")
+
 set(ZEEK_CXX_STD cxx_std_17 CACHE STRING "The C++ standard to use.")
 
+set(ZEEK_SANITIZERS "" CACHE STRING "Sanitizers to use when building.")
+
+set(CPACK_SOURCE_IGNORE_FILES "" CACHE STRING "CPack source ignore files.")
```
timwoj, line 69:
```suggestion
set(CPACK_SOURCE_IGNORE_FILES "" CACHE STRING "Files to be ignored by CPack")
```

configure
timwoj: I like this option since digging through CMakeCache.txt can be annoying. We could also update btest to use this instead as well.

timwoj (changes requested):

CMakeLists.txt
Neverlord: Fixed.

CMakeLists.txt
```diff
@@ -9,12 +9,68 @@ endif()
 
 project(Zeek C CXX)
 
+
+# We want to set ENABLE_DEBUG to ON by default if the build type is Debug.
+set(ENABLE_DEBUG_DEFAULT OFF)
+if ( NOT GENERATOR_IS_MULTI_CONFIG )
+  string(TOLOWER "${CMAKE_BUILD_TYPE}" build_type_lower)
+  if ( build_type_lower STREQUAL "debug" )
+    set(ENABLE_DEBUG_DEFAULT ON)
```
timwoj, line 18: Should this also be set if `RelWithDebInfo` is set? I can't remember all of what's different between those two modes.

Neverlord: `RelWithDebInfo` should use similar flags to `Release`, except that the binary still contains debug symbols to allow you to jump into a debugger (or generate meaningful stack traces). The `configure` script enabled this flag only for actual debug builds and I think keeping it this way makes sense.

configure
```diff
@@ -198,6 +169,14 @@ while [ $# -ne 0 ]; do
             echo "${usage}" 1>&2
             exit 1
             ;;
+        --show-config)
+            if [ ! -f "$builddir/CMakeCache.txt" ]; then
+                echo "Error: no CMake build found under '$builddir'." 1>&2
+                exit 1
+            fi
+            grep -E "^ENABLE_|^ZEEK_|^INSTALL_|^CMAKE_INSTALL_PRE|^CMAKE_C.*_FLAGS|^CMAKE_C.*_COMPILER|^CMAKE_BUILD" "$builddir/CMakeCache.txt" | grep -v ':INTERNAL'
```
awelzel, line 177: I've used/ran into LINKER_FLAGS previously that represents LDFLAGS usage:

```
$ cmake -S .  -LAH | grep CMAKE_EXE.*_FLAGS
CMAKE_EXE_LINKER_FLAGS:STRING=-fuse-ld=lld
...
```

timwoj: I added this as part of the merge.

Issue #2570, opened 2022-11-11
Windows: `setvbuf` has different semantics that on POSIX systems


timwoj:
I toyed with setting up some rudimentary tests with our external testing repo (zeek-testing) just to make sure some change we make doesn't inadvertently break windows builds while I get btest back into shape (see #2567).

I tried to read in one of the pcaps, and I ran into an assert in a call to `setvbuf` here: 

https://github.com/zeek/zeek/blob/a8fc63e182d40781ab6c15b80d1cf0b344037e29/src/File.cc#L212

The problem is that Microsoft's version of `setvbuf` doesn't match the semantics of the POSIX version. In the POSIX version, you can pass NULL for the buffer and zero for the length and it'll only set the mode. It'll automatically create the buffer underneath on the next call to read/write. In Windows, you're required to pass an actual buffer and a length between 2 and INT_MAX, unless you pass `_IONBF` which disables the buffering entirely.

As part of the Windows PR, I wrapped the call in an `#ifdef` block that passes `_IONBF` for those systems, but we should revisit this to see if there's a better way to do this on Windows and whether there's any sort of performance implications with turning off the buffering.

timwoj:
It looks like we're definitely not the first people to run into this: https://postgrespro.com/list/thread-id/1844585

Postgres opted to just disable buffering altogether for Windows, basically how it's done in our code: 

https://github.com/postgres/postgres/blob/684ffac8c7d2647d919495727d469c73129202ea/src/include/port.h#L352-L364

I imagine if something like Postgres is ok with whatever performance hit they're taking from that, we're probably ok with taking the same.

timwoj:
Alternatively, we could pass some arbitrary value for the length. That assumes we can come up with something reasonable that would map to the typical length of files we open here.

timwoj:
This was fixed in cc811eca29e688a8eb171e037cab251760695593.

PR #3002, opened 2023-04-30
Use workaround for setvbuf on Windows in DebugLogger/Extract file analyzer

timwoj:
We already have this workaround in src/File.cc, but apparently I missed a few uses of `setvbuf` in other places. They need the same workaround to avoid calling `abort`. This fixes some aborts when running btests on Windows. Note this doesn't do the performance investigation requested in https://github.com/zeek/zeek/issues/2570, it just fixes the additional aborts.

### DIFF
--- a/src/DebugLogger.cc
+++ b/src/DebugLogger.cc
@@ -56,7 +56,7 @@ void DebugLogger::OpenDebugLog(const char* filename)
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


timwoj:
> The existing File.cc version as a reporter->Error() in case of failure.

That's how it was prior to any of the Windows changes being made.

> Should this go to zeek::util

Yeah, that's probably a good idea. I'll do that before merging.

PR #3003, opened 2023-04-30
Fix smith-waterman sorting to follow correct Compare semantics

timwoj:
Windows currently throws an exception with the language.smith-waterman-test btest, resulting in a call to `abort()`. The sort comparator in the Smith-Waterman code triggers the exception because the comparator will return true if two elements are equal, where their std::sort implementation requires the comparator to return false in that case. Effectively it's expecting the same semantics as `std::less`. This PR fixes the comparator to only use `<` instead of `<=`, which fixes the exception. 

Unfortunately, it also changes the output from the test, so there's a consideration about whether this is correct or not.

### DIFF
--- a/src/SmithWaterman.cc
+++ b/src/SmithWaterman.cc
@@ -172,10 +172,7 @@ bool SubstringCmp::operator()(const Substring* bst1, const Substring* bst2) cons
 		return false;
 		}
 
-	if ( bst1->GetAlignments()[_index].index <= bst2->GetAlignments()[_index].index )
-		return true;
-
-	return false;
+	return (bst1->GetAlignments()[_index].index < bst2->GetAlignments()[_index].index);
 	}
 
 // A node in Smith-Waterman's dynamic programming matrix.  Each node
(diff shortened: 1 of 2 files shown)


awelzel (approved):
I think that's okay.

My skim of the code is that the sorting is based on the offsets in the longer string, but they are each  `5` for the following input:
```
xxAAxxAAxx - yyyyyAAyyyyy:
```

The following addition would use the index location in the other string as arbiter (and possibly make things more predictable/stable), but not sure it matters (and the (index + 1) % 2 obviously assumes things).

```
--- a/src/SmithWaterman.cc
+++ b/src/SmithWaterman.cc
@@ -172,10 +172,18 @@ bool SubstringCmp::operator()(const Substring* bst1, const Substring* bst2) cons
                return false;
                }
 
-       if ( bst1->GetAlignments()[_index].index <= bst2->GetAlignments()[_index].index )
+       int idx1 = bst1->GetAlignments()[_index].index;
+       int idx2 = bst2->GetAlignments()[_index].index;
+
+       if ( idx1 < idx2 )
                return true;
 
-       return false;
+       if ( idx1 > idx2 )
+               return false;
+
+       int oidx1 = bst1->GetAlignments()[(_index + 1) % 2].index;
+       int oidx2 = bst2->GetAlignments()[(_index + 1) % 2].index;
+       return oidx1 < oidx2;
        }
```

PR #3012, opened 2023-05-03
SSL: do not try to disable failed analyzer

0xxon:
Currently, if a TLS/DTLS analyzer fails with a protocol violation, we will still try to remove the analyzer later, which results in the following error message:

error: connection does not have analyzer specified to disable

Now, instead we don't try removing the analyzer anymore, after a violation occurred.


scripts/base/protocols/ssl/main.zeek
```diff
@@ -499,6 +499,15 @@ event analyzer_confirmation_info(atype: AllAnalyzers::Tag, info: AnalyzerConfirm
 		}
 	}
 
+event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo) &priority=5
+	{
+	if ( atype == Analyzer::ANALYZER_SSL || atype == Analyzer::ANALYZER_DTLS )
+		{
+		# analyzer errored out; prevent us from trying to remove it later
+		delete info$c$ssl$analyzer_id;
```
awelzel, line 507: Hmm.. could also see invoking `disable_analyzer()` with `err_if_no_conn=F` instead to suppress the error message?

There's an `analyzer_violation_info()` handler a few lines below with the same priority that ends-up in `finish()` and attempts to call `disable_analyzer()`, so the interaction of these two currently might not be obvious.

0xxon: Oh, I am blind. You are right - there are two handlers. I will fold it into the second one - that will also make this definitely deterministic.

I like that slightly better than suppressing the error message, but I guess one can have different opinions about that :)

scripts/base/protocols/ssl/main.zeek
```diff
@@ -499,6 +499,10 @@ event analyzer_confirmation_info(atype: AllAnalyzers::Tag, info: AnalyzerConfirm
 		}
 	}
 
+event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo) &priority=5
+	{
+	}
+
```
awelzel, line 505: That's now looking a bit odd, no? :-)

scripts/base/protocols/ssl/main.zeek
```diff
@@ -514,5 +518,9 @@ event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationI
 	{
 	if ( atype == Analyzer::ANALYZER_SSL || atype == Analyzer::ANALYZER_DTLS )
 		if ( info$c?$ssl )
+			{
+			# analyzer errored out; prevent us from trying to remove it later
+			delete info$c$ssl$analyzer_id;
 			finish(info$c, T);
```
awelzel, line 524: Nit: That `T` actually is a indication whether we want to call `disable_analyzer()`, but that bites with above deletion of `analyzer_id`, so would be less confusing to pass `F` here.

awelzel (approved):
The empty handler should go, the other it would be to avoid some potential confusion of future readers around intentions.

scripts/base/protocols/ssl/main.zeek
0xxon: I so thought I removed that *sigh*

PR #3014, opened 2023-05-04
Add `HookDispatchEvent()`

J-Gras:
This adds a new hook to handle event dispatching. In contrast to the existing `HookEnqueueEvent()`, this hook will be called when the event is actually about to be executed.


J-Gras:
> It would be good to motivate with a use-case and possibly contrast why `HookCallFunction()` didn't work out.

Good point! I was looking for a hook to deal with event metadata. In particular the source and the new timestamp. I figured, it would be possible to use `HookCallFunction()` and then use the `EventMgr`'s `Current...()` methods to obtain the desired info. However, that did not feel as clean as introducing the hook. In particular `HookCallFunction()` is called for _all_ functions and hooks as well.

awelzel (approved):
Seems convenient and avoids `HookCallFunction()` stunts :+1:

src/Event.cc
```diff
@@ -48,6 +48,11 @@ void Event::Dispatch(bool no_remote)
 	if ( src == util::detail::SOURCE_BROKER )
 		no_remote = true;
 
+	bool done = PLUGIN_HOOK_WITH_RESULT(HOOK_DISPATCH_EVENT, HookDispatchEvent(this, no_remote),
```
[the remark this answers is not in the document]
J-Gras: My thought here was that it could be surprising if `Dispatch()` ends up using another value for `no_remote` than the one passed to the hook. However, true that the caller provided value gets lost this way. I don't have a clear preference so feel free to change it.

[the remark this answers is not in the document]
ckreibich: It seems fine to me here to pass `no_remote` after the Broker case. Looks like a `no_remote = true` happens only in sufficiently few scenarios (I'm spotting `net_done` and `file_open`, but didn't hunt for long) that if a plugin really needs to dig into details, it can do so. Also, if we turned it around, plugin authors might still have to infer that `no_remote` was about to be flipped due to the Broker case. So I think it's fine.

I did wonder whether the `no_broker` state is something that this (or another) hook should be able to change, but I can't think of uses.

src/plugin/Manager.h
```diff
@@ -311,6 +311,18 @@ class Manager
 	 */
 	bool HookQueueEvent(Event* event) const;
 
+	/**
+	 * Hook that filters the dispatching of an event.
+	 *
+	 * @param event The event to be executed; it may be modified.
```
ckreibich, line 317: As I read this I wondered if that itself is a use-case — modifying the event at dispatch time vs enqueue time, say for time-tracking or other internal diagnostics.

src/plugin/Manager.h
```diff
@@ -311,6 +311,18 @@ class Manager
...
+	 *
+	 * @param no_remote True if the event should not be auto-published.
+	 *
+	 * @return Returns true if a plugin handled the dispatching; in that case
+	 * the plugin will have taken ownership.
```
ckreibich, line 322: I think you modeled that comment on `HookQueueEvent`, right? I wonder if taking ownership is as clear here, or if it even applies. `EventMgr::Dispatch()` unrefs the event, so that happens regardless. In `EventDispatch()` there's [this block](https://github.com/zeek/zeek/blob/master/src/Event.cc#L64-L66) further down that manually unrefs the `obj` member — I think this means that for a plugin to do this right, it would have to know about this, which feels wrong ... we should probably do that unref if `PLUGIN_HOOK_WITH_RESULT` returns `true`. And the second part of this sentence could just go.

Thoughts?

src/plugin/Plugin.h
```diff
@@ -1065,6 +1066,28 @@ class Plugin
...
+	 * @param event The event to be dispatched. The method can modify it in
+	 * place as long as it ensures matching types and correct reference
+	 * counting.
+	 *
+	 * @param no_remote True if the event should not be auto-published to
+	 * remote peers (e.g., if the event itself was received from remote).
+	 *
+	 * @return True if the plugin took charge of the event; in that case it
+	 * must have assumed ownership of the event and the event manager will not
+	 * do anything further with it. False otherwise.
```
ckreibich, line 1087: This touches the same ownership question I mentioned earlier — just flagging here too.

ckreibich (commented):
Thanks Jan! I'm fine with the relationship to `HookCallFunction` — I think there's enough event-specific thinking in Zeek that such a hook makes good sense. My main concern is around memory management as a plugin blocks the dispatch, see comments, so I'm not approving just yet.

src/plugin/Manager.h
J-Gras: Gotcha, I was more or less blindly following `HookQueueEvent`. I think you're right: In contrast to the event queuing, this hook fires when the event *is already dispatched* from the perspective of the manager. We might prevent that it gets executed by altering how the event itself behaves but the dispatching cannot be undone. I added the `Unref(obj)` as suggested and revised the docstrings.

ckreibich (approved):
This looks good to me. @rsmmr, any chance you could look this over too? You know plugins best, I think...

src/Event.cc
```diff
@@ -48,6 +48,15 @@ void Event::Dispatch(bool no_remote)
...
+	                                    false);
+	if ( done )
+		{
+		if ( obj )
+			Unref(obj);
+		return;
```
rsmmr, line 57: Formatting nit:

```suggestion
			Unref(obj);
		
		return;
```

rsmmr (commented):
Just for my understanding, I haven't been following what the use case is for hooking to dispatching vs queuing: in what cases is that timing difference important? 

I'm generally also fine with having a dedicated version for events, in addition to the existing function hook. However, I'm wondering if longer-term we might then want to move to a clean separation: one only for functions and hooks, and one only for events (i.e., change CallFunction to not execute for events). Not backwards-compatible obviously, so that would need a migration path.

J-Gras:
> Just for my understanding, I haven't been following what the use case is for hooking to dispatching vs queuing: in what cases is that timing difference important?

I have a use case where I would like to alter overall Zeek behaviour based on the event's metadata. To be more precise: I would like to use the timestamps attached to events to control time. In that case, it would be possible that multiple events are queued with significantly different timestamps. Hence, I looked into hooking to dispatching rather then queueing. Does that make sense?

rsmmr:
> In that case, it would be possible that multiple events are queued with significantly different timestamps. Hence, I looked into hooking to dispatching rather than queueing. Does that make sense?

I understand the use case now but it still feels very specific and somewhat ill-defined to me, and I'm wondering if we could support that differently (and better) through a new API to advance network time in a more controlled fashion that could be used from the existing queuing hook. I don't want to block this if you don't think so, merging is fine with me then; it's just that I have hard time seeing much other use cases for the new hook.

J-Gras:
> I understand the use case now but it still feels very specific and somewhat ill-defined to me, and I'm wondering if we could support that differently (and better) through a new API to advance network time in a more controlled fashion that could be used from the existing queuing hook. I don't want to block this if you don't think so, merging is fine with me then; it's just that I have hard time seeing much other use cases for the new hook.

I am happy to reiterate on the approach for network time management. I'll re-check whether it's feasible to stick with `HookCallFunction()` for the current prototype. So converting this to a draft for now.

PR #3015, opened 2023-05-04
ci: Add cleanup of zeek/zeek-dev public ECR repository

awelzel:
Script that batch-deletes images from zeek/zeek-dev. Added to .cirrus.yml so it runs for merges to master.

### DIFF
--- a/ci/public-ecr-cleanup.sh
+++ b/ci/public-ecr-cleanup.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+#
+# Script to batch-delete all untagged images from ECR public repositories,
+# defaulting to the zeek/zeek-dev repository.
+# First scans for manifest list images that are referencing other images and
+# deletes them, then deletes all remaining untagged images.
+set -eu
+
+if ! command -v aws >/dev/null; then
+    echo "missing aws command" >&2
+    exit 1
+fi
+
+REGISTRY_ID=${REGISTRY_ID:-103243056077}
+REPOSITORY_NAME=${REPOSITORY_NAME:-zeek-dev}
+BATCH_DELETE_SIZE=${BATCH_DELETE_SIZE:-50}
+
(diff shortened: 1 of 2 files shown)


awelzel:
@timwoj - I've tested locally as well as having the `only_if` condition commented out. Think this should do it. It's configured to run after merges to master. We could consider running regularly/cron but doesn't seem a difference either way.

Permissions of the Cirrus CI docker user have been extended to be able to delete images in zeek/zeek-dev.

ci/public-ecr-cleanup.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+#    application/vnd.docker.distribution.manifest.list.v2+json
+#    application/vnd.oci.image.index.v1+json
+#
+# These reference other images, so we need to delete them first as
+# otherwise the referenced images can not be deleted.
+IMAGE_DIGESTS=$(aws ecr-public describe-images \
+    --registry-id "${REGISTRY_ID}" \
+    --repository-name "${REPOSITORY_NAME}" \
+    --query 'imageDetails[?!imageTags && (contains(imageManifestMediaType, `manifest.list.v2`) || contains(imageManifestMediaType, `image.index.v1`))].{imageDigest: join(`=`, [`imageDigest`, imageDigest])}' \
```
timwoj, line 51: I looked at the StackOverflow link in the original issue, and I see the use of a `--filter "tagStatus=UNTAGGED"` filter. Is the above just another/better way to do the same?

awelzel: The link shows instructions for `aws ecr` rather than `aws ecr-public` and I didn't find `list-images` nor `--filter` for the latter, so `describe-images --query` it was.

timwoj:
Does this need to be backported into the release branches?

awelzel:
> Does this need to be backported into the release branches?

Nope - we only release zeek/zeek-dev off of master.

PR #3016, opened 2023-05-04
NTP: Detect out-of-order packets

awelzel:
The NTP mode provides us with the identity of the endpoints. For the simple CLIENT / SERVER modes, flip the connection if we detect orig/resp disagreeing with what the message says. This mainly results in the history getting a ^ and the ntp.log / conn.log showing the corrected endpoints.


awelzel:
@vpax - this would be a patch to NTP for detecting out-of-order packets and flipping the connection and including your pcap as a test-case. I think that would help with the specific scenario you pointed out, but wonder if you ran into more/other cases by now?

PR #3021, opened 2023-05-05
cluster/manager: Warn if there are no proxies

### LINKED ISSUE #2871: Add "no proxies configured" warning
A cluster without any proxies configured will seem to work, but we have a few scripts that do require the availability of proxy processes. We should consider adding a warning (produced by the manager?) if there's no proxy defined in the cluster-layout.

              This is pretty bad since nothing complains if you don't spin up proxies — looks like `Cluster::publish_hrw()` [returns `F`](https://github.com/zeek/zeek/blob/master/src/broker/messaging.bif#L229-L230) if there are no nodes in the pool, but we're not actually checking that in any of its uses.

_Originally posted by @ckreibich in https://github.com/zeek/zeek-docs/issues/177#issuecomment-1469145118_

awelzel:
This assumes Reporter::warning() makes it to stderr and is somewhat visible for the manager process.

Closes #2871

### DIFF
--- a/scripts/base/frameworks/cluster/nodes/manager.zeek
+++ b/scripts/base/frameworks/cluster/nodes/manager.zeek
@@ -23,3 +23,14 @@ redef Log::default_rotation_interval = 24 hrs;
 ## Use the cluster's delete-log script.
 redef Log::default_rotation_postprocessor_cmd = "delete-log";
 @endif
+
+event zeek_init()
+	{
+	if ( Cluster::get_node_count(Cluster::PROXY) == 0 )
+		{
+		Reporter::warning("No proxy nodes configured for this Zeek cluster. "
+		                  + "The software framework as well as certain external "
+		                  + "scripts rely on the presence of proxy nodes to function "
+		                  + "properly and should be considered required.");
+		}
+	}
--- a/testing/btest/scripts/base/frameworks/logging/field-extension-cluster-error.zeek
+++ b/testing/btest/scripts/base/frameworks/logging/field-extension-cluster-error.zeek
@@ -12,6 +12,7 @@
 redef Cluster::nodes = {
 	["manager-1"] = [$node_type=Cluster::MANAGER, $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT1"))],
 	["worker-1"]  = [$node_type=Cluster::WORKER,  $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT2")), $manager="manager-1", $interface="eth0"],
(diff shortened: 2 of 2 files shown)


testing/btest/scripts/base/frameworks/logging/field-extension-cluster-error.zeek
```diff
@@ -12,6 +12,7 @@
 redef Cluster::nodes = {
 	["manager-1"] = [$node_type=Cluster::MANAGER, $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT1"))],
 	["worker-1"]  = [$node_type=Cluster::WORKER,  $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT2")), $manager="manager-1", $interface="eth0"],
+	["proxy-1"]   = [$node_type=Cluster::PROXY,   $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT2")), $manager="manager-1"],
```
timwoj, line 15: Does this fix a test that was already returning the error? Should there be a test that does so intentionally?

scripts/base/frameworks/cluster/nodes/manager.zeek
```diff
@@ -23,3 +23,14 @@ redef Log::default_rotation_interval = 24 hrs;
...
 @endif
+
+event zeek_init()
+	{
+	if ( Cluster::get_node_count(Cluster::PROXY) == 0 )
+		{
+		Reporter::warning("No proxy nodes configured for this Zeek cluster. "
+		                  + "The software framework as well as certain external "
+		                  + "scripts rely on the presence of proxy nodes to function "
+		                  + "properly and should be considered required.");
```
timwoj, line 34: > and should be considered required.

I'm not a big fan of this. If something should be considered required, then I feel like we should actually require it. That's a breaking change though.

testing/btest/scripts/base/frameworks/logging/field-extension-cluster-error.zeek
0xxon: Just seconding the idea that having a test that intentionally raises the error would be neat.

0xxon:
In random questions - how difficult would it be to fix the scripts that currently require proxies to just work if we don't have any?

I see that at least the intel framework seems to specifically call out and work around this case.

awelzel:
> In random questions - how difficult would it be to, instead, fix the scripts that currently require proxies to just work if we don't have any?

Yeah, I think this is a good discussion to have, probably not too hard, though assumptions about proxy existing has come up before (the known scripts are something in policy that assume they exist).

Maybe #2871 is a better place for this discussion. I'm going to close the PR for now. I'm not overly happy with this either as is.

Issue #3019, opened 2023-05-05
Bumping Zeek version busts ccache cache, again


awelzel:
When only bumping the `VERSION` and running `./configure && make` with ccache enabled, I'm seeing a large part of the Zeek tree being rebuild from scratch again without using ccache.

The difference here is build time difference of ~1min 20 vs ~5min 40s after pulling in a version bump for an ASAN build.

Skimming Cirrus CI, all build step of all tasks are taking longer after a merge to master than they used to, too. Unfortunately I'm not aware of a dashboard that would clearly showed this as a regression :-(

Think we need some follow-up for: https://github.com/zeek/zeek/pull/2972

@Neverlord - is there an "easy" way to set `ZEEK_CONFIG_SKIP_VERSION_H` for all files within the Zeek source tree or minimally make that the semantics? We have a great number of `#include zeek/zeek-config.h` in-tree and unless they happen to be internal plugins (which isn't a given), they pull in zeek/zeek-version.h now and that busts the cache when the version is bumped (for no good reason).

Think `ZEEK_CONFIG_SKIP_VERSION_H` should be the default unless compiling "external dynamic plugins" or "extra static plugins".

References: #2776 #2802 #2806

awelzel:
Fixed by #3022, thanks!

awelzel:
I was curious if we actually have any visibility into the ccache effectiveness. The graphs in the Cirrus Web UI aren't super great, as they don't reach far back and don't allow to filter by individual tasks.

Cirrus CI does provide a GraphQL API that allows querying builds and their tasks pretty nicely:
https://cirrus-ci.com/explorer
```
{
  ownerRepository(owner: "zeek", name: "zeek", platform: "github") {
    id
    builds(branch: "master", last:100) {
      edges {
        node {
          changeMessage
          id
          status
          durationInSeconds
          clockDurationInSeconds
          buildCreatedTimestamp
        tasks {
          id
          creationTimestamp
          status
          name
          durationInSeconds
        }
        }
      }
    }
  }
}
```
With some wrangling of the result and focusing on (debian, fedora and macos), these got faster at the end of February (when the ccache changes were first added). It's now taking quite a bit longer. It's all really noisy though :-/
![task_duration](https://github.com/zeek/zeek/assets/2100509/f767f30f-1a75-4bdf-aa66-dffbf4ce47dc)

PR #3022, opened 2023-05-05
Skip version.h by default for Zeek sources

Neverlord:
Relates #3019.

@awelzel this should do the trick. Let me know if this doesn't fully address the issue.


awelzel (approved):
Tested and fixes it, thanks.

PR #3029, opened 2023-05-08
Clean up ZEEK_CONFIG_SKIP_VERSION_H issues

### LINKED ISSUE #3019: Bumping Zeek version busts ccache cache, again

### LINKED ISSUE #3023: plugins: Dynamic-link-time check missing for externally build plugins
Zeek has a "dynamic-link-time check" to detect loading external dynamic plugins compiled for a different Zeek version in Plugin.h.

This got slightly more intricate during #2802, adding `#if` checking of `ZEEK_PLUGIN_INTERNAL_BUILD` to avoid including zeek-version.h within Plugin.h (which in turn is included in many places in the tree).

https://github.com/zeek/zeek/blob/ab4f231232f160362eba5e3d51a56e2dee5448c7/src/plugin/Plugin.h#L133-L139

With #2972, the `ZEEK_PLUGIN_INTERNAL_BUILD` define was removed and I had missed the usage in Plugin.h. This results in the  "dynamic-link-time check" currently not being in place for external dynamic plugins.

As a test case, when building an external plugin against Zeek 5.2.0 and looking at the symbols of the .so with `nm`, it lists `zeek_version_5_2_0_plugin_7` as undefined (it will be satisfied during dynamic linking from the Zeek executable)
```
$ nm build/lib/Zeek-JavaScript.linux-x86_64.so | c++filt | grep zeek_version
                 U zeek_version_5_2_0_plugin_7
```

When compiling with latest master, the plugin's .so does not list a zeek_version symbol and there will be no dynamic-link error when loading this plugin with the wrong Zeek version. At best a different symbol can't be resolved, worst, random crashes entail.
```
$ nm build/lib/Zeek-JavaScript.linux-x86_64.so | c++filt | grep zeek_version
$ 
```

@Neverlord , I was tempted to reuse `ZEEK_CONFIG_SKIP_VERSION_H` but it's not quite the same: The dynamic link check is only required for external dynamic plugins, so the `! ZEEK_PLUGIN_INTERNAL_BUILD` was kind of right. Probably need a new define to skip or add it?

I can sketch a btest to cover this as a regression test if that would be helpful.

Neverlord:
@awelzel as discussed. 🙂

Closes #3023 and closes #3019 (hopefully for real this time).

### DIFF
--- a/src/plugin/Plugin.h
+++ b/src/plugin/Plugin.h
@@ -130,10 +130,12 @@ public:
 #endif
 	inline Configuration() __attribute__((always_inline))
 		{
-// Only bake in a ZEEK_PLUGIN_ZEEK_VERSION reference into external
-// plugins. The internal ones are in the same binary so the runtime
-// link check shouldn't be needed and we can avoid ccache busting.
-#if defined(ZEEK_PLUGIN_INTERNAL_BUILD) && ! ZEEK_PLUGIN_INTERNAL_BUILD
+// Only bake in a ZEEK_PLUGIN_ZEEK_VERSION reference into external plugins. The
+// internal ones are in the same binary so the runtime link check shouldn't be
+// needed and we can avoid ccache busting. The define gets set in the
+// CMakeLists.txt file for the Zeek::Internal target, which only exists when
+// building Zeek itself.
+#ifndef ZEEK_PLUGIN_SKIP_VERSION_CHECK
 		bro_version = ZEEK_PLUGIN_ZEEK_VERSION;
 		zeek_version = ZEEK_PLUGIN_ZEEK_VERSION;
 #endif
--- a/src/plugin/Plugin.h
+++ b/src/plugin/Plugin.h
@@ -13,8 +13,10 @@
 
 // Avoid ccache busting of Plugin.h for internal plugins by
 // only including zeek/zeek-version.h if we're building an
-// external plugin.
-#if defined(ZEEK_PLUGIN_INTERNAL_BUILD) && ! ZEEK_PLUGIN_INTERNAL_BUILD
+// external plugin. The define gets set in the CMakeLists.txt file
+// for the Zeek::Internal target, which only exists when
+// building Zeek itself.
+#ifndef ZEEK_PLUGIN_SKIP_VERSION_CHECK
 #include "zeek/zeek-version.h"
 // Remove the BRO define in v6.1.
 #define BRO_PLUGIN_BRO_VERSION ZEEK_VERSION_FUNCTION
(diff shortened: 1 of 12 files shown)


src/plugin/Plugin.h
```diff
@@ -130,10 +130,12 @@ class Configuration
...
-// Only bake in a ZEEK_PLUGIN_ZEEK_VERSION reference into external
-// plugins. The internal ones are in the same binary so the runtime
-// link check shouldn't be needed and we can avoid ccache busting.
-#if defined(ZEEK_PLUGIN_INTERNAL_BUILD) && ! ZEEK_PLUGIN_INTERNAL_BUILD
+// Only bake in a ZEEK_PLUGIN_ZEEK_VERSION reference into external plugins. The
+// internal ones are in the same binary so the runtime link check shouldn't be
+// needed and we can avoid ccache busting. The define gets set in the
+// CMakeLists.txt file for the Zeek::Internal target, which only exists when
+// building Zeek itself.
+#ifndef ZEEK_PLUGIN_SKIP_VERSION_CHECK
```
awelzel, line 138: suspect line 17 needs this change as well? CI isn't quite happy either :-(

https://github.com/zeek/zeek/pull/3029/files#diff-8cb93601057dda9faeaaa75290dd6153f18ab6c12e6a2653e399e764f4fe8600R17

awelzel: This would be a btest that works on Linux tested with Zeek 5.2.1:

https://github.com/zeek/zeek/commit/46f33182357a154138e2cf97c85e9c3d38945558

It does use `nm` to find the undefined zeek_version symbol, so it's a pretty low level test, but given we lost the feature possibly useful still?

testing/btest/plugins/zeek-version-link.zeek
```diff
@@ -1,5 +1,4 @@
 # @TEST-DOC: Poking at internals: Expect an undefined zeek_version_X_Y_Z_plugin symbol in the plugin's .so/.dynlib. If this test turns out to be brittle, remove it, but we lost the mechanism.
-# @TEST-REQUIRES: nm --version
```
timwoj: This will likely break on Windows. Of course, the test is probably already broken on Windows for other reasons, but this will break it more.

awelzel: I thought there are no plugins on windows anyhow?

timwoj: Yeah, I guess plugins are completely broken as is. Ignore this.

timwoj (approved):

PR #3031, opened 2023-05-09
Enhancements for Event-Tracing

vpax:
We've started using `-E` event tracing to generate scripts capturing network behavior that we need to edit for privacy considerations.  Doing so has turned up a number of bugs, which this PR addresses:
  - The tracer now includes at the end of the generated script a report of potentially sensitive constants that appear in the script, such as IP addresses and strings.  This list makes it easier for a script editor to locate all of the elements of concern and ensure they're modified.
  - Previously, the tracer just gave up in the presence of unsupported types such as `opaque` values, which have no feasible representation as a Zeek script constant.  These are now instead reflected in the generated script in a way that makes them easy to find and remedy (such as by skipping statements that use them if not needed for the targeted behavior).
  - Bug fixes:
    - "Unspecified" aggregates such as empty `set()`/`table()`/`vector()` constructors used to result in the generation of bad scripting code
    - IPv6 constants would result in bad code because the corresponding `Describe` method does not include `[]`s when printing them, which are needed in order to interpret them as Zeek scripting constants.
    - Running without a packet source would generate an extraneous `zeek_init()` event, leading to consistency problems.

I'll be putting in a companion `zeek-docs` PR reflecting the first two items shortly.

### DIFF
--- a/src/EventTrace.cc
+++ b/src/EventTrace.cc
@@ -213,12 +213,16 @@ void ValTrace::ComputeDelta(const ValTrace* prev, DeltaVector& deltas) const
 			// use the constant representation instead.
 			break;
 
+		case TYPE_ANY:
 		case TYPE_FILE:
 		case TYPE_OPAQUE:
-		case TYPE_ANY:
-			// These we have no way of creating as constants.
-			reporter->Error("cannot generate an event trace for an event of type %s",
-			                type_name(tag));
+			// If we have a previous instance, we can ignore this
+			// one, because we know it's equivalent (due to the
+			// test at the beginning of this method), and it's
+			// not meaningful to recurse inside it looking for
+			// interior changes.
+			if ( ! prev )
+				deltas.emplace_back(std::make_unique<DeltaUnsupportedCreate>(this));
 			break;
 
 		case TYPE_LIST:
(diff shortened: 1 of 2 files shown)


vpax:
Documentation PR: https://github.com/zeek/zeek-docs/pull/183

src/EventTrace.cc
```diff
@@ -871,63 +899,28 @@ const std::string& ValTraceMgr::ValName(const ValPtr& v)
...
-					}
-				}
-
-			else if ( tag == TYPE_FUNC )
-				rep = v->AsFunc()->Name();
-
-			else if ( tag == TYPE_TIME )
-				rep = std::string("double_to_time(") + std::to_string(v->AsDouble()) + ")";
+		val_names[v.get()] = GenValName(v);
+		find = val_names.find(v.get());
```
timwoj, line 903: I'm not sure how much it matters here but if you use `insert` or `insert_or_assign` for the actual insertion, you get back a pair that includes the iterator to the element inserted. That'd avoid the extra lookup.

timwoj (approved):
Just one suggestion for avoiding an extra map lookup. It might not matter though, so feel free to ignore it.

timwoj:
Is there anything here that needs to be called out in NEWS?

src/EventTrace.cc
vpax: Nice to know.  I've pushed this as a `--fixup`, I like how it visually streamlines the code somewhat (though also makes it a tad more cryptic due to more meta-characters).

vpax:
> Is there anything here that needs to be called out in NEWS?

I'm inclined to skip it.  There's little evidence of any uptake of event tracing to date - in particular, it was completely broken for processing anything with TLS traffic prior to these fixes, and I imagine also for a number of other protocols.  I'd feel differently if the PR was addressing Issues filed by others.

That said, happy to draft some text if you think it should be included.

PR #3032, opened 2023-05-09
Remove vendored CMake files from repo

timwoj:
This PR updates the cmake submodule to remove some files that are included in every CMake distribution that we support. It's unimportant for 6.0 though and can wait until after that branches.


rsmmr (approved):

timwoj:
I added updates from https://github.com/zeek/bifcl/pull/28 and https://github.com/zeek/binpac/pull/28 since those required updates due to the removal of FindBISON.

PR #3034, opened 2023-05-09
Prevent uid duplicates in supervisor mode

J-Gras:
For seeded runs, this incorporates the node names of supervised nodes into the uid seed so that different nodes will create different uids. Before, each node inherited the same seed and thus created the same uids leading to duplicates in the logs.


src/util.cc
```diff
@@ -2165,6 +2166,24 @@ uint64_t calculate_unique_id(size_t pool)
 			unique.pid = getpid();
 			unique.rnd = static_cast<int>(detail::random_number());
 
+			uid_instance = zeek::detail::HashKey::HashBytes(&unique, sizeof(unique));
+			++uid_instance; // Now it's larger than zero.
+			}
+		else if ( zeek::Supervisor::ThisNode() )
```
awelzel, line 2172: This couples the node name / supervisor framework quite in here.

As your test shows, running a supervisor setup in deterministic mode `-D` with `pcap_file` (even using same pcap) would produce different UIDs depending on the worker name. I realize this is exactly what you want and that it can make sense and seems convenient. However, changing `worker-1` to `worker-01` also results in different connection UIDs - that on the other hand seems surprising. Some indirection is probably warranted here.

First thought would be whether using per-worker/per-node `ZEEK_SEED_FILE` is a solution today? Certainly not as convenient and direct, but seems the mechanism to achieve this is there and workable.
Could also make `-G|--load-seeds` available in the supervisor Node config, but that's probably neither more nor less convenient.

@ckreibich ?

If something direct is needed/wanted still some preference for a new supervisor independent environment variable `ZEEK_RAND|SEED_EXTRA|SUFFIX|ADDL`  for indirection (I'm not sure about the name). This can be set in the runtime environment of the process to achieve the "individual determinism". This would provide the same convenience for non-supervisor setups/users as well.

(Ignore the nits below, I think whether this is more important)

J-Gras: > However, changing worker-1 to worker-01 also results in different connection UIDs - that on the other hand seems surprising.

I can't argue with that. It's intended but probably surprising.

> If something direct is needed/wanted still some preference for a new supervisor independent environment variable ZEEK_RAND|SEED_EXTRA|SUFFIX|ADDL (edit: maybe ZEEK_DETERMINISTIC_UID_SEED?) for indirection (I'm not sure about the name).

I would prefer the direct approach. Doesn't feel clean to create seed files for every node from a Zeek script just to start a cluster with independent nodes.

J-Gras: I was thinking along the same lines. Small idea for convenience: How about adding the possibility to specify just one value and use this for all seeds. Given that `-D` just sets everything to zero, I would assume this to be okay. So one could either specify `ZEEK_SEED=12 12 12 ... 12` or just `ZEEK_SEED=12`.

J-Gras:
Closing this in favour of #3038.

PR #3036, opened 2023-05-10
Add DTLS 1.3 support

0xxon:
This PR adds support for DTLS 1.3.

This mainly includes parsing the new DTLS 1.3 unified header (which causes a parse error when parsed with current versions of Zeek), and adding support for the connection_id extension.

This PR also fixes the logging of hello retry requests in the `ssl_history` field of ssl.log - where they so far were incorrectly logged as server hello messages.

It would be nice for this to make it into the next Zeek release.


src/analyzer/protocol/ssl/dtls-protocol.pac
```diff
@@ -4,11 +4,46 @@
...
+type UnifiedRecord(firstbyte: uint8, is_orig: bool) = record {
+	# sequence_number: bytestring &length=(sequence_number_length?2:1);
+	# If we have a CID, we do currently not try to parse anything, as the connection
+	# ID is variable length, with the length not given in this packet (but only in the hello message
+	# of the opposite side of the direction).
+	seqnum: case with_cid of {
+		false -> sequence_number: bytestring &length=(sequence_number_length?2:1);
+		true -> nothing1: bytestring &length=0;
+	} &requires(sequence_number_length) &requires(with_cid);
+	lenthfield: case (with_cid == false && length_present == true) of {
```
timwoj, line 34:
```suggestion
	lengthfield: case (with_cid == false && length_present == true) of {
```

src/analyzer/protocol/ssl/events.bif
```diff
@@ -589,10 +627,11 @@ event ssl_plaintext_data%(c: connection, is_client: bool, record_version: count,
 ## content_type: message type as reported by TLS session layer. Not populated for
 ##               SSLv2.
 ##
-## length: length of the entire message.
+## length: encrypted payload length of the message.
```
timwoj, line 630: This doesn't sound right to me. Maybe this?

```suggestion
## length: length of the encrypted payload in the message.
```

timwoj (approved):
Looks good to me. Just a couple of nits.

src/analyzer/protocol/ssl/dtls-protocol.pac
```diff
@@ -4,11 +4,46 @@
...
+	firstbyte: uint8;
+
+	cont: case $context.connection.choose_record_type(firstbyte) of {
+		false -> rec: SSLRecord(firstbyte, is_orig);
+		true -> unified: UnifiedRecord(firstbyte, is_orig);
+	};
+};
+
+type UnifiedRecord(firstbyte: uint8, is_orig: bool) = record {
+	# sequence_number: bytestring &length=(sequence_number_length?2:1);
```
awelzel, line 26: Nit. Is this needed here or could the comment go?

src/analyzer/protocol/ssl/dtls-protocol.pac
```diff
@@ -83,4 +118,11 @@ refine connection SSL_Conn += {
 		}
 		%}
 
+		function choose_record_type(firstbyte: uint8): bool
+			%{
+			uint16_t negotiated_version = zeek_analyzer()->GetNegotiatedVersion();
+			if ( negotiated_version == DTLSv13 && ( (firstbyte & 0x20) == 0x20 ) )
```
awelzel, line 124: This is testing the "first three high bits" being `001`, correct? Seems it needs to `&` with a mask of the first 3 bits `0xE0` and compare that result with `0x20` thereby asserting that the first two bits are actually `0`?
```suggestion
			return negotiated_version == DTLSv13 && ( (firstbyte & 0xE0) == 0x20 );
```

With the existing test `111` would still evaluate to true as the first 2 bits are ignored out.

Other nit, given this returns `bool` if the name was `is_unified_record()` that would help to know what the meaning is without reading a comment or the code.

awelzel (approved):
There's a bug with the `001` bit matching I think. Approving as after a fix seems good to go, :+1:

src/analyzer/protocol/ssl/events.bif
0xxon: That makes more sense. I probably will make it "length of the encrypted payload in the record" - which is even more correct, if that is ok with you :)

src/analyzer/protocol/ssl/dtls-protocol.pac
0xxon: the comment could go, thanks

src/analyzer/protocol/ssl/dtls-protocol.pac
0xxon: you are right, I am not entirely sure what I was thinking here. Thank you for noticing that.

Issue #3028, opened 2023-05-08
Connection flipping recreates c$id, removing extra fields


JustinAzoff:
This definitely may fall under "Don't do that", but this is certainly odd behavior:

```
module IDFlipBug;

redef record conn_id += {
	my_field: string &log &optional;
};

event new_connection(c: connection) &priority=4
	{
	c$id$my_field = "Hello";
	print "Adding my_field to", c$id;
	}
```

When ran on a pcap with the initial syn packet removed:

```
$ zeek -Cr  ~/work/pcaps/http_single_no_syn_high_port.pcap ./bug.zeek tuning/json-logs                                            
Adding my_field to, [orig_h=192.168.2.127, orig_p=28080/tcp, resp_h=192.168.2.185, resp_p=49420/tcp, my_field=Hello]
$ 
$ cat conn.log|jq .                                                                    
{
  "ts": 1620657542.468,
  "uid": "CvjXtL1vqLsPQGjMd2",
  "id.orig_h": "192.168.2.185",
  "id.orig_p": 49420,
  "id.resp_h": "192.168.2.127",
  "id.resp_p": 28080,
  "proto": "tcp",
  "duration": 0.017675161361694336,
  "orig_bytes": 87,
  "resp_bytes": 118,
  "conn_state": "SF",
  "missed_bytes": 0,
  "history": "^hADadfF",
  "orig_pkts": 5,
  "orig_ip_bytes": 355,
  "resp_pkts": 5,
  "resp_ip_bytes": 378
}
```

It appears that when the internals flip the connection - which I also notice doesn't have a `connection_flipped` type event, the `id` record is recreated from scratch, rather than being updated in place.  This removes any custom fields that may have been added to the record.

awelzel:
> This definitely may fall under "Don't do that", but this is certainly odd behavior:

This applies to the whole connection value, not just c$id. The new community-id-logging policy script is actually affected by this:

https://github.com/zeek/zeek/blob/c543387ce0d4949ed82e7c6f3dbfb9cd13dc562b/scripts/policy/protocols/conn/community-id-logging.zeek#L22-L26

> It appears that when the internals flip the connection - which I also notice doesn't have a connection_flipped type event, the id record is recreated from scratch, rather than being updated in place.

[...]

PR #3037, opened 2023-05-10
policy/community-id: Do not use new_connection()

awelzel:
Issue #3028 tracks how a flipped connections reset a connection's value including any state set during new_connection(). For the time being, update community-id functionality back to the original connection_state_remove() approach to avoid missing community_ids on flipped connections.

External test suite PRs:
https://github.com/zeek/zeek-testing-private/pull/15
https://github.com/zeek/zeek-testing/pull/13


ckreibich (approved):
The right fix for now, agree...

PR #3038, opened 2023-05-10
Introduce ZEEK_SEED_VALUES environment variable

awelzel:
For "individually different but deterministic" processes, specifying Zeek's seed as an environment variable eases setups as one can avoid creating extra files for each of the individual processes and instead inject the seeds as environment variable directly.

It is an error to specify the new ZEEK_SEED with the existing ZEEK_SEED_FILE environment variable together. There doesn't seem to be a clear preference for either. ZEEK_SEED takes precedence over deterministic (-D) mode like ZEEK_SEED_FILE does today already.


awelzel:
Alternative for #3034.

src/util.cc
```diff
@@ -404,19 +396,35 @@ void seed_random(unsigned int seed)
 		srandom(seed);
 	}
 
-void init_random_seed(const char* read_file, const char* write_file, bool use_empty_seeds)
+void init_random_seed(const char* read_file, const char* write_file, bool use_empty_seeds,
+                      std::string seed_string)
```
timwoj, line 400: Either pass this as const std::string& or turn it into a std::string_view.

src/util.cc
```diff
@@ -404,19 +396,35 @@ void seed_random(unsigned int seed)
...
+		}
+	else if ( ! seed_string.empty() )
+		{
+		if ( ! (seed_file = fmemopen(seed_string.data(), seed_string.size(), "r")) )
+			reporter->FatalError("Could not create seed file from string: %s", strerror(errno));
+		}
+
+	if ( seed_file )
+		{
+		// fclose() in read_random_seeds()...
```
timwoj, line 422: It seems more intuitive to move the `fclose` out here since this is where it's opened. You could wrap it in a `unique_ptr` so that it gets destructed properly on error, if you don't want to add two calls to `fclose`.

src/zeek-setup.cc
```diff
@@ -629,15 +629,22 @@ SetupResult setup(int argc, char** argv, Options* zopts)
 		supervisor_mgr = new Supervisor(std::move(cfg), std::move(*stem));
 		}
 
+	std::string seed_string;
+	if ( auto* seed_env = getenv("ZEEK_SEED") )
+		seed_string = seed_env;
+
 	const char* seed_load_file = getenv("ZEEK_SEED_FILE");
 
+	if ( seed_load_file && *seed_load_file && ! seed_string.empty())
```
timwoj, line 638: Does this need to take `-G` into account?

timwoj (changes requested):

timwoj:
Do we document the various environment variables anywhere? I don't see anything in zeek-docs for ZEEK_SEED_FILE, so maybe we don't?

src/util.cc
awelzel: Thanks - fmemopen wants a `void *` so either of those would require a `const_cast` then.

I see that `fmemopen()` isn't available on window, so I'll see about switching this to input streams instead.

src/zeek-setup.cc
awelzel: Yep, I can see that - I thought `-G` would just take precedence, but I like the conflict better.

J-Gras:
Just for the record, I would love to have the convenience feature of being able to specify `ZEEK_SEED="0"` instead of spelling out `ZEEK_SEED="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0"` 😉

src/util.cc
J-Gras: Just two random thoughts: 1) The function has become quite cluttered. How about separating the file-related stuff and the seed generation (e.g., by extracting the latter to its own function). 2) Maybe just setting the appropriate `seed_string` to get rid of the `use_empty_seeds` special case would improve readability.

awelzel:
> Just for the record, I would love to have the convenience feature of being able to specify `ZEEK_SEED="0"` instead of spelling out `ZEEK_SEED="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0"` wink

I'm actually forth and back renaming the whole thing to `ZEEK_SEED_VALUES` to make explicit it needs to have those 21 values. There's also a "seed" variable used to initialize only srandom() (the very first seed in the list). Accepting just a singular value could mean just setting that and doing "something" for the remaining 20 (maybe 0-initialization would be more intuitive).

So, from my side, this is not worth the extra documentation overhead and potential confusion that comes with the magic.

src/util.cc
awelzel: I can see, but I also didn't want to re-structure all that much in code where my IDE shows half the things ifdef'ed out :fearful:  While `init_random_seed()` is in zeek::detail, not breaking API compat if we don't need to was a bit in mind, too. Unless Tim thinks this has now crossed some threshold I'll keep it as is for now.

awelzel:
Closing and re-opening without stream usage after chatting with @timwoj .

PR #3040, opened 2023-05-11
remote_event_ts_compat: Fix NETWORK_TIMESTAMP to NetworkTimestamp

awelzel:
~The "check" was only meant to prevent going into the background running mode if things aren't looking good, but TEST-REQUIRES skips the test when that was the case.~


awelzel:
Pushed directly into master.

PR #3042, opened 2023-05-12
Introduce ZEEK_SEED_VALUES environment variable

awelzel:
For "individually different but deterministic" runs specifying Zeek's seed as an environment variable eases setups as one can avoid creating extra seed files for each of the individual processes.

It is an error to specify the new ZEEK_SEED_VALUES variable together with the existing ZEEK_SEED_FILE and -G. ZEEK_SEED takes precedence over deterministic mode (-D) like ZEEK_SEED_FILE does today already.


timwoj (approved):
Yeah, this looks better with it split out.

J-Gras:
Linking #3038 for reference.

PR #3047, opened 2023-05-16
Fix `get_active_node_count` for node types not present.

J-Gras:
Tiny fix to prevent `no such index (Cluster::active_node_ids[Cluster::node_type])` errors when calling `get_active_node_count()` for node types not present in the cluster.

### DIFF
--- a/scripts/base/frameworks/cluster/main.zeek
+++ b/scripts/base/frameworks/cluster/main.zeek
@@ -285,7 +285,7 @@ export {
 }
 
 # Track active nodes per type.
-global active_node_ids: table[NodeType] of set[string];
+global active_node_ids: table[NodeType] of set[string] &default=set();
 
 function nodes_with_type(node_type: NodeType): vector of NamedNode
 	{


scripts/base/frameworks/cluster/main.zeek
```diff
@@ -285,7 +285,7 @@ export {
 }
 
 # Track active nodes per type.
-global active_node_ids: table[NodeType] of set[string];
+global active_node_ids: table[NodeType] of set[string] &default=set();
```
awelzel, line 288: Oh. How would you feel about checking this `get_active_node_count()` explicitly instead? We only need the `0`, but this creates a new set on every missed lookup.  Re-using get_active_node_count() for worker_count avoids having to do that in multiple places.

I'm happy to fold it in.

```
--- a/scripts/base/frameworks/cluster/main.zeek
+++ b/scripts/base/frameworks/cluster/main.zeek
@@ -318,7 +318,7 @@ function Cluster::get_node_count(node_type: NodeType): count
 
 function Cluster::get_active_node_count(node_type: NodeType): count
        {
-       return |active_node_ids[node_type]|;
+       return node_type in active_node_ids ? |active_node_ids[node_type]| : 0;
        }
 
 function is_enabled(): bool
@@ -377,7 +377,7 @@ event Cluster::hello(name: string, id: string) &priority=10
 
        if ( n$node_type == WORKER )
-               worker_count = |active_node_ids[WORKER]|;
+               worker_count = get_active_node_count(WORKER);
 @pragma pop ignore-deprecations
        }
 
@@ -402,7 +402,7 @@ event Broker::peer_lost(endpoint: Broker::EndpointInfo, msg: string) &priority=1
 
                        if ( n$node_type == WORKER )
-                               worker_count = |active_node_ids[WORKER]|;
+                               worker_count = get_active_node_count(WORKER);
 @pragma pop ignore-deprecations

```

J-Gras: > We only need the 0, but this creates a new set on every missed lookup.

Oh, I missed that. Thought the default value will be created only once and stored unless `&default` uses a function. Feel free to switch to the explicit check.

Issue #2683, opened 2023-01-12
DataBlock::DataBlock: Heap-buffer overflow with overly large packets


awelzel:
Replaying the following pcap through a veth interface (fuzzin/fuzzout below) and listening on fuzzout with the standard libpcap source triggers the following heap-buffer-overflow. 

The pcap contains one packet with a length >14kbytes due to gro I suspect.

Running `zeek -r` on it works fine. Using `af_packet::fuzzout0` also doesn't trigger the crash, so seems something isolated to replay and libpcap.

The captured traffic is coming from reading a host interface (with gro/offloading enabled) then encapsulating the raw data with Geneve to forward it to a Zeek setup reachable on 127.0.0.1:6081. I just happened to have another Zeek listening on the local `lo` interface and observed the crash. I'm fairly certain this is mostly triggered by gro and setting overly large MTUs. it doesn't seem to trigger when stripping the Geneve encapsulation though.

[geneve-47101.pcap.zip](https://github.com/zeek/zeek/files/10400704/geneve-47101.pcap.zip)

````
# ip link add fuzzin0 type veth peer fuzzout0
# ip link set fuzzin0 mtu 20000
# ip link set fuzzout0 mtu 20000
# ip link set fuzzin0 up
# ip link set fuzzout0 up
# zeek -C -i fuzzout0 -f 'udp port (4789 or 6081)' LogAscii::use_json=T
<params>, line 1: listening on fuzzout
...
# tcpreplay -l 0 -i fuzzin0 ./geneve-47101.pcap

=================================================================
==4087843==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62500025a500 at pc 0x7f45f8424983 bp 0x7ffea20d9f90 sp 0x7ffea20d9740
READ of size 14080 at 0x62500025a500 thread T0
    #0 0x7f45f8424982 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:806
    #1 0x563c83a32ddb in zeek::DataBlock::DataBlock(unsigned char const*, unsigned long, unsigned long) ../src/Reassem.cc:24
    #2 0x563c83a35bb4 in zeek::DataBlockList::Insert(unsigned long, unsigned long, unsigned char const*, std::_Rb_tree_const_iterator<std::pair<unsigned long const, zeek::DataBlock> >) ../src/Reassem.cc:109
[...]

PR #3049, opened 2023-05-16
Fix issues with truncated packets inside tunnels inside tunnels

### LINKED ISSUE #2914: IPTunnel::build_inner_packet: Not keeping wire lengths intact
Stumbled over in relation to #2907 and #2683.

When building an inner packet, currently `packet->len`, the "wire length", is set to the same as the available "capture length":

https://github.com/zeek/zeek/blob/5718046b967e22ff375c55402e9a877163565719/src/packet_analysis/protocol/iptunnel/IPTunnel.cc#L185

That may not always be right. If the packet was truncated, the wire length of an inner packet should be computed based on the wire length of a more outer packet.

E.g., when analyzing a packet `eth-ip-udp-geneve-eth-ip-tcp-[snaplen-truncated]`, the wire length of the inner IP packet is currently set to the caplen left at that point, rather than to the "wire length" of the outer UDP packet minus the length of the Genve header.

The effect is that the IP analyzer finds the inner IP header invalid due to its header length disagreeing with the packet's wire length, creates a weird and stops. No further packet/protocol analysis is done and have no visibility into further protocols.

timwoj:
This PR reverts some earlier changes around handling truncated TCP and UDP packets inside tunnels-of-tunnels, and reimplements the handling of those packets in a different way. There's a few commits here that are general cleanup as well.

Fixes #2914


timwoj:
I'll take a look at that test failure tomorrow.

src/packet_analysis/protocol/udp/UDP.cc
```diff
@@ -223,11 +223,17 @@ void UDPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int remai
 	// port here because the orig/resp should have already swapped around based on
 	// likely_server_ports. This also prevents us from processing things twice if protocol
 	// detection has to be used.
-	ForwardPacket(std::min(len, remaining), data, pkt, ntohs(c->RespPort()));
```
awelzel, line 226: Same as above, think the `min` would need to stay as`len` can't be trusted.

src/packet_analysis/protocol/udp/UDP.cc
```diff
@@ -223,11 +223,17 @@ void UDPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int remai
...
+	ForwardPacket(len, data, pkt, ntohs(c->RespPort()));
 
 	// Also try sending it into session analysis.
 	if ( remaining >= len )
+		{
+		// Adjust the remaining by the difference between len and caplen in the packet, since
+		// we might not actually have all of the data available.
```
awelzel, line 232: But at the same time this means `remaining` was already out of sync with how much is left in `data` which seems very dangerous. It would be much easier if `remaining` is always guaranteed to be `<= bytes left in data`, or else seems this special case would also be required in TCP in case there were TCP in TCP tunnels?

src/packet_analysis/protocol/iptunnel/IPTunnel.cc
```diff
@@ -178,11 +178,16 @@ std::unique_ptr<Packet> build_inner_packet(Packet* outer_pkt, int* encap_index,
 	{
 	auto inner_pkt = std::make_unique<Packet>();
 
+	// Use the original packet's caplen difference to calculate a new caplen for the inner
+	// packet. Previously we just used the len for both values but that's likely not correct
+	// in most cases.
+	uint32_t cap_len = len - (outer_pkt->len - outer_pkt->cap_len);
```
awelzel, line 184: Hmm... maybe this is the crux: I think we should instead (or also) update the wire length of the inner packet (and use where appropriate) instead of having `ForwardPacket's `len` be the wire length.

awelzel (commented):
Hmm, I'm worried that's not quite it. The removal of std::min really makes me nervous as that brings us back to passing untrusted data through `ForwardPacket()`.

I left some comments, hope they help.

src/packet_analysis/protocol/tcp/TCP.cc
```diff
@@ -125,7 +125,7 @@ void TCPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int remai
 	pkt->session = c;
 
 	// Send the packet back into the packet analysis framework.
-	ForwardPacket(std::min(len, remaining), data, pkt);
+	ForwardPacket(len, data, pkt);
```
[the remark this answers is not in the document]
timwoj: The reason this breaks is because we take the `len` value passed into `GeneveAnalyzer::AnalyzePacket` and use it as the length value in the inner packet object. When it gets down in `IPAnalyzer::AnalyzePacket` it tries to compare that length value to the length in the actual IP header, sees that the one in the packet is too low, generates a weird, and bails out of processing any of the rest of the inner packet.

After reverting the above change, for example, in the first packet in geneve-vxlan-dns-truncated.pcap, `packet->len` ends up as 126 where the IP header says it should be 132 (which matches the len vs remaining difference).

We could fix `IPTunnel::build_inner_packet` to do some detection on whether it's IP and set the length based on that etc etc, but I think we'd be right back where we were with lying about how much data is actually remaining again.

src/packet_analysis/protocol/udp/UDP.cc
timwoj: It's probably the latter, that TCP also needs the change. It only came up in UDP here because that's the test cases we have.

timwoj:
The asan finding is because we're not checking the length/remaining values in PIA. Fixing it to do so causes the core.tunnels.geneve-truncated test to fail for the reasons discussed in the issue here: https://github.com/zeek/zeek/issues/2914#issuecomment-1531307240

src/packet_analysis/protocol/iptunnel/IPTunnel.cc
awelzel: See branch linked above. Resolving.

src/packet_analysis/protocol/tcp/TCP.cc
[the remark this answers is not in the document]
timwoj: Yeah, that looks good to me. Should I close this PR and open one against that? @ckreibich will have to approve it since we've both touched this one now.

Does this fix the asan crash with the other pcap?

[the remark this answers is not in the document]
awelzel: > Does this fix the asan crash with the other pcap?

You mean the one from the following issue?

https://github.com/zeek/zeek/issues/2683 

Yep, seems fine. I'm adding it as a btest. I'll rebase and create a new PR. I'm going to squash the fixup commit.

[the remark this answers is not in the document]
timwoj: Nah I meant the geneve-truncated pcap that Cirrus was having problems with on this branch. Looks like it's fine with the other branch.

PR #3062, opened 2023-05-19
"&analyze" attribute for @if conditionals

vpax:
This PR replaces https://github.com/zeek/zeek/pull/2985, reflecting the discussion there (and more extensively at https://github.com/zeek/zeek-docs/pull/181) to use an attribute for this type of conditional rather than a new directive. This PR also incorporates the coding review comments from @awelzel.

I'll be submitting a corresponding documentation PR shortly.


scripts/base/frameworks/cluster/main.zeek
```diff
@@ -347,7 +347,7 @@ function nodeid_topic(id: string): string
 	return nodeid_topic_prefix + id + "/";
 	}
 
-@if ( Cluster::is_enabled() )
+@if ( Cluster::is_enabled() ) &analyze
```
awelzel, line 350: Still think this `@if` should be dropped again and runtime checks are fine.

@J-Gras , do you remember why it was added in  3db8bb4a44e39ca87699060ee1317a871c6342b9?

The one reason I could see is a left-over from not being able to suppress deprecation warnings and putting it in place as a way to suppress them (because many tests don't run with Cluster::is_enabled()), but otherwise it's not obvious to me why this came in.

scripts/base/protocols/irc/dcc-send.zeek
```diff
@@ -44,26 +44,29 @@ function dcc_relay_topic(): string &is_used
 	return rval;
 	}
 
+const have_cluster = Cluster::is_enabled();
+const should_publish =
```
awelzel, line 48: I stumbled a bit over this pattern that's used elsewhere, no need to address because it's again in the naming and subjective area, but wanted to leave the thoughts.

* Are you worried that a `Cluster::is_enabled()` call is too expensive, and/or that caching seems a reasonable optimization?

* "have cluster" and "Cluster::is_enabled()", is close, but it's still different enough for me to go back the first few times to ensure it's just Cluster::is_enabled(). Is `const cluster_is_enabled =` too long? We could probably also cache a `Cluster::is_enabled_value` const?

* About "should_publish". Workers do call Broker::publish() just as well and seems if it's a global all Broker::publish() should be protected with it. Proxies / managers responsibility is to forward/broadcast/relay messages to workers, which I think this is representing? should_forward / should_relay?

src/zeek-setup.cc
```diff
@@ -183,6 +184,7 @@ zeek::plugin::Manager* zeek::plugin_mgr = nullptr;
 zeek::detail::RuleMatcher* zeek::detail::rule_matcher = nullptr;
 zeek::detail::DNS_Mgr* zeek::detail::dns_mgr = nullptr;
 zeek::detail::TimerMgr* zeek::detail::timer_mgr = nullptr;
+zeek::detail::ActivationManager* zeek::detail::activation_mgr;
```
timwoj, line 187: `= nullptr`

timwoj (changes requested):
Looks pretty straight-forward to me. I just had some general code quality comments, nothing major.

scripts/base/frameworks/cluster/main.zeek
J-Gras: > do you remember why it was added in https://github.com/zeek/zeek/commit/3db8bb4a44e39ca87699060ee1317a871c6342b9?

I think the intention was just to make things a little cleaner by skipping event handlers that won't be used anyway. Seems `Broker::peer_added` was (and is 🙄 ) the only handler that checks for `Cluster::is_enabled()`. For example, `Log::create_stream` is called unconditionally from `zeek_init`.

ckreibich (commented):
I've tinkered with this a bit locally and it all seems well.

As I was looking at all the `@if` -> `if` cases I was thinking it'd be nice to have some sort of mechanism to ensure that a condition's evaluation cannot change at runtime, both since it'd ensure we'd get the same effect as `@if`, and also because I'd imagine it could assist optimizations. Something like an explicitly cacheable predicate function. The added `const`s get close to that of course, they're just manual.

scripts/base/frameworks/cluster/main.zeek
awelzel: > However I agree that we shouldn't unconditionally create the log stream if it's clear it won't actually be used.  It'd be easy to adapt the zeek_init handler in that way.

Let me chime in here with a long-winded opinion that we should create that stream - this is much unrelated to "@if &analyze", sorry Vern.

* If we can avoid any ifs (static or dynamic) around cluster, IMO things are simpler as there's just one scenario to think about. Adding `@if` or `if` does not make things simpler.
* There won't actually be a cluster.log file created as long as no one `Log::write()`'s to it (I think not event a writer backend instantiated), so the stream shouldn't cost in the non-cluster mode.
* If you do get a `cluster.log` running in a non-cluster `zeek -r` setup, that's curious. A Log::write() happened that maybe shouldn't have. I think you do want to see it.
* But, if we don't create the log stream, a ``Log::write()`` to it is a silent dropped log message. Not great, would rather have that surprising cluster.log in non-cluster mode show up.
* Specifically because there's this ``Cluster::log()`` helper. Likely that's only used when `Cluster::is_enabled() == T`, but it's not obvious that it now won't work in non-cluster mode. ``Cluster::create_master()`` is particularly funky as it caters for the non-cluster case, too, but also has calls to `Cluster::log()`, but in the "right places".

* There's also at least one known schema generator that iterates over all active log streams and appears to have a non-cluster mode (not sure it's used). If we don't create the Cluster::LOG stream, we're hiding it form that tool when used as `zeek generate-log-schema.zeek` (does that fall under API/interface stability/regression? :-) )

So all in all, I think that stream being there even in a non-cluster setup has almost no cost, voids having that extra `@if` / `if`, avoids surprises also has been there in Zeek 5.2.

ckreibich: > Adding @if or if does not make things simpler.

Sure, but the cluster code is always loaded, but people aren't always running a cluster. I was thinking about this more earlier — I think the whole structure of that framework is a bit off. It should separate config state that needs to be there for any setup to reason about (as in, are we running a cluster or not?), and separate stuff that only matters when a cluster is actually running (like, a cluster log).

> If we don't create the Cluster::LOG stream, we're hiding it form that tool

Yes, I actually thought that's a plus. 

> and also has been there in Zeek 5.2 and prevoiusly.

Isn't the `Log::create_stream` currently guarded by `Cluster::is_enabled()`?

Anyway — not a hill I'm prepared to die on and I've spent all day reviewing PRs. You have my opinion. :-)

vpax: So I'm not sure I could follow all of that well enough to know whether I need to make a change to this facet of the PR. For now, I'll leave it as I originally* submitted it, we can circle back if more refinements needed. (* = well this whole PR is itself not "original", but you know what I mean ... :P)

scripts/base/protocols/irc/dcc-send.zeek
vpax: Happy to go with `cluster_is_enabled` (and now added to my pending update). Even though in this case it's trivial, I think of the base scripts as modeling patterns that script writers will often follow, and the pattern of "if you use the conditional more than once, put it in a `const` so it's immediately apparent that the two uses are the same" is a good one.

Regarding `should_publish`, @awelzel I take your comment to be more about is-the-current(pre-PR)-structure-in-fact-right. As the PR isn't about cluster structural logic but just about introducing the new feature where it's apt, I'm thinking I won't bite off the broader notion of just what `should_publish` should cover.

src/ActivationManager.h
```diff
@@ -0,0 +1,311 @@
...
+namespace zeek::detail
+	{
+
+using ScriptFuncPtr = IntrusivePtr<ScriptFunc>;
+using AttrVec = std::unique_ptr<std::vector<AttrPtr>>;
+
+class ActivationManager;
+
+/**
+ * Expresses an event (with one of the above types) occurring during an
```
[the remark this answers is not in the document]
vpax: Yep, thanks for flagging.  Fixed.

src/ActivationManager.h
```diff
@@ -0,0 +1,311 @@
...
+ * up with the conditional for the live parsing, as opposed to what *could*
+ * have happened (which is instead reflected in a set of ActivationEvent's).
+ */
+class Activation
+	{
+public:
+	Activation(ExprPtr cond, bool _is_activated, bool _parent_activated, int _cond_depth);
+	~Activation();
+
+	// True if we're in the part of the @activiate-if conditional for
```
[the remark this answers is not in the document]
vpax: Hah the typo let it elude my epic `grep -r` to squash all such instances. Thanks for catching!

src/parse.y
```diff
@@ -5,10 +5,10 @@
 // Switching parser table type fixes ambiguity problems.
 %define lr.type ielr
 
-%expect 211
+%expect 212
 
 %token TOK_ADD TOK_ADD_TO TOK_ADDR TOK_ANY
-%token TOK_ATENDIF TOK_ATELSE TOK_ATIF TOK_ATIFDEF TOK_ATIFNDEF
+%token TOK_ATENDIF TOK_ATELSE TOK_ATIF TOK_ATACTIVATEIF TOK_ATIFDEF TOK_ATIFNDEF
```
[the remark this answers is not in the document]
vpax: Yep, and the production for it below. Fixed.

src/scan.l
```diff
@@ -1000,6 +1039,13 @@ int yywrap()
 		entry_cond_depth.pop_back();
 		}
 
+	if ( entry_act_depth.size() > 0 )
```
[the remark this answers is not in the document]
vpax: Yes ... but I'm just following the pattern above and below it, so I'll leave this one be.

src/zeek-setup.cc
vpax: Rut-row!  A holdover from when it was a `unique_ptr` not requiring initialization.  Thanks for catching!

src/ActivationManager.h
```diff
@@ -0,0 +1,311 @@
...
+	// Adds a "subevent" to this event, only valid for events that are
+	// themselves conditionals.  Note that the subevent might itself
+	// be a (nested) conditional.
+	void AddSubEvent(std::shared_ptr<ActivationEvent> ae)
+		{
+		ASSERT(et == COND);
+		CurrSubEvents().push_back(std::move(ae));
+		}
+
+	// Changes this events accrural of subevents to correspond to its
```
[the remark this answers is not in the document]
vpax: Fixed.

src/ActivationManager.cc
```diff
@@ -0,0 +1,199 @@
+#include "zeek/ActivationManager.h"
+
+#include "zeek/Desc.h"
+
+using namespace std;
+
+namespace zeek::detail
+	{
+
+void ActivationEvent::Dump(int indent_level) const
```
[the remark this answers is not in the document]
vpax: Conceptually, I guess it does.  However I'd rather wait on doing so when we find that we actually need it. I created the method just to debug whether the events were being correctly characterized; indeed I could just remove it at this point, though it might still prove handy for debugging the C++ script compiler's support for conditionals when I get around to that ...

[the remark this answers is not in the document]
vpax: (in particular, would rather wait because it's non-trivial work to convert it)

src/ActivationManager.h
```diff
@@ -0,0 +1,311 @@
...
+		{
+		return activation_stack.empty() || activation_stack.back()->IsActivated();
+		}
+
+	// Returns the nesting level of @if &analyze's.  Used by the scanner
+	// to find dangling conditionals at the end of files.  Typed as an
+	// integer to match similar other structures in the scanner.
+	int ActivationDepth() const { return static_cast<int>(activation_stack.size()); }
+
+	// Tells the manager to being a new @if &analyze conditional.
```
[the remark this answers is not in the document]
vpax: fixed

scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -96,6 +96,10 @@ event Known::host_found(info: HostsInfo)
 		}
 	}
 
+const should_log_host =
```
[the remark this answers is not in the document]
vpax: While it's single use, I find it adds clarity to what's otherwise a bit of a visual mouthful (to mix metaphors :P):

```  
       ! Cluster::is_enabled() ||
        Cluster::local_node_type() == Cluster::PROXY;
```

testing/btest/language/at-if-analyze-invalid.zeek
```diff
@@ -0,0 +1,92 @@
+# @TEST-EXEC: cat %INPUT
```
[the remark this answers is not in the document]
vpax: good thought, now done

vpax:
@ckreibich @timwoj @awelzel  The latest push aims to resolve the pending issues, modulo those that I'd opt to defer per the above comment threads.  See watchya thunk.

scripts/base/frameworks/cluster/main.zeek
awelzel: > ... know whether I need to make a change to this facet of the PR.

Can leave as is. IMO removing the `@if`'s preferred, but resolving as that can be a follow-up or they'll just linger around.

scripts/base/protocols/irc/dcc-send.zeek
awelzel: > pattern of "if you use the conditional more than once, put it in a const so it's immediately apparent that the two uses are the same" is a good one.

 :+1: 

> @awelzel I take your comment to be more about is-the-current(pre-PR)-structure-in-fact-right.

Nah, I meant that the "should_publish" const name is a too generic for it's actual purpose. It's scoped small enough, not a module API and a naming topic, so I'm resolving this discussion.

timwoj (approved):

PR #3066, opened 2023-05-22
Revert part of previous CMake debug type fix

timwoj:
This PR reverts part of https://github.com/zeek/zeek/pull/3058. CI builds started failing after it was merged. After a bit of digging I realized that it was doing the wrong thing if you passed an all-lowercase string to `--build-type`, since the check added in that PR was looking specifically for `Debug`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,8 +11,12 @@ project(Zeek C CXX)
 
 # We want to set ENABLE_DEBUG to ON by default if the build type is Debug.
 set(ENABLE_DEBUG_DEFAULT OFF)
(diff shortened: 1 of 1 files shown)


ckreibich:
Thanks Tim. This stuff is so brittle, wow. Fwiw, we have another `CMAKE_BUILD_TYPE` check [further down](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L104-L110) that might also break if one passes `debug` as opposed to `Debug`.

ckreibich (approved):

timwoj:
> Fwiw, we have another CMAKE_BUILD_TYPE check [further down](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L104-L110) that might also break if one passes debug as opposed to Debug.

That one is in the Windows checks, which we don't manually set via user input anywhere.

PR #3072, opened 2023-05-23
Gracefully handle removal of analyzers without parent analyzer.

### LINKED ISSUE #3071: `disable_analyzer` can trigger an assertion failure/segfault if called on analyzer without parent
@JustinAzoff privately reported that the following triggered a segfault:

```zeek
	{
	local i = 0;
	while ( i < 200 )
		{
		print i, disable_analyzer(c$id, i, F);
		++i;
		}
	}
```

This removes e.g., `TCP` which has no parent analyzer and fails the assertion https://github.com/zeek/zeek/blob/614f1a9e5f8f10faf8019af31adea565774203cc/src/analyzer/Analyzer.cc#L462

This seems to be an issue in all supported versions.

bbannier:
Closes #3071.


bbannier (commented):
LGTM :shipit:.

Since I opened this PR I cannot approve it. Please squash this PR down to one commit on merge.

PR #3076, opened 2023-05-23
stats: Add zeek-net-packet-lag-seconds metric

awelzel:
While writing documentation about troubleshooting and looking a bit at the older stats.log, realized we don't have the packet lag metric exposed as metric/telemetry. Add it.

This is a Zeek instance lagging behind in network time ~6second because it's very overloaded:

    zeek_net_packet_lag_seconds{endpoint=""} 6.169406 1684848998092

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -161,6 +161,18 @@ New Functionality
 - Add logging metrics for streams (``zeek-log-stream-writes``) and writers
   (``zeek-log-writer-writes-total``).
 
+- Add networking metrics via the telemetry framework. These are enabled
+  when the ``misc/stats`` script is loaded.
+
+	zeek-net-dropped-packets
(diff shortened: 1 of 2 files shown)


timwoj (approved):

NEWS
```diff
@@ -161,6 +161,16 @@ New Functionality
 - Add logging metrics for streams (``zeek-log-stream-writes``) and writers
   (``zeek-log-writer-writes-total``).
 
+- Add networking metrics via the telemetry framework. These are enabled
+  when the ``misc/stats`` package is loaded and updated every 15 seconds
```
ckreibich, line 165: Total nits: I suggest "script" here instead of "package", and perhaps a reference to `Stats::Info` for context on where they come from.

Issue #3060, opened 2023-05-19
Support for negative vector indices


ynadji:
Many modern programming languages support negative index values for vectors for referring to entries from the end, i.e., `vec[-1]` is the last value, `vec[-2]` is the second-to-last value, and so on. Folks will expect this coming from these languages and it makes the code a bit more readable/succinct. For example:

```ruby
local stack: vector of count;
stack += 1;
stack += 2;
stack += 3;

# more readable (imo) equivalent to pop
local top = stack[-1];
stack = stack[0:-1];
```

etc.

awelzel:
> stack = stack[0:-1];

Slices with negative indices actually work, so not supporting negative indices even a bit inconsistent.

PR #3079, opened 2023-05-23
GH-3060: Support negative index lookups for vectors

timwoj:
This is a slight semantic change in that we don't support positive indexes up to the maximum size of an unsigned int, only to signed int. That should be fine though because we only support signed int values for slices as well.


awelzel (approved):
Cool. The 2**31 limit for vectors may be something to keep in mind, but today there's probably practical limitations of holding vectors that large (or larger).

PR #3089, opened 2023-05-26
Fix disappearing unit fields in Spicy type export.

### LINKED ISSUE #3083: Spicy unit types as Zeek record types fail at runtime
A user reported on Slack that using a Spicy type exported to Zeek fails at runtime. I was able to reduce this down to the following reproducer:

```ruby
# @TEST-REQUIRES: have-spicy
#
# @TEST-EXEC: spicyz -do export.hlto export.spicy export.evt >>output
# @TEST-EXEC: zeek -Cr $TRACES/http/pipelined-requests.trace export.hlto %INPUT
# @TEST-EXEC: btest-diff reporter.log

# @TEST-START-FILE export.spicy
module foo;

public type X = unit {
    x: uint8;
};
# @TEST-END-FILE

# @TEST-START-FILE export.evt
import foo;

protocol analyzer FOO over TCP:
    parse with foo::X,
    port 1/tcp-40000/tcp;

export foo::X;

on foo::X -> event foo::hello(self);
# @TEST-END-FILE

event foo::hello(x: foo::X)
	{ }
```

Running this I see the following in `reporter.log`:

```
1258577884.844956 analyzer error in <..>/spicy.export-types-usage/export.evt, line 9: Event parameter mismatch, no matching record field for field 'x'
```

Adding additional logging shows that `GlueCompiler::recordFields` sees zero unit items which seems wrong (would expect at least one for `x`, possibly more generated).

rsmmr:
When a Spicy unit type was used in an EVT event definition, it could
end up being cleared out at the time the logic for `export` got to see it.

Closes #3083.

Note: This needs backporting to the legacy plugin.


bbannier (approved):

bbannier:
@rsmmr, I added a patch to work around the failing CI. I filed #3094 to look into ways to really fix that issue (it was preexisting).

PR #3095, opened 2023-05-29
Simplify code generated for Spicy analyzer port ranges.

### LINKED ISSUE #3094: Failure when compiling Spicy analyzer which registers for a lot of ports
During #3089 we noticed that if a port analyzer registers a very broad port range, the analyzer compilation can fail in the C++ compiler with an internal error. The issue here seems to be that with compilers like gcc-11 having trouble compiling the generated `spicy_init` module which contains a constructor for a vector from 40000 `Port` values which are also constructed from integers.

One possible fix would be to teach spicy-plugin about port ranges so we can generate simpler code. This likely will also speed up compilation by a lot.

Here is the troublesome analyzer setup:

```ruby
# @TEST-REQUIRES: have-spicy
#
# @TEST-EXEC: spicyz -do export.hlto export.spicy export.evt
# @TEST-EXEC: zeek -Cr $TRACES/http/pipelined-requests.trace export.hlto %INPUT >>output
# @TEST-EXEC: test '!' -e reporter.log
#
# @TEST-DOC: Test type export end-to-end, with access from the Zeek side. Regression test for #3083.

# @TEST-START-FILE export.spicy
module foo;

public type X = unit {
    x: uint8;
};
# @TEST-END-FILE

# @TEST-START-FILE export.evt
import foo;

protocol analyzer FOO over TCP:
    parse with foo::X,
    port 1/tcp-40000/tcp;

export foo::X;

on foo::X -> event foo::hello(self);
# @TEST-END-FILE

event foo::hello(x: foo::X)
	{ print x; }
```

bbannier:
We previously would reprent port ranges from EVT files element-wise. This can potentially generate a lot of code (all on a single line though) which some versions of GCC seem to have trouble with, and which also causes JIT overhead.

With this patch we switch to directly representing ranges. Single ports are represented as ranges `[start, start]`.

Closes #3094.

### DIFF
--- a/scripts/spicy/zeek_rt.hlt
+++ b/scripts/spicy/zeek_rt.hlt
@@ -8,12 +8,15 @@ import hilti;
 public type Val = __library_type("::zeek::ValPtr");
 public type BroType = __library_type("::zeek::TypePtr");
 public type EventHandlerPtr = __library_type("::zeek::EventHandlerPtr");
+public type PortRange = __library_type("::zeek::spicy::rt::PortRange");
+
+declare public PortRange make_port_range(port begin_, port end_) &cxxname="zeek::spicy::rt::make_port_range" &have_prototype;
 
 type ZeekTypeTag = enum {
     Addr, Any, Bool, Count, Double, Enum, Error, File, Func, Int, Interval, List, Opaque, Pattern, Port, Record, String, Subnet, Table, Time, Type, Vector, Void
 } &cxxname="::zeek::spicy::rt::ZeekTypeTag";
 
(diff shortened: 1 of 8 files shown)


bbannier:
> Looks good, my only question is if we really need the new `PortRange` class & header, vs just typedefing that to a `std:pair`. But ok to merge this way if that's best.

I went with a dedicated class since that gives us better names for the fields as opposed to `std::pair`'s generic `first`/`second`. In any case, we probably still want a dedicated header since this is used in quite a number of places.

Issue #3093, opened 2023-05-26
Recent @if...&analyze change breaks zeekctl tests


timwoj:
See https://cirrus-ci.com/task/4808355248603136. The short gist is this:

```
[ #5] command.scripts ... failed
  % 'TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-timestamps-yyyy-mm-dd-hh-mm-ss btest-diff all.out' failed unexpectedly (exit code 1)
  % cat .diag
  == File ===============================
  manager scripts are ok.
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/cluster/./nodes/manager.zeek, line 22: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/cluster/./nodes/logger.zeek, line 67: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/cluster/./broker-stores.zeek, line 21: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/cluster/./broker-stores.zeek, line 25: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/intel/./cluster.zeek, line 20: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/intel/./cluster.zeek, line 24: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/intel/./cluster.zeek, line 76: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/intel/./cluster.zeek, line 97: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/sumstats/./cluster.zeek, line 62: @if &analyze inside conditional
    warning in /zeek/auxil/zeekctl/build/testing/test.25922/share/zeek/base/frameworks/sumstats/./cluster.zeek, line 210: @if &analyze inside conditional
[...]

PR #3097, opened 2023-05-29
at-if-analyze follow-up to avoid warnings

awelzel:
Proposal fixes for #3093 .

Seems like adding a test that  loads test-all-policies in cluster mode is something that should've been around for longer given that cluster mode is the actual important mode.


timwoj:
I cherry-picked in the first two commits from this (the test, and the fix for the test). I'm closing the rest of it since we have larger work to do with the &analyze feature.

PR #3102, opened 2023-06-05
Socks5 dpd sig fix

### LINKED ISSUE #3099: Issue with Zeek Socks DPD Signature
The Socks DPD signatures for v5 Socks do not appear to work correctly in Zeek.

The test case – socks-auth.pcap from the Zeek 5.2 distribution (in share/btest/data/pcaps):
Currently, this pcap will result in a Zeek Socks log because the traffic uses TCP port 1080. However, if this same traffic appeared on another port, a Zeek Socks log will not appear.
We tested this by creating test signatures that are the same as Zeek’s Socks DPD sigs, except that we removed “enable socks” form dpd_socks5_server and added “event <msg>” to both signatures so we know when they match. By doing this, we confirmed that the dpd_socks5_client signature matches, but the dpd_socks5_server signature does not match against this pcap. Examples of these test signatures are below.
 
The request is to fix the Socks DPD signature so that it matches typical socks traffic. Other references below.

References: Socks dpd.sig - https://github.com/zeek/zeek/blob/master/scripts/base/protocols/socks/dpd.sig (https://github.com/zeek/zeek/blob/master/scripts/base/protocols/socks/dpd.sig)

Test pcap file: /opt/zeek/share/btest/data/pcaps/socks-auth.pcap

 

Our version of the test signatures (ztest-event.sig):
signature ztest_dpd_socks5_client

```
{ ip-proto == tcp # Watch for a few authentication methods to reduce false positives. payload /^\x05.[\x00\x01\x02]/ tcp-state originator event "ztest_dpd_socks5_client (partial match)" }
 ```

signature ztest_dpd_socks5_server
```
{ ip-proto == tcp requires-reverse-signature ztest_dpd_socks5_client # Watch for a single authentication method to be chosen by the server or # the server to indicate the no authentication is required. payload /^\x05(\x00|\x01[\x00\x01\x02])/ tcp-state responder event "ztest_dpd_socks5_client + ztest_dpd_socks5_server (full match)" # enable "socks" }
 ```

Associated Zeek Script
```
@load-sigs ztest-event.sig

 

event zeek_init()

{ print "zeek_init(): Script is running!"; }
 

event zeek_done()

{ print "zeek_done(): Zeek is exiting!"; }
 

event signature_match(state: signature_state, msg: string, data: string)

{ print fmt("Signature Match: %s", msg); }
```

awelzel:
Not sure what `dpd_socks5_server` was meant to match, not quite what's in the RFC.

Closes #3099


awelzel:
@timwoj - potential 6.0 and 5.0 candidate.

PR #3103, opened 2023-06-06
cluster/main: Remove extra @if ( Cluster::is_enabled() )

awelzel:
These have been discussed in the context of "@if &analyze" [1] and am much in favor for not disabling/removing ~100 lines (more than fits on a single terminal) out from the middle of a file. There's no performance impact for having these handlers enabled unconditionally. Also, any future work on "@if &analyze" will look at them again which we could also skip.

This also reverts back to the behavior where the Cluster::LOG stream is created even in non cluster setups like in previous Zeek versions. As long as no one writes to it there's essentially no difference. If someone does write to Cluster::LOG, I'd argue not black holing these messages is better. Schema generators using Log::active_streams will continue to discover Cluster::LOG even if they run in non-cluster mode.

https://github.com/zeek/zeek/pull/3062#discussion_r1200498905


awelzel:
@timwoj - mostly a minor thing and sorry for being late, but IMO would be nice to align 6.0 with what we had previously. Yes, I'm very stubborn on that one :-)

PR #3110, opened 2023-06-09
Clarify subitem relationship in CMake configure summary.

bbannier:


awelzel (approved):
Think the intention was to indicate that tcmalloc/debugging relate to gperftools rather than being standalone options - okay to forgo this change maybe? I'm approving in case you want to move it in anyway. It doesn't matter enough to me :-)

bbannier:
> Think the intention was to indicate that tcmalloc/debugging relate to gperftools rather than being standalone options - okay to forgo this change maybe? I'm approving in case you want to move it in anyway. It doesn't matter enough to me :-)

Having never thought about this deeply, that interpretation never occurred to me from the way it looks. With that we could replace the existing `\t` with e.g., `  - ` to make clear this is a list nested under the previous item; alternatively I can drop this PR.

rsmmr (approved):

PR #3114, opened 2023-06-12
Remove deprecations tagged for 6.1

timwoj:


scripts/base/protocols/radius/consts.zeek
```diff
@@ -182,7 +182,7 @@ const attr_types: table[count] of string = {
 	[171] = "Delegated-IPv6-Prefix-Pool",
 	[172] = "Stateful-IPv6-Address-Pool",
 	[173] = "IPv6-6rd-Configuration"
-} &default=function(i: count): string { return fmt("unknown-%d", i); } &deprecated="Remove in v6.1.  Usage testing indicates this function is unused.";
+};
```
awelzel, line 185: I think we should reconsider those: It appears the tables aren't used in any script, but that doesn't mean they aren't useful (we're not removing them here) and having the &default here does make sense if someone was to start using the table.

In other words, a `&default=function` on a table maybe should not be considered unused?

src/Var.cc
```diff
@@ -195,18 +195,9 @@ static void make_var(const IDPtr& id, TypePtr t, InitClass c, ExprPtr init,
...
-
-		// The historical instances of these, such as the
-		// language/redef-same-prefixtable-idx.zeek btest, treat
-		// this as += rather than =, and with the initializer
-		// implicitly inside a list.
-		init = make_intrusive<ListExpr>(init);
-		c = INIT_EXTRA;
+		reporter->Error(util::fmt("Initialization not preceded by =/+=/-= is not allowed. (%s)",
+		                          obj_desc_short(init.get()).c_str()),
+		                init->GetLocationInfo());
```
awelzel, line 200: Haven't tried the effect, but could we `init->Error()` or so and return early? Seems any following error wouldn't (yet) be relevant or maybe even confusing?

Or tighten the parser to cause a syntax error, but the descriptive message here is likely more useful..

src/packet_analysis/protocol/gre/GRE.cc
```diff
@@ -46,12 +46,6 @@ bool GREAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
 		return false;
 		}
 
-	if ( ! BifConst::Tunnel::enable_gre )
-		{
-		Weird("GRE_tunnel", packet);
```
awelzel, line 51: Was that a mechanism to notify you that "enable_gre" was disabled but there was GRE traffic. Seems fine to let go, but wonder what the motivation was here. If the analyzer is disabled, why would it create a weird? :-)

testing/btest/scripts/base/frameworks/logging/rotate-ascii-logdir-leftover-log-rotation-shadow-files.zeek
```diff
@@ -28,7 +28,7 @@ function my_rotation_postprocessor(info: Log::RotationInfo) : bool
 	return T;
 	}
 
-redef LogAscii::logdir = "./logs";
 redef LogAscii::enable_leftover_log_rotation = T;
+redef Log::default_logdir = "./logs";
```
awelzel, line 32: There's two tests that should cover this (and the below with stale .shadow) files. Seems reasonable to delete the "rotate-ascii-logdir-leftover*" tests:

rotate-default-logdir-leftover-log-rotation-shadow-files.zeek
rotate-default-logdir-leftover-log-rotation-stale-shadow-files.zeek

testing/btest/scripts/base/frameworks/logging/rotate-ascii-logdir-leftover-log-rotation-stale-shadow-files.zeek
```diff
@@ -1,4 +1,4 @@
-# @TEST-DOC: Test that stale .shadow files are removed from LogAscii::logdir.
+# @TEST-DOC: Test that stale .shadow files are removed from Log::default_logdir.
```
awelzel, line 1: See comment above, I think there exists a test that covers this.

awelzel (commented):
Thanks for going through this. I left a few thoughts and will take a closer look and probably just approve then.

src/packet_analysis/protocol/gre/GRE.cc
timwoj: I dunno, that code existed long before it got split out into the separate analyzer. This is from the release/3.2 branch in Sessions.cc:

```
	case IPPROTO_GRE:
		{
		if ( ! zeek::BifConst::Tunnel::enable_gre )
			{
			Weird("GRE_tunnel", ip_hdr, encapsulation);
			return;
			}
```

It was just leftover from when the code was moved, and probably should have been removed then.

src/Var.cc
timwoj: This is the output difference from `language.init-mismatch` after changing it to `init->Error()` and returning. Because of the return the `b` variable doesn't exist anymore and we get an extra error message. Without the return, the variable would have been created but just wouldn't have been initialized.

```
  --- /dev/fd/63	2023-06-13 19:06:37
  +++ /dev/fd/62	2023-06-13 19:06:37
  @@ -18,3 +18,4 @@
   error in <...>/init-mismatch.zeek, line 34: type clash in assignment (s3 = set(s2))
   error in <...>/init-mismatch.zeek, line 36: pattern += op requires op to be a pattern (p += 3)
   error in <...>/init-mismatch.zeek, line 38: illegal table constructor element (1.2.3.4)
  +error in <...>/init-mismatch.zeek, line 7: undeclared variable (b)
```

timwoj: I switched it to use `init->Error()` but removed the return.

scripts/base/protocols/radius/consts.zeek
timwoj: Yeah I thought the same thing, but took them out since they were marked to be removed. I can revert those ones.

@vpax What do you think about not marking `default` functions as unused?

vpax: I believe that's marked as deprecated simply because there's no reference anywhere else to `attr_types`, _not_ because it erroneously gets marked as an unused function. If I'm wrong about that, then I'm wondering why the one above it works but it doesn't (and can look into that as an Issue).

timwoj: > I believe that's marked as deprecated simply because there's no reference anywhere else to attr_types, not because it erroneously gets marked as an unused function.

That's certainly possible. A quick grep shows that none of those variables are referenced anywhere else in the scripts or the source.

timwoj: I changed those functions to be marked `&is_used` instead of deprecated. I could see just removing them entirely but they're also documenting the protocol a bit so I left them in.

awelzel: > I believe that's marked as deprecated simply because there's no reference anywhere else to attr_types, not because it erroneously gets marked as an unused function.


`attr_types` not being used would be an argument for removing the whole table as Tim points out. The message about a `&default` function not being unused is pretty confusing unless you know implementation details. That's why I could see default functions being marked as implicitly used rather than explicitly needing the `&is_used` on it.

```
$ cat test.zeek 
module MyRadius;

const ids: table[count] of string = {
        [1] = "One",
} &default=function(c: count): string { return fmt("unknown-%d", c); };
$ zeek test.zeek 
warning in ./test.zeek, line 5: non-exported function does not have any callers (MyRadius::lambda_<17123251138430341678>)
```

It probably doesn't come up enough though, so `&is_used` seems fine :-)

vpax: Thanks @awelzel for doing a direct examination. Given the confusing nature of the message, I agree that ideally it'd be suppressed.  Feel free to file an issue on it and I'll take a look in my Copious Spare Time.  Also seems fine to just go with `&is_used` for now ...

testing/btest/language/redef-same-prefixtable-idx.zeek
```diff
@@ -4,12 +4,8 @@
 
 const my_table: table[subnet] of subnet &redef;
 
-redef my_table[3.0.0.0/8] = 1.0.0.0/8;
```
awelzel, line 7: I wonder if this was deprecated by accident, but it has been reported as deprecated for a bit, just not sure what the message ever meant for above case :-)
```
warning in ./my_table.zeek, line 3: initialization not preceded by =/+=/-= is deprecated (3.0.0.0/8 = 1.0.0.0/8)
```

awelzel: Vern explained that this is indeed supposed to go. Resolving.

PR #3116, opened 2023-06-13
Fix leftover-log-rotation in multi-logger setups

### LINKED ISSUE #3112: Left-over log rotation not multi-logger aware
Setting up a Zeek Cluster with 3 loggers in non-zeekctl mode and observed that "leftover log rotation" functionality does not tag a logger's name onto the rotated file.

Left-over log rotation runs at `InitPostScript()` time, but the `log_metadata` table is only populated during `zeek_init()`:

https://github.com/zeek/zeek/blob/857c15a293d07a66797f7f0d45270275f414ae51/scripts/base/frameworks/cluster/nodes/logger.zeek#L75-L83

I'm thinking of initializing `log_metadata` lazily now, or letting go of it completely again.

awelzel:
Populating log_metadata during zeek_init() is too late for the
leftover-log-rotation functionality, so do it at script parse time.

Also, prepend archiver_ to the log_metadata table and encoding function
due to being in the global namespace and to align with the
archiver_rotation_format_func. This hasn't been in a released
version yet, so fine to rename still.

Closes #3112


timwoj:
> @timwoj , this is 6.0.0 material.

Got it.

PR #3119, opened 2023-06-13
tests: Do not use archiver_rotation_format_func as postprocessor

awelzel:
This test triggered ubsan by putting a function with the wrong type as a post-processor into the .shadow file. Don't do that.

Likely Zeek should provide a better error message, but hand-crafting .shadow files isn't what is normally done and this is to fix the master build for now.

Follow-up for #3116


timwoj (approved):
The ARM build is failing for some reason, but it's not related to this PR. The change does fix the ubsan failure.

Issue #3090, opened 2023-05-26
CXXFLAGS used by external plugins: differences between 5.2.2 and master


awelzel:
Compiling external plugins with master uses different CXXFLAGS than with Zeek 5.2.2.

The following flags aren't passed to external plugins anymore (and the rapidjson include):
```
-Wall -Wno-unused -Wno-register -Werror=vla -g -I/opt/zeek-5.2.2-debug/include/zeek/3rdparty/rapidjson/include
```

Edit after Tim's reply: I had missed the following:
```
-DDOCTEST_CONFIG_DISABLE 
```

Mostly recording a ticket for discussion. Maybe can close as acceptable new behavior. For me it was the missing `-g` that lead at looking: Didn't have line numbers in crashdumps with an external plugin with `master`, but do get them with Zeek 5.2.2 (debug builds, both).

Sample output of `VERBOSE=1 make` between 5.2.2 and 6.0.0 dev660:

#### zeek version 5.2.2-debug
> ```/usr/bin/c++ -DDOCTEST_CONFIG_DISABLE -DZEEK_PLUGIN_INTERNAL_BUILD=false -DZeek_JavaScript_linux_x86_64_EXPORTS -I/opt/node-19.8/include/node -I/opt/node-19.8/include -I/home/awelzel/corelight-oss/zeekjs/src -I/home/awelzel/corelight-oss/zeekjs/build/src -I/home/awelzel/corelight-oss/zeekjs/build -I/opt/zeek-5.2.2-debug/include -I/opt/zeek-5.2.2-debug/include/binpac -I/opt/zeek-5.2.2-debug/include/zeek/3rdparty/rapidjson/include -std=c++17 -Wall -Wno-unused -Wno-register -Werror=vla -Wall -Wno-unused -g -DDEBUG -DBRO_DEBUG -Wno-register -Werror=vla -g -fPIC -o CMakeFiles/Zeek-JavaScript.linux-x86_64.dir/src/Plugin.cc.o -c /home/awelzel/corelight-oss/zeekjs/src/Plugin.cc```

#### zeek version 6.0.0-dev.660-debug
[...]

PR #3124, opened 2023-06-14
Force -std=c++17 mode for plugin targets, avoid using RequireCXX17.cmake

timwoj:
This ended up quite a bit messier than I had hoped it would, but it resolves the issues with both `-std=c++17` not being set at all and with duplicate setting of `-std=gnu++17` and `-std=c++17`. Fixes part of #3090.


bbannier (approved):
This looks good to me. The downside of this approach is that it introduces a lot of copy-n-paste code and will be something we need to remember whenever we add a new library or executable target; the alternative would be to globally set them which would introduce its own issues (what if we do want to deviate?).

PR #3125, opened 2023-06-15
June '23 ZAM script optimization maintenance

vpax:
Here's my ~monthly PR for script optimization, this time focused on ZAM (I have a separate one close to ready for `gen-C++`). I think the individual commits are fairly self-explanatory, though happy to clarify any of them as is helpful. The main thing I'd flag is the decision that script optimization just ignores the new `assert` statement, which fits with my view that the point of `assert`s are for debugging at the cost of performance, whereas script optimization already includes some instances of performance-at-the-cost-of-debugging.


timwoj (approved):

src/script_opt/ZAM/README.md
```diff
@@ -75,13 +75,12 @@ of the expression are modified by compiled scripts.
 
 ### Incompatibilities:
 
+* ZAM ignores `assert` statements.
+
```
awelzel, line 79: We had chatted via DM a bit, but assuming `assert` finds its way into `testing/btest` over time, do we need to mark every test that's using it (which might be most  simple ones over time) with a `@TEST-REQUIRES: test "${ZEEK_ZAM}" != "1"` ?

I suspect tests would still run even in ZAM mode, but nothing actually asserted/validated and potentially missing out on errors specific to running in ZAM mode.

vpax: You have a good point.  What comes to mind is if indeed we start using it for routine tests, down the road adding a mechanism that tells script optimization "please include the asserts rather than discarding them", which we'd turn on for running on the test suite.  That way we get both the performance benefits and the testing benefits.

PR #3127, opened 2023-06-15
CMakeLists: Add rapidjson/include to zeek_dynamic_plugin_base

awelzel:
threading/formatters/JSON.h has a rapidjson include. Extend the include directories of external plugins so they are setup to find these in Zeek's install tree.

Relates to #3090


Neverlord (approved):
I would also like to add `SYSTEM` here to suppress unnecessary warnings from 3rd party headers.

Neverlord (approved):

PR #3129, opened 2023-06-15
CMakeLists: Propagate DOCTEST defines to external plugins 

awelzel:
If Zeek has not been built with doctest enabled then it's not easy
to run a plugin's tests (and if they really wanted to they could
again undef the DOCTEST_CONFIG_DISABLE and provide their own doctest
main() implementation.

Relates to https://github.com/zeek/zeek/issues/3090


awelzel:
Testing with a Zeek built with `--disable-cpp-tests` or not, the flags used when compiling external plugin files are `-DDOCTEST_CONFIG_DISABLE` or `-DDOCTEST_CONFIG_SUPER_FAST_ASSERTS`.

timwoj (approved):

PR #3134, opened 2023-06-15
Update dce-rpc constants

timwoj:
Provided by @IAMUSConsulting in https://github.com/zeek/zeek/discussions/3133. I don't have enough knowlege of dce-rpc to verify that these are correct, if someone else can.


ckreibich (approved):
Thank you very much, @IAMUSConsulting. I pulled up the values in the MS docs and it all seems to check out.

PR #3137, opened 2023-06-16
Convert uses of varargs to variadic templates

timwoj:
This is something I've had sitting on a branch for a long time. I figured I'd dig it out, clean it up for the current state of the source, and open a draft PR for it.

This PR changes all of the uses (minus two) of varargs/va_list for logging functions into variadic templates/parameter packs. This is in preparation for potentially using libfmt/std::format for text formatting, and converting all of these into modern C++ facilitates that goal a bit. The two uses missing are util::fmt and util::vfmt, since those two would effectively be replaced by a different formatter method if the follow-on work takes place. There are a few other places where varargs are used, but they weren't logging-related so I left them alone for now.

There are some drawbacks to this approach:
1. We lose the ability to ensure that the format arguments for log functions are proper `printf`-style formats, due to the loss of the `__attribute__` tags applied to the functions. Clang 15.x gains this function on template functions, but not until then.
2. Build times will be longer, due to the addition of a bunch of template methods in heavily-used headers (Reporter.h, for example).

### DIFF
--- a/src/Reporter.h
+++ b/src/Reporter.h
@@ -223,7 +223,11 @@ public:
 			return;
 
 		char* buf;
+#ifdef HAVE_ASPRINTF
 		asprintf(&buf, fmt, std::forward<Args>(args)...);
+#else
+		util::asprintf(&buf, fmt, std::forward<Args>(args)...);
+#endif
 		DoSyslog(buf);
 		free(buf);
 		}
--- a/src/Reporter.h
+++ b/src/Reporter.h
@@ -2,7 +2,7 @@
 
 #pragma once
 
-#include <stdarg.h>
+#include <stdio.h>
 #include <list>
 #include <map>
 #include <string>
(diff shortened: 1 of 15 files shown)


src/Reporter.h
```diff
@@ -144,21 +217,53 @@ class Reporter
...
+		{
+		if ( run_state::reading_traces )
+			return;
+
+		char* buf;
+#ifdef HAVE_ASPRINTF
+		asprintf(&buf, fmt, std::forward<Args>(args)...);
+#else
+		util::asprintf(&buf, fmt, std::forward<Args>(args)...);
+#endif
```
bbannier, line 230: Just as a thought in passing while this is still in flux: the conditionally compiled code could also be in `utitl::asprintf` so this code could unconditionally call it without having to worry about wrappers or not; in that case `util::asprintf` would also be defined unconditionally.

One benefit of that approach would be that we'd end up with a very simple guideline for reviews, `asprintf` should never be called, unless you are implementing `asprintf`.

timwoj: I left it like this because `util::asprintf` uses `vasprintf`, which requires a va_list to work, which in turn requires `util::asprintf` to be a C-style variadic function. I could make two versions of the function conditionally wrapped in `#ifdef` block, but I don't see a way to call `vasprintf` otherwise.

timwoj: The more correct fix here would be to implement `vasprintf` in libunistd, so it's only Windows that's missing it.

timwoj:
Closing this PR as the work that I was planning on doing on top of it is a long way off, so there's no good reason to merge this.

Issue #3144, opened 2023-06-21
out_of_bound ASN1EncodingMeta: more_len 129 > 26 for GSSAPI analyzer


awelzel:
This has recently been reported on the public [Zeek Slack channel](https://zeekorg.slack.com/archives/CSZBXF6TH/p1687245878023269)
> Goodmorning. What is the reason that i have a lot of binpac exceptions: out_of_bound ASN1EncodingMeta: more_len 129 > 26 for GSSAPI analyzer... Is something i can do to stop it in dpd.log? Thanks

And then a day later privately reported as well. The private report comes with a pcap, but unfortunately can't be shared publicly.

Initial skimming in the parser around `more_len` hints at the fact that we do not support "high tag number" form and possibly that causes parsing to fail.

https://github.com/zeek/zeek/blob/46715dacfd2f955739d3bf4010c74fbf827ad946/src/analyzer/protocol/asn1/asn1.pac#L31-L33

http://luca.ntop.org/Teaching/Appunti/asn1.html

Mohan-Dhawan:
I might be entirely off the mark here as I do not know the whole issue. In my interactions with the ASN1 binpac parser, I found that several analyzer violations happened because GSSAPI did not handle the WRAP and MIC tokens (https://www.rfc-editor.org/rfc/rfc4121.html#section-4.2.6.1. The built-in analyzer only analyzes the NEG token. In my experiments, whenever there was a payload for either WRAP/MIC tokens, the analyzer threw violations.

0xxon:
Yup, you are indeed correct - this is a case where there are wrap/mic tokens present, which are missing the framing that is expected. From a short glance, it seems like it should not be hard to add support for these.

PR #3146, opened 2023-06-21
GSSAPI: basic support for MIC/WRAP tokens

0xxon:
When MIC/WRAP tokens are encountered, we now skip the message, instead of raising a parse error. The data in the messages is encrypted - so it does not seem work to raise an event at the moment.

Fixes GH-3144

### DIFF
--- a/src/analyzer/protocol/gssapi/gssapi-protocol.pac
+++ b/src/analyzer/protocol/gssapi/gssapi-protocol.pac
@@ -1,6 +1,16 @@
-
-type GSSAPI_NEG_TOKEN(is_orig: bool) = record {
+type GSSAPI_SELECT(is_orig: bool) = record {
 	wrapper  : ASN1EncodingMeta;
+	token: case tok_id of {
+		0x0404 -> mic_blob: bytestring &restofdata;
+		0x0504 -> wrap_blob: bytestring &restofdata;
+		default -> neg_token: GSSAPI_NEG_TOKEN(is_orig, is_init);
+	} &requires(is_init) &requires(tok_id);
+} &let {
+	is_init: bool = wrapper.tag == 0x60;
+	tok_id: uint32 = (wrapper.tag << 8) | wrapper.len;
+} &byteorder=littleendian;
+
+type GSSAPI_NEG_TOKEN(is_orig: bool, is_init: bool) = record {
 	have_oid : case is_init of {
 		true  -> oid    : ASN1Encoding;
 		false -> no_oid : empty;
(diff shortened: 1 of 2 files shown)


0xxon:
I sadly don't have public test data for this - thus there currently is no testcase.

src/analyzer/protocol/gssapi/gssapi-protocol.pac
```diff
@@ -1,6 +1,16 @@
...
 	wrapper  : ASN1EncodingMeta;
+	token: case tok_id of {
+		0x0404 -> mic_blob: bytestring &restofdata;
+		0x0504 -> wrap_blob: bytestring &restofdata;
+		default -> neg_token: GSSAPI_NEG_TOKEN(is_orig, is_init);
+	} &requires(is_init) &requires(tok_id);
+} &let {
+	is_init: bool = wrapper.tag == 0x60;
+	tok_id: uint32 = (wrapper.tag << 8) | wrapper.len;
+} &byteorder=littleendian;
```
awelzel, line 11: Above mentioned RFC has a lot of big-endian in it for some of the fields in the tokens, but probably only relevant if support was actually added.

awelzel (approved):

src/analyzer/protocol/gssapi/gssapi-protocol.pac
```diff
@@ -1,6 +1,16 @@
...
+type GSSAPI_SELECT(is_orig: bool) = record {
 	wrapper  : ASN1EncodingMeta;
+	token: case tok_id of {
+		0x0404 -> mic_blob: bytestring &restofdata;
+		0x0504 -> wrap_blob: bytestring &restofdata;
+		default -> neg_token: GSSAPI_NEG_TOKEN(is_orig, is_init);
+	} &requires(is_init) &requires(tok_id);
+} &let {
+	is_init: bool = wrapper.tag == 0x60;
+	tok_id: uint32 = (wrapper.tag << 8) | wrapper.len;
```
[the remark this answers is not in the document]
0xxon: Yup - I am essentially doing this, just the other way round.

src/analyzer/protocol/gssapi/gssapi-protocol.pac
0xxon: Yup - and currently I see no reason to actually parse the messages.

0xxon:
I know that it is super late, but in theory this one would be really nice for 6.0, @timwoj. This probably breaks parsing this kind of traffic in a lot of cases.

timwoj:
> I know that it is super late, but in theory this one would be really nice for 6.0, @timwoj. This probably breaks parsing this kind of traffic in a lot of cases.

/me sighs

😉

PR #3147, opened 2023-06-21
cirrus: Stop pushing container images to zeekurity

### LINKED ISSUE #3143: Stop publishing to Docker zeekurity organization, there is zeek now
This was discussed internally and the ideas was to not publish the 6.0 release to zeekurity anymore, only zeek

@bbannier suggested to remove the `latest` (and probably the `lts`) tags so that consumers receive an error upon pulling what would otherwise be a potentially outdated tag. Keeping versioned tags around should not hurt.

Deleted the tags through the UI seems easiest for this task.

awelzel:
Closes #3143

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -407,8 +407,6 @@ windows_task:
   # 2 hour timeout just for potential of building Docker image taking a while
   timeout_in: 120m
   windows_container:
-    # image: cirrusci/windowsservercore:cmake
-    # image: zeekurity/broker-ci-windows:latest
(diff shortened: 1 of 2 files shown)


timwoj (approved):
There's a reference to a zeekurity broker image in the Windows section. Can it go away as well?

awelzel:
This is for Zeek 6.0 so that the images do not go out.

For merging this into master we should make an announcement on community.zeek.org that the lts and latest tags will vanish.

PR #3149, opened 2023-06-26
from_json: Support function to normalize key names

### LINKED ISSUE #3142: Normalize JSON keys when matching against record types from `from_json` BIF.
Right now, the implementation of JSON deserialization for the new `from_json` BIF matches record types to JSON key-value pairs by iterating through the record fields and finding matching JSON keys. Unfortunately, this fails to correctly handle JSON record types with characters which are disallowed in Zeekscript record fields. A concrete example is trying to parse Manufacturer Usage Description files into Zeek using this new BIF; MUD includes fields with `-` and `:` in them, which can't currently be matched to Zeekscript record fields.

The solution would be to normalize JSON keys in some documented way, perhaps by 1) ignoring disallowed characters or 2) replacing disallowed characters with an underscore.

The relevant code is in the `TYPE_RECORD` branch of the `switch` on type tag in `BuildVal(const rapidjson::Value&, const TypePtr&)` in `zeek/src/Val.cc`. __[View code](https://github.com/zeek/zeek/blob/master/src/Val.cc#L1297-L1329)__

awelzel:
Closes #3142


src/Val.cc
```diff
@@ -1301,24 +1302,60 @@ static std::variant<ValPtr, std::string> BuildVal(const rapidjson::Value& j, con
...
+					}
 
-				if ( ! has_member || member_is_null )
+				if ( ! jval || jval->IsNull() )
 					{
 					if ( ! td_i->GetAttr(detail::ATTR_OPTIONAL) &&
 					     ! td_i->GetAttr(detail::ATTR_DEFAULT) )
 						return util::fmt("required field %s$%s is %s in JSON", t->GetName().c_str(),
-						                 td_i->id, member_is_null ? "null" : "missing");
+						                 td_i->id, jval ? "null" : "missing");
```
timwoj, line 1353: Previously the statement checked `jval && jval->IsNull()` via `member_is_null`, but now it checks if the jval itself is null. Is that intentional?

awelzel: Yes, but maybe this is a bit too clever... We only get here if `jval` is a nullptr (missing in the JSON object) or `jval->IsNull()` was true (it's a null JSON value). That's what `jval ? "null" : "missing"` uses.

Can bring back `has_member` and `member_is_null` bools, but they also seemed redundant at the time :-)

timwoj: Ah, yeah now that I look again the way you have it makes sense 👍🏼 Maybe add a comment there just for clarity.

awelzel: Done, are you otherwise fine with this PR?

timwoj (approved):

PR #3155, opened 2023-06-27
Rename magic __bro_plugin__ file to __zeek_plugin__

timwoj:
This is part of the on-going removal of Bro naming from zkg. Related PRs: https://github.com/zeek/zeek-docs/pull/198 and https://github.com/zeek/cmake/pull/86

### DIFF
--- a/src/plugin/Manager.cc
+++ b/src/plugin/Manager.cc
@@ -94,7 +94,7 @@ void Manager::SearchDynamicPlugins(const std::string& dir)
 
 	// Check if it's a plugin directory.
 
-	const std::string magic = dir + "/__bro_plugin__";
+	const std::string magic = dir + "/__zeek_plugin__";
 
 	if ( util::is_file(magic) )
 		{
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,12 @@ Breaking Changes
 
 - ``assert`` is now a reserved keyword for the new ``assert`` statement.
 
+- The ``__bro_plugin__`` file that gets generated as part of plugin builds was
+  renamed to ``__zeek_plugin__``. This will affect the ability for older
+  versions of ``zkg`` to use the ``zkg unload`` and ``zkg load`` commands. This
+  should only cause breakage for people using a version of ``zkg` that doesn't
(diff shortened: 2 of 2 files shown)


src/plugin/Manager.cc
```diff
@@ -94,10 +94,22 @@ void Manager::SearchDynamicPlugins(const std::string& dir)
...
+		magic = dir + "/__bro_plugin__";
+		}
 
 	if ( util::is_file(magic) )
 		{
+		if ( bro_plugin )
+			reporter->Warning(
+				"__bro_plugin__ is deprecated and will be unsupported in Zeek 7.1. Check to ensure "
+				"that the version of zkg in use is newer than x.y.z.");
+
```
awelzel, line 112: Even if I was the one that started the [discussion](https://github.com/zeek/cmake/pull/86/files#r1247997933): I'm now thinking it's not worth the hassle. The CMake part that generates `__bro_plugin__` or `__zeek_plugin__` is part of a Zeek installation tree so in Zeek it should be clear which one to expect.

The two scenarios that comes up:
* an older `zkg` version will not recognizing `__zeek_plugin__`: It cannot disable plugins. `zkg install` of a plugin should continue to work as expected as the initial `__zeek_plugin__` is created through cmake, not zkg. Unclear how many users actually use `zkg unload` for plugins and also, how many of those use a distro provided zkg. This is where I think "unlikely to hit anyone".

* A newer `zkg` not recognizing `__bro_plugin__` anymore. This seems like we we can/should add backwards compat on the zkg side for that, so that it moves either `__bro_plugin__` or `__zeek_plugin__` depending on which exists.

timwoj: I brought it up in Slack here: https://zeekorg.slack.com/archives/CSZBXF6TH/p1689193958220359

We'll see if there are any replies.

timwoj:
@ckreibich Did you have any comments on this? I'd like to merge it so I can do the zkg side of it.

awelzel:
Approved - I presume a NEWS entry might be good for users using non-bundled zkg packages.

PR #3158, opened 2023-06-30
Attempt to protect from unbounded DCE-RPC state

### LINKED ISSUE #3145: Unbounded per-connection state growth in dce-rpc scripts
Constant worker memory growth with eventual OOM crash has been reported in the public Zeek Slack: https://zeekorg.slack.com/archives/CSZBXF6TH/p1685587437627989

After deploying [per-conn-footprint](https://github.com/awelzel/zeek-per-conn-footprint) which regularly logs the footprint of all connections and first-level fields, a connection was identified with a `dce_rpc_backing` table and a footprint of 4mio, growing at around 1800/s.

```
# [footprint, uid, services, duration, details]
[3863086,1687255886.350536,"ClMdKa4h5LSvI4BeTc",["SMB","DCE_RPC","KRB","GSSAPI"],2735.635337114334,"start_time=1,krb=32,conn=32,duration=1,ftp_data_reuse=1,id=8,dce_rpc_state=3,dce_rpc_backing=3814351,uid=1,vlan=1,orig=12,smb_state=48543,history=1,removal_hooks=6,extract_orig=1,dce_rpc=17,service=12,known_services_done=1,resp=12,extract_resp=1"]
[3982057,1687255946.351189,"ClMdKa4h5LSvI4BeTc",["SMB","DCE_RPC","KRB","GSSAPI"],2795.6359899044037,"start_time=1,krb=32,conn=32,duration=1,ftp_data_reuse=1,id=8,dce_rpc_state=13,dce_rpc_backing=3934193,uid=1,vlan=1,orig=12,smb_state=47659,history=1,removal_hooks=6,extract_orig=1,dce_rpc=20,service=12,known_services_done=1,resp=12,extract_resp=1"]
[4113009,1687256006.351253,"ClMdKa4h5LSvI4BeTc",["SMB","DCE_RPC","KRB","GSSAPI"],2855.6360540390015,"start_time=1,krb=32,conn=32,duration=1,ftp_data_reuse=1,id=8,dce_rpc_state=13,dce_rpc_backing=4053416,uid=1,vlan=1,orig=12,smb_state=59408,history=1,removal_hooks=6,extract_orig=1,service=12,known_services_done=1,resp=12,extract_resp=1"]
```

Can't easily tell using the footprint if it is `c$dce_rpc_backing` table growing or `State$ctx_to_uuid` within that table (that should be bounded by ~16k~ 64k though). Looking through the scripts, no provisions for cleaning up either is there.

profile-20230616_0810

awelzel:
1) For SMB close requests, tear down any associated DCE-RPC
   analyzer if one exists.

2) Protect from fid_to_analyzer_map growing unbounded by introducing a
   new SMB::max_dce_rpc_analyzers limit and forcefully wipe the
   analyzers if exceeded. Propagate this to script land as event
   smb_discarded_dce_rpc_analyzers() for additional cleanup.

3) Add script level handling for `smb2_close_request()` to clean out dce_rpc_backing.

Closes #3145


timwoj:
I could see this going into the next set of patch releases.

timwoj:
@JustinAzoff Since you were poking at this one, do you want to review it?

src/analyzer/protocol/smb/smb-pipe.pac
```diff
@@ -49,6 +49,22 @@ refine connection SMB_Conn += {
...
+				if ( smb_discarded_dce_rpc_analyzers )
+					zeek::BifEvent::enqueue_smb_discarded_dce_rpc_analyzers(zeek_analyzer(), zeek_analyzer()->Conn());
+
+				for ( const auto& kv : fid_to_analyzer_map )
+					{
+					kv.second->Done();
+					delete kv.second;
+					}
+
+				fid_to_analyzer_map.clear();
```
timwoj, line 65: I'm curious whether we need to clear out all of them or if we could either just drop the new one or some sort of FIFO queue. The actual SMB connection doesn't drop, so would we just start making more new DCE analyzers to go with it?

awelzel: The reproducer (and presumed behavior in the wild) is that the same pipe is opened and closed again over and over again, allocating a new `fid` and analyzer sequentially. In that scenario, dropping all analyzers or doing fifo style discard makes no difference.

Further, that code path only kicks in if we're missing SMB close requests due to capture loss / packet drops. If we see all close requests the analyzers map should mostly contain a single entry only.

If we discard a fid that is then re-used over the same connection, yes, it'll just re-instantiate a new one. Think that's okay as it shouldn't be the normal case.

timwoj (approved):

Issue #3136, opened 2023-06-16
Add metric for loaded modules.


J-Gras:
It would be helpful to export the loaded modules via the telemetry framework. Currently, it would be possible to export loaded scripts. However, this is quite fine-grained.

awelzel:
In DM I had proposed it may be more interesting to expose information about modules to the Zeek script layer.

Optimally `global_ids()` would list/include them with type_name=module or some-such.  Exposing them via telemetry is then a custom script away and the extension may be useful elsewhere.

awelzel:
@J-Gras - with latest master, `global_ids()` can be used to find module names. For adding metrics, you could either have one metric per loaded module, but given it's completely static information (and unclear how you'd want to consume it), maybe better to stuff all names into a single label value or even expose it differently like an API endpoint or so.

```
@load base/frameworks/telemetry


# One metric per module, not suggested to do this.
global module_gf = Telemetry::register_gauge_family([
        $prefix="zeek",
        $name="module_info",
        $unit="1",
        $help_text="Indicates this Zeek module is available",
        $labels=vector("module_name"),
]);

# all modules stuffed into a single metric
global all_modules_gf = Telemetry::register_gauge_family([
        $prefix="zeek",
        $name="modules_info",
        $unit="1",
        $help_text="All available Zeek modules",
        $labels=vector("names"),
]);

event zeek_init() {
        local all_module_names: vector of string;
        for ( name, v in global_ids() ) {
                if ( v$type_name == "module" ) {
                        local module_name = gsub(name, /^module /, "");
                        Telemetry::gauge_family_set(module_gf, vector(module_name), 1.0);
                        all_module_names += module_name;
                }
        }

        # For the single metric approach - sorted names for predictability:
        sort(all_module_names, strcmp);
        Telemetry::gauge_family_set(all_modules_gf, vector(join_string_vec(all_module_names, ",")), 1.0);
}

```

```
[...]

PR #3161, opened 2023-07-04
Introduce Type and Val for modules

awelzel:
This relates to #3136. Instead of a new bif returning to expose the list of modules, this goes a bit further adding a ModuleType and Value.

```
    Introduce Type and Val for modules
    
    This adds shallow support for exposing modules to the script layer as
    identifiers and values.
    
    1) global_ids() now returns identifiers representing modules unless
       they are shadowed by global variables or types.
    
    2) Modules can be referenced in expressions. At this point there's
       not really anything one can do with them except for printing.
       Assignment, for example, does not work.
    
    A new warning is generated when global identifiers shadow modules. One such
    occurrence in the base scripts is Management::Node which is both a module
    and a record type. Seems like it would be nice to disambiguate this.
    
    This should allow to implement #3136 by external scripts using the new
    module_ids() bif.
```

If this is overdoing it and a simple `module_ids(): vector of string` seems sufficient to get access to module names I'm happy to track back.


awelzel:
Chatted with Robin and he's not fond of adding Val and Types because it's a can of worms. Will re-open a separate PR that modifies `global_ids()` in a more ad-hoc way, but should still provide enough functionality to cover #3136 .

PR #3162, opened 2023-07-04
global_ids(): Include module names

awelzel:
Ad-hoc include module names in the global_ids() table. Table values will have the type_name field set to "module" and their key in the table is prefixed with "module " to avoid clashes with existing global identifiers shadowing module names (Management::Node being an existing example).


awelzel:
@rsmmr - assigning you. I could also see now going back further and just having `module_names(): vector of string` BiF leave `global_ids()` alone and move. I'm not too fond of the `"module "`: As a user I'd likely want to strip it away and it's mostly an artifact to avoid the collision with identifiers.

src/zeek.bif
```diff
@@ -2129,10 +2129,31 @@ function global_ids%(%): id_table
...
+		auto rec = to_script_id_rec(id);
 		ids->Assign(std::move(id_name), std::move(rec));
 		}
 
+	// We do not track modules as identifiers, ad-hoc instantiate one
+	// that is re-used here.
+	auto mod_id = zeek::make_intrusive<zeek::detail::ID>("<<void>>", zeek::detail::SCOPE_GLOBAL, false /* is_export */);
+	mod_id->SetType(zeek::base_type(zeek::TYPE_VOID));
+	auto base_rec = to_script_id_rec(mod_id);
+	base_rec->Assign(0, "module");  // Assign type_name as "module"
```
rsmmr, line 2149: Seems feels a bit indirect. I get the desire to factor out the logic into the lambda, but I think it would be more readable to just initialize module-specific records here directly, then you'd not need the fake mod_id.

src/Var.cc
```diff
@@ -470,6 +470,18 @@ void add_type(ID* id, TypePtr t, std::unique_ptr<std::vector<AttrPtr>> attr)
 		id->SetAttrs(make_intrusive<Attributes>(std::move(*attr), tnew, false, false));
 	}
 
+static std::set<std::string, std::less<>> all_module_names;
```
rsmmr, line 473:
```suggestion
static std::set<std::string> all_module_names;
```

Should work without, no?

rsmmr (approved):
Looks good to me with just a couple minor comments.

Re/  `module_names()`: I can see it either way. I like it being part of global IDs because semantically they *are* global IDs; but I do see that's it's a bit awkward due to the potential naming conflicts, so I wouldn't object to `module_names()` either.

src/Var.cc
awelzel: Yep, works and I can remove it. [Sonarlint suggested to add](https://rules.sonarsource.com/cpp/RSPEC-6045/) std::less<> and my understanding is that it can void creating string objects when doing lookups using `const char *` or `string_view`.

It won't matter here and if it looks noisy unusual better to drop.

PR #3165, opened 2023-07-05
[Spicy] Support `switch` fields when exporting Spicy types to Zeek.

### LINKED ISSUE #3157: Incomplete Zeek-side type generation for exported Spicy unit with unit switch
The following test should work, but fails since the Zeek-side field for the conditionally filled field `foo` is not generated.

```ruby
# @TEST-EXEC: spicyz -do test.hlto test.spicy test.evt
# @TEST-EXEC: zeek -Cr ${TRACES}/udp-packet.pcap test.hlto main.zeek >output 2>&1

# @TEST-START-FILE test.spicy
module TEST;

public type Message = unit {
    switch (1) {
        * -> foo: bytes &eod;
    };
};
# @TEST-END-FILE

# @TEST-START-FILE test.evt
import TEST;
protocol analyzer spicy::Test over UDP:
    port 0/udp - 42000/udp,
    parse with TEST::Message;

export TEST::Message;

on TEST::Message -> event TEST_ZEEK::MessageEvt(self);
# @TEST-END-FILE

# @TEST-START-FILE main.zeek
event TEST_ZEEK::MessageEvt(message: TEST::Message)
	{ }
# @TEST-END-FILE
```

As of `6.1.0-dev.152` (ccfc68fccd9b0635210bf951ee6d4769f997d157) this emits the following runtime warning:

```
XXXXXXXXXX.XXXXXX analyzer error in <...>/test.evt, line 8: Event parameter mismatch, no matching record field for field 'foo'
```

This looks similar to #3083 with the additional twist of a unit switch.

rsmmr:
Closes #3157.


auxil/spicy
bbannier, line 1: This bump is not needed here (and bumps to a commit not on Spicy's `main` branch). Let's leave it out here, also to make backporting this patch easier.

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -1194,6 +1194,23 @@ struct VisitorZeekType : hilti::visitor::PreOrder<hilti::Result<hilti::Expressio
         return (! ids.empty() && ids.front().has_value()) ? ids.front()->namespace_() : hilti::ID();
     }
 
+    result_t create_record_type(const hilti::ID& ns, const hilti::ID& local,
+                                const std::vector<hilti::Expression>& fields) {
+        if ( fields.size() ) {
```
bbannier, line 1199: Nit
```suggestion
        if ( ! fields.empty() ) {
```

bbannier, line 1199: The `if`/`else` branches are only for logging and can do quite some work just iterating, even if logging is disabled.

WDYT about wrapping all of this in a `if ( hilti::logger().isEnabled(ZeekPlugin) ) {
...`?

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -1194,6 +1194,23 @@ struct VisitorZeekType : hilti::visitor::PreOrder<hilti::Result<hilti::Expressio
...
+            SPICY_DEBUG(hilti::util::fmt("Creating Zeek record type %s::%s with fields:", ns, local));
+
+            for ( const auto& f : fields )
+                SPICY_DEBUG(
+                    hilti::util::fmt("  %s",
+                                     f.as<hilti::expression::Ctor>().ctor().as<hilti::ctor::Tuple>().value()[0]));
```
bbannier, line 1205: Logging the ctr parameters will print e.g., `"foo"` instead of `foo`; we could do some more unwrapping here to get the actual source name.

I think it would also be nice to add `-D zeek` and storing in `output` in the new test so we exercise this logging code.

bbannier (approved):

auxil/spicy
rsmmr: That was accidental, will skip.

src/spicy/spicyz/glue-compiler.cc
rsmmr: > Logging the ctr parameters will print e.g., `"foo"` instead of `foo`; we could do some more unwrapping here to get the actual source name.

+1

> I think it would also be nice to add `-D zeek` and storing in `output` in the new test so we exercise this logging code.

I'll skip that since we aren't generally testing debug output elsewhere either.

PR #3170, opened 2023-07-05
ci/freebsd: Use pkg-static to avoid libssl.so.30 error

awelzel:
Starting a gcloud instance and looking around a bit, there's actually a pkg-static executable that works.


awelzel:
Oh well, that just starts failing later.

I think we essentially wait for this one until there's a new 14-snap image: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272354

Issue #3159, opened 2023-07-03
when() capture issue and segfault potential in script parser with nested when()'s


dopheide-esnet:
[test.zeek.txt](https://github.com/zeek/zeek/files/11941485/test.zeek.txt)

This started with master on 6/30/2023, I'm not sure of the exact commit yet.

The two errors are
[1]
error in ./test.zeek, line 40: Test::ip is used inside "when" statement but not captured (when [Test::r](Test::r = Broker::keys(Test::host_store$store)) { if (Test::r$status == Broker::SUCCESS && Test::r?$result) { for ([Test::ip] in Test::r$result as addr_set) { when [Test::ip, Test::res](Test::res = Broker::get(Test::host_store$store, to_any_coerceTest::ip)) { print fmt(%s, Test::res$result as string)} timeout Test::host_store_timeout }}} timeout Test::host_store_timeout )

[2]
error in ./test.zeek, line 29: no such local identifier: ip
Segmentation fault

The 'ip' variable comes from a for() loop inside.   I wasn't able to get the error to happen in a simple test case that doesn't use Broker.  It appears I can get around it by declaring a blank variable first, but the inconsistent behavior still bothers me.

Potential work around:
```
local ip: addr;
when [ip] (local r = Broker::keys(Test::host_store$store)){
```

dopheide-esnet:
Figured out a simpler example that doesn't require Broker, the 'ip' variable needs to be used in the conditional of the 2nd when().
```

module Test;

event zeek_init(){
	local myset: set[addr] = { 10.0.0.1, 10.0.0.2 };
	when [myset] (T){
		for (ip in myset){
			when [ip] (ip == ip){
				print fmt("%s",ip);
			}
		}
	}
}

```

PR #3172, opened 2023-07-05
Fix for nested "when" statements leading to confusion over scoping

vpax:
Closes https://github.com/zeek/zeek/issues/3159.


src/Var.cc
```diff
@@ -783,7 +787,11 @@ TraversalCode OuterIDBindingFinder::PreStmt(const Stmt* stmt)
 	auto ws = static_cast<const WhenStmt*>(stmt);
 
 	for ( auto& cl : ws->Info()->WhenExprLocals() )
-		outer_id_references.insert(const_cast<ID*>(cl.get()));
+		{
```
timwoj, line 790: Could add a comment here about what this is actually doing? I think I understand that it's only considering this a reference if didn't find the same id in the local scope, but it's not terribly obvious.

timwoj (changes requested):

src/Var.cc
vpax: Thanks for the nudge to revisit that code.  I think you'll find the replacement PR I'm about to submit aesthetically more pleasing ;-).

PR #3173, opened 2023-07-05
Define early_shutdown lambda earlier in zeek-setup, avoids build failure with gperftools

### LINKED ISSUE #3163: Building with gperftools seems broken
Building Zeek 6 with `--enable-perftools-debug` fails with a couple of messages like the following:
```
.../zeek/src/zeek-setup.cc:916:4: error: use of undeclared identifier 'early_shutdown'
                        early_shutdown();
                        ^
.../zeek/src/zeek-setup.cc:953:5: error: use of undeclared identifier 'early_shutdown'
                        early_shutdown();
                        ^
...
```

Patch suggested by @awelzel seems to fix the issue:
```
$ git diff src/zeek-setup.cc
diff --git a/src/zeek-setup.cc b/src/zeek-setup.cc
index 46d5eacef..621dfeb85 100644
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -906,9 +906,6 @@ SetupResult setup(int argc, char** argv, Options* zopts)
                dns_mgr->InitPostScript();
                trigger_mgr->InitPostScript();
 
-#ifdef USE_PERFTOOLS_DEBUG
-               }
-#endif
                set_signal_mask(false);
 
                if ( reporter->Errors() > 0 )
@@ -1112,6 +1109,10 @@ SetupResult setup(int argc, char** argv, Options* zopts)
                        g_frame_stack.pop_back();
                        }
 
+#ifdef USE_PERFTOOLS_DEBUG
+                }
+#endif
                if ( zeek_script_loaded )
                        {
                        // Queue events reporting loaded scripts.
```

timwoj:
Fixes #3163

### DIFF
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -774,6 +774,20 @@ SetupResult setup(int argc, char** argv, Options* zopts)
 	dbl_histogram_metric_type = make_intrusive<OpaqueType>("dbl_histogram_metric");
 	dbl_histogram_metric_family_type = make_intrusive<OpaqueType>("dbl_histogram_metric_family");
 
+	// After spinning up Broker, we have background threads running now. If
+	// we exit early, we need to shut down at least Broker to get a clean
+	// program exit. Otherwise, we may run into undefined behavior, e.g., if
+	// Broker is still accessing OpenSSL but OpenSSL has already cleaned up
+	// its state due to calling exit(). This needs to be defined here before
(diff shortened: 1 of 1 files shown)


awelzel (approved):
That works, my only worry here is that calling `broker->Terminate()` before `broker_mgr->InitPostScript()` will blow up and the lambda nicely avoided that, but

PR #3179, opened 2023-07-07
Add &default_insert attribute for tables

awelzel:
This is based on the discussion in zeek/zeek#2668. Using &default with tables
can be confusing as the default value is not inserted. The following example
prints an empty table at the end even new Service records was instantiated.

    type Service: record {
        occurrences: count &default=0;
        last_seen: time &default=network_time();
    };

    global services: table[string] of Service &default=Service();

    event zeek_init()
        {
        services["http"]$occurrences += 1;
        services["http"]$last_seen = network_time();

        print services;
        }

Changing above &default to &default_insert will insert the newly created
default value upon a missed lookup and act less surprising.

Docs PR: https://github.com/zeek/zeek-docs/pull/201


awelzel:
@vpax / @ckreibich / @rsmmr - as of naming: I like the "insert" part. It's also what the docs use for adding to a table. Either `&insert_default` or `&default_insert`. Went with the latter for now as in "default and insert".  If neither rings or looks nice, `&create_default` or any other proposal. For "create", I'm reading as "creating a default value" and that also happens for `&default` or "creating entry in table", but less obvious.

I do think the addition is useful. Just writing tests and comparing usage a bit, `&default=container()` on tables acts surprising and little useful.

One case where it came and a better &default would have helped: https://github.com/zeek/zeek/pull/2823#discussion_r1120646438

JustinAzoff:
ah!  

I never could understand why &default worked sometimes and not others.  Makes total sense now.

Things like `base/files/pe/consts.zeek` that do like

```
    const directories: table[count] of string = {
        [0]  = "Export Table",
        ...
        [15] = "Reserved"
    } &default=function(i: count):string { return fmt("unknown-%d", i); };
```

would want to continue using `default` but I think every time I've used `&default` I really wanted `&default_insert`.

Does `&default` only mostly work now for tracking things like counts because

```
++mytable[foo]
```

turns into
```
mytable[foo] = mytable[foo] + 1
``` 

?

awelzel:
>     } &default=function(i: count):string { return fmt("unknown-%d", i); };

Thanks for pointing that out. Vern brought up the same case in the discussion and why we probably shouldn't change `&default` to avoid surprising state growth.

> Does `&default` only mostly work now for tracking things like counts because

Yep. That should be it.

vpax:
> Went with the latter for now as in "default and insert"

I like `&default_insert`!

NEWS
```diff
@@ -31,6 +31,11 @@ New Functionality
   the break statement within ``assertion_failure()`` or ``assertion_result()``
   allows to suppress the default message.
 
+- Add a new ``&default_insert`` attribute for tables. This behaves as ``&default``
+  with the addition that the default value is inserted into the table upon a
+  failed lookup. Particularly for tables with nested container values, the
+  ``&default`` behavior of not inserting the value little useful.
```
rsmmr, line 37:
```suggestion
  ``&default`` behavior of not inserting the value can be of little use.
```

src/Attr.cc
```diff
@@ -359,8 +363,24 @@ void Attributes::CheckAttr(Attr* a)
...
 
+		case ATTR_DEFAULT_INSERT:
+			{
+			if ( ! type->IsTable() )
+				{
+				Error("&default_insert only applicable to tables");
+				break;
+				}
+			}
+			// fallthrough
```
rsmmr, line 374: I'd prefer copying the the call to `check_default_attr` here instead of the fall-through. The fall-through  makes code reading harder and is easy to miss when working on the code in the future (like adding another new attribute). (And seems the  check for having both attributes is fine to have at only one location.)

src/Val.cc
```diff
@@ -1878,6 +1878,10 @@ void TableVal::SetAttrs(detail::AttributesPtr a)
 		broker_store = c->AsStringVal()->AsString()->CheckString();
 		broker_mgr->AddForwardedStore(broker_store, {NewRef{}, this});
 		}
+
+	// Cache &default or &default_insert attribute.
+	if ( def_attr = attrs->Find(detail::ATTR_DEFAULT); ! def_attr )
+		def_attr = attrs->Find(detail::ATTR_DEFAULT_INSERT);
```
rsmmr, line 1884: General question about this: why do we need to cache the attribute given that we're already caching the value itself  in `def_val` as well? Is that only so that now we can say whether the default is of the new insert-type? If so, I'm wondering if there's another way to store that bit of information, and/or if it's worth doing the caching at all. I'm worried about making table val instances even larger than they already are; there can be many for some nested tables.

rsmmr (commented):
Looks all good, with some minor comments. 

In https://github.com/zeek/zeek/discussions/2668#discussioncomment-4990153, Christian suggested to make doc changes part of this, which makes sense I think.

src/Val.cc
awelzel: The  commit adding it came from the opposite angle: Removing repeated `GetAttr()` lookups in favor of caching/storing it once accepting the increased memory usage. I'll rework/revert to back to repeated lookups as this is what we do elsewhere.

rsmmr (approved):

PR #3181, opened 2023-07-08
Migrated some raw pointers to smart pointers

vpax:
This PR gets rid of old-style `StmtPList`s and replaces them with `std::vector<StmtPtr>`.  It also upgrades uses of `WhenInfo` pointers to `std::shared_ptr<WhenInfo>`. Both of these changes are laying groundwork for a much more involved set of changes aimed at reducing the memory footprint of ZAM-compiled scripts (which currently is quite high due to vestigial ASTs that could in fact be discarded ... if things are memory safe). (FYI, in the process of developing the latter I ran across some significant ZAM problems, so PRs for fixing those will likely come first.)

### DIFF
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -1677,7 +1677,7 @@ ValPtr StmtList::Exec(Frame* f, StmtFlowType& flow)
 	RegisterAccess();
 	flow = FLOW_NEXT;
 
-	for ( const auto& stmt_ptr : Stmts() )
+	for ( const auto& stmt_ptr : stmts )
 		{
 		auto stmt = stmt_ptr.get();
 
@@ -1702,7 +1702,7 @@ ValPtr StmtList::Exec(Frame* f, StmtFlowType& flow)
 
 bool StmtList::IsPure() const
 	{
-	for ( const auto& stmt : Stmts() )
+	for ( const auto& stmt : stmts )
 		if ( ! stmt->IsPure() )
 			return false;
 	return true;
@@ -1727,7 +1727,7 @@ void StmtList::StmtDescribe(ODesc* d) const
 			d->NL();
 			}
 
-		for ( const auto& stmt : Stmts() )
+		for ( const auto& stmt : stmts )
 			{
 			stmt->Describe(d);
 			d->NL();
@@ -1743,7 +1743,7 @@ TraversalCode StmtList::Traverse(TraversalCallback* cb) const
 	TraversalCode tc = cb->PreStmt(this);
 	HANDLE_TC_STMT_PRE(tc);
 
-	for ( const auto& stmt : Stmts() )
+	for ( const auto& stmt : stmts )
 		{
 		tc = stmt->Traverse(cb);
 		HANDLE_TC_STMT_PRE(tc);
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -576,12 +576,12 @@ static StmtTag get_last_stmt_tag(const Stmt* stmt)
 		return stmt->Tag();
 
 	const StmtList* stmts = stmt->AsStmtList();
-	int len = stmts->Stmts().length();
+	auto len = stmts->Stmts().size();
 
 	if ( len == 0 )
 		return STMT_LIST;
 
-	return get_last_stmt_tag(stmts->Stmts()[len - 1]);
+	return get_last_stmt_tag(stmts->Stmts()[len - 1].get());
 	}
 
 class FallthroughFinder : public TraversalCallback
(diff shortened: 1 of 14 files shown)


src/script_opt/Stmt.cc
```diff
@@ -971,7 +951,7 @@ void WhenInfo::UpdateIDs(Reducer* c)
 
 StmtPtr WhenStmt::Duplicate()
 	{
-	return SetSucc(new WhenStmt(new WhenInfo(wi)));
+	return SetSucc(new WhenStmt(std::make_shared<WhenInfo>(wi.get())));
```
timwoj, line 954: Making a second shared_ptr here but pointing at the same block of memory can be dangerous one of the two shared_ptrs goes fully out of scope and deletes it's memory. Is there any reason to not just pass the shared pointer down? It'd get `move`d but `WhenStmt`'s constructor already took a copy of the shared pointer (and updated the ref count) because it was passed by value.

src/Stmt.cc
```diff
@@ -1720,10 +1713,10 @@ void StmtList::StmtDescribe(ODesc* d) const
 	if ( ! d->IsReadable() )
 		{
 		AddTag(d);
-		d->AddCount(stmts->length());
+		d->AddCount(stmts.size());
 		}
 
-	if ( stmts->length() == 0 )
+	if ( stmts.size() == 0 )
```
timwoj, line 1719:
```suggestion
	if ( stmts.empty() )
```

src/script_opt/Stmt.cc
```diff
@@ -724,7 +713,7 @@ void StmtList::Inline(Inliner* inl)
 
 bool StmtList::IsReduced(Reducer* c) const
 	{
-	int n = Stmts().length();
+	auto n = Stmts().size();
```
timwoj, line 716: Why does `StmtList` call it's own `Stmts()` method if it has direct access to the object that method is returning already? This is kind of a general question across all of the code in `StmtList`.

timwoj (changes requested):

vpax:
@timwoj The latest push should address all of the above issues.

src/Stmt.h
```diff
@@ -481,23 +481,18 @@ class StmtList : public Stmt
...
 	// Optimization-related:
-	bool ReduceStmt(int& s_i, StmtPList* f_stmts, Reducer* c);
+	bool ReduceStmt(int& s_i, std::vector<StmtPtr>& f_stmts, Reducer* c);
 
-	void ResetStmts(StmtPList* new_stmts)
-		{
-		delete stmts;
-		stmts = new_stmts;
-		}
+	void ResetStmts(std::vector<StmtPtr>& new_stmts) { stmts = std::move(new_stmts); }
```
awelzel, line 495: Nothing to address, just staring ad references and std::move(). This is a bit of a funky API because the caller's vector is valid but unspecified (concretely empty) after a call which could be surprising.

It's not part of the public API (protected) and only used in script_opt so no reason to address, but wondering if a value arg may be good enough or if this is so performance critical? Anyhow, nothing to do.

src/Stmt.h
```diff
@@ -691,8 +686,7 @@ class WhenStmt final : public Stmt
 	{
 public:
 	// The constructor takes ownership of the WhenInfo object.
```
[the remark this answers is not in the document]
vpax: Thanks for catching!  Will elide.

src/Stmt.h
vpax: I guess I don't have a good mental model of just what `std::move` does. I would think that above it does a copy (and then clears `new_stmts`, so if I passed `new_stmts` as a non-reference there would be two copies, one onto the call stack and one from it. Is that not right?

Regarding performance critical, that method does indeed get banged on a lot, every time a sequence of statements changes while transforming the AST to reduced form, and because some of the transformations compound, this is effectively done over and over until there are no more changes.

bbannier: `std::move` is just a cast from e.g., `T` to a `T&&`. In this case a `T&&` on the RHS of the assignment allows `stmts` to directly take over the ownership of the backing storage behind `new_stmts` so that the assignment is just pointer assignments inside `vector`. After that `new_stmts` is a in a "valid but undefined state" so you cannot really do much with it anymore but e.g., assign to it again.

I believe moving from the arg is a good idea here, but since you very likely do not need the moved from arg in the calling scope, so I would instead accept a `std::vector<StmtPtr>` instead of a reference, and  `std::move` the arg at the call sites. That `std::move` will cause creation of a arg value from an rvalue which does the same efficient internal pointer assignments without deep copy like at the `std::move` here.

awelzel: > If a different name would help with the API funkiness, I'm fine with that. Haven't been able to come up with one myself. (Or I could just add a heads-up comment, though it's sort of directly apparent if you're looking at the declaration given the inlined definition.)

I think it's fine due to the special use, detail namespace and protected. Maybe trying Benjamin's suggestion, but the comment mainly stemmed from me derailing into "what if this was a public API / commonly used API".

vpax: Thanks for the sketch - just pushed the better approach.

timwoj (approved):

src/script_opt/Stmt.cc
vpax: FYI @timwoj this is actually different than it looked - the point isn't to duplicate the pointer, it's to create a new `WhenInfo` object that's initialized using the existing one.

Issue #3122, opened 2023-06-14
&default function on unused tables reported as unused


awelzel:
> Thanks @awelzel for doing a direct examination. Given the confusing nature of the message, I agree that ideally it'd be suppressed.  Feel free to file an issue on it and I'll take a look in my Copious Spare Time.  Also seems fine to just go with `&is_used` for now ...

_Originally posted by @vpax in https://github.com/zeek/zeek/pull/3114#discussion_r1229639810_

The usage analyzer reports a `&default` function of a never accessed table as unused. This is a bit confusing for users. Work-around is to put `&is_used` on them
```
$ cat test.zeek 
module MyRadius;

const ids: table[count] of string = {
        [1] = "One",
} &default=function(c: count): string { return fmt("unknown-%d", c); };
$ zeek test.zeek 
warning in ./test.zeek, line 5: non-exported function does not have any callers (MyRadius::lambda_<17123251138430341678>)
```

PR #3183, opened 2023-07-10
Fix to avoid complaining about attribute lambdas as "unused"

vpax:
Closes https://github.com/zeek/zeek/issues/3122.

### DIFF
--- a/src/script_opt/UsageAnalyzer.cc
+++ b/src/script_opt/UsageAnalyzer.cc
@@ -122,8 +122,12 @@ void UsageAnalyzer::FindSeeds(IDSet& seeds) const
 
 		// If the global is exported, or has global scope, we assume
 		// it's meant to be used, even if the current scripts don't
-		// use it.
-		if ( id->IsExport() || id->ModuleName() == "GLOBAL" )
+		// use it.  Similarly, lambdas are either attributes, in
+		// which case we should presume they're used, or appear inside
+		// of functions, so if not used there will be a broader
+		// complaint about the encompassing function.
+		if ( id->IsExport() || id->ModuleName() == "GLOBAL" ||
+		     std::string(id->Name()).find("lambda_<") != std::string::npos )
 			seeds.insert(id.get());
 		}
 	}


src/script_opt/UsageAnalyzer.cc
```diff
@@ -122,8 +122,12 @@ void UsageAnalyzer::FindSeeds(IDSet& seeds) const
...
 		// If the global is exported, or has global scope, we assume
 		// it's meant to be used, even if the current scripts don't
-		// use it.
-		if ( id->IsExport() || id->ModuleName() == "GLOBAL" )
+		// use it.  Similarly, lambdas are either attributes, in
+		// which case we should presume they're used, or appear inside
+		// of functions, so if not used there will be a broader
+		// complaint about the encompassing function.
+		if ( id->IsExport() || id->ModuleName() == "GLOBAL" ||
+		     std::string(id->Name()).find("lambda_<") != std::string::npos )
```
awelzel, line 130: Hey - I think this works for ones dinged in #3122 and probably suffices to avoid false positives, think that's good to merge unless you see an obvious fix for below.

Poking at the fix a bit harder, the following would be cases that aren't covered, but can see them as not idiomatic and probably can ignore. The &default usage on the record field is a bit surprising though, because that should be found, no?

```
module MyRadius;

function default_id(c: count): string {
        return fmt("unknown-%d", c);
}

const other_ids: table[count] of string = {
        [1] = "One",
} &default=default_id;

function default_id2(): string {
        return "";
}

const more_ids: table[count] of string = {
        [1] = "One",
} &default=default_id2() + "";


function gen_id(): string {
        return cat(rand(10000));
}

type R: record {
        id: string &default=gen_id();
};

event zeek_init() {
        print R();
}

warning in ./x.zeek, line 3: non-exported function does not have any callers (MyRadius::default_id)
warning in ./x.zeek, line 11: non-exported function does not have any callers (MyRadius::default_id2)
warning in ./x.zeek, line 20: non-exported function does not have any callers (MyRadius::gen_id)
[id=3890]
```

vpax: Dang, I clearly wasn't thinking about the problem in detail.  Let me mull over how hard it is to fix things correctly so it takes care of those, too.

awelzel: Vern / me chatted and have an idea for a more general fix, will update the PR tomorrow and see if that fits in well.

awelzel:
Closing. Opened #3186 as alternative though more verbose.

PR #3186, opened 2023-07-12
UsageAnalyzer: Collect identifiers found in attributes as seeds

### LINKED ISSUE #3122: &default function on unused tables reported as unused

awelzel:
This marks every identifier used within an attribute as seeds. The scenario this avoids is functions referenced through attributes on unused tables or record types (&default, &expire_func, ...) being dinged as unused as that's rather confusing.

Also adds test for the above and a light smoke test into language/ as it doesn't appear we had coverage here.

Closes #3122


awelzel:
@vpax - rather verbose extra Traverser approach: I tried using ProfileFunc but it acts a bit unexpected when an attribute expression is a lambda expression directly (doesn't yield its identifier), or when dealing with indirect calls (though that is a real corner case). I did switch to a "just collect all identifiers used within attributes" as you hinted at though.

Does that seem okay for now?

Also added some test coverage for the cases discussed. If you're thinking of a different approach/re-work down the road, at least those would help.

vpax:
> Does that seem okay for now?

Yep, this looks good. Bummed to hear ProfileFunc is close-but-not-there. The alternative would be to fix it to deal with the cases you mention (which might be in fact eventually need fixing for script optimization purposes), but I'm fine going with the standalone approach you've settled on.

timwoj (approved):

PR #3188, opened 2023-07-12
Add syntactic sugar for converting between vector and set

timwoj:
This PR adds support for converting between sets and vectors with constructs like:

```
local v1 = vector(1, 1, 1, 2, 2, 3, 3, 4);
local s1 = set(v1);

local s2 = set(1, 2, 3, 4);
local v2 = vector(s2);
```

For the set-to-vector, it only allows conversions for sets that have a single index type. Multiple index types will result in an error. It also does not allow for direct assignments like:

```
local v1 = vector(1, 1, 1, 2, 2, 3, 3, 4);
local s1: set[count] = v1;
```

This continues to result in an error.

@ynadji Is this kind of what you where thinking?


awelzel:
Hmm. Aren't both of test valid constructs today:

awelzel:
The one thought I had was adding unary `*` for splatting/unpacking: `set(*v)` and `vector(*s)` (or spread operator from Javascript `set(...v)`) and special case. But not sure it's worth it, and opens up. If any-style BiFs are acceptable, these could possibly be explored in an external plugin first as well.

JustinAzoff:
I recently came across something like this recently using input framework + paraglob where I had to do

```
        global dns_query_droplist: set[string] = set();
...

	local res: vector of string;

	for ( query in dns_query_droplist )
		res[|res|] = query;

	pg = paraglob_init(res);
```

vpax:
> Hmm. Aren't both of these valid constructs today to constuct `set[vector of X]` or `vector of set[X]`?

Yeah, that is a problem ... :-(  So maybe these need to be BiFs instead.

> `res[|res|] = query;`

(FYI, this can be done more idiomatically these days as `res += query`.)

JustinAzoff:
> (FYI, this can be done more idiomatically these days as res += query.)

Ah, I grabbed that from `policy/frameworks/management/util.zeek` which is what I found when I searched for `set_to_vector`

```
egrep -ri  '\[\|.*\|\].* =' scripts/
```
finds a few other instances.

timwoj:
I apparently didn't consider that this would be breaking existing semantics. I'll reimplement as BIFs.

PR #3198, opened 2023-07-18
-O gen-C++ generated code cleanup

timwoj:
This is a couple fixes for the code generated by `-O gen-C++`:

- Remove all uses of `override final` (which was slightly incorrect anyways) and mark all derived classes as `final` instead.
- Use `std::move` in constructors instead of copying everything, since all arguments are passed by value. This may result in a few unnecessary moves of primitive types, but those cases are minor overall.

It also fixes some issues with the maintenance scripts that were causing them to fail on Linux.


vpax (commented):
LGTM.

PR #3199, opened 2023-07-20
ZAM memory footprint reduction

vpax:
The Microsoft Zeek developers flagged that they're avoiding using ZAM because its memory footprint is too large.  I looked into this by measuring the amount of memory in use after all script processing completes, including potential ZAM compilation. On my Mac, without ZAM the total footprint is 69 MB at that point, and with ZAM it was 184 MB (ouch).

This PR reduces that figure to 79 MB by enabling compiled ASTs to be deleted once they're no longer needed. Along the way, it also introduces some minor streamlining and code cleanup, and fixes an issue with ZAM overlooking some functions that should be compiled that weren't (this fix raises overall ZAM compilation time on the base scripts, because now it's doing more work). Note, one of the significant memory-footprint savings is turning off script coverage analysis when using ZAM. This makes sense, because with ZAM, per-statement profiling doesn't apply given that the statements aren't what's being executed anyway.

The PR requires a tweak to `gen-zam`, per https://github.com/zeek/gen-zam/pull/9.


src/script_opt/ScriptOpt.cc
```diff
@@ -253,15 +249,15 @@ static void optimize_func(ScriptFunc* f, std::shared_ptr<ProfileFunc> pf, ScopeP
 
 	if ( analysis_options.gen_ZAM_code )
 		{
-		ZAM = new ZAMCompiler(f, pf, scope, new_body, ud, rc);
```
timwoj, line 256: Were we never deleting this before?

src/script_opt/ScriptOpt.cc
```diff
@@ -549,6 +545,25 @@ static void analyze_scripts_for_ZAM(std::unique_ptr<ProfileFuncs>& pfs)
 	finalize_functions(funcs);
 	}
 
+void clear_script_analysis()
+	{
+	IDOptInfo::ClearGlobalInitExprs();
+
+	// We need to explicitly clear out the optimization information
+	// associated with identifiers.  The have reference loops with
```
timwoj, line 553:
```suggestion
	// associated with identifiers.  They have reference loops with
```

src/script_opt/ScriptOpt.cc
vpax: Indeed, we weren't. I'm pretty sure this was because doing so led to crashes because of the lack of rigor in tracking AST elements, a broader problem than just ZAM (though it was the bulk of it).

src/script_opt/ScriptOpt.cc
vpax: Fixed!

PR #3203, opened 2023-07-22
change Trigger constructor to not potentially auto-delete itself

vpax:
The `Trigger` constructor has peculiar behavior in that it auto-`Unref()`s itself, which means that callers have used bare `new` expressions to construct them and then discard the result, seemingly leaking it (but not actually). This behavior makes it untrustworthy to use the returned pointer for anything, such as calling `SetName()` to associate a name with the trigger (needed by the compile-to-C++ script compiler).

This PR removes this behavior and documents the API as expecting construction via `make_intrusive<>`.

The one drawback is that this change is in non-`detail` space and yet effectively is an API-breaking change, since callers using the old style will now leak the `Trigger`s they create. It seems the only way to fix this is to introduce a _new_ constructor and deprecate the current one; but that new constructor will need to have something different about it, which seems a bit tortuous. I guess it could take some of the arguments in a different order.

Because of this issue, I'm marking this as a draft so we can discuss (@awelzel in particular).


timwoj (approved):
Like I said in Slack DMs, I'm not a huge fan of this whole API in general now that I look at it again. Having the constructor return a pointer that's valid but deactivated seems really weird to me. This fixes the immediate problem of the pointer being invalid though, so maybe we can revisit the API at some point later on.

vpax:
> Having the constructor return a pointer that's valid but deactivated seems really weird to me

One thing I didn't flag: _normally_ the object *is* active. It's only in the case where the condition is immediately true (no deferment necessary) that the object will be inactive. For sure most production uses of `when` don't have the immediately-true property. Because of that, I don't find the API all that peculiar.

awelzel:
> Because of this issue, I'm marking this as a draft so we can discuss (@awelzel in particular).

It's not a breaking change anymore due to a re-work / overload if I follow the conversation properly.

---

Though, I'm now looking and wondering: This seems solely about the "recently" introduced `SetName()` operation not being safe?

And I think there's more to it: There is an existing `loc` parameter for the `Trigger` constructor, but that's apparently completely unused and instead `Obj.location` is used. Maybe that was the motivation for `SetName()` in the first place?

So, assuming it's just about recently introduced `SetName()` not safe: How about adding a new constructor that takes an optional `const char *name`  instead of the unused `Location *loc` and then deprecate the old constructor and the `SetName()` accessor instead. Or, make `loc` actually work and see if that removes the need for `SetName()` within `script_opt/` ?

awelzel (changes requested):
Not so much "Request changes", but with the most recent comment maybe "another look might be worth it".

awelzel:
> Or, make loc actually work and see if that removes the need for SetName() within script_opt/ ?

I chatted a bit with Tim. I hadn't realized the C++ script_opt part formats a [location string into the produced C++ code](https://github.com/zeek/zeek/pull/3203/files#diff-85b5e2f8406d0c193657b888e8a01e54e25af7b0c4a2c064c60e4ea7c0ca4d0aL443). But I guess if the `loc` parameter would work, it could also render a Location object onto the stack and remove the need for `SetName()`.
```
Location loc{...};
Trigger(..., &loc);
```

timwoj:
> First, I was wrong in my comment that this is in non-detail space. It's in fact in zeek::detail::trigger:: space. (Not sure why an interior trigger namespace is needed - style appears to be inconsistent here.)

Ugh. I'm not sure why we did that, but it definitely should be `zeek::trigger::detail` to match everything else. We can fix it and alias it back temporarily, with a deprecation warning on the alias.

vpax:
> I'm not sure why we did that, but it definitely should be `zeek::trigger::detail` to match everything else

Either way, though, this means we don't need to support a deprecated interface, right?  More generally, are you good with the proposed plan above (which looks like Arne is)?

timwoj:
> > I'm not sure why we did that, but it definitely should be `zeek::trigger::detail` to match everything else
> 
> Either way, though, this means we don't need to support a deprecated interface, right? More generally, are you good with the proposed plan above (which looks like Arne is)?

The plan looks good to me. 👍🏼

vpax:
Okay, I've pushed a `fixup` that incorporates the plan as sketched above, except (per a quick side-discussion with Tim) I left in the deprecated old constructor, as that was simple to do and might help someone who's been messing around in `detail` space.

timwoj (approved):

timwoj:
I talked to @awelzel on Slack and he didn't have any other comments. I'll squash this and merge it.

awelzel (approved):

Issue #3184, opened 2023-07-11
Modbus: Implement Diagnostics (0x08) parsing (and other extensions)


awelzel:
@biswajitutil in the forum asked to support parsing the diagnostics function code:

https://community.zeek.org/t/modbus-parser-enhancement/7079/5

He provided a link to a SO post of someone dissecting this over TCP:
https://reverseengineering.stackexchange.com/questions/30386/interpreting-response-for-modbus-tcp-function-code-0x08-diagnostics

The large `modbus.trace` file we have in the repo actually contains examples of this code being used over TCP:
https://github.com/zeek/zeek/blob/master/testing/btest/Traces/modbus/modbus.trace

Additional comments were the following, though could consider splitting into separate tickets.

> we may need to support the way to get pdu length in case of read file record ( func code 0x14)
>
> MBAP(modbus applucation PDU) header’s transaction ID, data length, execution code, address if any etc.
>
> There should be support extracting delimeter from PDU in case of func code 0x8 and sub function code 0x01
>
> For file related access ( read or write), we need to extract File number, therefore if we know the function code, there should be option to get all possible values it returns alongside . ( eg.
If FC == 43 and sub FC == 14 :
Source address = extract from PDU)

Reference: https://modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf

biswajitutil:
Thanks @awelzel ! 
So just want to two points
1. Subfunc code available with func code 0x8(diagonistics) and 0x2b( Encapsulated Interface Transport) in modbus TCP.
2. PDU length should come with all type of read/write ops. 
Lastly, can we add write_quantity similar to read_quantity in [modbus_read_write_multiple_registers_request](https://docs.zeek.org/en/master/scripts/base/bif/plugins/Zeek_Modbus.events.bif.zeek.html#id-modbus_read_write_multiple_registers_request). I am not sure why write_quantity is not there.

Thanks
Biswa

timwoj:
> Subfunc code available with func code 0x8(diagonistics) and 0x2b( Encapsulated Interface Transport) in modbus TCP.

Does the data in the requests and responses matter or do you only care about the subfunction code for the diagnostics and the MEI code for the encapsulated transport?

biswajitutil:
[...]

PR #3204, opened 2023-07-26
Modbus extensions

timwoj:
This PR adds support for the following:

- Diagnostics requests and responses (FC=0x08)
- Encapsulation interface transport requests and responses (FC=0x2B)
- PDU length values for all packets
- Parsed record data for ReadFile/WriteFile events. We already had the parsing there for this data, but we weren't sending it on to the events.

There are two commits for diagnostics. One just returns the byte string from each message. The second returns a record differentiating the different possible fields (byte data, delimiter, and counter values). I couldn't decide which was best in that regard, so both are there for now. We can squash or drop one of them after discussion. Another possibility is to separate those three types of data into different events.

This covers most of the requests in #3184.


src/analyzer/protocol/modbus/events.bif
```diff
@@ -294,3 +294,42 @@ event modbus_read_fifo_queue_request%(c: connection, headers: ModbusHeaders, sta
...
 event modbus_read_fifo_queue_response%(c: connection, headers: ModbusHeaders, fifos: ModbusRegisters%);
 
+## Generated for a Modbus Diagnostics request.
+##
+## c: The connection.
+##
+## headers: The headers for the modbus function.
+##
+## diag: The diagnostics data, including subfunction-specific values.
+event modbus_diagnostics_request%(c: connection, headers: ModbusHeaders, diag: ModbusDiagnostics%);
```
awelzel, line 304: I can see the motivation for the `ModbusDiagnostics` record for extensible, but could also see a more primitive one instead not doing the conversion of `data` to stick closer to what the protocol is. Should be possible to build conversion above.
```
modbus_diagnostics_request(c: connection, headers, subfunction: count, data: string);
modbus_diagnostics_response(c: connection, headers, subfunction: count, data: string)
```

Then in script-land can render subfunction into a string representation (via const table) and convert `data` to count via `bytestring_to_count` or delimiter if wanted given subfunction.

src/analyzer/protocol/modbus/modbus-analyzer.pac
```diff
@@ -355,6 +355,154 @@ refine flow ModbusTCP_Flow += {
 		%}
 
 
+	# REQUEST FC=8
+	function deliver_DiagnosticsRequest(header: ModbusTCP_TransportHeader, message: DiagnosticsRequest): bool
+		%{
```
awelzel, line 360: As above, if we used a lower-level event, the conversions (and possibly also weird generations) could be done in Zeek script. Think the analyzer shouldn't have as much logic as it has now in the request/response functions and rather do this in the script layer.

The weird generation could possibly be converted to `&enforce` annotations and actually trigger analyzer violations or not encoded?

Mainly that we're not raising events if something is weird based on rules hard-coded in the analyzer seems a bit unfortunate.

awelzel (commented):
Hey @timwoj - thanks for picking this up, hope the comments are useful.

src/analyzer/protocol/modbus/events.bif
```diff
@@ -193,7 +193,7 @@ event modbus_write_multiple_registers_response%(c: connection, headers: ModbusHe
 ##
 ## .. note: This event is incomplete.  The information from the data structure
 ##          is not yet passed through to the event.
-event modbus_read_file_record_request%(c: connection, headers: ModbusHeaders%);
```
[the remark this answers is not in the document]
timwoj: Added definitions for the ones that I replaced. It's unfortunate that we don't have any test data for those.

scripts/base/init-bare.zeek
```diff
@@ -4436,8 +4436,48 @@ type ModbusHeaders: record {
...
+};
+
+type ModbusFileRecordResponses: vector of ModbusFileRecordResponse;
+
+type ModbusFileReference: record {
+	ref_type: count;
+	file_num: count;
+	record_num: count;
+	word_count: count;
+	register_value: string;
```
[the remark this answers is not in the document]
timwoj: Aligned both the pac and script-land code to match the spec.

src/analyzer/protocol/modbus/events.bif
timwoj: Yeah, that's what I had originally in the first commit, and switched to the record version later. It's easy enough to revert the second commit.

src/analyzer/protocol/modbus/modbus-analyzer.pac
timwoj: > The weird generation could possibly be converted to &enforce annotations and actually trigger analyzer violations or not encoded?

This is tough to do because the various validations depend what subfunction is selected. I could split it up into individual types in the .pac file, but that seems overkill when I can do the same validation on the C++ side.

> Mainly that we're not raising events if something is weird based on rules hard-coded in the analyzer seems a bit unfortunate.

The weirds being raised are based on the spec itself, so I'm not sure what you mean.

scripts/base/init-bare.zeek
```diff
@@ -4436,8 +4436,48 @@ type ModbusHeaders: record {
 	uid:           count;
 	## MODBUS function code
 	function_code: count;
+	## Full length of modbus PDU in bytes
+	pdu_length:    count;
```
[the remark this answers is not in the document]
timwoj: The length field in the header is the length of the data immediately following that field. The requested PDU length field is basically the length of everything after that header, which is a single byte field. I could see just removing the `pdu_length` field altogether (reverting that commit) and telling the user to just calculate it themselves from the `len` field in the header.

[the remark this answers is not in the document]
awelzel: > and telling the user to just calculate it themselves from the len field in the header.

That's what I meant, yes, include the original `len` rather than the derived field (still need to extend `ModbusHeaders` by `len`).

src/analyzer/protocol/modbus/modbus-analyzer.pac
awelzel: > The weirds being raised are based on the spec itself, so I'm not sure what you mean.

Would just always raise the `modbus_diagnostics_request` and response event, even if not spec conform but otherwise parsable.

A non-conformant  or development device with a new subfunction code added would not generate a diagnostics, running minimally into the `default` case even if structurally the request/response is valid and could be raised to script-land.

```
428                                 default:                                                                                                                                                                                                                                                                      
429                                         zeek::reporter->Weird("modbus_diag_unknown_request_subfunction",                                                                                                                                                                                                      
430                                                               zeek::util::fmt("%d", ${message.subfunction}));                                                                                                                                                                                                 
431                                         return false;                                                                                                                                                                                                                                                         
432                                 } 
```

If data format of known subfunction codes doesn't agree with spec, still possibly interesting to get those event into script-land. Also easier to extend/change the policy the in script layer - thinking of a `modbus_diagnostics.log` or so.

src/analyzer/protocol/modbus/events.bif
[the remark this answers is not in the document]
awelzel: Sorry, unresolving. Putting them into the `export {}` section of `Modbus` makes them available in `Modbus::`, not `GLOBAL::` - they also wouldn't be available in bare mode thought that would be minor.

Putting them directly under under the existing one in `events.bif` does it.

```
 event modbus_read_file_record_request%(c: connection, headers: ModbusHeaders, byte_count: count, refs: ModbusFileRecordRequests%);
+event modbus_read_file_record_request%(c: connection, headers: ModbusHeaders%) &deprecated="Remove in v7.1. Use the version that takes a byte_count and vector of references";
```

From here 15b294098c8c969af2d90dac6b78565bd0789f9f ;-)

scripts/base/init-bare.zeek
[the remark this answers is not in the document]
timwoj: Gotcha, implemented that.

src/analyzer/protocol/modbus/events.bif
timwoj: Reverted the second commit.

src/analyzer/protocol/modbus/modbus-analyzer.pac
timwoj: I don't feel super strongly about it, but we do have a bunch of prior places where we do the same thing with other analyzers (see, for example, all of the weirds reported by the GRE analyzer). They're not checking for violations of the parsing, but for violations of the data encountered while parsing, and bailing out of the analyzer at that point.

scripts/base/init-bare.zeek
```diff
@@ -4436,8 +4436,39 @@ type ModbusHeaders: record {
...
+	record_data: string;
+};
+
+type ModbusFileRecordResponses: vector of ModbusFileRecordResponse;
+
+type ModbusFileReference: record {
+	ref_type: count;
+	file_num: count;
+	record_num: count;
+	record_length: count;
```
awelzel, line 4465: One more nit, to align with the other `len` fields.

```suggestion
	record_len: count;
```

scripts/base/init-bare.zeek
```diff
@@ -4436,8 +4436,39 @@ type ModbusHeaders: record {
 	uid:           count;
 	## MODBUS function code
 	function_code: count;
+	## Length of the application PDU following the header plus
+	## one byte for the uid field.
+	len:           count;
```
awelzel, line 4441: Nit, move `len` just before `uid` as it appears on the wire.

src/analyzer/protocol/modbus/modbus-analyzer.pac
awelzel: Me neither. GRE might be historical due to being a packet analyzer? Compromise could be to generate the weird but fall through to raising the event regardless.

Then, if there's feedback that either isn't wanted (or weird generation in script-land allows easier customization) revisit.

Leaving it to you :-)

awelzel (approved):

timwoj:
Merged in 30b2afe2ad6308941a14b9fc153a3561054ef7fe

Issue #3205, opened 2023-07-27
UDP: padding incorrectly sent to downstream analyzers


0xxon:
2b9de839b0948c7de3eb5ed4a397194f96aae6b5 (https://github.com/zeek/zeek/pull/3080) introduced passing truncated data to downstream analyzers.

This has the unfortunate side-effect of also passing ethernet-level padding that is added after the UDP data to the downstream analyzers. This was raised on Slack in https://zeekorg.slack.com/archives/CSZBXF6TH/p1690340258929719.

While, in this case, this might not seem like a huge problem, there are setups in which local equipment does put information (that should be ignored) after the packet payload; an example for this are switch-inserted timestamps.

Slack message contains an example pcap; this behavior should be observable with any UDP packet that has a payload length that would change the ethernet frame length to under 60 bytes.

This is a regression in behavior and should definitely be fixed in 6.0; earlier versions should not be affected.

In the pcap provided on Slack, this bug causes an incorrect `syslog.log` to be created.

cc-ing @awelzel

PR #3206, opened 2023-07-27
Do not forward more than the remaining data to downstream UDP analyzer

0xxon:
This fixes a bug introduced in 2b9de839b0948c7de3eb5ed4a397194f96aae6b5 / GH-3080, which causes UDP padding to be sent to UDP based analyzers.

Fixes GH-3205.

Please credit "Hiep Long Ton" with finding this issue (see https://zeekorg.slack.com/archives/CSZBXF6TH/p1690508032030229?thread_ts=1690340258.929719&cid=CSZBXF6TH)


awelzel (approved):
Ouch, yes. I've put this on the 6.0.1 list.

Suspect the same should be done for TCP, too, if only for consistency:
https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/tcp/TCP.cc#L133

PR #3213, opened 2023-08-02
Do not forward padding to downstream TCP packet analyzer

0xxon:
This is similar to GH-3206. There do not seem to be practical consequences - but we should still fix it.

This also includes the udp-testcase that was forgotten in GH-3206.

We should probably also include this one in 6.0.


testing/btest/core/tcp-padding.zeek
```diff
@@ -0,0 +1,6 @@
+# Check that data contained in the UDP padding does not make it into protocol analysis
```
awelzel, line 1: Minor nit, suspect both of these are about ethernet padding.

```suggestion
# Check that data contained in the ethernet padding does not make it into protocol analysis
```

awelzel (approved):
Mind leaving a note on the tooling/source for the TCP pcap. Mostly for my information :-)

testing/btest/core/tcp-padding.zeek
0xxon: Thanks :)

0xxon:
The tooling was me, the pcap file spec, and a hex editor. I am not sure that's worth mentioning :)

PR #3215, opened 2023-08-03
Remove uses of cmake/FindRequiredPackage

timwoj:
This PR updates all of the submodules to deprecate/remove the FindRequiredPackage CMake macro in favor of direct calls to find_package(). This also updates a few uses in the Zeek repo itself.


rsmmr (approved):

timwoj:
I'm going to revert the python version increase, since it breaks some of the builds on CI.

Issue #2749, opened 2023-02-01
ThreadSanitizer reports a number of failures


timwoj:
This is a follow-on to #2057. 

The ThreadSanitizer job is running on CI again, but there were a bunch of failures in various tests while getting it working again. https://github.com/zeek/zeek/pull/2748 added the `ci/tsan_suppressions.txt` file that lists out the known failures. They're currently suppressed in CI builds so that the job passes, but we should investigate all of them to see if they can be fixed.

PR #3217, opened 2023-08-04
Upgrade ThreadSanitizer CI build to Ubuntu 22.04

timwoj:
This fixes the current build failure on CI with 20.04 after the Dockerfile upgrade in https://github.com/zeek/zeek/pull/3210. This also includes some new suppressions for new findings reported after the compiler upgrade. I opted to add the suppressions since we already have some there. They need to be investigated as part of https://github.com/zeek/zeek/issues/2749.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -390,7 +390,7 @@ ubsan_sanitizer_task:
 tsan_sanitizer_task:
   container:
     # Just uses a recent/common distro to run memory error/leak checks.
-    dockerfile: ci/ubuntu-20.04/Dockerfile
+    dockerfile: ci/ubuntu-22.04/Dockerfile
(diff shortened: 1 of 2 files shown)


awelzel (approved):
The new warnings sound scary, but presume these have always been there and also require profiling to be enabled.
```
  WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=229746)
    Cycle in lock order graph: M804872730922138176 (0x000000000000) => M807124530735823464 (0x000000000000) => M804872730922138176
```

timwoj:
> The new warnings sound scary, but presume these have always been there and also require profiling to be enabled.

I don't think we've ever seen one of these trigger in practice, but I assume if threadsanitizer can make them happen it's entirely possible.

PR #3230, opened 2023-08-11
Update zeek-aux submodule [nomail]

timwoj:
Updates zeek-aux to the latest master to pick up cmake FindRequiredPackage changes. This PR is just for CI coverage, but I don't expect it to fail. I'll merge it once it passes.


timwoj:
This failed on the Freebsd 13 build because Cirrus updated their image. The second commit moves from FreeBSD 12.3 to 12.4, and from 13.1 to 13.2. This fixes the build failure on Cirrus.

PR #3231, opened 2023-08-11
Force pre-commit to use python 3.9

timwoj:
This fixes the errors with shfmt-py and python 3.10. It's not a permanent fix, but it works for now until shfmt-py updates.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,10 @@ repos:
   rev: 3.3.1.8
   hooks:
     - id: shfmt
+      # We force python3.9 since as of its current version `shfmt-py` has a
+      # `setup.cfg` not conforming to PEP639 (`license_file` instead of
+      # `license_files`) which triggers a warning with newer Python versions.
+      language_version: "3.9"
       args: ["-w", "-i", "4", "-ci"]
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,9 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 #
+default_language_version:
+  python: python3.9
```
bbannier, line 5: Not a huge fan of this since it will fail if users do not have exactly that version available on their system (`pre-commit` will not install this interpreter, and the error is pretty opaque). Could you at least make it only specific to the shfmt hook, and ideally add a comment for why we do this (might help to have a similar comment in the GH action), e.g.,

```yaml
- repo: https://github.com/maxwinterstein/shfmt-py
  rev: 3.3.1.8
  hooks:
    - id: shfmt
      # We force python3.9 since as of its current version `shfmt-py` has a
      # `setup.cfg` not conforming to PEP639 (`license_file` instead of
      # `license_files`) which triggers a warning with newer Python versions.
      language_version: "3.9"
      args: ["-w", "-i", "4", "-ci"]
```

That way it might become easier for users to pin down any failures.

bbannier (approved):

PR #3237, opened 2023-08-17
IPBasedAnalyzer: Don't flip connections when destination is broadcast

### LINKED ISSUE #3235: Conn.log: Originator and Responder IPs/ports Not Flipping for Logged Responder Packets
### Capture wrong direction of communication

**To reproduce**
Version of zeek LTS: 5.0.9

Just run the pcap file enip_255_trunc.pcapng in attachment. The output is conn.log (also attached).

**Expected behavior**
The broadcast messages should be from 192.168.1.105 not from 255.255.255.255
[conn.log](https://github.com/zeek/zeek/files/12357398/conn.log)

This issue is resolved for icsnpp-enip plugin (https://github.com/cisagov/icsnpp-enip#source-and-destination-fields). 
[conn_Issue.zip](https://github.com/zeek/zeek/files/12357406/conn_Issue.zip)

awelzel:
Closes #3235

---

External baseline changes: https://github.com/zeek/zeek-testing/pull/14


awelzel:
Ouch, this has some ripple effects for the DHCP scripts. I think it's for the better to change this still, but the fixups in dhcp/main.zeek are a bit clunky - happy to hear thoughts if maybe this isn't such a great idea, or it makes sense to take a step back. Background is in #3235.

awelzel:
>  happy to hear thoughts if maybe this isn't such a great idea, or it makes sense to take a step back

I had chatted with Robin and it seemed reasonable to him. Undrafting and assigning him :-)

scripts/base/protocols/dhcp/main.zeek
```diff
@@ -246,12 +251,22 @@ event DHCP::aggregate_msgs(ts: time, id: conn_id, uid: string, is_orig: bool, ms
...
 		# an IP address.
-		if ( msg$yiaddr != 0.0.0.0 &&
-		     id$resp_h != 255.255.255.255 )
+		if ( msg$yiaddr != 0.0.0.0 )
 			{
-			log_info$server_addr = id$resp_h;
-			log_info$server_port = id$resp_p;
-			log_info$client_port = id$orig_p;
+			# If is_client is F and is_orig is T, it's a DHCP server
+			# broadcasting and orig_h is the server_addr, else resp_h.
```
rsmmr, line 257: The code looks good, but I'm wondering if the comment is correct: we're in the server case here already (because `is_client` is false), but couldn't we get here also for a non-broadcast connection initiated by the server?

rsmmr (approved):
I think this looks good, the new DHCP `conn.log` looks intuitive (here and in the external suite), and the tweaked DHCP logic seems right too (except my comment about the comment). Can't think of anything better for the latter either.

scripts/base/protocols/dhcp/main.zeek
awelzel: > but couldn't we get here also for a non-broadcast connection initiated by the server?

Yep. That's the "else resp_h" part from the comment for the second if branch. Server initiated "connections" to non-broadcast will have been flipped, so server is resp_h.  Can split the commentary between the branches if that helps?

rsmmr: Yep, just clarify it a bit.

PR #3238, opened 2023-08-17
pre-commit: Pin latest shfmt-py version

awelzel:
This allows users to run shfmt-py with Python > 3.9.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,13 +8,11 @@ repos:
   - id: clang-format
 
 - repo: https://github.com/maxwinterstein/shfmt-py
-  rev: 3.3.1.8
+  # Latest master revision as of 2023-08-17 allows to run with modern Python
+  # versions. Switch this back to a tag once there is one :-)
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -8,13 +8,11 @@ repos:
   - id: clang-format
 
 - repo: https://github.com/maxwinterstein/shfmt-py
-  rev: 3.3.1.8
+  # Latest master revision as of 2023-08-17 allows to run with modern Python
+  # versions. Switch this back to a tag once there is one :-)
+  rev: d4491d8
```
bbannier, line 13: Did you intentionally leave the pinning of a Python version in place in `.github/workflows/pre-commit.yml`? I think we could get rid of it since we do not really care about a particular version for these hooks now.

bbannier (approved):

.pre-commit-config.yaml
awelzel: Dropped before the merge.

PR #3240, opened 2023-08-18
Forward declare openssl structure instead of including header

timwoj:
@vpax had this come up while trying to build an external plugin that doesn't depend on openssl. It included `MIME.h` but didn't have the include path in the compilation for openssl, so it failed to find the openssl header required. This PR changes MIME.h to forward declare the required openssl structure and moves the actual include to the .cc file instead.

### DIFF
--- a/src/analyzer/protocol/mime/MIME.cc
+++ b/src/analyzer/protocol/mime/MIME.cc
@@ -2,6 +2,8 @@
 
 #include "zeek/zeek-config.h"
 
+#include <openssl/evp.h>
+
 #include "zeek/Base64.h"
 #include "zeek/NetVar.h"
 #include "zeek/Reporter.h"
--- a/src/analyzer/protocol/mime/MIME.h
+++ b/src/analyzer/protocol/mime/MIME.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <openssl/evp.h>
 #include <cassert>
 #include <cstdio>
 #include <queue>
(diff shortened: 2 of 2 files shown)


awelzel (commented):
Hmm - there's more `openssl` includes in our headers and this fixes just one instance.

Specifically the `OpaqueVal.h` header as it contains the opaque hash vals. Seems a plugin using `OpaqueVal.h` would trigger an issue, too?

There's also `ZeekOpenSSLHint` in CMakeLists.txt with the following comment:
```
 751 # Forward user-defined hint for OpenSSL to the plugins. Use a cache variable to 
 752 # make sure this variable survives CMake re-runs. Note: This variable is picked 
 753 # up in ZeekPluginConfig.cmake.in. 
```

So we might have provisions to find/use the right path. Is there a reference/place to see the errors that @vpax was getting? The plugin [zeek-more-hashes](https://github.com/zeek/zeek-more-hashes) uses OpaqueVal, does that compile for you, Vern?

timwoj:
I was hoping this was just something with the recent `find_package` changes I made on master. I tested building zeek-more-hashes against a 6.0 install though and it's failing there too.

timwoj:
> Note: This variable is picked up in ZeekPluginConfig.cmake.in. 

I fixed the issue with OPENSSL_ROOT_DIR not being set before the setting `ZeekOpenSSLHint, but there's further problems. Is `ZeekPluginConfig.cmake` supposed to be installed as part of the install? Right now it isn't, so you have to pass `--zeek-dist` to the plugin `configure` for it to be found. Even then though, we don't add the openssl path to the compile command and it still fails to find the header.

timwoj:
Looking at ZeekPluginConfig.cmake again, I don't think it helps at all. Setting `OPENSSL_ROOT_DIR` helps to ensure that `find_package` will find the same OpenSSL installation that Zeek used. If plugins never call that though, they're still not going to have the include or library paths available at build time.

timwoj:
Closing this PR as this isn't the right fix.

Issue #3242, opened 2023-08-22
Problematic behavior for enabling community id in notice.log


0xxon:
To enable the community ID in notice.log, one has to load `policy/frameworks/notice/community-id` _after_ loading `policy/protocols/conn/community-id-logging`. If one loads the notice script first, logging in `notice.log` will not be enabled, because the script checks if the community-id-logging script was loaded; if it was not, the script does not do anything.

I think we should change this behavor. First - currently `local.zeek` contains the two scripts in the opposite order. If a user just removes the comments to enable community-id logging in `conn.log` and `notice.log`, it will only be enabled in `conn.log`.

Furthermore, it seems weird that the `notice/community-id` script does not just load `conn/community-id-logging` and enable the community ID in both log files; to me this seems like the less problematic/surprising approach.

@ckreibich - I think you originally wrote this; is there some reason behind this choice? If not, I am happy to make a PR to change this.

This was found in this Slack conversation: https://zeekorg.slack.com/archives/CSZBXF6TH/p1692707161580799

ckreibich:
I did the original package, not the move into the distro. The package's reasoning for notices is [here](https://github.com/corelight/zeek-community-id#zeek-notice-support). It didn't auto-load the conn extension when loading the notice one — I think it just wasn't a priority at the time; no other reason.

I agree with both changes.

PR #3243, opened 2023-08-22
Community-id: load main script in notice script

0xxon:
This change makes the community-id script that adds the community id to
notice.log automatically load the main script if this was not already
loaded.

In the past, the script just did not perform any action if the main
script was not loaded.

This change also makes the notice script respect the seed/base64
settings that were set in the main script.

Fixes https://github.com/zeek/zeek/issues/3242


awelzel (approved):
Thanks!

PR #3245, opened 2023-08-23
PPP: Add PPP analyzer to handle LINKTYPE_PPP (0x9)

awelzel:
Using pcaps from https://interop.seemann.io/ as samples for QUIC protocol
data didn't produce a conn.log for the contained data. `tcpdump -r`
and Wireshark do show the contained IP/UDP packets. Teach Zeek how
to handle link type DLT_PPP 0x09 using a new PPP analyzer based on the
PPPSerial analyzer code.

Usual update to files/x509 baseline after adding new analyzer due
to enum values changing.


0xxon (approved):
assuming tests pass

J-Gras:
We have distinct analyzers for PPPSerial and PPPoE. I think it would make sense to either create a new PPP analyzer or merge all into a single PPP analyzer. Given that the idea of the packet framework is to expose composition to the user, I would be in favour of the first option.

awelzel:
> We have distinct analyzers for PPPSerial and PPPoE.

There seemed to overlap overlap to stuff them together, but I'll re-push a split version.

0xxon (approved):
Looks good - could you write a news entry when merging? :)

PR #3246, opened 2023-08-23
Update centos 7, centos stream 8, and opensuse leap 15.4 builds to a newer version of python

### LINKED ISSUE #3239: Require Python 3.7 or greater?
We recently bumped the required version of Python for standalone installs of btest up to Python 3.7. The btest script still runs fine if invoked manually with an older version of Python, since we didn't make a lot of changes that would have required the newer version. As part of the changes to remove uses of `FndRequiredPackage`, I was also bumping our required version of Python in other submodules to 3.7 as well. Unfortunately I ran into some issues with a few of the builds where it would fail to find that version, because the VM on Cirrus (and the base OS) still have some older version of Python by default.

Python 3.6 was marked as end-of-life back in 2021. Python 3.7 was actually marked end-of-life in June 2023 too. Should we consider moving our required version of Python up to something that's supported? That will require fixing the following VMs on Cirrus to have a newer version of Python installed than the stock version:

- ubuntu18 with python 3.6.9
- opensuse leap with python 3.6.15
- centos7 with an unknown version of python, since their old version of cmake doesn't report it. A docker container says it's 3.6.8.
- centosstream8 with python 3.6.8

timwoj:
This updates the listed CI docker images to newer versions of python, as an initial step to resolve #3239 

It also removes the Ubuntu 18 build since that version is EOL as of 2023/05/31.


0xxon (approved):

timwoj:
I made a note to myself to go through and do all of these upgrades on other repositories as well.

Issue #3191, opened 2023-07-13
Appending to vector of any with += crashes: Val::CONVERTER (record/vector)


awelzel:
This appears to have started with Zeek 5.0 and no one complained so far, so marking low priority. Zeek 4.2 doesn't abort on this.

```
# crash2.zeek
type R: record { r: string &default="r"; };
type S: record { s: string &default="s"; };

# any queue
global q: vector of any;

function keep_this(r: any) {
        q += r;  # this crashes since Zeek 5.0
        q[|q|] = r;  # this is fine.
}

event zeek_init()
        {
        keep_this(R());
        keep_this(S());
        print q;
        }

$ zeek crash2.zeek 
fatal error in <no location>: Val::CONVERTER (record/vector) ([r=r])
Aborted (core dumped)
```

vpax:
I have a hunch what's causing this, and if right then it's from a change I introduced, so I'm game to take a look.  That said, if you or someone else would like to, fine by me!

timwoj:
Changing it to `q += [r]` works as well. `BinaryExpr::Fold` is expecting the RHS of the expression to be a vector if the LHS is one, and the error you're seeing is thrown when it fails to convert the RHS object to a vector.

vpax:
I looked into this and it turns out that the core problem is an ambiguity. If `x` is a vector of `T` elements, then `x += y` is add-a-single-element if `y` has type `T`, and add-every-element if `y` has type `vector of T`. In the example above, `r` is of type `any`, so at run-time it could be _either_ a single element of type `T`, _or_ a `vector of T` that should be added element by element. (Or it could be the wrong type, period, leading to a run-time error.) The bug arises because in one context it's interpreted as a single element, and in another it's (incorrectly) assumed to be vector.

It's easy enough to fix this with some dynamic checking ... but it gives me pause to have a single statement that has somewhat varying semantics depending on what argument is passed to the function, as there are (perhaps quite unlikely) logic errors that this could mask. So I think my druthers would be to treat `x += y` for `y` of type `any` as requiring that `y` has (at run-time) a concrete type of `T`, and it'd be an error if it has the type `vector of T`, since I think the latter is surprising when one reads the statement. (If the developer wants the latter then they should declare `y` to have type `vector of any` to guarantee that it's a vector.)

That said, this isn't near-and-dear to my heart, I just want to know the best way to fix it :-).

awelzel:
> I looked into this and it turns out that the core problem is an ambiguity. 

[...]

PR #3249, opened 2023-08-24
Fixes for vectors of type "any" or having values of "any" assigned to elements

vpax:
This closes GH-3191. Along the way I noticed that assignments of `any` values to vector elements could silently fail, so I added reporting of these, and BTest coverage (as well as for the semantics agreed upon - I believe - in the GH-3191 thread).

There's a related PR for updating the documentation per https://github.com/zeek/zeek-docs/pull/207.


testing/btest/language/vector-any-append.zeek
```diff
@@ -19,4 +31,35 @@ event zeek_init()
 	append(v);
 	append(v);
 	print v;
+
+	local v2: vector of any;
+	v2 += |v|;
+	v2 += v;
+	print v2;
```
awelzel, line 38: Reproducing without baseline for illustration:
```
local v: vector of count = vector(1,2,3,4);
local v2: vector of any;
v2 += |v|
v2 += v;
assert v2 === [4, 0, 1, 2, 3];
```
This somehow doesn't feel right to me. The crux being that there's a `vector of any` on the LHS and so I interpret the RHS as a single element of type `any`, expecting `[4, [0,1,2,3]]`.

Think your proposal in #3191 was more about the RHS treatment to enable more of `add-every-element`. But, what if `+=` on a  `vector of any` is a special case and behaves as `add-single-element`? That would align more with prior Zeek 4.2's `vector of any` behavior and fix #3191. `any` is special `+=` on vectors is special and the above seems easier for me to explain (and happy to pick up the ticket from here).

That said, I believe I'm getting in the way and maybe @rsmmr / @ckreibich / @bbannier may have a look for separate intuition/idea.

bbannier: Since I was pinged, to me this behavior seems to be a natural extension with how vectors over non-`any` types behave, e.g.,

```zeek
local x1 = vector(0, 1, );
local x2 = vector(2, 3, );
x1 += x2;
assert(all_set(x1 == vector(0, 1, 2, 3, )));
```

The added caveat in the snippet you point out is that with this patch a `vector of any` and any other vector (e.g., the `vector of count` here) are implicitly covariant (since `any` and any other type are covariant). We are not doing anything with the `any` values in the `vector` here, so I think this should be a consistent use of fine (_The only operation allowed on a variable of type any is assignment._). IMO the semantics are unfortunately still too subtle.

Without knowing any of the history around this, my gut reaction to the special behavior of `vector`'s `+=` if vector element type and RHS type are covariant (i.e., that one can add a `count` value to a `vector of count` with `+=`) would be that it feels strange and might be unnecessary, e.g., `set` does not have this and can only have a compatible `set` on the RHS of `+=` (there's a dedicated `set` operator `add` to add single elements). If there was a discussion for adding `vector` `+=` today my argument would probably be to instead force the user to potentially create a RHS `vector` and handle that efficiently, e.g.,

```zeek
local xs: vector of count;
# xs += 1; # REJECTED: incompatible LHS and RHS.
xs += vector(1); # Works.
```

awelzel: I'm resolving because after pondering away from the computer I'm back at arguing based on the 4.2 behavior.

awelzel (approved):
Thanks Vern!

PR #3250, opened 2023-08-25
http: Prevent request/response de-synchronization and unbounded state…

### LINKED ISSUE #1705: half duplex http connections can cause unbounded state growth
If zeek sees half duplex http traffic it might result in unbounded state growth.

If zeek is seeing http replies but not requests, everything works OK. zeek will log the replies as they show up to the http log with the request fields unset.

if zeek is setting http requests but not replies, `c$http_state$pending` grows unbounded, and requests are not logged until the connection closes or zeek stops, at which point they all get logged at once.  This can be days later and the pending list could contain tens of thousands of requests.

Attached is http_many_one_connection.pcap, which is just from apachebench running against a server with the keepalive option enabled.

If you run it with this script and different filters you can see the problem:

```
global requests: count=0;
global replies: count=0;

event http_request(c: connection, method: string, original_URI: string,
                   unescaped_URI: string, version: string)
{
    ++requests;
}

event http_reply(c: connection, version: string, code: count, reason: string)
{
    ++replies;
}

hook HTTP::finalize_http(c: connection)
{
    print "Pending", |c$http_state$pending|;
}

event zeek_done()
{
    print "Requests", requests;
    print "Replies", replies;
}

```

```
justin@mbp:/tmp$ zeek -r ~/pcaps/http_many_one_connection.pcap ./h.zeek 
Pending, 0
Requests, 10000
Replies, 10000
justin@mbp:/tmp$ zeek -r ~/pcaps/http_many_one_connection.pcap ./h.zeek -f 'src port 8080'
Pending, 0
Requests, 0
Replies, 10000
justin@mbp:/tmp$ zeek -r ~/pcaps/http_many_one_connection.pcap ./h.zeek -f 'dst port 8080'
Pending, 10000
Requests, 10000
Replies, 0
```

This is somewhat similar to #289 The fix is likely different though.  We either need some sort of time or count (or both) threshold to cause pending messages to be flushed.

[http_many_one_connection.zip](https://github.com/zeek/zeek/files/6901156/http_many_one_connection.zip)

awelzel:
… growth

When http_reply events are received before http_request events, either through faking traffic or possible re-ordering, it is possible to trigger unbounded state growth due to later http_requests never being matched again with responses.

Prevent this by synchronizing request/response counters when late requests come in.

Also forcefully flush pending requests when http_replies are never observed either due to the analyzer having been disabled or because half-duplex traffic.

Fixes #1705


awelzel:
@JustinAzoff  - this isn't overly clever as you started in #1705 , maybe want to take a peek?

0xxon (approved):
Looks good to me

PR #3257, opened 2023-08-29
pre-commit: autoupdate

awelzel:
...except for clang-format, because versions after v13.0.0 have borked the Whitesmith formatting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,13 @@ repos:
   - id: clang-format
 
 - repo: https://github.com/maxwinterstein/shfmt-py
-  # Latest master revision as of 2023-08-17 allows to run with modern Python
-  # versions. Switch this back to a tag once there is one :-)
-  rev: d4491d8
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -8,15 +8,13 @@ repos:
...
-  # versions. Switch this back to a tag once there is one :-)
-  rev: d4491d8
+  rev: v3.7.0.1
   hooks:
     - id: shfmt
       args: ["-w", "-i", "4", "-ci"]
 
 - repo: https://github.com/pre-commit/mirrors-yapf
-  rev: v0.31.0
+  rev: v0.32.0
```
bbannier, line 17: https://github.com/pre-commit/mirrors-yapf is archived now and documents to use upstream instead. We should probably use

```yaml
- repo: https://github.com/google/yapf
  rev: v0.40.0
```

This brings in no additional formatting changes.

bbannier (approved):

PR #3259, opened 2023-08-30
testing/external: Exclude telemetry.log from baselines

awelzel:
The diffs produced by telemetry.log when introducing a weird or removing/adding protocol specific logs is overwhelming and distracting without providing value. Exclude telemetry.log similar to how we already exclude stats.log.

Some more targeted telemetry.log tests exists in the normal testing/btest suite and that appears more sensible.

Baseline updates:

https://github.com/zeek/zeek-testing-private/pull/17
https://github.com/zeek/zeek-testing/pull/16


bbannier (approved):
LGTM assuming CI passes and the external testing hash updates are reasonable.

PR #3261, opened 2023-08-30
RecordType: Fix deferred initialization of inner records with non-const default fields

### LINKED ISSUE #3260: Behavior changes due to deferring record field initialization
### Discussed in https://github.com/zeek/zeek/discussions/3212

<sup>Originally posted by **awelzel** August  2, 2023</sup>
Creating as a discussion as maybe this will never come up in practice, but I'm suspecting this is bug/regression material or minimally new subtle gotchas one needs to be aware of.

With [deferred initialization of aggregate record fields](https://github.com/zeek/zeek/pull/2938), any nested record fields using non-const default values (network_time() / current_time(), ...) are now evaluated at access time only.

awelzel:
RecordType: Fix deferred initialization of inner records with non-const default fields

With Zeek 6.0, nested records are lazily initialized. This caused subtle
behavior changes for non-const &default expressions of inner records.

Expression reading global state like network_time() or get_net_stats()
are evaluated at access time rather than construction time. Attempt to
fix by disallowing deferred initialization of any records that
have creation_inits that cannot be deferred. Complications revolve
around adding fields via redef and propagating the non-deferrability

Closes #3260


testing/btest/Baseline/core.record-field-defer-table-2/output
```diff
@@ -0,0 +1,19 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+0, my_seq()
+1, my_seq()
+2, my_seq()
+3, my_seq()
+4, printing
+4, s1, [seq=1, inner=[seq=3]]
+4, s2, [seq=2, inner=[seq=4]]
+4, tbl, {
+4, my_seq()
```
awelzel, line 10: I haven't looked what's happening here, but printing an element invokes the `&default` expression 3 times :man_shrugging: 

This happens with vanilla 6.0 and 5.2.

awelzel: Explanation/fix in #3270.

rsmmr (commented):
Have to say I'm not a fan of all this complexity. Hard to track what's going on, and that for something that probably won't matter often in practice. That said, I do agree that we should fix the unintended change in behavior. I'm wondering if there's something easier we can do. One idea: only defer initialization for records where all defaults are constants; and just fully initialize other types right away as we used to do. Could then just keep a flag per record if it's' deferrable.

awelzel:
> Have to say I'm not a fan of all this complexity. Hard to track what's going on, and that for something that probably won't matter often in practice.

I'm not a fan either, but neither do I like the gotcha (and there is the whole rebuild-parse-time-table-record-entries that's somewhere in the same corner).

> One idea: only defer initialization for records where all defaults are constants; 

...or there are no &defaults and no record fields that are themselves not-deferrable.

I think otherwise @vpax optimizations aren't effective anymore and this might all become mood.

> and just fully initialize other types right away as we used to do. Could then just keep a flag per record if it's' deferrable.

That flag is `creation_inits.empty()`, I believe, as all non-const &default's will end-up in there. The one tricky part is that this can change during parsing through non-const &default field addition via redef and the main complexity is fixing things up after the fact.

I'll try from a different angle - if we determine deferrability only once parsing has completed and consider it a separate optimization pass, this might become less complex.

awelzel:
Replaced by https://github.com/zeek/zeek/pull/3288

PR #3270, opened 2023-09-01
Avoid unnecessary &default evaluations of records

awelzel:
Using &default functions having print statements showed surprising extra and unnecessary invocations when

1) Iterating over tables/sets where the keys have non-const &default values. The &default expressions are evaluated for each val recovery. Fix by using lower-level AppendField() helper within CompositeHash.

2) The loop indices and vars of for-loops are initialized upon entering a function due to being added to the scope's inits, even if during for-loop execution these are then initialized. Fix by adding a new INIT_SKIP InitClass and using that for for-loops.


awelzel:
@vpax - wonder if you have opinions or initial reaction to those changes? As in the description, not sure if a InitClass INIT_SKIP is all that reasonable. Motivation is "locals" declared in for-loop heads only, we should not need to initialize them upon entering a event/function.

 I suspect CompositeHash might benefit from adapting ZVal's throughout, but this isn't attempting this and instead uses the existing `DoClone()` approach of `AppendField()`.

src/CompHash.cc
```diff
@@ -359,10 +359,11 @@ bool CompositeHash::RecoverOneVal(const HashKey& hk, Type* t, ValPtr* pval, bool
 
 					ASSERT(int(values.size()) == num_fields);
 
-					auto rv = make_intrusive<RecordVal>(IntrusivePtr{NewRef{}, rt});
+					auto rv = make_intrusive<RecordVal>(IntrusivePtr{NewRef{}, rt},
+					                                    false /* init_fields */);
 
 					for ( int i = 0; i < num_fields; ++i )
-						rv->Assign(i, std::move(values[i]));
+						rv->AppendField(std::move(values[i]), rt->GetFieldType(i));
```
vpax, line 366: If the field is optional, does this work okay because `values[i]` will be nil?

src/Var.cc
```diff
@@ -420,7 +420,9 @@ StmtPtr add_local(IDPtr id, TypePtr t, InitClass c, ExprPtr init,
 
 	else
 		{
-		current_scope()->AddInit(std::move(id));
+		if ( c != INIT_SKIP )
```
vpax, line 423: could this be done just by checking for `INIT_NONE` and not bothering to create a new initialization class? I forget whether `INIT_NONE` comes with some sort of activity that would make it inappropriate here.

vpax:
@awelzel The code changes look okay, though I had a couple of questions that I flagged.  I didn't look over the BTests other than superficially, let me know if you'd like a second pair of eyes on them or if they're sufficiently straightforward to skip in that regard.

vpax:
(oh, and: glad you spotted these inefficiencies!)

src/Var.cc
awelzel: For records, containers and vectors, INIT_NONE does initialize local variables
https://github.com/zeek/zeek/blob/6e6a2bee8af39fb03fd9807a30b6ac3500b1464e/src/Stmt.cc#L1782-L1795

Flipping this to INIT_NONE breaks the following.
```
event zeek_init()
   {
   local v: vector of count;
   v += 1;
   print v;
   }
```

Vectors, records, tables may well be used as index/yield types.

(For records it's a bit curious though, because it allows to construct an "empty" record with no fields set).

I'm resolving this one.

src/CompHash.cc
awelzel: Yes, values[i] will be nil for unset values. This is modeled after `RecordVal::DoClone()`.

awelzel:
> (oh, and: glad you spotted these inefficiencies!)

I'm not sure they would actually show up significantly: For example, I'm sure re-working CompHash to use ZVals would have an acutal impact. I was mainly irritated by surprising calls to the `&default=` functions :-)

@rsmmr - would you mind taking a peek?

rsmmr (approved):
Looks good. I would just suggest to add some comments to the tests on what's expected in terms of number of executions, as that's hard to infer from code/baselines.

PR #3271, opened 2023-09-01
Script Optimization maintenance for September 2023

vpax:
These changes are mainly for compatibility with GH-3249's tweaks to vector `+=` operations. In addition, GH-3199 wound up breaking low-level ZAM profiling, so this restores that.


src/script_opt/ScriptOpt.cc
```diff
@@ -554,6 +554,9 @@ void clear_script_analysis()
 	{
 	IDOptInfo::ClearGlobalInitExprs();
 
+	// Keep the functions around if we're debugging, so we can
+	// generate profiles.
+#ifndef DEBUG
```
awelzel, line 559: Seems strange to make this conditional on `DEBUG`, shouldn't profiling work in release/default builds, too?

How about:
```
if ( ! analysis_options.profile_ZAM )
    ...    
```

---

Eh, okay. After trying and looking a bit, the ZAM profiling functionality is actually gated on `DEBUG` throughout, so I understand now where this is coming from.

Yet, leaving a note: A new option/define named `ZEEK_ZAM_PROFILING` might be worth considering. There may be users wanting to run Release builds and accepting a minor overhead (it's just a few more bool checks when available but not enabled it seems), but still want to have the option to use `-O profile-ZAM` without re-compiling.

`DEBUG` may also add extra overhead and skew results around a bit when profiling what should otherwise be an optimized build?

I was scratching my head a bit using `-O ZAM -O profile-ZAM` and it only showed "did not execute" when trying with a release build  - might be worth saying "ZAM profiling unavailable" so.

Anyway, merging :+1:

awelzel (approved):

vpax:
Thanks for the reflections (and merging). Regarding introducing a `ZEEK_ZAM_PROFILING` option, I'm all for it ... I just don't know how to do it! (And for my early usage of the option, didn't need it - but I agree it's significantly better than the current approach.) In particular, it seems it needs (1) a `configure` option, (2) a default to _on_ when building `--enable-debug`, and otherwise a default to _off_. If you want to sketch the correct way to do that, I'll give it a go. I'm of course also fine if you just add it yourself at some point.

PR #3274, opened 2023-09-05
Add OpenSUSE Leap 15.5 to ci

0xxon:
This also adds 15.4 to the list of distributions that are skipped by
default - let me know if this is not a good idea. I assume that by now
most people will be running 15.5 (15.4 will be EOL in ~3 months).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -250,6 +250,14 @@ opensuse_leap_15_4_task:
     dockerfile: ci/opensuse-leap-15.4/Dockerfile
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
+  << : *SKIP_TASK_ON_PR
+
+opensuse_leap_15_5_task:
+  container:
+    # Opensuse Leap 15.5 EOL: ~Dec 2024
+    dockerfile: ci/opensuse-leap-15.5/Dockerfile
+    << : *RESOURCES_TEMPLATE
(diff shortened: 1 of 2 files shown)


0xxon:
Build fails due to https://github.com/zeek/zeek/issues/3275

awelzel:
Maybe `--with-python` is non-functional since switching away from PythonInter /PythonDev?

---

It's so weird... On Leap 15.5, the file `/usr/share/cmake/Modules/FindPython/Support.cmake` somewhere at the top has the following:
```
if (_${_PYTHON_PREFIX}_REQUIRED_VERSION_MAJOR EQUAL "3")
   set(_${_PYTHON_PREFIX}_VERSIONS 3.10 3.9 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0)
```

And if one manually patches the above to include 3.11, it'll find and use the 3.11 interpreter.

Alternatively, the following also does the trick because it'll check the `python3` executable first and be satisfied IIUC.
```
 set(ZEEK_PYTHON_MIN 3.5.0)
+set(Python_FIND_UNVERSIONED_NAMES FIRST)
 find_package(Python ${ZEEK_PYTHON_MIN} REQUIRED COMPONENTS Interpreter)
```

Neither is great I think.

awelzel:
Done in #3281 - closing.

PR #3277, opened 2023-09-06
Propagate failure reason in `spicy::decline_input`/`zeek::reject_protocol`

### LINKED ISSUE #3276: spicy::decline_input(reason: string) - reason not propagated
Using `spicy::decline_input()` triggers an analyzer_violation_info event. However, the `reason` field in the violation info record is always populated with "protocol rejected", not with the `reason` given. Also the `data` field containing the failing data isn't included.

The commented out `throw` works as expected and the analyzer violation shows "G aborts parsing" and includes the failing bytes in `data`.

```
module ParseBytes;

import spicy;

public type OneByte = unit {
    b: bytes &size=1 {
      if ( self.b != b"T" ) {
        spicy::decline_input("%s is not the right byte" % self.b);
        # throw "%s aborts parsing" % self.b;
      } else {
        spicy::accept_input();
      }
    }
};

public type ManyBytes = unit {
    : OneByte()[];
};


$  zeek -C -r request-line.pcap  ./parse-bytes.hlto ./parse-bytes.zeek misc/dump-events DumpEvents::include=/analyzer_/
1580991211.780489 analyzer_violation_info
                  [0] atype: AllAnalyzers::Tag = Analyzer::ANALYZER_MYPARSEBYTES
                  [1] info: AnalyzerViolationInfo = [reason=protocol rejected, c=[id=[orig_h=127.0.0.1, orig_p=59619/tcp, resp_h=127.0.0.1, resp_p=12345/tcp], ...], f=<uninitialized>, aid=7, data=<uninitialized>]

CUFEV13w31LnrYuTk8, T, G
```

bbannier:
Closes #3276.


src/spicy/manager.cc
```diff
@@ -586,8 +586,8 @@ static void hook_decline_input(const std::string& reason) {
 
     if ( auto x = cookie->protocol ) {
         auto tag = spicy_mgr->tagForProtocolAnalyzer(x->analyzer->GetAnalyzerTag());
-        SPICY_DEBUG(hilti::rt::fmt("rejecting protocol %s", tag.AsString()));
-        return x->analyzer->AnalyzerViolation("protocol rejected", nullptr, 0, tag);
+        SPICY_DEBUG(hilti::rt::fmt("rejecting protocol %s: %s", tag.AsString(), reason));
+        return x->analyzer->AnalyzerViolation(reason.c_str(), nullptr, 0, tag);
```
awelzel, line 590: (nothing to address - food for thought)

There's no additional context here to populate `data`?

Using `throw "error"` causes `data` in `analyzer.log` to be populated, so that has the required context. Is there a rule/guideline when/where to call `decline_input()` compared to `throw "error"` given (on first sight) the latter appears more useful?

And maybe (not for this PR) `decline_input()` could have a second parameter allowing to pass in the bytes that somehow failed/violated the protocol if they can't be inferred.

awelzel (approved):

src/spicy/manager.cc
bbannier: Interesting question.

The way things are currently modeled in the Spicy integration, Spicy-side runtime exceptions are often pretty directly tied to the input triggering it, <https://github.com/zeek/zeek/blob/7e11501d3cdee54482fcce9c9a640cabfe13d2c2/src/spicy/protocol-analyzer.cc#L73-L78> This probably makes sense for constructs like `&requires` which often directly relate to the input (but could `&parse-from` or similar). The `throw` you mentioned is also treated the same way even though it might be even less related. Ultimately we cannot tell whether there was relation.

IMO `spicy::decline_input` OTOH is a more high-level way to explicitly interact with the Zeek analyzer state machine and rejecting input is not something exceptional. One way it would be invoked would be that at some point the grammar writter checks invariants not directly encoded in the grammar and decides to opt out of further analysis. This argument however clashes with how we currently recommend in the package template to implement `%error` and call `spicy::decline_input` from it (or `zeek::reject_protocol` currently), <https://github.com/zeek/package-template/blob/312b2dea5c9de648861e445017c63414b43aa487/features/spicy-protocol-analyzer/analyzer/zeek_@ANALYZER_LOWER@.spicy@ALT-two-units@#L22-L24>. Since exceptions are cannot be caught in Spicy this would lead to two log entries in `analyzer.log`, one for the runtime exception which just bubbled up, and another one from the explicit `spicy::decline_input`. My gut feeling is that we do not want to recommend using `spicy::decline_input` in this way.

In this thinking the guidance would be to use `throw` similar to how you would use `&requires` for checking local invariants; `spicy::decline_input` would be for more higher-level validation.

awelzel: > Since exceptions cannot be caught in Spicy this would lead to two log entries in analyzer.log, one for the runtime exception which just bubbled up, and another one from the explicit spicy::decline_input

Yep, this actually happens and looks a bit confusing, too. So went without the `spicy::decline_input()` in `on %error` and only used explicit `throw`.

The other part around decline_input() is this from the doc string:

> Note that this does not automatically abort processing. If that’s desired, you need to trigger a parse error as well, e.g., by throwing an error, throw “<error message>”.

So when you check a high-level invariant, call `decline_input()`, it continues to parse afterwards. And a `throw` is required to make it stop (or some other `&eod` approach to eat the rest of the data?).

Maybe worth a ticket?

bbannier: > Maybe worth a ticket?

Yes please.

rsmmr (approved):
Changes looks good but agree that there's more here to think through.

PR #3282, opened 2023-09-07
Extend the timeouts on some of the supervisor btests

timwoj:
This fixes testing failures on Cirrus, where for some reason the timeouts are being hit and the tests are failing because zeek is being killed early. This shouldn't affect local test runs.


timwoj:
Unfortunately, this doesn't look like the right fix for this. Closing this PR to investigate further.

PR #3284, opened 2023-09-08
Move allocation of analyzer_hash to file_mgr

timwoj:
This avoids recreating this CompositeHash object every time a new AnalyzerSet is created. Since the object the hash is based on doesn't change, there's no reason to recreate it for each new file. For reference, the 2009-M57-day11-18.trace file creates 4738 new analyzer sets.

### DIFF
--- a/src/file_analysis/AnalyzerSet.cc
+++ b/src/file_analysis/AnalyzerSet.cc
@@ -22,10 +22,6 @@ static void analyzer_del_func(void* v)
 
 AnalyzerSet::AnalyzerSet(File* arg_file) : file(arg_file)
 	{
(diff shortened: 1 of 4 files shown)


ckreibich (commented):
Cool. Should we initialize the new `analyzer_hash` member in the constructor?

timwoj:
> Cool. Should we initialize the new `analyzer_hash` member in the constructor?

It could be set to nullptr in the header, but it has to wait to be initialized until after script parsing is done because we don't know the record type yet during the constructor.

ckreibich (approved):
Perfect, I did mean `nullptr` initialization. Thanks!

PR #3285, opened 2023-09-10
RecordType: Fix deferred initialization of inner records with non-const default fields - v2

awelzel:
Move back to a single field_inits vector for record types instead of the
creation_inits / deferred_inits split. Put deferral information into the
FieldInit class itself. All FieldInit classes are deferrable except for
ExprFieldInit and RecordFieldInit.The IsDeferrable() result for RecordFieldInit
is cached on the RecordType once parsing has completed.


awelzel:
Replaced by https://github.com/zeek/zeek/pull/3288

PR #3288, opened 2023-09-11
Fix deferred record initialization - v4

awelzel:
This keeps the current structure of `creation_inits` and `deferred_inits`, All record field inits are
treated as `creation_inits` during parsing and "promoted" to `deferred_inits` once parsing has completed
and that's applicable.

Prior attempt consolidating `creation/deferred_inits` and a `IsDeferrable()` helper on `FieldInit` indicated small but measurable extra overhead.


awelzel:
@rsmmr - could you see if this appears more reasonable? Not having the `creation|deferred_inits` split showed extra overhead in micro-benchmarks (and it didn't look all that better). 

This still looks a bit complex, but I feel a bit better just doing it as an optimization, rather than at `redef` time.

But, I also have second thoughts about the practical relevance of the examples in #3212. Maybe a more pragmatic approach could be to start warning on non-const &default usage in nested records and promote such constructs to errors down the road: Then users are forced to `&optional` or `&default` these fields. That would avoid the gotcha or non-trivial documentation, too.

Happy to chat directly about ways forward here.

src/Type.cc
```diff
@@ -1205,7 +1245,12 @@ void RecordType::AddField(unsigned int field, const TypeDecl* td)
 		TypeTag tag = type->Tag();
 
 		if ( tag == TYPE_RECORD )
-			init = std::make_unique<detail::RecordFieldInit>(cast_intrusive<RecordType>(type));
+			{
+			// Initially, put record fields into creation_inits. Once parsing has
+			// completed, they may move into deferred_inits. See DoneParsing().
+			auto rfi = std::make_unique<detail::RecordFieldInit>(cast_intrusive<RecordType>(type));
+			creation_inits.emplace_back(field, std::move(rfi));
+			}
```
rsmmr, line 1253: I believe it's ok/expected to leave `init` unset here, correct? Want to add comment saying so?

src/Type.h
```diff
@@ -734,9 +737,20 @@ class RecordType final : public Type
...
+	// contain any nested records that themselves are not deferrable,
+	// initialization can not be deferred, otherwise possible.
+	bool IsDeferrable() const;
+
+	static void DoneParsing();
+
 private:
 	RecordType() { types = nullptr; }
 
+	static std::vector<zeek::RecordTypePtr> record_types;
```
rsmmr, line 752: Instead of adding more hidden global state, suggest to use AST traversal instead, that's what's it's meant for: examining language elements after parsing.

rsmmr (commented):
Yeah, I like this much better, this is good, just one more suggestion.

src/Type.h
awelzel: Yeah, this is nicer. As in DM, requires a `const_cast`.

I made it a private class on RecordType so it can work directly with `creation_inits` of individual instances.

src/Type.cc
awelzel: Yep, it's the same pattern as for ExpressionFieldInit `efi` above. I've added a comment.

rsmmr (approved):

src/Type.cc
```diff
@@ -1116,6 +1126,59 @@ class VectorFieldInit final : public FieldInit
...
+class RecordType::CreationInitsOptimizer : public detail::TraversalCallback
+	{
+public:
+	detail::TraversalCode PreID(const detail::ID* id) override
+		{
+		if ( const auto& t = id->GetType() )
+			HANDLE_TC_TYPE_POST(t->Traverse(this));
+
+		return detail::TC_CONTINUE;
+		}
```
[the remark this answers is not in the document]
awelzel: `PreType()` is somewhat recent. `traverse_all()` will not by default invoke `PreType()`, so that `PreID()` trampoline is needed somewhere. It took quite a while to figure out and am now unsure if it's not bordering unexpected behavior.

In any case, I've switched to `PreTypedef()` after looking again. That's invoked for all IDs that are types, ends-up shorter and coincidentally/accidentally also allows to get rid of the `const_cast`.

PR #3292, opened 2023-09-13
Fix handling of module scope when checking exported Spicy types for collisions

### LINKED ISSUE #3279: Spicy enum export discrepancy: ZEEK_SPICY_MODULE_PATH vs loading hlto files
Compiling the following analyzer for testing https://github.com/mmguero-dev/icsnpp-genisys:
```
$ mkdir build && cd build
$ cmake ../
$ make
$ ls -lha genisys.hlto 
-rwxr-xr-x 1 awelzel awelzel 9.7M Sep  7 08:53 genisys.hlto
```

The analyzer.spicy file contains two public enums. `HeaderCode` and `Direction`, no explicit export.

Running with `ZEEK_SPICY_MODULE_PATH both are exported:
```
$ ZEEK_SPICY_MODULE_PATH=$(pwd) zeek -NN Zeek::Spicy
Zeek::Spicy - Support for Spicy parsers (*.hlto) (built-in)
    [Analyzer] Finger (ANALYZER_FINGER, enabled)
    [Analyzer] Syslog (ANALYZER_SYSLOG, enabled)
    [Analyzer] spicy_genisys_TCP (ANALYZER_SPICY_GENISYS_TCP, enabled)
    [Type] Genisys::HeaderCode
    [Type] Genisys::Direction
    Implements LoadFile (priority 0)
```

Pointing Zeek just at the genisys.hlto file without setting ZEEK_SPICY_MODULE_PATH exports just the `HeaderCode`.
```
$ zeek -NN Zeek::Spicy ./genisys.hlto 
Zeek::Spicy - Support for Spicy parsers (*.hlto) (built-in)
    [Analyzer] Finger (ANALYZER_FINGER, enabled)
    [Analyzer] Syslog (ANALYZER_SYSLOG, enabled)
    [Analyzer] spicy_genisys_TCP (ANALYZER_SPICY_GENISYS_TCP, enabled)
    [Type] Genisys::HeaderCode
    Implements LoadFile (priority 0)
```

Scripts referencing the enum fail due to not finding the identifiers:
```
$ zeek ./genisys.hlto ../analyzer/
error in ./../analyzer/./main.zeek, line 58: unknown identifier Genisys::Direction_CLIENT_TO_SERVER, at or near "Genisys::Direction_CLIENT_TO_SERVER"
```

bbannier:
When checking exported Spicy types for collisions with existing Zeek types we previously would also check whether they collide with names in global scope, i.e., we didn't provide a `no_global` arg to `detail::lookup_ID` which defaulted to false (since we also provided a module name I'd argue that the behavior of that function is confusing and probably error-prone -- like seen here).

This meant that e.g., a Spicy enum `foo::Direction` (automatically in implicit Spicy module scope) would be detected to collide with the existing Zeek `Direction` enum.

With this patch we use the `lookup_ID` API correctly and do not check against potential collisions with globals anymore since it is not needed.

Closes #3279.

### DIFF
--- a/src/spicy/manager.cc
+++ b/src/spicy/manager.cc
@@ -215,7 +215,7 @@ void Manager::registerPacketAnalyzer(const std::string& name, const std::string&
 void Manager::registerType(const std::string& id, const TypePtr& type) {
     auto [ns, local] = parseID(id);
 
-    if ( const auto& old = detail::lookup_ID(local.c_str(), ns.c_str()) ) {
+    if ( const auto& old = detail::lookup_ID(local.c_str(), ns.c_str(), true) ) {
         // This is most likely to trigger for IDs that other Spicy modules
         // register. If we two Spicy modules need the same type, that's ok as
         // long as they match.
--- a/testing/btest/spicy/export-type-collision-with-global.spicy
+++ b/testing/btest/spicy/export-type-collision-with-global.spicy
@@ -1,0 +1,20 @@
+# @TEST-REQUIRES: have-spicy
+#
+# @TEST-EXEC: spicyz -d -o foo.hlto foo.spicy foo.evt
+# @TEST-EXEC: zeek -NN Zeek::Spicy foo.hlto >output1 2>&1
+# @TEST-EXEC: ZEEK_SPICY_MODULE_PATH=$PWD zeek -B all -NN Zeek::Spicy >output2 2>&1
+# @TEST-EXEC: diff output1 output2 1>&2
+#
+# @TEST-DOC: This validates that an exported Spicy name with local part identical with a Zeek name in global scope does not clash. Regression test for #3279.
+
+# @TEST-START-FILE foo.spicy
+module foo;
+
+# `foo::Direction` has the same local part as Zeek's `::Direction`.
+public type Direction = enum { a };
+public type Other = enum { a };
+# @TEST-END-FILE
+
+# @TEST-START-FILE foo.evt
+import foo;
+# @TEST-END-FILE


src/spicy/manager.cc
```diff
@@ -215,7 +215,7 @@ void Manager::registerPacketAnalyzer(const std::string& name, const std::string&
 void Manager::registerType(const std::string& id, const TypePtr& type) {
     auto [ns, local] = parseID(id);
 
-    if ( const auto& old = detail::lookup_ID(local.c_str(), ns.c_str()) ) {
+    if ( const auto& old = detail::lookup_ID(local.c_str(), ns.c_str(), true) ) {
```
awelzel, line 218: From the description:

> i.e., we didn't provide a no_global arg to detail::lookip_ID which defaulted to true

... defaulted to false? And so global lookup was true? because no_global=false is actually true?

> I'd argue that the behavior of that function is confusing and probably error-prone

TBH, last time I looked at the bool flags of this API I couldn't make much sense of the motivation. I think passing `same_module_only=true` in addition to your current fix might make the explanation easier even if it's redundant with `no_global=true`.

---
This should fix the problem, so I'm approving. If you add another `, true`, that's fine, too :-)

awelzel (approved):

src/spicy/manager.cc
bbannier: Thanks for the careful reading, I'll fix the commit message to not add more confusion. I'll leave the args as is since I am fine with the defaults.

PR #3301, opened 2023-09-15
CMakeLists: Set Python3_EXECUTABLE hint for spicy

awelzel:
Sibling for https://github.com/zeek/spicy/pull/1528

@bbannier - do you think we'd actually need this? Given you backported the above to 1.8, not sure this is strictly required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -924,6 +924,12 @@ if (NOT DISABLE_SPICY)
         # disable it.
         set(HILTI_DEV_PRECOMPILE_HEADERS OFF)
 
+        # Remove in v6.2. Older versions of Spicy use find_package(Python3),
+        # accommodate by setting the Python3_EXECUTABLE hint.
+        if (Python_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


bbannier:
> do you think we'd actually need this? Given you backported the above to 1.8, not sure this is strictly required.

Like I wrote over in https://github.com/zeek/spicy/pull/1528#issuecomment-1720851508, no, this isn't strictly necessary. It would however fix this issue for Zeek today instead of whenever we put out the next spicy-1.8.x release for zeek-6.0.x, or spicy-1.9.0 for zeek-6.1.

CMakeLists.txt
```diff
@@ -924,6 +924,12 @@ if (NOT DISABLE_SPICY)
         # disable it.
         set(HILTI_DEV_PRECOMPILE_HEADERS OFF)
 
+        # Remove in v6.2: Older versions of Spicy use find_package(Python3).
+        #                 Accommodate by setting the Python3_EXECUTABLE hint.
```
bbannier, line 928: Fancy indention 👯‍♂️. If we want it this way, make it flush right, else just default to flush left like elsewhere.

bbannier (approved):

PR #3303, opened 2023-09-15
Correctly compute name for spicyz export with debug log.

bbannier:
In 36a6770e9891114b4f42dac3728040c67cc9411e we changed the way Spicy fields exported as
record fields are represented. This broke the `zeek` debug log which hardcoded a different representation.

This patch brings the generation of the debug log in line with the actual AST structure. We also enable debug logging in one test to validate that it works, at least for the case we hit here.

### DIFF
--- a/src/spicy/spicyz/glue-compiler.cc
+++ b/src/spicy/spicyz/glue-compiler.cc
@@ -1318,14 +1318,7 @@ struct VisitorZeekType : hilti::visitor::PreOrder<hilti::Result<hilti::Expressio
                 SPICY_DEBUG(hilti::util::fmt("Creating Zeek record type %s::%s with fields:", ns, local));
 
                 for ( const auto& f : fields )
-                    SPICY_DEBUG(hilti::util::fmt("  %s", f.as<hilti::expression::Ctor>()
-                                                             .ctor()
-                                                             .as<hilti::ctor::Tuple>()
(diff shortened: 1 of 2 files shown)


src/spicy/spicyz/glue-compiler.cc
```diff
@@ -1318,7 +1318,9 @@ struct VisitorZeekType : hilti::visitor::PreOrder<hilti::Result<hilti::Expressio
                 SPICY_DEBUG(hilti::util::fmt("Creating Zeek record type %s::%s with fields:", ns, local));
 
                 for ( const auto& f : fields )
-                    SPICY_DEBUG(hilti::util::fmt("  %s", f.as<hilti::expression::Ctor>()
+                    SPICY_DEBUG(hilti::util::fmt("  %s", f.as<hilti::expression::UnresolvedOperator>()
+                                                             .operands()[1]
+                                                             .as<hilti::expression::Ctor>()
```
rsmmr, line 1323: I'm wondering if we should just print the original expression `f` here, instead of trying to decompose it, which seems brittle in case of further changes in the future.

rsmmr: Yeah, not super pretty, but seems fine to me.

rsmmr (approved):

PR #3305, opened 2023-09-18
Move zeek/spicy-ldap into Zeek

### LINKED ISSUE #3234: Move zeek/spicy-ldap into Zeek
We should move https://github.com/zeek/spicy-ldap into Zeek proper.

bbannier:
Closes #3234.


bbannier:
@rsmmr The issue still missing here is that the LDAP parser leaves its traces in some files outside of its own baselines, e.g., event declarations. This was not an issue when migrating syslog or finger since they had a 100% compatible legacy implementation which was available if Spicy analyzers were not enabled. We do not have that for LDAP, so one solution would be to more aggressively sprinkle `@TEST-REQUIRES: have-spicy` among btests (which would effectively reduce test coverage for devs building with `--disable-spicy`).

rsmmr (approved):
Integration looks good to me. (Haven't looked at the analyzer itself.)

bbannier:
> Integration looks good to me. (Haven't looked at the analyzer itself.)

This PR only moves the existing analyzer into Zeek. I have heard that e.g., groups in Corelight (e.g., @J-Gras, @Mohan-Dhawan) ran into issues with spicy-ldap since the analyzer does not fully implement LDAP and as not defensive enough about cases it does not support. This could e.g., trigger nonsensical parsing results or worse pathological performance behavior.

Do you have a good suggestion on how to integrate such an analyzer? This PR enables it by default which likely would result in valuable feedback from users so we can improve the parser; an alternative approach would be to make the parser opt-in for now.

rsmmr:
> > Integration looks good to me. (Haven't looked at the analyzer itself.)
> 
> This PR only moves the existing analyzer into Zeek. I have heard that e.g., groups in Corelight (e.g., @J-Gras, @Mohan-Dhawan) ran into issues with spicy-ldap since the analyzer does not fully implement LDAP and is not defensive enough about cases it does not support. This could e.g., trigger nonsensical parsing results or worse pathological performance behavior.
> 
> Do you have a good suggestion on how to integrate such an analyzer? This PR enables it by default which likely would result in valuable feedback from users so we can improve the parser; an alternative approach would be to make the parser opt-in for now.

I think the best way to go would be running this for a while on real world traffic and looking through logs to see how noisy it is. If it's just an occasional "nonsensical result", that would be fine I'd say; but if it's frequent, then we need to either fix the parser (best) or make it opt-in.

Also, if it remains default-on, we should make sure there's a straight-forward way to turn it off if it causes trouble for someone.

bbannier:
> Also, if it remains default-on, we should make sure there's a straight-forward way to turn it off if it causes trouble for someone.

I have added an option to enable the plugin (defaulted to `T`). This should simplify turning the analyzers added by spicy-ldap to be turned off/on selectively.

src/analyzer/protocol/ldap/ldap.evt
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+
+protocol analyzer LDAP_TCP over TCP:
+    parse with LDAP::Messages,
+    ports { 389/tcp, 3268/tcp};
```
awelzel, line 5: I haven't looked much around, but this one came to mind (and slightly related to the enable option).

I had removed `ports` usage in the QUIC `.evt` mainly due to re-compiling to add or remove a port seemed a weird workflow when it can be done in script land just as well. Wonder if you have thoughts on generally promoting the previous pattern rather than `ports` in .evt files. Is there an API to prevent one of the port in an .evt file to be used?
 
Compare with:
https://github.com/zeek/zeek/blob/0b940d8b3ce8ebb5539b99b8eea460fd6eac3cdc/scripts/base/protocols/http/main.zeek#L150-L161

scripts/base/protocols/ldap/main.zeek
```diff
@@ -260,8 +263,13 @@ redef record connection += {
 
 #############################################################################
 event zeek_init() &priority=5 {
-  Log::create_stream(LDAP::LDAP_LOG, [$columns=Message, $ev=log_ldap, $path="ldap", $policy=log_policy]);
-  Log::create_stream(LDAP::LDAP_SEARCH_LOG, [$columns=Search, $ev=log_ldap_search, $path="ldap_search", $policy=log_policy_search]);
+  if ( LDAP::enable ) {
```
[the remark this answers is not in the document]
bbannier: My initial intention was more of a feature flag to opt out of a developing feature and less a convenience mechanism. I do agree that I made this look like other toggles (which might be more of a convenience).

> 

If there's an issue just with one of the transport protocols, the user need to fiddle by hand.

This is all in anticipation of the used parser potentially performing poorly. All analyzers added here share the same parser so I'd expect them to have similar parsing issues. I think a global feature flag would make sense here.

> Analyzers can be re-enabled at runtime, but this won't create the log streams if LDAP::enabled=F was set with the current approach ...
> ...
> Users (and we) should be able to do `redef Analyzer::disabled_a...`

I'd argue that for a feature flag we wouldn't need the ability to toggle it at runtime.

An additional complexity of Spicy analyzers in Zeek today is that `Analyzer::disable_analyzers` is not sufficient to disable them. This disables them on the script side of things, but still e.g., feeds analyzers with input. This is due to how Spicy analyzers are discovered in HLTO files and loaded via a [separate registration mechanism](https://github.com/zeek/zeek/blob/4af7d7027feb6af88b463443482bf2fdf18ac488/src/spicy/manager.cc#L437-L460) (this is also what e.g., [`Spicy::disable_protocol_analyzer` and similar functions](https://github.com/zeek/zeek/blob/4af7d7027feb6af88b463443482bf2fdf18ac488/src/spicy/manager.cc#L437-L460) interact with).

What we will ultimately need to fully support `Analyzer::disabled_analyzers` for Spicy-managed analyzers would be the functionality in `spicy::Manager` which can inspect that value at runtime and react on it. I didn't really plan to address that with this patch, but I can look into it.

src/analyzer/protocol/ldap/ldap.evt
bbannier: Doesn't `likely_server_ports` require reliable DPD signatures? With `ports` in the EVT file the analyzer would automatically be registered for a port, regardless of whether any of its signatures match (or if has any signatures at all).

scripts/base/protocols/ldap/main.zeek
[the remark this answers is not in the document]
awelzel: > My initial intention was more a feature flag to opt out of a developing feature and less a convenience mechanism. I do agree that I made this look like other toggles (which might be more of a convenience).

Yep, got it. I still think the following is all that should be needed to disable the new LDAP analyzers. We can document this as receipt in NEWS in case there are issues.
```
redef Analyzer::disabled_analyzers += {
  Analyzer::ANALYZER_LDAP_TCP,
  Analyzer::ANALYZER_LDAP_UDP,
```

The other side might be an expectation of `::enable` flags for other modules/analyzers.


> An additional complexity of Spicy analyzers in Zeek today is...
> What we will ultimately need to fully support ...

I actually thought that is functional and skimming the code, `toggleProtocolAnalyzer()` fowards to `analyzer_mgr->Enable/Disable()` which is all that `Analyzer::disable_analyzer()` or the `disabled_analyzers` mechanism should be doing (modulo the replaces cases). Have you tried? -I have not ;-)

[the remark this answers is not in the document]
bbannier: I tried and you are right, this works. Either I had a completely wrong model, or something changed here. I am removing the flag and will leave it up to users to just disable these analyzers.

src/analyzer/protocol/ldap/ldap.evt
bbannier: Yupp, that's the API the Spicy plugin [uses as well](https://github.com/zeek/zeek/blob/0b940d8b3ce8ebb5539b99b8eea460fd6eac3cdc/src/spicy/manager.cc#L694) so using `Analyzer::register_for_port(s)` should work, and be much easier redef'able if it was in script land (in particular since there seems to be no matching `unregister_for_port` for `Manager::UnregisterAnalyzerForPort`.

bbannier:
@initconf is this something you could take for a test run?

src/analyzer/protocol/ldap/ldap.evt
awelzel: I suggest to move the integrated (and possibly packaged version) of the analyzer to have ports/registration done in Zeek scripts then, even if it's more verbose. Besides redef, it also allows "in-a-pinch" hacking of base scripts to add/remove ports without needing to recompile.

May even consider adapting the package-template to include the necessary boiler plate and retire `ports` from `.evt` there.

@rsmmr - do you have differing opinions/thoughts?

rsmmr: Yeah, agree, doing this in script-land is preferable. I added the EVT support early on just for convenience, but it really doesn't play along well with Zeek's normal analyzer handling.

bbannier: I have moved the port configuration to script land.

scripts/base/protocols/ldap/__load__.zeek
```diff
@@ -0,0 +1,4 @@
+@if ( have_spicy_analyzers() )
```
awelzel, line 1: What would you think about doing `@ifdef ( Analyzer::ANALYZER_LDAP )` (or the _UDP / _TCP versions) here?

I've done that over on the QUIC side, because there's a possibility of a too old OpenSSL version. That is likely not the case for LDAP, but still wondering from the consistency side.

bbannier: I'd be in favor of that, but we have two analyzers here and I cannot come up with a form to combine `@ifdef` and `||`. That means we'd either need to check only for one analyzer, or have nested `@ifed`s, both of which are nasty enough that they'd warrent a comment much more complicated than `@if ( have_spicy_analyzers() )`.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,513 @@
...
+    message_id: int &log &optional;
+
+    # LDAP version
+    version: int &log &optional;
+
+    # normalized operations (e.g., bind_request and bind_response to "bind")
+    opcode: set[string] &log &optional;
+
+    # Result code(s)
+    result: set[string] &log &optional;
```
awelzel, line 56: Given `result`, `diagnostic_message`, `object`, and `argument` are all containers, I want to suggest pluralizing them to match other logs (conn: tunnel_parents, HTTP: tags, DNS: answers, TTLs, DHCP: msg_types).

Same for `scope`, `deref`, `base_object`, etc. in `Search`.

Can consider as follow-up with feedback from others.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,513 @@
...
+  global LDAP::message: event(c: connection,
+                              message_id: int,
+                              opcode: LDAP::ProtocolOpcode,
+                              result: LDAP::ResultCode,
+                              matched_dn: string,
+                              diagnostic_message: string,
+                              object: string,
+                              argument: string);
+
+  const PROTOCOL_OPCODES = {
```
awelzel, line 126: Conventionally these would go into a `ldap/const.zeek` file and `main.zeek` would be a bit shorter.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,513 @@
...
+
+  if ( atype == Analyzer::ANALYZER_LDAP_TCP ) {
+    c$ldap_proto = "tcp";
+  }
+
+}
+@endif
+#############################################################################
+event LDAP::message(c: connection,
+                    message_id: int,
```
awelzel, line 324: Quick googling indicates message_id is not negative, so this could be a count. Do you know why it wasn't/isn't one?

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,513 @@
...
+                    result: LDAP::ResultCode,
+                    matched_dn: string,
+                    diagnostic_message: string,
+                    object: string,
+                    argument: string) {
+
+  if (opcode == LDAP::ProtocolOpcode_SEARCH_RESULT_DONE) {
+    set_session(c, message_id, opcode);
+
+    if ( result != LDAP::ResultCode_Undef ) {
```
awelzel, line 335: This might look a bit more tidy if `ldap_searches[message_id]` and `ldap_messages[message_id]` was stored in a local variable upfront. Perf is another argument, through without a large test pcap hard to quantify.

Nothing to address, seems a consistent pattern across all event handlers.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,513 @@
...
+  if (authType == LDAP::BindAuthType_BIND_AUTH_SIMPLE) {
+    add c$ldap_messages[message_id]$opcode[BIND_SIMPLE];
+  } else if (authType == LDAP::BindAuthType_BIND_AUTH_SASL) {
+    add c$ldap_messages[message_id]$opcode[BIND_SASL];
+  }
+
+}
+
+#############################################################################
+event connection_state_remove(c: connection) {
```
awelzel, line 478: The conventional way is to use conn-removal hooks instead of connection_state_remove handlers (this handler is invoked also for non-ldap connections).

The package template was adapated, but I'm sure this code stems from before.

awelzel (commented):
I've left a few high-level comments/thoughts on the `main.zeek` file.

The record/field naming is the bigger topic as these are fairly user visible through policy hooks and the logs. My current understanding is that we could iterate till Zeek 7.0 without too much consequence though.

The rest is mostly thoughts for which we could create follow-ups.

Let me know if you'd rather have it merge as is and follow-up with a separate PR, or address some things.

scripts/base/protocols/ldap/main.zeek
bbannier: Unsure, but it looks like we only need half the range of `uint32` (though our `int` range is big enough). I tried changing it, but since this value comes from ASN.1 this is more than simply swapping the types. I'd prefer to leave this as a follow-up if it ever becomes relevant.

testing/btest/Baseline/scripts.base.protocols.ldap.attributes/conn.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	conn
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents
+#types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	count	string	count	count	count	count	set[string]
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	10.0.0.1	25936	10.0.0.2	3268	tcp	ldap_tcp	181.520479	258	188	RSTO	0	ShADdR	8	590	4	360	-
```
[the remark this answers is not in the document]
bbannier: This seems to be #3359 (just migrated from zeek/spicy-plugin). I am not sure how to do this in script land, so I'd prefer to defer this.

testing/btest/Baseline/coverage.record-fields/out.default
```diff
@@ -360,6 +360,39 @@ connection {
...
+        * id: record conn_id, log=T, optional=F
+            conn_id { ... }
+        * message_id: int, log=T, optional=T
+        * proto: string, log=T, optional=T
+        * result: set[string], log=T, optional=T
+        * result_count: count, log=T, optional=T
+        * scope: set[string], log=T, optional=T
+        * ts: time, log=T, optional=F
+        * uid: string, log=T, optional=F
+       }
```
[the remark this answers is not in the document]
bbannier: I removed `ldap_proto` and moved `ldap_searches` and `ldap_messages` into a new `LDAP::Info` (which also removes the weird namespacing-in-identifier.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,365 @@
...
+    id: conn_id &log;
+
+    # Message ID
+    message_id: int &log &optional;
+
+    # LDAP version
+    version: int &log &optional;
+
+    # normalized operations (e.g., bind_request and bind_response to "bind")
+    opcode: set[string] &log &optional;
```
awelzel, line 55: Oh, you did the pluralization for the other ones. Did you skip this one purposely?

scripts/base/protocols/ldap/main.zeek
```diff
@@ -0,0 +1,365 @@
...
+    diagnostic_messages: vector of string &log &optional;
+
+    #  a string representation of the search filter used in the query
+    filter: string &log &optional;
+
+    # a list of attributes that were returned in the search
+    attributes: vector of string &log &optional;
+  };
+
+  type Info: record {
```
awelzel, line 109: Convention is to call these `State`, usually `Info` is meant to be logged. See http or mqtt.

awelzel (approved):
I had two more follow-ups, but otherwise lets move this in?

awelzel (approved):

timwoj:
is there something docs related that should have merged with this?

From https://github.com/zeek/zeek/actions/runs/6472361980/job/17572847049 (and any other PR that included these changes):
```
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::ProtocolOpcode`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::ProtocolOpcode`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::ResultCode`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::ResultCode`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::SearchDerefAlias`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::SearchDerefAlias`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::SearchScope`"
WARNING: scripts/base/protocols/ldap/consts.zeek: unknown target for ":zeek:type:`LDAP::SearchScope`"
WARNING: scripts/base/protocols/ldap/main.zeek: unknown target for ":zeek:type:`LDAP::ProtocolOpcode`"
WARNING: scripts/base/protocols/ldap/main.zeek: unknown target for ":zeek:type:`LDAP::ProtocolOpcode`"
WARNING: scripts/base/protocols/ldap/main.zeek: unknown target for ":zeek:type:`LDAP::ResultCode`"
WARNING: scripts/base/protocols/ldap/main.zeek: unknown target for ":zeek:type:`LDAP::ResultCode`"
```

Issue #3306, opened 2023-06-21
Docs build fails with spicy built-in


timwoj:
Our docs build action passes `--disable-spicy` to configure. Since we're now including the spicy plugin in the Zeek repo and we have a couple of built-in spicy parsers, I thought I'd try a docs build with it enabled. The generation step using `ci/update-zeekygen-docs.sh` works fine, but `make html` in the docs submodule fails afterwards. A log of the make step is below.

```
mkdir -p build/html
sphinx-build -j auto -b html  . ./build/html
Running Sphinx v4.3.2
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 784 source files that are out of date
updating environment: [new config] 784 added, 0 changed, 0 removed
[...]

PR #3307, opened 2023-09-18
Fix Zeekygen with Spicy

rsmmr:
This fixes the issues in #3306 and also integrates our built-in Spicy analyzers into the generated output now, just like the BinPAC analyzers are.


src/spicy/manager.cc
```diff
@@ -47,6 +47,19 @@ static std::pair<std::string, std::string> parseID(const std::string& s) {
...
+    _module_info =
+        std::make_unique<zeekygen::detail::SpicyModuleInfo>(name, description, static_cast<time_t>(mtime.seconds()));
+}
+
+void Manager::registerSpicyModuleEnd() {
+    if ( ! _module_info )
+        return;
+
+    detail::zeekygen_mgr->AddSpicyModule(std::move(_module_info));
+    _module_info = 0;
```
timwoj, line 60: Use `nullptr` or `.reset()` instead of setting to zero.

testing/btest/spicy/zeekygen.zeek
```diff
@@ -0,0 +1,49 @@
...
+#
+# @TEST-EXEC: spicyz -D zeek -o test.hlto doc.spicy ./doc.evt >output 2>&1
+# @TEST-EXEC: cat output | grep 'module.s documentation' >output1
+# @TEST-EXEC: btest-diff output1
+#
+# @TEST-EXEC: unset ZEEK_DISABLE_ZEEKYGEN && zeek -X zeekygen.conf test.hlto %INPUT
+# @TEST-EXEC: cat protocol.rst  | sed -n '/_plugin-foo-bar/,/_plugin/p' | sed '$d' >output2
+# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff output2
+#
+# @TEST-DOC: Check that Spicy's tells Zeeygen about its analyzers.
```
timwoj, line 11:
```suggestion
# @TEST-DOC: Check that Spicy's tells Zeekygen about its analyzers.
```

src/spicy/manager.cc
```diff
@@ -47,6 +47,19 @@ static std::pair<std::string, std::string> parseID(const std::string& s) {
 
 Manager::~Manager() {}
 
+void Manager::registerSpicyModuleBegin(const std::string& name, const std::string& description, hilti::rt::Time mtime) {
+    _module_info =
+        std::make_unique<zeekygen::detail::SpicyModuleInfo>(name, description, static_cast<time_t>(mtime.seconds()));
```
bbannier, line 52: This doesn't handle (accidental?) nesting of calls to `register_spicy_module_begin`. At a minimum we should `assert(!_module_info)` here.

That being said, I wonder whether we need the complications of `begin`/`end` functions at all. Couldn't we instead if needed pass a module (e.g., via ID) to the various register functions so can trigger accumulation of state and then only have a single commit function by ID (currently in the end function)?

bbannier (commented):
Left only some high-level comments.

The interesting side-effect of switching on Spicy analyzers for Zeek is that the documentation for the legacy analyzers becomes inaccessible. This correctly reflects the defaults, but is also not ideal and I am not sure there is currently a good way around that (exposing documentation which depends on the Zeek configuration).

testing/btest/spicy/zeekygen.zeek
```diff
@@ -0,0 +1,49 @@
...
+    version : /[^-]*/;
+    dash    : /-/;
+    software: /[^\r\n]*/;
+};
+# @TEST-END-FILE
+
+# @TEST-START-FILE doc.evt
+
+%doc-id = Foo::Bar;
+%doc-description = "Just a \"test\" analyzer.h";
```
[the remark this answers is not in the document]
rsmmr: Yep, agree, we should be able to pull that out of EVT eventually. But I think we can postpone that. For now, this is just replicating the state we've had for the legacy analyzers, which take their description out of the the plugin's meta data, which is similarly filled as a simple string hardcoded in C++.

Same for the event documentation, btw: Right now, it's coming out of `.zeek` files manually prototyping the events once more; could pull that out of docstrings inside the EVT as well. But I'll likewise leave that for later.

src/spicy/manager.cc
rsmmr: I'll add the `assert`, nesting shouldn't be able to occur.

I thought about adding new parameters to the register functions, but figured bracketing them with a pair of functions enabling tracking of module info, would actually be easier / less complex.

rsmmr:
> The interesting side-effect of switching on Spicy analyzers for Zeek is that the documentation for the legacy analyzers becomes inaccessible. This correctly reflects the defaults, but is also not ideal and I am not sure there is currently a good way around that (exposing documentation which depends on the Zeek configuration).

I think that's fine. We want to move to a world where Spicy is always available and the disable-spicy flag will eventually go away. And for now, I don't think we want/should document the legacy analyzers anymore, so not worth the trouble keeping that working.

src/zeekygen/SpicyModuleInfo.h
```diff
@@ -0,0 +1,67 @@
...
+#include "zeek/plugin/Plugin.h"
+#include "zeek/zeekygen/Info.h"
+
+namespace zeek::zeekygen::detail
+	{
+
+/**
+ * Information about a Spicy EVT module.
+ */
+class SpicyModuleInfo : public Info
```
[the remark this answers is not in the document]
rsmmr: I wouldn't say this is a mix of `Info` and `Plugin`. It's an implementation of the `Info` API that happens to be similar to what a `Plugin` provides as well (but from a different source, i.e., Spicy). Merging it with Plugin would be more confusing than it helps. And I think the only place where it requires additional handling is the new `InfoMap`; but that seems consistent with how other implementations of the `Info` API handle it. In short: I think a new type is appropriate here. :-)

Issue #3262, opened 2023-08-31
Zeek dictionaries do not shrink


0xxon:
Zeek dictionaries currently never seem to be able to shrink in size. This can especially be noticeable when loading data with the Input framework; consider, e.g., a blocklist that temporarily balloons to several million IP addresses, and then is nearly completely emptied out again.

From what I can tell, the datastructures for Zeek dictionaries currently only support upsizing; when the amount of data inside a dictionary falls, the dictionary data structures are never freed.

This was noticed by @initconf; in this case the data structures for the (near empty) dictionary seem to occupy several Gb of RAM - which is not great, especially if it happens on a per-worker basis.

On a more general note - I wonder if we should consider switching out our dict implementation with something like Google::sparse/dense_hash_map. I only took a very cursory glance - but they do seem to support the one feature that we need: removal of elements while looping over a map, without it being invalidated.

initconf:
Thanks @0xxon  

Here attached is the script on how to reproduce   (I was using input-framework for this but I think any volumed dictionary use should show same behavior).
[ I need to work with about 2-5 Million IPs in a blacklist - thats when I noticed the issue flare up]

So here, 
1. I am loading 16M IPs using input-framework to a table (aka dictionary)  - Memory grows to 15GB 
2. Then I delete all but 10 entries - memory comes down to 6GB 
3. I reload all 16M IPs - memory again goes back up.  Here are the actions 

```
$ zeek ../test-input-framework-volume.zeek
end_of_data: name is ip_ban_feed source is blacklist
TOR:: NEW_COUNT: 16777216, CHANGE_COUNT: 0, REMOVED_COUNT: 0
size of tables: ip_ban_feed: 16777216

[ delete all but 10 entires ] 
end_of_data: name is ip_ban_feed source is blacklist
TOR:: NEW_COUNT: 0, CHANGE_COUNT: 0, REMOVED_COUNT: 16777206
size of tables: ip_ban_feed: 10

[ reload ] 
end_of_data: name is ip_ban_feed source is blacklist
TOR:: NEW_COUNT: 16777216, CHANGE_COUNT: 0, REMOVED_COUNT: 0
size of tables: ip_ban_feed: 16777216
q

^Creceived termination signal
end_of_data: name is ip_ban_feed source is blacklist
TOR:: NEW_COUNT: 10, CHANGE_COUNT: 0, REMOVED_COUNT: 0
size of tables: ip_ban_feed: 16777216
```
Here is memory using top 
```
[...]

PR #3312, opened 2023-09-19
POC: Replace Dictionary with std::map/unordered_map

timwoj:
This is a proof-of-concept I've been working on a little bit to prototype replacing our `Dict` type with `std::map`/`std::unordered_map` (or one of the many third-party hashtable/dictionary types). I started with `file_analysis::AnalyzerSet` since it's a very simple usage of `Dict`, and it's easy to rework to use a different type. It adds the following to all of the `Val` types:
- A `Hash()` method that computes a `std::size_t` hash value suitable for use in hash table types such as `std::unordered_map`.
- A 'IsSameAs()` method that computes whether two Val objects are semantically the same.

These methods for some of the Val types are currently unimplemented since I only finished the ones necessary to make the changes to `AnalyzerSet` work. The rest shouldn't be hard to fill out. The container types such as `ListVal`, `RecordVal`, etc parse down to their constituent parts to build up a correct representation of the hash or equality.

It also adds two methods to `ListVal` to be passed during the creation of the `unordered_map` in `AnalyzerSet`.

I haven't run any benchmarks on this at all, but I'm looking forward to seeing what @awelzel's recent changes to zeek-benchmarker says about it.

This is somewhat related to https://github.com/zeek/zeek/issues/3262.

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -41,6 +41,7 @@
 #include "zeek/broker/Manager.h"
 #include "zeek/broker/Store.h"
 #include "zeek/threading/formatters/detail/json.h"
+#include "zeek/util.h"
 
 using namespace std;
 
@@ -667,6 +668,16 @@ ValPtr AddrVal::DoClone(CloneState* state) {
     return {NewRef{}, this};
 }
 
+size_t AddrVal::Hash() const {
+    auto h = addr_val->MakeHashKey();
+    return h->Hash();
+}
+
+bool AddrVal::IsSameAs(const Val& other) const {
+    auto op = other.AsAddrVal();
+    return *addr_val == *op->addr_val;
+}
+
 SubNetVal::SubNetVal(const char* text) : Val(base_type(TYPE_SUBNET)) {
     subnet_val = new IPPrefix();
 
@@ -733,6 +744,16 @@ bool SubNetVal::Contains(const IPAddr& addr) const { return subnet_val->Contains
 ValPtr SubNetVal::DoClone(CloneState* state) {
     // Immutable.
     return {NewRef{}, this};
+}
+
+size_t SubNetVal::Hash() const {
+    auto h = subnet_val->MakeHashKey();
+    return h->Hash();
+}
+
+bool SubNetVal::IsSameAs(const Val& other) const {
+    auto op = other.AsSubNetVal();
+    return *subnet_val == *op->subnet_val;
 }
 
 StringVal::StringVal(String* s) : Val(base_type(TYPE_STRING)) { string_val = s; }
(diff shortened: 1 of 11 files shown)


timwoj:
There's an issue with the dump-events baseline between debian 11-based CI hosts and the debian 10 one, where if I fix the baseline on one, it breaks the other. I'm guessing it's just an ordering issue but I haven't looked too deeply.

src/Val.cc
```diff
@@ -2762,6 +2846,24 @@ void TableVal::RebuildTable(ParseTimeTableState ptts) {
         Assign(std::move(key), std::move(val));
 }
 
+size_t TableVal::Hash() const {
```
awelzel, line 2849: Just wondering, so this would allow to put `TableVal` into a `std::set` for example?

But if you get a reference to a specific entry and modify it by adding/removing an entry, the hash would change without the `std::set` noticing or putting it in a different bucket. Isn't this dangerous?

src/file_analysis/AnalyzerSet.cc
```diff
@@ -145,10 +155,15 @@ file_analysis::Analyzer* AnalyzerSet::InstantiateAnalyzer(const Tag& tag, Record
...
+void AnalyzerSet::Insert(file_analysis::Analyzer* a, const Tag& tag, RecordValPtr args) {
     DBG_LOG(DBG_FILE_ANALYSIS, "[%s] Add analyzer %s", file->GetID().c_str(),
             file_mgr->GetComponentName(a->Tag()).c_str());
-    analyzer_map.Insert(key.get(), a);
+
+    auto lv = make_intrusive<ListVal>(TYPE_ANY);
+    lv->Append(tag.AsVal());
+    lv->Append(std::move(args));
+
+    analyzer_map.insert({lv, a});
```
awelzel, line 166: Instead of `ListVal`, could the key be a `std::pair<zeek::Tag, RecordValPtr>` or `std::tuple` as the key instead?

Though, the fact that `args` is a pointer and can be mutated after having been used as a key into the `analyzer_map` seems dangerous/unexpected. Isn't it?

awelzel (commented):
My main thought is that computing a hash of a mutable container opens up a lot of possibilities to get things wrong :-) 

It seems you'd allow something that for example Python explicitly forbids:

```
In [4]: x = {}
In [5]: s = set([1,2,3])
In [6]: x[s] = 1
---------------------------------------------------------------------------
TypeError: unhashable type: 'set'
```

In Zeek script this works, but only because of the CompositeHash/RecoverVal logic. IIUC the approach here correctly, it would allow it, but then mutation the used key value results in fairly undefined behavior?

For non-container Vals, having hashing primitives available seems valuable though, for containers, I'd say no, but maybe that's me just being to timid.

JustinAzoff:
fwiw in python there's the not often used `frozenset` for this exact use case:

```
>>> x = {}
>>> s = frozenset([1,2,3])
>>> x[s] = 1
>>> x
{frozenset({1, 2, 3}): 1}
>>>
```

timwoj:
Closing this as I don't have the cycles to work on it. I might pick it back up at some point in the future, but no reason to leave the PR open until then.

PR #3315, opened 2023-09-20
Spicy: Fix support for exporting bitfields.

### LINKED ISSUE #3298: Add integration for Spicy `bitfield` fields mapped into parent scope
With spicy/spicy#1468 we added support for automatically mapping fields in anonymous `bitfield`s into the parent scope, e.g., for a Spicy value `foo` of type

```ruby
type Foo = unit {
  : bitfield(8) {
    x: 0..3;
    y: 4..7;
  };
};
```

we would resolve `foo.x` or `foo.y` to the field in the `bitfield`.

We should update our Spicy integration so it preserves this when types like `Foo` are automatically exported to Zeek.

rsmmr:
Anonymous bitfields now have their fields lifted into the surrounding record.

The template magic isn't exactly pretty, but the best we can do right now without larger changes. What we should eventually do is replace the `__visit()` methods for structs with usage of our type information.

Closes #3298.


src/spicy/spicyz/glue-compiler.cc
```diff
@@ -1475,6 +1507,23 @@ struct VisitorZeekType : hilti::visitor::PreOrder<hilti::Result<hilti::Expressio
...
+            if ( auto bf = f.type.tryAs<hilti::type::Bitfield>(); bf && f.is_anonymous ) {
+                for ( const auto& b : bf->bits() ) {
+                    auto ztype = createZeekType(b.itemType());
+                    if ( ! ztype )
+                        return ztype.error();
+
+                    fields.emplace_back(create_record_field(b.id(), *ztype, f.is_optional, export_.log));
+                }
+
+                continue;
```
bbannier, line 1520: This `continue` is hidden pretty well. Can we instead make this more explicit, e.g.,

```
if (is_bitfield) {
    ...
}

else if (is_anon) {
    ...
}

else {
    // Existing handling
}
```

src/spicy/runtime-support.h
```diff
@@ -818,6 +820,39 @@ inline ValPtr to_val(const T& t, TypePtr target) {
...
+    if ( target->Tag() != TYPE_RECORD )
+        throw TypeMismatch("bitfield", target);
+
+    auto rtype = cast_intrusive<RecordType>(target);
+
+    if ( sizeof...(Ts) - 1 != rtype->NumFields() )
+        throw TypeMismatch("bitfield", target);
+
+    auto rval = make_intrusive<RecordVal>(rtype);
+    int idx = 0;
```
bbannier, line 840:
```suggestion
    size_t idx = 0;
```

src/spicy/runtime-support.h
```diff
@@ -838,13 +873,35 @@ inline ValPtr to_val(const T& t, TypePtr target) {
...
             throw TypeMismatch(hilti::rt::fmt("no matching record field for field '%s'", name));
 
-        auto field = rtype->GetFieldType(idx);
-        std::string field_name = rtype->FieldName(idx);
+        // Special-case: Lift up anonymous bitfields (which always come as std::optionals).
+        if ( name == "<anon>" ) {
+            using X = typename std::decay<decltype(val)>::type;
+            if constexpr ( is_optional<X> ) {
+                if constexpr ( std::is_base_of<::hilti::rt::trait::isBitfield, typename X::value_type>::value ) {
+                    int j = 0;
```
bbannier, line 881:
```suggestion
                    size_t j = 0;
```

testing/btest/spicy/export-bitfield.zeek
```diff
@@ -0,0 +1,50 @@
...
+
+on foo::X -> event foo::hello(self);
+# @TEST-END-FILE
+
+event zeek_init() {
+    print record_fields("foo::X");
+}
+
+event foo::hello(x: foo::X)
+	{ print x; }
```
bbannier, line 49: It would be great to also baseline the type of `foo::X::bf`, e.g., like this:

```suggestion
event foo::hello(x: foo::X)
	{
	print record_fields(x);
	print record_fields(x$bf);
	}
```

bbannier (approved):

PR #3317, opened 2023-09-21
RecordVal/VectorVal: Embed record_val and vector_val

awelzel:
This should remove one malloc/free per created and destroyed record instance and avoid one extra pointer indirection to access fields.

Separate motivation of this changeset is to test `zeek-benchmarker`, actually :-)

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -3155,17 +3155,15 @@ RecordVal::RecordVal(RecordTypePtr t, bool init_fields) : Val(t), is_managed(t->
 	if ( run_state::is_parsing )
 		parse_time_records[rt.get()].emplace_back(NewRef{}, this);
 
-	record_val = new std::vector<std::optional<ZVal>>;
-
 	if ( init_fields )
 		{
-		record_val->resize(n);
+		record_val.resize(n);
 
 		for ( auto& e : rt->CreationInits() )
 			{
 			try
 				{
-				(*record_val)[e.first] = e.second->Generate();
+				record_val[e.first] = e.second->Generate();
 				}
 			catch ( InterpreterException& e )
 				{
@@ -3177,21 +3175,19 @@ RecordVal::RecordVal(RecordTypePtr t, bool init_fields) : Val(t), is_managed(t->
 		}
 
 	else
-		record_val->reserve(n);
+		record_val.reserve(n);
 	}
 
 RecordVal::~RecordVal()
 	{
-	auto n = record_val->size();
+	auto n = record_val.size();
 
 	for ( unsigned int i = 0; i < n; ++i )
 		{
-		auto f_i = (*record_val)[i];
+		auto f_i = record_val[i];
 		if ( f_i && IsManaged(i) )
 			ZVal::DeleteManagedType(*f_i);
 		}
-
-	delete record_val;
 	}
 
 ValPtr RecordVal::SizeVal() const
@@ -3206,7 +3202,7 @@ void RecordVal::Assign(int field, ValPtr new_val)
 		DeleteFieldIfManaged(field);
 
 		auto t = rt->GetFieldType(field);
-		(*record_val)[field] = ZVal(new_val, t);
+		record_val[field] = ZVal(new_val, t);
 		Modified();
 		}
 	else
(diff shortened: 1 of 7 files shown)


src/Val.h
```diff
@@ -1518,7 +1518,9 @@ class RecordVal final : public Val, public notifier::detail::Modifiable
 	RecordTypePtr rt;
 
 	// Low-level values of each of the fields.
-	std::vector<std::optional<ZVal>>* record_val;
+	//
+	// Lazily modified during GetField(), so mutable.
+	mutable std::vector<std::optional<ZVal>> record_val;
```
bbannier, line 1523: Just an observation and since you are already at it, `ZVal` can already encode an empty value (exactly what it's default ctor does), there shouldn't be a reason to additionally wrap it in an `optional` here (which adds another indirection at runtime). All you might need is to add an `explicit operator bool` to it, and adjust the usages from e.g., `*zval` to `zval`, and `zval->x` to `zval.x`.

src/Val.cc
```diff
@@ -3491,26 +3487,28 @@ ValPtr TypeVal::DoClone(CloneState* state)
...
 	yield_type = t->Yield();
 
 	auto y_tag = yield_type->Tag();
 	any_yield = (y_tag == TYPE_VOID || y_tag == TYPE_ANY);
 	managed_yield = ZVal::IsManagedType(yield_type);
 	}
 
+VectorVal::VectorVal(VectorTypePtr t, std::vector<std::optional<ZVal>>* vals) : VectorVal(t)
+	{
+	vector_val = std::move(*vals);
```
bbannier, line 3501: The previous API accepted `nullptr`.
```suggestion
	if ( vals )
		vector_val = std::move(*vals);
```

awelzel:
Posting some zeek-benchmarker results for the current version of this change.

Generally, runtime and memory usage stayed the same or got slightly better for most tests.

Except for  [micro-function-calls-many](https://github.com/zeek/zeek-benchmarker/blob/master/ci-based/scripts/microbenchmarks/function-calls/many.zeek) taking more CPU time.. I'm actually able to reproduce this locally with the default glibc allocator. Switching to jemalloc, however, the results are the same or better with the changes in this PR. And for reference, switching to jemalloc alone improves runtime for this test from ~31seconds to ~26seconds.

So, I'm shrugging this one off as artifact..

Screenshot-20230922-213755

Screenshot-20230922-213843

timwoj:
@vpax Do you remember the original reasoning for these not being directly embedded?

vpax:
> Do you remember the original reasoning for these not being directly embedded?

Yes, it's because back-in-the-day the `Val` base object held a `union` of all of the possible underlying `Val` values. In that context it was natural to use pointers rather than have objects directly present. When the underlying values were moved into the `Val` subclass objects, this was retained (which made the transition simpler).

src/Val.h
```diff
@@ -1673,26 +1672,17 @@ class VectorVal final : public Val, public notifier::detail::Modifiable
...
+	const RecordVal* RecordValAt(unsigned int index) const { return vector_val[index]->record_val; }
+	bool BoolAt(unsigned int index) const { return static_cast<bool>(vector_val[index]->uint_val); }
+	const StringVal* StringValAt(unsigned int index) const { return vector_val[index]->string_val; }
 	const String* StringAt(unsigned int index) const { return StringValAt(index)->AsString(); }
 
 	// Only intended for low-level access by internal or compiled code.
-	const auto& RawVec() const { return vector_val; }
-	auto& RawVec() { return vector_val; }
+	const std::vector<std::optional<ZVal>>& RawVec() const { return vector_val; }
+	std::vector<std::optional<ZVal>>& RawVec() { return vector_val; }
```
awelzel, line 1685: API break here intentionally: This previously returned `std::vector<std::optional<ZVal>>*&` (a reference to a pointer to a vector). Mainly taking the comment above as justification for breaking (and most users primarily only skipping the de-referencing of the pointer).

```
	// Only intended for low-level access by internal or compiled code.
```

src/Val.cc
```diff
@@ -3155,17 +3155,15 @@ RecordVal::RecordVal(RecordTypePtr t, bool init_fields) : Val(t), is_managed(t->
...
 	if ( init_fields )
 		{
-		record_val->resize(n);
+		record_val.resize(n);
 
 		for ( auto& e : rt->CreationInits() )
 			{
 			try
 				{
-				(*record_val)[e.first] = e.second->Generate();
```
timwoj, line 3168: These dereferences always bothered me, so I'm glad to see them gone.

timwoj (approved):
I think the API break is fine because it's noted that it's for internal use only.

PR #3323, opened 2023-09-22
Do not require cookie for `reject_protocol` during teardown.

bbannier:
A user reported that they ran into a situation where a parse error caused an assertion failure in `reject_protocol`. printf debugging points to the `CookieSetter`s in the `try`/`catch` blocks during processing already clearing the cookie RAII-style; since their `catch` blocks already send an analyzer violation no further sending of one from `reject_protocol` is required.

This patch replaces the assert in `reject_protocol` with a runtime check and an early return if the cookie is already cleared.


bbannier:
I am still not 100% sure how the user code arrived in this scenario and wasn't able to write a compact reproducer for test. I can dig further if we want that.

rsmmr (approved):

PR #3324, opened 2023-09-22
btest/dump-events: Run in bare mode, load conn and smtp scripts only

awelzel:
The dump-events baseline changes are pure noise and have spurred confusion for internal and external contributors. For example, adding new analyzers have perturbed orderings of sets holding analyzer tags.

Running in non-bare mode, the baselines change almost whenever any of the record types attached to connections change in the default scripts. This causes continuous and seemingly little useful updates to the baselines.

This change switches the test to run in bare mode and explicitly loads just base/protocols/conn and base/protocols/smtp. The intention is to test the misc/dump-events script, not all default scripts.

Protocol specific scripts that do want to leverage misc/dump-events for baseline creation of their own events can add additional/specific tests.

---

From #2770 :

> scripts.policy.misc.dump-events is a bit more concerning, it has the same order change but also there are some where md5 is changing from <uninitialized> to an MD5 hash.

https://github.com/zeek/zeek/pull/2770#issuecomment-1434216235


awelzel:
Preparation for #3305  and #3320, :crossed_fingers: not too contentious.

timwoj:
I wonder if this would also fix the issue I was seeing in https://github.com/zeek/zeek/pull/3312

awelzel:
> I wonder if this would also fix the issue I was seeing in #3312

If it does, it will be "fixed by ignoring" though, but you probably recognize that.

timwoj:
I'll be honest I've never really understood the point of this test. In the large majority of cases we (myself included) just update any changes that need to be made and push them through without checking what they actually were.

awelzel:
> I'll be honest I've never really understood the point of this test. In the large majority of cases we (myself included) just update any changes that need to be made and push them through without checking what they actually were.

Yeah, mostly in the same boat here, I just update the baseline and move one usually.

> I think if we're going to nerf it down to a very small subset of the events that it was outputting before, we should probably discuss whether we should keep it at all.

It's still testing the actual misc/dump-events functionality.

I can see protocol specific event dumping/testing using misc/dump-events for support, but as written in the description these should be protocol specific tests (enabling just an analyzer in question, loading just the scripts involved, bpf filtering). The current dump-events test is much too much of a catch-all and way too trigger happy.

@rsmmr / @ckreibich - mind approving this one? Benjamin's :tada: and Tim's statement make me believe this step would be for the better.

rsmmr (commented):
Matter of taste I think. The motivation for the test is primarily to act as a canary that something with events has changed. Often it's indeed less important to check every single line of the diff, and more a question of "did I expect any changes to Zeek's events?". But yeah, it is noisy and I understand that it can be confusing, too. So I'm fine either way.

JustinAzoff:
The ldap script added (kind of surprising) three different new fields to the connection record.

Those diffs probably wouldn't be so bad in word diff mode, but I don't think github supports that.

ckreibich:
I'm generally in favor of the bare-mode idea. I certainly respect the canary line of thought, but I think the balance between constant updates and valuable insights from baseline discrepancies has been pretty off. We could certainly try something else for a while and revisit the decision in the future.

I just looked through the git log of these baselines and let me flag another idea: I wonder whether we should keep the full `all-events-no-args.log`, but move the `all-events.log` family to bare mode (plus conn/smtp). Reading over the diffs it looks to me like the type and sequence of generated events is relatively interesting and ties more closely to the changes in code than the full-arg logs. Plus if we use these arguments to discover new fields in our popular types, I think we can do better with new tests that focus on highlighting these.

The plugin.hooks log is also fairly noisy but I think we should leave that alone for now.

What do you guys think?

Github does support word diffs, btw — like [here](https://github.com/zeek/zeek/pull/3305/files#diff-8dc23d8987f731df0d0a2f71e0e6e34b46f3f68d7d529ead69143c596f9aa3ec) — but somehow loses that ability in the all-events logs.

timwoj:
> Github does support word diffs, btw — like [here](https://github.com/zeek/zeek/pull/3305/files#diff-8dc23d8987f731df0d0a2f71e0e6e34b46f3f68d7d529ead69143c596f9aa3ec) — but somehow loses that ability in the all-events logs.

Usually the diffs for all-events.log and all-events-no-args.log are so large that GH hides them, which then loses the word diff when you expand it. As an aside that's probably a bug we should report to GH.

awelzel:
> I just looked through the git log of these baselines and let me flag another idea: I wonder whether we should keep the full all-events-no-args.log, but move the all-events.log family to bare mode (plus conn/smtp). Reading over the diffs it looks to me like the type and sequence of generated events is relatively interesting and ties more closely to the changes in code than the full-arg logs

> What do you guys think?

As in our direct conversation, there's still curious file_hash permutations that are hard to understand. Further, unless we use a "super pcap" that raises all possible events, the coverage won't be great. I'd still push towards protocol/analyzer specific/focused testing with specific pcaps.

I added a test in `coverage/` to output the connection record recursively. That should allow to detect changes to the record layout that Justin pointed out without too much "diff-fatigue".

testing/btest/coverage/record-fields.test
```diff
@@ -0,0 +1,55 @@
...
+	{
+	print fmt("%s%s {", indent, name);
+	local field_names: vector of string;
+	local fields = record_fields(rt);
+
+	for ( fn, _ in fields )
+		field_names += fn;
+	sort(field_names, strcmp);
+
+	for ( _, fn in field_names )
```
bbannier, line 22: TBH, I find this implementation which prints while recursively iterating pretty opaque and read-only. Moving that recursive part to a `record_type_to_string(rt: any): string` and collecting that from here might be more maintainable, e.g.,

```zeek
local fields_: table[string] of string;
for ( [fn], fr in record_fields(rt) )
	{ # ...
```

Not a blocker since this seems to do its job, but IMO still nice to have.

bbannier (approved):
LGTM and like a reasonable balance capturing most of the discussed trade-offs.

I left some technical comments on the new test.

testing/btest/coverage/record-fields.test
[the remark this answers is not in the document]
awelzel: Hmm, most other files have `.test` in that directory, but I'll move it over. There's already one that doesn't adhere.

testing/btest/coverage/record-fields.test
awelzel: I've extracted out one helper to determine the nested record type names, but given all the indenting the direct prints were easier.

awelzel:
> The plugin.hooks log is also fairly noisy but I think we should leave that alone for now.

I'm actually tempted to do the same: Load conn and http only given it's consuming `get.trace`. It's mostly the same arguments that apply.

PR #3329, opened 2023-09-25
Exclude dot directories when searching ZEEK_PLUGIN_PATH

bbannier:
Dot directories rarely contain anything we would want to load as a dynamic plugin. Even worse, they likely contain files with externally controlled lifetimes which might be removed while we are using them (see e.g., zeek/btest#98).

With this patch we do not search _discovered_ dot directories anymore. We continue to load from a user-specified `ZEEK_PLUGIN_PATH`, even if its name starts with a dot.

Since this patch changes previous behavior it is a **BREAKING CHANGE**.


src/plugin/Manager.cc
```diff
@@ -58,6 +58,10 @@ void Manager::SearchDynamicPlugins(const std::string& dir)
 	if ( dir.empty() )
 		return;
 
+	// Do not search dot directories.
+	if ( auto p = zeek::filesystem::path(dir); p.has_filename() && p.filename().c_str()[0] == '.' )
```
awelzel, line 62: Would move this down to line 142:

https://github.com/zeek/zeek/pull/3329/files#diff-b4ae5cd2712f7a5ffe288277de92b8b111e64b71219b542713d1f525dc5c76bbR151


Would also fix the test that does `ZEEK_PLUGIN_DIR=.`. Additionally, if someone does `ZEEK_PLUGIN_DIR=.tmp/just-do/.not-ask` we should let them.

bbannier: Thanks for the review, I also added a test validating that behavior.

awelzel (approved):
> Since this patch changes previous behavior it is a BREAKING CHANGE.

It may be hard to figure out if `.` dirs are used in a production environment, so a NEWS entry should be added before or during the merge.

We could add an env variable or maybe (not sure if possible) a script layer option to disable this behavior, but doesn't seem worth the effort unless someone complains loudly, tbh.

awelzel:
As discussed in DM, might be worth doing Spicy's recursive hlto search, too. The error in zeek/btest#98 contained a `ghc::filesystem` error which is more likely raised "over there".

bbannier:
> As discussed in DM, might be worth doing Spicy's recursive hlto search, too. The error in [zeek/btest#98](https://github.com/zeek/btest/issues/98) contained a `ghc::filesystem` error which is more likely raised "over there".

Done.

src/spicy/manager.cc
```diff
@@ -842,6 +843,39 @@ void Manager::searchModules(const std::string& paths) {
         auto it = hilti::rt::filesystem::recursive_directory_iterator(trimmed_dir, ec);
         if ( ! ec ) {
             while ( it != hilti::rt::filesystem::recursive_directory_iterator() ) {
+                // Reject loading of modules in dot directories or having a filename starting with a dot.
+                {
+                    // Compute a path relative to `dir`. This effectively removes a `dir`
+                    // prefix so that we allow dot directories as configure search paths.
+                    const auto& rel = hilti::rt::filesystem::proximate(it->path(), dir, ec);
```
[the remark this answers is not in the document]
bbannier: > Unless I'm missing something obvious, this is still traversing the full directory tree?

Yes, it was, and pruning paths after the fact. I was missing `disable_recursion_pending` and it does exactly what I want (and also allows loading modules from HLTO files starting with a dot). I added a test checking the behavior as well.

src/spicy/manager.cc
```diff
@@ -5,6 +5,7 @@
 #include <dlfcn.h>
 #include <glob.h>
 
+#include <algorithm>
```
awelzel, line 8: Can drop now.

testing/btest/spicy/module-path.spicy
```diff
@@ -1,9 +1,38 @@
...
+# @TEST-EXEC: mkdir a && mv .y a
+# @TEST-EXEC: echo "# Loading from module path '.' DOES NOT find module 'a/.y/z/x.hlto':" >>output
+# @TEST-EXEC: ZEEK_SPICY_MODULE_PATH=.y zeek -b Zeek::Spicy Spicy::enable_print=T >>output
+
+# @TEST-EXEC: mkdir b && mv a b
+# @TEST-EXEC: echo "# Loading from module path '.' DOES NOT find module 'b/a/.y/z/x.hlto':" >>output
+# @TEST-EXEC: ZEEK_SPICY_MODULE_PATH=.y zeek -b Zeek::Spicy Spicy::enable_print=T >>output
+
+# @TEST-EXEC: echo "# Sanity check: can load module from _file_ '.x.hlto' directly:" >>output
+# @TEST-EXEC: mv b/a/.y/z/x.hlto .x.hlto
```
awelzel, line 33: I could see this one go either way, but minimally the test confirms what the current code implements.

awelzel (approved):

testing/btest/spicy/module-path.spicy
bbannier: I think this should match what we do for explicitly requested paths in `ZEEK_MODULE_PATH` or `ZEEK_SPICY_MODULE_PATH` -- if a user requested a specific path he can get it, this PR just changes automatic behavior.

bbannier:
Assigned to you @awelzel for final steps. Thanks!

PR #3332, opened 2023-09-26
Topic/awelzel/ubuntu22 spicy task

awelzel:
Add two ubuntu22 variations where the external DHCP, DNS and HTTP Spicy analyzers are installed via zkg. The resulting artifacts is used for benchmarking.

Two new tests have been added for the benchmarker: DNS only and HTTP only pcaps. So improvements or regressions applicable to these analyzers should become visible.

If we want to save on Cirrus credits, the spicy_head task is mostly a nice to have. More regular spicy submodule bumps would also work.


.cirrus.yml
```diff
@@ -74,6 +74,17 @@ skip_task_on_pr: &SKIP_TASK_ON_PR
...
+
+benchmark_only_if_template: &BENCHMARK_ONLY_IF_TEMPLATE
+  # only_if condition for cron-triggered benchmarking tests.
+  # These currently do not run for release/.*
+  only_if: >
+    ( $CIRRUS_REPO_NAME == 'zeek' || $CIRRUS_REPO_NAME == 'zeek-security' ) &&
+    ( $CIRRUS_CRON == 'benchmark-nightly' ||
+      $CIRRUS_PR_LABELS =~ '.*fullci.*' ||
+      $CIRRUS_PR_LABELS =~ '.*benchmark.*' ||
+      $CIRRUS_BRANCH =~ 'topic/awelzel/ubuntu22-spicy-task' )
```
awelzel, line 86: To be removed during merge.

bbannier: I'd be curious what the ultimate condition would look like. Also curious whether it would make sense to trigger this even on PRs if something in `src/spicy` or `auxil/spicy` was touched (this would assume that analyzers do not change much).

bbannier (approved):
Looks reasonable. This task uses the same caching machinery we have elsewhere (`ccache` and Spicy compilation artifact caching via `HILTI_CXX_COMPILER`), so it should run in reasonable time.

As for the schedule, not sure it makes sense to run this nightly when we have a pretty good chance to detect whether something change in the Spicy-specific bits (`src/spicy`, `auxil/spicy`). We could use that in conjunction with [Cirrus conditional tasks](https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution).

.cirrus.yml
awelzel: I was forth and back to add `changesInclude()`, but I'm now at the point where I think the nightly signal is reasonable.

If for a certain spicy-bump or Zeek side change you do want to include the tasks, can put `benchmark` into the PR labels.

awelzel:
Added the following Cirrus Cron trigger, :crossed_fingers: 

Screenshot-20230927-165821

awelzel:
> As for the schedule, not sure it makes sense to run this nightly when we have a pretty good chance to detect whether something change in the Spicy-specific bits (src/spicy, auxil/spicy).

I would actually want to run the ubuntu22 task nightly as well to get a sense of noise/variation of the results. I should have folded that into this PR, ah well.

PR #3334, opened 2023-09-27
Var: Fix null-pointer deref on redefinition of lambdas

### LINKED ISSUE #3314: Redefining lambda leads to segfault
The following code leads to a segfault:

```zeek
event zeek_init() {
    local f = function() {};
    local f = function() {};
}
```

If instead `f` was defined as `local f = 0;` this would be rejected with `already defined (f)`. I would have expected something similar for lambdas.

I originally came across this when iterating over a `vector of function()` and the loop value outliving the loop so it clashed with a following local.

awelzel:
Closes #3314


testing/btest/Baseline/language.redeclaration-redefinition-errors-7/out
```diff
@@ -0,0 +1,2 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+error in <...>/redeclaration-redefinition-errors.zeek, line 2 and <...>/redeclaration-redefinition-errors.zeek, line 1: alternate function prototype already exists (event() and record {  })
```
timwoj, line 2: This result surprised me as neither of the two definitions are `record`s. Shouldn't this return something similar to test 5?

timwoj (changes requested):

testing/btest/Baseline/language.redeclaration-redefinition-errors-7/out
awelzel: Mostly an artifact due prototype/function parameters being represented as RecordTypes and a call to `Object::Error()` passing the RecordType as as `obj2` (which on first sight would be redundant given the error points out they are the same). Let me try pinpoint_only=true and add another test with actual parameters.

I've come across the `record { ... }` presentation for errors around function signatures before and they are indeed a bit unusual. So, I'm happy to squelch this one and be more sensitive to their occurrence in the future.

awelzel: I've re-worked it and made the `event(a:string; b:string;)` prototypes become `event(a:string, b:string)` while at it. This had been a bit of an irritation as well

timwoj: I wonder if that was a holdover from some prior way that scripts were written, and was just never fixed in the error output. It's good to have it match how the scripts actually look since now copy/paste grep works.

PR #3348, opened 2023-10-02

bbannier:


rsmmr (approved):

awelzel:
For what it's worth, this bump appears to have lowered RAM usage for processing HTTP requests `http_many` from ~1070MB to 977MB and improved processing times when DNS/HTTP spicy analyzers are in use.

Screenshot-20231004-133435

Screenshot-20231004-133826

bbannier:
Thanks for sharing this @awelzel, that's useful information.

The main thing the bumped Spicy version does it fix a big around parsers not trimming enough of their input (i.e., throwing away already processed and unneeded input data). I can see how the fix reduces RAM usage. I have more trouble understanding how this also improved throughput, but that's certainly welcome as well ;)

PR #3352, opened 2023-10-04
feat: zeek portable

r-rojo:
This change modifies necessary files to make zeek and zeek config "portable". This means that paths reported by zeek will paths relative to the zeek or zeek-config executable file instead of paths being hard-coded to those used at install time.
This makes it so that zeek installation tree can be copied to a different install prefix (as is done by dependency managers like vcpkg or conan).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -445,6 +445,7 @@ if (NOT ZEEK_SCRIPT_INSTALL_PATH)
     # set the default Zeek script installation path (user did not specify one)
     set(ZEEK_SCRIPT_INSTALL_PATH ${ZEEK_ROOT_DIR}/share/zeek)
 endif ()
+set(ZEEK_SCRIPT_RELATIVE_PATH "$ORIGIN/../share/zeek")
 
 if (NOT ZEEK_MAN_INSTALL_PATH)
     # set the default Zeek man page installation path (user did not specify one)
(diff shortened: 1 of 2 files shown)


ckreibich:
I share @bbannier's concerns over in zeek/spicy#1535 ... could you say a little more about what you're trying to achieve? Let's say that I `configure --prefix=/opt/foo`. Should I be able to say `mv /opt/foo /opt/bar`, and the whole installation still works?

I worry that this will be quite tough to pull off because path info propagates widely through the ecosystem, partially relying on cmake, partially on `zeek-config`, and partially by getting baked in. Do an install to some prefix and then recursively grep for that prefix in the installation — there's a lot. We've also historically found it very difficult to make such changes consistently and completely.

Before we explore this further let me ask: could there be simpler solutions that sidestep this, like moving the installation to other directories but keeping the original installation directory as a symlink? Something like management via the `alternatives` system? Or lightweight use of Linux namespacing to keep colliding installations from stepping on each other via `unshare`?

r-rojo:
@ckreibich , yes, you've pretty much summed it up but there's also the use case when an installation gets archived in a binary package registry. With the patches I've made (so far), I've been able to get it working (somwhat) using vcpkg to resolve/install zeek. Yes, you're right, there are a lot of absolute paths baked into an installation and vcpkg even warns about that too. As far as cmake config files, I haven't taken a close enough look at the files that get installed but I was taking it for granted that those files used relative paths when "finding zeek" (and resolving paths) from a different cmake project . I could get by with continuing to include patches with the vcpkg installation recipe, however, I think longer term it would be beneficial to everyone if zeek (and the ecosystem) installations were portable.

r-rojo:
I also wanted to add that I do have a patch for zeek-config (based on 6.0 release) but couldn't find it in the latest codebase.

bbannier:
> I also wanted to add that I do have a patch for zeek-config (based on 6.0 release) but couldn't find it in the latest codebase.

`zeek-config` is [generated by CMake](https://github.com/zeek/zeek/blob/0630bdbbd5f71e80a9bf31c3957e03ffeb184b58/CMakeLists.txt#L1175-L1177) from the template [here](https://github.com/zeek/zeek/blob/master/cmake_templates/zeek-config.in).

ckreibich:
@r-rojo I am going to go ahead and close this PR. I am not shooting down the idea, since I can see the utility! But, I also admit that I remain skeptical about doing this with an ecosystem as complex as Zeek's. So let's first talk this over and understand the requirements a little better. If we decide to move ahead, we'll need to figure out a model where all of the required changes live in a single toplevel branch with submodules pointing to the corresponding changes in those submodules so the whole thing becomes testable as one thing — this simply won't work as a set of independent PRs.

PR #3355, opened 2023-10-09
iosource/pcap: Support configurable buffer size

awelzel:
On Linux with a default ext4 or tmpfs filesystem, the default buffer size for
reading a pcap is chosen as 4k (strace/gdb validated). When reading large pcaps
containing raw data transfers, the syscall overhead for read becomes visible
in profiles. Support configurability of the buffer size and default to 128kb.

When processing a ~830M PCAP (16 UDP connections, each transferring ~50MB) in
bare mode, this change improves runtime from 1.39 sec to 1.29 sec. Increasing
the buffer further didn't provide a noticeable boost.


src/iosource/pcap/Source.cc
```diff
@@ -176,7 +178,32 @@ void PcapSource::OpenOffline()
...
+		}
+	else
+		{
+		if ( ! (f = fopen(props.path.c_str(), "rb")) )
+			{
+			Error(util::fmt("unable to open %s: %s", props.path.c_str(), strerror(errno)));
+			return;
+			}
+
+		if ( BifConst::Pcap::bufsize_offline_bytes > 0 )
```
bbannier, line 194: `BifConst::Pcap::bufsize_offline_bytes` is an unsigned so this check seems redundant.

awelzel: An unsigned int can still be 0. But maybe `!= 0` would be better?

awelzel:
Small, but visible improvement for the benchmarker tests reading from PCAPs:

Screenshot-20231009-141831

scripts/base/init-bare.zeek
```diff
@@ -5286,6 +5286,10 @@ export {
 	## interfaces.
 	const bufsize = 128 &redef;
 
+	## Number of bytes to use for buffering PCAP file read operations.
+	## Setting this to 0 uses operating system defaults (likely 4096).
+	const bufsize_offline_bytes = 131072 &redef;
```
bbannier, line 5291:
```suggestion
	const bufsize_offline_bytes = 128 * 1024 &redef;
```

scripts/base/init-bare.zeek
```diff
@@ -5286,6 +5286,10 @@ export {
 	## interfaces.
 	const bufsize = 128 &redef;
 
+	## Number of bytes to use for buffering PCAP file read operations.
+	## Setting this to 0 uses operating system defaults (likely 4096).
```
bbannier, line 5290: I feel like this needs to do more explaining on how to arrive at a locally good value (and TBH I am not convinced the new default couldn't lead to perf regressions in typical setups). Things to take into account are probably fs block size and the CPU cache sizes.

src/iosource/pcap/Source.cc
```diff
@@ -176,7 +178,32 @@ void PcapSource::OpenOffline()
...
 
-	pd = pcap_open_offline(props.path.c_str(), errbuf);
+	FILE* f = nullptr;
+	if ( props.path == "-" )
+		{
+		f = stdin;
+		}
+	else
+		{
+		if ( ! (f = fopen(props.path.c_str(), "rb")) )
```
bbannier, line 188: My version of clang-tidy complains about this construct
```
An assignment within an 'if' condition is bug-prone
```

The following form seems more readable:

```suggestion
		if ( f = fopen(props.path.c_str(), "rb"); ! f )
```

src/iosource/pcap/Source.cc
```diff
@@ -176,7 +178,32 @@ void PcapSource::OpenOffline()
...
+			{
+			Error(util::fmt("unable to open %s: %s", props.path.c_str(), strerror(errno)));
+			return;
+			}
+
+		if ( BifConst::Pcap::bufsize_offline_bytes != 0 )
```
bbannier, line 194: I feel comment calling out that this implements the behavior where `bufsize_offline_bytes=0` would lead to use of the system default buffer size here would be helpful.

src/iosource/pcap/Source.cc
```diff
@@ -176,7 +178,32 @@ void PcapSource::OpenOffline()
...
+			if ( util::detail::setvbuf(f, iobuf.data(), _IOFBF, iobuf.size()) != 0 )
+				{
+				Error(util::fmt("unable to setvbuf %s: %s", props.path.c_str(), strerror(errno)));
+				fclose(f);
+				return;
+				}
+			}
+		}
+
+	pd = pcap_fopen_offline(f, errbuf);
```
bbannier, line 206: This currently leaks `f` since we do not seem to call `::fclose` on it for the success path.

src/iosource/pcap/Source.h
```diff
@@ -44,6 +45,8 @@ class PcapSource : public PktSrc
 
 	pcap_t* pd;
 	struct pcap_stat prev_pstat = {0};
+
+	std::vector<char> iobuf;
```
bbannier, line 49: This should have a comment documenting what this is used for under what circumstances.

scripts/base/init-bare.zeek
awelzel: Hmm, if this causes a regression we learned something and it can be tweaked. IMO this is not at a level where CPU cache size should be relevant. The main effect is reduction the number of `read()` system calls by a factor of 32 for large pcaps in usual configurations (Linux + default ext4) at a minor increase of memory usage (128kb in 2023).

As a compromise, could default to `0`, but then I can almost keep it as a private patch for my own benefit :-)

src/iosource/pcap/Source.cc
awelzel: Yeah, that looks much better, thanks!

src/iosource/pcap/Source.cc
awelzel: `pcap_fopen_offline` takes ownership of `f`.

bbannier: That seems to be undocumented behavior, https://linux.die.net/man/3/pcap_open_offline.

awelzel: Actually - in the error case (when there's no actual pcap content in the file) we're leaking `f` I suspect. Let me see if this triggers on a sanitizer build.

scripts/base/init-bare.zeek
awelzel: @J-Gras chimed in mentioning he investigated a similar optimization for a PCAP load-balancer/splitter. He also provided a link to coreutils code that also uses 128kb today by default:

https://github.com/coreutils/coreutils/blob/9581c4b59d2e11cd090613cd300a833b4441d2b5/src/ioblksize.h#L24-L58

timwoj: Based on the script in that code, going even larger is better on an M1:

```
tim@coregeek2 build% for i in $(seq 0 10); do
     bs=$((1024*2**$i))
     printf "%7s=" $bs
     timeout --foreground -sINT 2 \
       gdd bs=$bs if=/dev/zero of=/dev/null 2>&1 \
         | sed -n 's/.* \([0-9.]* [GM]B\/s\)/\1/p'
   done
   1024=1.2 GB/s
   2048=2.4 GB/s
   4096=4.3 GB/s
   8192=8.0 GB/s
  16384=13.8 GB/s
  32768=20.8 GB/s
  65536=28.8 GB/s
 131072=34.9 GB/s
 262144=39.6 GB/s
 524288=41.7 GB/s
1048576=43.7 GB/s
```

I agree with the argument for going to a larger default though. The current default of 4k is likely too small for modern systems.

timwoj (approved):

PR #3361, opened 2023-10-10
RFC: Reformat code using Allman bracing style instead of Whitesmiths [skip ci]

timwoj:
This is an RFC version for switching our formatting to Allman mode. It's effectively just moving the braces one column to the left, but there may be other changes as well.


vpax:
It would be handy to see a version of these diffs that suppresses changes that are only due to whitespace, which right now overwhelmingly dominate looking at the changes. (Same for @bbannier 's example run.)

timwoj:
It's possible to do that from the Github interface by adding `?w=1` to the URL.

vpax:
> It's possible to do that from the Github interface by adding `?w=1` to the URL.

Aha! Great to know, thanks. The changes then look non-scary :-).

timwoj:
Closing this in favor of https://github.com/zeek/zeek/pull/3360

PR #3362, opened 2023-10-11
spicy/manager: Ensure Zeekygen knows identifier for registered types

awelzel:
Without this, Zeekygen won't generate documentation about the added
items as it can not resolve the identifier. Also, only register a
type as item with the Spicy plugin if there's no module currently
active.

---

I've also added it for events. As mentioned in the comment that won't be
pretty, but neither is just not including the names of the events in .evt files.
I haven't looked close enough to understand if we could derive a nicer
signature. I did drop the explicit `: void`.


awelzel:
@rsmmr , on second thought I'm happy to drop the hunk about registering dummy events with Zeekygen. This might actually be more misleading than just not showing them at all.

src/spicy/manager.cc
```diff
@@ -616,7 +619,7 @@ void Manager::InitPostScript() {
     for ( const auto& [name, id] : _events ) {
         if ( ! id->GetType() ) {
             auto args = make_intrusive<RecordType>(new type_decl_list());
-            auto et = make_intrusive<FuncType>(std::move(args), base_type(TYPE_VOID), FUNC_FLAVOR_EVENT);
+            auto et = make_intrusive<FuncType>(std::move(args), nullptr, FUNC_FLAVOR_EVENT);
```
rsmmr, line 622: This doesn't change anything, does it?

rsmmr (approved):

src/spicy/manager.cc
awelzel: Uh, I missed the comment, sorry. With `base_type(TYPE_VOID)` Zeekygen rendered events as: `x: event(): void`, while with `nullptr`, it renders them as the usual `x: event()` style instead.

It doesn't make a difference currently (after removing the identifier registration of the dummy events), because these are never rendered.

We do reset yield for FUNC_FLAVOR_EVENT to nullptr, too:

https://github.com/zeek/zeek/blob/01bf1d47642b7bd9cc59bda06065570cb2eae303/src/Var.cc#L680-L690

https://github.com/zeek/zeek/blob/01bf1d47642b7bd9cc59bda06065570cb2eae303/src/Type.h#L484-L488

PR #3363, opened 2023-10-11
ci/update-zeekygen-docs.sh: Do output stderr by default

awelzel:
stderr was only produced when there was a real failure as it had been a
bit noisy previously, but that has hidden actual problems. Now that there is
less noise, just output stderr all the time.

---

This is using module event groups to disable `zeek_init()` handlers in the management and control framework modules. They expect a certain environment / invocation and otherwise cause errors.

### DIFF
--- a/scripts/zeekygen/__load__.zeek
+++ b/scripts/zeekygen/__load__.zeek
@@ -21,7 +21,14 @@
 
 @load ./example.zeek
 
-event zeek_init()
+event zeek_init() &priority=1000
 	{
+	# Disable events in modules that use zeek_init() to do stuff and may
+	# fail when run under zeekygen. For the purpose of zeekygen, we could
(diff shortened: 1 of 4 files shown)


scripts/zeekygen/__load__.zeek
```diff
@@ -21,7 +21,15 @@
 
 @load ./example.zeek
 
-event zeek_init()
+event zeek_init() &priority=1000
 	{
+	# Disable module events that use zeek_init() to do stuff and may
+	# fail when run under zeekygen.. For the purpose of zeekygen, we
```
bbannier, line 27:
```suggestion
	# fail when run under zeekygen. For the purpose of zeekygen, we
```

bbannier (approved):
LGTM modulo the test failures.

Issue #3364, opened 2023-10-11
Stable Files::Info$analyzers order even when analyzers are added


awelzel:
When a new analyzer is added, there's a pretty good chance that the order of the `analyzers` field in the `files.log` changes in baselines due to the internal int values changing. With the `&ordered` attribute on `set` we should be able to remove this artifact.

Most recent commits to the baseline of the mentioned test are the following (and introducing QUIC with #3320 will add to this list). They mostly are unrelated to the x509/files.log.

```
$ git log --oneline testing/btest/Baseline/scripts.base.files.x509.files/files.log
f172febbc Move spicy-ldap into Zeek protocol analyzer tree
ee12a7a6e PPP: Add PPP analyzer to handle LINKTYPE_PPP (0x9)
7e88a2b3f Add basic LLC, SNAP, and Novell 802.3 packet analyzers
1843e2daa Update btest baselines to reflect the use of local address ranges.
d17329c95 Update seemingly-unrelated btests
```

I had looked at this a while ago and pushed an idea to use `&ordered` in branch `topic/awelzel/sort-mime-types-to-analyzers`. Might be a starting point.

PR #3365, opened 2023-10-11
files/main: Make files.log analyzers order predictable

awelzel:
It's unneeded baseline changes prevention week, so I'm trying this one, too.

More seriously, IMO having deterministic/predictable order of entries in the
analyzers field in files.log seems very reasonable and almost easier to explain
than why in a TSV or JSON the order may change between `zeek -r` runs.

---

Introducing new packet analyzers may change the underlying enum value of
File analyzers. This has been observed a few times after adding packet
or protocol analyzers and either is just churn, or causes some amount of
confusion why re-ordering happens.

    https://github.com/zeek/zeek/pull/2340
    https://github.com/zeek/zeek/pull/2770

Make the order in which File analyzers are added to the fa_file object
predictable by switching to a vector instead of a set and pre-sorting
using strcmp(). Additionally, use &ordered on the analyzers field of
Files::Info.

The sort approach is used as we can not easily change the existing API
of register_for_mime_type() that is taking a set. While the vector/sort
may seem inefficient, this should be a handful of analyzers at most
and the cost usually is incurred at startup.

Replicating one comment, the order in which file analyzers are processed
also becomes predictable:

> On the one hand it's the analyzers field order in the log that becomes stable, but on the other it's also having a defined/predictable order in which file analyzers are processed. That should make it more attainable to debug or reproduce issues that rely on the order of file analyzers being processed.

Closes https://github.com/zeek/zeek/issues/3364


src/file_analysis/AnalyzerSet.cc
```diff
@@ -20,7 +20,7 @@ static void analyzer_del_func(void* v)
 	delete a;
 	}
 
-AnalyzerSet::AnalyzerSet(File* arg_file) : file(arg_file)
+AnalyzerSet::AnalyzerSet(File* arg_file) : file(arg_file), analyzer_map(zeek::DictOrder::ORDERED)
```
awelzel, line 23: @timwoj - this would probably collide with your `std::map` work. As in the PR description, I think there's value in having the order of analyzer be predictable.

awelzel: And even more: A `std::map` wouldn't be ordered by insert order which is leveraged here.

I'm leaving this open as I'm thinking it's net positive. On the one hand it's the analyzers field order in the log that becomes stable, but on the other it's also having a defined/predictable order in which file analyzers are processed. That should make it more attainable to debug or reproduce issues that rely on the order of file analyzers being processed.

timwoj: The thing about "ordered" mode in our existing dictionary is that it's implemented by just keeping a vector with the insertion order. When you iterate over the dictionary, it instead iterates over that vector and does lookups against the dictionary. We could do the same for any implementation using one of the std maps.

rsmmr: Not sure I like this change. Couple things: (1) it could be considered a feature, not a bug, to have an unstable order: allowing the order to change may reveal bugs that happen to be invisible in particular order; and (2) I'm sure we have plenty other places where we could make the same argument, and changing it just here but not elsewhere feels inconsistent and potentially confusing to me.

scripts/base/frameworks/files/main.zeek
```diff
@@ -61,7 +61,7 @@ export {
 		depth: count &default=0 &log;
 
 		## A set of analysis types done during the file analysis.
-		analyzers: set[string] &default=string_set() &log;
+		analyzers: set[string] &ordered &log;
```
rsmmr, line 64: (Wouldn't it  be nice if `&ordered` just sorted them alphabetically right away ...)

scripts/base/frameworks/files/main.zeek
```diff
@@ -335,7 +335,7 @@ global registered_protocols: table[Analyzer::Tag] of ProtoRegistration = table()
 
 # Store the MIME type to analyzer mappings.
 global mime_types: table[Files::Tag] of set[string];
-global mime_type_to_analyzers: table[string] of set[Files::Tag];
+global mime_type_to_analyzers: table[string] of vector of Files::Tag;
```
rsmmr, line 338: This may be a matter of taste, but I'm wondering if we should leave this a `set`, and instead just convert into a sorted vector on output (i.e., when adding it to the logged `analyzers` field).

rsmmr (commented):
I generally agree with the benefits of making this stable, though I'm wondering about doing it just here vs. having the same issue elsewhere. It doesn't bother me much to do it for a specific case in script-land, but more so in C++ land.

Plus, I think for processing order there's a benefit of randomization to reveal bugs that could remain otherwise hidden.

Another crazy idea: could the logging framework just internally sort sets before it outputs them? It doesn't even need to be alphabetically or anything, just something stable.

scripts/base/frameworks/files/main.zeek
awelzel: We have a ticket with `&sorted`.

awelzel:
> Another crazy idea: could the logging framework just internally sort sets before it outputs them? It doesn't even need to be alphabetically or anything, just something stable.

Yeah, that could fit with the `&sorted` [idea here](https://github.com/zeek/zeek/issues/2208) and more of a post-processing/output change. I start to like that, too.

As in DM: Thanks for looking. I'm closing the PR for now. It's a pet peeve of mine, but it must have been around for a long time and not bothering too much. I might take a fresh look next time it comes up.

PR #3366, opened 2023-10-11
Docker builder CI fixes

timwoj:
This does two things:

- Adds a timeout for the http/https connections from `apt` commands to the docker files. This should help force the connections to fail and retry if they're taking a long time to complete.
- Split the `apt` commands in the docker files in to separate `RUN` commands so as to get better diagnostics about which step is failing to complete.

This is obviously hard to test since they're timeouts, and they're very spurious on CI.

### DIFF
--- a/docker/final.Dockerfile
+++ b/docker/final.Dockerfile
@@ -3,7 +3,12 @@
 # Final layer containing all artifacts.
 FROM debian:bookworm-slim
 
+# Allow apt to retry 3 times before failing.
 RUN echo 'Acquire::Retries "3";' > /etc/apt/apt.conf.d/80-retries
+
+# Force apt to timeout retrieval of data after 3 minutes.
+RUN echo 'Acquire::http::timeout "180";' > /etc/apt/apt.conf.d/99-timeouts
+RUN echo 'Acquire::https::timeout "180";' >> /etc/apt/apt.conf.d/99-timeouts
 
 RUN apt-get -q update \
  && apt-get install -q -y --no-install-recommends \
--- a/docker/final.Dockerfile
+++ b/docker/final.Dockerfile
@@ -3,6 +3,10 @@
 # Final layer containing all artifacts.
 FROM debian:bookworm-slim
 
+# Make the shell split commands in the log so we can determine reasons for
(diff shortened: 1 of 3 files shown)


docker/final.Dockerfile
```diff
@@ -19,8 +24,8 @@ RUN apt-get -q update \
      python3-minimal \
      python3-git \
      python3-semantic-version \
-     python3-websocket \
- && apt-get clean \
+     python3-websocket
+RUN apt-get clean \
  && rm -rf /var/lib/apt/lists/*
```
bbannier, line 29: The purpose of this step is to remove cached files from `apt-get` from the layer. If you add a separate `RUN` instructions the files are not present in that layer anymore, but will be in the previous layer. This increases the size of the image (probably inconsequential in the builder image, but not good for this final image).

timwoj: Ah, true. I'll drop out that commit.

docker/final.Dockerfile
```diff
@@ -3,8 +3,13 @@
 # Final layer containing all artifacts.
 FROM debian:bookworm-slim
 
+# Allow apt to retry 3 times before failing.
 RUN echo 'Acquire::Retries "3";' > /etc/apt/apt.conf.d/80-retries
 
+# Force apt to timeout retrieval of data after 3 minutes.
+RUN echo 'Acquire::http::timeout "180";' > /etc/apt/apt.conf.d/99-timeouts
+RUN echo 'Acquire::https::timeout "180";' >> /etc/apt/apt.conf.d/99-timeouts
+
```
awelzel, line 12: If you add the following to the Dockerfile, you can make the shell used by `RUN` use `-x` debug output which should help identify the command that hangs.
```
SHELL ["/bin/sh", "-c", "-x"] 
```

Additionally may make sense to place the following into the task's environment so buildkit doesn't collapse output, too:
```
BUILDKIT_PROGRESS=plain
```

timwoj: Added those. The new logs in https://cirrus-ci.com/task/5244127424544768 look better.

awelzel (approved):

PR #3367, opened 2023-10-11
Add registration for GRE-over-UDP

### LINKED ISSUE #3344: GRE-over-UDP support
[RFC 8086](https://www.rfc-editor.org/rfc/rfc8086) defines GRE over UDP, for which we've received [a request](https://zeekorg.slack.com/archives/CSZBXF6TH/p1695944452273349). This is basically just
```
const gre_ports = { 4754/udp } &redef;
PacketAnalyzer::register_for_ports(PacketAnalyzer::ANALYZER_UDP, PacketAnalyzer::ANALYZER_GRE, gre_ports);
```
plus a test case, unless I'm overlooking something @timwoj?

I'm attaching a quick synthetic [pcap](https://github.com/zeek/zeek/files/12755047/gre-over-udp.zip) with a single Eth/IP/UDP/GRE/IP/UDP packet, for testing.

If anyone reading this has a real example they could share, that'd be great.

The RFC also mentions encapsulation in DTLS (port 4755) — not sure we can do much with that but perhaps we can note down the protocol association somehow. @0xxon might know more.

@JustinAzoff, fyi.

timwoj:
Fixes #3344


testing/btest/Baseline/core.tunnels.gre-over-udp/conn.log
```diff
@@ -0,0 +1,14 @@
...
+#empty_field	(empty)
+#unset_field	-
+#path	conn
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents
+#types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	set[string]
+XXXXXXXXXX.XXXXXX	C4J4Th3PJpwUYZZ6gc	172.17.0.5	45690	1.1.1.1	53	udp	dns	0.000158	52	0	S0	T	F	0	D	2	108	0	0	ClEkJM2Vm5giqnMf4h
+XXXXXXXXXX.XXXXXX	CUM0KZ3MLUfNB0cl11	172.17.0.5	47478	192.0.78.150	80	tcp	http	0.090287	72	0	SH	T	F	0	SADF	6	332	0	0	ClEkJM2Vm5giqnMf4h
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.0.107	48282	192.168.5.1	4754	udp	-	0.000158	116	0	S0	T	T	0	D	2	172	0	0	-
+XXXXXXXXXX.XXXXXX	CtPZjS20MLrsMUOJi2	192.168.0.107	49714	192.168.5.1	4754	udp	-	0.090287	356	0	S0	T	T	0	D	6	524	0	0	-
```
awelzel, line 13: There's something off with this :-)

The tunnel_parent `ClEkJM2Vm5giqnMf4h` is used for *both* inner connections. If you look at the pcap, however, the inner DNS connection is tunneled within the `orig_p=48282` connection, while HTTP is tunneled within the `orig_p=49714` connection (src_port chosen as hash of inner flow).

Screenshot-20231012-091409

Screenshot-20231012-090918

So, there are two tunnel connections, but we only show a single one in `tunnel.log`. What is more, that `tunnel.log` uid isn't actually in the `conn.log`.

This is probably a more fundamental limitation around the current IP/GRE tunneling logic that needs to be made aware of the underlying connection.

I'd be okay to merge as is, but we should create a follow-up to straighten this out. Expectation would be two tunnel.log entries and two distinct tunnel_parents in the conn.log.

scripts/base/packet-protocols/udp/main.zeek
```diff
@@ -1,5 +1,8 @@
 module PacketAnalyzer::UDP;
 
-#event zeek_init() &priority=20
-#	{
-#	}
+const gre_ports = { 4754/udp } &redef;
```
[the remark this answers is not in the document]
timwoj: I considered it being in gre instead of UDP, and opted to follow what we did with the ip analyzer. I agree though that it should follow what the other tunnels are doing. We really ought to standardize one of of the two (either parents register children or children register against parents).

testing/btest/Baseline/core.tunnels.gre-over-udp/conn.log
timwoj: I don't think this is related to the change in this PR. I agree we should move this to a new issue and fix it separately.

JustinAzoff: For a GRE over UDP tunnel where the source port is a hash of the inner flow, it actually would be nice if there was only a single tunnel log entry, and not one for every single inner connection.  After all, there is actually only one tunnel.  I've seen this issue a bit with vxlan, where every conn log entry also has a corresponding vxlan and tunnel log, it's not ideal.

awelzel: Yes, but with the current model it would be more correct to have the individual outer connections. I do like the thought of removing the src port from the connection's tuple for these tunnels, but it would be quite different behavior.

testing/btest/Baseline/core.tunnels.gre-over-udp/conn.log
```diff
@@ -0,0 +1,14 @@
...
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	conn
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents
+#types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	set[string]
+XXXXXXXXXX.XXXXXX	C4J4Th3PJpwUYZZ6gc	172.17.0.5	45690	1.1.1.1	53	udp	dns	0.000158	52	0	S0	T	F	0	D	2	108	0	0	ClEkJM2Vm5giqnMf4h
+XXXXXXXXXX.XXXXXX	CUM0KZ3MLUfNB0cl11	172.17.0.5	47478	192.0.78.150	80	tcp	http	0.090287	72	0	SH	T	F	0	SADF	6	332	0	0	ClEkJM2Vm5giqnMf4h
```
awelzel, line 11: Could try running the test with `-C`  and see if  `SADF` looks better. I didn't pay too much attention recording the pcap and it was mostly virtual interfaces, so not surprised checksums dont work.

awelzel (approved):
I've created the followup #3378 and one more nit for the test.  :ship:

testing/btest/Baseline/core.tunnels.gre-over-udp/conn.log
timwoj: Zeek doesn't seem to care about the checksums actually. A normal run of zeek against the pcap doesn't report anything, and the btest doesn't report any difference when running with `-C`.

awelzel: Hmm, the pcap really contains just one side of the connection, so that must have been a fail on my side recording the pcap. Let me check if I can do a better one.

PR #3368, opened 2023-10-12
quic: Avoid bytes for VariableLengthInteger

awelzel:
Allocation of bytes objects due to parsing and usage of pack and the invocation of to_uint() showed significantly in profiles (3.3% sample matches). Switch to a more procedural approach to avoid the allocation overhead.

From zeek/spicy-quic/pull/13

---

Mainly testing workflow maintaining zeek/spicy-quic and the version in Zeek.

@bbannier , mind waving this one through assuming CI is happy?


bbannier (approved):
>@bbannier , mind waving this one through assuming CI is happy?

No waiving of review here, but after diligent review: My suggestion from zeek/spicy-quic#13 still looks great, and even better with `first_byte` no longer named ;)

PR #3371, opened 2023-10-12
protocol/quic/decrypt_crypto: Reuse OpenSSL context objects

awelzel:
It is not necessary to allocate and free the context objects used for HKDF and AES all the time, they can be re-used. The main assumption here is no cross-thread usage, but this should be guaranteed even with the fibers: QUIC_decrypt_crypto_payload() always runs to completion.

A pcap with ~12k QUIC connections had ~15% samples in QUIC_decrypt_crypto_payload. After this change it is down to 5% of samples. The improvement in runtime is ~16%, 12.2 seconds to 10.2 seconds.

From zeek/spicy-quic#14

### DIFF
--- a/src/analyzer/protocol/quic/decrypt_crypto.cc
+++ b/src/analyzer/protocol/quic/decrypt_crypto.cc
@@ -2,6 +2,18 @@
 // Copyright (c) 2023 by the Zeek Project. See COPYING for details.
 
 /*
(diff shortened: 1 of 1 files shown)


awelzel:
Benchmarker run indicates pcap-quic-12k is ~10% faster with this change:

Screenshot-20231013-081124

bbannier (approved):

PR #3373, opened 2023-10-13

bbannier:


awelzel (approved):
No visible benchmark changes for the pcap-quic benchmark tests, but I presume the chunk changes would be more visible for long TCP streams, rather than UDP packets? The nighly `spicy` test with HTTP might show something.

Screenshot-20231013-130142

PR #3380, opened 2023-10-16
Fix Dispatcher::Lookup shared_ptr copy + micro optimizations

awelzel:
...and various smaller fixes from discussion in #3379 in separate commits.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,10 @@ Zeek 6.2.0
 Breaking Changes
 ----------------
 
+- The methods ``Dispatcher::Lookup()`` and ``Analyzer::Lookup()`` in the packet_analysis
+  namespace were changed to return a reference to a std::shared_ptr instead of a copy
+  for performance reasons.
+
 New Functionality
 -----------------
 
--- a/src/packet_analysis/Analyzer.cc
+++ b/src/packet_analysis/Analyzer.cc
@@ -53,23 +53,47 @@ bool Analyzer::IsAnalyzer(const char* name) {
     return packet_mgr->GetComponentName(tag) == name;
 }
 
-AnalyzerPtr Analyzer::Lookup(uint32_t identifier) const { return dispatcher.Lookup(identifier); }
+const AnalyzerPtr& Analyzer::Lookup(uint32_t identifier) const { return dispatcher.Lookup(identifier); }
 
-bool Analyzer::ForwardPacket(size_t len, const uint8_t* data, Packet* packet, uint32_t identifier) const {
-    auto inner_analyzer = Lookup(identifier);
-    if ( ! inner_analyzer ) {
-        for ( const auto& child : analyzers_to_detect ) {
-            if ( child->DetectProtocol(len, data, packet) ) {
-                DBG_LOG(DBG_PACKET_ANALYSIS, "Protocol detection in %s succeeded, next layer analyzer is %s",
-                        GetAnalyzerName(), child->GetAnalyzerName());
-                inner_analyzer = child;
-                break;
-            }
+// Find the next inner analyzer using identifier or via DetectProtocol(),
+// otherwise return the default analyzer.
+const AnalyzerPtr& Analyzer::FindInnerAnalyzer(size_t len, const uint8_t* data, Packet* packet,
+                                               uint32_t identifier) const {
+    const auto& identifier_based_analyzer = Lookup(identifier);
+    if ( identifier_based_analyzer )
+        return identifier_based_analyzer;
+
+    const auto& detect_based_analyzer = DetectInnerAnalyzer(len, data, packet);
+    if ( detect_based_analyzer )
+        return detect_based_analyzer;
+
+    return default_analyzer;
+}
+
+// Find the next inner analyzer via DetectProtocol(), otherwise the default analyzer.
+const AnalyzerPtr& Analyzer::FindInnerAnalyzer(size_t len, const uint8_t* data, Packet* packet) const {
+    const auto& detect_based_analyzer = DetectInnerAnalyzer(len, data, packet);
+    if ( detect_based_analyzer )
+        return detect_based_analyzer;
+
+    return default_analyzer;
+}
+
+// Return an analyzer found via DetectProtocol() for the given data, else nil.
+const AnalyzerPtr& Analyzer::DetectInnerAnalyzer(size_t len, const uint8_t* data, Packet* packet) const {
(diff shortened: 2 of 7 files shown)


src/packet_analysis/Analyzer.cc
```diff
@@ -68,34 +68,60 @@ bool Analyzer::IsAnalyzer(const char* name)
 	return packet_mgr->GetComponentName(tag) == name;
 	}
 
+const AnalyzerPtr& Analyzer::LookupAnalyzer(uint32_t identifier) const
```
J-Gras, line 71: Naming nit as already discussed: I'd prefer to not reuse the name `LookupAnalyzer` here. Alternatives might be `DoLookup` or `Dispatch`. That being said, I wouldn't mind a breaking change using `Lookup`.

src/packet_analysis/Analyzer.cc
```diff
@@ -68,34 +68,60 @@ bool Analyzer::IsAnalyzer(const char* name)
...
 		}
+	return nil;
+	}
 
-	if ( ! inner_analyzer )
-		inner_analyzer = default_analyzer;
+bool Analyzer::ForwardPacket(size_t len, const uint8_t* data, Packet* packet,
+                             uint32_t identifier) const
+	{
+	const auto& analyzer = LookupAnalyzer(len, data, packet, identifier);
```
J-Gras, line 122: The other naming nit: Staying with `inner_analyzer` or something like `next_analyzer` would be more descriptive here.

src/packet_analysis/Analyzer.h
```diff
@@ -10,13 +10,17 @@
...
+class Analyzer;
+using AnalyzerPtr = std::shared_ptr<Analyzer>;
 
 /**
  * Main packet analyzer interface.
  */
 class Analyzer
 	{
 public:
+	static inline AnalyzerPtr nil = nullptr;
```
J-Gras, line 22: Reminds me of Pascal 😄

src/packet_analysis/Analyzer.cc
awelzel: Yeah, seems more reasonable to just break the API.

@timwoj , I'm going to push that and move Dispatcher into zeek::packet_analysis::detail. It doesn't look like a user could actually get direct access to a `Dispatcher` instance or put their own implementations in place if they wanted to, so this is a total implementation detail anyhow.

awelzel: > and move Dispatcher into zeek::packet_analysis::detail. 

Actually, nope. That causes a ton of fallout because we have much `detail::ABC` references within the `packet_analysis` namespace and they all break when introducing a `packet_analysis::detail` namespace :sigh:

awelzel:
@timwoj , I dropped the two top commits GetTag / TunnelEncapsulation pre-allocation and assigned you for thoughts around naming or API break. Thanks!

src/packet_analysis/protocol/iptunnel/IPTunnel.cc
```diff
@@ -183,12 +183,8 @@ std::unique_ptr<Packet> build_inner_packet(Packet* outer_pkt, int* encap_index,
 	uint32_t consumed_len = outer_pkt->cap_len - inner_cap_len;
 	uint32_t inner_wire_len = outer_pkt->len - consumed_len;
 
-	pkt_timeval ts;
-	ts.tv_sec = static_cast<time_t>(run_state::current_timestamp);
-	ts.tv_usec = static_cast<suseconds_t>(
-		(run_state::current_timestamp - static_cast<double>(ts.tv_sec)) * 1000000);
-
-	auto inner_pkt = std::make_unique<Packet>(link_type, &ts, inner_cap_len, inner_wire_len, data);
+	auto inner_pkt = std::make_unique<Packet>(link_type, &outer_pkt->ts, inner_cap_len,
```
timwoj, line 186: Does this work correctly for multiply-nested packets, if one of the nests has a weird timestamp? The previous method would always use the timestamp for the outermost frame, but now it'll potentially use one from an inner frame that might not match.

timwoj (commented):
This looks good to me minus the one question. I like the change to const references, since the actual shared_ptr shouldn't change at all in the caller. I'm also fine with the API break since these are all internal methods and shouldn't be used in other places.

I saw that you tried to move `Dispatcher` to the detail namespace but ran into problems. Would that be good to do as a follow-up outside of this work?

src/packet_analysis/Analyzer.cc
awelzel: > Naming nit as already discussed: I'd prefer to not reuse the name LookupAnalyzer here. Alternatives might be DoLookup or Dispatch. That being said, I wouldn't mind a breaking change using Lookup.

Ah, sorry Jan. I changed the `Dispatcher::LookupAnalyzer` one back to `Lookup` as that one is definitely internal. I now realize this was for `Analyzer`. Let me mull it over, though `LookupAnalyzer` as the public API seems preferable to the other names. But yes, naming... If the name re-use concern is with the private methods, those being `LookupInner` and `DetectInner` may be an improvement. I mostly came from an public API angle, the rest can always easily be changed.

src/packet_analysis/protocol/iptunnel/IPTunnel.cc
awelzel: I didn't think too much as this aligns with `ProcessEncapsulatedPacket` that do the same if `pkt` is provided. Also, think it's more intuitive to determine the timestamp based on input parameters rather than global state (would make unit testing easier as well).

I'm unclear where we would get a "weird" timestamp from. At least currently, it seems the `ts` field would anyhow be the same as the global network time  for `build_inner_packet()` users.

This borders the scenario where tunnel metadata contains timestamps of inner packets. Though that would certainly require more surgery and you wouldn't want to use the global network time then either.

Going to put something about the timestamp in the docstring, but otherwise resolving.

NEWS
```diff
@@ -9,6 +9,10 @@ Zeek 6.2.0
 Breaking Changes
 ----------------
 
+- The methods ``Dispatcher::Lookup()`` and ``Analyzer::Lookup()`` in the packet_analysis
+  namespace were changed to return reference to a std::shared_ptr instead of a copy
```
timwoj, line 13:
```suggestion
  namespace were changed to return a reference to a std::shared_ptr instead of a copy
```

timwoj (approved):
One minor nit. Feel free to merge after fixing.

PR #3381, opened 2023-10-16
performance improvements for the LDAP ASN.1 analyzer

pauldokas:
performance improvements for the LDAP ASN.1 analyzer


awelzel:
Feedback from the team that developed this patch:

> string concatenation in spicy is expensive. it creates a copy, whereas bytes can be appended to.

That seems to be the motivation to use an intermediate bytes object that's decoded during `%done`.

awelzel (approved):

awelzel:
@timwoj - this is for inclusion in 6.1.

timwoj:
> @timwoj - this is for inclusion in 6.1.

Cherry-picked. Thanks for flagging it.

PR #3391, opened 2023-10-19
Speed up code for exporting Spicy types to Zeek

### LINKED ISSUE #3370: Generated code for exporting Spicy types with many deps extremely slow
A user reported extremely slow compiles and possibly stack overflows at runtime in `__hlt::spicy_init::zeek_preinit()` for the code generated in `spicy_init` (performing registration of exported Spicy types and handlers with Zeek). I suspect bad perf in this particular case might be due to us generating deeply nested calls around `create_*_type` and `create_record_field` in `register_type` (for the worst case we generated a `register_type` line around 300k characters long).

The deep nesting is an artifact of us generating this code recursively. I wonder if generating more flattened code would perform better. The simplest version of that would probably open scopes for each `record_type` and emit needed types into it; the final call the `record_type` could then be built with references to these variables instead of creating temporaries. A more advanced version could e.g., use a local, per-module registry which we'd first populate and then reference.

rsmmr:
- Spicy: Unroll Zeek type registrations.
- Spicy: Avoid creating Zeek types multiple times.

Closes #3370.

### DIFF
--- a/src/spicy/spicyz/glue-compiler.cc
+++ b/src/spicy/spicyz/glue-compiler.cc
@@ -8,6 +8,7 @@
 
 #include <hilti/ast/all.h>
 #include <hilti/ast/builder/all.h>
+#include <hilti/ast/builder/expression.h>
 #include <hilti/base/preprocessor.h>
 #include <hilti/base/util.h>
 #include <hilti/compiler/unit.h>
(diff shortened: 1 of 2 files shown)


timwoj:
@rsmmr Is this something you'd like to get into 6.1?

bbannier (approved):
What a nice, compact change. I see big compile time improvements for the code which lead to #3370 (cannot validate that the stack overflow at runtime is fixed, but the generated code is much more compact).

PR #3393, opened 2023-10-19
Fix mmdb pointer destruction

pbcullen:
The mmdb references needs to use reset() instead of release() so that the destructor gets called and the stale DB is actually cleaned up. Otherwise, the pointer is leaked and the stale DB remains in memory.

### DIFF
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -4011,7 +4011,7 @@ static void report_mmdb_msg(const char* format, ...)
 
 	va_list al;
 	va_start(al, format);
-	std::string msg = zeek::util::fmt(format, al);
+	std::string msg = zeek::util::vfmt(format, al);
 	va_end(al);
 
 	zeek::reporter->Info("%s", msg.data());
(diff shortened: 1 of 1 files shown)


ckreibich:
Thanks Peter!

We should add a test case since as far as I can tell the `mmdb_*` functionality is essentially untested. There's a [Perl package](https://metacpan.org/pod/MaxMind::DB::Writer) and [Go module](https://github.com/maxmind/mmdbwriter) for creating dummy Maxmind DBs. We could then make `lookup_location()`/`lookup_autonomous_system()` calls to baseline the rotation behavior. This should gate on maxmind support being built in. Looks like in CI we do so for our Ubuntu builds.

If you don't have cycles for this we can handle this in a subsequent PR, Peter.

pbcullen:
@ckreibich Unfortunately, I don't have the bandwidth to work on the test case for this. I'd also lobby that this is a serious enough bug to make it into a 6.0.x release, if possible.

awelzel:
Superseded by #3401 - @pbcullen , I cherry-picked your commit there.

awelzel:
#3401 is merged.

PR #3394, opened 2023-10-20
event: Reinitialize EventMgr's flare after fork() from stem

### LINKED ISSUE #3190: Potential performance issue with Supervisor Framework
Using Zeek's Supervisor Framework to spawn clusters, I noticed that they don't scale as expected. Increasing the number of workers increases the total CPU time required to process the same amount of traffic rather significantly. For my tests I am working with Zeek 6.

To reproduce the issue, I am using the following script to process a PCAP replayed via a veth interface (`zeekplay`):
```Zeek
redef Cluster::retry_interval = 1sec;
redef Log::default_rotation_interval = 0secs;

event zeek_init()
        {
        if ( ! Supervisor::is_supervisor() )
                return;

        Broker::listen("127.0.0.1", 9999/tcp);

        local cluster: table[string] of Supervisor::ClusterEndpoint;
        cluster["manager"] = [$role=Supervisor::MANAGER, $host=127.0.0.1, $p=10000/tcp];
        cluster["logger"] = [$role=Supervisor::LOGGER, $host=127.0.0.1, $p=10001/tcp];
        cluster["proxy"] = [$role=Supervisor::PROXY, $host=127.0.0.1, $p=10002/tcp];

        local i = 0;
        while ( ++i <= 16 )
                {
                cluster[fmt("worker-%s", i)] = [
                        $role=Supervisor::WORKER, $host=127.0.0.1,
                        $p=count_to_port(10003 + i, tcp),
                        $interface="af_packet::zeekplay"];
                }

        for ( n, ep in cluster )
                {
                local sn = Supervisor::NodeConfig($name=n);
                sn$cluster = cluster;
                sn$directory = n;

                if ( ep?$interface )
                        sn$interface = ep$interface;

                local res = Supervisor::create(sn);

                if ( res != "" )
                        print fmt("supervisor failed to create node '%s': %s", n, res);
                }
        }
```
The above script creates 16 workers using AF_Packet to listen on the veth. Looking at the flame graph of a worker, pipe reads caught my attention:
perf-16w-live-supervised-topspeed
The pipe reads belong to the flare that is used in the event manager (the underlying file descriptor is used to make the event manager an IO source that can be polled):
https://github.com/zeek/zeek/blob/b6bff8aa37f3b51839b5d8dae606a09759e31e8c/src/Event.cc#L193-L195
Spending a lot of time in `native_queued_spin_lock_slowpath` doesn't seem desirable. Talking to @awelzel, we realized that it's not only the read but also the writes to the pipe that take a lot of time. Overall, about 22% is spent in spin locks here (search for `spin_lock_slowpath`). Unfortunately, it looks like this overhead scales with the number of workers.

Running a cluster with the same setup using `zeekctl`, the pipes don't run into these spin locks:
perf-16w-live-topspeed

My next idea was that forking as done in supervisor mode somehow leads to all workers using the same pipe. While I would have expected this to cause even more trouble, I played around with `lsof` to take a look at the pipes used by each process. From what I understand, there is only one pipe, that is shared between all nodes. This pipe is probably the one used for the communication with the supervisor's stem process. So the nodes don't seem to share the _same_ pipes. However, maybe the problem is still somehow related to the way the processes are forked.

awelzel:
Because EventMgr is defined globally as an object (rather than a global
pointer to an EventMgr object), its pipe is created even before main()
is entered. This further means that in the fork-based supervisor setup,
all Zeek processes created from the top-level supervisor process share
the same pipe object for the EventMgr. In turn, whenever any of the
processes enqueued an event, the flare was fired and ready for reading
on all other processes in the cluster, causing much contention and
unneeded overhead.

Closes #3190

---

Added a `InitPostFork()` that only runs on supervised nodes post fork(). If we have more cases where this matters we can make it more generic or look at `pthread_atexit`. As long as it's just the EventMgr, didn't seem worth it.

After this fix, worker CPU usage is comparable with the Python supervisor mentioned in #3190 

Credits to @J-Gras for the report.

### DIFF
--- a/src/Event.cc
+++ b/src/Event.cc
@@ -207,4 +207,12 @@ void EventMgr::InitPostScript()
 		reporter->FatalError("Failed to register event manager FD with iosource_mgr");
 	}
 
+void EventMgr::InitPostFork()
+	{
+	// Re-initialize the flare, closing and re-opening the underlying
+	// pipe FDs. This is needed so that each Zeek process in a supervisor
+	// setup has its own pipe instead of them all sharing a single pipe.
+	queue_flare = zeek::detail::Flare{};
+	}
+
 	} // namespace zeek
--- a/src/Event.h
+++ b/src/Event.h
@@ -118,6 +118,9 @@ public:
 	const char* Tag() override { return "EventManager"; }
 	void InitPostScript();
 
+	// Initialization to be done after a fork() happened.
+	void InitPostFork();
+
 	uint64_t num_events_queued = 0;
 	uint64_t num_events_dispatched = 0;
 
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -554,7 +554,13 @@ SetupResult setup(int argc, char** argv, Options* zopts)
 	auto stem = Supervisor::CreateStem(options.supervisor_mode);
 
 	if ( Supervisor::ThisNode() )
+		{
+		// If we get here, we're a supervised node that just returned
+		// from CreateStem() after being forked from the stem.
 		Supervisor::ThisNode()->Init(&options);
+
+		event_mgr.InitPostFork();
+		}
 
 	script_coverage_mgr.ReadStats();
 


timwoj (approved):

timwoj:
I'm approving this one, but I think we could also have a conversation about why EventMgr is the only manager object we create as a global object vs a pointer like the others.

PR #3401, opened 2023-10-23
Fix mmdb pointer destruction + tests

awelzel:


src/zeek.bif
```diff
@@ -4071,26 +4071,29 @@ MMDB_lookup_result_s MMDB::Lookup(const struct sockaddr* const sa)
...
-	std::chrono::time_point<Clock> now = Clock::now();
 
 	if ( lookup_error )
 		return true;
 
-	// Only perform stat once per 5 minutes.
-	using Min = std::chrono::minutes;
-	if ( std::chrono::duration_cast<Min>(now - last_check).count() < 5 )
+
+	static double mmdb_stale_check_interval = zeek::id::find_val("mmdb_stale_check_interval")->AsInterval();
```
bbannier, line 4079: The `static` here looks like a microoptimization to me which introduces subtleties. Are you sure it is needed and we cannot just get a fresh value each time?

testing/btest/core/mmdb/temporary-error.zeek
```diff
@@ -0,0 +1,85 @@
+# @TEST-DOC: Test a few error cases (corruped or removed MMDB database).
```
bbannier, line 1:
```suggestion
# @TEST-DOC: Test a few error cases (corrupted or removed MMDB database).
```

bbannier (approved):

bbannier:
> Mostly wondering if the spicy/frameworks change looks okay to you and maybe you have other thoughts on the whole changeset?

I suspect we had a load of `base/misc/version` in there to make version checks work in bare mode. It looks that conditional code is not there anymore, but the `@load` lingered; removing it seems fine to me.

> Also, do know what to do or where to do the following? Is that a Zeek issue, or should I rebase?
> ...
> https://github.com/zeek/zeek/actions/runs/6624781532/job/17994449351

You changed `spicy/init-framework.zeek` which is part of the generated docs so they need to be regenerated.

```console
cd doc
make check-spicy-docs
./devel/spicy/autogen-spicy-docs spicy-tftp
# Stage and commit updated files.
```

src/zeek.bif
awelzel: This might be called often enough and we do that elsewhere. Will keep it in.

https://github.com/zeek/zeek/blob/254d7ebfa2f39c26bfd18c79f7b6f525eb713fc6/src/zeek.bif#L5094

awelzel:
@timwoj , I've already put this on the 6.0.x issues list. I think 6.1 would be good as well. @ckreibich, this may have gone in a bit faster than you had hoped, so feel free to veto or request fixups :-)

src/zeek.bif
JustinAzoff: Yes. In common configurations a mmdb lookup might be done 4 times for each connection.

Issue #3408, opened 2023-10-25
CMake still doesn't pass along third-party package information for plugin builds


J-Gras:
Trying to build https://github.com/zeek/zeek-more-hashes using latest master, it seems I am still running into #3247 on macOS:
```
$ ./configure && make
...
/usr/local/zeek/include/zeek/OpaqueVal.h:11:10: fatal error: 'openssl/md5.h' file not found
#include <openssl/md5.h>
```
As recommended by @awelzel, I am opening this as a new issue.

Neverlord:
I can reproduce it. I'll take a look.

awelzel:
With #3420  this should be fixed for 6.2 and the future, but it's somewhat unreasonable for back-porting to 6.0.

Looking at this again, however, we have the following snippet in Zeek's top-level CMakeLists.txt:

https://github.com/zeek/zeek/blob/6d6a95e761693bad7734b92d395b22dba0e8bacc/CMakeLists.txt#L294-L296

But it's placed before `find_package(OpenSSL)` ever ran.. So wonder if naively moving this down would be an okay 6.0 fix, but immediately remove it with 6.2, too.

```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,9 +291,6 @@ target_include_directories(
     zeek_dynamic_plugin_base
     INTERFACE $<INSTALL_INTERFACE:include> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
               $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
-if (OPENSSL_INCLUDE_DIR)
-    target_include_directories(zeek_dynamic_plugin_base INTERFACE "${OPENSSL_INCLUDE_DIR}")
-endif ()
 target_link_libraries(zeek_dynamic_plugin_base INTERFACE Threads::Threads)
 add_library(Zeek::DynamicPluginBase ALIAS zeek_dynamic_plugin_base)
 set_target_properties(zeek_dynamic_plugin_base PROPERTIES EXPORT_NAME DynamicPluginBase)
@@ -761,6 +758,10 @@ if (NOT BINARY_PACKAGING_MODE)
     set(BUILD_SHARED_LIBS yes)
 endif ()
 
+if (OPENSSL_INCLUDE_DIR)
+    target_include_directories(zeek_dynamic_plugin_base INTERFACE "${OPENSSL_INCLUDE_DIR}")
+endif ()
+

```

[...]

PR #3409, opened 2023-10-25
Force OpenSSL as a dependency on all plugins

Neverlord:
Relates #3408.

Here's the TLDR:

- We forward OpenSSL paths to the plugins (and finding OpenSSL works for me).
- Nothing forces plugins to pick up OpenSSL as an actual dependency (yet).

This patch changes the latter and forces OpenSSL as a dependency on all plugins even if they don't actually need it.

The alternative solution is to have plugins opt-in to OpenSSL if they really do need it. They can do this by using `DEPENDENCIES OpenSSL::SSL OpenSSL::Crypto`, e.g.,

```
 zeek_add_plugin(
      Zeek
        MoreHashes
      DEPENDENCIES OpenSSL::SSL OpenSSL::Crypto
      SOURCES
        ...
```

If we want to go this way, just close this PR.

I'd personally prefer the second option and not have OpenSSL forced on all plugins. However, I also don't know how many plugins we might break, so my opinion might be naïve.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,7 +294,8 @@ target_include_directories(
 if (OPENSSL_INCLUDE_DIR)
     target_include_directories(zeek_dynamic_plugin_base INTERFACE "${OPENSSL_INCLUDE_DIR}")
 endif ()
-target_link_libraries(zeek_dynamic_plugin_base INTERFACE Threads::Threads)
+target_link_libraries(zeek_dynamic_plugin_base INTERFACE Threads::Threads OpenSSL::SSL
+                                                         OpenSSL::Crypto)
 add_library(Zeek::DynamicPluginBase ALIAS zeek_dynamic_plugin_base)
 set_target_properties(zeek_dynamic_plugin_base PROPERTIES EXPORT_NAME DynamicPluginBase)
 install(TARGETS zeek_dynamic_plugin_base EXPORT ZeekTargets)


awelzel:
> This patch changes the latter and forces OpenSSL as a dependency on all plugins even if they don't actually need it.
>
> The alternative solution is to have plugins opt-in to OpenSSL if they really do need it. They can do this by using DEPENDENCIES OpenSSL::SSL OpenSSL::Crypto, e.g.,

The thing is: `zeek-more-hashes` has no dependency on OpenSSL. It's only adding MurmurHash. What it does do is include Zeek's `OpaqueVal.h` and *that* has the include of <openssl/md5.h>. I suspect the plugin that @J-Gras  started with doesn't require anything from OpenSSL either, but it does want to create an OpaqueVal to add new types to Zeek scripts.

There are some other "public" headers that include OpenSSL and I'd argue that all plugins should get the dependency implicitly until we maybe have hidden them away. The alternative of asking plugin authors to depend on OpenSSL even though they don't need it seems worse.

timwoj:
> The thing is: zeek-more-hashes has no dependency on OpenSSL. It's only adding MurmurHash. What it does do is include Zeek's OpaqueVal.h and that has the include of <openssl/md5.h>.

This is my issue with this kind of change as well. We're fixing it by forcing everyone to need OpenSSL, even if they're not using it. 

I've had a couple of notes written down to do one of two things:

1. Move all of the child `Opaque` classes out into separate headers. If you need MD5, include the header for that and know you'll need OpenSSL. That means people that just want `OpaqueVal` can just include `OpaqueVal.h` and not get any of the extras. This would be a breaking change though because you have to add includes for those headers to the places that do actually need them.
2. Remove the dependencies from OpaqueVal.h and replace them with `void*` workarounds and typedefs for the various types that it needs. `OpaqueVal.cc` could then include the header that it actually needs to build.

Neverlord:
> 2. Remove the dependencies from OpaqueVal.h and replace them with `void*` workarounds and typedefs for the various types that it needs. `OpaqueVal.cc` could then include the header that it actually needs to build.

I went with this approach in https://github.com/zeek/zeek/pull/3420. I didn't go with `void*`, though, because you'll get in trouble very easily since every pointer type is implicitly convertible to it.

awelzel:
Closing in favor of #3420.

PR #3410, opened 2023-10-26
generic-analyzer-fuzzer: Support NextPacket() fuzzing, too.

### LINKED ISSUE #3398: Generic analyzer fuzzer only works for TCP based analyzers
The current `generic-analyzer-fuzzer.cc` uses `NextStream()` to deliver data to the analyzer being fuzzed.

https://github.com/zeek/zeek/blob/1aa9de28110a9f930e0296f4cb815926996102bc/src/fuzzers/generic-analyzer-fuzzer.cc#L74-L80

The current setup uses this generic fuzzer for UDP-based analyzers as well (Radius, DHCP, NTP, SNMP, Syslog, Sip and others). These only implement `DeliverPacket()`. The current fuzzer setup is not effective (not doing anything really) for UDP based analyzers.

Coverage data from OSS-Fuzz indicates they are "hit", but that's probably only indirectly through the lower-level packet-fuzzer.

awelzel:
A number of analyzers that we've been fuzzing with the generic-analyzer-fuzzer setup do not implement DeliverStream() and instead only work with DeliverPacket() (ntp, syslog, sip, radius, ...). Calling DeliverStream() on those is pretty much a noop and fuzzing not effective.

This change adds support to fuzz DeliverPacket(). Whether to use packet or stream fuzzing is configured through a define via CMake.

This is still somewhat limited in that analyzers that support both, DeliverPacket() and DeliverStream(), only one path is fuzzed.

Closed #3398

---

Also add a new fuzzer for DTLS. Mainly used that one for checking that packet based fuzzing starts :-)


src/fuzzers/generic-analyzer-fuzzer.cc
```diff
@@ -8,85 +18,191 @@
...
+		zeek::detail::ConnKey key(conn_id);
+		zeek::Connection* conn = new zeek::Connection(key, network_time_start, &conn_id, 1, &p);
+		conn->SetTransport(proto);
+		zeek::session_mgr->Insert(conn);
+		return conn;
+		}
+
+	std::tuple<zeek::analyzer::Analyzer*, zeek::packet_analysis::IP::SessionAdapter*,
+	           zeek::Connection*>
+	Setup()
```
timwoj, line 65: That definitely formatted weirdly.

src/fuzzers/CMakeLists.txt
```diff
@@ -94,27 +102,29 @@ add_generic_analyzer_fuzz_target(pop3)
 add_generic_analyzer_fuzz_target(smtp)
 
 add_generic_analyzer_fuzz_target(dce_rpc)
-add_generic_analyzer_fuzz_target(dhcp)
+add_generic_analyzer_fuzz_target(dhcp packet)
```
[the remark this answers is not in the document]
awelzel: As in DM moving to tcp/udp. I'm naming the environment variable `_PROTO` instead of `_TYPE` so we can add another dimension later if wanted.

src/fuzzers/generic-analyzer-fuzzer.cc
```diff
@@ -8,85 +17,191 @@
...
+#include "zeek/packet_analysis/protocol/udp/UDPSessionAdapter.h"
 #include "zeek/session/Manager.h"
 
 // Simple macros for converting a compiler define into a string.
 #define VAL(str) #str
 #define TOSTRING(str) VAL(str)
 
-static zeek::Connection* add_connection()
+static const char* FUZZ_ANALYZER_NAME = TOSTRING(ZEEK_FUZZ_ANALYZER);
+static const char* FUZZ_ANALYZER_PROTO = TOSTRING(ZEEK_FUZZ_ANALYZER_PROTO);
```
timwoj, line 30: How about `FUZZ_ANALYZER_TRANSPORT`? That's more concise.

timwoj (changes requested):

timwoj (approved):

timwoj:
Since this was merged, coverage numbers reported by oss fuzz went from 49.07% to 49.57%. I expect they'll probably go up a bit more before stabilizing.

PR #3413, opened 2023-10-26
ci/macos/prepare: Install libmaxminddb

awelzel:


bbannier:
This is part of #3414.

PR #3414, opened 2023-10-26
Fix tests so they work both with GNU and BSD tools

bbannier:


testing/btest/core/mmdb/temporary-error.zeek
```diff
@@ -15,15 +15,32 @@ redef mmdb_dir = "./mmdb";
...
 
-	local asn_fn = safe_shell_quote(mmdb_dir + "/GeoLite2-ASN.mmdb");
-	local city_fn = safe_shell_quote(mmdb_dir + "/GeoLite2-City.mmdb");
+event zeek_init()
+	{
+	# Set the initial modification time for the MMDBs.
+	for ( db in vector(asn_fn, city_fn, asn_fn_backup, city_fn_backup) )
+		{
+		if ( ! piped_exec(fmt("test -f %s && touch -d %s %s", db, timestamp(pkt), db),
+		    "") )
```
awelzel, line 36: Mind making that line longer instead of the line-break?

awelzel (approved):
Thanks!

testing/btest/core/mmdb/temporary-error.zeek
bbannier: This was zeek-format, but sure.

ckreibich: FWIW, I don't think zeek-format is production ready and I don't think we should be using it to format our code at this point in time.

bbannier: @ckreibich since I do not regularly work on Zeek's scripts I personally certainly heavily depend on zeek-format to get the formatting at least roughly right. In this instance manual reformatting probably looks better, but TBH, unless zeek-format completely gets something wrong I am usually fine with its behavior (e.g., zeek/zeekscript#62 is something I run in more often).

PR #3416, opened 2023-10-26
Support for indexing "table[pattern] of X" with strings

vpax:
This PR provides an implementation for indexing a `table[pattern] of X` with a `string` to yield a `vector of X` of all of the associated yields, per https://github.com/zeek/zeek/discussions/3326. The implementation involves adding a (simple) internal regular expression operator, `||` to support efficient matching.

I'll put in an associated documentation PR next.


src/re-parse.y
```diff
@@ -32,14 +33,17 @@ void yyerror(const char msg[]);
...
 
 %destructor { delete $$; } <mach_val>
 
 %%
 flexrule	:	re
 			{ $1->AddAccept(1); zeek::detail::nfa = $1; }
 
+		|  disjunction
+			{ zeek::detail::nfa = $1; }
+
```
awelzel, line 46: Hey @vpax - I was staring at this for probably way too long, but am wondering: Can we instead re-use the `Specific_RE_Matcher::CompileSet()` that already exists and remove the `||` parsing "trick" in REs?

Mostly because I was right away trying whether `/||a||b/` is now a supported pattern. It's not in script land, but it is accepted for signature files and also accessible via the `RE_Matcher` API.

The regex `||a||b` is valid in other languages (tried Python, JavaScript, Ruby) and while it might not be super sensible, it does also match the empty string.

Was there a compelling reason not to use CompileSet()? I've worked out the patch below and it does pass your new test-case (not including re-parse.* revert for brevity). Unfortunately, I'm not well versed here, so I might be missing the obvious.

```--- a/src/RE.cc
+++ b/src/RE.cc
@@ -194,6 +194,10 @@ bool Specific_RE_Matcher::CompileSet(const string_list& set, const int_list& idx
        dfa = new DFA_Machine(nfa, EC());
        ecs = EC()->EquivClasses();
 
+       // dfa took ownership
+       Unref(nfa);
+       nfa = nullptr;
+
        return true;
        }
 
@@ -523,13 +527,17 @@ RE_DisjunctiveMatcher::RE_DisjunctiveMatcher(const std::vector<const RE_Matcher*
        {
        matcher = std::make_unique<detail::Specific_RE_Matcher>(detail::MATCH_EXACTLY);
 
-       std::string disjunction;
-       for ( auto re : REs )
-               disjunction += std::string("||") + re->PatternText();
+       zeek::detail::string_list sl;
+       zeek::detail::int_list il;
+
+       for ( const auto* re : REs )
+               {
+               sl.push_back(const_cast<char*>(re->PatternText()));
+               il.push_back(sl.size());
+               }
 
-       matcher->SetPat(disjunction.c_str());
-       auto status = matcher->Compile();
-       ASSERT(status);
+       if ( ! matcher->CompileSet(sl, il) )
+               reporter->FatalError("failed compile set for disjunctive matcher");
        }


```

src/Val.cc
```diff
@@ -1761,6 +1761,100 @@ static void find_nested_record_types(const TypePtr& t, std::set<RecordType*>* fo
...
+		}
+	~TablePatternMatcher() { Clear(); }
+
+	void Insert(ValPtr pat, ValPtr yield) { Clear(); }
+	void Remove(ValPtr pat) { Clear(); }
+
+	void Clear()
+		{
+		delete matcher;
+		matcher = nullptr;
```
awelzel, line 1783: Opportunity for another unique_ptr and `reset()`? Wouldn't need a destructor then either.

src/re-parse.y
vpax: The key requirement is being able to associate different accepting numbers with the different elements of the disjunction. The place I saw to do that was in the RE parser. If `CompileSet` has the same effect, then yeah fine to go that route instead!

awelzel: Yeah, it's using the values provided in the idx list as accepting numbers for the individual REs (`il` in the patch above).

https://github.com/zeek/zeek/blob/4d3e9de3d309c77f0ee7a55447ad66eb6e6887e4/src/RE.cc#L180-L183

src/RE.h
```diff
@@ -113,6 +114,17 @@ class Specific_RE_Matcher
...
+	// A disjunction is a collection of regular expressions (that under
+	// the hood are matches as a single RE, not serially) for which
+	// the match operation returns *all* of the matches.  Disjunctions
+	// are constructed using the internal "||" RE operator, and the
+	// matches are returned as indices into the position, left-to-right,
+	// of which REs matched. IMPORTANT: the first RE is numbered 1, not 0.
+	//
+	// Note that there's no guarantee regarding the ordering of the
+	// returned matches if there is more than one.
+	void MatchDisjunction(const String* s, std::vector<int>& matches);
```
awelzel, line 126: AFAICT, the new code for this is pretty much directly `MatchAll()`, but then also collecting the numbers of all accepted states. Would an overloaded `MatchAll` maybe do the job, or would you prefer to stick with the disjunction term?

awelzel (commented):
This is really cool :-)


The disjunction `||` in re-parse vs `CompileSet()` question was the biggie. The `MatchDisjunction` vs a more flexibel `MatchAll()` maybe another.

The rest is somewhere between nits and ideas, hopefully useful.

Were you thinking of handing this off, or would you like to do another round?

vpax:
Per offline discussion, @awelzel is going to pick this up and move it forward when he has some cycles.

awelzel:
> Per offline discussion, @awelzel is going to pick this up and move it forward when he has some cycles.

Closing this PR, there's #3426 as follow-up.

PR #3420, opened 2023-10-28
Avoid OpenSSL header dependencies

Neverlord:
Followup to the discussion in #3409. Instead of forcing OpenSSL as a dependency on modules, this PR attempts to remove OpenSSL includes from public Zeek headers.

There are some OpenSSL includes left in Zeek headers:

- file_analysis/analyzer/x509/OCSP.h
- file_analysis/analyzer/x509/X509Common.h

Is it safe to assume that plugins don't include these Zeek headers? Otherwise, we'll need to get rid of the OpenSSL includes in those files as well.

### DIFF
--- a/src/OpaqueVal.cc
+++ b/src/OpaqueVal.cc
@@ -12,6 +12,9 @@
 
 #include <broker/data.hh>
 #include <broker/error.hh>
+#include <openssl/evp.h>
+#include <openssl/md5.h>
+#include <openssl/sha.h>
 #include <memory>
 
 #include "zeek/CompHash.h"
@@ -20,8 +23,17 @@
 #include "zeek/Reporter.h"
 #include "zeek/Scope.h"
 #include "zeek/Var.h"
+#include "zeek/digest.h"
 #include "zeek/probabilistic/BloomFilter.h"
 #include "zeek/probabilistic/CardinalityCounter.h"
+
+#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
+inline void* EVP_MD_CTX_md_data(const EVP_MD_CTX* ctx) { return ctx->md_data; }
+#endif
+
+#if ( OPENSSL_VERSION_NUMBER < 0x30000000L ) || defined(LIBRESSL_VERSION_NUMBER)
+#include <openssl/md5.h>
+#endif
 
 namespace zeek {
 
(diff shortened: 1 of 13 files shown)


Neverlord:
Also relates relates #3408.

Neverlord:
Rebased to resolve the reformatting conflicts.

src/OpaqueVal.cc
```diff
@@ -193,16 +205,189 @@ StringValPtr HashVal::DoGet() {
...
+
+void do_clone(MD5Val::StatePtr out, MD5Val::StatePtr in) { hash_copy(to_digest_ptr(out), to_digest_ptr(in)); }
+
+void do_feed(MD5Val::StatePtr ptr, const void* data, size_t size) {
+    detail::hash_update(to_digest_ptr(ptr), data, size);
+}
+
+void do_get(MD5Val::StatePtr ptr, u_char* digest) { detail::hash_final_no_free(to_digest_ptr(ptr), digest); }
+
+std::pair<const char*, size_t> do_get_bytes(MD5Val::StatePtr ptr) {
```
awelzel, line 234: Given that the callers of `do_get_bytes()` all subsequently create a std::string, could this just return a std::string directly (though internal, also makes ownership questions obvious)

awelzel: And nit, there's `do_get`, do wonder if this should be `do_serialize()` and `do_unserialize()` such that it aligns with where they are used and not too close to the existing ones.

src/OpaqueVal.cc
```diff
@@ -193,16 +205,189 @@ StringValPtr HashVal::DoGet() {
...
-void HashVal::digest_one(EVP_MD_CTX* h, const Val* v) {
+void do_set_bytes(SHA256Val::StatePtr ptr, const char* bytes, size_t len) { memcpy(ptr, bytes, len); }
+
+void do_destroy(SHA256Val::StatePtr ptr) { delete to_native_ptr(ptr); }
+
+#endif
+
+} // namespace
+
+MD5Val::MD5Val() : HashVal(md5_type) { memset(&ctx, 0, sizeof(ctx)); }
```
awelzel, line 386: This appears reminiscent. ctx is null initialized.

awelzel (approved):
@Neverlord , mind doing the fixups / renames that popped out during a light skim and re-push to see what a "fullci" CI run says (I labeled the PR as such, but it needs a re-push to be effective).

Also, it would be good to write in `NEWS` under breaking changes or changed functionality what someone needs to do if they previously used `zeek::detail::hash_init()` and friends from `digest.h` (return type changes mostly and the length constants `ZEEK_*DIGEST_LENGTH` mostly?). Also, if a plugin user does explicitly require the right OpenSSL headers, what do they need to put in their CMakeLists.txt file? The `DEPENDENCIES OpenSSL::SSL OpenSSL::Crypto` part?

awelzel:
> Are you sure about that? Calling out a change for `zeek::detail`? That's by definition private API and must not be called from 3rd parties. At least that's my understanding of what a `detail` namespace is.
> 

We can leave it out, sure. Figured it could've been a "popular API" to easily create hashes. Community-id uses/used it, but maybe no-one else:

https://github.com/corelight/zeek-community-id/blob/3bcf3e4ce411a60f689a24607678bedba67a6654/src/communityid.bif#L13-L19

> Yes. However, thinking about it, aren't the plugins loaded by Zeek at runtime via `dlopen`? This means the OpenSSL symbols are already loaded and just adding `OPENSSL_INCLUDE_DIR` to build the plugin should suffice. Didn't try it, though.

Link dependency may not be necessary. Putting a cookbook approach how/where to get access to the `OPENSSL_INCLUDE_DIR` in CMakeLists.txt of a plugin would be good (I wouldn't know). Maybe it's just what community-id did before? Is that missing some hints from the Zeek installation?

```
include(MacDependencyPaths)
find_package(OpenSSL REQUIRED)
include_directories(BEFORE ${OPENSSL_INCLUDE_DIR})
```

Neverlord:
> Link dependency may not be necessary. Putting a cookbook approach how/where to get access to the `OPENSSL_INCLUDE_DIR` in CMakeLists.txt of a plugin would be good (I wouldn't know). Maybe it's just what community-id did before? Is that missing some hints from the Zeek installation though?
> 
> ```
> include(MacDependencyPaths)
> find_package(OpenSSL REQUIRED)
> include_directories(BEFORE ${OPENSSL_INCLUDE_DIR})
> ```

That `find_package` shouldn't be necessary. [Our package file](https://github.com/zeek/zeek/blob/master/src/ZeekConfig.cmake.in#L21) should already make sure that OpenSSL has been found by CMake. So I think it'd be just:

```cmake
target_include_directories(<your-target> SYSTEM BEFORE PRIVATE "${OPENSSL_INCLUDE_DIR}")
```

ckreibich:
Thank you guys. I have finally (as usual — apologies!) caught up on this PR. I have a question about our previous OpenSSL code that I cannot really make sense of. The new code consistently switches to OpenSSL's high-level digest functions — that makes sense. But we used to say:
```
#if ( OPENSSL_VERSION_NUMBER < 0x30000000L ) || defined(LIBRESSL_VERSION_NUMBER)
    EVP_MD_CTX* ctx;
#else
    MD5_CTX ctx;
#endif
```
What was the goal of this? It says that for _older_ OpenSSL versions we use the high-level API, but for newer ones we _don't_ — that strikes me as backward. Does anyone know why we did this?

A few other comments:

More context in the commit messages, and perhaps more modular but smaller commits, would be helpful. This was a pretty big change and all we'll have in the future is "Avoid OpenSSL header dependencies". This is always aspirational and of course not fatal, but so much of Zeek is under-documented that I'll gladly take erring on the side of verbosity. We also have far too many commits that say what's in the changeset, but not why.

Regarding the `NEWS` coverage, I think it captures the matter nicely, thank you.

@awelzel — I couldn't really parse this ...

> Putting a cookbook approach how/where to get access to the OPENSSL_INCLUDE_DIR in CMakeLists.txt of a plugin would be good (I wouldn't know).

... but I think we do need to expand docs on writing plugins at this point — at a minimum to explain how our new cmake API works, but ideally also to cover the overall approach to dependencies. (I really like the `DEPENDENCIES` approach, btw!)

To answer your question about plugins, @Neverlord: yes, plugins get hooked in via `dlopen`, and specifically with `RTLD_NOW`, so not with lazy resolution of symbols.

Regarding this:

> Maybe it's just what community-id did before? Is that missing some hints from the Zeek installation though?

Don't read too much into what Community ID did there :-) — I recall that all of the OpenSSL stuff landed iteratively and with bugfixes. I also don't think it was principled enough to even consider the problem of multiple OpenSSL installations, etc.

awelzel:
> The new code consistently switches to OpenSSL's high-level digest functions 

This hasn't changed. For newer OpenSSL versions it still uses the hash specific methods. But, there's now a single `#if-#else-#endif` implementing small hook functions for the respective version instead of the repeated `#if` across the files.

Compare these two do_init() functions:

https://github.com/zeek/zeek/blob/cc646f625ff6ce8f9ea033a7aef117cec0efed50/src/OpaqueVal.cc#L224

https://github.com/zeek/zeek/blob/cc646f625ff6ce8f9ea033a7aef117cec0efed50/src/OpaqueVal.cc#L314-L318

> What was the goal of this? It says that for older OpenSSL versions we use the high-level API, but for newer ones we don't — that strikes me as backward. Does anyone know why we did this?

Hearing @0xxon say that this is needed to serialize/unserialize the "state of the hash computation" and transferring it via broker.

> More context in the commit messages, and perhaps more modular but smaller commits, would be helpful. This was a pretty big change and all we'll have in the future is "Avoid OpenSSL header dependencies". This is always aspirational and of course not fatal, but so much of Zeek is under-documented that I'll gladly take erring on the side of verbosity. We also have far too many commits that say what's in the changeset, but not why.

Ack. I'll try to find a balance of lines changed vs intricacy vs commit-message details to bring that up in future reviews.

> @awelzel — I couldn't really parse this ...
>> Putting a cookbook approach how/where to get access to...

I was just fishing for what to put into NEWS to help the few plugin writers that do need access to OpenSSL's include path now that these are not implicitly added anymore. Switched to DM with Dominik which is where we (re-)found the `INCLUDE_DIRS` for `zeek_add_plugin()`. Agree on the under-documentedness.

Neverlord:
@ckreibich in this case, we actually want the "low-level" API. `EVP_MD_CTX*` is a level of indirection that we don't want or need. We have concrete hash types (like MD5) and want to pull out the current state (bytes) or restore (deserialize) a previous state.

0xxon:
> Hearing @0xxon say that this is needed to serialize/unserialize the "state of the hash computation" and transferring it via broker.

Yup, exactly that. But - I think we should get rid of that feature. It is not supported by OpenSSL anymore, and we can only still support it by this abuse of APIs that are not meant to be used anymore. This also potentially breaks if someone is using a different engine.

The last time I proposed that, there was some resistance to this though. It also would make the hash opaqueval-types unique - all other types are serializeable.

ckreibich:
Thanks a bunch for the clarifications, folks — I get it now. :+1:

PR #3426, opened 2023-11-01
Support for indexing "table[pattern] of X" with strings (follow-up)

### LINKED ISSUE #3402: Matching sets of regular expressions
### Discussed in https://github.com/zeek/zeek/discussions/3326

<sup>Originally posted by **vpax** September 22, 2023</sup>
@JustinAzoff and I were discussing how we could significantly speed up some of our code that currently loops through sets of `pattern`s if Zeek could tell us _which_ pattern(s) were matched.  For example, it would be neat if code like this:
```
global a: table[pattern] of count;

a[/foo/] = 3;
a[/bar/] = 7;
a[/(foo|bletch)/] = 9;

print a["x"], a["foo"], a["bar"], a["bletch"];
```
would print
```
[], [3, 9], [7], [9]
```
My recollection was that Zeek in fact has this information internally but it gets lost in the layering used for matching `pattern`s. I took a look in the internals and sure enough that's the case, and in fact this is leveraged by the signature framework to tell which of several signatures have matched.

I took a quick stab at prototyping and got the above example to work with < 150 lines of new code (`topic/vern/pattern-tables-work`).  Assuming that given its feasibility, there's interest in developing this further (which I'd hope, as it really is helpful in some situations!), several questions come to mind:

- is `table[pattern] of X` the right way to expose this feature to users? It's certainly convenient for some cases.
- should the match be _exact_, _anywhere_, or user-selectable (maybe via some syntax)? My prototype implements _exact_, which seems best as one can always add `.*` on either side to make a pattern _anywhere_.
- are we okay with not supporting `/.../i` or `/.../s` distinctions? (The prototype just loses these.) It would not be hard to make them apply to _all_ patterns, but due to the way the implementation works they would be difficult to support for only some of the patterns but not others. Un-pretty.

BTW I'm perfectly happy for someone else to pick up the prototype and build out the rest - just would like to see some functionality along these lines available sometime soon!</div>

awelzel:
Closes #3402, replaces #3416.

@vpax - this is your PR, reformatted to the new style and followed up with individual commits. I'm happy to squash them all into your original commits if you'd prefer or keep it separated. Mostly done for documentation.

I've dropped most of the RE.h / RE.cc changes as it seemed most things we wanted to do existed. Not sure that was quite the right call. I'm assigning Robin for review, maybe he has thoughts.


rsmmr (commented):
I only took a quick look but want to point out that this could become a performance rabbit hole: if people put lots of complex regular expressions into a table, we'll compile them all into a single NFA/DFA, which can easily explode in numbers of states---something that will be quite intransparent to users. We have the same challenge with regexes in signatures. There we group them and limit how large each group can get, which at least somewhat mitigates the issue.

rsmmr:
> lots of complex regular expressions into a table

Actually they don't even need to be that complex, just unbounded (`.*`).

awelzel:
> I only took a quick look but want to point out that this could become a performance rabbit hole: if people put lots of complex regular expressions into a table, we'll compile them all into a single NFA/DFA, which can easily explode in numbers of states---something that will be quite intransparent to users. We have the same challenge with regexes in signatures. There we group them and limit how large each group can get, which at least somewhat mitigates the issue.

Thanks for the input. I've found the `sig_max_group_size` mitigation and we could/can do that here, too.

Would @ckreibich [ideas of using bifs](https://github.com/zeek/zeek/discussions/3326#discussioncomment-7169654)) help the intransparency a bit? It wouldn't fix the issue you're bringing up, but possibly provide more visibility that there's more going on.

> I wonder whether BiFs would be the natural way to allow the different use cases like matching vs searching, while keeping table return value types consistent. This could be pattern_table_search(), pattern_table_match(), etc, returning vectors of the value type.

I suspect the `bif` would have to have a `vector of any` return type, unfortunately. But I'm also at only 50% for the special IndexExpr due to the return type not being consistent with the table's yield type, and no easy parameterization possible.

But then, if a bif is fine, this can all be done via an opaque type as well. Thinking`pararegex_init(t: table[pattern] of any)`, `pararegex_match(pr: opaque of pararegex, s: string): vector of any`, etc. Maybe there's some more room for discussing...

src/RE.h
```diff
@@ -96,6 +96,11 @@ class Specific_RE_Matcher {
     // to the matching expressions.  (idx must not contain zeros).
     bool CompileSet(const string_list& set, const int_list& idx);
 
+    // For use with CompileSet() to collect indices of all matched
+    // expressions into the matches vector. Behaves as MatchAll(),
+    // consuming the complete input string.
```
ckreibich, line 101: Might be good to expand that comment a bit since the `matches` result argument isn't something that's in the other (public) `MatchAll()` implementations.

ckreibich (commented):
I've read this over and tinkered with it a bit, and I think it's great.

As you may recall I was initially worried that adding further special-casing of tables would be too confusing, but I think the alternative (of dedicated BiFs) isn't any better since it would either involve a fair bit of `any` or not be as practical as the immediate vector return — so I'm sold. :-)

I wondered whether there could be a dedicated table-like type — something like `matchtable`, perhaps — that would also accommodate the `subnet` use case, but that doesn't work since while `table[subnet]` shares the apply-to-all-entries lookup pattern it yields only a single member. (If on the other hand we ever wanted a mode where a `table[subnet]` yields all matching entries, that might be another story.)

A very minor nit that might just be me but that gave me pause while tinkering is that asking `if ("foobar" in t)` for a `table[pattern]` seems backward — it's not whether the string is in the table, but whether any of the patterns are in the string. But that's just a matter of getting used to.

One request, though: I think we should align the new functionality with the behavior of `set[subnet]`. This works:
```
global s: set[subnet] = {
        192.168.1.0/24, 10.0.0.0/8,
};

print 192.168.1.1 in s;
```
but if you try the same for `set[pattern]` it breaks, since that's not handled yet:
```
global s: set[pattern] = {
        /.*foo.*/, /bar/,
};

print "foobar" in s;
```
This yields `type clash (pattern and foobar) ...`. I think this would be quite helpful to add, since I find it common during development to flip-flop between tables and sets as the correct target type for a feature, so more consistency would reduce surprise.

awelzel:
> 
> A very minor nit that might just be me but that gave me pause while tinkering is that asking `if ("foobar" in t)` for a `table[pattern]` seems backward

Vern's initial version didn't have `in` implemented (`|t["foobar"]| > 0` is the replacement). I thought `in` would be nice to work, but now I have second thoughts.

> One request, though: I think we should align the new functionality with the behavior of set[subnet].

If `in` didn't work on `table[pattern]`, would that still be relevant? Vern mentioned `set_to_pattern()` for not supporting `set[pattern]`. Currently, there's some inconsistency though, agreed.

So either (1) go back to the "bare" version where `in` doesn't work on the table and leave `set[pattern]` alone, or (2) extend  properly to `set[pattern]` so `in` works in both cases. Can see either with (2) being somewhat more consistent - IMO only.

ckreibich:
Fwiw, my "backwardness" comment was really just a nit — to me it's not a showstopper. I land of the side of supporting `in` simply because it seems odd to have those tables/sets and not support `in`, and because `in` is also supported for `table[subnet]` and `addr`s.

awelzel:
>  I land of the side of supporting in simply because it seems odd to have those tables/sets and not support in, and because in is also supported for table[subnet] and addrs.

I've added this for "normal scripts".

@vpax , would you mind doing the script_opt part as a follow-up, there's currently no failing tests (maybe should tweak CI at some point to actually do a full ZAM run).

@rsmmr - re-requesting a review. I'm up for nits and renaming, but also more fundamental changes ;-)

vpax:
I'm fine doing the script-opt part of this. (FYI, you can always do it locally via `btest -a zam ...`)

src/Expr.cc
```diff
@@ -2382,6 +2382,17 @@ IndexExpr::IndexExpr(ExprPtr arg_op1, ListExprPtr arg_op2, bool arg_is_slice, bo
     if ( IsError() )
         return;
 
+    if ( op1->GetType()->Tag() == TYPE_TABLE ) { // Check for a table[pattern] being indexed by a string
+        const auto& table_type = op1->GetType()->AsTableType();
+        const auto& rhs_type = op2->GetType()->AsTypeList()->GetTypes();
+        if ( table_type->IsPatternIndex() && table_type->Yield() && rhs_type.size() == 1 &&
+             IsString(rhs_type[0]->Tag()) ) {
+            is_pattern_table = true;
+            SetType(make_intrusive<VectorType>(op1->GetType()->Yield()));
```
rsmmr, line 2391: Vector preferable over set?

src/Type.cc
```diff
@@ -389,6 +389,13 @@ bool IndexType::IsSubNetIndex() const {
     return false;
 }
 
+bool IndexType::IsPatternIndex() const {
+    const auto& types = indices->GetTypes();
+    if ( types.size() == 1 && types[0]->Tag() == TYPE_PATTERN )
+        return true;
+    return false;
```
rsmmr, line 396:
```suggestion
    return types.size() == 1 && types[0]->Tag() == TYPE_PATTERN;
```

rsmmr: And maybe inline this?

src/Val.cc
```diff
@@ -1453,6 +1563,9 @@ void TableVal::Init(TableTypePtr t, bool ordered) {
     else
         subnets = nullptr;
 
+    if ( table_type->IsPatternIndex() )
+        pattern_matcher = new detail::TablePatternMatcher(this, table_type->Yield());
+
```
rsmmr, line 1568: I realize that this is following the existing pattern, but moving to `unique_ptr` would be nice (maybe as a follow up covering the existing cases too).

src/Val.cc
```diff
@@ -1420,6 +1421,115 @@ static void find_nested_record_types(const TypePtr& t, std::set<RecordType*>* fo
...
+// Support class for returning multiple values from a table[pattern]
+// when indexed with a string.
+class detail::TablePatternMatcher {
+public:
+    TablePatternMatcher(const TableVal* _tbl, TypePtr _yield) : tbl(_tbl) {
+        vtype = make_intrusive<VectorType>(std::move(_yield));
+    }
+
+    void Insert(ValPtr pat, ValPtr yield) { Clear(); }
+    void Remove(ValPtr pat) { Clear(); }
```
rsmmr, line 1433: Are these just supposed to clear the state on insert/assign? Then the method names seem misleading, I'd just skip the methods and call clear directly.

src/Val.cc
```diff
@@ -1420,6 +1421,115 @@ static void find_nested_record_types(const TypePtr& t, std::set<RecordType*>* fo
...
+    // they'll frequently be constructed just once), and also keeps us
+    // from having to re-build the matcher on every insert/delete in
+    // the common case that a whole bunch of those are done in a single
+    // batch.
+    std::unique_ptr<detail::Specific_RE_Matcher> matcher = nullptr;
+
+    // Maps matcher values to corresponding yields. When building the
+    // matcher we insert a nil at the head to accommodate how
+    // disjunctive matchers use numbering starting at 1 rather than 0.
+    std::vector<ValPtr> matcher_yields;
```
rsmmr, line 1467: So this is the keeping a `Val*` to every yield value in the table so that we can then index into it when we get a match? Looks a bit unfortunate that that additional state is necessary, both memory-wise (though not too bad) and also maintainance-wise (need to rebuild it on changes). However, I also don't see any better approach right now, so ok I guess.

rsmmr (approved):
Just small observations, otherwise LTGM (I didn't look at `script_opt/*`).

src/Expr.cc
vpax: The plus of a vector is that if the developer knows there can only be one match, they can get to it with `[0]`, whereas we don't have a convenient "give me an arbitrary set element" operator.

src/Val.cc
vpax: FWIW, it has that form so that the "how" of what `TablePatternMatcher` responds to inserts and removals remains isolated to its internals. (Originally it did more than just `Clear()`, so the isolation was more material.)

src/Expr.cc
awelzel: Also cheaper as we don't need to serialize values into and from a set, allows to mutate the values in case of non-atomic yield types. Non-lossy, too, if there's a use-case for the same value for different patterns - summing up scores of matched patterns?

src/Val.cc
vpax: IIRC it's just a ValPtr to an existing `Val*` that the table has in its dictionary anyway.

src/Val.cc
awelzel: I had almost removed them, will now do.

src/Val.cc
awelzel: To me sounded like templating the AcceptSet/Index type. Then we could associate ValPtr/Val* directly with matching expressions. I'm resolving - this could be interesting if we find more use cases though.

PR #3427, opened 2023-11-02
Deprecations for removals in v7.1

awelzel:
Few things I had in mind for a while.

Recently `policy/tuning/default` also came up as a good candidate. Just not sure about a good way forward there.


src/EventRegistry.h
```diff
@@ -68,8 +68,8 @@ class EventRegistry final {
     // themselves.
     void SetErrorHandler(std::string_view name);
 
-    string_list UnusedHandlers();
-    string_list UsedHandlers();
+    [[deprecated("Remove in v7.1 - Unused handlers are now found via UsageAnalyzer.")]] string_list UnusedHandlers();
+    [[deprecated("Remove in v7.1 - UsedHandlers() is not reliable.")]] string_list UsedHandlers();
```
timwoj, line 72: You changed the one usage of `UsedHandlers` to `AllHandlers`. Should the deprecation warning suggest the same?

timwoj (changes requested):

timwoj (approved):

PR #3428, opened 2023-11-03
Fix HTTP signature framework not supporting '$'

xb-anssi:
The `http-request-body` or `http-reply-body` conditions of the signature framework never match when the pattern uses a `$`, which makes it impossible to distinguish an HTTP body such as AB from one such as ABCD using the signature framework.

This adds a test which demonstrates the issue by having signatures with and without `$` and showing that those with `$` never match against a test PCAP with a few requests.

The fix is to let the signature framework know when the message body ends, which currently happens on EndOfData().


rsmmr (approved):
Nice catch, this looks like an oversight. 

The double delivery isn't good either but yeah we can look at that separately.

PR #3429, opened 2023-11-03
Upgrade c-ares submodule to 1.21.0

timwoj:
This updates us from 1.20.1 to 1.21.0. The changelog is here: https://c-ares.org/changelog.html#1_21_0

The big notable change is the new memory-safe parser architecture.


timwoj:
I pay attention to their releases, but that's about it.

PR #3433, opened 2023-11-06
Script optimization maintenance for November 2023

vpax:
Along with fixing some script optimization bugs (and warnings about signed/unsigned comparisons), this month's maintenance includes a number of optimization improvements. The most significant of these is more aggressive optimization of inlined function calls, which is also laying the groundwork for a large PR for collapsing event handlers that I'll be putting together once this is merged.


awelzel:
This tickles a bad_ref() followed by a null-pointer deref of `zeek::reporter` when the zeekjs plugin is enabled on Fedora 38. I'm sure it's an assumption or issue in the plugin.

```
(gdb) bt
#0  0x00005555579b72c2 in zeek::Reporter::DoLog (this=this@entry=0x0, prefix=prefix@entry=0x555555e9b920 "internal error", event=..., out=<optimized out>, conn=conn@entry=0x0, addl=addl@entry=0x0, location=<optimized out>, time=<optimized out>, postfix=<optimized out>, fmt=<optimized out>, ap=<optimized out>)
    at /home/awelzel/corelight-oss/zeek/src/Reporter.cc:483
#1  0x00005555579be322 in zeek::Reporter::InternalError (this=this@entry=0x0, fmt=fmt@entry=0x555555e9fc80 "bad reference count [%d]") at /home/awelzel/corelight-oss/zeek/src/Reporter.cc:211
#2  0x00005555579eb8ce in zeek::bad_ref (type=type@entry=2) at /home/awelzel/corelight-oss/zeek/src/Obj.cc:184
#3  0x0000555557fd3bd5 in zeek::Unref (o=<optimized out>) at /home/awelzel/corelight-oss/zeek/src/include/zeek/Obj.h:168
#4  zeek::IntrusivePtr<zeek::Obj>::~IntrusivePtr (this=0x7fffe6002800, __in_chrg=<optimized out>) at /home/awelzel/corelight-oss/zeek/src/include/zeek/IntrusivePtr.h:117
#5  std::_Destroy<zeek::IntrusivePtr<zeek::Obj> > (__pointer=0x7fffe6002800) at /usr/include/c++/12/bits/stl_construct.h:151
#6  std::_Destroy_aux<false>::__destroy<zeek::IntrusivePtr<zeek::Obj>*> (__last=0x7fffe6002860, __first=0x7fffe6002800) at /usr/include/c++/12/bits/stl_construct.h:163
#7  std::_Destroy<zeek::IntrusivePtr<zeek::Obj>*> (__last=0x7fffe6002860, __first=<optimized out>) at /usr/include/c++/12/bits/stl_construct.h:196
#8  std::_Destroy<zeek::IntrusivePtr<zeek::Obj>*, zeek::IntrusivePtr<zeek::Obj> > (__last=0x7fffe6002860, __first=<optimized out>) at /usr/include/c++/12/bits/alloc_traits.h:850
#9  std::vector<zeek::IntrusivePtr<zeek::Obj>, std::allocator<zeek::IntrusivePtr<zeek::Obj> > >::~vector (this=0x555558593560 <zeek::detail::AST_objs>, __in_chrg=<optimized out>) at /usr/include/c++/12/bits/stl_vector.h:730
#10 0x00007fffe985d4dd in __run_exit_handlers (status=0, listp=0x7fffe99f1820 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at ./stdlib/exit.c:116
#11 0x00007fffe985d61a in __GI_exit (status=<optimized out>) at ./stdlib/exit.c:146
#12 0x00007fffe98461d1 in __libc_start_call_main (main=main@entry=0x5555582de0cc <main(int, char**)>, argc=argc@entry=4, argv=argv@entry=0x7fffffffccd8) at ../sysdeps/nptl/libc_start_call_main.h:74
#13 0x00007fffe9846285 in __libc_start_main_impl (main=0x5555582de0cc <main(int, char**)>, argc=4, argv=0x7fffffffccd8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffccc8) at ../csu/libc-start.c:360
#14 0x00005555560da8a1 in _start ()
(gdb) print locations
Cannot access memory at address 0x10
(gdb) print zeek::reporter
$1 = (zeek::Reporter *) 0x0
(gdb) select 9

```

vpax:
@awelzel thanks a bunch for running that one down!

src/Expr.cc
```diff
@@ -116,6 +116,7 @@ int Expr::num_exprs = 0;
 Expr::Expr(ExprTag arg_tag) : tag(arg_tag), paren(false), type(nullptr) {
     SetLocationInfo(&start_location, &end_location);
     opt_info = new ExprOptInfo();
+    register_new_expr(ThisPtr());
```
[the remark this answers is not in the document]
vpax: The pinning only during main parsing is a good thought - that would keep things under our control. We could also extend `ThisPtr()` to work on `const` Exprs/Stmts, in which case when adding bare pointers to containers the AST analysis could also stash an IntrusivePtr version, rather than doing it wholesale for every possible AST node.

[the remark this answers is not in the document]
awelzel: > The pinning only during main parsing is a good thought

I actually meant that the pinning could be local to script_opt via an extra traversal. Just before doing the "critical" stuff, have a guard/scope class that keeps the AST objects alive until destruction.
```
+    class PinAstObjs : public TraversalCallback {
+    public:
+        PinAstObjs() { zeek::detail::traverse_all(this); };
+        TraversalCode PreStmt(const Stmt* stmt) override {
+            ast_objs.push_back(const_cast<Stmt*>(stmt)->ThisPtr());
+            return TC_CONTINUE;
+        }
+
+        TraversalCode PreExpr(const Expr* expr) override {
+            ast_objs.push_back(const_cast<Expr*>(expr)->ThisPtr());
+            return TC_CONTINUE;
+        }
+
+        std::vector<zeek::IntrusivePtr<Obj>> ast_objs;
+    };
+
+    PinAstObjs pin_ast_objs;
+
```

That costs an extra traversal, but doubt that matters too much. Also suspect ZAM is traversing the AST anyhow, so could also do it there. New hooks in Stmt/Expr constructors stashing away pointers isn't great.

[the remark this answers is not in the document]
vpax: I was thinking further about this and I believe it's actually hazardous. The problem is that during script optimization the AST evolves in an ongoing fashion, so new nodes get constructed and then potentially discarded. The danger is that a pointer to one of those winds up in a container, and it won't be caught by a particular AST traversal since it popped up in the flow of optimization subsequent to the traversal.

This has me thinking that a `ThisPtr()` that works for `const` objects is the best way to go. Here's a thought on how perhaps to do that safely: add an `Obj` wrapper class that will hold an `IntrusivePtr<Obj>` inside it, constructed from a `const Obj*`, but won't allow any access to the pointer. That way, there's no temptation to sneak around `const`-ness but we still address the pointer mis-aliasing issue.  What do you think?

src/script_opt/ScriptOpt.cc
```diff
@@ -56,7 +56,7 @@ bool is_lambda(const ScriptFunc* f) { return lambdas.count(f) > 0; }
 
 bool is_when_lambda(const ScriptFunc* f) { return when_lambdas.count(f) > 0; }
 
-const FuncInfo* analyze_global_stmts(Stmt* stmts) {
+size_t analyze_global_stmts(Stmt* stmts) {
```
awelzel, line 59: First thought: Could this new contract be avoided by handing out a pointer to a persistent instance of FuncInfo for global stmts that is updated after optimization with the contents of whats in `funcs[global_stmts_handle]`?

Hmm, after trying, that's still weird, but same as before.. The other thought I'd have is let go of `analyze_global_stmts()` and fold it into `analyze_scripts()` via an extra optional stmt pointer for the global statements and let that return the optimized init_stmts entry point instead.

That should remove the need for a contract completely. The new function with a parameter where there's only one valid way to call it seems nice.

---
```
diff --git a/src/script_opt/ScriptOpt.cc b/src/script_opt/ScriptOpt.cc
index bb203277c..d17e2f885 100644
--- a/src/script_opt/ScriptOpt.cc
+++ b/src/script_opt/ScriptOpt.cc
@@ -37,6 +37,9 @@ static std::unordered_map<const ScriptFunc*, LambdaExpr*> lambdas;
 static std::unordered_set<const ScriptFunc*> when_lambdas;
 static ScriptFuncPtr global_stmts;
 
+static size_t global_stmts_handle = -1;
+static FuncInfo global_stmts_func_info = {nullptr, nullptr, nullptr, 0};
+
 void analyze_func(ScriptFuncPtr f) {
     // Even if we're analyzing only a subset of the scripts, we still
     // track all functions here because the inliner will need the full list.
@@ -56,7 +59,7 @@ bool is_lambda(const ScriptFunc* f) { return lambdas.count(f) > 0; }
 
 bool is_when_lambda(const ScriptFunc* f) { return when_lambdas.count(f) > 0; }
 
-size_t analyze_global_stmts(Stmt* stmts) {
+FuncInfo* analyze_global_stmts(Stmt* stmts) {
     // We ignore analysis_options.only_{files,funcs} - if they're in use, later
     // logic will keep this function from being compiled, but it's handy
     // now to enter it into "funcs" so we have a FuncInfo to return.
@@ -72,11 +75,11 @@ size_t analyze_global_stmts(Stmt* stmts) {
     global_stmts->AddBody(stmts->ThisPtr(), empty_inits, sc->Length());
 
     funcs.emplace_back(global_stmts, sc, stmts->ThisPtr(), 0);
-    return funcs.size() - 1;
+    global_stmts_handle = funcs.size() - 1;
+    global_stmts_func_info = funcs[global_stmts_handle];
+    return &global_stmts_func_info;
 }
 
-const FuncInfo& get_global_stmts(size_t global_ind) { return funcs[global_ind]; }
-
 void add_func_analysis_pattern(AnalyOpt& opts, const char* pat) {
     try {
         std::string full_pat = std::string("^(") + pat + ")$";
@@ -595,6 +598,9 @@ void analyze_scripts(bool no_unused_warnings) {
 
     if ( reporter->Errors() > 0 )
         reporter->FatalError("Optimized script execution aborted due to errors");
+
+    // May have been replaced/updated.
+    global_stmts_func_info = funcs[global_stmts_handle];
 }
```

src/Expr.h
```diff
@@ -1502,6 +1508,8 @@ class ListExpr : public Expr {
     ExprPtr Reduce(Reducer* c, StmtPtr& red_stmt) override;
     StmtPtr ReduceToSingletons(Reducer* c) override;
 
+    bool AllConstantOps() const;
```
awelzel, line 1511: Alternative to a new name might be to make HasConstantOps() virtual and override for ListExpr?

src/script_opt/Expr.cc
```diff
@@ -2223,14 +2268,21 @@ StmtPtr ListExpr::ReduceToSingletons(Reducer* c) {
     return red_stmt;
 }
 
+bool ListExpr::AllConstantOps() const {
```
awelzel, line 2271: Doesn't seem  overly optimization specific, so could go in `src/Expr.cc`, too.

awelzel (commented):
Thanks for removing the hooks in the constructors. Doesn't make script_opt prettier :-/ I had one more look at the analyze_global_stmts() contract and left a comment. Would be nice to make that simpler.

Haven't really looked at the `src/script_opt` parts. Mainly just what happens to the rest.

src/Expr.h
vpax: Yeah, that's a better name. In principle it could be virtual and overridden for `ListExpr`, but to do that in a non-surprising way would entail implementing it for a bunch of other `Expr` subclasses too, which I'd rather avoid at this point given it's not needed in any other context.

src/script_opt/Expr.cc
vpax: Done!

awelzel (approved):
Will squash the fixups/rework and merge.

src/script_opt/ScriptOpt.cc
awelzel: I though getting rid of the "conditional two-step API" might be nice, but this is all detail/zeek-setup.cc and works, so :man_shrugging:

vpax: Thanks again for the help!

> getting rid of the "conditional two-step API" might be nice

Ah I see now. Yeah, would be slightly tidier to pass in the global statements to the call to `analyze_scripts()`. Will keep this in mind as a tweak next time I touch `zeek-setup.cc`. (The need for the special call later to execute the global statements remains, however, so it will still have a somewhat unobvious API.)

awelzel: > The need for the special call later to execute the global statements remains, however, so it will still have a somewhat unobvious API.

That's where I thought `analyzer_scripts()` could just return them (they might be the original ones, too). I might have missed another optimization related call in zeek-setup.cc though - you have a better picture there.

PR #3445, opened 2023-11-08
Update btest TestDirs to include misc and opt

awelzel:
...and fix the alloc-dealloc ASAN report as well as stale baseline updates.

I ran some `-O ZAM` and it ran into an alloc-dealloc-mismatch quickly. I remembered adding `zam/basic.test` a while go that should've caught this, but missed adding it to the TestDirs entry in btest.cfg.


vpax (commented):
LGTM

PR #3449, opened 2023-11-10
ZAM Optimization: Coalescence of Event Handlers

vpax:
This PR implements a new ZAM optimization, "coalescence" of event handlers that have multiple bodies into a single body that is optimized as a unit. Hyperfine runs show that the gains are significant: running the base scripts against a large DNS PCAP executes about 4% faster with this enabled.

Most of the changes are fairly straightforward. The main complexity comes from event groups, which can enable handlers to be turned on or off dynamically at run-time. If an event handler has _some_ of its bodies turned off but not _all_ of them, then it can't use the optimized-as-a-unit version of those handlers. This is dealt with by introducing the notion that `ScriptFunc` objects can, in addition to their regular vector of `Body` objects, also include an "alternate" body, which is only used when all of the event handlers are enabled.

With this approach, one semantic difference remains: with interpreted execution, if an event handler itself deactivates (or activates) its fellow handlers, currently that change has effect immediately, suppressing or enabling the execution on the current event of any lower-priority handlers. With ZAM optimization, those effects won't occur until after processing the current event. To my thinking, it's reasonable to state that it's undefined whether enabling/disabling event groups during event handler execution affects execution of subsequent handlers for the current event.


awelzel:
> The main complexity comes from event groups, which can enable handlers to be turned on or off dynamically at run-time.

I haven't looked much in detail, but suspect native plugins installing their own custom Stmt subclasses into function bodies will bite with coalescing as well. ZeekJS is the one that is out there, adding its own Stmt subclass as a body for each registered callback. I have another try-something-out-plugin which installs C++ event handlers and called out to wasmtime that used the same approach. In general, any plugin wanting to install "native" event handlers currently 

The ability to inject new Stmt types and execute "anything" provides a great extension hook for Zeek IMO. It's understood this is all `zeek::detail` and "you're on your own territory", but it would be nice to not dry it up.

The two plugins above use STMT_ANY for their Stmt subclasses. I wonder if we could alias that as `STMT_EXTERN` / `STMT_EXTENSION` / `STMT_PLUGIN` and void coalescing if any of the top-level statements in bodies looks as such.


> With this approach, one semantic difference remains: with interpreted execution, if an event handler itself deactivates (or activates) its fellow handlers, currently that change has effect immediately, suppressing or enabling the execution on the current event of any lower-priority handlers. [...] To my thinking, it's reasonable to state that it's undefined whether enabling/disabling event groups during event handler execution affects execution of subsequent handlers for the current event.

It has been leveraged to disable `zeek_init()` event handlers of modules to avoid errors/warnings when creating documentation. May consider it a trick and not too strong of an opinion, but it's been useful in that case.

https://github.com/zeek/zeek/blob/abc32b3b464e60098a8a78438d54895916dca725/scripts/zeekygen/__load__.zeek#L24-L34

vpax:
> I wonder if we could alias that as `STMT_EXTERN` / `STMT_EXTENSION` / `STMT_PLUGIN` and void coalescing if any of the top-level statements in bodies looks as such

I like that as a clean approach.

> It has been leveraged to disable `zeek_init()` event handlers of modules to avoid errors/warnings when creating documentation

My thinking had been that these sorts of uses could just use flags, but in your example they're not cooperating event handlers so hmmmm. Your particular example will still work because the coalescence code specifically skips `zeek_init()` (because it's always too big to coalesce anyway). One thought would be an attribute that if added to any handler for an event prevents any coalescence; this would be analogous to the `STMT_EXTERN` or such as framed above, and perhaps the two could use the same mechanism.

src/Func.cc
```diff
@@ -568,24 +568,62 @@ void ScriptFunc::AddBody(StmtPtr new_body, const std::vector<IDPtr>& new_inits,
...
+void ScriptFunc::EnablementChanged(bool all_enabled) {
+    if ( alt_body.empty() )
+        return;
+
+    if ( all_enabled )
+        SwitchToAlternative();
+    else
+        SwitchToRegular();
+}
+
```
awelzel, line 626: Instead of adding the "alternative bodies" and the state management in Func/ScriptFunc, could you see an approach where the EventHandler's complete FuncPtr is replaced with a script_opt specific Func subclass that either executes the inlined body, or delegates execution to the original function?

Sketch below, it compiles, but nothing tested :-) Should be enough to show the idea. It does need this new `HasEnabledAllBodies()` helper and the corresponding state set during `UpdateFuncBodies()`, but that seems fine and less intrusive and possibly useful elsewhere, too.

Probably some extra-overhead due to the indirection, but keeps the Func/ScriptFunc parts more tidy and doubtful the extra indirection matters significantly. I understand this is all just because of EventGroups :smiling_face_with_tear:

```
diff --git a/src/Func.h b/src/Func.h
index d0026e245..63c3e0f21 100644
--- a/src/Func.h
+++ b/src/Func.h
@@ -84,6 +84,8 @@ public:
      */
     bool HasEnabledBodies() const { return ! bodies.empty() && has_enabled_bodies; };
 
+    bool HasEnabledAllBodies() const { return ! bodies.empty() && has_enabled_all_bodies; };
+
     /**
      * Called when there's been a change in which bodies are enabled.
      * @param all_enabled  true if now all bodies are enabled, false otherwise.
@@ -176,6 +178,7 @@ private:
     // expose accessors in the zeek:: public interface.
     friend class EventGroup;
     bool has_enabled_bodies = true;
+    bool has_enabled_all_bodies = true;
 };
 
 namespace detail {
diff --git a/src/script_opt/Inline.cc b/src/script_opt/Inline.cc
index 27e05b1c6..51971961e 100644
--- a/src/script_opt/Inline.cc
+++ b/src/script_opt/Inline.cc
@@ -245,8 +245,11 @@ void Inliner::CoalesceEventHandlers(ScriptFuncPtr func, const std::vector<Func::
 
     PostInline(oi, func, true);
 
-    new_scope = func->GetScope();
-    func->SetAlternativeBody(merged_body, new_scope, func->FrameSize());
+    auto inlined_func = make_intrusive<AlternateScriptFunc>(merged_body, new_scope, func->FrameSize(), func);
+    // Replace the function for that EventHandler the delegating one.
+    auto* eh = zeek::event_registry->Lookup(func->Name());
+    eh->SetFunc(inlined_func);
+
     funcs.emplace_back(func, new_scope, merged_body, 0);
 
     auto pf = std::make_shared<ProfileFunc>(func.get(), merged_body, true);
diff --git a/src/script_opt/ScriptOpt.h b/src/script_opt/ScriptOpt.h
index 204993885..0360207fc 100644
--- a/src/script_opt/ScriptOpt.h
+++ b/src/script_opt/ScriptOpt.h
@@ -174,6 +174,32 @@ protected:
     bool is_alt = false;
 };
 
+class StmtList;
+using StmtListPtr = zeek::IntrusivePtr<StmtList>;
+
+// ScriptFunc subclass running a single inlined body if possible,
+// otherwise delegates to the original function with multiple bodies
+class AlternateScriptFunc : public ScriptFunc {
+public:
+    AlternateScriptFunc(zeek::detail::StmtListPtr merged_body, ScopePtr scope, int frame_size, ScriptFuncPtr orig_func)
+        : orig_func(orig_func) {
+        AddBody(merged_body, frame_size);
+        SetScope(scope);
+    };
+
+    ValPtr Invoke(zeek::Args* args, Frame* parent) const override {
+        // If the original function has all bodies enabled, run our
+        // inline one, otherwise delegate
+        if ( orig_func->HasEnabledAllBodies() )
+            return ScriptFunc::Invoke(args, parent);
+
+        return orig_func->Invoke(args, parent);
+    }
+
+private:
+    ScriptFuncPtr orig_func;
+};
+

```

vpax: > Instead of adding the "alternative bodies" and the state management in Func/ScriptFunc, could you see an approach where the EventHandler's complete FuncPtr is replaced with a script_opt specific Func subclass

Yeah, that's cleaner. I'll put together a new PR with that approach, thanks for flagging.

PR #3452, opened 2023-11-11
Bind scan_path to the scope; avoid heap allocation

Neverlord:
@awelzel I think that Coverity warning was a bit overzealous. However, `scan_path` shouldn't be heap allocated in the first place, so I think it's still a worthwhile change.


awelzel (approved):
Oh, this looks like a great improvement.

PR #3454, opened 2023-11-14
Add jq to dockerfile

mbispham:
Whenever using zeek docker I always install jq - https://github.com/jqlang/jq. In zeek RTD use of jq is widespread. Its ~1MB. These are the dependencies for jq:

- libtool
- make
- automake
- autoconf

### DIFF
--- a/docker/final.Dockerfile
+++ b/docker/final.Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get -q update \
  && apt-get install -q -y --no-install-recommends \
      ca-certificates \
      git \
+     jq \
      libmaxminddb0 \
      libnode108 \
      libpython3.11 \


bbannier:
Thanks for the patch @mbispham! While I agree that `jq` is useful, including "useful" tools is a slippery slope. Right now this image intentionally provides a minimal Zeek installation only, e.g., it even omits dependencies to compile Spicy analyzers, so that it can be a building block for differing use cases.

My gut feeling is that we wouldn't want to include `jq`.

ckreibich:
Slippery slope duly noted, but I think Matti has a point here. `jq` is such standard tooling that I expect it to be available, just like awk & grep. I'm actually surprised our btests don't require it anywhere, and `zeek-testing-cluster` does require it, so we [add it in CI](https://github.com/zeek/zeek/blob/master/.cirrus.yml#L683-L684).

The image is currently 548MB, and `jq` adds 1MB. I think that's acceptable.

Btw I'm happy to revisit the idea of adding a script in the image that adds other classes of tooling, like a build environment for plugins. I've repeatedly looked for the right minimal set of packages to install. Something to easily grab our test pcaps would also be nice.

docker/final.Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apt-get -q update \
  && apt-get install -q -y --no-install-recommends \
      ca-certificates \
      git \
+     jg \
```
ckreibich, line 21: I think you meant "jq". :-)

ckreibich (changes requested):

docker/final.Dockerfile
mbispham: Face meet palm!

bbannier:
> Slippery slope duly noted, but I think Matti has a point here. `jq` is such standard tooling that I expect it to be available, just like awk & grep. I'm actually surprised our btests don't require it anywhere, and `zeek-testing-cluster` does require it, so we [add it in CI](https://github.com/zeek/zeek/blob/master/.cirrus.yml#L683-L684).

The idea behind this image is not to be a CI image for Zeek, but instead provide enough Zeek to be useful. Currently all dependencies we install are directly required by Zeek or its components (e.g., we only install the pretty heavy Python toolchain since it is required for e.g., `zkg`, and it is so tightly coupled with Zeek that installing it and its dependencies later is not really feasible). We only provide `awk` and `grep` since they are part of the slimmed down Debian base image we use, but technically I don't think we necessarily want them either (I had users ask me why we didn't use e.g., Alpine which comes with a much smaller base image size by using busybox instead of the "real" tools).

> The image is currently 548MB, and `jq` adds 1MB. I think that's acceptable.
> 
> Btw I'm happy to revisit the idea of adding a script in the image that adds other classes of tooling, like a build environment for plugins. I've repeatedly looked for the right minimal set of packages to install. Something to easily grab our test pcaps would also be nice.

If we have to add anything that would be my preferred approach since users who prefer a "fuller experience" can explicitly opt into the additional size requirements without having to impose them on everyone. When working with this container I personally know exactly what I need, so I probably wouldn't use them, but we'd have all freedom to go completely wild in them (vim, emacs, gdb, lldb, different gcc& clang versions, ...).

ckreibich:
> The idea behind this image is not to be a CI image for Zeek

Yes, definitely. Adding `jq` also wouldn't remove the need to install it in the CI setup, since that's happening in the Github runner, not in our image. My point is more that it is just a matter of time before `jq` is a required bit of the tooling and at 1MB I don't think it's worth much debate. The past 1GB vs 500MB, that mattered. `git` pulling in ~100MB just so you can run `zkg`, that arguably matters too, but I think we're in agreement that it's worth it for a toolchain that works out of the box.

Do others have input here? It might just be me — I just find that I install `jq` all the time.

ckreibich (approved):

awelzel:
Not a too strong opinion, so I've asked @mbispham to rebase and squash. Will merge then.

That said, I never installed `jq` before, but I definitely miss `less` fwiw ;-)

I think it could be an interesting exercise to build an actual slim/bare version that lets go of `git`, `zkg`, `zeekctl`, `btest` with the goal of providing a container image that runs an individual Zeek processes and package/process management happens externally. But it's not being asked, so :man_shrugging: 

> Something to easily grab our test pcaps would also be nice.

Bind-mount? Putting something into the image to fetch pcaps remotely seems off.

```
~/zeek$ docker run -v$(pwd)/testing/btest/Traces:/Traces --rm -it zeek/zeek:6.1
root@a7464b696ae1:/# zeek -r /Traces/http/get.trace
```

I can see a script to install dev dependencies as being convenient. At the same time, a local custom container image isn't overly difficult and provides better caching behavior, too.
```
FROM zeek/zeek-dev
RUN apt-get update && apt-get install -y --no-install-recommends libpcap-dev and all the things

$ docker build -t my-zeek-dev .
```

PR #3458, opened 2023-11-15
strings.bif: Implement do_split_string() using RE_Match_State

### LINKED ISSUE #3455: `split_string_n` returns incorrect match if pattern includes `^`
I would expect exactly one element for `split_string_n("test", /^est/, T, 1)` since the pattern does not match the input. Instead Zeek returns `[t, est, ]`.

This got reported by a user out of band since they saw unexpected results for `match_pattern("test", /^est/)` (expectation: no match, behavior: match found).

awelzel:
This allows better control of BOL and EOL. MatchPrefix() / LongestMatch() always start with BOL.

Closes #3455.

---

Including some prep work in `RE.h / RE.cc...

### DIFF
--- a/src/RE.cc
+++ b/src/RE.cc
@@ -268,6 +268,7 @@ bool RE_Match_State::Match(const u_char* bv, int n, bool bol, bool eol, bool cle
 
         // Initialize state and copy the accepting states of the start
         // state into the acceptance set.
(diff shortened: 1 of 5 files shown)


awelzel:
Replaced by #3459 which is less intrusive and seems more reasonable.

PR #3459, opened 2023-11-16
strings.bif/do_split_string: Pass bol and eol to MatchPrefix()

### LINKED ISSUE #3455: `split_string_n` returns incorrect match if pattern includes `^`

awelzel:
This allows better control of BOL and EOL. MatchPrefix() / LongestMatch()
always start with BOL.

Closes #3455

### DIFF
--- a/src/RE.cc
+++ b/src/RE.cc
@@ -318,7 +318,7 @@ bool RE_Match_State::Match(const u_char* bv, int n, bool bol, bool eol, bool cle
     return accepted_matches.size() != old_matches;
 }
 
-int Specific_RE_Matcher::LongestMatch(const u_char* bv, int n) {
+int Specific_RE_Matcher::LongestMatch(const u_char* bv, int n, bool bol, bool eol) {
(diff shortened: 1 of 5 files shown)


timwoj (approved):
LGTM. This fix is definitely smaller than the other one. Are any of the bugfix commits worth putting into a separate PR?

awelzel:
> This fix is definitely smaller than the other one.  Are any of the bugfix commits worth putting into a separate PR?

I think so, yes. But `RE_Match_State`  sole usage in `RuleMatcher` makes me worried that reasonable looking fixes have unexpected effects on expectations there.

I'm following up with the `sub`, `gsub` fixes first mentioned on #3455.

PR #3460, opened 2023-11-16
ZAM Optimization: Coalescence of Event Handlers (v2)

vpax:
This is an update/replacement for https://github.com/zeek/zeek/pull/3449, following Arne's suggestion of subclassing `ScriptFunc` rather than adding a bunch of internal cruft to track whether to use the coalesced body or the individual bodies. The main changes from the previous PR are in `Func.h`/`Func.cc` (along with a couple of new BTests to ensure that the functionality works as expected), with modest additions to `script_opt/` stuff, mostly in `Inline.cc` and `ScriptOpt.h`.


src/EventRegistry.cc
```diff
@@ -172,11 +172,20 @@ void EventGroup::UpdateFuncBodies() {
...
+        bool func_changed = false;
+        for ( auto& b : func->bodies ) {
+            auto prev = b.disabled;
             b.disabled = std::any_of(b.groups.cbegin(), b.groups.cend(), is_group_disabled);
+            if ( prev != b.disabled )
+                func_changed = true;
+        }
+
+        if ( ! func_changed )
+            continue;
```
awelzel, line 184: I'll rewrite this a bit to track the all some/all body state via `|=` and `&=` within the loop, and removing the `any_of` / `all_of` part.

src/script_opt/Inline.cc
```diff
@@ -120,33 +122,171 @@ void Inliner::Analyze() {
...
+        // it takes more time to compile it than to just run it via the
+        // interpreter, it's a lose.
+        static std::string zeek_init_name = "zeek_init";
+        if ( func->Name() == zeek_init_name )
+            continue;
+
+        const auto& body = f.Body();
+
+        if ( func->GetKind() == Func::SCRIPT_FUNC && func->GetBodies().size() > 1 ) {
+            if ( event_handlers.count(func) == 0 )
```
awelzel, line 159: I'll condense this to the following during the merge. `operator[]` will default construct an entry when missing.
```
++event_handlers[func];
```

src/script_opt/Inline.cc
```diff
@@ -120,33 +122,171 @@ void Inliner::Analyze() {
...
+        // We use a special scope name so that when debugging issues we can
+        // see that a given variable came from coalescing event handlers.
+        auto p = install_ID(vi->Name(), "<event>", false, false);
+        p->SetType(vi->GetType());
+        param_ids.push_back(std::move(p));
+    }
+
+    auto new_scope = pop_scope();
+    auto orig_scope = func->GetScope();
+    func->SetScope(new_scope);
```
awelzel, line 222: This modifies the original function? I presume that's what wanted (though it's a bit surprising). Merging as is, but maybe worth another look.

testing/btest/scripts/base/utils/dir.test
```diff
@@ -1,4 +1,9 @@
 # @TEST-EXEC: btest-bg-run zeek zeek -b ../dirtest.zeek
+#
+# The following helps eliminate race conditions when Zeek startup is
+# slow due to ZAM compilation.
+# @TEST-EXEC: sleep 5
```
awelzel, line 5: I'll remove this unconditional `sleep 5` and increase below 10 seconds to 20 seconds. Effect should be similar.

src/script_opt/Inline.cc
vpax: Good catch - that's vestigial. Just pushed a fixup.

awelzel: I've picked the fixup last minute/second during the merge. I'll repush your branch and remove it from there so GitHub hopefully shows this PR as merged.

JustinAzoff:
HI!

Does this do the right thing if this script is ran?

```
redef exit_only_after_terminate=T;

event my_event() &priority=-10
	{
	print "first instance, lower priority";
	terminate();
	}

event my_event() &priority=10
	{
	print "second instance, higher priority";
	local c: conn_id;
	print c$orig_h;
	}

event zeek_init()
	{
	# Make sure execution of both handlers happens properly.
	schedule 1msec { my_event() };
	}
```

It should output

```
second instance, higher priority
expression error in ././foo.zeek, line 13: field value missing (c$orig_h)
first instance, lower priority
received termination signal
```

But I wonder if the expression error in the coalesced body will prevent the first instance from running?

vpax:
In general, script optimization reserves the right to behave differently for some error situations. For example, from the ZAM `README.md`:
> * The optimizer assumes you have ensured initialization of your variables.
If your script uses a variable that hasn't been set, the compiled code may
crash or behave aberrantly.

JustinAzoff:
That sounds reasonable, but I'm thinking about a particular case where that may be problematic.

Is it true that if a 3rd party script does something like handle `connection_state_remove`, and that script has an error, then potentially all other event handlers for `connection_state_remove` may not run?

basically the difference in "blast radius" of

> If your script uses a variable that hasn't been set, the compiled code **for your script** may crash or behave aberrantly.

and

> If your script uses a variable that hasn't been set, the compiled code **across all scripts** may crash or behave aberrantly.

vpax:
I see your point ... but the problem would still arise if the 3rd party script leads to a crash, rather than the more benign "behave aberrantly", due to its deficiencies.

One way to address this is that script optimization _does_ provide hooks to say "optimize only these files" (or functions), and the coalescence logic knows not to intermingle stuff that's optimized vs. stuff that isn't. So you could do the equivalent of `--optimize-file=/usr/local/zeek/share` (say) and it'll only optimized the installed scripts, not anything in addition.

Issue #3431, opened 2023-11-04
Allow plugins to convey exta state in `zeek::Packet`


ckreibich:
Some custom packet sources benefit from tucking additional metadata onto instances of `zeek::Packet`, for example the source's own flow ID that a packet belongs to. Such extensions currently require patching the Zeek sources, and it would be nice to support this via standard plugin mechanisms.

The `Packet` instance currently passed throughout the sources starts out as `current_packet` in the `PktSrc` base class. It should be possible to make that instance dynamic and access extra state via safe downcasts elsewhere in the plugin code, but details will take a closer look. Clearly that code is quite hot so details will matter.

PR #3468, opened 2023-11-20
Use std::move() where appropriate

timwoj:
This fixes a bunch of Coverity findings (too many to list)

### DIFF
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -891,7 +891,7 @@ void DNS_Mgr::Event(EventHandlerPtr e, const DNS_MappingPtr& old_dm, DNS_Mapping
         event_mgr.Enqueue(e, BuildMappingVal(old_dm), BuildMappingVal(new_dm));
 }
 
-ValPtr DNS_Mgr::BuildMappingVal(const DNS_MappingPtr& dm) {
+RecordValPtr DNS_Mgr::BuildMappingVal(const DNS_MappingPtr& dm) {
     if ( ! dm_rec )
         return nullptr;
 
--- a/src/iosource/Packet.h
+++ b/src/iosource/Packet.h
@@ -74,7 +74,7 @@ public:
      */
     Packet(int link_type, pkt_timeval* ts, uint32_t caplen, uint32_t len, const u_char* data, bool copy = false,
            std::string tag = "") {
-        Init(link_type, ts, caplen, len, data, copy, tag);
+        Init(link_type, ts, caplen, len, data, copy, std::move(tag));
     }
 
     /**
--- a/src/DNS_Mgr.h
+++ b/src/DNS_Mgr.h
@@ -32,6 +32,7 @@ class Val;
 class ListVal;
 class TableVal;
 class StringVal;
+class RecordVal;
 
 template<class T>
 class IntrusivePtr;
@@ -39,6 +40,7 @@ using ValPtr = IntrusivePtr<Val>;
 using ListValPtr = IntrusivePtr<ListVal>;
 using TableValPtr = IntrusivePtr<TableVal>;
 using StringValPtr = IntrusivePtr<StringVal>;
+using RecordValPtr = IntrusivePtr<RecordVal>;
 
 } // namespace zeek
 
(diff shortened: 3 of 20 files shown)


timwoj:
There's a lot of changes here that probably don't do anything in the end, and coverity is just being obnoxious about. I'll dig into it a bit more.

src/zeek-setup.cc
```diff
@@ -582,7 +582,7 @@ SetupResult setup(int argc, char** argv, Options* zopts) {
 
     if ( options.supervisor_mode ) {
         Supervisor::Config cfg = {};
-        cfg.zeek_exe_path = zeek_exe_path;
+        cfg.zeek_exe_path = std::move(zeek_exe_path);
```
awelzel, line 585: I guess I understand where this comes from, but if someone started to use `zeek_exe_path` further down in this very long function, this has great bug/confusion potential (this applies to other places as well). Performance here is irrelevant, so much I'd almost suggest another call to `util::detail::get_exe_path(argv[0])` rather than this `std::move()`.

For pleasing coverity sure, but it's very noisy and I can't say for the better (but then my limit for surprises is also pretty low).

src/DNS_Mgr.cc
```diff
@@ -906,7 +906,7 @@ ValPtr DNS_Mgr::BuildMappingVal(const DNS_MappingPtr& dm) {
     r->Assign(4, h ? std::move(h) : make_intrusive<StringVal>("<none>"));
     r->Assign(5, dm->AddrsSet());
 
-    return r;
+    return std::move(r);
```
awelzel, line 909: Do you know if it suggests this also for `std::shared_ptr`. Wouldn't the compiler recognize that it can move?

Neverlord: This move is only necessary because this is a type conversion. This is actually `ValPtr{std::move(dm)}`. If the function would return the same type, the compiler would move automatically (guaranteed copy elision) and adding a `move` would actually be a pessimization.

timwoj: > This move is only necessary because this is a type conversion. This is actually ValPtr{std::move(dm)}. If the function would return the same type, the compiler would move automatically (guaranteed copy elision) and adding a move would actually be a pessimization.

Ah that explains that then. I was trying to figure out why exactly the compiler wouldn't automatically elide these. As for the pessimization, a decently-new version of Clang complains when you override the elision so those are easy to catch. I ran into one already where I misread the section that needed to be changed from the Coverity finding.

src/zeek-setup.cc
timwoj: Yeah, this one isn't going to make or break any sort of performance. I'll drop it and mark the Coverity issue as ignored.

timwoj: @vpax echoed both of your concerns here as well: (a) there's plenty of places where this could trip someone up making later changes where a value as already been `move`d and (b) these add a lot of visual noise.

src/DNS_Mgr.cc
timwoj: > Unfortunately we really can't change all of these signatures to return the matching types.

I would entertain going through and fixing all of the ones in `detail` namespaces at least though.

timwoj:
I stripped this down to a few low-hanging fruit and left the larger bulk of the warnings alone. I could still see making the changes for any return value where we're returning to a `ValPtr` or from a bif, since those are `return` statements and you shouldn't be using the value anymore.

src/iosource/Packet.h
```diff
@@ -74,7 +74,7 @@ class Packet {
      */
     Packet(int link_type, pkt_timeval* ts, uint32_t caplen, uint32_t len, const u_char* data, bool copy = false,
            std::string tag = "") {
-        Init(link_type, ts, caplen, len, data, copy, tag);
+        Init(link_type, ts, caplen, len, data, copy, std::move(tag));
```
awelzel, line 77: I've come across `tag` in another context. It seems completely unused in-tree. Maybe we can deprecate it separately assuming #3431 will provide a different approach.

awelzel (approved):

PR #3469, opened 2023-11-22
CMakeLists: Fix OPENSSL_INCLUDE_DIR reference (for release/6.0)

awelzel:
Seems the intention was to always add OPENSSL_INCLUDE_DIR to the zeek_dynamic_plugin_base target, but finding of the OpenSSL package and setting OPENSSL_INCLUDE_DIR happens much later, so this was never functional.

Relates to #3408 - this is the fix for Zeek 6.0.3, #3420 is the better fix for the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,9 +287,6 @@ target_include_directories(
     zeek_dynamic_plugin_base
     INTERFACE $<INSTALL_INTERFACE:include> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
               $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
-if (OPENSSL_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1033,6 +1030,9 @@ endif ()
 # --with-openssl (which may be common).
 include_directories(BEFORE ${OPENSSL_INCLUDE_DIR})
 
+# Propagate non-standard OpenSSL includes to dynamic plugins.
+target_include_directories(zeek_dynamic_plugin_base INTERFACE "${OPENSSL_INCLUDE_DIR}")
```
Neverlord, line 1034: Could use this instead to suppress warnings from OpenSSL headers and make sure they are searched before other include paths:

```suggestion
target_include_directories(zeek_dynamic_plugin_base SYSTEM BEFORE INTERFACE "${OPENSSL_INCLUDE_DIR}")
```

Doesn't really matter that much, though. I can't remember ever seeing warnings from OpenSSL headers. So approving anyways.

Neverlord (approved):

timwoj (approved):

PR #3470, opened 2023-11-22
Spicy: allow providing file id in zeek::file_begin

0xxon:
Allow spicy parsers to generate their own file IDs and provide them to Zeek. This duplicates functionality that is currently possible (and used) by some binpac-based analyzers. One example for an analyzer creating its own file IDs is the SSL analyzer.


src/spicy/runtime-support.cc
```diff
@@ -774,10 +776,10 @@ std::string rt::fuid() {
     throw ValueUnavailable("fuid() not available in current context");
 }
 
-std::string rt::file_begin(const std::optional<std::string>& mime_type) {
+std::string rt::file_begin(const std::optional<std::string>& mime_type, const std::optional<std::string>& fuid) {
```
bbannier, line 779: Since down the line we take ownership anyway let's take a value here and `std::move` on `push` below.

src/spicy/runtime-support.cc
```diff
@@ -681,9 +681,11 @@ void rt::protocol_handle_close(const ProtocolHandle& handle) {
     c->analyzer->RemoveChildAnalyzer(handle.id());
 }
 
-rt::cookie::FileState* rt::cookie::FileStateStack::push() {
+rt::cookie::FileState* rt::cookie::FileStateStack::push(std::optional<std::string> fid_provided) {
     auto _ = hilti::rt::profiler::start("zeek/rt/file-stack-push");
-    auto fid = file_mgr->HashHandle(hilti::rt::fmt("%s.%d", _analyzer_id, ++_id_counter));
+    auto fid = fid_provided.value_or(file_mgr->HashHandle(hilti::rt::fmt("%s.%d", _analyzer_id, ++_id_counter)));
+    if ( find(fid) )
+        throw InvalidValue(hilti::rt::fmt("Duplicate file id %s provided", fid));
```
bbannier, line 688: We incur the duplicate check even if we asked Zeek to create a hash for us; I don't think we need to validate whether it works here ;)

```suggestion
    if (fid_provided && find(*fid_provided)) 
        throw InvalidValue(hilti::rt::fmt("Duplicate file id %s provided", *fid_provided));

    auto fid = fid_provided.value_or(file_mgr->HashHandle(hilti::rt::fmt("%s.%d", _analyzer_id, ++_id_counter)));
```

src/spicy/runtime-support.cc
```diff
@@ -681,9 +681,11 @@ void rt::protocol_handle_close(const ProtocolHandle& handle) {
...
     _stack.emplace_back(fid);
```
bbannier, line 689: Since you are at it:

```suggestion
    _stack.emplace_back(std::move(fid));
```

0xxon:
Please note that this needs merging of the docs subrepo on merge.

testing/btest/spicy/file-analysis-data-in-with-fuid.zeek
```diff
@@ -0,0 +1,63 @@
...
+	dec += spicy::base64_finish(d);
+
+	print self.file_id;
+	zeek::file_data_in(dec);
+    }
+    version : /[^-]*/;
+    dash    : /-/;
+    software: /[^\r\n]*/;
+
+    var file_id: string = zeek::file_begin("application/x-x509-ca-cert", "FaAaAaAaAaAaAaAaAa");
```
bbannier, line 35: Since this seems to be a straight copy of of `file-analysis-data-in`, what do you think about extending that test instead? One way to do that would be to instead initializing `file_id` in the decl to implement an `%init` hook which sets `file_id`, e.g.,

```ruby
on SSH::Banner::%init { self.file_id = ...}
```

You could move that init hook to its own file and then modify the test to pass that file via `%INIT` and handle the two test cases with `@TEST-START-NEXT` (requires naming the Zeek script currently passed as `%INPUT`).

I personally would be a big fan of using a string I can remember and type as fid, e.g., `MY_FILE_ID` instead of `FaAaAaAaAaAaAaAaAa`.

bbannier (changes requested):

src/spicy/runtime-support.cc
bbannier: Even with my suggestion we'd always invoke `HashHandle` (even if a value was provided). It would be great to avoid that even though it might require fatter conditionals.

src/spicy/runtime-support.cc
0xxon: As just discussed in chat - this is sadly not possible, since `spicyz` expects that this is passed by reference.

testing/btest/spicy/file-analysis-data-in-with-fuid.zeek
0xxon: Sounds good. I admittedly also considered doing something like this - but didn't do it because it complicated the base test quite a bit.

testing/btest/Baseline/spicy.file-analysis-data-in/output
```diff
@@ -1,3 +1,5 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 FM47gX3vI5ofQPm1li
 FZjUS57tUkGFTibv3
+FaAaAaAaAaAaAaAaAa
```
bbannier, line 4: Nit: leakage between test cases. Making this two cases should fix that.

0xxon: Ok, that I kind of agree with. I admittedly was lazy here :)

testing/btest/spicy/file-analysis-data-in.zeek
```diff
@@ -9,6 +10,13 @@
 # @TEST-EXEC: cat files.log | zeek-cut sha1 filename >files.log.tmp && mv files.log.tmp files.log
 # @TEST-EXEC: btest-diff files.log
 #
+# @TEST-EXEC: cat ssh.spicy ssh-2.spicy > ssh-test.spicy
```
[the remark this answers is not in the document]
0xxon: No, no specific reason. I find the other way to be slightly easier to understand *shrug*

PR #3474, opened 2023-11-27
[Spicy] Use shared ptr without atomic refcounts

bbannier:
This the Spicy bump accompanying zeek/spicy#1600.


bbannier:
We won't move forward with this ATM.

PR #3475, opened 2023-11-27
TLS: Update cipher consts and keyexchange parsing

0xxon:
Update cipher consts.

Furthermore some past updates have been applied to scriptland, but it was not considered that some of these also have to be applied to binpac code, to be able to correcly parse the ServerKeyExchange message.

(As a side-note - this was discovered due to a test discrepancy with the Spicy parser)


0xxon:
Test failure should be independent of this - looks like OpenSSL added a previously missing string.

PR #3478, opened 2023-11-29
signatures: Support custom event via [event_name] syntax

### LINKED ISSUE #3403: More flexible signature "event" interface
### Discussed in https://github.com/zeek/zeek/discussions/3372

<sup>Originally posted by **vpax** October 12, 2023</sup>
Currently, Zeek's signature framework supports the generation of a single event:
```
event signature_match(state: signature_state, msg: string, data: string)
```
and when writing signatures, all that's specified is the `msg` that goes with that event:
```
signature my-first-sig {
    ip-proto == tcp
    dst-port == 80
    payload /.*root/
    event "Found root!"
}
```
We've been developing Zeek scripts that need to dispatch on the particular signature matched. Currently this requires a bunch of comparisons of `msg` (and these can happen in different instances of `signature_match` event handlers, depending on the modularity associated with the content). An extension to deal with this would be to allow the `event` specifier in a signature to optionally be the name of an event to generate, rather than a string to associate with it:
```
signature my-first-sig {
    ip-proto == tcp
    dst-port == 80
    payload /.*root/
    event my_first_sig_event
}
```
(the `event` value could instead be a string like now, in which case `signature_match` is generated), or perhaps:
```
signature my-first-sig {
    ip-proto == tcp
    dst-port == 80
    payload /.*root/
    event "Found root!"
    zeek_event my_first_sig_event
}
```
where the given event would have to have a common type signature (which for the second example could be the same as the `signature_match` event).

This seems handy and pretty straightforward to implement. Thoughts?</div>

awelzel:
This change allows to specify a per signature specific event, overriding
the default signature_match event. It further removes the message
parameter from such events if not provided in the signature.

Closes #3403

---

zeek-docs: https://github.com/zeek/zeek-docs/pull/234


awelzel:
@timwoj , mind looking at this one. @vpax , this is what was discussed on #3403 , maybe want to skim? Usage looks reasonably nice (given the constraints).

timwoj (approved):

src/RuleAction.cc
```diff
@@ -15,23 +15,63 @@ using std::string;
...
+        reporter->Error("unknown event '%s' specified in rule", event_name);
+        return;
+    }
+
+    // Register non-script usage to make the UsageAnalyzer happy.
+    zeek::event_registry->Register(event_name, false /*is_from_script*/);
+
+    static const auto& signature_match_params = signature_match->GetFunc()->GetType()->ParamList()->GetTypes();
+    // Fabricated signature_match2() parameters for type checking.
+    static const std::vector<zeek::TypePtr> signature_match2_params = {signature_match_params[0],
```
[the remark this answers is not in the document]
vpax: (Oh it's because you're skipping `msg` - I had in my head that it would come last ...)

src/RuleAction.cc
```diff
@@ -15,23 +15,63 @@ using std::string;
...
+    }
+}
 void RuleActionEvent::DoAction(const Rule* parent, RuleEndpointState* state, const u_char* data, int len) {
-    if ( signature_match )
-        event_mgr.Enqueue(signature_match, IntrusivePtr{AdoptRef{}, rule_matcher->BuildRuleStateValue(parent, state)},
-                          make_intrusive<StringVal>(msg),
-                          data ? make_intrusive<StringVal>(len, (const char*)data) : val_mgr->EmptyString());
+    if ( handler ) {
+        zeek::Args args;
+        args.reserve(msg ? 3 : 2);
```
vpax, line 55: Is it worth doing the `reserve`? I figure it's only needed for code that gets banged on a lot and so you don't want the vector object doing extra work, or wasting memory.

src/RuleAction.cc
```diff
@@ -15,23 +15,63 @@ using std::string;
...
+    }
 }
 
-void RuleActionEvent::PrintDebug() { fprintf(stderr, "	RuleActionEvent: |%s|\n", msg); }
-
-RuleActionMIME::RuleActionMIME(const char* arg_mime, int arg_strength) {
-    mime = util::copy_string(arg_mime);
-    strength = arg_strength;
+void RuleActionEvent::PrintDebug() {
+    fprintf(stderr, "	RuleActionEvent: |%s (%s)|\n", msg->CheckString(), handler->Name());
```
vpax, line 69: Will `msg` always be defined here?

src/RuleAction.h
```diff
@@ -24,32 +31,31 @@ class RuleAction {
...
     void PrintDebug() override;
 
-    std::string GetMIME() const { return mime; }
+    const std::string& GetMIME() const { return mime; }
 
     int GetStrength() const { return strength; }
 
 private:
-    const char* mime;
+    std::string mime;
```
vpax, line 58: is improving `mime` here needed for this change, or just a makes-sense-to-tidy-while-we're-in-there tweak?

NEWS
```diff
@@ -69,6 +69,37 @@ New Functionality
...
+
+	event found_root(state: signature_state, data: string) { }
+
+  Passing an additional message parameter to a custom event is possible with the
+  following syntax. The custom event's parameters need to align with those for the
+  ``signature_match()` event:
+
+	signature event-with-msg {
+		payload /.*root/
+		event [found_root_with_msg] "the-message"
```
vpax, line 94: Perhaps remind the reader that `"the-message"` can instead be a string-valued Zeek expression (or just a variable?). I was initially surprised at the `message_as_id` BTest because of this

vpax:
@awelzel looks good, I just have some micro-questions, per what I've now entered above

src/RuleAction.cc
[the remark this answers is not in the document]
awelzel: Which is a bit of a quirk, but I don't think too bad. I changed the comment a bit.

src/RuleAction.cc
awelzel: This is probably not banged as much as other places, but std::vector allocation strategy is custom.

Not doing the `reserve()` result in 2, respective 3 allocate calls (at least with the std library selected here): https://godbolt.org/z/h1ceoerTd.

Using `reserve()` results in a single allocation in either case. I'll keep it.

src/RuleAction.cc
awelzel: Ouch, thanks!

src/RuleAction.h
awelzel: Yep, separate commit and has the "makes sense part" :-)

vpax:
@awelzel thanks for the clarifications, LGTM

Issue #3439, opened 2023-11-07
Update `Tunnel::max_depth` default to reflect modern encapsulation environments


ckreibich:
Per [discussion on Slack](https://zeekorg.slack.com/archives/CSZBXF6TH/p1699037435105629) our current default value of 2 is too low for modern environments, where a combo of VXLAN & Geneve is quite common. The `max_depth` variable defines the cutoff at which Zeek will stop decapsulating further.

We're looking for input on the number of lower-layer encapsulation levels that folks see in production to guide a choice of a new default — please chime in here if you can.

@awelzel, fyi.

awelzel:
I've surveyed the logs of five non-cloudy production deployments and none had "exceeded_tunnel_max_depth" in the weird_stats for the past year. Neither @initconf nor @dopheide-esnet have reported seeing it.

On Slack, where this [first came up recently](https://zeekorg.slack.com/archives/CSZBXF6TH/p1699454716025919?thread_ts=1699039132.123169&cid=CSZBXF6TH), the user reported a value of 3 being sufficient for their needs.

> Sure, it is a multi account AWS environment using the VPC mirroring feature with an AWS load balancer to direct the traffic from multiple accounts.
> Yes, the weird logs were generated when using default settings.

> A value of 3 was sufficient to get past the layers of encapsulation.

Based on that, I'd vote 4 for now, but would certainly be good to hear if others see "exceeded_tunnel_max_depth" and have done something about.

---

I do wonder if a weird is the right messaging in general. A `x` or `_` or `?` in a tunnel's history for packets not fully decapsulated would make it directly visible in a popular field. Maybe.

ckreibich:
I'm comfortable with a highly scientific choice of 4 based on your investigation. :-)

> A x or _ or ? in a tunnel's history for packets not fully decapsulated

As per our chat I like this a lot. I'd lean toward non-alphabet characters so the notion of upper/lowercase for directionality doesn't apply to them — so `?`, or perhaps `+`.

multicast-bytes:
Great conversation. 
In uncomfortable from a network forensics standpoint, where packets are on the network and analysis/recording is missed due to parameters such as these. 

Thus, I like the idea of using history to inform users that tunnel depth limit(s) where reached.

Indeed, I hit this issue myself whilst working in AWS where using their native AWS Traffic Mirroring via a AWS GWLB resulted in packets 3x encapsulated so to speak.

[...]

PR #3480, opened 2023-11-30
Bump Tunnel::max_depth to 4, support "X" in history

awelzel:
I'm happy to split this into two PRs, but they both were discussed in the #3439 ticket.

* Bump Tunnel::max_depth to 4
* Commits to lift history to `Session` and support a new history code/character 'X' when there's something "knowingly unkown" or some "didn't look further here" situation. Usually they should come with a weird or analyzer violation, too, but the history is a more popular field and directly visible.


testing/btest/Baseline/core.tunnels.max-depth-exceeded/conn.log.cut
```diff
@@ -0,0 +1,5 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+uid	id.orig_h	id.resp_p	id.resp_h	id.resp_p	proto	history	service	tunnel_parents
+CHhAvVGS1DHFjwGM9	1.1.1.1	4789	1.1.1.9	4789	udp	D	vxlan	-
+ClEkJM2Vm5giqnMf4h	2.2.2.2	4789	2.2.2.9	4789	udp	D	vxlan	CHhAvVGS1DHFjwGM9
+C4J4Th3PJpwUYZZ6gc	3.3.3.3	4789	3.3.3.9	4789	udp	D?	-	ClEkJM2Vm5giqnMf4h
```
awelzel, line 5: I could also see `x` and `X`  to allow for directionality (more interesting if it's extended to other meanings). The `?` at the end of the history may also look confusing when asking questions.

What's the meaning of D??  - what is the meaning of `D?` ?

timwoj: My vote is for x/X

timwoj (approved):

PR #3489, opened 2023-12-04
Use std::move in return values from bif methods to avoid copies

timwoj:
PR for benchmark pass

### DIFF
--- a/src/reporter.bif
+++ b/src/reporter.bif
@@ -169,7 +169,7 @@ function Reporter::get_weird_sampling_whitelist%(%): string_set
 		auto idx = zeek::make_intrusive<zeek::StringVal>(el);
 		set->Assign(std::move(idx), nullptr);
 		}
-	return set;
+	return std::move(set);
 	%}
 
 ## Sets the weird sampling whitelist
@@ -206,7 +206,7 @@ function Reporter::get_weird_sampling_global_list%(%): string_set
 		auto idx = zeek::make_intrusive<zeek::StringVal>(el);
 		set->Assign(std::move(idx), nullptr);
 		}
-	return set;
+	return std::move(set);
 	%}
 
 ## Sets the weird sampling global list
--- a/src/stats.bif
+++ b/src/stats.bif
@@ -57,7 +57,7 @@ function get_net_stats%(%): NetStats
 	if ( stat.filtered )
 		r->Assign(n++, stat.filtered.value());
 
-	return r;
+	return std::move(r);
 	%}
 
 ## Returns Zeek traffic statistics.
@@ -107,7 +107,7 @@ function get_conn_stats%(%): ConnStats
 
 	r->Assign(n++, zeek::detail::killed_by_inactivity);
 
-	return r;
+	return std::move(r);
 	%}
 
 ## Returns Zeek process statistics.
@@ -164,7 +164,7 @@ function get_proc_stats%(%): ProcStats
 	r->Assign(n++, static_cast<uint64_t>(ru.ru_oublock));
 	r->Assign(n++, static_cast<uint64_t>(ru.ru_nivcsw));
 
-	return r;
+	return std::move(r);
 	%}
 
 ## Returns statistics about the event engine.
@@ -191,7 +191,7 @@ function get_event_stats%(%): EventStats
 	r->Assign(n++, event_mgr.num_events_queued);
 	r->Assign(n++, event_mgr.num_events_dispatched);
 
-	return r;
+	return std::move(r);
 	%}
 
 ## Returns statistics about reassembler usage.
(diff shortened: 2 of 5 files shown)


timwoj:
Yep, this one isn't worth the effort. I'll mark all of these Coverity findings as `Ignore`.

bbannier:
> Yep, this one isn't worth the effort. I'll mark all of these Coverity findings as `Ignore`.

Even though this might not make a performance difference it might still be worthwhile to do these cleanups. Zeek uses C++17 and `std::move` is exactly the tool needed to trigger the right constructors for these cases. If this would have been in C++ and not bif files clang-tidy might even have reported the previous pattern as inefficient similar to what Coverity did. Just muting these checks leaves performance on the table and if done globally could even introduce inefficiencies across the board (something we are working on improving in Spicy all the time to not die by a 1000 cuts).

timwoj:
> > Yep, this one isn't worth the effort. I'll mark all of these Coverity findings as `Ignore`.
> 
> Even though this might not make a performance difference it might still be worthwhile to do these cleanups. Zeek uses C++17 and `std::move` is exactly the tool needed to trigger the right constructors for these cases. If this would have been in C++ and not bif files clang-tidy might even have reported the previous pattern as inefficient similar to what Coverity did. Just muting these checks leaves performance on the table and if done globally could even introduce inefficiencies across the board (something we are working on improving in Spicy all the time to not die by a 1000 cuts).

I agree, but others have concerns that this adds visual noise to the code. The benchmark pass says that it's basically a wash performance-wise. Some are higher, some are lower, but none significantly higher or lower except the data center pcap which is about 2.5% slower.

awelzel:
> I agree, but others have concerns that this adds visual noise to the code. The benchmark pass says that it's basically a wash performance-wise. Some are higher, some are lower, but none significantly higher or lower except the data center pcap which is about 2.5% slower.

You probably would need to add a new micro-benchmark that focuses just on one of the changed BiFs, and then probably also use hyperfine instead of the benchmarker setup, and even then I'd be surprised if you see much.

timwoj:
Alright, I stepped through what's actually being executed here, using the `to_subnet` bif as an example. In this case, the 

Existing code: no `std::move` in the bif, `BifReturnVal` constructor calls `AdoptRef{}`:

- `IntrusivePtr` move constructor
- `BifReturnVal` constructor
- `IntrusivePtr` `AdoptRef` constructor
- `IntrusivePtr` destructor (for the function object)

Changing `BifReturnVal` constructor to use `std::move`:

- `IntrusivePtr` move constructor
- `BifReturnVal` constructor
- `IntrusivePtr` move constructor
- `IntrusivePtr` pass-by-value copy constructor
- `IntrusivePtr` destructor (for the function object)

Changing `to_substr` to return via `std::move`:

- `IntrusivePtr` move constructor
- `BifReturnVal` constructor
- `IntrusivePtr` `AdoptRef` constructor
- `IntrusivePtr` destructor (for the function object)

Basically, it appears the compiler is already doing the most efficient thing (calling the `intrusivePtr` move constructor). One could argue that creating the extra `AdoptRef` object doesn't need to happen. My guess is that Coverity is complaining about calling the `BifReturnVal` constructor pass-by-value, even tho the argument was moved.

awelzel (approved):
Chatted with Tim. I'm leaning on the "mostly noise" side and will probably not do this naturally in new code, but my opinion isn't strong enough, so... approving.

@ckreibich / @bbannier - any other thoughts?

bbannier (approved):
> I'm leaning on the "mostly noise" side and will probably not do this naturally in new code, ...

To me this sentiment still sounds like one should prefer a declaration `int foo(std::string x)` over `int foo(const std::string& x)` due to the "noise" of `const &` while I think we all agree that that would be silly and leaving performance on the table for absolutely no reason.

Since even noticing inefficient code as fixed here can be tricky we'd ideally we would have a warning enabled which rejects such inefficient code automatically (we have `-Werror` builds now), but I am not sure that exists.

Thanks for doing this @timwoj!

awelzel:
One completely different thought:

The `BifReturnVal` that this is related to has been introduce to transparently deal with `Val*` and `ValPtr` returns from BIFs.

Support for the  BifReturnVal `Val *` constructor was deprecated and removed with Zeek 4.1 (0618be792f210e29af8145e7e52225d26c27946e) so I think we can assume all bifs return ValPtr.

Couldn't we change `bifcl` today to render `ValPtr` from BIFs instead and ditch BifReturnVal.h altogether?

From a prior PR (#3468 / https://github.com/zeek/zeek/pull/3468#discussion_r1400228758), this won't help with the coverity warnings, but seems a nice cleanup as a follow up.

PR #3491, opened 2023-12-05
spicy/runtime-support: Use Vector::unsafeBegin() for args iteration

awelzel:
Use unsafe iteration from https://github.com/zeek/spicy/pull/1609

### DIFF
--- a/src/spicy/runtime-support.cc
+++ b/src/spicy/runtime-support.cc
@@ -197,7 +197,8 @@ void rt::raise_event(const EventHandlerPtr& handler, const hilti::rt::Vector<Val
 
     Args vl = Args();
     vl.reserve(args.size());
-    for ( const auto& v : args ) {
+    for ( auto it = args.unsafeBegin(); it != args.unsafeEnd(); it++ ) {
+        const auto& v = *it;
         if ( v )
             vl.emplace_back(v);
         else


awelzel:
Iterating over a `hilti::rt::Vector` instance results `std::shared_ptr creation` and atomic instruction overhead for every iterated element. This switched `raise_event()` to the "unsafe" interface iterating directly over the underlying vector to avoid the unneeded overhead.


For a tiny Spicy analyzer, results in significant overhead reduction.
```
|:---|---:|---:|---:|---:|
| `taskset -c 0 /opt/zeek-dev-prod/bin/zeek -D -C -b  protocols/bench  -r ../issues/bench-traffic/bench-traffic-32-10000.pcap  Bench::t=spicy` | 5.684 ± 0.064 | 5.633 | 5.847 | 1.07 ± 0.02 |
| `taskset -c 0 /home/awelzel/corelight-oss/zeek/prod-build/src/zeek -D -C -b  protocols/bench  -r ../issues/bench-traffic/bench-traffic-32-10000.pcap  Bench::t=spicy` | 5.318 ± 0.069 | 5.248 | 5.460 | 1.00 |
```

Following flamegraph zooming into a done hook. In this case `spicy::rt::raise_event` is completely dominated by 1) iteration and `std::shared_ptr` overhead and 2) EventMgr's flare `write()` system call  (which has just been removed with #3483).

Screenshot from 2023-12-05 16-55-16

auxil/spicy
bbannier, line 1: This needs to be replaced with a bump of Spicy.

bbannier (approved):

PR #3492, opened 2023-12-05
ZAM optimization fixes & improvements

vpax:
The AST optimization used by ZAM makes heavy use of CSE = Common Subexpression Elimination, which means identifying when a subexpression (e.g., `c$ssl`) is computed multiple times and later instances can be replaced by the temporary variable used to hold the first instance of the computation. The trick in doing so is to make sure that the reuse is _safe_, i.e., that nothing relevant has changed since the first instance was computed (in our example, that `c` hasn't been reassigned, nor `c$ssl`). It's easy to take a conservative approach to safety (for example, assume that any function call could modify any aggregate or non-local variable), but doing so misses out on many optimization opportunities. This gets particularly important given how `&default` attributes on `table`s and `record`s can result in function calls (as can `&on_change` and `&default_insert`) ... and those function calls can themselves modify globals or trigger further function calls by accessing `table`s or constructing `record`s, for example. A conservative approach would be to not allow any CSE across table references or modifications, or record construction/coercion, even though the reality is that it's almost never the case that any associated attribute functions have problematic side effects.

Previously, AST optimization was both overly conservative and incorrect (missed some situations where side effects could invalidate CSE). The heart of this large PR is to completely replace the previous technique with one that much more accurately identifies side effects. Running the default scripts on a large DNS trace, this change eliminates 40% of the run-time loads of subrecords from records (like `c$ssl` above) beyond what the previous approach yielded.

Along the way I had to fix a number of bugs, some of them unrelated to the main work but happening to be tickled by its results. In addition, one of the bugs turned out to make the optimizer do _much_ more work than it needed to (as well as leading it to overly conservative conclusions). Removing this speeds up the total `-O ZAM` startup overhead by a bit more than 40% - quite striking!


src/Func.h
```diff
@@ -258,6 +262,12 @@ class ScriptFunc : public Func {
     StmtPtr CurrentBody() const { return current_body; }
     int CurrentPriority() const { return current_priority; }
 
+    /**
+     * If this function is an instantiated lambda, returns the primary from
+     * which it was constructed. Otherwise, returns this function itself.
+     */
+    const ScriptFunc* Primary() const { return primary ? primary.get() : this; }
```
awelzel, line 269: Hmm, isn't this information already tracked in the AST via `LambaExpr::PrimaryFunc()`? There's also this comment in `LambdaExp::Eval()`:

https://github.com/zeek/zeek/blob/d1e7c8b81ced00378bf430348283ea866c62f2cf/src/Expr.cc#L4250-L4252

It's a bit indirect, but couldn't script_opt leverage that instead of introducing a new API in ScriptFunc - which didn't need to know so far it's a lambda - and for which there's seemingly little value generally?

The following seems to pass the `opt` tests (not sure they cover anything here though) and would allow to skip the `ScriptFunc` changes. Would that work for you?

```
diff --git a/src/script_opt/ProfileFunc.cc b/src/script_opt/ProfileFunc.cc
index 5df70eb6a..fc5abc6cb 100644
--- a/src/script_opt/ProfileFunc.cc
+++ b/src/script_opt/ProfileFunc.cc
@@ -779,6 +779,7 @@ void ProfileFuncs::ComputeBodyHashes(std::vector<FuncInfo>& funcs) {
     for ( auto& l : lambdas ) {
         auto pf = ExprProf(l);
         func_profs[l->PrimaryFunc().get()] = pf;
+        lambda_primaries[l->Name()] = l->PrimaryFunc().get();
         ComputeProfileHash(pf);
     }
 }
@@ -1349,7 +1350,8 @@ bool ProfileFuncs::AssessSideEffects(const SideEffectsOp* se, SideEffectsOp::Acc
 }
 
 std::shared_ptr<SideEffectsOp> ProfileFuncs::GetCallSideEffects(const ScriptFunc* sf) {
-    sf = sf->Primary();
+    if ( lambda_primaries.count(sf->Name()) > 0 )
+        sf = lambda_primaries[sf->Name()];
 
     auto sf_se = func_side_effects.find(sf);
     if ( sf_se != func_side_effects.end() )
diff --git a/src/script_opt/ProfileFunc.h b/src/script_opt/ProfileFunc.h
index af1206623..27c68a215 100644
--- a/src/script_opt/ProfileFunc.h
+++ b/src/script_opt/ProfileFunc.h
@@ -531,6 +531,9 @@ protected:
     // to regular functions, and for that it suffices.
     std::unordered_map<const ScriptFunc*, std::shared_ptr<ProfileFunc>> func_profs;
 
+    // Map lambda names to their primary functions
+    std::unordered_map<std::string, const ScriptFunc*> lambda_primaries;
+
     // Tracks side effects associated with script functions. If we decide in
     // the future to associate richer side-effect information with BiFs then
     // we could expand this to track Func*'s instead.

```

src/Stmt.h
```diff
@@ -122,6 +122,7 @@ class IfStmt final : public ExprStmt {
     StmtPtr DoReduce(Reducer* c) override;
 
     bool NoFlowAfter(bool ignore_break) const override;
+    bool CouldReturn(bool ignore_break) const override;
```
awelzel, line 125: This could've also been realized as a `TraversalCallback`, no? (and probably `NoFlowAfter`, too).

Not a too strong opinion, but I'm mostly looking at things from the perspective of "what if someone wanted to do this out-of-tree, do we provide the right tools" and "we should use these tools in-tree as well".

vpax: I can see some of these going either way. If you look at the `TraversalCallback`s used for script optimization, they generally have (1) `switch` statements to determine "which node is it this time", along with (2) logic that implements analysis that a node can't readily do locally for itself. For `CouldReturn`, using traversal will (a) do a whole bunch of unnecessary calls, (b) have some complexity due to needing to track potentially nested `switch` nodes (which change the `ignore_break` setting). In this case, I also wanted to leave room for nodes caching the result, which the traversal framework doesn't cleanly support since it's only for `const` objects.

src/Func.h
vpax: Thanks for flagging (and for the diffs!), that's much tidier. I've now pushed this change, assuming I didn't screw up my `git rebase` fu.

src/script_opt/FuncInfo.cc
```diff
@@ -0,0 +1,506 @@
...
+//
+// Broker::*
+//	These can manipulate unspecified (at script level) records.
+//
+// Cluster::publish_hrw
+// Cluster::publish_rr
+//	These call script functions to get topic names.
+//
+// Log::__write
+//	Calls log policy functions.
```
awelzel, line 476: The new `Log::__delay_finish()` invokes callbacks

src/script_opt/ZAM/Compile.h
```diff
@@ -53,7 +52,7 @@ class OpaqueVals {
 
 class ZAMCompiler {
 public:
-    ZAMCompiler(ScriptFunc* f, std::shared_ptr<ProfileFunc> pf, ScopePtr scope, StmtPtr body,
+    ZAMCompiler(ScriptFunc* f, ProfileFuncs& pfs, std::shared_ptr<ProfileFunc> pf, ScopePtr scope, StmtPtr body,
```
awelzel, line 55: Nothing to do, because this is down in the weeds and not an API for someone to use.  Mostly amusing how this is using a raw pointer, a reference, a few std::shared_ptrs and then an IntrusivePtrs, too.

I guess `ProfileFuncs` could just be a raw pointer, it's just as unsafe as storing the reference, no?

awelzel (commented):
Approving, I haven't looked at `script_opt/` except for what GitHub didn't collapse and that only skimmed very quickly. I hope you have no expectations here.

src/Attr.h
```diff
@@ -150,5 +150,8 @@ class Attributes final : public Obj {
 // it will be returned in err_msg.
 extern bool check_default_attr(Attr* a, const TypePtr& type, bool global_var, bool in_record, std::string& err_msg);
 
+// Returns the script-level name associated with an attribute, e.g. "&default".
+extern const char* attr_name(AttrTag t);
```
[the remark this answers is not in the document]
vpax: Heh ... yeah, yet another sometimes-consistent-sometimes-not ...

src/script_opt/FuncInfo.cc
vpax: Thanks for flagging! (And I'm pleased to see that the new ZAM maintenance helper script indeed correctly flags it, and others, as requiring inspection.)

The nice thing about how the code deals with BiFs is new/unanalyzed ones are by default treated as "who knows what side effects this might have", so if it took me a while to get around to looking over `Log::__delay_finish()` there wouldn't have been any harm.

src/script_opt/ZAM/Compile.h
vpax: Oof, yeah it is unnecessarily cluttered. I've noted this for potential streamlining when I do this month's script optimization (which I've put on hold until this PR is resolved & merged, given the number of changes it introduces).

PR #3493, opened 2023-12-06
iosource: Specify libkqueue dependency for subdir lib

awelzel:
Since a while my build has been spilling the following warnings:

    [18/1687] Building C object auxil/c-ares/src/lib/CMakeFiles/c-ares.dir/ares__addrinfo2hostent.c.o
    cc1: warning: zeek/prod-build/libkqueue-build/include: No such file or directory [-Wmissing-include-dirs]

My take is that FindKqueue extends the include directories globally and
tickles this warning because c-ares is built first. Grepping around,
<sys/event.h> is only included in iosource/Manager.cc, so we should
be able to reduce the exposure just to the iosource subdir.

---

To go with: https://github.com/zeek/cmake/pull/101

### DIFF
--- a/src/iosource/CMakeLists.txt
+++ b/src/iosource/CMakeLists.txt
@@ -1,5 +1,9 @@
 zeek_add_subdir_library(
     iosource
+    INCLUDE_DIRS
+    ${LIBKQUEUE_INCLUDE_DIRS}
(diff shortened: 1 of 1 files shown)


Neverlord (approved):
LGTM 👍

PR #3496, opened 2023-12-08
Monthly script optimization maintenance - Dec 23

vpax:
This month's script optimization maintenance mainly focuses on the recent introduction of indexing `table[pattern] of T` values with `string`s, and "delayed" logging. There's also more minor tidying, including (inspired by a comment of @awelzel's) reducing the number of different types of pointers used by script optimization classes, a change that's quite straight forward but touches a lot of files.

There are two substantive changes to the main (non-script-optimization) code. The first fixes a subtle bug. When Zeek terminates, it tells the `notifier::detail` `Registry` used by `Trigger` objects (to be informed about changes to values they're interested in) to likewise terminate. When the `Registry` terminates, it in turn tells each `Trigger` object registered with it to terminate. However it turns out that the `Trigger` terminations can lead to `Frame` resets, and those in turn can delete _other_ `Trigger` objects. If those other objects are _also_ in the `Registry` and it hasn't yet terminated them, then during the `Registry`'s subsequent access to them (to terminate them) it can access now-deallocated memory. (This problem only manifests in the BTests when using ZAM. I haven't been able to fully figure out why. The interpreter and ZAM set up somewhat different `Trigger` constructions for nested `when` statements. Both constructions appear to work, and at a glance the ZAM one actually looks more likely to be fundamentally correct than the interpreter one.)

Probably the best fix for this problem would be to rework the `Registry` to use smart pointers, so it can safely hold on to its list-of-`Trigger`s-to-terminate even if telling them to terminate leads to other `Trigger`s being `Unref`'d. And this would in fact be easy if the `Registry` dealt directly with `Trigger`s, because then it could just use `IntrusivePtr<Trigger>`. However, instead it deals with an intermediary class, `Receiver`, and `Trigger` objects use multiple inheritance to be both `Obj` and `Receiver` subclasses. It turns out that the only `Receiver` subclasses in the code base are `Trigger`s, so we could in fact just get rid of `Receiver` and switch the `Registry` to dealing with `IntrusivePtr<Trigger>` objects. I didn't go this route (yet), however, because I didn't know whether there's a desire to keep the functionality available for future use by something other than `Trigger` objects. So the fix I introduced was to have `Trigger`s explicitly tell the `Registry` when they're all done (they already implicitly did so), and the `Registry` remembers that fact and uses it to avoid accessing a now-invalid `Receiver`/`Trigger` pointer. Happy to rework with the cleaner approach if there's agreement that it's the better way to go.

Second (and much more minor), some recently introduced `TableVal` methods required `const StringValPtr&` arguments, for which ZAM code had to go out of its way to construct an `IntrusivePtr` from a bare underlying pointer. These methods now support both `const StringValPtr&` and `const StringVal*` versions, which saves it from having to do so.


src/Notifier.cc
```diff
@@ -58,13 +70,20 @@ void Registry::Modified(Modifiable* m) {
 }
 
 void Registry::Terminate() {
-    std::set<Receiver*> receivers;
```
awelzel, line 61: I haven't tried, and this might be completely off, but could rewriting the original `Terminate()` as something like the following work without the extra final_receivers? Intuitively by the comments, Triggers should fully unregister themselves and at the end of it registrations be empty (or this results in an endless loop :sweat: )

```
while ( ! registrations.empty() ) {
  const auto& it = registrations.begin();
  it->second->Terminate();
}
```

src/Val.h
```diff
@@ -869,12 +869,14 @@ class TableVal final : public Val, public notifier::detail::Modifiable {
     // For a table[pattern], return a vector of all yields matching
     // the given string.
     // Causes an internal error if called for any other kind of table.
-    VectorValPtr LookupPattern(const StringValPtr& s);
+    VectorValPtr LookupPattern(const StringValPtr& s) { return LookupPattern(s.get()); }
+    VectorValPtr LookupPattern(const StringVal* s);
```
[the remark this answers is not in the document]
vpax: Point taken. I'll push a reversion shortly.

Also, in the process of testing this I found a more significant issue in how script optimization's AST analysis identifies functions that might be called indirectly. That led to some significant revamping of the tracking-of-function-properties functionality (all inside `src/script_opt/`) which is part of what I'm pushing, too.

src/Notifier.cc
vpax: Oooh much better. I've included this in my latest push. (I kept the new `Notifier` API for a `Receiver` to report "hey I'm all done" as it seems reasonable to have that in any case.)

I'm finished with new changes, barring other points to consider. Thanks for the comments!

vpax: (oh and ASAN confirmed that that change is fine)

awelzel: > (oh and ASAN confirmed that that change is fine)

The ASAN test failed in CI, actually. Let me comment on one `Unref()` where I'm not sure where it went :-)

src/Trigger.cc
```diff
@@ -373,10 +373,7 @@ void Trigger::Register(Val* val) {
 void Trigger::UnregisterAll() {
     DBG_LOG(DBG_NOTIFIERS, "%s: unregistering all", Name());
 
-    for ( const auto& o : objs ) {
-        notifier::detail::registry.Unregister(o.second, this);
-        Unref(o.first);
```
awelzel, line 378: This `Unref()` has been lost? Wouldn't mind sticking with the old variant, given all the comments how the Registry isn't doing reference management.

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -989,7 +989,8 @@ bool GlueCompiler::compile() {
     if ( ! _doc_id.empty() ) {
         auto mtime = hilti::expression::Ctor(hilti::ctor::Time(hilti::rt::time::current_time()));
         preinit_body.addCall("zeek_rt::register_spicy_module_begin",
-                             {hilti::builder::string(_doc_id), hilti::builder::string(_doc_description), mtime});
+                             {hilti::builder::string_mut(_doc_id), hilti::builder::string_mut(_doc_description),
```
awelzel, line 992: Bit unsure why the spicy changes ended up here, but can probably straightened out during a final rebase or merge.

src/Trigger.cc
vpax: Ah - lost in the switch to the other API. I'll revert (and remove the API, since this shows that it's problematic).

I'm guessing CI ASAN found this via leaks as opposed to memory problems? Alas ASAN on my Mac doesn't do leaks ...

src/spicy/spicyz/glue-compiler.cc
vpax: Just pushed fixups for the `Trigger` stuff, with luck it's now sound.

The spicy changes may have been a cut&paste glitch on copying a commit hash. If not, then I Have No Idea

src/Trigger.cc
awelzel: > I'm guessing CI ASAN found this via leaks as opposed to memory problems?

Ah yes, those were all LeakSanitizer report.

src/script_opt/FuncInfo.h
```diff
@@ -1,17 +1,24 @@
...
+extern bool is_special_script_func(std::string func_name);
+
+// An idempotent function returns the same value when called with the
+// same arguments (and has no meaningful side effects in terms of script-level
+// or Zeek-internal state).
+extern bool is_idempotent(std::string func_name);
+
+// Whether the given function (currently, just BiFs) has no Zeek-script-level
+// side effects.
+extern bool has_no_script_side_effects(std::string func_name);
```
awelzel, line 22: Usuing `const std::string& func_name` or `std::string_view` would avoid a copy of a string. Probably irrelevant where this is used, but thought I'd mention it.

If the source is `const char*`, then yes, this works, though I believe `std::map<std::string, ...>` can be made (or does) support `const char*` lookups transparently, too.

src/script_opt/FuncInfo.cc
```diff
@@ -6,461 +6,507 @@
...
+// we use a bitmask-style approach so we can accommodate future attributes
+// that might overlap.
+
+// BiF Functions that are not listed are assumed to have Unknown side effects.
+// (These are described in comments after the table definition.)  Script
+// functions that are not listed as assumed to not be "special", i.e. known
+// to the event engine.
+
+// Does not change script-level state (though may change internal state).
+#define ATTR_NO_SCRIPT_SIDE_EFFECTS 0x1
```
awelzel, line 23: Nothing to address and it's certainly subjective, but positive terms are generally easier to reason about and help avoiding double negatives.

```
if ( ! has_no_script_side_effects(b->Name()) )
            return false;
```
For someone just skimming this and not immersed, it's mind-twisting :-)

src/script_opt/FuncInfo.h
vpax: thanks for flagging, I've noted this for next month's maintenance

src/script_opt/FuncInfo.cc
vpax: likewise, noted this for next month's maintenance

PR #3499, opened 2023-12-13
SMTP: Add BDAT support

### LINKED ISSUE #3264: SMTP BDAT support
As per discussion over in #2487 we should fix Zeek's inability to deal with `BDAT` verbs. There are several options, outlined in the discussion, starting from minimally just skipping these instructions (though I'd hope we can do better).

I'm marking this a bug because SMTP parsing of `BDAT` sessions tends to derail pretty badly.

I'm putting this on the roadmap for 6.2 — thoughts welcome.

---

https://datatracker.ietf.org/doc/html/rfc3030

awelzel:
Closes #3264


awelzel:
For the BINARYMIME part of RFC3030, I'm interpreting it as an "ask" to the receiver to not transform anything within the BDAT chunks, but it should still be MIME messages within it, so wouldn't need to treat it differently.

>    It is especially important when using BINARYMIME to ensure that the
   MIME message itself is properly formed.

src/analyzer/protocol/smtp/BDAT.cc
```diff
@@ -0,0 +1,301 @@
...
+    DBG_LOG(DBG_ANALYZER, "BDAT: NextChunk size=%" PRIi64 " last=%d", chunk_size, chunk_type == ChunkType::Last);
+    assert(remaining_chunk_size == 0);
+    cur_chunk_type = chunk_type;
+    remaining_chunk_size = chunk_size;
+}
+
+void SMTP_BDAT_Analyzer::DeliverStream(int len, const u_char* data, bool is_orig) {
+    analyzer::Analyzer::DeliverStream(len, data, is_orig);
+    assert(mail != nullptr);
+    assert(len <= RemainingChunkSize());
```
0xxon, line 47: Is it possible to create a case where this triggers?

If yes - this should probably raise a weird instead.

0xxon (approved):
Looks good to me.

src/analyzer/protocol/smtp/BDAT.cc
awelzel: Made it a weird and truncated len instead. It shouldn't trigger given the `SMTP_Analyzer` usage, but remember being uneasy having the `assert` back when I added it :-)

Thanks!

PR #3501, opened 2023-12-13
rule-parse: Remove [event_name] syntax, deprecate msg as identifier

awelzel:
As suggested by Robin. Thanks.


NEWS
```diff
@@ -73,32 +73,31 @@ New Functionality
   to raise instead of ``signature_match()``. This can be more efficient in certain
   scenarios compared to funneling every match through a single event.
 
-  The new syntax is to put the name of the event in brackets before the string
-  or identifier used as message. As an extension, it is possible to only provide
-  the bracketed event name. In this case, the framework expects the event's
-  parameters to consist of only state and data as follows:
+  The new syntax is to put the name of the event before the string used for the
+  ``msg`` argument. As an extension, it is possible to only provide an event name.
```
rsmmr, line 77:
```suggestion
  ``msg`` argument. As an extension, it is possible to only provide an event name, skipping ``msg``.
```

NEWS
```diff
@@ -73,32 +73,31 @@ New Functionality
...
-		event [found_root]
+		event found_root
 	}
 
 	event found_root(state: signature_state, data: string) { }
 
-  Passing an additional message parameter to a custom event is possible with the
-  following syntax. The custom event's parameters need to align with those for the
-  ``signature_match()` event:
+  Using the msg parameter with a custom event looks as follows. The custom event's
```
rsmmr, line 88:
```suggestion
  Using the ``msg`` parameter with a custom event looks as follows. The custom event's
```

NEWS
```diff
@@ -73,32 +73,31 @@ New Functionality
...
 	}
 
 	event found_root_with_msg(state: signature_state, msg: string, data: string) { }
 
-  The message can also be specified as a Zeek side identifier, in which case
-  its initial value will be passed to the custom events. This is identical
-  to the behavior with the default ``signature_match()`` event.
+  Note, message can currently be specified as a Zeek identifier, which is disambiguated
+  behind the scenes for the first variant. Specifying ``msg`` as a Zeek identifier
+  has further been deprecated.
```
rsmmr, line 100:
```suggestion
  Note, the message argument can currently still be specified as a Zeek identifier refering to a script-level string value, which is disambiguated
  behind the scenes for the first variant. Specifying ``msg`` as a Zeek identifier
  has been deprecated with the new event support and will be removed in the future. 
```

rsmmr (approved):
Thanks. :-)

PR #3502, opened 2023-12-14
CI updates

ckreibich:
I'm taking the liberty to remove a few distros that will EOL before the end of the year, since we won't be changing that much before 12/31. This includes FreeBSD 12, see [here](https://www.freebsd.org/security/#sup).

I'm not yet going near the headache distros that we've considered dropping since they will EOL by the time Zeek 7 comes out. I'm thinking we could do that once 6.2 is out, but thoughts are welcome.


ckreibich:
I have changes for Spicy too, but its CI stance is sufficiently different from Zeek that I'm usually not sure how far to go in the tweaking when I make a CI pass. @rsmmr, @bbannier, feel free to let me know — happy to put up a PR there too while I'm at it.

.cirrus.yml
```diff
@@ -209,7 +209,7 @@ centos7_task:
 
 debian12_task:
   container:
-    # Debian 12 (bookworm) EOL: (not yet released)
+    # Debian 12 (bookworm) EOL: TBD
     dockerfile: ci/debian-12/Dockerfile
```
awelzel, line 213: Seems reasonable to remove SKIP_TASK_ON_PR and add it to debian11 (and 10?)

awelzel: Could also see bumping the arm_debian11 build to debian12 instead.

awelzel (approved):

awelzel:
Given the broker change, Could see bumping ubuntu23 to from 23.04 to 23.10.

.cirrus.yml
ckreibich: Thanks, done.

PR #3506, opened 2023-12-14
ldap: Use scalar values in logs where appropriate

### LINKED ISSUE #3504: ldap.log/ldap_search.log: Use scalar values where "more" appropriate
The following fields within the ldap logs are represented as sets/vectors where in practice some only ever hold a single value (MessageInfo$results or opcodes).

MessageInfo:
```
    # normalized operations (e.g., bind_request and bind_response to "bind")
    opcodes: set[string] &log &optional;

    # Result code(s)
    results: set[string] &log &optional;

    # result diagnostic message(s)
    diagnostic_messages: vector of string &log &optional;

    # object(s)
    objects: vector of string &log &optional;

    # argument(s)
    arguments: vector of string &log &optional;
```

SearchInfo:
```
    # sets of search scope and deref alias
    scopes: set[string] &log &optional;
    derefs: set[string] &log &optional;

    # base search objects
    base_objects: vector of string &log &optional;

    # number of results returned
    result_count: count &log &optional;

    # Result code (s)
    results: set[string] &log &optional;

    # result diagnostic message(s)
    diagnostic_messages: vector of string &log &optional;

    #  a string representation of the search filter used in the query
    filter: string &log &optional;

    # a list of attributes that were returned in the search
    attributes: vector of string &log &optional;
```

Change as many of these over to scalar strings to remove some burden from downstream systems and consumers to deal with arrays in the resulting log.

awelzel:
Skimming through the RFC, the previous approach of having containers for most
fields seems unfounded for normal protocol operation. The new weirds could just
as well be considered protocol violations. Outside of duplicated or missed data
they just shouldn't happen for well-behaved client/server behavior.
Additionally, with non-conformant traffic it would be trivial to cause
unbounded state growth and immense log record sizes.

Unfortunately, things have become a bit clunky now.

Closes #3504

---

zeek-testing-private PR: https://github.com/zeek/zeek-testing-private/pull/19


scripts/base/protocols/ldap/main.zeek
```diff
@@ -186,75 +187,107 @@ event LDAP::message(c: connection,
...
 
     if ( result != LDAP::ResultCode_Undef ) {
-      if ( ! searches?$results )
-        searches$results = set();
-      add searches$results[RESULT_CODES[result]];
+      local sresult_str = RESULT_CODES[result];
+      if ( sm?$result && sm$result != sresult_str ) {
+        Reporter::conn_weird("LDAP_search_result_change", c,
+                             fmt("%s: %s -> %s", message_id, sm$result, sresult_str), "LDAP");
+        sm$result = cat_sep("|", "", sm$result, sresult_str);
```
awelzel, line 204: Hmm, after writing the NEWS entry, I'm now strongly leaning towards just overwrite the existing value. The concatenation approach directly translates into unbounded state growth and potentially the creation of immense log entries (something which the existing logic suffers from as well).

awelzel:
@bbannier - would you be willing to look and possibly sign off?

0xxon:
This looks good to me in principle. But - I want to raise that while this makes things better for new users of the protocol, this obviously also is a bunch of renaming that will probably impact people who already are using the ldap analyzer.

I am not opposed to all - I just wonder if it might be worthwhile asking people in Slack, or on our forum if they are using this - and how much this change would mess with their existing pipelines.

awelzel:
> I am not opposed to all - I just wonder if it might be worthwhile asking people in Slack, or on our forum if they are using this - and how much this change would mess with their existing pipelines.

Agree. There's been some internal communication that this should be the right move, but yes, nothing public.

Instead of asking, I'm tempted to just announce it. We've introduced the LDAP analyzer with Zeek 6.1. It's never been in an LTS release before and we actually have the following snippet in NEWS:

```
This analyzer is enabled by default. The
analyzer's events and its ``ldap.log`` and ``ldap_search.log`` should be      
considered preliminary and experimental until the arrival of Zeek's next      
long-term-stable release (7.0).
```

0xxon:
Oh, cool. I missed that it never was in a non-LTS release. With that - I am on board with this :)

0xxon (approved):

PR #3507, opened 2023-12-14
minor search-ability improvement

stevesmoot:
I looked and looked and finally asked on slack, so adding keywords...


timwoj (approved):

timwoj:
Updated spicy docs as part of the merge.

PR #3508, opened 2023-12-14
Some streamlining for constructing TableVal's

vpax:
Currently when a `TableVal` is constructed it creates a corresponding `CompHash` (and deletes it when destructed), even though that value is actually static for the associated `TableType`. This PR shifts its construction to the `TableType`, and also does some additional minor compute-it-once-because-it's-static tweaks.

On a good-sized DNS PCAP (500K requests), _hyperfine_ finds this shaves 2% off of execution time.

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -1568,7 +1568,6 @@ void TableVal::Init(TableTypePtr t, bool ordered) {
     if ( table_type->IsPatternIndex() )
         pattern_matcher = std::make_unique<detail::TablePatternMatcher>(this, table_type->Yield());
 
-    table_hash = new detail::CompositeHash(table_type->GetIndices());
     if ( ordered )
         table_val = new PDict<TableEntryVal>(DictOrder::ORDERED);
     else
@@ -1581,7 +1580,6 @@ TableVal::~TableVal() {
     if ( timer )
         detail::timer_mgr->Cancel(timer);
 
-    delete table_hash;
     delete table_val;
     delete expire_iterator;
 }
@@ -1756,7 +1754,7 @@ bool TableVal::AddTo(Val* val, bool is_first_init, bool propagate_ops) const {
         auto* v = tble.value;
 
         if ( is_first_init && t->AsTable()->Lookup(k.get()) ) {
-            auto key = table_hash->RecoverVals(*k);
+            auto key = GetTableHash()->RecoverVals(*k);
             // ### Shouldn't complain if their values are equal.
             key->Warn("multiple initializations for index");
             continue;
@@ -2081,7 +2079,7 @@ bool TableVal::UpdateTimestamp(Val* index) {
     return true;
 }
 
-ListValPtr TableVal::RecreateIndex(const detail::HashKey& k) const { return table_hash->RecoverVals(k); }
+ListValPtr TableVal::RecreateIndex(const detail::HashKey& k) const { return GetTableHash()->RecoverVals(k); }
 
 void TableVal::CallChangeFunc(const ValPtr& index, const ValPtr& old_value, OnChangeType tpe) {
     if ( ! change_func || ! index || in_change_func )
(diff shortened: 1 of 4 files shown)


src/Val.cc
```diff
@@ -2717,19 +2715,25 @@ unsigned int TableVal::ComputeFootprint(std::unordered_set<const Val*>* analyzed
...
     if ( it == parse_time_table_record_dependencies.end() )
-        return;
+        return affected_table_types;
 
     auto& table_vals = it->second;
 
-    for ( auto& tv : table_vals )
+    for ( auto& tv : table_vals ) {
         parse_time_table_states[tv.get()] = tv->DumpTableState();
+        affected_table_types.insert(tv->table_type.get());
```
awelzel, line 2733: Matter of taste I guess, but I'd have tended to leave `SaveParseTimeTableSate()` and `RecordType::AddFields()` unmodified and add an extra loop in `RebuildParseTimeTables()`:
```
 
 void TableVal::RebuildParseTimeTables() {
+    std::set<TableType*> affected_table_types;
+    for ( auto& [tv, _] : parse_time_table_states )
+        affected_table_types.insert(tv->table_type.get());
+
+    for ( auto* tt : affected_table_types )
+        tt->RegenerateHash();
+
     for ( auto& [tv, ptts] : parse_time_table_states )
         tv->RebuildTable(std::move(ptts));
```

Don't think runtime matters at this stage and easier to justify getting away without comments what is returned and why there's the hash regeneration in the middle of `AddFields()` now ;-)

awelzel (approved):
Nice! The zeek-benchmarker doesn't show all that much, but for this one I guess we'd need a microbenchmark that creates a few thousand or million tables in a tight loop.

Left one thought and I'd switch to a unique_ptr for table_hash prior to merging unless you'd want to do that. Main quirk I ran into trying that was that it requires an empty constructor in `Type.cc`, otherwise the compiler tries to create it in `Type.h` where `CompositeHash` isn't a complete type.

vpax:
> The zeek-benchmarker doesn't show all that much, but for this one I guess we'd need a microbenchmark that creates a few thousand or million tables in a tight loop.

I'm curious, what's the workload for the benchmarker? The numbers I measured were just the stock scripts running on the DNS PCAP. Profiling showed that quite a bit of time was spent in table construction - in particular, in the `set(hk)` constructor in `register_removal_hook()` - so the gains appeared w/o a specific stressing of creating tables.

vpax:
> I'd switch to a unique_ptr for table_hash prior to merging

Cool, please go ahead and do so. I tried making it a unique_ptr but couldn't figure out how to get the compiler happy without including headers (which ultimately led to dependency loops and other headaches). Would like to learn the pattern for this!

src/Val.cc
vpax: > Matter of taste I guess, but I'd have tended to leave `SaveParseTimeTableSate()` and `RecordType::AddFields()` unmodified and add an extra loop in `RebuildParseTimeTables()`

I'm completely happy for you to restructure that too if you wish!

awelzel:
> > The zeek-benchmarker doesn't show all that much, but for this one I guess we'd need a microbenchmark that creates a few thousand or million tables in a tight loop.
> 
> I'm curious, what's the workload for the benchmarker?

It runs `zeek -r` with the stock scripts for a number of pcaps, including a DNS only one (though only ~85k dns.log entries). It also runs a number of microbenchmark scripts to completion. Comparison is against the most recent master build. This all runs on a non-dedicated EC2, so I'm not surprised small gains go under in the noise, particularly when logging threads are involved etc.

I've added 3 new microbenchmarks and there's definitely an improvement.
https://github.com/zeek/zeek-benchmarker/pull/18

(micro-table-ops-create, micro-record-ops-connection-create - green is master, yellow the PR)

image


Primarily this is useful to avoid introducing striking regressions, I wouldn't necessarily recommend looking there for tiny improvements. `hyperfine` locally likely more useful.

PR #3511, opened 2023-12-15
smb: Fix &read_expire not in effect due to &default=string_set() usage

awelzel:
The SMB::State$recent_files field is meant to have expiring entries. However, due to usage of &default=string_set(), the &read_expire attribute is not respected causing unbounded state growth. Simply remove &default altogether.

Thanks to ya-sato on Slack for reporting!

Likely related: zeek/zeek-docs#179

### DIFF
--- a/scripts/base/protocols/smb/main.zeek
+++ b/scripts/base/protocols/smb/main.zeek
@@ -157,7 +157,7 @@ export {
 		## A set of recent files to avoid logging the same
 		## files over and over in the smb files log.
 		## This only applies to files seen in a single connection.
-		recent_files : set[string] &default=string_set() &read_expire=3min;
+		recent_files : set[string] &default=set() &read_expire=3min;
 	};
 
 	## Everything below here is used internally in the SMB scripts.


vpax:
BTW, the changes here remind me that there are a number of places in the base scripts where there are unnecessary `&default`s or initializations. Things like: `local v: vector = vector()`, and the equivalent for some container record fields.

awelzel:
> BTW, the changes here remind me that there are a number of places in the base scripts where there are unnecessary `&default`s or initializations. Things like: `local v: vector = vector()`, and the equivalent for some container record fields.

The SMB state is even more curious. The other fields are all marked as `&optional`, but where State is created, there's the following, making the &optional somewhat questionable.

https://github.com/zeek/zeek/blob/d1d9b9a1be9e7d2f16ca33f6d6d932ed3f2fbd87/scripts/base/protocols/smb/smb2-main.zeek#L18-L26

---

However, this PR is meant as a single-line backportable 6.0 fix as it was user reported, so not too open of doing anything else but the least invasive fix that makes `&read_expire` actually function.

0xxon (approved):

PR #3514, opened 2023-12-18
segment_profiling: Remove SegmentProfiler and load_sample event

awelzel:
While it seems interesting functionality, this hasn't been documented,
maintained or knowingly leveraged for many years.

There are various other approaches today, too:

* We track the number of event handler invocations regardless of
  profiling. It's possible to approximate a load_sample event by
  comparing the result of two get_event_stats() calls. Or, visualize
  the corresponding counters in a Prometheus setup to get an idea of
  event/s broken down by event names.

* HookCallFunction() allows to intercept script execution, including
  measuring the time execution takes.

* The global call_stack and g_frame_stack can be used from plugins
  (and even external processes) to walk the Zeek script stack at certain
  points to implement a sampling profiler.

* USDT probes or more plugin hooks will likely be preferred over Zeek
  builtin functionality in the future.

Relates to #3458


NEWS
```diff
@@ -126,6 +129,9 @@ Changed Functionality
 Removed Functionality
 ---------------------
 
+- The "segment_profiling" functionality and ``load_sample`` event have been removed
```
timwoj, line 132: This doesn't need to be duplicated. Adding it to the Breaking Changes section is enough.

timwoj (approved):
I have no qualms with removing this. Like you said it hasn't been maintained in a while, and if it's causing known performance issues it'd be ok to go.

PR #3487, opened 2023-12-04
Integrate Broker's new variant type

Neverlord:
WIP; don't merge yet. Filing for CI coverage.

PR #3519, opened 2023-12-22
Prepare Broker manager for broker::variant

Neverlord:
These are basically the changes from #3487 minus the actual switch to `broker::variant`. The idea is to keep Zeek compatible with both versions of Broker for now (with and without `variant`) to allow us to switch back (just in case we need to).

- With `broker::data`, we always have actual `std::string` objects that we can pass to C functions expecting a null-terminated string. However, `broker::variant` will return a `std::string_view` where we have previously received a `std::string`. Hence, we add an extra level of indirection that ensures that views are converted to null-terminated strings and also use `c_str()` where we have previously used `data()`. The former is not present on a `std::string_view`. Using this member function instead acts as an extra level of insurance that we do not accidentally pass the bytes from a view to a C function.
- Switch from error and status views to actual error and status objects. The view types from Broker only work with `broker::data` and thus won't be available with `broker::variant`.

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -1,11 +1,14 @@
 #include "zeek/broker/Manager.h"
 
 #include <broker/broker.hh>
+#include <broker/config.hh>
 #include <broker/configuration.hh>
 #include <broker/zeek.hh>
 #include <unistd.h>
 #include <cstdio>
 #include <cstring>
+#include <string>
+#include <string_view>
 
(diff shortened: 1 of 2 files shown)


ckreibich (approved):
Very sorry for the delay Dominik! LGTM. And thank you for the commit message. :-)

ckreibich:
I'm just rebasing real quick (no conflicts) and will then merge.

PR #3530, opened 2024-01-05
ldap: Fix substring filter parsing and rendering

awelzel:
The initial (prefix) and final (suffix) strings are specified individually with a variable number of "any" matches that can occur between these. The previous implementation assumed a single string and rendered it as *<string>*.

Reported and PCAP provided by @martinvanhensbergen, thanks!

Closes zeek/spicy-ldap#27


bbannier (approved):
Thanks!

Like discussed offline, it would be great to backport this patch to zeek/spicy-ldap while we still maintain it (probably until the zeek-7.0 timeframe).

PR #3533, opened 2024-01-06
Backward compatibility for OpaqueVal serialization

Neverlord:
External plugins depend on the API for `OpaqueVal`. This set of changes brings back the previous signature for the `Serialize` and `Unserialize` member functions. The new set of functions that operate on the recently added `BrokerData` API were renamed accordingly and use a `Data` suffix to distinguish between the old and new interface.

For the transition period, `OpaqueVal` now has two "sets" of serialization functions: old and new (using the suffix). By default, the new functions call the old API and then convert to the new types. Hence, plugins that override the old set of member functions will continue to work. New code should only override the new set of functions.

Since the macro `DECLARE_OPAQUE_VALUE` (a convenience macro for adding a default set of member functions to a subtype of `OpaqueVal`) might be used by 3rd parties, the macro has been "restored" to its previous behavior, i.e., it will override the old set of member functions. The new macro `DECLARE_OPAQUE_VALUE_V2` is similar but overrides the new set of functions instead.

The class `BloomFilter` uses the same member function signatures as `OpaqueVal` for serialization. Hence, the same old/new split was implemented to keep the APIs consistent.

### DIFF
--- a/src/OpaqueVal.h
+++ b/src/OpaqueVal.h
@@ -94,7 +94,7 @@ private:
 /**
  * Legacy macro to insert into an OpaqueVal-derived class's declaration. Overrides the "old" serialization methods
  * DoSerialize and DoUnserialize.
- * @deprecated Use DECLARE_OPAQUE_VALUE_DATA instead.
+ * @deprecated Use DECLARE_OPAQUE_VALUE_DATA instead. Remove in v7.1.
  */
 #define DECLARE_OPAQUE_VALUE(T)                                                                                        \
     friend class zeek::OpaqueMgr::Register<T>;                                                                         \
@@ -169,7 +169,7 @@ protected:
     friend class OpaqueMgr;
 
     /**
-     * @deprecated Override DoSerializeData instead.
+     * @deprecated Override DoSerializeData instead. Remove in v7.1.
      */
     virtual broker::expected<broker::data> DoSerialize() const;
 
@@ -183,7 +183,7 @@ protected:
     virtual std::optional<BrokerData> DoSerializeData() const;
 
     /**
-     * @deprecated Override DoUnserializeData instead.
+     * @deprecated Override DoUnserializeData instead. Remove in v7.1.
      */
     virtual bool DoUnserialize(const broker::data& data);
 
--- a/src/OpaqueVal.h
+++ b/src/OpaqueVal.h
@@ -111,8 +111,8 @@ private:
 #define DECLARE_OPAQUE_VALUE_V2(T)                                                                                     \
     friend class zeek::OpaqueMgr::Register<T>;                                                                         \
     friend zeek::IntrusivePtr<T> zeek::make_intrusive<T>();                                                            \
-    std::optional<zeek::BrokerData> DoSerializeData() const override;                                                      \
-    bool DoUnserializeData(zeek::BrokerDataView data) override;                                                            \
+    std::optional<zeek::BrokerData> DoSerializeData() const override;                                                  \
+    bool DoUnserializeData(zeek::BrokerDataView data) override;                                                        \
     const char* OpaqueName() const override { return #T; }                                                             \
     static zeek::OpaqueValPtr OpaqueInstantiate() { return zeek::make_intrusive<T>(); }
 
@@ -143,7 +143,7 @@ public:
     /**
      * @copydoc Serialize
      */
-     std::optional<BrokerData> SerializeData() const;
+    std::optional<BrokerData> SerializeData() const;
 
     /**
      * Reinstantiates a value from its serialized Broker representation.
--- a/src/OpaqueVal.h
+++ b/src/OpaqueVal.h
@@ -6,6 +6,7 @@
 #include <unistd.h>
 #endif
 
+#include <broker/expected.hh>
 #include <paraglob/paraglob.h>
 #include <sys/types.h> // for u_char
 #include <optional>
@@ -17,6 +18,10 @@
 #include "zeek/telemetry/Counter.h"
 #include "zeek/telemetry/Gauge.h"
 #include "zeek/telemetry/Histogram.h"
+
+namespace broker {
+class data;
+}
 
 namespace zeek {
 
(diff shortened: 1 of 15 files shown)


src/OpaqueVal.h
```diff
@@ -117,41 +138,61 @@ class OpaqueVal : public Val {
...
+    static OpaqueValPtr UnserializeData(BrokerListView data);
 
 protected:
     friend class Val;
     friend class OpaqueMgr;
 
+    /**
+     * @deprecated Override DoSerializeData instead.
+     */
+    virtual broker::expected<broker::data> DoSerialize() const;
```
awelzel, line 173: If this are marked with `[[deprecated]]`, will external plugins receive a deprecation warning when overriding it? If yes, we should do this and suppress the warnings for calling `DoSerialize()` in the default `DoSerializeData()` implementation in-tree (same for `DoUnserialize()`).

The doxygen `@deprecated` won't yield a compiler warning and so the casual user won't be notified explicitly.

src/OpaqueVal.h
```diff
@@ -86,12 +91,28 @@ class OpaqueMgr {
...
+    broker::expected<broker::data> DoSerialize() const override;                                                       \
+    bool DoUnserialize(const broker::data& data) override;                                                             \
+    const char* OpaqueName() const override { return #T; }                                                             \
+    static zeek::OpaqueValPtr OpaqueInstantiate() { return zeek::make_intrusive<T>(); }
+
+/**
+ * Macro to insert into an OpaqueVal-derived class's declaration. Overrides the "new" serialization methods
+ * DoSerializeData and DoUnserializeData.
+ */
+#define DECLARE_OPAQUE_VALUE_V2(T)                                                                                     \
```
awelzel, line 111: `_DATA` instead of `_V2`? Neither overly meaningful, but the former would be somewhat closer/related.

src/OpaqueVal.h
```diff
@@ -117,41 +138,61 @@ class OpaqueVal : public Val {
      * @return the broker representation, or an error if serialization
      * isn't supported or failed.
      */
-    std::optional<BrokerData> Serialize() const;
+    [[deprecated("use SerializeData instead")]] broker::expected<broker::data> Serialize() const;
```
awelzel, line 141:
```suggestion
    [[deprecated("Remove in v7.1: use SerializeData instead")]] broker::expected<broker::data> Serialize() const;
```

src/OpaqueVal.h
```diff
@@ -117,41 +138,61 @@ class OpaqueVal : public Val {
...
+     std::optional<BrokerData> SerializeData() const;
 
     /**
      * Reinstantiates a value from its serialized Broker representation.
      *
      * @param data Broker representation as returned by *Serialize()*.
      * @return unserialized instances with reference count at +1
      */
-    static OpaqueValPtr Unserialize(BrokerDataView data);
+    [[deprecated("use UnserializeData instead")]] static OpaqueValPtr Unserialize(const broker::data& data);
```
awelzel, line 154:
```suggestion
    [[deprecated("Remove in v7.1: use UnserializeData instead")]] static OpaqueValPtr Unserialize(const broker::data& data);
```

src/probabilistic/BloomFilter.h
```diff
@@ -105,8 +105,11 @@ class BloomFilter {
      */
     virtual std::string InternalState() const = 0;
 
-    std::optional<BrokerData> Serialize() const;
-    static std::unique_ptr<BloomFilter> Unserialize(BrokerDataView data);
+    [[deprecated("use SerializeData instead")]] broker::expected<broker::data> Serialize() const;
```
awelzel, line 108:
```suggestion
    [[deprecated("Remove in v7.1: use SerializeData instead")]] broker::expected<broker::data> Serialize() const;
```

src/probabilistic/BloomFilter.h
```diff
@@ -105,8 +105,11 @@ class BloomFilter {
...
+    [[deprecated("use UnserializeData instead")]] static std::unique_ptr<BloomFilter> Unserialize(
```
awelzel, line 109:
```suggestion
    [[deprecated("Remove in v7.1: use UnserializeData instead")]] static std::unique_ptr<BloomFilter> Unserialize(
```

src/OpaqueVal.h
Neverlord: I had the same thought, but it seems like there's no way to deprecate overriding a virtual member function: https://stackoverflow.com/questions/49011990/c-deprecation-warning-when-overriding-a-deprecated-virtual-method. At least I don't know how we could get a warning for that. 🙁

I've added the `deprecated` Doxygen tag here to have at least *some* in-source documentation on that. Happy to hear better ideas.

awelzel: Oh, this is sad. For someone not reading NEWS this will pretty much end-up having the same effect as just breaking the interface, but I guess we can make the argument that we've documented it somewhere.

The other ideas I'm coming up revolve around adding some explicit info to derived class that they support the BrokerData serialization and minimally we could warn-once at runtime for those that don't. Probably not worth it.

src/OpaqueVal.h
```diff
@@ -117,41 +138,62 @@ class OpaqueVal : public Val {
...
      */
-    static OpaqueValPtr Unserialize(BrokerListView data);
+    static OpaqueValPtr UnserializeData(BrokerListView data);
 
 protected:
     friend class Val;
     friend class OpaqueMgr;
 
+    /**
+     * @deprecated Override DoSerializeData instead.
```
awelzel, line 172: Could you also add the "Remove in v7.1" to all the `@deprecated` ones? I think Tim is grepping for that when doing the removal passes.

awelzel (approved):
@Neverlord - mind adding a NEWS entry under Deprecations for this? Possibly emphasize the fact that there's no compiler deprecation warning produced.

src/OpaqueVal.h
awelzel: Suppose introducing a complete new `OpaqueVal` subclass (overridding the deprecated methods with final) and deprecating the current `OpaqueVal` could be a way out, but that would imply another name change and might be just too much.

Neverlord:
> @Neverlord - mind adding a NEWS entry under Deprecations for this? Possibly emphasize the fact that there's no compiler deprecation warning produced.

@awelzel I've added an entry under breaking changes, even though there's "technically" a deprecation cycle. But since there's not going to be a warning for it, I figured it might be more appropriate to document it this way. Let me know if you think otherwise.

PR #3536, opened 2024-01-08
parse: Support @if conditionals in record definitions

### LINKED ISSUE #3531: @if directives don't work in record definitions
The following generates a syntax error:
```
type r: record {
	a: count;
@if ( ! Version::at_least("6.1.0") )
	b: double;
@endif
};
```
This is unfortunate as I'm trying to update a script to work with two different versions of Zeek. (This is in the context of the change to `files.log` removing some record fields.) The problem is quite akin to #2075 which required a parser tweak to support.

Also, AFAICT the documentation doesn't indicate any restrictions on where `@if`s can appear ...

awelzel:
Closes #3531


awelzel:
@vpax - mind looking this over? I'm not 100% sure about use-cases that you ran into (as per DM), but seems reasonable to allow this.

vpax (approved):
Nice! It all looks good, no comments. I especially enjoyed the `Even more strange` stress-test :P

vpax:
@bbannier yes I think there's good agreement on that as preferred style. But supporting large (script) code bases across multiple versions can sometimes benefit from this sort of handy ugliness

PR #3541, opened 2024-01-09
Known: Keep &create_expire on local tables/sets valid

### LINKED ISSUE #3540: Known::hosts &create_expire attribute removed during node_up/node_down handling
Jordan Barnartt reported too few known_hosts entries after upgrading to 6.1.

The following commit switched to using the non-broker-store implementation by default with 6.0 to avoid performance issues with the broker store implementation: f6d3ed2395c29b0fbf5e7647c32ef09d7e52c1f7

However, this implementation is buggy in that `node_up` and `node_down` reset the `Known::hosts` to a set that has no `&create_expire` attribute anymore and "log all hosts every 24h" logic stops working.

https://github.com/zeek/zeek/blob/f1e94594f91bce2ac181bba043c92586be1f1fd4/scripts/policy/protocols/conn/known-hosts.zeek#L60

https://github.com/zeek/zeek/blob/f1e94594f91bce2ac181bba043c92586be1f1fd4/scripts/policy/protocols/conn/known-hosts.zeek#L124

We have this documented as a gotcha, but I start to wonder if it might make sense to runtime warn when expire attributes are removed during an assignment.

Applies to the other Known scripts, too:
https://github.com/zeek/zeek/blob/f1e94594f91bce2ac181bba043c92586be1f1fd4/scripts/policy/protocols/conn/known-services.zeek#L190

https://github.com/zeek/zeek/blob/f1e94594f91bce2ac181bba043c92586be1f1fd4/scripts/policy/protocols/ssl/known-certs.zeek#L149

awelzel:
After switching the known scripts away from broker stores, the
&create_expire value of the local tables/sets of the known scripts
wasn't in effect due to Cluster::node_up() and Cluster::node_down()
re-assigning these without keeping the &create_expire attribute
intact. This broke the "log hosts every 24h" behavior.

Closes #3540

### DIFF
--- a/scripts/policy/protocols/conn/known-hosts.zeek
+++ b/scripts/policy/protocols/conn/known-hosts.zeek
@@ -121,7 +121,7 @@ event Cluster::node_up(name: string, id: string)
 		return;
 
 	# Drop local suppression cache on workers to force HRW key repartitioning.
-	Known::hosts = set();
+	clear_table(Known::hosts);
 	}
 
 event Cluster::node_down(name: string, id: string)
@@ -133,7 +133,7 @@ event Cluster::node_down(name: string, id: string)
 		return;
 
 	# Drop local suppression cache on workers to force HRW key repartitioning.
-	Known::hosts = set();
+	clear_table(Known::hosts);
 	}
 
 event Known::host_found(info: HostsInfo)
--- a/scripts/policy/protocols/conn/known-services.zeek
+++ b/scripts/policy/protocols/conn/known-services.zeek
@@ -187,7 +187,7 @@ event Cluster::node_up(name: string, id: string)
 		return;
 
 	# Drop local suppression cache on workers to force HRW key repartitioning.
-	Known::services = table();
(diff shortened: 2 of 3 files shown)


vpax:
@awelzel From a quick read of what you sketch, yes, I imagine we can discard the IDOptInfo. I'm doing my monthly ZAM maintenance now and will add it to the list. If you have a simple reproducer that would be handy, but if not then I can flag when you should try retesting

vpax:
@awelzel please see if this patch does the trick.
[id-opt-info.patch.txt](https://github.com/zeek/zeek/files/13882343/id-opt-info.patch.txt)

awelzel:
> @awelzel please see if this patch does the trick.

It does, I had a similar one locally, just wasn't sure if that broke something in ZAM land :-)

I've updated the changesest here to use `clear_table()` and won't pick it up. If you carry it with your maintenance batch that seems fine.

0xxon (approved):

PR #3546, opened 2024-01-11
Script optimization maintenance for Jan 2024

vpax:
This PR attends to a number of script optimization maintenance issues. Most of these were driven by trying out ZAM optimization on a broader set of scripts than present in the existing BTests and base scripts, so I added regression tests corresponding to a number of those cases. The change that most affects non _script_opt/_ code concerns concretizing empty `vector()` values. This benefits interpreted script execution too, as it enables better run-time type-checking.


src/script_opt/ZAM/BuiltInSupport.h
```diff
@@ -36,12 +36,12 @@ class CatArg {
...
 class FixedCatArg : public CatArg {
 public:
-    FixedCatArg(const TypePtr& t);
+    FixedCatArg(TypePtr t);
 
     void RenderInto(ZVal* zframe, int slot, char*& res) override;
 
 protected:
-    const TypePtr& t;
+    TypePtr t;
```
awelzel, line 44: Seems you could continue to use `FixedCatArg(const TypePtr& t)` as that will be copied, but either works.

scripts/base/protocols/ldap/main.zeek
```diff
@@ -203,7 +203,7 @@ event LDAP::message(c: connection,
     }
 
     if ( diagnostic_message != "" ) {
-      if ( ! sm?$diagnostic_message && sm$diagnostic_message != diagnostic_message ) {
+      if ( sm?$diagnostic_message && sm$diagnostic_message != diagnostic_message ) {
```
awelzel, line 206: Ouch and thanks.

awelzel (commented):
Mostly nits / suggestions.

src/Type.h
```diff
@@ -366,6 +366,15 @@ class IndexType : public Type {
...
         : Type(t), indices(std::move(arg_indices)), yield_type(std::move(arg_yield_type)) {
+        // "indices" might be nil if we're deferring construction of the type
+        // for "-O use-C++" initialization.
+        if ( indices )
+            CheckForSpecialIndices();
+        else
+            is_subnet_index = is_pattern_index = false; // placeholders
+    }
+
+    void CheckForSpecialIndices() {
```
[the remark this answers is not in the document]
vpax: I see your point. I went with your second suggestion.

src/Val.cc
```diff
@@ -1889,44 +1889,56 @@ ValPtr TableVal::Default(const ValPtr& index) {
...
 
         try {
-            return def_val->Clone();
+            result = def_val->Clone();
         } catch ( InterpreterException& e ) { /* Already reported. */
-        }
 
-        Error("&default value for table is not clone-able");
-        return nullptr;
+            if ( ! result ) {
```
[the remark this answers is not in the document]
vpax: Ah, yes, vestigial. Now streamlined.

src/Val.cc
```diff
@@ -1889,44 +1889,56 @@ ValPtr TableVal::Default(const ValPtr& index) {
...
-    if ( ! result ) {
-        Error("no value returned from &default function");
-        return nullptr;
+    auto rt = result->GetType();
+    if ( rt->Tag() == TYPE_VECTOR && rt->Yield()->Tag() == TYPE_ANY ) {
+        auto rv = result->AsVectorVal();
+        auto table_vector_yield = GetType()->Yield()->Yield();
+        if ( rv->Size() == 0 && table_vector_yield && table_vector_yield->Tag() != TYPE_ANY )
+            rv->Concretize(table_vector_yield);
     }
```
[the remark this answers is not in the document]
vpax: Good suggestion, this definitely makes the code more clear.

src/script_opt/ZAM/BuiltInSupport.h
vpax: I like it being a `TypePtr` in the constructor declaration because that way it's more apparent that the object is going to hang onto the pointer. `const TypePtr&` is more suggestive that it just uses it during construction but not later. (Not that the caller should actually care about this distinction, though ...)

vpax:
I've pushed fixups for the threads above. I also added a few more ZAM fixes/improvements that came up in the ZAM evaluation I've been doing in parallel with this work.

src/script_opt/ZAM/BuiltInSupport.h
awelzel: A yes, I like that, thanks!

src/Expr.cc
```diff
@@ -2601,14 +2601,9 @@ VectorValPtr index_slice(VectorVal* vect, int _first, int _last) {
 
     zeek_int_t first = get_slice_index(_first, len);
     zeek_int_t last = get_slice_index(_last, len);
-    zeek_int_t sub_length = last - first;
 
-    if ( sub_length >= 0 ) {
-        result->Resize(sub_length);
```
awelzel, line 2607: I'm going to drop this commit. The pre `Resize()` does allocate the required memory at once while the `Assign()` may reallocate a few times. There's also `Reserve()` which would allow replacing `Assign()` with `Append()`, but more generally wonder if this could work and be even faster when switched to ZVal's.

See also. https://github.com/zeek/zeek/pull/3478#discussion_r1415692021

src/Val.cc
```diff
@@ -1889,45 +1889,54 @@ ValPtr TableVal::Default(const ValPtr& index) {
...
     if ( def_val->GetType()->Tag() != TYPE_FUNC || same_type(def_val->GetType(), GetType()->Yield()) ) {
         if ( def_attr->GetExpr()->IsConst() )
             return def_val;
 
         try {
-            return def_val->Clone();
-        } catch ( InterpreterException& e ) { /* Already reported. */
+            result = def_val->Clone();
+        } catch ( InterpreterException& e ) {
+            Error("&default value for table is not clone-able");
```
awelzel, line 1901: I think my suggestion failed :-) I'll move this one block out and keep the `if ( ! result )` to handle the situation where `->Clone()` returns nil. If I read things right, would result in a segfault on the `result->GetType()` further down now.

awelzel, line 1901: I think my suggestion failed :-) I'll move this one block out and keep the `if ( ! result )` to handle the situation where `->Clone()` returns nil. If I read things right, would result in a segfault on the `result->GetType()` further down now.

awelzel (approved):
Merging with the two comments addressed in the force-pushed version of the branch.

src/Expr.cc
vpax: Thanks, makes sense.

src/Val.cc
vpax: Ah now I understand (I think) what you meant. Appreciate you going ahead and addressing it!

PR #3547, opened 2024-01-11
Fixes to MMDB handling and introduce `mmdb.bif` / `MMDB.{cc,h}`

ckreibich:
I didn't really mean to work on this, but a couple of things happened. First, I noticed that the `core/mmdb/temporary-error.zeek` btest fails when you happen to have Maxmind data installed in one of the fallback locations Zeek considers. I noticed that these are hardwired in the core, which doesn't seem right. As I set out to fix this, I also noticed that the `mmdb_open_location_db()` and `mmdb_open_asn_db()` BiFs are untested, and have a bug: when you open a DB via them, and it turns stale, Zeek silently switches away from them to the script-level configuration for the DBs (via `mmdb_dir` — which may well not be configured). And, I noticed that this is a fair bit of code, including a whole class, all uncommented and free-floating in zeek.bif.

So this PR does a couple of things:
- It moves the MMDB functionality into a minimal `mmdb.bif`, plus `MMDB.{cc.h}`, so the bulk of the code is now formatted. It's not in the detail namespace, so becomes part of the internal API — seems okay to me.
- It moves the additional list of directories in which Zeek looks for MMDB files into the policy layer, making them configurable. I am not sure this is required functionality (spelling out one path for the location and one for the ASN DBs seems sufficient), but for now it's backward-compatible and configurable, and the `temporary-error` test is fixed in the presence of a system-wide MMDB file. The name of the files themselves is now configurable too, no longer insisting on `GeoLite2-City.mmdb` etc.
- It makes any MMDB file identified in the configuration (via the `mmdb_dir` approach or via explicit `mmdb_open_*_db()`) "sticky", meaning that once Zeek has located a DB file, it will stick to it across re-openings. This was easiest to do by moving more state from static globals into the `MMDB` class, which is likely a good thing, but also made the changeset larger.

The commits change things gradually to make it easier to track changes — the code initially moves as-is from the .bif into .cc, and only gets refactored there.

I think there are corner cases this code still isn't super careful about. If a DB file is busted or gone, the code likely tries too hard to keep opening it. (Removal of a file, once successfully loaded, likely shouldn't matter at all.) I decided not to pursue these further for now.

It also looks like Maxmind DB handling isn't in the manual. Seems like a good one for the "Popular customizations" section. I'll link a ticket tomorrow.


src/MMDB.cc
```diff
@@ -0,0 +1,359 @@
...
+
+        if ( OpenFile(d.data()) )
+            return true;
+    }
+
+    const auto& mmdb_dir_fallbacks_val = zeek::detail::global_scope()->Find("mmdb_dir_fallbacks")->GetVal();
+    auto* vv = mmdb_dir_fallbacks_val->AsVectorVal();
+
+    for ( unsigned int i = 0; i < vv->Size(); ++i ) {
+        auto d = std::string(vv->StringAt(i)->CheckString()) + "/" + mmdb_city_db;
```
awelzel, line 211:
```suggestion
        auto d = vv->StringAt(i)->ToStdString() + "/" + mmdb_city_db;
```

src/MMDB.cc
```diff
@@ -0,0 +1,359 @@
...
+        if ( OpenFile(d.data()) )
+            return true;
+    }
+
+    for ( unsigned int i = 0; i < vv->Size(); ++i ) {
+        auto d = std::string(vv->StringAt(i)->CheckString()) + "/" + mmdb_country_db;
```
awelzel, line 217:
```suggestion
        auto d = vv->StringAt(i)->ToStdString() + "/" + mmdb_country_db;
```

src/MMDB.cc
```diff
@@ -0,0 +1,359 @@
...
+
+bool MMDB::OpenFile(const char* a_filename) {
+    filename = a_filename;
+    Close();
+
+    if ( 0 != stat(a_filename, &file_info) ) {
+        return false;
+    }
+
+    int status = MMDB_open(a_filename, MMDB_MODE_MMAP, &mmdb);
```
awelzel, line 88:
```suggestion
    int status = MMDB_open(filename.c_str(), MMDB_MODE_MMAP, &mmdb);
```

Nit, but I was a bit wondering if a_filename is dangling (but must have been in the prior implementation).

awelzel (approved):
Thanks for the cleanup! Guess it would be nice to fix the issue in 6.0, but seems a bit too much and wasn't user or production environment that was affected, so not putting it on the list.

src/MMDB.cc
ckreibich: Yeah. I did a quick test and the library does copy the string. But why risk it — I changed it to `filename`, thanks.

ckreibich:
Thanks Arne, I've pushed a round of updates to modernize. I won't merge just yet so you can look it over, since I've not worked with our "modern" APIs lately.

We may want to strengthen the wording in the [IntrusivePtr wiki page](https://github.com/zeek/zeek/wiki/IntrusivePtr-Overview) to make clear that it's an requirement for things in `zeek::`.

awelzel (approved):

awelzel:
> We may want to strengthen the wording in the [IntrusivePtr wiki page](https://github.com/zeek/zeek/wiki/IntrusivePtr-Overview) to make clear that it's an requirement for things in `zeek::`.

Hmm, mostly a personal hunch to avoid introducing new raw pointer APIs unless there's a good performance argument. There might be a better place for "Zeek public API" considerations, but that page seems fine, too.

PR #3557, opened 2024-01-16
More std::string_view usage in various BIFs

timwoj:
This is a follow-on to https://github.com/zeek/zeek/pull/3554 that switches to using `ToStdStringView()` for more BIFs that were previously calling `ToStdString()`.

In strings.bif:
- The actual work for the `starts_with` and `ends_with` BIFs are moved out into separate utility methods so they can be reused more easily in other BIFs.
- `starts_with` changes to closely match the `ends_with` changes from https://github.com/zeek/zeek/pull/3554.
- `remove_prefix` and `remove_suffix` were both heavily reworked to just call the two new utility methods.
- `count_substr` just had a direct function call change.

In zeek.bif:
- `Scope::Find()` takes a string_view argument so we can just pass that in instead of calling `ToStdString()` and then having it be converted as part of the function call.

Some perf numbers:

- `starts_with`:
```
event zeek_init() {
	local x = 0;
	while (++x < 5000000) {
		starts_with("sdf.fasdfa.asfs.dfas.fsfdsf.com", "abcde");
	}
}

Benchmark 1: ./src/zeek.master -b ~/Desktop/t.zeek
  Time (mean ± σ):     15.907 s ±  0.063 s    [User: 15.687 s, System: 0.203 s]
  Range (min … max):   15.830 s … 16.030 s    10 runs
 
Benchmark 2: ./src/zeek -b ~/Desktop/t.zeek
  Time (mean ± σ):     14.455 s ±  0.033 s    [User: 14.229 s, System: 0.207 s]
  Range (min … max):   14.401 s … 14.525 s    10 runs
 
Summary
  ./src/zeek -b ~/Desktop/t.zeek ran
    1.10 ± 0.01 times faster than ./src/zeek.master -b ~/Desktop/t.zeek
```

- `remove_prefix`:
```
event zeek_init() {
	local x = 0;
	while (++x < 5000000) {
		remove_prefix("abcabcabcabcabcabcabcabcabcabc", "abc");
	}
}

Benchmark 1: ./src/zeek.master -b ~/Desktop/t.zeek
  Time (mean ± σ):     23.462 s ±  0.052 s    [User: 23.243 s, System: 0.201 s]
  Range (min … max):   23.410 s … 23.568 s    10 runs
 
Benchmark 2: ./src/zeek -b ~/Desktop/t.zeek
  Time (mean ± σ):     25.367 s ±  0.032 s    [User: 25.128 s, System: 0.219 s]
  Range (min … max):   25.306 s … 25.425 s    10 runs
 
Summary
  ./src/zeek.master -b ~/Desktop/t.zeek ran
    1.08 ± 0.00 times faster than ./src/zeek -b ~/Desktop/t.zeek
```

- `remove_suffix`:
```
event zeek_init() {
	local x = 0;
	while (++x < 5000000) {
		remove_suffix("abcabcabcabcabcabcabcabcabcabc", "abc");
	}
}

Benchmark 1: ./src/zeek.master -b ~/Desktop/t.zeek
  Time (mean ± σ):     38.084 s ±  0.109 s    [User: 37.797 s, System: 0.265 s]
  Range (min … max):   37.905 s … 38.224 s    10 runs
 
Benchmark 2: ./src/zeek -b ~/Desktop/t.zeek
  Time (mean ± σ):     25.462 s ±  0.098 s    [User: 25.211 s, System: 0.228 s]
  Range (min … max):   25.348 s … 25.652 s    10 runs
 
Summary
  ./src/zeek -b ~/Desktop/t.zeek ran
    1.50 ± 0.01 times faster than ./src/zeek.master -b ~/Desktop/t.zeek
```

### DIFF
--- a/src/strings.bif
+++ b/src/strings.bif
@@ -1355,8 +1355,14 @@ function rfind_str%(str: string, sub: string, start: count &default=0, end: int 
 ##
 function starts_with%(str: string, sub: string%) : bool
 	%{
-	string s = str->ToStdString();
-	return zeek::val_mgr->Bool(s.find(sub->ToStdString()) == 0);
+	if ( sub->Len() > str->Len() )
+		return zeek::val_mgr->Bool(false);
+
+	auto sub_s = sub->ToStdStringView();
+	auto s = str->ToStdStringView();
+	auto start_s = std::string_view{s.data(), sub_s.size()};
+
+	return zeek::val_mgr->Bool(start_s == sub_s);
 	%}
 
 ## Returns whether a string ends with a substring.
--- a/src/strings.bif
+++ b/src/strings.bif
@@ -1247,8 +1247,8 @@ function reverse%(str: string%) : string
 ##
 function count_substr%(str: string, sub: string%) : count
 	%{
-	string s = str->ToStdString();
-	string sub_s = sub->ToStdString();
+	auto s = str->ToStdStringView();
+	auto sub_s = sub->ToStdStringView();
 
 	size_t count = 0;
 	size_t pos = s.find(sub_s);
--- a/src/strings.bif
+++ b/src/strings.bif
@@ -1562,8 +1562,8 @@ function remove_prefix%(str: string, sub: string%) : string
 	// This could just use repeated calls to lstrip(), except for a couple of reasons:
 	// 1) lstrip() creates a StringVal at the end, and that would mean repeated recreation of objects
 	// 2) lstrip() searches for any character in the string, not the string as a whole.
-	string s = str->ToStdString();
-	string sub_s = sub->ToStdString();
+	auto s = str->ToStdStringView();
+	auto sub_s = sub->ToStdStringView();
 
 	size_t pos = s.find(sub_s);
 	if ( pos != 0 )
@@ -1584,11 +1584,14 @@ function remove_prefix%(str: string, sub: string%) : string
 function remove_suffix%(str: string, sub: string%) : string
 	%{
 	// See the note in removeprefix for why this doesn't just call rstrip.
-	string s = str->ToStdString();
-	string sub_s = sub->ToStdString();
+	auto s = str->ToStdStringView();
+	auto sub_s = sub->ToStdStringView();
 
 	size_t pos = s.rfind(sub_s);
 	size_t next_pos = s.size() - sub_s.size();
+
+	if ( pos != next_pos )
+		return zeek::IntrusivePtr<zeek::StringVal>(NewRef{}, str);
 
 	while ( pos == next_pos )
 		{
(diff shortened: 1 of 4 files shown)


awelzel:
For `remove_prefix`, any reason it got slower than master, or some copy/paste issue?
```
Summary
  ./src/zeek.master -b ~/Desktop/t.zeek ran
    1.08 ± 0.00 times faster than ./src/zeek -b ~/Desktop/t.zeek
```

src/strings.bif
```diff
@@ -1556,39 +1572,24 @@ function remove_prefix%(str: string, sub: string%) : string
...
 	// 1) lstrip() creates a StringVal at the end, and that would mean repeated recreation of objects
 	// 2) lstrip() searches for any character in the string, not the string as a whole.
-	string s = str->ToStdString();
-	string sub_s = sub->ToStdString();
+	auto s = str->ToStdStringView();
+	auto sub_s = sub->ToStdStringView();
 
-	size_t pos = s.find(sub_s);
-	if ( pos != 0 )
-		return zeek::IntrusivePtr<zeek::StringVal>(NewRef{}, str);
```
awelzel, line 1564: That optimization (not allocating a new string when the old one doesn't start with prefix is gone now?)

Not quite sure why it's otherwise 8% slower in the microbenchmark (and might not matter in the grand scheme), possibly the function calls and repeated string_view creation?

timwoj:
> For `remove_prefix`, any reason it got slower than master, or some copy/paste issue?
> 
> ```
> Summary
>   ./src/zeek.master -b ~/Desktop/t.zeek ran
>     1.08 ± 0.00 times faster than ./src/zeek -b ~/Desktop/t.zeek
> ```

As I mentioned in DMs, this is probably because of the pessimistic nature of the test I ran, where the target string is just repeats of the substring. In a more optimistic case where there are no matches, the new version is ~6% faster and in a slightly-less-optimistic version where it removes the prefix 3 times from a longer string, it's ~3% faster.

timwoj:
Another interesting finding is that the significant performance increase for `remove_suffix` is only happening on my M1 Pro Mac. It's ~10% slower on my Intel NUC.

timwoj:
I changed everything around a bit, and now the numbers are more consistent (and faster than master) on all platforms. `starts_with` is 5-7% faster and `remove_prefix` is 6-9% faster on both platforms

`remove_suffix` is about even on Intel for the pessimistic case but 10% faster on Apple M1. It's about 20% faster on both for the optimistic case, owing to the fact that there wasn't an early-exit check for if the string wasn't found like `remove_prefix` already had. The sort-of-optimistic case is about 7% faster on both.

In the files below, file 1 is the pessimistic case, file 2 is the optimistic case, and file 3 is the other one.

[intel.txt](https://github.com/zeek/zeek/files/14017571/intel.txt)
[mac.txt](https://github.com/zeek/zeek/files/14017572/mac.txt)

awelzel (approved):

Issue #3545, opened 2024-01-10
Make BIFs return ValPtr directly


timwoj:
Pulling out from @awelzel in #3489:

The BifReturnVal that this is related to has been introduce to transparently deal with Val* and ValPtr returns from BIFs.

Support for the BifReturnVal Val * constructor was deprecated and removed with Zeek 4.1 (https://github.com/zeek/zeek/commit/0618be792f210e29af8145e7e52225d26c27946e) so I think we can assume all bifs return ValPtr.

Couldn't we change bifcl today to render ValPtr from BIFs instead and ditch BifReturnVal.h altogether?

From a prior PR (https://github.com/zeek/zeek/pull/3468 / https://github.com/zeek/zeek/pull/3468#discussion_r1400228758), this won't help with the coverity warnings, but seems a nice cleanup as a follow up.

PR #3560, opened 2024-01-17
Make BIFs just return ValPtr directly instead of BifReturnVal

timwoj:
Fixes https://github.com/zeek/zeek/issues/3545.

This didn't actually turn out to be that difficult. I removed `BifReturnVal` altogether, but another option would be to change it to just be an alias to `ValPtr` instead. That would add a deprecation path as well. `BifReturnVal` was in the `zeek::detail` namespace though, so deprecation isn't required.


src/Hash.h
```diff
@@ -37,12 +37,11 @@ class ODesc;
 namespace zeek::detail {
 
 class Frame;
-class BifReturnVal;
 
 } // namespace zeek::detail
 
 namespace zeek::BifFunc {
-extern zeek::detail::BifReturnVal md5_hmac_bif(zeek::detail::Frame* frame, const zeek::Args*);
+extern zeek::ValPtr md5_hmac_bif(zeek::detail::Frame* frame, const zeek::Args*);
```
awelzel, line 44: Can drop the extern.

src/zeek.bif
```diff
@@ -2721,7 +2721,7 @@ function to_addr%(ip: string%): addr
 		}
 
 	delete [] s;
-	return std::move(ret);
+	return ret;
```
awelzel, line 2724: Hmm... I wonder if the `ValPtr ret` pattern could be used elsewhere, too, to avoid the noisy `return std::move(r)`: If `ret` was instead `AddrValPtr`, coverity would flag the `std::move()` as necessary, correct?

Nothing to address here, was just taken by surprise why now there's a return in BiFs that doesn't need `std::move()` anymore.

awelzel (approved):
This (the change in bifcl specifically) will break code that invoked the generated bifs directly. That seems worth noting in the NEWS, but not a reason to hold back IMO.

PR #3562, opened 2024-01-18
Backport of CI updates to 6.0

ckreibich:
This is a bunch of cherry-picks to bring `release/6.0` in line with what we run in CI on `master`. It skips the following:
- The recent `-Werror` introduction
- The broader benchmarker configuration (it already runs for Ubuntu 22, though)
- Spicy special-casing via the `ubuntu22_spicy_task` and `ubuntu22_spicy_head_task` targets — these seem more appropriate for `master`


timwoj:
I updated the docs submodule so it should pass now if you want to rebase.

.cirrus.yml
```diff
@@ -96,6 +96,7 @@ ci_template: &CI_TEMPLATE
 
   init_external_repos_script: ./ci/init-external-repos.sh
 
+  pre_build_script: ./ci/pre-build.sh
```
timwoj, line 92: Is this used by anything in 6.0? On master it's used in the `ubuntu22_spicy_head_task` task, but we don't have that task on 6.0 (or it wasn't added by this PR).

timwoj (changes requested):
One comment on something I don't think we need. The rest looks fine to me.

.cirrus.yml
ckreibich: Yeah you're right, let me drop it. Thanks!

timwoj (approved):

PR #3564, opened 2024-01-18
RFC: Helpers for raw Val* to IntrusivePtr conversions

awelzel:
I've been dabbling a bit with ideas for helpers to get from `Val*` to `AbcValPtr`, or from a "typed Val pointer", like `AddrVal*`, to an `AddrValPtr`. This mostly comes up in the context of BiFs, where Bifcl generates a preamble of raw pointers, but certain Zeek APIs are now using IntrusivePtr.

One pattern in Bifs with `any` parameters, the safe pattern is something like:
```
Val* v = ...;
if ( v->GetType()->Tag() == TYPE_RECORD )
    // bail
auto rv =  IntrusivePtr{NewRef{}, v->AsRecordVal()};
```

The other pattern is the following, which causes an abort if v isn't a RecordVal.
```
auto rv = IntrusivePtr{NewRef{}, v->AsRecordVal()};
```

The other, to get from a `AddrVal*` to an AddrValPtr is usually:
```
AddrVal* ap = ...;
auto av = zeek::IntrusivePtr{zeek::NewRef{}, ap};
```

---
This PR proposes `try_to_ptr()` and `to_ptr()` helpers to contain the above logic, so it's visually not too nosiy. For the 3 examples above - the PR contains a few examples for usages.

```
auto rv = try_to_ptr<RecordVal>(v)
if ( ! rv )
   // bail

auto rv = as_ptr<RecordVal>(v);  // fatal error if v isn't a RecordVal

auto av = to_ptr(ap);  // type of av automatically IntrusivePtr<AddrVal>;
```

Given I have no gut feeling for C++'sh things and naming is difficult, happy to hear any feedback. I've not worried about performance at all, there might be an extra ref-count increment here and there. This doesn't need to go in and happy for critcial feedback, it's just that I find some of the code around these conversions repetitive/clumsy. It could be argued Bifcl should generate a ValPtr preamble, too :man_shrugging: , but unsure we'll have that in the near future.

### DIFF
--- a/src/Val.h
+++ b/src/Val.h
@@ -488,6 +488,8 @@ private:
 
 class AddrVal final : public Val {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_ADDR;
+
     explicit AddrVal(const char* text);
     explicit AddrVal(const std::string& text);
     ~AddrVal() override;
@@ -538,6 +540,8 @@ private:
 
 class StringVal final : public Val {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_STRING;
+
     explicit StringVal(String* s);
     StringVal(std::string_view s);
     StringVal(int length, const char* s);
@@ -607,6 +611,8 @@ private:
 
 class PatternVal final : public Val {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_PATTERN;
+
     explicit PatternVal(RE_Matcher* re);
     ~PatternVal() override;
 
@@ -722,6 +728,8 @@ protected:
 
 class TableVal final : public Val, public notifier::detail::Modifiable {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_TABLE;
+
     explicit TableVal(TableTypePtr t, detail::AttributesPtr attrs = nullptr);
 
     ~TableVal() override;
@@ -912,7 +920,7 @@ public:
     ValPtr Remove(const detail::HashKey& k, bool* iterators_invalidated = nullptr);
 
     // Returns a ListVal representation of the table (which must be a set).
-    ListValPtr ToListVal(TypeTag t = TYPE_ANY) const;
+    ListValPtr ToListVal(zeek::TypeTag t = TYPE_ANY) const;
 
     // Returns a ListVal representation of the table (which must be a set
     // with non-composite index type).
@@ -1081,6 +1089,8 @@ inline constexpr bool is_zeek_val_v = is_zeek_val<T>::value;
 
 class RecordVal final : public Val, public notifier::detail::Modifiable {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_RECORD;
+
     explicit RecordVal(RecordTypePtr t, bool init_fields = true);
 
     ~RecordVal() override;
@@ -1469,6 +1479,8 @@ private:
 
 class EnumVal final : public detail::IntValImplementation {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_ENUM;
+
     ValPtr SizeVal() const override;
 
 protected:
@@ -1502,6 +1514,8 @@ protected:
 
 class VectorVal final : public Val, public notifier::detail::Modifiable {
 public:
+    static constexpr auto TypeTag = zeek::TYPE_VECTOR;
+
     explicit VectorVal(VectorTypePtr t);
     VectorVal(VectorTypePtr t, std::vector<std::optional<ZVal>>* vals);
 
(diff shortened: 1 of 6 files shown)


src/Val.h
```diff
@@ -1753,4 +1767,55 @@ extern void concretize_if_unspecified(VectorValPtr v, TypePtr t);
 
 } // namespace detail
 
+struct to_ptr_auto_type {};
+
+template<typename T, typename RefTag = zeek::NewRef>
+zeek::IntrusivePtr<T> try_to_ptr(zeek::Val* val) {
```
Neverlord, line 1773: Unless I miss something, this is basically to `cast_intrusive` what `dynamic_cast` is to `static_cast`.

Just from reading the function name, I would expect the function to actually return a pointer (not a smart pointer). But it's quite the opposite. The function also conflates two operations:

1. Conversion of a raw pointer to a smart pointer.
2. (Safe) downcast from a base type (`Val`) to a derived type.

Semantically, this is basically `dynamic_cast_intrusive(IntrusivePtr{RefTag{}, val})`, whereas `dynamic_cast_intrusive` would be:

```c++
template<class T, class U>
IntrusivePtr<T> dynamic_cast_intrusive(IntrusivePtr<U> p) noexcept {
    return {AdoptRef{}, dynamic_cast<T*>(p.release())};
}
```

You're using `val->As<T*>()` instead of actually using `dynamic_cast`, but this is just me trying to get a handle on your ideas.

So the first question I have is: why not make this a generic utility for `IntrusivePtr`? It seems like the `As` functions are just type-checked downcast versions that we could just as well do with `dynamic_cast`. Please correct me if I miss something, but the implementations from `CONVERTER` in `Val.cc` just do a type check and then a downcast. Is that type check really faster than `dynamic_cast`?

Stepping up, there are three versions here:

1. `try_to_ptr`, which is basically `dynamic_cast` for `IntrusivePtr`.
2. `to_ptr`, which is also a `dynamic_cast` but this will terminate (?) the application if the conversion fails.
3. `as_ptr` is the same as `to_ptr`, only that it expects a smart pointer instead of a regular pointer.

So maybe we could just add two functions to `IntrusivePtr` to achieve the same but with a more generic interface?

```c++
// Converts `T` to `Derived` and throws / terminates the application if the conversion fails.
template <typename Derived, typename Base>
IntrusivePtr<Derived> subtype(IntrusivePtr<Base> ptr);

// Tries to convert `T` to `Derived` and returns `nullptr` if the conversion fails.
template <typename Derived, typename Base>
IntrusivePtr<Derived> subtype_cast(IntrusivePtr<Base> ptr);
```

The idea here is similar to `narrow` and `narrow_cast` from the GSL: one assumes that the conversion will succeed (and throws otherwise) and the other does the casting and returns the result regardless (which may be `null` here).

As to functions taking "raw" pointers: I would still leave the conversion to `IntrusivePtr` in. So `try_to_ptr(ptr)` would become `subtype_cast(IntrusivePtr{NewRef{}, ptr})`, which is slightly longer but at least to me it would read better. And if we use this pattern a lot, maybe we can introduce something like `NewRef::From(ptr)` or just `new_ref(ptr)`.

Just my thoughts. 🙂

awelzel: > The function also conflates two operations:

That was intentional - helpers to go from plain `Val*` to a `ConcreteValPtr` in one step. And the `try_` version that doesn't terminate the process on tag mismatch to cover the pattern of first checking the type, then doing the conversion and also going to an IntrusivePtr.

We have a whole set of `AsConcreteExprPtr` implementations for `Expr` and it's subtypes:
https://github.com/zeek/zeek/blob/13fde341d25302f46901693a05d191591781c6ea/src/Expr.h#L231-L240

Essentially, this was an attempt to provide the same functionality for the `Val` side of things, but without having all the explicit implementations. I could also see us adding a whole slew of `AsConcreteValPtr` functions to mirror this - though I don't like it.

> Is that type check really faster than dynamic_cast?
> So maybe we could just add two functions to IntrusivePtr to achieve the same but with a more generic interface?

I read a bit around and sounded as if `dynamic_cast` goes through rtti and that would then result in string comparisons of the involved type names, so didn't even go near it. Otherwise yes, it could be generic.

---

I think `subtype_cast<X>(NewRef::From(ptr))` would be okay (disregarding the implementation which IMO should go trough `->GetType()->Tag()`). If `NewRef::From` returns an `IntrusivePtr<AddrVal>` given an `AddrVal*`, this alone might already be a nice addition.

awelzel:
I'm closing this. Thanks @Neverlord  for the feedback. When tempted next time, I might try `NewRef::From()` helper as a first step.

PR #3226, opened 2023-08-09
Explicitly link Zeek executable against Spicy libraries in binary packaging mode.

bbannier:
Closes #3177.

PR #3565, opened 2024-01-18
Make sure Spicy symbols are available.

### LINKED ISSUE #3177: Statically-linked zeek binary from Debian_12 debs missing symbol(s) used by spicy
Steps to reproduce (uses a provided Dockerfile, but likely could produce on any Debian 12 system using the 6.0.0 release .deb files from opensuse.org)

1. Download this [Dockerfile](https://github.com/zeek/zeek/files/11970377/Dockerfile.txt)
    * This Dockerfile is a bare-bones bookworm-slim image that installs minimum packages needed to download the Zeek 6.0.0 debs from [opensuse.org](https://download.opensuse.org/repositories/security:/zeek/Debian_12)
1. `docker build -t zeek-spicy-crash .`
1. `docker run -Pit --rm --entrypoint=/bin/bash zeek-spicy-crash`
1. `zkg install --skiptests --force https://github.com/cisagov/icsnpp-synchrophasor`
    * I'm not sure what is special about this parser, but one of my other parsers, `cisagov/icsnpp-genisys`, does not reproduce the problem
1. `zeek -NN local`

Observed behavior:

```
# zeek -NN local
[libhilti] Fatal error: could not open library path "/opt/zeek/lib/zeek/plugins/packages/icsnpp-synchrophasor/synchrophasor.hlto":
failed to load library "/opt/zeek/lib/zeek/plugins/packages/icsnpp-synchrophasor/synchrophasor.hlto":
/opt/zeek/lib/zeek/plugins/packages/icsnpp-synchrophasor/synchrophasor.hlto:
undefined symbol: _ZN5hilti2rt4real6unpackERKNS0_6stream4ViewENS1_4TypeENS0_9ByteOrderE
```

Expected behavior: For `zeek` to run and output the list of available analyzers

Discussed in Slack in a thread in the #spicy channel with @bbannier and @0xxon

bbannier:
The Spicy/HILTI runtime libraries are modelled as object libraries and linked into the `spicy` and `hilti` libraries. These libraries are then linked into `zeek_objs` which is another object library linked into the `zeek` executable (target `zeek_exe`). On some platforms this transitive link of object libraries is broken with `--binary-package` which causes creation of static archives, even with our minimum required CMake version which already contains many fixes for object libraries.

With this patch we now explicitly linked the Spicy/HILTI runtime libraries into `zeek_exe`.

Closes #3177.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -959,8 +959,10 @@ if (NOT DISABLE_SPICY)
         # instead explicitly branch on `BINARY_PACKAGING_MODE` here.
         if (BINARY_PACKAGING_MODE)
             hilti_link_object_libraries_in_tree(zeek_exe PRIVATE)
+            spicy_link_object_libraries_in_tree(zeek_exe PRIVATE)
         else ()
             hilti_link_libraries_in_tree(zeek_exe PRIVATE)
+            spicy_link_libraries_in_tree(zeek_exe PRIVATE)
         endif ()
 
         set(HAVE_SPICY yes)


src/CMakeLists.txt
```diff
@@ -536,6 +536,16 @@ zeek_target_link_libraries(zeek_objs)
 
 if (HAVE_SPICY)
     target_link_libraries(zeek_objs PRIVATE hilti spicy)
+
+    # On some platforms the transitive link of the Spicy and HILTI object
+    # libraries via e.g., `hilti` and `zeek_objs` into `zeek_exe` does not work,
+    # see #3177. Make it explicit to be sure.
+    if (BINARY_PACKAGING_MODE)
+        target_link_libraries(
```
bbannier, line 544: I am not really sure whether this extra conditional makes this slip through CI, but would still fail similar to the last failing run on this PR on some platforms, but at least from what I see this fixes the issue I was seeing with the reproducer and custom-built packages. @0xxon, would you mind taking this on a test run through the packaging pipeline?

src/CMakeLists.txt
```diff
@@ -536,6 +536,16 @@ zeek_target_link_libraries(zeek_objs)
...
+            zeek_exe PRIVATE $<IF:$<CONFIG:Debug>,hilti-rt-debug-objects,hilti-rt-objects>
+                             $<IF:$<CONFIG:Debug>,spicy-rt-debug-objects,spicy-rt-objects>)
+    endif ()
+
```
awelzel, line 548: The top-level CMakeLists.txt has this:

https://github.com/zeek/zeek/blob/6d10082cc35d9b0b4ad527cc320ff7d7ee597445/CMakeLists.txt#L956-L964

Isn't that doing the same already, but with a helper? There isn't a call to `spicy_link_object_libraries_in_tree` though. Just a wild guess.

It would be good to explain the new target_link_libraries and the quoted code - superficially they seem to do the same thing.

bbannier: Excellent point, it looks like that place place is missing the link against Spicy. I'll check whether doing that instead would fix the issue as well.

bbannier: That fixed it.

awelzel (approved):
Oh, this is basically just a follow-up for this one?

https://github.com/zeek/zeek/pull/3226

Seems a good candidate for 6.0.3 and 6.1? if @timwoj could pick this up still?

bbannier:
> Oh, this is basically just a follow-up for this one?

Yes exactly, #3226 was the initial attempt for a fix, but didn't fully address #3177 so we reopened the issue.

PR #3568, opened 2024-01-19
SMTP/BDAT: Fix huge chunk size overflow issues

awelzel:
A few things as a follow-up:

* I totally borked comparison and working with int, int64_t and uint64_t
* The ContentLine analyzer does not support plain delivery of chunks > int64_t max.
* Use strtoull and check if the input was too large

Running the fuzzer locally the same issue found by OSS-Fuzz (Issue 65933) triggered after a few minutes. Nothing else came up for some 30 minutes now, so I hopefully this has the worst squelched.


src/analyzer/protocol/smtp/SMTP.cc
```diff
@@ -857,6 +858,15 @@ bool SMTP_Analyzer::ProcessBdatArg(int arg_len, const char* arg, bool orig) {
         return false;
     }
 
+    // The ContentLine analyzer only supports int64_t, but BDAT could deal
+    // with uint64_t sized chunks. Weird if the chunk size is larger and
+    // do not configure the ContentLine analyzer for plain delivery.
+    if ( chunk_size > std::numeric_limits<int64_t>::max() ) {
```
timwoj, line 867: Could this cause us to drop valid packets? Should we open an issue to investigate fixing the ContentLine analyzer?

src/analyzer/protocol/smtp/BDAT.cc
```diff
@@ -18,13 +18,20 @@ struct BDATCmd parse_bdat_arg(int length, const char* arg) {
         return r;
     }
 
+    errno = 0;
```
timwoj, line 21: Does this need to be set if we don't use it?

src/analyzer/protocol/smtp/SMTP.cc
awelzel: I think it's fine. 2**63 is is 8192 petabytes - here it's definitely weird and suspect immensely huge in other places, too. 
```
In [8]: 2**63/(1024*1024*1024*1024*1024)
Out[8]: 8192.0
```

timwoj (approved):

PR #3575, opened 2024-01-23
Remove setting non-existent session history for IPTunnel

timwoj:
OSS-Fuzz found this one, which was added in 7325bc945f1625614e358deb6421f787fe16a6e7. IP-in-IP tunnels won't have a session because those are set from either the UDP or TCP analyzers, which are never invoked if we go directly from IP to an tunnel. We can safely remove setting the history here since the tunnel never gets logged as a connection.


awelzel (approved):
And you confirmed the other places all pre-check `packet->session`.

Thanks for the fix, would still suggest to add the following as a regression somewhere to the `max_depth` test suite.

```
zeek -C -r testing/btest/Traces/tunnels/gre-within-gre.pcap Tunnel::max_depth=1
AddressSanitizer:DEADLYSIGNAL
==303159==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000090 (pc 0x55e609f3050c bp 0x7ffe0de2cce0 sp 0x7ffe0de2c610 T0)
```

PR #3576, opened 2024-01-23
Switch all of the Windows Conan configuration to vcpkg

timwoj:
This PR switches our Conan configuration used on Windows for third-party packages to use vcpkg instead. The primary reason for this is that Conan's CMake support is definitely not considered first-party, and lags behind changes that they make in the core pretty heavily. For example, support for Conan 2.0 which has been out for almost a year now is still considered "experimental" in conan-cmake. We have an open issue to switch away from the existing deprecated generators in Conan 1.x, but switching to Conan 2.x is difficult. On top of that, a number of packages fail to build correctly (or even get through CMake configuration), which throws a wrench in using any of them. For example, OpenSSL fails and has been a [reported problem](https://github.com/conan-io/conan/issues/12012) that they've been kicking down the road for a long while.

The one significant change is that we have to include vcpkg as a submodule to Zeek for versioning support, but the submodule is small so it's not a big deal.

I've added OpenSSL as a vcpkg submodule instead of using the one installed via Chocolatey due to some build issues during testing with it not finding the library. It bloats the build times of on Windows a bit on CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,7 +133,7 @@ if (MSVC)
     # Set PCAP_ROOT_DIR to point at the installation from vcpkg. A later call
     # to FindPCAP.cmake will fill in the rest of the necessary variables.
     if (NOT PCAP_ROOT_DIR)
-        set(PCAP_ROOT_DIR "${VCPKG_INSTALLED_DIR}")
+        set(PCAP_ROOT_DIR ${VCPKG_INSTALLED_DIR}/x64-windows-static)
     endif ()
 
     set(LIBPCAP_PCAP_COMPILE_NOPCAP_HAS_ERROR_PARAMETER false)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,17 @@ cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
 if (WIN32)
     # Enable usage of CMAKE_MSVC_RUNTIME_LIBRARY variable
     cmake_policy(SET CMP0091 NEW)
+
+    # I tried to just use CMAKE_SOURCE_DIR and CMAKE_CURRENT_SOURCE_DIR
+    # but it's not setting the path correctly and so the toolchain
+    # variable doesn't get passed down to submodules like libkqueue
+    # correctly. Instead get the absolute path to the vcpkg.cmake file
+    # and use that.
+    get_filename_component(_toolchain ./auxil/vcpkg/scripts/buildsystems/vcpkg.cmake ABSOLUTE)
+
+    # This needs to happen before the project() call below so that it
+    # doesn't need to be manually passed on the command line.
+    set(CMAKE_TOOLCHAIN_FILE ${_toolchain} CACHE STRING "Vcpkg toolchain file")
 endif ()
 
 project(Zeek C CXX)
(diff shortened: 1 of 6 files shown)


timwoj:
This is ready for review.

timwoj:
After fixing https://github.com/zeek/zeek/pull/3587, I rebased and removed the openssl change.

ci/windows/build.cmd
```diff
@@ -7,5 +7,5 @@ call "c:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliar
 mkdir build
 cd build
 
-cmake.exe .. -DCMAKE_BUILD_TYPE=release -DENABLE_ZEEK_UNIT_TESTS=yes -G Ninja
+cmake.exe .. -DCMAKE_BUILD_TYPE=release -DCMAKE_TOOLCHAIN_FILE="auxil/vcpkg/scripts/buildsystems/vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="x64-windows-static" -DENABLE_ZEEK_UNIT_TESTS=yes -G Ninja
```
timwoj, line 10: The `DCMAKE_TOOLCHAIN_FILE` part here might not be necessary since we're setting it during CMake.

philrz:
@timwoj asked me on Slack to test these changes and I can confirm that they seem to work for me. Specifically, we have a repo where we run steps in GitHub Actions based on the Windows instructions from https://docs.zeek.org/en/master/install.html against a checkout of the Zeek source tree to create artifacts to bundle with Brimcap/Zui. I adjusted the steps based on this PR as shown in https://github.com/brimdata/build-zeek/compare/main...zeek-3576 and the artifact built without complaint and I was able to use it to generate logs from a large test pcap as usual.

As long as I'm chiming in, I'll note that when these steps become official and there's a GA release that requires them it'd be ideal to update the Windows steps at https://docs.zeek.org/en/master/install.html at that time. I'm now Subscribed to this PR so I'll try to remember as well and put up a docs PR if someone else doesn't get to it first.

CMakeLists.txt
```diff
@@ -5,6 +5,17 @@ cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
 if (WIN32)
     # Enable usage of CMAKE_MSVC_RUNTIME_LIBRARY variable
     cmake_policy(SET CMP0091 NEW)
+
+    # I tried to just use CMAKE_SOURCE_DIR and CMAKE_CURRENT_SOURCE_DIR
+    # but it's not setting the path correctly and so it the toolchain
```
ckreibich, line 10:
```suggestion
    # but it's not setting the path correctly and so the toolchain
```

CMakeLists.txt
```diff
@@ -5,6 +5,17 @@ cmake_minimum_required(VERSION 3.15.0 FATAL_ERROR)
...
+    # variable doesn't get passed down to submodules like libkqueue
+    # correctly. Instead get the absolute path to the vcpkg.cmake file
+    # and use that.
+    get_filename_component(_toolchain ./auxil/vcpkg/scripts/buildsystems/vcpkg.cmake ABSOLUTE)
+
+    # This needs to happen before the project() call below so that it
+    # doesn't need to be manually passed on the ccommand line.
```
ckreibich, line 17:
```suggestion
    # doesn't need to be manually passed on the command line.
```

CMakeLists.txt
```diff
@@ -119,33 +130,23 @@ if (MSVC)
...
-    if (ZEEK_STANDALONE)
-        include(${CMAKE_SOURCE_DIR}/cmake/conan.cmake)
-        conan_cmake_autodetect(settings)
-        # Install packages from conanfile
-        conan_cmake_install(PATH_OR_REFERENCE ${CMAKE_SOURCE_DIR}/ci/windows/conanfile_windows.txt
-                            BUILD missing SETTINGS ${settings})
-    endif ()
-
-    # Set LibPCAP to point to libpcap binaries.
+    # Set PCAP_ROOT_DIR to point at the installation from vcpkg. A later call
```
ckreibich, line 133: The install section of the docs currently spells out chocolatey instructions. Looks like we'll need to update those, right? I'm not fully clear on how they relate.

CMakeLists.txt
```diff
@@ -119,33 +130,23 @@ if (MSVC)
...
+        set(PCAP_ROOT_DIR ${VCPKG_INSTALLED_DIR}/x64-windows-static)
     endif ()
 
     set(LIBPCAP_PCAP_COMPILE_NOPCAP_HAS_ERROR_PARAMETER false)
 
-    # Set ZLib to point at the right variable.
+    # Find zlib installed by vcpkg.
     find_package(ZLIB)
-    set(ZLIB_LIBRARY ${ZLIB_LIBRARIES})
+    set(ZLIB_LIBRARY ZLIB::ZLIB)
```
ckreibich, line 143: Interesting, I like how tidy this looks. Let me make sure I get the workflow: it's the inclusion of the toolchain file for vcpkg that kicks this all off; this then pulls in vcpkg.json to install c-ares, libpcap, and zlib, and it populates `ZLIB::ZLIB` and `c-ares::cares` with the results. Is that roughly right?

ckreibich (commented):
I don't know vcpkg so this is mostly questions to help me understand — thanks...

CMakeLists.txt
timwoj: Correct. Vcpkg's CMake integration automatically creates the `ZLIB::ZLIB` and `c-ares::cares` library targets. The libpcap package for vcpkg doesn't do that though, so we still have to handle it like a bare library/include path.

CMakeLists.txt
timwoj: We still need the libraries that are installed by chocolatey. You just don't need to install conan anymore.

ckreibich (approved):
Thanks Tim. Consider this a non-expert approval. :-)

PR #3578, opened 2024-01-24
testing/scripts: Use sed -E everywhere

awelzel:
I'm always a bit worried to use sed -E anywhere, because the canonifiers give the impression it won't work everywhere consistently. My manpage says sed -E should be preferred for portability, so lets remove the sed -r / sed -E differentiation assuming it's just a thing from the past.

### DIFF
--- a/testing/scripts/diff-clean-doctest
+++ b/testing/scripts/diff-clean-doctest
@@ -5,13 +5,6 @@
 # - The version number, which becomes "x.y.z"
 # - Spacing in the report, which we normalize to single spaces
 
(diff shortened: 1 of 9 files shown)


bbannier (approved):
Thanks for the cleanup! Can't believe what I am saying, but this looks better 😅

grigorescu:
Best I can tell, `-E` was added to sed 18 years ago: https://git.savannah.gnu.org/gitweb/?p=sed.git;a=commitdiff;h=3a8e165ab02487c372df217c1989e287625ce0ae

Think we're safe.

PR #3580, opened 2024-01-24
websocket: Add Spicy parser version, too.

awelzel:
The Spicy analyzer is added as a child analyzer when enabled and the WebSocket.cc logic dispatches between the BinPac and Spicy version.

It substantially slower when tested against a somewhat artificial 2.4GB PCAP compared to the BinPac analyzer. The first flamegraph indicates that the unmask() function stands out with 35% of all samples, and above it shared_ptr samples.


src/analyzer/protocol/websocket/websocket.spicy
```diff
@@ -0,0 +1,121 @@
...
+function unmask(inout f: Frame, chunk: bytes): bytes {
+  if ( ! f.mask )
+    return chunk;
+
+  local unmasked_chunk: bytes;
+  local i = 0;
+  # This is very very expensive :-( Creates a new shared_ptr for
+  # each byte we iterate over it seems. Visible in the first
+  # flamegraph I looked at.
+  for ( c in chunk ) {
```
awelzel, line 26: This iteration is done using safe iterators on the underlying bytes object, instantiating shared_ptrs over and over. It's by far the dominating perf sink here. 70% of samples fall into shared_ptr constructor and destructors and it's 30% of the total runtime for my testing.

Screenshot from 2024-01-24 20-45-24

src/analyzer/protocol/websocket/websocket.spicy
```diff
@@ -0,0 +1,121 @@
...
+
+  : void {
+    self.payload_len = self.payload_len1;
+    if ( self?.payload_len2 )
+      self.payload_len = self.payload_len2;
+    else if ( self?.payload_len8 )
+      self.payload_len = self.payload_len8;
+  }
+
+  masking_key: uint8[] &size=4 if (self.mask);
```
awelzel, line 65: This being an `uint8[]` allows `masking_key[x]` indexing, while a `bytes` object would require `*masking_key.at(i)` which took roughly 20% more runtime when I tested it.

ckreibich (approved):
Looks great, thank you! It's also a nice template for how to introduce alternative parsers in the future — thinking of TLS, in particular.

src/analyzer/protocol/websocket/websocket.spicy
```diff
@@ -0,0 +1,120 @@
...
+
+const OPCODE_CONTINUATION = 0x00;
+const OPCODE_TEXT         = 0x01;
+const OPCODE_BINARY       = 0x02;
+const OPCODE_CLOSE        = 0x08;
+const OPCODE_PING         = 0x09;
+const OPCODE_PONG         = 0x0a;
+
+# Are bytes copied like a string in C++, or is that ref counted and cheap
+# like it would be in Python?
```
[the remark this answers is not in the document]
bbannier: No need to guess here: `f` is passed as `inout` so we generate code taking it as mutable ref; `chunk` is `in` so it becomes a const ref argument. If you are ever curious, just look at the generated code, e.g., here

```ruby
module foo;

type Frame = unit {};

function unmask( inout f: Frame, chunk: bytes): bytes {
    return b"";
}
```

```console
# Disable optimizations with `-g` to avoid removing unused code.
$ spicyc -cg foo.spicy
..
static auto unmask(::hilti::rt::ValueReference<__hlt::foo::Frame>& f, const ::hilti::rt::Bytes& chunk) -> ::hilti::rt::Bytes;
..
```

src/analyzer/protocol/websocket/websocket.spicy
bbannier: We discussed this at length already. The issue here is that iteration over a `bytes` container is by design safe in Spicy, i.e., even in a for each loop we still check for element validity. This is needed since the loop body could invalidate elements (also the reason a `for_each` method taking a callback on `bytes` could not avoid that).

Like discussed, if performance of this piece is critical one approach would be to move it to a C++ function; if this code wouldn't be called during `&chunked` parsing it very likely would also be possible to more directly express this in the parsing code.

bbannier (commented):
I would suggest to either capture your comments around Spicy usage in the parser in Spicy issues and reference them, or directly documenting why a certain workaround was picked.

src/analyzer/protocol/websocket/websocket.spicy
awelzel: @bbannier - I've implemented unmask in C++ and just doing that improved overall runtime by 1.8x (from 22 down to 12 seconds). Mind checking out the unmask.cc file and signing-off if it doesn't looks too troublesome?

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2024 by the Zeek Project. See COPYING for details.
+
+// Import HILTI
+#include <hilti/rt/libhilti.h>
+
+namespace hlt_websocket::WebSocket {
+
+// Implement XOR unmasking of WebSocket frames due to being very
+// hot and inefficient in pure Spicy.
```
bbannier, line 9: The comment is unnecessarily judgemental (e.g., I'd expect that being able to write that function in Spicy code was an efficient use of your time, but you particularly care about throughput here).

```suggestion
// Implement XOR unmasking of WebSocket frames in C++ since this code is very hot.
```

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2024 by the Zeek Project. See COPYING for details.
+
+// Import HILTI
```
bbannier, line 3: This just repeats what the code does.

```suggestion
```

src/analyzer/protocol/websocket/websocket.spicy
```diff
@@ -0,0 +1,143 @@
...
+const OPCODE_TEXT         = 0x01;
+const OPCODE_BINARY       = 0x02;
+const OPCODE_CLOSE        = 0x08;
+const OPCODE_PING         = 0x09;
+const OPCODE_PONG         = 0x0a;
+
+# Spicy implementation of WebSocket unmasking.
+#
+# This isn't used because it's too slow. Instead, it's
+# implemented in C++ unmask.cc as fast_unmask().
```
bbannier, line 18: Do we need to keep this around?

```suggestion
# This isn't used because since this code is pretty hot.
# Instead we use `fast_unmask()` implemented in C++.
```

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
...
+    constexpr size_t masking_key_size = 4;
+
+    if ( masking_key.size() != masking_key_size )
+        throw hilti::rt::RuntimeError(hilti::rt::fmt("wrong masking_key size %ld", masking_key.size()));
+
+    const uint64_t mi = masking_key_idx;
+    std::array<uint8_t, masking_key_size> unsafe_masking_key;
+
+    size_t i = 0;
+    for ( auto it = masking_key.unsafeBegin(); it != masking_key.unsafeEnd(); it++ )
```
bbannier, line 24: Can use in-place increment.

```suggestion
    for ( auto it = masking_key.unsafeBegin(); it != masking_key.unsafeEnd(); ++it )
```

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
...
+        unsafe_masking_key[i++] = *it;
+
+    std::string unmasked;
+    unmasked.resize(chunk.size());
```
bbannier, line 28: You know the final size, so can directly allocate it.

```suggestion
    std::string unmasked(chunk.size(), '\0');
```

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
...
+    const auto& chunk_str = chunk.str();
+
+    for ( i = 0; i < chunk_str.size(); i++ )
```
bbannier, line 31: Let's not reuse the index variable across loops.

```suggestion
    for ( size_t i = 0; i < chunk_str.size(); i++ )
```

src/analyzer/protocol/websocket/unmask.cc
```diff
@@ -0,0 +1,37 @@
...
+// hot and inefficient in pure Spicy.
+//
+// https://github.com/zeek/spicy/issues/1663
+::hilti::rt::Bytes fast_unmask(const hilti::rt::integer::safe<uint64_t>& masking_key_idx,
+                               const hilti::rt::Vector<hilti::rt::integer::safe<uint8_t>>& masking_key,
+                               const hilti::rt::Bytes& chunk) {
+    constexpr size_t masking_key_size = 4;
+
+    if ( masking_key.size() != masking_key_size )
+        throw hilti::rt::RuntimeError(hilti::rt::fmt("wrong masking_key size %ld", masking_key.size()));
```
bbannier, line 18: `RuntimeError` is just a base class and we probably have a type fitting better.

- if this is due to some error in the input and you want to recover from it use a `RecoverableFailure`
- if this is due to a misuse of this function use a `UsageError`

src/analyzer/protocol/websocket/unmask.cc
awelzel: Ah, perfect.

PR #3583, opened 2024-01-25
Complete association of location information with AST nodes / ZAM generated code

vpax:
ZAM script optimization creates a bunch of new AST nodes derived from the original AST. To tie these to the correct location information, ZAM introduced the notion of the "original" node associated with a new node (for both `Stmt` and `Expr` objects). Doing this at the node-level was handy during earlier development when debugging ZAM's AST transformations. However, now that those are solid, the associations are overly complex (all that's fundamentally needed these days is the location information), and also incomplete (some new nodes wind up having `no_location` for their location).

This PR removes the notion of "original" node and replaces it with a simple mechanism for derived nodes to use to mirror the location information of their source nodes. The changes are lengthy mainly because they fix up a bunch of instances where a new node did not in fact correctly track its source. (The PR includes an `ASSERT` in the ZAM compiler that ensures that every non-trivial node has a non-trivial location associated with it.)

There are also a couple of secondary changes similar in spirit: (1) for inlined functions, tracking their original names, (2) for specialized ZAM instructions that replace certain BiFs, tracking the original function call to the BiF.

(All of this is preparation for work-in-progress on ZAM supporting generating fine-grained flame graphs for profiling execution.)

### DIFF
--- a/src/Expr.cc
+++ b/src/Expr.cc
@@ -446,7 +446,7 @@ ExprPtr NameExpr::MakeLvalue() {
     if ( id->IsOption() && ! in_const_init )
         ExprError("option is not a modifiable lvalue");
 
-    return make_intrusive<RefExpr>(ThisPtr());
+    return with_location_of(make_intrusive<RefExpr>(ThisPtr()), this);
 }
 
 void NameExpr::Assign(Frame* f, ValPtr v) {
@@ -1038,9 +1038,9 @@ void BinaryExpr::PromoteOps(TypeTag t) {
         bt2 = op2->GetType()->AsVectorType()->Yield()->Tag();
 
     if ( bt1 != t )
-        op1 = make_intrusive<ArithCoerceExpr>(op1, t);
+        op1 = with_location_of(make_intrusive<ArithCoerceExpr>(op1, t), op1);
     if ( bt2 != t )
-        op2 = make_intrusive<ArithCoerceExpr>(op2, t);
+        op2 = with_location_of(make_intrusive<ArithCoerceExpr>(op2, t), op2);
 }
 
 void BinaryExpr::PromoteType(TypeTag t, bool is_vector) {
@@ -1059,7 +1059,7 @@ void BinaryExpr::PromoteForInterval(ExprPtr& op) {
         SetType(base_type(TYPE_INTERVAL));
 
     if ( op->GetType()->Tag() != TYPE_DOUBLE )
-        op = make_intrusive<ArithCoerceExpr>(op, TYPE_DOUBLE);
+        op = with_location_of(make_intrusive<ArithCoerceExpr>(op, TYPE_DOUBLE), op);
 }
 
 bool BinaryExpr::CheckForRHSList() {
@@ -1078,7 +1078,7 @@ bool BinaryExpr::CheckForRHSList() {
 
             for ( auto i = 1U; i < rhs_exprs.size(); ++i ) {
                 ExprPtr re_i = {NewRef{}, rhs_exprs[i]};
-                op2 = make_intrusive<BitExpr>(EXPR_OR, op2, re_i);
+                op2 = with_location_of(make_intrusive<BitExpr>(EXPR_OR, op2, re_i), op2);
             }
 
             SetType(op1->GetType());
@@ -1098,9 +1098,9 @@ bool BinaryExpr::CheckForRHSList() {
         }
 
         if ( lhs_t->IsTable() )
-            op2 = make_intrusive<TableConstructorExpr>(rhs, nullptr, lhs_t);
+            op2 = with_location_of(make_intrusive<TableConstructorExpr>(rhs, nullptr, lhs_t), op2);
         else
-            op2 = make_intrusive<SetConstructorExpr>(rhs, nullptr, lhs_t);
+            op2 = with_location_of(make_intrusive<SetConstructorExpr>(rhs, nullptr, lhs_t), op2);
     }
 
     else if ( lhs_t->Tag() == TYPE_VECTOR ) {
@@ -1109,7 +1109,7 @@ bool BinaryExpr::CheckForRHSList() {
             return false;
         }
 
-        op2 = make_intrusive<VectorConstructorExpr>(rhs, lhs_t);
+        op2 = with_location_of(make_intrusive<VectorConstructorExpr>(rhs, lhs_t), op2);
     }
 
     else {
@@ -1422,7 +1422,7 @@ AddToExpr::AddToExpr(ExprPtr arg_op1, ExprPtr arg_op2)
         if ( IsArithmetic(bt1) ) {
             if ( IsArithmetic(bt2) ) {
                 if ( bt2 != bt1 )
-                    op2 = make_intrusive<ArithCoerceExpr>(std::move(op2), bt1);
+                    op2 = with_location_of(make_intrusive<ArithCoerceExpr>(op2, bt1), op2);
 
                 SetType(t1);
             }
@@ -2007,9 +2007,9 @@ CondExpr::CondExpr(ExprPtr arg_op1, ExprPtr arg_op2, ExprPtr arg_op3)
         if ( BothArithmetic(bt2, bt3) ) {
             TypeTag t = max_type(bt2, bt3);
             if ( bt2 != t )
-                op2 = make_intrusive<ArithCoerceExpr>(std::move(op2), t);
+                op2 = with_location_of(make_intrusive<ArithCoerceExpr>(op2, t), op2);
             if ( bt3 != t )
-                op3 = make_intrusive<ArithCoerceExpr>(std::move(op3), t);
+                op3 = with_location_of(make_intrusive<ArithCoerceExpr>(op3, t), op3);
 
             if ( is_vector(op1) )
                 SetType(make_intrusive<VectorType>(base_type(t)));
@@ -2031,9 +2031,11 @@ CondExpr::CondExpr(ExprPtr arg_op1, ExprPtr arg_op2, ExprPtr arg_op3)
                 auto tt3 = op3->GetType<TableType>();
 
                 if ( tt2->IsUnspecifiedTable() )
-                    op2 = make_intrusive<TableCoerceExpr>(std::move(op2), std::move(tt3));
+                    op2 = with_location_of(make_intrusive<TableCoerceExpr>(op2, std::move(tt3)), op2);
+
                 else if ( tt3->IsUnspecifiedTable() )
-                    op3 = make_intrusive<TableCoerceExpr>(std::move(op3), std::move(tt2));
+                    op3 = with_location_of(make_intrusive<TableCoerceExpr>(op3, std::move(tt2)), op3);
+
                 else if ( ! same_type(op2->GetType(), op3->GetType()) )
                     ExprError("operands must be of the same type");
             }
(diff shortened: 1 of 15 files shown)


src/Expr.cc
```diff
@@ -4653,14 +4731,20 @@ ExprPtr check_and_promote_expr(ExprPtr e, TypePtr t) {
     TypeTag t_tag = t->Tag();
 
     if ( t_tag == TYPE_ANY ) {
-        if ( e_tag != TYPE_ANY )
-            return make_intrusive<CoerceToAnyExpr>(e);
+        if ( e_tag != TYPE_ANY ) {
+            auto c = make_intrusive<CoerceToAnyExpr>(e);
+            c->SetLocation(e);
```
awelzel, line 4736: I haven't checked all the call-sites, but this looks like a lot of churn easy to miss to keep consistent.

If we do this for most cases now, could we just move it into the `UnaryExpr` and `BinaryExpr` constructor to ensure it? Essentially, make new expressions use their first op's location implicitly. If in a given situation that's not appropriate, they can still do `SetLocation()` to something else.

```
diff --git a/src/Expr.cc b/src/Expr.cc
index 20d5042b6..94d687b0f 100644
--- a/src/Expr.cc
+++ b/src/Expr.cc
@@ -502,6 +502,8 @@ TraversalCode ConstExpr::Traverse(TraversalCallback* cb) const {
 UnaryExpr::UnaryExpr(ExprTag arg_tag, ExprPtr arg_op) : Expr(arg_tag), op(std::move(arg_op)) {
     if ( op->IsError() )
         SetError();
+
+    SetLocationInfo(arg_op->GetLocationInfo());
 }
 
 ValPtr UnaryExpr::Eval(Frame* f) const {
diff --git a/src/Expr.h b/src/Expr.h
index c4c60e156..443806a5f 100644
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -552,6 +552,8 @@ protected:
             return;
         if ( op1->IsError() || op2->IsError() )
             SetError();
+
+        SetLocationInfo(op1->GetLocationInfo());
     }
```

src/Expr.h
```diff
@@ -365,16 +367,8 @@ class Expr : public Obj {
...
-    // rather than an ExprPtr to emphasize that the access is read-only.
-    const Expr* Original() const { return original ? original->Original() : this; }
-
-    // Designate the given Expr node as the original for this one.
-    void SetOriginal(ExprPtr _orig) {
-        if ( ! original )
-            original = std::move(_orig);
-    }
+    void SetLocation(const ObjPtr& orig_obj) { SetLocation(orig_obj.get()); }
+    void SetLocation(const Obj* orig_obj) { SetLocationInfo(orig_obj->GetLocationInfo()); }
```
awelzel, line 371: If we need to keep them, maybe `SetLocationFrom` or so, but if we can move the location setting into the constructors, I think these should just. For the remaining ones the explicit version works and it's not too bad to justify the helpers on first sight.
```
other->SetLocationInfo(obj->GetLocationInfo());
```

src/Stmt.cc
```diff
@@ -1896,17 +1896,38 @@ void WhenInfo::Build(StmtPtr ws) {
...
+    one_test->SetLocation(cond);
+    two_test->SetLocation(cond);
+    empty->SetLocation(cond);
+    test_cond->SetLocation(cond);
+    do_test->SetLocation(cond);
+    else_branch->SetLocation(cond);
+    do_bodies->SetLocation(cond);
+    dummy_return->SetLocation(cond);
+    shebang->SetLocation(cond);
+    lambda->SetLocation(cond);
```
awelzel, line 1910: There might be a more compact way, but this would work (and using the more verbose Obj methods rather than the specialized SetLocation helper).
```
+    for ( const auto& e : std::vector<ExprPtr>{true_const, param, one_test, two_test, lambda} )
+        e->SetLocationInfo(cond->GetLocationInfo());
+
+    for ( const auto& s :
+          std::vector<StmtPtr>{empty, test_cond, do_test, else_branch, do_bodies, dummy_return, shebang} )
+        s->SetLocationInfo(cond->GetLocationInfo());
```

src/Stmt.cc
```diff
@@ -1933,6 +1956,13 @@ void WhenInfo::BuildInvokeElems() {
     invoke_cond = make_intrusive<ListExpr>(one_const);
     invoke_s = make_intrusive<ListExpr>(two_const);
     invoke_timeout = make_intrusive<ListExpr>(three_const);
+
+    one_const->SetLocation(cond);
+    two_const->SetLocation(cond);
+    three_const->SetLocation(cond);
+    invoke_cond->SetLocation(cond);
+    invoke_s->SetLocation(cond);
+    invoke_timeout->SetLocation(cond);
```
awelzel, line 1965: Maybe for-range loop as above?

src/StmtBase.h
```diff
@@ -144,16 +146,8 @@ class Stmt : public Obj {
...
-    // rather than a StmtPtr to emphasize that the access is read-only.
-    const Stmt* Original() const { return original ? original->Original() : this; }
-
-    // Designate the given Stmt node as the original for this one.
-    void SetOriginal(StmtPtr _orig) {
-        if ( ! original )
-            original = std::move(_orig);
-    }
+    void SetLocation(const ObjPtr& orig_obj) { SetLocation(orig_obj.get()); }
+    void SetLocation(const Obj* orig_obj) { SetLocationInfo(orig_obj->GetLocationInfo()); }
```
awelzel, line 150: Similar as in Expr: I want to suggest we don't need these on the `Stmt` themselves given they are essentially `stmt->SetLocationInfo(other->GetLocationInfo())` which is just a bit more verbose and there aren't too many call-sites in `Stmt.cc`.

I realize there's extensive usage in script_opt, but you could have a free helper `set_location(Obj *, Obj *)`, too.

src/StmtBase.h
```diff
@@ -202,11 +189,6 @@ class Stmt : public Obj {
     mutable double last_access;    // time of last execution
     mutable uint32_t access_count; // number of executions
 
-    // The original statement from which this statement was
-    // derived, if any.  Used as an aid for generating meaningful
-    // and correctly-localized error messages.
-    StmtPtr original = nullptr;
```
awelzel, line 208: Nice.

awelzel (changes requested):
Mostly wondering if we can trim this a bit.

src/Expr.cc
vpax: Unfortunately that change is at odds to with how the parser manages location information associated with AST nodes. It calls `set_location()` prior to constructing a node, which then automatically sets the node's location (in the `Obj` constructor). The parser's location is more precise than a node can obtain from its operands ilke in the example above - for example, in the expression `! foo`, the parser sets the location to start at `!` and end at the end of `foo`, whereas the example would have it only cover `foo`.

Regarding "easy to miss to keep consistent", there's an `ASSERT` in the ZAM compiler that complains if a node doesn't have _some_ sort of location associated with it (which is what happens for nodes constructed during optimization, unless they're associated with a predecessor like the logic in the PR laboriously does).

src/Expr.h
vpax: I agree that's a better name, if we keep the approach in the PR.

src/Stmt.cc
vpax: Nice! Definitely more pleasing than a whole list of calls.

src/StmtBase.h
vpax: Cool, I'm fine with streamlining this

src/Expr.cc
awelzel: > whereas the example would have it only cover foo.

Example here being the proposal I made. I think I understand now. An opposite situation currently: When an AddExpr is instantiated during parsing and we get down into `PromoteOps()`, the `ArithCoerceExpr`s created there for each of the ops will have a location covering the whole AddExpr, not the location of just the individual operands.

I have no good idea. Something that comes to mind is making location setting chain-able, but I'm not sure that just falls into the category of playing tricks to avoid the temporary/extra code.

Paging @bbannier for intuition.

In Expr.h (alternative names "WithLocationOf" or "AdoptLocationOf/From" or "CopyLocationOf").

```
+    ExprPtr UseLocationOf(const ExprPtr& other) {
+        SetLocationInfo(other->GetLocationInfo());
+        return ThisPtr();
+    }
+
```

```
--- a/src/Expr.cc
+++ b/src/Expr.cc
@@ -446,9 +446,7 @@ ExprPtr NameExpr::MakeLvalue() {
     if ( id->IsOption() && ! in_const_init )
         ExprError("option is not a modifiable lvalue");
 
-    auto r = make_intrusive<RefExpr>(ThisPtr());
-    r->SetLocation(this);
-    return r;
+    return make_intrusive<RefExpr>(ThisPtr())->UseLocationOf(ThisPtr());
 }
 
 void NameExpr::Assign(Frame* f, ValPtr v) {
@@ -1039,16 +1037,11 @@ void BinaryExpr::PromoteOps(TypeTag t) {
     if ( is_vec2 )
         bt2 = op2->GetType()->AsVectorType()->Yield()->Tag();
 
-    if ( bt1 != t ) {
-        auto orig_op1 = op1;
-        op1 = make_intrusive<ArithCoerceExpr>(op1, t);
-        op1->SetLocation(orig_op1);
-    }
-    if ( bt2 != t ) {
-        auto orig_op2 = op2;
-        op2 = make_intrusive<ArithCoerceExpr>(op2, t);
-        op2->SetLocation(orig_op2);
-    }
+    if ( bt1 != t )
+        op1 = make_intrusive<ArithCoerceExpr>(op1, t)->UseLocationOf(op1);
+
+    if ( bt2 != t )
+        op2 = make_intrusive<ArithCoerceExpr>(op2, t)->UseLocationOf(op2);
 }

Comment (awelzel) [reply src/Expr.cc:4736]: Benjamin suggested if anything to replace the class method and use a a free standing helper instead.
```
ExprPtr with_location(const ExprPtr&, Location *l);
ExprPtr with_location_of(const ExprPtr&, ExprPtr&l);

op1 = with_location(make_intrusive<ArithCoerceExpr>(op1, t), op1->GetLocationInfo())
op1 = with_location_of(make_intrusive<ArithCoerceExpr>(op1, t), op1)
```

Though neither this nor the verbose version are appealing or clear why really needed. Removing setting of an Obj's location based on parser state is likely the crux and would help disentangle (explicitly set locations `parse.y` on created nodes), but that's likely much bigger surgery.

@vpax, think I'd prefer the `with_location_of()` helper variant, but feel free to chime in if you think that's not actually helping. It's mostly just using a function argument as temporary with a bit more expressiveness :man_shrugging:

Comment (vpax) [reply src/Expr.cc:4736]: `with_location_of()` WFM, I'll set about updating when I have some cycles

**Comment by vpax** [comment]
I've now pushed changes reflecting the thread above, i.e. mainly to switch to helper functions for setting location. This has simplified the changes and in some places also the original code. I believe I've addressed all of the pending issues, but let me know if some remain.

**Comment by awelzel** [comment]
Thanks Vern. I've pushed one follow up commit removing the individual `with_location_of` implementations with this template placed into `Obj.h`.

```
+// Helper for updating e's location to the one used by o, returning e.
+template<typename T, typename U>
+T with_location_of(T e, const U& o) {
+    e->SetLocationInfo(o->GetLocationInfo());
+    return e;
+}
+
```

If that seems off to someone, happy to do a follow-up .

awelzel (approved):

PR #3584, opened 2024-01-26
Fix an MMDB error I introducd in the recent factoring

ckreibich:
Failing to check the value of `result.found_entry` essentially causes an arbitrary DB entry, depending on DB structure, to be returned.

With many thanks to @dopheide-esnet for hitting this one!


testing/btest/core/mmdb/lookup.zeek
```diff
@@ -0,0 +1,28 @@
+# @TEST-DOC: Test basic DB lookups for success/failure.
+#
+# @TEST-REQUIRES: grep -q "#define USE_GEOIP" $BUILD/zeek-config.h
```
awelzel, line 3: We have `zeekl-config --have-geoip` now, no? :-)

awelzel (approved):

testing/btest/core/mmdb/lookup.zeek
ckreibich: Oh yeah totally. :) That spot is what originally prompted me to add it. But last night it was close to midnight and I wanted to ship this thing. Happy to work it over today.

PR #3591, opened 2024-01-31
spicy/diff-remove-timestamp: Fix missing -e

awelzel:
This got lost when converting to sed -E.

### DIFF
--- a/testing/scripts/spicy/diff-remove-timestamps
+++ b/testing/scripts/spicy/diff-remove-timestamps
@@ -2,4 +2,4 @@
 #
 # Replace anything which looks like timestamps with XXXs (including the #start/end markers in logs).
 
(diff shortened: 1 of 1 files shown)


awelzel:
I went through 73458927ba4293d6260b8ab1995732c8183ff96c and this is the only `-e` that mistakenly removed. I haven't looked much further, but we have two copies of diff-remove-timestamp and the one from `spicy/` is the one that's installed, but the one further up is the one that's used during tests IIUC.

bbannier:
Interesting that we use an installed file instead of bundling something with the package, I guess that was for convenience. It exposes packages to this kind of issue though which isn't great.

awelzel:
> Interesting that we use an installed file instead of bundling something with the package, I guess that was for convenience. It exposes packages to this kind of issue though which isn't great.

There's something else off, let me take another look.

And agree on the re-using, don't think we should encourage that.

awelzel:
The spicy-dns installation fails due to baseline changes for error reporting:

Screenshot from 2024-01-31 14-25-47

That requires normalization/fixup in that repo.

PR #3598, opened 2024-02-03
Do not default to proxy nodes in `Broker::PoolSpec`

ckreibich:
This requires pool creation to spell out a spec explicitly, which the only code using these types already does. There's no reason for pools to automatically refer to proxies.

I came upon this via @ynadji's comment on proxies in the extracted Zeekygen docs [on Slack](https://zeekorg.slack.com/archives/CSZBXF6TH/p1706892270800569) — this change will help with that too.

### DIFF
--- a/scripts/base/frameworks/cluster/pools.zeek
+++ b/scripts/base/frameworks/cluster/pools.zeek
@@ -23,7 +23,7 @@ export {
 	## A pool specification.
 	type PoolSpec: record {
 		## A topic string that can be used to reach all nodes within a pool.
-		topic: string &default = "";
+		topic: string;
 		## The type of nodes that are contained within the pool.
 		node_type: Cluster::NodeType;
 		## The maximum number of nodes that may belong to the pool.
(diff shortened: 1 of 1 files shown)


scripts/base/frameworks/cluster/pools.zeek
```diff
@@ -25,7 +25,7 @@ export {
 		## A topic string that can be used to reach all nodes within a pool.
 		topic: string &default = "";
```
awelzel, line 26: Have you considered un-defaulting this one, too? I'm not versed with the topics. Would not changing this and subscribing to an empty topic be a wild-card match, or even an error?

awelzel (approved):

scripts/base/frameworks/cluster/pools.zeek
ckreibich: I did! I was torn. I do think it would be better not to default it, and in terms of our current `PoolSpec` use it's just as safe/risky to change it as `node_type`. But the semantics of empty topics aren't clear to me either, so I didn't go near it.

I did some digging today and the details might only start to matter down in CAF — I couldn't find explicit treatment or objection to empty topics at the higher levels. But it's clear that the topic is not meant to remain empty, so I come out in favor of un-defaulting it.

awelzel (approved):

PR #3600, opened 2024-02-05
Netcontrol updates - better logging for catch and release, new hook

0xxon:
This PR has a couple of small updates for NetControl.

It extends the logging for catch-and-release, adding more information in a variety of cases that were not logged before; it also logs the plugin name for operations for which this makes sense.

This PR also adds a new NetControl::rule_added_policy hook, which can be used to change rules when they are added to NetControl. this can be, e.g., used in situations where rules have to be synced between several plugins, with dependencies on the plugin states.


awelzel (approved):
It's not possible for me to make sense of the test+baselines combination, but the changes make sense on a high level :man_shrugging: 

Out of curiosity, do you know many users of this framework exists?

scripts/policy/frameworks/netcontrol/catch-and-release.zeek
```diff
@@ -270,11 +272,40 @@ event rule_added(r: Rule, p: PluginState, msg: string)
...
+event rule_exists(r: Rule, p: PluginState, msg: string)
+	{
+	if ( !cr_check_rule(r) )
+		return;
+
+	local ip = subnet_to_addr(r$entity$ip);
+	local bi = blocks[ip];
+
+	local log = populate_log_record(ip, bi, INFO);
+	log$plugin = p$plugin$name(p);
```
awelzel, line 290: The only thought I had was to add a new `populate_named_log_record(p: PluginState, ...)` or so that calls the original `populate_log_record()` and does the enrichment, but not sure it's worth it.

0xxon:
Thank you :)

I sadly do not know of a lot of people who use this. The main known user is @initconf. They use it _very_ extensively, managing at least thousands of blocks with it, with a quite active turnover.

So - this code does get a lot of regular exercise.

I think the reason for this being not more popular is that it's a bit hard to integrate in your infrastructure - and you have to be ok with doing automatic blocking.

PR #3601, opened 2024-02-05
Fix build of Alpine CI image

timwoj:
Alpine has similar issues as Debian 12 with regards to python packages. This changes how we install them to be consistent with the other images. It also removes a file that was mistakingly committed.


awelzel (approved):

ci/fedora-38/Dockerfile~
bbannier, line 1: Would it make sense to add e.g., `*~` to a global `.gitignore`?

awelzel: No too strong opinion, but then .swo, .swp, too?

I have this in my global gitignore:
```
$ cat ~/.gitignore
.vscode
*.swp
*.swo
*~
```

bbannier: I do have some stuff in there as well, but was more wondering about the `.gitignore` at the root of the repo. There is already vscode-specific stuff in there.

PR #3602, opened 2024-02-06
Stmt: Introduce STMT_EXTERN

awelzel:
It's currently possible for plugin's to implement their own statement
subclasses and override the Exec() implementation. This has been leveraged
by ZeekJS [1] and zeek-perf-support [2] as well as a private WASM plugin.
All of these used STMT_ANY as the tag of their own statement subclasses.

With STMT_EXTERN, we make the possibility to add external code into the AST
somewhat more supported. It's all in detail space and plugin authors have
no guarantee for stability, but it seems such a powerful extension point
that IMO we should keep it.

I'm conscious there's the broader topic how this interacts with ZAM
optimization like in-lining or rewriting of statements. However, this
already applies to the STMT_ANY usage of the mentioned plugins.

[1] https://github.com/corelight/zeekjs
[2] https://github.com/zeek/zeek-perf-support

### DIFF
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -52,6 +52,7 @@ const char* stmt_name(StmtTag t) {
         "ZAM",
         "null",
         "assert",
+        "extern",
     };
 
     return stmt_names[int(t)];
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -55,6 +55,12 @@ const char* stmt_name(StmtTag t) {
         "extern",
     };
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+    if ( int(t) == STMT_ANY )
+        return "any";
+#pragma GCC diagnostic pop
+
     return stmt_names[int(t)];
 }
 
--- a/src/StmtEnums.h
+++ b/src/StmtEnums.h
@@ -31,7 +31,8 @@ enum StmtTag {
     STMT_ZAM,           // a ZAM function body
     STMT_NULL,
     STMT_ASSERT,
-#define NUM_STMTS (int(STMT_ASSERT) + 1)
+    STMT_EXTERN, // for custom Stmt subclasses provided by plugins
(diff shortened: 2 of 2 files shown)


awelzel:
Alternative name suggestion would be `STMT_PLUGIN`, but I liked "extern" a bit better. Open to other suggestions.

@vpax - any general thoughts if this will be major hindrance for ZAM down the road? I tried to lay out the motivation in the description, it would be nice to keep the extension mechanism, but use ZAM optimizations (I recently saw some perf numbers from Justin of interpreted and ZAM execution... and oh my!)

vpax:
@awelzel Huh those sorts of "off-label" usages of `STMT_ANY` weren't on my radar, and the fact that its `enum` is assigned to `-1` had led me to assume it's not in use. If it's kosher to use it, we should add a BTest which does so, as I expect that presently it will crash ZAM.

Regarding `STMT_EXTERN` - which I agree is a better way to handle the need - I can't see any way that ZAM can support compiling ASTs that include those (unsurprisingly) ... _however_ there's already machinery for deciding "this function body is not compilable" (which was handy back when ZAM wasn't feature-complete), so by adding recognition of `STMT_EXTERN` to that predicate, ZAM will know to skip such bodies. It has logic to still Do Its Thing for all the bodies that _are_ compilable, so this should work okay in terms of interoperability. (Similarly for `-O gen-C++`, which already knows to skip function bodies that are influenced by `@if` conditionals.)

awelzel:
> I can't see any way that ZAM can support compiling ASTs that include those (unsurprisingly)

This PR is meant to move the unsurprising part over to: Yes, it's a possibility for plugins to inject custom statements. For ZeekJS, it's adding to `Func.body` top-level statement and skipping the whole body the right thing.


@rsmmr - assigning you for an intuition if that change opens up too many questions, or is a reasonable step away from the STMT_ANY misuse.

ckreibich:
Heh — well, I just saw you're actually already deprecating `STMT_ANY`, so that addresses my question.

PR #3605, opened 2024-02-07
Do not use snprintf() for hex printing

awelzel:
I've been staring at native/script flamegraphs for the zeek-perf-support demo a bit and snprintf() showing up caught my eye. bytestring_to_hexstr() also has an extra copy of the resulting string that can be removed. Tested with pure SSL traffic, looks to be around ~3% so seems worth doing.

|:---|---:|---:|---:|---:|                    
| `taskset -c 0,1 zeek  -C -r ./much-alexa-https-top-100.pcap -f "port 443" Log::default_writer=Log::WRITER_NONE` | 5.201 ± 0.037 | 5.156 | 5.270 | 1.00 |
| `taskset -c 0,1 /opt/zeek-dev-prod/bin/zeek  -C -r ./much-alexa-https-top-100.pcap -f "port 443" Log::default_writer=Log::WRITER_NONE` | 5.392 ± 0.034 | 5.349 | 5.455 | 1.04 ± 0.01 |

### DIFF
--- a/src/net_util.cc
+++ b/src/net_util.cc
@@ -8,7 +8,9 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <memory>
 
+#include "zeek/3rdparty/doctest.h"
 #include "zeek/IP.h"
 #include "zeek/IPAddr.h"
 #include "zeek/Reporter.h"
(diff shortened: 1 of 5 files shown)


JustinAzoff:
Speaking of printf for hex, I've also noticed how this is done in `fmt_mac`, which is called twice per connection, even though often nothing ever looks at the result.

awelzel:
> Speaking of printf for hex, I've also noticed how this is done in fmt_mac, which is called twice per connection, even though often nothing ever looks at the result.

I've pushed one more commit and it's improving syn-only perf by 1-3% (-b base/protocols/conn and ZAM so more visible...). I'm not quite sure that's really worth it, but it adds a few doctests, so maybe that counts as improvement, too :-)

|:---|---:|---:|---:|---:|
| `taskset -c 0,1 zeek -OZAM -b base/protocols/conn  -C -r 1mio-syn-packets.pcap  Log::default_writer=Log::WRITER_NONE` | 14.300 ± 0.055 | 14.219 | 14.341 | 1.00 |
| `taskset -c 0,1 /opt/zeek-dev-prod/bin/zeek -OZAM -b base/protocols/conn  -C -r 1mio-syn-packets.pcap  Log::default_writer=Log::WRITER_NONE` | 14.676 ± 0.066 | 14.612 | 14.768 | 1.03 ± 0.01 |

awelzel:
@bbannier was asking if the std::unique_ptr is really worth it or whether std::string would be good enough. Forth and back, I'm going to leave it in as while marginal, is reproducibly better.

src/net_util.cc
```diff
@@ -157,21 +159,69 @@ const char* fmt_conn_id(const uint32_t* src_addr, uint32_t src_port, const uint3
...
-                 m[7]);
+    int elen = (len == 6 || (m[6] == 0 && m[7] == 0)) ? 6 : 8;
+    auto slen = 2 * elen + (elen - 1);
+    auto buf = std::make_unique<uint8_t[]>(slen + 1);
+    auto* bufp = reinterpret_cast<char*>(buf.get());
+
+    for ( int i = 0; i < elen; i++ ) {
+        zeek::util::bytetohex(m[i], &bufp[i * 2 + i]);
+        if ( i < elen - 1 )
+            bufp[i * 2 + i + 2] = ':';
```
JustinAzoff, line 220: I wonder if there is an easy way to pre-fill the buffer with ':' when it is allocated, then the `if` and assignment can be removed.  I doubt it matters much performance wise but it would it be simpler that way?

awelzel: Might be easier to read, yeah. Pre-populating a large blob (not applicable here) could be significant, so will not do until others have strong opinions :-)

src/zeek.bif
```diff
@@ -3175,14 +3175,16 @@ function bytestring_to_hexstr%(bytestring: string%): string
...
 
-	hexstr[0] = 0;
 	for ( zeek_uint_t i = 0; i < len; ++i )
-		snprintf(hexstr + (2 * i), 3, "%.2hhx", bytes[i]);
+		zeek::util::bytetohex(bytes[i], &hexstr[2 * i]);
+	hexstr[2 * len] = 0;
 
-	return zeek::make_intrusive<zeek::StringVal>(hexstr);
+	// Adopt the memory...
+	String* s = new String(1, reinterpret_cast<uint8_t*>(hexstr_buf.release()), 2 * len);
```
timwoj, line 3186:
```suggestion
	String* s = new String(true, reinterpret_cast<uint8_t*>(hexstr_buf.release()), 2 * len);
```

src/net_util.cc
timwoj: You can use `std::fill`, but I don't know what the performance implications of that are. `memset` might be enough.

src/net_util.cc
```diff
@@ -157,21 +159,69 @@ const char* fmt_conn_id(const uint32_t* src_addr, uint32_t src_port, const uint3
...
+    CHECK(len4 == 2 * 8 + 7);
+    CHECK(memcmp(buf4.get(), "01:02:03:04:05:06:07:08", len4 + 1) == 0);
+}
+
+std::pair<std::unique_ptr<uint8_t[]>, int> fmt_mac_bytes(const unsigned char* m, int len) {
     if ( len < 8 && len != 6 ) {
-        *buf = '\0';
-        return buf;
+        auto buf = std::make_unique<uint8_t[]>(1);
+        buf.get()[0] = '\0';
```
timwoj, line 208: The `.get()` shouldn't be needed here. `unique_ptr` implements `operator[]` itself.

timwoj (changes requested):

src/zeek.bif
awelzel: Done, and also for the other occasions in Conn.cc

timwoj (approved):

PR #3612, opened 2024-02-21
signatures: Fix ISO 9960 signature

awelzel:
Reference: https://community.zeek.org/t/zeek-installation-failing-to-mime-type-an-iso-file/7289/2


timwoj (approved):
The magic numbers (the byte sizes) could maybe use a comment, but looks fine to me otherwise.

PR #3613, opened 2024-02-21
Clear ftp.log fuid after logging

awelzel:
This was reported through Corelight channels and also recently came up in the forum:

https://community.zeek.org/t/ftp-fuid-logged-for-wrong-data-connection/7276

```
A user reported being confused about the fuid association of subsequent
FTP commands when a data transfer has completed. It seems reasonable to
unset fuid upon logging a FTP command which had a fuid.

The current behavior results in the PORT or PASV commands after a RETR or STOR
to have the fuid of the prior file transfer. Similarly, any CWD or DEL commands
following a file transfer will unnecessarily be logged with the fuid of the
prior file transfer.

This tickles the baselines for the private testing PCAP a lot, primarily
because there data connections in that pcap are never established properly.
E.g, the fuids FzDzid1Dxm9srVKHXf and FEfYX73q5C6GEQZXX9 have been re-used
for multiple commands.

This may look like we're losing information, but the fuids vanishing
in the normal btests belong to a LIST command that isn't logged by
default into ftp.log. If it was, the fuid would be attached to it.
```

Private baseline updates:

https://github.com/zeek/zeek-testing-private/pull/20


awelzel:
I could see the clearing being toggle'ble via an option if this appears too controversial, but the overzealous fuid association seems pretty misleading actually, so would avoid that route if possible.

scripts/base/protocols/ftp/main.zeek
```diff
@@ -212,8 +212,9 @@ function ftp_message(c: connection)
 	# values after logging.
 	delete s$mime_type;
 	delete s$file_size;
-	# Same with data channel.
+	# Same with data channel and fuid;
```
timwoj, line 215: Remove this comment and change the one above to say something like "The following fields are specific".

timwoj (approved):
The approach sounds fine to me. We shouldn't be logging an fuid for commands not associated with the file. I had one nit about a comment.

PR #3615, opened 2024-02-21
zeek.bif: Add load_path() helper

awelzel:
Relates to #3594. This helper can be used to determine the path that will be used by @load, if at all. Usage example:

    @if ( load_path("my-package") != "" )
    @load my-package
    @endif


testing/btest/bifs/load_path.zeek
```diff
@@ -0,0 +1,49 @@
...
+		return r;
+	local parts = split_string(r, /\//);
+	return join_string_vec(parts[-4:], "/");
+	}
+
+print "load_path base/protocols/conn", path_tail(load_path("base/protocols/conn"));
+print "load_path protocols/conn (empty expected, no __load__.zeek)", load_path("protocols/conn");
+print "load_path protocols/conn/vlan-logging", path_tail(load_path("protocols/conn/vlan-logging"));
+
+print "load_path pkg1", load_path("pkg1");
```
bbannier, line 32: Can you sprinkly in some tests which exercise relative loads (involving `./`, `../`, a mix of them)?

awelzel: The relative loads are dependent on the parsing state (which file is currently being parsed?) and not a great fit for the BiF. Though, that make me wonder if there's a better name, or the docs need to emphasize this.

awelzel:
Renamed to find_in_zeekpath()

PR #3616, opened 2024-02-21
zeek.bif: Add find_in_zeekpath() helper

awelzel:
Relates to #3594. This helper can be used to determine the path that will be used by @load, if at all. Usage example:

    @if ( find_in_zeekpath("my-package") != "" )
    @load my-package
    @endif

Replaces #3615, renaming and adding error on absolute and relative paths.


src/zeek.bif
```diff
@@ -5328,3 +5328,35 @@ function table_pattern_matcher_stats%(tbl: any%) : MatcherStats
 
 	return std::move(result);
 	%}
+
+## Determine the path used by a non-relative @load directive.
+##
+## This function is package aware: Passing *package* will yield the
+## path to *package.zeek*, *package/__load__.zeek* or an empty string
+## if neither can be found. Note that passing a relative path is an
```
timwoj, line 5336: Should this be "passing a relative or absolute path"? The if statement below is checking for a path starting with `/` and has a similar error message.

testing/btest/bifs/find_in_zeekpath.zeek
```diff
@@ -0,0 +1,59 @@
...
+	local parts = split_string(r, /\//);
+	return join_string_vec(parts[-4:], "/");
+	}
+
+print "find_in_zeekpath base/protocols/conn", path_tail(find_in_zeekpath("base/protocols/conn"));
+print "find_in_zeekpath protocols/conn (empty expected, no __load__.zeek)", find_in_zeekpath("protocols/conn");
+print "find_in_zeekpath protocols/conn/vlan-logging", path_tail(find_in_zeekpath("protocols/conn/vlan-logging"));
+
+print "find_in_zeekpath pkg1", find_in_zeekpath("pkg1");
+print "find_in_zeekpath pkg1.zeek", find_in_zeekpath("pkg1");
```
timwoj, line 37:
```suggestion
print "find_in_zeekpath pkg1.zeek", find_in_zeekpath("pkg1.zeek");
```

timwoj (changes requested):

JustinAzoff:
I still like the idea of

```
@if ( can_load("the-package") )
@load the-package
@endif
```

Comparing the return value to "" seems a bit awkward.  Is there a use case where the resulting path is useful?

awelzel:
> Comparing the return value to "" seems a bit awkward. Is there a use case where the resulting path is useful?

Yeah, it would look better. It mostly came from the angle that `@load` also works for relative paths, too and a `can_load()` being over-promising here. As for the return value, my thought was the next possibly question would be what the actual path ended up being (could be interesting if your ZEEKPATH became very complex, or want to determine installation directories based on init-bare.zeek's location or some such tricks), so why not yield that instead.

I do agree that the new helper doesn't make the `can_load()` use-case obvious, but that also seemed relatively niche/advanced enough to allow for the `!= ""` awkwardness. If you/anyone feels much stronger, I'll add it.

> It's just needs a BiF - why not support both?

Given `find_in_zeekpath()`, `can_load()` seemed small enough to be provided by an external script. Though, not sure I understand your comment fully.

timwoj (approved):
:shipit:

PR #3617, opened 2024-02-21
Allow delete statement for tables, sets and vectors

awelzel:
Relates to #3472. This allow "delete tbl" as an alternative for clear_table(tbl). Also works for vectors.


timwoj (approved):
LGTM. While reviewing, I found this comment on `VectorVal::Resize` in Val.h. I don't think it's true anymore, since we just store vectors as `std::vector` internally.

```
    // Is there any way to reclaim previously-allocated memory when you
    // shrink a vector?  The return value is the old size.
```

One thought is that `Resize()` could short-cut through `std::vector::clear()` if size is zero, but I don't know the ramifications for the managed ZAM code there.

vpax:
> One thought is that `Resize()` could short-cut through `std::vector::clear()` if size is zero, but I don't know the ramifications for the managed ZAM code there.

@timwoj which managed ZAM code do you have in mind? I'm not seeing an issue off-hand.

timwoj:
> @timwoj which managed ZAM code do you have in mind? I'm not seeing an issue off-hand.

Oh wait, scratch that. I was looking at something else (I searched for `VectorVal::Re` and `Remove` came up before `Resize` 🤦🏼 ). I think a shortcut through `clear()` should be fine.

vpax:
> Oh wait, scratch that. I was looking at something else

Gotcha. FYI, the code in `Remove` is about `ZVal` stuff. That's not in fact ZAM-specific but now a standard part of the `Val` internals.

awelzel:
> One thought is that Resize() could short-cut through std::vector::clear() if size is zero

I didn't do this. As of memory reclaiming, there's `shrink_to_fit()` and maybe that could make sense when resizing to 0. It's much dependent on usage patterns though, so :man_shrugging:

Issue #3620, opened 2024-02-23
Mozilla CA list outdated


AmalieDue:
It seems like the Mozilla CA list (mozilla-ca-list.zeek) and the Google CT list (ct-list.zeek) are not up to date. They have not been updated in 10 months. Furthermore, it seems like the tools used to update the lists are not publicly available. @0xxon, can the lists be updated? And is it possible to get access to the tools used to update the lists, or do I need to write my own?  

Thanks!

0xxon:
Hi,

sorry - you are right, I forgot to do that for the current release. I will remedy this shortly - and we will put it onto the release checklist.

The tool that we use for conversion is publically available here: https://github.com/zeek/zeek-aux/blob/master/devel-tools/gen-mozilla-ca-list.rb

The current best way to use it is to download a recent release of NSS, and point it at the contained certdata.txt.

0xxon:
I updated the process description in https://github.com/zeek/zeek/wiki/Release-Process and added the generation of the files to the release checklist.

PR #3621, opened 2024-02-23
Update Mozilla CA and Google CT list and related tests.

0xxon:
Test updates are necessary due to removed CT logs.

Fixes GH-3620.

This should make it into 6.2.0 if possible.


awelzel (approved):

awelzel:
> This should make it into 6.2.0 if possible.
cc @timwoj 

Have also put it on the 6.0.x list.

0xxon:
Looking at this, I am actually a tad unsure if we want to pull it into 6.0. I can see an argument for it - but it also changes behavior in ways that not everyone might expect, and we have traditionally not updated this in LTS releases after they came out.

Issue #3622, opened 2024-02-23
State explosion in regex engine when using huge repetition counts


bbannier:
I am seeing a stack overflow on my system when running on a simple PCAP, e.g.,

```console
$ zeek -Cr testing/btest/Traces/ssh/single-conn.trace
[3]    34066 segmentation fault  zeek -Cr testing/btest/Traces/ssh/single-conn.trace
```

I bisected this to commit d2409dd43260e9e909fb0acc323817ebde76da8b where just having `file-magic /^.{32769}CD001/` for `file-iso9660`'s signature is sufficient to trigger this.

Looking at the backtrace I see a total of about 200k frames repeating

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603ff0)
  * frame #0: 0x00000001009e5744 zeek`zeek::detail::EpsilonState::~EpsilonState(this=<unavailable>) at NFA.h:79
    frame #1: 0x00000001009e5788 zeek`zeek::detail::EpsilonState::~EpsilonState(this=0x000060001a24dea0) at NFA.h:79:7
    frame #2: 0x0000000100003b9c zeek`zeek::Unref(o=0x000060001a24dea0) at Obj.h:176:9
    frame #3: 0x0000000100cde354 zeek`zeek::detail::NFA_State::~NFA_State(this=0x00006000161768f0) at NFA.cc:51:13
    frame #4: 0x0000000100cde3e4 zeek`zeek::detail::NFA_State::~NFA_State(this=0x00006000161768f0) at NFA.cc:48:25
    frame #5: 0x0000000100cde410 zeek`zeek::detail::NFA_State::~NFA_State(this=0x00006000161768f0) at NFA.cc:48:25
    frame #6: 0x0000000100003b9c zeek`zeek::Unref(o=0x00006000161768f0) at Obj.h:176:9
    frame #7: 0x0000000100cde354 zeek`zeek::detail::NFA_State::~NFA_State(this=0x000060001a24ddb0) at NFA.cc:51:13
    frame #8: 0x00000001009e57b8 zeek`zeek::detail::EpsilonState::~EpsilonState(this=0x000060001a24ddb0) at NFA.h:79:7
    frame #9: 0x00000001009e575c zeek`zeek::detail::EpsilonState::~EpsilonState(this=0x000060001a24ddb0) at NFA.h:79:7
[...]

PR #3623, opened 2024-02-23
Prevent regex state in iso9660 signature.

bbannier:
Regex matches with huge repetition counts seem to trigger a state explosion in the regex engine which can lead to e.g., stack overflows and also seems to slow down performance significantly.

This patch works around this by not restricting the repetition count. This should lead to strictly more false positives, but at least makes this signature useful and not disruptive.

Relates to #3622.


awelzel:
Uff, as  posted in #3622, `/^.{34817}CD001/` also tanks performance for HTTP only pcaps really badly. @rsmmr , @vpax, do you have an intuition what happens internally? Is it expected that large repetition values are not good?

> This patch works around this by not restricting the repetition count. This should lead to strictly more false positives, but at least makes this signature useful and not disruptive.

It's actually not useful unless a user bumps default_file_bof_buffer_size to something that would accomodate the 32k offset (default is 4k). I'm tempted to suggest we remove the signature completely and point that out in NEWS?

Thoughts?

vpax:
Any use of `{` _count_ `}` requires _count_ states (at least) to do the counting, so this isn't surprising. For these particular uses, certainly it'd be much better to have a bit of logic around signature match size to finalize the decision.

scripts/base/frameworks/files/magic/general.sig
```diff
@@ -299,15 +299,27 @@ signature file-windows-minidump {
...
+# at offset 1: 16 * 2048 + 1 = 32769
 signature file-iso9660 {
         file-mime "application/x-iso9660-image", 99
-        file-magic /^.{32769}CD001/
+
+        # We know we expect the magic string at byte 34817, but cannot use a
+        # quantifier here since it would trigger state explosion in the regex
+        # engine, see #3622.
+        #
+        # TODO(bbannier): match for `/^.{34817}CD001/` once the regex engine is fixed.
```
awelzel, line 310: I'd be fine with commenting out the whole signature as the normal case won't work by default anyhow due to the limited bof buffer, leaving a comment that it can be done using separate scripts. Doesn't sound there is a good fix and the `.*` does result in the full bof being scanned for every file I believe.

awelzel: Benjamin, want me to pick this up, or you good with commenting them out completely, too?

I'm also thinking, we could place the signature and a `default_file_bof_buffer_size` redef into `policy/frameworks/signatures` to provide an easy (and functional) way for users to load it if they really want without having it in `base/`. It's not optimal, but seems better and more explicit than what we had in the past.

bbannier: Feel free to take this over. From an outside perspective I would have expected signatures to be independent from whatever `default_file_bof_buffer_size` was set to, and for signature matching to possibly provide gradually improve as more data becomes available.

bbannier:
Superseded by #3627.

PR #3628, opened 2024-02-27
spicy/zeekygen: Remove mtime from generated code

awelzel:
Zeekygen implements its own make-style update logic to prevent re-creation of files that have not changed. To fulfill this, we currently encode the current time into spicyz generated .cc files.

This degrades ccache efficiency for built-in analyzers and also for all .evt files compiled during testing. Switch SpicyModuleInfo to return current time instead. This results in the re-generation of documentation files unconditionally when running Zeekygen, but that seems more acceptable IMO.

Generally wonder if Zeekygen should produce output unconditionally and if we need to clobber prevention, compare with the content of the existing file.


src/zeekygen/SpicyModuleInfo.h
```diff
@@ -45,13 +45,12 @@ class SpicyModuleInfo : public Info {
     void AddBifItem(const std::string& id, plugin::BifItem::Type type) { bif_items.emplace_back(id, type); }
 
 private:
-    time_t DoGetModificationTime() const override { return mtime; }
+    time_t DoGetModificationTime() const override { return time(nullptr); }
```
bbannier, line 48: Do you have a feel for how the modification time as returned by `GetModificationTime` is used? I see only a few implementations of `DoGetModificationTime` and no users of `GetModificationTime` this repo and I wonder if we have code generators or consumers elsewhere.

Interestingly `zeek::zeekygen::detail::PackageInfo` uses an [implementation which sometimes returns the start of the epoch](https://github.com/zeek/zeek/blob/961bac02202a146432f9edc94f1f6e97079db386/src/zeekygen/PackageInfo.cc#L47-L48) and if possible here I'd think that be preferrable to something which is always interpreted as outdated.

awelzel: > and no users of GetModificationTime in this repo and I wonder if we have code generators or consumers elsewhere.

Individual targets call `IsUpToDate()` and if the input files haven't changed (based on mtime), they don't produce output.

https://github.com/zeek/zeek/blob/961bac02202a146432f9edc94f1f6e97079db386/src/zeekygen/Manager.h#L245-L268


> Interestingly zeek::zeekygen::detail::PackageInfo uses an [implementation which sometimes returns the start of the epoch](https://github.com/zeek/zeek/blob/961bac02202a146432f9edc94f1f6e97079db386/src/zeekygen/PackageInfo.cc#L47-L48) and if possible here I'd think that be preferrable to something which is always interpreted as outdated.

Only if there's no readme text, so that seems okay.

Returning 0 is semantically not right: If an `.evt` file is modified, it would not re-generate documentation involving that .evt file. The flip side, with this change we now always re-generate the involved documentation.

I *think* this all comes from `sphinx-build` also being `mtime` based and not touching existing files will prevent re-regeneration of HTML/PDF downstream. Which I can see being nice, but I don't like the complexity, that's why I made the comment of:


> Generally wonder if Zeekygen should produce output unconditionally and if we need to clobber prevention, compare with the content of the existing file.


 I could maybe see returning the mtime of the involved `.hlto` file or the Zeek executable, but it'll look "a bit ugly" :-)

bbannier: > Individual targets call `IsUpToDate()` and if the input files haven't changed (based on mtime), they don't produce output.

Ah, this makes a lot of sense. I missed this since it was in a template.

> Returning 0 is semantically not right: If an `.evt` file is modified, it would not re-generate documentation involving that .evt file. The flip side, with this change we now always re-generate the involved documentation.

Makes sense.

bbannier (approved):
This seems like a reasonable approach to me. We might now need slightly more time to rebuild the documentation (I'd suspect that this is rare), but the saving time during compilation seems more than outweigh that.

PR #3630, opened 2024-02-28
RE/Specific_RE_Matcher: Make Compile() return false for syntax errors

awelzel:
When creating RE_Matcher instances at runtime and verifying the pattern
compiles via Compile(), the syntax_error flag wasn't respected and
Compile() would return true even for some invalid regular expressions.

For example, compiling /a{1,b}/, Compile() would return true even though
it produced a reporter error while parsing complaining about b not
being valid.

This patch improves the error handling, so that calling Compile() returns
false whenever zeek::detail::synerr() was called while a pattern was
parsed. The use-case is creation of patterns at runtime based on
JavaScript strings. These might be entered or received at runtime via
an API. This change allows to be a bit more robust to detect invalid
input and raising exceptions to notify the user.

This also move syntax_error and csize out of global scope.

If RE_Matcher was to be used as an actual API, we likely should squelch
the reporter errors and mark it as not thread safe, but this is a small
step forward.

### DIFF
--- a/src/RE.cc
+++ b/src/RE.cc
@@ -27,6 +27,8 @@ extern void RE_done_with_scan();
 namespace zeek {
 namespace detail {
 
+extern bool re_syntax_error;
+
 Specific_RE_Matcher::Specific_RE_Matcher(match_type arg_mt, bool arg_multiline)
     : mt(arg_mt), multiline(arg_multiline), equiv_class(NUM_SYM) {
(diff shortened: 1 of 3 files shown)


awelzel:
@timwoj , I realized I had this in draft. Mind looking a bit? Not for 6.2, just came up while trying to integrate Zeek's pattern with JavaScript and vice versa.

PR #3631, opened 2024-02-29
Port Spicy integration to new AST API.

rsmmr:
This goes with https://github.com/zeek/spicy/pull/1462.


src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -280,7 +280,7 @@ public type LongHeaderPacket = unit {
 };
 
 # A QUIC Frame.
-public type Frame = unit(header: LongHeaderPacket, from_client: bool, crypto_sink: sink) {
+public type Frame = unit(header: LongHeaderPacket, from_client: bool, inout crypto_sink: sink) {
```
bbannier, line 283: It seems that before passing an immutable value was fine since our refs implemented interior mutability (i.e., a const ref did seem to allow mutable access to the wrapped data) ... is that correct? If it is, I am not sure I saw something explicitly checking the previous or newer semantics explicitly in a test, but I think that would be great to have.

Separately I wonder whether it would make sense to more generally allow `copy` parameters in Spicy like we already do in HILTI. That would allow users to use narrower semantics (i.e., to specify that something is passed to be modified in a function body, but not require outside-visible changes). That might not be the semantics needed here, but I'd imagine it could be useful elsewhere.

src/spicy/spicyz/driver.cc
```diff
@@ -31,28 +32,45 @@ using Driver = ::zeek::spicy::Driver;
...
+    }
 
-        if ( module == hilti::ID("hilti") || module == hilti::ID("spicy_rt") || module == hilti::ID("zeek_rt") )
+    void operator()(hilti::declaration::Type* n) final {
+        if ( module.empty() || module == hilti::ID("hilti") || module == hilti::ID("spicy_rt") ||
+             module == hilti::ID("zeek_rt") )
             return;
 
+        assert(! n->type()->type()->typeID() ||
+               n->type()->type()->typeID() == hilti::ID(module, n->id())); // ensure consistent IDs
```
bbannier, line 59: Yikes, all the `type`s are quite a mouthful 😬

src/spicy/spicyz/driver.cc
```diff
@@ -195,6 +213,8 @@ hilti::Result<hilti::Nothing> Driver::compile() {
 
     SPICY_DEBUG("Running Spicy driver");
 
+    _error = hilti::Nothing();
```
bbannier, line 216: Is it correct that any previous `_error` value would only matter if `compile` is invoked multiple times which shouldn't happen? In that case I'd simply return an error here instead of silently resetting it, e.g.,
```suggestion
    if ( _error )
        return hilti::result::Error(hilti::util::fmt("compilation triggered with previous error: %s", _error.error()));
```

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -134,23 +131,24 @@ static std::string extract_string(const std::string& chunk, size_t* i) {
...
 
     if ( token == hilti::ID("uint64") )
-        return hilti::type::UnsignedInteger(64);
+        return builder->typeUnsignedInteger(64);
 
     throw ParseError("mismatching type");
 }
 
-static hilti::type::function::Parameter extract_parameter(const std::string& chunk, size_t* i) {
+static std::shared_ptr<hilti::type::function::Parameter> extract_parameter(Builder* builder, const std::string& chunk,
```
bbannier, line 150: Can we use the type alias in `declaration` here? It resolves to exactly the same type.

```suggestion
static hilti::type::function::ParameterPtr extract_parameter(Builder* builder, const std::string& chunk,
```

src/spicy/spicyz/glue-compiler.h
```diff
@@ -100,25 +101,25 @@ struct SpicyModule {
     std::set<hilti::rt::filesystem::path> evts; /**< EVT files that refer to this module. */
 
     // Generated code.
-    std::optional<hilti::Module> spicy_module; /**< the ``BroHooks_*.spicy`` module. */
+    std::shared_ptr<hilti::declaration::Module> spicy_module; /**< the ``BroHooks_*.spicy`` module. */
```
bbannier, line 104:
```suggestion
    hilti::ModulePtr spicy_module; /**< the ``BroHooks_*.spicy`` module. */
```

src/spicy/spicyz/glue-compiler.h
```diff
@@ -127,8 +128,8 @@ struct Event {
     // TODO: The following aren't set yet.
 
     // Code generation.
-    std::optional<::spicy::declaration::UnitHook> spicy_hook; /**< The generated Spicy hook. */
-    std::optional<hilti::declaration::Function> hilti_raise;  /**< The generated HILTI raise() function. */
+    std::shared_ptr<::spicy::declaration::UnitHook> spicy_hook; /**< The generated Spicy hook. */
```
bbannier, line 131:
```suggestion
    ::spicy::type::unit::item::UnitHookPtr spicy_hook; /**< The generated Spicy hook. */
```

src/spicy/spicyz/glue-compiler.h
```diff
@@ -162,6 +163,18 @@ class GlueCompiler {
...
+     * Returns the AST context in use. Only available once the driver has
+     * initialized the glue compiler.
+     */
+    auto context() { return _driver->context()->astContext().get(); }
+
+    /**
+     * Returns the AST builder in use. Only available once the driver has
+     * initialized the glue compiler.
+     */
+    auto builder() { return static_cast<::spicy::Builder*>(_driver->builder()); }
```
bbannier, line 176: Nit: This made me briefly wonder whether `context` or `builder` returned values instead of references. Specifying `auto*` (or slightly weird here: `auto&`) instead of `auto` would have made that clearer for me 🤷

src/spicy/spicyz/glue-compiler.h
```diff
@@ -269,8 +282,7 @@ class GlueCompiler {
 
     std::map<hilti::ID, std::shared_ptr<glue::SpicyModule>> _spicy_modules;
 
-    std::vector<std::pair<hilti::ID, std::optional<hilti::ID>>>
-        _imports;                                            /**< imports from EVT files, with ID and optional scope */
+    std::vector<std::pair<hilti::ID, hilti::ID>> _imports;   /**< imports from EVT files, with ID and optional scope */
```
bbannier, line 285: The scope is not optional anymore so let's update the docstring, e.g.,

```suggestion
    std::vector<std::pair<hilti::ID, hilti::ID>> _imports;   /**< imports from EVT files, with ID and scope */
```

testing/btest/spicy/tuple-arg.zeek
```diff
@@ -27,7 +27,7 @@ public type Banner = unit {
     dash    : /-/;
     software: /[^\r\n]*/;
 
-    unset_field: uint64 if ( False );
+    unset_field: bytes &eod if ( False );
```
bbannier, line 30: Not completely clear to me what this does and whether it is related to the baseline change. The test seems to still fail for me.

bbannier (approved):

src/analyzer/protocol/quic/QUIC.spicy
rsmmr: > It seems that before passing an immutable value was fine since our refs implemented interior mutability (i.e., a const ref did seem to allow mutable access to the wrapped data) ... is that correct?

That's correct I think (though not sure if it's the only case where currently we "miss" the constness). I'll see that I add a test on the Spicy side.

> Separately I wonder whether it would make sense to more generally allow copy parameters 

Yeah, I can see providing that.

src/spicy/spicyz/driver.cc
rsmmr: It's not as bad as it looks. :-)

src/spicy/spicyz/glue-compiler.h
rsmmr: Well, it is still optional but that would now be an empty ID. I'll clarify the comment.

testing/btest/spicy/tuple-arg.zeek
rsmmr: The issue is in the EVT file, which passes `unset_field` to the `ssh::banner` event, which expected as `bytes`.

PR #3632, opened 2024-03-02
Monthly script optimization maintenance: March 2024

vpax:
This month's maintenance is pretty minor:

1. In some testing I found a bug where if `x` is a `vector of vector of T`, then `x += vector()` did _not_ append an empty `vector of T` to `x`, but instead did nothing (because it added all of the empty vector's _elements_ to `x`). The fix for this is leveraging Zeek's internal tracking of "unspecified" vectors (different from vectors of a known type that happen to be empty), and treat `+=` of unspecified vectors as being append-an-element in such situations.
2. Both ZAM and compile-to-C++ needed updates to support the use of `delete` to clear entire tables/vectors.
3. When working on ZAM profiling, I found a bug in printing ZAM instructions of a certain rare flavor. These now print correctly.
4. Quite minor tweaks for compile-to-C++ due to earlier changes for comprehensive AST location information.


testing/btest/language/vector.zeek
```diff
@@ -229,4 +229,16 @@ event zeek_init()
 	test_case( "negative index", v24[-1] == 5 );
 	test_case( "negative index", v24[-3] == 3 );
 
+	# For a vector-of-vectors, += of an empty vector should append it as
+	# a single element, not all of its elements (= nothing gets appended).
+	local v25: vector of vector of count;
+	v25 += vector();
```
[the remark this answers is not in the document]
vpax: A `vector()` constructor has an "unspecified" type (which is what the change to `Expr.cc` leverages), which means it's a vector-of-whatever's-needed. This works because it's empty, so there are no yield values to worry about not having the right type. Let me know if that answers your question ...

src/Expr.cc
```diff
@@ -1429,8 +1429,13 @@ AddToExpr::AddToExpr(ExprPtr arg_op1, ExprPtr arg_op2)
         // We need the IsVector(bt2) check in the following because
         // same_type() always treats "any" types as "same".
         if ( IsVector(bt2) && same_type(t1, t2) ) {
-            SetType(t1);
-            return;
+            // If the RHS is an unspecified vector and the LHS is a
+            // vector-of-vector-of... then treat this as adding an
+            // element, not as an append.
+            if ( t1->Yield()->Tag() != TYPE_VECTOR || ! t2->AsVectorType()->IsUnspecifiedVector() ) {
```
awelzel, line 1435:
```
global vv: vector of vector of count;

global empty: vector of count = vector();

vv += vector(vector(1), vector(2));
vv += vector(vector());
vv += vector(3);
vv += vector();
vv += vector(4);
vv += empty;
vv += vector(5);
print "vv", vv;
```
With 6.1:
```
$ zeek test.zeek 
vv, [[1], [2], [], [3], [4], [], [5]]
```

With this change:
```
$ zeek test.zeek 
vv, [[1], [2], [], [3], [], [4], [], [5]]
```

That makes sense.

I find the conditions, nesting and comments pretty hard to grasp, `! is unspecified` also triggers double negation for me :sweat_smile:  Maybe a helper to make this higher-level?
```
 
+static bool want_element_wise_vector_append(const TypePtr& lhs, const TypePtr& rhs) {
+    if ( ! (IsVector(lhs->Tag()) && IsVector(rhs->Tag())) )
+        return false;
+
+    if ( ! same_type(lhs, rhs) )
+        return false;
+
+    // LHS is not a vector-of-vector, do element-wise-append.
+    if ( lhs->Yield()->Tag() != TYPE_VECTOR )
+        return true;
+
+    if ( rhs->AsVectorType()->IsUnspecifiedVector() )
+        // This is a vector-of-vector-of-X += vector() construct.
+        // It is *not* treated as element-wise-append of an
+        // empty RHS, instead append an empty vector to the LHS.
+        return false;
+
+    return true;
+}
+
 AddToExpr::AddToExpr(ExprPtr arg_op1, ExprPtr arg_op2)
     : BinaryExpr(EXPR_ADD_TO, std::move(arg_op1), std::move(arg_op2)) {
     if ( IsError() )
@@ -1426,18 +1446,14 @@ AddToExpr::AddToExpr(ExprPtr arg_op1, ExprPtr arg_op2)
     }
 
     else if ( IsVector(bt1) ) {
-        // We need the IsVector(bt2) check in the following because
-        // same_type() always treats "any" types as "same".
-        if ( IsVector(bt2) && same_type(t1, t2) ) {
-            // If the RHS is an unspecified vector and the LHS is a
-            // vector-of-vector-of... then treat this as adding an
-            // element, not as an append.
-            if ( t1->Yield()->Tag() != TYPE_VECTOR || ! t2->AsVectorType()->IsUnspecifiedVector() ) {
-                SetType(t1);
-                return;
-            }
+        // Treat += of two vectors as appending each element
+        // of the RHS to the LHS if types agree.
+        if ( want_element_wise_vector_append(t1, t2) ) {
+            SetType(t1);
+            return;
         }
 
+        // Append RHS as element to LHS.
         is_vector_elem_append = true;
 
         bt1 = t1->AsVectorType()->Yield()->Tag();

```

awelzel (approved):

src/Expr.cc
vpax: Good thought! Now pushed.

PR #3641, opened 2024-03-07
Expr: Handle TYPE_VOID in SizeExpr and AssignExpr::Typecheck()

### LINKED ISSUE #3640: Indexing of sets allowed in some contexts
While `set`s and `table`s share a lot of common syntax, it doesn't make sense to index a set. For example, this:
```
global x = set(3.4.5.6, 9.8.7.6);
print x[3.4.5.6];
```
correctly produces:
```
error in <stdin>, line 2: value of type void illegal (print x[3.4.5.6])
```
However, the following works:
```
local x = set(5, 3);
local y: any = x[3];
print y;
```
printing `T` (!), though generating a run-time error if the index isn't present, rather than `F`. (It also doesn't work at all if `y` has type `bool` rather than `any`.)

That one is fairly contrived, but the following bit us today:
```
global x = set(3.4.5.6, 9.8.7.6);
print |x|;
print |x[3.4.5.6]|;
```
which prints:
```
2
1
```
rather than complaining in the second `print` that `x` cannot be indexed. This bit us because rather than using `print`, the `|x[3.4.5.6]|` equivalent appeared in an `if` statement, where what was _intended_ was `|x|`. (It wasn't a simple variable but rather an involved expression, which made the error harder to spot.)

You'd think this would be easy to fix, but the headache is that `add` and `delete` can take similar expressions, and for them the "indexing" is in fact okay since they're not accessing a value but rather controlling whether it's present. I  thought of an approach where the parser tracks whether it's inside an `add`/`delete`, and while it will work in practice, I can contrive situations for which it doesn't, alas.

awelzel:
@vpax reported surprising behavior when working with "void values". While these are not exposed to script land, plumb the places he pointed out are causing confusing behavior.

Closes #3640.


vpax:
For a set of pointwise fixes (which might be as best as we can readily do), this looks good - thanks!

timwoj (approved):

PR #3642, opened 2024-03-07
Introduce plugable cluster backends

awelzel:
This is draft/rfc material. I'm happy to hear feedback on terminology and if people see anything conceptually off. Initially this work was meant to result in some documentation/prototype, but it's been reasonably straight-forward, I could even see something like this going in, just so there's no large diff hanging around.

#### High-level

There's a new `Cluster::publish()` method that acts like `Broker::publish()`. `Cluster::publish()`, however, may use an alternative cluster backend if enabled. There's also new `Cluster::subscribe()` and `Cluster::unsubscribe()` bifs.

By default, these new cluster bifs use the existing broker implementation.

However, it's now possible to `redef` Cluster::backend` to another cluster backend. This PR provides one possible example that is  using [NATS](https://nats.io/). It's not decided that NATS will be Zeek's future cluster backend, it's just very easily to get setup with and their C library is available as package on Ubuntu :sweat_smile: 

Additionally, `redef Cluster::serializer` allows to change the encoding of Zeek events. The added implementation `BROKER_BIN_V1` directly leverages broker's functions for serialization/de-serialization.

The commit titles should give an impression about the steps involved. A summary of the new components:

* CLUSTER_SERIALIZER
  * Serializing a `cluster::detail::Event` instance (basically FuncValPtr/EventHandler, zeek::Args and (todo) metadata) into a byte buffer
  * De-serializing byte buffers back into a `cluster::detail::Event` instance that can then be queued
  * There's one implementation which is directly re-using the existing broker::data and broker binary v1 format. [Leveraging the JSON one should be straight-forward.](https://github.com/zeek/broker/issues/391).

* CLUSTER_BACKEND
  * Mainly defines an API for publishing and subscribe.
  * Receives a CLUSTER_SERIALIZER instance upon construction.
  * Stored on zeek::cluster::backend which is used by the Cluster bifs.
  * No stores, no logging, just Zeek events

#### Missing

* CI. This PR adds a NATS.io cluster backend and that upsets baselines. It would also introduce a new optional dependency. Either the code is moved to a plugin proper, or could be made opt-in and not built by default. Not sure. It's a bit sad to not build it by default because a number of baselines fall apart :cry: 
* Update existing `Broker::publish()` in the base scripts unless they are broker specific (think acld or zeekctl where broker Python bindings are involved).
* Logging. @rsmmr mentioned logging could possibly be just an event. With the batching of individual log writes that may be efficient enough (transporting the encoded batch as a zeek::StringVal that's deserialized at the receiver). One wouldn't handle that event in script land, but could setup a C++ based event handler directly. Could be an interesting debug channel though..
* Feedback


awelzel:
> Logging. @rsmmr mentioned logging could possibly be just an event. With the batching of individual log writes that may be efficient enough (transporting the encoded batch as a zeek::StringVal that's deserialized at the receiver). One wouldn't handle that event in script land, but could setup a C++ based event handler directly. Could be an interesting debug channel though..

Had chatted with Robin. In the end this appears more like a work-around and using a dedicated message type more sensible. Potentially as a separate interface.

The round-robin/batching logic of log writes currently lives in broker/Manager. Lifting this logic into logging/Manager allows other transports to benefit as well - maybe not the round-robin part, as NATS for example has a [queue](https://docs.nats.io/nats-concepts/core-nats/queue) feature that could be leveraged.

JustinAzoff:
Really exciting!

for this part:

```
CLUSTER_BACKEND

    Mainly defines an API for publishing and subscribe.
    Receives a CLUSTER_SERIALIZER instance upon construction.
```

so it sounds like we have a (in python)

```
class Backend:
    def __init__(self, serializer):
        self.serializer = serializer

    def publish(self, event, args):
        blob = self.serialize(event, args)
        self.do_whatever(blob)
```

Would it make more sense for it to be

```
class Backend:
    def publish(self, blob):
        self.do_whatever(blob)
```
and then somewhere else in zeek we have like...

```
class Communicator:
    def __init__(self, serializer, backend):
        self.serializer = serializer
        self.backend = backend

    def publish(self, event, args):
        blob = self.serializer.serialize(event, args)
        self.backend.publish(blob)
```

Basically, making a cluster backend know nothing about zeek events or serialization, it would only have to implement methods for moving blobs between processes, and not even have to be aware of serialization.  Passing in the serializer gets things mostly there, but then all the function signatures and api surface will still be "zeeky" vs only needing to deal with things like `std::vector<uint8_t>`.  The one area this doesn't work so well with is sending large values, it can be nice to have a way to do things like

```
serializeTo(fd, value);
deserializeFrom(fd, &value);
```

instead of needing the temporary buffer

```
serialize(value, &buf);
write(fd, buf);
read(fd, &buf);
deserialize(buf, &value);
```

Do you think the backend api would expose the concept of `peer`ing? I see the comment says

```
     * A cluster backend should initialize itself based on script variables,
     * register any IO sources and possibly start connections with a central
     * broker or peers.
```

Currently that's done in the script layer using `Broker::peer`, and one neat thing about that is you could build out new cluster topologies just by adding new scripts.

awelzel:
> Passing in the serializer gets things mostly there, but then all the function signatures and api surface will still be "zeeky" vs only needing to deal with things like std::vector<uint8_t>

Yeah, I could see the non-zeeky `DoXyz()` in an abstract helper class for backends, but think I'd keep the option open to still override `Xyz()` if you really wanted to. But good callout that's different of what's done in other frameworks, I was more focused on top-level interface that the BiFs will use.

> Do you think the backend api would expose the concept of peering? I see the comment says
>
>      * A cluster backend should initialize itself based on script variables,
>      * register any IO sources and possibly start connections with a central
>      * broker or peers.

Ah, that's a bit stale. The two sketches for NATS and ZeroMQ actually start connecting through `zeek_init()`, otherwise just running `zeek -e print 1` connects to stuff :-)

> Do you think the backend api would expose the concept of peering? 

Generally, I'm tending to have it decoupled no assumption on peering and connectivity between nodes (I think the reliance on that concept with Broker makes it difficult to use). `Cluster::publish()` publishes to a topic and all nodes subscribing to that topic will get the message (if they're up and connected and subscribed). If that's given, `Cluster::node_up` can be implemented via broadcasting on a discovery topic.

For NATS, that's shown here:

https://github.com/zeek/zeek/blob/34ecee9ab401d7bb498e550a45b9757997cd6e3e/scripts/policy/frameworks/cluster/backend/nats/connect.zeek

https://github.com/zeek/zeek/blob/34ecee9ab401d7bb498e550a45b9757997cd6e3e/scripts/policy/frameworks/cluster/backend/nats/main.zeek#L101-L138

If you have more thoughts, happy to switch to email or Zoom!

awelzel:
I am closing this for the time being. Some of the pre-work is in #3937 and the rest is actively being worked on. CI isn't close to passing, so no need to keep this open.

PR #3643, opened 2024-03-08
Add a method to register an event handler to a std::function via C++

timwoj:
While working on the telemetry work, I realized that I was adding two events that simply called a BIF method so I could do all of the handling in C++. That seemed very off to me, in that I couldn't add an event handler from C++ that just called a C++ method directly. This PR adds a way to register such event handlers.

### DIFF
--- a/src/EventRegistry.h
+++ b/src/EventRegistry.h
@@ -12,6 +12,7 @@
 #include <vector>
 
 #include "zeek/IntrusivePtr.h"
+#include "zeek/ZeekArgs.h"
 
 namespace zeek {
 
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -53,6 +53,7 @@ const char* stmt_name(StmtTag t) {
         "null",
         "assert",
         "extern",
+        "std-function",
     };
 
 #pragma GCC diagnostic push
@@ -2081,6 +2082,19 @@ TraversalCode WhenStmt::Traverse(TraversalCallback* cb) const {
 
     tc = cb->PostStmt(this);
     HANDLE_TC_STMT_POST(tc);
+}
+
+ValPtr StdFunctionStmt::Exec(Frame* f, StmtFlowType& flow) {
+    zeek::Args args = *f->GetFuncArgs();
+
+    if ( func.index() == 0 ) {
+        flow = FLOW_NEXT;
+        std::get<0>(func)(args);
+    }
+    else
+        std::get<1>(func)(args, flow);
+
+    return nullptr;
 }
 
 } // namespace zeek::detail
--- a/src/StmtEnums.h
+++ b/src/StmtEnums.h
@@ -31,8 +31,9 @@ enum StmtTag {
     STMT_ZAM,           // a ZAM function body
     STMT_NULL,
     STMT_ASSERT,
-    STMT_EXTERN, // for custom Stmt subclasses provided by plugins
-#define NUM_STMTS (int(STMT_EXTERN) + 1)
+    STMT_EXTERN,       // for custom Stmt subclasses provided by plugins
+    STMT_STD_FUNCTION, // statements that call a std::function from c++
+#define NUM_STMTS (int(STMT_STD_FUNCTION) + 1)
 };
 
 enum StmtFlowType {
--- a/src/Func.cc
+++ b/src/Func.cc
@@ -143,6 +143,11 @@ void Func::AddBody(detail::StmtPtr /* new_body */, const std::vector<detail::IDP
     Internal("Func::AddBody called");
 }
 
+void Func::AddBody(detail::StdFunctionStmt::FunctionVariant body, int priority) {
+    auto stmt = zeek::make_intrusive<detail::StdFunctionStmt>(std::move(body));
+    AddBody(stmt, {}, priority);
+}
+
 void Func::SetScope(detail::ScopePtr newscope) { scope = std::move(newscope); }
 
 FuncPtr Func::DoClone() {
--- a/src/Func.h
+++ b/src/Func.h
@@ -114,6 +114,7 @@ public:
     void AddBody(detail::StmtPtr new_body, const std::vector<detail::IDPtr>& new_inits, size_t new_frame_size,
                  int priority = 0);
     void AddBody(detail::StmtPtr new_body, size_t new_frame_size);
+    void AddBody(detail::StdFunctionStmt::FunctionVariant body, int priority = 0);
 
     virtual void SetScope(detail::ScopePtr newscope);
     virtual detail::ScopePtr GetScope() const { return scope; }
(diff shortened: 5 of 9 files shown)


src/EventRegistry.h
```diff
@@ -99,6 +100,16 @@ class EventRegistry final {
...
 
+    /**
+     * Registers an event handler that calls a C++ std::function instead of a script-land event handler.
+     *
+     * @param name The name of the event handler to lookup/register.
+     * @param priority The priority of the function body in the handler.
+     * @param f The std::function or lambda to call when the event is handled.
+     * @return true if the registration was successful, false otherwise.
+     */
+    bool Register(std::string_view name, int priority, std::function<void(const zeek::Args&)> f);
```
awelzel, line 111: IMO, I don't think we should have that API on the registry. Registry is to get handlers, not necessarily containing logic to modify them. The `Register()` behavior here also disagrees with the existing Register() method which actually installs a handler if it doesn't exist.

https://github.com/zeek/zeek/pull/3643/files#diff-3b5342aaebade88c649afd23944bf7ea51f98eebd0c7f859188cb064ea6ba232R50

To me, if anything, this seems more like a helper/convenience function that could live in utils.h.

```
bool add_std_func_to_handler(std::string_view name, int priority, std::function<void(const zeek::Args&)> f) {
   if ( auto handler = Lookup(name) ) {
        handler->GetFunc()->AddBody(std::move(f), priority);
        return true;
   }
   return false;
}
```

And I could actually see not even offering that unless it ends up being a very often used pattern.

I know the first implementation you had started with this API and maybe you're hanging on to it? :-)

src/Stmt.cc
```diff
@@ -2071,4 +2071,13 @@ TraversalCode WhenStmt::Traverse(TraversalCallback* cb) const {
     HANDLE_TC_STMT_POST(tc);
 }
 
+ValPtr StdFunctionStmt::Exec(Frame* f, StmtFlowType& flow) {
+    flow = FLOW_NEXT;
+    zeek::Args args = *f->GetFuncArgs();
+
+    func(args);
```
awelzel, line 2078: I do wonder if we should support the `StmtFlowType& flow` in the std::function as well. Maybe with some sort of overload/variant so that you would not need to have it. Not sure if that's worth it.

Then you could actually use StdFunctionStmt with hook handlers and actually "break" from them.

awelzel (commented):
I'm a fan. Modulo the EventRegistry API.

If StmtFlowType support is added, I'm almost tempted to re-use this in ZeekJS :-)

I wonder if @rsmmr should double check on what you're doing here :-)

src/StmtEnums.h
```diff
@@ -32,6 +32,7 @@ enum StmtTag {
     STMT_NULL,
     STMT_ASSERT,
     STMT_EXTERN, // for custom Stmt subclasses provided by plugins
+    STMT_NATIVE, // statements that call a std::function from c++
```
awelzel, line 35: Seeing STMT_CPP above, maybe that could be re-used. Though probably better to have a new one. I start to like your earlier and more specific `STMT_STD_FUNCTION` more than the opaque `STMT_NATIVE`.. sorry ;-)

src/EventRegistry.h
timwoj: I was hanging onto it as a convenience method, but I agree that it's not necessary. I can drop it.

src/StmtEnums.h
timwoj: `STMT_CPP` is for script-opt, so it probably isn't appropriate here.

timwoj: I renamed it back to STMT_STD_FUNCTION and also fixed the array of strings that I missed before.

src/Stmt.cc
timwoj: I can't override it because the function gets stored, so StdFunctionStmt would have to store both versions. I could see the function definition just always taking the flow though. Implementers would need to know to set it to `FLOW_NEXT` if they don't want to handle it though.

awelzel: I was thinking something like the following, allowing users to provide either variant and then calling the right one at Exec() time. Untested and not so sure about the names, but think this should work.

```
diff --git a/src/Stmt.cc b/src/Stmt.cc
index a75d768a8..362805005 100644
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -2076,7 +2076,10 @@ ValPtr StdFunctionStmt::Exec(Frame* f, StmtFlowType& flow) {
     flow = FLOW_NEXT;
     zeek::Args args = *f->GetFuncArgs();
 
-    func(args);
+    if ( func.index() == 0 )
+        std::get<0>(func)(args);
+    else
+        std::get<1>(func)(args, flow);
 
     return nullptr;
 }
diff --git a/src/Stmt.h b/src/Stmt.h
index 8d9bd1ac7..bdcfb7b88 100644
--- a/src/Stmt.h
+++ b/src/Stmt.h
@@ -782,9 +782,12 @@ protected:
 
 // Statement that calls a std::function. These can be added to a Func body
 // to directly all a C++ method.
+using StdFunction =
+    std::variant<std::function<void(const zeek::Args&)>, std::function<void(const zeek::Args&, StmtFlowType& flow)>>;
+
 class StdFunctionStmt : public Stmt {
 public:
-    StdFunctionStmt(std::function<void(const zeek::Args&)> f) : Stmt(STMT_STD_FUNCTION), func(std::move(f)) {}
+    StdFunctionStmt(StdFunction f) : Stmt(STMT_STD_FUNCTION), func(std::move(f)) {}
 
     ValPtr Exec(Frame* f, StmtFlowType& flow) override;
 
@@ -796,7 +799,7 @@ public:
     TraversalCode Traverse(TraversalCallback* cb) const override { return TC_CONTINUE; }
 
 private:
-    std::function<void(const zeek::Args&)> func;
+    StdFunction func;
 };
```

awelzel: > I could see the function definition just always taking the flow though.

Or that, might be pragmatic enough. We could set FLOW_NEXT before calling func, too.

rsmmr (commented):
Interesting idea, I can see that being useful (even though it feels a bit like a layering violation :-)

One question: I was  a bit surprised to see the new statement type for this, I think I'd have expected a new function `Kind` instead. What's advantage of the statement approach?

timwoj:
> One question: I was a bit surprised to see the new statement type for this, I think I'd have expected a new function `Kind` instead. What's advantage of the statement approach?

Since it's hooking into an `EventHandler`, that's already going to have a ScriptFunc attached to it. This just adds another body to the the func object that already exists. If I was to add another func type, it'd have to override any existing function that exists for the event handler.

src/Stmt.cc
timwoj: Looks ok with the variant. I pushed a commit with that.

Issue #3647, opened 2024-03-14
Binary builds broken after recent Spicy work


0xxon:
All binary builds are currently broken.

Example error message:

```
FAILED: src/analyzer/protocol/ldap/ldap___linker__.cc src/analyzer/protocol/ldap/ldap_spicy_init.cc src/analyzer/protocol/ldap/ldap_spicy_hooks_LDAP.cc src/analyzer/protocol/ldap/ldap_LDAP.cc src/analyzer/protocol/ldap/ldap_ASN1.cc /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap___linker__.cc /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap_spicy_init.cc /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap_spicy_hooks_LDAP.cc /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap_LDAP.cc /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap_ASN1.cc
cd /home/johanna/zeek/src/analyzer/protocol/ldap && /usr/bin/cmake -E env ZEEK_SPICY_LIBRARY_PATH=/home/johanna/zeek/scripts/spicy ASAN_OPTIONS=:detect_leaks=0 /home/johanna/zeek/build/src/spicy/spicyz/spicyz -L /home/johanna/zeek/auxil/spicy/hilti/lib -L /home/johanna/zeek/auxil/spicy/spicy/lib -x /home/johanna/zeek/build/src/analyzer/protocol/ldap/ldap ldap.spicy ldap.evt asn1.spicy
[internal-error] unknown operator 'generic::CastedCoercion'
```

To regenerate, use any linux version with the following configure flags:

```
./configure --binary-package --enable-static-broker --enable-static-binpac --disable-broker-tests --build-type=Release --generator=Ninja
```

I (apparently mistakenly) remembered that we added this to CI in the past. As this does not seem to be the case, I will just add something to CI that uses these build flags, and alerts on this kind of problem early.

0xxon:
Ok, this does also fail in CI if one uses the exact same parameters, as GH-3648 adds :)

The run at https://cirrus-ci.com/task/6155383379591168?logs=build#L3067 has a more verbose error message:

```
[error] /zeek/src/analyzer/protocol/finger/finger.spicy:17:22-17:31: unsupported operator: <Finger::Request>.whois
[...]

PR #3648, opened 2024-03-14
Add a CI test with the same options as we used for binary builds.

0xxon:
We already do have a test that check static builds, but none that tests the exact same build flags we use for binary builds on OBS. This commit fixes this oversight.

Related to GH-3647


awelzel:
Merged with https://github.com/zeek/zeek/pull/3650

PR #3650, opened 2024-03-14
Fix binary build of `spicyz`.

### LINKED ISSUE #3647: Binary builds broken after recent Spicy work

rsmmr:
We need to link against the Spicy toolchain's object libraries but
were linking against static libraries instead.

Closes #3647.

### DIFF
--- a/src/spicy/spicyz/CMakeLists.txt
+++ b/src/spicy/spicyz/CMakeLists.txt
@@ -10,7 +10,18 @@ set_target_properties(spicyz PROPERTIES CXX_EXTENSIONS OFF)
 
 target_include_directories(spicyz PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
-target_link_libraries(spicyz PRIVATE hilti spicy)
+if (BINARY_PACKAGING_MODE)
+    # Ensure we link against the HILTI/Spicy toolchain libraries (should be
+    # default, but isn't set here for some reason).
+    set(HAVE_TOOLCHAIN ON)
+
+    hilti_link_object_libraries_in_tree(spicyz PRIVATE)
+    spicy_link_object_libraries_in_tree(spicyz PRIVATE)
+else ()
+    hilti_link_libraries_in_tree(spicyz PRIVATE)
+    spicy_link_libraries_in_tree(spicyz PRIVATE)
+endif ()
+
 prefer_configured_spicy_include_dirs(spicyz)
 
 install(TARGETS spicyz DESTINATION ${CMAKE_INSTALL_BINDIR})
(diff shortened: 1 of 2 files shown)


rsmmr:
I could reproduce it on my Mac, and this solves it for me. @0xxon can you check if that works for the binary builds as well?

bbannier:
> I could reproduce it on my Mac, and this solves it for me. @0xxon can you check if that works for the binary builds as well?

Could you cherry-pick into this PR and rerun with the `fullci` tag?[^1]

[^1]: Set the tag before pushing so it is picked up by the triggered CI run.

rsmmr:
> Could you cherry-pick #3648 into this PR and rerun with the `fullci` tag?[1](#user-content-fn-1-f205bf4de25ad72b8fcf286b452de9c9)

Good idea, done.

bbannier (approved):

src/spicy/spicyz/CMakeLists.txt
```diff
@@ -10,7 +10,18 @@ set_target_properties(spicyz PROPERTIES CXX_EXTENSIONS OFF)
 
 target_include_directories(spicyz PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
-target_link_libraries(spicyz PRIVATE hilti spicy)
+if (BINARY_PACKAGING_MODE)
```
awelzel, line 13: Just commenting for the record, nothing to change.

I'm somewhat confused by this (and have been a bit before with https://github.com/zeek/zeek/pull/3565)

This seems to assume BINARY_PACKAGING_MODE never uses an externally provided Spicy and it's possible to link against the in-tree objects. The other thought, are we just linking object files into `spicyz` to full-fill some "rpaths-not-allowed" rule? Seems like a dishonest solution to that :-)

Fedora documents it's okay to use RPATH for internal libraries that are used by executables to enable code sharing. Could term hilti/spicy as such IMO.
https://fedoraproject.org/wiki/RPath_Packaging_Draft#Rpath_for_Internal_Libraries

PR #3651, opened 2024-03-14
CI: Remove CentOS 7 and Debian 10 builds

timwoj:
According to our [Platform Support Policy](https://github.com/zeek/zeek/wiki/Platform-Support-Policy):

Debian: Current release and prior LTS one
Debian 12 came out last year and Debian 11 is the prior LTS. Debian 10 is going EOL before we will release 7.0 so we can remove the builds now.

CentOS: Current release and prior one still receiving maintenance updates
CentOS Stream 9 is out and CentOS Stream 8 is still receiving maintenance updates. CentOS 7 is going EOL before we will release 7.0 so we can remove the builds now.

Updates should probably be made to the [OS Support Matrix](https://github.com/zeek/zeek/wiki/Zeek-Operating-System-Support-Matrix), but could be done after 7.0 is released.


timwoj:
We should also make a similar PR in the broker repo.

awelzel (approved):

ckreibich:
> We should also make a similar PR in the broker repo.

Yeah, thanks Tim. And zeek-archiver and zeek-aux and binpac, technically. I'm looking into migrating zeek-archiver into zeek-aux right now, though, so hopefully at least that one won't be necessary any longer. More on that as soon as I have a branch to review.

And it's interesting: normally we should also backport the CI change to the latest LTS branch, though since this one is forward-looking to the Zeek 7 release, that's not the case right now.

Issue #3573, opened 2024-01-22
Spicy: Make `replaces` cover calls to `InstantiateAnalyzer`


J-Gras:
While this works in script-land, calls to `InstantiateAnalyzer` in C++ will not pick up the new [replaced] analyzer, for example:
https://github.com/zeek/zeek/blob/13fde341d25302f46901693a05d191591781c6ea/src/analyzer/protocol/smtp/SMTP.cc#L388
In my case, I had to use the new analyzer name in the instantiate call.

_Originally posted by @J-Gras in https://github.com/zeek/zeek/issues/3569#issuecomment-1903578604_

awelzel:
As mentioned in #3654, deprecating "port" / "ports" points out that ports registered though `Analyzer::register_for_ports()` are not respected when used together with `replaces`.

That's boiling down to InstantiateAnalyzer(), but tag based which may add a bit of a twist to this:
https://github.com/zeek/zeek/blob/a45489de90a78c2815d6589b03ddd398251f0001/src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc#L177-L181

(and I'm tempted to open a PR to rename `port` and `ports` to *analyzer_tag* :-)

PR #3654, opened 2024-03-15
Deprecate port usage in .evt files

### LINKED ISSUE #3442: Deprecate specifying ports in Spicy EVT files
Once can currently specify ports a Spicy analyzer should trigger on in EVT files; at the same time Spicy analyzers can also registered in Zeek script land like any other analyzer.

Registering from EVT files requires custom C++ code and careful consideration so it works with all expectations from script land; the script land registration already works correctly automatically.

We should deprecate and eventually drop support for specifying ports via EVT files. The package template already stopped suggesting specifying ports in the EVT file with zeek/package-template#27.

Related: removing support for `%port` in Spicy, zeek/spicy#1587.

awelzel:
As in the commit message, this points out that "replaces" isn't functioning for ports of the replaced analyzer registered via `Analyzer::register_for_port()`. This could probably be covered via #3573.

Closes #3442.


bbannier:
Closes #3442.

PR #3658, opened 2024-03-19
Remove variant from StdFunctionStmt

timwoj:
The variant ended up conflicting with std::bind, which resulted in failures on the btest invoking it. Change back to a single function that takes a flow, and default it to a value in Exec.


ckreibich:
For reference: https://github.com/zeek/zeek/pull/3643/files

PR #3660, opened 2024-03-20
fix ZAM "cat" of double/time values

vpax:
ZAM has its own set of `cat()` instructions that replace calls to the BiF. These were missing some special-casing that the BiF (indirectly) does to render `double`/`time` values with a trailing `.0` if they're exactly equivalent to an integer. This PR fixes that.

### DIFF
--- a/src/script_opt/ZAM/BuiltInSupport.cc
+++ b/src/script_opt/ZAM/BuiltInSupport.cc
@@ -68,10 +68,18 @@ void FixedCatArg::RenderInto(ZVal* zframe, int slot, char*& res) {
             break;
 
         case TYPE_DOUBLE:
-        case TYPE_TIME:
-            n = modp_dtoa2(z.AsDouble(), res, 6);
+        case TYPE_TIME: {
+            auto d = z.AsDouble();
+            n = modp_dtoa2(d, res, 6);
             res += n;
+
+            if ( util::approx_equal(d, nearbyint(d), 1e-9) && std::isfinite(d) && ! strchr(tmp, 'e') ) {
+                // disambiguate from integer
+                *(res++) = '.';
+                *(res++) = '0';
+            }
             break;
+        }
 
         case TYPE_PATTERN:
             text = z.AsPattern()->AsPattern()->PatternText();
--- a/src/util.cc
+++ b/src/util.cc
@@ -2550,6 +2550,32 @@ TEST_CASE("util split") {
     }
 }
 
+TEST_CASE("util approx_equal") {
+    CHECK(approx_equal(47.0, 47.0) == true);
+    CHECK(approx_equal(47.0, -47.0) == false);
+    CHECK(approx_equal(47.00001, 47.00002) == false);
+    CHECK(approx_equal(47.00001, 47.00002, 1e-5) == true);
+    CHECK(approx_equal(47.0, -47.0, 1e2) == true);
+    CHECK(approx_equal(47.0, -47.0, 94 + 1e-10) == true);
+    CHECK(approx_equal(47.0, -47.0, 94) == false);
+
+    constexpr auto inf = std::numeric_limits<double>::infinity();
+    CHECK_FALSE(approx_equal(inf, inf));
+    CHECK_FALSE(approx_equal(-inf, inf));
+    CHECK_FALSE(approx_equal(inf, -inf));
+    CHECK_FALSE(approx_equal(inf, inf, inf));
+
+    constexpr auto qnan = std::numeric_limits<double>::quiet_NaN(); // There's also `signaling_NaN`.
+    CHECK_FALSE(approx_equal(qnan, qnan));
+    CHECK_FALSE(approx_equal(-qnan, qnan));
+    CHECK_FALSE(approx_equal(qnan, -qnan));
+}
+
+/**
+ * Returns whether two double values are approximately equal within some tolerance value.
+ */
+bool approx_equal(double a, double b, double tolerance) { return std::abs(a - b) < std::abs(tolerance); }
+
 } // namespace zeek::util
 
 extern "C" void out_of_memory(const char* where) {
--- a/src/util.h
+++ b/src/util.h
@@ -572,6 +572,15 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
 std::string json_escape_utf8(const char* val, size_t val_size, bool escape_printable_controls = true);
 
 /**
+ * Checks for values that are approximately equal.
+ * @param a first value to compare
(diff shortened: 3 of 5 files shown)


src/script_opt/ZAM/BuiltInSupport.cc
```diff
@@ -68,10 +68,24 @@ void FixedCatArg::RenderInto(ZVal* zframe, int slot, char*& res) {
...
         case TYPE_DOUBLE:
-        case TYPE_TIME:
-            n = modp_dtoa2(z.AsDouble(), res, 6);
+        case TYPE_TIME: {
+            auto d = z.AsDouble();
+            n = modp_dtoa2(d, res, 6);
             res += n;
+
+            // Taken from ODesc::Add(double d, bool no_exp)
+            auto approx_equal = [](double a, double b, double tolerance = 1e-6) -> bool {
```
timwoj, line 77: We should move this to util, since I can see it being useful in other places.

timwoj (approved):

src/script_opt/ZAM/BuiltInSupport.cc
vpax: Okay, I'll update with that change.

vpax: @timwoj fixup now pushed

src/util.cc
```diff
@@ -2550,6 +2550,24 @@ TEST_CASE("util split") {
...
+    CHECK(approx_equal(47.0, 47.0) == true);
+    CHECK(approx_equal(47.0, -47.0) == false);
+    CHECK(approx_equal(47.00001, 47.00002) == false);
+    CHECK(approx_equal(47.00001, 47.00002, 1e-5) == true);
+    CHECK(approx_equal(47.0, -47.0, 1e2) == true);
+    CHECK(approx_equal(47.0, -47.0, 94 + 1e-10) == true);
+    CHECK(approx_equal(47.0, -47.0, 94) == false);
+    CHECK(approx_equal(1 / 0.0, 2 / 0.0) == false);
+    CHECK(approx_equal(1 / 0.0, 2 / 0.0, 3 / 0.0) == false);
+    CHECK(approx_equal(0.0 / 0.0, 0.0 / 0.0) == false);
```
timwoj, line 2563: The MSVC compiler will actually complain about divide-by-zeroes by default:

```
FAILED: src/CMakeFiles/zeek_objs.dir/util.cc.obj 
"C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\bin\Hostx86\x64\cl.exe"  /nologo /TP -DCARES_STATICLIB -DDOCTEST_CONFIG_NO_MULTITHREADING -DDOCTEST_CONFIG_SUPER_FAST_ASSERTS -DZEEK_CONFIG_SKIP_VERSION_H -DZEEK_PLUGIN_SKIP_VERSION_CHECK -Isrc\include -Isrc -I..\src\include -I..\src -I. -I..\auxil\out_ptr\include -I..\auxil\rapidjson\include -I..\auxil\paraglob\include -I..\auxil\highwayhash -Iauxil\broker\libbroker -I..\auxil\broker\libbroker -I"C:\Program Files\OpenSSL-Win64\include" -Ivcpkg_installed\x64-windows-static\include -Iauxil\binpac\lib -I..\auxil\binpac\lib -I..\auxil\libunistd\regex -I..\auxil\libunistd\unistd /DWIN32 /D_WINDOWS /GR /EHsc /O2 /Ob2 /DNDEBUG /MT -MT /guard:cf /Z7 /J /wd4068 -std:c++17 /showIncludes /Fosrc\CMakeFiles\zeek_objs.dir\util.cc.obj /Fdsrc\CMakeFiles\zeek_objs.dir\ /FS -c ..\src\util.cc
..\src\util.cc(2561): error C2124: divide or mod by zero
```

I believe there's a checker for ubsan for it too, but we don't enable ubsan jobs for PR builds.

timwoj (changes requested):

src/util.cc
vpax: so should we just leave those (and the NaNs) out? Certainly seems simplest

bbannier: One can explicitly construct infinity and NaN values so we could (and should) leave in a test, e.g.,

```cpp
constexpr auto inf = std::numeric_limits<double>::infinity();
CHECK_FALSE(approx_equal(inf, inf));
CHECK_FALSE(approx_equal(-inf, inf));
CHECK_FALSE(approx_equal(inf, -inf));
CHECK_FALSE(approx_equal(inf, inf, inf));

constexpr auto qnan = std::numeric_limits<double>::quiet_NaN(); // There's also `signalining_NaN`.
CHECK_FALSE(approx_equal(qnan, qnan));
CHECK_FALSE(approx_equal(-qnan, qnan));
CHECK_FALSE(approx_equal(qnan, -qnan));
```

src/util.cc
```diff
@@ -2550,6 +2550,24 @@ TEST_CASE("util split") {
...
+}
+
+bool approx_equal(double a, double b, double tolerance) {
+    auto v = a - b;
+    return v < 0 ? -v < tolerance : v < tolerance;
+}
```
bbannier, line 2569: Not originally introduced here, but since this patch makes it part of the public API let's fix the behavior for negative epsilon, e.g.,

```suggestion
bool approx_equal(double a, double b, double tolerance) { return std::abs(a - b) < std::abs(tolerance); }
```

src/util.h
```diff
@@ -571,6 +571,15 @@ std::string json_escape_utf8(const std::string& val, bool escape_printable_contr
...
 std::string json_escape_utf8(const char* val, size_t val_size, bool escape_printable_controls = true);
 
+/**
+ * Checks for values that are approximately equal.
+ * @param a first value to compare
+ * @param b second value to compare
+ * @param tolerance how close they need to be to deem them "approximately equal"
+ * @return true if `a` is within the given tolerance of `b`, false otherwise
+ */
+bool approx_equal(double a, double b, double tolerance = 1e-6);
```
bbannier, line 581: Not sure about this one, but for general `double` values the default epsilon seems pretty big. What do you think about using e.g., [`std::numeric_limits<double>::epsilon()`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) as default value and update the callers to pass in context-specific bigger value where they did not specify an epsilon before?

timwoj:
@bbannier I added all of your suggestions to this, do you want to make a second pass at it?

timwoj (approved):

bbannier (approved):

PR #3664, opened 2024-03-27
Minor cmake tweaks to fix library build

ckreibich:
Building with `ZEEK_STANDALONE=OFF` (a library-only build of Zeek, which came in via Microsoft) hit two minor hiccups, one a typo in a target, and the other a violation of the assumption that `zeek_exe` always exists.

Also removes a vestigial bit of Conan logic.


0xxon (approved):
Looks good to me - assuming the mac os build failure has nothing to do with it :)

PR #3665, opened 2024-03-28
CI: Specify the xcode version of the macOS Sonoma instance

timwoj:
A recent update to the base image by Cirrus caused our build to break. https://github.com/cirruslabs/cirrus-ci-docs/issues/1272 says that it's a problem with the base OS itself, and they recommended we try to xcode image instead.


ckreibich (approved):
Thanks Tim!

PR #3666, opened 2024-03-30
fix: set SignatureFile.full_path if invalid

r-rojo:
This sets SignatureFile::full_path member variable to the full path if SignatureFile::file member is relative and full_path is invalid. This uses the SignatureFile::load_location to derive the full path from the file path relative to the load_location.

This is needed when the full_path is passed in to either HookLoadFile and/or HookLoadFileExtended.

### DIFF
--- a/src/RuleMatcher.cc
+++ b/src/RuleMatcher.cc
@@ -241,8 +241,8 @@ bool RuleMatcher::ReadFiles(const std::vector<SignatureFile>& files) {
         // if load is relative to the load location, and we don't have a valid full_path
         // let's fix it by setting the full_path to load location and relative path
         if ( f.file.find("./") == 0 && (! f.full_path.has_value() || f.full_path == "") ) {
-          auto full_path = std::string(f.load_location.filename) + "/" + f.file);
-          f.full_path = util::detail::normalize_path(full_path);
+            auto full_path = std::string(f.load_location.filename) + "/" + f.file;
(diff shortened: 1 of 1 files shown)


ckreibich:
Thanks Rodrigo — I don't fully understand what this achieves. In the current code (prior to your PR), the bit that establishes `full_path` ([here](https://github.com/zeek/zeek/blob/master/src/RuleMatcher.cc#L241-L242)) seems to be doing its job? I've installed `zeek-amadey-detector`, and see this when I print `f.file` and `f.full_path` before/after it:
```
XXX before './amadey.sig' '/home/christian/inst/opt/zeek/share/zeek/site/packages/zeek-amadey-detector/amadey.sig'
XXX after './amadey.sig' '/home/christian/inst/opt/zeek/share/zeek/site/packages/zeek-amadey-detector/amadey.sig'
```
For signature files given on the command line (`-s ./foo.sig`), I see:
```
XXX before './foo.sig' '<none>'
XXX after  './foo.sig' './foo.sig'
```
So this looks fine to me. Can you say more about where you see unexpected input in your plugin? How is `full_path` invalid?

Ideally, add a test that shows what's not working, since we'll need that anyway.

ckreibich:
@r-rojo do you have any updates on this? I'm inclined to close the ticket at this point.

ckreibich:
I'm closing this for now — let us know if you have updates, Rodrigo.

PR #3669, opened 2024-04-02
Raise configure error message for unsupported archives

0xxon:
Github lets users download archives of our repos. These do not contain the necessary submodules. We regularly encounter users who stumble across this.

We already do have an error message that is raised when a non-recursive git checkout was done. This commit adds an error message for a non-git download that does not contain the necessary files.

### DIFF
--- a/configure
+++ b/configure
@@ -145,6 +145,25 @@ Zeek. To check out the required subdirectories, please execute:
     exit 1
 fi
 
+if [ ! -e "$sourcedir/cmake/COPYING" ] && [ ! -d "$sourcedir/.git" ]; then
+    echo "\
+You seem to be missing the content of the cmake directory.
+
+This typically means that you downloaded a non-release archive from github.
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -145,6 +145,25 @@ Zeek. To check out the required subdirectories, please execute:
...
+archive using one of the links at https://zeek.org/get-zeek/.
+
+If you want to get the current development version of Zeek, please use git to
+clone our repository.
+
+See https://docs.zeek.org/en/master/install.html#retrieving-the-sources for
+instructions.
+" >&2
+    exit 1
+fi
```
bbannier, line 165: This repeats the same check for the "magic file" `cmake/COPYING`. Let's use a single condition checking for absence of that file, and then have two cases under that, one for when the `.git` directory is present and one where it is missing.

bbannier (approved):

Issue #3224, opened 2023-08-08
Support for configurable localversion part in Zeek's version


awelzel:
For distributors, appending an additional version string to Zeek's version can be useful to track any custom patches applied or other build/tree specific information.

Currently, to include additional information into the output of `zeek --version` a custom patch or `sed` in the build process is required.

Idea would be:
```
$./configure --localversion="myversion+4"
...
$ zeek --version
zeek version 6.0.0-myversion+4
```

ckreibich:
Originalism vs pragmatism? :-)

> you cannot modify Zeek in any way while still calling it Zeek

There's also the flip-side, which is that you have to call it something else if you do modify it. Seems to me that the proposal provides a good compromise: it makes it clear that a modification is present, but implies that it's one that remains essentially Zeek. I also see the practical utility — I'd likely use this at times during development.

A potential alternative might be to add content in the `zeek -V` output. Not sure that's better because it's less visible, but the field naming could imply the use case (perhaps via a section like `distributor-metadata`). 

How would this get along with existing modifications, like `zeek version 6.1.0-dev.8-debug`? Just append?

ckreibich:
I'm putting this on the agenda for the LT call this coming Thursday, just to make sure. My take is to put it in the version string and simply dash-append to whatever else is already there.

ckreibich:
> My take is to put it in the version string and simply dash-append to whatever else is already there.

We're good to go with this approach.

ckreibich:
This originally came in via @pbcullen — Peter, if this is still pressing for you, is this something you could see contributing?

pbcullen:
I'm finally starting to look at this. If a localversion is supplied, I'm thinking ZEEK_VERSION_FULL in CMakeLists.txt should reflect that, but wondering if that might break something? It would affect the VERSION_C_IDENT value.  Also, should ZEEK_VERSION_NUMBER in CMakeLists.txt differ from that calculated using the major, minor, patch values? Maybe something like add 1000000 to it?

ckreibich:
This is pretty subtle ... thanks for checking first. I don't think we should change `ZEEK_VERSION_FULL`, because it only captures version information right now (directly based on the `VERSION` file in the source tree). Our handling of debug builds looks close to what we'd like: the `zeek_version()` function manually adds `-debug` to the returned string that gets shown to the user, and cmake adds it to `VERSION_C_IDENT`, which means it ends up in the internal string identifying the Zeek build a plugin gets compiled for.

[...]

PR #3673, opened 2024-04-03
Support for configurable localversion

pbcullen:
Adds a new --localversion option to configure.
The value of localversion becomes part of the Zeek version string. Docker builds allow an environment variable (LOCALVERSION) to set the localversion string.

https://github.com/zeek/zeek/issues/3224


scripts/base/misc/version.zeek
```diff
@@ -92,6 +92,8 @@ function parse(version_string: string): VersionDescription
 
 		vs = gsub(vs, /-debug$/, "");
 		vs = gsub(vs, /-(beta|rc|dev)[0-9]*/, "");
+		# Drop the optional local version piece
+		vs = gsub(vs, /-[a-zA-Z0-9]+$/, "");
```
ckreibich, line 96: How about we add a field `localversion` to the record and fill that in too?

ckreibich (commented):
Very nice Peter — I would not have thought of the Docker Makefile. I only have two suggestions, let me know what you think.

scripts/base/misc/version.zeek
pbcullen: While working on this, I think I've found some abiguity. In the version string "2.5.2-beta5-12-debug", what does the "12" signify? And, how could that be distinguished from a localversion?

pbcullen: It seems that "12" is the number of commits relative to the beta|rc|dev. And, it seems that a "." is currently used between the beta|rc|dev and the commit count. If that's all true, I'd propose disallowing a "-" as a separator for that grouping. Perhaps there should be a "rule" that a "-" is only used to separate the extra portions of the version string? (beta|dev|rc, localversion, debug). @ckreibich thoughts?

ckreibich: Nice catch Peter. Yes, that rule is the way to go. We changed the dev versioning scheme in 2019, [here](https://github.com/zeek/zeek-aux/commit/6b9ab346f9243270e25d41bdd67c62b9cd5fc2d7), so "2.5.2-beta5-12-debug" would not occur today — it would be "2.5.2-beta5.12-debug". So I think proceeding as planned and enforcing that localversion doesn't contain dashes is all we need.

CMakeLists.txt
```diff
@@ -573,6 +575,12 @@ set(VERSION_C_IDENT "${ZEEK_VERSION_FULL}_plugin_${API_VERSION}")
 string(REGEX REPLACE "-[0-9]*$" "_git" VERSION_C_IDENT "${VERSION_C_IDENT}")
 string(REGEX REPLACE "[^a-zA-Z0-9_\$]" "_" VERSION_C_IDENT "${VERSION_C_IDENT}")
 
+set(ZEEK_VERSION_FULL_LOCAL "${ZEEK_VERSION_FULL}")
+if (NOT ZEEK_VERSION_LOCAL STREQUAL "")
+    set(ZEEK_VERSION_FULL_LOCAL "${ZEEK_VERSION_FULL_LOCAL}-${ZEEK_VERSION_LOCAL}")
```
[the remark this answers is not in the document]
pbcullen: Sounds good. I've made the change in CMakeLists.txt to check for dashes in ZEEK_VERSION_LOCAL.

scripts/base/misc/version.zeek
ckreibich: Thanks Peter! That macOS failure is unrelated to your PR and we'll sort it out.

timwoj: That was fixed by https://github.com/zeek/zeek/pull/3676. This just needs to be rebased to pick it up.

ckreibich: Yeah I just saw. Thank you both for being so quick. I'm approving.

ckreibich (approved):

ckreibich:
Peter I've merged this via #3690 because I'm seeing each of the commits twice in the history if I go with yours — I think you may have done a merge instead of a rebase? It's still your commits, just cherry-picked. I verified that the trees are the same and configuring a localversion does work.

PR #3678, opened 2024-04-09
Make sure that vcpkg isn't preferred if pcap_root_dir is passed in

timwoj:
A user discovered that if you pass in `PCAP_ROOT_DIR` to cmake on Windows, cmake ignores it and still prefers the vcpkg installation of libpcap. This PR fixes this by first flushing the cached variables set by vcpkg, and then adding the vcpkg paths to the ignore list when searching for the libpcap paths.

A branch with the same name needs to be merged in the cmake repo.


ckreibich (approved):
Looks good to me Tim. Subtle stuff.

PR #3680, opened 2024-04-10
Don't attempt to stop or flush disabled writer frontends

### LINKED ISSUE #3662: zeek dumps core when attempting log creation in non-writeable directory.
zeek will dump core when attempting to write logs in a directory that does not have write permissions.
This problem was found in error.

**Steps to reproduce**:
```
1 - cd to a non writeable directory. /opt
2 - zeek -r <pcap file> ('reproducer.tgz' attached)
```

**Expected results:**
Process exits cleanly

**Actual results:**
Process dumps core.

**Summary:**
The backtraces produced by this problem are by and large consistent. 
However, there were a few cases that were different. Associated backtraces are attached 'zeek-bt.txt'. 

I compiled zeek with AddressSanitizer and when executing with the attached pcap 'reproducer.tgz' all 
backtraces are consistent and produce the following:

```
==90178==ERROR: AddressSanitizer: heap-use-after-free on address 0x61f000023099 at pc 0x000002cee56f bp 0x7ffd3059bcf0 sp 0x7ffd3059bce8
READ of size 1 at 0x61f000023099 thread T0
    #0 0x2cee56e in std::__atomic_base<bool>::load(std::memory_order) const /usr/include/c++/12/bits/atomic_base.h:488
    #1 0x2cee56e in std::atomic<bool>::operator bool() const /usr/include/c++/12/atomic:87
    #2 0x32b0133 in zeek::threading::BasicThread::Terminating() const /home/zeek-user/src/zeek/src/include/zeek/threading/BasicThread.h:106
    #3 0x32af335 in zeek::threading::MsgThread::SendIn(zeek::threading::BasicInputMessage*, bool) /home/zeek-user/src/zeek/src/threading/MsgThread.cc:327
    #4 0x278e089 in zeek::threading::MsgThread::SendIn(zeek::threading::BasicInputMessage*) /home/zeek-user/src/zeek/src/include/zeek/threading/MsgThread.h:66
    #5 0x2831093 in zeek::logging::WriterFrontend::FlushWriteBuffer() /home/zeek-user/src/zeek/src/logging/WriterFrontend.cc:212
    #6 0x2830295 in zeek::logging::WriterFrontend::Stop() /home/zeek-user/src/zeek/src/logging/WriterFrontend.cc:136
    #7 0x280756c in zeek::logging::Manager::Terminate() /home/zeek-user/src/zeek/src/logging/Manager.cc:1794
    #8 0x2cdeff4 in terminate_zeek /home/zeek-user/src/zeek/src/zeek-setup.cc:381
    #9 0x2ce7c12 in zeek::detail::cleanup(bool) /home/zeek-user/src/zeek/src/zeek-setup.cc:1075
    #10 0x187c218 in main /home/zeek-user/src/zeek/src/main.cc:113
    #11 0x7fb11a82950f in __libc_start_call_main (/lib64/libc.so.6+0x2950f)
    #12 0x7fb11a8295c8 in __libc_start_main_impl (/lib64/libc.so.6+0x295c8)
    #13 0x187ba14 in _start (/home/zeek-user/src/zeek/build/src/zeek+0x187ba14)
```
Full AddressSanitizer logs are attached as zeek-AddressSanitizer-full.txt

The problem occurs in **zeek::threading::BasicThread::Terminating()**  executing code:

` bool Terminating() const { return terminating; } `

which is accessing memory that was freed by destructor  **zeek::logging::writer::detail::Ascii::~Ascii()** .

[reproducer.tgz](https://github.com/zeek/zeek/files/14776098/reproducer.tgz)
[zeek-bt.txt](https://github.com/zeek/zeek/files/14776099/zeek-bt.txt)
[zeek-AddressSanitizer-full.txt](https://github.com/zeek/zeek/files/14776156/zeek-AddressSanitizer-full.txt)

timwoj:
When a writer fails to start up, the backend tells the frontend to disable itself via a call to `WriterFrontend::SetDisable()`. The backend may be marked invalid and deleted by the threading manager at that point, but the pointer is still valid in the frontend. Most operations in the frontend check for this state before they try to do anything, but `Stop()` and `FlushWriteBuffer()` both don't. If they attempt to use the backend once it's been deleted, it runs into a use-after-free and crashes.

Fixes #3662

### DIFF
--- a/src/logging/WriterFrontend.cc
+++ b/src/logging/WriterFrontend.cc
@@ -133,6 +133,11 @@ WriterFrontend::~WriterFrontend() {
 }
 
 void WriterFrontend::Stop() {
+    if ( disabled ) {
+        CleanupWriteBuffer();
+        return;
+    }
+
     FlushWriteBuffer();
     SetDisable();
 
@@ -204,6 +209,11 @@ void WriterFrontend::Write(int arg_num_fields, Value** vals) {
 }
 
 void WriterFrontend::FlushWriteBuffer() {
+    if ( disabled ) {
+        CleanupWriteBuffer();
+        return;
+    }
+
     if ( ! write_buffer_pos )
         // Nothing to do.
         return;
@@ -259,6 +269,18 @@ void WriterFrontend::DeleteVals(int num_fields, Value** vals) {
         delete vals[i];
 
     delete[] vals;
+}
+
+void WriterFrontend::CleanupWriteBuffer() {
+    if ( ! write_buffer || write_buffer_pos == 0 )
+        return;
+
+    for ( int j = 0; j < write_buffer_pos; j++ )
+        DeleteVals(num_fields, write_buffer[j]);
+
+    delete[] write_buffer;
+    write_buffer = nullptr;
+    write_buffer_pos = 0;
 }
 
 } // namespace zeek::logging
--- a/src/logging/WriterFrontend.h
+++ b/src/logging/WriterFrontend.h
@@ -206,6 +206,9 @@ protected:
     static const int WRITER_BUFFER_SIZE = 1000;
     int write_buffer_pos;             // Position of next write in buffer.
     threading::Value*** write_buffer; // Buffer of size WRITER_BUFFER_SIZE.
+
+private:
+    void CleanupWriteBuffer();
 };
 
 } // namespace zeek::logging


src/logging/WriterFrontend.cc
```diff
@@ -133,6 +133,16 @@ WriterFrontend::~WriterFrontend() {
...
+        for ( int j = 0; j < write_buffer_pos; j++ )
+            DeleteVals(num_fields, write_buffer[j]);
+
+        delete[] write_buffer;
+        write_buffer = nullptr;
+        write_buffer_pos = 0;
+        return;
+    }
+
     FlushWriteBuffer();
```
ckreibich, line 146: Since this calls `FlushWriteBuffer()` anyway, you could just drop that change here — unless you wanted it in both methods for robustness (in which case you could make it a private method).

ckreibich (approved):
I can confirm that this fixes the problem over here, in an ASAN-enabled build.

src/logging/WriterFrontend.cc
timwoj: It's more about bailing out of `Stop` early if it's disabled. I hoisted it into a private method. I could call `FlushWriteBuffer` and then check for `disabled`, but it'd mean that `Stop` is relying on a side-effect of `FlushWriteBuffer` for the cleanup which isn't ideal.

PR #3685, opened 2024-04-15
Bump Spicy to latest `main`

bbannier:
The new reference parameter handling requires slight adjustments to the QUIC analyzer. These changes make it incompatible with earlier versions, but since this analyzer is in tree we do not need to worry about this (we leave cleanup of the remaining `SPICY_VERSION` checks for later).


bbannier:
Superseded by #3686.

PR #3686, opened 2024-04-15
Bump Spicy

rsmmr:


bbannier (approved):
I had previously opened by #3685 but didn't get to mark it ready-for-review. It looks like this makes a more surgical update, and also includes the removal of version-dependent code.

rsmmr:
> I had previously opened by #3685 but didn't get to mark it ready-for-review. It looks like this makes a more surgical update, and also includes the removal of version-dependent code.

Oh, sorry, I had missed that. But yeah, with the Spicy-side changes, this becomes more natural.

PR #3691, opened 2024-04-17
Add PacketFilter::remove_exclude function

pbcullen:
This adds a public function to the PacketFilter framework that allows the caller to remove an exclude filter by filter id.


timwoj (approved):

timwoj:
The change itself looks fine. The test failures are just because the line numbers changed on the error message output, so you just need to update those baselines.

pbcullen:
OK. I've got a local btest run going now. Will update soon.

pbcullen:
Baseline has been updated.

PR #3698, opened 2024-04-24
 Spicy: Map Spicy's `Protocol` enum to Zeek's `transport_proto`.

rsmmr:
- **Spicy: Map Spicy's `Protocol` enum to Zeek's `transport_proto`.**
- **Spicy: Cleanup some runtime code.**


src/spicy/runtime-support.h
```diff
@@ -929,36 +920,35 @@ inline ValPtr to_val(const T& t, TypePtr target) {
...
     // We'll usually be getting an int64_t for T, but allow other signed ints
     // as well.
-    static_assert(std::is_signed<std::underlying_type_t<T>>{});
-    auto it = static_cast<int64_t>(t);
+    static_assert(std::is_signed<std::underlying_type_t<typename T::Value>>{});
+    auto it = static_cast<int64_t>(t.value());
+
+    if ( target->GetName() == "transport_proto" )
+        // Special case: map enum values to Zeek's semantics.
+        return to_val_for_transport_proto(it, target);
```
bbannier, line 951: We can now convert any Spicy enum to a `transport_proto`. That seems at the very least fragile.

I'd suggest that we 1) check that we actually have a port value (either by adding an additional condition to this conditional, or by throwing an error if we see a mismatching type), and pass a proper `Protocol::Value` to `to_val_for_transport_proto` instead of an `int64_t`, e.g., here

```suggestion
    // Special case: map enum values to Zeek's semantics.
    if ( target->GetName() == "transport_proto" ) {
        if ( ! std::is_same_v<T, hilti::rt::Protocol::Value> )
            throw TypeMismatch(hilti::rt::demangle(typeid(t).name()), target);
        return to_val_for_transport_proto(static_cast<hilti::rt::Protocol::Value>(it), target);
    }
```

src/spicy/runtime-support.h
```diff
@@ -929,36 +920,35 @@ inline ValPtr to_val(const T& t, TypePtr target) {
...
-#endif
-
-    return to_val(proto, target);
+/** Maps HILTI's `Protocol` enum to Zeek's `transport_proto` enum. */
+inline ValPtr to_val_for_transport_proto(int64_t val, const TypePtr& target) {
+    switch ( val ) {
+        case hilti::rt::Protocol::TCP: return id::transport_proto->GetEnumVal(::TransportProto::TRANSPORT_TCP);
+        case hilti::rt::Protocol::UDP: return id::transport_proto->GetEnumVal(::TransportProto::TRANSPORT_UDP);
+        case hilti::rt::Protocol::ICMP: return id::transport_proto->GetEnumVal(::TransportProto::TRANSPORT_ICMP);
+        default: return id::transport_proto->GetEnumVal(::TransportProto::TRANSPORT_UNKNOWN);
```
bbannier, line 929: It's unfortunate that nothing makes explicit that we map `Undef` to `TRANSPORT_UNKNOWN`. If we keep accepting an `int64_t` here I'd at least add a fallthrough so we have it in the source (and can e.g., see it if we search for references),

```suggestion
        case hilti::rt::Protocol::Undef: [[fallthrough]];
        default: return id::transport_proto->GetEnumVal(::TransportProto::TRANSPORT_UNKNOWN);
```

If we'd instead accepted a `Protocol::Value` we could get rid of the `default` case altogether.

bbannier (approved):

src/spicy/runtime-support.h
bbannier: > If we'd instead accepted a Protocol::Value we could get rid of the default case altogether.

Actually our enum types are over int so we need to keep the `default`.

PR #3702, opened 2024-04-24
Avoid calling typecasts in Val when we have direct access to the underlying value object

timwoj:
There are a number of places in the Val.cc file where we call an `As...` method inside a `Val` implementation but the `Val` being implemented already has access to the raw object. This causes an unnecessary `static_cast` from one of the set of `UNDERLYING_ACCESSOR_DEF` macros defined in Val.h. This PR fixes these cases to use the raw object directly instead.

I'm opening this as a draft to get results from the benchmarker service before asking for reviews.

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -742,22 +742,18 @@ StringVal::~StringVal() { delete string_val; }
 
 ValPtr StringVal::SizeVal() const { return val_mgr->Count(string_val->Len()); }
 
-int StringVal::Len() const { return AsString()->Len(); }
+int StringVal::Len() const { return string_val->Len(); }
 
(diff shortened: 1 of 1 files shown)


timwoj:
Benchmark results:

Screenshot 2024-04-25 at 10 59 39 AM

Overall a wash. Some tests are higher and some are lower, both significantly and not.

timwoj:
I still think this is worth doing. It's cleaner to avoid the conversion in classes that have direct access to their data variables.

0xxon (approved):

0xxon:
Agreed, this looks reasonable to do to me.

PR #3703, opened 2024-04-25
Spicy: Query Zeek scriptland for file handles.

### LINKED ISSUE #3440: Spicy ignores Zeek-side FUID generation
Spicy currently does not call the Zeek script-side code that can be used for file uid generation; specifically it seems to ignore the functions that are set in `Files::register_protocol`. The Zeek C++-side functionality that should be called somewhere is `file_mgr->GetFileID`. This steps is currently completely skipped, instead Spicy generates its own IDs in `rt::cookie::FileStateStack::push()` - https://github.com/zeek/zeek/blob/9b1f3b58386c110302747d2f49d5a0bcfb334bd4/src/spicy/runtime-support.cc#L684-L689

In slightly different, but related things - it is somewhat common for binpac parsers to generate their own file-IDs and pass these to the files framework. This is currently also not possible in Spicy.

rsmmr:
Like traditional file analyzers, we now query Zeek's
`get_file_handle()` event for handles when a connection begins
analyzing an embedded file. That means that Spicy-side protocol
analyzers that are forwarding data into file analysis now need to call
Zeek's `Files::register_protocol()` and provide a callback for
computing file handles. If that's missing, Zeek will now issue a
warning. This aligns with the requirements Zeek's traditional protocol
analyzers. (If the EVT file defines a protocol analyzer to `replace`
an existing one, that one's `register_protocol()` will be consulted.)

Because Zeek's `get_file_handle()` event requires a current
connection, if a Spicy file analyzer isn't directly part of a
connection context (e.g., with nested files), we continue to use
hardcoded, built-in file handle. Scriptland won't be consulted in
that case, just like before.

Closes #3440.


testing/btest/spicy/file-analysis-data-in-concurrent.zeek
```diff
@@ -1,12 +1,23 @@
...
+
+function get_file_handle(c: connection, is_orig: bool): string
+	{
+	return cat(c$uid, ++i);
+	}
+
+event zeek_init()
+	{
+	Analyzer::register_for_port(Analyzer::ANALYZER_SPICY_SSH, 22/tcp);
+	Files::register_protocol(Analyzer::ANALYZER_SSH, [$get_file_handle=SSH::get_file_handle]); # use tag of replaced analyzer
```
bbannier, line 19: It isn't really clear to me why we cannot use the tag of this analyzer (from just the line above; trying that gives me _some result_ as well); could you add an expanded comment block explaining why we need to use the tag of the replaced analyzer (probably copy-and-pasted to everywhere we have the EOL comment now)?

NEWS
```diff
@@ -58,6 +58,13 @@ Changed Functionality
   ``frameworks/signatures/iso-9660`` which also increases the BOF buffer sufficiently.
   Note, doing so may increase memory and CPU usage significantly.
 
+- If a Spicy protocol analyzers feeds data into file analysis, it now
+  needs to call Zeek's `Files::register_protocol()` and provide a
+  callback for computing file handles. If that's missing, Zeek will
+  issue a warning. While this was not necessary in previous versions,
+  it aligns with the same requirement for traditional analyzers and
+  enables customizing file handles for protocol-specific semantics.
```
bbannier, line 66: Is there a way we could provide a sensible default in the template for protocol analyzers?

src/spicy/runtime-support.cc
```diff
@@ -689,8 +689,21 @@ rt::cookie::FileState* rt::cookie::FileStateStack::push(std::optional<std::strin
     std::string fid;
     if ( fid_provided )
         fid = *fid_provided;
-    else
-        fid = file_mgr->HashHandle(hilti::rt::fmt("%s.%d", _analyzer_id, ++_id_counter));
+    else {
+        auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
```
bbannier, line 693:
```suggestion
        auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
        assert(cookie);
```

bbannier (approved):

testing/btest/spicy/file-analysis-data-in-concurrent.zeek
rsmmr: What result do you get? When I tried it, the `get_file_handle` just wouldn't be called, and instead I'd get a warning that the `get_file_handle()` handler was missing.

The reason that we need the replaced analyzer's tag is the `spicy_mgr->tagForProtocolAnalyzer(..)` call in the code, the translation happens there. We could skip it but then existing scripts written for the original analyzer wouldn't work, which isn't desired. And we can't check for both tags because the 1st failure triggers that warning (and that's coming from scriptland).

NEWS
rsmmr: I was wondering about that too. The thing is that this doesn't apply to most protocol analyzers, only those that feed data into file analysis. I guess we could leave some commented out code in the template?

testing/btest/spicy/file-analysis-data-in-concurrent.zeek
bbannier: You are right, I see that warning too, but also seemingly valid file IDs, e.g., for `spicy/file-data-in-at-offset.zeek`

```patch
-FU6P513gQKLVP6iwmf
-FU6P513gQKLVP6iwmf
+XXXXXXXXXX.XXXXXX warning in /.../zeek/scripts/base/frameworks/files/main.zeek, line 523: get_file_handle() handler missing for Analyzer::ANALYZER_SSH
+FPFmBApS7zfa2mU78
+FPbnGx3L0aN5H4YLPa
```

With that it probably make sense to stick with your current approach. Is this something which could be fixed as part of #3573?

NEWS
bbannier: > I guess we could leave some commented out code in the template?

Yes, I believe that would be helpful, ideally with that comment explaining how `get_file_handle` would need to behave (maybe even expanding a bit of what is in the Zeek docs for e.g., `get_file_handle` which we should reference). I don't think we e.g., want to introduce code which counts files into the default template, and it seems hard to provide a reusable function.

testing/btest/spicy/file-analysis-data-in-concurrent.zeek
rsmmr: > You are right, I see that warning too, but also seemingly valid file IDs

Ok, got it. That's because it falls back to the previous built-in scheme to be able to continue still. That also means that existing analyses will keep working as before, people will just start seeing that warning now until scriptland gets updated--which seems better than breaking code.

> With that it probably make sense to stick with your current approach. Is this something which could be fixed as part of https://github.com/zeek/zeek/issues/3573?

Not quite following here: the PR's behavior for what tag to use in scriptland is so that `replaces` works transparently, which #3573 aims for too in a different context. So we wouldn't change anything further that this PR introduces. Unless I'm missing your point?

NEWS
rsmmr: Ok, I'll work on the template.

testing/btest/spicy/file-analysis-data-in-concurrent.zeek
bbannier: I think I got myself confused and don't know enough about what happens in scriptland elsewhere, but that this the tag we use here makes a difference at all tells me that something `replaces` doesn't completely replace the analyzer reachable under a tag (#3573 is about reaching an analyzer via its name which is probably a different but related aspect). I am closing this.

Issue #3430, opened 2023-11-04
Make Zeekygen warnings more easy to filter


ckreibich:
Zeekygen's warning messages are currently tricky to deal with because they're not immediately recognizable. It's handy to be able to filter these (to only see those, or see none of them) depending on the application setting. This could simply be a change to `:zeekygen::detail::DbgAndWarn()` that adds a `[zeekygen]` prefix to the reporter warnings and debug log entries.

ckreibich:
Tim flagged that while the code change is very manageable we're lacking tests for Zeekygen when things go wrong. So fixing this should include such testing, too.

PR #3705, opened 2024-04-25
Label zeekygen warnings as such

pbcullen:
Label warnings from zeekygen as such which makes them much easier to filter out, if desired.

### DIFF
--- a/src/zeekygen/Manager.cc
+++ b/src/zeekygen/Manager.cc
@@ -25,7 +25,7 @@ static void DbgAndWarn(const char* msg) {
         // be confusing.
         return;
 
-    reporter->Warning("%s", msg);
+    reporter->Warning("[zeekygen] %s", msg);
     DBG_LOG(DBG_ZEEKYGEN, "%s", msg);
 }
 


pbcullen:
@timwoj asked about testing. I ran all btests, but there must not be anything that triggers this. Let me see if I can figure out how to do that.

ckreibich:
Like @timwoj I just tried to produce zeekygen warnings and mostly failed. That's because most of them catch internal inconsistencies during parsing that one shouldn't really be able to trigger (which could mean they perhaps shouldn't just be warnings). Ironically, the few warnings I managed to produce also crash Zeek with some super simple scenarios that I'm about to open a separate ticket for. I can confirm that the existing btests don't operate at that level; they mostly verify generated ReST content. None of them generate those warnings.

IIRC, I originally created #3430 on your behalf Peter. So I have a few suggestions/requests:

- Could you look into what triggered those warnings and post a few examples?
- Zeek supports a `ZEEK_DISABLE_ZEEKYGEN` environment variable. We use it [in our btests](https://github.com/zeek/zeek/blob/master/testing/btest/btest.cfg#L36), except for Zeekygen's own tests. Could this be a way for you to suppress the warnings altogether?
- If we do end up prefixing Zeekygen-related reporter messages, we should do that consistently because there are additional places in the Zeekygen code that produce them. Many of them are errors, but not all:
```
$ cd src/zeekygen/
$ grep reporter- * | grep -v Error
Manager.cc:    reporter->Warning("%s", msg);
PackageInfo.cc:        reporter->InternalWarning("Zeekygen failed to open '%s': %s", readme_file.c_str(), strerror(errno));
PackageInfo.cc:        reporter->InternalWarning("Zeekygen error reading '%s': %s", readme_file.c_str(), strerror(errno));
ScriptInfo.cc:                reporter->InternalWarning("Zeekygen failed to get mtime of %s", it->c_str());
Target.cc:            reporter->InternalWarning("Zeekygen ID lookup failed: %s\n", it->GetID().c_str());
Target.cc:        reporter->InternalWarning(
Target.cc:                reporter->Warning("Failed to unlink %s: %s", f.c_str(), strerror(errno));
```

stevesmoot:
IDK if this is one of them but
# make sure we have erad loaded
@load ./erad.zeek
gets
warning in ./main.zeek, line 1: Discarded extraneous Zeekygen comment: make sure we have erad loaded
which self-documents as ZG, but doesnt have this style.

ckreibich:
@stevesmoot thanks, say more? As-is, this outputs nothing for me with any version I try. Was this with `-X`, and if so what was your zeekygen config? What Zeek version?

Does it go away when you run with `ZEEK_DISABLE_ZEEKYGEN=1` in your environment?

Fwiw, the only warning I could force was "discarded extraneous comment" as well. We could consider adding unit tests to Zeekygen to make testing stuff easier.

stevesmoot:
I may have misunderstood, those are from zeek. Fairly near tip but it shouldn’t matter

timwoj:
This is a candidate for the next set of patch releases.

pbcullen:
After further investigation, I think I'm going to close this PR. It seems likely that ZEEK_DISABLE_ZEEKYGEN would suffice.

timwoj:
Should we close https://github.com/zeek/zeek/issues/3430 as well?

ckreibich:
I'm thinking the original issue still stands, and adding a bit of a facade to the Zeekygen manager to prefix all of its messages would be easy and helpful. So I suggest we leave that open for now.

PR #3706, opened 2024-04-25
Fix ICMP caplen violations that could trigger out-of-bounds buffer access in rule matching

### LINKED ISSUE #3671: Crashes due to Invalid Pointer Read in RE_Match_State::Match with ICMP packets
### Bottom Line
Zeek happens to crash due to memory access violation when invoking zeek::detail::RE_Match_State::Match within ICMP Analyzer with an invalid packet length.

### Stack Trace
[0x0]   ntdll!ZwAlpcSendWaitReceivePort+0x14   0x7cbb8fd268   0x7ffedf355222   
[0x1]   ntdll!SendMessageToWERService+0x1ca   0x7cbb8fd270   0x7ffedf355009   
[0x2]   ntdll!ReportExceptionInternal+0xdd   0x7cbb8fd3d0   0x7ffede233ddc   
[0x3]   KERNEL32!WerpReportFaultInternal+0x3d8   0x7cbb8fdf40   0x7ffede2339ca   
[0x4]   KERNEL32!WerpReportFault+0xbe   0x7cbb8fe060   0x7ffedce955e0   
[0x5]   KERNELBASE!UnhandledExceptionFilter+0xbdb80   0x7cbb8fe0a0   0x7ffedf4046d5   
[0x6]   ntdll!RtlpThreadExceptionFilter+0x83   0x7cbb8fe1d0   0x7ffedf3eb4c6   
[0x7]   ntdll!RtlUserThreadStart$filt$0+0x94   0x7cbb8fe1d0   0x7ffedf3eb4c6   
[0x8]   ntdll!__C_specific_handler+0x96   0x7cbb8fe210   0x7ffedf3fff4f   
[0x9]   ntdll!RtlpExecuteHandlerForException+0xf   0x7cbb8fe280   0x7ffedf36eb64   
[0xa]   ntdll!RtlDispatchException+0x424   0x7cbb8fe2b0   0x7ffedf3ff82e   
[0xb]   ntdll!KiUserExceptionDispatch+0x2e   0x7cbb8fea00   0x7ff75ff1c8d2   
**[0xc]   zeek::detail::RE_Match_State::Match+0xe2   0x7cbb8ff1a0   0x7ff75ff140c6   
[0xd]   zeek::detail::RuleMatcher::Match+0xc6   0x7cbb8ff240   0x7ff75fb81e29   
[0xe]   zeek::packet_analysis::ICMP::ICMPSessionAdapter::MatchEndpoint+0x39   0x7cbb8ff340   0x7ff75fb83c43   
[0xf]   zeek::packet_analysis::ICMP::ICMPAnalyzer::DeliverPacket+0x503   0x7cbb8ff390   0x7ff75fb8f789**   
[0x10]   zeek::packet_analysis::IP::IPBasedAnalyzer::AnalyzePacket+0x4a9   0x7cbb8ff450   0x7ff75fcddca5   
[0x11]   zeek::packet_analysis::Analyzer::ForwardPacket+0x295   0x7cbb8ff5f0   0x7ff75fb911ec   
[0x12]   zeek::packet_analysis::IP::IPAnalyzer::AnalyzePacket+0x8dc   0x7cbb8ff6d0   0x7ff75fcddca5   
[0x13]   zeek::packet_analysis::Analyzer::ForwardPacket+0x295   0x7cbb8ff7d0   0x7ff75fb94962   
[0x14]   zeek::packet_analysis::IEEE802_11::IEEE802_11Analyzer::AnalyzePacket+0x122   0x7cbb8ff8b0   0x7ff75fcddca5   
[0x15]   zeek::packet_analysis::Analyzer::ForwardPacket+0x295   0x7cbb8ff8f0   0x7ff75fcdb875   
[0x16]   zeek::packet_analysis::Manager::ProcessPacket+0x185   0x7cbb8ff9d0   0x7ff75ff0ed18   
[0x17]   zeek::run_state::detail::dispatch_packet+0x348   0x7cbb8ffb50   0x7ff75fd34b44   
[0x18]   zeek::iosource::PktSrc::Process+0x24   0x7cbb8ffc00   0x7ff75ff0f5b8   
[0x19]   zeek::run_state::detail::run_loop+0x128   0x7cbb8ffc30   0x7ff76047c894   

### Root Cause
The "len" parameter passed by ICMPAnalyzer::DeliverPacket to zeek::detail::RE_Match_State::Match at [ICMP.cc:](https://github.com/zeek/zeek/blob/da0c1ca089660e8a0a496211e48d79c779cc2f4d/src/packet_analysis/protocol/icmp/ICMP.cc#L120) with the pointer "data" is too high:
`adapter->MatchEndpoint(data, len, is_orig);
`
Then, at [zeek::detail::RE_Match_State::Match](https://github.com/zeek/zeek/blob/master/src/RE.cc#L317) the crash happens when reading out of bounds ("bv" is "data"):
`ec = ecs[*(bv++)];
`

"len" and "data" are read from the IP header of the packet ([ICMP.cc](https://github.com/zeek/zeek/blob/da0c1ca089660e8a0a496211e48d79c779cc2f4d/src/packet_analysis/protocol/icmp/ICMP.cc#L61)):
```
const u_char* data = pkt->ip_hdr->Payload();
int len = pkt->ip_hdr->PayloadLen();
```

In the memory dump collected from one of the crashes, the value of "len" is 106:
image

However, "data" points to address 0x000002335e6c5f9e, and we can see that the actual length allocated at this address is less than 106 (0x6a):
image

image

### Resolution
ICMP Analyzer should generate a "weird" when the packet's payload size from the IP header is larger than the actual size of the packet's payload.

ckreibich:
I considered adding a testcase based on the reproducer in the issue, but it feels odd to add something that technically relies on libpcap's buffer sizing to trigger. Holler if you disagree, though.

I looked at the other IP-based packet parsers and didn't see obviously similar patterns. Not directly related, but I came across this while digging through the code: it looks suspicious that [RuleMatcher::InitEndpoint()](https://github.com/zeek/zeek/blob/master/src/RuleMatcher.cc#L679-L768) has a `caplen` argument but never uses it.

Resolves #3671.

### DIFF
--- a/src/packet_analysis/protocol/icmp/ICMP.cc
+++ b/src/packet_analysis/protocol/icmp/ICMP.cc
@@ -92,11 +92,14 @@ void ICMPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int rema
     c->SetLastTime(run_state::current_timestamp);
     adapter->InitEndpointMatcher(ip.get(), len, is_orig);
 
-    // Move past common portion of ICMP header.
+    // Move past common portion of ICMP header. BuildConnTuple() verified that
+    // the header is fully present.
     data += 8;
     remaining -= 8;
     len -= 8;
 
+    // The ICMP session adapter only uses len to signal endpoint activity, so
+    // caplen vs len does not matter.
     adapter->UpdateLength(is_orig, len);
 
     if ( ip->NextProto() == IPPROTO_ICMP )
@@ -112,12 +115,12 @@ void ICMPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int rema
     // handling those properly.
     pkt->session = c;
 
-    ForwardPacket(len, data, pkt);
+    ForwardPacket(std::min(len, remaining), data, pkt);
 
     if ( remaining >= len )
         adapter->ForwardPacket(len, data, is_orig, -1, ip.get(), remaining);
 
-    adapter->MatchEndpoint(data, len, is_orig);
+    adapter->MatchEndpoint(data, std::min(len, remaining), is_orig);
 }
 
 void ICMPAnalyzer::NextICMP4(double t, const struct icmp* icmpp, int len, int caplen, const u_char*& data,


src/packet_analysis/protocol/icmp/ICMPSessionAdapter.cc
```diff
@@ -51,8 +51,13 @@ void ICMPSessionAdapter::UpdateLength(bool is_orig, int len) {
     int& len_stat = is_orig ? request_len : reply_len;
     if ( len_stat < 0 )
         len_stat = len;
-    else
+    else {
         len_stat += len;
+#ifdef DEBUG
+        if ( len_stat < 0 )
+            reporter->Warning("wrapping around for ICMP %s length", is_orig ? "request" : "reply");
+#endif
```
bbannier, line 59: Signed integer overflow is undefined behavior in C++ and if we do overflow all bets are off, so this check is too late. If potential overflow is a concern we have to _always check for it before we trigger UB_, e.g.,

```cpp
        // Check for potential overflow.
        //
        // NOTE: Checking `MAX_INT - len` is insufficient since it can
        // underflow for possible negative `len` and/or `len_stat`. Treat
        // potential signs of `len` separately.
        //
        // TODO(bbannier): None of this would be necessary if both `len_stat`
        // and `len` would arrive here unsigned. In that case we wouldn't need
        // to deal with UB since unsigned overflow is defined in C++. This
        // probably would require new validations where these values are
        // extracted.
        if ( len >= 0 ) {
            if ( std::numeric_limits<int>::max() - len > len_stat )
                reporter->FatalError("OH MY!");
        }
        else {
            if ( std::numeric_limits<int>::min() - len > len_stat )
                reporter->FatalError("OH MY!");
        }
        len_stat += len;
```

timwoj (approved):

Issue #3700, opened 2024-04-24
Additional `conn_id` fields don't make it to the files.log


JustinAzoff:
I thought this was related to #3028 again but it is a different issue.

Given this script:

```
redef record conn_id += {
        ## The VLAN that the connection is seen on.
        hello:       string &log &optional;
};

event new_connection(c: connection) &priority=4
{
	c$id$hello = "world";
}
```

The `hello` field shows up in all logs except for the files.log.  I believe this is the correct fix to make this work:

```
diff --git a/scripts/base/frameworks/files/main.zeek b/scripts/base/frameworks/files/main.zeek
index baefdfeec..31da1a7b5 100644
--- a/scripts/base/frameworks/files/main.zeek
+++ b/scripts/base/frameworks/files/main.zeek
@@ -603,7 +603,7 @@ event file_state_remove(f: fa_file) &priority=-10
                # by reference in Log::write() rather than by copy.
                local info = |f$conns| > 1 ? copy(f$info) : f$info;
                info$uid = c$uid;
-               info$id = cid;
+               info$id = c$id;
                Log::write(Files::LOG, info);
                }
        }
```

Since `c$id` is the record that was updated in the `new_connection` event and `cid` is the internal conn_id record that never got the update.

0xxon:
Yup, that fix looks right to me - I opened a PR.

PR #3710, opened 2024-04-29
Fix cid propagation into files.log

0xxon:
Changes to the connection id were not propagated to files.log in all cases.

Fixes GH-3700

I did not add a test case as this is highly specific and did not seem worthwhile in this case - but I am happy to add it if so desired.


timwoj (approved):

timwoj:
Is this worth pulling into the upcoming patch releases?

awelzel:
> Is this worth pulling into the upcoming patch releases?

+1 - thanks!

PR #3711, opened 2024-04-29
Refining Val "footprint" to equate long strings with multiple objects

vpax:
When using the relatively new `val_footprint()` BiF to assess relative memory impact, we're finding that sometimes we miss major memory issues arising from very long strings. This PR tweaks the "footprint" logic to convert the length of a string `s` into an additional footprint impact of `floor(|s| - sizeof(Val))`.

Note that because `sizeof(Val)` can vary across different environments, the BTest for this just makes sure that with a long string, the footprint grows, rather than printing its exact value.

I'll be putting in a companion PR for the documentation shortly.


vpax:
Doc PR now available: https://github.com/zeek/zeek-docs/pull/260

timwoj (approved):

PR #3712, opened 2024-04-29
Move the options from policy/tuning/defaults to actual Zeek defaults, deprecate that package

### LINKED ISSUE #3389: Deprecate tuning/defaults policy package.  Remove with Zeek 7.1
All the tuning/defaults package does is:
```
redef FileExtract::default_limit = 104857600;
redef frag_timeout = 5 min;
```

...and produce the warning about empty `Site::local_nets`.

The latter has become mood after using private address spaces as default. We likely won't extend the list of modified options (it hasn't happened in the past 10 years).

Christian's idea for the former were to use them as defaults generally. We could also open-code them in `local.zeek` directly if that's too controversial (the file extraction limit was moved from the module to tuning/defaults in 17d0ecd388581d702c27a4eea259cc5bcc0ca465, for example).

timwoj:
Fixes #3389

This PR moves all of the settings from the tuning/defaults policy into their original definition locations, using them as the new defaults for Zeek. The reasoning here is that cluster users were already loading this policy by default from local.zeek, and so were already using all of these settings as they are. It also removes the settings from those policy files, and marks them all as deprecated for removal in v7.1.


ckreibich (approved):
Thanks Tim, LGTM! My only request is to make sure there's a NEWS entry for this one. I took a look at the docs and I see nothing where they'd contradict the new defaults.

PR #3713, opened 2024-04-30
Cleanup Spicy's `replaces`

rsmmr:
- **Move enabled/disabled functionality from analyzers into `Component` base class API.**
- **Add component API to transparently remap one component to another one.**
- **Spicy: Wenn replacing an analyzer, add a component mapping.**
- **Fix packet analyzer replacement.**


rsmmr:
@bbannier @timwoj Not quite sure who's the best to review so tagging both of you.

src/analyzer/Manager.cc
```diff
@@ -110,7 +110,7 @@ void Manager::DumpDebug() {
 void Manager::Done() {}
 
 bool Manager::EnableAnalyzer(const zeek::Tag& tag) {
-    Component* p = Lookup(tag);
+    Component* p = Lookup(tag, false);
```
bbannier, line 113: Do we somewhere else take mappings into account when toggling an analyzer? I think there should be a way to do that (worst case: through an extra argument here and the bifs calling it). I am thinking of a use case where a user loads e.g., spicy-http (which replaces the building `HTTP`), and they should have a way to transparently specify from scriptland (e.g., `local.zeek`) that they do not want HTTP analysis without having to be aware of all possible mappings.

Similarly below.

src/file_analysis/file_analysis.bif
```diff
@@ -41,7 +41,7 @@ function Files::__set_reassembly_buffer%(file_id: string, max: count%): bool
 ## :zeek:see:`Files::enable_analyzer`.
 function Files::__enable_analyzer%(tag: Files::Tag%) : bool
 	%{
-	auto c = zeek::file_mgr->Lookup(tag->AsEnumVal());
+	auto c = zeek::file_mgr->Lookup(tag->AsEnumVal(), false);
```
bbannier, line 44: Similar argument as for above, I believe there should be a way to transparently toggle all mapped analyzers.

src/packet_analysis/Manager.cc
```diff
@@ -82,14 +82,14 @@ AnalyzerPtr Manager::GetAnalyzer(const std::string& name) {
 }
 
 bool Manager::EnableAnalyzer(EnumVal* tag) {
-    Component* c = Lookup(tag);
+    Component* c = Lookup(tag, false);
```
bbannier, line 85: Similarly, it would be great to allow users to transparently work on all mapped analyzers.

src/plugin/ComponentManager.h
```diff
@@ -119,24 +119,59 @@ class ComponentManager {
...
+     *
+     * @param old The original component tag.
+     * @param new_ The new component tag.
+     */
+    void AddComponentMapping(const zeek::Tag& old, const zeek::Tag& new_) {
+        if ( old != new_ ) {
+            component_mapping_by_src[old] = new_;
+            component_mapping_by_dst[new_] = old;
+        }
+    }
```
bbannier, line 160: Does this actually work when stacking multiple EVT `replaces` on top of each other, e.g., in this order have

- `spicy-http` `replaces` `HTTP`, and 
- `my-http` `replace` `HTTP` as well
   - even worse: `my-http` `replaces` `HTTP`

I feel we really need  a stack to model this, indexed by the most basic replacement (in above example: `HTTP`).

src/plugin/ComponentManager.h
```diff
@@ -119,24 +119,59 @@ class ComponentManager {
...
+
+    /**
+     * Returns true if a given component has a mapping to different one in place.
+     *
+     * @param tag The component tag to check.
+     */
+    auto HasComponentMapping(const zeek::Tag& tag) { return component_mapping_by_src.count(tag); }
```
bbannier, line 167:
```suggestion
    auto HasComponentMapping(const zeek::Tag& tag) const { return component_mapping_by_src.count(tag); }
```

src/plugin/ComponentManager.h
```diff
@@ -119,24 +119,59 @@ class ComponentManager {
...
+
+    /**
+     * Returns true if a given component is mapped to from a different one.
+     *
+     * @param tag The component tag to check.
+     */
+    auto ProvidesComponentMapping(const zeek::Tag& tag) { return component_mapping_by_dst.count(tag); }
```
bbannier, line 174: With the `auto` return type this would return a `size_t` from `count`, suggest to type explicitly.

```suggestion
    bool ProvidesComponentMapping(const zeek::Tag& tag) const { return component_mapping_by_dst.count(tag); }
```

bbannier (commented):
I wrote down a couple points regarding the API, but I cannot comment on how this fits into Zeek internals around e.g., lifecycle management.

src/plugin/ComponentManager.h
timwoj: Alternatively, changing it to `!= 0` would be more explicit about what's happening here.

src/plugin/Component.h
```diff
@@ -118,6 +118,25 @@ class Component {
...
+    /**
+     * Returns true if the component is currently enabled and hence
+     * available for use.
+     */
+    bool Enabled() const { return enabled; }
+
+    /**
+     * Enables or disables this component. Derived classes may override this if
+     * they need to initiate additional actions, but must then call the base
+     * class version.
```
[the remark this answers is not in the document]
rsmmr: Well, in principle yes, but (a) unfortunately pretty much nothing of the Zeek-side analyzer life-cycle logic is documented anywhere anyways and (b) while this moves `SetEnabled()` into the base class, it doesn't change semantics re/ `Initialize` compared to what we already had before for analyzers. Given those, I'm not really inclined to go further here, as I don't think it would help much.

src/analyzer/Analyzer.cc
```diff
@@ -801,4 +801,30 @@ TEST_SUITE("Analyzer management") {
...
+        auto ssh_is_imap = zeek::analyzer_mgr->InstantiateAnalyzer("SSH", conn.get());
+        CHECK_EQ(ssh_is_imap->GetAnalyzerTag(), imap->GetAnalyzerTag()); // SSH is now IMAP
+
+        // orderly cleanup through connection
+        auto* tcp = new zeek::packet_analysis::TCP::TCPSessionAdapter(conn.get());
+        conn->SetSessionAdapter(tcp, nullptr);
+        tcp->AddChildAnalyzer(ssh);
+        tcp->AddChildAnalyzer(imap);
+        tcp->AddChildAnalyzer(ssh_is_imap);
+        conn->Done();
```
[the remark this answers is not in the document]
rsmmr: > None of this will run if any of the above REQUIRE fails.

That cleanup logic is mainly there to avoid the leak checker triggering on the test. If the test fails for another reason (like the `REQUIRE`), the leak doesn't really matter.

>  were great if testing with zeek::Connection would be easier and safer 

Agree, it would be great. :-) But that feels out of scope here, it's an existing issue, the new test just follows the existing pattern.

src/analyzer/Manager.cc
rsmmr: Toggling HTTP as a protocol is a different thing from toggling a specific HTTP analyzer. This API has always been referring to specific analyzers, i.e., here one particular implementation of HTTP, that can be toggled. That's what both the analyzer tags und names refer to. So this doesn't change anything, nor should mismatch with expectations afaict.

src/file_analysis/file_analysis.bif
rsmmr: Same answer here too.

src/plugin/ComponentManager.h
rsmmr: Sorry, I don't want to go there; at least not for now until we have an actual need/use-case for more complexity here. There aren't that many situations where `replaces` applies in the first place; the standard case is using a Spicy analyzer instead of a built-in, and there are only so many possible instances of that. I don't really see that we need to support chains of analyzers replacing each other; if you have multiple Spicy analyzers doing the same thing, just load the one you need. Otherwise this gets easily really complex and out-of-hand: next thing would be cycle-detection: what if each analyzer disables another one so that eventually have none anymore?

src/plugin/Component.h
```diff
@@ -64,7 +64,7 @@ class Component {
     /**
      * Destructor.
      */
-    virtual ~Component() = default;
+    virtual ~Component();
```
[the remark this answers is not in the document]
rsmmr: I had changed it to fix a missing vtable but just tried reversion and it's working fine with `default` now.

src/analyzer/Analyzer.cc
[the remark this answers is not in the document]
bbannier: FTR, I ran the test with the cleanup code removed and it failed even without LSAN. Fine either way though.

src/plugin/ComponentManager.h
rsmmr: That I can do!

src/analyzer/Analyzer.cc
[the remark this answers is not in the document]
bbannier: I ran `zeek --test` which gave me an `abort()` triggered by this not very RAII-friendly code (to be fair, and it flagged exactly what was the problem, i.e., that `Conn::Done` was not called 🤷):

https://github.com/zeek/zeek/blob/966d411ed52ff23a48c9eca88ef7aca9068b1b02/src/Conn.cc#L71-L72

timwoj (approved):
Looks good to me after the revisions. Re: the unit test we run our unit tests on CI under ASAN, so any leak checking will definitely happen there. I agree that Conn could have better RAII handling, but this isn't the place to fix it.

PR #3716, opened 2024-05-02
Enable ZAM operation specifications to reside in multiple files

vpax:
This PR is a companion to https://github.com/zeek/cmake/pull/107 (which should be merged at the same time). It's a simple change to how ZAM operation specifications are maintained, in support of future factoring.


timwoj (approved):
I merged the other two PRs already and updated the submodules here. If it passes CI I'll merge this one too.

PR #3720, opened 2024-05-04
Making the SSH analyzer robust to half-duplex connections

vpax:
Currently if the SSH analyzer processes a half-duplex SSH connection, it generates zillions of `analyzer.log` entries full of BinPAC exceptions because the SSH version isn't in an expected state. This PR adds a `default` to the switch that causes these exceptions, where the default just gobbles up the rest of the input. (Better maybe would be to disable the analyzer, but that's beyond my feeble BinPAC syntax chops. Happy to revise if someone can provide that.)

While the vanilla diff makes it look like quite a bit changed in the `.pac` file, it's really just one line. The big block later in the file was mis-indented, which I've remedied.

This fix is very helpful in some operational environments, where `analyzer.log` can wind up being bigger than `ssh.log` (!).


timwoj (approved):
This looks like it requires an update to the zeek-testing-private repo.

src/analyzer/protocol/ssh/ssh-protocol.pac
```diff
@@ -37,6 +37,7 @@ type SSH_Key_Exchange(is_orig: bool) = record {
 	key_ex: case $context.connection.get_version() of {
 		SSH1 -> ssh1_msg : SSH1_Key_Exchange(is_orig, packet_length);
 		SSH2 -> ssh2_msg : SSH2_Key_Exchange(is_orig, packet_length);
+		default -> terminate : bytestring &restofdata &transient;
```
awelzel, line 40: Chatting with @J-Gras: He observes this causing the SSH analyzer to not produce violations, when testing with HTTP traffic on port 22, for example.

For 7.0 we're aiming to revert this.

> This fix is very helpful in some operational environments, where analyzer.log can wind up being bigger than ssh.log (!).

It's true that analyzer.log can produce a lot of output and might be too noisy for operational environments. Violations should only be logged until the SSH analyzer is disabled via dpd.zeek. To reduce the volume of the log, there's other mechanisms. These range from disabling the log completely to ignoring the SSH analyzer specifically. Also, log policy hooks could probably help for rate limiting, too.

After further discussion and pondering, prior to 3769ed6c6687e5ecf93c3668bebe3c3f7da3d0fb, `get_version()` would return what had been seen from the client.

One idea for a different fix of the half-duplex case would be to change `get_version()` to return either client or server version if `version_` wasn't determined yet and weird if it's the special 1.99 case.

vpax: I agree that the original fix is wrong and should be reverted. I like the idea you float about the `get_version()` fix. Another that would work (but I don't know how to do in BinPAC) is have the `default` inform DPD about an analyzer mismatch and detach.

PR #3727, opened 2024-05-08
Spicy: Fix service reporting for replaced analyzers.

rsmmr:
We accidentally applied analyzer mappings when looking up an
analyzer's name from scriptland.


bbannier (approved):
Thanks for the fast fix!

Does it make sense to add a test for this?

rsmmr:
> Does it make sense to add a test for this?

Yeah, done.

PR #3729, opened 2024-05-08
CI: Replace Ubuntu 23.10 with Ubuntu 24.04

timwoj:
Ubuntu 24.04 came out in April 2024, and replaces Ubuntu 23.10 as the next LTS. This also updates the asan and ubsan jobs to use Ubuntu 22. They were previously on Ubuntu 20.


ckreibich (approved):
LGTM Tim — as usual we should mirror this in broker / binpac / zeek-aux as well, and backport. If you can't get to that I can do it, just let me know.

PR #3738, opened 2024-05-14
Bump spicy submodule to v1.10.1

timwoj:


bbannier (approved):
The failure for `ubuntu22_spicy_head` is expected since it explicitly bumps `auxil/spicy` to its `HEAD` version. Spicy's `HEAD` is incompatible with release/6.2.

Since that CI job is more a canary for upcoming incompatible changes it might make sense to not run it at all for release branches, or also not when bumping Spicy itself.

PR #3740, opened 2024-05-15
CI: Disable spicy_head task for release branches

timwoj:
This PR disables the spicy-head task on CI for release branches. This test builds against spicy's HEAD commit, which may rely on other changes being in the branch under test. This is especially true on release branches.


.cirrus.yml
```diff
@@ -324,7 +324,12 @@ ubuntu22_spicy_head_task:
   upload_binary_artifacts:
     path: build.tgz
   benchmark_script: ./ci/benchmark.sh
-  << : *BENCHMARK_ONLY_IF_TEMPLATE
+  only_if: >
+    ( $CIRRUS_REPO_NAME == 'zeek' || $CIRRUS_REPO_NAME == 'zeek-security' ) &&
+    ( $CIRRUS_BRANCH !=~ 'release/.*' ) &&
+    ( $CIRRUS_CRON == 'benchmark-nightly' ||
+      $CIRRUS_PR_LABELS =~ '.*fullci.*' ||
+      $CIRRUS_PR_LABELS =~ '.*benchmark.*' )
```
bbannier, line 332: This probably works, but seems overly complicated to me.

```
    ( $CIRRUS_REPO_NAME == 'zeek' || $CIRRUS_REPO_NAME == 'zeek-security' ) &&
```

What is the use of checking the name of the repo? It would seem this task would be fine to run in forks, regardless of how they are named.


```
    ( $CIRRUS_CRON == 'benchmark-nightly' ||
      $CIRRUS_PR_LABELS =~ '.*fullci.*' ||
      $CIRRUS_PR_LABELS =~ '.*benchmark.*' )
```

I would argue that this task still makes no sense and would very likely fail on e.g., PRs labeled `fullci` or `benchmark` iff they are on a release branch.

WDYT about instead just using [`skip`](https://cirrus-ci.org/guide/writing-tasks/#:~:text=the%20skip%20keyword%20allows%20to%20skip%20execution%20of%20a%20task%20and%20mark%20it%20as%20successful.%20for%20example%2C%20you%20may%20want%20to%20skip%20linting%20if%20no%20source%20files%20have%20changed%20since%20the%20last%20successful%20run.%20)? The way it is documented is a little wonky, but it would have thethe effect that the task is not run, but can still be triggered manually from the Cirrus UI (though by default it shouldn't be useful at all), i.e., replace the whole `only_if` with

```
  skip: $CIRRUS_BRANCH ==~ 'release/.*'
```

bbannier (approved):

.cirrus.yml
timwoj: I was basically copying the condition from the `BENCHMARK_ONLY_IF_TEMPLATE` setting, plus checking for release branches, since that's what already existed for that job. I could see just adding the `skip` though. I'll switch it over to that.

PR #3741, opened 2024-05-15
CI: Use llvm-cov-18 on Cirrus for building coverage data

timwoj:
Coverage builds broke after the change to use clang-18 for sanitizer builds. The reason is that the default `gcov` binary which comes from gcc expects data to be in a different version than the data generated by clang. Instead we should be using `llvm-gcov-18 gcov` in place of the `gcov` command when doing Cirrus builds.

### DIFF
--- a/testing/coverage/code_coverage.sh
+++ b/testing/coverage/code_coverage.sh
@@ -21,6 +21,21 @@ BASE="$(cd "$CURR" && cd ../../ && pwd)"
 TMP="${CURR}/tmp.$$"
 mkdir -p $TMP
 
+GCOV_CMD=""
+if [ -n "${CIRRUS_TASK_NAME}" ]; then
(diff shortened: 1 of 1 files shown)


testing/coverage/code_coverage.sh
```diff
@@ -21,6 +21,21 @@ BASE="$(cd "$CURR" && cd ../../ && pwd)"
 TMP="${CURR}/tmp.$$"
 mkdir -p $TMP
 
+GCOV_CMD=""
+if [ -n "${CIRRUS_TASK_NAME}" ]; then
+    GCOV_CMD="$(which llvm-cov-18)"
```
ckreibich, line 26: Just wondering — would `llvm-cov` suffice here, to be future-proof?

ckreibich (approved):
LGTM Tim. Much better to check for gcov before doing any work, too.

PR #1255, opened 2020-10-27
topic/vern/ast duplicate

vpax:
I'm trying to carve up the massive script optimization work into more bite-size pieces.  This PR is for adding AST duplication functionality.  It's a precursor to adding inlining of function bodies.

I've isolated the new `Stmt`/`Expr` method definitions into a new subdirectory, `script_opt`, as Robin had previously suggested.  The changes to `Stmt.h` and `Expr.h` relevant to script optimization are set off from the other class declarations, as a model of how we'll do this going forward.  It would be nice to know soon whether this style will be acceptable, so I can know whether to use it as I work in parallel on migrating more elements of the main optimization branch.

There are also a number of simple changes to other files, mainly to tie error messages to the original script locations where they occur, rather than to the location information available at the time of duplication; and a couple of bug fixes where types were incorrectly set but in ways that didn't lead to manifesting problems prior to adding duplication.

By the way, I branched this off  of topic/vern/brovalunion-change (https://github.com/zeek/zeek/pull/1231), but now that this branch is done, I think it can easily stand alone, if desirable.

PR #3743, opened 2024-05-16
Changing "add" and "delete" to be expressions rather than statements

vpax:
In https://github.com/zeek/zeek/discussions/3684 we discussed (among other things) possible syntax for operations associated with a new `list` container type. The front-runner for popping elements from either the front or the back of a `list` is using `delete` (`delete L[0]` for the front and `delete L[-1]` for the back). Since a very common operation would be pop-list-element-and-do-something-with-it, it would be helpful if one could write `elem = delete L[0]`, for example, to pop the front of `L` and return it in `elem`. However, to do that `delete` needs to be an AST _expression_ node. Currently, it's an AST _statement_ node, and those don't have values associated with them.

This PR lays the groundwork for such a change. It changes `delete` AST nodes to be _expressions_ that (potentially) return values. For symmetry, because it would be a bit weird to do otherwise, it also likewise changes `add` AST nodes.

The question is then what values should these expressions return? This PR kicks that can down the road: they currently behave as "void" types that don't return anything.

While the change is conceptually simple (and backward compatible), it winds up moving a bunch of code from various `Stmt` files into `Expr` files, along with small accompanying tweaks, so quite a bit gets touched even though nothing deep is going on.
 
I've also included two other minor changes. One of these is a bug fix to prevent bad initializers from leading to crashes in some circumstances. The other is to change expressions that render as words to include a space separating that word from their operand. This last (isolated to a separate commit) leads to a bunch of BTest changes that look voluminous but are in fact trivial.


src/Expr.cc
```diff
@@ -42,6 +42,8 @@ const char* expr_name(ExprTag t) {
         "-",
         "+",
         "-",
+        "add ",
```
timwoj, line 45: Maybe add a comment to the header of this method that explains why some of them have spaces.

src/Expr.h
```diff
@@ -700,6 +706,41 @@ class AddExpr final : public BinaryExpr {
...
+class AggrAddDelExpr : public UnaryExpr {
+public:
+    explicit AggrAddDelExpr(ExprTag _tag, ExprPtr _e) : UnaryExpr(_tag, std::move(_e)) {}
+
+    bool IsPure() const override { return false; }
+
+    // Optimization-related:
+    bool IsReduced(Reducer* c) const override { return HasReducedOps(c); }
+    bool HasReducedOps(Reducer* c) const override { return op->HasReducedOps(c); }
+    ExprPtr Reduce(Reducer* c, StmtPtr& red_stmt) override;
```
timwoj, line 719: This is implemented for the script-opt case. Should it have a no-op implementation for the non-script-opt case? Same for the duplicate methods below.

src/script_opt/ZAM/Expr.cc
```diff
@@ -78,6 +82,56 @@ const ZAMStmt ZAMCompiler::CompileAppendToExpr(const AppendToExpr* e) {
...
+        int field = op->AsFieldExpr()->Field();
+        return DelFieldVi(aggr, field);
+    }
+
+    auto index_list = op->GetOp2();
+
+    if ( index_list->Tag() != EXPR_LIST )
+        reporter->InternalError("non-list in \"delete\"");
+
+    auto internal_ind = std::unique_ptr<OpaqueVals>(BuildVals(index_list->AsListExprPtr()));
```
timwoj, line 131: Is this `unique_ptr` just because `DelTableV0` doesn't destroy the pointer?

src/Expr.cc
vpax: Make sense, will do

src/Expr.h
vpax: I'm not following. The usual pattern for `Expr.h` is to include these methods for script optimization (which is all that they're used for). The implementations are in `script_opt/Expr.cc`. The functionality isn't called when running the interpreter.

src/script_opt/ZAM/Expr.cc
vpax: Yes. `DelTableV0` is auto-generated by `gen-zam`, and those methods work on raw pointers.

(Note, this code is old, it just shows up here because it was moved from `src/script_opt/ZAM/Stmt.cc` to here.)

src/Expr.cc
vpax: Now pushed.

src/Expr.h
vpax: @timwoj I believe this is now the one pending issue. If easier to chat about over Zoom, just let me know.

timwoj: Ok, I see what's happening here now. I can't say that I'm a fan of splitting the implementation of these classes into two places. It just leads to having to dig around for an implementation instead of looking in a single place. It's been like that though, so I guess leave it as it is.

timwoj (approved):
Squashed and rebased. I'll merge it after it gets through CI.

src/Expr.h
vpax: FWIW, that split was originally asked for by Robin, per https://github.com/zeek/zeek/pull/1255

PR #3745, opened 2024-05-17
Copy timestamp from file object

mvhensbergen:
In some cases, e.g. running zeek on short pcaps as opposed to continuous packet streams, network_time() may not equal the time that was used when generating the file object.

This results in the pe.log entry having a different timestamp than its corresponding files.log entry which is strange as they refer to the exact same file.

By explicitly copying the files original timestamp in the pe section we ensure that there can be no discrepancy between the two which helps correlation.


martinvanhensbergen:
I noticed this behaviour on zeek 5.2. I am not too sure if this patch is consistent with the zeek philosopy of always using network_time for ts, but I believe that this change may be warranted as the pe.log is a derivative log of the main files.log

0xxon:
Do you randomly have a pcap that you can share that shows this behavior (and could be added as a test-case)?

mvhensbergen:
I have isolated a TCP stream that triggers this behaviour and also verified that it is shows this behaviour on zeek 6.2.1. However, I do not have clearance (yet) to disclose it but I will try to see if I can (or create a different pcap to see if triggers).

Even without a pcap, what is your opinion on this change? I'd say that since pe.log is a derived log from files.log it makes sense to copy the timestamp from the original object. At the same time I haven't looked into why network_time would give two different values.

mvhensbergen:
I have created a new pcap and will share shortly.

mvhensbergen:
```
root@e8e60950cd7a:/# cat pe.log
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	pe
#open	2024-05-23-14-03-49
#fields	ts	id	machine	compile_ts	os	subsystem	is_exe	is_64bit	uses_aslr	uses_dep	uses_code_integrity	uses_seh	has_import_table	has_export_table	has_cert_table	has_debug_data	section_names
#types	time	string	string	time	string	string	bool	bool	bool	bool	bool	bool	bool	bool	bool	bool	vector[string]
1716472371.825234	FyNlOv4xQHBf68TIf5	I386	1157435187.000000	Windows XP x64 or Server 2003	WINDOWS_GUI	T	F	F	T	F	T	T	T	T	T	.text,.data,.rsrc,.reloc
#close	2024-05-23-14-03-49
root@e8e60950cd7a:/# grep FyNl files.log
1716472371.824801	FyNlOv4xQHBf68TIf5	C1UTf6KAHtghsBHF1	192.168.2.1	54101	192.168.2.55	80	HTTP	0	PE	application/x-dosexec	-	0.088195	T	F	759632	759632	0	0	F	-	-	-	-	-	-	-
root@e8e60950cd7a:/# 

```

[pe_files_timestamp.pcap.zip](https://github.com/zeek/zeek/files/15417989/pe_files_timestamp.pcap.zip) zeek -Cr /pe_files_timestamp.pcap

mvhensbergen:
Hi! I will try and make a testcase tomorrow and add it to this PR. 

On a side note, is the fact that we see different time values explainable? I guess it has something to do with the pcap being very short and thus the network time function behaves a bit differently?

0xxon:
> On a side note, is the fact that we see different time values explainable? I guess it has something to do with the pcap being very short and thus the network time function behaves a bit differently?

I honestly am not sure :)

0xxon:
Actually, I kind of like your solution, so I'm just going to keep it, without the comment.

In case you need this in the future, or if you are curious: you can set the canonification script that btest-diff uses by setting the `TEST_DIFF_CANONIFIER` environment variable. Some tests do this to change it to a different canonification (like, e.g., sorting) - or to turn it off.

PR #3746, opened 2024-05-18
Fix for suppressing SMB logging of previously-logged files

vpax:
`scripts/base/protocols/smb/main.zeek` has logic to avoid generating multiple logs for the exact same file. The logic used for this incorporates all of the times associated with the file (creation, modification, ...) _including_ access time. However, when the same exact file is read multiple times, the one thing that might change in its meta-data is access time. This simple patch tweaks the logic to not include access time.


timwoj (commented):
Is there any concern with the repeated access not being indicative of someone trying to check whether a file has changed? This change would prevent that kind of detection, wouldn't it?

vpax:
I don't believe it would. The point is that each time they check, the access time will be updated to reflect that they looked.

timwoj (approved):

PR #3750, opened 2024-05-23
Update TLS consts, mainly new named curves.

0xxon:
This updates the TLS consts, and also adds a test for X25519Kyber768Draft00 (post-quantum key agreement)


timwoj (approved):
This is good to get in before 7.0.

PR #3752, opened 2024-05-24
Add BiFs to retrieve numeric ID for a given connection and analyzer

ckreibich:
See `NEWS` addition for the details. This came out of a support thread on Slack, where I realized that as far as I could discern there's not currently a way to alter analyzer activation via `disable_analyzer()` at this level.

For consistency I've changed the reporter error function called in `disable_analyzer()` since it provides better context.


NEWS
```diff
@@ -39,6 +39,27 @@ New Functionality
...
+		{
+		if ( no_http_for_this_conn_wanted(c) )
+			{
+			local aid = lookup_conn_analyzer_id_by_tag(c$id, Analyzer::ANALYZER_HTTP);
+			if ( aid > 0 )
+				disable_analyzer(c$id, aid, T, T);
+			}
+		}
+
+  The ``lookup_conn_analyzer_id_by_name()`` BiF does the same, but looks up the
```
vpax, line 60: Can't this one just be done via composition, using the first one plus the existing `Analyzer::get_tag()`?

0xxon (approved):

NEWS
ckreibich: True! Thank you for flagging that. I'll drop that one.

ckreibich: I've renamed the one BiF remaining to `lookup_connection_analyzer_id()`, for symmetry with the existing `lookup_connection()`.

PR #3755, opened 2024-05-25
CI updates

ckreibich:
CentOS Stream 8 is about to EOL so I'm including its removal here.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -166,17 +166,17 @@ env:
 # Linux EOL timelines: https://linuxlifecycle.com/
 # Fedora (~13 months): https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
 
+fedora40_task:
+  container:
+    # Fedora 40 EOL: Around May 2025
+    dockerfile: ci/fedora-40/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
+
 fedora39_task:
   container:
     # Fedora 39 EOL: Around Nov 2024
     dockerfile: ci/fedora-39/Dockerfile
-    << : *RESOURCES_TEMPLATE
-  << : *CI_TEMPLATE
-
-fedora38_task:
-  container:
-    # Fedora 38 EOL: Around May 2024
-    dockerfile: ci/fedora-38/Dockerfile
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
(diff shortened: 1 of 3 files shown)


timwoj:
The `asan_sanitizer` build failed here because of the coverage issues, which should be fixed by https://github.com/zeek/zeek/pull/3756 now.

ckreibich:
I pushed a rebase just to iron that one out. I'll look at the Fedora 40 one shortly.

timwoj:
Can we land the centos part of this and handle fedora 40 later?

awelzel:
@bbannier / @ckreibich - I've pushed a follow-up to compile sqlite3.c with `-Wno-error`. Seems sensible to me in general given we just ship vanilla upstream.

bbannier:
> @bbannier / @ckreibich - I've pushed a follow-up to compile sqlite3.c with `-Wno-error`. Seems sensible to me in general given we just ship vanilla upstream.

The `-Werror` addition seems fine to me though none of this would be an issue if we didn't build external deps with our own flags, e.g., with our custom vcpkg setup, I'd love to see that some day.

I cannot comment on the sqlite changes. The branch's commit message talks about bumping to 3.45.0, and the merge about 3.46.0. There are no instructions on how this patch was generated (e.g., what was the previous commit), so I cannot validate the bump.

awelzel:
> The -Werror addition seems fine to me though none of this would be an issue if we didn't build external deps with our own flags, e.g., with our custom vcpkg setup, I'd love to see that some day.

Thanks. I'm going to move this forward then.

awelzel (approved):

ckreibich:
Thanks @awelzel, I think that's the right call. I've left a post on the [SQLite forum](https://www.sqlite.org/forum/forumpost/f68e85b2c9) — perhaps they'll pick it up.

Fwiw, I did not imagine the `-O3` we talked about ... it's the default optimization level for build type "release" on my system (via `CMAKE_CXX_FLAGS_RELEASE`), and it's what triggers the warning:
```
$ gcc --version
gcc (GCC) 14.1.1 20240607 (Red Hat 14.1.1-5)

$ gcc -c -Wall -Werror -O2 sqlite3.c
$

$ gcc -c -Wall -Werror -O3 sqlite3.c
In function ‘sqlite3Strlen30’,
    inlined from ‘sqlite3ColumnSetColl’ at sqlite3.c:121296:10:
sqlite3.c:34653:28: error: ‘strlen’ reading 1 or more bytes from a region of size 0 [-Werror=stringop-overread]
34653 |   return 0x3fffffff & (int)strlen(z);
      |                            ^~~~~~~~~
In function ‘sqlite3ColumnSetColl’:
cc1: note: source object is likely at address zero
cc1: all warnings being treated as errors
```

PR #3758, opened 2024-06-01
Follow-up fixes for telemetry framework

timwoj:
- The `running_under_test` flag was not being set for the scripts.base.frameworks.logging.telemetry test. For most systems this was working fine, but broke on centos-stream 8. This could honestly be set at the top level of btest instead of setting it in individual tests, but I left it as-is for now.
- Fix a race condition in the services.json endpoint when calling `id::find_val`. Instead, assume that the list of nodes won't change during runtime, build the json at startup, and return that to the endpoint.
- Remove the prefix column from the telemetry log. It's redundant since it's included in the full name of the metric.
- Rework building the vector of label values when building the records for telemetry logging so that they're in the same order as the label names. This could be further improved by possibly just making prometheus-cpp return a std::map, but it works for right now to fix the existing breakage.


timwoj:
> Rework building the vector of label values when building the records for telemetry logging so that they're in the same order as the label names. This could be further improved by possibly just making prometheus-cpp return a std::map, but it works for right now to fix the existing breakage.

Thinking about this one a little more deeply, I think it would be better to move the label name/value fields together either up into the metric record or down into the opt record. It would mean I could just use the names and values directly from the prometheus-cpp object instead.

timwoj:
> Thinking about this one a little more deeply, I think it would be better to move the label name/value fields together either up into the metric record or down into the opt record. It would mean I could just use the names and values directly from the prometheus-cpp object instead.

I went ahead and implemented this in the last commit. It required some changes to the family registration methods in script-land, which need documentation in NEWS, but overall it feels more natural when querying the data.

src/telemetry/Manager.cc
```diff
@@ -259,6 +259,32 @@ static bool compare_histograms(const std::optional<ZVal>& a, const std::optional
     return comparer(a, b, metric_record_type);
 }
 
+static VectorValPtr build_label_values_vector(const std::vector<prometheus::ClientMetric::Label>& prom_labels,
+                                              const VectorValPtr& record_label_names) {
+    static auto string_vec_type = zeek::id::find_type<zeek::VectorType>("string_vec");
+    auto label_values_vec = make_intrusive<VectorVal>(string_vec_type);
+
+    // This feels really bad, since it's an O(m*n) search to bulld the vector,
```
ckreibich: If we change the label storage in the `Metric` record to a `table[string] of string` we'd avoid this scan, no? I'm worried that the separation of labels into two vectors, one in `MetricOpts` and one in `Metric`, is always going to be tricky for the user.

testing/btest/Baseline/scripts.base.frameworks.logging.telemetry/telemetry.log
```diff
@@ -10,7 +10,7 @@
 XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_stream_writes_total	module,stream	Conn,Conn::LOG	34.0
 XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_stream_writes_total	module,stream	DNS,DNS::LOG	34.0
 XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_stream_writes_total	module,stream	HTTP,HTTP::LOG	14.0
-XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_writer_writes_total	writer,module,stream,filter-name,path	default,Conn,conn,Conn::LOG,Log::WRITER_ASCII	30.0
-XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_writer_writes_total	writer,module,stream,filter-name,path	default,DNS,dns,DNS::LOG,Log::WRITER_ASCII	23.0
-XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_writer_writes_total	writer,module,stream,filter-name,path	default,HTTP,http,HTTP::LOG,Log::WRITER_ASCII	10.0
+XXXXXXXXXX.XXXXXX	zeek	counter	zeek_log_writer_writes_total	writer,module,stream,filter-name,path	Log::WRITER_ASCII,Conn,Conn::LOG,-,conn	30.0
```
ckreibich: Heh. So we did test it, we just missed it.

scripts/base/init-bare.zeek
```diff
@@ -5832,8 +5824,16 @@ export {
 		## A :zeek:see:`Telemetry::MetricOpts` record describing this metric.
 		opts: MetricOpts;
 
+		## The label names (also called dimensions) of the metric. When
+		## instantiating or working with concrete metrics, corresponding
+		## label values have to be provided. Examples of a label might
+		## be the protocol a general observation applies to, the
+		## directionality in a traffic flow, or protocol-specific
+		## context like a particular message type.
+		label_names: vector of string &default=vector();
```
ckreibich, line 5833: Yeah here's where you're losing me Tim — having both label names and label values in separate vectors doesn't look very practical. Wouldn't it be easier for this to be a `table[string] of string`?

scripts/base/init-bare.zeek
```diff
@@ -5802,14 +5802,6 @@ export {
 		## Documentation for this metric.
 		help_text: string &optional;
 
-		## The label names (also called dimensions) of the metric. When
-		## instantiating or working with concrete metrics, corresponding
-		## label values have to be provided. Examples of a label might
-		## be the protocol a general observation applies to, the
-		## directionality in a traffic flow, or protocol-specific
-		## context like a particular message type.
-		labels: vector of string &default=vector();
```
ckreibich, line 5811: Sorry Tim — what's the benefit of removing the labels here and then passing them into the various APIs separately? Can't we just leave them here?

I was just wondering whether a set would be better — then there's no implication of ordering, and it's easier to check membership.

src/telemetry/ProcessStats.cc
```diff
@@ -143,9 +156,7 @@ process_stats get_process_stats() {
         result.vms = vmsize_bytes;
         result.cpu = static_cast<double>(utime_ticks + stime_ticks) / ticks_per_second;
 
-        zeek::filesystem::path fd_path{"/proc/self/fd"};
-        result.fds =
```
ckreibich, line 147: I'd love to understand where this actually goes haywire in the `fd` directory — it's a pretty amazing failure mode! It's a pity because this looks so much nicer.

ckreibich (commented):
All great stuff Tim, I think it's only the handling of labels and their values where we're not on the same page yet. Let me know your thoughts there.

src/telemetry/Manager.cc
timwoj: No, because this is scanning through the `vector<pair<string,string>>` that prometheus-cpp returns from the call to `Collect()`. If they'd return a map instead, it would be cleaner.

testing/btest/Baseline/scripts.base.frameworks.logging.telemetry/telemetry.log
timwoj: Yeah... 😬

scripts/base/init-bare.zeek
timwoj: I could go either way. I was mostly just trying to remove the duplication of storing the names, but I agree having both results in cleaner function calls.

src/telemetry/ProcessStats.cc
timwoj: Same. The only thing I can think is that either we get into a mode where the `directory_iterator` is creating file descriptors as it loops, or the `opendir`/`readdir` version caches the current list of file descriptors before we start looping so it's always over a fixed set.

scripts/base/init-bare.zeek
timwoj: Assuming this field stays, would it still be filled in via the collect methods along with the new table in the other record types?

scripts/base/init-bare.zeek
timwoj: One problem with this is that tables can't have `&log` attributes, so we'll have to find another way to log them.

CMakeLists.txt
```diff
@@ -338,15 +338,20 @@ add_zeek_dynamic_plugin_build_interface_include_directories(
...
+    ${CMAKE_BINARY_DIR}/auxil/broker/libbroker
+    ${CMAKE_BINARY_DIR}/auxil/prometheus-cpp/core/include)
 
 # threading/formatters/JSON.h includes rapidjson headers and may be used
 # by external plugins, extend the include path.
 target_include_directories(zeek_dynamic_plugin_base SYSTEM
                            INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/rapidjson/include>)
+target_include_directories(
+    zeek_dynamic_plugin_base SYSTEM
+    INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/prometheus-cpp/include>)
```
awelzel, line 354: This might be opening a can of worms. We've previously attempted to remove rapidjson includes in public headers so that external plugins aren't exposed or aware of it. The rapidjson tweak above is meant to vanish with 7.1.

https://github.com/zeek/zeek/pull/3131

For OpenSSL, Dominik had done something similar
https://github.com/zeek/zeek/pull/3420

On first sight, this seems like a step towards exposing plugins to an external library again.

Also tangents the question if in the far future this dependency should from a distribution package ([fedora](https://packages.fedoraproject.org/pkgs/libprometheus-cpp/libprometheus-cpp-devel/fedora-40.html), [debian sid](https://packages.debian.org/sid/prometheus-cpp-dev)) rather than bundled.

awelzel: I realize hiding it likely results in a bunch of abstraction, so feel free to bounce.

Neverlord: Shouldn't forward declarations in our headers suffice?

bbannier: This looks like a fix for https://github.com/zeek/zeek/issues/3760 btw.

src/telemetry/Manager.cc
timwoj: That said, we've already forked prometheus-cpp to add the async interface, so modifying it to return a map instead shouldn't be too hard.

CMakeLists.txt
timwoj: > Shouldn't forward declarations in our headers suffice?

The `Family` template type causes some issues. I was able to remove one set of forward declarations out of the Manager in 1a66c203005384975c1b38698cf929c746c3aca9, but the includes remain in the instrument types.

awelzel: Hmm, oh well - I'm going to close.

I did see that moving the `telemetry/*` includes from `src/OpaqueVal.h` to `Opaques.h` could be useful. It should fix existing `src/OpaqueVal.h` users.

Though it won't change the situation much if someone wants to use telemetry from a plugin.

src/telemetry/Manager.cc
timwoj: This was superceded by the change that moves the labels to the other records, so I'm going to resolve this comment.

scripts/base/init-bare.zeek
ckreibich: Okay Tim, thanks. I was picturing an iteration over the table to populate the `Info` record. I'll fiddle with it some more and we can iterate again on just that aspect if really needed — better to get the other fixes in place now.

ckreibich (approved):

PR #3765, opened 2024-06-04
Add spicy based SSL analyzer as an optional component

0xxon:
This PR adds the spicy-based SSL analyzer to Zeek - and disables it by default.

It is hidden behind the `--enable-spicy-ssl` compile-time option.

The current analyzer is pretty feature complete - but not quite at the full level of the current binpac analyzer. Having feature parity is the final goal.

Pulling it into master would make it easier to keep in sync with Zeek developments - and it also makes it easier for it to be used, e.g., for performance tests.


src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+  DTLSv13   = 0xFEFC
+};
+
+const UNKNOWN_VERSION : uint16 = 0x0000;
+const SSLv2 = 0x0002;
+const SSLv3  = 0x0300;
+const TLSv10 = 0x0301;
+const TLSv11 = 0x0302;
+const TLSv12 = 0x0303;
+const TLSv13 = 0x0304;
```
bbannier, line 187: The constants here seem to duplicate some of the values in `TLSVersion` above, but e.g., `TLSv13` seems to be missing. Could you get rid of the constants and exclusively use the enum values instead?

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+  channel_id_new = 30032,
+  padding_old = 35655,
+  renegotiation_info = 65281
+};
+
+type ECPointFormat = enum {
+ uncompressed = 0,
+ ansiX962_compressed_prime = 1,
+ ansiX962_compressed_char2 =2
+};
```
bbannier, line 96: This is completely unused.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+  sect233r1 = 7, sect239k1 = 8, sect283k1 = 9,
+  sect283r1 = 10, sect409k1 = 11, sect409r1 = 12,
+  sect571k1 = 13, sect571r1 = 14, secp160k1 = 15,
+  secp160r1 = 16, secp160r2 = 17, secp192k1 = 18,
+  secp192r1 = 19, secp224k1 = 20, secp224r1 = 21,
+  secp256k1 = 22, secp256r1 = 23, secp384r1 = 24,
+  secp521r1 = 25,
+  arbitrary_explicit_prime_curves = 0xFF01,
+  arbitrary_explicit_char2_curves = 0xFF02
+};
```
bbannier, line 116: Completely unused.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+ peer_not_allowed_to_send = 2
+};
+
+type NameType = enum {
+  host_name = 0
+};
+
+type AlertLevel = enum {
+  warning= 1, fatal= 2
+};
```
bbannier, line 129: Used only in commented out code.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+  decode_error = 50,
+  decrypt_error = 51,
+  export_restriction_RESERVED = 60,
+  protocol_version = 70,
+  insufficient_security = 71,
+  internal_error = 80,
+  user_canceled = 90,
+  no_renegotiation = 100,
+  unsupported_extension = 110
+};
```
bbannier, line 157: Used only in commented out code.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+
+type HashAlgorithm = enum {
+  none = 0, md5= 1, sha1= 2, sha224= 3, sha256= 4, sha384= 5,
+  sha512 = 6
+};
```
bbannier, line 162: Used only in commented out code.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+
+type SignatureAlgorithm = enum {
+  anonymous = 0, rsa= 1, dsa= 2, ecdsa= 3
+};
```
bbannier, line 166: Used only in commented out code.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+const DTLSv10 = 0xFEFF;
+# DTLSv11 does not exist.
+const DTLSv12 = 0xFEFD;
+const DTLSv13 = 0xFEFC;
+
+type ClientCertificateType = enum {
+  rsa_sign = 1, dss_sign = 2, rsa_fixed_dh = 3, dss_fixed_dh = 4,
+  rsa_ephemeral_dh_RESERVED = 5, dss_ephemeral_dh_RESERVED = 6,
+  fortezza_dms_RESERVED = 20
+};
```
bbannier, line 198: Unused.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+import zeek;
+
+public function convert_server_names(snl: SSL::ServerNameList) : vector<bytes> {
+  local out: vector<bytes>;
+
+  for ( i in snl.server_name_list )
+    out.push_back(i.host_name);
+
+  return out;
+}
```
bbannier, line 1775: My gut reaction to seeing all these conversion functions returning lists is wondering whether we couldn't instead directly transform e.g., `ServerNameList` values to `vector<bytes>` on parsing with a `&convert`. That way we would perform this conversion exactly once.

Here and below.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+  version: uint16;
+
+  switch ( is_dtls_version(self.version) ) {
+    True -> dtlsfragment : DTLSRecordFragment(self.content_type, handshakesink, alertsink, msg, sh);
+    False -> tlsfragment : TLSRecordFragment(self.content_type, handshakesink, alertsink, msg, sh);
+  };
+
+  on content_type {
+    # print "Content type", self.content_type;
+  }
```
bbannier, line 694: This hook will not magically disappear but still be invoked (and the code for that will also be generated and compiled). If you want to keep the commented out code please comment out the full hook.

Here and elsewhere.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+    ContentType::application_data -> {
+      switch ( self.encrypted ) {
+        False -> appdata: bytes &size=self.length;
+        True -> cryptdata: bytes &size=self.length;
+      };
+    }
+    ContentType::change_cipher_spec -> ccs : bytes &size=self.length;
+    ContentType::heartbeat -> hn: Heartbeat(sh, self.length);
+    ContentType::alert -> : bytes &size=self.length -> alertsink;
+    * -> unhandled : bytes &size=self.length;
```
bbannier, line 734: You can use `skip` here to simply jump over `self.length` bytes,

```suggestion
    * -> unhandled: skip bytes &size=self.length;
```

There are few more instances where `skip` can be used.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,1901 @@
...
+## it is not necessary to send it and some implementations just don't.
+##
+## So - what we do here is that we enable the encrypted flag when we get
+## the first application data in a connection that negotiated TLS 1.3.
+##
+## This is correct insofar as the packet will be encrypted. We sadly loose
+## a bit of context here - we can't really say when we get the first packet
+## that uses the final cryptographic key material - and will contain content
+## data. We just don't have that information available in TLS 1.3 anymore.
+function determine_encryption_on(pr: PlaintextRecord, content_type: uint8, handshakesink: sink&, alertsink: sink&, inout msg: Message, inout sh: Share) : bool {
```
bbannier, line 786: The parameter `msg` is not used in the body.

bbannier (commented):
Since the parser is huge I only reviewed it on a superficial level. In particular I did not check whether we do work parsing and storing data which never makes it into a log (might be since this parser seems close to specs).

I fear that due to how this analyzer currently needs to be enabled, and if enabled does not pass tests seems to mean that we basically only make sure that the analyzer compiles (but I see no CI currently running with the analyzer enabled). To close out this work we need to track followup work to add a CI for it, and to bring it to feature parity with the binpac analyzer (if this is not done this analyzer would likely go stale, but I have seen users use experimental Spicy analyzers instead of tested binpac ones because "new"). Could you file such an issues?

Please run https://github.com/bbannier/spicy-lint which performs some very simple checks. It currently reports candidates for `skip`, empty hooks, and an unused parameter. I made fixes to the linter yesterday so every one of its reports should be relevant.

You might also want to run https://github.com/bbannier/spicy-format to clean up some formatting warts. I made a release yesterday and it should now support all constructs used here.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: Just for my understanding - is the conversion currently performed more than once?

My understanding is that currently this is performed when the event `ssl_extension_server_name` is raised (via the evt file). Which is raised once, and should perform this conversion once.

Same for the others - I don't think that the conversions are performed repeatedly.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: You are right that this is duplicate. Due to the way that these are used in the analyzer, it is easier to get rid of the enum, and instead just use the consts everywhere.

It is possible to use the enums instead - but this would necessitate casts to their int-value in essentially all locations. Which at the very least looks ugly/bulky.

src/analyzer/protocol/ssl/spicy/SSL.spicy
bbannier: Ah okay, I didn't fully trace the code. This was more a comment on style from reviewing internal code which called potentially expensive conversion functions repeatedly.

I'd say if something _always_ needs conversion (even if only once) my gut reaction would be to do that during parsing; `&convert` is exactly the tool to change from "on wire representation for parsing" to "user-consumable representation". One benefit is that should one suddenly expose something in more places the conversion already happened and one does not pay for it again; another benefit is that parsed datum and conversion are close together (e.g., here we could probably change `convert_server_names` to work on `ServerNameList::server_name_list` instead of the full unit and then `&convert` just that field (or even strip away `length` by adding a `&convert` to the unit.

If we sometimes need conversion and sometimes don't the situation is more tricky, but I'd probably still cache it in a unit var which I'd populate from a field hook as soon as the data becomes available.

This is basically trying to balance "accidentally converting repeatedly" and "computing a value which is not needed" (e.g., if there is no EVT hook mapping to expose a computed value).

0xxon:
This npw also supports SSLv2. Most baseline differences are caused by this being more exact than the binpac analyzer and will disappear once #3898 is merged.

The external test baselines have a difference that seems to be caused by https://zeekorg.slack.com/archives/C012CQV1K29/p1724336458281689

src/analyzer/protocol/ssl/CMakeLists.txt
```diff
@@ -1,34 +1,39 @@
...
+        ssl-protocol.pac
+        ssl-defs.pac
+        proc-certificate.pac
+        PAC
+        dtls.pac
+        ssl-dtls-analyzer.pac
+        dtls-analyzer.pac
+        ssl-dtls-protocol.pac
+        dtls-protocol.pac
+        ssl-defs.pac)
```
bbannier, line 35: Do you think it would make sense to at least test here that the Spicy analyzer continues building?, e.g., add here

```cmake
    if (NOT DISABLE_SPICY)
        # Even if we are using the binpac SSL analyzer make sure the Spicy
        # analyzer builds successfully. We use a debug build (`-d`) to perform
        # a faster build.
        add_custom_command(
            OUTPUT spicy-ssl.hlto
            COMMENT "Compiling Spicy SSL analyzer"
            COMMAND spicyz -d spicy/SSL.spicy spicy/SSL.evt spicy/support.cc -o
                    ${CMAKE_CURRENT_BINARY_DIR}/spicy-ssl.hlto
            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
            DEPENDS spicyz spicy/SSL.spicy spicy/SSL.evt spicy/support.cc)
        add_custom_target(spicy-ssl ALL DEPENDS spicy-ssl.hlto)
    endif ()
```

This could allow faster local iteration when e.g., bumping Spicy or modifying Zeek's Spicy integration without outright having to switch to the Spicy SSL analyzer.

src/analyzer/protocol/rdp/RDP.h
```diff
@@ -3,7 +3,6 @@
 #include "zeek/analyzer/protocol/pia/PIA.h"
```
bbannier, line 3: Since you are at it, this one is also unused.

```suggestion
```

src/analyzer/protocol/ssl/functions.bif
```diff
@@ -13,13 +17,16 @@
 ## Returns: T on success, F on failure.
 function set_ssl_established%(c: connection%): bool
 	%{
+#ifndef ENABLE_SPICY_SSL
+/* not implemented for spicy ssl */
```
bbannier, line 21:
```suggestion
/* not implemented for Spicy ssl */
```

src/analyzer/protocol/ssl/functions.bif
```diff
@@ -34,13 +41,16 @@ function set_ssl_established%(c: connection%): bool
 ## Returns: T on success, F on failure.
 function set_secret%(c: connection, secret: string%): bool
 	%{
+#ifndef ENABLE_SPICY_SSL
+/* not implemented for spicy ssl */
```
bbannier, line 45:
```suggestion
/* not implemented for Spicy ssl */
```

src/analyzer/protocol/ssl/functions.bif
```diff
@@ -55,13 +65,16 @@ function set_secret%(c: connection, secret: string%): bool
 ## Returns: T on success, F on failure.
 function set_keys%(c: connection, keys: string%): bool
 	%{
+#ifndef ENABLE_SPICY_SSL
+/* not implemented for spicy ssl */
```
bbannier, line 69:
```suggestion
/* not implemented for Spicy ssl */
```

testing/btest/core/analyzer-confirmation-violation-info.zeek
```diff
@@ -1,4 +1,5 @@
 # @TEST-DOC: The SSL analyzer picks up on the traffic in pppoe-over-qing, but then raises analyzer_violation_info
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 2: Assuming this is due to missing parity: could you add a comment/TODO explaining what is missing to not require this?

testing/btest/coverage/bare-load-baseline.test
```diff
@@ -9,6 +9,7 @@
 # below does. Don't ask. :-)
 
 # @TEST-REQUIRES: $SCRIPTS/have-spicy  # This test logs loaded scripts, so disable it if Spicy and it associated plugin is unavailable.
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 12: Can you add a comment why we don't run this with the Spicy SSL analyzer?

testing/btest/coverage/default-load-baseline.test
```diff
@@ -8,6 +8,7 @@
 # below does. Don't ask. :-)
 
 # @TEST-REQUIRES: ${SCRIPTS}/have-spicy
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 11: Ditto.

testing/btest/plugins/hooks.zeek
```diff
@@ -1,5 +1,6 @@
 # @TEST-REQUIRES: test "${ZEEK_ZAM}" != "1"
 # @TEST-REQUIRES: ${SCRIPTS}/have-spicy  # This test logs loaded scripts, so disable it if Spicy and the associated plugin are unavailable.
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 3: Ditto.

testing/btest/scripts/base/protocols/ssl/dtls-13.test
```diff
@@ -1,5 +1,6 @@
 # This tests a normal SSL connection and the log it outputs.
 
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 3: Ditto.

testing/btest/scripts/base/protocols/ssl/dtls-stun-dpd.test
```diff
@@ -1,3 +1,4 @@
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 1: Ditto.

testing/btest/scripts/base/protocols/ssl/dtls.test
```diff
@@ -1,5 +1,6 @@
 # This tests a normal SSL connection and the log it outputs.
 
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 3: Ditto.

testing/btest/scripts/policy/protocols/ssl/decryption.zeek
```diff
@@ -1,4 +1,5 @@
 # @TEST-REQUIRES: grep -q "#define OPENSSL_HAVE_KDF_H" $BUILD/zeek-config.h
+# @TEST-REQUIRES: ! grep -q "#define ENABLE_SPICY_SSL" $BUILD/zeek-config.h
```
bbannier, line 2: Ditto.

src/analyzer/protocol/ssl/spicy/SSL.evt
```diff
@@ -0,0 +1,87 @@
+protocol analyzer SSL over TCP:
```
bbannier, line 1: Do we want a copyright header here?

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
+module SSL;
```
bbannier, line 1: Ditto.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+    channel_id_new = 30032,
+    padding_old = 35655,
+    renegotiation_info = 65281,
+};
+
+# type ECPointFormat = enum {
+#  uncompressed = 0,
+#  ansiX962_compressed_prime = 1,
+#  ansiX962_compressed_char2 =2
+# };
```
bbannier, line 106: There's a couple instances of commented out code here, not sure whether you left them in deliberately.

src/analyzer/protocol/ssl/spicy/SSL.evt
```diff
@@ -0,0 +1,87 @@
...
+on SSL::Alert_message -> event ssl_alert($conn, SSL::get_direction(sh), self.level, self.description);
+
+on SSL::Heartbeat -> event ssl_heartbeat($conn, SSL::get_direction(sh), length, self.tpe, self.payload_length, self.data);
+
+on SSL::CertificateStatus -> event ssl_stapled_ocsp($conn, $is_orig, self.response);
+
+on SSL::CertificateRequest if ( SSL::uses_signature_and_hashalgorithm(sh) ) -> event ssl_certificate_request($conn, SSL::get_direction(sh), self.certificate_types, self.supported_signature_algorithms.supported_signature_algorithms_converted, self.certificate_authorities);
+on SSL::CertificateRequest if ( ! SSL::uses_signature_and_hashalgorithm(sh) ) -> event ssl_certificate_request($conn, SSL::get_direction(sh), self.certificate_types, SSL::create_empty_sigmature_algorithms(), self.certificate_authorities);
+
+on SSL::DirectionCheck::%done if ( self.was_flipped ) -> event ssl_connection_flipped($conn);
```
bbannier, line 87:
```suggestion
on SSL::DirectionCheck if ( self.was_flipped ) -> event ssl_connection_flipped($conn);
```

src/analyzer/protocol/ssl/spicy/support.cc
```diff
@@ -0,0 +1,63 @@
...
+    if ( ! c )
+        throw zeek::spicy::rt::ValueUnavailable("connection not available");
+
+    zeek::ODesc file_handle;
+    file_handle.AddRaw("Analyzer::ANALYZER_SSL");
+    file_handle.Add(c->analyzer->Conn()->StartTime());
+    file_handle.AddRaw(is_client ? "T" : "F", 1);
+    c->analyzer->Conn()->IDString(&file_handle);
+
+    file_handle.Add((uint32_t)pos);
```
bbannier, line 24:
```suggestion
    file_handle.Add(pos.Ref());
```

src/analyzer/protocol/ssl/spicy/support.cc
```diff
@@ -0,0 +1,63 @@
...
+#include <hilti/rt/libhilti.h>
+
+#include "zeek/Desc.h"
+#include "zeek/file_analysis/Manager.h"
+#include "zeek/spicy/cookie.h"
+#include "zeek/spicy/runtime-support.h"
+
+std::string ssl_get_certificate_fuid(const hilti::rt::Bool& is_client, const hilti::rt::integer::safe<uint32_t>& pos) {
+    auto cookie = static_cast<zeek::spicy::rt::Cookie*>(hilti::rt::context::cookie());
+    assert(cookie);
```
bbannier, line 12: Needs `#include <cassert>`.

src/analyzer/protocol/ssl/spicy/support.cc
```diff
@@ -0,0 +1,63 @@
...
+    file_handle.AddRaw("Analyzer::ANALYZER_SSL");
+    file_handle.Add(c->analyzer->Conn()->StartTime());
+    file_handle.AddRaw("F");
+    c->analyzer->Conn()->IDString(&file_handle);
+    file_handle.Add("ocsp");
+    std::string file_id = zeek::file_mgr->HashHandle(file_handle.Description());
+    return file_id;
+}
+
+bool ssl_is_partial_tcp() {
```
bbannier, line 47: Do you want to add a TODO here to potentially make this available for all users in the `zeek` Spicy module?

bbannier (approved):
Made a quick pass and flagged some minor issues. Feel free to ignore them at will.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+    var invalid_dtls_version_count: uint32;
+    #  var skipping: bool;
+    var client_encrypted: bool;
+    var server_encrypted: bool;
+    var both_sides_encrypted_first_time: bool;
+    var established: bool;
+    var client_certificate_depth: uint32;
+    var server_certificate_depth: uint32;
+
+    on %init {
```
awelzel, line 582: Total flyby. `%init` doesn't actually do anything when used as %context. Seems everything is initialized to default values anyhow?

https://github.com/zeek/spicy/issues/1502 (think this can be closed).

I'm also wondering if we should come up with a consistent name for structs used for `%context`, I've used `Ctx` for LDAP. In QUIC it's `ConnectionIDInfo` for historical reasons, I'd use Ctx today. Ctx or Context seems the least amount of cognitive overhead.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+type SupportedVersions = unit(sh: Share) {
+    length: uint8;
+    versions: uint16[self.length / 2];
+};
+
+# If the server sends it, this is the authoritative version. Set it.
+type OneSupportedVersion = unit(inout sh: Share) {
+    version: uint16;
+
+    on version {
```
awelzel, line 1207: Any reason not to fold that hook with the field?

```suggestion
   version: uint16 {
                set_version(self.version, sh);
    }
```

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+type ServerHelloChoice = unit(len: uint64, msg: Message, inout sh: Share) {
+    direction_check: DirectionCheck(sh, False); # should be sent by responder
+    sv: bitfield(16) {
+        server_version0: 0..7;
+        server_version1: 8..15;
+        server_version: 0..15;
+    };
+    var negotiated_version: uint16;
+
+    switch (self.negotiated_version) {
```
awelzel, line 1064: Is it true that `self.negotiated_version` here is dependent on the `on sv` execution at the bottom? I'm seeing this pattern elsewhere and wonder if this is required for some reason? If not, this appears  obfuscated as reading top to bottom doesn't really work.

Wouldn't it be easier to follow if the hook was directly following the `sv` bitfield?

```
bitfield(16) {
...
} {
    ...
    if (self.sv.server_version1 == 0x7F) {
        self.negotiated_version = 0x7f00;
    ...
}

switch (self.negotiated_version) {
}
```

The pattern of having `on field { }` at the end of the unit is used pretty often, so wonder generally if there's a reason rather than following them directly with the parsed field.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+        Extensions::application_layer_protocol_negotiation -> application_layer_protocol_negotiation: ProtocolNameList(sh) &max-size=self.length if(self.length > 0);
+        Extensions::status_request -> status_request: StatusRequest(self.length) &max-size=self.length if(self.length > 0);
+        Extensions::signed_certificate_timestamp -> signed_certificate_timestamp: SignedCertificateTimestampList(sh) &max-size=self.length if(self.length > 0);
+        Extensions::key_share_old -> key_share_old: KeyShare(client_hello, sh, self.length) &max-size=self.length if(self.length > 0);
+        Extensions::key_share -> key_share: KeyShare(client_hello, sh, self.length) &max-size=self.length if(self.length > 0);
+        Extensions::supported_versions -> supported_versions: SupportedVersionsSelector(sh, client_hello) &max-size=self.length if(self.length > 0);
+        Extensions::psk_key_exchange_modes -> psk_key_exchange_modes: PSKKeyExchangeModes(sh) &max-size=self.length if(self.length > 0);
+        Extensions::pre_shared_key -> pre_shared_key: PreSharedKey(sh, client_hello) &max-size=self.length if(self.length > 0);
+        * -> unknown: bytes &size=self.length;
+    };
```
awelzel, line 1142: Are those `&max-size` and `&if` all same? Think they could work on the outer switch, too? Saw this somewhere else, too.

Also wonder if you could go with `&parse-from=self.raw` instead of the `set_input()`, so that wouldn't require random access. Not sure that makes a huge difference though.

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -0,0 +1,2058 @@
...
+        }
+        ContentType::change_cipher_spec -> ccs: bytes &size=self.length;
+        ContentType::heartbeat -> hn: Heartbeat(sh, self.length);
+        ContentType::alert -> : bytes &size=self.length -> alertsink;
+        * -> unhandled: skip bytes &size=self.length;
+    };
+    trigger_two: void;
+
+    on length {
+        self.encrypted = determine_encryption_on(self, content_type, handshakesink, alertsink, sh);
```
awelzel, line 805: This `self.encrypted` is used on 792 above? Similar as for other comments, any reason not do add it directly as:
```
var encrypted: bool;
length: uint16  {
    self.encrypted = determine_encryption_on(self, content_type, handshakesink, alertsink, sh);
}
```

awelzel (commented):
Just skimming through.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: Yup, I did. It's for referencing/potential future extensions. They are work to search for - and there does not seem to be any reason to not just keep them in there commented out.

testing/btest/coverage/bare-load-baseline.test
0xxon: Will do. I just was too lazy to add comments for the more-obvious-testcases, that just change on every major change in Zeek.

src/analyzer/protocol/ssl/spicy/SSL.spicy
bbannier: I think I agree with @awelzel's sentiment here, having "smallish" hooks at the end makes the flow harder to follow for no good reason. Being able to read a parser top-down is useful and makes side effects easier to understand.

In this case the hook directly influences parsing of following fields and IMO hiding that away in a hook "elsewhere" makes this less transparent; since the code is small enough I'd tend to just declare it inline. (potentially different story if we e.g., fill up some unit var only used in an EVT hook). This is also relevant when thinking about error recovery where being able to quickly see which side effects parsing a field has makes it easier to reason about the state of fields we see; IMO an out-of-line field hook has a tendency to be mistaken for a unit `%done` hook.

Overall I think when possible it is worth taking advantage of Spicy's ability to mix declarative (fields) and procedural code.

.cirrus.yml
```diff
@@ -228,6 +229,17 @@ debian12_binary_task:
   env:
     ZEEK_CI_CONFIGURE_FLAGS: *BINARY_CONFIG
 
+debian12_spicy_ssl_task:
+  container:
```
awelzel, line 233: Further down is a `ubuntu22_spicy_task`. It would be sufficient to add `--enable-spicy-ssl` there I believe? The nice side-effect is that the task also submits to zeek-benchmarker, allowing some automatic recording of performance differences between the BinPAC and Spicy versions.

src/analyzer/protocol/ssl/CMakeLists.txt
0xxon: That's a great idea, I am going to add this.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: Thank you - I totally was not aware of this.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: Sounds nice, but doesn't work:

```
[error] /Users/johanna/zeek/main/src/analyzer/protocol/ssl/spicy/SSL.spicy:1109:5-1127:28: attribute '&max-size' is not supported here
```

.cirrus.yml
0xxon: That sounds good :)

PR #3770, opened 2024-06-06
Add Telemetry::metrics_address option

### LINKED ISSUE #3769: Feature Request:  Add a bind_address option for telemetry
It would be handy to have a bind_address option for the Telemetry framework to specify which address the prometheus endpoints(s) listen on.

timwoj:
Fixes #3769

### DIFF
--- a/scripts/base/frameworks/telemetry/options.zeek
+++ b/scripts/base/frameworks/telemetry/options.zeek
@@ -6,6 +6,10 @@ module Telemetry;
 # to see the options without needing the rest.
 
 export {
+	## Address used to make metric data available to Prometheus scrapers via
+	## HTTP.
+	const metrics_address = getenv("ZEEK_DEFAULT_LISTEN_ADDRESS") &redef;
+
 	## Port used to make metric data available to Prometheus scrapers via
 	## HTTP.
 	const metrics_port = 0/unknown &redef;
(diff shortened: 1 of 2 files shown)


scripts/base/frameworks/telemetry/options.zeek
```diff
@@ -6,6 +6,10 @@ module Telemetry;
 # to see the options without needing the rest.
 
 export {
+	## Address used to make metric data available to Prometheus scrapers via
+	## HTTP.
+	const metrics_address: addr = 127.0.0.1 &redef;
```
ckreibich, line 11: Tim I wonder whether we should switch this over to using the `ZEEK_DEFAULT_LISTEN_ADDRESS` environment variable, like [for Broker](https://github.com/zeek/zeek/blob/master/scripts/base/frameworks/broker/main.zeek#L27).

The benefit would be that anything that already works by adjusting that variable will transparently apply here, and it seems pretty likely to me that these will broadly match (if you run a cluster for which you want Broker to listen globally, you'll want that for telemetry too, and vice versa). As-is, I imagine a lot of cluster users will want to adjust it somewhere to listen globally.

timwoj: Ah yeah, that would be a good choice for the default. We were defaulting to localhost in the existing code, so I opted for that.

timwoj:
I know that Broker just uses a string to store the result of `getenv` and relies on broker to tell whether the address is valid. We can do the same with prometheus-cpp's error logging. I modified our error message to include the address it tried to bind to though, so now we get something like this:

```
% ./src/zeek -i en0 -e 'redef Telemetry::metrics_address="abcde";redef Telemetry::metrics_port=9991/tcp'
fatal error in <command line>, line 3: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports. Attempted to bind to abcde:9991.
```

ckreibich (approved):
Right on!

PR #3772, opened 2024-06-06
Check for 'zeekctl check' before trying to start up prometheus

### LINKED ISSUE #3771: zeekctl check fails with frameworks/telemetry/prometheus loaded: fatal error: Failed to setup Prometheus endpoint
When loading `load frameworks/telemetry/prometheus` via `local.zeek` in a zeekctl setup, running `zeekctl check` or `zeekctl deploy` while the cluster is already up triggers the following errors. Workaround is `zeekctl stop` before `check` or `deploy`.

```
root@ubuntu-01:/opt/zeek-dev# ./bin/zeekctl check
logger-1 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


manager scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


proxy-1 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


worker-1-1 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


worker-1-2 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


worker-1-3 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports


worker-1-4 scripts failed.
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports
```

There's an env called `ZEEKCTL_CHECK_CONFIG` or `ZEEKCTL_DISABLE_LISTEN` which could possibly be leveraged to avoid binding the port in that scenario.

https://github.com/zeek/zeek/blob/master/scripts/base/frameworks/cluster/setup-connections.zeek#L60-L63

timwoj:
`zeekctl check` runs all of the `InitPostScript` methods before exiting. This means that if the cluster is already running and prometheus is already initialized, it'll try to start up the underlying webserver a second time and fail due to the ports being in use. Check for one of the environment variables set during `check` before trying to initialize it.

Fixes #3771

### DIFF
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -70,16 +70,18 @@ void Manager::InitPostScript() {
             }
         }
 
-        try {
-            prometheus_exposer = std::make_unique<prometheus::Exposer>(prometheus_url, 2, callbacks);
+        if ( ! getenv("ZEEKCTL_CHECK_CONFIG") ) {
+            try {
+                prometheus_exposer = std::make_unique<prometheus::Exposer>(prometheus_url, 2, callbacks);
 
-            // CivetWeb stores a copy of the callbacks, so we're safe to delete the pointer here
-            delete callbacks;
-        } catch ( const CivetException& exc ) {
-            reporter->FatalError("Failed to setup Prometheus endpoint: %s\n", exc.what());
+                // CivetWeb stores a copy of the callbacks, so we're safe to delete the pointer here
+                delete callbacks;
+            } catch ( const CivetException& exc ) {
+                reporter->FatalError("Failed to setup Prometheus endpoint: %s\n", exc.what());
+            }
+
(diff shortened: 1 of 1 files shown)


ckreibich (approved):
LGTM, I think it's the right variable to pick here. I just looked through the check logic a bit and I think the whole concept is something we could move over into Zeek more fully some time.

PR #3773, opened 2024-06-07
Suppress a known data race during civetweb shutdown

timwoj:
This was reported in https://github.com/civetweb/civetweb/issues/861, but has apparently cropped up again after it was fixed once. Since it only happens during shutdown, we're ok to suppress it for now.


awelzel (approved):
From skimming the issue, we could try defining `STOP_FLAG_NEEDS_LOCK` in the TSAN build, but not sure that's worth it (and would anyhow be different from the default build).

PR #3777, opened 2024-06-11
Fix coverage builds

awelzel:
This does two things:
* Compile coverage builds with with `-fprofile-update=atomic` to avoid negative counters
* Unset CCACHE_BASEDIR from the asan_sanitizer task to avoid cmake rewriting everything as relative paths and confusing our gcov setup (we only ever compile in `/zeek` in CI, so CCACHE_BASEDIR usage isn't really useful generally).

Goes with https://github.com/zeek/cmake/pull/112

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -397,6 +397,8 @@ asan_sanitizer_task:
     CXXFLAGS: -DZEEK_DICT_DEBUG
     ZEEK_CI_CONFIGURE_FLAGS: *ASAN_SANITIZER_CONFIG
     ASAN_OPTIONS: detect_leaks=1:detect_odr_violation=0
+    # Use absolute paths for coverage files.
+    CCACHE_BASEDIR:
 
 ubsan_sanitizer_task:
   container:


awelzel:
I've tested within an interactive Cirrus CI using `coveralls-lcov --dry-run -v` and that seemed all sensible :crossed_fingers:

.cirrus.yml
```diff
@@ -397,6 +397,8 @@ asan_sanitizer_task:
     CXXFLAGS: -DZEEK_DICT_DEBUG
     ZEEK_CI_CONFIGURE_FLAGS: *ASAN_SANITIZER_CONFIG
     ASAN_OPTIONS: detect_leaks=1:detect_odr_violation=0
+    # Use absolute paths for coverage files.
```
bbannier, line 400: Could you add a little more context here, e.g.,

```suggestion
    # Use absolute paths for coverage files since gcov gets confused by relative paths.
```

cmake
bbannier, line 1: Just flagging: Support `-fprofile-update` has only [landed with ccache-4.10](https://github.com/ccache/ccache/pull/1408), is that the version we are using? Even if we use earlier versions I'd expect ccache simply refusing to cache which might be an okay fallback for us.

bbannier (approved):

cmake
awelzel: Oh, thanks for checking. Ubuntu 24.04 (fairly new) only has ccache 4.9.1, so this will disable caching :-1: 

I'm conflicted. asan_sanitizer task currently has `--disable-spicy`, but once that goes, probably need to revisit or upgrade ccache.

PR #3781, opened 2024-06-12
SSH half-duplex robustness revert/tweak

awelzel:
As mentioned at https://github.com/zeek/zeek/pull/3720/files#r1636314375, this reverts #3720, this reverts the original fix and tweaks `get_version()` to use whatever we have.

@J-Gras , can you see how this fares with the pcap you have? Is there a lot of SSH-1.99 included? :-/


awelzel:
@vpax / @J-Gras  - I now realize the `update_kex()` method, too, relies packets from both sides to determine how key exchange is parsed. Think teaching `update_kex()` or `get_state()` to some "unknown" and ignore each packet until seeing `MSG_NEWKEYS` could alleviate this, but I'm unsure this is all worth it :-(

Somewhat related: https://github.com/zeek/spicy/issues/1328

You could also experiment with setting `DPD::max_violations[ANALYZER_SSH] = 1 (or maybe 0)` - that should disable the analyzer after the first violation.

J-Gras:
@awelzel, the larger test PCAP successfully identifies 216,684 connections as SSH. The `ssh.log` has 260,171 entries, with `SSH-1.99` popping up 2,876 times. With this patch, the size of the `analyzer.log` is reduced from 203,613 to 113,514. The failure reason is now mostly a variant of `binpac exception: out_of_bound: ssh_string:val` instead of `binpac exception: invalid index for case: SSH_Key_Exchange`.

Given the above results, I'd agree that this path doesn't seem worth it. Right now the logs feel slightly harder to interpret.

When I experimented with setting `DPD::max_violations` to one, I encountered situations where the same analyzer got attached twice. This was because the analyzer was first attached based on the port, threw a violation and then got detached. As this happen early in the connection, the analyzer got attached again by a DPD signature hitting a couple of packets later just to fail again.

Long story short: I believe introducing means to reduce the output to `analyzer.log` would make most sense here. Talking to @vpax, it would also be nice if we could detach the SSH analyzer as early as possible, i.e. once we know that further analysis won't succeed.

awelzel:
> Long story short: I believe introducing means to reduce the output to analyzer.log would make most sense here. Talking to @vpax, it would also be nice if we could detach the SSH analyzer as early as possible, i.e. once we know that further analysis won't succeed.

Yep, agree - for 7.0 I've opened #3792 for a plain revert. There's a pattern in the SSL analyzer to call `AnalyzerViolation()` followed by `SetSkip()` - maybe this could be leveraged also in SSH to have the violation raised, but then also ignore any further input, though technically that's not the same as actually detaching the analyzer (but wouldn't trigger the DPD issue you ran into).

PR #3785, opened 2024-06-14
Spicy: Disallow repeating replacements of the same analyzer.

### LINKED ISSUE #3783: Analyzer disabled upon startup
I compiled Zeek v7 in debug mode and got the below output in `debug.log`
```
0.000000/1718178868.242260 [spicy] spicy::FOO_TCP replaces existing protocol analyzer FOO
0.000000/1718178868.242265 [dpd] Disabling analyzer FOO
0.000000/1718178868.242273 [spicy] spicy::FOO_UDP replaces existing protocol analyzer FOO
0.000000/1718178868.242278 [dpd] Disabling analyzer spicy_FOO_TCP
```
The relevant snippet in the `evt` file is shown below
```
protocol analyzer spicy::FOO_TCP over TCP:
    parse with foo::TCP_PDUs,
    replaces FOO;

protocol analyzer spicy::FOO_UDP over UDP:
    parse with foo::UDP_PDUs,
    replaces FOO;
```

rsmmr:
We now reject EVT files that attempt to replace the same built-in
analyzer multiple times as doing so would be ill-defined and not very
intuitive in what exactly it means.

Closes #3783.


src/spicy/manager.cc
```diff
@@ -897,14 +897,21 @@ void Manager::disableReplacedAnalyzers() {
...
             SPICY_DEBUG(hilti::rt::fmt("%s is supposed to replace protocol analyzer %s, but that does not exist",
                                        info.name_analyzer, replaces));
 
             continue;
         }
 
+        auto tag = component->Tag();
+        if ( analyzer_mgr->HasComponentMapping(tag) )
+            reporter->FatalError(
+                "%s: protocol analyzer %s is already mapped to a different analyzer; cannot replace an analyzer "
```
bbannier, line 911: Nit: It's a bit unfortunate that we currently cannot tell the user which previous analyzer also made a replacement since there is no API for us to query that. If we had e.g., 

```cpp
// Returns component mapped over this one if there is a mapping defined, else `nullptr`.
`Component* GetComponentMapping(const zeek::Tag& tag) const`
```

we could do that (and replace all uses of `HasComponentMapping` as well).

bbannier (approved):

src/spicy/manager.cc
rsmmr: Yeah, I thought about that but then figured it's fine, this should be rare enough in the first place

PR #3788, opened 2024-06-18
Spicy: Provide runtime API to access Zeek-side globals.

rsmmr:
This allows to read Zeek global variables from inside Spicy code. The
main challenge here is supporting all of Zeek's data type in a
type-safe manner.

The most straight-forward API is a set of functions
`get_<type>(<id>)`, where `<type>` is the Zeek-side type
name (e.g., `count`, `string`, `bool`) and `<id>` is the fully scoped
name of the Zeek-side global (e.g., `MyModule::Boolean`). These
functions then return the corresponding Zeek value, converted in an
appropriate Spicy type. Example:

    Zeek:
        module Foo;

        const x: count = 42;
        const y: string = "xxx";

    Spicy:
        import zeek;

        assert get_count("Foo::x") == 42;
        assert get_string("Foo::y") == b"xxx"; # returns bytes(!)

For container types, the `get_*` function returns an opaque types that
can be used to access the containers' values. An additional set of
functions `as_<type>` allows converting opaque values of atomic
types to Spicy equivalents. Example:

    Zeek:
        module Foo;

        const s: set[count] = { 1, 2 };
        const t: table[count] of string = { [1] = "One", [2] = "Two" }

    Spicy:

        # Check set membership.
        local set_ = zeek::get_set("Foo::s");
        assert zeek::set_contains(set_, 1) == True

        # Look up table element.
        local table_ = zeek::get_table("Foo::t");
        local value = zeek::table_lookup(t, 1);
        assert zeek::as_string(value) == b"One"

There are also functions for accessing elements of Zeek-side vectors
and records.

If any of these `zeek::*` conversion functions fails (e.g., due to a
global of that name not existing), it will throw an exception.

Design considerations:

    - We support only reading Zeek variables, not writing. This is
      both to simplify the API, and also conceptually to avoid
      offering backdoors into Zeek state that could end up with a very
      tight coupling of Spicy and Zeek code.

    - We accept that a single access might be relatively slow due to
      name lookup and data conversion. This is primarily meant for
      configuration-style data, not for transferring lots of dynamic
      state over.

    - In that spirit, we don't support deep-copying complex data types
      from Zeek over to Spicy. This is (1) to avoid performance
      problems when accidentally copying large containers over,
      potentially even at every access; and (2) to avoid the two sides
      getting out of sync if one ends up modifying a container without
      the other being able to see it.


rsmmr:
@bbannier Documentation is still missing but take a look at the API and see if that looks like a good approach.

src/spicy/runtime-support.h
```diff
@@ -59,15 +61,22 @@ class Unsupported : public UsageError {
...
+class ParameterMismatch : public TypeMismatch {
 public:
-    TypeMismatch(const std::string_view& msg, std::string_view location = "")
-        : UsageError(hilti::rt::fmt("Event parameter mismatch, %s", msg)) {}
-    TypeMismatch(const std::string_view& have, const TypePtr& want, std::string_view location = "")
-        : TypeMismatch(_fmt(have, want)) {}
+    ParameterMismatch(const std::string_view& msg, std::string_view location = "")
+        : TypeMismatch(hilti::rt::fmt("Event parameter mismatch, %s", msg)) {}
+    ParameterMismatch(const std::string_view& have, const TypePtr& want, std::string_view location = "")
+        : ParameterMismatch(_fmt(have, want)) {}
```
bbannier, line 79: Nit: Lets fix this to pass `string_view` by value here.

```suggestion
    ParameterMismatch(std::string_view msg, std::string_view location = "")
        : TypeMismatch(hilti::rt::fmt("Event parameter mismatch, %s", msg)) {}
    ParameterMismatch(std::string_view have, const TypePtr& want, std::string_view location = "")
        : ParameterMismatch(_fmt(have, want)) {}
```

scripts/spicy/zeek.spicy
```diff
@@ -147,54 +147,115 @@ public function protocol_handle_close(handle: ProtocolHandle): void &cxxname="ze
...
+public function get_port(id: string): port &cxxname="zeek::spicy::rt::get_port";
+public function get_record(id: string): ZeekRecord &cxxname="zeek::spicy::rt::get_record";
+public function get_set(id: string): ZeekSet &cxxname="zeek::spicy::rt::get_set";
+public function get_string(id: string): bytes &cxxname="zeek::spicy::rt::get_string";
+# public function get_subnet(id: string): network &cxxname="zeek::spicy::rt::get_subnet";
+public function get_table(id: string): ZeekTable &cxxname="zeek::spicy::rt::get_table";
+public function get_time(id: string): time &cxxname="zeek::spicy::rt::get_time";
+public function get_vector(id: string): ZeekVector &cxxname="zeek::spicy::rt::get_vector";
+
+public function get_value(id: string): ZeekVal &cxxname="zeek::spicy::rt::get_value";
```
bbannier, line 226: Maybe more general, we could return an `optional<ZeekVal>` here so users have a way to work around missing values instead of facing an exception.

scripts/spicy/zeek.spicy
```diff
@@ -147,54 +147,115 @@ public function protocol_handle_close(handle: ProtocolHandle): void &cxxname="ze
...
+
+public function as_address(v: ZeekVal): addr &cxxname="zeek::spicy::rt::as_address";
```
bbannier, line 228: Similarly, these could return our a `result<T>` now that we have that available (with the current exception message as message).

bbannier (commented):
I like this, especially the way it is currently limited (no writing from Spicy to Zeek; pretty explicit API without too much magic).

Currently this raises exceptions at runtime if Zeek script and Spicy parser code are out of sync which is hard to handle. I wonder whether wrapping things in `optional` and `result` wouldn't be easier to work with.

rsmmr:
> Currently this raises exceptions at runtime if Zeek script and Spicy parser code are out of sync which is hard to handle. I > > wonder whether wrapping things in `optional` and `result` wouldn't be easier to work with.

Funny, this is actually what I had in the first version I wrote: I was using `result` (being happy to have a use case :-). However, I didn't end up liking it, for two reasons:

- Main reason: I couldn't see how it helps. It makes usage more clunky (need to check and dereference, potentially handle error) but without much benefit: what's the user going to do in the case that an error comes back? Most likely that's indeed a mismatch between Zeek and Spicy, without much opportunity to recover, so the Spicy code would probably just report an error in some other way then and probably need to abort. Maybe one could substitute some default value if an error comes back to be able to proceed, but is that really helpful in the case of a clear mistake somewhere that should rather be fixed? Plus, will Spicy users actually add all that error handling logic for what I'm pretty sure they'd consider unlikely to happen (rather than just always directly dereference and get the exception that way)? So I ended up thinking that triggering errors behind the scenes through exceptions makes usage more convenient without loosing much.

- Secondary reason: we're doing exceptions already for the value conversion from Spicy to Zeek, so doing exceptions for the other way round seems more consistent, too.

rsmmr:
I'm going to use `result` for table lookups: that seems a valid case of might-not-exist.

bbannier:
> I'm going to use `result` for table lookups: that seems a valid case of might-not-exist.

To be fair, IMO record access or even access to globals could also fail, even if Zeek and Spicy code are in sync when running in bare mode or the Zeek scripts are not loaded for other reasons.

rsmmr:
> To be fair, IMO record access or even access to globals could also fail, even if Zeek and Spicy code are in sync when running in bare mode or the Zeek scripts are not loaded for other reasons. 

Right, it sure can happen, but what would one *do* if that comes back with an error?

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+/** Retrieves the value of Zeek record field. Throws on error. */
+inline ::zeek::ValPtr record_field(const std::string& name, const std::string& index) {
+    return record_field(get_record(name), index);
+}
+
+/** Check if a Zeek record has a field's value set. Throws on errors. */
+inline hilti::rt::Bool record_has_field(const zeek::spicy::rt::ValRecordPtr& v, const std::string& field) {
+    auto index = v->GetType()->AsRecordType()->FieldOffset(field.c_str());
+    if ( index < 0 )
+        throw InvalidValue(util::fmt("no such field in record type: %s", field.c_str()));
```
bbannier, line 1208: It would seem to fit the function's purpose and really simple for us to return `false` if a record does not have a certain field, e.g., in the test

```ruby
assert ! zeek::record_has_field("Bar::record_", "does_not_exist");
```

This would allow users to check whether a field is present in a record or not which might change dynamically depending on which scripts are loaded.

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+::hilti::rt::Bool table_contains(const std::string& name, const T& key) {
+    return table_contains(get_table(name), key);
+}
+
+/**
+ * Retrieves a value from a Zeek table. Returns an error value if the key does
+ * not exist. Throws on other errors.
+ */
+template<typename T>
+::hilti::rt::Result<::zeek::ValPtr> table_lookup(const zeek::spicy::rt::ValTablePtr& v, const T& key) {
```
bbannier, line 1249: Not sure we need the error message here and couldn't just go with `std::optional<ValPtr>`. It should be clear that a key is not present if no value is returned. Similar for the overload below, and on the script side.

bbannier (approved):
This looks like a great addition to me.

I left a couple smaller comments.

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+ * a ``TypeMismatch`` exception on mismatch.
+ */
+inline auto check_type(const ValPtr& v, ::zeek::TypeTag type_tag, const char* expected) {
+    if ( v->GetType()->Tag() != type_tag )
+        type_mismatch(v, expected);
+}
+
+} // namespace detail
+
+/** Strong typedef for a Zeek record value. */
```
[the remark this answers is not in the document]
rsmmr: Maybe we mean something different with "strong typedef", I'll just change to "Type for ..."

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+inline auto type_mismatch(const ValPtr& v, const char* expected) {
+    throw TypeMismatch(util::fmt("type mismatch in Zeek value: expected %s, but got %s", expected,
+                                 ::zeek::type_name(v->GetType()->Tag())));
+}
+
+/**
+ * Helper to check the type of Zeek value against an expected type tag, raising
+ * a ``TypeMismatch`` exception on mismatch.
+ */
+inline auto check_type(const ValPtr& v, ::zeek::TypeTag type_tag, const char* expected) {
```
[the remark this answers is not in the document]
rsmmr: Yeah, it's tricky to fit this into a single `as_type()`. But it's strictly internal, so seems fine.

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+inline ::hilti::rt::Interval as_interval(const ValPtr& v) {
+    detail::check_type(v, TYPE_INTERVAL, "interval");
+    return ::hilti::rt::Interval(v->AsInterval(), hilti::rt::Interval::SecondTag{});
+}
+
+/** Converts a Zeek `port` value to its Spicy equivalent. Throws on error. */
+inline ::hilti::rt::Port as_port(const ValPtr& v) {
+    detail::check_type(v, TYPE_PORT, "port");
+    auto p = v->AsPortVal();
+    return ::hilti::rt::Port(p->Port(), p->PortType());
```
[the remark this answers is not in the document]
rsmmr: Will do. I'm surprised now that that even compiles without a cast ...

src/spicy/runtime-support.h
rsmmr: Ack, agree.

src/spicy/runtime-support.h
```diff
@@ -984,4 +994,293 @@ inline ValPtr to_val(const T& t, const TypePtr& target) {
...
+    detail::check_type(v, TYPE_DOUBLE, "double");
+    return v->AsDouble();
+}
+
+/** Converts a Zeek `enum` value to its Spicy equivalent. Throws on error. */
+inline std::string as_enum(const ValPtr& v) {
+    detail::check_type(v, TYPE_ENUM, "enum");
+    // Zeek returns the name as "<module>::<enum>", we just want the enum name.
+    return hilti::rt::rsplit1(v->GetType()->AsEnumType()->Lookup(v->AsEnum()), "::").second;
+}
```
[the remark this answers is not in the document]
bbannier: Thanks for the details, I agree that a `string` as implemented is probably the best choice with what we have available.

> I don't see "Spicy equivalent" for any of the ``{get,as}_enum` functions?

Right above we have `/** Converts a Zeek `enum` value to its Spicy equivalent. Throws on error. */`. It is the generic repeated docstring, only in C++ code, so not a big issue.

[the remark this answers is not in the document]
rsmmr: (Ah, I was looking at the `zeek.spicy` docstring.)

PR #3790, opened 2024-06-18
Topic/awelzel/3682 bad pipe op

awelzel:


src/threading/MsgThread.cc
```diff
@@ -165,6 +165,52 @@ bool ReporterMessage::Process() {
...
+class MsgThread_IOSource : public iosource::IOSource {
+public:
+    explicit MsgThread_IOSource(MsgThread* thread) : thread(thread) {
+        if ( ! iosource_mgr->RegisterFd(flare.FD(), this) )
+            reporter->FatalError("Failed to register MsgThread fd with iosource_mgr");
+
+        SetClosed(false);
+
+        // Register as lifetime managed IO source.
+        iosource_mgr->Register(this, true);
```
awelzel, line 178: This isn't great as it'll increase the `sources` list by however many log writers are active (they are all MsgThreads). This could easily triple (or worse) `sources.size()`. The `sources` vector is iterated over in `FindReadySources()` twice, so it's fairly hot.

awelzel:
Replacing this with #3791.

PR #3791, opened 2024-06-18
threading/MsgThread: Decouple IO source and thread

### LINKED ISSUE #3682: Crash 6.2.0 : Flare.cc: zeek::detail::bad_pipe_op
I have been seeing intermittent crashes in 6.2.0 with the following coredump:

For reference: 
```
(gdb) bt
#0  0x0000000803888bca in thr_kill () from /lib/libc.so.7
#1  0x0000000803886f94 in raise () from /lib/libc.so.7
#2  0x00000008037ff489 in abort () from /lib/libc.so.7
#3  0x0000000000cec2d1 in zeek::Reporter::FatalErrorWithCore (this=<optimized out>, fmt=<optimized out>) at /usr/local/src/zeek-6.0.1/src/Reporter.cc:154
#4  0x0000000000c8966f in zeek::detail::bad_pipe_op (which=0x4d1b9d "read", signal_safe=<optimized out>) at /usr/local/src/zeek-6.0.1/src/Flare.cc:76
#5  0x0000000000c896d5 in zeek::detail::Flare::Extinguish (this=0x8a4558ca4, signal_safe=false) at /usr/local/src/zeek-6.0.1/src/Flare.cc:153
#6  0x0000000000d9a04b in zeek::threading::MsgThread::Process (this=0x8a4558800) at /usr/local/src/zeek-6.0.1/src/threading/MsgThread.cc:517
#7  0x0000000000d16ed6 in zeek::run_state::detail::run_loop () at /usr/local/src/zeek-6.0.1/src/RunState.cc:349
#8  0x00000000007506c8 in main (argc=<optimized out>, argv=<optimized out>) at /usr/local/src/zeek-6.0.1/src/main.cc:102
```

I'll try to see how I can replicate it - so far different physical clusters have showing this crash - albeit very intermittent - I don't see a predictive pattern yet!

awelzel:
A MsgThread acting as an IO source itself can result in the scenario where the threading manager's heartbeat timer deletes a terminated MsgThread instance, but at the same time this instance is in the list of ready IO sources as determined by the IO loop in the current iteration.

Fix this by decoupling the IO source from the thread and clean them up separately. The threading manager now also tracks MsgThread IO sources and deletes them only when they've been closed and therefore unregistered from the IO loop. Unregistration only happens during Process() which should have the necessary guarantees.

Fixes #3682

### DIFF
--- a/src/threading/Manager.cc
+++ b/src/threading/Manager.cc
@@ -58,8 +58,12 @@ void Manager::Terminate() {
         delete *i;
     }
 
+    for ( auto* iosource : io_sources )
+        delete iosource;
+
     all_threads.clear();
     msg_threads.clear();
+    io_sources.clear();
     terminating = false;
 }
 
@@ -71,9 +75,10 @@ void Manager::AddThread(BasicThread* thread) {
         StartHeartbeatTimer();
 }
 
-void Manager::AddMsgThread(MsgThread* thread) {
+void Manager::AddMsgThread(MsgThread* thread, iosource::IOSource* iosource) {
     DBG_LOG(DBG_THREADING, "%s is a MsgThread ...", thread->Name());
     msg_threads.push_back(thread);
+    io_sources.push_back(iosource);
 }
 
 void Manager::KillThreads() {
@@ -115,6 +120,16 @@ void Manager::SendHeartbeats() {
 
         t->Join();
         delete t;
+    }
+
+    for ( auto it = io_sources.begin(); it != io_sources.end(); /**/ ) {
+        auto* src = *it;
+        if ( ! src->IsOpen() ) {
+            delete src;
+            it = io_sources.erase(it);
+        }
+        else
+            ++it;
     }
 }
 
--- a/src/threading/Manager.h
+++ b/src/threading/Manager.h
@@ -121,8 +121,9 @@ protected:
      * MsgThread constructor makes sure to do so.
      *
      * @param thread The thread.
+     * @param iosource The IO source of the thread.
      */
-    void AddMsgThread(MsgThread* thread);
+    void AddMsgThread(MsgThread* thread, iosource::IOSource* iosource);
 
     void Flush();
 
@@ -142,6 +143,9 @@ private:
 
     using msg_thread_list = std::list<MsgThread*>;
     msg_thread_list msg_threads;
+
+    using io_source_list = std::list<iosource::IOSource*>;
+    io_source_list io_sources;
 
     bool did_process; // True if the last Process() found some work to do.
     double next_beat; // Timestamp when the next heartbeat will be sent.
--- a/src/threading/MsgThread.cc
+++ b/src/threading/MsgThread.cc
@@ -165,6 +165,72 @@ bool ReporterMessage::Process() {
     return true;
 }
 
+
+// This is the IO source used by MsgThread.
+//
+// The lifetime of the IO source is decoupled from
+// the thread. The thread may be terminated prior
+// to the IO source being properly unregistered and
+// forgotten by the IO manager. Specifically the
+// threading manager would delete an IO source which
+// the IO manager still believed to be ready.
+//
+// See issue #3682 for more details.
+class MsgThread_IOSource : public iosource::IOSource {
+public:
+    explicit MsgThread_IOSource(MsgThread* thread) : thread(thread) {
+        if ( ! iosource_mgr->RegisterFd(flare.FD(), this) )
+            reporter->FatalError("Failed to register MsgThread fd with iosource_mgr");
+
+        SetClosed(false);
+    }
+
+    ~MsgThread_IOSource() {
+        if ( IsOpen() ) {
+            if ( thread )
+                reporter->Warning("Have thread %s set in MsgThread_IOSource", thread->Name());
+
+            if ( ! iosource_mgr->UnregisterFd(flare.FD(), this) )
+                reporter->FatalError("Failed to unregister MsgThread fd from iosource_mgr");
+        }
+    }
+
+    void Process() override {
+        flare.Extinguish();
+
+        if ( thread )
+            thread->Process();
+        else {
+            // When there's no thread anymore, unregister
+            // this source from the IO manager and mark
+            // it as closed. The threading manager will then
+            // reap it during heartbeat processing or shutdown.
+            if ( ! iosource_mgr->UnregisterFd(flare.FD(), this) )
+                reporter->FatalError("Failed to unregister MsgThread fd from iosource_mgr");
+
+            SetClosed(true);
+        }
+    }
+
+    const char* Tag() override { return thread ? thread->Name() : "<MsgThread_IOSource orphan>"; }
+
+    double GetNextTimeout() override { return -1; }
+
+    void Fire() { flare.Fire(); };
+
+    // Fire the flare one more time so that
+    // the IO manager will call Process() and
+    // SetClosed(true).
+    void Close() {
+        thread = nullptr;
+        flare.Fire();
+    }
+
+private:
+    MsgThread* thread = nullptr;
+    zeek::detail::Flare flare;
+};
+
 } // namespace detail
 
 ////// Methods.
@@ -179,18 +245,20 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
     child_finished = false;
     child_sent_finish = false;
     failed = false;
-    thread_mgr->AddMsgThread(this);
 
-    if ( ! iosource_mgr->RegisterFd(flare.FD(), this) )
-        reporter->FatalError("Failed to register MsgThread fd with iosource_mgr");
-
-    SetClosed(false);
+    io_source = new detail::MsgThread_IOSource(this);
+    thread_mgr->AddMsgThread(this, io_source);
 }
 
 MsgThread::~MsgThread() {
-    // Unregister this thread from the iosource manager so it doesn't wake
-    // up the main poll anymore.
-    iosource_mgr->UnregisterFd(flare.FD(), this);
+    // Unregister this thread from the IO source so we don't
+    // get Process() callbacks anymore. The IO source is
+    // freed by separately by the threading manager after its
+    // last Process() invocation.
+    if ( io_source ) {
+        io_source->Close();
+        io_source = nullptr;
+    }
 }
 
 void MsgThread::OnSignalStop() {
@@ -253,7 +321,13 @@ void MsgThread::OnWaitForStop() {
 }
 
 void MsgThread::OnKill() {
-    SetClosed(true);
+    // Ensure the IO source is closed and won't call Process() on this
+    // thread anymore. The thread got killed, so the threading manager will
+    // remove it forcefully soon.
+    if ( io_source ) {
+        io_source->Close();
+        io_source = nullptr;
+    }
 
     // Send a message to unblock the reader if its currently waiting for
     // input. This is just an optimization to make it terminate more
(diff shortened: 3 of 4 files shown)


awelzel:
The spurious "unexpected pipe read failure: Bad file descriptor" this should fix is really the good case as it'll cause a crash. But this is use-after-free, so there's all kind of other things that could go wrong.


@initconf / @leres - would you be able to deploy this branch in your environment and see if it fixes the aborts?

initconf:
> The spurious "unexpected pipe read failure: Bad file descriptor" this should fix is really the good case as it'll cause a crash. But this is use-after-free, so there's all kind of other things that could go wrong.
> 
> @initconf / @leres - would you be able to deploy this branch in your environment and see if it fixes the aborts?

Ack ! Let me get this branch running - Its interesting that most of this crashes I've noticed were on INT-DMZ cluster specifically - must have something to do with internal traffic ! I'll test it out for stability etc !

leres:
> @initconf / @leres - would you be able to deploy this branch in your environment and see if it fixes the aborts?

Oops, the diff I get from  ee5c455 is quite far from applying to 6.0.4.
```
zinc 54 % patch -p1 < /tmp/p
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/threading/Manager.cc b/src/threading/Manager.cc
|index 9bcdc925f4..09be785025 100644
|--- a/src/threading/Manager.cc
|+++ b/src/threading/Manager.cc
--------------------------
Patching file src/threading/Manager.cc using Plan A...
Hunk #1 failed at 58.
Hunk #2 failed at 75.
Hunk #3 failed at 121.
3 out of 3 hunks failed--saving rejects to src/threading/Manager.cc.rej
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/threading/Manager.h b/src/threading/Manager.h
|index 875e35290a..0930182aa0 100644
|--- a/src/threading/Manager.h
|+++ b/src/threading/Manager.h
--------------------------
Patching file src/threading/Manager.h using Plan A...
Hunk #1 failed at 121.
Hunk #2 failed at 144.
2 out of 2 hunks failed--saving rejects to src/threading/Manager.h.rej
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/threading/MsgThread.cc b/src/threading/MsgThread.cc
|index a1d2e5e3da..dafa67d174 100644
|--- a/src/threading/MsgThread.cc
|+++ b/src/threading/MsgThread.cc
--------------------------
Patching file src/threading/MsgThread.cc using Plan A...
Hunk #1 failed at 165.
Hunk #2 failed at 245.
Hunk #3 failed at 321.
Hunk #4 failed at 419.
Hunk #5 failed at 493.
5 out of 5 hunks failed--saving rejects to src/threading/MsgThread.cc.rej
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/threading/MsgThread.h b/src/threading/MsgThread.h
|index 55b8f0ba1d..45b05b955b 100644
|--- a/src/threading/MsgThread.h
|+++ b/src/threading/MsgThread.h
--------------------------
Patching file src/threading/MsgThread.h using Plan A...
Hunk #1 failed at 27.
Hunk #2 failed at 42.
Hunk #3 failed at 210.
Hunk #4 succeeded at 287 with fuzz 2 (offset 5 lines).
Hunk #5 failed at 368.
4 out of 5 hunks failed--saving rejects to src/threading/MsgThread.h.rej
done
```

awelzel:
> Oops, the diff I get from https://github.com/zeek/zeek/commit/ee5c4555e3e390f9a488403a417c6452a4783419 is quite far from applying to 6.0.4.

Yep, the Zeek tree was re-formatted shortly after 6.0, so that's expected. Do you have a 6.2 setup around?

leres:
> Yep, the Zeek tree was re-formatted shortly after 6.0, so that's expected. Do you have a 6.2 setup around?

Not really. I'm the maintainer for the FreeBSD security/zeek port and I've been keeping it on the 6.0.X train. I run this version (via the port) on two boxes.

Are there any plans to back-port this to 6.0?

awelzel:
> Are there any plans to back-port this to 6.0?

It's a candidate, yes. I can backport it to 6.0 and point you at a branch if that helps.

awelzel:
> That would be awesome, thank you!

I hope this does it - do you see this crash often in your environment?

https://github.com/zeek/zeek/commit/8c337bd7693a2002dcfe8a15b35dc92eb9e78de9.patch

leres:
> > That would be awesome, thank you!
> 
> I hope this does it - do you see this crash often in your environment?

I had 5 crashes in May, 2 so far in June. This was for a system that does dns and http. The other that only does dns didn't seem to crash.

Ok, the FreeBSD 6.0.4_1 port now has that fix (plus the unicode validation fix).

src/threading/Manager.h
```diff
@@ -143,6 +144,9 @@ class Manager {
     using msg_thread_list = std::list<MsgThread*>;
     msg_thread_list msg_threads;
 
+    using io_source_list = std::list<iosource::IOSource*>;
+    io_source_list io_sources;
```
rsmmr, line 148: Better to avoid new raw pointers / manual memory mgmt. Seems the manager is owning these, so suggest to use `std::list<std::unique_ptr<iosource::IOSource>>`. That will then take care of the other clean-up pieces as well, like in `terminate()`. 

(I realize we do manual memory mgt for the threads, too, but I believe that's just a left-over from the past and we should move those over to smart pointers as well.

src/threading/MsgThread.cc
```diff
@@ -179,18 +245,20 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
...
     child_sent_finish = false;
     failed = false;
-    thread_mgr->AddMsgThread(this);
-
-    if ( ! iosource_mgr->RegisterFd(flare.FD(), this) )
-        reporter->FatalError("Failed to register MsgThread fd with iosource_mgr");
 
-    SetClosed(false);
+    io_source = new detail::MsgThread_IOSource(this);
+    thread_mgr->AddMsgThread(this, io_source);
```
rsmmr, line 250: If `MsgThread_IOSource` becomes a `unique_ptr` inside the manager (see other comment), either use `make_unique()` plus `std::move()` here, or move the instantiation to the manager, too, by replacing `AddMsgThread` with, say, `MsgThread_IOSource* CreateMsgThread(this)` and putting the `make_unique` inside `CreateMsgThread` and they just have it return a non-owning raw pointer.

src/threading/Manager.cc
```diff
@@ -116,6 +121,16 @@ void Manager::SendHeartbeats() {
         t->Join();
         delete t;
     }
+
+    for ( auto it = io_sources.begin(); it != io_sources.end(); /**/ ) {
+        auto* src = *it;
+        if ( ! src->IsOpen() ) {
```
rsmmr, line 127: Just polishing:

```suggestion
        if ( auto* src = *it; ! src->IsOpen() ) {
```

src/threading/MsgThread.cc
```diff
@@ -165,6 +165,72 @@ bool ReporterMessage::Process() {
...
+//
+// The lifetime of the IO source is decoupled from
+// the thread. The thread may be terminated prior
+// to the IO source being properly unregistered and
+// forgotten by the IO manager. Specifically the
+// threading manager would delete an IO source which
+// the IO manager still believed to be ready.
+//
+// See issue #3682 for more details.
+class MsgThread_IOSource : public iosource::IOSource {
```
rsmmr, line 179: The name feels a bit clunky to me. Could move the class into `MsgThread` so that it becomes `MsgThread::IOSource`, or into namespace, e.g., `threading::IOSource`?

src/threading/MsgThread.cc
```diff
@@ -165,6 +165,72 @@ bool ReporterMessage::Process() {
...
+public:
+    explicit MsgThread_IOSource(MsgThread* thread) : thread(thread) {
+        if ( ! iosource_mgr->RegisterFd(flare.FD(), this) )
+            reporter->FatalError("Failed to register MsgThread fd with iosource_mgr");
```
rsmmr, line 183: `FatalError` a user-level error message. I'd either rephrase it to make it at least somewhat better understandable by users, or, if it's something that really shouldn't happen, make it an `InternalError` instead. More of these elsewhere.

src/threading/MsgThread.cc
```diff
@@ -165,6 +165,72 @@ bool ReporterMessage::Process() {
...
+
+        SetClosed(false);
+    }
+
+    ~MsgThread_IOSource() {
+        if ( IsOpen() ) {
+            if ( thread )
+                reporter->Warning("Have thread %s set in MsgThread_IOSource", thread->Name());
```
rsmmr, line 191: Similar to above; here, and elsewhere: `InternalWarning`?

rsmmr (commented):
The separation makes sense to me, I like that.  I've left some comments on memory mint, but a higher-level point: I'm wondering if we should leave the memory mgmt for the new IOSources to the `iosource_mgr`. The code isn't calling `iosource_mgr->Register()`. If it did, the manager could take care of the instance's life-time, avoiding the kind of problem that started this all. Plus, that `Register()` takes care of some additional initialization, like calling `InitSource()` (which isn't overridden, so no change, but could be surprising in the future if that doesn't get called). In other words, now that we have "clean" IOSource that's no longer depending on the thread, also treat it like a standard IOSource.

awelzel:
> I'm wondering if we should leave the memory mgmt for the new IOSources to the iosource_mgr. The code isn't calling iosource_mgr->Register(). If it did, the manager could take care of the instance's life-time, avoiding the kind of problem that started this all.

Yep - agree conceptually. I had attempted this in a prior [PR here](https://github.com/zeek/zeek/pull/3790/files), but was worried that the life-time management overhead could end-up high (`FindReadySources()` is very hot, so the once a second cleanup seemed nice), Though that might be unfounded in practice.

Using `Register()` for short-lived IO sources today actually leaks memory and isn't quite working, so that would be good to fix anyhow.

Let me resurrect the other PR and fold in your feedback where appropriate.

awelzel:
> Yep - agree conceptually. I had attempted this in a prior [PR here](https://github.com/zeek/zeek/pull/3790/files), but was worried that the life-time management overhead could end-up high (FindReadySources() is very hot, so the once a second cleanup seemed nice), Though that might be unfounded in practice.

I've tested this a bit and the extra overhead might be visible in `zeek -r` scenarios. Still seems the better approach as it shouldn't matter overly much in actual cluster scenarios, so thanks for the nudge. Possibly in the future we could differentiate between 'fd-only' and 'timeout-only' IO sources. This could help avoid over all the sources in [FindReadySources()](https://github.com/zeek/zeek/blob/43804fa3b5ccedb181a6da3c1f199155de772766/src/iosource/Manager.cc#L136-L139).

initconf:
Yes, seek-test has 6.2.x setup in place. Sorry I haven’t gotten to running the branch yet.

rsmmr:
> I've tested this a bit and the extra overhead might be visible in `zeek -r` scenarios. Still seems the better approach as it shouldn't matter overly much in actual cluster scenarios, so thanks for the nudge. Possibly in the future we could differentiate between 'fd-only' and 'timeout-only' IO sources. This could help avoid over all the sources in [FindReadySources()]

Yes, agree that this is something the manager should handle/optimize internally. An IOSource could either tell it explicitly what kind it is, or maybe the manager could have some heuristic internally to split the sources into two groups based on behavior (with the former clearly being easier / less risky).

awelzel:
Replaced by #3791.

@initconf - if you haven't yet deployed, could you try the other PR? @leres, might have been a bit too early to pull into FreeBSD ports - the final patch will likely be much different.

PR #3795, opened 2024-06-21
Add `join_any_vec` function

ynadji:
If I have a vector of elements that are not `string`s, I'd like to join them into a `string` without having to build a new `vector of string`s to use `join_string_vec`. This PR adds a bif function `join_any_vec` that does this automatically. Optionally, a `stringifier` function can be provided that allows more fine-grained control over how the elements are turned into strings, which defaults to `cat`.

The `join_string_vec` implementation already does this, but the function declaration prevents other vector types from being passed as an argument. As such, there's a lot of duplication between the two functions. I'll happily take suggestions to address this!

I wanted to add a similar `join_any_set` function, but `set[any]` is invalid. If there's a clever way to do this I'd happily do that as well.


awelzel:
> I'd like to join them into a string without having to build a new vector of strings to use join_string_vec. 

Just out of curiosity, do you use `join_any_vec()` in a context where it would be prohibitive to have the intermediary vector, or is this more for, say, debug output?

I'm not overly opposed, just wondering what you're doing :-)

> As such, there's a lot of duplication between the two functions. I'll happily take suggestions to address this!

You could introduce a `static` helper in `strings.bif` that's used by both functions and pass it the stringifier (as Func::nil or nullptr when it's `cat`), it returns the StringVal.

```
  auto e = v->ValAt(...);
  if ( stringifier )
    e = stringifier->Invoke(v)

  e->Describe().
```

There's other static helpers (search for "should_strip()").

awelzel:
@ynadji - for the time being I'm closing this. There's always an option to add such a BiF via a plugin. If you find it's a very popular (internal or external), that would certainly be a good data point for including it.

PR #3796, opened 2024-06-25
Remove internal `Supervisor::ClusterEndpoint` struct

ckreibich:
This implements the brief discussion we had [here](https://zeekorg.slack.com/archives/CTGDG83B8/p1711501157059819). Removing that internal struct removes one place in which we need to keep changes to `Cluster::Node` consistent. This would be nice to get in prior to my upcoming change to add the metrics port, since it'll mean a smaller changeset.

I tried this in a couple of ways. One alternative was to keep the internal representation as a `RecordValPtr` to the script-layer `ClusterEndpoint` instance throughout (instead of keeping a JSON string), but that does not work because the stem doesn't have the corresponding script-layer context at the time it receives the serialized `NodeConfig`. I rather like that the meat of the mapping logic now lives in the script layer.

### DIFF
--- a/scripts/base/frameworks/cluster/supervisor.zeek
+++ b/scripts/base/frameworks/cluster/supervisor.zeek
@@ -1,0 +1,57 @@
+##! Cluster-related functionality specific to running under the Supervisor
+##! framework.
+
+@load base/frameworks/supervisor/api
+
+module Cluster::Supervisor;
+
+export {
+	## Populates the current node's :zeek:id:`Cluster::nodes` table from the
+	## supervisor's node configuration in :zeek:id:`Supervisor::NodeConfig`.
+	##
+	## Returns: true if initialization completed, false otherwise.
+	global __init_cluster_nodes: function(): bool;
+}
+
+function __init_cluster_nodes(): bool
+	{
+	local config = Supervisor::node();
+
+	if ( |config$cluster| == 0 )
+		return F;
+
+	local rolemap: table[Supervisor::ClusterRole] of Cluster::NodeType = {
+		[Supervisor::LOGGER] = Cluster::LOGGER,
+		[Supervisor::MANAGER] = Cluster::MANAGER,
+		[Supervisor::PROXY] = Cluster::PROXY,
+		[Supervisor::WORKER] = Cluster::WORKER,
+	};
+
+	local manager_name = "";
+	local cnode: Cluster::Node;
+	local typ: Cluster::NodeType = Cluster::NONE;
+
+	for ( node_name, endp in config$cluster )
+		{
+		if ( endp$role == Supervisor::MANAGER )
+			manager_name = node_name;
+		}
+
+	for ( node_name, endp in config$cluster )
+		{
+		if ( endp$role in rolemap )
+			typ = rolemap[endp$role];
+
+		cnode = [$node_type=typ, $ip=endp$host, $p=endp$p];
+@pragma push ignore-deprecations
+		if ( endp?$interface )
+			cnode$interface = endp$interface;
+@pragma pop ignore-deprecations
+		if ( |manager_name| > 0 && cnode$node_type != Cluster::MANAGER )
+			cnode$manager = manager_name;
+
+		Cluster::nodes[node_name] = cnode;
+		}
+
+	return T;
+	}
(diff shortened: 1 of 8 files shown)


src/supervisor/Supervisor.cc
```diff
@@ -1252,24 +1254,8 @@ Supervisor::NodeConfig Supervisor::NodeConfig::FromRecord(const RecordVal* node)
...
-        if ( iface )
-            ep.interface = iface->AsStringVal()->ToStdString();
-
-        const auto& pcap_file = rv->GetField("pcap_file");
-
-        if ( pcap_file )
-            ep.pcap_file = pcap_file->AsStringVal()->ToStdString();
-
-        rval.cluster.emplace(name, std::move(ep));
+        auto re = std::make_unique<RE_Matcher>("^_");
```
ckreibich, line 1257: I'm unsure whether this is needed. I kept it because it matches the existing use down in line 1321, but it's not clear to me why it's there either.

src/Val.cc
```diff
@@ -330,7 +330,10 @@ TableValPtr Val::GetRecordFields() {
 }
 
 // This is a static method in this file to avoid including rapidjson's headers in Val.h because
-// they're huge.
+// they're huge. For records, two arguments control the JSON result: only_loggable, if true,
+// skips any fields that don't have the &log attribute. The regular expression matcher, if
+// given, is used to strip the first match on any record field name in the resulting output.
+// See the to_json() BiF for context.
```
awelzel, line 336: Moving that comment to `Val::ToJSON()` in `Val.h` would be visible in the headers we ship, too, and potentially more useful.

src/supervisor/Supervisor.h
```diff
@@ -233,15 +204,16 @@ class Supervisor : public iosource::IOSource {
...
         /**
-         * The Cluster Layout definition.  Each node in the Cluster Framework
+         * The Cluster layout definition.  Each node in the Cluster Framework
          * knows about the full, static cluster topology to which it belongs.
-         * Entries in the map use node names for keys.
+         * Entries in the map use node names for keys and the JSON representations
+         * of script-layer Supervisor::ClusterEndpoint records as values.
          */
-        std::map<std::string, ClusterEndpoint> cluster;
+        std::map<std::string, std::string> cluster;
```
awelzel, line 216: Only thought is whether we could just keep the whole JSON blob in a `std::string` - is there any advantage to having the `std::map` ?

awelzel (approved):
I'm approving. Left some nits and wonder if `NodeConfig.cluster` could be a std::string storing a blob rather than the `std::map` that's a bit structured but which don't necessarily need IIUC.

src/supervisor/Supervisor.h
ckreibich: Thanks Arne. There was no reason other than focusing the change on `ClusterEndpoint` instances, but I think you're right — with that change the map itself isn't required. I'll make an edit pass, also for your other suggestions.

ckreibich:
Arne I've worked these all in and stacked it onto #3804 since the whole thing was getting too big. I'll straighten out the branches during the merge.

Switching the JSON representation to the whole table meant that we'd need to be able to ... read tables, which is functionality we could use anyway. I think this makes it a nice overall changeset. But holler if you still see anything.

scripts/base/frameworks/cluster/supervisor.zeek
```diff
@@ -0,0 +1,57 @@
...
+@load base/frameworks/supervisor/api
+
+module Cluster::Supervisor;
+
+export {
+	## Populates the current node's :zeek:id:`Cluster::nodes` table from the
+	## supervisor's node configuration in :zeek:id:`Supervisor::NodeConfig`.
+	##
+	## Returns: true if initialization completed, false otherwise.
+	global __init_nodes: function(): bool;
```
awelzel, line 13: Nit.

```suggestion
	global __init_cluster_nodes: function(): bool;
```

awelzel (approved):

awelzel:
> Switching the JSON representation to the whole table meant that we'd need to be able to ... read tables

Eh.. thanks for going down the rabbit hole.

scripts/base/frameworks/cluster/supervisor.zeek
ckreibich: Done.

PR #3797, opened 2024-06-25

bbannier:


awelzel (approved):
Is that prepping for 7.0? Anything major that should be called out in Zeek's NEWS file regarding this bump?

I'm thinking the accessors for Zeek variables might be a good candidate, though that was all on the Zeek side anyhow?

bbannier:
> Is that prepping for 7.0? Anything major that should be called out in Zeek's NEWS file regarding this bump?
> 
> I'm thinking the accessors for Zeek variables might be a good candidate, though that was all on the Zeek side anyhow?

Yes, this is prep for a Zeek RC, we will cut a Spicy release near then end of the RC period. I updated the Spicy-side NEWS file with the interesting bits there, the Zeek-side plugin changes should go into the Zeek NEWS file. This was usually done by the Zeek release manager, at the minimum we should say "bump to spicy-x.y.z".

PR #3804, opened 2024-07-01
JSON serialization / ingestion improvements

ckreibich:
While iterating on #3796 I decided to split this out into a separate PR.


ckreibich:
The edits to the tests here are actually quite small, just testing that the roundtrip through `to_json()`/`from_json()` works. But I've split out the single large test functions into individual ones, for the various index types.

src/Val.cc
```diff
@@ -1023,34 +1072,69 @@ static std::variant<ValPtr, std::string> BuildVal(const rapidjson::Value& j, con
...
+                    rapidjson::Document idxstr;
+                    idxstr.Parse(it->name.GetString(), it->name.GetStringLength());
+
+                    std::variant<ValPtr, std::string> idx;
+
+                    if ( tl->GetTypes().size() > 1 )
+                        idx = BuildVal(idxstr, tl, key_func);
+                    else if ( IsQuotedJSONType(tl->GetPureType()) )
+                        // Parse this with the quotes the string came with. This
+                        // mirrors the quote-stripping in BuildJSON().
```
awelzel, line 1115: Not sure this comment fits - there's not quotes in `it->name` IIUC for addr or enum. Maybe `IsQuotedJSONType()` could be `HasQuotesStripped()` or so.

awelzel (approved):

src/Val.cc
ckreibich: I've renamed the function to `UsesJSONStringType()`, perhaps that's better? I was trying to capture the list of Zeek types that use a JSON string for rendering.

PR #3807, opened 2024-07-04
CI: Use OSX Sonoma only, fix Threads::Threads imported target

awelzel:


awelzel:
@Neverlord , the OSX builds with CMake version 3.30 fail with the following message currently:

```
-- Configuring done (22.9s)
CMake Error at CMakeLists.txt:317 (target_link_libraries):
  The link interface of target "zeek_dynamic_plugin_base" contains:

    Threads::Threads

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

Does the following seem right to you? I'm not quite sure why this starts failing now - some of the bundled projects (broker, prometheus) use `find.*Threads` in CMake, so we just got lucky?

```
index 13ea5a5fe..6a2f3dcff 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,7 @@ target_include_directories(
     zeek_dynamic_plugin_base
     INTERFACE $<INSTALL_INTERFACE:include> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
               $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>)
+find_package(Threads REQUIRED)
 target_link_libraries(zeek_dynamic_plugin_base INTERFACE Threads::Threads)
 add_library(Zeek::DynamicPluginBase ALIAS zeek_dynamic_plugin_base)
 set_target_properties(zeek_dynamic_plugin_base PROPERTIES EXPORT_NAME DynamicPluginBase)
```

awelzel:
>  the OSX builds with CMake version 3.30 fail with the following message currently:

Installed a custom CMake 3.30 and the same error happens on Linux with that version, too.

Issue #3595, opened 2024-01-31
mkdir() BiF generates errors on failure rather than warnings


vpax:
I have a `zeek_init` that tries to create a directory to use to stash some data later. In some configurations, this fails, which is okay and there's logic in the handler to deal with it. However, because `mkdir()` has:
```
zeek::emit_builtin_error(zeek::util::fmt("cannot create directory '%s': %s", filename,
strerror(error)));
```
the failure gets marked as a run-time error and the script aborts with an initialization error. Seems it should either not generate a message (putting the burden of doing so on the caller, who should check the return status), or if it does, that it should be a warning rather than a harder error.

awelzel:
> or if it does, that it should be a warning rather than a harder error.
> I'd leave that all to the return handling. I've marked it a bug. Thanks for flagging.

We could also go the route of adding `err_on_fail: bool` parameter that similarly exists for `disable_analyzer()`. Mostly coming from the corner that error checking might not be very-well-established given many BiFs do produce diagnostic output. Not an overly strong opinion.

PR #3810, opened 2024-07-04
mkdir() BiF optionally generates error on failure

cknill:
Addresses #3595.  

I added an extra argument to the BiF mkdir() called err_on_fail and it defaults to false.  It could default to true, but I reason that if folks want all the noise of diagnostics, they can ask for it.  That may be wrongheaded of me, so, if desired I can swap out the default value easily.  

While I was in the zeek.bif, I noticed that the BiF functions: rmdir, unlink, and rename behave like mkdir.  It may be worthwhile having those functions do as mkdir does.  I could add that on to this PR or have it in a separate one; however you all like to organize things.

### DIFF
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -4393,13 +4393,15 @@ function flush_all%(%): bool
 ##
 ## f: The directory name.
 ##
+## err_on_fail: Emits an error message when true, the default value is false.
+##
 ## Returns: True if the operation succeeds or if *f* already exists,
 ##          and false if the file creation fails.
 ##
 ## .. zeek:see:: active_file open_for_append close write_file
 ##              get_file_name set_buf flush_all enable_raw_output
 ##              rmdir unlink rename
-function mkdir%(f: string%): bool
+function mkdir%(f: string, err_on_fail: bool &default=F%): bool
 	%{
 	const char* filename = f->CheckString();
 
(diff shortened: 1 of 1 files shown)


awelzel:
> I added an extra argument to the BiF mkdir() called err_on_fail and it defaults to false. It could default to true, but I reason that if folks want all the noise of diagnostics, they can ask for it. 

I'm forth and back that if users have to opt-in to error reporting, they are already aware they care and could do proper error handling.  But.. they won't have access to `errno` / `strerror`, so maybe there's still value. My own hunch would be to default to `true`, but @ckreibich may have more/stronger opinions.

cknill:
> My own hunch would be to default to `true`

I went down memory lane over the weekend where I made similar decisions and I remembered how the floodgates of angry emails opened on me and as a result I think defaulting to true is probably the right thing to do. 

Furthermore, I am beginning to question the wisdom of adding the extra function argument to mkdir. I typed it a bunch of times over the weekend and it really gets annoying. Apologies if what I type next is a faux pas, I haven't been in the Zeek community long enough to know better; I wonder if it may be better to add a global variable instead of an extra function parameter and then a user redefines it to turn off emitting error messages on failure of directory operations. 

> But.. they won't have access to `errno` / `strerror`, so maybe there's still value. 

I don't think it would be too hard to add in errno and strerror, unless this has already been quashed for reasons a quick Github issues search didn't reveal.  If there's no reason not to add those, I would be happy to take that on.

ckreibich:
I'm in favor of turning the `emit_builtin_error()` cases in this family of BiFs into reporter warnings. @cknill, if you're still around (sigh, I am sorry for the silence here!) would you be up for updating the PR to do that?

I don't think these functions are sufficiently special to expose the error reporting via an additional argument. There are other BiFs that conceivably fit that line of thought, and it's hard to see us adding such an argument wherever that scenario comes up.

If we were to provide a `strerror()` equivalent this should likely be at the reporter level, not the system's — i.e., this should return to you the most recent error message that went through the reporter instance.

ckreibich:
Cool, hi! :-) Yes to both. You can just use `zeek::reporter->Warning()`, like elsewhere in the BiFs. Thanks!

ckreibich:
Last call, @cknill. :-)

ckreibich:
Closing in favor of #4222.

PR #3812, opened 2024-07-08
Telemetry support for the management framework 

ckreibich:
This introduces the `metrics_port` support (per #3701) for the management framework. Similar to our additions to zeekctl, this includes auto-enumeration of the metrics ports when desired, and enabled by default. It also pulls in updates for zeek-client and to zeek-cluster-testing for corresponding test cases.

It has two minor tweaks to the pre-existing telemetry code:
- Hardening against `Telemetry::metrics_address` remaining empty and Civetweb thus launching with a broken URL of `:<port>`, which I hit by accident
- A few comment tweaks where content was outdated

This does not yet include a zeek-docs documentation update.

See commit messages for details.


scripts/policy/frameworks/management/controller/main.zeek
```diff
@@ -372,6 +389,83 @@ function config_assign_ports(config: Management::Configuration)
...
+	sort(nodes, function [roles] (n1: Management::Node, n2: Management::Node): int
+		{ return config_nodes_compare(n1, n2, roles); });
+
+	for ( i in nodes )
+		{
+		node = nodes[i];
+
+		# Find next available port ...
+		while ( p in ports_set )
+			++p;
```
rsmmr, line 474: Here, or maybe elsewhere[1]: should we check that we don't end up conflicting with other ports Zeek knows/uses? Starting at 9000 makes it unlikelier, but somebody could be changing defaults for some ports.

[1] "elsewhere" would also allow to check manually configured ports, in addition to auto-assigned.

scripts/policy/frameworks/management/controller/main.zeek
```diff
@@ -186,6 +186,24 @@ global g_configs: table[ConfigState] of Management::Configuration
...
+		{
+		if ( inst$name in g_instances_known
+		    && inst$host == 0.0.0.0
+		    && g_instances_known[inst$name]$host != 0.0.0.0 )
+			inst$host = g_instances_known[inst$name]$host;
+
+		add instances[inst];
+		}
+
+	config$instances = instances;
```
rsmmr, line 205: Just out of curiosity: why can't we modify the values in `config$instances` directly?

rsmmr (approved):

scripts/policy/frameworks/management/controller/main.zeek
ckreibich: It just doesn't work that way for sets due to how we store the indexes in the underlying dict. By the time we get an `inst`, it's not the one in the table itself.

See here: https://try.zeek.org/#/tryzeek/saved/f34d8aee749142ff8d1e11237ab22a4e

If this were a table, we could modify its values (as opposed to the keys) directly.

scripts/policy/frameworks/management/controller/main.zeek
ckreibich: Oh that's a good point — we can do more cross-checking, at least against the configured (or auto-assigned) Broker ports the agent just processed before this. I'll work that in, it's easy.

Beyond that I think this just is a bit best-effort since we cannot know what other ports the user's loaded scripts may open, say via some Zeek package. I did wonder about ports already taken due to other (non-Zeek) services, but decided to table that at least for now.

scripts/policy/frameworks/management/controller/main.zeek
rsmmr: Ah, right, got it.

PR #3814, opened 2024-07-08
Move prof.log metrics to telemetry framework

### LINKED ISSUE #3081: Add prof.log metrics to telemetry
Some of the `prof.log` data produced would be useful to be exposed via the telemetry framework and Prometheus directly. Either by leveraging the C++ API or scripting.

Reference to method writing out this data:
https://github.com/zeek/zeek/blob/master/src/Stats.cc#L72

The implementation might be in the subsystems directly: E.g. updating counters when producing or consuming messages for threads, rather than hooking into the existing `ProfileLogger::Log()` method.

              This sounds like something that a ticket could be opened for as a future enhancement.

_Originally posted by @timwoj in https://github.com/zeek/zeek-docs/pull/186#discussion_r1204509867_

timwoj:
This PR makes telemetry framework metrics out of most of the metrics that are included in `prof.log`. It skips the "expensive" metrics. This doesn't need to land before 7.0.

Fixes #3081


src/threading/MsgThread.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <atomic>
+#include <memory>
```
bbannier, line 4: Since you are at it, many of these are unused, and we only seem to need these includes:

```cpp
#include <memory>

#include "zeek/threading/BasicThread.h"
#include "zeek/threading/Queue.h"
```

src/threading/MsgThread.cc
```diff
@@ -227,6 +225,83 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
 
     io_source = new detail::IOSource(this);
 
+    cnt_sent_in_metric = telemetry_mgr->CounterInstance("zeek", "msg_thread_msgs_sent_in", {{"thread_name", Name()}},
+                                                        "Number of messages sent into thread");
+    cnt_sent_out_metric = telemetry_mgr->CounterInstance("zeek", "msg_thread_msgs_sent_out", {{"thread_name", Name()}},
```
bbannier, line 230: These two names where he before, but I don't think we need Hungarian notation here (i.e., drop the `cnt_` prefixes). Ideally we'd have consistent naming, see my previous comments.

Also check below for consistent naming.

src/threading/MsgThread.cc
```diff
@@ -227,6 +225,83 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
...
+                                                         "Number of messages sent from thread");
+    pending_in_metric = telemetry_mgr->GaugeInstance("zeek", "msg_thread_msgs_pending_in", {{"thread_name", Name()}},
+                                                     "Number of pending messages sent into thread", "",
+                                                     [this]() -> prometheus::ClientMetric {
```
bbannier, line 234: The captured `this` in these is even harder to follow than previous instances since we don't have a clear global we could reference. If e.g., each `MsgThread` had a unique ID we could check whether the instance is still valid by capturing the ID and then going via the global `thread_mgr` (checked), but we don't have one. Do you have any other ideas for making this less fragile?

src/threading/MsgThread.cc
```diff
@@ -227,6 +225,83 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
 
     io_source = new detail::IOSource(this);
```
awelzel, line 226: Would keep this close to the `io_mgr->Register()` call (or move the Register() call up).

src/threading/MsgThread.cc
```diff
@@ -227,6 +225,83 @@ MsgThread::MsgThread() : BasicThread(), queue_in(this, nullptr), queue_out(nullp
...
 
+    cnt_sent_in_metric = telemetry_mgr->CounterInstance("zeek", "msg_thread_msgs_sent_in", {{"thread_name", Name()}},
```
awelzel, line 228: Personally don't like the shortening of msgs, so would expand and also flip so that the unit (messages) is to the right, also dropping `sent`.
```
zeek_msg_thread_in_messages
zeek_msg_thread_out_messages
```

For the pending
```
zeek_msg_thread_messages_in_pending
zeek_msg_thread_messages_out_pending
```
---

However, and kind of more important than naming: With short lived threads (as for example used with the reproducer of #3682 with `Exec::run`), is this causing unbounded state growth over time because these metrics are kept around and labels based on an ever incrementing thread ID?

Is it possible to produce these metrics on demand for threads that are still alive? I've done that elsewhere with the golang library, but it might not fit our model here.

awelzel: Yeah, and using the reproducer from #3682 and subsequently scraping the metrics endpoints crashes an ASAN build:

```
root@tinkyx1:/home/awelzel/corelight-oss/zeek# zeek -C -i lo Telemetry::metrics_port=4222/tcp issues/3682-flare-bad-pipe-op/test.zeek 
<params>, line 1: listening on lo                                                                                      
                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                               
1720614889.519241 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                                       
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                           
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                              
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                           
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                                                                                                                                                                                                                                                                                      
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                                                                                                                                                                                                                                                                                      
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                   
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                   
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
1720614890.519514 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                   
1720614890.526037 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
1720614890.526037 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                                       
1720614890.526037 error in <params>, line 1: /cmd/does/not/exist |/Input::READER_RAW: Child process exited with non-zero return code 127                                                                                             
==732145==ERROR: AddressSanitizer: heap-use-after-free on address 0x61f000040fb8 at pc 0x5597ab7f533c bp 0x7fd3d05b1950 sp 0x7fd3d05b1940                                                                                                                                                                                                                                                                                                                                                     
READ of size 8 at 0x61f000040fb8 thread T3                                                                                                                                                                                                     
    #0 0x5597ab7f533b in std::operator-(std::_Deque_iterator<zeek::threading::BasicInputMessage*, zeek::threading::BasicInputMessage*&, zeek::threading::BasicInputMessage**> const&, std::_Deque_iterator<zeek::threading::BasicInputMessage*, zeek::threading::BasicInputMessage*&, zeek::threading::BasicInputMessage**> const&) /usr/include/c++/11/bits/stl_deque.h:356
    #1 0x5597ab7f533b in std::deque<zeek::threading::BasicInputMessage*, std::allocator<zeek::threading::BasicInputMessage*> >::size() const /usr/include/c++/11/bits/stl_deque.h:1234
    #2 0x5597ab7f533b in std::queue<zeek::threading::BasicInputMessage*, std::deque<zeek::threading::BasicInputMessage*, std::allocator<zeek::threading::BasicInputMessage*> > >::size() const /usr/include/c++/11/bits/stl_queue.h:209        
    #3 0x5597ab7f533b in zeek::threading::Queue<zeek::threading::BasicInputMessage*>::Size() /home/awelzel/corelight-oss/zeek/src/include/zeek/threading/Queue.h:219
    #4 0x5597ab7f533b in operator() /home/awelzel/corelight-oss/zeek/src/threading/MsgThread.cc:236                                                                                                                                            
    #5 0x5597ab7f533b in __invoke_impl<prometheus::ClientMetric, zeek::threading::MsgThread::MsgThread()::<lambda()>&> /usr/include/c++/11/bits/invoke.h:61                                                                                                                                                                                                                                                                                                                                       #6 0x5597ab7f533b in __invoke_r<prometheus::ClientMetric, zeek::threading::MsgThread::MsgThread()::<lambda()>&> /usr/include/c++/11/bits/invoke.h:116                                                                                      
    #7 0x5597ab7f533b in _M_invoke /usr/include/c++/11/bits/std_function.h:291                                                                                                                                                                                                                                                                                                                                                                                                                    #8 0x5597ac027146 in std::function<prometheus::ClientMetric ()>::operator()() const /usr/include/c++/11/bits/std_function.h:590                                                                                                                                                                                                                                                                                                                                                           
    #9 0x5597ac027146 in prometheus::Gauge::Collect() const /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/src/gauge.cc:40                                                                                                                                                                                                                                                                                                                                                        
    #10 0x5597ac008c5a in prometheus::Family<prometheus::Gauge>::CollectMetric(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_str
ing<char, std::char_traits<char>, std::allocator<char> > > > > const&, prometheus::Gauge*) const /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/src/family.cc:110             
    #11 0x5597ac0124ec in prometheus::Family<prometheus::Gauge>::Collect() const /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/src/family.cc:102
    #12 0x5597ac035ff3 in CollectAll<std::vector<std::unique_ptr<prometheus::Family<prometheus::Gauge> > > > /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/src/registry.cc:21                                                                                                                                                                                                                                                                                                    
    #13 0x5597ac035ff3 in prometheus::Registry::Collect() const /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/src/registry.cc:51                                                                                                  
    #14 0x5597ac0a3378 in prometheus::detail::CollectMetrics(std::vector<std::weak_ptr<prometheus::Collectable>, std::allocator<std::weak_ptr<prometheus::Collectable> > > const&) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/pull/src/metrics_collector.cc:20
    #15 0x5597ac093161 in prometheus::detail::MetricsHandler::handleGet(CivetServer*, mg_connection*) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/pull/src/handler.cc:155                                                                                                                                                                                                                                                                                                           
    #16 0x5597ac056aa2 in CivetServer::requestHandler(mg_connection*, void*) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/CivetServer.cpp:235                                                                                                                                                                                                                                                                                                                      #17 0x5597ac07b02a in handle_request /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:15037                                                                                                                                                                                                                                                                                                                                                         
    #18 0x5597ac07e550 in handle_request_stat_log /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:6569                                                                                                                                                                                                                                                                                                                                                 
    #19 0x5597ac07e550 in process_new_connection /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:19480                                                                                                                                                                                                                                                                                                                                                 
    #20 0x5597ac07e550 in worker_thread_run /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:19887                                                                                                                                                                                                                                                                                                                                                      
    #21 0x5597ac07e550 in worker_thread /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:19948                                                                                                                                                                                                                                                                                                                                                          
    #22 0x7fd3d6094ac2 in start_thread nptl/pthread_create.c:442                                                                                                                                                                                                                                                                                                                                                                                                                              
    #23 0x7fd3d612684f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f)                                                                                                                                                                             
                                                                                                                                                                                                                                               
0x61f000040fb8 is located 824 bytes inside of 3304-byte region [0x61f000040c80,0x61f000041968)                                                                                                                                                 
freed by thread T0 here:                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
    #0 0x7fd3dc6b724f in operator delete(void*, unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:172                                                                                                                       
    #1 0x5597aab80e91 in zeek::input::reader::detail::Raw::~Raw() /home/awelzel/corelight-oss/zeek/src/input/readers/raw/Raw.cc:60                                                                                                                                                                                                                                                                                                                                                            
    #2 0x5597ab7e2516 in zeek::threading::Manager::SendHeartbeats() /home/awelzel/corelight-oss/zeek/src/threading/Manager.cc:137                                                                                                              
    #3 0x5597ab7e2876 in zeek::threading::detail::HeartbeatTimer::Dispatch(double, bool) /home/awelzel/corelight-oss/zeek/src/threading/Manager.cc:20
    #4 0x5597ab645e02 in zeek::detail::TimerMgr::DoAdvance(double, int) /home/awelzel/corelight-oss/zeek/src/Timer.cc:173
    #5 0x5597ab646066 in zeek::detail::TimerMgr::Advance(double, int) /home/awelzel/corelight-oss/zeek/src/Timer.cc:86
    #6 0x5597ab646151 in zeek::detail::TimerMgr::Process() /home/awelzel/corelight-oss/zeek/src/Timer.cc:93
    #7 0x5597ab5b9487 in zeek::run_state::detail::run_loop() /home/awelzel/corelight-oss/zeek/src/RunState.cc:289
    #8 0x5597abf91c7b in main /home/awelzel/corelight-oss/zeek/src/main.cc:93                                          
    #9 0x7fd3d6029d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58                              
                                                                                                                       
previously allocated by thread T0 here:                                                                                                                                                                                                        
    #0 0x7fd3dc6b61e7 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:99                                                                                                                                  
    #1 0x5597aab87ddb in zeek::input::reader::detail::Raw::Instantiate(zeek::input::ReaderFrontend*) /home/awelzel/corelight-oss/zeek/src/include/zeek/input/readers/raw/Raw.h:29                                                              
    #2 0x5597aab20172 in zeek::input::Manager::CreateBackend(zeek::input::ReaderFrontend*, zeek::EnumVal*) /home/awelzel/corelight-oss/zeek/src/input/Manager.cc:187                                                                           
    #3 0x5597aab5d5c7 in zeek::input::ReaderFrontend::ReaderFrontend(zeek::input::ReaderBackend::ReaderInfo const&, zeek::EnumVal*) /home/awelzel/corelight-oss/zeek/src/input/ReaderFrontend.cc:36                                            
    #4 0x5597aab237f0 in zeek::input::Manager::CreateStream(zeek::input::Manager::Stream*, zeek::RecordVal*) /home/awelzel/corelight-oss/zeek/src/input/Manager.cc:252                                                                                                                                                                                                                                                                                                                        
    #5 0x5597aab2ca3f in zeek::input::Manager::CreateEventStream(zeek::RecordVal*) /home/awelzel/corelight-oss/zeek/src/input/Manager.cc:377                                                                                                   
    #6 0x5597ab3eb1c3 in zeek::BifFunc::Input::__create_event_stream_bif(zeek::detail::Frame*, std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > > const*) /home/awelzel/corelight-oss/zeek/src/input/input.bif:27
    #7 0x5597ab3a7e9f in zeek::detail::BuiltinFunc::Invoke(std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >*, zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Func.cc:739          
    #8 0x5597ab31937f in zeek::detail::CallExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:4123                                                                                                            
    #9 0x5597ab5fd47d in zeek::detail::ReturnStmt::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1404                                                                                   
    #10 0x5597ab6015a5 in zeek::detail::StmtList::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1439                                                                                    
    #11 0x5597ab3b22e8 in zeek::detail::ScriptFunc::Invoke(std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >*, zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Func.cc:386          
    #12 0x5597ab31937f in zeek::detail::CallExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:4123                                                                                                           
    #13 0x5597ab60043a in zeek::detail::ExprStmt::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:389                                                                                     
    #14 0x5597ab6015a5 in zeek::detail::StmtList::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1439                                                                                    
    #15 0x5597ab3b22e8 in zeek::detail::ScriptFunc::Invoke(std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >*, zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Func.cc:386          
    #16 0x5597ab31937f in zeek::detail::CallExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:4123                                                                                                           
    #17 0x5597ab309224 in zeek::detail::AssignExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:2404                                                                                                         
    #18 0x5597ab2fbf7e in zeek::detail::UnaryExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:544                                                                                                           
    #19 0x5597ab5fd47d in zeek::detail::ReturnStmt::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1404                                                                                  
    #20 0x5597ab60077c in zeek::detail::IfStmt::DoExec(zeek::detail::Frame*, zeek::Val*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:455                                                                         
    #21 0x5597ab6004c2 in zeek::detail::ExprStmt::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:392                                                                                     
    #22 0x5597ab6015a5 in zeek::detail::StmtList::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1439                                                                                    
    #23 0x5597ab3b22e8 in zeek::detail::ScriptFunc::Invoke(std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >*, zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Func.cc:386
    #24 0x5597ab31937f in zeek::detail::CallExpr::Eval(zeek::detail::Frame*) const /home/awelzel/corelight-oss/zeek/src/Expr.cc:4123
    #25 0x5597ab6519c3 in zeek::detail::trigger::Trigger::Eval() /home/awelzel/corelight-oss/zeek/src/Trigger.cc:246
    #26 0x5597ab655157 in zeek::detail::trigger::Trigger::Trigger(std::shared_ptr<zeek::detail::WhenInfo>, std::unordered_set<zeek::detail::ID const*, std::hash<zeek::detail::ID const*>, std::equal_to<zeek::detail::ID const*>, std::allocator<zeek::detail::ID const*> > const&, std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >, double, zeek::detail::Frame*, zeek::detail::Location const*) /home/awelzel/corelight-oss/zeek/src/Trigger.cc
:142                                                                                                                   
    #27 0x5597ab633d17 in zeek::IntrusivePtr<zeek::detail::trigger::Trigger> zeek::make_intrusive<zeek::detail::trigger::Trigger, std::shared_ptr<zeek::detail::WhenInfo>&, std::unordered_set<zeek::detail::ID const*, std::hash<zeek::detail::ID const*>, std::equal_to<zeek::detail::ID const*>, std::allocator<zeek::detail::ID const*> > const&, std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >&, double&, zeek::detail::Frame*&, zeek::deta
il::Location*&>(std::shared_ptr<zeek::detail::WhenInfo>&, std::unordered_set<zeek::detail::ID const*, std::hash<zeek::detail::ID const*>, std::equal_to<zeek::detail::ID const*>, std::allocator<zeek::detail::ID const*> > const&, std::vector<zeek::IntrusivePtr<zeek::Val>, std::allocator<zeek::IntrusivePtr<zeek::Val> > >&, double&, zeek::detail::Frame*&, zeek::detail::Location*&) /home/awelzel/corelight-oss/zeek/src/include/zeek/IntrusivePtr.h:179
    #28 0x5597ab62ca02 in zeek::detail::WhenStmt::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1967
    #29 0x5597ab6015a5 in zeek::detail::StmtList::Exec(zeek::detail::Frame*, zeek::detail::StmtFlowType&) /home/awelzel/corelight-oss/zeek/src/Stmt.cc:1439

Thread T3 created by T0 here:                                                                                          
    #0 0x7fd3dc658685 in __interceptor_pthread_create ../../../../src/libsanitizer/asan/asan_interceptors.cpp:216
    #1 0x5597ac05f0dd in mg_start_thread_with_id /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:5746
    #2 0x5597ac08817d in mg_start2 /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/civetweb.c:21148
    #3 0x5597ac053307 in CivetServer::CivetServer(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, CivetCallbacks const*, void const*) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/3rdparty/civetweb/src/CivetServer.cpp:463
    #4 0x5597ac088e51 in std::unique_ptr<CivetServer, std::default_delete<CivetServer> > prometheus::detail::make_unique<CivetServer, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, CivetCallbacks const*&>(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic
_string<char, std::char_traits<char>, std::allocator<char> > > >&&, CivetCallbacks const*&) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/core/include/prometheus/detail/future_std.h:12
    #5 0x5597ac088e51 in prometheus::Exposer::Exposer(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, CivetCallbacks const*) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/pull/src/exposer.cc:23
    #6 0x5597ac0899e8 in prometheus::Exposer::Exposer(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, CivetCallbacks const*) /home/awelzel/corelight-oss/zeek/auxil/prometheus-cpp/pull/src/exposer.cc:19
    #7 0x5597aa961ea5 in std::_MakeUniq<prometheus::Exposer>::__single_object std::make_unique<prometheus::Exposer, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, int, CivetCallbacks*&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, int&&, CivetCallbacks*&) /usr/include/c++/11/bits/unique_ptr.h:962
    #8 0x5597aa961ea5 in zeek::telemetry::Manager::InitPostScript() /home/awelzel/corelight-oss/zeek/src/telemetry/Manager.cc:78
    #9 0x5597ab122d0b in zeek::detail::setup(int, char**, zeek::Options*) /home/awelzel/corelight-oss/zeek/src/zeek-setup.cc:802
    #10 0x5597abf9192c in main /home/awelzel/corelight-oss/zeek/src/main.cc:52
    #11 0x7fd3d6029d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

SUMMARY: AddressSanitizer: heap-use-after-free /usr/include/c++/11/bits/stl_deque.h:356 in std::operator-(std::_Deque_iterator<zeek::threading::BasicInputMessage*, zeek::threading::BasicInputMessage*&, zeek::threading::BasicInputMessage**> const&, std::_Deque_iterator<zeek::threading::BasicInputMessage*, zeek::threading::BasicInputMessage*&, zeek::threading::BasicInputMessage**> const&)

```

src/threading/MsgThread.cc
awelzel: And it blows up, as shown above. Think it would be better to collect metrics in the manager and return for alive threads, only.

awelzel (changes requested):
Mostly commenting about naming and removing some redundant instances. There's `promtool check metrics` that does bring up the `_count` endings.

The MsgThread ones are tricky. It's essentially not allowed to have unbounded labels and it's exactly doing that. I'd probably drop that one and maybe creating summary stats over all threads if that's needed (and/or doing it for logging threads explicitly where we have bounded naming choices).

src/threading/MsgThread.h
timwoj: DebugLogger.h is needed for `void Debug(DebugStream stream, const char* msg);` but the rest can go.

src/Timer.cc
```diff
@@ -97,6 +98,41 @@ void TimerMgr::InitPostScript() {
         iosource_mgr->Register(this, true);
 
     dispatch_all_expired = zeek::detail::max_timer_expires == 0;
+
+    peak_size_metric = telemetry_mgr->CounterInstance("zeek", "timer_peak_count", {}, "Peak number of timers", "",
```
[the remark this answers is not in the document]
timwoj: I agree that `_total` suffix isn't right, but it's a counter in that it never goes down. I guess a gauge that never goes down still counts. I can remove it though, as you're right that someone could calculate it if they want.

src/Timer.cc
```diff
@@ -97,6 +98,41 @@ void TimerMgr::InitPostScript() {
...
+                                                   "Lag between current network time and last expired timer", "",
+                                                   [this]() -> prometheus::ClientMetric {
+                                                       prometheus::ClientMetric metric;
+                                                       metric.gauge.value =
+                                                           run_state::network_time - this->last_timestamp;
+                                                       return metric;
+                                                   });
+
+    std::shared_ptr<telemetry::GaugeFamily> family =
+        telemetry_mgr->GaugeFamily("zeek", "timer_count", {"type"}, "Number of timers for a certain type");
```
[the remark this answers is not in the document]
timwoj: I think if the others end up being `zeek_timers` and `zeek_timers_lag_seconds`, this one should be `zeek_timers_pending` just to keep with the theme.

src/Timer.cc
```diff
@@ -97,6 +98,41 @@ void TimerMgr::InitPostScript() {
...
+
+    cumulative_num_metric =
+        telemetry_mgr->GaugeInstance("zeek", "timer_cumulative_num", {}, "Cumulative number of timers", "",
+                                     [this]() -> prometheus::ClientMetric {
+                                         prometheus::ClientMetric metric;
+                                         metric.gauge.value = static_cast<double>(this->CumulativeNum());
+                                         return metric;
+                                     });
+
+    lag_time_metric = telemetry_mgr->GaugeInstance("zeek", "timer_lag_time", {},
```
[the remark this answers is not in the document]
timwoj: Changed to `timers_lag` and set the `unit` value to seconds.

src/packet_analysis/protocol/tcp/Stats.cc
```diff
@@ -4,34 +4,66 @@
...
+        case analyzer::tcp::TCP_ENDPOINT_CLOSED: return "closed";
+        case analyzer::tcp::TCP_ENDPOINT_RESET: return "reset";
+    }
+
+    return "<unknown>";
+}
+
 TCPStateStats::TCPStateStats() {
+    std::shared_ptr<telemetry::GaugeFamily> family =
+        telemetry_mgr->GaugeFamily("zeek", "tcp_stats", {"orig", "resp"}, "");
```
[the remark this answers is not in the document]
timwoj: Looking at them that way, I agree. We print them as a table in prof.log so it's more compact. I'll drop them out of this PR and make a note to see if I can reimplement them as a script.

src/telemetry/Manager.cc
```diff
@@ -114,13 +114,21 @@ void Manager::InitPostScript() {
...
                               });
 
-    cpu_gauge = GaugeInstance("process", "cpu", {}, "Total user and system CPU time spent", "seconds",
-                              []() -> prometheus::ClientMetric {
-                                  auto* s = get_stats();
-                                  prometheus::ClientMetric metric;
-                                  metric.gauge.value = s->cpu;
-                                  return metric;
-                              });
+    user_cpu_gauge = GaugeInstance("process", "user_cpu", {}, "Total user CPU time spent", "seconds",
```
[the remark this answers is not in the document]
timwoj: @bbannier Suggested the same

src/broker/Manager.cc
```diff
@@ -1706,7 +1744,12 @@ const Stats& Manager::GetStatistics() {
     statistics.num_stores = data_stores.size();
     statistics.num_pending_queries = pending_queries.size();
 
-    // The other attributes are set as activity happens.
+    statistics.num_events_incoming = static_cast<size_t>(num_events_incoming_metric->Value());
+    statistics.num_events_outgoing = static_cast<size_t>(num_events_outgoing_metric->Value());
+    statistics.num_logs_incoming = static_cast<size_t>(num_logs_incoming_metric->Value());
+    statistics.num_logs_outgoing = static_cast<size_t>(num_logs_outgoing_metric->Value());
+    statistics.num_ids_incoming = static_cast<size_t>(num_ids_incoming_metric->Value());
+    statistics.num_ids_outgoing = static_cast<size_t>(num_ids_outgoing_metric->Value());
```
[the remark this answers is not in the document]
timwoj: I don't feel like we have to cast these frequently enough to add an extra template for it.

src/threading/MsgThread.cc
timwoj: > However, and kind of more important than naming: With short lived threads (as for example used with the reproducer of #3682 with `Exec::run`), is this causing unbounded state growth over time because these metrics are kept around and labels based on an ever incrementing thread ID?

Lifetime of telemetry objects gets somewhat tricky. We keep a shared_ptr to every instrument in the family objects, which are in turn kept around in the manager. I spent some time looking at cleanup to where deleting an instrument flows through and cleans up the instance from the family plus the handle in prometheus-cpp. The latter is so that prom-cpp will stop trying to call `Collect` on a dead instrument, which is what crashed above. Unfortunately, this breaks script-land because functions like `counter_family_inc` expect to always return the same handle object and for that handle object to stick around (see the script/base/frameworks/telemetry/basic.zeek btest).

awelzel: FWIW, outside of code Zeek's internal cleanup, it's not a good pattern to have unbounded labels for metrics like in this case with `thread_name` - another take would be to accept that handles stay around forever and are never cleaned up (that's what broker did).

Alternatively, could the `Collect()` callback be decoupled from an instrument and produce them on-demand (I haven't looked at all how/if this could work).

timwoj: > it's not a good pattern to have unbounded labels for metrics like in this case with thread_name - another take would be to accept that handles stay around forever and are never cleaned up (that's what broker did).

Yeah, I saw that comment. I was focusing on the cleanup first since it's what's causing the crash you saw. The `MsgThread` object doesn't exist anymore, one of the `Collect`s gets called, and it tries to dereference a member variable that doesn't exist anymore. At the very least, we should remove the handle from prometheus-cpp so that doesn't continue trying to call the `Collect` methods once the thread is dead. Leaving the Zeek instrument objects around feels bad, but without the actual collection it's just memory at that point instead of a crash.

As for the unbounded name, I agree that's not good. I could set a max length on the name used but then it runs into the possibility of two threads with the same start of their name being combined into a single metric. Maybe something like 20 characters is enough to avoid that though.

> Alternatively, could the Collect() callback be decoupled from an instrument and produce them on-demand (I haven't looked at all how/if this could work).

`Collect` is a called internally by prometheus-cpp and expects to only get data back for one instrument (as a `ClientMetric` object). We could do tricks with custom `prometheus::Collectable` objects and inserting them by hand into the `Exposer` object but that feels like you're basically skipping around how prometheus-cpp works at that point.

timwoj: > At the very least, we should remove the handle from prometheus-cpp so that doesn't continue trying to call the Collect methods once the thread is dead. Leaving the Zeek instrument objects around feels bad, but without the actual collection it's just memory at that point instead of a crash.

Another possible way would be to leave the prom-cpp handle in place, but instead store off the last value from the callback method into the instrument and clear out the callback method. Pulls from prometheus would still return the last-known value for dead threads. I'm not sure which is more correct from the prometheus side though.

timwoj: I pushed a few commits for the comments in this thread. They do the following:
- Enforce a 20-character limit for thread names used in naming metrics
- Renames the msg thread metrics to be more in-line with what you mentioned above, while also grouping some of them and using `direction` labels to differentiate in/out versions instead of having separate metrics for those.
- Add/call a method at MsgThread cleanup time that removes the handle from prometheus-cpp. This leaves the metric object behind, but prometheus won't try to report it anymore.

awelzel: > Enforce a 20-character limit for thread names used in naming metrics

I'm not sure about that approach. Too short will quickly overlap, too large and it's a lot. I'm also not quite sure what the user would immediately do with it either. We could just do it for the first 1000 threads, but that's also very arbitrary. After pondering (see below), maybe transposing and making it buckets instead is useful?

```
$ echo -n "thread-9999999999999" | wc -c
20
```
It's a lot of threads/metrics :-)

Separate thoughts - not sure it's "easy" in a threads safe way (but maybe computing in the heartbeat timer is good), what if the threading manager would produce the following (and individual thread would not have any metrics).

* Sum of messages_in/out sent over all threads (counter, total)
* Sum of message_in/out pending over all threads (gauge)
* Number of current threads (gauge) and threads ever created metrics (counter)
* Something like a "gauge histogram" (I googled a bit and didn't find something obvious): Basically, track counts of threads that have more than >100, >1000 >5000, >10000, >50000, > 100000 messages pending in/out (not sure about the right bounds, but that seems a a bound and fairly condensed way while still providing useful information? One doesn't know the thread-name, but the numbered one isn't useful anyhow and event less if it's truncated significantly.
  ```
  zeek_msgthread_pending_messages_in = 12311
  zeek_msgthread_pending_messages_in_bucket{lt="100"} = 3
  zeek_msgthread_pending_messages_in_bucket{lt="1000"} = 4
  zeek_msgthread_pending_messages_in_bucket{lt="10000"} = 5
  zeek_msgthread_pending_messages_in_bucket{lt="100000"} = 5
  ```

  > them and using direction labels to differentiate in/out versions instead of having separate metrics for those.

  The process exporter doesn't do that for network packets or disk io. The rule of thumb is that `avg()` and `sum()` over dimensions should be meaningful (except for histograms), so seems okay to have them separate.

What do you think about the buckets?

Not everything from prof.log might fit prometheus well and the per thread ones might just be such a case given that we can have short-lived threads.

> Add/call a method at MsgThread cleanup time that removes the handle from prometheus-cpp. This leaves the metric object behind, but prometheus won't try to report it anymore.

Would that still "leak" memory if you create 1000 of threads per second?

timwoj: I like the idea of "summary" metrics for the threads as a whole vs individually.

> Not everything from prof.log might fit prometheus well and the per thread ones might just be such a case given that we can have short-lived threads.

I was beginning to think the same thing. Another thought along those lines (and in line with the change I made to dump dead threads) was that prof.log only shows output for active threads. Granted, it queries a lot more often than Prometheus would, but it does lend weight to that decision. Of course if all of those metrics go away it doesn't really matter.

> Would that still "leak" memory if you create 1000 of threads per second?

It would, because all of the zeek instruments are left behind in a map in the telemetry manager.

timwoj:
```
# HELP zeek_msgthreads_threads_total Total number of threads
# TYPE zeek_msgthreads_threads_total counter
zeek_msgthreads_threads_total 4
# HELP zeek_msgthreads_in_messages_total Number of inbound messages received
# TYPE zeek_msgthreads_in_messages_total counter
zeek_msgthreads_in_messages_total 0
# HELP zeek_msgthreads_out_messages_total Number of outbound messages sent
# TYPE zeek_msgthreads_out_messages_total counter
zeek_msgthreads_out_messages_total 0
# HELP zeek_msgthreads_pending_in_messages Pending number of inbound messages
# TYPE zeek_msgthreads_pending_in_messages gauge
zeek_msgthreads_pending_in_messages 0
# HELP zeek_msgthreads_pending_out_messages Pending number of outbound messages
# TYPE zeek_msgthreads_pending_out_messages gauge
zeek_msgthreads_pending_out_messages 0
# HELP zeek_msgthreads_pending_messages_in_buckets Number of threads with pending inbound messages split into buckets
# TYPE zeek_msgthreads_pending_messages_in_buckets gauge
zeek_msgthreads_pending_messages_in_buckets{lt="inf"} 0
zeek_msgthreads_pending_messages_in_buckets{lt="10000"} 0
zeek_msgthreads_pending_messages_in_buckets{lt="1000"} 0
zeek_msgthreads_pending_messages_in_buckets{lt="10"} 0
zeek_msgthreads_pending_messages_in_buckets{lt="100"} 0
zeek_msgthreads_pending_messages_in_buckets{lt="1"} 4
# HELP zeek_msgthreads_pending_messages_out_buckets Number of threads with pending outbound messages split into buckets
# TYPE zeek_msgthreads_pending_messages_out_buckets gauge
zeek_msgthreads_pending_messages_out_buckets{lt="inf"} 0
zeek_msgthreads_pending_messages_out_buckets{lt="10000"} 0
zeek_msgthreads_pending_messages_out_buckets{lt="1000"} 0
zeek_msgthreads_pending_messages_out_buckets{lt="10"} 0
zeek_msgthreads_pending_messages_out_buckets{lt="100"} 0
zeek_msgthreads_pending_messages_out_buckets{lt="1"} 4
```

Some things to point out:
- `zeek_msgthreads_threads_total` is the total number of threads, not just message threads.
- The `lt="1"` bucket is simply there to catch threads that don't have any pending messages. I could see removing it though, since you could calculate it from the others. It just looked strange that the numbers didn't add up.
- The `lt="inf`` bucket is similar to `histogram` instruments infinity bucket, intended to catch values higher than the last bucket.

timwoj:
There's still some findings with thread sanitizer that I need to track down here. Most of it boils down to the callbacks in the threading manager, and data races from civetweb querying various global variables and other state.

awelzel:
> Most of it boils down to the callbacks in the threading manager, and data races from civetweb querying various global variables and other state.

The fact that we have a thread querying all these makes this little surprising (and it is a bit scary).

I'm just throwing it out there, but def not for this PR and maybe never: Wonder if there could be an IO source that the civet-web thread transfers responsibility to. Then the main thread would run the collection by processing that IO source and resumes the thread civet-web thread when the result is available.

timwoj:
I rebased on top of the threading changes from the other PR and squashed everything down. I also noticed I missed reverting some atomic changes in MsgThread, so I fixed those. There's some minor timing increases in the benchmark numbers (connection-create especially) so I need to look into those, but this is ready for another review.

src/Timer.cc
```diff
@@ -97,6 +97,34 @@ void TimerMgr::InitPostScript() {
         iosource_mgr->Register(this, true);
 
     dispatch_all_expired = zeek::detail::max_timer_expires == 0;
+
+    cumulative_num_metric = telemetry_mgr->GaugeInstance("zeek", "timers", {}, "Cumulative number of timers", "",
```
awelzel, line 101: That should be a counter, not a gauge.

src/session/Manager.cc
```diff
@@ -75,7 +75,25 @@ class ProtocolStats {
 
 } // namespace detail
 
-Manager::Manager() { stats = new detail::ProtocolStats(); }
+Manager::Manager() {
+    stats = new detail::ProtocolStats();
+    killed_by_inactivity_metric =
+        telemetry_mgr->CounterInstance("zeek", "sessions_killed_by_inactivity", {},
```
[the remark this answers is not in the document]
awelzel: It would be nice to name that one such that it ends in "sessions_total", so `zeek_killed_by_inactivity_sessions_total`, or move to the labeled variant for the future>

src/telemetry/Counter.cc
```diff
@@ -21,6 +23,8 @@ double Counter::Value() const noexcept {
     return handle.Value();
 }
 
+void Counter::ClearHandle() { family->Remove(&handle); }
```
awelzel, line 26: Wasn't able to find the user of `ClearHandle()`, so would suggest to remove it. It's existence might be an incentive to link less about labels and whether unbounded metrics are produced, which we don't want.

Also for Gauge.

src/telemetry/Counter.cc
```diff
@@ -1,9 +1,11 @@
 #include "zeek/telemetry/Counter.h"
 
+#include "zeek/RunState.h"
```
awelzel, line 3: Leftover?

src/telemetry/Gauge.cc
```diff
@@ -1,5 +1,7 @@
 #include "zeek/telemetry/Gauge.h"
 
+#include "zeek/RunState.h"
```
awelzel, line 3: Leftover?

src/Timer.h
```diff
@@ -110,7 +117,6 @@ class TimerMgr final : public iosource::IOSource {
      * @param timer the timer to cancel
      */
     void Cancel(Timer* timer) { Remove(timer); }
-
```
awelzel, line 113: Hmm? :-)

src/broker/Manager.cc
```diff
@@ -716,7 +753,7 @@ bool Manager::PublishLogWrite(EnumVal* stream, EnumVal* writer, string path, int
     lb.msgs[topic].add(std::move(msg));
 
     if ( lb.message_count >= log_batch_size )
-        statistics.num_logs_outgoing += lb.Flush(bstate->endpoint, log_batch_size);
+        num_logs_outgoing_metric->Inc(static_cast<double>(lb.Flush(bstate->endpoint, log_batch_size)));
```
awelzel, line 756: Maybe on two lines for clarity?

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+    num_threads_metric =
+        telemetry_mgr->GaugeInstance("zeek", "msgthread_active_threads", {}, "Number of active threads", "",
+                                     []() -> prometheus::ClientMetric {
+                                         prometheus::ClientMetric metric;
+                                         metric.gauge.value =
+                                             thread_mgr ? static_cast<double>(thread_mgr->all_threads.size()) : 0.0;
+                                         return metric;
+                                     });
+
+    total_threads_metric = telemetry_mgr->CounterInstance("zeek", "msgthreads_threads", {}, "Total number of threads");
```
awelzel, line 92:
```suggestion
    total_threads_metric = telemetry_mgr->CounterInstance("zeek", "msgthread_threads", {}, "Total number of threads");
```

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+    total_messages_in_metric =
+        telemetry_mgr->CounterInstance("zeek", "msgthreads_in_messages", {}, "Number of inbound messages received", "",
```
awelzel, line 94:
```suggestion
        telemetry_mgr->CounterInstance("zeek", "msgthread_in_messages", {}, "Number of inbound messages received", "",
```

And also for all the others? Or `msgthread_active_threads` should be plural, but like the singular one better.

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+                for ( const auto& t : thread_mgr->msg_threads ) {
+                    t->GetStats(&thread_stats);
+
+                    thread_mgr->current_bucketed_messages.sent_in_total += thread_stats.sent_in;
+                    thread_mgr->current_bucketed_messages.sent_out_total += thread_stats.sent_out;
+                    thread_mgr->current_bucketed_messages.pending_in_total += thread_stats.pending_in;
+                    thread_mgr->current_bucketed_messages.pending_out_total += thread_stats.pending_out;
+
+                    for ( auto idx : pending_bucket_brackets ) {
+                        if ( thread_stats.pending_in <= idx ) {
```
awelzel, line 63: If you compare `<=` here, then we should change the label to `le` rather than `lt`. Will make a comment there, too.

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+        telemetry_mgr->GaugeInstance("zeek", "msgthreads_pending_out_messages", {},
+                                     "Pending number of outbound messages", "", []() -> prometheus::ClientMetric {
+                                         auto* s = get_message_thread_stats();
+                                         prometheus::ClientMetric metric;
+                                         metric.gauge.value = static_cast<double>(s->pending_out_total);
+                                         return metric;
+                                     });
+
+    pending_message_in_buckets_fam =
+        telemetry_mgr->GaugeFamily("zeek", "msgthreads_pending_messages_in_buckets", {"lt"},
```
awelzel, line 129:
```suggestion
        telemetry_mgr->GaugeFamily("zeek", "msgthread_pending_messages_in_buckets", {"le"},
```

src/threading/Manager.cc
```diff
@@ -22,6 +22,8 @@ void HeartbeatTimer::Dispatch(double t, bool is_expire) {
 
 } // namespace detail
 
+std::vector<uint64_t> pending_bucket_brackets = {1, 10, 100, 1000, 10000, std::numeric_limits<uint64_t>::infinity()};
```
awelzel, line 25: Should this be `static` or in an anonymous namespace if we have one?

The zeekier way would probably to have this in BifConst as a vector of count (so it could also be disabled if its empty), though not sure it's worth it really, but it might make testing easier.

I'd also randomly add 100000, but that's just a hunch :-)

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+                MsgThread::Stats thread_stats;
+                for ( const auto& t : thread_mgr->msg_threads ) {
+                    t->GetStats(&thread_stats);
+
+                    thread_mgr->current_bucketed_messages.sent_in_total += thread_stats.sent_in;
+                    thread_mgr->current_bucketed_messages.sent_out_total += thread_stats.sent_out;
+                    thread_mgr->current_bucketed_messages.pending_in_total += thread_stats.pending_in;
+                    thread_mgr->current_bucketed_messages.pending_out_total += thread_stats.pending_out;
+
+                    for ( auto idx : pending_bucket_brackets ) {
```
awelzel, line 62: Total nit, the `idx` I read as vector index (0...buckets.size()). Maybe `lim` or `upper` would be easier to reason/skim.

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+                                     });
+
+    pending_message_in_buckets_fam =
+        telemetry_mgr->GaugeFamily("zeek", "msgthreads_pending_messages_in_buckets", {"lt"},
+                                   "Number of threads with pending inbound messages split into buckets");
+    pending_message_out_buckets_fam =
+        telemetry_mgr->GaugeFamily("zeek", "msgthreads_pending_messages_out_buckets", {"lt"},
+                                   "Number of threads with pending outbound messages split into buckets");
+
+    for ( auto idx : pending_bucket_brackets ) {
```
awelzel, line 135:
```suggestion
    for ( auto upper : pending_bucket_brackets ) {
```
?

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+        if ( idx == std::numeric_limits<uint64_t>::infinity() )
+            idx_str = "inf";
+        else
+            idx_str = std::to_string(idx);
+
+        current_bucketed_messages.pending_in[idx] = 0;
+        current_bucketed_messages.pending_out[idx] = 0;
+
+        pending_message_in_buckets[idx] =
+            pending_message_in_buckets_fam->GetOrAdd({{"lt", idx_str}}, [idx]() -> prometheus::ClientMetric {
```
awelzel, line 146:
```suggestion
            pending_message_in_buckets_fam->GetOrAdd({{"le", idx_str}}, [idx]() -> prometheus::ClientMetric {
```

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+                auto* s = get_message_thread_stats();
+                prometheus::ClientMetric metric;
+                metric.gauge.value = static_cast<double>(s->pending_in.at(idx));
+                return metric;
+            });
+        pending_message_out_buckets[idx] =
+            pending_message_out_buckets_fam->GetOrAdd({{"lt", idx_str}}, [idx]() -> prometheus::ClientMetric {
```
awelzel, line 153:
```suggestion
            pending_message_out_buckets_fam->GetOrAdd({{"le", idx_str}}, [idx]() -> prometheus::ClientMetric {
```

src/threading/Manager.cc
```diff
@@ -36,6 +38,127 @@ Manager::~Manager() {
...
+                auto* s = get_message_thread_stats();
```
awelzel, line 154: Nothing to do: I just thought that a callback that'n not bound to a given counter/gauge would work nicely here.

awelzel (approved):
I had another set of comments. See what you think. 

I skimmed the perf results and `micro-table-ops-create` stands out with 2-3%, but not sure if that's really related.

I suspect we should document somewhere, that callbacks are running in the main thread now: If you don't know many questions arise about how this is safe :-)

src/broker/Manager.cc
timwoj: That's just how clang-format formats it.

src/threading/Manager.cc
timwoj: Switched it to `<` and left the label as `lt`. It's basically just so all of the zero values fall into the `1` bucket.

src/threading/Manager.cc
timwoj: Changed to `upper_limit`

src/broker/Manager.cc
awelzel: I thought of having an intermediate value that's then passed to `Inc()` :-)

src/threading/Manager.cc
awelzel: Oh, I wasn't explicit enough. The Prometheus histogram uses the "le" label for buckets, so I figured staying consistent with that would've been nice.

PR #3815, opened 2024-07-09
Fix for --display-cmake option in configure

### LINKED ISSUE #3767: `--display-cmake` option from `configure` modifies build directory
The help text on the the option states:

> don't create build configuration, just output final CMake invocation

This isn't necessarily true. If you invoke this without a build directory existing, it creates one. If you invoke with an existing build directory, it touches the directory in some way and the next build will re-run cmake. I'm not sure if there are any changes actually made to the existing directory.

cknill:
Fixes  #3767

The configure script was setting up the build tree before checking if --display-cmake was set. It looks like it was as simple as moving that build tree setup further down the script.

### DIFF
--- a/configure
+++ b/configure
@@ -458,6 +458,19 @@ if [ -z "$CMakeCommand" ]; then
     fi
 fi
 
+echo "Using $(cmake --version | head -1)"
+echo
+if [ -n "$CMakeGenerator" ]; then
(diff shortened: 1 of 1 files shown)


awelzel (commented):
Thanks, merged.

PR #3820, opened 2024-07-09
CI: Use ccache and a single CPU when building spicy analyzers for btests

timwoj:
Some of the CI tasks are struggling with OOM issues when running tests. @bbannier dug a bit and it appears that the spicy jobs are causing extra load by using the maximum number of CPUs. This is compounded by the fact that we run as many btest jobs as we can, up to the number of CPUs. This PR tries to reduce the load first by using ccache for spicy builds, and second by only using a single CPU for the spicy plugin builds.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -59,7 +59,9 @@ function run_btests {
 
     pushd testing/btest
 
-    ZEEK_PROFILER_FILE=$(pwd)/.tmp/script-coverage/XXXXXX \
+    HILTI_CXX_COMPILER_LAUNCHER=ccache \
+        HILTI_JIT_PARALLELISM=1 \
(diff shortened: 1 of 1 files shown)


ckreibich (approved):

awelzel:
We already have the following snippet in btest.cfg, enabling ccache if Zeek was configured with ccache, too.

https://github.com/zeek/zeek/blob/b995924b217ad8cc6ba96fddb1638eafd2bbcd5b/testing/btest/btest.cfg#L16

I'm tempted to suggest moving HILTI_JIT_PARALLELISM=1 to btest.cfg as well, `btest -j spicy` after a Spicy bump is pretty much guaranteed to knock my box out (using gcc) if I'm not using memory confinement.

PR #3817, opened 2024-07-09
Track disabling analyzers in `analyzer.log`

J-Gras:
This adds an option to track disabling of analyzers in `analyzer.log`.

PR #3824, opened 2024-07-10
Log disabled analyzers by default

J-Gras:
This is a follow-up on #3817 requested by @awelzel to turn on logging of disabled analyzers by default. For the websocket btests, I was surprised to see that the SSH analyzer gets somehow disabled. I assume this is intentional but it would be nice to verify.


awelzel:
> For the websocket btests, I was surprised to see that the SSH analyzer gets somehow disabled. I assume this is intentional but it would be nice to verify.

There is `SSH::disable_analyzer_after_detection` (similarly as for SSL), so that seems expected.

I'm torn - for debugging purposes defaulting it to on is nice, but otherwise not so sure. I think going with #3817 is sufficient.

J-Gras:
> There is `SSH::disable_analyzer_after_detection` (similarly as for SSL), so that seems expected.

Interesting this did not pop up for any other SSH-related test.

PR #3826, opened 2024-07-11
ldap: Handle integrity-only KRB wrap tokens

awelzel:
Mostly staring at the PCAPs and opened a few RFCs. For now, only if the MS_KRB5 OID is used and accepted in a bind response, start stripping KRB5 wrap tokens for both, client and server traffic.

Would probably be nice to forward the GSS-API data to the analyzer...

Closes zeek/spicy-ldap#29.


awelzel:
@bbannier - mind taking a look. I was forth and back, but eventually removed the "magic" MessageWrapper logic. I believe we should rather explicitly support SASL/GSS-API, then `&try / backtrack()` parse the input. At least for the pcaps we have, this works reasonably well.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-  var sasl_success: bool = False;
-  : SASLLayer &try if ( self.success == False ) {
-    if ( $$.success == True ) {
-      self.sasl_success = True;
+  ctx_flags :bitfield(8) {
+    send_by_acceptor: 0;
+    sealed: 1;
+    acceptor_subkey: 2;
+  };
+  filler: b"\xff";
```
bbannier, line 168:
```suggestion
  filler: skip b"\xff";
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-  # now (in the long run we need to be parsing SASL/GSSAPI instead, in which case encrypted payloads
-  # are just another message type).
-
-  # SASLLayer (see unit above) just consumes bytes &until=b"\x30" or backtracks if it isn't found
-  # and sets a success flag we can use later to decide if those bytes contain a parsable message.
-  var sasl_success: bool = False;
-  : SASLLayer &try if ( self.success == False ) {
-    if ( $$.success == True ) {
-      self.sasl_success = True;
+  ctx_flags :bitfield(8) {
```
bbannier, line 163:
```suggestion
  ctx_flags: bitfield(8) {
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-  # we should re-add it to the remainder if the delimiter was found. If the delimiter was not found, we
-  # leave the remainder empty, but note that the bytes must be consumed either way to avoid stalling the
-  # parser and causing an infinite loop error.
-  : bytes &eod if ( self.success == False )  {
-    if ( self.sasl_success == True ) {
-      self.remainder = b"\x30" + $$;
-    }
-  }
+  snd_seq: uint64;
+  header_e: bytes &size=self.header_ec;
```
bbannier, line 182: Could mark these `skip` as well for now.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-    if ( $$.success == True ) {
-      self.sasl_success = True;
+  ctx_flags :bitfield(8) {
+    send_by_acceptor: 0;
+    sealed: 1;
+    acceptor_subkey: 2;
+  };
+  filler: b"\xff";
+  ec: uint16; # extra count
+  rrc: uint16 { # right rotation count
```
bbannier, line 170: Nit: no need to give this a name since we only reference the field value in its hook, we could use `$$` there.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-  # If we still didn't manage to parse a message (so the &try resulted in another backtrack()) then
-  # this is probably an encrypted LDAP message, so skip it
+  switch ( self?.krb_wrap_token && ! self.krb_wrap_token.ctx_flags.sealed ) {
+    True -> : Message(ctx)[] &eod;
+    * -> : skip bytes &eod;
+  } &size=self.switch_size;
 
-} &convert=self.message;
+  # Consume the wrap token trailer, if any.
+  trailer_e: bytes &size=self.krb_wrap_token.trailer_ec if (self?.krb_wrap_token);
```
bbannier, line 213:
```suggestion
  trailer_e: skip bytes &size=self.krb_wrap_token.trailer_ec if (self?.krb_wrap_token);
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-      if ($$.body.seq.submessages[1]?.application_id) {
-        self.opcode = cast<ProtocolOpcode>(cast<uint8>($$.body.seq.submessages[1].application_id));
-        self.applicationBytes = $$.body.seq.submessages[1].application_data;
-      }
-    }
+  seqHeader: ASN1::ASN1Header &requires=($$.tag.class == ASN1::ASN1Class::Universal && $$.tag.type_ == ASN1::ASN1Type::Sequence);
+
+  : void {
+    self.seqHeaderLen = self.offset();
+  }
```
bbannier, line 231:
```suggestion
  seqHeader: ASN1::ASN1Header &requires=($$.tag.class == ASN1::ASN1Class::Universal && $$.tag.type_ == ASN1::ASN1Type::Sequence) {
    self.seqHeaderLen = self.offset();
  }
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -267,13 +264,18 @@ public type Message = unit {
     ProtocolOpcode::INTERMEDIATE_RESPONSE   -> INTERMEDIATE_RESPONSE:   NotImplemented(self);
     ProtocolOpcode::MOD_DN_REQUEST          -> MOD_DN_REQUEST:          NotImplemented(self);
     ProtocolOpcode::SEARCH_RESULT_REFERENCE -> SEARCH_RESULT_REFERENCE: NotImplemented(self);
-  } &parse-from=self.applicationBytes if ( self.opcode );
+  } &size=self.protocolOp.len.len;
 
-  on %error {
+  # Eat the controls field if it exists.
+  : skip bytes &size=self.seqHeader.len.len - (self.offset() - self.seqHeaderLen);
```
bbannier, line 270: Are there ways where we could genuinely see the subtraction underflow? If we could let's instead use `&size=0` here (checking requires casting the values to `int64`); if we cannot see that let's maybe add explicit `&requires` checking the values as soon we have them available since for a user "integer underflow" looks like a bug, but `&requires failed` does less.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -126,125 +126,122 @@ public type Result = unit {
...
-      }
-    }
+  seqHeader: ASN1::ASN1Header &requires=($$.tag.class == ASN1::ASN1Class::Universal && $$.tag.type_ == ASN1::ASN1Type::Sequence);
+
+  : void {
+    self.seqHeaderLen = self.offset();
+  }
+
+  messageID_header: ASN1::ASN1Header &requires=($$.tag.class == ASN1::ASN1Class::Universal && $$.tag.type_ == ASN1::ASN1Type::Integer);
+  messageID_body: ASN1::ASN1Body(self.messageID_header, False) {
```
bbannier, line 234: No need to name this,

```suggestion
 : ASN1::ASN1Body(self.messageID_header, False) {
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -267,13 +264,18 @@ public type Message = unit {
     ProtocolOpcode::INTERMEDIATE_RESPONSE   -> INTERMEDIATE_RESPONSE:   NotImplemented(self);
     ProtocolOpcode::MOD_DN_REQUEST          -> MOD_DN_REQUEST:          NotImplemented(self);
     ProtocolOpcode::SEARCH_RESULT_REFERENCE -> SEARCH_RESULT_REFERENCE: NotImplemented(self);
-  } &parse-from=self.applicationBytes if ( self.opcode );
+  } &size=self.protocolOp.len.len;
```
bbannier, line 267: NB: Ugly, but this is really due to how the ASN.1 types look. In a more ideal world they would extract more "natural" values. Similar for `self.seqHeader.len.len` below.

bbannier (commented):
Without having validated this against the spec from the implementation side this looks all reasonable, I left only minor style comments.

The last patch removing `MessageWrapper` looks scary, but with `Message` being able to parse `NotImplemented` now this seems to make sense. Can we add a test against some PCAP which tickles the extension point handled by that?

src/analyzer/protocol/ldap/ldap.spicy
awelzel: Think I prefer to keep the name. The `$$` can be fairly cryptic.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: > Are there ways where we could genuinely see the subtraction underflow?

Not genuinely, don't think so. Let me put a `&requires`.

src/analyzer/protocol/ldap/ldap.spicy
bbannier: I mean, the compressed field name also only made sense to me with the comment. 🤷

awelzel:
> Can we add a test against some PCAP which tickles the extension point handled by that?

I've got an LDAP StartTLS PCAP now and that seemed to work (it tickled a left-over backtrack() elsewhere however). I can add that as a test - though tempted to actually support it then :-)

Ok, I have a trace for adding entries via the addRequest which isn't implemented either. I'll use that one.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: This is subtly different :-( staying with the `: void`.

bbannier (approved):
Thanks for adding the tests, looks reasonable for me from a technical point of view. I didn't validate against the spec.

PR #3827, opened 2024-07-11
Fix warning about grealpath when running 'make dist' on Linux

timwoj:
This fixes the following error if you run `make dist` on Linux:

```
tim@nuc release-7.0% make dist
error: cannot run grealpath: No such file or directory
fatal: run_command returned non-zero status for auxil/bifcl
.
error: cannot run grealpath: No such file or directory
fatal: run_command returned non-zero status for auxil/bifcl
.
```


awelzel (approved):
Think `--relative-to=.` would be fine as well, just `--relative-to=` isn't great :-)

PR #3830, opened 2024-07-12
Update cmake and broker submodules to fix configure failures with FETCHCONTENT_FULLY_DISCONNECTED flag

timwoj:


ckreibich (approved):
I'm looking into the cluster failure, which is irrelevant to this PR.

PR #3832, opened 2024-07-16

bbannier:
This patch bump Spicy to the latest development snapshot. This introduces a backwards-incompatible change in that it removes support for a never officially supported syntax to specify unit fields (so I would argue: not strictly a breaking change).


bbannier:
I'd argue that this one should go into both `master` and `release/7.0`, does that still make sense to you as well @rsmmr?

rsmmr:
Seems fine to me.

timwoj (approved):
The Windows build here isn't important, and it's failing for Cirrus infrastructure reasons.

PR #3833, opened 2024-07-17
ldap: Implement extendedRequest/extendedResponse and StartTLS support

awelzel:
* Adds support for extendedRequest / extendedResponse parsing and StartTLS
* Fixes limitations introduced during #3826 around SASL authentication that doesn't continue encrypted
* Reintroduces a heuristic to detect encrypted messages after SASL bindResponse

The last one was reported by @dopheide-esnet - I've unfortunately not been able to figure out how to configure OpenLDAP to act like the private pcap he shared.


awelzel:
> I've unfortunately not been able to figure out how to configure OpenLDAP to act like the private pcap he shared.

Pushed in the latest commit now.

src/analyzer/protocol/ldap/ldap_zeek.spicy
```diff
@@ -0,0 +1,13 @@
...
+import zeek;
+
+on LDAP::ExtendedResponse::%done priority=-5 {
+  if ( ctx.messageMode == LDAP::MessageMode::TLS )
+    zeek::protocol_begin("SSL");
+}
+
+on LDAP::TlsForward::chunk {
+  zeek::protocol_data_in(zeek::is_orig(), self.chunk);
+}
```
bbannier, line 13: I find that having `ExtendedResponse` open the handle and `TlsForward` use it (via an implicit handle) to be confusing and potentially error prone. Is there are reason you couldn't open the handle on `TlsForward::%init`? If that is impossible I'd at least try to store the handle as a unit var and forward it to users like `TlsForward` as a unit parameter.

src/analyzer/protocol/ldap/ldap.evt
```diff
@@ -41,3 +41,18 @@ on LDAP::SearchRequest -> event LDAP::search_request($conn,
...
+                                                         self.requestValue);
+
+on LDAP::ExtendedResponse -> event LDAP::extended_response($conn,
+                                                           message.messageID,
+                                                           message.result_.code,
+                                                           self.responseName,
+                                                           self.responseValue);
+
+# Once switched into MessageMode::TLS, we won't parse messages anymore,
+# so this is raised just once.
```
bbannier, line 57: Likely only true as long as the analyzer does not perform e.g., resynchronization, or is refactored where TLS parsing does not consume all data anymore.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -980,16 +1066,69 @@ type AbandonRequest = unit(inout message: Message) {
...
+  };
+
+  switch ( message.opLen > self.offset() ) {
+    True -> : ExtendedResponseEntry(self);
+    * -> : void;
+  };
+
+  on %done {
+    # Client had requested StartTLS and it was successful? Switch to SSL.
+    if ( ctx.startTlsRequested && message.result_.code == ResultCode::SUCCESS )
```
bbannier, line 1128: What happens the request failed? Not sure this is done elsewhere already, but ideally we'd handle it here explicitly as well and transition the unit to some failed state. If this cannot happen we could `assert` that.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -980,16 +1066,69 @@ type AbandonRequest = unit(inout message: Message) {
...
+  : Result {
+    message.result_ = $$;
+  }
+
+  # Try to parse two ASN1 entries if there are bytes left. Both
+  # are optional and identified by context specific tagging.
+  switch ( message.opLen > self.offset() ) {
+    True -> : ExtendedResponseEntry(self);
+    * -> : void;
+  };
```
bbannier, line 1119: I'd write this with a field `if`, e.g.,

```suggestion
    : ExtendedResponseEntry(self) if (message.opLen > self.offset());
```

Similar for the `ExtendedResponseEntry` below.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -980,16 +1066,69 @@ type AbandonRequest = unit(inout message: Message) {
...
-# type ExtendedRequest = unit(inout message: Message) {
-#
-# };
+  # If there's more byte to parse, it's the requestValue.
+  switch ( message.opLen > self.offset() ) {
+    True -> : ASN1::ASN1Message(False) &requires=($$.head.tag.class == ASN1::ASN1Class::ContextSpecific) {
+      self.requestValue = $$.application_data;
+    }
+    * -> : void;
+  };
```
bbannier, line 1082: Could be a field with `if`.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -980,16 +1066,69 @@ type AbandonRequest = unit(inout message: Message) {
...
-#
-# };
+
+  on %done {
+   # Did the client request StartTLS?
+   #
+   # https://datatracker.ietf.org/doc/html/rfc4511#section-4.14.1
+   if ( self.requestName == "1.3.6.1.4.1.1466.20037" )
+     ctx.startTlsRequested = True;
+  }
```
bbannier, line 1091: Any reason we cannot do this in the hook for `requestName` already?

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -130,29 +130,100 @@ public type Result = unit {
...
 };
 
 
+#-----------------------------------------------------------------------------
+type MaybeEncrypted = unit(ctx: Ctx&) {
+  # A plaintext LDAP message always starts with at least 3 bytes and the first
+  # byte is 0x30 for the sequence. A SASL encrypted message starts with a 4 byte
+  # length field. The heuristic here is that if we see 0x30, assume it's
+  # unencrypted. This should be pretty good, as we deal with a LDAP message
+  # of minimally size 0x30000000 (~768 MB) and that seems unlikely.
```
bbannier, line 172: Maximally? 😅

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -130,29 +130,100 @@ public type Result = unit {
...
+  var start: iterator<stream>;
+  var saslLen: uint64;
+  var mech: bytes;
+
+  on %init {
+    self.start = self.input();
+    # Don't have starts_with() on string, work around that.
```
bbannier, line 179: Let's open a feature request and link it here.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -130,29 +130,100 @@ public type Result = unit {
...
+    }
+  }
+
+  # Rewind the input.
+  : void {
+    self.set_input(self.start);
+  }
+
+  # One recursion to parse with the new ctx.messageMode setting.
+  : MessageDispatch(ctx);
```
bbannier, line 212: Nit: the dispatch feels slightly scary as this unit consumes no data and in some unfortunate future could recurse indefinitely. Might be worth adding an `assert ctx.messageMode != MessageMode::MAYBE_ENCRYPTED` to the rewinding hook.

Also `MaybeEncrypted` will not expose its payload since the `MessageDispatch` is an anon field. Would it make sense to name it, or maybe even `&convert=self.dispatched` when `MaybeEncrypted` is done parsing to transparently remove the indirection?

bbannier (approved):

scripts/base/protocols/ldap/main.zeek
```diff
@@ -258,6 +258,9 @@ event LDAP::message(c: connection,
       }
 
       m$object = object;
+
+      if ( opcode == LDAP::ProtocolOpcode_EXTENDED_REQUEST )
```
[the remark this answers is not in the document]
awelzel: The `EXTENDED_RESPONSE` is already in there, so that once the server replies to the request, the entry is logged.

src/analyzer/protocol/ldap/ldap_zeek.spicy
awelzel: I like the `%init` route, let me try.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: The failed state means that client/server continue communicate in plaintext, so it's not a failed state. Presumably both sides close down the connection, but there's not real failed state we need to deal with.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: I'm adding the `assert`.

I'm not following the consequences of keeping MessageDispatch anonymous. The `Message` within `MessageDispatch` still generates events properly I believe.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: I meant if the first byte `0x30` heuristic is wrong (it's encrypted instead of us guessing plaintext), then the length field starts with 0x30 and the encrypted size would be between `0x30 00 00 00` and `0x30 FF FF FF` which would be at least 768MB and unlikely large for a LDAP message.

Let me rephrase.

src/analyzer/protocol/ldap/ldap.spicy
awelzel: Hmm, I had adapted, end up introducing an issue liking `%done` better. While we don't use `requestValue`, it would be possible in `%done` to use it.

src/analyzer/protocol/ldap/ldap.spicy
bbannier: > I'm not following the consequences of keeping MessageDispatch anonymous. The Message within MessageDispatch still generates events properly I believe.

Yes, this does not affect events which are triggered via units down the parse tree, but is more about what "interface"  `MaybeEncrypted` exposes; its current shape completely dictated by parsing. Since `MessageDispatch` contains useful data it might make sense to expose it to users, but we currently have none 🤷. Feel free to drop.

PR #3840, opened 2024-07-21
telemetry: Deprecate prometheus.zeek policy script

awelzel:
With Cluster::Node$metrics_port being optional, there's not really a need for the extra script. New rule, if a metrics_port is set, the node will attempt to listen on it.

Users can still redef Telemetry::metrics_port *after* base/frameworks/telemetry was loaded to change the port defined in cluster-layout.zeek.

### DIFF
--- a/scripts/base/frameworks/telemetry/__load__.zeek
+++ b/scripts/base/frameworks/telemetry/__load__.zeek
@@ -1,3 +1,1 @@
 @load ./main
-
-@load base/frameworks/cluster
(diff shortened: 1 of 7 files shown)


awelzel:
@ckreibich / @timwoj - as mentioned in Slack, think something like this would make sense. So for zeekctl one sets MetricsPort in zeekctl.cfg and all nodes start listening automatically. Loading an extra policy script seems unnecessary.

If we fold that into 7.0, we can remove things in 7.1. Otherwise, would re-target for 8.1, though that would be a bit sad.

timwoj (approved):
I agree that we should pull this into 7.0 since that's where all of these changes were introduced anyways.

awelzel:
> I agree that we should pull this into 7.0 since that's where all of these changes were introduced anyways.

Merged into master and pushed the cherry-pick into release/7.0.

Issue #3839, opened 2024-07-20
"non-void function returning without a value" when returning from a `when`-block


ckreibich:
I bumped into this while fiddling with `Broker::exists()`, but the following is a shorter example:
```
event zeek_init() {
        when ( local res = lookup_addr(127.0.0.1) ) {
                return;
        }
}
```
When running this:
```
$ zeek ./test.zeek exit_only_after_terminate=T
warning in <params>, line 1: non-void function returning without a value: lambda_<14219442610740974060>
```
This seems wrong, or at least misleading. When printing the function in question at the time of the warning, via `Describe()`, I get:
```
{
if (1 == when-param-12)
        return (to_any_coerceres = lookup_addr(127.0.0.1));

if (2 == when-param-12)
        {
        return ;
        }

return (to_any_coerceT);
}
```
So I suspect that discrepancy in returns is what's triggering the warning. I noticed that `WhenInfo::Build()` has [a comment](https://github.com/zeek/zeek/blob/master/src/Stmt.cc#L1810-L1823) that says this might be addressed elsewhere, but I'm not sure where that might be, or how it works.

@vpax, you're likely closest to this?

vpax:
Interesting, I was just today bumping up against a similar (not the same) problem with `when`s. I'll take this for next month's script optimization maintenance (it's not ZAM-related, but those changes were made in support of script optimization).

bbannier:
Fixed by e960c29acb9d81da69ced871ab5054c3e2152d3b.

PR #3846, opened 2024-07-24
Script optimization maintenance for August 2024 (early)

vpax:
Due to upcoming PTO, I attended to next month's script optimization early. I took the opportunity to address GH-3839 (a very simple fix), and to add a minor optimization for comparing boolean values to `T` or `F`.

### DIFF
--- a/src/script_opt/Expr.cc
+++ b/src/script_opt/Expr.cc
@@ -1238,10 +1238,28 @@ ExprPtr EqExpr::Reduce(Reducer* c, StmtPtr& red_stmt) {
     if ( IsHasElementsTest() )
         return BuildHasElementsTest()->Reduce(c, red_stmt);
 
-    if ( GetType()->Tag() == TYPE_BOOL && same_singletons(op1, op2) ) {
-        bool t = Tag() == EXPR_EQ;
-        auto res = with_location_of(make_intrusive<ConstExpr>(val_mgr->Bool(t)), this);
-        return res->Reduce(c, red_stmt);
+    if ( GetType()->Tag() == TYPE_BOOL ) {
+        if ( same_singletons(op1, op2) ) {
+            bool t = Tag() == EXPR_EQ;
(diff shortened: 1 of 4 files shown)


testing/btest/language/when.zeek
```diff
@@ -26,6 +26,13 @@ event zeek_init()
 	when [h] ( local hname3 = lookup_addr(h) ) {}
 	timeout to + 2sec {}
 
+	# The following used to generate a spurious warning, so it's here
+	# as a regression test.
+	when ( local res = lookup_addr(127.0.0.1) )
+		{
+		return;
+		}
```
bbannier, line 34: Since the warning would go to stderr it would never make it into the baseline, could you adjust above

```patch
-# @TEST-EXEC: mv test1/.stdout out
+# @TEST-EXEC: cat test1/.{stdout,stderr} >> out
```

src/script_opt/Expr.cc
```diff
@@ -1238,10 +1238,28 @@ ExprPtr EqExpr::Reduce(Reducer* c, StmtPtr& red_stmt) {
...
-        auto res = with_location_of(make_intrusive<ConstExpr>(val_mgr->Bool(t)), this);
-        return res->Reduce(c, red_stmt);
+    if ( GetType()->Tag() == TYPE_BOOL ) {
+        if ( same_singletons(op1, op2) ) {
+            bool t = Tag() == EXPR_EQ;
+            auto res = with_location_of(make_intrusive<ConstExpr>(val_mgr->Bool(t)), this);
+            return res->Reduce(c, red_stmt);
+        }
+
+        if ( op1->GetType()->Tag() == TYPE_BOOL ) {
```
awelzel, line 1248: I've merged as is, but a comment would've been nice. The `t = ! t` part is pretty magic, particularly without context what optimization is being employed

vpax: Thanks. I'll make a note to add that next time around

PR #3848, opened 2024-07-24
Fix handling of zero-length SMB2 error responses

timwoj:
It's possible to receive SMB2 error responses with the byte_length field set to zero. When we tried to parse these though, we still attempted to retrieve the rest of the data out to the end of the packet instead of skipping it. This sets the length of the "stuff" we're supposed to parse to what the packet reports, which allows it to properly skip process those zero-byte PDUs. Previously these PDUs would cause a binpac exception, which caused binpac to drop any further PDUs in the packet on the floor. 

This will likely cause performance decreases on networks that see a lot of these kinds of PDUs. Now that they're being processed correctly, there are extra events being processed and so event drain happens more frequently.

Thanks @JustinAzoff for the find during the 7.0 release candidate testing. This should probably be merged into the 7.0 release branch as well as master, since it fixes a feature that was added during 7.0 (multi-PDU parsing).


NEWS
```diff
@@ -113,7 +113,10 @@ New Functionality
   environment variable configures the addition.
 
 - SMB2 packets containing multiple PDUs now correctly parse all of the headers,
-  instead of just the first one and ignoring the rest.
+  instead of just the first one and ignoring the rest. This has the potential to
+  cause decreased performance on SMB2-heavy networks. Since the remaining PDUs
```
awelzel, line 117: I think I'd reword as "This may increase worker CPU usage on SMB2-heavy networks." and probably skip the parsing and more events part. Seems a bit redundant or too detailed.

awelzel (approved):

PR #3850, opened 2024-07-24
Add contributors to 7.0.0 NEWS entry

ckreibich:
I'm still treating @vpax as a given here, but holler if we should mention you. :-)


timwoj (approved):

timwoj:
Make sure those changes go into master too, to keep it consistent at least for the first release.

PR #3858, opened 2024-07-29
Add packet analyzer history to `unknown_protocols.log`

J-Gras:
Experimenting with `unknown_protocols.log` in production environments, it became clear that more context would be helpful to interpret the log. In particular, the lineage of analyzers that processed a packet before an unknown protocol is encountered would provide valuable context. This changes adds a new analyzer history field to track that information. This change seems rather invasive so I am opening it as a draft looking for feedback.

Unfortunately, some quick benchmarking shows a measurable overhead:
```
$ hyperfine -w 1 "~/zeek-installs/zeek-7.1.0-master/bin/zeek -r 2009-M57-day11-18.pcap" "~/zeek-installs/zeek-7.1.0-analyzer-history/bin/zeek -r 2009-M57-day11-18.pcap"
Benchmark 1: ~/zeek-installs/zeek-7.1.0-master/bin/zeek -r 2009-M57-day11-18.pcap
  Time (mean ± σ):     12.974 s ±  0.066 s    [User: 13.213 s, System: 2.328 s]
  Range (min … max):   12.890 s … 13.122 s    10 runs
 
Benchmark 2: ~/zeek-installs/zeek-7.1.0-analyzer-history/bin/zeek -r 2009-M57-day11-18.pcap
  Time (mean ± σ):     13.381 s ±  0.059 s    [User: 13.614 s, System: 2.326 s]
  Range (min … max):   13.308 s … 13.485 s    10 runs
 
Summary
  ~/zeek-installs/zeek-7.1.0-master/bin/zeek -r 2009-M57-day11-18.pcap ran
    1.03 ± 0.01 times faster than ~/zeek-installs/zeek-7.1.0-analyzer-history/bin/zeek -r 2009-M57-day11-18.pcap
```


awelzel:
> Unfortunately, some quick benchmarking shows a measurable overhead:

How much worse are those numbers for encapsulated or maybe double vlan tagged traffic? I usually test with `-b -C`, this won't change the absolute time, but for raw performance of the lower level packet path increases the relative slow down.

If you fix (or disable) the failing btests, your build would be submitted to the benchmarker where there's the 2009-M57-day11-18.pcap genve+vxlan encapsulated being tested, too, as well as some other pcaps.

---

Have you considered tracking an analyzer stack outside of packets in the manager as `std::vector<Analyzer*>` or so? A `zeek::Tag` I believe is at least 3x times as large as a raw pointer, so that alone is expensive as in the fast path you don't want the Tag. Adding a `std::vector` to `Packet` is also controversial for performance. Further, the first 3 pushes to it would [result in a new malloc/realloc call with glibc](https://github.com/zeek/zeek/pull/3478#discussion_r1415692021).

J-Gras:
Closing this in favor of #3879.

PR #3859, opened 2024-07-29
Process metric callbacks from the main-loop thread

timwoj:
This avoids the callbacks from being processed on the worker thread spawned by Civetweb. It fixes data race issues with lookups involving global variables, amongst other threading issues.

This was pulled out of https://github.com/zeek/zeek/pull/3814. Unfortunately none of the existing metrics triggered the ThreadSanitizer findings from this, but if you look at that PR you can see that the tsan build passes again after this commit.

### DIFF
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -6,6 +6,7 @@
 
 // CivetServer is from the civetweb submodule in prometheus-cpp
 #include <CivetServer.h>
+#include <prometheus/collectable.h>
 #include <prometheus/exposer.h>
 #include <prometheus/registry.h>
 #include <rapidjson/document.h>
@@ -16,19 +17,32 @@
 
 #include "zeek/3rdparty/doctest.h"
 #include "zeek/ID.h"
+#include "zeek/RunState.h"
 #include "zeek/ZeekString.h"
 #include "zeek/broker/Manager.h"
+#include "zeek/iosource/Manager.h"
 #include "zeek/telemetry/ProcessStats.h"
 #include "zeek/telemetry/Timer.h"
+#include "zeek/telemetry/consts.bif.h"
 #include "zeek/telemetry/telemetry.bif.h"
 #include "zeek/threading/formatters/detail/json.h"
 
 namespace zeek::telemetry {
 
-Manager::Manager() { prometheus_registry = std::make_shared<prometheus::Registry>(); }
+/**
+ * Prometheus Collectable interface used to insert Zeek callback processing
+ * before the Prometheus registry's collection of metric data.
+ */
+class ZeekCollectable : public prometheus::Collectable {
+public:
+    std::vector<prometheus::MetricFamily> Collect() const override {
+        telemetry_mgr->WaitForPrometheusCallbacks();
+        return {};
+    }
+};
 
-// This can't be defined as =default because of the use of unique_ptr with a forward-declared type
-// in Manager.h
+Manager::Manager() : IOSource(true) { prometheus_registry = std::make_shared<prometheus::Registry>(); }
+
 Manager::~Manager() {}
 
 void Manager::InitPostScript() {
@@ -75,7 +89,9 @@ void Manager::InitPostScript() {
 
         if ( ! getenv("ZEEKCTL_CHECK_CONFIG") ) {
             try {
-                prometheus_exposer = std::make_unique<prometheus::Exposer>(prometheus_url, 2, callbacks);
+                prometheus_exposer =
+                    std::make_unique<prometheus::Exposer>(prometheus_url, BifConst::Telemetry::civetweb_threads,
+                                                          callbacks);
 
                 // CivetWeb stores a copy of the callbacks, so we're safe to delete the pointer here
                 delete callbacks;
@@ -83,6 +99,13 @@ void Manager::InitPostScript() {
                 reporter->FatalError("Failed to setup Prometheus endpoint: %s. Attempted to bind to %s.", exc.what(),
                                      prometheus_url.c_str());
             }
+
+            // This has to be inserted before the registry below. The exposer
+            // processes the collectors in order of insertion. We want to make
+            // sure that the callbacks get called and the values in the metrics
+            // are updated before prometheus-cpp scrapes them.
+            zeek_collectable = std::make_shared<ZeekCollectable>();
+            prometheus_exposer->RegisterCollectable(zeek_collectable);
 
             prometheus_exposer->RegisterCollectable(prometheus_registry);
         }
@@ -130,6 +153,21 @@ void Manager::InitPostScript() {
                                   return metric;
                               });
 #endif
+
+    iosource_mgr->RegisterFd(collector_flare.FD(), this);
+}
+
+void Manager::Terminate() {
+    // Notify the collector condition so that it doesn't hang waiting for
+    // a collector request to complete.
+    collector_cv.notify_all();
+
+    // Shut down the exposer first of all so we stop getting requests for
+    // data. This keeps us from getting a request on another thread while
+    // we're shutting down.
+    prometheus_exposer.reset();
+
+    iosource_mgr->UnregisterFd(collector_flare.FD(), this);
 }
 
 // -- collect metric stuff -----------------------------------------------------
@@ -543,6 +581,39 @@ HistogramPtr Manager::HistogramInstance(std::string_view prefix, std::string_vie
     auto lbls = Span{labels.begin(), labels.size()};
     auto bounds_span = Span{bounds.begin(), bounds.size()};
     return HistogramInstance(prefix, name, lbls, bounds_span, helptext, unit);
+}
+
+void Manager::ProcessFd(int fd, int flags) {
+    std::unique_lock<std::mutex> lk(collector_cv_mtx);
+
+    collector_flare.Extinguish();
+
+    prometheus_registry->UpdateViaCallbacks();
+    collector_response_idx = collector_request_idx;
+
+    lk.unlock();
+    collector_cv.notify_all();
+}
+
+void Manager::WaitForPrometheusCallbacks() {
+    std::unique_lock<std::mutex> lk(collector_cv_mtx);
+
+    ++collector_request_idx;
+    uint64_t expected_idx = collector_request_idx;
+    collector_flare.Fire();
+
+    // It should *not* take 5 seconds to go through all of the callbacks, but
+    // set this to have a timeout anyways just to avoid a deadlock.
+    bool res = collector_cv.wait_for(lk,
+                                     std::chrono::microseconds(
+                                         static_cast<long>(BifConst::Telemetry::callback_timeout * 1000000)),
+                                     [expected_idx]() {
+                                         return telemetry_mgr->collector_response_idx >= expected_idx ||
+                                                zeek::run_state::terminating;
+                                     });
+
+    if ( ! res )
+        fprintf(stderr, "Timeout waiting for prometheus callbacks\n");
 }
 
 } // namespace zeek::telemetry
--- a/src/telemetry/Manager.h
+++ b/src/telemetry/Manager.h
@@ -9,8 +9,10 @@
 #include <string_view>
 #include <vector>
 
+#include "zeek/Flare.h"
 #include "zeek/IntrusivePtr.h"
 #include "zeek/Span.h"
+#include "zeek/iosource/IOSource.h"
 #include "zeek/telemetry/Counter.h"
 #include "zeek/telemetry/Gauge.h"
 #include "zeek/telemetry/Histogram.h"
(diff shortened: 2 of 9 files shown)


timwoj:
Apparently tsan is throwing errors for other reasons that didn't show up on the other PR. I'll look into those.

src/telemetry/Manager.cc
```diff
@@ -545,6 +575,25 @@ HistogramPtr Manager::HistogramInstance(std::string_view prefix, std::string_vie
     return HistogramInstance(prefix, name, lbls, bounds_span, helptext, unit);
 }
 
+void Manager::Process() {
+    std::unique_lock<std::mutex> lk(collector_cv_mtx);
+
+    collector_flare->Extinguish();
+    prometheus_registry->UpdateViaCallbacks();
```
awelzel, line 582: This is where I was wondering if we could let go of the custom patching of prometheus-cpp:

If we kept track of callbacks in the telemetry manager one level up, could we execute them here directly?

src/telemetry/Manager.cc
```diff
@@ -545,6 +575,25 @@ HistogramPtr Manager::HistogramInstance(std::string_view prefix, std::string_vie
...
+    collector_cv.notify_all();
+}
+
+void Manager::WaitForPrometheusCallbacks() {
+    std::unique_lock<std::mutex> lk(collector_cv_mtx);
+    collector_flare->Fire();
+
+    // It should *not* take 5 seconds to go through all of the callbacks, but
+    // set this to have a timeout anyways just to avoid a deadlock.
+    collector_cv.wait_for(lk, std::chrono::seconds(5));
```
awelzel, line 594: Should this handle a timeout return value and print to stderr if there is one? A BifConst variable would be nice for testing really small timeouts to try and force a timeout.

For the latter point, `Manager::Process()` holds the lock while executing callbacks - if that actually takes too long, are we in danger of running into:

> Right after wait_for returns, lock.owns_lock() is true, and lock.mutex() is locked by the calling thread. If these postconditions cannot be satisfied[[1]](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for#cite_note-1), calls [std::terminate](https://en.cppreference.com/w/cpp/error/terminate).

There's also no predicate we're looping on, so we may run into spurious wakeups, too. Maybe this needs some counter/queue to understand if callback processing has completed? This should also help avoiding to hold the lock in `Manager::Process()`: Only hold it to increase a counter or send something over the queue.

src/telemetry/Manager.h
```diff
@@ -250,6 +267,11 @@ class Manager final {
     std::unique_ptr<prometheus::Exposer> prometheus_exposer;
 
     std::string cluster_json;
+
+    std::shared_ptr<ZeekCollectable> zeek_collectable;
+    std::unique_ptr<zeek::detail::Flare> collector_flare;
```
awelzel, line 272: No strong opinion, but flare could probably be embedded, too.

src/telemetry/Manager.cc
```diff
@@ -16,20 +17,40 @@
...
+}
 
-// This can't be defined as =default because of the use of unique_ptr with a forward-declared type
-// in Manager.h
-Manager::~Manager() {}
+Manager::~Manager() {
+    // Shut down the exposer first of all so we stop getting requests for
+    // data. This keeps us from getting a request on another thread while
+    // we're shutting down.
+    prometheus_exposer.reset();
```
awelzel, line 52: Suspect if currently a request is processed, doing this is still unsafe, but don't have an immediate answer.

awelzel (commented):
I like it, but clearly it's also hairy :-) Added some comments.

src/telemetry/Manager.cc
timwoj: We can't, because for things like counters there isn't any way to directly set the value of the instrument. I can for gauges, but not for counters.

src/telemetry/Manager.cc
timwoj: I dug down into the civetweb code and their `mg_stop` method correctly handles this case. Since deleting the exposer is effectively just deleting the civetweb handler, it should be fine.

src/telemetry/Manager.h
timwoj: I left it here mostly because of how easy our `Flare` type is to use.

src/telemetry/Manager.cc
timwoj: See if that matches what you were thinking. The idea of returning early if the flag is set is that we shouldn't bother to run the callbacks a second time if they're already running. Just let the second request wait until it's done and return the same data.

src/telemetry/Manager.cc
awelzel: Ah, but if we were to produce `prometheus::MetricFamily` instances from callbacks (maybe indirectly), then we maybe could get along without needing to patch? E.g., the callback based approach wouldn't set Counter/Gauge instances and instead yield the actual data directly (or append to a vector passed by reference).

Nothing for now, but it seems prometheus-cpp actually has enough API to implement callbacks through the `Collectable` without needing to patch.

src/telemetry/Manager.cc
awelzel: Hmm, that's not it.

The `ProcessFd()` runs on in the main-thread through the IO loop sequentially, so the early return should never be hit and is confusing.

The current change doesn't actually wait for the callbacks to complete. Sprinkling a few fprintfs, you get output which shows the callbacks run *after* the thread has fallen through the condition variable:

```
diff --git a/src/telemetry/Manager.cc b/src/telemetry/Manager.cc
index b370db8758..429bc04d37 100644
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -586,22 +586,32 @@ void Manager::ProcessFd(int fd, int flags) {
 
     collector_flag = true;
     collector_flare->Extinguish();
+    std::fprintf(stderr, "running callbacks\n");
     prometheus_registry->UpdateViaCallbacks();
 
     collector_flag = false;
     lk.unlock();
+    std::fprintf(stderr, "notify!\n\n");
     collector_cv.notify_all();
 }
 
 void Manager::WaitForPrometheusCallbacks() {
+    std::fprintf(stderr, "wait for prometheus callbacks\n");
     std::unique_lock<std::mutex> lk(collector_cv_mtx);
     collector_flare->Fire();
 
     // It should *not* take 5 seconds to go through all of the callbacks, but
     // set this to have a timeout anyways just to avoid a deadlock.
-    bool res = collector_cv.wait_for(lk, std::chrono::seconds(5), []() { return ! telemetry_mgr->collector_flag; });
+    bool res = collector_cv.wait_for(lk, std::chrono::seconds(5), []() {
+        bool p = ! telemetry_mgr->collector_flag;
+        std::fprintf(stderr, "pred %d\n", p);
+        return p;
+    });
+
     if ( ! res )
         fprintf(stderr, "Timeout waiting for prometheus callbacks\n");
+
+    std::fprintf(stderr, "wait for prometheus callbacks done\n");
 }
 
 } // namespace zeek::telemetry
```

```
# zeek -C -i lo Telemetry::metrics_port=4222/tcp  Log::default_writer=Log::WRITER_NONE
<params>, line 1: listening on lo

wait for prometheus callbacks
pred 1
wait for prometheus callbacks done
running callbacks
notify!

wait for prometheus callbacks
pred 1
wait for prometheus callbacks done
running callbacks
notify!

wait for prometheus callbacks
pred 1
wait for prometheus callbacks done
running callbacks
notify!

```

awelzel: This is the local patch that seems to work alright with lots of printf'ing.
```
diff --git a/src/telemetry/Manager.cc b/src/telemetry/Manager.cc
index b370db8758..8f7213431b 100644
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -156,11 +156,21 @@ void Manager::InitPostScript() {
 }
 
 void Manager::Terminate() {
+    {
+        // Notify threads that we're shutting down.
+        //
+        // Don't think we really need the lock here?
+        std::unique_lock<std::mutex> lk(collector_cv_mtx);
+        std::fprintf(stderr, "wake em up term=%d\n", zeek::run_state::terminating);
+        collector_cv.notify_all();
+    }
     // Shut down the exposer first of all so we stop getting requests for
     // data. This keeps us from getting a request on another thread while
     // we're shutting down.
+    std::fprintf(stderr, "reset exposer\n");
     prometheus_exposer.reset();
 
+    std::fprintf(stderr, "unregister\n");
     iosource_mgr->UnregisterFd(collector_flare->FD(), this);
     collector_flare.reset();
 }
@@ -579,29 +589,37 @@ HistogramPtr Manager::HistogramInstance(std::string_view prefix, std::string_vie
 }
 
 void Manager::ProcessFd(int fd, int flags) {
-    if ( collector_flag )
-        return;
-
     std::unique_lock<std::mutex> lk(collector_cv_mtx);
-
-    collector_flag = true;
     collector_flare->Extinguish();
+
+    std::fprintf(stderr, "running callbacks\n");
     prometheus_registry->UpdateViaCallbacks();
+    response_idx = request_idx; // Let pending threads know we've completed the update.
+    std::fprintf(stderr, "running callbacks updated to %lu\n", request_idx);
 
-    collector_flag = false;
     lk.unlock();
+    std::fprintf(stderr, "notify!\n");
     collector_cv.notify_all();
 }
 
 void Manager::WaitForPrometheusCallbacks() {
     std::unique_lock<std::mutex> lk(collector_cv_mtx);
+    ++request_idx;
+    uint64_t expected_idx = request_idx;
+    std::fprintf(stderr, "wait for prometheus callbacks %lu\n", expected_idx);
     collector_flare->Fire();
 
     // It should *not* take 5 seconds to go through all of the callbacks, but
     // set this to have a timeout anyways just to avoid a deadlock.
-    bool res = collector_cv.wait_for(lk, std::chrono::seconds(5), []() { return ! telemetry_mgr->collector_flag; });
+    bool res = collector_cv.wait_for(lk, std::chrono::seconds(5), [expected_idx]() {
+        bool p = telemetry_mgr->response_idx >= expected_idx || zeek::run_state::terminating;
+        std::fprintf(stderr, "predicate %d for %lu!\n", p, expected_idx);
+        return p;
+    });
     if ( ! res )
         fprintf(stderr, "Timeout waiting for prometheus callbacks\n");
+
+    std::fprintf(stderr, "wait for prometheus callbacks done idx=%lu\n", expected_idx);
 }
 
 } // namespace zeek::telemetry
diff --git a/src/telemetry/Manager.h b/src/telemetry/Manager.h
index 0627971f0e..3983416e3d 100644
--- a/src/telemetry/Manager.h
+++ b/src/telemetry/Manager.h
@@ -275,7 +275,9 @@ private:
     std::unique_ptr<zeek::detail::Flare> collector_flare;
     std::condition_variable collector_cv;
     std::mutex collector_cv_mtx;
-    bool collector_flag = false;
+    // Only modified under collector_cv_mtx!
+    uint64_t request_idx = 0;
+    uint64_t response_idx = 0;
 };
 
 } // namespace zeek::telemetry

```

awelzel: That commit would add a few BifConst values. Not expecting anyone to change them in, but they should make testing easier for us if there's something going wrong:

https://github.com/zeek/zeek/commit/a7d0d52071114a613672ec37c54718afbe003b6a

timwoj: Ah, ok that makes more sense. I integrated that. Thanks for the BifConst commit. I cherry-pbicked that in too.

src/telemetry/Manager.cc
```diff
@@ -160,6 +155,16 @@ void Manager::InitPostScript() {
     iosource_mgr->RegisterFd(collector_flare->FD(), this);
 }
 
+void Manager::Terminate() {
+    // Shut down the exposer first of all so we stop getting requests for
+    // data. This keeps us from getting a request on another thread while
+    // we're shutting down.
```
awelzel, line 163: I believe this should have the `collector_cv.notify_all()` call here, otherwise I saw a 5 second shutdown hang once in a while when running a `while true;  do curl <metrics endpoint>; done`. It terminated cleanly, but essentially the threads were waiting for the 5 seconds timeout, just to find that the predicate is now fulfilled as Zeek is terminating.

It's basically the `.reset()` and `destructor` waiting for all requests to finish.

src/telemetry/Manager.cc
```diff
@@ -545,6 +579,40 @@ HistogramPtr Manager::HistogramInstance(std::string_view prefix, std::string_vie
...
+    uint64_t expected_idx = collector_request_idx;
+    collector_flare.Fire();
+
+    // It should *not* take 5 seconds to go through all of the callbacks, but
+    // set this to have a timeout anyways just to avoid a deadlock.
+    bool res = collector_cv.wait_for(lk,
+                                     std::chrono::microseconds(
+                                         static_cast<long>(BifConst::Telemetry::callback_timeout * 1000000)),
+                                     [expected_idx]() {
+                                         bool p = telemetry_mgr->collector_response_idx >= expected_idx ||
```
awelzel, line 607: Your version without the extra `p` looked better :-)

awelzel (approved):
One more comment about `Terminate()`, but otherwise :ship:

PR #3867, opened 2024-08-06
Bump Spicy to a version with much less inlining

bbannier:


bbannier:
Not inline the code seems to to degrade throughput of Spicy analyzer by 15-30%.

PR #3868, opened 2024-08-06
Threading Manager Metrics Follow-up

awelzel:
@timwoj -  there were a few rough edges when testing the threading metrics specifically locally.

Mind checking the commits and messages to see if this makes sense to you?

### DIFF
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -162,7 +162,9 @@ void Manager::InitPostScript() {
                               });
 #endif
 
-    iosource_mgr->RegisterFd(collector_flare.FD(), this);
+    if ( ! iosource_mgr->RegisterFd(collector_flare.FD(), this) ) {
+        reporter->FatalError("Failed to register telemetry collector descriptor");
+    }
 }
 
 void Manager::Terminate() {
--- a/src/threading/Manager.cc
+++ b/src/threading/Manager.cc
@@ -56,8 +56,6 @@ void Manager::InitPostScript() {
                 for ( const auto& t : thread_mgr->msg_threads ) {
                     t->GetStats(&thread_stats);
 
-                    thread_mgr->current_bucketed_messages.sent_in_total += thread_stats.sent_in;
-                    thread_mgr->current_bucketed_messages.sent_out_total += thread_stats.sent_out;
                     thread_mgr->current_bucketed_messages.pending_in_total += thread_stats.pending_in;
                     thread_mgr->current_bucketed_messages.pending_out_total += thread_stats.pending_out;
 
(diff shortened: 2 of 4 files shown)


timwoj:
> This might be strange initially, but aligns with the Prometheus histogram approach (though we're using gauges here).

I'm not sure this is true. The Prometheus docs do say this:

> The reason is that the histogram buckets are [cumulative](https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram). The le="0.3" bucket is also contained in the le="1.2" bucket

Except that contradicts both:

1) how prometheus-cpp does it: https://github.com/jupp0r/prometheus-cpp/blob/6492e820cdff7e7345d46d82b43735aaea542098/core/src/histogram.cc#L40-L49
2) how we do it in our test output (which is now based on the prometheus-cpp methods, but hasn't changed from how Broker did it): https://github.com/zeek/zeek/blob/c845c7cce01255b412cc4e6dbf5ece83c0c6c4a4/testing/btest/Baseline/scripts.base.frameworks.telemetry.basic/out#L23-L24

It's certainly possible that prometheus-cpp is wrong here and I should file an issue there to have it fixed, but traditionally the buckets haven't been cumulative in our output.

awelzel:
> > This might be strange initially, but aligns with the Prometheus histogram approach (though we're using gauges here).
> 
> I'm not sure this is true. The Prometheus docs do say this:
> 
> > The reason is that the histogram buckets are [cumulative](https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram). The le="0.3" bucket is also contained in the le="1.2" bucket
> 
> Except that contradicts both:
> 
> 1. how prometheus-cpp does it: https://github.com/jupp0r/prometheus-cpp/blob/6492e820cdff7e7345d46d82b43735aaea542098/core/src/histogram.cc#L40-L49

During exposition in Collect() it is computing the cumulative value for the individual ClientMetrics. I suspect broker did the same on Prometheus export:

https://github.com/jupp0r/prometheus-cpp/blob/6492e820cdff7e7345d46d82b43735aaea542098/core/src/histogram.cc#L85

> 2. how we do it in our test output (which is now based on the prometheus-cpp methods, but hasn't changed from how Broker did it): https://github.com/zeek/zeek/blob/c845c7cce01255b412cc4e6dbf5ece83c0c6c4a4/testing/btest/Baseline/scripts.base.frameworks.telemetry.basic/out#L23-L24
> 
> It's certainly possible that prometheus-cpp is wrong here and I should file an issue there to have it fixed, but traditionally the buckets haven't been cumulative in our output.

For buckets of a real histogram, agree. But here we're essentially populating the individual gauges / ClientMetrics of the exposition.

timwoj (approved):

PR #3870, opened 2024-08-06
broker: Deprecate MakeEvent(ValPList*)

awelzel:
The variadic broker messaging BIFs currently convert `@ARGS@` into a ValPList before passing it on to MakeEvent(). This appears historic plumbing. Implement the same functionality using iterators and do the extra copying in the now deprecated MakeEvent(ValPList*).

Further, make passing a frame optional as not all callers may have one available.

---

@Neverlord / @timwoj - I've been staring a lot at some of the variadic BiFs used by broker and wanted to see if you have opinions about moving to a model where the BiF arguments are passed as two iterators rather than the ValPList (which seems somewhat historic).

Basically, replace:
```
RecordVal* MakeEvent(
        ValPList* args, zeek::detail::Frame* frame);
```
with
```
zeek::RecordValPtr MakeEvent(ArgsIt first, ArgsIt last, zeek::detail::Frame* frame);
```

where `ArgsIt` is std::vector<ValPtr>::const_iterator. As `@ARGS@` in BiFs is a `std::vector<ValPtr>`. Switching to iterators allows slicing without copying (though maybe there's a better pattern than that?)

### DIFF
--- a/src/Span.h
+++ b/src/Span.h
@@ -2,7 +2,6 @@
 
 #pragma once
 
-#include <array>
 #include <cstddef>
 #include <iterator>
 #include <type_traits>
@@ -127,9 +126,6 @@ private:
 
 template<class T>
 Span(T*, size_t) -> Span<T>;
-
-template<class Iter>
-Span(Iter, Iter) -> Span<typename std::iterator_traits<Iter>::value_type>;
 
 template<class T, size_t N>
 Span(T (&)[N]) -> Span<T>;
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -853,16 +853,29 @@ bool Manager::AutoUnpublishEvent(const string& topic, Val* event) {
 }
 
 RecordVal* Manager::MakeEvent(ValPList* args, zeek::detail::Frame* frame) {
-    auto rval = new RecordVal(BifType::Record::Broker::Event);
+    // Deprecated MakeEvent() version using ValPList - requires extra copy.
+    zeek::Args cargs;
+    cargs.reserve(args->size());
+    for ( auto* a : *args )
+        cargs.push_back({zeek::NewRef{}, a});
+
+    return MakeEvent(ArgsSpan{cargs}, frame)->Ref()->AsRecordVal();
+}
+
+zeek::RecordValPtr Manager::MakeEvent(ArgsSpan args, zeek::detail::Frame* frame) {
+    scoped_reporter_location srl{frame};
+    return MakeEvent(args);
+}
+
+zeek::RecordValPtr Manager::MakeEvent(ArgsSpan args) {
+    auto rval = zeek::make_intrusive<RecordVal>(BifType::Record::Broker::Event);
     auto arg_vec = make_intrusive<VectorVal>(vector_of_data_type);
     rval->Assign(1, arg_vec);
-    Func* func = nullptr;
-    scoped_reporter_location srl{frame};
+    const Func* func = nullptr;
 
-    for ( auto i = 0; i < args->length(); ++i ) {
-        auto arg_val = (*args)[i];
-
-        if ( i == 0 ) {
+    for ( size_t index = 0; index < args.size(); index++ ) {
+        const auto& arg_val = args[index];
+        if ( index == 0 ) {
             // Event val must come first.
 
             if ( arg_val->GetType()->Tag() != TYPE_FUNC ) {
@@ -877,10 +890,10 @@ RecordVal* Manager::MakeEvent(ValPList* args, zeek::detail::Frame* frame) {
                 return rval;
             }
 
-            auto num_args = func->GetType()->Params()->NumFields();
+            auto num_args = static_cast<size_t>(func->GetType()->Params()->NumFields());
 
-            if ( num_args != args->length() - 1 ) {
-                Error("bad # of arguments: got %d, expect %d", args->length(), num_args + 1);
+            if ( num_args != args.size() - 1 ) {
+                Error("bad # of arguments: got %zu, expect %zu", args.size() - 1, num_args + 1);
                 return rval;
             }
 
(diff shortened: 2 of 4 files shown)


src/broker/Manager.cc
```diff
@@ -813,14 +813,32 @@ bool Manager::AutoUnpublishEvent(const string& topic, Val* event) {
...
     auto arg_vec = make_intrusive<VectorVal>(vector_of_data_type);
     rval->Assign(1, arg_vec);
     Func* func = nullptr;
-    scoped_reporter_location srl{frame};
 
-    for ( auto i = 0; i < args->length(); ++i ) {
-        auto arg_val = (*args)[i];
+    auto it = first;
+
+    int argc = std::distance(first, last);
```
Neverlord, line 838: This is a narrowing bug on 64-bit platforms (`distance` returns `ptrdiff_t`, not `int`). Simply use `auto` instead:

```suggestion
    auto argc = std::distance(first, last);
```

src/broker/Manager.cc
```diff
@@ -813,14 +813,32 @@ bool Manager::AutoUnpublishEvent(const string& topic, Val* event) {
...
+
+    for ( auto i = 0; it != last; ++i, ++it ) {
+        const auto& arg_val = *it;
```
Neverlord, line 841: Why bother about incrementing the iterator here? We already have the index.

```suggestion
    for ( ptrdiff_t index = 0; index < argc; ++index ) {
        const auto& arg_val = first[index];
```

Nitpicks:
- since we use the `for`-loop variable as an index, I'd name it as such
- `std::distance` returns `ptrdiff_t`, so that's we should use here as well (or `decltype(argc)` if you rather infer it)
- `it` can be dropped entirely

src/broker/Manager.cc
```diff
@@ -839,16 +857,16 @@ RecordVal* Manager::MakeEvent(ValPList* args, zeek::detail::Frame* frame) {
...
+                Error("bad # of arguments: got %d, expect %d", argc - 1, num_args + 1);
                 return rval;
             }
 
             rval->Assign(0, func->Name());
             continue;
         }
 
-        const auto& got_type = (*args)[i]->GetType();
+        const auto& got_type = (*it)->GetType();
```
Neverlord, line 869: We already have an alias for that:

```suggestion
        const auto& got_type = arg_val->GetType();
```

src/broker/Manager.cc
```diff
@@ -861,9 +879,9 @@ RecordVal* Manager::MakeEvent(ValPList* args, zeek::detail::Frame* frame) {
         RecordValPtr data_val;
 
         if ( same_type(got_type, detail::DataVal::ScriptDataType()) )
-            data_val = {NewRef{}, (*args)[i]->AsRecordVal()};
+            data_val = {NewRef{}, (*it)->AsRecordVal()};
```
Neverlord, line 882:
```suggestion
            data_val = {NewRef{}, arg_val->AsRecordVal()};
```

src/broker/Manager.cc
```diff
@@ -861,9 +879,9 @@ RecordVal* Manager::MakeEvent(ValPList* args, zeek::detail::Frame* frame) {
...
         else
-            data_val = BrokerData::ToRecordVal((*args)[i]);
+            data_val = BrokerData::ToRecordVal(*it);
```
Neverlord, line 884:
```suggestion
            data_val = BrokerData::ToRecordVal(arg_val);
```

src/broker/messaging.bif
```diff
@@ -208,13 +204,10 @@ function Cluster::publish_rr%(pool: Pool, key: string, ...%): bool
...
-	const auto& bif_args = @ARGS@;
-	ValPList args(bif_args->size() - 2);
+	auto first = @ARGS@->begin();
+	std::advance(first, 2);
 
-	for ( auto i = 2u; i < bif_args->size(); ++i )
-		args.push_back((*bif_args)[i].get());
-
-	auto rval = publish_event_args(args, topic->AsString(), frame);
+	auto rval = publish_event_args(first, @ARGS@->end(), topic->AsString(), frame);
```
Neverlord, line 210: This can be done in one shot:


```c++
auto rval = publish_event_args(std::next(@ARGS@->begin(), 2), @ARGS@->end(), topic->AsString(), frame);
```

src/broker/messaging.bif
```diff
@@ -251,12 +244,9 @@ function Cluster::publish_hrw%(pool: Pool, key: any, ...%): bool
...
-	const auto& bif_args = @ARGS@;
-	ValPList args(bif_args->size() - 2);
-
-	for ( auto i = 2u; i < bif_args->size(); ++i )
-		args.push_back((*bif_args)[i].get());
+	auto first = @ARGS@->begin();
+	std::advance(first, 2);
 
-	auto rval = publish_event_args(args, topic->AsString(), frame);
+	auto rval = publish_event_args(first, @ARGS@->end(), topic->AsString(), frame);
```
Neverlord, line 250: Could also use `std::next` instead, since `first` isn't used otherwise.

Neverlord (commented):
Looks like a good improvement to me. 👍
Just a few comments on the C++ side.

awelzel:
> Looks like a good improvement to me. 👍

Thanks @Neverlord !

After some other exploration, tempted to switch to a `Span<ValPtr>` instead of `std::vector<ValPtr>::const_iterator`: IIUC, this could be interesting to avoid allocation of a `std::vector` totally in certain cases.

src/Span.h
```diff
@@ -63,6 +63,11 @@ class Span {
              class = std::enable_if_t<std::is_convertible_v<const Data*, T*>>>
     Span(const Container& xs) noexcept : memory_block(xs.data()), num_elements(xs.size()) {}
 
+    // This only works with contiguous iterators like those from std::vector or std::string!
+    template<class Iter, class Data = typename Iter::value_type,
+             class = std::enable_if_t<std::is_convertible_v<const Data*, T*>>>
+    Span(Iter first, Iter last) noexcept : memory_block(&first[0]), num_elements(last - first) {}
+
```
awelzel, line 70: @Neverlord - I suspect this isn't exactly safe or great. If you have some enable_if magic to check for `Iter == std::vector::const_iterator` or some such, that might prevent some trouble.

Can also change the users to use the pointers, but that feels bit clunkier than the iterators.

Neverlord: We already have to overload above that takes a container and queries `data()` and `size()`. Why aren't we just using that? I suppose it's because `@ARGS@` in a BIF only has iterator access? But why not extend the interface so that we can do something like `@ARGS@->asSpan()`? I think that would avoid the iterator clunkiness.

Neverlord: If it's just a vector, why not `ArgsSpan{*@ARGS@}`?

Neverlord: Maybe to clarify, I don't suggest to not use iterators at all, but I would cut this constructor and refactor code places like this:

```
auto ev = zeek::broker_mgr->MakeEvent({@ARGS@->begin(), @ARGS@->end()});
```

to simply

```
auto ev = zeek::broker_mgr->MakeEvent({*@ARGS@->begin()});
```

Neverlord (approved):
👍

PR #3871, opened 2024-08-06
7.1 deprecation removals

timwoj:


src/rule-parse.y
```diff
@@ -199,21 +199,12 @@ rule_attr:
 				{
 				const char *msg = id_to_str($2);
 				if ( ! zeek::util::streq(msg, "<error>") )
-					zeek::reporter->Deprecation(zeek::util::fmt("Remove in v7.1: Using an identifier for msg is deprecated (%s:%d)",
-					                                            current_rule_file, rules_line_number+1));
+					zeek::reporter->Error("Idenfitiers for messages cannot be used for rule events (%s:%d)",
+					                      current_rule_file, rules_line_number+1);
```
timwoj, line 203: I'm not sure about the wording here, if anyone has suggestions.

scripts/policy/tuning/__load__.zeek
```diff
@@ -1,2 +0,0 @@
-##! This loads the default tuning
-@load ./defaults
```
timwoj, line 2: I left this empty load script behind but it could easily be removed as well. We didn't deprecate loading the package itself, so removing the script would cause errors.

src/rule-parse.y
awelzel: We only support event identifiers, so would replace  the whole `else` of `if ( is_event($2) )` with:

```
Error("Identifier %s not an event (%s:%d)", $2, ...);
```

... and not doing the AddAction() anymore below.

src/threading/formatters/JSON.h
```diff
@@ -4,12 +4,6 @@
 
 #include <memory>
 
-#define RAPIDJSON_HAS_STDSTRING 1
-// Remove in v7.1 when removing NullDoubleWriter below and also remove
-// rapidjson include tweaks from  CMake's dynamic_plugin_base target.
```
awelzel, line 9: We should now be able to drop all the following ones: Plugins should not need to know about rapidjson anymore.

https://github.com/zeek/zeek/blob/97fa7cdc0a49869ee6605fac9cfc15f11d8c855b/CMakeLists.txt#L343

https://github.com/zeek/zeek/blob/97fa7cdc0a49869ee6605fac9cfc15f11d8c855b/CMakeLists.txt#L351-L354

https://github.com/zeek/zeek/blob/97fa7cdc0a49869ee6605fac9cfc15f11d8c855b/CMakeLists.txt#L1013-L1014

This should make make rapidjson an implementation detail and plugins won't be exposed to it anymore.

src/OpaqueVal.h
```diff
@@ -132,28 +119,11 @@ class OpaqueVal : public Val {
...
     ~OpaqueVal() override = default;
 
-    /**
-     * Serializes the value into a Broker representation.
-     *
-     * @return the broker representation, or an error if serialization
-     * isn't supported or failed.
-     */
-    [[deprecated("Remove in v7.1: use SerializeData instead")]] broker::expected<broker::data> Serialize() const;
-
```
awelzel, line 142: Think we should call this one out again in the NEWS. There's already a NEWS entry for when `DoSerialize()` and `DoUnserialize() was deprecated, but the caveat was that plugin authors won't get compiler warnings.

With these now removed, plugins with OpaqueVal implementations will just fail to compile, so might be good to pointing that out again.

testing/btest/Baseline/coverage.bare-mode-errors/errors
```diff
@@ -1,15 +1,2 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 ### NOTE: This file has been sorted with diff-sort.
-warning in <...>/__load__.zeek, line 1: deprecated script loaded from <...>/__load__.zeek:2 "Remove in v7.1 The policy<...>/defaults package is deprecated. The options set here are now the defaults for Zeek in general.";
```
awelzel, line 3: The corresponding test has a few `grep -v` lines for other deprecated scripts. Same for `test-all-policy-cluster`.

I've added them, but I start to like the approach of including the warnings, too, will do next time!

awelzel (approved):
Some comments here and there. Lots of stuff gone, nice :-)

src/rule-parse.y
timwoj: Removing the check for `<error>` results in a weird error message in one of the tests, so I left that in.

timwoj: This is the result of removing all of it (from signatures.custom-event-errors):

```
   error: Error in signature (./id.sig:19): unknown script-level identifier (non_existing_event)
  error: Identifier <error> is not an event (./id.sig:19)
```

PR #3872, opened 2024-08-07
Extensive ZAM regularization/simplification/extension

vpax:
This large PR updates ZAM optimization with a number of changes aimed at (1) improving and reguliarizing its templating language, including catching errors in operation templates, and (2) providing more flexibility for future broader leveraging of ZAM optimization. The changes are split into thematic commits, though there is some overlap between themes due to the difficulty of untangling related code, and there are some fixups and other messiness (commit-wise, not in the end result) due to needing to merge this long-running branch with changes over the past two months.

These changes require a companion set of changes to auxil/gen-zam to be merged at the same time.

The main elements:

- ZAM instruction templates can now be typed per-operand using "op-types", which then automatically expands $-parameters to have the correct accessors, and in more cases automatically generates memory management for assignments that need it. The typing also expands what used to be treated generically as "integer constant", dividing those into their specific uses (e.g., branch target, global or loop index), with enforcement of compliance with the use. These changes enable removing the "internal-binary-op", "op-accessor", "op1-accessor", and "op2-accessor" template specifiers.

- The above change also facilitates ZAM now not requiring any special access to private ZVal methods or members.

- The template keyword "type" has been changed to "class" (to avoid confusion with "op-types"), and extended so that "classes" can be used to list multiple operand classes to iterate over, whereas previously these would require separate templates.

- The new template keyword "predicate" specifies a template for an expression that can either yield a value for assignment, or a test (or a negated test) for a conditional branch. This allows collapsing what used to be multiple templates into a single template.

- The "eval-pre" specifier has been replaced by "precheck" and "precheck-action" specifiers.

- ZAM instructions now track their associated control flow information. An internal option, set via the environment variable $ZEEK_NO_ZAM_CONTROL_FLOW_OPT, turns off optimization of control flow.

- The new maintenance option "-O validate-ZAM" performs a set of consistency checks on the evaluation code generated by gen-zam. There's an associated BTest to make sure this runs when doing "-a zam" BTests.

- Various methods and functions have been added or made public to enable other script optimization code to leverage elements of ZAM optimization and execution.

- AST profiling now includes ZAM instructions and more robustly tracks identifiers and types. Such profiling also now only optionally computes hash values, so that re-profiling can retain original hash values if desired.

- Identifiers associated with parameters now track their associated declaration record, enabling ZAM profiling to robustly identify them instead of relying on (buggy) heuristics.

- The generic AST traversal framework now ensures the safety of traversing recursive types. Previously this had to be done separately by each traverser.

- ZAM now compiles "type" switches in a manner to leverage "value" switches, rather than using separate if-then logic.

- ZAM instruction templates are now factored into ~16 files rather than a single 3K-LOC file.

- The logic for skipping functions that don't require script optimization because they've been fully inlined has been made more robust.

- Minor tweaks:
	- Added a type-checking fix for vector-of-string operations
	- Added a ZVal constructor for booleans
	- Added a method to clear ListVal's to enable their reuse
	- Provided a smart-pointer accessor for a FileVal's value
	- Fixed a nit in base/protocols/krb/main.zeek
	- Fixed a potential inadvertent interpretor exception in CallExpr::IsPure()
	- Fixed issues with script optimization of coerce-to-any expressions
	- Fixed tracking of whether a capture requires deep-copying
	- Fixed "-O report-C++", which requires generation of profiles
	- Scripts compiled to C++ can now directly generate run-time warnings
	- Provided CPPCompile with access to Type internals for more efficient code generation

I will put in the associated `gen-ZAM` PR right after this one.


vpax:
Companion `gen-zam` PR: https://github.com/zeek/gen-zam/pull/14

I'm not sure what's up with the docs generation check faililng ...

src/ID.h
```diff
@@ -95,6 +94,11 @@ class ID final : public Obj, public notifier::detail::Modifiable {
 
     void ClearVal();
 
+    // *If* this identifier is a function parameter, return the record
+    // associated with its declaration, otherwise nil.
+    RecordTypePtr GetParamRecord() const { return param_record; }
+    void SetParamRecord(const RecordTypePtr& pr) { param_record = pr; }
```
awelzel, line 100: This is quirky, but I think I understand where it comes from (the now removed comment about renaming was useful, maybe that could be added back here?

After long staring:

>             if ( profiled_func_t && id->GetParamRecord() == profiled_func_t->Params() )

Is it too naive to do the following:
In `ProfileFunc::ProfileFunc`, go via `func->GetScope()->OrderedVars()` and take the first `func->GetType()-Params().NumFields()` from it. Shouldn't those be the IDs representing parameters? If `id` is in those, it's a function parameter.


If that's no dice, I think would vote for attaching the full `FuncPtr` to the ID rather than just the params. Just so that you could have more context if needed.

src/Type.cc
```diff
@@ -603,6 +614,9 @@ SetType::~SetType() = default;
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 FuncType::Capture::Capture(detail::IDPtr _id, bool _deep_copy) : id(std::move(_id)), deep_copy(_deep_copy) {
     is_managed = id ? ZVal::IsManagedType(id->GetType()) : false;
+    if ( ! is_managed )
+        // For non-managed types, deep copying isn't applicable.
+        deep_copy = false;
```
awelzel, line 619: Is this a bug fix that we should pick-up for Zeek 7.0 or early?

awelzel (dismissed):
Hey Vern - had a few thought on the non-script-opt changes.

I don't have the capacity to review the CPP or ZAM changes, if that concerns anyone before we merge this we should chat :-)

src/Type.cc
vpax: It's not needed for what shipped in Zeek 7.0, but it matters for some upcoming refinements in how lambdas are dealt with for script optimization

src/Traverse.h
```diff
@@ -41,11 +41,9 @@ class TraversalCallback {
...
     virtual TraversalCode PostDecl(const ID*) { return TC_CONTINUE; }
 
-    // A caution regarding using the next two: when traversing types,
-    // there's a possibility of encountering a (directly or indirectly)
-    // recursive record.  So you'll need some way of avoiding that,
-    // such as remembering which types have already been traversed
-    // and skipping via TC_ABORTSTMT when seen again.
+    // Note, if a type is recursive then the first time it's encountered
+    // it will be traversed, but when re-visited it'll simply return
+    // TC_CONTINUE.
```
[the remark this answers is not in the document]
vpax: Per our offline discussion, I reverted this and just implemented per-traverser checking, which is simpler than the alternatives

src/Val.h
```diff
@@ -682,6 +684,12 @@ class ListVal final : public Val {
      */
     void Append(ValPtr v);
 
+    /**
+     * Empties the list.
+     * @param v  the value to append.
+     */
+    void Clear() { vals.clear(); }
```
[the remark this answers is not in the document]
vpax: Good catch, thanks! Now addressed

src/ID.h
vpax: Your suggestion does indeed work, likewise thanks! I was concerned it might run into problems due to inlining and/or lambdas, but I've tested those and it remains valid, and much simpler

vpax:
The conflict is simple to resolve by just discarding the inclusion of that file here, since the same change was already made in https://github.com/zeek/zeek/pull/3886/ (but with a new function following it, hence the diff doesn't cleanly apply).

awelzel (approved):
Approving.

I've rebased and squashed fixups and yanked out reverts and the spurious zeekctl, spicy, doc submodule updates.

PR #3873, opened 2024-08-07
ci/ubuntu-24.04: Use ccache 4.10.2

awelzel:
The ccache version shipped with Ubuntu 24.04 does not yet recognize --fprofile-update=atomic, install one that does.

Now that the asan_sanitizer build also includes building Spicy and running the spicyz test suite, ccache is quite important.

Reference ccache/ccache#1408 and zeek/zeek#3777.

### DIFF
--- a/ci/ubuntu-24.04/Dockerfile
+++ b/ci/ubuntu-24.04/Dockerfile
@@ -4,7 +4,7 @@ ENV DEBIAN_FRONTEND="noninteractive" TZ="America/Los_Angeles"
 
 # A version field to invalidate Cirrus's build cache when needed, as suggested in
 # https://github.com/cirruslabs/cirrus-ci-docs/issues/544#issuecomment-566066822
-ENV DOCKERFILE_VERSION 20240528
+ENV DOCKERFILE_VERSION 20240807
 
 RUN apt-get update && apt-get -y install \
     bc \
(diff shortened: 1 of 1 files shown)


ci/ubuntu-24.04/Dockerfile
```diff
@@ -36,8 +36,35 @@ RUN apt-get update && apt-get -y install \
...
+    libxxhash-dev \
+    libzstd-dev \
   && apt autoclean \
   && rm -rf /var/lib/apt/lists/*
 
 RUN pip3 install --break-system-packages junit2html
 RUN gem install coveralls-lcov
+
+# Download and build a newer ccache version that recognizes -fprofile-update=atomic
+# which is used when building with --coverage.
```
bbannier, line 49: Are we compiling our own version since we have to run this on platforms other than x86_64 also? I am asking since there is are [prebuilt binaries](https://ccache.dev/download.html) for that platform. If that is the reason let's at least capture it in a comment.

awelzel: Uhm, no reason other than I compiled from source to fix it locally for me and doing the same in the Dockerfile.

I've re-pushed the branch. Now uses the pre-built ccache version. wdyt?

bbannier (approved):

PR #3879, opened 2024-08-12
Add packet analyzer history to unknown_protocols.log

J-Gras:
This is a rework of #3858.

---

Experimenting with `unknown_protocols.log` in production environments, it became clear that more context would be helpful to interpret the log. In particular, the lineage of analyzers that processed a packet before an unknown protocol is encountered would provide valuable context. This changes adds a new analyzer history field to track that information.


src/event.bif
```diff
@@ -885,8 +885,12 @@ event Pcap::file_done%(path: string%);
 ##
 ## first_bytes: A certain number of bytes at the start of the unknown protocol's header.
 ##
+## analyzer_history: The chain of packet analyzers that processed the packet up to this
+##                   point. This includes the history of encapsulating packets in case
+##                   of tunneling.
+##
 ## .. zeek:see:: UnknownProtocol::first_bytes_count
-event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: string%);
+event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: string, analyzer_history: string_vec%);
```
awelzel, line 893: Two thoughts:
* We need to keep the old `unknown_protocol()` event and deprecate it for 8.1 (see recent commit 401a07403628f32c09eb03c57dba5e93fa35681b). Even if you're the only user ;-)
* A bit nit-picky, wondering about returning `vector of PacketAnalyzer::Tag` rather than the string version. The tag seems nicer API wise. For getting a short history as you have right now, this needs fiddling with the stringified tag in script land unfortunately, so I can certainly see the convenience if you want to push back on that idea.

src/packet_analysis/Manager.h
```diff
@@ -176,6 +176,13 @@ class Manager : public plugin::ComponentManager<Component> {
      */
     uint64_t GetUnprocessedCount() const { return total_not_processed; }
 
+    /**
+     * Tracks the given analyzer for the current packet's analyzer history.
+     * The packet analyzer history is implemented in form of a stack, which is reset on a
+     * call to ProcessPacket() but maintained throughout calls to ProcessInnerPacket().
+     */
+    void TrackAnalyzer(AnalyzerPtr analyzer) { analyzer_stack.push_back(std::move(analyzer)); }
```
awelzel, line 184: Suggest `PushAnalyzer()` as it implies `stack` without needing to check the docs.

Could you add `@analyzer` doxygen documentation?

src/event.bif
awelzel: Ah, I'm seeing `unknown_protocol()` already gets an analyzer name instead of a tag. Ignore point two and lets stick with a string for consistency.

awelzel:
The zeek-benchmarker run indicates there's no performance overhead visible. Also tested locally and it's cheap enough :+1: 

It's interesting to me that your prior version with `std::vector<std::shared_ptr<Analyzer>>` had basically no overhead either, but prefer to avoid constructing them still and go via raw pointers as per DM. You could constify the pointers if you wanted.

With the old `unknown_protocols()` event deprecated, happy to merge!

image

src/event.bif
J-Gras: I've added deprecation for the old event signature.

FWIW, my gut feeling was that the tag version could also come with a measurable performance hit because the Tags need to be translated in script-land. I've [implemented it](https://github.com/J-Gras/zeek/tree/topic/jgras/packet-analyzer-history-rework-tags) but couldn't see any significant impact for the given workload in my local tests:
```
$ hyperfine -w 1 -m 20 "~/zeek-installs/zeek-7.1.0-analyzer-history-rework-tags/bin/zeek -CD -r 2009-M57-day11-18.pcap" "~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap"
Benchmark 1: ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-tags/bin/zeek -CD -r 2009-M57-day11-18.pcap
  Time (mean ± σ):      2.707 s ±  0.044 s    [User: 2.208 s, System: 1.985 s]
  Range (min … max):    2.660 s …  2.860 s    20 runs
 
Benchmark 2: ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap
  Time (mean ± σ):      2.698 s ±  0.065 s    [User: 2.205 s, System: 1.978 s]
  Range (min … max):    2.584 s …  2.874 s    20 runs
 
Summary
  ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap ran
    1.00 ± 0.03 times faster than ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-tags/bin/zeek -CD -r 2009-M57-day11-18.pcap
```

src/packet_analysis/Manager.h
J-Gras: I've added the doxygen documentation.

Regarding renaming the method: I'd argue that the history being a stack could be seen as an implementation detail. I'd prefer calling it `TrackAnalyzer()`. If that's a blocker I'd be okay with `PushAnalyzerToHistory()` as well :)

J-Gras:
Thanks for taking a look at the benchmarker! I've seen the similar results for my flaky local tests:
```
$ hyperfine -w 1 -m 20 "~/zeek-installs/zeek-7.1.0-master/bin/zeek -CD -r 2009-M57-day11-18.pcap" "~/zeek-installs/zeek-7.1.0-analyzer-history-rework/bin/zeek -CD -r 2009-M57-day11-18.pcap" "~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap"
Benchmark 1: ~/zeek-installs/zeek-7.1.0-master/bin/zeek -CD -r 2009-M57-day11-18.pcap
  Time (mean ± σ):      2.696 s ±  0.063 s    [User: 2.194 s, System: 1.971 s]
  Range (min … max):    2.628 s …  2.925 s    20 runs
 
Benchmark 2: ~/zeek-installs/zeek-7.1.0-analyzer-history-rework/bin/zeek -CD -r 2009-M57-day11-18.pcap
  Time (mean ± σ):      2.713 s ±  0.044 s    [User: 2.218 s, System: 1.976 s]
  Range (min … max):    2.601 s …  2.791 s    20 runs
 
Benchmark 3: ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap
  Time (mean ± σ):      2.720 s ±  0.065 s    [User: 2.215 s, System: 1.978 s]
  Range (min … max):    2.548 s …  2.898 s    20 runs
 
Summary
  ~/zeek-installs/zeek-7.1.0-master/bin/zeek -CD -r 2009-M57-day11-18.pcap ran
    1.01 ± 0.03 times faster than ~/zeek-installs/zeek-7.1.0-analyzer-history-rework/bin/zeek -CD -r 2009-M57-day11-18.pcap
    1.01 ± 0.03 times faster than ~/zeek-installs/zeek-7.1.0-analyzer-history-rework-raw/bin/zeek -CD -r 2009-M57-day11-18.pcap
```

awelzel (approved):

PR #3885, opened 2024-08-13
Make ssl_history work for SSLv2 handshakes/connections

0xxon:
It turns out that the ssl_history field never was populated with C/S for SSLv2 connections, or connections using the SSLv2 handshake. In our testcases, the latter is especially common - with connections up to TLS1 using the old SSLv2 client hello for backwards compatibility.

This change resolves this issue. As the history is not by default enabled in a lot of locations, baseline impact is minor.

Needs merge of the external testing repos.


awelzel:
> git clone https://github.com/zeek/zeek-testing && ./scripts/sync-repo $(basename https://github.com/zeek/zeek-testing)
> Cloning into 'zeek-testing'...
> Moving 'zeek-testing' to e3a2d006118261fe49e10ac44f6668204e21ccd4
> fatal: reference is not a tree: e3a2d006118261fe49e10ac44f6668204e21ccd4

Did you forget to push?

scripts/base/protocols/ssl/main.zeek
```diff
@@ -282,6 +282,11 @@ event ssl_client_hello(c: connection, version: count, record_version: count, pos
 		c$ssl$session_id = bytestring_to_hexstr(session_id);
 		c$ssl$client_ticket_empty_session_seen = F;
 		}
+
+	# add manually for SSLv2, since the handshake_message event is not raised, as there is no handshake protocol.
```
awelzel, line 286:
```suggestion
	# add client hello letter for SSLv2 to history, since the handshake_message event is not raised, as there is no handshake protocol.
```

Total nit, didn't immediately know what "c" stands for :-) similarly for "s" below. Take or reject :-)

awelzel (approved):

PR #3886, opened 2024-08-13
-O gen-C++ initialization improvements (and some other tweaks)

vpax:
I've been using `-O gen-C++` script optimization on some large script collections, which revealed that some facets of how the generated code performs its (massive) initializations could take a huge amount of time to compile. This PR addresses two instances of the problem, both regarding large `std::vector` initializers. Compilation is much slower if the individual initializers are non-trivial C++ objects as opposed to very simple types (`int`s and `struct`s, in particular), so the changes switch to using the latter.

Running on the diverse set of scripts also exposed a number of previously-untickled bugs, which the PR also fixes. Finally, it includes some low-level changes for tidy-ness and enabling more flexible use of some compile-time and run-time components.

### DIFF
--- a/src/script_opt/CPP/Func.h
+++ b/src/script_opt/CPP/Func.h
@@ -8,9 +8,7 @@
 #include "zeek/Func.h"
 #include "zeek/script_opt/ProfileFunc.h"
 
-namespace zeek {
-
-namespace detail {
+namespace zeek::detail {
 
 // A subclass of Func used for lambdas that the compiler creates for
 // complex initializations (expressions used in type attributes).
@@ -42,11 +40,6 @@ public:
 
     const std::string& Name() { return name; }
 
-    // Sets/returns a hash associated with this statement.  A value
-    // of 0 means "not set".
-    p_hash_type GetHash() const { return hash; }
-    void SetHash(p_hash_type h) { hash = h; }
-
     // The following only get defined by lambda bodies.
     virtual void SetLambdaCaptures(Frame* f) {}
     virtual std::vector<ValPtr> SerializeLambdaCaptures() const { return std::vector<ValPtr>{}; }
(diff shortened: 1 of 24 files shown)


awelzel (approved):
I would merge but with the caveat that I'm not even looking at src/script_opt.

I wonder if we should open a discussion whether we should rather spend time making it feasible to compile to C++ via plugin functionality, or accept for now that src/script_opt gets a free pass in terms of review.

vpax:
> I wonder if we should open a discussion whether we should rather spend time making it feasible to compile to C++ via plugin functionality, or accept for now that src/script_opt gets a free pass in terms of review.

I'm not following your thoughts here. Is it that implementing it via plugin functionality would change the expectations of the degree of review that the code has received?

awelzel:
> I'm not following your thoughts here. Is it that implementing it via plugin functionality would change the expectations of the degree of review that the code has received?

I was thinking that if "Zeek script to C++" conversion was an external plugin (separate repo, separate build, ..., experimental status), you could work on that without any review. IMO not very different (to me at least) than merging changes in `src/script_opt` un-reviewed. But, it would likely slow you down as it's more difficult to add or change existing Zeek APIs.

src/script_opt/CPP/Func.h
```diff
@@ -85,6 +77,7 @@ class CPPLambdaFunc : public ScriptFunc {
     // Methods related to sending lambdas via Broker.
     std::optional<BrokerData> SerializeCaptures() const override;
     void SetCaptures(Frame* f) override;
+    void SetCapturesVec(std::unique_ptr<std::vector<ZVal>> _captures_vec) { captures_vec = std::move(_captures_vec); }
```
awelzel, line 80: I'm not seeing where this is used in the context of this PR and am also thinking this could just go into `class Func`, I'll do that as a fixup and make the field private again. If it breaks something, you might need to add a fixup.

PR #3887, opened 2024-08-14
Add DNS TKEY event

### LINKED ISSUE #3857: DNS TKEY support
Reference: [RFC-2930 - Secret Key Establishment for DNS (TKEY RR)](https://datatracker.ietf.org/doc/html/rfc2930)

PCAP harvested from [CTU-SME-11](https://zenodo.org/records/7958259) dataset (Experiment-VM-Microsoft-Windows7AD-1):
```
1676937749.535171       CixPbKPKgWzgi7VR9       192.168.1.106   50138   192.168.1.108   53      DNS_RR_unknown_type     249     F       zeek    DNS
```

There's a `dns.log` entry and `conn.log` also contains service `dns`, but not having the weird produced would be good, and possibly a dns_TKEY event.

```
#types  time    string  addr    port    addr    port    enum    count   interval        string  count   string  count   string  count   string  bool    bool    bool    bool    count   vector[string]  vector[interval]        bool
1676937749.533770       CixPbKPKgWzgi7VR9       192.168.1.106   50138   192.168.1.108   53      tcp     52640   0.001401        1068-ms-7.309-2c6e448.7a9463b8-b109-11ed-26a3-080027f220e5      1       C_INTERNET      249     TKEY    0       NOERROR F       F       F       F       0       <unknown type=249>      0.000000        F
```

[ctu-sme-11-win7ad-1-dns-tcp-tkey.pcap.gz](https://github.com/user-attachments/files/16393494/ctu-sme-11-win7ad-1-dns-tcp-tkey.pcap.gz)

evantypanski:
Closes #3857 

This not only removes the weird, but also adds an event :)

I couldn't easily find sites that use DNS TKEY, so I couldn't get an organic test. I just used the pcap provided by @awelzel in the issue for the test, but it's not exactly "exhaustive"

There are other minor points I have with style, but not even sure they're worth calling out so I'll defer that for now


testing/btest/Baseline/scripts.base.protocols.dns.tkey/output
```diff
@@ -0,0 +1,3 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
awelzel, line 1: could you unset  TEST_DIFF_CANONIFIER= for `output`, then inception and expiration timestamps won't be canonified to XXXXXXXXXX.XXXXXX :-)

Also, printing the individual fields explicitly, and using `|ans$key_data|` is easier to grasp.

Mostly FYI, nothing to change: Zeek script has `assert` now, but many of the existing tests aren't using it and we still do a lot of baselines.

testing/btest/scripts/base/protocols/dns/tkey.zeek
```diff
@@ -0,0 +1,10 @@
+# @TEST-EXEC: zeek -b -C -r $TRACES/dns/tkey.pcap %INPUT > output
+# @TEST-EXEC: btest-diff dns.log
+# @TEST-EXEC: btest-diff output
+
+@load policy/protocols/dns/auth-addl
```
awelzel, line 5: Any reason this policy script is loaded?

```suggestion
@load base/protocols/dns
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1660,6 +1704,23 @@ RecordValPtr DNS_MsgInfo::BuildEDNS_COOKIE_Val(struct EDNS_COOKIE* opt) {
...
 }
 
+RecordValPtr DNS_MsgInfo::BuildTKEY_Val(struct TKEY_DATA* tkey) {
+    static auto dns_tkey = id::find_type<RecordType>("dns_tkey");
+    auto r = make_intrusive<RecordVal>(dns_tkey);
+
+    r->Assign(0, query_name);
+    r->Assign(1, answer_type);
+    r->Assign(2, tkey->alg_name);
+    r->AssignTime(3, double(tkey->inception));
```
awelzel, line 1714: I want to suggest the following to be consistent with size_t below, but I realize there's more of that in `DNS.cc`.

```suggestion
    r->AssignTime(3, static_cast<double>(tkey->inception));
```

Do you have thoughts/opinions here?

scripts/base/init-bare.zeek
```diff
@@ -4083,6 +4083,21 @@ type dns_edns_cookie: record {
...
 
+## A DNS TKEY record.
+##
+## .. zeek:see:: dns_TKEY
+type dns_tkey: record {
+	query: string;		##< Query.
+	qtype: count;		##< Query type.
+	alg_name: string;	##< Algorithm name.
+	inception: time;	##< Time when signed.
+	expiration: time;	##< Time when the signed RR expires.
```
awelzel, line 4094: Would be good to update to something else than the TSIG ones.

```suggestion
	inception: time;	##< Requested or provided start of validity interval for keying material.
	expiration: time;	##< Requested or provided end of validity interval for keying material.
```
https://datatracker.ietf.org/doc/html/rfc2930#section-2.4

testing/btest/scripts/base/protocols/dns/tkey.zeek
```diff
@@ -0,0 +1,10 @@
+# @TEST-EXEC: zeek -b -C -r $TRACES/dns/tkey.pcap %INPUT > output
+# @TEST-EXEC: btest-diff dns.log
```
awelzel, line 2: Could add a `TEST-EXEC: test ! -f weird.log` to ensure the prior weird isn't generated anymore (an won't be in the future without noticing).

awelzel (commented):
Nice Evan! Seems you found your way around :-)

I only have nits, happy to merge afterwards.

You could add an entry about this in the NEWS file, too.

testing/btest/scripts/base/protocols/dns/tkey.zeek
evantypanski: So with the policy script there are two `TKEY` events triggered in the baseline (with field-by-field output):

```
TKEY
query, 1068-ms-7.309-2c6e448.7a9463b8-b109-11ed-26a3-080027f220e5
qtype, 3
alg_name, gss-tsig
inception, 1676938156.0
expiration, 1677024556.0
mode, 3
rr_error, 0
key_data size, 3073
is_query, 1
TKEY
query, 1068-ms-7.309-2c6e448.7a9463b8-b109-11ed-26a3-080027f220e5
qtype, 1
alg_name, gss-tsig
inception, 1676938118.0
expiration, 1677024518.0
mode, 3
rr_error, 0
key_data size, 185
is_query, 0
```

With `base/protocols/dns` there's only one:

```
TKEY
query, 1068-ms-7.309-2c6e448.7a9463b8-b109-11ed-26a3-080027f220e5
qtype, 1
alg_name, gss-tsig
inception, 1676938118.0
expiration, 1677024518.0
mode, 3
rr_error, 0
key_data size, 185
is_query, 0
```

Since I couldn't confirm why the first one was dropped I went with the one that gave more to test, but happy to change that if there's a particular reason I should know.

src/analyzer/protocol/dns/DNS.cc
evantypanski: I prefer `static_cast` - just forgot to swap it over after I found out about `AssignTime` :)

testing/btest/scripts/base/protocols/dns/tkey.zeek
awelzel: Ah! Seems the `dns_skip_all_addl` defaults to `T`. Can check the `tsig.zeek` test setting it to `F`:
```
redef dns_skip_all_addl = F;
```

Hmm, I wonder a bit if for TKEY this flag should be ignored (the addl rr seems a crucial part), but that's for another time.

awelzel (approved):

PR #3888, opened 2024-08-15
spicy/protocol-analyzer: Move Done() after EndOfData()

awelzel:
We've got crash reports involving Spicy analyzers that appear to be caused by `ProtocolAnalyzer::Done()` or `EndOfData()` instantiating new child analyzers *after* `analyzer::tcp::TCP_ApplicationAnalyzer::Done();` was called, resulting in the following code being executed:

https://github.com/zeek/zeek/blob/ac5cbcc43e7527a22513ec7ce5685b2bdeefbb98/src/analyzer/Analyzer.cc#L117-L122

The call to `->Done()` on `new_children` in turn may reference the `TCP_SessionAdapter`, e.g. through `AnalyzerViolation()`, resulting in access to already freed memory.

This feels like a stop gap. We've done something similar for the HTTP analyzer in the past, but I wonder if there's a better idea? 6ef9423f3cff13e6c73f97eb6a3a27d6f64cc320

### DIFF
--- a/src/spicy/protocol-analyzer.cc
+++ b/src/spicy/protocol-analyzer.cc
@@ -131,8 +131,6 @@ void TCP_Analyzer::Init() {
 }
 
 void TCP_Analyzer::Done() {
-    ProtocolAnalyzer::Done();
-
     EndOfData(true);
     EndOfData(false);
 
--- a/src/spicy/protocol-analyzer.cc
+++ b/src/spicy/protocol-analyzer.cc
@@ -131,11 +131,15 @@ void TCP_Analyzer::Init() {
 }
 
 void TCP_Analyzer::Done() {
-    analyzer::tcp::TCP_ApplicationAnalyzer::Done();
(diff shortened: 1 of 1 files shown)


rsmmr:
> This feels like a stop gap. We've done something similar for the HTTP analyzer in the past, but I wonder if there's a better idea?

As stop-gap I'm approving if that seems to fix the crashes, but I'm wondering if that code needs to move out the destructor, over to some point where it's more well defined. The destructor should just release any pending memory, but not be able to trigger new activity. Or: Maybe we could even just simply skip the call to `Done()` on any analyzers that are still pending in `new_children` at destructor time, although I'm kind of guessing that this is there for a reason ...

rsmmr (approved):

awelzel:
@pbcullen reports this doesn't fix the crashes they observe. The culprit is DPD attaching a new analyzer during `Done()` instead. They're doing more testing.

PR #3889, opened 2024-08-16
ScriptOpt: Ensure global statements have non-null scope

awelzel:
The ProfileFunc() logic assumed that GetScope() returned a non-nullptr. This holds except for the synthetic global statements function.

Fix the latter and add an assert, also add a name to the type so it's easier to recognize in a debugger what's going on, otherwise the name is "".

This was found by UBSAN due to it seeing the ->OrderedVars() call on a nullptr. Elsewhere, num_params == 0 shielded from that access and so didn't lead to crashes.

### DIFF
--- a/src/script_opt/ProfileFunc.cc
+++ b/src/script_opt/ProfileFunc.cc
@@ -80,6 +80,8 @@ ProfileFunc::ProfileFunc(const Expr* e, bool _abs_rec_fields) {
 void ProfileFunc::Profile(const FuncType* ft, const StmtPtr& body) {
     num_params = ft->Params()->NumFields();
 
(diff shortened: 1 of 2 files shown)


vpax:
Seems an easier fix is to just gate the loop that spins through the parameters on the scope pointer being non-nil ... ?

vpax:
> Seems an easier fix is to just gate the loop that spins through the parameters on the scope pointer being non-nil ... ?

Per an offline discussion, this would work, but it's tidy to have a scope associated with the global-statements pseudo-function. Given that, this looks good to me

vpax (approved):
Per the comment thread, this LGTM

Issue #3881, opened 2024-08-13
Reintroduce .evt port/ports handling as Zeek event


awelzel:
The idea is to provide an event from Spicy to Zeek that reports all available analyzers, including their meta data. We can then provide a standard handler that registers them for any ports provided inside the *.evt and *.spicy files. We won't be readding `port` to the analyzer template (because defining the ports Zeek-side is more idiomatic), but can use it inside the tutorial to keep things simpler (with a footnote pointing to how you'd do it more idiomatically).

awelzel:
Closing - change from Robin has been backported to 7.0.1.

PR #3891, opened 2024-08-19
Un-deprecate Spicy's `port`

rsmmr:
This goes back to supporting `port` in EVT files, per #3881.

I ended up taking a simpler route that just raises one event per
analyzer port, with no customization points, because (1) that's what
we already do for file analyzers' MIME types, and (2) it's simple
enough that I think we can/should backport this to 7.0.

This reverts the previous commit that removed support for `port` field
but it doesn't revert the earlier, larger commit where we deprecated
it (except for removing the actual deprecation warning). In
particular, I'm not reverting all the changes to the tests that came
with the deprecation, since they're either still useful or at least
don't hurt.

- **Revert "Remove deprecated port/ports fields for spicy analyzers"**
- **Spicy: Register well-known ports through an event handler.**


timwoj (approved):

scripts/base/frameworks/spicy/init-framework.zeek
```diff
@@ -47,12 +47,18 @@ export {
...
 # Marked with &is_used to suppress complaints when there aren't any
 # Spicy file analyzers loaded, and hence this event can't be generated.
-# The attribute is only supported for Zeek 5.0 and higher.
 event spicy_analyzer_for_mime_type(a: Files::Tag, mt: string) &is_used
     {
     Files::register_for_mime_type(a, mt);
     }
 
+# Marked with &is_used to suppress complaints when there aren't any
+# Spicy protocol analyzers loaded, and hence this event can't be generated.
```
awelzel, line 56: Mostly nit, if we were to get a handle to the event handler via `zeek::event_registry->Register("spicy_analyzer_for_port", false /*is_from_script*/);` unconditionally during `InitPostScript()` or so, wouldn't need the `&is_used` annotation.

src/spicy/manager.cc
```diff
@@ -699,14 +702,44 @@ void Manager::InitPostScript() {
...
+        auto register_analyzer_for_port = [&](auto tag, const hilti::rt::Port& port_) {
+            SPICY_DEBUG(hilti::rt::fmt("  Scheduling analyzer for port %s", port_));
+
+            // Well-known ports are registered in scriptland, so we'll raise an
+            // event that will do it for us through a predefined handler.
+            zeek::Args vals = Args();
+            vals.emplace_back(tag.AsVal());
+            vals.emplace_back(zeek::spicy::rt::to_val(port_, base_type(TYPE_PORT)));
+            EventHandlerPtr handler = event_registry->Register("spicy_analyzer_for_port");
+            event_mgr.Enqueue(handler, vals);
```
awelzel, line 714: Another nit, there's a vaargs function, that voids needing the explicit args (and I wonder if the lambda could go then).
```
event_mgr.Enqueue(handler, tag.AsVal(), zeek::spicy::rt::to_val(port_, base_type(TYPE_PORT)));
```

Ah, but that pattern is used for mimes, too. Never mind.

src/spicy/manager.h
```diff
@@ -85,6 +85,7 @@ class Manager : public zeek::plugin::Plugin {
      *
      * @param name name of the analyzer as defined in its EVT file
      * @param proto analyzer's transport-layer protocol
+     * @param prts well-known ports for the analyzer; it'll be activated automatically for these
```
awelzel, line 88:
```suggestion
     * @param ports well-known ports for the analyzer; it'll be activated automatically for these
```

src/spicy/manager.h
```diff
@@ -354,7 +357,7 @@ class Manager : public zeek::plugin::Plugin {
         bool operator==(const ProtocolAnalyzerInfo& other) const {
             return name_analyzer == other.name_analyzer && name_parser_orig == other.name_parser_orig &&
                    name_parser_resp == other.name_parser_resp && name_replaces == other.name_replaces &&
-                   protocol == other.protocol && linker_scope == other.linker_scope;
+                   protocol == other.protocol && ports == other.ports && linker_scope == other.linker_scope;
```
awelzel, line 360: Not sure that matters, but `port 21/tcp, port 22/tcp` is different from `port 22/tcp, port 21/tcp`. Wonder if the parsing code should sort/normalize?

awelzel (approved):
This is so much nicer :-)

scripts/base/frameworks/spicy/init-framework.zeek
rsmmr: Yeah, seems fine to me as it is as well though.

src/spicy/manager.cc
rsmmr: Used elsewhere in the Spicy code too iirc, so I'll leave it for now, could clean up/modernize some other time.

src/spicy/manager.h
rsmmr: Good point, will do.

Issue #3856, opened 2024-07-26
DNS Dynamic Update: Produces weird and is not confirmed


awelzel:
The following PCAP contains two DNS connections with opcode "UPDATE (5)". We purposely don't log these in dns.log. Instead of ignoring, however, the following weirds are produced and they aren't confirmed as DNS connections (so conn.log has an empty service field):

```
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       name    addl    notice  peer    source
#types  time    string  addr    port    addr    port    string  string  bool    string  string
1676940453.027625       Cdgarz4QLOH6h1qEE7      192.168.1.106   62763   192.168.1.108   53      DNS_zero_rdlength       -       F       zeek    DNS
1676968917.288586       Cwzjvm1eGfcAQZXr84      192.168.1.105   62763   192.168.1.108   53      DNS_zero_rdlength       -       F       zeek    DNS
```

Not sure we want to support all of [DNS Dynamic Update](https://datatracker.ietf.org/doc/html/rfc2136), but squelching the weirds would be good.

[ctu-sme-11-win7ad-1-dns-tcp-62763.pcap.gz](https://github.com/user-attachments/files/16392148/ctu-sme-11-win7ad-1-dns-tcp-62763.pcap.gz)

PCAP harvested from [CTU-SME-11](https://zenodo.org/records/7958259) dataset where a large amount of weirds produces was for DNS_zero_rdlength.

What Wireshark thinks:
![Screenshot from 2024-07-26 15-04-08](https://github.com/user-attachments/assets/0b17d200-ee20-4819-9c50-6ed699510a38)

timwoj:
Looking at this just very briefly, we handle everything in the DNS analyzer as if it's opcode 0 (a standard query). Opcode 1 (an inverse query) might work as well. All of the other opcodes are very unlikely work correctly. A quick fix here, assuming we don't want to add support for these messages, would be to just check for opcode 0 and bail out. If we want to we could confirm the analyzer as well, but we wouldn't have done any other processing to verify that the data was correct.

timwoj:
The above PR adds a new weird for unhandled opcodes, but does not add support for Dynamic Update or the other missing opcodes. I'll leave this issue open for now.

timwoj:
I spent a little time looking at this today. It doesn't seem too much of a stretch to support dynamic update.

[...]

PR #3892, opened 2024-08-20
Add weird for unhandled opcodes in DNS analyzer

timwoj:
This fixes part of #3856. It changes to report a weird for the unhandled opcode instead of a nonsensical one. The sessions still aren't marked as DNS in conn.log.


awelzel:
Seems m57-long contains NETBIOS over UDP 137 that ended up parsed as DNS with opcodes 5/6. They actually look interesting and include release/registration of hostnames, though we never logged this somewhere properly, just marked it DNS. So this may or may not be better now :man_shrugging:

Here's two pcaps extracted:

[netbios-release-registration-192.168.1.104-137.pcap.gz](https://github.com/user-attachments/files/16688080/netbios-release-registration-192.168.1.104-137.pcap.gz)
[netbios-registration-169.254.225.22-137.pcap.gz](https://github.com/user-attachments/files/16688081/netbios-registration-169.254.225.22-137.pcap.gz)

timwoj:
> Seems m57-long contains NETBIOS over UDP 137 that ended up parsed as DNS with opcodes 5/6. They actually look interesting and include release/registration of hostnames, though we never logged this somewhere properly, just marked it DNS. So this may or may not be better now

Because of course NetBIOS would break things 🤦🏼 

There's nothing in these packets that indicate that this is NetBIOS outside of (1) the port number and (2) the type field in the queries/records blocks being set to `32`. For (2), checking for that won't necessarily work because that value overlaps with other valid values (see footnote [1] in https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4).

timwoj:
I added a check for the NetBIOS port to the interpreter and updated the external tests. A branch of the same name needs to be merged there as well.

awelzel (commented):
Left some nits. Were you thinking of closing the associated issue? I'd actually keep it open as this doesn't add DNS UPDATE support (just improves the weirds).

timwoj:
> Left some nits. Were you thinking of closing the associated issue? I'd actually keep it open as this doesn't add DNS UPDATE support (just improves the weirds).

No, I'd leave the issue open until we decide whether to support the other DNS opcodes.

awelzel (approved):

PR #3893, opened 2024-08-21
logging: Replace threading::Value** with std::vector

awelzel:
While looking at the logging API/code: When converting from `RecordVal` to the `threading::Value**` construct, there are individual `new threading::Value(...)` calls for every field. Instead, it's also possible to allocate the needed `threading::Value`s as a single array in one go, reducing pressure on the malloc implementation. While at it, also switching to `std::vector` for automatic memory management.

This is a RFC PR to a) run the benchmarker and b) get a feeling if others would be okay changing what I think are mostly internal APIs on WriterFrontend and WriterBackend. I added backwards compat, but these are "slow implementations" doing copies that maybe we shouldn't even provide.

For a 500k syn packet pcap (10kpps), with `-C -D -b base/protocols/conn` using glibc's malloc, processing is roughly 17% faster:

|:---|---:|---:|---:|---:|
| `taskset -c 1 /opt/zeek-dev-prod-orig/bin/zeek --no-unused-warnings -D -C -b -r /media/awelzel/T7/pcaps/awelzel/500k-syns-slow.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE` | 16.604 ± 0.073 | 16.559 | 16.688 | 1.21 ± 0.01 |
| `taskset -c 1 /opt/zeek-dev-prod/bin/zeek --no-unused-warnings -D -C -b -r /media/awelzel/T7/pcaps/awelzel/500k-syns-slow.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE` | 13.768 ± 0.109 | 13.685 | 13.892 | 1.00 |

Using jemalloc, with this branch it's maybe 1%-2% faster, but by far not as impressive:

|:---|---:|---:|---:|---:|
| `taskset -c 1 /opt/zeek-dev-prod-orig/bin/zeek --no-unused-warnings -D -C -b -r /media/awelzel/T7/pcaps/awelzel/500k-syns-slow.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE` | 10.203 ± 0.026 | 10.184 | 10.232 | 1.02 ± 0.01 |
| `taskset -c 1 /opt/zeek-dev-prod/bin/zeek --no-unused-warnings -D -C -b -r /media/awelzel/T7/pcaps/awelzel/500k-syns-slow.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE` | 10.050 ± 0.102 | 9.983 | 10.168 | 1.00 |

---

PCAPs (not so exciting except for 500k-syns-slow).

Screenshot from 2024-08-21 14-23-16

Logging micro-benchmarks show improvement (table-complex-key most left is likely noise, shouldn't be impacted by this changeset).
Screenshot from 2024-08-21 14-24-15

### DIFF
--- a/src/logging/Manager.cc
+++ b/src/logging/Manager.cc
@@ -1158,9 +1158,10 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
         if ( zeek::plugin_mgr->HavePluginForHook(zeek::plugin::HOOK_LOG_WRITE) ) {
             // The current HookLogWrite API takes a threading::Value**.
             // Fabricate the pointer array on the fly. Mutation is allowed.
-            std::vector<threading::Value*> vals(rec.size());
-            for ( size_t i = 0; i < rec.size(); i++ )
-                vals[i] = &rec[i];
+            std::vector<threading::Value*> vals;
+            vals.reserve(rec.size());
+            for ( auto& v : rec )
+                vals.emplace_back(&v);
 
             bool res =
                 zeek::plugin_mgr->HookLogWrite(filter->writer->GetType()->AsEnumType()->Lookup(
--- a/src/logging/Manager.cc
+++ b/src/logging/Manager.cc
@@ -1153,21 +1153,25 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
         }
 
         // Alright, can do the write now.
+        auto rec = RecordToLogRecord(stream, filter, columns.get());
 
-        threading::Value** vals = RecordToFilterVals(stream, filter, columns.get());
+        if ( zeek::plugin_mgr->HavePluginForHook(zeek::plugin::HOOK_LOG_WRITE) ) {
+            // The current HookLogWrite API takes a threading::Value**.
+            // Fabricate the pointer array on the fly. Mutation is allowed.
+            std::vector<threading::Value*> vals(rec.size());
+            for ( size_t i = 0; i < rec.size(); i++ )
+                vals[i] = &rec[i];
 
-        if ( ! PLUGIN_HOOK_WITH_RESULT(HOOK_LOG_WRITE,
-                                       HookLogWrite(filter->writer->GetType()->AsEnumType()->Lookup(
-                                                        filter->writer->InternalInt()),
-                                                    filter->name, *info, filter->num_fields, filter->fields, vals),
-                                       true) ) {
-            DeleteVals(filter->num_fields, vals);
+            bool res =
+                zeek::plugin_mgr->HookLogWrite(filter->writer->GetType()->AsEnumType()->Lookup(
+                                                   filter->writer->InternalInt()),
+                                               filter->name, *info, filter->num_fields, filter->fields, &vals[0]);
+            if ( ! res ) {
+                DBG_LOG(DBG_LOGGING, "Hook prevented writing to filter '%s' on stream '%s'", filter->name.c_str(),
+                        stream->name.c_str());
 
-#ifdef DEBUG
-            DBG_LOG(DBG_LOGGING, "Hook prevented writing to filter '%s' on stream '%s'", filter->name.c_str(),
-                    stream->name.c_str());
-#endif
-            return true;
+                return true;
+            }
         }
 
         assert(w != stream->writers.end());
@@ -1175,7 +1179,7 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
 
         // Write takes ownership of vals.
         assert(writer);
-        writer->Write(filter->num_fields, vals);
+        writer->Write(std::move(rec));
 
 #ifdef DEBUG
         DBG_LOG(DBG_LOGGING, "Wrote record to filter '%s' on stream '%s'", filter->name.c_str(), stream->name.c_str());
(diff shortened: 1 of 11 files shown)


src/logging/Manager.h
```diff
@@ -268,23 +268,25 @@ class Manager : public plugin::ComponentManager<Component> {
...
      *
-     * @param num_fields The number of log values to write.
+     * @param rec Representation of the log record to write.
+
+     * @return Returns true if the record was processed successfully.
      *
-     * @param vals An array of log values to write, of size num_fields.
-     * The method takes ownership of the array.
+     * TODO: Extend and stabilize API with 8.0. With pluggable cluster backends,
+     *       we may also include the filter name and support batched log records.
```
rsmmr, line 287: Take out, like above.

src/logging/WriterFrontend.h
```diff
@@ -8,6 +8,70 @@ namespace zeek::logging {
...
+ *
+ * \see WriterFrontend::Write
+ */
+class WriteBuffer {
+public:
+    /**
+     * Constructor.
+     */
+    explicit WriteBuffer(size_t buffer_size) : buffer_size(buffer_size) {}
+    /**
```
rsmmr, line 26:
```suggestion

    /**
```

src/logging/WriterFrontend.h
```diff
@@ -8,6 +8,70 @@ namespace zeek::logging {
...
+     * Push a record to the buffer.
+     *
+     * @param record The records vals.
+     */
+    void PushRecord(LogRecord&& record) { records.emplace_back(std::move(record)); }
```
rsmmr, line 31: Suggestion on terminology: Given that this a *write* buffer, maybe call this either `WriteRecord` or even simply `Write`?

src/logging/WriterFrontend.h
```diff
@@ -90,7 +154,30 @@ class WriterFrontend {
      *
      * This method must only be called from the main thread.
      */
-    void Write(int num_fields, threading::Value** vals);
+    [[deprecated("Remove in v8.1. Usedetail::LogRecord instead.")]] void Write(int num_fields, threading::Value** vals);
```
rsmmr, line 157:
```suggestion
    [[deprecated("Remove in v8.1. Use detail::LogRecord instead.")]] void Write(int num_fields, threading::Value** vals);
```

rsmmr: Do we need this deprecation, or can the method just go?

awelzel: Leftover. thanks!

src/threading/SerialTypes.cc
```diff
@@ -82,31 +82,88 @@ std::string Field::TypeName() const {
...
+
+        delete[] val.val.vector_val.vals;
+    }
+}
+
+Value::Value(const Value& other) {
+    if ( this == &other )
+        return;
+
+    free_value(this);
```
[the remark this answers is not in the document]
awelzel: Oh, I missed this. Thanks! I'll revert the `free_value()` change.

src/logging/Manager.cc
```diff
@@ -1153,29 +1153,33 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
...
 
         // Alright, can do the write now.
+        detail::LogRecord rec = RecordToLogRecord(stream, filter, columns.get());
+
+        if ( zeek::plugin_mgr->HavePluginForHook(zeek::plugin::HOOK_LOG_WRITE) ) {
+            // The current HookLogWrite API takes a threading::Value**.
+            // Fabricate the pointer array on the fly. Mutation is allowed.
+            std::vector<threading::Value*> vals(rec.size());
+            for ( size_t i = 0; i < rec.size(); i++ )
+                vals[i] = &rec[i];
```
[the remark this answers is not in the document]
awelzel: The pattern was to use index-assignment for the raw pointer vectors, and `emplace_back()` if a vector holds an actual `threading::Value`.

Your suggestion also requires a change to `.reserve(rec.size())` instead of the `vector(rec.size())`.

I'll change to emplace_back(), just wasn't completely random :-)

src/threading/SerialTypes.cc
```diff
@@ -82,31 +82,88 @@ std::string Field::TypeName() const {
...
 
+Value::Value(Value&& other) noexcept {
+    present = other.present;
+    type = other.type;
+    subtype = other.type;
+    line_number = other.line_number;
+
+    val = other.val; // take ownership.
+    other.val.set_val.size = 0;
+    other.val.set_val.vals = 0;
```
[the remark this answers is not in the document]
awelzel: > Is set_val assigned here just so that we reset the whole union to null?

Yes. The std::move() will just be a copy. The `_val()` is good. I think the proposed code isn't actually valid. Thanks!

src/threading/SerialTypes.cc
```diff
@@ -82,31 +82,88 @@ std::string Field::TypeName() const {
...
+    if ( this == &other )
+        return;
+
+    free_value(this);
+
+    type = other.type;
+    subtype = other.subtype;
+    present = other.present;
+
+    // Deal with "simple types".
```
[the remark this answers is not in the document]
awelzel: Thanks!

awelzel:
> Just smaller comments. I like how this is able to remove some of the manual memory mgmt, too.

Think I covered them all. This got better, too.

rsmmr (approved):

PR #3896, opened 2024-08-22
Analyzer: Do not add child analyzers when finished

awelzel:
Depending on an analyzer's implementation, its Done() method may
attempt to access analyzer or connection state when executing.
When this happens in the destructor of the parent analyzer during
the process of destructing a connection, this state may have been
deleted, resulting in use-after-free crashes or worse memory
corruption.

The following cases have been observed in the wild for when this happens.

  * PIA matching during Done() for undelivered TCP data enables a Spicy
    based analyzer which in turn attempts to raise an analyzer violation
    during Done()->EndOfData().

  * Spicy analyzers attaching new analyzers during their Done() processing
    which in turn attempt to use TCP() (to call FindChild()) during Done()
    while the analyzer tree / connection is being destructed.

The second scenario was previously found to happen in the HTTP analyzer
and fixed with 6ef9423f3cff13e6c73f97eb6a3a27d6f64cc320.

Plug these scenarios by short-circuiting AddChildAnalyzer() if the analyzer
or connection have finished or are being finished.

### DIFF
--- a/src/analyzer/Analyzer.cc
+++ b/src/analyzer/Analyzer.cc
@@ -113,19 +113,7 @@ void Analyzer::CtorInit(const zeek::Tag& arg_tag, Connection* arg_conn) {
 
 Analyzer::~Analyzer() {
     assert(finished);
-
-    // Make sure any late entries into the analyzer tree are handled (e.g.
-    // from some Done() implementation).
-    LOOP_OVER_GIVEN_CHILDREN(i, new_children) {
-        if ( ! (*i)->finished )
-            (*i)->Done();
-    }
-
-    // Deletion of new_children done in separate loop in case a Done()
-    // implementation tries to inspect analyzer tree w/ assumption that
-    // all analyzers are still valid.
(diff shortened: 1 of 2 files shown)


src/analyzer/Analyzer.cc
```diff
@@ -117,6 +117,8 @@ Analyzer::~Analyzer() {
     // Make sure any late entries into the analyzer tree are handled (e.g.
     // from some Done() implementation).
     LOOP_OVER_GIVEN_CHILDREN(i, new_children) {
+        reporter->InternalWarning("%s has new child %s in analyzer destructor", fmt_analyzer(this).c_str(),
+                                  fmt_analyzer(*i).c_str());
```
awelzel, line 121: Drop the whole block ,replace with `assert(new_children.empty())` after testing in production.

rsmmr (approved):

awelzel:
Merging now after @pbcullen has tested in their testing environment. Thanks!

PR #3897, opened 2024-08-22
TCP_Reassembler: Fix IsOrig() position in Match() call

awelzel:
Found during a debug session with @rsmmr.

### DIFF
--- a/src/analyzer/protocol/tcp/TCP_Reassembler.cc
+++ b/src/analyzer/protocol/tcp/TCP_Reassembler.cc
@@ -273,7 +273,10 @@ void TCP_Reassembler::MatchUndelivered(uint64_t up_to_seq, bool use_last_upper) 
         if ( b.upper > last_reassem_seq )
             break;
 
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/tcp/TCP_Reassembler.cc
```diff
@@ -273,7 +273,10 @@ void TCP_Reassembler::MatchUndelivered(uint64_t up_to_seq, bool use_last_upper)
         if ( b.upper > last_reassem_seq )
             break;
 
-        tcp_analyzer->Conn()->Match(zeek::detail::Rule::PAYLOAD, b.block, b.Size(), false, false, IsOrig(), false);
+        // Note: Even though this passes bol=false, at the point where
+        // this code runs, the matcher is re-initialized resulting in
+        // undelivered data implicitly being bol-anchored.
```
rsmmr, line 278:
```suggestion
        // undelivered data implicitly being bol-anchored. It's unclear if that was intended, but there's hardly a right way here, so that seems ok.
```

rsmmr (approved):

PR #3899, opened 2024-08-26
Improve packet lag calculation

J-Gras:
Using network_time to calculate packet lag will produce wrong results when there is no packet available but network time does not (yet) fall back to wall clock. This introduces `get_current_packet_ts()`. For the BiF I thought about returning a record with an optional timestamp field that isn't set in case there is no packet. However, I wasn't sure it would be wort the effort.


src/zeek.bif
```diff
@@ -3797,6 +3797,31 @@ function get_current_packet_header%(%) : raw_pkt_hdr
...
+## Returns: The currently processed packet's timestamp.
+##
+## .. zeek:see:: get_current_packet get_current_packet_header
+##
+## .. note::
+##
+##      For ``get_current_packet_ts()`` the same limitations as for
+##      :zeek:see:`get_current_packet` apply. In particular, the return value
+##      should be considered undefined when called within event handlers raised
+##      via :zeek:see:`event`, :zeek:see:`schedule` or by recipient of Broker
```
awelzel, line 3811: Hmm, yet, the user added is in `check_stats()` - that's a scheduled event, and the baselines verify the output of 0.0.

---

After looking a bit: `get_current_packet()` returns `pcap_packet` including timestamp information that should allow to build the timestamp in script land without a new BiF? I forget if there was another reason to have a separate BiF?

```

```

I think this comment could be relaxed, actually:

```
 ##      The return value of ``get_current_packet()`` further should be considered
 ##      undefined when called within event handlers raised via :zeek:see:`event`,
 ##      :zeek:see:`schedule` or by recipient of Broker messages.
```
`Packet::ToVal()` actually returns a zero-initialized structure if there wasn't a packet, I had missed that.

awelzel: Alternative proposal in DM was to add a BiF `get_packet_lag()` that handles it all internally.

ckreibich (changes requested):
Sorrry for the delay here — do we actually document anywhere why & when Zeek updates network time to current time? It's mentioned in the NEWS, but without doing so, I think users will have a hard time understanding `network_time()` vs `get_current_packet_ts()`.

If you're on board with adding that, I'm (mildly) in favor of `get_current_packet_ts()`, mainly because we already have `get_current_packet_header()` (i.e., variations of retrieving packet info), and because it strikes me as potentially more useful to have in other settings than only having `get_packet_lag()`.

I'd then just add `get_packet_lag()` as a script-layer function — i.e., I see no harm in providing both. Hardly a lot of code. :-)

J-Gras:
I've added `get_packet_lag()` as a script-layer function and provided some more details in the doc-strings of the new functions 🙂

awelzel (approved):

PR #3904, opened 2024-08-28
RunState: Drop broker_mgr->Active() usage

awelzel:
...and deprecate it. The communication_enabled flag doesn't really make a difference in how the IO loop works.


timwoj (approved):
There was a reason I added that way back when but heck if I can remember why. I vaguely recall the loop hanging if the only thing going on was broker communication, but no packet traffic. If that situation works after removing this, I see no reason to keep checking it.

PR #3907, opened 2024-08-30
Update c-ares to the latest release

timwoj:
It's been a while since we updated this submodule and in the meantime they've been very busy. They added a new memory-safe parsing architecture and a new event loop, and in the process deprecated about half of the existing API. This PR takes advantage of the former, but we're skipping the latter since we need access to the sockets for our IO loop. This is probably best-reviewed commit-by-commit since each is independent from the others.

I added a couple of TODOs to the TXT response handling. I believe it's possible to get multiple TXT responses from a single query, but we've never handled it before so they're mostly just informational notes. It could be an opportunity for future work that wouldn't be too hard, along with handling all of the alias names for a PTR request.


timwoj:
I bumped this to v1.34.2 and rebased it on master.

awelzel (commented):
Hey - I haven't tried to run, but I think there's some dangers with the new aliases handling? Could we drop that for now?

Do you have a DNS stress-test Zeek script available that does various different lookups? Maybe we should commit that to the code-base or leave a pointer so can execute it with an ASAN build.

awelzel (approved):

PR #3911, opened 2024-09-02
Bump clang-format

bbannier:
This patch contains a bump of the configured clang-format version from
17.0.3 to 18.1.8 and automatically generated C++ source updates. The
main difference we are seeing from this is fixes for previously
incomplete reformats.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -1645,36 +1645,34 @@ FILE* open_file(const string& path, const string& mode) {
     return rval;
 }
 
-TEST_CASE("util path ops"){
+TEST_CASE("util path ops") {
 #ifdef _MSC_VER
 // TODO: adapt these tests to Windows paths
 #else
-    SUBCASE("SafeDirname"){SafeDirname d("/this/is/a/path", false);
-CHECK(d.result == "/this/is/a");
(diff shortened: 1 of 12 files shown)


awelzel (approved):

src/util.cc
```diff
@@ -2303,8 +2301,7 @@ static void strerror_r_helper(char* result, char* buf, size_t buflen) {
         buf[buflen - 1] = 0;
 }
 
-static void strerror_r_helper(int result, char* buf, size_t buflen) { /* XSI flavor of strerror_r, no-op. */
-}
+static void strerror_r_helper(int result, char* buf, size_t buflen) { /* XSI flavor of strerror_r, no-op. */ }
```
awelzel, line 2304: That one, the above, and their use is pretty strange :-)

awelzel:
Generally would argue we shouldn't do bumps of formatters / reformats just after releasing an LTS (rather do them in the last dev cycle or before cutting the next LTS), but this seems trivial.

src/util.cc
bbannier: Yep, both the current code and the original commit fc33bf2014 are slightly light on details 😉

PR #3918, opened 2024-09-05
pre-commit: Add spicy-format for new Spicy parsers

awelzel:
This adds spicy-format and re-formats the smaller analyzers. QUIC, LDAP and Websocket are excluded and should be re-formatted during the 7.2 cycle or just before 8.0, so that cherry-picking fixes isn't harder than it should be.

#3914 and #3765 (IIUC) used spicy-format and once this pre-commit hook is merged, will automatically be format checked and formatted during `pre-commit` execution.

Thanks @bbannier !


src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -15,9 +15,9 @@ public type Message = unit {
     on %done {
         spicy::accept_input();
 
-        if ( self?.prio ) {
+        if (self?.prio) {
             self.severity = (self.prio.value & 0x07);
-            self.facility = (self.prio.value & 0x03f8) >> 3;
+            self.facility = (self.prio.value & 0x03f8)>>3;
```
awelzel, line 20: Seems worth a fix :-)

bbannier: Shift operators are currently unhandled and adding support for them would be great. The rule that would need to be expanded is https://github.com/bbannier/spicy-format/blob/8843f49a2291fb8f4d2f82724b4d76e1752659fd/src/query.scm#L196-L214

bbannier: https://github.com/bbannier/spicy-format/pull/127 has landed with spicy-format-0.15.0 so this should format like expected if you bump the version in `.pre-commit-config.yaml`.

bbannier (approved):

PR #3926, opened 2024-09-11
Pre-commit - remove spicy-format for the moment

0xxon:
This currently errors during compilation, because pre-commit does not pass --locked during compilation, without a way to enable it.

See https://github.com/pre-commit/pre-commit/issues/3162 for details.


bbannier:
Ultimately this is a design issue in `pre-commit`, but if we wanted we could work around this with https://github.com/bbannier/spicy-format/pull/133. Like I wrote there this might fail again spuriously until spicy-format's transitive tree-sitter dependency from https://github.com/tweag/topiary is bumped to tree-sitter-0.23.0 (which was supposed to fix these issues after all 🤦).

PR #3929, opened 2024-09-12
telemetry: Move callbacks to Zeek

awelzel:
Now that we run callbacks on the main loop, we can move callback support
for Counter and Gauge instances directly into Zeek and don't need to patch
prometheus-cpp anymore.

This pins prometheus-cpp back to 1.2.4

Needs to be done in `broker/auxil`, too!

### DIFF
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -557,42 +557,38 @@ void DNS_Mgr::InitPostScript() {
 
     cached_hosts_metric =
         telemetry_mgr->GaugeInstance("zeek", "dnsmgr_cache_entries", {{"type", "host"}},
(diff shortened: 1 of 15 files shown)


awelzel:
@timwoj , this breaks the callback API as it moves to `std::function<double()>` to avoid the `prometheus::ClientMetric` dependencies. It also moves type into detail so we can be more fancy in the future without feeling bad.

awelzel:
@timwoj suggested to add this to the 7.0.2 pile. It's a bit unusual and and API change, but will avoid callback implementors exposure to prometheus-cpp (and the API has only been there since the just released 7.0.0).

timwoj (approved):
Agreed with it going into 7.0.2.

PR #3931, opened 2024-09-12
teredo/gtpv1: Replace connection_state_remove with conn-removal-hooks

awelzel:
Running a 1mio syn packet pcap in baremode, the bifs `remove_teredo_connection()` and `remove_gtpv1_connection()` show up at about 1-2% in flamegraphs. That's because they are called via a `connection_state_remove()` for every connection, even for those not even UDP. Worse, this happens even if the Teredo or GTPv1 analyzer is disabled.

ZAM has even added [specialized ops for these two BiFs](https://github.com/zeek/zeek/blob/ba91de59b0b42126c2603eac9a27aa7cdc5aea75/src/script_opt/ZAM/OPs/ZBI.op#L3-L27).

Switch the cleanup from connection_state_remove() to a connection removal hook. This way it's only done when state was initialized for a give connection. In the case of the 1 mio syn packet pcap that's never.

|:---|---:|---:|---:|---:|
| `taskset -c 0 base/zeek -D -b -C -r  <...>/1mio-syn-packets.pcap Log::default_writer=Log::WRITER_NONE` | 9.320 ± 0.026 | 9.294 | 9.347 | 1.03 ± 0.01 |
| `taskset -c 0 zeek -D -b -C -r  <...>/1mio-syn-packets.pcap Log::default_writer=Log::WRITER_NONE` | 9.045 ± 0.043 | 8.997 | 9.075 | 1.00 |

As an aside, if we could just attache that information to the protocol analyzer tree, the we wouldn't need to do this, but I don't think we have a nice mechanism to do that.


vpax:
Nice, will note this for next month's script optimization maintenance

PR #3934, opened 2024-09-16
Fix `|...|` operator for inner ports/enums

### LINKED ISSUE #3835: Mismatch between `|...|` operator and what's returned for `enum`s and `port`s
The `|...|` operator [only allows for one of three possible yield types](https://github.com/zeek/zeek/blob/0ba80d13b49fc1d4603e67fc3b4807cc333376bc/src/Expr.cc#L1331-L1340), `any`, `double`, and `count`. However, [`port`](https://github.com/zeek/zeek/blob/0ba80d13b49fc1d4603e67fc3b4807cc333376bc/src/Val.cc#L597) and [`enum`](https://github.com/zeek/zeek/blob/0ba80d13b49fc1d4603e67fc3b4807cc333376bc/src/Val.cc#L3134) values return `int` types. That discrepancy causes the [attached program](https://github.com/user-attachments/files/16288493/size-count-val-discrepancy.zeek.txt) to crash with a low-level `ZVal` inconsistency error.

For `port` this is easy to fix because it really should be returning a `count` in all cases. However for `enum` it's not, because one of the `EnumType` constructors allows specifying a `zeek_int_t` value to associate with a new `enum` element. So probably the right fix here is to (1) make `port` return a `count`, and (2) extend `|...|` to recognize the `enum`s return `int`s.

---

```
type my_enum: enum {
	enum1
};

type Info: record {
	my_action: count &default = |enum1|;
} &redef;

event zeek_init()
        {
	local i: Info;
	print i$my_action;
        }
```

evantypanski:
Fixes #3835 

The `|...|` (sizeof) operator applies to ports/enums. That has returned a count, but it causes an internal error when used as an inner value in a record. This fixes that internal error for ports by just making the sizeof operator return a count.

Enums could technically be negative before this change, but that is rejected at parse time. It seems reasonable to modify the enum value to only be non-negative, which makes the sizeof operator easier since it can just return a count. Changing that to return an int would potentially break scripts that use the sizeof operator and assign it to a count.

This could technically break code that internally sets the enum value to a negative value, but I don't think that's a very likely use.

Everything here changes the enum value, but not the type, to use non-negative values. The type has a couple more nuances: since `Lookup` returns a negative value when not found, it's easier to keep that as an int to avoid narrowing conversions and extra checks


testing/btest/language/sizeof.zeek
```diff
@@ -80,6 +90,9 @@ print fmt("Double %s: %f", d, |d|);
 # Size of enum: returns numeric value of enum constant.
 print fmt("Enum %s: %d", ENUM3, |ENUM3|);
 
+# Within a record, enum sizeof should still be ok
+print fmt("Enum in record: %d", |with_enum$e|);
```
awelzel, line 94: I'm a bit lost, why we apply `||` here. Isn't the actual `&default` value more interesting? Suggest to add the non-sizeof values, too, minimally to show that they end up producing the same.

```suggestion
print fmt("Enum in record: %d %d", with_enum$e, |with_enum$e|);
```

src/Val.h
```diff
@@ -1497,20 +1496,20 @@ class RecordVal final : public Val, public notifier::detail::Modifiable {
     const std::vector<bool>& is_managed;
 };
 
-class EnumVal final : public detail::IntValImplementation {
+class EnumVal final : public detail::UnsignedValImplementation {
```
awelzel, line 1499: Hmm, not comfortable doing this. It's breaking the API.

Any user calling `AsInt()` on an `ValPtr` today is valid (one example is in EventTrace.cc, another in logging/Manager.cc and possibly some in external plugins). After this change `AsInt()` will cast to the wrong underlying implementation. Things may still happen to work out, but it's confusing, no? Also, would expect UBSAN to be unhappy (can label the PR fullci and re-push to see if it would fail).

Can we keep the IntValImplementation and cap enum values a 0x7FFFFFFF...?

Alternatively, as mentioned in the ticket, sizeof enum/port doesn't seem to be documented, so IMO, deprecation could be another direction to take.

awelzel (changes requested):
Don't think we can change the parent class of EnumVal just like that. This is in the public Zeek namespace and a breaking change without users realizing.

src/Val.h
evantypanski: It's fair to say that it may break some cases. But isn't it just as confusing that a value that is [enforced to be non-negative](https://github.com/zeek/zeek/blob/2907d9feeec4c107f7b728ff16ba1107287bd285/src/parse.y#L1144) is represented as a signed value? It makes a confusing case where an internal API can allow negative values technically, even though it's certainly not the intent. It's already assumed `EnumVal` will always be non-negative; making that concrete seems like the *less* confusing option.

I pushed the small test change and will see what fullci says here, but it seems like those cases can be easily fixed. More details on what to fix/how to fix it could go in the breaking changes section. If that's wrong then this can get scrapped easily.

Wouldn't it also be wrong to grab the enum value `AsInt` and then actually get a negative value? We already assume it won't be negative (if it's `-1` for example, then [`Lookup`'s result](https://github.com/zeek/zeek/blob/ba91de59b0b42126c2603eac9a27aa7cdc5aea75/src/Type.h#L831-L833) will be completely invalid). Wouldn't having a possibly-negative type introduce other subtle bugs? Wouldn't making `EnumVal` always unsigned solidify assumptions that were already made?

I don't really like deprecating it, if only because `SizeVal` is a [virtual method in `Val`](https://github.com/zeek/zeek/blob/ba91de59b0b42126c2603eac9a27aa7cdc5aea75/src/Val.h#L129-L131) so it makes sense to me to just implement it for all of them. But that's a pretty weak argument on my part

For what it's worth, the fix without the breaking change could just copy [int's version](https://github.com/zeek/zeek/blob/2907d9feeec4c107f7b728ff16ba1107287bd285/src/Val.cc#L216-L220) - the operator is also referred to as the "absolute value operator" for other types in the docs, so it makes more sense to me that it'd return an absolute value for enums, too.

Oh, also, regardless we should cap enum values at the max int value IMO. Or at least within an enum type. It still seems confusing to use `AsInt`, to me, since it can't be signed by design after that. But it'd at least help without introducing a breaking change (or as much of a breaking change, it could still potentially break something, technically). That still seems like it's trying to address the problem without actually addressing it, though

tl;dr: I'd be happy to just close this and make the simple change that's the fix for the problem without a bigger breaking change. I just also think this is an opportunity to make `EnumVal` more intuitive and less bug prone for users. I'd personally prefer making `EnumVal` express its non-negative reality.

awelzel: > I just also think this is an opportunity to make EnumVal more intuitive and less bug prone for users. I'd personally prefer making EnumVal express its non-negative reality.

I totally agree with your statements around making the underlying implementation unsigned. I'm solely coming from the angle that there are existing users calling `AsInt()` on a `ValPtr` and they now invoke undefined behavior (see the downcasting errors in the ubsan CI job for internal ones). And also, that we try to avoid changing the API in the public `zeek::` namespace without a deprecation cycle.

> but it seems like those cases can be easily fixed.

EnumVal lives in the public `zeek::` namespace (and is an established/old API). Cases and callers might live in code we do not control, have no access or even knowledge of. We usually try to deprecate and raise compilation 
warnings (via e.g. `[[deprecated]]`) for changes to the API for a whole LTS cycle (in this case it would be throughout 8.0) so users have time to adapt. I'm just not sure how that would/could look and whether it's worth it - maybe deprecating `AsInt()` on Val and implementing it just for `IntVal` (but not sure this works out or is great).

evantypanski: Well, that's fair enough. It really doesn't feel like anything is deprecated, so deprecating `AsInt()` for generic values seems like a pretty heavy-handed approach to get this one change in. I'll just close this and do the easy way that just directly fixes that issue. I was just trying to sneak in something to make this case cleaner while also solidifying some assumptions

For what it's worth, I still really like this change and would like to find a way to get the underlying enum value as unsigned 🙊  but I'll keep that on the back burner a little

PR #3946, opened 2024-09-19
Honor password logging

martinvanhensbergen:
The FTP policy allows adding extra commands to be logged, including the PASS command. When doing so, the actual plaintext password will be logged as argument of the command even though password_capture may be set to F.

With this change, if password capture is false, the password as part of the PASS argument will also be hidden.


scripts/base/protocols/ftp/main.zeek
```diff
@@ -169,9 +169,20 @@ function set_ftp_session(c: connection)
 		}
 	}
 
+function should_hide_password(s: Info) : bool
+	{
+	if ( ! s$capture_password &&
```
awelzel, line 174: You could condense to this to:
```
return ! s$capture_password && to_lower(s$user) !in guest_ids;
```

scripts/base/protocols/ftp/main.zeek
```diff
@@ -169,9 +169,20 @@ function set_ftp_session(c: connection)
...
+			{
+			return T;
+			}
+	return F;
+	}
+
 function ftp_message(c: connection)
 	{
 	if ( ! c?$ftp ) return;
+	local password_hidden_string:string = "<hidden>";
```
awelzel, line 185:
```suggestion
	local password_hidden_string = "<hidden>";
```

scripts/base/protocols/ftp/main.zeek
```diff
@@ -193,15 +204,20 @@ function ftp_message(c: connection)
 		s$reply_msg = s$reply_msg[:max_reply_msg_length];
 		}
 
+	# In case of logging PASS command, the password will appear in arg column and might need hiding
+	if ( s$command == "PASS" &&
+		 should_hide_password(s))
```
awelzel, line 209: Suggest to use a single line and remove the braces. Also below.

awelzel (changes requested):
Hey @martinvanhensbergen - thanks!

Have only nits - mind taking a look?

Would you be up for squashing some of the "Fix" commits before a merge? Otherwise can do so, too.

martinvanhensbergen:
Hi @awelzel  thanks for reviewing and the suggested changes. I will make the changes and squash and cleanup a little later this week!

martinvanhensbergen:
Hi @awelzel 

I have applied the changes in the code as you suggested. I was wondering if squashing is necessary as i'd say this is done automatically by github on merge?

Regards
Martin

awelzel:
@martinvanhensbergen  - thanks!
> I have applied the changes in the code as you suggested. I was wondering if squashing is necessary as i'd say this is done automatically by github on merge?

We don't use the GitHub merge button :-) If you don't want to squash yourself, can also do during the merge by hand.

martinvanhensbergen:
If you would be able to do the merging & squash then that would be much appreciated ❤️ Thanks for the code suggestions though; it looks a bit cleaner indeed :)

awelzel:
Merged with c36ae2f1a3e572676456dfb1230d776b9d0631bc.

PR #3949, opened 2024-09-24
Avoid duplicating warnings when reading table

### LINKED ISSUE #3692: Input manager on large files does not suppress a large quantity of errors
When a malformed large input file is loaded into Zeek, I'd expect the framework to limit the errors encountered so to not flood output, based on the info found under: [https://docs.zeek.org/en/master/frameworks/input.html#broken-input-data](url)

This doesn't seem to be the case, as the reporter.log file fills up with every error encountered while loading. On a table file containing almost a million entries, most of them faulty, an error is reported for each case resulting in a massive flood.

A simple reproduction can be shown, with an input file generated with:

`(echo '#fields\tindicator\tindicator_type\tmeta.source'; yes hello|head -n 1000) > intel.dat`

and a zeek script loading it:

`cat intel.zeek
redef Intel::read_files += {
	fmt("%s/intel.dat", @DIR)
};`

Running this script results in a flood of output:

```zeek intel.zeek 

warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 2: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1
warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 3: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1
warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 4: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1
warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 5: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1
warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 6: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1
warning: /private/tmp/./intel.dat/Input::READER_ASCII: /private/tmp/./intel.dat, line 7: Not enough fields in line 'hello' of /private/tmp/./intel.dat. Found 0 fields, want positions 1 and -1

...

```

I've attached the two sample intel files here, along with an additional table.zeek which calls the input framework directly to load a table on the intel.dat file. Both zeek scripts result in a reporter.log containing every error encountered.
[testFiles.zip](https://github.com/zeek/zeek/files/15029190/testFiles.zip)

evantypanski:
Fixes #3692

The `FailWarn` method can set warnings to get suppressed, but this was the *single* case within Zeek that did not suppress warnings on that call (via the default parameter). I don't see a reason for it not to suppress warnings in git blame anywhere, so I think it makes sense to just suppress warnings until a valid row is found. This won't occur if it should fail on invalid lines, since it'll just abort immediately after.

Related, that default parameter may be better defaulting to `true` since that's what most call sites within Zeek use, but probably not worth the deprecation and whatnot.


src/input/readers/ascii/Ascii.cc
```diff
@@ -345,7 +344,8 @@ bool Ascii::DoUpdate() {
                 FailWarn(fail_on_invalid_lines,
                          Fmt("Not enough fields in line '%s' of %s. Found "
                              "%d fields, want positions %d and %d",
-                             line.c_str(), fname.c_str(), pos, fit.position, fit.secondary_position));
+                             line.c_str(), fname.c_str(), pos, fit.position, fit.secondary_position),
+                         ! fail_on_invalid_lines);
```
awelzel, line 348: The `! fail_on_invalid_lines` doesn't make a difference, so passing `true` seems less cryptic to me. What do you think?

src/input/readers/ascii/Ascii.cc
```diff
@@ -395,8 +393,10 @@ bool Ascii::DoUpdate() {
             delete[] fields;
             continue;
         }
+        // If there's no error, then it makes sense to report the next error.
```
awelzel, line 396: If every other line is invalid this doesn't improve the situation much, but agree this works in the case where few or a lot of consecutive lines are broken.

I do wonder if it would make sense to count failures and output a summary on `StopWarningSuppression()`, a user might at least then have an indication if it was 3 lines failing, or 10000 lines - right now it's pretty noticeable spill of warnings.

```
warning: ..<...>/Input::READER_ASCII: ../input.log, Suppressed 120121 warnings"
```

awelzel (commented):
I wonder if we should do the same change for the config reader?

Also see the comment about how much was suppressed - could you see that as useful?

The sqlite one doesn't use FailWarn, but just bails on the first error IIUC (nothing to change)

src/input/readers/ascii/Ascii.cc
evantypanski: That decision was to avoid the case where failing on invalid lines would cause an error, then later a completely unrelated error gets suppressed. Since it aborts immediately it shouldn't make a difference for this function, but it may make a difference to the caller, theoretically.

Though, admittedly no where else seems to do anything like that, so it could just get changed back. I thought it would be nice to have conditionally (and do that sort of thing more).

As a side note, the backend just suppresses and unsuppresses warnings unconditionally, it would be nice to have some memory of `suppress_warnings` and reset it - so calling `DoUpdate()` won't always stop suppression. This is just one tiny case of trying to be a little more clever with suppression logic, but it's also not really necessary

src/input/readers/ascii/Ascii.cc
evantypanski: Seems nice to have, I'll experiment with this

evantypanski:
I added a couple things based on this:

1) Stopping warning suppression will now print how many warnings got suppressed (if above 0) - note this includes 1 warning, which seems a little weird, but I personally don't mind it
2) the config reader got switched to use `FailWarn` and suppress warnings. Seems like the warnings were made before suppression was an option, so it makes sense they should use it.

SQLite could get swapped as well, but obviously would require a little more effort because it does abort immediately, so meh.

I'm pretty iffy on the implementations here at least, but if they seem reasonable then I'm ok with them

src/input/readers/config/Config.cc
```diff
@@ -167,10 +166,12 @@ bool Config::DoUpdate() {
         return true;
     }
 
+    bool error = false;
```
awelzel, line 169: It does not seem like `error` is read? Can it be removed?

awelzel: Also, seeing the test change makes me think that for config, we should probably not suppress: There shouldn't be 1000s of lines in a config file and it's likely better for a user to see all the problematic ones.

Mind reverting? :-/

awelzel (approved):
> note this includes 1 warning, which seems a little weird, but I personally don't mind it

The resulting baseline changes for `missing-file-initially` did look a bit strange. I think the internal `OpenFile()` function shouldn't use `fail_onf_file_problem`, because that causes wrong behavior in `DoUpdate()` or `DoHeartbeat()` method.

```
diff --git a/src/input/readers/ascii/Ascii.cc b/src/input/readers/ascii/Ascii.cc
index 6d862fd1ca..0601af5b4a 100644
--- a/src/input/readers/ascii/Ascii.cc
+++ b/src/input/readers/ascii/Ascii.cc
@@ -139,14 +139,14 @@ bool Ascii::OpenFile() {
     if ( ! file.is_open() ) {
         FailWarn(fail_on_file_problem, Fmt("Init: cannot open %s", fname.c_str()), true);
 
-        return ! fail_on_file_problem;
+        return false;
     }
 
     if ( ReadHeader(false) == false ) {
         FailWarn(fail_on_file_problem, Fmt("Init: cannot open %s; problem reading file header", fname.c_str()), true);
 
         file.close();
-        return ! fail_on_file_problem;
+        return false;
     }
 
     if ( ! read_location ) {

```

With this applied and the config reader changes reverted, I think that's good to go :rocket:

src/input/readers/config/Config.cc
evantypanski: Agreed :) and leaving `error` around was an oopsie.

PR #3950, opened 2024-09-24
Add numeric field to conn.log for IP protocol

### LINKED ISSUE #3915: sessions for non-analyzed protocols should still be tracked with a conn entry
Zeek's current behavior of silently ignoring traffic for which there is no analyzer is quite unexpected to new users. The assumption is that the conn log should contain data similar netflow or firewall logs.

There are many protocols where the mere presence of the protocol is a very interesting finding even if we aren't interested in protocol details. For example, is someone else's equipment trying to speak OSPF to your routers? Is there IP traffic using an uncommon or unregistered protocol number on your network? How much traffic is going over IPSEC connections on your network? Currently, zeek is not providing visibility into these behaviors in the conn log.

I would like to propose that at a minimum we create conn entries for all IP protocols with a minimum of a 3-tuple (orig_h, resp_h, and proto) and use similar logic to ICMP to determine when a connection has ended (ie. timeout). This should be enough to bring the conn log to parity with firewall logs with respect to traffic other than TCP, UDP, and ICMP.

timwoj:
Fixes #3915 

This PR does the following:

- Add a new field to conn.log called proto_id that has the numeric protocol identifier
- Leave the port fields set to zero for unhandled protocols
- Leave the proto field set to unknown_transport
- Add a policy file that you can load to get an additional column with the string version of the protocol

One thought I had here is that we're effectively adding the value twice. We add a `proto` field to the conn tuple record and then we also add the `protocol_id` field to the conn record. The `proto` field in the tuple isn't logged by default as part of the conn record. Maybe that could change and we could remove the `protocol_id` field instead. This is especially noticeable in the changes to the sqlite btest, where I had to add both values even though they're the same.


scripts/base/packet-protocols/ip/main.zeek
```diff
@@ -1,5 +1,10 @@
 module PacketAnalyzer::IP;
 
+export {
+	## Default analyzer
+	const default_analyzer: PacketAnalyzer::Tag = PacketAnalyzer::ANALYZER_UNKNOWN_IP_TRANSPORT &redef;
```
ckreibich, line 5: (Same naming concern here)

scripts/base/protocols/conn/main.zeek
```diff
@@ -158,6 +158,9 @@ export {
 		## *uid* values for any encapsulating parent connections
 		## used over the lifetime of this inner connection.
 		tunnel_parents: set[string] &log &optional;
+		## The numeric identifier for the transport protocol for this
+		## connection.
```
ckreibich, line 162: This needs more detail to explain the relationship to `proto`, and I would suggest `ip_proto` for the name.

scripts/policy/protocols/dhcp/software.zeek
```diff
@@ -45,7 +45,8 @@ event DHCP::log_dhcp(rec: DHCP::Info)
 		local id: conn_id = [$orig_h=rec$assigned_addr,
 		                     $orig_p=rec$client_port,
 		                     $resp_h=rec$server_addr,
-		                     $resp_p=rec$server_port];
+		                     $resp_p=rec$server_port,
+				     $proto=17]; # DHCP is typically UDP
```
ckreibich, line 49: Did we consider defaulting the `proto` field in `conn_id` to something (e.g. whatever we agree to use for an unknown value), to avoid breaking any code that instantiates a `conn_id` record explicitly? Or did we conclude that it's sufficiently rare?

src/Conn.h
```diff
@@ -60,7 +60,7 @@ struct ConnTuple {
     uint32_t src_port = 0;
     uint32_t dst_port = 0;
     bool is_one_way = false; // if true, don't canonicalize order
-    TransportProto proto = TRANSPORT_UNKNOWN;
+    uint8_t proto;
```
ckreibich, line 63: `uint8_t` makes sense from the IP header's perspective, but my `/etc/protocols` shows 262 for multipath TCP, though with the caveat that it's only used internally by the kernel. It likely doesn't matter for our use case, and we could broaden later as needed, but I wanted to flag it.

src/IPAddr.cc
```diff
@@ -60,26 +60,29 @@ ConnKey::ConnKey(Val* v) {
...
     else {
         // While it's not a conn_id, it may have equivalent fields.
         orig_h = vr->FieldOffset("orig_h");
         resp_h = vr->FieldOffset("resp_h");
         orig_p = vr->FieldOffset("orig_p");
         resp_p = vr->FieldOffset("resp_p");
+        proto = vr->FieldOffset("proto");
 
-        if ( orig_h < 0 || resp_h < 0 || orig_p < 0 || resp_p < 0 ) {
+        if ( orig_h < 0 || resp_h < 0 || orig_p < 0 || resp_p < 0 || proto < 0 ) {
```
ckreibich, line 80: `proto` is a `uint8_t` so less than zero cannot happen

src/TunnelEncapsulation.h
```diff
@@ -46,6 +46,7 @@ class EncapsulatingConn {
           src_port(0),
           dst_port(0),
           proto(TRANSPORT_UNKNOWN),
+          proto_id(255),
```
ckreibich, line 49: We used 256 earlier, more reason to alias it to a constant signifying "unknown".

ckreibich (changes requested):
Apologies for taking so long to pick this up, folks. I don't think it's ready quite yet, per my review comments, but I am still in agreement that the new protocol value is the way to go, and in the conn.log.

We should add a policy script to filter conn.log back to the previous form, to give people an out in case we're overlooking something. I've not tried but I believe the logging framework provides all the primitives needed.

I'll tag the PR as needing docs since we'll need to explain this one in a fair bit of detail.

scripts/base/init-bare.zeek
```diff
@@ -1300,21 +1301,27 @@ const non_analyzed_lifetime = 0 secs &redef;
...
 ##
-## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout set_inactivity_timeout
+## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout unknown_ip_inactivity_timeout set_inactivity_timeout
 const icmp_inactivity_timeout = 1 min &redef;
 
+## If a flow with an unknown IP-based protocol is inactive, time it out after
+## this interval. If 0 secs, then don't time it out.
+##
+## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout icmp_inactivity_timeout set_inactivity_timeout
+const unknown_ip_inactivity_timeout = 1 min &redef;
```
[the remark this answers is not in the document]
timwoj: More that it tracks with the udp, tcp, and icmp values that already exist.

src/Conn.h
timwoj: Good point, bumped it up to a uin16_t and set the "unknown" value to 65535.

src/TunnelEncapsulation.h
timwoj: Added a constant to IP.h, used it in Conn.h and TunnelEncapsulation.h.

scripts/policy/protocols/conn/remove-ip-proto-log.zeek
```diff
@@ -0,0 +1,17 @@
+#! This script filters the ip_proto field out of the conn.log.
+
+@load base/protocols/conn
+
+module Conn;
+
+event zeek_init() &priority=-100
+	{
+	local filter = Log::get_filter(Conn::LOG, "default");
```
awelzel, line 9: We could also do the following:

```
redef Conn::Info$ip_proto -= { &log }
```

This would hide `ip_proto` from any logic that computes/creates conn.log schmeas based on the presence of `&log`. Also works for all filters rather than just the default filter.

Should we also  disable tracking of unknown/unsupported protocols here. If ip_proto isn't logged, maybe we shouldn't even log and track such connections, so they are fully excluded from conn.log?

```
redef PacketAnalyzer::IP::default_analyzer = NONE;  # or whatever works
# or maybe a BifConst that is checked in the unknown ip analyzer instead.
redef PacketAnalyzer::IP::track_unknown|unsupported_connections = F;
```

ckreibich: Yes please, to both — at last a use-case for that `&log` redef! :-) This is also nicer than iterating over all currently defined filters.

The effect should be as if that whole feature were not there, which means we should not just suppress the column, but also avoid the additional connections. I initially thought of just filtering via a policy hook, but disabling the analyzer (or part of it) would be much nicer.

timwoj: I assume this should also disable the protocol name policy if someone opted to load both policies?

timwoj: I don't really like the idea of setting it to `NONE` because someone else might have redef'd it already, and our setting it to `NONE` would break that. I'll opt for the variable but that still means we're calling an analyzer to do no work.

Re: loading both policies, given there's no way to conditionally redef a record at startup, I think we just live with the fact that someone can do that. If they load the names policy and the removal policy, they won't get connections with unknown protocols in the log but they'll still get the `ip_proto_name` field that should hopefully match the `proto` field.

timwoj: > I decided that this PR is the one that sets it to something other than NONE, so setting it back to that is fine.

There's actually no good way to unset a default analyzer. Since we're passing a Tag, there isn't something you can pass that indicates "ignore this". A "none" tag would have to be added somewhere. I opted for just calling `disable_analyzer` on the unknown IP analyzer instead.

> 
> Re: loading both policies, given there's no way to conditionally redef a record at startup, I think we just live with the fact that someone can do that. If they load the names policy and the removal policy, they won't get connections with unknown protocols in the log but they'll still get the `ip_proto_name` field that should hopefully match the `proto` field.

I added some extra docs to the protocol strings policy script about this. I'm also considering renaming that script to `ip_proto_strings.zeek` since the current name isn't descriptive enough.

scripts/test-all-policy.zeek
```diff
@@ -95,6 +95,8 @@
 @load protocols/conn/known-hosts.zeek
 @load protocols/conn/known-services.zeek
 @load protocols/conn/mac-logging.zeek
+@load protocols/conn/protocol-strings.zeek
+@load protocols/conn/remove-ip-proto-log.zeek
```
awelzel, line 99: Another nit, given it's more than just logging now, maybe `disable-unknown-ip-proto-support.zeek`.

I'm thinking of the NEWS entry: To track only supported/known IP protocols, load ``protocols/conn/disable-unknown-ip-protocol-support.zeek`` (or something nicer/shorter.

scripts/policy/protocols/conn/protocol-strings.zeek
```diff
@@ -0,0 +1,169 @@
...
+	[137] = "mpls-in-ip",
+	[138] = "manet",
+	[139] = "hip",
+	[140] = "shim6",
+	[141] = "wesp",
+	[142] = "rohc",
+	[143] = "ethernet",
+	[144] = "aggfrag",
+	[145] = "nsh"
+};
```
awelzel, line 161:
```suggestion
} &redef;
```

And wonder if this could go into some consts in base? Then it's easier re-usable by someone that doesn't want to change conn.log, but still make use of the consts?

scripts/policy/protocols/conn/protocol-strings.zeek
```diff
@@ -0,0 +1,169 @@
...
+
+event connection_state_remove(c: connection) {
```
awelzel, line 163: Maybe `new_connection()` with similar motivation as #4021

src/communityid.bif
```diff
@@ -49,8 +49,8 @@ function community_id_v1%(cid: conn_id, seed: count &default=0, do_base64: bool
 
             break;
         case TRANSPORT_UNKNOWN:
-            emit_builtin_error("CommunityID: unknown transport layer", cid);
-            return zeek::make_intrusive<zeek::StringVal>("");
+            hash_proto = ntohs(orig_port->Port());
```
awelzel, line 52: Is this left-over from the initial version using orig_port->Port() to encode the proto? Could use cid's proto field now?

I think we also need to ignore the ports in the `digest_update()` calls below according to the [spec](https://github.com/corelight/community-id-spec).

testing/btest/Baseline/bifs.community_id.v1/out
```diff
@@ -1,11 +1,11 @@
...
+PASS: expected '1:HhA1B+6CoLbiKPEs5nhNYN4XWfk=', got '1:HhA1B+6CoLbiKPEs5nhNYN4XWfk=' ([orig_h=1.2.3.4, orig_p=1122/tcp, resp_h=5.6.7.8, resp_p=3344/tcp, proto=6], seed=1)
+PASS: expected '1:OShq+iKDAMVouh/4bMxB9Sz4amw=', got '1:OShq+iKDAMVouh/4bMxB9Sz4amw=' ([orig_h=1.2.3.4, orig_p=1122/udp, resp_h=5.6.7.8, resp_p=3344/udp, proto=17], seed=1)
+PASS: expected '1:9pr4ZGTICiuZoIh90RRYE2RyXpU=', got '1:9pr4ZGTICiuZoIh90RRYE2RyXpU=' ([orig_h=1.2.3.4, orig_p=8/icmp, resp_h=5.6.7.8, resp_p=0/icmp, proto=1], seed=1)
+PASS: expected '1:IO27GQzPuCtNnwFvjWALMHu5tJE=', got '1:IO27GQzPuCtNnwFvjWALMHu5tJE=' ([orig_h=fe80:1:203:405:607:809:a0b:c0d, orig_p=128/icmp, resp_h=fe80:1011:1213:1415:1617:1819:1a1b:1c1d, resp_p=129/icmp, proto=1], seed=1)
+FAIL: expected '', got '1:sryUGlncUGOU9hpA1tO/7SLUdWU=' ([orig_h=1.2.3.4, orig_p=0/unknown, resp_h=5.6.7.8, resp_p=0/unknown, proto=65535], seed=0)
+FAIL: expected '', got '1:lSXdOyOzJ1psY9gnKXzWdOBI6uo=' ([orig_h=fe80:1:203:405:607:809:a0b:c0d, orig_p=0/unknown, resp_h=fe80:1011:1213:1415:1617:1819:1a1b:1c1d, resp_p=0/unknown, proto=65535], seed=1)
```
awelzel, line 11: Hmm... I guess if `proto==65535`, then it should not return anything. Might be useful to add some tests that use a valid but not supported proto.

scripts/test-all-policy.zeek
```diff
@@ -95,6 +95,8 @@
 @load protocols/conn/known-hosts.zeek
 @load protocols/conn/known-services.zeek
 @load protocols/conn/mac-logging.zeek
+@load protocols/conn/protocol-strings.zeek
```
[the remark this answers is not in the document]
timwoj: Yep, mentioned this in https://github.com/zeek/zeek/pull/3950#discussion_r1837442733

src/communityid.bif
timwoj: > Is this left-over from the initial version using orig_port->Port() to encode the proto? Could use cid's proto field now?

Fixing this fixed the `FAIL`s you mentioned below.

> I think we also need to ignore the ports in the digest_update() calls below according to the [spec](https://github.com/corelight/community-id-spec).

Agreed that it's not matching the spec, but I'll fix that separate from this PR as it's out of scope.

ckreibich: The new approach [here](https://github.com/zeek/zeek/pull/3950/files#diff-788f371ac608aa4080cb2be2d0e6374b34a69b10482f8c4c95ca14322d7f0eecR59) casting down `cid_proto` to 8 bits strikes me as right. We have to keep the algorithm compatible because it's not just about old vs new IDs, it's also about IDs computed elsewhere on the same traffic. Also, the 8-bit value will be correct — I can't imagine Community ID will ever encounter in-kernel values like that 262 for MPTCP.

scripts/base/protocols/conn/main.zeek
```diff
@@ -158,6 +158,11 @@ export {
 		## *uid* values for any encapsulating parent connections
 		## used over the lifetime of this inner connection.
 		tunnel_parents: set[string] &log &optional;
+		## For IP-based connections, this contains the protocol
+		## identifier passed in the IP header. This is different
+		## from the ``proto`` field in that this value comes
```
ckreibich, line 163: Nit: to render right this has to be `*proto*`, ReST style. (See e.g. `conn_state` higher up.)

scripts/base/protocols/conn/main.zeek
```diff
@@ -281,6 +286,8 @@ function set_conn(c: connection, eoc: bool)
 		if ( c$history != "" )
 			c$conn$history=c$history;
 		}
+
+	c$conn$ip_proto = c$id$proto;
```
ckreibich, line 290: `c$id$proto` uses 65535 as default, while `c$conn$ip_proto` is optional, so should this be protected via an `if ( c$id$proto != 65535 )`?

scripts/policy/protocols/conn/disable-unknown-ip-proto-support.zeek
```diff
@@ -0,0 +1,11 @@
...
+##! logging of connections with unknown IP protocols.
+
+@load base/protocols/conn
+@load base/frameworks/analyzer/main
+
+redef record Conn::Info$ip_proto -= { &log };
+
+event zeek_init() {
+	Analyzer::disable_analyzer(PacketAnalyzer::ANALYZER_UNKNOWN_IP_TRANSPORT);
+}
```
ckreibich, line 11: Total nit but it'd be nice to end with a newline.

scripts/policy/protocols/conn/ip-proto-name-logging.zeek
```diff
@@ -0,0 +1,171 @@
...
+	[137] = "mpls-in-ip",
+	[138] = "manet",
+	[139] = "hip",
+	[140] = "shim6",
+	[141] = "wesp",
+	[142] = "rohc",
+	[143] = "ethernet",
+	[144] = "aggfrag",
+	[145] = "nsh"
+};
```
ckreibich, line 161: https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml also shows something called "Homa" for 146, fwiw.

scripts/policy/protocols/conn/ip-proto-name-logging.zeek
```diff
@@ -0,0 +1,171 @@
...
+
+event new_connection(c: connection) &priority=5 {
+	# In case we're the first access
+	Conn::set_conn(c, F);
+	if ( c$conn?$ip_proto && c$conn$ip_proto in protocol_names ) {
+		c$conn$ip_proto_name = protocol_names[c$conn$ip_proto];
+	} else {
+		c$conn$ip_proto_name = "unknown";
```
ckreibich, line 169: `ip_proto_name` is optional, so just don't assign in the unknown case?

src/IPAddr.cc
```diff
@@ -60,26 +60,29 @@ ConnKey::ConnKey(Val* v) {
...
     }
     else {
         // While it's not a conn_id, it may have equivalent fields.
         orig_h = vr->FieldOffset("orig_h");
         resp_h = vr->FieldOffset("resp_h");
         orig_p = vr->FieldOffset("orig_p");
         resp_p = vr->FieldOffset("resp_p");
+        proto = vr->FieldOffset("proto");
 
         if ( orig_h < 0 || resp_h < 0 || orig_p < 0 || resp_p < 0 ) {
```
ckreibich, line 80: Should this cover `proto < 0` too?

src/TunnelEncapsulation.cc
```diff
@@ -14,6 +14,7 @@ EncapsulatingConn::EncapsulatingConn(Connection* c, BifEnum::Tunnel::Type t)
       src_port(c->OrigPort()),
       dst_port(c->RespPort()),
       proto(c->ConnTransport()),
+      proto_id(c->KeyProto()),
```
ckreibich, line 17: In the other places we named the field `ip_proto` when living alongside an existing `proto` field, so I'd suggest doing this here too.

src/zeek.bif
```diff
@@ -315,6 +315,7 @@ zeek::RecordValPtr zeek::detail::build_dummy_conn_record()
 	id_val->Assign(1, zeek::val_mgr->Port(ntohs(0), TRANSPORT_UDP));
 	id_val->Assign(2, zeek::make_intrusive<zeek::AddrVal>((unsigned int) 0));
 	id_val->Assign(3, zeek::val_mgr->Port(ntohs(0), TRANSPORT_UDP));
+    id_val->Assign(4, 255);
```
ckreibich, line 318: Indentation buglet. Also, since this `&defaults` to 65535, do we need this?

(I wonder why once upon a time we chose `TRANSPORT_UDP` in the above, not `TRANSPORT_UNKNOWN`, odd.)

scripts/base/init-bare.zeek
```diff
@@ -213,11 +213,12 @@ type transport_proto: enum {
...
-	orig_h: addr;	##< The originator's IP address.
-	orig_p: port;	##< The originator's port number.
-	resp_h: addr;	##< The responder's IP address.
-	resp_p: port;	##< The responder's port number.
-} &log;
+	orig_h: addr &log;	##< The originator's IP address.
+	orig_p: port &log;	##< The originator's port number.
+	resp_h: addr &log;	##< The responder's IP address.
+	resp_p: port &log;	##< The responder's port number.
+	proto: count &default=65535;  ##< The transport protocol ID. Defaults to 65535 as an "unknown" value.
```
ckreibich, line 220: I'm tempted to add a script-layer `UNKNOWN_IP_PROTO` constant to mirror the core, but I think it doesn't matter much — the default will mean that users generally won't use the value explicitly. But it would help avoid proliferating the literal. Feel free to handle as you see fit.

ckreibich (changes requested):
This all looks great to me Tim, I just have a few low-level things.

Oh and one Spicy-related comment: I see a [conn_id() function](https://github.com/zeek/zeek/blob/master/scripts/spicy/zeek.spicy#L40) in spicy.zeek, which follows the current rationale of ports conveying protocols. I haven't looked into whether it also needs updating. I suspect it could happen in a follow-up PR if needed at all.

src/IPAddr.cc
timwoj: `proto` is unsigned, so it can't ever be < 0.

src/zeek.bif
timwoj: > (I wonder why once upon a time we chose `TRANSPORT_UDP` in the above, not `TRANSPORT_UNKNOWN`, odd.)

I hadn't noticed that. I wonder if this should be set to `IPPROTO_UDP` then?

scripts/policy/protocols/conn/protocol-strings.zeek
timwoj: I moved this up to init-bare into a new `IP` module. The `Conn` module didn't feel like the right place for it if it's being made more generic. I'll add a test for the `&default` tomorrow.

scripts/base/init-bare.zeek
ckreibich: I like it!

src/zeek.bif
ckreibich: I wondered too. But the only use for this is as dummy return value when `lookup_connection()` fails. I don't think that's supposed to look like a UDP flow. Most existing uses of `lookup_connection()` don't actually anticipate that it can fail. So all in all I'd still lean toward leaving it out.

scripts/base/init-bare.zeek
timwoj: I'm actually going to leave this one out. It results in some cascading `#include` issues in C++ (Conn.h requires const.bif.netvar_h which requires Val.h...) which end up being more changes than it's worth.

src/IPAddr.cc
ckreibich: Actually Tim, I misread the code higher up. None of this is about the value of the proto, it's about checking the RecordVal's fields. In line 63, proto should also be an int, and the `FieldOffset()` check can then theoretically be negative for it. And we're assigning 4 above, not a proto value.

timwoj: Fixed

ckreibich (approved):
Alright Tim! :rocket:

PR #3951, opened 2024-09-25
Integrate new Broker auto-disconnecting feature

Neverlord:
Zeek-side PR for https://github.com/zeek/broker/pull/429

### DIFF
--- a/scripts/base/frameworks/broker/main.zeek
+++ b/scripts/base/frameworks/broker/main.zeek
@@ -86,6 +86,24 @@ export {
 	## ZEEK_BROKER_MAX_THREADS environment variable overrides this setting.
 	const max_threads = 1 &redef;
 
+	## Max number of items we buffer at most per peer. What action to take when
+	## the buffer reaches its maximum size is determined by
+	## `peer_overflow_policy`.
+	const peer_buffer_size = 2048 &redef;
+
+	## Configures how Broker responds to peers that cannot keep up with the
+	## incoming message rate. Available strategies:
+	## - disconnect: drop the connection to the unresponsive peer
+	## - drop_newest: replace the newest message in the buffer
+	## - drop_oldest: removed the olsted message from the buffer, then append
+	const peer_overflow_policy = "disconnect" &redef;
+
+	## Same as `peer_buffer_size` but for WebSocket clients.
+	const web_socket_buffer_size = 512 &redef;
+
+	## Same as `peer_overflow_policy` but for WebSocket clients.
+	const web_socket_overflow_policy = "disconnect" &redef;
+
 	## The CAF scheduling policy to use.  Available options are "sharing" and
 	## "stealing".  The "sharing" policy uses a single, global work queue along
 	## with mutex and condition variable used for accessing it, which may be
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -294,6 +294,36 @@ void Manager::InitPostScript() {
     options.disable_forwarding = ! get_option("Broker::forward_messages")->AsBool();
     options.use_real_time = use_real_time;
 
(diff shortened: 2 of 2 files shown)


scripts/base/frameworks/broker/main.zeek
```diff
@@ -86,6 +86,24 @@ export {
 	## ZEEK_BROKER_MAX_THREADS environment variable overrides this setting.
 	const max_threads = 1 &redef;
 
+    ## Max number of items we buffer at most per peer. What action to take when
+	## the buffer reaches its maximum size is determined by
```
ckreibich, line 90: Just some indentation nits here.

scripts/base/frameworks/broker/main.zeek
```diff
@@ -86,6 +86,24 @@ export {
...
+	## `peer_overflow_policy`.
+	const peer_buffer_size = 2048 &redef;
+
+    ## Configures how Broker responds to peers that cannot keep up with the
+    ## incoming message rate. Available strategies:
+	## - disconnect: drop the connection to the unresponsive peer
```
ckreibich, line 96: Dominik does this then transparently trigger a `Broker::peer_lost` event, ideally with a message that indicates the cause? I think so, since the Broker manager should see the status change, but I wanted to check. This'll be a good thing to cover in cluster.log / broker.log, and it's the part where we need to see whether the cluster heals when this happens.

scripts/base/frameworks/broker/main.zeek
```diff
@@ -86,6 +86,24 @@ export {
...
+	## - drop_newest: replace the newest message in the buffer
+	## - drop_oldest: removed the olsted message from the buffer, then append
+    const peer_overflow_policy = "disconnect" &redef;
```
ckreibich, line 99: I wonder whether we should provide some tooling to help people understand their current buffer use. We have some of this in telemetry, but it might help to have a policy script that periodically reports buffer occupancy to broker.log, or some such. (Needn't happen in this PR)

ckreibich (commented):
Cool that the Zeek-side changes are so small for this. Just a few initial comments, I haven't looked at the Broker side yet.

scripts/base/frameworks/broker/main.zeek
Neverlord: This will trigger a `peer_removed` event. Listening on that event is how I've implemented a test for this: https://github.com/zeek/broker/blob/issue/gh-426/tests/btest/peering/disconnect-on-overload.cc (I plan to use that test program in a btest eventually, but it can be used manually already).

I don't think it'll have a dedicated error message, though. That's something we would need to figure out as a followup and/or use the new Broker logger integration for.

scripts/base/frameworks/broker/main.zeek
Neverlord: That new buffer is per peer. We currently only have a metric for the entire chain. We could (should?) also measure how much we buffer per peer, but that metric would have to use the Broker endpoint ID as a label dimension to be meaningful - and I don't know whether that'd be useful to the user. If we wrap it into a script that "resolves" the endpoint IDs to node names maybe? That's probably a separate discussion, though. :)

scripts/base/frameworks/broker/main.zeek
Neverlord: Sorry, tabs vs. spaces fail.

ckreibich:
I've nudged CI again on this one Dominik to hopefully get some clarity on the failures. I haven't yet run it locally — more soon. :crossed_fingers:

ckreibich:
@Neverlord let me close this in favor of #3981, which has all of the work in this branch but adds Zeek-level handling of the Broker events.

PR #3952, opened 2024-09-26
zeek-setup: Support enabling debug streams through env variable

awelzel:
For debugging btests, it can be convenient to enable debug streams by setting an environment variable rather than editing zeek invocations and adding -B selectively.

Sample use case:

    $ export ZEEK_DEBUG_LOG_STREAMS=all
    $ btest -d core/failing-test.zeek
    $ less .tmp/core/failing-test/debug.log

Zeek's -B option and ZEEK_DEBUG_LOG_STREAMS are additive.

### DIFF
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -556,9 +556,13 @@ SetupResult setup(int argc, char** argv, Options* zopts) {
         reporter->FatalError("-u incompatible with --no-unused-warnings");
 
 #ifdef DEBUG
-    if ( options.debug_log_streams ) {
+    if ( options.debug_log_streams )
         debug_logger.EnableStreams(options.debug_log_streams->data());
 
+    if ( const auto* streams = getenv("ZEEK_DEBUG_LOG_STREAMS") )
+        debug_logger.EnableStreams(streams);
+
+    if ( debug_logger.HasEnabledStreams() ) {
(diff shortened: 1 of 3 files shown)


src/zeek-setup.cc
```diff
@@ -556,9 +556,13 @@ SetupResult setup(int argc, char** argv, Options* zopts) {
         reporter->FatalError("-u incompatible with --no-unused-warnings");
 
 #ifdef DEBUG
-    if ( options.debug_log_streams ) {
+    if ( options.debug_log_streams )
         debug_logger.EnableStreams(options.debug_log_streams->data());
 
+    if ( getenv("ZEEK_DEBUG_LOG_STREAMS") )
+        debug_logger.EnableStreams(getenv("ZEEK_DEBUG_LOG_STREAMS"));
```
bbannier, line 563: Nit: can reuse previous the result:

```suggestion
    if ( auto streams = getenv("ZEEK_DEBUG_LOG_STREAMS") )
        debug_logger.EnableStreams(streams);
```

bbannier (approved):
Looks consistent to me.

I was first wondering whether this was a good approach at all since by definition `btest.cfg` must not reset the value which could unexpectedly (?) cause tests to behave differently depending on whether this env var is set or not (which is exactly the use case behind this). That's already an issue with `ZEEK_DEBUG_LOG_STDERR` though which seems similar, so I guess this all makes sense.

PR #3953, opened 2024-09-27

bbannier:


rsmmr (approved):

awelzel:
@bbannier  / @rsmmr - this comes a late, but looking at the OS-perf history, this bump caused the pcap-quic-16-50mb test to run some 17% slower (3.13sec increased to 3.68sec). The pcap-quic-12k also got slower, but only 3-4%. Feels regression enough to take a closer look?

Screenshot from 2024-10-23 11-15-01

PR #3954, opened 2024-09-27
Install `procps` in OpenSuse Leap images

bbannier:
The tests `core.sigterm-regular` and `core.sigterm-stdin` rely on `ps` to be present which is not the case anymore on OpenSuse Leap; install it explicitly there.


awelzel (approved):
Thanks!

PR #3959, opened 2024-10-02
input/Raw: Rework GetLine()

### LINKED ISSUE #3957: Input Raw: 100% CPU usage when line is not immediately terminated with \n
A user reported 100% CPU usage of a ReaderBackend thread when executing a long-running process and reading its stdout/stderr process. I think the following reproduces it well:

```
#!/bin/sh
#
echo -n "aaa"
sleep 0.55
echo -n "aaa"
sleep 0.55
echo  "aaa"

sleep infinity
```
This isn't great, there's nothing special about this program except that "aaaaaaaaa\n" is produced "slowly". That seems to knock the input thread into an endless loop. There's also reallocation with every loop iteration (4096 bytes) which shouldn't be needed either.

```

type OneLine: record { s: string; is_stderr: bool; };

event one_line(d: Input::EventDescription, tpe: Input::Event, l: string, is_stderr: bool) {
        print "one_line", is_stderr, l;
}

event zeek_init() {
        Input::add_event([
                $name="run",
                $source="./run.sh |",
                $reader=Input::READER_RAW,
                $fields=OneLine,
                $mode=Input::STREAM,
                $ev=one_line,
                $want_record=F,
                $config=table(["read_stderr"] = "1"),
        ]);
}
```

I presume we should yield from `GetLine()` with no data and resume reading later.

awelzel:
This isn't a straightforward fix, unfortunately. The existing GetLine() implementation didn't deal well with input that's incrementally produced where individually read chunks wouldn't end with the separator.

The prior implementation increased the buffer each time it failed to find a separator in the current buffer, but then also ended up not searching the full new buffer size for the terminator, doing that endlessly.

This change reworks the Raw reader to rely only on bufpos for reading and searching purposes and skip reallocation if the buffer size if it wasn't actually exhausted.

Closes #3957


ckreibich (commented):
Yikes, this whole class is terrifying. I've made it through your fix Arne and it looks good to me, but I still need to look over the tests. I'll return to this tomorrow and will make sure this gets in soon in case this runs into next week.

I wonder whether it would in the end be simpler to base the approach on `std::getline()` to offload buffering chores, and work around the fact that it only supports a character delimiter by checking the continuation when that character is found — a bit like the comment in 440. Haven't tried though.

awelzel:
>  to offload buffering chores,

Using a `std::string` instead of manual buffer management might also help enough. @bbannier pointed that out, too. I didn't go there as currently the threading values directly use the allocated buffer. While with `std::string` would require an extra copy - but that might just be fine.

`std::getline()` sounds interesting, but googling a bit, it may provide some challenges with with non-blocking input (pipes for executed commands). Also not sure how easy the single delimiter limitation is in the general case..

bwerthmann:
I've tested https://github.com/zeek/zeek/pull/3959/commits/a6e644865eb392ba9abf3bafc566179cb54e008d applied to Zeek 7 for a number of days and it seems to fix the issue as reported.

ckreibich (approved):
LGTM. This function should document its return argument and side effects somewhere, since they're subtle. But that's true for the whole class anyway, and unrelated to this changeset.

ckreibich:
Just rebasing here to keep the commit closer to the merge since some time has passed. Merging momentarily.

ckreibich:
Btw I'm surprised to see that `sleep infinity` is sufficiently portable, but it appears so...

PR #3960, opened 2024-10-02
Fix input error test nondeterminism

evantypanski:
Test introduced in #3949 caused nondeterminism (spotted by @awelzel [here](https://cirrus-ci.com/task/5479888096854016?logs=test#L2021))

It seems like other similar tests get by because they have more "stuff" before they call `terminate()` most likely. But, to be safe, just removing the "received termination signal" line seems like the best approach.

also fixed tabs vs. spaces :)


awelzel:
> But, to be safe, just removing the "received termination signal" line seems like the best approach.

I'm thinking putting a valid entry at the end might have also worked :-)

awelzel (approved):

PR #3961, opened 2024-10-02
telemetry: Invoke Telemetry::sync() only at scrape/collection time

### LINKED ISSUE #3947: Provide network-time-independent Telemetry sync hook
Currently, the existing `Telemetry::sync` hook is called using Zeek-script `schedule`, which is based on network time:

https://github.com/zeek/zeek/blob/b22ec065680a43c6a484416a807f4c3d6a5d9304/scripts/base/frameworks/telemetry/main.zeek#L496

Network time is driven by packet timestamps. If Zeek is overloaded in live mode, packets might not be processed in real-time and network time "slows-down" compared to actual wall-clock. As a consequence metrics might not be updated at the desired frequency. This is particularly bad for metrics that are intended to diagnose overload situations (e.g. packet lag).

A proposed solution by @awelzel is to introduce a hook in the new Telemetry framework that is called when Prometheus scrapes metrics. So how about adding a `Telemetry::scrape` hook?

awelzel:
This stops invoking Telemetry::sync() via a scheduled event and instead only invokes it on-demand. This makes metric collection network time independent and lazier, too.

With Prometheus scrape requests being processed on Zeek's main thread now, we can safely invoke the script layer Telemetry::sync() hook.

Closes #3947


awelzel:
@J-Gras - could you see if this would work for you?

ckreibich:
I realize this is just a draft, but a few thoughts since this is close to follow-ups for the framework that Tim and I discussed:
* The in-core metrics already have the notion of scrape-time callbacks (at metric granularity), and we considered expanding that concept to the script layer but ran out of time for 7.0. I think it is tempting to just use a single hook for this so I like this quite a bit — more Zeek-y and probably easier to use overall.
* I think as-is this breaks telemetry.log unless `policy/frameworks/telemetry/log.zeek` invokes the `sync()` hook?
* There's also the script-level `Telemetry::collect_metrics()` — haven't checked whether that's currently covered but it too has an expectation of seeing current values

NEWS
```diff
@@ -93,6 +93,10 @@ Changed Functionality
 * The ASCII input reader now suppresses warnings for consecutive invalid lines,
   producing a summary of total suppressions once a valid line is encountered.
 
+* The `Telemetry::sync()` hook is now invoked lazily. Either when the metrics
+  of a node are scraped via the HTTP endpoint, or one of the collect method
+  invoked from Zeek script.
```
J-Gras, line 98:
```suggestion
  of a node are scraped via the HTTP endpoint, or one of the collect methods is
  invoked from Zeek script.
```

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -263,22 +263,14 @@ export {
...
-	## Telemetry sync hook.
+	## Interval at which the :zeek:see:`Telemetry::sync` hook is invoked.
 	##
-	## This hook is invoked every :zeek:see:`Telemetry::sync_interval`
-	## for script writers to synchronize or mirror metrics with the
-	## telemetry subsystem. For example, when tracking table or value
-	## footprints with gauges, the value in question can be set on an actual
-	## :zeek:see:`Telemetry::Gauge` instance during execution of this hook.
+	## By default, the hook is invoked on demand, setting this option to
+	## a positive interval allows to invoke it regularly, too.
```
J-Gras, line 269:
```suggestion
	## a positive interval allows to invoke it regularly, too. Note that regular
	## invocations are relative to Zeek's network time.
```

src/telemetry/Manager.cc
```diff
@@ -256,6 +257,19 @@ static bool compare_histograms(const std::optional<ZVal>& a, const std::optional
     return comparer(a, b, metric_record_type);
 }
 
+void Manager::InvokeTelemetrySyncHook() {
+    static const auto sync_hook = zeek::id::find_func("Telemetry::sync");
+
+    if ( sync_hook->Flavor() != FUNC_FLAVOR_HOOK )
+        reporter->InternalError("Telemetry::hook not a hook?");
```
J-Gras, line 264:
```suggestion
        reporter->InternalError("Telemetry::sync not a hook?");
```

J-Gras (commented):
> @J-Gras - could you see if this would work for you?

This should do it, thanks a lot for picking it up! I've left comments on some minor wording nits that caught my eye.

> - I think as-is this breaks telemetry.log unless `policy/frameworks/telemetry/log.zeek` invokes the `sync()` hook?

Indeed, with `sync_interval = 0sec` being the new default, this looks like it could break the telemetry log.

> - There's also the script-level `Telemetry::collect_metrics()` — haven't checked whether that's currently covered but it too has an expectation of seeing current values

My understanding is that the hook is invoked for all collect methods from within their implementation in the telemetry manager.

awelzel:
>> @J-Gras - could you see if this would work for you?
>
> This should do it, thanks a lot for picking it up! I've left comments on some minor wording nits that caught my eye.

Great, thank you. They look good.

>> I think as-is this breaks telemetry.log unless policy/frameworks/telemetry/log.zeek invokes the sync() hook?
>
> Indeed, with sync_interval = 0sec being the new default, this looks like it could break the telemetry log.

The log logic has it's own schedule (`Telemetry::log_interval`). As mentioned below, `collect_metrics()` will invoke `Telemetry::sync()` on-demand, so that should continue to work.

>> There's also the script-level Telemetry::collect_metrics() — haven't checked whether that's currently covered but it too has an expectation of seeing current values
> My understanding is that the hook is invoked for all collect methods from within their implementation in the telemetry manager.
>

Yeah, `Telemetry::sync` is invoked during `collect_metrics()` so that the returned metrics are up-to-date.

awelzel:
The other thought I had was if individual script-level metrics could have a callback, instead of the global `Telemetry::sync` one.
```
global my_counter = Telemetry::counter_with_callback(counter_family, function(): double { return 42.0; }, labels)
```

But maybe that's just an additional feature for down the road.

scripts/base/frameworks/telemetry/main.zeek
awelzel: I'd think this applies to most every interval we have anywhere and may or may not be problematic. But yeah, can add it, given the external visibility.

ckreibich:
> The other thought I had was if individual script-level metrics could have a callback

That's what Tim and I had considered, but looking at the sync hook I actually like that better. It strikes me as more Zeek-y.

ckreibich:
I added the docs label because we should cover this in the manual.

scripts/policy/misc/stats.zeek
```diff
@@ -130,9 +130,18 @@ global packet_lag_gf = Telemetry::register_gauge_family([
     $help_text="Difference of network time and wallclock time in seconds.",
 ]);
 
+global network_time_cf = Telemetry::register_gauge_family([
```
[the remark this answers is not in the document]
awelzel: As discussed in person, initially due to the `_total` suffix, but after searching a bit, will rename to `zeek_net_timestamp_seconds` and keep the gauge.

https://prometheus.io/docs/practices/naming/
https://www.robustperception.io/are-increasing-timestamps-counters-or-gauges/

PR #3964, opened 2024-10-04
Fix flaky `missing-file-initially` test

evantypanski:
@awelzel (again) found a flaky test from #3949 - result in centosstream9 [here](https://cirrus-ci.com/task/5259252405633024?logs=test#L2029)

The test got flaky probably from #3949 on centosstream9 CI. You can replicate that behavior by increasing the sleep time when waiting for the file such that the test will attempt to read the missing file again. Since the one second wait for file is glacially slow for this, speeding it up should mean that the file gets created sooner and so the test won't try to open the file again. But, it's always still technically possible, since the test will wait for 10 seconds and the heartbeat seems to be 1 second. At least if that happens, it's probably a bug or massive slowdown of some kind.

### DIFF
--- a/testing/scripts/wait-for-file
+++ b/testing/scripts/wait-for-file
@@ -7,17 +7,21 @@ if [[ $# -ne 2 ]]; then
     exit 1
 fi
 
+SLEEP_INTERVAL=0.1
+SLEEP_INTERVAL_MS=100
+
 wait_file=$1
 max_wait=$2
-wait_count=0
+# Avoid floating point arithmetic by using milliseconds
+wait_countdown=$((${max_wait}000 / ${SLEEP_INTERVAL_MS}))
 
 while [[ ! -e $wait_file ]]; do
-    let "wait_count += 1"
+    let "wait_countdown -= 1"
 
-    if [[ $wait_count -ge $max_wait ]]; then
+    if [[ $wait_countdown -le 0 ]]; then
         echo >&2 "error: file '$wait_file' does not exist after $max_wait seconds"
         exit 1
     fi
 
-    sleep 1
+    sleep $SLEEP_INTERVAL
 done


evantypanski:
I guess there are some cases that pass locally but not here again. Well well well

evantypanski:
The "fix" is in the `wait-for-file` script - the rest of the changes are to switch it from taking the seconds to wait to cycles to wait (where a cycle is 1/10 of a second).

I originally used `bc` to do math and not require that change (because math is bash's worst enemy), but seems like that wasn't installed everywhere. I'd rather not deal with that so I just changed all tests that use the script to use 10x the number as the cycles argument - meh.

testing/scripts/wait-for-file
```diff
@@ -3,21 +3,22 @@
 # Sleeps until a file comes into existence.
 
 if [[ $# -ne 2 ]]; then
-    echo >&2 "usage: $0 <file to wait for> <max secs to wait>"
+    echo >&2 "usage: $0 <file to wait for> <max cycles to wait>"
```
awelzel, line 6: I'd leave this at `max secs` to sleep (it's easier to reason about).

The maybe something like:
```
sleep_until=$(date "-s +$max_wait sec" +%s)
...
if [[ $(date +%s) -gt $sleep_until ]] ;then
  break;
fi
```
Your `bc` approach also looked fine, was there something off with it?

This maybe does the wait_count's well enough?
```
$ echo 'x=(0.0 / 0.1 + 0.5); scale=0; if (x/1 > 0) { x/1 } else { 1 }' | bc
```

awelzel: The issue with the `date` comparison is a bit that if the time goes into the past by a few years, it'll sleep very long, and it involves many more `date` invocations.


This would be okay, too, I guess:
```
sleep_interval_ms=100
wait_count = $(( ${max_wait}000 / ${sleep_interval_ms} ))
```

evantypanski: I actually kind of prefer something like this, though it's slightly weird. I'll push that and see what you/CI think

awelzel (approved):
I have a shell-check warning locally on that let, will fix it up during the merge (wants `x=$((x - 1))`)

Thanks!

PR #3968, opened 2024-10-08
Factoring of script optimization headers

vpax:
In response to a suggestion of @ckreibich's, this PR factors the previously massive `Compile.h` header associated with both `src/script_opt/ZAM/` and `src/script_opt/CPP/` into per-.cc-headers. In the process I also reflowed a bunch of comments, and made some minor rearrangements for consistency. There are no semantic changes, however, just a lot of declarations moving around.


ckreibich (approved):
Thanks Vern, just rebasing to pull up the commits since some time has passed. Will merge shortly.

PR #3971, opened 2024-10-09
Improvements in FNV1A hash functions

martinvanhensbergen:
- Add FNV1a64 hash function
- make hash function type aware
- add unit tests


martinvanhensbergen:
I see that the testcases fail. From what I can tell (but I will investigate more later today) is that the function was indeed used in creating identifiers and ids. I think due to the values being a bit different now that some ordering of logs have changed and that an update of the baseline will do. Do you share that quick analysis or might there be something else involved?


Also I want to change the code a bit to not have the constants as parameter but separately defined first.

awelzel:
>  Do you share that quick analysis or might there be something else involved?

I think it's this call to fnv1a32 causing pool-related differences as it's now computing a different hash value for the string input.

https://github.com/zeek/zeek/blob/2e576b058d10a6bea31db1959afcfa0e7e84bc36/scripts/base/frameworks/cluster/pools.zeek#L271-L272

src/zeek.bif
```diff
@@ -922,6 +922,41 @@ function paraglob_equals%(p_one: opaque of paraglob, p_two: opaque of paraglob%)
 	);
 	%}
 
+%%{
+enum FNV_HASH_TYPE { FNV1a32, FNV1a64};
+
+uint64_t fnva(zeek::Val* input, uint64_t offset, uint64_t prime, FNV_HASH_TYPE type ) {
+	zeek::ODesc desc(zeek::DESC_READABLE);
```
awelzel, line 929: I'm forth and back suggesting to revert this to DESC_BINARY so the type is included in the hash. If someone want's READABLE, they can do the following explicitly now `fnv1a32(cat(val))`.

```
assert fnv1a32(123) == fnv1a32("123")
assert fnv1a32(123.0) == fnv1a32("123.0")
assert fnv1a32(T) == fnv1a32("T")
```

src/zeek.bif
```diff
@@ -922,6 +922,41 @@ function paraglob_equals%(p_one: opaque of paraglob, p_two: opaque of paraglob%)
...
+		}
+
+	uint64_t rval = offset;
+	for ( auto i = 0; i < length; ++i )
+	{
+		rval ^= (uint64_t) bytes[i];
+		rval *= prime;
+
+		if (type == FNV1a32)
+			rval = rval % (1ULL << 32);
```
awelzel, line 953: You could template fnva with the type and skip this I think - also allows to drop the type argument.

```
-uint64_t fnva(zeek::Val* input, uint64_t offset, uint64_t prime, FNV_HASH_TYPE type ) {
+template<typename IntType>
+IntType fnva(zeek::Val* input, IntType offset, IntType prime) {
        zeek::ODesc desc(zeek::DESC_READABLE);
 
        auto length = 0;
@@ -943,14 +944,11 @@ uint64_t fnva(zeek::Val* input, uint64_t offset, uint64_t prime, FNV_HASH_TYPE t
                        length = desc.Len();
                }
 
-       uint64_t rval = offset;
+       IntType rval = offset;
        for ( auto i = 0; i < length; ++i )
        {
-               rval ^= (uint64_t) bytes[i];
+               rval ^= bytes[i];
                rval *= prime;
-
-               if (type == FNV1a32)
-                       rval = rval % (1ULL << 32);
        }
 
        return rval;
```

Use as follows:
```
@@ -967,10 +965,10 @@ uint64_t fnva(zeek::Val* input, uint64_t offset, uint64_t prime, FNV_HASH_TYPE t
 ## .. zeek:see:: hrw_weight
 function fnv1a32%(input: any%): count
        %{
-       uint64_t offset = 2166136261;
-       uint64_t prime = 16777619;
+       uint32_t offset = 2166136261;
+       uint32_t prime = 16777619;
 
-       uint64_t hash = fnva(input, offset, prime, FNV1a32);
+       auto hash = fnva<uint32_t>(input, offset, prime);
        return zeek::val_mgr->Count(hash);
```

src/zeek.bif
martinvanhensbergen: I have made the changes as you suggested and added testcases for the int/double/string 123 versions to check they have different values.

src/zeek.bif
martinvanhensbergen: Made the change as you suggested!

martinvanhensbergen:
I have made the changes as discussed and also updated the baselines of a few other tests now that they use a different id value caused by the change of the hash function.

src/zeek.bif
```diff
@@ -922,6 +922,37 @@ function paraglob_equals%(p_one: opaque of paraglob, p_two: opaque of paraglob%)
...
+template<typename IntType>
+IntType fnva(zeek::Val* input, IntType offset, IntType prime) {
+	zeek::ODesc desc(zeek::DESC_BINARY);
+
+	auto length = 0;
+	const u_char* bytes = NULL;
+
+	if (input->GetType()->Tag() == zeek::TYPE_STRING)
+		{
+			length = ((zeek::StringVal*) input) -> Len();
```
0xxon, line 935: super small nit - the indentation here is wrong, the length and wollowing lines are supposed to be at the same leel as the `{`.

We can just fix this on merging.

Same below.

src/zeek.bif
```diff
@@ -922,6 +922,37 @@ function paraglob_equals%(p_one: opaque of paraglob, p_two: opaque of paraglob%)
...
+		{
+			input->Describe(&desc);
+			bytes = desc.Bytes();
+			length = desc.Len();
+		}
+
+	IntType rval = offset;
+	for ( auto i = 0; i < length; ++i )
+	{
+		rval ^= bytes[i];
```
0xxon, line 948: Also indentation - here `{` should be indented one more level.

0xxon (approved):

PR #3972, opened 2024-10-11
Prevent non-Modbus on port 502 to be reported as Modbus

### LINKED ISSUE #3962: TCP packets get parsed as Modbus
Dear Zeek team,

I’ve encountered an issue with how Zeek parses Modbus traffic on port 502. Currently, all traffic on port 502 is parsed as Modbus if an entire PDU from both sides is successfully parsed, even if the traffic isn't valid Modbus TCP. This leads to cases where invalid TCP traffic is misidentified as Modbus traffic. In my application, I frequently encounter odd data and invalid function codes, making it difficult to distinguish between genuine Modbus traffic and erroneously parsed data.
The Modbus Messaging on TCP/IP Implementation Guide (V1.0b) specifies that the MBAP header should contain a Protocol Identifier and a Length field. By incorporating this information into the Modbus analyzer, we could prevent incorrectly parsed TCP traffic by verifying that the Protocol Identifier (PID) and length are correct.
image

 
I’ve examined the relevant code and believe the necessary changes can be made in zeek/src/analyzer/protocol/modbus/modbus-analyzer.pac, specifically in the function deliver_ModbusTCP_PDU(message: ModbusTCP_PDU): bool at line 101. This function should return false if the MBAP headers do not contain the correct PID or if the length does not match the body length (plus 1 for the Unit Identifier).

image
image

However, I’m unsure about how to access the header and body fields within this function. From what I understand of the .pac language, we should be able to access these through message.header and message.body. If this is correct, we can verify the Protocol Identifier and Length fields as part of the header and ensure they match the PDU body.

Could you confirm if the header and body fields are correctly defined and accessible in deliver_ModbusTCP_PDU? If this structure is valid, I’d be happy to proceed with implementing the fix and submitting a pull request.

zambo99:
This pull request prevents most non-Modbus TCP traffic on port 502 to be reported as Modbus in conn.log as well as in modbus.log. To do so, we have introduced a new function  in the Modbus flow that checks that some specific fields of the (supposedly) Modbus header are compatible with values specified in the specs.

The function is then used in guards around the different callback functions of the parser, to prevent the generation of incorrect Modbus events.

Closes #3962


zambo99:
Hi, is there something wrong with this pull request? Do I have to explicitly suggest a reviewer?

src/analyzer/protocol/modbus/modbus-analyzer.pac
```diff
@@ -81,13 +81,20 @@ refine connection ModbusTCP_Conn += {
 		%{
 		return confirmed && orig_pdu && resp_pdu;
 		%}
+
+	function IsValidHeader( header: ModbusTCP_TransportHeader ) : bool
+        %{
+        // Protocol Identifier must be 0 as per specs
+        // Length of Modbus APDU must be at least 2 (Unit Identifier and Function Code)
+        return ${header.pid} == 0 && ${header.len} >= 2;
+        %}
```
evantypanski, line 90: Nit: The rest of the file uses tabs, but the new function bodies use spaces. Could they be swapped to tabs to be consistent?

src/analyzer/protocol/modbus/modbus-analyzer.pac
```diff
@@ -102,21 +109,21 @@ refine flow ModbusTCP_Flow += {
...
-		if ( ! connection()->IsConfirmed() )
-			{
-			connection()->SetConfirmed();
-			connection()->zeek_analyzer()->AnalyzerConfirmation();
-			}
+        if ( ! connection()->IsConfirmed() && connection()->IsValidHeader( ${message.header} ) )
+            {
+            connection()->SetConfirmed();
+            connection()->zeek_analyzer()->AnalyzerConfirmation();
+            }
```
evantypanski, line 118: Nit again: Could these be switched to tabs? not spaces

evantypanski (commented):
> is there something wrong with this pull request?

Nothing is wrong, I just think the people who would be a little more equipped to review this are out right now. I only have pretty superficial comments which I left, but it shouldn't be too much longer until we get a better review. Sorry for the delay.

I haven't really done much (or anything really) with these binpac grammars so I don't think I'd be the best reviewer here

src/analyzer/protocol/modbus/modbus-analyzer.pac
```diff
@@ -81,13 +81,20 @@ refine connection ModbusTCP_Conn += {
 		%{
 		return confirmed && orig_pdu && resp_pdu;
 		%}
+
+	function IsValidHeader( header: ModbusTCP_TransportHeader ) : bool
```
awelzel, line 85: @zambo99 - thanks for the PR. Could you see if the following patch does the same?

```
diff --git a/src/analyzer/protocol/modbus/modbus-protocol.pac b/src/analyzer/protocol/modbus/modbus-protocol.pac
index ebd1805815..5b566e6729 100644
--- a/src/analyzer/protocol/modbus/modbus-protocol.pac
+++ b/src/analyzer/protocol/modbus/modbus-protocol.pac
@@ -90,8 +90,8 @@ type ModbusTCP_PDU(is_orig: bool) = record {
 
 type ModbusTCP_TransportHeader = record {
        tid: uint16; # Transaction identifier
-       pid: uint16; # Protocol identifier
-       len: uint16; # Length of everything after this field
+       pid: uint16 &enforce(pid == 0); # Protocol identifier
+       len: uint16 &enforce(len >= 2); # Length of everything after this field
        uid: uint8;  # Unit identifier (previously 'slave address')
        fc:  uint8;  # MODBUS function code (see function_codes enum)
 } &byteorder=bigendian, &let {
```

This would trigger analyzer violations instead of skipping event raising if it wasn't a valid header. It should also result in appropriate analyzer.log / dpd.log entries.

src/analyzer/protocol/modbus/modbus-analyzer.pac
```diff
@@ -102,21 +109,21 @@ refine flow ModbusTCP_Flow += {
...
+            }
 
 		return true;
 		%}
 
 	# EXCEPTION
 	function deliver_Exception(header: ModbusTCP_TransportHeader, message: Exception): bool
 		%{
-		if ( ::modbus_exception )
+		if ( ::modbus_exception && ( connection()->IsConfirmed() || connection()->IsValidHeader(header) ) )
```
awelzel, line 126: I *think* if we use `&enforce` this can go and result in the wanted behavior still.

src/analyzer/protocol/modbus/modbus-analyzer.pac
zambo99: Indeed, with the two &enforce attributes added to the parser the effect is the same.
Your solution is more elegant. However, I think we should at least keep the btest I made in my patch to ensure a basic level of regression testing.
How shall we proceed? Do I make a new pull request?

zambo99:
Indeed, with the two &enforce attributes added to the parser the effect is the same.
Your solution is more elegant. However, I think we should at least keep the btest I made in my patch to ensure a basic level of regression testing.
How shall we proceed? Do I make a new pull request?

src/analyzer/protocol/modbus/modbus-analyzer.pac
awelzel: > However, I think we should at least keep the btest I made in my patch to ensure a basic level of regression testing.

Yeah, definitely.

> How shall we proceed? Do I make a new pull request?

You can also remove the commits with the `IsValidHeader()` addition and add a new commit with the enforce hunk. Then force-push.

You could consider adding `analyzer.log` to the baselines which should show the violations being raised.

awelzel (approved):

awelzel:
Thank you for your contribution @zambo99 ! I've marked this for 7.0.4, too.

PR #3974, opened 2024-10-11
Enshrine `.evt` raising event order in a test

evantypanski:
It seems like, given two event triggers in a .evt file, that the order they are parsed is the order the events will be raised via `raise_event` later. This isn't groundbreaking, but this behavior isn't anywhere in a test nor is it documented (from what I can tell). At the very least, some user might notice this behavior and rely on it, which I don't think is unreasonable (that user is me). So, this modifies the existing test that raises two of the same event and makes it also test the order, both with an assertion and by removing the `sort` call.

For what it's worth, I'd also like to document this behavior, but it's fine if it's not. I looked into the spicy plugin history for insights about this, but didn't find any before hitting another import from another repo, which goes before the GH import of Spicy, but I tried :)

And, for good measure: is this a safe assumption? Digging through the code I see nothing that would make the second event fire before the first

### DIFF
--- a/testing/btest/Baseline/spicy.hook-priority/output
+++ b/testing/btest/Baseline/spicy.hook-priority/output
@@ -2,4 +2,6 @@
 Spicy: highest prio
 Spicy: default prio
 Spicy: lowest prio
+Zeek: highest prio, [x=default]
 Zeek: default prio, [x=default]
+Zeek: lowest prio, [x=default]
--- a/testing/btest/spicy/hook-priority.zeek
+++ b/testing/btest/spicy/hook-priority.zeek
@@ -12,9 +12,19 @@ event zeek_init()
 	Analyzer::register_for_port(Analyzer::ANALYZER_FOO, 80/tcp);
 	}
 
+event foo_last(x: foo::X)
+	{
+	print "Zeek: lowest prio", x;
+	}
+
 event foo(x: foo::X)
 	{
 	print "Zeek: default prio", x;
+	}
+
+event foo_first(x: foo::X)
+	{
+	print "Zeek: highest prio", x;
 	}
 
 # @TEST-START-FILE foo.spicy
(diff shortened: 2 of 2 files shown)


bbannier:
I am not sure there is a huge upside in tying us to a particular behavior since we provide knobs to control the execution order (hook and event priorities). The current behavior seems more a coincidence and hard to control given how many moving pieces are involved.

Conceptually related, at least on my system I do not see this ordering when running pure Spicy hooks, e.g.,

```ruby
# foo1.spicy
module foo1;

public type X = unit {
    a: uint8;
    b: uint8;
};

on X::a {
    print "1";
}

on X::a {
    print "2";
}
```

```console
$ echo foo | spicy-driver foo1.spicy -d
2
1
```

versus when just dropping a field

```ruby
# foo2.spicy

module foo2;

public type X = unit {
    a: uint8;
};

on X::a {
    print "1";
}

on X::a {
    print "2";
}
```

```console
echo foo | spicy-driver foo2.spicy -d
1
2
```

To make the ordering of these Spicy-side hooks deterministic one would use [hook priorities](https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#:~:text=if%20multiple%20implementations%20are%20provided%20for%20the%20same%20hook%2C%20by%20default%20it%20remains%20undefined%20in%20which%20order%20they%20will%20execute.%20if%20a%20particular%20order%20is%20desired%2C%20you%20can%20specify%20priorities%20for%20your%20hook%20implementations%3A).
 
The order in which Zeek-side events are raised depends on the implicit priority of the LHS Spicy hooks (`priority=-1000`), or the `&priority` of the RHS Zeek event if overridden (this seems undocumented); these priorities get then plugged through to Zeek and should control the order in which Spicy hooks are handled.

rsmmr:
>  is this a safe assumption?

The order is a result of the current code generation, which I believe indeed just happens to go through the `event` entries in order, generating a function call for each. However, I'd rather not depend on this and prefer leaving the order undefined. That's also why the existing test has the `sort`, so that order explicitly doesn't matter.

 There is actually a way to control order if needed: you can add `&priority=N` to the end of the EVT `event` definition  (similar to Spicy hooks and Zeek events). Turns out that's apparently neither documented nor tested--oops.  So I suggest we document that as the way to get a deterministic order, and leave semantics undefined in the absence of `&priority`.

evantypanski:
Well, I didn't see that `&priority` can be used when I was searching through the code, I just saw it default to `-1000` :)

I changed this to test `&priority` in evt files instead. I think this is better, I just didn't see it. I'll try to find a good place to document it, too.

bbannier (approved):

PR #3975, opened 2024-10-15
Fix CI jobs failing to run due to unsupported use of `pip install`

bbannier:
The GH action image `ubuntu-latest` recently switched to a version which does not support running `pip install` without `--break-system-packages` anymore. This PR fixes affected workflows.


bbannier:
I will make similar changes for the _Generate Documentation_ job.

rsmmr (approved):

PR #3981, opened 2024-10-18
Support for "slow nodes": un- and re-peering of nodes that Broker drops to due excessive message I/O lag

ckreibich:
This supersedes #3951, building on @Neverlord's work. It currently also stacks on top of #3980 until that's merged.


ckreibich:
This is nearly ready but I've marked it as a draft until some final tweaks are in place, including finalizing the Broker-side PR, and getting #3980 merged first.

scripts/base/frameworks/cluster/slow-peers.zeek
```diff
@@ -0,0 +1,105 @@
+##! This implements handling of "slow peers", i.e. nodes that fall so far behind
+##! in message I/O that the local node decides to unpeer them.  Zeek captures
+##! this occurrence in logging, eventing, and via telemetry.
+
+@load base/frameworks/telemetry
+
+module Cluster;
```
awelzel, line 7: Could we move this logic into framworks/broker instead? Also wondering how much the existing `broker.log` covers the logging done here.

I'm also a bit wary about the `Cluster::node_slow()` - we're not introducing users and wonder if a less generic `Broker::peer_forced_disconnect` (or something better named - I find even "slow" subjective) would be sufficient.

ckreibich: Ah yeah — I did wonder about Broker vs Cluster frameworks for this one. I landed on the Cluster framework because the `node_up` and `node_down` events fit well to me with the fact that a node is overwhelmed. That's also why I chose "slow node" as the moniker for it, and I'm happy to change it. Basically, there's the Broker-specific fact that it un-peered a node due to backpressure, and the cluster-level observation that a node is overwhelmed.
 
Broker.log currently captures the event via its existing logging of removed/lost peerings in status updates, if you know the message string to look for. If we had more explicit handling of the reasons why things happen in Broker, there'd be nothing new to do at that level. (I realized while working on this that there is generally more detail available to Broker that Zeek does not report, for example when a peer rejects a peering's new TCP connection because it's redundant to an already existing peering — etc.)

I just noticed that `Cluster::node_down` only triggers as a derivative of `Broker::peer_lost` (i.e. from the perspective of the node connected to), but not `Broker::peer_removed` (the perspective of the node establishing the connection). If we support this from either end, then the Cluster framework will automatically report nodes depeered due to backpressure as down, and we don't need to do anything in there.

Let me try the above, and add an explicit backpressure-induced de-peering event to Broker.

ckreibich:
Okay, I've reworked this quite a bit. The re-peering lives in the Broker scripts, and the cluster framework just adds logging/telemetry since it has node-naming context. The corresponding test cases live in the cluster testsuite as linked above, and this also has additional Broker changes per zeek/broker#436.

I'm going to leave this in draft mode for the time being, to see how it fares in CI.

scripts/base/frameworks/broker/backpressure.zeek
```diff
@@ -0,0 +1,35 @@
...
+		Reporter::error(fmt("Missing network info to re-peer with %s", endpoint$id));
+		return;
+	}
+
+	# Re-establish the peering so Broker's reconnect behavior kicks in once
+	# the other endpoint catches up. Broker will periodically re-try
+	# connecting as necessary. If the other endpoint originally connected to
+	# us, our attempt will fail (since we attempt to connect to the peer's
+	# ephemeral port), but in that case the peer will reconnect with us once
+	# it recovers.
```
awelzel, line 29: Is it correct that a node that disconnected a "connecting node" will endlessly attempt to connect to the other node's stale ephemeral port though and produce connection errors continuously?

awelzel: I see the comment below and don't have a better idea, unless `endpoint` could somehow be enriched with information whether the peer was originally "accepted" or "peered_to".

If the `cluster/setup-connections.zeek` / `Broker::peer()` functionality was idempotent, maybe could just re-run that, but don't think this works just like this.

So, don't know what to do.

src/broker/Manager.cc
```diff
@@ -297,6 +297,36 @@ void Manager::DoInitPostScript() {
     options.disable_forwarding = ! get_option("Broker::forward_messages")->AsBool();
     options.use_real_time = use_real_time;
 
+    options.peer_buffer_size = get_option("Broker::peer_buffer_size")->AsCount();
+    auto peer_overflow_policy = get_option("Broker::peer_overflow_policy")->AsString()->CheckString();
```
awelzel, line 301: Nit, could use `ToStdString()` and just  `==` below rather than `util::streq`.

awelzel (commented):
This is nice, I left a few thoughts for maybe one more iteration, but :+1: 

I'm not so sure about the optimistic `Broker:peer()`, but maybe can figure out something more clever in the future.

testing/btest/bifs/sleep.zeek
```diff
@@ -0,0 +1,21 @@
+# Verifies sleep()'s reported latencies.
+#
+# @TEST-EXEC: zeek -b %INPUT 2>out
+# @TEST-EXEC: btest-diff out
```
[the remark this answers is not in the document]
ckreibich: This was mainly to trigger a failure when `assert()` produces output. `assert()` does not error out, right? I felt this was easier than using an `assert_failure()` hook, but let me know if I'm not holding this right.

scripts/base/frameworks/cluster/broker-backpressure.zeek
```diff
@@ -0,0 +1,33 @@
+# Notifications for Broker-reported backpressure overflow.
+
+@load base/frameworks/telemetry
+
+module Cluster;
+
+global broker_backpressure_overflows_cf = Telemetry::register_counter_family([
+    $prefix="zeek",
+    $name="broker-backpressure-overflows",
```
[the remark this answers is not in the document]
ckreibich: Good thought, yeah. If one changes the backpressure policy, the `peer_removed` event won't trigger in the first place, so this is specific to that situation. Let me change it.

src/broker/Manager.cc
ckreibich: Could you live with doing that in a follow-up? Dominik likely did that just to stay consistent with the rest of the use in that function, and I'm wondering whether we should just ditch `get_option()` in favor of `id::find_const()`, or some such. What do you think? I can tuck it on here if you prefer.

awelzel: Not a biggie!

scripts/base/frameworks/broker/backpressure.zeek
awelzel: > All in all I feel I can live with this for now, assuming the assumption that backpressure problems are rare holds. But I'm happy to look into this in a follow-up.

The symptom for users would be "spurious" connection attempts on nodes to addr/ports that aren't anywhere in the cluster layout. This might be rather confusing, but follow-up / past 7.1 :+1:

testing/btest/bifs/sleep.zeek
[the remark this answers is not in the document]
awelzel: You're right. Because we run with `ZEEK_ALLOW_INIT_ERRORS=1`, failing asserts in `zeek_init()` won't cause Zeek to exit with a failure.

Once `assert` testing is more established, might look for another pattern. If `zeek` would exit with failure, we could just rely on `.stderr` being populated. That's also shown by default when running `btest -d` on failure.

awelzel (approved):
:rocket:

ckreibich:
Just a final rebase here with Broker and the testsuite switched to their master/main branches.

PR #3985, opened 2024-10-21
Fix issues with Modbus message logging

### LINKED ISSUE #3984: Issues with Modbus message logging
Dear Zeek team,

During our testing of Zeek's Modbus logging capabilities we have discovered three issues:

1.  Some exception responses (e.g., READ_COILS_EXCEPTION) are logged twice: once without and once with the exception message. This is due to an inconsistency between the parser and the main.zeek script. The modbus_message() event prevents emitting logs for exception responses that receive also the modbus_exception() event. In the example above, the READ_COILS_EXCEPTION generates the modbus_exception(), but is also logged by the code in the modbus_message() event.
2. Some exception responses (e.g., PROGRAM_484_EXCEPTION) are not logged. For the same reason, some exception responses configured not to be logged in the modbus_message() event do not even generate a modbus_exception() event, resulting in the exception response not being logged at all.
3. Some reserved (but known) function codes (e.g., PROGRAM_UNITY) are logged as unk-xxx (e.g., unk-90), while it would be possible to log their known name.

Given that all Modbus exception responses share the same payload, we believe it is possible to modify the Modbus parser to parse all exception responses (i.e., a Modbus message with the MSB of the function code set) and generate the corresponding modbus_exception() event. In this way, the main.zeek script logic can also be simplified by preventing logging any exception response in the modbus_message() event, and logging all exception responses in the modbus_exception() event.

Additionally, it would be possible to improve the logging by making more coherent the function code enumeration in const.zeek, so that for every known request/response function code, also the corresponding exception response is present.

Finally, it would also be possible to add new reserved (but known) function codes to the enumeration in const.zeek.

We have a PR ready for this, and will create it shortly.

zambo99:
This pull request fixes three issues with Zeek's Modbus message logging:

1 - Some exception responses (e.g., READ_COILS_EXCEPTION) are logged
    twice: once without and once with the exception message.
2 - Some exception responses (e.g., PROGRAM_484_EXCEPTION) are not
    logged.
3 - Some known but reserved function codes (e.g., PROGRAM_UNITY) are
    logged as unk-xxx (e.g., unk-90), while it would be possible to
    log their known name.

To address these inconsistencies, the modbus parser has been updated to parse all exception responses (i.e., all responses where the MSB of the function code is set) using the already defined Exception message.
Also, the Modbus main.zeek script has been updated to consistently demand logging exception responses to the specialized modbus_exception event, rather than logging some exception responses in the modbus_message event and others in the modbus_exception event. Finally, the consts.zeek script has been updated to make sure that for every known function code, the corresponding exception code was also present, and to expand the list of known function codes.

Closes #3984


scripts/base/protocols/modbus/consts.zeek
```diff
@@ -44,13 +47,33 @@ export {
 		[0x85] = "WRITE_SINGLE_COIL_EXCEPTION",
 		[0x86] = "WRITE_SINGLE_REGISTER_EXCEPTION",
 		[0x87] = "READ_EXCEPTION_STATUS_EXCEPTION",
+		[0x88] = "DIAGNOSTICS_EXCEPTION",
```
awelzel, line 50: I find the exception codes in function_codes a bit redundant, duplicated.

This does the same locally and allows to remove all the `_EXCEPTION` entries. It's possibly a bit slower, but I'm not worried here. What do you think?

```
--- a/scripts/base/protocols/modbus/main.zeek
+++ b/scripts/base/protocols/modbus/main.zeek
@@ -46,6 +46,25 @@ event zeek_init() &priority=5
        Analyzer::register_for_ports(Analyzer::ANALYZER_MODBUS, ports);
        }
 
+function build_func(func: count): string
+       {
+       local masked = func & ~0x80;
+
+       # If the function code is in function_codes, use it. Also,
+       # if the masked value isn't in function_codes, use function_codes
+       # &default functionality.
+       if ( func in function_codes || masked !in function_codes )
+               return function_codes[func];
+
+       local s = function_codes[masked];
+
+       # Suffix exceptions with _EXCEPTION.
+       if ( func & 0x80 == 0x80 )
+               s += "_EXCEPTION";
+
+       return s;
+       }
+
 event modbus_message(c: connection, headers: ModbusHeaders, is_orig: bool) &priority=5
        {
        if ( ! c?$modbus )
@@ -56,7 +75,7 @@ event modbus_message(c: connection, headers: ModbusHeaders, is_orig: bool) &prio
        c$modbus$ts   = network_time();
        c$modbus$tid = headers$tid;
        c$modbus$unit = headers$uid;
-       c$modbus$func = function_codes[headers$function_code];
+       c$modbus$func = build_func(headers$function_code);
        ## If this message is from the TCP originator, it is a request. Otherwise,
        ## it is a response.
        c$modbus$pdu_type = is_orig ? "REQ" : "RESP";
@@ -65,7 +84,7 @@ event modbus_message(c: connection, headers: ModbusHeaders, is_orig: bool) &prio

```

awelzel (commented):
@zambo99 - sorry for the delay. This looks nice, I like the simplification for exception handling.

zambo99:
Hi Arne,

Given the solution you propose functionally achieves the same result, I think we can go with that. I do not see any downside to it.

I'll implement it and upload a new patch.

awelzel (approved):
Thanks! I'm going to revert the doc submodule update, I'm sure that's spurious.

If the `build_func()` ends up not being great for someone, can reconsider. For now I like the fact that we don't duplicate the exception values in the table.

zambo99:
Thank you!

And yes, any submodule update was totally accidental. :-)

PR #3986, opened 2024-10-22
Bump zeekjs to 0.12.1

### LINKED ISSUE #3978: Segfault in zeek 7 with JavaScript support under alpine
Hello,

I have a Docker image based on alpine that includes a Zeek installation. When processing trace files with a script that loads some JavaScript code and only after Zeek is already running for a few seconds there is a segfault when the stream of trace files is interrupted.
To recreate this behavior I tried to reduce the Docker image to the relevant components and run a script that creates a named pipe which Zeek is reading from while running a script that is empty but ends on `.js` to initialize the Node.js environment/JavaScript engine. After running for 10 seconds a pcap-header with no following packet records is written to the pipe with `cat`
 which closes the pipe afterwards and should lead to Zeek ending normally but causes a segfault instead.

This is the Dockerfile:
```dockerfile
FROM alpine:3.20

RUN apk add --no-cache \
    make \
    g++ \
    python3-dev \
    linux-headers \
    openssl-dev \
    libpcap-dev \
    zlib-dev \
    fts-dev \
    cmake \
    clang \
    bison \
    bash \
    swig \
    flex \
    flex-dev \
    flex-libs \
    nodejs-dev \
    nodejs-libs

WORKDIR /build
RUN wget --quiet -O /build/zeek.tar.gz "https://github.com/zeek/zeek/releases/download/v7.0.3/zeek-7.0.3.tar.gz" \
    && tar xf /build/zeek.tar.gz

WORKDIR /build/zeek-7.0.3

# Fix PAGE_SIZE already being defined as macro
RUN sed -i '147a#ifdef PAGE_SIZE\n#undef PAGE_SIZE\n#endif' auxil/spicy/3rdparty/fiber/src/fiber.c

RUN ./configure --build-type=Debug \
    && make -j 4 \
    && make install

ENV PATH $PATH:/usr/local/zeek/bin

WORKDIR /test

COPY runZeek.sh /test/runZeek.sh
CMD ["/test/runZeek.sh"]
```

And this is the `runZeek.sh` script:
```bash
#!/bin/sh
set -x
mkfifo pipe
touch empty.js
echo -n "1MOyoQIABAAAAAAAAAAAAA0AAQABAAAA" | base64 -d > header.pcap
ulimit -c unlimited
CORE_PATTERN="$(cat /proc/sys/kernel/core_pattern)"
CORE_DIR=${CORE_PATTERN%/*}
mkdir -p $CORE_DIR
zeek -b -C -r pipe empty.js &
sleep 10
cat header.pcap > pipe
wait
ls -lh $CORE_DIR
```

Inspecting a coredump reveales that the segfault is caused by the libnode library and running the script above with the official Docker image or a self compiled zeek version under ubuntu doesn't lead to a segfault but I am kinda restricted to alpine images.

awelzel:
7ec0953 ci: Bump actions/upload-artifact
    8e5914d ci/docker: Bump distros, update some OBS repos
    129b737 Fix crash during shutdown

Closes #3978


bbannier (approved):
🚀

PR #3992, opened 2024-10-24
Uniform auto-generated metrics ports w/ mgmt framework.

dopheide-esnet:
With the current code, if you have a multi-node cluster, the auto-generated metrics_ports assigned by the controller will continue to increment from a single starting point.   So for example, if you have a 3 node cluster and a starting port of 9101 you may end up with:

host1: ports 9101-9112   (manager, logger, proxy, 8 workers)
host2: ports 9113-9120   (8 workers)
host3: ports 9121-9128   (8 workers)

This pull request attempts to make those more uniform:
host1: ports 9101-9112   (manager, logger, proxy, 8 workers)
host2: ports 9101-9108   (8 workers)
host3: ports 9101-9108   (8 workers)

Note, this would currently break things if for some reason you had multiple instances running on a single host.  However, in that case you could just set the metrics_port manually for each cluster node.

### DIFF
--- a/scripts/policy/frameworks/management/controller/main.zeek
+++ b/scripts/policy/frameworks/management/controller/main.zeek
@@ -414,15 +414,17 @@ function config_assign_metrics_ports(config: Management::Configuration)
 		[Supervisor::WORKER] = 3,
 	};
 
-	local p = port_to_count(Management::Controller::auto_assign_metrics_start_port);
-	local ports_set: set[count];
+        local instance_metrics_start_port: table[string] of count;
+        local instance_ports_set: table[string] of set[count];
 	local node: Management::Node;
 
 	# Pre-populate agents ports, if we have them:
 	for ( inst in config$instances )
 		{
+		instance_metrics_start_port[inst$name] = port_to_count(Management::Controller::auto_assign_metrics_start_port);
+		instance_ports_set[inst$name] = {};
 		if ( inst?$listen_port )
-			add ports_set[port_to_count(inst$listen_port)];
+			add instance_ports_set[inst$name][port_to_count(inst$listen_port)];
 		}
 
 	# Pre-populate nodes with pre-defined metrics ports, as well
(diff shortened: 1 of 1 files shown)


0xxon:
I think this is something that @ckreibich should weight in on.

Is there a reason for this change, besides cosmetics? I am honestly a tad worried that it might not be super unusual for people to want to run multiple instances on a single node (for, e.g., different network interfaces) - and this change seems to make that more awkward.

dopheide-esnet:
My main goal was making it easier to uniformly configure our prometheus collector.   I think this could be expanded slightly to account for both cases (multi-node vs multi-instance on one node).

dopheide-esnet:
This latest version should support both cases whether you have multiple instances on a single node, or spread across nodes.  Metrics_port numbering will start at the same place for each unique instance$host (addr).

0xxon:
That sounds good to me. As @ckreibich is most familiar with this, I think he should take a peek :)

ckreibich:
Sincere apologies (again) for letting this sit so long. This certainly looks good to me. My main concerns right now are whether we should align Broker port assignment (in a subsequent PR), and whether this actually passes the cluster testsuite (which has a bunch of Prometheus checks), since it looks like it got skipped here. I'm looking into that.

ckreibich:
Cool, looks like this worked — but looking at one of the tests, I am not immediately sure why. :-) Digging into that one for a bit.

dopheide-esnet:
Lemme know if there's anything I can help with.

ckreibich:
Okay Dop I see what's going on. The fact that the test I mentioned on Slack (`prometheus-multihost`) continues to pass is indeed a bug. Here's what happens: when agents connect to the controller, it's perfectly fine to push a cluster config that just mentions the instances by name, with no IP address information. In that case the code (incorrectly) uses 0.0.0.0, treating the whole thing as one, so each instance's enumeration doesn't restart from the start port.

The logic needed already exists, it just is applied too late — after the enumeration. That's an easy fix. I'm thinking of creating a new PR with your commits and this additional fix, and then also pull in a testsuite update that adjusts the `prometheus-multihost` test. Does that sound good?

dopheide-esnet:
Yes, please go for it.  Thanks!

PR #3993, opened 2024-10-25
Skip modbus protocol testing with Spicy SSL enabled

awelzel:
@0xxon , think for the time we can skip the test?

This also introduces a helper similarly what we do for other features (af-packet, spicy, javascript, ...)


0xxon (approved):
Yup, sounds good to me.

PR #3995, opened 2024-10-28
account for spaces encoding to plus signs in sqli regex detection

cooper-grill:
If I run a quick test through DVWA and send `' OR 1=1#` over a GET request, the URI ends up looking like `/full/uri?id='+OR+1=1#`. This tool encodes spaces as plus signs (+), and this PR should address cases like this.

To summarize the changes, every time we see `[[:blank:]\x00-\x1f]` (or similar), add a `\+` to it.

Line [51](https://github.com/zeek/zeek/blob/fc42074c7cd2380396ede65da6fdad4da2bae9d4/scripts/policy/protocols/http/detect-sqli.zeek#L51) is the only pattern group that intentionally matches on a plus sign, so the space-to-plus-sign match is not added to the clause before the plus sign match (see columns 113 and 133).

### DIFF
--- a/scripts/policy/protocols/http/detect-sqli.zeek
+++ b/scripts/policy/protocols/http/detect-sqli.zeek
@@ -46,11 +46,11 @@ export {
 
 	## Regular expression is used to match URI based SQL injections.
 	const match_sql_injection_uri =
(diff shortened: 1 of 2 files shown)


awelzel:
@cooper-grill - thanks! Would you be able to capture a pcap while running the GET request agains DVWA and add a test? Or attach the pcap here so a test can be added during the merge.

Thanks!

cooper-grill:
@awelzel hopefully these tests are in the right place! Attached is the reference PCAP
[sql injection 2.pcapng.gz](https://github.com/user-attachments/files/17556338/sql.injection.2.pcapng.gz)

awelzel (approved):

awelzel:
Thank you for your contribution! This has been merged.

PR #3997, opened 2024-10-29
Fix Clang 19 deprecation failure

### LINKED ISSUE #3994: 7.0.3 build errors with FreeBSD 15 current and clang 19.1.2
pkg fallout just started complaining about zeek. Here's a representative example error:
```
/usr/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
  820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
/wrkdirs/usr/ports/security/zeek/work/zeek-7.0.3/src/DFA.cc:257:23: note: in instantiation of template class 'std::basic_string<unsigned char>' requested here
  257 |     for ( auto& entry : states ) {
      |                       ^
/usr/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
      |                             ^
```

The failed build log with all errors is here:

https://pkg-status.freebsd.org/beefy18/data/main-amd64-default/p645a502a4f60_s8b2e7da7085/logs/zeek-7.0.3.log

evantypanski:
This is a draft PR to get some CI runs, this kind of scares me with unintended consequences so having a list would be nice ;) .

Fixes #3994

Clang 19 with libc++ started failing to compile because the default implementation of `std::char_traits` was removed, making uses of `std::char_traits<unsigned char>` invalid (by consequence, also `std::basic_string<unsigned char>`). This was more of used for convenience before, but it should be roughly the same behavior with `char`.

Note that I initially wanted to just replace `u_char` with `char`, but that's probably bad because openssl expects `unsigned char` (and *lots* of other stuff) :/ That's pretty reasonable, though.

See relevant LLVM commits:

https://github.com/llvm/llvm-project/commit/aeecef08c385b1e4955155dd649a2d3724463849

https://github.com/llvm/llvm-project/commit/08a0faf4cd32bce6c51027ea9b5ec351747995b4

### DIFF
--- a/src/DFA.cc
+++ b/src/DFA.cc
@@ -265,9 +265,9 @@ DFA_State_Cache::~DFA_State_Cache() {
 DFA_State* DFA_State_Cache::Lookup(const NFA_state_list& nfas, DigestStr* digest) {
     // We assume that state ID's don't exceed 10 digits, plus
     // we allow one more character for the delimiter.
-    auto id_tag_buf = std::make_unique<u_char[]>(nfas.length() * 11 + 1);
+    auto id_tag_buf = std::make_unique<char[]>(nfas.length() * 11 + 1);
     auto id_tag = id_tag_buf.get();
-    u_char* p = id_tag;
+    char* p = id_tag;
 
     for ( int i = 0; i < nfas.length(); ++i ) {
         NFA_State* n = nfas[i];
@@ -287,7 +287,7 @@ DFA_State* DFA_State_Cache::Lookup(const NFA_state_list& nfas, DigestStr* digest
     // HashKey because the data is copied into the key.
     hash128_t hash;
     KeyedHash::Hash128(id_tag, p - id_tag, &hash);
-    *digest = DigestStr(reinterpret_cast<const unsigned char*>(hash), 16);
+    *digest = DigestStr(reinterpret_cast<const char*>(hash), 16);
 
     auto entry = states.find(*digest);
     if ( entry == states.end() ) {
--- a/src/DFA.cc
+++ b/src/DFA.cc
@@ -2,8 +2,6 @@
 
 #include "zeek/DFA.h"
 
-#include "zeek/zeek-config.h"
-
 #include "zeek/Desc.h"
 #include "zeek/EquivClass.h"
 #include "zeek/Hash.h"
--- a/src/analyzer/protocol/ssl/SSL.cc
+++ b/src/analyzer/protocol/ssl/SSL.cc
@@ -6,7 +6,6 @@
 #include <vector>
 
 #include "zeek/Reporter.h"
-#include "zeek/analyzer/Manager.h"
 #include "zeek/analyzer/protocol/ssl/events.bif.h"
 #include "zeek/analyzer/protocol/ssl/ssl_pac.h"
 #include "zeek/analyzer/protocol/ssl/tls-handshake_pac.h"
@@ -276,7 +275,7 @@ bool SSL_Analyzer::TryDecryptApplicationData(int len, const u_char* data, bool i
 
         // FIXME: should we change types here?
         const u_char* encrypted = data;
-        size_t encrypted_len = len;
+        int encrypted_len = len;
 
         if ( is_orig )
             c_seq++;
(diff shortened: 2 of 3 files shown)


evantypanski:
I think this is ok, I was expecting more blowups honestly.

@leres this branch should build for you if you want to test it. I was able to reproduce with Clang 19's libc++. The first commit fixes it

CI failures seem unrelated

src/analyzer/protocol/ssl/SSL.cc
```diff
@@ -266,21 +265,21 @@ bool SSL_Analyzer::TryDecryptApplicationData(int len, const u_char* data, bool i
...
+        int encrypted_len = len;
 
         if ( is_orig )
             c_seq++;
         else
             s_seq++;
 
         // AEAD nonce, length 12
-        std::basic_string<unsigned char> s_aead_nonce;
+        std::string s_aead_nonce;
```
awelzel, line 282: Switching s_aead_nonce and s_aead_tag to a std::vector<u_char> allows to remove most of the casts here.

https://github.com/zeek/zeek/commit/b04164b78f38eab03b35cf873cced49435e811e1

I'd be okay merging as is, this can be done as a follow-up, too.

src/DFA.cc
```diff
@@ -287,7 +285,7 @@ DFA_State* DFA_State_Cache::Lookup(const NFA_state_list& nfas, DigestStr* digest
     // HashKey because the data is copied into the key.
     hash128_t hash;
     KeyedHash::Hash128(id_tag, p - id_tag, &hash);
-    *digest = DigestStr(reinterpret_cast<const unsigned char*>(hash), 16);
+    *digest = DigestStr(reinterpret_cast<const char*>(hash), 16);
```
awelzel, line 288: Nothing to do - I guess it would be nice to be able to use `hash128_t` directly within a `std::map`, then we wouldn't even need the string. But that requires more surgery :-)

awelzel (approved):
Waiting for your thoughts on switching to std::vector<u_char> within SSL, but otherwise :+1:

src/analyzer/protocol/ssl/SSL.cc
evantypanski: agreed, I think adding this in is closer than using `std::string`. Added it to the first commit

src/DFA.cc
evantypanski: Yeah I saw that too, there's a lot of cleanup for these files just scanning through so I only did the low hanging fruit in the second commit (which, admittedly, wasn't much. A lot of narrowing conversions are around here though, too)

src/analyzer/protocol/ssl/SSL.cc
```diff
@@ -310,28 +314,28 @@ bool SSL_Analyzer::TryDecryptApplicationData(int len, const u_char* data, bool i
...
 
         // AEAD tag
-        std::basic_string<unsigned char> s_aead_tag;
+        byte_buffer s_aead_tag;
         if ( is_orig )
             s_aead_tag = fmt_seq(c_seq);
         else
             s_aead_tag = fmt_seq(s_seq);
 
+        assert(s_aead_tag.size() == 13);
```
0xxon, line 326: Is there a reason for the move of this assert?

Just looking at the code, it seems like it might always trigger after this change.

awelzel: if size wasn't at least 13, accessing `s_aead_tag[12]` would be invalid, so seemed reasonable to pre-assert before potentially accessing out-of-bounds.

evantypanski: I think it won't because it's reserved in `fmt_seq` which is always called https://github.com/zeek/zeek/blob/7b3bc891f4b4e132b8ec37b9b2d1b7f5235b2769/src/analyzer/protocol/ssl/SSL.cc#L39

@awelzel can confirm since that's from the [suggested commit](https://github.com/zeek/zeek/commit/b04164b78f38eab03b35cf873cced49435e811e1), but that was my understanding

PR #3999, opened 2024-10-29
CI: Add Ubuntu Clang task

evantypanski:
~This will be a draft because I couldn't get the cirrus CLI working on the config file here so it may take a try or two 🙊~

This feels more like a band-aid solution - it still requires very manual intervention to get the next builds, but at least this way we don't rely on whatever Clang/libc++ happens to ship with FreeBSD for the only libc++ testing. It seems like sanitizer builds don't use libc++ (they use clang), but I'd rather a dedicated Linux/Clang/libc++ task at least.

Related #3997 - if the new images were there, that issue would have been caught long before it was a real issue. This is intended to fail until that's merged

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -368,7 +368,7 @@ macos_sonoma_task:
 freebsd14_task:
   freebsd_instance:
     # FreeBSD 14 EOL: Nov 30 2028
-    image_family: freebsd-14-0
+    image_family: freebsd-14-1
     << : *FREEBSD_RESOURCES_TEMPLATE
 
   prepare_script: ./ci/freebsd/prepare.sh
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -268,6 +268,18 @@ ubuntu24_task:
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
 
+# Same as above, but using Clang and libc++
+ubuntu24_clang_libcpp_task:
+  container:
+    # Ubuntu 24.04 EOL: Jun 2029
+    dockerfile: ci/ubuntu-24.04/Dockerfile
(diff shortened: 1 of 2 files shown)


evantypanski:
Here are the failing builds from before the base branch swap to confirm these CI failures would have caught #3994 :

https://cirrus-ci.com/task/4849732365844480

https://cirrus-ci.com/task/6714504086552576

Now the base of this PR is the compilation failure fix - they can probably get merged together if they're good.

For the record, I upgraded the FreeBSD image, but the EOL seems different. Since it seems like that didn't matter in the past, I just kept it. I can just drop that commit if we'd rather not deal with that, though..

.cirrus.yml
```diff
@@ -268,6 +268,18 @@ ubuntu24_task:
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
 
+# Same as above, but using Clang and libc++
+ubuntu24_clang_libcpp_task:
+  container:
+    # Ubuntu 24.04 EOL: Jun 2029
+    dockerfile: ci/ubuntu-24.04/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
```
awelzel, line 277: I'm going to add `SKIP_TASK_ON_PR` - we'll still start to get errors/failures during nightly builds which should be good enough.

awelzel (approved):

timwoj:
> Clang/libc++ happens to ship with FreeBSD for the only libc++ testing. It seems like sanitizer builds don't use libc++ (they use clang), but I'd rather a dedicated Linux/Clang/libc++ task at least.

Moving the sanitizer builds to Clang was something I was tinkering with a while back, but at least for the asan build it broke everything about how we do code coverage. I ended up backing it all out.

PR #4000, opened 2024-10-30
RuleMatcher: Use a single list for tracking pattern_matches

awelzel:
@rsmmr - as discussed on #3996. How does that look, other thoughts?

### DIFF
--- a/src/RuleMatcher.cc
+++ b/src/RuleMatcher.cc
@@ -187,14 +187,24 @@ RuleEndpointState::RuleEndpointState(analyzer::Analyzer* arg_analyzer, bool arg_
     pia = arg_PIA;
 }
 
+const RuleEndpointState::RulePatternMatch* RuleEndpointState::FindRulePatternMatch(const Rule* r) const {
+    const auto it =
+        std::find_if(pattern_matches.begin(), pattern_matches.end(), [r](const auto& m) { return m.rule == r; });
+    if ( it != pattern_matches.end() )
+        return &(*it);
+
+    return nullptr;
+}
+
+void RuleEndpointState::AddRulePatternMatch(Rule* r, const u_char* data, int data_len, MatchPos end_of_match) {
+    pattern_matches.emplace_back(r, data, data_len, end_of_match);
+}
+
 RuleEndpointState::~RuleEndpointState() {
     for ( auto matcher : matchers ) {
         delete matcher->state;
         delete matcher;
     }
-
-    for ( auto text : matched_text )
-        delete text;
 }
 
 RuleFileMagicState::~RuleFileMagicState() {
@@ -865,11 +875,9 @@ void RuleMatcher::Match(RuleEndpointState* state, Rule::PatternType type, const 
                 continue;
 
             // Remember that all patterns have matched.
-            if ( ! state->matched_by_patterns.is_member(r) ) {
-                state->matched_by_patterns.push_back(r);
-                String* s = new String(data, data_len, false);
-                state->matched_text.push_back(s);
-                state->matched_text_end_of_match.push_back(match_end_pos - pre_match_pos);
+            if ( ! state->FindRulePatternMatch(r) ) {
+                MatchPos end_of_match = match_end_pos - pre_match_pos;
+                state->AddRulePatternMatch(r, data, data_len, end_of_match);
             }
 
             DBG_LOG(DBG_RULES, "And has not already fired");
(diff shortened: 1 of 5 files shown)


src/RuleMatcher.cc
```diff
@@ -865,11 +875,10 @@ void RuleMatcher::Match(RuleEndpointState* state, Rule::PatternType type, const
                 continue;
 
             // Remember that all patterns have matched.
-            if ( ! state->matched_by_patterns.is_member(r) ) {
-                state->matched_by_patterns.push_back(r);
-                String* s = new String(data, data_len, false);
-                state->matched_text.push_back(s);
-                state->matched_text_end_of_match.push_back(match_end_pos - pre_match_pos);
+            const auto* pattern_match = state->FindRulePatternMatch(r);
+            if ( pattern_match == nullptr ) {
```
rsmmr, line 879: Style nit, but I think we generally skip explicit comparisons to `nullptr`:

```suggestion
            if ( ! pattern_match ) {
```

src/RuleMatcher.cc
```diff
@@ -893,8 +902,8 @@ void RuleMatcher::FinishEndpoint(RuleEndpointState* state) {
 
     ExecPureRules(state, true);
 
-    loop_over_list(state->matched_by_patterns, i)
-        ExecRulePurely(state->matched_by_patterns[i], state->matched_text[i], state, true);
+    std::for_each(state->pattern_matches.begin(), state->pattern_matches.end(),
+                  [this, state](auto& m) { ExecRulePurely(m.rule, m.text, state, true); });
```
rsmmr, line 906: This is my personal preference, so YMMV, but I do prefer a simple for-loop over `std::for_each` any time ...

src/RuleMatcher.cc
```diff
@@ -997,11 +1006,11 @@ void RuleMatcher::ExecRule(Rule* rule, RuleEndpointState* state, bool eos) {
                 return;
             }
 
+
```
rsmmr, line 1009:
```suggestion
```

src/RuleMatcher.h
```diff
@@ -183,13 +213,8 @@ class RuleEndpointState {
...
     rule_hdr_test_list hdr_tests;
 
-    // The follow tracks which rules for which all patterns have matched,
-    // in a parallel list the (first instance of the) corresponding
-    // matched text, and in another parallel list the offset of the
-    // end of the last pattern match.
-    rule_list matched_by_patterns;
-    bstr_list matched_text;
-    match_offset_list matched_text_end_of_match;
+    // The follow all pattern matches for rules for which all patterns have matched.
```
rsmmr, line 216:
```suggestion
    // The following tracks all pattern matches for rules for which all patterns have matched.
```

src/RuleMatcher.cc
rsmmr: Actually, don't even need the tmp:

```
if ( ! state->FindRulePatternMatch(r) ) {
```

src/RuleMatcher.h
```diff
@@ -168,6 +169,35 @@ class RuleEndpointState {
     RuleEndpointState(analyzer::Analyzer* arg_analyzer, bool arg_is_orig, RuleEndpointState* arg_opposite,
                       analyzer::pia::PIA* arg_PIA);
 
+    // Tracking pattern matches for a given Rule.
+    class RulePatternMatch {
```
rsmmr, line 173: Optional: if all members are public, I personally like making it a `struct` to indicate so.

src/RuleMatcher.h
```diff
@@ -168,6 +169,35 @@ class RuleEndpointState {
...
+    public:
+        RulePatternMatch(Rule* rule, const u_char* data, int data_len, MatchPos end_of_match)
+            : rule(rule), text(new String(data, data_len, false)), end_of_match(end_of_match) {}
```
rsmmr, line 176: Suggest to use `unique_ptr` for `text` to avoid manual memory mgmt.

rsmmr: Or, actually, do we need this on the heap at all, vs. just storing a `String` directly inside the `class`?

rsmmr (approved):
Look good, just small notes.

src/RuleMatcher.cc
awelzel: Yep, thanks for mentioning!

PR #4008, opened 2024-11-04
btest: Add detect-sqli test script

awelzel:
...I added the baselines during the merge, just not the test itself.


evantypanski (approved):
😃

PR #4009, opened 2024-11-04
ZAM fix for inlining empty function bodies

vpax:
Today we tripped upon a ZAM bug that's only tickled when the `else` branch of a conditional is a call to an empty function that gets inlined. This PR fixes the bug and adds a regression BTest.


vpax:
Huh - standalone, that doesn't fail for me, but in the BTest indeed it does. Thanks for flagging! Will close this and look into it further.

PR #4014, opened 2024-11-06
Cleanup diff-remove-abspath, add better support for Windows paths

timwoj:
This makes some minor cleanup changes to diff-remove-abspath, plus adds better handling of Windows paths. The new handling isn't perfect (Windows has a *lot* of path variations), but it's good enough to cover our btests. These changes fix roughly 300 btests, bringing the number down from just under 700 to just under 400.

There's still some path oddities with Windows in the btests. In a couple of them, `reporter.log` has double backslash characters for all of the path separators. I dug into it a bit, and it looks like the Ascii formatter used by the Ascii writer is adding them, but I didn't dig far enough to figure out why.


timwoj:
Oh god, `sed` on macOS doesn't support `\t` tokens?

ckreibich (approved):
Approving, based on a mixture of parsing those JS docs and taking your word for it. ;-)

I'm wondering whether we should move this filter to Python. We already need it so we can run btest in the first place, and I'd imagine with pathlib this is a pretty clean one-liner.

PR #4015, opened 2024-11-07
Spicy: Do not raise an analyzer error when a connection is missing a regular tear-down.

### LINKED ISSUE #4007: Spicy analyzer raises protocol violation on read beyond connection end
Spicy and Binpac behave differently when a connection ends while a protocol unit is parsed. In Binpac, the connection is just silently closed, while Spicy seems to raise an protocol violation at the very end of the connection. This seems undesirable to me, as there are a variety of reasons why connections can be closed unexpectedly.

To reproduce this, run the short test in the private testsuite with the Spicy SSL analyzer enabled and check analyzer.log. Alternatively I have a pcap with the single offending connection that I can share. This pcap can be run with

```
zeek -r out.pcap "SSL::disable_analyzer_after_detection=F"
```

and will result in an analyzer.log entry containing

```
not enough bytes for skipping (0 bytes available) (/Users/johanna/zeek/main/src/analyzer/protocol/ssl/spicy/SSL.spicy:783:25-783:55)
```

rsmmr:
So far, when Zeek didn't see a connection's regular tear-down (e.g.,
because its state timed-out before we got to the end), we'd still
signal a regular end-of-data to Spicy parsers. As a result, they would
then typically raise a parse error because they were probably still
expecting data and would now declare it missing. That's not very
useful because semantically it's not really a protocol issue if the
data just doesn't make it over to us; it's a transport-layer issue
that Zeek already handles elsewhere. So we now switch to signaling
end-of-data to Spicy analyzers only if the connection indeed shuts
down regularly. This is also matches how BinPAC handles it.

This also comes with a test exercising various combinations of
end-of-data behavior so that we ensure consistent/desired behavior.

Closes #4007.


0xxon (approved):
Thank you very much. This looks good to me, assuming the tests pass.

PR #4020, opened 2024-11-08
ci: Add asan and ubsan zam tasks

awelzel:
This adds two new tasks (mirroing asan und ubsan) running `btest -a zam` if the labels fullci or zamci are used for PRs, or else during the regular nightly runs.


bbannier (approved):
Thanks! 🚀

PR #4021, opened 2024-11-08
policy/community-id: Populate conn$community_id in new_connection()

awelzel:
This wasn't possible before #3028 was fixed, but now it's safe to set the value in new_connection() and allow other users access to the field much earlier. We do not have to deal with connection_flipped() because the community-id hash is symmetric.


awelzel:
Thanks to @pauldokas  for mentioning.

PR #4022, opened 2024-11-11
ZAM fixes for ASAN issues

vpax:
This PR includes two types of fixes found using ASAN for ZAM BTest and large PCAP runs: one set involves vector operations, the other records with `any` fields. (One of the fixed problems affects ordinary interpreted execution rather than ZAM.)

One of the ZAM fixes currently results in a not-very-useful error message because it lacks location information. Fixing this is on my ZAM maintenance to-do list.


awelzel:
@vpax - I'm rebasing on latest master and force-pushing this/your branch. I've also labeled the PR zamci which will now result in `btest -a zam` running on Linux with ASAN and UBSAN in CI.

awelzel (approved):

PR #4024, opened 2024-11-11
Fix nullptr deref in Spicy accept/decline input

evantypanski:
Seems like this is a continuation of #4006

I was getting assertions in `hook_decline_input` running the Redis analyzer [on all ports](https://github.com/evantypanski/spicy-redis/tree/topic/etyp/all-ports) on [4SICS-GeekLounge-151022.pcap](https://share.netresec.com/s/gw6Y2QzJHqDD5pr/download/4SICS-GeekLounge-151022.pcap)

is there any better debug output that could be used? I don't think the accept input one has any usable info :/

### DIFF
--- a/src/spicy/manager.cc
+++ b/src/spicy/manager.cc
@@ -593,25 +593,25 @@ static ::TransportProto transport_protocol(const hilti::rt::Port port) {
 }
 
 static void hook_accept_input() {
-    auto cookie = static_cast<rt::Cookie*>(hilti::rt::context::cookie());
-    assert(cookie);
-
-    if ( auto x = cookie->protocol ) {
-        auto tag = spicy_mgr->tagForProtocolAnalyzer(x->analyzer->GetAnalyzerTag());
-        SPICY_DEBUG(hilti::rt::fmt("confirming protocol %s", tag.AsString()));
-        return x->analyzer->AnalyzerConfirmation(tag);
+    if ( auto cookie = static_cast<rt::Cookie*>(hilti::rt::context::cookie()) ) {
+        if ( auto x = cookie->protocol ) {
+            auto tag = spicy_mgr->tagForProtocolAnalyzer(x->analyzer->GetAnalyzerTag());
+            SPICY_DEBUG(hilti::rt::fmt("confirming protocol %s", tag.AsString()));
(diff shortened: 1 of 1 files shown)


src/spicy/manager.cc
```diff
@@ -593,25 +593,25 @@ static ::TransportProto transport_protocol(const hilti::rt::Port port) {
...
 static void hook_accept_input() {
-    auto cookie = static_cast<rt::Cookie*>(hilti::rt::context::cookie());
-    assert(cookie);
-
-    if ( auto x = cookie->protocol ) {
-        auto tag = spicy_mgr->tagForProtocolAnalyzer(x->analyzer->GetAnalyzerTag());
-        SPICY_DEBUG(hilti::rt::fmt("confirming protocol %s", tag.AsString()));
-        return x->analyzer->AnalyzerConfirmation(tag);
+    if ( auto cookie = static_cast<rt::Cookie*>(hilti::rt::context::cookie()) ) {
+        if ( auto x = cookie->protocol ) {
```
bbannier, line 597: This feels more like a one-off workaround and we would need something similar for all accesses to `protocol`. I think the real bug is that we somehow manage to set `cookie`, but do not set `protocol`.

I think the `Cookie` should come in properly set up from the outside depending on what kind of analyzer we have. Could you check values for `EndpointState`, `PacketState` and `FileState` (they all take a `Cookie` in their ctrs and pass it down)? I think the real error is somewhere up there in their uses.

evantypanski: I don't quite understand since the only change is swapping out the assertion checking for `cookie` with an `if` like #4006 - I think in this case it's not set because there's a synchronization failure on the responder side, then the originator rejecting the input later gets a null cookie.

So, it doesn't seem like there's anything unexpected happening here to me, just seems like it's a missed case assuming the cookie is set. Or am I missing something?

bbannier: Sorry, I got confused. The check for `cookie->protocol` was already there, and you just fixed the access to `cookie` which is definitely a change we want.

bbannier (approved):

src/spicy/manager.cc
evantypanski: Github diffs strike again :)

PR #4025, opened 2024-11-11
PktSrc/RunState: Scale on first_wallclock and move pseudo realtime logic to RunState

awelzel:
@vpax observed spurious failures with ASAN+ZAM+pseudo-realtime use.

I *think* this is mainly because `zeek_start_time` is used for scaling of pseudo time which might differ significantly from `first_wallclock` (which is the wallclock at which the first packet was consumed).

This PR moves away from zeek_start_time. While at it, move some of the pseudo realtime logic from PktSrc into RunState, which seems the better place.

Anyone brave enough to review? It's pseudo-realtime, so I don't worry about it too much - in live deployments it should never be used.

### DIFF
--- a/src/RunState.cc
+++ b/src/RunState.cc
@@ -438,14 +438,17 @@ double current_timestamp = 0.0;
 static int _processing_suspended = 0;
 
 void suspend_processing() {
-    if ( _processing_suspended == 0 )
+    if ( _processing_suspended == 0 ) {
+        DBG_LOG(DBG_MAINLOOP, "processing suspended");
         reporter->Info("processing suspended");
+    }
 
     ++_processing_suspended;
 }
 
 void continue_processing() {
     if ( _processing_suspended == 1 ) {
+        DBG_LOG(DBG_MAINLOOP, "processing continued");
         reporter->Info("processing continued");
         detail::current_wallclock = util::current_time(true);
     }
--- a/src/RunState.cc
+++ b/src/RunState.cc
@@ -129,8 +129,6 @@ void update_network_time(double new_network_time) {
 static bool should_forward_network_time() {
     // In pseudo_realtime mode, always update time once
     // we've dispatched and processed the first packet.
-    // run_state::detail::first_timestamp is currently set
-    // in PktSrc::ExtractNextPacketInternal()
     if ( pseudo_realtime != 0.0 && run_state::detail::first_timestamp != 0.0 )
         return true;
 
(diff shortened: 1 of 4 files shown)


src/RunState.cc
```diff
@@ -399,10 +406,19 @@ void delete_run() {
...
 
 double check_pseudo_time(const Packet* pkt) {
+    assert(pkt->time > 0.0);
+    assert(first_wallclock > 0.0);
+    assert(first_timestamp > 0.0);
     double pseudo_time = pkt->time - first_timestamp;
     double ct = (util::current_time(true) - first_wallclock) * pseudo_realtime;
 
-    current_pseudo = pseudo_time <= ct ? zeek_start_time + pseudo_time : 0;
+    current_pseudo = pseudo_time <= ct ? first_wallclock + pseudo_time : 0;
```
awelzel, line 415: This used first_wallclock rather than zeek_start_time now.

awelzel:
I've manually triggered `asan_sanitizer_zam` for this PR and the last failing test is now the following which Vern is tracking elsewhere.
```
[#2] core.file-analyzer-violation [zam] ... ok
[#4] core.expr-exception [zam] ... failed
  % 'zeek -b -r $TRACES/wikipedia.trace base/protocols/ftp base/protocols/http base/frameworks/reporter /zeek/testing/btest/.tmp/core.expr-exception/expr-exception.zeek >output' failed unexpectedly (exit code 1)
  % cat .stderr
  1300475168.783842 runtime error in /zeek/testing/btest/.tmp/core.expr-exception/expr-exception.zeek, line 10: field value missing: $ftp
  1300475168.915940 runtime error in /zeek/testing/btest/.tmp/core.expr-exception/expr-exception.zeek, line 10: field value missing: $ftp
```

Previously a few of the `--pseudo-realtime` based cluster tests failed due to ZAM taking too long to compile/optimize.

src/RunState.h
```diff
@@ -70,6 +70,7 @@ extern void suspend_processing();
 extern void continue_processing();
 bool is_processing_suspended();
 
+[[deprecated("Remove with v8.1. Use run_state::current_pseudo directly, but you probably should not")]]
 extern double current_packet_timestamp();
```
rsmmr: Let me just double check: Iirc, it's called `current_packet_timestamp()` because it used to give you whatever semantically was the current timestamp, either the actual time of a live packet or the pseudo time in pseudo-realtime-mode. In other words, the caller didn't need to worry about which mode we're in. 

That is no longer the case, correct? I see that it now always returns `current_pseudo`  (as you say) and I assume that's not set to anything reasonable when not in PRM?

rsmmr (approved):
Look ok and reasonable to me, except that t I don't get why this would trigger/solve any ASAN issues?

awelzel:
>  I don't get why this would trigger/solve any ASAN issues?

Ah, sorry. It fixes test failures when running with ZAM under ASAN. The startup sequence was slow enough for Zeek to expire the connection created by the first packet (zeek_start_time being multiple seconds in the past making, upsetting timer management).

src/RunState.h
awelzel: > That is no longer the case, correct?

Correct and I actually don't see when it ever was - 69da2d7b1d215832f1d948cefb4d01e7ea698a03.

And yes, current_pseudo is only set with pseudo realtime. So I figured I'd just remove the generic accessor. `current_packet_wallclock()` is similarly, but it does more that I din't dare to touch for now.

PR #4028, opened 2024-11-12
CI updates

ckreibich:
I'm doing a quick update round since I may be seeing some build problems on Fedora 41.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -387,7 +387,7 @@ freebsd14_task:
 freebsd13_task:
   freebsd_instance:
     # FreeBSD 13 EOL: January 31, 2026
-    image_family: freebsd-13-3
+    image_family: freebsd-13-4
     << : *FREEBSD_RESOURCES_TEMPLATE
 
   prepare_script: ./ci/freebsd/prepare.sh
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -174,17 +174,17 @@ env:
 # Linux EOL timelines: https://linuxlifecycle.com/
 # Fedora (~13 months): https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle
 
+fedora41_task:
+  container:
+    # Fedora 41 EOL: Around Nov 2025
+    dockerfile: ci/fedora-41/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
+
 fedora40_task:
   container:
     # Fedora 40 EOL: Around May 2025
     dockerfile: ci/fedora-40/Dockerfile
-    << : *RESOURCES_TEMPLATE
-  << : *CI_TEMPLATE
-
-fedora39_task:
-  container:
-    # Fedora 39 EOL: Around Nov 2024
-    dockerfile: ci/fedora-39/Dockerfile
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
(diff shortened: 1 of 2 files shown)


ckreibich:
@awelzel I _think_ this is due to Fedora switching to Node 22, see [here](https://cirrus-ci.com/task/4661200397533184?logs=build#L2768-L2795)?

awelzel:
> @awelzel I think this is due to Fedora switching to Node 22, see [here](https://cirrus-ci.com/task/4661200397533184?logs=build#L2768-L2795)?

Thanks, there's some deprecations incoming, but this should do for the time being: https://github.com/corelight/zeekjs/pull/92

0xxon (approved):

0xxon:
Oh, actually, Fedora 41 has some... interesting SSL test failures. Let me take a look at those.

I have a suspicion that they might use a signature algorithm that is no longer supported in the OpenSSL version that Fedora 41 uses.

0xxon:
Ok, after digging at this for too long - one has to call `update-crypto-policies --set LEGACY` for this to still allow certificates with SHA1 signatures - which we have in this test.

It looks like way this is implemented is that Fedora has a patch to OpenSSL that allows them to explicitly disallow all SHA1 signatures. https://fedoraproject.org/wiki/Changes/OpenSSLDistrustSHA1SigVer has more details.

This will come back in a couple of years - at which point we might just have to disregard these tests :)

I am going to push a dockerfile update in a bit.

awelzel:
> This will come back in a couple of years - at which point we might just have to disregard these tests :)

Doesn't this potentially affect any Fedora 41 users now when using validate-certs.zeek?

Could we do something like what's mentioned in the docs and spill a warning via reporter with a suggestion to use update-crypto-policies? Not necessarily for this PR, but seems a worthwhile follow-up?

https://fedoraproject.org/wiki/SHA1SignaturesGuidance#Detecting_SHA-1_signature_support_at_runtime

> Your application may want to detect whether SHA-1 signatures are supported at runtime. Unfortunately, OpenSSL currently doesn't offer a dedicated API to do this. For now, we recommend that you attempt verification of a known-good SHA-1 signature to determine whether they're supported. We're planning to work with OpenSSL upstream to potentially introduce a more suitable API for supported digests in signatures in the future.

0xxon:
I think calling this out in NEWS is fair - and we should do it for people using Fedora.

That being said - I am not sure that I would even recommend peope to use update-crypto-policies. There is no good reason for SHA1 certificates to still exist. I expect that the only people that this will be interesting for are those with special reequirements - like replaying old traces (which is how we noticed it).

Thus, I am also not sure that I think it's worthwhile to write a detection for this. We could totally write a small Zeek script that finds out that this is going on - but honestly - in most cases this is a good thing...

ckreibich:
I'm downgrading the policy change to `FEDORA40` since it is a smaller changeset than `LEGACY` vs the default one, and the one they recommend at least for several releases to come.

I'm okay with that change for this PR so we get CI going again, but I don't think it is the whole answer. If our tests rely on trust in SHA-1 to work, that's a requirement of the test, not something that the platform should need to work around — so we should fix them or add `@TEST-REQUIRES` tests.  I'll open a follow-up ticket.

I was about to add NEWS coverage but it's awkward since there isn't a Zeek change to cover. I'm thinking of calling it out in the installation section in the manual instead, and it's perhaps the kick in the rear that I needed to get the new FAQs out the door. :-)

PR #4029, opened 2024-11-13
Backporting ZAM fixes for 7.0.x

vpax:
This PR collects together a number of ZAM fixes for incorporation into 7.0.x. The first, for the ZAM `cat` built-in, is a direct cherry-pick. The next 3 are ports of recent issues, all of which required a fair amount of translation due to the major changes in `gen-zam` and the corresponding ZAM operation descriptions that have been subsequent to 7.0. The last one started as a cherry-pick but it turned out that it relied on elements of some previous recent changes, and one custom component that goes away with the `gen-zam` changes.

On my laptop these changes pass the full test suite, and all of the `-a zam` test suite other than one trivial difference that I suspect reflects a ZAM baseline issue in 7.0 itself, and in any case isn't of any concern.


timwoj (commented):
This all looks fine to me after comparing it to the PRs the changes are based on. The only question I had was whether the tests for  f693f221922ce029c4bcb77df7e8d30112b2fe54 (from https://github.com/zeek/zeek/pull/4022/files) should be backported as well.

vpax:
> The only question I had was whether the tests for [f693f22](https://github.com/zeek/zeek/commit/f693f221922ce029c4bcb77df7e8d30112b2fe54) (from https://github.com/zeek/zeek/pull/4022/files) should be backported as well.

My thinking was to keep things as simple/minimal as possible, but if you'd prefer to have the tests then I'll go ahead and add them.

timwoj:
> > The only question I had was whether the tests for [f693f22](https://github.com/zeek/zeek/commit/f693f221922ce029c4bcb77df7e8d30112b2fe54) (from https://github.com/zeek/zeek/pull/4022/files) should be backported as well.
> 
> My thinking was to keep things as simple/minimal as possible, but if you'd prefer to have the tests then I'll go ahead and add them.

When we cherry-pick in PRs we generally bring over the whole set of changes, so it's preferable to have them. Plus it keeps the release branch stable over time if the tests exist.

vpax:
> When we cherry-pick in PRs we generally bring over the whole set of changes, so it's preferable to have them. Plus it keeps the release branch stable over time if the tests exist.

Okay these are now ported over. And it was (of course) a good suggestion, as it flagged a glitch in one of the earlier manual ports.

timwoj (approved):

PR #4031, opened 2024-11-13
CI: Fedora 41 + Ubuntu 24.10

0xxon:
Also update Fedora version on PR to 41.


0xxon:
Ok, I missed that Christian did the same with F41 - so let me try this again...

PR #4033, opened 2024-11-13
Pluggable Cluster Backends - Part 1

awelzel:
This mostly adds infrastructure to install serializer and backend components.

This PR has no impact on existing code paths.


timwoj (changes requested):

src/cluster/Serializer.h
```diff
@@ -0,0 +1,119 @@
...
+#include <string>
+#include <vector>
+
+#include "zeek/Span.h"
+#include "zeek/logging/Types.h"
+
+namespace zeek {
+
+template<class T>
+class IntrusivePtr;
```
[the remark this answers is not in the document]
awelzel: Indeed! Leftovers when I was dabbling with adding `Val` serialization. Thanks!

src/logging/Types.cc
```diff
@@ -0,0 +1,50 @@
...
+                               std::string arg_path)
+    : stream_id(std::move(arg_stream_id)),
+      writer_id(std::move(arg_writer_id)),
+      filter_name(std::move(arg_filter_name)),
+      path(std::move(arg_path)) {
+    stream_name = obj_desc_short(stream_id.get());
+    writer_name = obj_desc_short(writer_id.get());
+}
+
+LogWriteHeader& LogWriteHeader::operator=(const LogWriteHeader& other) = default;
```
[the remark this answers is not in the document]
awelzel: I want to say I did this because technically the header needs a `Val.h` include to instantiate the defaulted ones, otherwise you get these kind of errors:
```
/home/awelzel/corelight-oss/zeek/src/include/zeek/IntrusivePtr.h:117:17: error: no matching function for call to 'Unref'
  117 |                 Unref(ptr_);
      |                 ^~~~~
/home/awelzel/corelight-oss/zeek/src/include/zeek/logging/Types.h:40:5: note: in instantiation of member function 'zeek::IntrusivePtr<zeek::EnumVal>::~IntrusivePtr' requested here
   40 |     LogWriteHeader() = default;
      |     ^
/home/awelzel/corelight-oss/zeek/src/include/zeek/Obj.h:172:13: note: candidate function not viable: cannot convert argument of incomplete type 'pointer' (aka 'zeek::EnumVal *') to 'Obj *' for 1st argument
  172 | inline void Unref(Obj* o) {
      |             ^     ~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
```

By moving `= default` into Types.cc, we can do the `#include "zeek/Val.h" just there.

Am I being to avoiding of that "zeek/Val.h" include?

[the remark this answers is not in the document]
awelzel: I've removed the move constructor, but otherwise left a comment why it's looking like that. If you think that's not sensible, lets chat :-)

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+    virtual void DoTerminate() = 0;
+
+    /**
+     * Given the arguments for the \a Cluster::make_event() function, create
+     * a script-level record value that represents the event.
+     *
+     * The default implementation produces the \a Cluster::Event script type,
+     * but may be overridden by implementations.
+     *
+     * Note that this method exists primarily for the broker backend. Do not
```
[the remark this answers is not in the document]
awelzel: I think that would be the idea moving forward, but right now it's "broker does it's own thing in a few places" which I didn't want to touch. I extended the comment to contain "removal" in the future.

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+     *
+     * It's safe to call this method from other threads.
+     */
+    void QueueForProcessing(QueueMessages&& qmessage);
+
+    void Process() override;
+
+    double GetNextTimeout() override { return -1; }
+
+    /**
```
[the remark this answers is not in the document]
awelzel: I'll update to this, I think that's better, but feel free to wordsmith :-)
```
418      * Classes deriving from ThreadedBackend and providing their                                                                                                                                                                                                                                                                                                                          
419      * own DoInitPostScript() method should invoke the ThreadedBackend's                                                                                                                                                                                                                                                                                                                  
420      * implementation to register themselves as a non-counting                                                                                                                                                                                                                                                                                                                            
421      * IO source with the IO loop.    
```

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+                                    zeek::Span<zeek::logging::detail::LogRecord> records);
+
+    /**
+     * Send out a serialized log batch.
+     *
+     * A backend implementation may use the values from \a header to
+     * construct a topic to write the logs to.
+     *
+     * Semantics of this call are "fire-and-forget". An implementation should
+     * ensure the message is enqueue for delivery, but may not have been send out
```
ckreibich, line 326:
```suggestion
     * ensure the message is enqueued for delivery, but may not have been sent,
```

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+     * let alone received by the destination when this call returns.
+     *
+     * Sharding log writes to multiple receivers (logger nodes) is
+     * backend specific. Broker, for example, involves script layer
+     * cluster pool concepts, but other backends may use native mechanisms
+     * that may be more efficient.
+     *
+     * @param header The header describing the writer frontend where the records originate from.
+     * @param buf The serialized log batch. This is the message payload.
```
ckreibich, line 335: (Here too, add `@param` for `format`)

src/cluster/Component.h
```diff
@@ -0,0 +1,131 @@
...
+
+class BackendComponent : public plugin::Component {
+public:
+    using factory_callback = std::unique_ptr<Backend> (*)(std::unique_ptr<EventSerializer>,
+                                                          std::unique_ptr<LogSerializer>);
+
+    /**
+     * Constructor.
+     *
+     * @param name The name of cluster backend. An Zeek script level enum
```
ckreibich, line 21:
```suggestion
     * @param name The name of cluster backend. A Zeek script-level enum
```

src/cluster/Component.h
```diff
@@ -0,0 +1,131 @@
...
+
+
+class EventSerializerComponent : public plugin::Component {
+public:
+    using factory_callback = std::unique_ptr<EventSerializer> (*)();
+
+    /**
+     * Constructor.
+     *
+     * @param name The name of cluster backend. An Zeek script level enum
```
ckreibich, line 61:
```suggestion
     * @param name The name of cluster backend. A Zeek script-level enum
```

src/cluster/Component.h
```diff
@@ -0,0 +1,131 @@
...
+};
+
+class LogSerializerComponent : public plugin::Component {
+public:
+    using factory_callback = std::unique_ptr<LogSerializer> (*)();
+
+    /**
+     * Constructor.
+     *
+     * @param name The name of cluster backend. An Zeek script level enum
```
ckreibich, line 100:
```suggestion
     * @param name The name of cluster backend. A Zeek script-level enum
```

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+#include "zeek/cluster/Serializer.h"
+#include "zeek/iosource/IOSource.h"
+#include "zeek/logging/Types.h"
+
+namespace zeek {
+
+class Val;
+class FuncVal;
+
+using ValPtr = IntrusivePtr<Val>;
```
ckreibich, line 26: I think this line can go?

src/cluster/Backend.cc
```diff
@@ -0,0 +1,276 @@
...
+    return zeek::cluster::detail::Event{handler, std::move(*checked_args), timestamp};
+}
+
+zeek::RecordValPtr Backend::DoMakeEvent(zeek::ArgsSpan args) {
+    static const auto& any_vec_type = zeek::id::find_type<zeek::VectorType>("any_vec");
+    static const auto& event_record_type = zeek::id::find_type<zeek::RecordType>("Cluster::Event");
+    auto rec = zeek::make_intrusive<zeek::RecordVal>(event_record_type);
+
+    if ( args.size() < 1 ) {
+        zeek::reporter->Error("not enough arguments to Cluster::make_event()");
```
ckreibich, line 81: `Cluster::make_event()` isn't a thing yet? Also applies to similar comments in the header file. It was probably just too tedious to split cleanly, and that's okay ... but I'm flagging in case it would make sense to introduce more of this in Part 2.

src/logging/Types.h
```diff
@@ -0,0 +1,98 @@
...
+ * send via the cluster to a remote node.
+ */
+using LogRecord = std::vector<threading::Value>;
+
+/**
+ * A struct holding all necessary information that relates to
+ * log writes for a given path. These values are constant over
+ * the lifetime of a \a WriterFrontend.
+ */
+struct LogWriteHeader {
```
ckreibich, line 36: What's the intuition for "Header" in the name?

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -280,6 +280,14 @@ export {
 	## Returns: a topic string that may used to send a message exclusively to
 	##          a given cluster node.
 	global nodeid_topic: function(id: string): string;
+
+	## Untyped remote event.
+	type Event: record {
+		## The function to be invoked as an event on the remote side.
```
ckreibich, line 286: Would it be better here to say "the event handler", even if it's technically a function down in the guts of the script layer? I'm flagging this because for script developers functions, hooks, and events are generally distinct entities.

testing/btest/Baseline/scripts.base.files.x509.files/files.log
```diff
@@ -7,10 +7,10 @@
...
-XXXXXXXXXX.XXXXXX	F1H4bd2OKGbLPEdHm4	ClEkJM2Vm5giqnMf4h	192.168.4.149	60624	74.125.239.129	443	SSL	0	SHA256,X509,SHA1,MD5	application/x-x509-ca-cert	-	0.000000	F	F	1032	-	0	0	F	-	9e4ac96474245129d9766700412a1f89	d83c1a7f4d0446bb2081b81a1670f8183451ca24	a047a37fa2d2e118a4f5095fe074d6cfe0e352425a7632bf8659c03919a6c81d
-XXXXXXXXXX.XXXXXX	Fgsbci2jxFXYMOHOhi	ClEkJM2Vm5giqnMf4h	192.168.4.149	60624	74.125.239.129	443	SSL	0	SHA256,X509,SHA1,MD5	application/x-x509-ca-cert	-	0.000000	F	F	897	-	0	0	F	-	2e7db2a31d0e3da4b25f49b9542a2e1a	7359755c6df9a0abc3060bce369564c8ec4542a3	3c35cc963eb004451323d3275d05b353235053490d9cd83729a2faf5e7ca1cc0
+XXXXXXXXXX.XXXXXX	FgN3AE3of2TRIqaeQe	CHhAvVGS1DHFjwGM9	192.168.4.149	60623	74.125.239.129	443	SSL	0	X509,SHA256,SHA1,MD5	application/x-x509-user-cert	-	0.000000	F	F	1859	-	0	0	F	-	7af07aca6d5c6e8e87fe4bb34786edc0	548b9e03bc183d1cd39f93a37985cb3950f8f06f	6bacfa4536150ed996f2b0c05ab6e345a257225f449aeb9d2018ccd88f4ede43
```
ckreibich, line 10: Where does the reordering stem from?

ckreibich (changes requested):
This looks great Arne — I just have a few questions, the main one being about the `Backend` APIs that are specific to Broker. The change requests are all minor stuff.

src/cluster/Backend.h
```diff
@@ -0,0 +1,457 @@
...
+class Event {
+public:
+    // When an event is published from script land, the handler is known
+    // as a FuncVal. When an event is deserialized from a message, it's
+    // easier to use an EventHandler from the registry so the event can
+    // be enqueued directly.
+    //
+    // It seems there's no direct translation between EventHandlerPtr and
+    // FuncValPtr without going through the global scope or event registry,
+    // so this is done via a variant here.
```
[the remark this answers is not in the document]
awelzel: > Also, this isn't technically doing any translating, it just allows working with either type — right?

Yeah, this is the motivation. If such an event is created from the script layer (via publish() or make_event()), we're receiving a FuncValPtr and that is enough for serialization and validation purposes.

On the flip-side, to enqueue a received remote event, we require a EventHandlerPtr and there's an (undocumented...) requirement that deserialization produces that.

The variant exists to work with both cases, but I'm starting to lean to just lookup the EventHandlerPtr through the registry - doubt it'll make a substantial perf difference but remove some weirdness. Thanks!

src/cluster/Backend.h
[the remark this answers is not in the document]
awelzel: Related to Tim's question if that should eventually go away:

`Broker::make_event()` returns a `Broker::Event` and `Cluster::make_event()` will also return a `Broker::Event` when running with the Broker backend enabled. (Motivation here is that you can `s/Broker::make_event()/Cluster::make_event()/g` and not jeopardize Broker requiring an extra conversion.

> When would I need to reimplement it?

If `Cluster::Event` doesn't fit your purpose, but hopefully never. The motivation is a bit reversed, however: The new backends don't use `Broker::Event`.

If we switch broker over to using `Cluster::Event` and moving serialization out of broker, this can go away, but I didn't want to go there at this point.

src/cluster/Backend.cc
awelzel: > It was probably just too tedious to split cleanly, and that's okay

Too tedious fits: I basically added whole files, accepting that there will be forward references in places. I could see pieces of `cluster.bif` being added here already, but not used.

src/logging/Types.h
awelzel: Ha, good one :-)

When a cluster node sends out a batch of log records, the message starts with a description of the batch to allow stateless interpretation of the batch (stream, writer, path, schema/fields). That's the "Header". And the "LogWrite" is more like "LogBatch".

Maybe "LogBatchHeader", let me know what you think, I'm a bit stuck in a box here :-)

testing/btest/Baseline/scripts.base.files.x509.files/files.log
awelzel: I guess it's just old news to me that whenever something major changes with enums, this file needs an update :-)


https://github.com/zeek/zeek/pull/2340#issuecomment-1216306937
https://github.com/zeek/zeek/pull/2770#issuecomment-1434330771

I've tried before to fix, but it's not very simple:
https://github.com/zeek/zeek/issues/3364
https://github.com/zeek/zeek/pull/3365

src/cluster/Component.h
awelzel: Thanks! I think I also fixed up all non-fitting references to cluster backend.

src/cluster/Backend.h
awelzel: It's used below for ArgsSpan - I'll reorder a bit.

src/cluster/Backend.h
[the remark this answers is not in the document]
awelzel: After pondering: I start to think this is confusing and complex and `Cluster::make_event()` should produce the same event for all backends.

My only worry is that something like the following (which today is working but seemingly barely used) would break at runtime when using a non-broker backend.
```
local e = Broker::make_event(...)
Cluster::publish_hrw(topic, key, e);
```

But maybe a `Reporter::error()` is enough for these cases. And even more maybe we could attempt to detect these via AST traversal.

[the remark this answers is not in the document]
awelzel: I have removed `MakeEvent()` and `DoMakeEvent()` from the `Backend` class and moved most logic into BifSupport.cc. This feels better even with the above being a bit quirky.

The fixups live in `topic/awelzel/pluggable-cluster-backends-part4-with-fixups`, but I'll fold them in soon.

src/cluster/Backend.h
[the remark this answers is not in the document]
awelzel: I've removed the variant to do the lookup where needed instead.

src/logging/Types.cc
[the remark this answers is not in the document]
timwoj: Gah, I've run into that before. I think what you did is sensible. Resolving.

src/cluster/Backend.cc
ckreibich: No problem, thanks.

src/logging/Types.h
ckreibich: Thanks! Believe me when I say that I always at least hesitate when I bring up naming stuff. :-)

I think what tripped me up was that the members don't really have anything to do with a header, and I see now that it's just context that the serializer then turns into a "header" write. So perhaps something like `LogWriteInfo`. But, no biggie.

testing/btest/Baseline/scripts.base.files.x509.files/files.log
ckreibich: Oooh wow. That's a pretty detached change to surface here. Thanks.

src/cluster/Backend.h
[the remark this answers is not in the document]
ckreibich: Cool, thanks a bunch. I get it now. And I like that iteration a lot. :+1:

ckreibich (approved):
Right on, I'm on to Part 2 then! :-)

src/logging/Types.h
awelzel: I kept this one for now, but certainly up for post-merge-rename!

PR #4037, opened 2024-11-14
 Pluggable Cluster Backends - Part 3

awelzel:
Mostly logging related changes:

* WriterFrontends have a new LogWriteHeader
* Non-broker cluster backends re-use the WriterFrontend log write buffering and flushing
  * That one is a bit hairy (logging/WriterFrontend: Add logic for non-broker cluster backends)
* Non-broker cluster backends will use a new `WritesFromRemote()` entry point in the logging manager
* Introduces a log serializer using SerialTypes named binary-serialization-format. Similar to broker, but the whole batch uses SerialTypes logic. This can be replaced in the future with something more efficient.


src/logging/Manager.cc
```diff
@@ -1731,6 +1731,79 @@ WriterFrontend* Manager::CreateWriterForFilter(Filter* filter, const std::string
                         from_remote, filter->name);
 }
 
+bool Manager::WritesFromRemote(const detail::LogWriteHeader& header, std::vector<detail::LogRecord>&& arg_records) {
+    auto records = std::move(arg_records);
```
timwoj, line 1735: I'm confused by the arguments here. I understand the header being const-reference since you're only using the values out of it and not copying it. The rvalue-reference for arg_records doesn't make as much sense. If you're going to immediately `std::move` it, why not make it pass-by-value and have the caller std::move into it? Then you don't need the extra `std::move` at the start of the method.

timwoj (changes requested):
Minor nits and one confusion about argument types. Assuming I'm just confused and it's fine, I'll approve after.

src/logging/Manager.cc
awelzel: I've used the rvalue here as the method is intended to be a sink and taking ownership, plus preventing the caller from mistakenly creating a copy (which for a batch of log writes might be significant).

Elsewhere there exists the following which broker currently calls - it's somewhat consistent with that also (though I was the one that added this)
```
bool WriteFromRemote(EnumVal* id, EnumVal* writer, const std::string& path, detail::LogRecord&& rec);
```

> Then you don't need the extra std::move at the start of the method.

I only added this one as sonarlint warned that I never moved `arg_records` so I added that one. I can also just remove it assuming nothing else warns.

The actual move happens per record further down:
```
    for ( auto& r : records )
        w->second->writer->Write(std::move(r));
```

I like the rvalue to indicate the passing of ownership, but if you think that's quirky, maybe we can change both APIs as a follow-up?

awelzel: > I only added this one as sonarlint warned that I never moved arg_records so I added that one. I can also just remove it assuming nothing else warns.

I've removed the `auto record = std::move(...)` for now.

timwoj: > I like the rvalue to indicate the passing of ownership, but if you think that's quirky, maybe we can change both APIs as a follow-up?

I think this is fine. Maybe a note on the function comment indicating that that's the case.

src/logging/Manager.h
```diff
@@ -287,6 +288,20 @@ class Manager : public plugin::ComponentManager<Component> {
      */
     bool WriteFromRemote(EnumVal* id, EnumVal* writer, const std::string& path, detail::LogRecord&& rec);
 
+    /**
+     * Writes out a batch of log entries received from remote nodes.
+
```
ckreibich, line 293: Maybe clarify how this relates to `WriteFromRemote` since their comments sound so similar ("writes out log entries" vs "writes out a batch of log entries"). Unlike with `DoPublishlogWrites` there's an ambiguity with "write" and "writes" working as a verb and/or a noun. Maybe `WriteBatchFromRemote()`? (Definitely minor.)

Btw did the formatter put these empty lines here or did you do this? :-) It's a bit unsual.

src/logging/Manager.cc
```diff
@@ -1730,6 +1731,88 @@ WriterFrontend* Manager::CreateWriterForFilter(Filter* filter, const std::string
...
+                            filter->name.c_str(), header.stream_name.c_str());
+            return false;
+        }
+    }
+
+    w = stream->writers.find(wpp);
+    assert(w != stream->writers.end());
+
+    // Write each record individually. This results in the
+    // frontends buffering to be in effect.
```
ckreibich, line 1805: I don't fully follow this comment — you mean it builds up buffering in the frontend because we don't flush explicitly? (Is it "frontend's buffering"?)

ckreibich (approved):
LGTM Arne, I only have questions and suggestions that don't rise to the level of a change request. I'm a bit nervous about special-casing the Broker backend / `broker_mgr` in more places in the code, but that seems unavoidable to be on the safe side for now. I'm curious whether it's already clear to you how replacing `broker_mgr` will unfold, or whether surprises might lurk in that still. Let's talk about that some time.

awelzel:
> I'm curious whether it's already clear to you how replacing broker_mgr will unfold, or whether surprises might lurk in that still. Let's talk about that some time.

Currently, Broker does log write load balancing per-record and buffering internally. Other backends rely on the frontend's log record buffering (1000 records, 1 second by default) and should distribute full batches among loggers in whichever way they see fit.

My idea for unifying was to switch broker to the same model.

src/logging/Manager.cc
awelzel: It's "frontend's buffering", yes. And yes, if 8 nodes send log batches of a small size at about the same time, the logger would still just flush its own buffer once, rather than 8 times.

src/cluster/serializer/binary-serialization-format/Plugin.cc
```diff
@@ -0,0 +1,19 @@
...
+using namespace zeek::plugin::Zeek_Binary_Serializer;
+using namespace zeek::cluster;
+
+zeek::plugin::Configuration Plugin::Configure() {
+    AddComponent(new LogSerializerComponent("ZEEK_BIN_V1", []() -> std::unique_ptr<LogSerializer> {
+        return std::make_unique<cluster::detail::BinarySerializationFormatLogSerializer>();
+    }));
+
+    zeek::plugin::Configuration config;
+    config.name = "Zeek::Binary_Serializer";
```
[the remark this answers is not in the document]
awelzel: IMO that's okay. There's other plugin names like `Zeek::MySQL`, `Zeek::File`, ` Zeek::Root` where it might not be clear that they provide. One can introspect with `zeek -NN` to get the components a plugin actually provides.

I could also see event serialization being built with Zeek's serialization format if we really wanted to (or log serialization with broker's format).

src/cluster/serializer/binary-serialization-format/Serializer.cc
```diff
@@ -0,0 +1,191 @@
...
+#include "zeek/SerializationFormat.h"
+#include "zeek/Type.h"
+#include "zeek/Val.h"
+#include "zeek/cluster/Serializer.h"
+#include "zeek/logging/Types.h"
+#include "zeek/threading/SerialTypes.h"
+
+using namespace zeek::cluster;
+
+// #define SERIALIZER_DEBUG(...) fprintf(stderr, __VA_ARGS__)
```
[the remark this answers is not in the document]
awelzel: I've switched this over to use the following so that `zeek -B plugin-Zeek-Binary_Serializer` enables the messages instead.
```
#define SERIALIZER_DEBUG(...) PLUGIN_DBG_LOG(zeek::plugin::Zeek_Binary_Serializer::plugin, __VA_ARGS__)
```

Thanks for poking.

src/cluster/serializer/binary-serialization-format/Serializer.cc
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+    }
+
+    char* data;
+    uint32_t len;
+    len = fmt.EndWrite(&data);
+
+    // Copy result into buffer. Would be nice to serialize directly into the buffer,
+    // but the SerializationFormat doesn't really allow for that.
```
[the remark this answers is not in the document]
awelzel: Broker also copies, but into a `std::string`: - within cluster `vector<std::byte>` is used as a container rather than `std::string`. If we we keep with SerializationFormat rather than using something else altogether might be worth changing this to serialize into an appropriate container.

https://github.com/zeek/zeek/blob/2c52e533b222357b5180acbb7407431dce2939ee/src/broker/Manager.cc#L693-L695

src/logging/Manager.h
awelzel: I like `WriteBatchFromRemote()` and added a bit of text to contrast it with `WriteFromRemote()`.

> Btw did the formatter put these empty lines here or did you do this? :-) It's a bit unsual.

Thanks! Fixed.

PR #4040, opened 2024-11-15
Define ordering on `Rule` instances and use on sets in `RuleMatcher`

ckreibich:
Establishing reliable ordering fixes a test failure we're seeing on Alpine for the `signatures/tcp-end-of-match` btest, since discrepancies in rule match traversal could lead to discrepancies in corresponding event ordering.

Since the Rule code is fairly pointer-heavy, I decided to stick with that, rather than changing the types of the set members. Maybe there's a nicer C++ way to do this, feel free to chime in.

I also wondered whether we could just use vectors in these two settings. Seems a little safer to stick with sets to avoid duplicates.


src/RuleMatcher.cc
```diff
@@ -842,7 +843,12 @@ void RuleMatcher::Match(RuleEndpointState* state, Rule::PatternType type, const
     // matched patterns per connection (which is a plausible assumption).
 
     // Find rules for which patterns have matched.
-    set<pair<Rule*, MatchPos>> rule_matches;
+    auto cmp = [](pair<Rule*, MatchPos> a, pair<Rule*, MatchPos> b) {
+        if ( *a.first == *b.first )
+            return a.second < b.second;
+        return *a.first < *b.first;
+    };
+    set<pair<Rule*, MatchPos>, decltype(cmp)> rule_matches(cmp);
```
awelzel, line 851: Thanks for looking and fixing. Rule's apparently have a unique `Index()` which should be consistent across platforms (haven't tried).

That would allow to do the following instead:
```
     // Find rules for which patterns have matched.
-    set<pair<Rule*, MatchPos>> rule_matches;
+    map<decltype(Rule::idx), std::pair<Rule*, MatchPos>> rule_matches;
```

I doubt this matters performance wise, but this would sort by Index() which is only a 4 byte unsigned int.

ckreibich: I saw those and was suspicious about the notion of an index for ordering, but it might be fine. I can take a look.

ckreibich: I took a look and the rule indexes are determined the parser, numbering them up as it encounters them. That seems safe, so I switched it over. CI agreed in my fork, fingers crossed. Thanks Arne.

src/RuleMatcher.cc
```diff
@@ -850,13 +850,12 @@ void RuleMatcher::Match(RuleEndpointState* state, Rule::PatternType type, const
         Rule* r = Rule::rule_table[aidx - 1];
 
         if ( AllRulePatternsMatched(r, mpos, accepted_matches) )
-            rule_matches.insert(make_pair(r, mpos));
+            rule_matches[r->Index()] = make_pair(r, mpos);
```
awelzel, line 853: This is subtly different in that `insert()` didn't override existing entries, if it doesn't matter, that's probably okay.

awelzel (approved):
One nit, but looks good.

src/RuleMatcher.cc
ckreibich: I tried that but it complains about storing a reference to a pointer, so I settled on that one.

ckreibich:
Argh, I just pushed from the wrong machine, apologies. Fixing shortly.

src/RuleMatcher.cc
bbannier: Did you try this?

```cpp
            rule_matches.insert({r->Index(), {r, mpos}});
```

The equivalent change on l.668 would have been

```cpp
            rule_matches.insert({r->Index(), r});
```

PR #4042, opened 2024-11-15
Minor include cleanups

awelzel:
Mostly motivated the input.h one in logging being curious, but also superfluous broker/Manager.h includes. And moving the doctest.h header at the bottom seemed also nice.

Comments welcome :-)


timwoj (approved):
:shipit:

PR #4043, opened 2024-11-15
ZAM community_id_v1() fix, ZAM baseline updates

awelzel:
Follow up for `$proto` addition to `conn_id`.


awelzel:
@vpax - any insight about ZAM vs non-ZAM making a difference here: https://github.com/zeek/zeek/pull/4043/commits/a75ade222bed3951c98a91cd329786e093a494b1 ? I'd have guessed non-ZAM would/could also defer record field field initialization, but maybe that's not actually the case.

vpax:
> any insight about ZAM vs non-ZAM making a difference here: [a75ade2]

That is pretty peculiar. What was the previous behavior that was different ZAM/non-ZAM?

awelzel:
> That is pretty peculiar. What was the previous behavior that was different ZAM/non-ZAM?

Hmm, there's no previous behavior. For non-ZAM, calling community_id_v1() as follows, `GetFieldAs<CountVal>(5)` - this is the new `$proto` field, defaulting `&default=65535`, return 65535. For ZAM, it returns `0` (supposedly unset and not checked).
```
$ zeek -e 'print community_id_v1([$orig_h=1.1.1.1, $orig_p=0/unknown, $resp_h=8.8.8.8, $resp_p=0/unknown])'
```

I just tested with the explicit `conn_id` constructor and non-ZAM and ZAM lead to bogus/random values using `GetFieldAs()`.
```
$ zeek -OZAM -e 'print community_id_v1(conn_id($orig_h=1.1.1.1, $orig_p=0/nknown, $resp_h=8.8.8.8, $resp_p=0/unknown))'
1:DnXyl2YS1Z4Ko2Is3P9q9d+lX7M=
$ zeek -OZAM -e 'print community_id_v1(conn_id($orig_h=1.1.1.1, $orig_p=0/unknown, $resp_h=8.8.8.8, $resp_p=0/unknown))'
1:Fk3zASMH3gLT+R9ApI3J74tqNi4=
$ zeek  -e 'print community_id_v1(conn_id($orig_h=1.1.1.1, $orig_p=0/unknown, $resp_h=8.8.8.8, $resp_p=0/unknown))'
1:nhOCe3BTiOcaPFPaBWIoV8EiNHg=
$ zeek  -e 'print community_id_v1(conn_id($orig_h=1.1.1.1, $orig_p=0/unknown, $resp_h=8.8.8.8, $resp_p=0/unknown))'
```

Calling `GetFieldAs<>` for something that has a default or is unset - seems `GetField()` is the safer approach here.

vpax:
Huh so the question is more like Why did the original work with the Interpreter, rather than Why did it not work with ZAM. Peculiar! But I'll assume no ZAM maintenance action needed here.

awelzel:
> Huh so the question is more like Why did the original work with the Interpreter, rather than Why did it not work with ZAM. Peculiar! But I'll assume no ZAM maintenance action needed here.

Nothing needed. Some output contained `coerce`. I haven't looked, but possibly ZAM coerces `[$orig_h=...]` differently to `conn_id` than the interpreter does, and that's visible in BiFs.

Otherwise this PR is just baseline updates in `Baseline.zam` if someone would want to leave a :+1:

timwoj (approved):

PR #4044, opened 2024-11-15
Add btests for ip_proto policy files

timwoj:
This adds two new btests to test the policy files for the ip_proto fields added in https://github.com/zeek/zeek/pull/3950. Instead of comparing the conn.log files as a whole, it uses `jq` to just look at the fields we care about.


testing/btest/scripts/policy/protocols/conn/disable-unknown-ip-proto-support.zeek
```diff
@@ -0,0 +1,8 @@
+# @TEST-DOC: Test the policy for removing the unknown IP protocol field and connections from conn.log
+# @TEST-EXEC: zeek -C -r $TRACES/cisco-fabric-path.pcap %INPUT
```
ckreibich, line 2: Just a detail but it looks like you don't actually need `-C` for that trace, here and below.

ckreibich (approved):

testing/btest/scripts/policy/protocols/conn/disable-unknown-ip-proto-support.zeek
timwoj: Ah, I had pulled that invocation from the `core.cisco-fabric-path` test so I'll check/fix that one too.

PR #4046, opened 2024-11-16
Don't forward out of the unknown IP transport analyzer

timwoj:
We already accept that the packets coming into this analyzer won't be processed, so forwarding out of it will just result in a failure. Forwarding will also report a weird for every packet which just results in extra noise.

Thanks @ckreibich for pointing out the overload of weirds.


ckreibich (approved):
Thanks Tim!

PR #4047, opened 2024-11-18
RFC: plugin: Add InitPreExecution()

awelzel:
Currently, plugins do not have a way to further inspect or even mutate script functions after ZAM optimization ran. One use-case here is zeek-perf-support [1]. This plugin wraps Stmt instances of functions, events and hooks hooks with a small assembly stub to support JIT map files [2] and for integration with perf tools.

This change introduces a new InitPreExecution() hook that runs after ZAM optimization completed, just before the zeek_init() event is enqueued.

Additionally, remove the existing CPP_activation_hook. It doesn't seem to be used. If it becomes necessary in the future, the new InitPreExecution() hook can be leveraged instead.

[1] https://github.com/zeek/zeek-perf-support
[2] https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/jit-interface.txt

### DIFF
--- a/src/script_opt/ScriptOpt.cc
+++ b/src/script_opt/ScriptOpt.cc
@@ -26,7 +26,6 @@ AnalyOpt analysis_options;
 std::unordered_set<const Func*> non_recursive_funcs;
 
 void (*CPP_init_hook)() = nullptr;
-void (*CPP_activation_hook)() = nullptr;
 
 // Tracks all of the loaded functions (including event handlers and hooks).
 static std::vector<FuncInfo> funcs;
--- a/src/plugin/Manager.cc
+++ b/src/plugin/Manager.cc
@@ -484,6 +484,14 @@ void Manager::InitPostScript() {
         (*i)->InitPostScript();
 }
 
+void Manager::InitPreExecution() {
+    assert(init);
+
+    for ( plugin_list::iterator i = Manager::ActivePluginsInternal()->begin();
+          i != Manager::ActivePluginsInternal()->end(); i++ )
+        (*i)->InitPreExecution();
+}
+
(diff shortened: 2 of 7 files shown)


awelzel:
@rsmmr - requested a reviewmostly to get an opinion whether that seems like an interesting extension?

src/script_opt/ScriptOpt.cc
```diff
@@ -26,7 +26,6 @@ AnalyOpt analysis_options;
 std::unordered_set<const Func*> non_recursive_funcs;
 
 void (*CPP_init_hook)() = nullptr;
-void (*CPP_activation_hook)() = nullptr;
```
rsmmr, line 29: So this already wasn't used anymore?

rsmmr (commented):
Seems fine to me. It's getting a bit confusing what exactly happens when but if this is a hook needed to support new functionality, I don't see anything wrong with adding it.

rsmmr (approved):

src/script_opt/ScriptOpt.cc
awelzel: Yeah, seems this in-tree user isn't needed anymore. I've pinged Vern about it. If the `CPP_activation_hook` is required down the road, can use the plugin hook instead.

awelzel:
> Seems fine to me. It's getting a bit confusing what exactly happens when but if this is a hook needed to support new functionality, I don't see anything wrong with adding it.

I've tested the this new hook with the `zeek-perf-support` plugin and using the `InitPreExecution()` script allows to support JIT files also for ZAM:

Screenshot from 2024-11-21 14-35-20

I'll move ahead and merge it. If there's better names coming up, can change before 8.0 :-)

PR #4048, opened 2024-11-18
sumstat/non-cluster: Move last epoch processing to zeek_done()

awelzel:
@Sheco reported that standalone epoch processing may exclude scheduled events when the final sumstat epoch runs before. For example, this easily happens when attempting to do sumstat observations within connection_state_remove().

Delay final epoch processing to zeek_done() instead.

This doesn't deal with the clustered version - this would need something more elaborate and potentially a mechanism to delay the shutdown of other cluster nodes until/after sumstat processing completed.


Sheco:
A quick glance tells me it looks good, I'll have to continue playing with this a bit more tomorrow. 

Thank you for the quick response and patch @awelzel .

0xxon (approved):
Thank you for this PR.

This seems fine to me. In general, this also points out that the way that sumstats handles the shutdown of Zeek is underdocumented - it would be neat to change that and point out in the documentation how Zeek will behave in clustered/non-clustered mode.

PR #4049, opened 2024-11-18
CI: Use test.sh script for running ZAM tests, but disable parts of it

timwoj:
This ensures that the normal test setup is run, which enables greedy mode on Cirrus allowing us to over-provision the testing. This should help speed up the ZAM builds a bit.

This cuts the time for the ZAM ASAN builds roughly in half.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -40,6 +40,11 @@ function banner {
 }
 
 function run_unit_tests {
+    if [[ ${ZEEK_CI_SKIP_UNIT_TESTS} -eq 1 ]]; then
+        printf "Skipping unit tests as requested by task configureation\n\n"
+        return 0
+    fi
+
     banner "Running unit tests"
 
     pushd build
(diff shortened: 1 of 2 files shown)


ci/test.sh
```diff
@@ -57,17 +63,27 @@ function prep_artifacts {
 function run_btests {
     banner "Running baseline tests: zeek"
 
+    if [[ -n "${ZEEK_CI_ZAM}" ]]; then
+        ALTERNATIVE_ARG="-a zam"
```
awelzel, line 67: Maybe `local alternative_arg` given it's just used in this function? And possibly initialize with `""`.

awelzel (approved):
Nice! I wasn't aware we use a larger `-j`.

ci/test.sh
```diff
@@ -57,17 +63,27 @@ function prep_artifacts {
...
+        ${BTEST} -z ${ZEEK_CI_BTEST_RETRIES} -d -A -x btest-results.xml -j ${ZEEK_CI_BTEST_JOBS} ${ALTERNATIVE_ARG} || result=1
     make coverage
     prep_artifacts
     popd
     return 0
 }
 
 function run_external_btests {
+    # Skip external btests when running against ZAM
+    if [[ -n "${ZEEK_CI_ZAM}" ]]; then
```
awelzel, line 82: After looking again, instead ZEEK_CI_ZAM, could also do something like the following and set appropriately in the tasks. Slight preference for that as it seems more generic.
```
ZEEK_CI_SKIP_EXTERNAL_BTESTS
ZEEK_CI_SKIP_UNIT_TESTS
ZEEK_CI_BTEST_EXTRA_ARGS
```

timwoj: Oh yeah, I like that too.

timwoj:
I'm planning to cherry-pick this into the release/7.0 branch once it's merged since we have the ZAM CI builds there now.

PR #4050, opened 2024-11-19
Bump zeekjs to 0.13.2

awelzel:
41ec748 cmake: Use foreach range
    2b90a56 tests: Update new-event to run with 7.1+, remove ip_proto elsewhere
    a596f6f Makefile: Update outdated commands
    b954661 Enable `cmake/FindNodejs.cmake` to find more libnode versions


timwoj (approved):
Approving, pending CI finishing though I doubt it would fail

PR #4056, opened 2024-11-21
Fix conversion of intervals through JSON

timwoj:
I'm lifting this out of the storage framework work since it doesn't need to be in there exclusively. This fixes the conversion of `interval` values from JSON to `IntervalVal` objects. It also adds a simple btest to test round-trip conversion of Vals through JSON.


src/Val.cc
```diff
@@ -938,10 +938,32 @@ static std::variant<ValPtr, std::string> BuildVal(const rapidjson::Value& j, con
...
         case TYPE_INTERVAL: {
-            if ( ! j.IsNumber() )
-                return mismatch_err();
+            if ( j.IsNumber() )
+                return make_intrusive<IntervalVal>(j.GetDouble());
+
+            if ( j.IsString() ) {
+                auto parts = util::split(j.GetString(), " ");
+                if ( parts.size() != 2 )
+                    return "wrong interval format, must be 'value {day|hr|min|sec|msec|usec}'";
```
awelzel, line 947: For the purpose of the storage framework, I think `to_json()` should render intervals as numbers/doubles instead. If this means adding a new parameter (e.g. intervals_as_double=T or some more generic/extensible options record/struct).

Contrast, e.g., with `time` which doesn't render into ISO format, but instead unix timestamp in seconds.

For interoperability with external system, the verbose string representation makes little sense - consumers would need to parse this explicitly. Using double/float is less annoying.

awelzel: The parsing also needs to be more more fancy :-(

```
$ zeek -e 'print double_to_interval(2*24*3600 + 2*3600 +  2*60 + 2*1.0 + 2*0.1 + 2*0.0001)'
2.0 days 2.0 hrs 2.0 mins 2.0 secs 200.0 msecs 200.000021 usecs
```

I think it makes sense to have this within Zeek still, but lets try to avoid putting the burden on others, too?

timwoj: > The parsing also needs to be more more fancy :-(

This is fixed/tested in the first two fixup commits

> For the purpose of the storage framework, I think to_json() should render intervals as numbers/doubles instead. If this means adding a new parameter (e.g. intervals_as_double=T or some more generic/extensible options record/struct).

The problem with this is that `to_json` just calls `Val::BuildJSON` underneath via `Val::ToJSON`. Adding a type-specific flag to all of those methods breaks the generic-ness of that whole call-chain. I could see making a new version of `to_json` that takes a record of options and passing a `RecordVal` through to the C++, but I worry about having to do a bunch of repeat record lookups on a potential logging hotpath. Unfortunately the only other easy way to do this - using a `global` flag - is a no-go as well.

I made the most recent changes to `BuildJSON` to use rapidjson clear back in 3.0, but the original decision to log the broken-out version of `interval` goes back to `json.bro` before then.

testing/btest/Baseline/scripts.base.utils.json-roundtrip/output
```diff
@@ -0,0 +1,25 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+Valid conversion of Foo: 1
+
+T
+T
+T
+T
+F
```
awelzel, line 8: This one looks a bit off, is it `	print f$n == f2_v$n;` ?

testing/btest/scripts/base/utils/json-roundtrip.test
```diff
@@ -0,0 +1,102 @@
+# @TEST-DOCS: Test round-trip JSON encoding and decoding using the Zeek methods
```
awelzel, line 1: Thanks for adding this! Could also see this one in `bifs/`, next to the existing `from_json.zeek` test. I thought `base/utils` is used for what's actually located in `base/utils`.

src/zeek.bif
```diff
@@ -5103,14 +5103,18 @@ function anonymize_addr%(a: addr, cl: IPAddrAnonymizationClass%): addr
 ##                       rendered name. The default pattern strips a leading
 ##                       underscore.
 ##
+## interval_as_double: If true, interval values will be logged as doubles
```
awelzel, line 5106:
```suggestion
## interval_as_double: If T, interval values will be logged as doubles
```

src/zeek.bif
```diff
@@ -5103,14 +5103,18 @@ function anonymize_addr%(a: addr, cl: IPAddrAnonymizationClass%): addr
...
+##                     instead of the broken-out version with units.
```
awelzel, line 5107: Would be good to mention type string. Maybe this?
```suggestion
##                     instead of the broken-out version with units as strings.
```

src/Val.cc
```diff
@@ -938,10 +949,39 @@ static std::variant<ValPtr, std::string> BuildVal(const rapidjson::Value& j, con
         }
 
         case TYPE_INTERVAL: {
-            if ( ! j.IsNumber() )
-                return mismatch_err();
+            if ( j.IsNumber() )
+                return make_intrusive<IntervalVal>(j.GetDouble());
+
+            if ( j.IsString() ) {
+                auto parts = util::split(j.GetString(), " ");
```
awelzel, line 956: I'm not sure we care all that much, but IIRC superfluous spaces would yield empty parts. Could filter out empty parts to support more than one spaces as separators. Mostly relevant if someone crafts input by hand.

awelzel:
> Adding a type-specific flag to all of those methods breaks the generic-ness of that whole call-chain. 

Ah, that's a good point. I see you did anyway - to some degree `only_loggable` and `re` apply only to records already.

I didn't realize the entry is `Val::ToJSON()` - I thought it's freestanding (`BuildJSON()` is) - I guess if it was a non-member function `utils::to_json(const ValPtr& v)` or so, this would be mood.

>  I could see making a new version of to_json that takes a record of options and passing a RecordVal through to the C++, but I worry about having to do a bunch of repeat record lookups on a potential logging hotpath. Maybe I'm just overthinking it though.

I was thinking converting the options RecordVal to a plain struct that's then passed down. Though that one extra parameter for now seems fine.

awelzel (approved):
Mostly nits, feel free to reject or address.

I like where this ended!

testing/btest/scripts/base/utils/json-roundtrip.test
timwoj: There's also `scripts/base/utils/json.test`

awelzel: Fair! Resolved it.

Issue #4052, opened 2024-11-20
Test `core.expr-exception` reliably fails under ZAM with ASAN


bbannier:
The test `core.expr-exception` fails reliably under ZAM with ASAN due to [detected memory leaks](https://github.com/user-attachments/files/17830204/test.log), see e.g., https://cirrus-ci.com/task/6065381939347456.

awelzel:
@vpax - mind referencing the issue number in a commit message (e.g `Fixes #4052`) once you get around poking at the `core.expr-exception` leak? Thanks!

PR #4059, opened 2024-11-22
Disable core.expr-execption btest under ZAM to fix CI builds

awelzel:
Relates to #4052.

For the time being I think we should disable the test for ZAM, so that `fullci` and nightly runs end up green.


testing/btest/Baseline/core.expr-exception/reporter.log
bbannier, line 1: This seems unrelated.

bbannier (approved):
The new `@TEST-REQUIRES` looks good to me, but we probably shouldn't include the baseline change which seems unrelated.

PR #4064, opened 2024-11-27
zeek-testing: Add Experiment-VM-Microsoft-Windows7AD-1 PCAP a…

awelzel:
…nd baselines

The pcap comes from the following dataset [1]:

    CTU-SME-11: a labeled dataset with real benign and malicious network
    traffic mimicking a small medium-size enterprise environment

[1] https://zenodo.org/records/7958259

---
Goes with https://github.com/zeek/zeek-testing/pull/18


0xxon:
As that is a public trace, I would just go put it into our public test suite.

The private test suite is typically just used for traces that we cannot publically share. As this one is already downloadable by everyone, that should not be an issue.

0xxon (approved):

PR #4070, opened 2024-11-29
Fix -O dump-final-ZAM to work in conjunction with -O no-ZAM-opt

vpax:
The original implementation of `-O dump-final-ZAM` mis-assumed that the regular full ZAM optimization had been done, which includes doubling up where local variables live in the `ZVal` frame used for execution. That failed when also specifying `-O no-ZAM-opt` to skip final ZAM optimization. This PR remedies the issue.

### DIFF
--- a/src/script_opt/ZAM/Driver.cc
+++ b/src/script_opt/ZAM/Driver.cc
@@ -404,16 +404,19 @@ void ZAMCompiler::Dump() {
         }
     }
     else if ( analysis_options.dump_final_ZAM ) {
-        ASSERT(remapped_frame);
-
         printf("\nFrame for %s:\n", func->GetName().c_str());
 
-        for ( auto i = 0U; i < shared_frame_denizens.size(); ++i ) {
-            printf("frame[%d] =", i);
-            for ( auto& id : shared_frame_denizens[i].ids )
-                printf(" %s", id->Name());
-            printf("\n");
+        if ( remapped_frame ) {
+            for ( auto i = 0U; i < shared_frame_denizens.size(); ++i ) {
+                printf("frame[%d] =", i);
+                for ( auto& id : shared_frame_denizens[i].ids )
+                    printf(" %s", id->Name());
+                printf("\n");
+            }
         }
+        else
+            for ( const auto& elem : frame_layout1 )
+                printf("frame[%d] = %s\n", elem.second, elem.first->Name());
     }
 
     if ( ! insts2.empty() )


vpax:
That dang fixup was meant for a different branch, reverting it here

src/script_opt/ZAM/Driver.cc
```diff
@@ -404,16 +404,19 @@ void ZAMCompiler::Dump() {
...
-        ASSERT(remapped_frame);
-
         printf("\nFrame for %s:\n", func->GetName().c_str());
 
-        for ( auto i = 0U; i < shared_frame_denizens.size(); ++i ) {
-            printf("frame[%d] =", i);
-            for ( auto& id : shared_frame_denizens[i].ids )
-                printf(" %s", id->Name());
-            printf("\n");
+        if ( remapped_frame ) {
```
awelzel, line 409: I wonder if adding `"remapped_frame=%d", remapped_frame` to above printf would add a bit more information to the output when remapping happened and when it didn't.

(...without actually knowing what that means :-) )

awelzel (commented):
Would merge, but wonder if you'd want to add a test into `btest/opt` that that uses some variations of `-O dump-final-ZAM`. Not necessarily to check the output (unless it's stable enough), but maybe to at least assert no crash/assertions are triggered.

vpax:
> Would merge, but wonder if you'd want to add a test into `btest/opt` that that uses some variations of `-O dump-final-ZAM`. Not necessarily to check the output (unless it's stable enough), but maybe to at least assert no crash/assertions are triggered.

I thought about this but it seemed more bloat than benefit, given all the test can do is fail to crash. The reason is that they're both options only meant for maintainers to use, and you have to use them together. If you're a maintainer and there's a regression, you'll just fix it at that point.

src/script_opt/ZAM/Driver.cc
vpax: The frame is always remapped unless low-level ZAM optimization is turned off (which is what `-O no-ZAM-opt` does). As this is output for the maintainer, it should be immediately apparent to them whether the frame has been remapped.

BTW, remapping refers to where local variables reside in the frame. Initially each is mapped to its own slot (like the interpreter does). ZAM's low-level optimization identifies which of these don't overlap and thus can share a slot. This often makes the frame much smaller due to the heavy use in script optimization of temporary variables, which turns out to have appreciable performance gains.

PR #4071, opened 2024-11-29
Add location information to ZAM run-time errors about vector length mismatches

vpax:
As noted in https://github.com/zeek/zeek/pull/4022, that commit added a fix for a ZAM ASAN issue that introduced a not-great error message. At the time I thought that making the message coherent would take non-trivial work, but it turns out it was in fact trivial. This PR provides that fix; I did it as a `--fixup` but if that's not something done across merges, let me know and I'll do it standalone.

### DIFF
--- a/src/script_opt/ZAM/ZBody.cc
+++ b/src/script_opt/ZAM/ZBody.cc
@@ -622,7 +622,8 @@ static void vec_exec(ZOp op, TypePtr t, VectorVal*& v1, const VectorVal* v2, con
     auto n = vec2.size();
 
     if ( vec3.size() != n ) {
-        ZAM_run_time_error(util::fmt("vector operands are of different sizes (%d vs. %d)", int(n), int(vec3.size())));
+        ZAM_run_time_error(z.loc,
+                           util::fmt("vector operands are of different sizes (%d vs. %d)", int(n), int(vec3.size())));
         return;
     }
 


src/script_opt/ZAM/ZBody.cc
```diff
@@ -622,7 +622,8 @@ static void vec_exec(ZOp op, TypePtr t, VectorVal*& v1, const VectorVal* v2, con
     auto n = vec2.size();
 
     if ( vec3.size() != n ) {
-        ZAM_run_time_error(util::fmt("vector operands are of different sizes (%d vs. %d)", int(n), int(vec3.size())));
+        ZAM_run_time_error(z.loc,
+                           util::fmt("vector operands are of different sizes (%d vs. %d)", int(n), int(vec3.size())));
```
awelzel, line 626: Nit: Mind switching `%d` to `%zu` (for size_t) and removing the C-style casts?

awelzel (changes requested):
Fixup commits are meant to be squashed into prior commits. That's okay for commits not already merged into `master`. Fixing up commits already merged into master would imply rewriting history and force-pushing `master` which in most every case is unreasonable.

So yes, changing the commit title to something standalone/new is the right call.

vpax:
Closing this PR as now superseded by https://github.com/zeek/zeek/pull/4072.

PR #4073, opened 2024-11-30
Skip optimization of functions with AST nodes unknown to script optimization

vpax:
Per a comment of @awelzel's, this PR provides a structure for recognizing functions whose bodies have AST nodes of types that script optimization doesn't know how to compile. I didn't include a BTest because by their nature it's hard to construct such function bodies, but I did test that the functionality works if I deliberately omit one of the AST types from those listed. I also extended the logic in the AST inliner to avoid inlining any such function.

Note, the listing of AST node tags is done with clang-formatting turned off so that it's easy to visually compare the list with those in `Expr.h` and `StmtEnums.h`.

### DIFF
--- a/src/script_opt/ProfileFunc.cc
+++ b/src/script_opt/ProfileFunc.cc
@@ -101,7 +101,7 @@ ProfileFunc::ProfileFunc(const Expr* e, bool _abs_rec_fields) {
 }
 
 TraversalCode ProfileFunc::PreStmt(const Stmt* s) {
-    stmts.push_back(s);
+    stmts.push_back({NewRef{}, const_cast<Stmt*>(s)});
 
     switch ( s->Tag() ) {
         case STMT_INIT:
@@ -185,7 +185,7 @@ TraversalCode ProfileFunc::PreStmt(const Stmt* s) {
 }
 
 TraversalCode ProfileFunc::PreExpr(const Expr* e) {
-    exprs.push_back(e);
+    exprs.push_back({NewRef{}, const_cast<Expr*>(e)});
 
     TrackType(e->GetType());
 
@@ -867,11 +867,11 @@ void ProfileFuncs::ComputeProfileHash(std::shared_ptr<ProfileFunc> pf) {
         h = merge_p_hashes(h, p_hash(ov[i]->Name()));
 
     h = merge_p_hashes(h, p_hash("stmts"));
-    for ( auto i : pf->Stmts() )
+    for ( auto& i : pf->Stmts() )
         h = merge_p_hashes(h, p_hash(i->Tag()));
 
     h = merge_p_hashes(h, p_hash("exprs"));
-    for ( auto i : pf->Exprs() )
+    for ( auto& i : pf->Exprs() )
         h = merge_p_hashes(h, p_hash(i->Tag()));
 
     h = merge_p_hashes(h, p_hash("ids"));
--- a/src/script_opt/ScriptOpt.cc
+++ b/src/script_opt/ScriptOpt.cc
@@ -657,4 +657,123 @@ void profile_script_execution() {
 
 void finish_script_execution() { profile_script_execution(); }
 
+// For now, we have equivalent concerns between ZAM and compile-to-C++.
+bool has_AST_node_unknown_to_script_opt(const ProfileFunc* prof, bool /* is_ZAM */) {
+    // Note that the following sets are not comprehensive across the
+    // standard tags, because some tags are only generated *by* script
+    // optimization
+    // clang-format off
+    static const std::set<StmtTag> known_stmts = {
+        STMT_PRINT,
+        STMT_EVENT,
+        STMT_EXPR,
+        STMT_IF,
+        STMT_WHEN,
+        STMT_SWITCH,
+        STMT_FOR,
+        STMT_NEXT,
+        STMT_BREAK,
+        STMT_RETURN,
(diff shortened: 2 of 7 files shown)


src/script_opt/ScriptOpt.cc
```diff
@@ -657,4 +657,123 @@ void profile_script_execution() {
...
+        // STMT_CATCH_RETURN,
+        // STMT_CHECK_ANY_LEN,
+        // STMT_CPP,
+        // STMT_ZAM,
+        STMT_NULL,
+        STMT_ASSERT,
+        // STMT_EXTERN,
+        // STMT_STD_FUNCTION,
+    };
+    // clang-format on
```
awelzel, line 691: I'm not so sure about the maintainability here.

Isn't it just `STMT_EXTERN` and `STMT_STD_FUNCTION` (and maybe STMT_CPP?) that need to be recognized from the original AST?

If you want to keep the full list, maybe some static assert / constexpr comparing NUM_STMTS and NUM_EXPRS with the number of statements listed here could detect inconsistencies at compilation?

src/script_opt/ScriptOpt.cc
```diff
@@ -657,4 +657,123 @@ void profile_script_execution() {
...
+        // EXPR_INDEX_ASSIGN,
+        // EXPR_FIELD_LHS_ASSIGN,
+        // EXPR_REC_ASSIGN_FIELDS,
+        // EXPR_REC_ADD_FIELDS,
+        // EXPR_REC_CONSTRUCT_WITH_REC,
+        // EXPR_FROM_ANY_VEC_COERCE,
+        // EXPR_ANY_INDEX,
+        // EXPR_SCRIPT_OPT_BUILTIN,
+        // EXPR_NOP,
+    };
```
awelzel, line 769: Do we need it for expressions, too? I'd think if ZAM optimization runs into a expression it doesn't know, could either hard-abort with an error message, or fallback to interpreted execution?

src/script_opt/ProfileFunc.cc
```diff
@@ -867,11 +867,11 @@ void ProfileFuncs::ComputeProfileHash(std::shared_ptr<ProfileFunc> pf) {
         h = merge_p_hashes(h, p_hash(ov[i]->Name()));
 
     h = merge_p_hashes(h, p_hash("stmts"));
-    for ( auto i : pf->Stmts() )
+    for ( auto& i : pf->Stmts() )
```
awelzel, line 870:
```suggestion
    for ( const auto& i : pf->Stmts() )
```

src/script_opt/ZAM/Support.cc
```diff
@@ -117,6 +117,12 @@ bool file_mgr_set_reassembly_buffer(StringVal* file_id, uint64_t max) {
 bool ZAM_error = false;
 
 bool is_ZAM_compilable(const ProfileFunc* pf, const char** reason) {
+    if ( has_AST_node_unknown_to_script_opt(pf, true) ) {
+        if ( reason )
+            *reason = "unknown AST node type";
```
awelzel, line 122: Might make sense to get the offending node type out of the helper (or change to std::pair<bool, reason: string>`) so that the message provides a hint which stmt/expr it was which seems important.

src/script_opt/ScriptOpt.cc
vpax: > Isn't it just `STMT_EXTERN` and `STMT_STD_FUNCTION` (and maybe STMT_CPP?) that need to be recognized from the original AST?

In principle it's more than that, per the others that are commented out. One could however rely on the fact that things like `STMT_CATCH_RETURN` shouldn't show up in a candidate for script optimization since it's only generated during script optimization, and then go with the simple list of what's-not-compilable. It seems though that that path is going to fail the next time one of those is added and forgets to update them here.

We could add comments to `StmtEnums.h` that when adding a node there, it should be added to `ScriptOpt.cc` too. Do you prefer that?

src/script_opt/ScriptOpt.cc
vpax: > Do we need it for expressions, too? I'd think if ZAM optimization runs into a expression it doesn't know, could either hard-abort with an error message, or fallback to interpreted execution?

It will likely fail with an internal inconsistency error. The same is true for statements, which is why I went down this path in the first place - I'm not seeing a substantive difference between statements and expressions here.

src/script_opt/ZAM/Support.cc
vpax: I see the benefit. Here I wasn't looking to open up that API, just to add an element to its functionality.

src/script_opt/ProfileFunc.cc
vpax: and re the first one: oops, thanks for catching!

src/script_opt/ZAM/Support.cc
awelzel: Oh, I just tried and now see const `char** reason` as output parameter which would require memory management for a formatted string. Resolving.

src/script_opt/ScriptOpt.cc
awelzel: I've added a commit using `static_assert()` so that compilation fails if new Stmt or Expr enums are added. I'll merge this PR if CI looks sensible.

awelzel (approved):

PR #4075, opened 2024-12-01
Remove obsolete Broker compatibility layer

Neverlord:
Since the transition to broker::variant has been long finalized, there is no more need to be able to go back to a pre-variant version of Broker. Hence, we can drop various utilities that allow Zeek to run with older Broker releases.

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -1,8 +1,9 @@
 #include "zeek/broker/Manager.h"
 
-#include <broker/broker.hh>
 #include <broker/config.hh>
 #include <broker/configuration.hh>
+#include <broker/endpoint.hh>
+#include <broker/variant.hh>
 #include <broker/zeek.hh>
 #include <unistd.h>
 #include <cstdio>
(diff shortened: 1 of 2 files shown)


src/broker/Manager.cc
```diff
@@ -1,8 +1,9 @@
 #include "zeek/broker/Manager.h"
 
-#include <broker/broker.hh>
 #include <broker/config.hh>
 #include <broker/configuration.hh>
+#include <broker/endpoint.hh>
+#include <broker/variant.hh>
```
awelzel, line 6: Unrelated and not for this PR: I wonder if adding a `clang-format` rule to visually separate `<broker/...>` includes from `std` includes might be a nice cleanup.

src/broker/Manager.cc
```diff
@@ -31,47 +32,15 @@
...
     return std::move(broker::get<broker::vector>(tmp));
 }
 
-#endif
-
-// Converts a string_view into a string to make sure that we can safely call `.c_str()` on the result.
-template<class View>
-std::enable_if_t<std::is_same_v<std::decay_t<View>, std::string_view>, std::string> c_str_safe(View&& arg) {
-    return std::string{arg};
-}
```
awelzel, line 67: Is the `c_str_safe()` cleanup related to `broker::data` removal? Seems a separate commit  would make review a bit easier and give a better idea of the individual changes in the history.

Up for splitting up the existing single commit?

src/broker/Manager.cc
timwoj: Agreed. Our default setup for zeek puts the standard headers first, followed by anything else. It should probably be the same for broker, but with broker headers instead of zeek headers.

src/broker/Manager.cc
Neverlord: > Is the c_str_safe() cleanup related to broker::data removal?

Yes, a `broker::data` contains actual `std::string` objects while a `broker::variant` only grants access to characters via `std::string_view`. Hence, we need to copy things into temporary `std::string` objects when using the Zeek formatting functions.

src/broker/Manager.cc
Neverlord: I'll do that as a followup.

src/broker/Manager.cc
Neverlord: > Up for splitting up the existing single commit?

Done.

awelzel (approved):

PR #4079, opened 2024-12-03
ZAM support for (optionally) keeping "assert" statements

vpax:
ZAM has historically optimized `assert` statements away (analogous to C++ optimization). That makes sense for usage of `assert`s to sanity-check run-time logic. However it doesn't fit with the use of `assert`s for testing, which is a style we've started to use since Zeek added script-level support for `assert`. This PR addresses that discrepancy by adding an option to tell ZAM to keep `assert` statements; Zeek's BTest config turns on the corresponding environment variable so that tests written using `assert`s now come within the purview of ZAM CI.

While the change seems conceptually simple, the implementation grew in complexity due to a number of issues:
- `assert` statements include an optional "message" expression, and the original implementation supported the evaluation of the expression potentially throwing an exception. To support that in ZAM would have required transforming those expressions into lambdas, which seems like a lot of implementation just to tidy up errors that the script writer wrote in a particular context. In discussion, @awelzel (the original implementor) agreed that it makes sense to remove dealing with those exceptions. (I checked the docs and didn't find any mention of the functionality, so I don't believe this is a breaking change.)
- For script optimization, it's helpful for `Stmt` nodes to have a degree of regularity. To that end, `AssertStmt` is now derived from `ExprStmt` (for its main test expression) rather than being fully stand-alone.
- Running ZAM on some of the `assert` BTests highlighted an issue with ZAM's event handler coalescence (merging multiple event handler bodies via inlining, to allow cross-body optimization). Such coalescence means that if one of the event handler bodies causes a run-time exception, the later bodies won't execute. To enable full compatibility in this regard, ZAM now has a `-O no-event-handler-coalescence` option to prevent the inlining.


awelzel:
> To enable full compatibility in this regard, ZAM now has a -O no-event-handler-coalescence option to prevent the inlining.

Should this be the default though? For interpreted execution, my mental model is that each handler runs within its own `try { ... } catch (...) {}` block. A runtime error in handler A has no effect on B (unless B has some actual dependency on fields or data set by A).

src/script_opt/Inline.cc
```diff
@@ -166,7 +166,8 @@ void Inliner::Analyze() {
         inline_ables[func] = f.Profile();
     }
 
-    CoalesceEventHandlers();
+    if ( ! analysis_options.no_eh_coalescence )
+        CoalesceEventHandlers();
```
awelzel, line 170: Could've squashed that in the commit that added the option, too.

src/Stmt.h
```diff
@@ -516,9 +516,12 @@ class AssertStmt final : public Stmt {
     StmtPtr DoReduce(Reducer* c) override;
 
 private:
-    ExprPtr cond;
     std::string cond_desc;
     ExprPtr msg;
+
+    // Statement to execute before evaluating "msg". Only used for script
+    // optimization.
+    StmtPtr msg_setup_stmt;
```
awelzel, line 524: I'm not following if this is left-over from attempted exception handling, or required because it'll be set during `DoReduce()` via:
```
        msg = msg->ReduceToSingleton(c, msg_setup_stmt);
```

I'm just making things up, but could `AssertStmt::DoReduce()` return a `StmtList` prepended with the setup stmt?

This isn't a biggie, it's just that it's one more optimization specific field/accessor and I find it not easy to follow why it exists - if you say no dice or complicated, no worries.

awelzel (approved):
I'd merge, but left a few comments/thoughts.

src/script_opt/Inline.cc
vpax: > Could've squashed that in the commit that added the option, too.

Oops yeah not sure how that wound up being separate.

src/Stmt.h
vpax: Yeah, it needs to be separate. The reason is that it should only be evaluated if the assertion is going to be reported, and the statement might not be valid if it's not going to be reported. For example,
```
assert ! r?$b, fmt("r$b unexpectedly set: %s", r$b);
```

vpax:
> EDIT: Saw the incompatibilities change now - I guess 👍 if this doesn't worry you.

Yeah I mulled this over, and then recalled that there's already a notion of incompatibilities. Ones that arise only in error contexts don't give me too much heartburn.

src/Stmt.h
vpax: Hmmm I may see a bug in how that's implemented. However I'd rather tackle that post-merge, if you're okay with that.

PR #4080, opened 2024-12-03
ZAM optimization now removes hook calls to hooks without any bodies

vpax:
This PR adds to ZAM script optimization the removal of calls to hooks that don't have any bodies to execute. This has the nice benefit of making it cost-free for developers using script optimization to add in hooks to their scripts even if those hooks might often not be used.

### DIFF
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -1458,6 +1458,7 @@ protected:
     bool IsFoldableBiF() const;
     bool AllConstArgs() const;
     bool CheckForBuiltin() const;
+    bool IsEmptyHook() const;
     ExprPtr TransformToBuiltin();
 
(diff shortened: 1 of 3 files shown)


awelzel (approved):
I wonder if this could just be applied to the original AST, too.

This does put a stop at registering hooks at runtime when running under ZAM, but I guess that's fine. It's technically possible to do so today with ZeekJS or from other plugins.

vpax:
> I wonder if this could just be applied to the original AST, too.

It's trickier because the interpreter doesn't have a framework for updating ASTs.

> This does put a stop at registering hooks at runtime when running under ZAM

Yeah ... I can imagine down the line having a setting that says to omit this optimization for that reason.

PR #4085, opened 2024-12-04
ZAM support for pattern equality/inequality operations

vpax:
This PR adds ZAM support for the recently introduced relationals for comparing one pattern to another. Doing so requires a (minor) enhancement to `gen-zam`, which is tracked [here](https://github.com/zeek/gen-zam/pull/15). That change pointed up a couple of places where some type-checking/coercion was a bit loose (and now broken), so those are fixed here too.


timwoj (approved):
Thanks for turning this one around so fast!

Issue #4068, opened 2024-11-27
Generic parsing for HTTP UPGRADE connections not containing sub protocols


blightzero:
The http_connection_upgrade() event and the websocket protocol parser allow the analysis of traffic in a websocket connection. However, some APIs do not use any particular protocol after an HTTP 101 UPGRADE response. For example the docker HTTP API makes a lot of use of HTTP upgrade messages to facilitate the interaction of the docker client and the docker server. The connection is upgraded and then the stdin and stdout/stderr of the respective program are connected via the established HTTP connection. This happens without any particular sub protocol in place that transports the data. There is currently no obvious way in zeek to further analyze or log events inside this stream.

One application of this would be docker honeypots where it would be great to analyze the interaction of actors with the docker container. For that it would be awesome if it was possible to extract and match the traffic using zeek.

The Upgrade response has Content-Type `application/vnd.docker.raw-stream` and the Upgrade protocol `tcp`:
![Screenshot from 2024-11-27 09-56-52](https://github.com/user-attachments/assets/94334876-bb09-40ef-9f83-6bb0fe5a4c39)

A minimal pcap with the respective HTTP UPGRADE and following interaction via the docker API can found in the attached pcap zip:
[docker_http_uprade.zip](https://github.com/user-attachments/files/17939740/docker_http_uprade.zip)

For further reference see slack discussion with @awelzel : https://zeekorg.slack.com/archives/CTGDG83B8/p1732622761199559

awelzel:
@blightzero - I'm closing this after the inclusion of #4088 and your positive feedback on Slack. Please re-open if you end up having a more complex use-case.

PR #4088, opened 2024-12-05
Add StreamEvent analyzer and demo usage in HTTP::upgrade_analyzers

awelzel:
In #4068 , @blightzero reported not knowing which analyzer to attach to an HTTP connection upgraded to "tcp" to do further analysis of the data. Their use-case is a HTTP/TCP connection representing stdin/stdout of containers.

For the provided pcap, an appropriate analyzer can be attached via `HTTP::upgrade_analyzers["tcp"]`, but AFAICT, there's no analyzer within Zeek that would allow raw stream data access.

This PR adds a RawStream analyzer that can be leveraged to capture stream data in script land. In contrast to `tcp_contents`, the analyzer can be dynamically added to existing or upgraded connections. I *think* it can also be enabled via DPD if one really wanted to, something that `tcp_contents` doesn't do either.

---

There's a second part to #4068 in that a `Upgrade: tcp` isn't very specific. The pcap provided in #4068 has a useful Content-Type in the response that could be used for pivoting/dispatching. The analyzer would still be needed though.
```
HTTP::upgrade_analyzers_content_type["tcp", "application/vnd.docker.raw-stream"] = Analyzer::ANALYZER_RAWSTREAM.
```


awelzel:
@rsmmr - requesting a review from you in case I missed something. Also feel like you might have an opinion here :-)

src/analyzer/protocol/rawstream/Plugin.cc
```diff
@@ -0,0 +1,23 @@
+// See the file  in the main distribution directory for copyright.
```
rsmmr, line 1: `"COPYING"` got lost in that sentence.

src/analyzer/protocol/rawstream/README
```diff
@@ -0,0 +1,4 @@
+# TCP application analyzer for handing raw stream data to script-land.
+#
+# This is similar to tcp_contents(), but can be triggered via DPD,
+# well-known-port registration, or HTTP and WebSocket upgrade mechansisms
```
rsmmr, line 4: I'd start this more directly in terms of what the analyzer does, not with a comparison to `tcp_contents()`. Then maybe afterwards do the compare to `tcp_contents()`.

src/analyzer/protocol/rawstream/events.bif
```diff
@@ -0,0 +1,27 @@
+## Generated for every DeliverStream() invocation on an instance of the Raw analyzer.
```
rsmmr, line 1: Better not to phrase this in terms of internal methods, but instead describe directly what it does.

src/analyzer/protocol/rawstream/events.bif
```diff
@@ -0,0 +1,27 @@
...
+##
+## This is a low-level event to inspect all data from the originator
+## and responder for debugging or logging of plain-text interactive
+## sessions.
```
rsmmr, line 5: I'd add a warning that it's potentially expensive if substantial amounts of data end up going through scriptland.

src/analyzer/protocol/rawstream/events.bif
```diff
@@ -0,0 +1,27 @@
...
+## c: The connection.
+##
+## is_orig: T if this is originator-side input.
+##
+## data: The raw data
+##
+## .. zeek:see:: rawstream_undelivered
+event rawstream_deliver%(c: connection, is_orig: bool, data: string%);
+
+## Generated for every Undelivered() invocation on an instance of the Raw analyzer.
```
rsmmr, line 16: Like above.

src/analyzer/protocol/rawstream/RawStream.cc
```diff
@@ -0,0 +1,25 @@
...
+namespace zeek::analyzer::rawstream {
+
+RawStream_Analyzer::RawStream_Analyzer(Connection* conn) : analyzer::tcp::TCP_ApplicationAnalyzer("RawStream", conn) {}
+
+
+void RawStream_Analyzer::DeliverStream(int len, const u_char* data, bool orig) {
+    analyzer::tcp::TCP_ApplicationAnalyzer::DeliverStream(len, data, orig);
+
+    auto s = len > 0 ? zeek::make_intrusive<StringVal>(len, reinterpret_cast<const char*>(data)) :
+                       zeek::val_mgr->EmptyString();
```
rsmmr, line 16: Just `return` if `len` is zero?

src/analyzer/protocol/rawstream/RawStream.cc
```diff
@@ -0,0 +1,25 @@
...
+    BifEvent::enqueue_rawstream_deliver(this, Conn(), orig, std::move(s));
+}
+void RawStream_Analyzer::Undelivered(uint64_t seq, int len, bool orig) {
+    analyzer::tcp::TCP_ApplicationAnalyzer::Undelivered(seq, len, orig);
+
+    BifEvent::enqueue_rawstream_undelivered(this, Conn(), orig, seq, len);
```
rsmmr, line 22: Same here on `len==zero` (not sure if that even happens, but for consistency in case if ever did). Or the other way around: always pass even zero to script-land for both delivered and undelivered.

rsmmr (approved):
Looks useful to me, and indeed nothing we'd already support in any other way. Two high-level points:

- my main worry is performance if this ended up sending larger amounts of data through scriptland (same as `tcp_contents()`, though this is arguably better because more tailored). We could think about further use-cases to support more directly that don't need going through scripts (like recording the data to disk).
- What do you think about naming this just `Stream` analyzer rather than `RawStream`? (I'd suggest `Contents` actually, but that's kind of taken with `tcp_contents()` and hence confusing.)

awelzel:
> my main worry is performance if this ended up sending larger amounts of data through scriptland (same as tcp_contents(), though this is arguably better because more tailored). We could think about further use-cases to support more directly that don't need going through scripts (like recording the data to disk).

I'll expand the docstring as you suggested and concretely mention that a native parser is usually the better choice. There's also the concern of multiple `stream_deliver(...)` users needing to coordinate based on connection state, but that's really a new concept.


> What do you think about naming this just Stream analyzer rather than RawStream? (I'd suggest Contents actually, but that's kind of taken with tcp_contents() and hence confusing.)

Thanks, yes, I wasn't particularly fond with the "raw" part.

I realized we have a similar `DataEvent` analyzer on the file analysis side, actually. I'll go with `stream_event` / `StreamEvent` - `Analyzer::ANALYZER_STREAM_EVENT` seem pretty obvious. Events will be `stream_deliver(...)` and `stream_undelivered()`.

src/analyzer/protocol/rawstream/RawStream.cc
awelzel: I need to admit I don't really no when we'd get `0`, but thought propagating it to script-land seemed reasonable in case it means something, so I'll keep the zero cases.

PR #4091, opened 2024-12-06
Add license-header check and update .h and .cc files

awelzel:
Spicy-inspired pre-commit hook to ensure a license header exists - for now this is limited to .h and .cc files.

Is this something people think is worth having/doing in Zeek, too?

The first commit adds back a missing `"COPYING"` word in a number of files via a simple `sed`. I think that makes sense to pick up one way or the other.


bbannier:
I think this is useful (well, I added a similar check to Spicy 😅).

FWIW, we are much more aggressive over there in that we require a license header for almost everything (we allow absence in files like e.g., not distributed tests, documentation files, selected redistributed files tracked in tree, or`.gitmodules`). This is done with the idea that it is better to do this too often than to miss it (@0xxon might have more fact-based take). I think it would definitely make sense here to require a license header in Zeek script files as well.

timwoj:
I don't see anything wrong with this either. I'm admittedly bad about adding the headers to new files, so having the reminder would be nice.

timwoj (approved):

PR #4093, opened 2024-12-06
Add HTTP::upgrade_content_type_analyzers

awelzel:
Follow-up for #4088 - allow selecting the analyzer during a HTTP upgrade based on Upgrade and Content-Type headers.

I think this makes some sense given the report in #4068, albeit it's not very generic, so looking for feedback whether this seems fine, or we should rethink this.

The most generic version would likely be a new scripting layer hook and draining of the event loop similar to what we currently do with the WebSocket analyzer configuration, but that's already a bit contentious, too.


awelzel:
Closing after chatting with @rsmmr - the `[Upgrade, Content-Type]` selection might be very Docker specific. We'll wait for further feedback from users if `HTTP::upgrade_analyzers` is sufficient.

PR #4095, opened 2024-12-06
Update python magic numbers for signature matching

timwoj:
This adds the magic numbers of Python bytecode for versions 3.8 - 3.14-prerelease. It also splits the python signatures out into a separate file, similar to what we do for Java bytecode.


awelzel (approved):
Unless we have a test, maybe could pick a file from a `__pycache__` directory and do something like "testing/btest/scripts/base/frameworks/notice/file-info-no-conns.zeek" uses to see if the mimetype detection works?

LGTM modulo the test failures in plugin/hooks.zeek

PR #4097, opened 2024-12-09

bbannier:
This PR should be merged when the first RC for `release/7.1` (does not yet exist) is cut. At that time we should also make sure that it correctly reflects the current status of Spicy's development branch.


timwoj:
Any reason not to pull this in before I branch 7.1?

timwoj (approved):

PR #4099, opened 2024-12-09
DNS/dns_binds_rr: Fix complte to complete typo, switch to count

awelzel:
From my reading in the docs the complete_flag should only ever be a single byte, so add a weird for when it is longer, but use count as the new type.

Relates to https://github.com/zeek/zeek/pull/4098#discussion_r1876277106


timwoj (approved):
🤦🏼

PR #4107, opened 2024-12-10
Some minor NEWS additions for v7.1

timwoj:


NEWS
```diff
@@ -192,6 +198,10 @@ Changed Functionality
   Previously, ``network_time()`` was used. This matters if ``Broker::publish()``
   is called within scheduled events or called within remote events.
 
+- The SSL analyzer now reports the correct version of SSLv2 connections as
+  passed during the hello messages. Zeek previously reported it as v2 always,
+  no matter was was passed.
```
0xxon, line 203: I would phrase this slightly different. How about

The SSL analyzer now reports the correct version when an SSLv2 client hello is used. Zeek previously always reported these as v2, even when the v2 client hello indicated support for a later version of SSL.

I know it's splitting hairs a bit, but the semantics are quite important :)

0xxon (changes requested):

NEWS
timwoj: Done, thanks

0xxon (approved):

PR #4110, opened 2024-12-11
Add Spicy analyzers to Coverity builds

### LINKED ISSUE #4103: Add Spicy to Coverity build
We currently disable Spicy for the Coverity build.

https://github.com/zeek/zeek/blob/1c42bfc7156800c1407d1487a80f3d8e00efda0e/.github/workflows/coverity-scan.yml#L46

As we move more analyzers in Zeek to Spicy we should add enable Spicy to get coverage. Ideally such a Coverity run would cover generated analyzers, as well as the runtime code the use.

timwoj:
I believe the original reason we disabled this was that Spicy builds used to take a very long time and lots of memory to complete. That's not the case anymore. Unfortunately we can't verify that this is fine locally because of how Coverity works, but it doesn't hurt to just add it and remove it later if it's bad. We run these builds every three days via a cron job, so we'll have to wait for the next one to trigger.

Fixes #4103

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: "recursive"
 
@@ -43,7 +43,7 @@ jobs:
             wget
 
       - name: Configure
-        run: ./configure --build-type=debug --disable-broker-tests --disable-spicy
+        run: ./configure --build-type=debug --disable-broker-tests
 
       - name: Fetch Coverity Tools
         env:
@@ -59,8 +59,8 @@ jobs:
 
       - name: Build
         run: |
-          export PATH=`pwd`/coverity-tools/bin:$PATH
-          ( cd build && cov-build --dir cov-int make -j $(nproc) )
+          export PATH="$PWD/coverity-tools/bin":$PATH
+          ( cd build && cov-build --dir cov-int make -j "$(nproc)" )
           cat build/cov-int/build-log.txt
 
       - name: Submit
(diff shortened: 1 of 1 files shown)


bbannier (approved):
Thanks! 🚀

Are there any path-based suppression set in the Coverity UI which would exclude results from `auxil/spicy`?

Since the workflow uses a pretty old Ubuntu release and compiler setup I confirmed that Zeek w/ Spicy actually builds there (in the `ubuntu:20.04` Docker image, not the GH image). This all works fine.

If you want and since you are at it, we could bump the used GH actions and fix some shell nastiness,

```patch
diff --git a/.github/workflows/coverity-scan.yml b/.github/workflows/coverity-scan.yml
index 2a6706a27..6bb7eef9c 100644
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: "recursive"
 
@@ -59,8 +59,8 @@ jobs:
 
       - name: Build
         run: |
-          export PATH=`pwd`/coverity-tools/bin:$PATH
-          ( cd build && cov-build --dir cov-int make -j $(nproc) )
+          export PATH="$PWD/coverity-tools/bin":$PATH
+          ( cd build && cov-build --dir cov-int make -j "$(nproc)" )
           cat build/cov-int/build-log.txt
 
       - name: Submit
@@ -70,9 +70,9 @@ jobs:
           cd build
           tar czf myproject.tgz cov-int
           curl \
-            --form token=${COVERITY_TOKEN} \
+            --form token="${COVERITY_TOKEN}" \
             --form email=zeek-commits-internal@zeek.org \
             --form file=@myproject.tgz \
-            --form "version=`cat ../VERSION`" \
-            --form "description=`git rev-parse HEAD`" \
+            --form "version=$(cat ../VERSION)" \
+            --form "description=$(git rev-parse HEAD)" \
             https://scan.coverity.com/builds?project=Bro
```

timwoj:
> Are there any path-based suppression set in the Coverity UI which would exclude results from auxil/spicy?

Yes, but let's wait to see what the findings look like first. The UI exclusions are just filtering what it shows on the UI, not what it actually finds from the scan.

> Since the workflow uses a pretty old Ubuntu release and compiler setup I confirmed that Zeek w/ Spicy actually builds there (in the ubuntu:20.04 Docker image, not the GH image). This all works fine.

I saw that too. Their tooling should be fine on ubuntu 22. I ran the `cov-build` part of it just now and it ran all the way through. I scrolled through the build output too, but I didn't try uploading the results. I'll push a separate update for that.

PR #4112, opened 2024-12-11
Pre-compute the node topics for all pool entries.

JustinAzoff:
A zeek script profile showed a small percentage of time spent in Cluster::node_topic, but this never changes and can be cached.

### DIFF
--- a/scripts/base/frameworks/cluster/pools.zeek
+++ b/scripts/base/frameworks/cluster/pools.zeek
@@ -18,6 +18,8 @@ export {
 		site_id: count;
 		## Whether the node is currently alive and can receive work.
 		alive: bool &default=F;
+		## The pre-computed result from Cluster::node_topic
+		topic: string;
 	};
 
 	## A pool specification.
@@ -172,7 +174,7 @@ function hrw_topic(pool: Pool, key: any): string
 
 	local site = HashHRW::get_site(pool$hrw_pool, key);
 	local pn: PoolNode = site$user_data;
-	return Cluster::node_topic(pn$name);
+	return pn$topic;
 	}
 
 function rr_topic(pool: Pool, key: string): string
@@ -198,7 +200,7 @@ function rr_topic(pool: Pool, key: string): string
 
 		if ( pn$alive )
 			{
-			rval = Cluster::node_topic(pn$name);
+			rval = pn$topic;
 			break;
 			}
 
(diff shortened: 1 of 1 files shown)


awelzel (approved):
LGTM, thanks. Sharing the flamegraph likely difficult?

JustinAzoff:
I don't think there's anything in the flame graph that's sensitive, will check.  I can always filter it for /base if there are any custom scripts mentioned.

I did work out how to do a little benchmark this morning:

```
@load base/frameworks/cluster
@load base/frameworks/cluster/pools

module Cluster;

event zeek_init()
{
	init_pool_node(Cluster::logger_pool, "logger-01");
	init_pool_node(Cluster::logger_pool, "logger-02");
	init_pool_node(Cluster::logger_pool, "logger-03");

	Cluster::logger_pool$nodes["logger-01"]$alive=T;
	Cluster::logger_pool$nodes["logger-02"]$alive=T;
	Cluster::logger_pool$nodes["logger-03"]$alive=T;

	local n = 0;
	while (++n < 20000000) {
		Cluster::rr_topic(Cluster::logger_pool, "Cluster::rr_log_topic");
	}
}
```

Not trying to be super accurate here since it should be a big change, or not:

```
[nix-shell:~/src/zeek]$ time zeek ./pool-test.zeek

real	0m42.661s
user	0m42.080s
sys	0m0.439s

[nix-shell:~/src/zeek]$ git checkout -
Switched to branch 'topic/justin/pools-cache-node-topic'
Your branch is up to date with 'origin/topic/justin/pools-cache-node-topic'.

[nix-shell:~/src/zeek]$ time zeek ./pool-test.zeek

real	0m33.351s
user	0m32.604s
sys	0m0.737s
```

odd thing is with ZAM enabled, the runtime is the same.  @vpax Does ZAM optimize this out somehow?  I guess it was still worth doing since ZAM isn't on-by-default yet, but very nice to see that ZAM may make minor hand optimizations like this obsolete.

vpax:
Ah here's what's going on. ZAM is able to identify that the topic is constant because it's constructed by `constant + constant + constant` (where all of those are string constants), which it folds. So this is the perils of using microbenchmarks in the face of a decent optimizer.

vpax:
(I confirmed this by using `--dump-final-ZAM` as well as `--optimize-file=` to limit compilation to just the benchmark script)

JustinAzoff:
Ah, I wouldn't say perils... Usually I do the little microbenchmarks to determine if the change even matters.. With ZAM off it makes a big difference.  I suspected it might inline the function call, but didn't realize it could also optimize the string concatenations too!

The full results I get with/without are:

|  | ZAM Off  | ZAM On |
| ------------- | ------------- | - | 
| Master  | 42s  | 14s |
| Branch  | 33s  | 14s | 

So.. either way ZAM wins :-)

PR #4113, opened 2024-12-12
Fix ZAM's implementation of `Analyzer::name()` BiF

ckreibich:
ZAM's analyzer lookup implementation deviates from Zeek's in consideration of remappings, which means that running ZAM produced conn.log service names prefixed with "spicy_" whereas regular Zeek doesn't.

I'm mildly uneasy now that there might be other such problems, because I'm pretty sure that for a while we made updates to BiFs without even realizing that there are ZAM counterparts to keep in sync. But scanning `BuiltIn.cc` I'm thinking those should only be a handful, and in a quick scan I didn't find further discrepancies.

@Mohan-Dhawan, @vpax, fyi.


awelzel (approved):
Added to 7.0.5.

Welp, thanks. I guess centralizing the implementation(s) would make some sense, or figuring out why calling BiFs is so slow.

vpax:
Great to see that this had an easy fix!

Given there aren't a lot of known-to-ZAM-opt BiFs, we could consider annotating the existing ones with comments to that effect.

PR #4120, opened 2024-12-12
Fix warning about Broker::publish call

timwoj:
This fixes the following warning when running ci/update-zeekygen-docs.sh:

> warning in /home/tim/projects/zeek-master/scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek, line 29: Non-broker cluster backend configured and Broker manager inactive. Did you mean to use Cluster::publish() instead of Broker::publish()?

Strangely, this change doesn't cause any changes in the btest baselines, which I was definitely expecting.

### DIFF
--- a/scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
+++ b/scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
@@ -26,7 +26,7 @@ event node_fully_connected(name: string, id: string, resending: bool)
 		event cluster_started();
 
 		for ( topic in Cluster::broadcast_topics )
-			Broker::publish(topic, Cluster::Experimental::cluster_started);
+			Cluster::publish(topic, Cluster::Experimental::cluster_started);
 		}
 	}
 


ckreibich:
This one is a little subtle — I wonder whether this is actually a case of not properly recognizing whether we're running clustered at all, because I doubt there's really a non-Broker backend configured. I think Arne steered clear of the scripts in `nodes-experimental/` because they're specific to the connection establishment behavior we have with Broker. I'll check a few things.

ckreibich:
Okay so the cause is that the conditional [`@load` of ZeroMQ](https://github.com/zeek/zeek/blob/master/scripts/test-all-policy.zeek#L14-L18) in `test-all-policy.zeek`, which the Zeekygen configuration pulls in, switches the backend to ZeroMQ but we load the manager's experimental cluster code after that, and it's built for Broker — hence the warning. A couple of things:
- Afaik the only point of loading scripts for Zeekygen is to generate documentation, not to absolutely load all scripts including ones commented out in `test-all-policy`. The `node-experimental/manager.zeek` script does not add documentation other than the header comment, so it wouldn't be a big deal to skip it. The same holds for `frameworks/cluster/backend/zeromq/connect.zeek`. We could start a section in `zeekygen/__init__.zeek` that lists things that _currently_ don't have documentation, so we don't lose sight of them.
- For `test-all-policy` and Zeekygen we should probably switch the backend back to Broker after the ZeroMQ stuff, like `test-all-policy-cluster.test` does [here](https://github.com/zeek/zeek/blob/master/testing/btest/coverage/test-all-policy-cluster.test#L19-L21). I tried this and this works (as in, it silences the warning).
- For both `test-all-policy` and Zeekygen there's an assumption that one can (more or less) load "everything". With multiple backends built in, this would mean being able to switch through those at load/runtime to cover all script content — I'm not sure that's going to work? We might at some point need multiple invocations, with varying (backend) configurations.

@awelzel  let me know if I'm missing something there ... a bit subtle.

awelzel:
Thanks for the catch Tim.


> I think Arne steered clear of the scripts in nodes-experimental/ because they're specific to the connection establishment behavior we have with Broker.

Correct. I'd actually for the time being keep the `Broker::publish()` in those scripts. At least `node_fully_connected()` is Broker specific. `cluster_started()` could make some sense for other backends, too, but their implementation is likely different (and becomes difficult with cluster-layout-less clusters, but that's far out, though closer with a central message broker).

> I'm not sure that's going to work? We might at some point need multiple invocations, with varying (backend) configurations.

If it's only about script documentation, we should probably stick to a single invocation.

Maybe ignoring redefinitions from certain (all?) files might make sense instead so that the documentation doesn't show "redefinition" just because another backend's script was loaded via test-all-policy.

https://docs.zeek.org/en/master/scripts/base/init-bare.zeek.html#id-Cluster::backend

Screenshot from 2024-12-13 09-36-53


To me this is mostly a quirk today, but if someone feels strongly could also skip the redefs when running under zeekygen explicitly in `zeromq/main.zeek` (or have a new `zeromq/redefs.zeek` file and that's then not loaded under zeekygen).

> @awelzel let me know if I'm missing something there ... a bit subtle.

To add to the subtle things, we disable running event handlers of certain modules under zeekygen:

https://github.com/zeek/zeek/blob/94f64a6dc6574a9cc4a01afbbbabfbaa2599943a/scripts/zeekygen/__load__.zeek#L28-L39

I'd actually prefer adding `Cluster::Experimental` here (ZeroMQ is already there). The warning about `Broker::publish()` doesn't strictly mean it's inappropriate to use.

awelzel:
> To me this is mostly a quirk today, but if someone feels strongly could also skip the redefs when running under zeekygen explicitly in zeromq/main.zeek (or have a new zeromq/redefs.zeek file and that's then not loaded under zeekygen).

@bbannier mentioned to me that this is wanted: Show all possible locations of redefinitions in the documentation.

awelzel:
Replaced by #4125. Closing.

PR #4121, opened 2024-12-12
Fixes for -O gen-standalone-C++

vpax:
Per https://github.com/zeek/zeek/pull/4096, recent changes to doing more comprehensive AST profiling for script optimization broke `-O gen-standlone-C++` functionality. This PR restores it. Along the way it streamlines the common case of there not being any "global statements" (a requirement for standalone C++), which in turn led to identifying some BTests that in need of tweaks (either a `TEST-REQUIRES` to ignore them, or other changes to accommodate having multiple different bodies).


src/script_opt/CPP/Exprs.h
```diff
@@ -142,6 +142,13 @@ std::unordered_map<const EnumType*, std::unordered_map<int, int>> enum_val_mappi
...
+struct EnumMappingInfo {
+    int enum_type; // as a global offset
+    std::string enum_name;
+    bool create_if_missing;
+};
+
 // For each entry in "enum_mapping", the EnumType (as a global offset) and
 // name associated with the mapping.
-std::vector<std::pair<int, std::string>> enum_names;
+std::vector<EnumMappingInfo> enum_names;
```
awelzel, line 154: Total flyby, but shouldn't this have an `extern` in case Exprs.h is included in multiple .cc files? There's no `#pragma once`, so not sure, maybe this is on purpsoe.

awelzel (approved):
Very light skim.

src/script_opt/CPP/Exprs.h
vpax: Those headers are all included in only one place, as part of the `CPPCompile` class (via `Compile.h`). I split them out a while back when a reviewer pointed out it would be helpful if the different source files for various `CPPCompile` methods had corresponding headers focused on those methods (and associated member variables).

PR #4123, opened 2024-12-13
Removing no-longer used BTest alternatives and their associated baselines

vpax:
Talking with @ckreibich recently, he mentioned that there are a bunch of additional BTest alternatives for intermediary stages of script optimization, and they're no longer up to date. I checked whether these are still relied on (indirectly) by the `-a zam` and the answer is No. This PR removes them.


ckreibich (approved):
Now _that_'s a large PR I like! :-)

PR #4124, opened 2024-12-13
Simplifying ZAM instructions for executing "assert" statements

vpax:
The fix for ZAM's new support for `assert` statements in https://github.com/zeek/zeek/pull/4089 left around some now-redundant logic from the original implementation. This PR removes it, simplifying the code a bit.


timwoj (approved):

timwoj:
I approved this before noticing there were some test failures on the CI build. Do you want to look into those?

vpax:
> test failures on the CI build

They're both failing with
```
% 'btest-bg-wait 10' failed unexpectedly (exit code 1)
```
First thing to try is cranking that 10 to say 30, as we've encountered before issues with ZAM compilation taking long enough that the timeouts aren't long enough. If that's easy for you to do, please give it a try - if you're booked up, I'll see if I can get to it later today.

awelzel:
Okay, the timeouts might be a bit aggressive for ZAM, but the deterministic failure in both tasks is actually the following:

```
[#8] opt.validate-ZAM [zam] ... failed on retry #1
  % 'btest-diff output' failed unexpectedly (exit code 1)
  % cat .diag
  1248 valid, 1890 tested, 437 skipped
  --- /dev/fd/63	2024-12-13 07:09:20.347793710 +0000
  +++ /dev/fd/62	2024-12-13 07:09:20.347793710 +0000
  @@ -1 +1 @@
  -1248 valid, 1892 tested, 437 skipped
  +1248 valid, 1890 tested, 437 skipped
```

Totally naively: What do those numbers mean? Do they provide "testing value"? Would `-O validate-ZAM` fail with `exit(1)` or fatal error if something is actually "bad"? Because then we should probably ditch the output/baseline to avoid unnecessary churn.

Also, not sure this needs the `ZEEK_ZAM = 1` check, doesn't `-O validate-ZAM` anyhow imply that? We should just run it unconditionally on all platforms?

PR #4125, opened 2024-12-13
zeekygen: Disable Cluster::Experimental module

awelzel:
The nodes-experimental/manager.zeek file ends up calling Broker::publish() unconditionally, resulting in a warning. Skip running that code when generating documentation.

Replaces #4120

### DIFF
--- a/scripts/zeekygen/__load__.zeek
+++ b/scripts/zeekygen/__load__.zeek
@@ -32,6 +32,7 @@ event zeek_init() &priority=1000
 	# probably disable all modules, too.
 	disable_module_events("Control");
 	disable_module_events("Cluster::Backend::ZeroMQ");
+	disable_module_events("Cluster::Experimental");
 	disable_module_events("Management::Agent::Runtime");
 	disable_module_events("Management::Controller::Runtime");
 	disable_module_events("Management::Node");


scripts/zeekygen/__load__.zeek
```diff
@@ -32,6 +32,7 @@ event zeek_init() &priority=1000
 	# probably disable all modules, too.
 	disable_module_events("Control");
 	disable_module_events("Cluster::Backend::ZeroMQ");
+	disable_module_events("Cluster::Experimental");
```
bbannier, line 35: I don't think `Cluster::Experimental` fits with the others here (_Disable events in modules that use zeek_init() to do stuff and may fail when run under zeekygen_, haven't validated that that is actually true).

The warning we want to suppress comes from

https://github.com/zeek/zeek/blob/94f64a6dc6574a9cc4a01afbbbabfbaa2599943a/scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek#L29

and is raised from

https://github.com/zeek/zeek/blob/94f64a6dc6574a9cc4a01afbbbabfbaa2599943a/src/broker/messaging.bif#L64-L66

Is there any reason to not just skip emission of the warning there if we run under zeekygen, e.g.,

```patch
-       if ( zeek::broker_mgr != zeek::cluster::backend && ! zeek::broker_mgr->Active() )
+       if ( zeek::broker_mgr != zeek::cluster::backend && ! zeek::broker_mgr->Active()  && ! zeek::detail::zeekygen_mgr->IsEnabled() )
                zeek::reporter->Warning("Non-broker cluster backend configured and Broker manager inactive. "
                                         "Did you mean to use Cluster::publish() instead of Broker::publish()?");
```

awelzel: `node_fully_connected()` is scheduled a bit further down via zeek-init.
```
event zeek_init() &priority=-15
	{
	# Make sure the manager recognizes itself as ready if no
	# connections have to be initiated.
	if ( |connectees_pending| == 0 )
		event node_fully_connected(Cluster::node, Broker::node_id(), F);
	}
```

> Is there any reason to not just skip emission of the warning there if we run under zeekygen, e.g.,

Not particularly liking that one. Agree that the proposed one is fairly ad-hoc, but this one feels even more ad-hoc :-)

zeekygen_mgr is also *always* enabled, so the proposal would disable the warning in all default configurations (zeekctl, manual zeek invocations, ...), not just when generating docs.

bbannier: > zeekygen_mgr is also always enabled, so the proposal would disable the warning in all default configurations (zeekctl, manual zeek invocations, ...), not just when generating docs.

Ah that's unfortunate and I missed it. I also now see that even with this line we still emit documentation for the module 🤷

bbannier (approved):

PR #4131, opened 2024-12-16
Expr: Fix folding of pattern values to support == and !=

### LINKED ISSUE #4084: Comparing vectors of patterns reports a fatal error
In 7.0 the following reports a type error:

```
print vector(/foo/, /bar/) == vector(/x/, /bar/);
```

```
error in <stdin>, line 1: illegal comparison (vector(/^?(foo)$?/, /^?(bar)$?/) == vector(/^?(x)$?/, /^?(bar)$?/))
```

After https://github.com/zeek/zeek/commit/7670a6f5896271e6670d96313f44a7ee850a2018, it's now reporting an error that's a bit less informative:

```
fatal error in <stdin>, line 1: BinaryExpr::PatternFold (vector(/^?(foo)$?/, /^?(bar)$?/) == vector(/^?(x)$?/, /^?(bar)$?/))
```

awelzel:
The fatal error is actually triggered at runtime, so it's a bit dangerous for users, but not sure there's many use-cases to compare vectors of patterns.

Closes #4084


awelzel:
@vpax - could you give a hint (or a hunk) to fix ZAM support, too?

timwoj (approved):
I was going to pick this one up once I got these releases sorted. Thanks for beating me to it.

vpax:
@awelzel looks like this didn't work for ZAM because my previous fix was boneheaded-ly incomplete. Try this patch:
```
diff --git a/src/script_opt/ZAM/OPs/rel-exprs.op b/src/script_opt/ZAM/OPs/rel-exprs.op
index 92c35b7ce..f4ed1a023 100644
--- a/src/script_opt/ZAM/OPs/rel-exprs.op
+++ b/src/script_opt/ZAM/OPs/rel-exprs.op
@@ -29,7 +29,7 @@ eval-type P	strcmp($1->Get()->PatternText(), $2->Get()->PatternText()) == 0
 eval-mixed P S	$1->MatchExactly($2->AsString())
 
 rel-expr-op NE
-op-type I U D S T A N F
+op-type I U D S T A N F P
 vector
 eval $1 != $2
 eval-type S	Bstr_cmp($1->AsString(), $2->AsString()) != 0
@@ -37,6 +37,7 @@ eval-type T	! $1->EqualTo(*$2)
 eval-type A	$1->AsAddr() != $2->AsAddr()
 eval-type N	$1->AsSubNet() != $2->AsSubNet()
 eval-type F	$1->GetName() != $2->GetName()
+eval-type P	strcmp($1->Get()->PatternText(), $2->Get()->PatternText()) != 0
 eval-mixed P S	! $1->MatchExactly($2->AsString())
 
 # Note, canonicalization means that GE and GT shouldn't occur
```

awelzel:
> @awelzel looks like this didn't work for ZAM because my previous fix was boneheaded-ly incomplete. Try this patch:

Thanks, that seems to have done the trick. I had to do a full rebuild, however, or else I tickled a segfault. Is there maybe a missing dependency between `.op` files and the output? Mostly leaving the comment as a note to hopefully remember when something is weird :-)

timwoj:
Generated files can sometimes need to run the build twice (once to force regeneration, once to guild the regenerated files). I've seen that happen with binpac/bifcl code as well.

vpax:
> I had to do a full rebuild, however, or else I tickled a segfault. Is there maybe a missing dependency between `.op` files and the output?

There's some form of dependency that Ninja doesn't identify properly, so it won't always regenerate everything that depends on `gen-zam` output. I sometimes just do `rm build/src/ZAM*` prior to a build to address this

timwoj:
I cherry-picked this into the release/7.1 branch as well.

PR #4134, opened 2024-12-17
Add spicy runtime support current_analyzer(), use it to set id in events

### LINKED ISSUE #4090: current_analyzer() in events from Spicy-based analyzers returns 0
The following script, prints `0` for `current_analyzer()`  when handling the `PostgreSQL::ssl_request()`, while it prints `5` for the `ssl_client_hello()`.

```
event PostgreSQL::ssl_request(c: connection) {
        print c$uid, "PostgreSQL::ssl_request", current_analyzer();
}

event ssl_client_hello(c: connection, version: count, record_version: count, possible_ts: time, client_random: string, session_id: string, ciphers: index_vec, comp_methods: index_vec) {
        print c$uid, "ssl_client_hello", current_analyzer();
}
```
```
zeek -Cr testing/btest/Traces/postgresql/psql-aws-ssl-require.pcap ./s.zeek 
C3Ol5q4STvFiNSB93, PostgreSQL::ssl_request, 0
C3Ol5q4STvFiNSB93, ssl_client_hello, 5
```

There might be some plumbing missing to set the analyzer_id on the events when queueing them.

timwoj:
Fixes #4090


src/spicy/runtime-support.cc
```diff
@@ -206,7 +206,7 @@ void rt::raise_event(const EventHandlerPtr& handler, const hilti::rt::Vector<Val
             throw InvalidValue("null value encountered after conversion");
     }
 
-    event_mgr.Enqueue(handler, std::move(vl));
+    event_mgr.Enqueue(handler, std::move(vl), util::detail::SOURCE_LOCAL, rt::current_analyzer()->GetID());
```
awelzel, line 209: If `rt::raise_event()` is called for a file analyzer or packet analyzer this will crash, I guess?

src/spicy/runtime-support.cc
```diff
@@ -221,6 +221,18 @@ TypePtr rt::event_arg_type(const EventHandlerPtr& handler, const hilti::rt::inte
     return zeek_args[idx];
 }
 
+zeek::analyzer::Analyzer* rt::current_analyzer() {
```
awelzel, line 224: As the other comment, there's file and packet analyzers, too. either `current_protocol_analyzer()`, or maybe `current_analyzer_id()` that happens to work for protocol analyzers and return `0` otherwise?

But probably @rsmmr  should mull this over :-)

bbannier: ➕ for making this work for all kinds. With that change most of the `spicy` test pass, but 4 still fail since some events don't seem to get raised. This might not be a purely Spicy problem anymore though.

src/spicy/runtime-support.cc
```diff
@@ -221,6 +221,24 @@ TypePtr rt::event_arg_type(const EventHandlerPtr& handler, const hilti::rt::inte
...
 }
 
+zeek::analyzer::ID rt::current_analyzer_id() {
+    auto _ = hilti::rt::profiler::start("zeek/rt/current_analyzer_id");
+
+    if ( auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie()) ) {
+        if ( auto x = cookie->protocol ) {
+            return x->analyzer->GetID();
+        }
+        else if ( auto x = cookie->file ) {
```
awelzel, line 231: Zeek's packet analysis analyzers don't set a current analyzer - e.g  [vxlan_packet](https://github.com/zeek/zeek/blob/991bc9644dd9b77f09fff992658f10500dce77e1/src/packet_analysis/protocol/vxlan/VXLAN.cc#L61-L67). For the one file analyzer we have (PE) there's also nothing obvious that sets the current analyzer.

Additionally, I'm slightly nervous that the IDs among the different analyzer kinds overlap. E.g. calling `disable_analyzer(c$id, current_analyzer())` within a packet analysis event (VXLAN, GENEVE, ...) or file analysis event would always attempt to disable a protocol analyzer which may by end-up working by accident.

So I'm starting to be conflicted if maybe returning 0 for non-protocol analyzers here makes more sense for the time being. Alternatively, whether protocol, packet and file analyzers should have separate ID spaces so that conflicts are impossible and one can tell by the numeric value what kind it is (similar to what we do with port spaces).

I'm approving anyhow as maybe I'm overthinking this.

awelzel (approved):
There's a CI failure for a baseline. I'm a bit timid and left a comment about ID collisions, but no good intuition if that's worth the trouble.

src/spicy/runtime-support.cc
timwoj: One problem is that the docs for `current_analyzer` don't specify what kind of analyzer it's returning an ID for:

```
## Returns the ID of the analyzer which raised the current event.
##
## Returns: The ID of the analyzer which raised the current event, or 0 if
##          none.
function current_analyzer%(%) : count
```

For protocol and file analyzers, there's only ever one instance of the analyzer so an ID is meaningful. The static counter could be moved up into the `zeek::id` namespace instead of having a separate counter for those two. 

For packet analyzers, the ID is set when the analyzer is instantiated. This means it's not really an ID for the type of analyzer but for that one instance. Is that really meaningful?

PR #4135, opened 2024-12-18
Fix for ZAM optimization of assigning a record field to result of "in" operation

vpax:
ZAM has single instructions for a number of record field assignments of the form `r$field = x op y`. However, these are only easy to implement for operations where `x` and `y` are the same type and don't require specialized handling. That's not the case for the `in` operator, but it was listed as one of the operations eligible for single-instruction. This PR removes it from that list, and adds a regression test. (I also inspected the other operations listed for this optimization and confirmed that there are corresponding ZAM instructions.)


awelzel:
Marked for 7.0.6. @timwoj likely also for 7.1-rc2.

PR #4137, opened 2024-12-18
Make empty container expansion more intuitive

### LINKED ISSUE #3933: Incorrect handling of init expression on set of sets
I would expect the following code to either work at runtime or be outright rejected:

```zeek
# foo.zeek
event zeek_init()
	{
	local vs: set[set[string]] = { [ set() ] };
	assert |vs| == 1;
	}
```

```console
$ zeek -b foo.zeek
error in foo.zeek, line 4: assertion failure: sizeof vs == 1
fatal error: errors occurred while initializing
```

I originally saw this with an initialization of the form

```zeek
local vs: set[count, set[string, bool]] = { [ 1, set() ] };
```

In a [Slack discussion](https://zeekorg.slack.com/archives/CTGDG83B8/p1726145403845739) there was an argument that this should not be supported at all and rejected since values in a `set` should be immutable (which a `set` is not). I'd argue that the usage here which does not modify the contained values at all could still be supported; we could still reject mutating operations on `vs` at runtime.

evantypanski:
Closes #3933

The issue is actually different from what the issue states IMO - sets of sets are completely supported and [even tested](https://github.com/zeek/zeek/blob/991bc9644dd9b77f09fff992658f10500dce77e1/testing/btest/language/set.zeek#L247-L270). It's just the syntax for initializing them shouldn't have square brackets (so it shouldn't be `{[set()]}`, it should be `{set()}` to add the empty set first).

Nonetheless, this case is still unintuitive. What happens is that list expansion gets triggered, so it tries to expand the `[set()]` and bring what's inside into the `{}`. But, evaluating `[set()]` just gets an empty expression, so you're left with just `{}`, which creates an empty set.

I could've evaluated the inside to find `set()` calls and treated those as important, then lifted them out, but that requires some knowledge about the context that I wasn't really happy with. Instead, a list will just never expand to 0 elements - if it would, it will just add the unexpanded list in. This has the unfortunate consequence of making `{[set()]}` invalid but something with multiple elements like `{[1, set()]}` is valid, but it seems cleaner to me.


evantypanski:
> ⚠️ Failed to start an instance: FAILED_PRECONDITION: Monthly compute limit exceeded!

:'(

evantypanski:
> can you split the formatting of the btest out into a separate commit?

sure no worries :) I just rolled with it when my editor automatically did it. I disabled that now for Zeek script because it's a little annoying though.

The baseline for the init-mismatch test still sucks because of globals ordering before everything else though, unfortunately

awelzel:
>  This has the unfortunate consequence of making {[set()]} invalid but something with multiple elements like {[1, set()]} is valid, but it seems cleaner to me.

I looked a bit - the following works with 7.0 (expands to `[b, 2], [c, 2]), but causes a runtime segfault with this PR:
```
event zeek_init() {
        local cs2 = { [set(), 1], [set("b", "c"), 2] };
        print type_name(cs2), cs2;
}
```

There's also the following which IMO should work according to the comments, but doesn't:
```
         local ssss: set[set[string], string] = {[set("a"), "b"]}; # E: inconsistent types in set constructor (set(a, b))                                                                
         print type_name(ssss), ssss;
```

This works after the following ad-hoc hunk (that I'm not sure is all that good):
```
--- a/src/Expr.cc
+++ b/src/Expr.cc
@@ -3083,8 +3083,12 @@ static bool expand_op_elem(ListExprPtr elems, ExprPtr elem, TypePtr t) {
         index = elem->GetOp1();
         yield = elem->GetOp2();
     }
-    else
+    else {
+        if ( t && t->IsSet() )
+            t = t->AsTableType()->GetIndices();
+
         index = elem; // this is a set - no yield
+    }

```

I'm also wondering if we should just skip expansion of unspecified sets right before rather than checking the evaluated result that this PR does. What'd your take be on that?
```
@@ -3109,6 +3113,10 @@ static bool expand_op_elem(ListExprPtr elems, ExprPtr elem, TypePtr t) {
         }
 
         if ( ie_i->GetType()->IsSet() ) {
+
+            // An unspecified set() is never expanded.
+            if ( ie_i->GetType()->AsTableType()->IsUnspecifiedTable() )
+                break;
+
             // Check for this set corresponding to what's expected
             // in this location, in which case it shouldn't be
             // expanded.
@@ -3139,12 +3147,6 @@ static bool expand_op_elem(ListExprPtr elems, ExprPtr elem, TypePtr t) {
             return false;
         }
 
-        // If this expands to nothing, assume that is important and don't expand.
-        if ( v->AsTableVal()->Size() == 0 ) {
-            elems->Append(elem);
-            return false;
-        }
-
```

Mostly motivated by the following which I'd guess would result in a typing conflict if it were to use the empty set, but it never sees that (not sure how to provoke it).
 
```
global x: set[string] = set("b", "c");

event zeek_init()
        {
        delete x["b"];
        delete x["c"];
        local ss: set[count, string] = {[1, set("a")], [2, x]};
        print "ss", type_name(ss), ss;
        }
```

There's more strange things happening if the global initializer is removed.

awelzel:
I'm also wondering a bit if initialize `{ ["a"] }` should be allowed for `set[string]`. For `set[string, string]`, `{ ["a", "b"], ...}` is required (and also for `table[string, string] of X`). Allowing `{ ["a"] }` would make it somewhat consistent with `table[string] of X` which requires `{ ["a"] = X() }` rather than `{ "a" = X() }`. But, not for this PR :-)

vpax:
My two cents (caveat: just from a skim of the above) is that we should move away, stylistically, from `{ ... }` initializers and towards explicit `set(...)` or `table(...)` initializers. So it doesn't give me much heartburn if `{...}` initializers behave in somewhat counterintuitive ways, as it's hard to get them right ...

evantypanski:
I tried walking a tightrope between not touching expansion and making this particular case better. But now with Arne's nudging I found cases that expansion causes more issues - this is regardless of whether `{...}` is used or not. Though, that's mainly from types where `set` has multiple types like what Arne pointed out.

Just as an added issue for fun, this slightly modified example from Arne gets me a fatal error (and probably shouldn't):

```
local ssss: set[set[string], string] = set(set("a"), "b");
```

```
$ zeek test.zeek
error in ./test.zeek, line 3 and string: incompatible types (set[string] and string)
fatal error in error: Type::AsTableType (error/table) (error)
[1]    82150 abort      zeek test.zeek
```

I think I need to sit on this for a day or two, I think. I'm not particularly a fan of that expansion.

PR #4138, opened 2024-12-18
Document `get_tag` to ensure that `name` exists

evantypanski:
This caused confusion and I don't think it's very intuitive. If called with a name that does not exist, this returns without a value in the script. Changing that seems like it could be more deprecation work - it has [specific handling](https://github.com/zeek/zeek/blob/991bc9644dd9b77f09fff992658f10500dce77e1/src/analyzer/analyzer.bif#L76) that will end up returning a `nullptr`. At least documenting it is better than nothing.

### DIFF
--- a/scripts/base/frameworks/analyzer/main.zeek
+++ b/scripts/base/frameworks/analyzer/main.zeek
@@ -100,6 +100,10 @@ export {
 
 	## Translates an analyzer's name to a tag enum value.
 	##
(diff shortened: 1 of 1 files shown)


timwoj (approved):

timwoj:
Just as a note, for simple docs changes like this you can add the `[skip ci]` tag to the commit message to keep Cirrus from doing builds.

PR #4139, opened 2024-12-20

bbannier:


timwoj:
The one build failure is probably because of https://github.com/zeek/zeek/pull/4140. The arm_debian12 build disables spicy. I'm somewhat surprised that doesn't fail on master too.

PR #4143, opened 2025-01-01
Fix incorrect uses of `zeek:see`

bbannier:
This fixes instances where `zeek:see` was used incorrectly so it was not rendered correctly. All these instances have been found by looking for `zeek:see` in the generated HTML where it should not be visible anymore.

I also removed a doc reference to `paraglob_add` which never existed.


evantypanski (approved):
🥳

PR #4153, opened 2025-01-07
Prefer stdout over stderr for `--help` messages

evantypanski:
is there a reason the `zeek --help` output goes to stderr? I just tried to `grep` it and couldn't so I fixed it

### DIFF
--- a/src/Options.cc
+++ b/src/Options.cc
@@ -72,145 +72,144 @@ bool fake_dns() { return getenv("ZEEK_DNS_FAKE"); }
 
 extern const char* zeek_version();
 
-void usage(const char* prog, int code) {
-    fprintf(stderr, "zeek version %s\n", zeek_version());
+void prompt_for_help(const char* prog) { fprintf(stderr, "Try '%s --help' for more information.\n", prog); }
 
-    fprintf(stderr, "usage: %s [options] [file ...]\n", prog);
-    fprintf(stderr, "usage: %s --test [doctest-options] -- [options] [file ...]\n", prog);
-    fprintf(stderr, "    <file>                          | Zeek script file, or read stdin\n");
-    fprintf(stderr, "    -a|--parse-only                 | exit immediately after parsing scripts\n");
-    fprintf(stderr, "    -b|--bare-mode                  | don't load scripts from the base/ directory\n");
-    fprintf(stderr, "    -c|--capture-unprocessed <file> | write unprocessed packets to a tcpdump file\n");
-    fprintf(stderr, "    -d|--debug-script               | activate Zeek script debugging\n");
-    fprintf(stderr, "    -e|--exec <zeek code>           | augment loaded scripts by given code\n");
-    fprintf(stderr, "    -f|--filter <filter>            | tcpdump filter\n");
-    fprintf(stderr, "    -h|--help                       | command line help\n");
-    fprintf(stderr, "    -i|--iface <interface>          | read from given interface (only one allowed)\n");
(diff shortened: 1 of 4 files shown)


bbannier:
I was initially confused with this as well, many tools in the Zeek org default to showing usage info on stderr (e.g., `spicyc` as well). The guideline in e.g., the GNU coding style is for explicitly requested usage info (e.g., via `--help`) to go to stdout; if usage info is shown due to incorrect args it might make sense to show it on stderr instead, but I am unsure how much a hassle implementing that would be.

src/Options.cc
```diff
@@ -554,7 +553,10 @@ Options parse_cmdline(int argc, char** argv) {
                 break;
 
             case '?':
-            default: usage(zargs[0], 1); break;
+            default:
+                fprintf(stderr, "ERROR: Unrecognized argument %s\n", zargs[optind - 1]);
+                usage(zargs[0], 1);
```
awelzel, line 558:
```suggestion
                exit(1);
```

I'm not a fan of the `usage()` spill personally either (slightly related): https://github.com/zeek/zeek/commit/55655f5ea11e48d5de2696ff49f5577849465b4e).

How about suppressing usage (inspiration from grep)?
```
$ grep -Y | grep more
grep: invalid option -- 'Y'
Usage: grep [OPTION]... PATTERNS [FILE]...   // think we can skip.
Try 'grep --help' for more information.
```

And possibly also `zeek -r` or other cases with missing arguments if it's easy:
```
$ grep -C
grep: option requires an argument -- 'C'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

This might break some people's workflows/tests, but :man_shrugging:

evantypanski: I changed it to look like this:

```
$ zeek --aoeu
ERROR: Unrecognized option --aoeu
Try 'zeek --help' for more information.
```

and:

```
$ zeek -r
ERROR: Option -r requires an argument.
Try 'zeek --help' for more information.
```

thoughts?

awelzel (approved):
I'm a fan, but assigning Christian and Robin to make the call.

ckreibich (approved):
Yep I like this too, thanks Evan!

If you're fired up for this kind of stuff, the help output itself could use some polish (for future PRs):
- It does not mention that you can assign variables on the command line, as in `zeek -r some.pcap LogAscii::use_json=T` etc
- Tim suggested a while back that we should do a scan for any additional environment variables to flag in the output.
- The output is a bit inconsistent when environment variables aren't set — some show as `()`, some as `(not set)`, some as `(off)`, etc.

Perhaps whenever you're next waiting for a build, or some such. :-)

stevesmoot:
> It does not mention that you can assign variables on the command line, as in zeek -r some.pcap LogAscii::use_json=T etc

I thought that was covered with -e and is deprecated backwards compat to not require -e.

ckreibich:
They're subtly different — `-e` really is code, while the assignment at the end of the line implies a variable assignment/redef. This works:
```
$ zeek -r http.trace LogAscii::use_json=T
```
... but this doesn't:
```
$  zeek -r http.trace -e 'LogAscii::use_json=T'
error in <command line>, line 1: const is not a modifiable lvalue (LogAscii::use_json)
```
You'd need to say:
```
$  zeek -r http.trace -e 'redef LogAscii::use_json=T'
```
On the flip-side you can do things like `zeek -e 'print "Hi Smoot"'`.

Also, I forgot that you can also tuck plugin loads onto the invocation. Details [here](https://github.com/zeek/zeek/blob/master/src/Options.cc#L568-L578).

PR #4154, opened 2025-01-07
Add recommended non-routable subnets

JW-Corelight:
224.0.0.0/24 (and 6to4 conversion 2002:e000::/40) from RFC5771	"Multicast Local Network Control Block" defined as non-routable.

239.0.0.0/8 (and 6to4 conversion 2002:ef00::/24) from RFC2365 "Administratively Scoped IP Multicast"

fec0::/10 from RFC3879 "Deprecated Site Local Addresses"

### DIFF
--- a/scripts/base/utils/site.zeek
+++ b/scripts/base/utils/site.zeek
@@ -76,6 +76,16 @@ export {
 		# 203.0.113.0/24 as a 6to4 address, see :rfc:`5737` and :rfc:`3056`
 		[2002:cb00:7100::]/40,
 
+		# Multicast Local Network Control Block, see :rfc:`5771`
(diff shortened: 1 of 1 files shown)


timwoj:
This changes the `local_resp` field in a bunch of the btests. Can you go through and update them?

ckreibich:
All of your prefixes look valid for inclusion to me, so I suggest we squeeze. :-) It's interesting, because I've only been monitoring the two IANA registries mentioned in site.zeek, and the ones you're adding aren't on that.

The baseline updates are a bit involved here, since they also include external testsuites. We can take it from here, Jeremy.

timwoj:
Superseded by #4164

Issue #4061, opened 2024-11-23
EDNS Functionality


anthonykasza:
This issue represents 2 complected problems. This issue relates to #1021.

Problems
=======
1. There is no EDNS data from query messages passed to scriptland for processing/logging. 
 
2. There are some EDNS events currently implemented which are supposed to pass EDNS data from reply messages to scriptland. They are `dns_EDNS_addl, dns_EDNS_cookie, dns_EDNS_ecs, dns_EDNS_tcp_keepalive`. However, it seems at least one, `dns_EDNS_ecs`, does not operate properly. 

Demonstration
===========
Both of these problems can be verified with the following script to debug scriptland and the following dig command to generate DNS traffic.

```
dig +subnet=1.1.1.1/24 example.com @8.8.8.8
``` 

```
event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qclass: count, original_query: string) {
  print "dns_request";
  print msg;
  print "";
}

event dns_EDNS_addl(c: connection, msg: dns_msg, ans: dns_edns_additional) {
  print "dns_EDNS_addl";
  print ans;
  print "";
}

event dns_EDNS_ecs(c: connection, msg: dns_msg, opt: dns_edns_ecs) {
  print "dns_EDNS_ecs";
  print opt;
  print "";
}

event dns_message(c: connection, is_orig: bool, msg: dns_msg, len: count) {
  print "dns_message";
  print msg;
  print "";
}
```

How I've failed
===========
- I tried defining a new event and enqueueing it from DNS.cc but had trouble handling the events in scriptland. I believe this could have been caused by `SendReplyOrRejectEvent`.
- I also looked at adding this functionality to the spicy-dns analyzer. I think there needs to be a RDType::OPT added to the ResourceRecord unit. My knowledge of Spicy is limited, unfortunately.

Suggestions
=========
A single event could be raised, similar to `ssl_extension`, when EDNS is present in a message. Then parsing of the EDNS OPT data could be handled in a straightforward manner from scriptland in an external package similar to [ssl-extensions](https://github.com/anthonykasza/ssl-extensions). 

To match `ssl_extension`, I propose the following prototype where "code" represent DNS EDNS0 Option Codes as defined by IANA.
```dns_edns_option(c: connection, is_query: bool, code: count, val: string)```

References
=======

List of currently allocated OPT codes by IANA:

https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11

[...]

PR #4155, opened 2025-01-07
Raise warnings when for DNS events that are not raised due to p_all_addl

0xxon:
By default, dns_skip_all_addl is set to false. This causes several events to not be raised. This change emits warnings when a user defines event handlers for events that will not be raised.

Furthermore, it adds notes about this behavior to the documentation. We also introduce a new BIF, `is_event_handled`, which checks if an event is handled.

Fixes GH-4061


testing/btest/Baseline/scripts.base.protocols.dns.event-handler-warning/.stderr
```diff
@@ -0,0 +1,6 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+warning in <...>/check-event-handlers.zeek, line 15: Used event 'dns_EDNS_tcp_keepalive' will not be raised because 'dns_skip_all_addl' is true
```
awelzel, line 2: I think for making noise this is good. I might have tried in C++ `InitPostScript()` in the DNS analyzer which would've allowed to get the locations of the actual handlers, but that'd be for polishing the UX. Script-land is also nice.

src/zeek.bif
```diff
@@ -5015,6 +5015,22 @@ function generate_all_events%(%) : bool
...
 	%}
 
+## Check if an event is handled. Typically this means that a script defines an event.
+## This currently is mainly used to warn when events are defined that will not be used
+## in certain conditions.
+##
+## event_name: event name to check
+##
+## returns: true if the named event is handled.
+function is_event_handled%(event_name: string%) : bool
```
awelzel, line 5025: Have you considered using `ev: any` instead of a string to prevent some potential string typoing by being able to use the event identifier directly (at the expense of needing dynamic type checking within the BiF).
```
if ( ! is_event_handled(dns_EDNS_addl) ) { }
```

In any case, with `string` or `any`, this should call `builtin_error` if name or identifier isn't an event to notice usage errors (or typos). For probing if an event exists, once can already use `@ifdef ( dns_EDNS_addl )`.

src/analyzer/protocol/dns/events.bif
```diff
@@ -608,6 +633,11 @@ event dns_TKEY%(c: connection, msg: dns_msg, ans: dns_tkey%);
 ## msg: The parsed DNS message header.
 ##
 ## ans: The parsed TSIG reply.
+#
+## .. note::
+##
+##      Note that this event will only be raised if ``dns_skip_all_addl``
```
awelzel, line 639: Slightly nicer for generated documentation:

```suggestion
##      Note that this event will only be raised if :zeek:see:`dns_skip_all_addl`
```

src/zeek.bif
0xxon: I am happy to add the builtin error.

nope, I have not considered making is possible to use this with an event handler directly. Unless you have strong feelings about this, I also don't see a super good reason to do this once the builtin error is added.

awelzel: No overly strong opinion. Resolving. Can add identifier support down the road if wanted without breaking existing users, too.

awelzel (approved):
Approving. There's one more test failure about ZAM BiF tracking.

Some background how to fix is here: https://github.com/zeek/zeek/pull/4105

@vpax - could we treat unknown BiFs as having side-effects by default and keep some maintenance list outside of the btest infrastructure?

I'm also wondering if annotating functions with `&pure` or `&no_side_effect` or `&idempotent` instead would be a bit easier. Not just for us, also for new contributors. In the future, BiFs added by external plugins could also inform ZAM about their attributes/properties this way.

vpax:
> could we treat unknown BiFs as having side-effects by default and keep some maintenance list outside of the btest infrastructure?

I _think_ that's indeed how they'll be treated. The BTest failure will occur but not the mis-use.

Regarding `&pure` etc. yeah I'd like that! A lot better than the maintainer having eyeball it. If we go down this path, at least some of the properties are documented in `FuncInfo.h` (I think that's the right place, haven't double-checked). There wound up being more than I had initially expected, such as idempotent-but-don't-call-until-Zeek-has-initialized.

> keep some maintenance list outside of the btest infrastructure

Likewise sounds fine as long as it's pretty streamlined to figure out when to update it. (That might wind up being equivalent to the current BTest approach, but if you have thoughts for how to do it more readily, great!)

PR #4161, opened 2025-01-08
Update COPYING to "now" :)

evantypanski:
happy 2024 and 2025!

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2023 by the Zeek Project. See LICENSE for details.
+# See the file "COPYING" in the main distribution directory for copyright.
 
 ---
 Language:        Cpp
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,10 +8,9 @@ repos:
     name: Check for license headers
     entry: ./ci/license-header.py
     language: python
-    types_or:
-      - "c"
-      - "c++"
-    exclude: '^(testing/btest/plugins/.*|testing/builtin-plugins/.*)$'
+    files: '\.(h|c|cpp|cc|spicy|evt)$'
+    types: [file]
+    exclude: '^(testing/btest/(Baseline|plugins|spicy|scripts)/.*|testing/builtin-plugins/.*)$'
 
 - repo: https://github.com/pre-commit/mirrors-clang-format
   rev: 'v19.1.4'
--- a/COPYING
+++ b/COPYING
@@ -1,4 +1,4 @@
-Copyright (c) 1995-2023, The Regents of the University of California
+Copyright (c) 1995-now, The Regents of the University of California
 through the Lawrence Berkeley National Laboratory and the
 International Computer Science Institute. All rights reserved.
 
--- a/ci/license-header.py
+++ b/ci/license-header.py
@@ -5,20 +5,15 @@ import sys
 
 exit_code = 0
 
-pat1 = re.compile(
+copyright_pat = re.compile(
     r"See the file \"COPYING\" in the main distribution directory for copyright."
 )
 
-# This is the copyright line used within Spicy plugin and popular in
-# Spicy analyzers.
-pat2 = re.compile(r"Copyright \(c\) 2... by the Zeek Project. See COPYING for details.")
-
 
 def match_line(line):
-    for pat in [pat1, pat2]:
-        m = pat.search(line)
-        if m is not None:
-            return True
+    m = copyright_pat.search(line)
+    if m is not None:
+        return True
 
     return False
 
--- a/scripts/base/protocols/ldap/main.zeek
+++ b/scripts/base/protocols/ldap/main.zeek
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
+# See the file "COPYING" in the main distribution directory for copyright.
 
 @load base/frameworks/reporter
 @load base/protocols/conn/removal-hooks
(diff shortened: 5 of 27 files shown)


timwoj:
I was looking at this the other day and got hung up on one question, then never asked it.

@awelzel @0xxon Is there a reason why the websocket and SSL spicy analyzers use their own copyright headers instead of pointing at the COPYING file like the rest of the Zeek source?

awelzel:
>  Is there a reason why the websocket and SSL spicy analyzers use their own copyright headers instead of pointing at the COPYING file like the rest of the Zeek source?

That's the copyright line style used in the Spicy project and there's probably some lineage from some `.spicy` file/template somewhere. LDAP/QUIC also have that one, but came in from an existing external project, so that might be justified.

I'm okay updating (and adding one to postgresql :sweat_smile: )

timwoj:
> LDAP/QUIC also have that one, but came in from an existing external project, so that might be justified.

Ah yeah, I just grepped for `2024`, but grepping for `Copyright` shows bunch of others as well.

0xxon:
I honestly still think that we should just replace it with - now, or just get rid of the year completely.

Precedents: 

https://github.com/facebook/react - no year
https://github.com/pytorch/pytorch/blob/main/LICENSE - no end year (just a blank space)
https://github.com/chromium/chromium/blob/main/LICENSE - year not updated (just lists 2015)
https://github.com/torvalds/linux/blob/master/COPYING - no year

evantypanski:
I pushed stuff that changes some notable analyzers to point to the top-level COPYING and changed its end date to "now" (the beginning date seems useful but also probably completely unnecessary).

I kept the non-Zeek copyrights because I don't know their background and didn't add *all* of the missing copyright comments, just updated obvious ones

happy to tweak more, though, that's just what conclusion I came to from that discussion

bbannier:
Just to surface a discussion I had with @awelzel some time ago, over in https://github.com/zeek/spicy we use a pre-commit hook [[1]](https://github.com/zeek/spicy/blob/83dd14183eb4747342f9000ff670e2e7384b77c5/.pre-commit-config.yaml#L15-L21)[[2]](https://github.com/zeek/spicy/blob/83dd14183eb4747342f9000ff670e2e7384b77c5/tests/Scripts/license-header.py) to ensure consistent license headers everywhere (we have a small allowlist for files which can stay without license header, but try to add license comments to every file which allows comments). Maybe it would be worthwhile to have a similar check for https://github.com/zeek/zeek to make it easier to keep things consistent.

ckreibich:
I don't feel strongly about the available options to change these to, but certainly ditching the current year is great since we'll never maintain them consistently. (Remember the guy who mass-created the typo-fix PRs? Here's one bot idea to unleash on the interwebs... :-)

I did a bit of background reading and mostly managed to stay awake, and what gather is that it's all nebulous. There are people who argue that the meaning of a year range needs to be explained somewhere, that one should instead itemize every year in which the code gets published, and other fun stuff. The BSD license templates I find online all use a single year, but don't elaborate on whether that's meant as a starting point, whether that year should be updated annually, expanded into a range/list, or what. 

All in all I think that removing the statements entirely and replacing them with a reference to `COPYING` (as you already do in much of the PR) is the best immediate option, and where we need dates, I'm inclined to go for "- now"  or "- present" because it at least attempts to convey continuous relevance, and it's handy to see when we introduced something.

Along those lines, do you want to push these tweaks a little further locally in the Zeek tree Evan? I see the following in `scripts/`:
```
spicy/zeek_file.spicy:# Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
spicy/zeek.spicy:# Copyright (c) 2020-2021 by the Zeek Project. See LICENSE for details.
base/protocols/ldap/main.zeek:# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
```
... and caught one in `.clang-format`.

We have ranges in many other places too, and since we have some many repos, this is quite a chore...

awelzel:
> Just to surface a discussion I had with @awelzel some time ago, over in https://github.com/zeek/spicy we use a pre-commit hook [[1]](https://github.com/zeek/spicy/blob/83dd14183eb4747342f9000ff670e2e7384b77c5/.pre-commit-config.yaml#L15-L21)[[2]](https://github.com/zeek/spicy/blob/83dd14183eb4747342f9000ff670e2e7384b77c5/tests/Scripts/license-header.py) to ensure consistent license headers everywhere

An adapted form exists in Zeek already since a few weeks.

@evantypanski , mind adding this hunk (or something compatible) so .evt and .spicy files are included, too?

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 2deb32b668..6352a4bf10 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,10 +8,9 @@ repos:
     name: Check for license headers
     entry: ./ci/license-header.py
     language: python
-    types_or:
-      - "c"
-      - "c++"
-    exclude: '^(testing/btest/plugins/.*|testing/builtin-plugins/.*)$'
+    files: '\.(h|c|cpp|cc|spicy|evt)$'
+    types: [file]
+    exclude: '^(testing/btest/(Baseline|plugins|spicy)/.*|testing/builtin-plugins/.*)$'
```

If you caught all the Spicy-style copyright lines, feel free to remove the extra regex for it in `ci/license-header.py`, too.

evantypanski:
pre-commit hook was changed and I updated any obvious ones I found, but I can update any more that get spotted too

note that docs would be the next logical place to take this - the copyright notice at the bottom has the wrong year and that year should probably be completely removed. Probably goes for both Zeek and Spicy docs

awelzel (approved):

PR #4165, opened 2025-01-10
Support for record extensions when using -O gen-standalone-C++

vpax:
This is the last of PRs related to compiling scripts to standalone C++. It covers an issue that I only found after putting in the earlier PRs: if a script extends a record using `redef record += ...` then the standalone C++ initialization code needs to add those fields explicitly. To enable the compiler to figure out if this is needed, we add a `Location` associated with each record field.

### DIFF
--- a/src/Type.h
+++ b/src/Type.h
@@ -618,11 +618,16 @@ public:
 
     const detail::AttrPtr& GetAttr(detail::AttrTag a) const { return attrs ? attrs->Find(a) : detail::Attr::nil; }
 
+    const detail::Location* GetLocationInfo() const { return &loc; }
+
     void DescribeReST(ODesc* d, bool roles_only = false) const;
 
     TypePtr type;
     detail::AttributesPtr attrs;
     const char* id = nullptr;
+
+private:
+    detail::Location loc = detail::GetCurrentLocation();
 };
 
 using type_decl_list = PList<TypeDecl>;
--- a/src/script_opt/CPP/RuntimeInits.cc
+++ b/src/script_opt/CPP/RuntimeInits.cc
@@ -383,11 +383,18 @@ TypePtr CPP_TypeInits::BuildRecordType(InitsManager* im, ValElemVec& init_vals, 
     auto r = cast_intrusive<RecordType>(inits_vec[offset]);
     ASSERT(r);
 
-    if ( r->NumFields() == 0 ) {
+    auto addl_fields = init_vals[2];
+
+    if ( addl_fields > 0 || r->NumFields() == 0 ) {
+        // We shouldn't be adding fields if the record doesn't have any
+        // existing fields - that would reflect an initialization botch.
+        if ( addl_fields > 0 && r->NumFields() == 0 )
+            reporter->InternalError("record unexpectedly empty when adding fields");
+
         type_decl_list tl;
 
         auto n = init_vals.size();
-        auto i = 2U;
+        auto i = 3U + addl_fields * 3;
 
         while ( i < n ) {
             auto s = im->Strings(init_vals[i++]);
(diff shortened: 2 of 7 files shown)


src/Type.h
```diff
@@ -623,6 +623,7 @@ class TypeDecl final {
     TypePtr type;
     detail::AttributesPtr attrs;
     const char* id = nullptr;
+    detail::Location loc = detail::GetCurrentLocation();
```
awelzel, line 626: Could you instead extend the non-default `TypeDecl` constructor a parameter and default it to to the current location? Mostly thinking that it'd be nice to have some way to controlthe location of a new `TypeDecl` even if you don't have a direct use.

Also, making the member private and adding a `GetLocationInfo()` returning a pointer to it would make it consistent with `class Obj`, which would look more streamlined when locations from either type are used / compared.

src/script_opt/CPP/RuntimeInits.cc
```diff
@@ -383,11 +383,16 @@ TypePtr CPP_TypeInits::BuildRecordType(InitsManager* im, ValElemVec& init_vals,
     auto r = cast_intrusive<RecordType>(inits_vec[offset]);
     ASSERT(r);
 
-    if ( r->NumFields() == 0 ) {
+    auto addl_fields = init_vals[2];
+
+    if ( addl_fields > 0 || r->NumFields() == 0 ) {
+        // We shouldn't be adding fields if the record doesn't have any
+        // existing fields - that would reflect an initialization botch.
+        ASSERT(addl_fields == 0 || r->NumFields() > 0);
```
awelzel, line 391: `FatalError()` or `InternalError`? This shouldn't be hot, so better to fail in non-debug bulids, too, no?

src/script_opt/CPP/Types.cc
```diff
@@ -189,9 +189,37 @@ shared_ptr<CPP_InitInfo> CPPCompile::RegisterType(const TypePtr& tp) {
...
+        if ( tp->Tag() == TYPE_RECORD ) {
+            auto tr = tp->AsRecordType();
+            for ( auto i = tr->NumOrigFields(); i < tr->NumFields(); ++i ) {
+                auto fd = tr->FieldDecl(i);
+                if ( filename_matches_opt_files(fd->loc.filename) ) {
+                    if ( addl_fields == 0 )
+                        addl_fields = i;
+                }
+                else if ( addl_fields > 0 ) {
+                    reporter->Error(
```
awelzel, line 207: Seems `FatalError()` works for Error + exit(1).

I'm not 100% sure what's going on here. If a user forgets to add a file that redefs to `only_files`, this triggers an error? It would probaby be helpful to output the filename from `fd->loc.filename` in the message?

src/Type.h
vpax: Making it private and adding `GetLocationInfo()` sounds good. Adding a constructor parameter seems a bit crowded to me given it already has a default parameter, but how about instead adding `SetLocationInfo()` to further reflect the `Obj` interface?

awelzel: > Adding a constructor parameter seems a bit crowded to me given it already has a default parameter, but how about instead adding SetLocationInfo() to further reflect the Obj interface?

Fair. Don't worry about `SetLocationInfo()` then and use the default initializer as you have :+1: If needed, can add it later, too.

src/script_opt/CPP/RuntimeInits.cc
vpax: Good point, fixed in the fixup I just pushed

src/script_opt/CPP/Types.cc
vpax: Yeah you're right it's fairly baffling. I added a comment above to explain the considerations, and expanded the error message to include the name of the record with the issue. (I'll add the filename too if it subsequently proves helpful, not sure whether it's needed given the record name.)

vpax:
I've done a fixup push to address the review comments, let me know if more is needed.

Also, in the process of further testing these I found yet another standalone issue, but I'll attend to it in a separate PR.

awelzel (approved):

PR #4169, opened 2025-01-12
Fix incomplete includes in Spicy SSL analyer C++ code

bbannier:
This appears to have been broken by
feec451bcee7c459bc9a93e39ae18dc41515ac17.


bbannier:
The failing test in `zeekctl_debian11` should be fixed by #4168.

PR #4170, opened 2025-01-13
Prevent copies in various places

bbannier:


auxil/spicy
bbannier, line 1: This still points to https://github.com/zeek/spicy/pull/1958, should eventually point to Spicy's `main`.

timwoj (approved):

PR #4172, opened 2025-01-14
Fix Coverity reports in Spicy generated code

bbannier:


awelzel (approved):
The `std::move()` in Initialize() is a bit weird as these are "consts", but defined locally due to laziness. If they were marked `const`, would that squelch coverity, too?

bbannier:
> The `std::move()` in Initialize() is a bit weird as these are "consts", but defined locally due to laziness. If they were marked `const`, would that squelch coverity, too?

I suspect it would, as would probably also making them `static`. That said, right now we create new values for each invocation and drop them onto the floor.

PR #4181, opened 2025-01-17
session/Manager::FindConnection() hardening

awelzel:
We silently broke users constructing conn_id records manually and
subsequently using them with lookup_connection() or connection_exists().

This is an attempt to at least report a runtime error about the situation
so it doesn't go completely unnoticed.

---

@timwoj, what do you think about this? Do you see other ways? Encoding errors in the `ConnKey.transport` feels a bit hacky, but it shouldn't take more space than the extra `bool` value, but also allows to not truncate the 2**16-1 value from script land that might be in there. 

---

Thanks to @pbcullen for reporting.


timwoj (approved):
I think encoding it there is fine. We already track the UNKNOWN value there, so the validation is fine also.

PR #4183, opened 2025-01-20
Ignore mDNS broadcasts.

jbaggs:
Ignore mDNS broadcasts from local devices. I'm pretty sure this is not what the original script authors wanted to detect, and it's not really "external", but also isn't likely to be in names a local DNS server is authoritative for.

### DIFF
--- a/scripts/policy/protocols/dns/detect-external-names.zeek
+++ b/scripts/policy/protocols/dns/detect-external-names.zeek
@@ -15,17 +15,18 @@ export {
 		## **must** be set appropriately for this detection.
 		External_Name,
 		};
+
+	## Default is to ignore mDNS broadcasts.
+	option skip_resp_host_port_pairs: set[addr, port] = { [[224.0.0.251, [ff02::fb]], 5353/udp] };
 }
 
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=-3
 	{
-	local multicast: set[addr] = { 224.0.0.251, [ff02::fb] };
 
 	if ( |Site::local_zones| == 0 )
 		return;
 
-	# Ignore mdns.
-	if ( c$id$resp_h in multicast && c$id$resp_p == 5353/udp )
+	if ( [c$id$resp_h, c$id$resp_p] in skip_resp_host_port_pairs )
 		return;
 
 	# Check for responses from remote hosts that point at local hosts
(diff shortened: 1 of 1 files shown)


awelzel (changes requested):
Hey @jbaggs,

looks good to me as a default. Just at the off-chance someone relies on this or wants to further customize to ignore a certain DNS server. Could you make an option for this:
```
module DNS;
export {
        option: skip_resp_host_port_pairs: set[addr, port] = { [[224.0.0.251, [ff02::fb]], 5353/udp] };
}

...

    if ( [c$id$resp_h, c$id$resp_p] in skip_resp_host_port_pairs )
       return;
```

Unclear there's a use, but this way it's configurable. If you have a better option name, go for it :-)

0xxon:
Hm. I really wonder about this one.

I completely agree that mDNS was not what we had in mind when we came up with dns.log. However, I am also not sure that completely excluding it by default is the way to go here.

If you drop Zeek into an environment, mDNS will give potentially quite interesting hints about what kind of machines are active on the local network and might allow you to later go back and see what kind of device is responsible for some traffic. Loosing that completely seems a bit harsh.

I might be off base here - but I think this change might deserve wider discussion.

jbaggs:
> Hm. I really wonder about this one.
> 
> I completely agree that mDNS was not what we had in mind when we came up with dns.log. However, I am also not sure that completely excluding it by default is the way to go here.
> 
> If you drop Zeek into an environment, mDNS will give potentially quite interesting hints about what kind of machines are active on the local network and might allow you to later go back and see what kind of device is responsible for some traffic. Loosing that completely seems a bit harsh.
> 
> I might be off base here - but I think this change might deserve wider discussion.

I agree that looking at MDNs broadcasts could be of interest, but it shouldn't be triggering an "external name" message. It would be easy enough to add an MDNs broadcast script with appropriate messages.

awelzel:
@jbaggs - thanks for adapting. Any chance you have some sample traffic that could be added as a test? (Can do during the merge, would just need a pcap).

jbaggs:
> @jbaggs - thanks for adapting. Any chance you have some sample traffic that could be added as a test? (Can do during the merge, would just need a pcap).
> 
> I wonder if an additional "heuristic" could be to check for `.local` or non `.local`, too.

I wouldn't rely on `.local`. The multicast IPs and port should be sufficient. No pcap. I'm about at the end of doing free work here.

0xxon:
I added a testcase with example traffic in #4215 and, while doing this, also added IPv6 support. That PR contains the original commit in its history.

I think that completely supersedes this PR - let me know if that is not the case.

PR #4186, opened 2025-01-22
Merge duplicate module sections in init-bare.zeek

timwoj:
This is something that's been bothering me for a bit. In init-bare.zeek, we have 9 sections marked `module GLOBAL`, not including the section at the beginning that is automatically in the global module. We also have two sections marked `module Tunnel`. This PR squashes the two `Tunnel` sections into one.

The larger change rearranges definitions from some of the `GLOBAL` sections into others, reducing it down to 3 additional `GLOBAL` sections. I left the ones where the definitions in that section depended on something being defined in a `module` section directly before, plus the ending section that initializes BIFs.

There's more cleanup that could be done in this file, especially deprecating a lot of definitions from the `GLOBAL` module and putting them in the modules where they make sense. For example, there's a block of Modbus types that could be moved into a Modbus module, and a huge block DNS types that could be moved into a DNS module.

I don't expect that this PR will improve performance since switching modules is fairly fast and this only affects startup performance during parsing, but it makes more sense to me to have as few blocks as needed.

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -144,8 +144,8 @@ type string_mapper: function(s: string): string;
 ## :zeek:see:`file_magic`
 type mime_match: record {
 	strength: int;    ##< How strongly the signature matched.  Used for
-	                  ##< prioritization when multiple file magic signatures
-	                  ##< match.
+			  ##< prioritization when multiple file magic signatures
+			  ##< match.
 	mime:     string; ##< The MIME type of the file magic signature match.
 };
 
@@ -197,10 +197,10 @@ const allow_network_time_forward = T &redef;
 ## A connection's transport-layer protocol. Note that Zeek uses the term
 ## "connection" broadly, using flow semantics for ICMP and UDP.
 type transport_proto: enum {
-    unknown_transport,	##< An unknown transport-layer protocol.
-    tcp,	##< TCP.
-    udp,	##< UDP.
-    icmp	##< ICMP.
+	unknown_transport,	##< An unknown transport-layer protocol.
+	tcp,	##< TCP.
+	udp,	##< UDP.
+	icmp	##< ICMP.
 };
 
 ## A connection's identifying 4-tuple of endpoints and ports.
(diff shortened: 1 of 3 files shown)


scripts/base/init-bare.zeek
```diff
@@ -389,6 +373,225 @@ type endpoint_stats: record {
...
+##
+## .. zeek:see:: io_poll_interval_default
+const io_poll_interval_live = 10 &redef;
+
+## Whether Zeek is being run under test. This can be used to alter functionality
+## while testing, but should be used sparingly.
+const running_under_test: bool = F &redef;
+
+module FTP;
+
```
awelzel, line 583: While here, maybe add an `export {` section ? Mostly for consistency with other `module`'s. Same for `SMTP` below.

scripts/base/init-bare.zeek
```diff
@@ -6138,59 +6179,7 @@ export {
...
-## .. zeek:see:: io_poll_interval_default
-const io_poll_interval_live = 10 &redef;
-
-## Whether Zeek is being run under test. This can be used to alter functionality
-## while testing, but should be used sparingly.
-const running_under_test: bool = F &redef;
+@load base/bif/event.bif
 
 global done_with_network = F;
 event net_done(t: time)
```
awelzel, line 6185: While here, mind doing the following?

```suggestion
event net_done(t: time)
        {
        done_with_network = T;
        }
```

I'm always wondering why that indentation is even more special than what we use elsewhere.

scripts/base/init-bare.zeek
```diff
@@ -6138,59 +6179,7 @@ export {
...
-##
-## .. zeek:see:: io_poll_interval_default
-const io_poll_interval_live = 10 &redef;
-
-## Whether Zeek is being run under test. This can be used to alter functionality
-## while testing, but should be used sparingly.
-const running_under_test: bool = F &redef;
+@load base/bif/event.bif
 
 global done_with_network = F;
```
awelzel, line 6184: Also wondering if the state variable could be deprecated, but feel free to disregard.

awelzel (approved):

scripts/base/init-bare.zeek
timwoj: From a quick grep, nothing is using it.

PR #4187, opened 2025-01-22
Add type expressions to Zeek script

### LINKED ISSUE #4173: Script type expression support
Inspiration from https://github.com/zeek/zeek/pull/4163#discussion_r1910076791

There already are [`TypeVal`](https://github.com/zeek/zeek/blob/b2222e97a134c6c6506bcc7fcafae955017e4cfa/src/Val.h#L1522) but no `TypeExpr` to match that. The expression version would be useful so you could, say, pass a primitive type into a function that expects a `TypeVal` and use its value.

I don't think it's a big deal to split the `type` production in the grammar up into cases that can easily produce a `TypeVal` without conflicts and those that cannot. Then at least some types could be passed directly as expressions, making it not necessary to create some arbitrary placeholder like `type str: string` when passing types into functions.

Adding `type` into the `expr` production added a lot of conflicts, but `TOK_STRING` added none, so my guess is we could find a number of good tokens to use that would just make this new `TypeExpr` without a bunch of conflicts.

evantypanski:
Closes #4173

This is mostly meant to make this snippet:

```
type str: string;
# ...
my_bif(str);
```

Into

```
my_bif(string)
```

I wanted to add a `type` type accessible via the frontend, but that seemed like a lot since then we'd have to get the parse-time type from a value... which then starts exploding. For this, at least we can guarantee to have the value if it's getting used. If not, you still get the error like "not a Zeek type (anything)"

few tiny things I'm just not very familiar with what the "Best practice" should be, regarding stuff like `ExprDescribe` output, script_opt stuff, etc. so feel free to nitpick this


src/Expr.cc
```diff
@@ -478,6 +478,15 @@ ExprPtr NameExpr::MakeLvalue() {
 }
 
 void NameExpr::Assign(Frame* f, ValPtr v) {
+    // Type types cannot be reassigned because that could require parse-time control
+    // flow to resolve a variable's type
+    if ( id->GetType()->Tag() == TYPE_TYPE ) {
+        if ( (id->IsGlobal() && id->GetVal()) || (! id->IsGlobal() && f->GetElementByID(id)) ) {
+            ExprError("type type cannot be assigned");
```
vpax, line 485: Can this check be done statically rather than dynamically?

src/Expr.h
```diff
@@ -109,6 +109,7 @@ enum ExprTag : int {
     EXPR_FROM_ANY_VEC_COERCE,
     EXPR_ANY_INDEX,
     EXPR_SCRIPT_OPT_BUILTIN,
+    EXPR_TYPE,
```
vpax, line 112: per the comment above this block, these are meant for internal expressions created during AST transformation, so the tag should come earlier

src/Expr.h
```diff
@@ -1672,6 +1673,25 @@ class CoerceFromAnyExpr : public UnaryExpr {
     ExprPtr Duplicate() override;
 };
 
+/**
```
vpax, line 1676: Nit, not the usual comment style used in headers

src/script_opt/ScriptOpt.cc
```diff
@@ -804,7 +804,8 @@ bool has_AST_node_unknown_to_script_opt(const ProfileFunc* prof, bool /* is_ZAM
         // EXPR_ANY_INDEX,
         // EXPR_SCRIPT_OPT_BUILTIN,
         // EXPR_NOP,
-#define SCRIPT_OPT_NUM_EXPRS  70
+        // EXPR_TYPE,
```
vpax, line 807: (and move it here too)

vpax:
Have you tried running `-a zam` BTests with this? In particular, I would expect that ZAM needs some tweaks to support it, and it would be nice to know where these arise. I guess the CI failures above point these up

testing/btest/language/type-expr-assignment.zeek
```diff
@@ -0,0 +1,10 @@
+# @TEST-DOC: Ensure redefining a type expression ID is an error
+# @TEST-EXEC: zeek -b %INPUT
```
awelzel, line 2: If you felt like actually ensuring the failiing exit code, can do the following (we don't do that a lot, so feel free to disregard).
```suggestion
# @TEST-EXEC-FAIL: unset ZEEK_ALLOW_INIT_ERRORS && zeek -b %INPUT
```

testing/btest/language/type-expr.zeek
```diff
@@ -0,0 +1,19 @@
...
+	local str = string;
+	local my_string: str = "hi there :)";
+	print my_string;
+	print my_global_string;
+
+	local integer = int;
+	local my_int: integer = 41;
+	my_int += 1;
+	print my_int;
+	}
```
awelzel, line 19: I presume it'll just work, but as one motivation was for bifs taking types, you could add another test for e.g. `from_json`:
```
zeek -e 'type str: string; print from_json("\"a\"", str)'
[v=a, valid=T]
zeek -e 'type countx: count; print from_json("1234", countx)'
[v=1234, valid=T]
```

Or `type_name(string)` could also be interesting :-)

src/Expr.h
0xxon: I actually think this is fine, from my inerpretation of our [style guide](https://github.com/zeek/zeek/wiki/Coding-Style-and-Conventions), this should be the preferred style for something like that. We are not very consistent about it yet.

To add to this - at least for newly added functionality, I think it would be neat to add documentation for the invidual functions. I am not sure it makes entirely much sense for this specific case, so I am leaving that up to you :)

src/parse.y
```diff
@@ -1265,24 +1282,12 @@ type:
 				$$ = new VectorType({AdoptRef{}, $3});
 				}
 
-	|	TOK_FUNCTION func_params
-				{
-				set_location(@1, @2);
-				$$ = $2;
-				}
-
 	|	TOK_EVENT '(' formal_args ')'
```
awelzel, line 1285: Any reason you kept `event`?

The following works and prints `type`, but it doesn't work for hook and function types, so my thinking is that it would be more consistent if it wouldn't work for `event` either (or make it work for all of them), just due to their close relationships, even if technically possible.

```
event zeek_init()
        {
        local et = event();
        print type_name(et);
        local etc = event(c: count);
        print type_name(etc);

        # local ht = hook(c: count);
        # local ft = function(c: count): bool;
        }
```

evantypanski:
> Have you tried running -a zam BTests with this? 

Indeed, and I ran the given tests with `zeek -b type-expr.zeek -O ZAM` and they give the expected output. I can't tell you why it works, though :)

I believe that the test failure is fixed in #4188 - that is unrelated

testing/btest/language/type-expr-assignment.zeek
evantypanski: Ah I didn't know that, thanks! I was pretty confused (though moving it to a static check means it's not needed)

src/parse.y
evantypanski: Reason was `simple_type` was all of the types that didn't make a conflict within `expr` - but I moved it out because having some reason beyond "it's easy to parse" makes sense :)

vpax:
> Indeed, and I ran the given tests with `zeek -b type-expr.zeek -O ZAM` and they give the expected output. I can't tell you why it works, though :)

Well it works because:
```
Skipping compilation of zeek_init due to unknown AST node type
```
😏

evantypanski:
> Skipping compilation of zeek_init due to unknown AST node type

Ah I'm happy that exists! At least this is new (and only narrowly useful) functionality

And the CI failure seems unrelated to me here

testing/btest/language/type-expr.zeek
```diff
@@ -0,0 +1,23 @@
+# @TEST-DOC: Test valid use of type expressions in scripts
+# @TEST-EXEC: zeek -b %INPUT
+# @TEST-EXEC: TEST_DIFF_CANONIFIER= btest-diff .stdout
+
+global global_str = string;
+global my_global_string: global_str = "hey!!";
+
+event zeek_init()
+	{
+	local str = string;
```
awelzel, line 10: Feel free to push back for this PR, but wonder if allowing `type` as "type"
```
type X: record {
    t: type;
};

function x(a: type): string {
        return type_anme(a);
}

event zeek_init() {
        local t: type = string;
        print x(t);
}
```

That would allow to change the `from_json()` signature to use `type` rather than `any`.

awelzel (approved):
I'm approving, recognizing ZAM support isn't fully there. No strong opinion whether that should be a follow up PR or done in this PR.

Also left another comment around allowing `type` as "type" which could be nice, but possibly separate.

Won't merge until someone chimed in.

testing/btest/language/type-expr.zeek
evantypanski: I indeed tried that, it seems technically possible, but the main issue is that `type` would need to be known statically. It technically is, but a function like `x` you point out errors out before even trying to figure out what value `a` will have. So in order to do this, we'd need to do some propagation to figure out what `a` is at parse time, OR treat types entirely as first-class citizens so more types can be figured out dynamically (so we don't need a million definitions of `x` for each type)

I actually really wanted to have this because it would make generics possible! (or something similar) I even wonder how this would be with `redef` structs and whatnot if going beyond what `any` can do... but that's for another time

```
function generic_func(ty: type, a: ty) {
    print a;
    print type_name(a);
}

event zeek_init() {
    generic_func(string, "hello:)"); # Prints "hello:)" then "string"
}
```

evantypanski:
Closing in favor of #4194. Not using this one so that this approach is still available for public, though.

new approach just uses `ConstExpr`

PR #4188, opened 2025-01-23
script_opt: Use emit_builtin_error() for Lookup-Conn

awelzel:
Still create a separate baseline file because ZAM doesn't provide detailed location information.

Patch for ZBI.op provided by @vpax


evantypanski (approved):
this makes #4187 pass with `btest -a zam` when rebased on it so I'm good with it :)

PR #4191, opened 2025-01-23
Fix options parsing segfault

evantypanski:
A command like this would segfault (particular number of arguments and last option expects an argument but is not given one):

```
zeek -b test.zeek --debug
```

The issue was that `getopt_long` was using a null element to determine what the end of the options array is. If it saw a non-null element after `--debug` it would say it's the argument for optarg, even if it's beyond `zeek_args.size()`. Instead, just make sure the array is null-terminated.

EDIT: This is probably dependent on the `libc` implementation, by the way, since it doesn't seem like that's intended from documentation :)

### DIFF
--- a/src/Options.cc
+++ b/src/Options.cc
@@ -424,10 +424,16 @@ Options parse_cmdline(int argc, char** argv) {
     opterr = 0;
 
     // getopt may permute the array, so need yet another array
-    auto zargs = std::make_unique<char*[]>(zeek_args.size());
+    //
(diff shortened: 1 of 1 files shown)


timwoj (approved):
> This is probably dependent on the libc implementation, by the way, since it doesn't seem like that's intended from documentation :)

It must be. Both macOS and glibc on ubuntu 24 don't seem to care about it and just report an error:

```
build% zeek -b test.zeek --debug
ERROR: Option --debug requires an argument.
Try 'zeek --help' for more information.
```

Either way, seems fine to me.

PR #4193, opened 2025-01-23
Fix errors from rst linting on the generated docs

timwoj:
This fixes a bunch of ReST style issues, mostly for consistency across all of the docs.

### DIFF
--- a/scripts/base/frameworks/input/main.zeek
+++ b/scripts/base/frameworks/input/main.zeek
@@ -24,10 +24,10 @@ export {
 		STREAM = 2
 	};
 
-	## The default input reader used. Defaults to `READER_ASCII`.
+	## The default input reader used. Defaults to :zeek:see:`Input::READER_ASCII`.
 	option default_reader = READER_ASCII;
 
-	## The default reader mode used. Defaults to `MANUAL`.
+	## The default reader mode used. Defaults to :zeek:see:`Input::MANUAL`.
 	option default_mode = MANUAL;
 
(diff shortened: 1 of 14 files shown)


scripts/base/frameworks/spicy/init-bare.zeek
```diff
@@ -23,7 +23,7 @@ export {
 # doc-options-end
 
 # doc-types-start
-    ## Result type for `Spicy::resource_usage()`. The values reflect resource
+    ## Result type for ``Spicy::resource_usage()``. The values reflect resource
```
awelzel, line 26: `:zeek:see:` ?

scripts/base/frameworks/input/main.zeek
```diff
@@ -24,10 +24,10 @@ export {
 		STREAM = 2
 	};
 
-	## The default input reader used. Defaults to `READER_ASCII`.
+	## The default input reader used. Defaults to ``READER_ASCII``.
```
awelzel, line 27: :zeek:see: ?

and below for `default_mode` ?

awelzel (approved):
Thanks!

scripts/base/frameworks/input/main.zeek
timwoj: It's always bothered me that Zeek puts the enum values into the module's namespace instead of one below that, but changing that would be a massive change. For example `Input::MANUAL` instead of `Input::Mode::MANUAL`.

PR #4194, opened 2025-01-24
Make types into constants

### LINKED ISSUE #4173: Script type expression support

evantypanski:
Closes #4173

I just decided to redo #4187 but making types a constant expression rather than its own expression. Using a separate PR so that discussion is available, though.

The real reason I changed this is because now ZAM doesn't need changes (from what I can tell)

I'll leave implementing a frontend component for `type` types for a later date

Note that `TypeExpr` technically has some niceties if we were to propagate. (There would be one less layer of indirection, you wouldn't have to look at its value to get the type so easier to do propagate the type down. And you can tell if it's a type from its expression tag. ¯\\\_(ツ)_/¯). When I started I wanted to do a bit more with type expressions, but it exploded a bit as I did, so this is fine too. I should've done this initially.


src/parse.y
```diff
@@ -1265,45 +1288,49 @@ type:
...
 				set_location(@1, @3);
-				$$ = new OpaqueType($3);
+				$$ = new FuncType({AdoptRef{}, $3}, nullptr, FUNC_FLAVOR_EVENT);
 				}
 
 	|	resolve_id
 			{
-			if ( ! $1 || ! ($$ = $1->IsType() ? $1->GetType().get() : nullptr) )
+			if ( $1 && $1->GetType()->Tag() == TYPE_TYPE )
+				$$ = $1->GetType()->AsTypeType()->GetType()->Ref();
```
evantypanski, line 1332: Note that we can remove this, then the only thing that this adds is types passing in to BIFs (so it lets you do `type_name(string)` but not `local ty = string;`).

That may be better if it's a `ConstExpr` since this line kind of conflates a dynamic type value with a statically known type

So I'm okay removing this just to make the new `ConstExpr` make sense

src/parse.y
```diff
@@ -765,6 +765,10 @@ expr:
 				reporter->Error("index slice assignment may not be used"
 				                      " in arbitrary expression contexts, only"
 				                      " as a statement");
+			// Type types cannot be reassigned because that could require parse-time
```
vpax, line 768: Can you do this instead by marking the identifier as `const`? That might be a bit cleaner/more comprehensive (but I haven't thought much about this so could have something wrong)

testing/btest/language/type-expr-assignment.zeek
```diff
@@ -0,0 +1,11 @@
+# @TEST-DOC: Ensure redefining a type expression ID is an error
+# @TEST-EXEC-FAIL: zeek -b %INPUT
+# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff .stderr
+
+event zeek_init()
+	{
+	local str = string;
+	str = count;
+	local my_string: str = "aoeu"; # This will still be a string
+    print my_string;
```
vpax, line 10: (micro-nit: indentation inconsistency)

vpax:
Really like the simplicity of this!

vpax:
> @vpax do you see any ZAM (or other) concerns with `ConstExpr` now including type vals?

I don't. If there are some, they should be fairly easy to resolve - introducing a new type of AST node that's always a bunch of work, adjusting use of a current node often doesn't take much.

src/parse.y
timwoj: > Note that we can remove this, then the only thing that this adds is types passing in to BIFs (so it lets you do type_name(string) but not local ty = string;).

I'm trying to think of when you would ever do this in a Zeek script, but C++ allows type-aliasing like this at the function level too with `using` statements. Not quite the same as assigning a type to a variable though.

evantypanski:
I just ditched all of the extra complexities, this now will basically only work with BIFs, you can't do anything with a variable assigned to a type.

The error messages for this being used incorrectly (like `local ty = string; local x: ty = "abc";`) are a little rough ("not a Zeek type (ty)"). If that's too much of a deal breaker I can add in a few extra checks, I just don't like it because they'll need to be in a few places and if that behavior changes it's not obvious where you'd have to update in code (namely for locals, globals, consts, and regular assignments...).

timwoj (approved):
I like how simple this ended up. We can always add the variable-assignment bits later if there's a use for them, but this solves the current need.

vpax:
FYI, I confirmed that the generated ZAM code makes sense, so looking good on that end

PR #4197, opened 2025-01-27
Pack some classes for better memory usages

timwoj:
Opening this as a draft PR to get benchmark numbers

- Analyzer: Reduce from 208 bytes to 192 bytes, remove one cache line
- EventGroup: Reduce from 104 bytes to 96 bytes
- Packet: Reduce from 188 bytes to 192 bytes, remove one cache line
- threading::Value: Reduce from 48 bytes to 40 bytes
- ConnTuple: push hole to the end of struct
- TCP_Reassembler: Reduce from 240 bytes to 232 bytes

### DIFF
--- a/src/Conn.h
+++ b/src/Conn.h
@@ -59,8 +59,8 @@ struct ConnTuple {
     IPAddr dst_addr;
     uint32_t src_port = 0;
     uint32_t dst_port = 0;
-    bool is_one_way = false; // if true, don't canonicalize order
     uint16_t proto = UNKNOWN_IP_PROTO;
+    bool is_one_way = false; // if true, don't canonicalize order
 };
 
 static inline int addr_port_canon_lt(const IPAddr& addr1, uint32_t p1, const IPAddr& addr2, uint32_t p2) {
--- a/src/analyzer/protocol/tcp/TCP_Reassembler.h
+++ b/src/analyzer/protocol/tcp/TCP_Reassembler.h
@@ -105,10 +105,10 @@ private:
     bool did_EOF;
     bool skip_deliveries;
 
-    uint64_t seq_to_skip;
-
-    bool in_delivery;
     analyzer::tcp::TCP_Flags flags;
+    bool in_delivery;
+
+    uint64_t seq_to_skip;
 
     FilePtr record_contents_file; // file on which to reassemble contents
 
--- a/src/threading/SerialTypes.h
+++ b/src/threading/SerialTypes.h
@@ -126,9 +126,8 @@ struct Field {
  * those Vals supported).
  */
 struct Value {
-    TypeTag type;         //! The type of the value.
-    TypeTag subtype;      //! Inner type for sets and vectors.
-    bool present = false; //! False for optional record fields that are not set.
+    TypeTag type;    //! The type of the value.
+    TypeTag subtype; //! Inner type for sets and vectors.
 
     struct set_t {
         zeek_int_t size;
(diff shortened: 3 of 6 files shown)


src/Conn.h
```diff
@@ -59,8 +59,8 @@ struct ConnTuple {
     IPAddr dst_addr;
     uint32_t src_port = 0;
     uint32_t dst_port = 0;
-    bool is_one_way = false; // if true, don't canonicalize order
     uint16_t proto = UNKNOWN_IP_PROTO;
+    bool is_one_way = false; // if true, don't canonicalize order
```
awelzel, line 63: If we wanted to play tricks, could probably also use a bit from proto and encode UNKNOWN_IP_PROTO as 2**15-1 instead. Just resolve unless this tickles you as interesting. I don't know if that's really worth it though.

src/analyzer/protocol/tcp/TCP_Reassembler.h
```diff
@@ -105,10 +105,10 @@ class TCP_Reassembler final : public Reassembler {
     bool did_EOF;
     bool skip_deliveries;
```
awelzel, line 106: Here, but also for Analyzer: Would bitfields for all those bools in these classes help to squeeze out a bit more? But then, even for 1mio concurrent TCP connections this would maybe save 4MB, maybe some more for connections with very many analyzers, but seems 10's of MB rather than 100's of them.

src/threading/SerialTypes.h
```diff
@@ -126,9 +126,8 @@ struct Field {
  * those Vals supported).
  */
 struct Value {
-    TypeTag type;         //! The type of the value.
-    TypeTag subtype;      //! Inner type for sets and vectors.
-    bool present = false; //! False for optional record fields that are not set.
+    TypeTag type;    //! The type of the value.
+    TypeTag subtype; //! Inner type for sets and vectors.
```
awelzel, line 130: TypeTag are both 4 bytes each due to being ints by dfeault, right? I'm not sure about the implications, but the following might save 6 bytes (minus padding - might make sense to move present up if that's done). We may have a few thousand threading::Vals flying around, so that might  be worth it.

```
diff --git a/src/Type.h b/src/Type.h
index e7dc3c52a2..fbd5b5b0b0 100644
--- a/src/Type.h
+++ b/src/Type.h
@@ -59,7 +59,7 @@ public:
 } // namespace detail
 
 // Zeek types.
-enum TypeTag {
+enum TypeTag : uint8_t {
```

awelzel (commented):
> Packet: Reduce from 188 bytes to 192 bytes, remove one cache line

Not following :-)

Comments mostly food for thought. Not sure it's worth it, happy to approve, too.

src/analyzer/protocol/tcp/TCP_Reassembler.h
timwoj: It'd likely lose some performance with accessing the bitfields vs the discrete bools. Like you said the memory savings probably aren't worth it at that point.

src/threading/SerialTypes.h
timwoj: `InternalTypeTag` being a `uint16_t` is strange also, since it's only 9 values. Let me poke at that.

timwoj: This one ends up changing a lot of other objects too, so I'm gonna make a note of it and leave it out for now. There's definitely space savings we could make across all of the `enum`s we have in the code. I'll take a look at it more deeply the next time I do this.

awelzel (approved):
:+1: 

I'm very curious by the `pahole` output and whether rearranging fields in Analyzer that are used together during `DeliverStream()` to fall into the same cacheline would make a visible difference. E.g., moving less hot fields (signature, conn, id, tag, prevented?) elsewhere.

PR #4199, opened 2025-01-29
Topic/awelzel/4136 cluster websocket support

awelzel:
Opening for a light CI run.


src/cluster/websocket/WebSocket.h
```diff
@@ -0,0 +1,301 @@
...
+     * Main processing function of the dispatcher.
+     *
+     * This runs on Zeek's main thread.
+     */
+    void Process(WebSocketEvent&& event);
+
+    void Process(WebSocketOpen& open);
+    void Process(WebSocketSubscribeFinished& fin);
+    void Process(WebSocketMessage& msg);
+    void Process(WebSocketClose& close);
```
bbannier, line 207: Not sure why we need mutable refs, maybe pass like we pass `WebsocketEvent` above?

```suggestion
    void Process(WebSocketOpen&& open);
    void Process(WebSocketSubscribeFinished&& fin);
    void Process(WebSocketMessage&& msg);
    void Process(WebSocketClose&& close);
```

src/cluster/websocket/WebSocket.cc
```diff
@@ -0,0 +1,480 @@
...
+    // Process any queued messages now.
+    for ( auto& msg : entry.queue ) {
+        assert(entry.msg_count > 1);
+        Process(msg);
+    }
+}
+
+void WebSocketEventDispatcher::HandleSubscriptions(WebSocketClientEntry& entry, std::string buf) {
+    rapidjson::Document doc;
+    doc.Parse(buf.c_str());
```
bbannier, line 351: Nit: No need to pass a string copy.

```suggestion
void WebSocketEventDispatcher::HandleSubscriptions(WebSocketClientEntry& entry, std::string_view buf) {
    rapidjson::Document doc;
    doc.Parse(buf.data(), buf.size());
```

src/cluster/websocket/WebSocket.cc
```diff
@@ -0,0 +1,480 @@
...
+
+    for ( const auto& subscription : subscriptions ) {
+        if ( ! entry.backend->Subscribe(subscription, cb) ) {
+            zeek::reporter->Error("Subscribe for WebSocket client failed!");
+            QueueReply(WebSocketCloseReply{entry.wsc, 1011, "Could not subscribe. Something bad happened!"});
+        }
+    }
+}
+
+void WebSocketEventDispatcher::HandleEvent(WebSocketClientEntry& entry, std::string buf) {
```
bbannier, line 392:
```suggestion
void WebSocketEventDispatcher::HandleEvent(WebSocketClientEntry& entry, std::string_view buf) {
```

src/cluster/websocket/WebSocket.cc
```diff
@@ -0,0 +1,480 @@
...
+        reporter->Error("WebSocket message from non-existing WebSocket client %s", id.c_str());
+        return;
+    }
+
+    auto& entry = it->second;
+    const auto& wsc = entry.wsc;
+    entry.msg_count++;
+
+    WS_DEBUG("Message %zu size=%zu from %s (%s:%d) backend=%p", entry.msg_count, msg.msg.size(), wsc->getId().c_str(),
+             wsc->getRemoteIp().c_str(), wsc->getRemotePort(), entry.backend.get());
```
bbannier, line 463:
```suggestion
    WS_DEBUG("Message %llu size=%zu from %s (%s:%d) backend=%p", entry.msg_count, msg.msg.size(), wsc->getId().c_str(),
             wsc->getRemoteIp().c_str(), wsc->getRemotePort(), entry.backend.get());
```

src/cluster/websocket/WebSocket.h
```diff
@@ -0,0 +1,301 @@
...
+
+    std::unique_ptr<WebSocketEventDispatcher> dispatcher;
+};
+
+
+/**
+ * TLS configuration for a WebSocket server.
+ */
+struct TLSOptions {
+    bool enable = true;
```
bbannier, line 271: See related comment on corresponding Zeek record.

src/cluster/websocket/WebSocket.h
```diff
@@ -0,0 +1,301 @@
...
+    bool enable_peer_verification = false;
+    std::string cert_file;
+    std::string key_file;
+    std::string ca_file;
+    std::string ciphers;
+};
+
+struct ServerOptions {
+    std::string host;
+    int port = -1;
```
bbannier, line 281: Nit: Probably by design, but `-1` is not a good default port. Could this be e.g., unsigned and default to `0` (typical value for: pick anything)?

src/cluster/websocket/WebSocket.h
```diff
@@ -0,0 +1,301 @@
...
+    int ping_interval = 5;
```
bbannier, line 282: What is the time unit?

src/cluster/websocket/WebSocket.h
```diff
@@ -0,0 +1,301 @@
...
+
+    /**
+     * @return true if all subscriptions have an active status.
+     */
+    bool AllSubscriptionsActive() const;
+
+private:
+    bool acked = false;
+    std::map<std::string, bool> subscriptions_state;
+    std::vector<std::string> subscriptions;
```
bbannier, line 119: We only store this value to (marginally?) speed up `GetSubscriptions`. I'd suggest to either just compute it on the fly there, or if really needed introduce a small helper map type which has a constant `keys` function which would internally make sure `subscription_state` and `subscriptions` are consistent (and which would return some set).

My gut feeling is that this is not critical for performance, and if users have a massive number of subscriptions we would first see slowdown elsewhere.

bbannier (commented):
Overall this makes a lot of sense to me.

I was at first wondering whether we really needed to model the different ws message types, but it seems to make the handling easier to follow, so I think I get that now.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -306,7 +331,7 @@ bool ZeroMQBackend::DoPublishLogWrites(const logging::detail::LogWriteHeader& he
     // * The serialized log write itself.
     std::array<zmq::const_buffer, 4> parts = {
         zmq::const_buffer{message_type.data(), message_type.size()},
-        zmq::const_buffer(my_node_id.data(), my_node_id.size()),
+        zmq::const_buffer(NodeId().data(), NodeId().size()),
```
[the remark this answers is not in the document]
awelzel: Yeah, good call.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -92,32 +88,33 @@ void ZeroMQBackend::DoInitPostScript() {
...
         zeek::id::find_val<zeek::StringVal>("Cluster::Backend::ZeroMQ::connect_xsub_endpoint")->ToStdString();
+    connect_xpub_nodrop =
+        zeek::id::find_val<zeek::BoolVal>("Cluster::Backend::ZeroMQ::connect_xpub_nodrop")->AsBool() ? 1 : 0;
     listen_log_endpoint =
         zeek::id::find_val<zeek::StringVal>("Cluster::Backend::ZeroMQ::listen_log_endpoint")->ToStdString();
+
+    linger_ms = static_cast<int>(zeek::id::find_val<zeek::IntVal>("Cluster::Backend::ZeroMQ::linger_ms")->AsInt());
     poll_max_messages = zeek::id::find_val<zeek::CountVal>("Cluster::Backend::ZeroMQ::poll_max_messages")->Get();
     debug_flags = zeek::id::find_val<zeek::CountVal>("Cluster::Backend::ZeroMQ::debug_flags")->Get();
+    proxy_io_threads = zeek::id::find_val<zeek::CountVal>("Cluster::Backend::ZeroMQ::proxy_io_threads")->Get();
```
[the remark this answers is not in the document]
awelzel: The ZMQ API is using `int` - I'm going to cast this one. If someone configures `> 2^32` IO threads, that's on them :-)

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -327,11 +327,58 @@ export {
 		## The arguments for the event.
 		args: vector of any;
 	};
+
+	## The TLS options for a WebSocket server.
+	type WebSocketTLSOptions: record {
+		## Whether TLS is enabled at all.
```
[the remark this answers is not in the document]
awelzel: sound good to me!

src/cluster/Manager.cc
```diff
@@ -27,3 +42,21 @@ std::unique_ptr<LogSerializer> Manager::InstantiateLogSerializer(const zeek::Enu
     const LogSerializerComponent* c = LogSerializers().Lookup(tag);
     return c ? c->Factory()() : nullptr;
 }
+
+bool Manager::ListenWebSocket(const std::string& addr, int port, const cluster::websocket::TLSOptions& tls_options) {
+    WebSocketServerKey key{addr, port};
+
+    if ( websocket_servers.count(key) != 0 ) {
+        zeek::emit_builtin_error(zeek::util::fmt("Already listening on %s:%d", addr.c_str(), port));
```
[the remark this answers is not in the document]
awelzel: Done.

src/cluster/websocket/WebSocket.h
awelzel: That change results in a stack overflow. Happy to debug with you :-)

src/cluster/websocket/WebSocket.cc
awelzel: Went with PRIu64.

src/cluster/OnLoop.h
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * There's currently no explicit way to transfer a result back. If this is needed,
+ * have the queueing thread block on a semaphore or condition variable and update
+ * it from Process().
+ *
+ * Note that QueueForProcessing() puts the queueing thread to sleep if there's
+ * too many items in the queue.
+ */
+template<class Proc, class Work>
+class OnLoopProcess : public zeek::iosource::IOSource {
```
ckreibich, line 34: If I understand it correctly this is basically a message queue from threads into Zeek. I really like that abstraction, but what's the intuition for "OnLoop"? Is it that this feeds into Zeek's I/O loop, from some externally established thread, or that it runs in the I/O loop of such threads?

src/cluster/websocket/WebSocket.cc
```diff
@@ -0,0 +1,487 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+// Implement Broker's WebSocket client handling in Zeek.
```
ckreibich, line 3: Ah — this has an important detail! At this stage you're aiming to replicate Broker's current functionality but within Zeek. I kept wondering about this while reading.

I'm not sure if this is the right spot, but somewhere it'd be nice to have documentation of how the running server(s) manifest in the Zeek process. If I list the Zeek process's threads, what do the webserver ones look like?

src/zeek-setup.cc
```diff
@@ -378,6 +378,7 @@ static void terminate_zeek() {
     notifier::detail::registry.Terminate();
     log_mgr->Terminate();
     input_mgr->Terminate();
+    cluster::manager->Terminate();
```
ckreibich, line 381: This is interesting — now that we have this call I wondered whether the active backend and the cluster manager itself could shut down via one call. But the manager doesn't actually know the _active_ backend, am I seeing that correctly?

src/cluster/websocket/WebSocket.cc
```diff
@@ -0,0 +1,487 @@
...
+// XXX: How is this going to work with class broker? With
+// ZeroMQ, each WebSocket client has its own XPUB/XSUB
+// connectivity to a central broker and similarly with NATS.
+// But with broker we need to do something different.
+// Maybe connect to the local endpoint.
+//
+// We cannot actually instantiate a Broker backend :-(
+//
+// We could also have InitPostScript() recognize Broker
+// and start its internal server instead.
```
ckreibich, line 261: I'm not following this, let's talk this over in person. I thought that in the next iteration of the refactoring Broker becomes a backend like the others, but I'm probably just missing the point.

ckreibich (commented):
Very nice, and nicely incrementally readable via the individual commits — thanks! Tests look great too. I mostly just have a few questions in the comments.

I am getting lost regarding the overall architecture though ... I could use a sketch that explains how event propagation works across local events, backends, and the webserver(s). I really hope I didn't miss a previous doc or some such? :-)

For example: 
- How does the websocket handshake work? (Fine to point to Broker's docs if it's the same right now)
- The local Zeek generates an event on a topic that a websocket client and another node have subscribed to — how does that event get there?
- A websocket client generates an event on the connection it has with a Zeek node, and another node has subscribed to it — how does it get there? Etc.

I _think_ I get the role of `EventHandlingStrategy` in these but it'd be nice to see how it all fits together. Let's talk it over in person and then figure out where to put this.

I'd probably get the NEWS entry started at this point even if we're going to iterate in the near future.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -55,6 +55,13 @@ export {
 	## By default, this is set to ``T`` on the manager and ``F`` elsewhere.
 	const run_proxy_thread: bool = F &redef;
 
+	## How many IO threads to configure for the ZeroMQ context that
+	## acts as a central broker.
+
+	## See ZeroMQ's `ZMQ_IO_THREADS documentation <http://api.zeromq.org/4-2:zmq-ctx-set#toc4>`_
```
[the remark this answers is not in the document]
awelzel: I've linked the doc. I have to test implications, but it seemed good to have it configurable.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -446,6 +445,9 @@ void ZeroMQBackend::Run() {
                     } while ( ! result );
                 }
             }
+            else {
+                ZEROMQ_THREAD_PRINTF("inproc: error: expected 1 or 4 parts, have %zu!\n", msg.size());
```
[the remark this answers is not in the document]
awelzel: stderr, yes. Mostly because reporter wouldn't be thread safe and this really should not happen :crossed_fingers:

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -327,11 +327,62 @@ export {
...
+	## WebSocket server options to pass to :zeek:see:`Cluster::listen_websocket`.
+	type WebSocketServerOptions: record {
+		## The host address to listen on.
+		listen_host: string;
+		## The port the WebSocket server is supposed to listen on.
+		listen_port: port;
+		## The TLS options used for this WebSocket server. By default,
+		## TLS is disabled. See also :zeek:see:`Cluster::WebSocketTLSOptions`.
+		tls_options: WebSocketTLSOptions &default=WebSocketTLSOptions();
+	};
```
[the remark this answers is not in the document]
awelzel: The parsing of WebSocketClient messages happens in: `WebSocketEventDispatcher::HandleEvent()`. That currently uses: `broker::format::json::v1::decode()`.

I wouldn't necessarily configure the format per server. WebSocket clients going to `/v1/messages/json` will speak v1 format, clients going to `/v2` would speak a newer format, both should be supported by the implementation. Maybe disabling support for `v1` could be an option, but that could also be a global policy.

src/cluster/websocket/WebSocket.cc
awelzel: > Ah — this has an important detail! At this stage you're aiming to replicate Broker's current functionality but within Zeek. I kept wondering about this while reading.

Hmm, yes, that was the plan, so `Broker::listen_websocket()` can be replaced with `Cluster::listen_websocket()` without clients needing a change.

> I'm not sure if this is the right spot, but somewhere it'd be nice to have documentation of how the running server(s) manifest in the Zeek process. If I list the Zeek process's threads, what do the webserver ones look like?

Hmm, this is mostly dictated by the IXWebSockets library, there'll be at least one server thread and one thread per client. Some threads used to send replies to clients and depending on the cluster backend more threads :-)

src/cluster/websocket/WebSocket.cc
awelzel: The `InitPostScript()` part was about switching `Cluster::listen_websocket()` to be `Broker::listen_websocket()` instead, but I should just be bold enough to make the native Zeek level WebSocket work properly.

This would then also allows to change the JSON format for WebSocket clients, even if Broker is used as a Backend.

In summary, after this PR, `Cluster::listen_websocket()` with broker as cluster backend will not be functional, but there's a sketch of a way forward.

src/cluster/OnLoop.h
awelzel: > Is it that this feeds into Zeek's I/O loop, from some externally established thread,

This. And particularly running code on the main thread.

The name is not great, but it stuck.

I also had chatted with @bbannier and he was a bit afraid that everyone builds their own things to this. I acknowledge this. We settled somewhat on keeping an eye out if this happens elsewhere and unify or come up with something better. E.g. @Neverlord's recent logging changes have a similar class, to run logging in the main-thread, but the tasks come from within broker/caf threads. That's why it's in `src/cluster/` instead of somewhere more common.

src/cluster/OnLoop.h
```diff
@@ -0,0 +1,175 @@
...
+            flare.Fire();
+
+        --queuers;
+    }
+
+private:
+    zeek::detail::Flare flare;
+    std::mutex mtx;
+    std::list<Work> queue;
+    size_t max_queue_size = 10;
```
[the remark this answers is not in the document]
awelzel: Removed.

src/cluster/OnLoop.h
```diff
@@ -0,0 +1,175 @@
...
+    double GetNextTimeout() override { return -1; };
+
+    /**
+     * Queue the given Work item to be processed on Zeek's main thread.
+     *
+     * If there's too many items in the queue, this method sleeps using
+     * std::this_thread::sleep() for the *block_duration* passed to the
+     * constructor.
+     *
+     * Don't call this from the main thread, you might end-up deadlocking!
```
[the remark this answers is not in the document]
awelzel: Added via abort(), thanks!

src/cluster/websocket/WebSocket.cc
ckreibich: There's a blog post in walking through all of Zeek's threads at some point. :-)

src/cluster/websocket/WebSocket.cc
ckreibich: Thanks for the explanation and walking me through it in person today, I get it now.

ckreibich (approved):
I think we should merge this at this point — best to be able to start tinkering and iterating at this point. Thanks Arne! :+1:

timwoj (approved):
That additional fix for macOS python stuff on CI looks good to me.

Issue #4082, opened 2024-12-04
Proposal - change DPD behavior on analyzer error after protocol violation


0xxon:
When there is a protocol violation during a connection, currently a detected protocol will be removed from the `service` field of `conn.log`. This holds true, even when the protocol in question already was confirmed. This also holds true when we are really far into a connection - imagine a HTTP connection that has gigabytes of exchanged data; if there is an error at the end of it, the service field will be removed.

This means that, currently, we can end up in situations where a protocol is confirmed & a protocol log is written - but there is no trace of the detected protocol in `conn.log`; one has to look at `analyzer.log`, `dpd.log`, or the respective protocol log files to figure out that there was some additional potential protocol activity in an affected connection.

This behavior was detected during the development of the Spicy based SSL analyzers; there are some testcases in our test suite that have protocol violation at the end of a short connection. Binpac does not raise errors for protocol violations due to a connection ending, which is why this behavior was probably not noticed and potentially not as important so far.

Some members of the development team had a discussion about this in a recent meeting. It was generally felt that the current behavior is probably, at the very least, unintuitive and that we should change it.

I propose to change the behavior to always list a detected protocol in the `conn.log` `service`-field, once it has been confirmed.

I ran out entire test suite with a change that implements this; there is a relatively low number of test changes. On a first glance, most of these are expected, and caused by tests for protocol error cases.

PR #4200, opened 2025-01-29
DPD: restore original behavior, tweak IRC analyzer, alawys log confirmed analyzers, add failed service tracking

0xxon:
This PR contains a set of changes which tries to address #4082. Specifically, these changes are:

## Restore old DPD analyzer removal behavior

From my understanding, and looking back at old versions of Zeek (pre 2.0), the desired behavior of DPD is as follows:

When an analyzer that has not been confirmed raises a protocol violation, it is immediately removed from the connection. This allows the analyzer to be in a "tasting" phase at the beginning of the connection, and to error out quickly once it realizes that it was attached to a connection not containing the desired protocol.

In b4b990cfb56ccb60846f39abe8c9c33a1ebdfb64, a bug was introduced that caused non-confirmed analyzers to never be removed. This was recognized as a problem years later, and in #477 a fix was applied that made non-confirmed analyzers fail after 5 errors. From my memory this was discovered due to performance issues. This fix is a bit of a weird band-aid; from looking back there never was a lot of discussion about this change, and the values were chosen pretty arbitrarily.

This PR restores the original behavior and removes the DPD::max_violations setting completely. There are minimal changes to baselines due to this change; I will explain these below.

## Always list confirmed protocols in conn.log

At the moment, protocols are removed from conn.log when a violation occurs, even after they have been confrmed. This seems weird, as a protocol log probably exists - and one should only raise an analyzer confirmation when one is reasonably sure about the protocol. It seems more reasonable to document the fact that the analyzer was confirmed, and keep it in conn.log.

This PR changes this behavior to always add confirmed analyzers to conn.log. This change leads to minimal, reasonable changes to our test baseline - especially tests that test violations now have the protocol in conn.log.

There is one exception to this, which is the IRC analyzer.

## IRC analyzer protocol confirmation behavior fix

The IRC analyzer, so far, was extremely liberal in raising a protocol confirmation - this was raised essentially immediately when a connection was established.

This PR changes this behavior and looks for valid commands before confirming a connection. There are no baseline diffs due to this behavior change.

## Make services field ordered

This PR changes the services field in conn.log to be ordered. Services are listed in the order in which they are confirmed. This makes it possible to argue about protocol composition; e.g. mysql with an StartTLS upgrade will now be always logged as mysql,ssl. Another example from the test suites is http,websocket,ssh.

## New options for tracking failed services

This PR adds two possibilities to track confirmed services that later failed in conn.log. 

Setting ``DPD::track_removed_services_in_connection`` to true causes these services to be logged in conn.log with a "-" in front. This is in addition to the confirmed service. So, if a connection has an SSL analyzer that attaches and then fails, it is logged as "ssl,-ssl"

Alternatively, setting the option ``Conn::log_violating_services`` to true causes a new field to be added to conn.log, which contains the list of violating services.

## Test suite changes

The changes in the test suites are generally minor. In some cases, there are now less analyzer.log entries, because a protocol analyzer is now removed earlier.

In one specific case, there are now more analyzer.log entries - connections where the originator and the responder speak two different protocols. Consider the case of an SSL client connecting to a HTTP server on port 80. This automatically causes the HTTP analyzer to attach. When the client hello is sent, the SSL analyzer also attaches; however, the HTTP analyzer raises a violation and fails.

In an observed case, the HTTP server will respond with a "400, Bad request" to the client hello. This causes the HTTP dpd signature to match, which causes a replay of the connection - and another dpd violation. The HTTP response is logged correctly.

This is not optimal, but I think is ok for the moment. I wonder if it might make sense to partially fail analyzers - and let each direction of a connection fail independently from the other direction; however, this would be a pretty major change.


scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -32,13 +32,16 @@ export {
 	## Ignore violations which go this many bytes into the connection.
 	## Set to 0 to never ignore protocol violations.
 	option ignore_violations_after = 10 * 1024;
+
+	## Add removed services to conn.log, with a - in front of them.
+	option track_removed_services_in_connection = F;
```
awelzel, line 37: Naming and debatable, but wonder if `Conn::track_service_removals` is shorter and just as descriptive compared to `DPD::track_removed_services_in_connection`, even if it's in the other module (which still makes some sense though as that's where the log is created). Possibly also easier to find, given one would likely look into conn/main.zeek for logging of conn.log. A comment that the actual logic happens in `dpd.zeek` probably would help.

src/analyzer/protocol/irc/IRC.h
```diff
@@ -75,6 +75,14 @@ class IRC_Analyzer final : public analyzer::tcp::TCP_ApplicationAnalyzer {
      */
     std::vector<std::string> SplitWords(const std::string& input, char split);
 
+    /**
+     * Checks if a passed string is a valid command for an IRC client.
+     *
+     * \param command command to check
+     * \return true if command is valid
+     */
+    static bool IsValidClientCommand(const std::string& command);
```
awelzel, line 84: No strong opinion and the C++ gurus might disagree, but if it's private static, maybe can just make it a static/anonymous namespace function in `IRC.cc`?

awelzel (commented):
This looks nice and surprised about the little baseline changes in fact :+1: 

I have a fairly strong opinion about moving the `log_violating_services const &redef` a policy script, the rest is mostly nits / take-a-look-what-you-think comments.

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -127,6 +130,18 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
...
+    if ( ! confirmed && orig )
+        // confirm if we see a known IRC command
+        if ( command == "ADMIN" || command == "AWAY" || command == "CNOTICE" || command == "CPRIVMSG" ||
+             command == "INFO" || command == "INVITE" || command == "JOIN" || command == "KICK" || command == "LIST" ||
+             command == "MOVE" || command == "NAMES" || command == "NICK" || command == "NOTICE" || command == "PASS" ||
+             command == "PART" || command == "PING" || command == "PRIVMSG" || command == "SERVER" ||
+             command == "SETNAME" || command == "TOPIC" || command == "USER" || command == "WHOIS" ||
+             command == "WHOWAS" || command == "STARTTLS" || command == "PONG" ) {
+            AnalyzerConfirmation();
+            confirmed = true;
```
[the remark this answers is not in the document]
0xxon: Ah, fair.

testing/btest/Baseline/scripts.base.protocols.quic.chromium/conn.log.cut
```diff
@@ -1,3 +1,3 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
 ts	uid	history	service
-XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	Dd	quic,ssl
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	Dd	ssl,quic
```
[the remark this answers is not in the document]
0xxon: Yup, we could revisit that. If you think it's worth it I can just create a bug so that we don't forget about this.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -32,13 +32,16 @@ export {
...
 }
 
 redef record connection += {
 	dpd: Info &optional;
 	## The set of services (analyzers) for which Zeek has observed a
 	## violation after the same service had previously been confirmed.
-	service_violation: set[string] &default=set();
+	service_violation: set[string] &optional &default=set() &ordered;
```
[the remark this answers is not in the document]
0xxon: Yup, it itotally is

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -130,17 +143,10 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
...
-             command == "PART" || command == "PING" || command == "PRIVMSG" || command == "SERVER" ||
-             command == "SETNAME" || command == "TOPIC" || command == "USER" || command == "WHOIS" ||
-             command == "WHOWAS" || command == "STARTTLS" || command == "PONG" ) {
-            AnalyzerConfirmation();
-            confirmed = true;
-        }
+    if ( ! confirmed && orig && IsValidClientCommand(command) ) {
+        AnalyzerConfirmation();
+        confirmed = true;
+    }
```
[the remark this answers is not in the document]
0xxon: Unless you have strong feelings - I don't feel like polluting weird.log with that. I am sure that there is non-standard stuff - but as this is only thought to catch IRC connections reaonably, and these are the by far most likely commands, it probably doesn't matter much.

If we do care, we still can look through irc.log.

src/analyzer/protocol/irc/IRC.h
0xxon: Yup, I could :)

This is purely for visibility, and there is no drawback that I am aware of to do it this way.

src/analyzer/protocol/irc/IRC.cc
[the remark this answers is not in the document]
awelzel: No strong opinion.  I'm not sure there'll be something in `irc.log` by default, but maybe one could get to it via `irc_message/irc_request` . If you feel this is polluting weird.log, let me resolve the comment.

It's also ... IRC.

testing/btest/Baseline/scripts.base.protocols.quic.chromium/conn.log.cut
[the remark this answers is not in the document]
awelzel: This PR adds `&ordered` actually does, so this reversal is "real" :-) @0xxon , issue would be perfect, so maybe someone else will look at it, rather than it being in the back of my mind.

JustinAzoff:
> From my memory this was discovered due to performance issues.

if it helps, yes.. it was a performance issue.  I was working on a sensor that was seeing a lot of traffic misidentified as DTLS, and the DTLS analyzer was raising an analyzer violation event or whatever it was for every single packet.  The thing that made it super hard to track down was that the dpd log was suppressing the duplicates errors, so there wasn't an outward facing indication that this was the problem.

testing/btest/Baseline/scripts.base.protocols.quic.chromium/conn.log.cut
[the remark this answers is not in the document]
0xxon: Will create an issue. And yes, we do introduce the ordering, exactly like you invision it, @JustinAzoff - that's part of this PR.

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -5,6 +5,7 @@
 #include "zeek/analyzer/protocol/irc/IRC.h"
```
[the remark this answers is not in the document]
0xxon: Agreed.

scripts/base/protocols/conn/main.zeek
```diff
@@ -32,6 +38,11 @@ export {
 		## colons. Protocols listed are in the order in which they are
 		## confirmed.
 		service:      string          &log &optional;
+		## List of protocols in a connection that raised protocol violations
+		## causing the analyzer to be removed.
+		## Protocols are listed in order that they were removed.
+		## Only present if ``log_violating_services`` is true.
+		service_violation: vector of string  &log &optional;
```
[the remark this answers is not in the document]
0xxon: I will refactor the service_violation addition to a policy script.

scripts/base/frameworks/analyzer/dpd.zeek
0xxon: I would like for this to remain in this module, so that dpd.zeek does not have to include conn.zeek.

I am happy to add a comment to conn/main.zeek. Is that good enough for you?

0xxon:
> > From my memory this was discovered due to performance issues.
> 
> if it helps, yes.. it was a performance issue.

Thank you for confirming :)

scripts/base/frameworks/analyzer/dpd.zeek
awelzel: Including `conn/main.zeek` isn't great. You could open the `Conn` module within `dpd.zeek` just to introduce the option.

I find the current name awkward and better fitting in `Conn` given the potential of this being a non-niche/interesting/popular option (I agree with you that default `T` could be considered) . But, I'm easily stubborn, so leaving it up to you or someone else chiming in - resolving :-)

testing/btest/Baseline/scripts.base.protocols.quic.chromium/conn.log.cut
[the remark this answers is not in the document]
0xxon: Filed as https://github.com/zeek/zeek/issues/4201

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -26,15 +26,6 @@ export {
...
 	};
 
-	## Ongoing DPD state tracking information.
-	type State: record {
-		## Current number of protocol violations seen per analyzer instance.
-		violations: table[count] of count;
-	};
-
-	## Number of protocol violations to tolerate before disabling an analyzer.
-	option max_violations: table[Analyzer::Tag] of count = table() &default = 5;
```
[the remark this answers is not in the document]
0xxon: I will add max_violations back. I think if you rely on dpd$state, you might just want to fail :)

scripts/base/frameworks/analyzer/dpd.zeek
0xxon: Using the conn namespace in this is a fun little hack, but I am not entirely sure about that :)

If this option actually becomes popular, we could just flup the default, and no one ever needs to use it.

0xxon:
I think the current state addresses all PR feedback so far :)

scripts/base/protocols/conn/main.zeek
```diff
@@ -27,8 +27,10 @@ export {
 		id:           conn_id         &log;
 		## The transport layer protocol of the connection.
 		proto:        transport_proto &log;
-		## An identification of an application protocol being sent over
-		## the connection.
+		## The identification of the application protocol(s) being sent over
+		## the connection. Can list more than one protocol separated with
+		## colons. Protocols listed are in the order in which they are
```
awelzel, line 32: `s/colons/commas/`.

Maybe a bit shorter: `The confirmed applications protocol(s) of the connection`. Multiple protocols are listed in the order in which they are confirmed, separated by commas.

I wonder if the order is necessary to call out - it not being ordered hadn't been called out and is arguably less obvious.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -120,7 +120,16 @@ event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationI
 	if ( ignore_violations_after > 0 && size > ignore_violations_after )
 		return;
 
-	disable_analyzer(c$id, aid, F);
+	local disabled = disable_analyzer(c$id, aid, F);
+
+	# add "-service" to the list of services on removal due to violation, if analyzer was confirmed before
+	if ( track_removed_services_in_connection && disabled && Analyzer::name(atype) in c$service )
+		{
+		local rname = fmt("-%s", Analyzer::name(atype));
```
awelzel, line 128: `cat("-", name(atype))` or event `"-" + ...` is likely faster than the format string machinery if you wanted to replace that.

scripts/policy/protocols/conn/failed-services.zeek
```diff
@@ -0,0 +1,27 @@
...
+module Conn;
+
+redef record Conn::Info += {
+	## List of protocols in a connection that raised protocol violations
+	## causing the analyzer to be removed.
+	## Protocols are listed in order that they were removed.
+	service_violation: vector of string  &log &optional;
+};
+
+# Not using connection removal hook, as this has to run for every connection.
```
awelzel, line 17: Fine to leave as is given it's a policy script.

The optimization would be to handle analyzer violations with a low priority and if `c$service_violation` is non empty, register the removal hook.

`connection_state_remove()` will also run on any syn/rst/syn-ack/rst/... connections or unknown data connections for which we even found a protocol.

scripts/test-all-policy.zeek
```diff
@@ -98,6 +98,7 @@
 @load misc/unknown-protocols.zeek
 @load protocols/conn/community-id-logging.zeek
 @load protocols/conn/disable-unknown-ip-proto-support.zeek
+@load protocols/conn/failed-services.zeek
```
awelzel, line 101: OCD: To adhere to the existing extension scripts: `service-violation-logging.zeek`? 

```
community-id-logging.zeek
vlan-logging.zeek
mac-logging.zeek
ip-proto-name-logging.zeek
```
:face_holding_back_tears:

awelzel (approved):
Only nits, thanks for addressing everything.

I'm approving, but think @rsmmr and/or @ckreibich should definitely look this through, too.

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -127,6 +142,10 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
     // Extract parameters.
     string params = myline;
 
+    if ( ! AnalyzerConfirmed() && orig && IsValidClientCommand(command) ) {
```
ckreibich, line 145: There's logic [further down](https://github.com/zeek/zeek/blob/master/src/analyzer/protocol/irc/IRC.cc#L493-L502) in this file that looks at the command length to determine validity. Is that obsolete now that we have an explicit command list?

scripts/base/protocols/conn/main.zeek
ckreibich: I agree here that shorter would be better — perhaps "A comma-separated list of confirmed protocol(s). With :zeek:see:`DPD::track_removed_services_in_connection`, the list includes the same protocols prefixed with "-" to record that Zeek dropped them due to parsing violations."

ckreibich (changes requested):
This is awesome, I just have a few minor comments/questions. Also
- I think one can't comment on the commit messages, so let me do that here: 2f712c3c has a typo, "This introduces ian options"
- Given the significance and subtlety, could you word in the `NEWS` update right away?
- I'm going to tag this as needing docs since we should cover this minimally in the section in the conn.log, and/or ideally a new one on the whole analyzer framework.

bbannier:
> * Given the significance and subtlety, could you word in the `NEWS` update right away?
> 
> * I'm going to tag this as needing docs since we should cover this minimally in the section in the conn.log, and/or ideally a new one on the whole analyzer framework.

I 100% concur. Like I mentioned in our discussion offline, this should work for a majority of cases, but can be pretty subtle when things are a little weird. I think we should make sure to communicate that

1. Analyzers are presented as an ordered list, but in reality analyzers form a tree. If multiple analyzers work on identical data the order of `services` might be weird.
2. We implicitly rely on parent analyzers confirming before their children. While we can control this to some extent for in-tree analyzers it does not compose nicely (e.g., a parent might be unaware on when their children could confirm). The only way to reliably work with this would be to hold back sending out any data until one can confirm (at least true for forwarding into other analyzers, potentially also applies to events).
3. `connection$service` is not fully controlled by us, so results can be inconsistent if it is mutated not by us.

I checked whether (2) could be taken care off by making sure we maintain the tree order when inserting new services in `DPD::analyzer_confirmation_info`. Unfortunately the parent analyzer is not available from scriptland, but at least in theory we could move the implementation to C++ so we can ensure that parent analyzers are always ordered before their children in `service` (modulo misuse a la (3)).

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -42,6 +43,18 @@ inline void IRC_Analyzer::SkipLeadingWhitespace(string& str) {
         str = str.substr(first_char);
 }
 
+bool IRC_Analyzer::IsValidClientCommand(const std::string& command) {
+    static const std::unordered_set<std::string_view> validCommands =
+        {"ADMIN",   "AWAY",     "CNOTICE", "CPRIVMSG", "CONNECT", "DIE",    "ENCAP",   "ERROR",    "INFO",
```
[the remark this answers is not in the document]
0xxon: That is correct.

There are a couple of more commands; they are extremely esoteric. Most IRC sessions will start with a NICK command.

This also does not prevent parsing in any way, just confirmation.

src/analyzer/protocol/irc/IRC.cc
0xxon: Not really - the list is only used for confirmation to be minimally invasive.

0xxon:
> > * Given the significance and subtlety, could you word in the `NEWS` update right away?
> > * I'm going to tag this as needing docs since we should cover this minimally in the section in the conn.log, and/or ideally a new one on the whole analyzer framework.
> 
> I 100% concur. Like I mentioned in our discussion offline, this should work for a majority of cases, but can be pretty subtle when things are a little weird. I think we should make sure to communicate that

If you think this is a problem/the semantics are too subtle, I am also fine just not documenting the fact that the service field is ordered, but instead still claiming that it does not have a reliable order.

That way if one knows how Zeek works internally, one can take this as an indication of what happened, and otherwise things are still not any worse than before.

awelzel:
I wonder if there's a disconnect on the subtlety around the changes in this PR:

* confirmed-then-violating analyzers are not removed from c$service anymore: Pretty big, could have even argued to make this opt-out for a few releases and see if someone gets upset by the new default. But @0xxon's arguments that the current state was never intended seems good enough to move this forward.

* c$service is now insert ordered: IMO, really really useful, but not overly concerning. Previously there was no guaranteed order. Now it's in confirmation order. Weird/subtle now becomes reliably visible (e.g. #4201). What's controversial?

> If you think this is a problem/the semantics are too subtle, I am also fine just not documenting the fact that the service field is ordered, but instead still claiming that it does not have a reliable order.

To me it seems fine to define `service` as being in the order as the analyzer_confirmation events within an analyzer tree for a connection happen. Maybe documenting that the order is not guaranteed to be stable across Zeek versions (we fix or we break things). Personally, it's not clear to me why we'd try to do something clever based on the tree-order of analyzers. This is just as subtle if your expectation is event order (e.g. looking at dump-events confirmation events).

To me, the bigger issue is that if you have something like HTTP->WebSocket->HTTP, you do not see the second HTTP being confirmed within history. That should work if we started tracking `[tag, aid]` instead, but it's more surgery.

As of 2): If a parser has understood enough of the protocol that it instantiates new child analyzers and/or starts DPD, it seems like a good point / heuristic to confirm its own protocol. It'd be great to have examples where this is not the case, but for that we'd go with `&ordered` first.

bbannier:
To me showing confirmed-then-violating analyzers is completely non-controversial (only minor but acceptable downside: we overload the name since our schema forces us to use a magic `-` since TSV already struggles with sets and we don't have much room for more metadata there).

Re: ordered, I think this is great and very valuable for users, but I also have the feeling that users will naturally expect this to correctly reflect the different protocol layers (e.g., "http,websocket" -> Websocket nested in HTTP). You also do this when you say

> To me, the bigger issue is that if you have something like HTTP->WebSocket->HTTP, you do not see the second HTTP being confirmed within history. That should work if we started tracking [tag, aid] instead, but it's more surgery.

This angle is not about "confirmation order", but about nesting. Correctly reflecting nesting is hard since it requires knowing when downstream analyzers confirm. With my Spicy trainer's hat on, what is the guidance on when to comfirm? I have the feeling "before forwarding any data/raising any events" is not always workable since it might require buffering potentially unbounded amounts of data even beyond what forwarding from a Spicy analyzer might require. I do not have a concrete example, but could totally see this happening, imagine e.g., a protocol with an identifier for the "next" analyzer followed by a huge payload, and only then enough structure to be reasonably sure about the protocol; a downstream analyzer might be able to confirm much faster after a couple dozen bytes if it e.g., expects some magic, but why would we expect the writer of the outer analyzer to even know that?

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -79,12 +75,11 @@ event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationI
...
-	# If the service hasn't been confirmed yet, don't generate a log message
-	# for the protocol violation.
-	if ( analyzer !in c$service )
+	# If the service hasn't been confirmed yet, or already failed,
+	# don't generate a log message for the protocol violation.
+	if ( analyzer !in c$service || analyzer in c$service_violation )
 		return;
 
-	delete c$service[analyzer];
 	add c$service_violation[analyzer];
```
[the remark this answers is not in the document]
0xxon: You raise a good point. Currently c$service_violation is used in exactly one place - the logging of violated services. It seems to me to make more sense to only log services that actually were removed due to their violation, aligning it with the way that "-analyzer" works. I will address this in a separate commit.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -79,12 +75,11 @@ event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationI
 
 	local c = info$c;
 	local analyzer = Analyzer::name(atype);
-	# If the service hasn't been confirmed yet, don't generate a log message
-	# for the protocol violation.
-	if ( analyzer !in c$service )
+	# If the service hasn't been confirmed yet, or already failed,
+	# don't generate a log message for the protocol violation.
+	if ( analyzer !in c$service || analyzer in c$service_violation )
```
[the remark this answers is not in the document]
0xxon: I disagree - `dpd.log` exists for the dpd state machine; once it removed the analyzer that's it.

`analyzer.log` will continue logging these repeat violations, so no information is logged.

Note that this is not really a change in functionality, as previously the service often was removed from `c$service` - which also stopped `dpd.log` entries to be generated.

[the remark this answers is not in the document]
0xxon: I kind of agree with you, but I still don't want to introduce a new change that will mess with how `dpd.log` looks like; the current way it works makes kind-of-sense. Getting rid of the already-violated-test causes a larger change of the baselines than keeping it here.

However, I agree also with you. `dpd.log` as it exists at the moment is not completely internally consistent. Plus - with analyzer.log existing there also is not much of a reason for it to exist at all anymore.

My plan is to do a followup PR that slightly extends analyzer.log, and gets rid of `dpd.log`. (With a deprecation cycle during which one can switch it on again). This should also remedy all of these concerns.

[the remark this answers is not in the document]
ckreibich: I look forward to seeing that one. I'm starting to sound like a broken record on this one but I'll flag again that analyzer.log is entirely undocumented, and I feel that the distinction of dpd.log and analyzer.log isn't particularly helpful to the user.

ckreibich:
Fwiw Arne's comments [above](https://github.com/zeek/zeek/pull/4200#issuecomment-2633537474) are essentially my take.

Putting aside for a moment the fact that I think this new behavior is strictly better than what's there right now, if this goes in now, it's only in master. Several of our power users will chew on it then, 7.2 is quite a ways off, and we can additionally flag this via the pre-release development update or a dedicated blog post. There's plenty of opportunity to iterate. I care about the NEWS update because we usually leave docs-writing until just before a release, so there's a quick explanation available at merge time.

Another thing: I'm always a bit dismayed when our own work (as opposed to a user contribution) triggers this level of discussion in a PR — it suggests to me we missed an opportunity to hash things out earlier.

scripts/base/frameworks/analyzer/dpd.zeek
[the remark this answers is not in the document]
0xxon: This is now updated, using a different mechanism for tracking. The service_violation record is retained for compatibility, but no longer used for this purpose.

scripts/policy/protocols/conn/failed-services.zeek
0xxon: This is updated and now uses the ``Analyzer::disabling_analyzer`` hook.

scripts/base/protocols/conn/main.zeek
0xxon: This now uses the proposed text.

scripts/test-all-policy.zeek
0xxon: As part of addressing some of the last comments, the column was renamed to ``failed_services``. The script is now called ``failed-services-logging``.

scripts/policy/protocols/conn/failed-services.zeek
awelzel: `Analyzer::disabling_analyzer hook` doesn't work correctly for this. It is invoked for all `disable_analyzer()` calls,  also for SSH and SSL when their analyzer's are disabled due to encryption starting.

Indeed, the hook was added to prevent this default disabling from 3rd party scripts for certain connections that want to peek at encrypted connections for just a bit longer, without needing to change the global options SSH/SSL connections affecting all connections.

A perfectly fine SSL connection reports failed_service `ssl` with this implementation.
```
$ zeek -C -r testing/btest/Traces/tls/tls13_wolfssl.pcap LogAscii::use_json=T Analyzer::Logging::include_disabling=T Analyzer::Logging::include_confirmations=T protocols/conn/failed-service-logging

cat analyzer.log 
{"ts":1607280745.836694,"cause":"confirmation","analyzer_kind":"protocol","analyzer_name":"SSL","uid":"C3hMwx1BFkXQxQ00yf","id.orig_h":"192.168.186.133","id.orig_p":43056,"id.resp_h":"192.168.186.134","id.resp_p":9090}
{"ts":1607280745.845941,"cause":"disabled","analyzer_kind":"protocol","analyzer_name":"SSL","uid":"C3hMwx1BFkXQxQ00yf","id.orig_h":"192.168.186.133","id.orig_p":43056,"id.resp_h":"192.168.186.134","id.resp_p":9090}

$ jq < conn.log 
{
  "ts": 1607280745.834631,
  "uid": "C3hMwx1BFkXQxQ00yf",
  ...
  "ip_proto": 6,
  "failed_service": [
    "ssl"
  ]
}
```

I'm totally okay with the `connection_state_remove` hook, feel free to revert :-)

awelzel:
> update to the policy script to change the column name to failed_servive

"Naming corner" and not a default column, so ignore me if this is just annoying, but: "failed_service" is new and very generic terminology and may mean various things to various people. Did a HTTP service return a 500 status? That's a failing service, right?

service_violations seemed okay to me given the "violation" part. The explicit/long version could be `disabled_violating_analyzers`. Either are a bit more specific and non-generic.

0xxon:
The problem with service_violations is that the name is overloaded - service violations are tracked separately (in ``c$service_violations``) and include services that were not removed.

``disabled_violating_analyzers`` on the other hand seems a tad long to me.

NEWS
```diff
@@ -22,6 +22,46 @@ New Functionality
...
+  detection (DPD) changed. Now, a violation that is raised by an analyzer
+  before it is confirmed will immediately disable the analyzer. This adjusts
+  the behavior back to the historically desired state, and aligns it with
+  the treatment of confirmed analyzers.
+
+  As a consequence of this, the option ``DPD::max_violations`` is no longer used.
+  It will be retained till Zeek 8.1 to prevent script errors, and raises a
+  deprecation warning.
+
+  The way failed service interact with the ``service`` field in the connection
```
ckreibich, line 43:
```suggestion
  The way failed services interact with the ``service`` field in the connection
```

NEWS
```diff
@@ -22,6 +22,46 @@ New Functionality
...
+  the ``service`` field in the connection log, and the entry will be retained
+  even if the analyzer raises a violation later.
+
+  To extend the visibility of protocol violations, a new option
+  ``DPD::track_removed_services_in_connection`` was added. Enabling it causes
+  failed analyzers to be logged to the ``service`` field of the connection log,
+  with a prepended "-". So a connection that attached the ``ssl`` analyzer
+  which later failed due to a protocol error will be logged as ``ssl,-ssl``.
+
+  Alternatively, this change also adds a new policy script,
```
ckreibich, line 60:
```suggestion
  This change also adds a new policy script,
```

ckreibich (approved):
Thanks! I'm approving. Just a few nits in the NEWS file.

0xxon:
Thank you - I updated the news file.

PR #4204, opened 2025-01-31
Make enc_part value from kerberos response available to scripts

philipp-tg:
This change makes the enc_part value of Kerberos responses available to scripts.
This field contains encrypted session information.
It is useful to know this value because it can be used to detect if an account has an insecure password or if an outdated encryption algorithm is being used.


timwoj (approved):

awelzel:
@philipp-tg - thanks for your contribution!

PR #4210, opened 2025-02-03
testing/btest: Use OPENSSL_ENABLE_SHA1_SIGNATURES

### LINKED ISSUE #4035: Test failures on platforms that distrust SHA-1 signatures
While adding Fedora 41 to CI (#4028) we noticed that its default crypto policy [no longer trusts SHA-1 signatures](https://fedoraproject.org/wiki/Changes/OpenSSLDistrustSHA1SigVer). This means:

- For users running certificate validation (`validate-certs.zeek`), validations involving SHA-1 will fail. While unlikely in current traffic, it's certainly possible.
- Our own tests rely on SHA-1 in validations, specifically the following:

```
[  8%] bifs.x509_verify ... failed
[ 84%] scripts.base.protocols.ssl.ocsp-stapling ... failed
[ 91%] scripts.policy.protocols.ssl.validate-ocsp ... failed
[ 91%] scripts.policy.protocols.ssl.validate-certs ... failed
```

This looks like a workable predicate (but my `openssl` fu is weak, so suggestions welcome):
```
$ openssl genpkey -out privkey.pem -algorithm rsa
$ openssl rsa -in privkey.pem -outform PEM -pubout -out pubkey.pem
$ touch foo
$ openssl dgst -sha1 -sign privkey.pem -out foo.signed foo # Fails when SHA-1's not trusted
```

We should offer users guidance on this and, I'd argue, fix our own tests so they either don't require SHA-1 for signature verification, or guard them with a suitable predicate.

awelzel:
This reverts the call to update-crypto-policies in the Fedora 41 image
and instead sets OPENSSL_ENABLE_SHA1_SIGNATURES in the individual tests.
This allows RHEL 10 or Fedora 41 users to run the tests in question
without needing to fiddle with system settings.

Fixes #4035

...and a commit to remove OpenSSL 1.0 support in x509_verify.zeek


timwoj (approved):
Looks fine to me. Maybe add a comment to the tests about why that variable gets set. Copying the one in subdir-btest.cfg would be enough.

0xxon:
I just noticed that you set this in btest.cfg for the external tests.

Could we just do that for the normal btests too? That way we don't have to set system specific variables in individual btests. That feels a bit awkward to me.

awelzel:
Eh, I just merged... I pondered the same but found it nicer that having the environment variable in the tests provides an explicit tag which tests/pcaps are affected. Further, if one copy&pastes the `TEST-EXEC` line directly. I sometimes do this for  easier debugging, the global environment variable that's now missing is not obvious and it might take a while to figure it out.

I had the same thought for external baselines, but given they have `OPENSSL_ENABLE_MD5_VERIFY=1` seemed more consistent to set it globally there - I'd rather change that to explicit usage within the test TBH.

PR #4211, opened 2025-02-03
Make --save-seeds imply deterministic mode

### LINKED ISSUE #4209: Using `--save-seeds` on its own is insufficient to make IDs reproducible
Zeek has a `--save-seeds` option which can be used to dump random seeds to a file. However, this does not seem sufficient to make things like random IDs reproducible, e.g.,

```console
$ zeek --save-seeds seeds.txt testing/btest/Traces/smtp.trace
```

produces logs, but attempting to reproduce the result produces different IDs

```console
$ zeek --load-seeds seeds.txt testing/btest/Traces/smtp.trace
```

The only way I found to make `--save-seeds` behave reproducible was to also use `--load-seeds` at the same time which lead to identical IDs. This does seem pointless to me though.

```console
$ zeek --load-seeds seeds.txt --save-seeds seeds2.txt testing/btest/Traces/smtp.trace
$ diff seeds.txt seeds2.txt. # Passes since files are identical.
```

awelzel:
See commits for a bit of context.

Closes #4209


0xxon (changes requested):
After looking at this for a while, I wonder if we could make the control flow a bit clearer. At the moment, there is a lot of indirection in the `init_random_seed` function. The `seed_done` (or `deterministic`) variable means two things. First, it tracks if the random seed has to be initialized. Furthermore it is co-opted later as the argument to zeek_srandom, which is purely used to determine if `zeek_rand_deterministic` deterministic should be set to true.

Furthermore, seed itself is partially used for that purpose too, though you already got rid of one of the use-cases for that, which also would have lead to some weird behavior in the past.

In any case - what do you think about the idea of just setting `zeek_rand_deterministic` directly in `init_random_seed`. It seems like this is where it belongs in any case. That also can be used to make the control flow clearer, and completely get rid of the `seed_done`/`deterministic` variable by just replacing it by an if statement.

My proposal would be something along these lines:

```patch
From 608c1cb9a3e5c461005b164d473f37aa5f6496c1 Mon Sep 17 00:00:00 2001
From: Johanna Amann <johanna@corelight.com>
Date: Tue, 4 Feb 2025 10:31:03 +0000
Subject: [PATCH] Proposal

---
 src/util.cc | 48 ++++++++++++++++++++++--------------------------
 1 file changed, 22 insertions(+), 26 deletions(-)

index 604867721..34691a9f3 100644
--- a/src/util.cc
+++ b/src/util.cc
@@ -363,9 +363,8 @@ static long int zeek_rand_state = 0;
 static bool first_seed_saved = false;
 static unsigned int first_seed = 0;
 
-static void zeek_srandom(unsigned int seed, bool deterministic) {
+static void zeek_srandom(unsigned int seed) {
     zeek_rand_state = seed == 0 ? 1 : seed;
-    zeek_rand_deterministic = deterministic;
 
     srandom(seed);
 }
@@ -380,26 +379,28 @@ void seed_random(unsigned int seed) {
 void init_random_seed(const char* read_file, const char* write_file, bool use_empty_seeds,
                       const std::string& seed_string) {
     std::array<uint32_t, zeek::detail::KeyedHash::SEED_INIT_SIZE> buf = {};
-    size_t pos = 0; // accumulates entropy
-    bool deterministic = false;
     uint32_t seed = 0;
 
-    if ( read_file ) {
-        if ( ! read_random_seeds(read_file, &seed, buf) )
-            reporter->FatalError("Could not load seeds from file '%s'.", read_file);
+    if ( write_file )
+        // run in deterministic mode when we write a file
+        zeek_rand_deterministic = true;
 
-        deterministic = true;
+    if ( read_file || use_empty_seeds || ! seed_string.empty() ) {
+        // if a seed is provided - run Zeek in deterministic mode
+        zeek_rand_deterministic = true;
+
+        if ( read_file ) {
+            if ( ! read_random_seeds(read_file, &seed, buf) )
+                reporter->FatalError("Could not load seeds from file '%s'.", read_file);
+        }
+        else if ( ! seed_string.empty() ) {
+            if ( ! fill_random_seeds(seed_string, &seed, buf) )
+                reporter->FatalError("Could not load seeds from string");
+        }
     }
-    else if ( ! seed_string.empty() ) {
-        if ( ! fill_random_seeds(seed_string, &seed, buf) )
-            reporter->FatalError("Could not load seeds from string");
+    else {              // no seed provided
+        size_t pos = 0; // accumulates entropy
 
-        deterministic = true;
-    }
-    else if ( use_empty_seeds )
-        deterministic = true;
-
-    if ( ! deterministic ) {
 #ifdef HAVE_GETRANDOM
         // getrandom() guarantees reads up to 256 bytes are always successful,
         assert(sizeof(buf) < 256);
@@ -437,18 +438,13 @@ void init_random_seed(const char* read_file, const char* write_file, bool use_em
             reporter->FatalError("Could not read enough random data. Wanted %d, got %zu",
                                  zeek::detail::KeyedHash::SEED_INIT_SIZE, pos);
 
-        if ( ! seed ) {
-            for ( size_t i = 0; i < pos; ++i ) {
-                seed ^= buf[i];
-                seed = (seed << 1) | (seed >> 31);
-            }
+        for ( size_t i = 0; i < pos; ++i ) {
+            seed ^= buf[i];
+            seed = (seed << 1) | (seed >> 31);
         }
     }
 
-    if ( write_file )
-        deterministic = true;
-
-    zeek_srandom(seed, deterministic);
+    zeek_srandom(seed);
 
     if ( ! first_seed_saved ) {
         first_seed = seed;
-- 
2.39.5 (Apple Git-154)
```

awelzel:
@0xxon - I've replaced my patches with yours. Will merge if tests pass.

0xxon:
Sounds good to me - thank you

PR #4215, opened 2025-02-04
Update detect-external-names with option and testcase

0xxon:
This supersedes GH-4183

In addition to it, this adds IPv6 support and a testcase.


scripts/policy/protocols/dns/detect-external-names.zeek
```diff
@@ -39,3 +39,24 @@ event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priori
 		        $identifier=cat(a,ans$query)]);
 		}
 	}
+
+# exactly the same as above.
```
awelzel, line 43: If it's exactly the same, might make sense to promote the original handler to a function that's then called from either.

awelzel (approved):
Thanks for picking this up!

PR #4217, opened 2025-02-05
Preserve Original Casing in DNS Queries & HTTP Headers to Prevent Data Obfuscation

kayes817:
Originally in Zeek, hostnames in DNS queries were converted to lowercase, reflecting DNS's case-insensitivity. HTTP headers underwent a process that transformed them into uppercase.

While these conventions align with protocol specifications, adversaries can exploit them. 
Example: If Base64-encoded data is exfiltrated through DNS queries, altering the case could obscure the encoded content. 

Methods like these would circumvent detection rules designed to identify Base64 data. This commit emphasizes the importance of observing actual data rather than relying solely on interpreted formats.

### DIFF
--- a/src/analyzer/protocol/dns/DNS.cc
+++ b/src/analyzer/protocol/dns/DNS.cc
@@ -185,11 +185,6 @@ bool DNS_Interpreter::ParseQuestion(detail::DNS_MsgInfo* msg, const u_char*& dat
     if ( dns_event && ! msg->skip_event ) {
         String* original_name = new String(name, name_end - name, true);
 
-        // Downcase the Name to normalize it
-        for ( u_char* np = name; np < name_end; ++np )
-            if ( isupper(*np) )
-                *np = tolower(*np);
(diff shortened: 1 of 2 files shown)


awelzel:
Hey @kayes817 - thanks for looking at Zeek and being up for improving it.

For DNS and HTTP headers, there's two even declarations that allow you to get the original casing already:

```
event http_header(c: connection, is_orig: bool, original_name: string, name: string, value: string) { print "header", c$uid, is_orig, original_name, name, value; }

event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qclass: count, original_query: string); 
```

`original_name` and `original_query` allow you to get the original casing. If you want these to appear in, e.g. dns.log or http.log, you can extend the DNS::Info, respective HTTP::Info record and populate it in a separate script.

I'm going to close this PR. If you have questions, reach out on Slack or Discourse (http://zeek.org/community)

PR #4220, opened 2025-02-05
Fix TSAN reportings for ZeroMQ

awelzel:


src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -227,7 +218,9 @@ bool ZeroMQBackend::DoInit() {
     // Setup connectivity between main and child thread.
     main_inproc.bind("inproc://inproc-bridge");
     child_inproc.connect("inproc://inproc-bridge");
-    self_thread = std::thread(self_thread_fun, this);
+
+    // Thread is joined in backend->DoTerminate(), backend outlives it.
```
bbannier, line 222: Since this is not a `jthread`, but just a `thread` it will not autojoin on destruction; let's add a destructor for `ZeroMQ` so we do not leak this thread under any circumstances. If `DoTerminate` is idempotent we might be able to just call it in the dtr for cleanup (in a `try`/`catch` to prevent throwing from a dtr).

And if I understand correctly, as written this change here just fixes style, but doesn't change behavior at all.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -356,7 +349,10 @@ bool ZeroMQBackend::DoPublishLogWrites(const logging::detail::LogWriteHeader& he
 }
 
 void ZeroMQBackend::Run() {
-    util::detail::set_thread_name(zeek::util::fmt("zmq-%p", this));
+    char name[4 + 2 + 16 + 1]; // zmq-0x<8byte pointer in hex><nul>
```
bbannier, line 352: This is default-initializing the elements (i.e., they'll have random values), let's instead zero-initialize them.

```suggestion
    char name[4 + 2 + 16 + 1]{}; // zmq-0x<8byte pointer in hex><nul>
```

For me `printf("%p", this)` seems to return a a string with 12 values, so properly initializing the buffer would prevent us potentially reading unit memory. The docs for `set_thread_name` state

>  OS limitations typically truncate the name to 15 bytes maximum.

so a big part might be thrown away anyway 🤷.

bbannier (approved):

src/cluster/backend/zeromq/ZeroMQ.cc
awelzel: > seems to return a string of length 11

This will depend on the address of `this`, with 8 byte pointers this could be larger, but yes, the name is truncated by the OS.

> properly initializing the buffer would prevent us potentially reading uninit memory

`snprintf()` will null terminate (so the extra null assignment I added is superflous, actually) - so this shouldn't be a concern.

src/cluster/backend/zeromq/ZeroMQ.cc
awelzel: Correct, no change in functionality. Let me see about the destruction, thanks!

PR #4224, opened 2025-02-06
SSL main.zeek - fix indentation

0xxon:
No functional changes


awelzel (approved):
There's no users in-tree users or tests for SSL::delay_log or SSL::undelay_log. Do you know any users? Maybe these could be deprecated in favor of `Log::delay()`, `Log::delay_finish()`?

0xxon:
I do not know if anyone uses these :)

PR #4225, opened 2025-02-07
DPD changes - ZAM baseline updates

0xxon:
See GH-4200 for related PR


bbannier (approved):
LGTM, so if CI passes 🚀.

That said, the baselines in `script.policy.misc.dump-events` are simply to big to review them in detail, pretty sure this was discussed before and decided this works like designed 🤷.

0xxon:
It's the same changes that already were reviewed in #4200 - just that this time they are for the ZAM btest directory.

0xxon:
Also - can I just note that the ZAM tests seem a bit slow.

They aborted twice so far with instance rescheduling.

0xxon:
And OOM killed.

This is starting to feel a bit silly...

ckreibich:
> the ZAM tests seem a bit slow.

I looked at the test runtimes for those runs and yeah that's rough ... am also surprised: the regular ASAN build took [33 mins](https://cirrus-ci.com/task/6387581849436160), and the ASAN one with ZAM that eventually worked out took [just under 36 mins](https://cirrus-ci.com/task/6157056459145216), so that's actually not too bad given the ZAM startup overhead.

But the earlier (aborted) runs were way way slower, so I'm going to blame those CI runners here. I can't immediately comment on the OOM, but I've been finding Cirrus instances generally prone to delays/rescheduling lately, regardless of their workload.

PR #4226, opened 2025-02-08
Fix crash related to Broker stores

Neverlord:
This has been reported in chat and is caused by a race in Broker where the store proxy object can be invalid if the store object is discarded by Broker before the proxy is constructed.

### DIFF
--- a/src/broker/Manager.cc
+++ b/src/broker/Manager.cc
@@ -1724,6 +1724,12 @@ detail::StoreHandleVal* Manager::MakeClone(const string& name, double resync_int
     auto handle = new detail::StoreHandleVal{*result};
     Ref(handle);
 
+    if ( ! handle->proxy.valid() ) {
+        Error("Failed to create clone for data store %s", name.c_str());
+        Unref(handle);
+        return nullptr;
+    }
+
     data_stores.emplace(name, handle);
     if ( ! iosource_mgr->RegisterFd(handle->proxy.mailbox().descriptor(), this) )
         reporter->FatalError("Failed to register broker clone mailbox descriptor with iosource_mgr");


awelzel:
> This has been reported in chat and is caused by a race in Broker where the store proxy object can be invalid if the store object is discarded by Broker before the proxy is constructed.

Do you have an explanation (or also hunch) when/why this can happen? Is it timeout based? Error reporting seems better than crashing, but IIUC the store won't be functional and it'd be good to recover from that. Or else, have some explanation for the user what the underlying problem is and how it could be fixed.

Neverlord:
> Do you have an explanation (or also hunch) when/why this can happen?

It's basically a small race between creating the store object and the proxy. If the proxy creation fails, it means that the store object got discarded in the meantime by Broker. I don't know why that could happen, though. Maybe we can re-run the setups that broke prior and see what stores fail in what conditions? At least they shouldn't crash now.

0xxon:
Should we perhaps crash Zeek with a `FatalErrror` or `FatalErrorWithCore` when this happens? That prevents a store from (relatively silently) not working, and would at least give a reason to users what happened.

awelzel (approved):
Are the metrics available differently now, or created lazily? I'm a bit confused by the removal of them, though it's just a side-effect of the broker bump I assume.

Neverlord:
> Are the metrics available differently now, or created lazily? I'm a bit confused by the removal of them, though it's just a side-effect of the broker bump I assume.

The metrics change is an unrelated thing. Long story short, those metrics are useless (always 0) we're working on a replacement.

ckreibich:
@initconf is hitting this one in production with the 7.2 RC1 right now.

initconf:
> @initconf is hitting this one in production with the 7.2 RC1 right now.

Corresponding with Christian on testing channel thread!  Ping me if you need more info including backtraces etc from the core dumps. Now I am at point where I can replicate this as will !!!

PR #4228, opened 2025-02-12
Set original/outer packet flags to reflect inner packet results

mnhsrj:
Zeek's IPTunnel analyzer does not set the 'processed', 'dump_packet' and 'dump_size' flags on the outer packet. As a consequence, when Zeek is run with -c (dump unprocessed packets), the outer packet is dumped even when it is accounted for in tunnel.log. When Zeek is run with -w (essentially dump all packets), the outer packet is not dumped.

Looks like for analyzing a tunnelled packet, the IPTunnel analyzer creates a new packet to handle the inner packet. 
However, the flags set by session analyzers on the inner packet , such as 'processed', 'dump_packet' and 'dump_size', are not properly propagated back to the outer packet.

This commit attempts to address this issue by propagating these flags from the inner packet to the outer packet.

### DIFF
--- a/src/iosource/Packet.cc
+++ b/src/iosource/Packet.cc
@@ -46,6 +46,7 @@ void Packet::Init(int arg_link_type, pkt_timeval* arg_ts, uint32_t arg_caplen, u
         data = arg_data;
 
     dump_packet = false;
+    dump_size = 0;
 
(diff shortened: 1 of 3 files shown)


timwoj (approved):
This looks reasonable to me. I'll fix the failure from pre-commit during the merge.

PR #4233, opened 2025-02-15
Update BIFs to_int() and to_count() to accept optional `base` argument

### LINKED ISSUE #4076: Provide a way to convert hex strings into numeric type
There is no way to convert a hex string (`"0x1"`) into a numeric type like `count` or `int`. Making `to_count`/`to_int` accept hex representation or introducing a `hexstr_to_count` would be nice.

markoverholser:
Allows for more easy conversions of strings representing (for example) hexadecimal, octal, or binary numbers
Fixes #4076

"Baby's first PR," let me know if I need to bolster anything.


src/zeek.bif
```diff
@@ -2675,12 +2675,12 @@ function double_to_count%(d: double%): count
...
 ##
 ## .. zeek:see:: to_addr to_int to_port to_subnet
-function to_count%(str: string%): count
+function to_count%(str: string, base: count &default=10%): count
 	%{
 	const char* s = str->CheckString();
 	char* end_s;
 
-	uint64_t u = (uint64_t) strtoull(s, &end_s, 10);
+	uint64_t u = (uint64_t) strtoull(s, &end_s, base);
```
bbannier, line 2683: Nit: since we are touching this already we could fix this to use a proper cast, e.g.,

```suggestion
	uint64_t u = static_cast<uint64_t>(strtoull(s, &end_s, base));
```

bbannier (commented):
This looks good to me, it just seems to need an update of some test baseline elsewhere; running the full test suite with `btest -u` should take care of that.

Somewhat tangentially related, while playing with this I noticed that the failure handling of these functions is somewhat broken in that we never check the error code of `strtoull`. This means I get silently incorrect results from at least `to_int`, e.g.,

```zeek
	print to_int("FFF", 2);  # Prints '0' w/o error.
```

We should fix that and also add some more combinations to the tests (valid and invalid, choice of binary/octal/decimal/hex as base). I defer to @awelzel or @timwoj on whether we want to force that on you 😬 or do it later.

markoverholser:
I'm happy to add more test cases, that seems like something I can cobble together, though I won't likely get to it until after wrapping up The Big Thing on My Plate this week. If you want me to look at `strtoull()` and how I can adjust `to_int()` and `to_count()` to recognize failures and then deal with them appropriately, I'll probably need some more guidance/coaching since I don't have the familiarity with the code base that you do.

markoverholser:
I just read up on `strtoull()` a bit, it just returns a `0` when the conversion fails, or whatever the max for a ULL is if the converted value is out of bounds, but I don't see any way to tell the difference between it returning a `0` because it failed and it returning a `0` because, say, you passed in `"0"`. Do you see something I don't?

timwoj:
> I just read up on `strtoull()` a bit, it just returns a `0` when the conversion fails, or whatever the max for a ULL is if the converted value is out of bounds, but I don't see any way to tell the difference between it returning a `0` because it failed and it returning a `0` because, say, you passed in `"0"`. Do you see something I don't?

`strtoull` will return a `0` for a failed conversion, but will also set `errno` to EINVAL.

markoverholser:
@timwoj this is obviously a design pattern I hadn't encountered but seems like it's pretty fundamental to error handling in C/C++. It shows my naivety. There's a reason I called this "baby's first PR." Thanks for your patience with me. I may not be immediately valuable, but I believe I have potential.

@bbannier since you're working another PR surrounding this, I'm not sure what my next steps are, if any. Do you need anything else from me? I believe I fixed up the remaining tests that were failing from my work.

testing/btest/bifs/to_count.zeek
```diff
@@ -19,6 +19,9 @@ event zeek_init()
 	print to_count("7");
 	print to_count("");
 	print to_count("-5");
+	print to_int("10101100", 2);
+	print to_int("43", 8);
+	print to_int("C3", 16);
```
timwoj, line 24: Should these be tests for `to_count`? Maybe add a test for the `0xFF` form of hex numbers as well.

timwoj (changes requested):
If you can handle the error case from `strotull` as already discussed that'd be good.

bbannier:
> If you can handle the error case from `strotull` as already discussed that'd be good.

My intention was for that to be taken care of by #4236. The existing and now extended error handling strategy here is slightly unusual since we not directly inspect error codes, but instead check side what is consumed, but I believe this should be fine. Are we missing any other errors which we should handle @timwoj?

testing/btest/bifs/to_count.zeek
markoverholser: Yes, that was a copy/pasta, thanks for catching. And having more tests for the "0x" format sounds good, I added one to each.

timwoj:
> Are we missing any other errors which we should handle @timwoj?

There's also the overflow cases where `stroull` will return `ULLONG_MAX` but I agree the error cases can be handled separately.

timwoj (approved):

0xxon:
I caught this too late - this also needs an update of the function argument documentation for the BIFs.

PR #4236, opened 2025-02-18
Align error handling in `to_int` with existing behavior of `to_count`

bbannier:


timwoj (approved):
I'm ok with trimming both sides of the string. It doesn't make sense to be ok with one and not the other. I am somewhat surprised we don't have a trim method in util.h considering how many other string utilities we have.

0xxon:
As a side-note to this - it seems like this behavior should also be documented in the function documentation :)

ckreibich:
Docs updated via #4247 (thx Arne for flagging), let me adjust the tagging.

Issue #4237, opened 2025-02-18
Invalid RDP binpac exception


MMChrisHinshaw:
[rdp-no-cookie-mstshash-filtered.pcap.zip](https://github.com/user-attachments/files/18852237/rdp-no-cookie-mstshash-filtered.pcap.zip)

analyzer.log

```
Binpac exception: binpac exception: string mismatch at /src/zeek/src/analyzer/protocol/rdp/rdp-protocol.pac:81: \x0aexpected pattern: "Cookie: mstshash\\="\x0aactual data: "\x01"
```

The above occurs with a pcap that I will attach shortly on `try.zeek.org` and zeek 7.1.0 (as well as the old version that I'm currently using... 5.2.2).

Per [this document](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/18a27ef9-6f9a-4501-b000-94b1fe3c2c10), the RDP Connection Request states the Cookie: mstshash= in the message is: 

```
cookie (variable): An optional and variable-length ANSI character string terminated by a 
0x0D0A two-byte sequence. This text string MUST be "Cookie: mstshash=IDENTIFIER", where 
IDENTIFIER is an ANSI character string (an example cookie string is shown in section 4.1.1). 
The length of the entire cookie string and CR+LF sequence is included in the X.224 
Connection Request Length Indicator field. This field MUST NOT be present if the 
routingToken field is present.
```

It looks like the `rdp-protocol.pac` assumes that the cookie field will always be present, but it is optional.

0xxon:
I think I have a working fix for this.

Just to confirm - is it ok to use the pcap you provided as a test-case for Zeek?

MMChrisHinshaw:
Yes, please feel free to use that pcap.

0xxon:
The full proposed patch is in https://github.com/zeek/zeek/pull/4241. If you happen to have the ability to test this in live traffic, that would be awesome. It would be especially great to hear if this leads to any false positives for you, or to an increase in RDP related dpd.log/analyzer.log entries.

MMChrisHinshaw:
I will do what I can to test this, but cannot guarantee how soon it can be accomplished.  

Please do not let that hold you up in releasing this change. I really appreciate the quick turn around.

PR #4241, opened 2025-02-21
RDP: cookie is optional

0xxon:
This addresses GH-4237

The change is relatively small, and should not impact parsing of real RDP connections in any way; all test suites still pass. 

However, this change also means that parsing now is more forgiving, and there is a chance that some non-RDP-data might be able to sneak by the parser, and the paser might be willing to interpret is as RDP. On the flip-side, the parser currently rejects some RDP flows that are RDP and can be continued to be parsed.

This could probably benefit from some public testing - so 7.2 might be a good target for it. If we can find people that are able to keep an eye to changes to rdp.log / analyzer.log entries for RDP, that would also be great.


sethhall (approved):
Looks good to me

src/analyzer/protocol/rdp/rdp-protocol.pac
```diff
@@ -74,18 +74,27 @@ type Data_Block = record {
...
+} &byteorder=littleendian;
+
 type Connect_Request(cotp: COTP) = record {
 	destination_reference: uint16;
 	source_reference:      uint16;
 	flow_control:          uint8;
-	cookie_mstshash:       RE/Cookie: mstshash\=/;
-	cookie_value:          RE/[^\x0d]*/;
-	cookie_terminator:     RE/\x0d\x0a/;
+	cookie_mstshash:       RE/(Cookie: mstshash\=){0,1}/;
```
awelzel, line 86: Does that work in BinPac?

```suggestion
	cookie_mstshash:       RE/(Cookie: mstshash\=)?/;
```

awelzel (approved):

src/analyzer/protocol/rdp/rdp-protocol.pac
0xxon: It might, let me check. I tried a different approach first that didn't - which is why I ended up with `{0,1}`.

timwoj:
I installed this on a machine at LBL and it's definitely catching more RDP traffic. The gzips for rdp.log are about 4x the size since I restarted. I scrolled through the last couple of days of rdp.logs and I don't see anything crazy that I'd consider bad traffic.

PR #4244, opened 2025-02-24
Allow to redef ports of built-in analyzers

J-Gras:


0xxon (changes requested):
Thanks for this, I think it's neat to be able to redef these.

However I think the ports should not be an option. From my point of view, options should be used for settings that are changeable during runtime. However, if the ports are changed during runtime using `Config::set_value`, this will not update the ports that the analyzers listen to.

Hence I think those should also only be redef-able consts.

Also - as those all are moving inside the documentation-visible part, the comments should potentialy be extended to state the protocol name, or they should be added for the files in which no comment is attached at all :) 

I also wonder a bit if we should rename it from ports to something like well_known_ports, know that this is public-facing - but I don't have a strong opinion about this.

It would also be neat if there was a test for at least one protocol to check that this works in practice.

I apologize about the wall of comments :)

awelzel:
Thanks for doing this! In addition to @0xxon , think it's nicer to have the `redef` out of the `export` section similar to what we do in the package-template (which also uses `const &redef` for similar reasons then outlined):

https://github.com/zeek/package-template/blob/f3026e7a9e19e2bce2239fb80cbe1440a684e326/features/spicy-protocol-analyzer/scripts/main.zeek%40ALT-one-unit%40#L46

scripts/base/protocols/dhcp/main.zeek
```diff
@@ -112,6 +112,11 @@ export {
 	## Event that can be handled to access the DHCP
 	## record as it is sent on to the logging framework.
 	global log_dhcp: event(rec: Info);
+
+	# 67/udp is the server's port, 68/udp the client.
+	# 4011/udp seems to be some proxyDHCP thing.
+	option ports = { 67/udp, 68/udp, 4011/udp };
+	redef likely_server_ports += { 67/udp };
```
awelzel, line 119: Seems that's the only one where we can't use ports. How about splitting into `client_ports` and `server_ports` just in case .

0xxon:
Implementing a change-handler would be the greatest way to deal with this. However, I am not entirely sure that's easily doable - I don't think we have the necessary BIFs. Furthermore, this would probably also necessitate changing likely_server_ports to an option, which the change_handler then also potentially should update.

awelzel:
>> How about implementing a change handler?
>
> Implementing a change-handler would be the greatest way to deal with this.

My two cents: Doesn't seem worth the complexity and potential for new bugs and not-functioning-as-expected-now scenarios.

Doing it via a change handler puts a soft requirement on other functionality to support the same. It'd also propagate to  external analyzers to have this logic, too. So first reaction is that while we could potentially implement it, simple and easy is the better idea for maintenance and the ecosystem. Do you guys think there's an actual use-case / scenario for adding/removing ports for analyzers at runtime regularly that wouldn't be covered by schedule_analyzer()?. Seems restarting Zeek is totally fine if this is generally about a static configuration that barely changes.

0xxon:
Actually - I just realized that, redef-ing the ports is, potentially, a tad problematic in general.

Specifically, redef-ing one of the port-sets for the protocols will not lead to `likely_server_ports` to be updated, even when using redefs. From my understandong it will have the original value of ports.

I guess that `likely_server_ports` is not super important, and not used widely - but it does seem a bit awkward.

One could change the type of this and the way that it is collected - but that would break every protocol parser that's out at the moment.

ckreibich:
`likely_server_ports` is arguably important in that Zeek uses it to flip connections on partial flows. So I like the idea of building out something that keeps this consistent.

This reminds me a bit of `Site::local_nets` vs `Site::private_address_space` — we did use change handlers there to keep them in sync.

@J-Gras are you still interested in working on this?

J-Gras:
> @J-Gras are you still interested in working on this?

@ckreibich, I am happy to give it another shot if there is definitive agreement on the approach. It's quite cumbersome to go through this again and again.

ckreibich:
Okay cool Jan, let's make sure we're all on the same page then. I finally (sorry!) looked at this and here's my take:

- `likely_server_ports` is currently used to implement [`IPBasedAnalyzer::IsLikelyServerPort()`](https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc#L124-L141), which in turn is used by the TCP and UDP packet analyzers to potentially flip endpoints. It's also used in the script layer in the [Conn module](https://github.com/zeek/zeek/blob/master/scripts/base/protocols/conn/inactivity.zeek#L35).
- `likely_server_ports` is in lots of third-party packages — basically nearly all protocol analyzers, also scan detection. I don't think we should break these, even after a deprecation cycle, if we can avoid it.
- `likely_server_ports` is updated by both packet- and protocol analyzers (see e.g. [here](https://github.com/zeek/zeek/blob/master/scripts/base/packet-protocols/teredo/main.zeek#L20-L27) for an example of the former).
- The `Analyzer::register_for_ports()` and `PacketAnalyzer::register_for_ports()` calls receive nearly exactly the same set of ports as `likely_server_pors` in aggregate, with the exception of DHCP, already flagged above, where the analyzer is currently registered for more ports than are likely servers.
- That said I found two packages, GQUIC and Amazon's S7, that define `ports` and `likely_server_ports` but don't pass them to an analyzer's port registration.

So we already harvest nearly the exact content of `likely_server_ports` via those two calls. If we extend the signature slightly, perhaps like this ...
```
global register_for_ports: function(tag: Analyzer::Tag, server_ports: set[port], addl_ports: set[port]) : bool;
```
... we have a way to address the DHCP case, and these calls can update `likely_server_ports` automatically (as a table, it doesn't complain about the update at runtime from the script layer, or we could do it in the core).

Since this happens at runtime it reflects any per-analyzer redef'd port sets. Users can still redef `likely_server_ports` if they want to, but no longer need to. We can remove the explicit `likely_server_ports` additions for each analyzer throughout our code.

Does that sound okay to you guys?

awelzel:
> So we already harvest nearly the exact content of likely_server_ports via those two calls. If we extend the signature slightly, perhaps like this ...

Nice! My summary is that your proposal would mean no changes to base scripts or external packages around port registration. Modulo the opportunity to remove the explicit redef of `likely_server_ports`, as `server_ports` is added to `likely_server_ports` behind the scenes.

> Does that sound okay to you guys?

I'm a fan. Definitely better than adding a new function! `addl_ports` maybe as `non_server_ports`, so the name implies a bit that it won't be added to `likely_server_ports` without reading all of the doc string :-)

ckreibich:
@J-Gras, what's your availability for this these days? Should we take it on?

J-Gras:
@ckreibich, happy to pick it up again. Am I correct that we go for https://github.com/zeek/zeek/pull/4244#issuecomment-2933023650 and replace `addl_ports` by `non_server_ports`?

ckreibich:
Sounds right to me, yep!

ckreibich:
I'm going to close this Jan. We have a way forward per the above, so when you find time for it, please just open a new PR. (And, if you know you won't find the time, let us know and we can look into it!)

PR #4246, opened 2025-02-24
Update ZAM baselines for to_int/to_count base changes

timwoj:
Some day I'll remember to check the ZAM baselines when merging PRs.


0xxon:
As this happens a lot, and not just to you - should we perhaps just include this in the default set that is run on PRs? :)

vpax:
> small task that checks if zam baselines for test that has been updated exist

Most of the time the ZAM BTests use the default baseline (as one would hope!). The exceptions are rare, ~5%.

timwoj:
I added the asan ZAM job by default. I'll work on the pre-commit idea separately, since I'd like to get master building again.

awelzel:
> I added the asan ZAM job by default. I'll work on the pre-commit idea separately, since I'd like to get master building again.

We could also modify a non-ASAN job to run `btest -a zam` in addition to the default test without having a separate task.

The ASAN tasks use `--build-type=debug + address sanitizer + coverage` which likely makes them particularly slow.


Seems ASAN taks take 50 or 60 min (non-ZAM vs ZAM) while other tasks take ~15min, so we could even test moving ASAN away from PRs and see how we fare with that.

timwoj:
> We could also modify a non-ASAN job to run `btest -a zam` in addition to the default test without having a separate task.
> 
> The ASAN tasks use `--build-type=debug + address sanitizer + coverage` which likely makes them particularly slow.
> 
> Seems ASAN taks take 50 or 60 min (non-ZAM vs ZAM) while other tasks take ~15min, so we could even test moving ASAN away from PRs and see how we fare with that.

The reason we were using the ASAN/USBAN jobs as the base for ZAM was because both had actually caught memory leaks and UB before.

.cirrus.yml
```diff
@@ -276,6 +276,20 @@ ubuntu24_task:
...
+  container:
+    # Ubuntu 24.04 EOL: Jun 2029
+    dockerfile: ci/ubuntu-24.04/Dockerfile
+    << : *RESOURCES_TEMPLATE
+  << : *CI_TEMPLATE
+  env:
+    ZEEK_CI_SKIP_UNIT_TESTS: 1
+    ZEEK_CI_SKIP_EXTERNAL_BTESTS: 1
+    ZEEK_CI_BTEST_EXTRA_ARGS: -a zam
+    # Use a lower number of jobs due to OOM issues with ZAM tasks
```
awelzel, line 290: Without ASAN this might not be needed. Maybe remove before merge and see how it behaves?

awelzel (approved):

PR #4252, opened 2025-02-27
Spicy: Fix passing vectors to Zeek.

### LINKED ISSUE #4250: Spicy: Vectors of enums fail to compile `to_val` call
This was reported [on Zeek's public slack](https://zeekorg.slack.com/archives/C012CQV1K29/p1740599323101159)

The following code defines a vector of enums and then passes it into an event:

```
# enumerr.spicy
module EnumErr;

type E = enum {
    One,
    Two,
};

public type Data = unit {
    var vec: vector<E>;
};
```

```
# enumerr.evt
protocol analyzer spicy::EnumErr over UDP:
    parse with EnumErr::Data,
    port 69/udp;

# This passes in the vector
on EnumErr::Data -> event enumerr::test(self.vec);
```

This fails at C++ compilation via `spicyz` (many candidates omitted for brevity except for one I found important):

```
spicyz -o enumerr.hlto enumerr.spicy enumerr.evt
/private/var/folders/w6/h7tg594x0yng0_yzvcsn6mv00000gn/T/spicy_hooks_EnumErr_61ba3b74eaadb444-cea7c1cb06d05765.cc:88:23: error: no matching function for call to 'to_val'
   88 |     args.emplace_back(::zeek::spicy::rt::to_val((*__self).vec, ::zeek::spicy::rt::event_arg_type((*__hlt::spicy_hooks_EnumErr::__zeek_handler_enumerr_test_216b05a4de4a8f01), ::hilti::rt::integer::safe<std::uint64_t>{0U})));
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~

...

/Users/etyp/.local/zeek/include/zeek/spicy/runtime-support.h:699:15: note: candidate template ignored: could not match 'allocator' against 'Allocator'
  699 | inline ValPtr to_val(const hilti::rt::Vector<T>& v, const TypePtr& target) {
      |               ^
```

Seems like the issue is that enums use [Spicy's `Allocator`](https://github.com/zeek/spicy/blob/2f46bd6c5d41e2eb93d88dda6468b966ab07bce5/hilti/runtime/include/types/vector.h#L52-L53) not `std::allocator`. `uint8` (which works) doesn't define one so it defaults to `std::allocator`.

rsmmr:
We missed the allocator argument for the C++-side conversion
template, which let vector's of certain types not compile.

Closes #4250.


src/spicy/runtime-support.h
```diff
@@ -502,8 +502,8 @@ template<typename K, typename V>
 ValPtr to_val(const hilti::rt::Map<K, V>& s, const TypePtr& target);
 template<typename T>
 ValPtr to_val(const hilti::rt::Set<T>& s, const TypePtr& target);
-template<typename T>
-ValPtr to_val(const hilti::rt::Vector<T>& v, const TypePtr& target);
+template<typename T, typename Allocator>
+ValPtr to_val(const hilti::rt::Vector<Allocator>& v, const TypePtr& target);
```
evantypanski, line 506: Should this not keep `T`? Or am I confused?

```suggestion
ValPtr to_val(const hilti::rt::Vector<T, Allocator>& v, const TypePtr& target);
```

evantypanski (approved):

PR #4258, opened 2025-03-04
Switch to built-in CMake support for IWYU and clang-tidy

timwoj:
This adds to two configure flags called `--enable-iwyu` and `--enable-clang-tidy` for adding `include-what-you-use` and `clang-tidy` output to builds. It uses the built-in functionality of CMake to add them directly to the `zeek_objs` build target. This also removes the existing hand-written `clang-tidy` target. That one ran separately from a regular build and it wasn't obvious to developers that it existed.

If this gets merged, I'll likely go make similar changes to the various submodules that were using the `FindClangTidy.cmake` script. Eventually that script can go away.


0xxon:
This looks good to me - I just have one side-question. Is include-what-you-use something we should run in one of our CI tests?

timwoj:
For reference, here's the log from a build with `--enable-iwyu` and no other changes: [iwyu.log](https://github.com/user-attachments/files/19076968/iwyu.log)

ckreibich (approved):
LGTM Tim. It might be nice to have some output in the configuration stage that indicates when these features get enabled (as the prometheus-cpp output does for IWYU).

Issue #4219, opened 2025-02-05
`zeek --parse-only` has surprising side-effect of potentially triggering DNS resolutions


bbannier:
A probably naive expectation would be that `zeek --parse-only` only parses the passed scripts and reports any syntax errors; I would not expect it to have any other effects. With that mental picture I implemented syntax checking for [ALE's Zeek linter for vim](https://github.com/dense-analysis/ale/blob/master/ale_linters/zeek/zeek.vim) and also in [my language server for Zeek](https://github.com/bbannier/zeek-language-server/blob/763c2ea6d1c1f1c29fcf51cb5c61799ad1d56f41/src/zeek.rs#L127-L132). Both these tools trigger the syntax check when a file is saved, but users might e.g., configure their editor to auto-save changes periodically or on editor exit. **EDIT**: just checked zeek-language-server and I actually also automatically trigger a syntax check when a file is opened, so just opening a file with that editor integration on and `zeek` in path is sufficient to trigger the side effects.

Currently merely parsing a Zeek script with `--parse-only` _can have_ side effects since we might resolve hostname literals during parsing. This means that for security-sensitive users `zeek --parse-only` is not the right tool to check for syntax errors since it can leak information during development, see #4216.

If we do not want to by default disable resolving of hostname literals via #4216 we should fix `--parse-only` to be free of side-effect.

awelzel:
Done with #4261

PR #4261, opened 2025-03-04
scan.l: Deprecate DNS resolution of hostname literals

### LINKED ISSUE #4216: Consider disabling hostname literal resolution by default
Currently Zeek will resolve hostname literals to a `set[addr]`, i.e., merely parsing a hostname might trigger a DNS request.

```zeek
event zeek_init()
	{
	print example.com;
	}
```

While this is a very powerful feature it also introduces a side channel which users might be unaware of; I suspect that for a majority of users the potential risks of this feature might not be worth it at all, and some security-conscious users might even want this feature unavailable by policy at their sites.

We should provide a way to make this feature toggleable (personal opinion: default to OFF and enabled explicitly). What makes this tricky is that the resolution happens during parsing already, so to control this with e.g., a redef'able constant via policy script might require some gymnastics.

awelzel:
This also skips DNS lookups when running with zeek --parse-only.

Closes #4216 #4219


src/scan.l
```diff
@@ -682,7 +683,19 @@ F	RET_CONST(zeek::val_mgr->False()->Ref())
 
 "0x"{HEX}+	RET_CONST(zeek::val_mgr->Count(static_cast<zeek_uint_t>(strtoull(yytext, 0, 16))).release())
 
-({H}".")+{HTLD}		RET_CONST(zeek::detail::dns_mgr->LookupHost(yytext).release())
+({H}".")+{HTLD} {
+	zeek::TableValPtr result = zeek::detail::dns_mgr->empty_addr_set();
+
+	if ( ! zeek::detail::parse_only ) {
+		const char *allow = getenv("ZEEK_DNS_ALLOW_SCRIPT_RESOLVE");
+		if ( ! zeek::fake_dns() && (! allow || std::string(allow) == "0") )
```
0xxon, line 691: Is the `zeek::fake_dns` case in here to allow tests to work without setting the environment variable, or is there a different reason?

awelzel: Mostly that and the fact that ZEEK_FAKE_DNS won't do any real DNS resolutions, so seemed okay to allow.

But it's a good point, I can also see ditching `zeek::fake_dns()` here and adapt the tests - seems more consistent - is that what you had in mind?

0xxon: Yup, that was where my question was going,mostly for consistency.

src/scan.l
```diff
@@ -682,7 +682,19 @@ F	RET_CONST(zeek::val_mgr->False()->Ref())
 
 "0x"{HEX}+	RET_CONST(zeek::val_mgr->Count(static_cast<zeek_uint_t>(strtoull(yytext, 0, 16))).release())
 
-({H}".")+{HTLD}		RET_CONST(zeek::detail::dns_mgr->LookupHost(yytext).release())
+({H}".")+{HTLD} {
+	zeek::TableValPtr result = zeek::detail::dns_mgr->empty_addr_set();
+
+	if ( ! zeek::detail::parse_only ) {
```
bbannier, line 688: Any reason we suppress the error with `--parse-only`? I used `--parse-only` for very basic linting in some editor integrations, and it would be great if we could surface it in that mode, before deployment.

src/scan.l
```diff
@@ -682,7 +682,19 @@ F	RET_CONST(zeek::val_mgr->False()->Ref())
...
+		const char *allow = getenv("ZEEK_DNS_ALLOW_SCRIPT_RESOLVE");
+		if ( ! allow || std::string(allow) == "0" )
+			zeek::reporter->Error("Prevented DNS lookup of '%s' at script parsing time. Set the ZEEK_DNS_ALLOW_SCRIPT_RESOLVE environment variable to enable this feature.", yytext);
```
bbannier, line 691: Nit: wondering whether we shouldn't hidfe `ZEEK_DNS_ALLOW_SCRIPT_RESOLVE` in the docs, and instead suggest one of the `lookup_hostname*` functions.

src/scan.l
awelzel: Hmm, fair. If one actually wants to use the feature, they'll need to inject the env variable into their development environment, but maybe that's their issue then.

This also behaves different with `--parse-only` now - though I don't think I care all that much :smiling_face_with_tear: 
```
@if ( |zeek.org| > 0 )
        print 1;
@endif
```

Let me ditch the parse_only special case!

src/scan.l
awelzel: I doubt this is used enough. It's also unclear if `lookup_hostname()` is the right tool (hard to use because it's async) - it might be input or config framework instead.

awelzel:
@vpax - after discussing this topic in a team call, we were thinking of deprecating and remove DNS lookups of host literals altogether with Zeek 8.1. This PR instead disables the feature by default, but provides a way to enable it again.

What's your take on this feature these days? Any gripes if it goes?

vpax:
I agree that the current form is too problematic vs. its utility. It would be helpful to know whether anyone uses it, I can imagine some do when initializing sets/tables of addresses, which was the original use-case. If so then there might be alternative syntax that eases the migration from the deprecated syntax

src/zeek.bif
```diff
@@ -4080,6 +4080,25 @@ function lookup_hostname%(host: string%) : addr_set
...
+##
+## .. zeek:see:: lookup_addr
+##
+## .. note::
+##
+##      This is a blocking call. You should use :zeek:see:`lookup_hostname`
+##      unless for initialization or testing purposes.
+##
+## .. zeek:see:: lookup_addr lookup_hostname
+function lookup_hostname_sync%(host: string%) : addr_set
```
bbannier, line 4097: Nit: like discussed offline we could use another name for this: #4163 seems to go with dedicated namespaces for async/blocking versions, but for this one we could just go to a truly ugly name since using this is a red flag, e.g., `lookup_hostname_blocking` or similar.

bbannier (approved):

src/zeek.bif
awelzel: I'll go with `blocking_lookup_hostname`, as we have `lookup_hostname_txt` with another suffix. Looks scary enough :-)

PR #4263, opened 2025-03-04
Add code of conduct and contributing to repo.

0xxon:
Adding this should add checkmarks next to the respective items of https://github.com/zeek/zeek/community.

Documents are direct copies from our wiki.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,1 @@
+Our code of conduct is published at https://zeek.org/community-code-of-conduct/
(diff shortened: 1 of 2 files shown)


timwoj (approved):
Are you planning to remove the versions from the wiki? My opinion is that they should go and these should be the sole versions.

0xxon:
I was not going to do that. I think we are linking to them from several places, and they are arguably a bit easier to discover.

Plus - this spans all ouf our projects, not just the Zeek repo.

timwoj:
> I was not going to do that. I think we are linking to them from several places, and they are arguably a bit easier to discover.

In that case, should these files just link to the Wiki pages instead of copying the text?

0xxon:
That's fair - I updated them to contain links.

awelzel (approved):

timwoj (approved):
Re-approving 😄

PR #4269, opened 2025-03-06
Allow to log speculative service in service field

J-Gras:
This implements an option to add the speculative service to conn.log's service field. The speculative service is logged using a "~" prefix.


J-Gras:
> For example, what does., "http,~http" mean?
> 
> The http analyzer was attached, but it also matched late once the signature buffer was exhausted and couldn't be attached?

Yes. However, the fact that the HTTP analyzer cannot be attached on a late match anymore is irrelevant, because it is already attached in this case. So, I admit adding `~http` while `http` is already present doesn't provide a real benefit.

I changed the logic to omit adding the speculative service to the service field, if it already contains the properly confirmed service. I've also extended the test to cover a case that adds `~http`.

awelzel:
> This implements an option to add the speculative service to conn.log's service field. The speculative service is logged using a "~" prefix.

My thinking - and chatting with parts of the team - it's not a good direction to stuff more special encoding into service. I'm not a fan of the violation `-` prefixing option already and filed it under "turn on historic behavior that hopefully isn't used much" feature.

Adding more of these mechanisms/options might give the impression that it's an established pattern to do so. I'ts not great for downstream users to have the contents of service vary structurally and semantically (?) based on a given Zeek's instance configuration. By default service contains a comma-separated list of analyzer/service names: This is easy, adding more entropy into service is not.

A potential `service_history` field could cover speculative-service, if you're up for that? :slightly_smiling_face:

J-Gras:
> A potential service_history field could cover speculative-service, if you're up for that? 🙂

I just implemented that based on a spontaneous suggestion. I'll defer to the idea originators to take on the debate.

awelzel:
> I just implemented that based on a spontaneous suggestion. I'll defer to the idea originators to take on the debate.

Closing as Jan mentioned via DM to wait for further input for a week, or else disregard for the time being. Thanks.

PR #4270, opened 2025-03-07
Fix printing type constants like `double`

### LINKED ISSUE #4268: "type" expressions don't have a printed form
Executing `print double;` in Master prints an empty string. Seems it should provide the name of the type.

evantypanski:
Fixes #4268


src/Val.cc
```diff
@@ -3202,7 +3202,13 @@ ValPtr EnumVal::DoClone(CloneState* state) {
...
 }
 
-void TypeVal::ValDescribe(ODesc* d) const { d->Add(type->AsTypeType()->GetType()->GetName()); }
+void TypeVal::ValDescribe(ODesc* d) const {
+    const auto& ty = type->AsTypeType()->GetType();
+    if ( ty->GetName().empty() )
+        d->Add(type_name(ty->Tag()));
+    else
+        d->Add(ty->GetName());
+}
```
awelzel, line 3211: WDYT about using `type->Describe(ODesc *)` instead?

evantypanski: Yeah I'll change to that. I figured for something tiny changing less would be preferred and I didn't want to change existing type vals, but turns out with a name a lot will just print the name anyway... so that's a wash :)

awelzel (approved):

Issue #4239, opened 2025-02-19
Spicy: Run rt functions on main stack instead of fiber


awelzel:
This is a long shot, but writing this down minimally for documentation. If this is too much out there, feel free to close, but it's basically a blocker of using Spicy and JavaScript safely together right now.

Using ZeekJS together with the spicy-zip analyzer results in "stack overflow" exceptions when hooking `file_state_remove` or other file related events triggered by the file manager.

The culprit here is that the V8 engine has it's own stack overflow detection, based on offsets in the main stack. The `rt::file_begin`, `rt::file_end`, `rt::file_data_in`, etc. functions may drain Zeek's event loop while switched to a fiber stack. Once that happens and JavaScript handlers are executed, the V8 stack overflow detection triggers as execution is happening on a completely different stack, aborting execution.

One attempt to fix this is to reset the stack offset when draining events: https://github.com/corelight/zeekjs/pull/98

This seems to work, but I've observed one other crash due to garbage collection runs happening on a fiber stack and GC doing something with stack pointers (unsure about about the details - but likely a similar culprit).

V8 offers a `v8::Locker` class that can be used to deal with this scenario when different threads are at play that access the same V8 isolate (the threads will have separate stacks as well), but the internals do require real threads.

I'm wondering if it would be possible to run execution of Zeek API calls (minimally the file manager related one due to the potential to drain the event loop) on the main stack instead. It shouldn't require a full fiber save/restore I believe, but instead "invoke this on the main stack", then resume execution on the fiber from which we came.

(slightly related could be zeek/spicy#1974, as stack based profiling would have less chance to be confused by the fiber stacks, at least for Zeek related code execution)

@rsmmr - thoughts?

awelzel:
> Oh, this is nasty ...

I know :-( We can chat in person a bit next week. After finding this all out, I find it a bit brave to run Zeek API functions on the fiber stacks, but I guess as long as there's nothing else in Zeek that's stack-sensitive, that'll work out.

I'm wondering if JavaScript execution could be done on a dedicated stack via `makecontext()` / `swapcontext()` actually. Not sure that would work and it's definitely also in the "nasty" department, though maybe not as bad as working with the fiber stacks.

> Before we go down that route, another idea: if we had a way to suppress the event draining while inside Spicy, we could instead drain immediately after returning from Spicy if needed.

[...]

PR #4273, opened 2025-03-10
JavaScript on separate thread

awelzel:
Run JavaScript on a separate thread for interoperability with Spicy
fiber stacks.

Closes https://github.com/zeek/zeek/issues/4239


testing/btest/javascript/file-sniff.js
bbannier, line 1: Like you wrote, this test passes, bumped `auxil/zeekjs` or not. Not really sure I agree with the reasoning from the commit message ("still makes sense to test these events") since we don't really want to test all events neither.

No strong opinion, so feel free to resolve without any action.

bbannier (approved):

PR #4274, opened 2025-03-10
Fix return types in docs for double_to_count and double_to_int

timwoj:
The comment blocks for these two methods got reversed somehow.


awelzel (approved):

timwoj:
I forgot to mention it, but thanks to @anthonykasza for pointing this out.

Issue #3917, opened 2024-09-05
Remove need for topology awareness with broker


awelzel:
I'm splitting this from #3597.

Prototyping ZeroMQ and NATS cluster backends, a `Cluster::publish(worker_topic, ...)` on a worker node is guaranteed to reach all online nodes subscribing to a `worker_topic` (how is an implementation detail - in these cases it's just a central broker/message bus).

The way Broker is configured by default in Zeek today does not provide this. IMO it should - users should not be exposed to the cluster topology.

---

              > To raise `request_message` on the workers, the common pattern I've seen is that on the manager node an explicit re-publish happens. And that pattern is what I'm stamping as topology awareness.

Yikes, that's not how Broker (or pub/sub in general) is supposed to work. I'm more than happy to discuss / dig into this, but let's not hijack this ticket. 🙂

_Originally posted by @Neverlord in https://github.com/zeek/zeek/issues/3597#issuecomment-1998021971_

PR #4282, opened 2025-03-13
Broker: Support "global" publish/subscribe

awelzel:
This introduces a new `Cluster::enable_global_pub_sub` option which will cause `cluster/setup-connections` to establish Broker peerings with all nodes of the same type, too.

In a 100 worker cluster, this essentially means that every node has 99 TCP connections to all the other workers.

Relates to #3917

---

Very drafty, it's something I had chatted about with Christian.


awelzel:
I'm going to close this for the time being - @ckreibich , this was an attempt to enable global pub/sub in broker by connecting nodes of the same type, too.

PR #4284, opened 2025-03-15
CI: Change macOS runner to Sequoia

timwoj:
Cirrus is reporting this for macOS builds:

> Only ghcr.io/cirruslabs/macos-runner:sequoia is allowed. Automatically upgraded.

This bumps our builds to Sequoia to resolve it.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -377,12 +377,12 @@ alpine_task:
 # Cirrus only supports the following macos runner currently, selecting
 # anything else automatically upgrades to this one.
 #
-#    ghcr.io/cirruslabs/macos-runner:sonoma
+#    ghcr.io/cirruslabs/macos-runner:sequoia
 #
 # See also: https://cirrus-ci.org/guide/macOS/
-macos_sonoma_task:
+macos_sequoia_task:
   macos_instance:
-    image: ghcr.io/cirruslabs/macos-runner:sonoma
+    image: ghcr.io/cirruslabs/macos-runner:sequoia
   prepare_script: ./ci/macos/prepare.sh
   << : *CI_TEMPLATE
   << : *MACOS_ENVIRONMENT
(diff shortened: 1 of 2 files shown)


awelzel:
I've pushed a fixup to this branch:
```
diff --git a/ci/init-external-repos.sh b/ci/init-external-repos.sh
index 7f27f582be..161791de59 100755
--- a/ci/init-external-repos.sh
+++ b/ci/init-external-repos.sh
@@ -51,9 +51,9 @@ if [[ -n "${CIRRUS_CI}" ]] && [[ "${CIRRUS_REPO_OWNER}" == "zeek" ]] && [[ ! -d
 
     banner "Trying to clone zeek-testing-private git repo"
     echo "${ZEEK_TESTING_PRIVATE_SSH_KEY}" >cirrus_key.b64
-    if [ "${CIRRUS_TASK_NAME}" == "macos_ventura" -o "${CIRRUS_TASK_NAME}" == "macos_sonoma" ]; then
+    if [[ "${CIRRUS_TASK_NAME}" =~ ^macos_ ]]; then
```

awelzel:
Grrr... The active http test fails and after a bit of staring, it's because the `httpd.py` scripts usage of `http.server` ends up doing a `getfqdn("127.0.0.1")` internally and that times out the test.

```
% python3 -c 'import socket, time; s=time.time(); print(socket.getfqdn("127.0.0.1")); e=time.time(); print(e - s)'
localhost
35.00473499298096
```

I don't know anything about OSX and DNS/hostname resolutions, but this seems more like a configuration issue? 35.0 seconds looks like a timeout of some sort. Tim, any clues? Or for Cirrus?

awelzel (approved):
@timwoj , if the follow up looks fine, feel free to merge. I've closed the ticket on the Cirrus side as the timeout doesn't happen right now anymore.

timwoj:
I googled around a bit regarding that lookup, and I think it could have been a configuration issue on their runner that they fixed. I guess we'll keep it in mind if it ever pops up again. DNS lookups are weird on macOS, since they use their own custom caching DNS server internally.

PR #4287, opened 2025-03-18
Add analyzer registration from VLAN to VNTAG

### LINKED ISSUE #4286: vntag inside vlan isn't handled
Received a pcap that isn't currently processed by zeek, it appears that we don't handle vntag inside of vlan.

[vntag_vlan_sandwich_clean.pcap.zip](https://github.com/user-attachments/files/19298675/vntag_vlan_sandwich_clean.pcap.zip)

Appears that we are missing

```
PacketAnalyzer::register_packet_analyzer(PacketAnalyzer::ANALYZER_VLAN, 0x8926, PacketAnalyzer::ANALYZER_VNTAG);
```

timwoj:
As @JustinAzoff suggested, we were missing this to allow going from VLAN to VNTAG. We already had the other way around.

Fixes #4286


awelzel:
Could we add the pcap from #4286 . It ends with `_clean.pcap`, so I guess that's good to use. @JustinAzoff ?

PR #4288, opened 2025-03-18
Bring in Redis analyzer

evantypanski:
this is a simple copy&paste of the [Redis analyzer repo](https://github.com/evantypanski/spicy-redis). I'm curious how CI handles it to start so it'll be a draft

I'm okay if this doesn't land in Zeek proper and instead this gets closed and it stays a separate package. This was just simple enough to bring up and get a PR in for visibility.

Still need to look over it again to see if there's anything I'd do differently. More/better events might be nice


evantypanski:
I'll just mark this ready but may add a bit of small tinkering (like to the Zeek scripts). I made a `spicy-decls.zeek` rather than `spicy-events.zeek` because there are some types from Spicy in the events, that could be possible to change but I like it for big records like for `SET`

This logs every command, which may be a little excessive. Events could use more refinement, too. Suggestions would be great.

There also was a bit of back and forth to get this to not trigger on a lot of traffic, since client commands can be completely unserialized and we'll never enumerate every possible command. Part of me worries a little about this catching non-Redis traffic or just spamming analyzer violations if DPD catches something, but that shouldn't be super common from the serialized server data in the DPD signature.

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,254 @@
...
+			{
+			if ( info$c?$redis_state )
+				{
+				info$c$redis_state$violation = T;
+				}
+			}
+		}
+	}
+
+function new_redis_session(c: connection): Info
```
ckreibich, line 85: An `Info` record is more like a request/response transaction, no? "Session" to me would suggest multiple transactions.

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+event analyzer_violation_info(atype: AllAnalyzers::Tag,
+    info: AnalyzerViolationInfo)
+	{
+	if ( atype == Analyzer::ANALYZER_SPICY_REDIS )
+		{
+		if ( info?$c )
+			{
+			if ( info$c?$redis_state )
```
ckreibich, line 77: I'd suggest compressing things like this into a single condition until we need to separate them. The reason is performance — more code means more work for the interpreter (which here really might be trivial, but it's something to keep in mind about Zeek scripting in general, particularly when adding function calls).

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,254 @@
...
+		delete c$redis_state$pending[c$redis_state$current_response];
+		}
+	}
+
+hook finalize_redis(c: connection)
+	{
+	if ( c$redis_state$violation )
+		{
+		# If there's a violation, make sure everything gets deleted
+		delete c$redis_state;
```
ckreibich, line 241: I don't believe you need this. `finalize_redis` is a removal hook, so we already know that the connection is about to get thrown away, which'll take care of the state connected to it.

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,254 @@
...
+		## Timestamp for when the activity happened.
+		ts: time &log;
+		## Unique ID for the connection.
+		uid: string &log;
+		## The connection's 4-tuple of endpoint addresses/ports.
+		id: conn_id &log;
+		## The Redis command.
+		cmd: Command &log;
+		## The response for the command.
+		response: ServerData &log &optional;
```
ckreibich, line 26: This is nitpicky Evan, apologies — looking around other analyzers we usually either say "request/response" or "command / reply". You actually use "request" and "response" in the state record below. The Redis docs actually match this pattern, too!

And, for logs there's the fact that fields get flattened out, so these things become prefixes to serveral log fields. This currently leads to `cmd.command`, which is a bit awkward. So I'm landing on a mild preference for "request"/"response", but don't feel strongly.

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,254 @@
...
+	global finalize_redis: Conn::RemovalHook;
+
+	type State: record {
+		## Pending requests.
+		pending: table[count] of Info;
+		## Current request in the pending queue.
+		current_request: count &default=0;
+		## Current response in the pending queue.
+		current_response: count &default=0;
+		## Ranges where we do not expect a response.
```
ckreibich, line 41: Would help here to say why — I am no longer sure... :-) Would it be better to have a record to express a range explicitly than via vectors? That would simplify `is_last_interval_closed()` too.

scripts/base/protocols/redis/spicy-decls.zeek
```diff
@@ -0,0 +1,94 @@
+##! Events and records generated by the Redis analyzer.
+
```
ckreibich, line 2: If you put this in the `Redis` module then you can skip the `Redis::` prefixes below, which is a bit more idiomatic. I actually didn't know that exports work this way, too!

ckreibich (changes requested):
Okay Evan here's a first round of feedback to get this going again — sorry again for the delay here! Let me know your thoughts. I'll stay on it this time, oath.

I have two minor requests regarding the pcaps:
- Some of them are a little on the large side (>200KB). Nothing terrible, and we have a handful of larger ones, but it'd be good to check if there's an easy way to cut down them down if they don't have to be that big.
- The test cases shouldn't do `-C` to ignore checksums out of convenience. We should instead fix the checksums in the pcaps. An easy way to do this is to run them through `tcprewrite -C -i <input> -o <output>`.

scripts/base/protocols/redis/main.zeek
evantypanski: I think the point was to not log pending commands if it gets to the finalize hook, but it's kind of a roundabout way to do it. I turned it into a return

ckreibich (approved):
Ship it! :rocket:

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,258 @@
...
+		## Timestamp for when the activity happened.
+		ts: time &log;
+		## Unique ID for the connection.
+		uid: string &log;
+		## The connection's 4-tuple of endpoint addresses/ports.
+		id: conn_id &log;
+		## The Redis command.
+		request: Command &log;
+		## The response for the command.
+		response: ServerData &log &optional;
```
awelzel, line 26: Given the spec says `command` and `reply`, IMO we should stick with those names rather than introducing `request` and `response`? Maybe shorten `command` to `cmd`.

Could see making `Command.command` field `Command.name`, so it ends up as `cmd.name` in the log.

scripts/base/protocols/redis/spicy-decls.zeek
```diff
@@ -0,0 +1,97 @@
...
+		## The value, if this command is known to have a value
+		value: string &log &optional;
+		## The command in an enum if it was known
+		known: KnownCommand &optional;
+	};
+
+	## Generic server data returned from the server.
+	type ServerData: record {
+		## Was this an error?
+		err: bool &log;
```
awelzel, line 58: For discussion, maybe? IMO at least in the log this should be inverted to a `success` field (possibly within `Info` rather than in `Reply` so that the command/reply pair indicates success or failure)

If a user wants to query for successful responses using this log, their query is going to be `reply.err=F`, which doubly negative and makes me shudder (might be a me problem, too). It's also nicer to see `"success": true` in a log, rather than `"err": false`.

For the case of "no clue what happened with the command", it seems better to have "reply.success" unset, rather than "reply.err" unset.

There's other logs that have an artificial `success` field (e.g. mysql). `err` or `error` seem less common.

awelzel (commented):
I looked through the events and log a bit and have a few comments. I think they are more than just nitpicking, but could consider another iteration if you'd rather move this in now after Christian's approval.

scripts/base/protocols/redis/main.zeek
awelzel: Ah, I see Christian asked the same :-)

scripts/base/protocols/redis/main.zeek
```diff
@@ -0,0 +1,258 @@
...
+		{
+		# All 3 CLIENT commands we care about have 3 elements
+		if ( |command$raw| == 3 )
+			{
+			if ( to_lower(command$raw[2]) == "on" )
+				{
+				# If the last range is open, close it here. Otherwise, noop
+				if ( |c$redis_state$no_response_ranges| > 0 )
+					{
+					local range = c$redis_state$no_response_ranges[|c$redis_state$no_response_ranges|
```
awelzel, line 155:
```suggestion
					local range = c$redis_state$no_response_ranges[-1];
```

Thought this works today.

scripts/base/protocols/redis/spicy-decls.zeek
evantypanski: Yeah I think success seems clearer, it's mostly just easier to think about an error field getting set when making the parser, but the user would find the opposite more intuitive, so it's probably a good swap

src/analyzer/protocol/redis/resp.evt
```diff
@@ -0,0 +1,21 @@
+# See the file "COPYING" in the main distribution directory for copyright.
+
+protocol analyzer spicy::Redis over TCP:
```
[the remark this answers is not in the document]
evantypanski: I recently learned about this convention and fixed the package template/docs, but not here, oops

https://github.com/zeek/package-template/issues/47

scripts/base/protocols/redis/spicy-decls.zeek
```diff
@@ -0,0 +1,97 @@
...
+##
+## command: The command sent to the server.
+global client_command: event(c: connection, command: Command);
+
+## Generated for every response sent by the Redis server to the client.
+##
+## c: The connection.
+##
+## data: The server data sent to the client.
+global server_data: event(c: connection, data: ServerData);
```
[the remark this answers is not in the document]
evantypanski: > Further, could consider dropping the err field in ServerData in favor of different events: Redis::reply(data: ServerData (ReplyData?)) (implies success) and Redis::error(data: string). There's e.g. smb1_error(), mysql_error() or krb_error() events.

I did consider it a bit, but I wanted all server data to get sent to a single event because it made `CLIENT REPLY` easier to handle. Maybe I'll tinker with it, splitting out might not be that unclean

src/analyzer/protocol/redis/redis.spicy
```diff
@@ -0,0 +1,369 @@
...
+function parse_command(raw: vector<bytes>): Command {
+    assert |raw| >= 1;
+    local cmd = command_from(raw[0]);
+    local parsed: Command = [$raw = raw, $command = raw[0], $key = Null, $value = Null, $known = cmd];
+    if (!cmd)
+        return parsed;
+
+    if (|raw| >= 2) {
+        switch (*cmd) {
+            case KnownCommand::KEYS:
```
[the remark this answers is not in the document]
evantypanski: Well, extending the command would mean

1) Potentially having a command with a space (... currently mostly unhandled 🙊 )
2) Defining which spot (if any) has a source key based on the command
3) Defining which spot (if any) has a destination key based on the command

Then this has to be tokenized by Spicy before getting sent to scriptland, so Zeek would have to "parse" a stream of tokens into commands itself (which may have different formats depending on if it's inline or not)

I think it could be put into scriptland, but unfortunately it's not just a simple map from the first word to the command, it's really a second stage of parsing - hence why it's in Spicy. Maybe one day. It seems to conceptually belong in Spicy to me, but being able to extend the recognized commands seems really nice too

src/analyzer/protocol/redis/resp.evt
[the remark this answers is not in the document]
ckreibich: Oh interesting, I wasn't aware of the effect of the `spicy::` prefix here. Indeed, the service field entries should never have the `spicy_` prefix. (This reminded me of #4113, which I thought was the whole story about those prefixes.)

scripts/base/protocols/redis/main.zeek
ckreibich: Yeah, since Zeek 6. I just stumbled over that yesterday when trying to write code that works with 5.2! :-)

evantypanski:
@ckreibich sorry I went back in and [implemented more](https://github.com/zeek/zeek/pull/4288/commits/dcd4942762b9a032bd4e91ac4f78323e7d52d7a1), hope you don't mind!!

evantypanski:
Per discussion with @awelzel, the fixup commit at the end moves replies back into a record for 3 main reasons:

1) Backwards compatibility in case we later handle/log arrays or maps in some capacity
2) Nicer on logs to know it's the reply data (and so that it's internally consistent with command)
3) Allows unset strings rather than empty strings when the data is a map or array or something

testing/btest/Baseline/scripts.base.protocols.redis.pubsub/redis.log
```diff
@@ -0,0 +1,13 @@
...
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	redis
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	cmd.name	cmd.key	cmd.value	success	reply.value
+#types	time	string	addr	port	addr	port	string	string	string	bool	string
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	127.0.0.1	56162	127.0.0.1	6379	SUBSCRIBE	-	-	T	-
+XXXXXXXXXX.XXXXXX	ClEkJM2Vm5giqnMf4h	127.0.0.1	56163	127.0.0.1	6379	PUBLISH	-	-	T	-
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	127.0.0.1	56162	127.0.0.1	6379	-	-	-	T	-
```
awelzel, line 12: For a follow-up, what's going on here?

testing/btest/Baseline/scripts.base.protocols.redis.django-cloud/redis.log
```diff
@@ -0,0 +1,168 @@
...
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	SET	:1:factorial_47	258623241511168180642964355153611979969197632389120000000000	T	OK
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	SET	:1:factorial_48	12413915592536072670862289047373375038521486354677760000000000	T	OK
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	SET	:1:factorial_49	608281864034267560872252163321295376887552831379210240000000000	T	OK
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	SET	:1:factorial_50	30414093201713378043612608166064768844377641568960512000000000000	T	OK
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	SET	:1:factorial_50	30414093201713378043612608166064768844377641568960512000000000000	T	30414093201713378043612608166064768844377641568960512000000000000
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	GET	:1:factorial_50	-	T	30414093201713378043612608166064768844377641568960512000000000000
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.4	50044	18.234.186.95	10625	GET	:1:factorial_50	-	T	(empty)
```
awelzel, line 65: Just skimming, think there's something with the command/reply correlation off: Wireshark indicates `GET` returns a value and the `SET` above in line 63 doesn't return a value.

Screenshot from 2025-05-24 15-50-33

testing/btest/Baseline/scripts.base.protocols.redis.stream/redis.log
```diff
@@ -0,0 +1,14 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	redis
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	cmd.name	cmd.key	cmd.value	success	reply.value
+#types	time	string	addr	port	addr	port	string	string	string	bool	string
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	127.0.0.1	49992	127.0.0.1	6379	XADD	-	-	T	1729622832637-0
```
awelzel, line 10: It seems XADD has a key, though the value might be complex. We likely don't yet parse the key?

awelzel (approved):
:ship:

The logs look like there's something not fully right with cmd/reply correlation, but probably easier to deal with and review as follow-up PRs.

Could you also add a NEWS entry about the analyzer and the log? Can see QUIC, LDAP or Postgres entries for a bit of inspiration, but definitely add your own spin to it!

testing/btest/Baseline/scripts.base.protocols.redis.pubsub/redis.log
evantypanski: an array `message` from the server on the subscriber's connection

```
@data 127.0.0.1-56162-127.0.0.1-6379-tcp-resp 48
*3
$7
message
$10
my_channel
$8
hello :)
```

It's an edge case that I... haven't quite looked into how it needs to be handled :) It might just be best to immediately log them, I don't think they should be correlated with `CLIENT REPLY` stuff and it'll break that. If that's blocking I can hold

PR #4289, opened 2025-03-18
Update banner parsing of SSH analyzer

0xxon:
The banner parsing of the SSH analyzer at the moment is both a bit too open, and a bit too restrictive.

Specifically, the SSH analyzer currently will happily parse any protocol that starts with two lines of text, and assume the lines are versions. This is obviously not optimal, as it will pollute the ssh log with non-ssh entries.

On the other hand, the SSH analyzer does not correctly deal with cases where the SSH server does not send the banner in the first line of the connection.

This PR updates the parsing so that:
* the format of SSH banners is enforced. Specifically an SSH banner needs to start with `SSH-`, otherwise it is not valid. The SSH parser itself requires this format to continue parsing.
* on the server side, the SSH analyzer will accept lines that are sent before the banner. This is not common, but is supported by the protocol. This also adds a new event `ssh_server_pre_banner_data` that contains the data in such pre-banner-lines

This change means that non-ssh-banners no longer show up in the `ssh_server_version` and `ssh_client_version` events. I was not sure if it might be interesting to be able to add a new event that show the non-client-banner that's passed in to the SSH analyzer, to find what non-ssh-protocols are spoken on port 22. If that's desired, I am happy to add such an event. 

This requires merging of the private test suite.


0xxon:
As a side-note - I also created a package that should log servers that output data before a normal SSH handshake.

That's at https://github.com/0xxon/zeek-sshprebannermessage

src/analyzer/protocol/ssh/events.bif
```diff
@@ -336,3 +336,14 @@ event ssh2_gss_init%(c: connection, is_orig: bool%);
...
 event ssh2_rsa_secret%(c: connection, is_orig: bool%);
+
+## SSH servers can send textual data to the client before sending
+## a banner. The primary use case of this are error messages of TCP
+## wrappers.
+##
+## As this event happens before the SSH banner is exchanged, it is
+## possible that it contains data from different protocols; e.g. if
+## an SSH client connects to a non-SSH-server.
+##
```
awelzel, line 347: For documentation:

```
c: The connection.

data: The pre-banner data.
```

src/analyzer/protocol/ssh/ssh-protocol.pac
```diff
@@ -20,16 +20,32 @@ proc: bool = $context.connection.inc_encrypted_byte_count_in_current_segment();
...
+
+# SSH servers can have banners before their SSH version. Which... fun.
+type SSH_Version_Server = record {
+	version: RE/(SSH-.*)?/;
+	# only UTF-8 data. This doesn't catch all bad cases, but some
+	nonversiondata: RE/([^[\xC0-\xC1]|[\xF5-\xFF]])*/;
+} &let {
+	update_state   : bool = $context.connection.update_state(KEX_INIT, false) &if(sizeof(version) > 0);
+	update_version : bool = $context.connection.update_version(version, false) &if(sizeof(version) > 0);
+} &oneline;
```
awelzel, line 41: Fancy :-)

testing/btest/Baseline/scripts.base.protocols.ssh.pre-banner/.stdout
```diff
@@ -0,0 +1,36 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+ _____________
+< Hi stranger >
```
awelzel, line 3: Hi!

awelzel (approved):
Nice!

src/analyzer/protocol/ssh/events.bif
0xxon: Thanks - no idea how I missed that.

src/analyzer/protocol/ssh/ssh-analyzer.pac
```diff
@@ -58,19 +58,26 @@ const char* fingerprint_md5(const unsigned char* d)
...
+	function proc_ssh_version_server(msg: SSH_Version_Server): bool
+		%{
+		if ( ${msg.version}.length() > 0 )
 			{
-			zeek::BifEvent::enqueue_ssh_client_version(connection()->zeek_analyzer(),
+			zeek::BifEvent::enqueue_ssh_server_version(connection()->zeek_analyzer(),
 				connection()->zeek_analyzer()->Conn(),
 				to_stringval(${msg.version}));
 			}
-		else if ( ssh_server_version )
```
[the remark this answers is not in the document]
0xxon: You are right, there is no specific reason to do that.

They will always be true as used in our default scripts, but *shrug*.

PR #4297, opened 2025-03-19
Quick follow-ups for storage framework merge

timwoj:
These are few quick follow-up items from the storage framework merge yesterday:

- Add `libhiredis-dev` to the set of packages installed for the generate-docs workflow. Otherwise, the Redis docs don't get included because the backend doesn't get built.
- Remove installing `expected-lite` for plugins, which was breaking OBS.
- Fix some minor Coverity findings in the SQLite backend.
- Merge the `OperationResultCallback` code into `ResultCallback`, and remove the former. `OpenResultCallback` was just a super-set of `OperationResultCallback` and so `OperationResultCallback` doesn't need to exist.
- Add NEWS entry and removed giant block of commits from CHANGES
- Fix a couple of data races with non-native expiration reported by tsan
- Removed unnecessary type aliases in the btests, which I never took care of after https://github.com/zeek/zeek/pull/4194 got merged.
- Fixed SQLite `put` operations when overwriting is disabled so it returns the right status code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -354,10 +354,6 @@ target_include_directories(
     zeek_dynamic_plugin_base SYSTEM
     INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/prometheus-cpp/include>)
 
-target_include_directories(
-    zeek_dynamic_plugin_base SYSTEM
-    INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/expected-lite/include>)
-
 # Convenience function for adding an OBJECT library that feeds directly into the
 # main target(s).
 #
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,27 @@ New Functionality
 
 - The new ``is_valid_subnet()`` function mirrors ``is_valid_ip()``, for subnets.
 
+- A new Storage framework was merged into the Zeek tree. The intention with this framework
+  is to eventually replace the storage functionality that Broker provides, including
+  direct storage via calls such as ``Cluster::create_store`` and ``Broker::put_unique`` as
+  well as storage-backed tables via the ``&backend`` attribute. This is an initial version
+  for testing, and will be expanded upon in the future. The current state of the framework
+  is as follows:
+
+  - A new API was added for storage backend plugins.
+
+  - Script-level functions for opening and closing backends, and insertion, retrieval, and
+    erasure of elements are available.
+
+  - Backends can support both asynchronous mode (using ``when`` statements) and
+    synchronous mode (blocking until the operation copmletes). BIF methods were added
+    under new ``Storage::Async`` and ``Storage::Sync`` modules for these two modes. The
+    modes can be used interchangeably with the same backend handle.
+
+  - SQLite and Redis backends exist in the Zeek tree by default. We are working on a
+    backend for NATS that will be available as an external plugin, but it is not quite
+    ready yet. Both of the existing backends support usage in a cluster environment.
+
 Changed Functionality
 ---------------------
 
--- a/src/storage/Manager.cc
+++ b/src/storage/Manager.cc
@@ -89,7 +89,7 @@ OperationResult Manager::OpenBackend(BackendPtr backend, OpenResultCallback* cb,
     return res;
 }
 
-OperationResult Manager::CloseBackend(BackendPtr backend, OperationResultCallback* cb) {
+OperationResult Manager::CloseBackend(BackendPtr backend, ResultCallback* cb) {
     // Expiration runs on a separate thread and loops over the vector of backends. The mutex
     // here ensures exclusive access. This one happens in a block because we can remove the
     // backend from the vector before actually closing it.
--- a/src/storage/Manager.cc
+++ b/src/storage/Manager.cc
@@ -23,7 +23,7 @@ void detail::ExpirationTimer::Dispatch(double t, bool is_expire) {
     // in the interim.
     if ( ! expire_running.test_and_set() ) {
         DBG_LOG(DBG_STORAGE, "Starting new expiration thread");
-        storage_mgr->expiration_thread = std::jthread([]() { storage_mgr->Expire(); });
+        storage_mgr->expiration_thread = std::jthread([t]() { storage_mgr->Expire(t); });
     }
 
     storage_mgr->StartExpirationTimer();
@@ -107,17 +107,16 @@ OperationResult Manager::CloseBackend(BackendPtr backend, ResultCallback* cb) {
     return res;
 }
 
-void Manager::Expire() {
+void Manager::Expire(double t) {
     // Expiration runs on a separate thread and loops over the vector of backends. The mutex
     // here ensures exclusive access.
     std::unique_lock<std::mutex> lk(backends_mtx);
 
     DBG_LOG(DBG_STORAGE, "Expiration running, have %zu backends to check", backends.size());
 
-    double current_network_time = run_state::network_time;
     for ( auto it = backends.begin(); it != backends.end() && ! run_state::terminating; ++it ) {
         if ( (*it)->IsOpen() )
-            (*it)->Expire(current_network_time);
+            (*it)->Expire(t);
     }
 
     expire_running.clear();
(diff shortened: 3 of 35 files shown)


src/storage/Manager.cc
```diff
@@ -114,8 +114,6 @@ void Manager::Expire(double current_network_time) {
     // here ensures exclusive access.
     std::unique_lock<std::mutex> lk(backends_mtx);
 
-    DBG_LOG(DBG_STORAGE, "Expiration running, have %zu backends to check", backends.size());
-
```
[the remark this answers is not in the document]
timwoj: I don't think I've ever encountered having to run DBG_LOG in a separate thread, or at least not since we've had tsan to complain about. It probably should be thread-safe though.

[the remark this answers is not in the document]
timwoj: I opened https://github.com/zeek/zeek/issues/4302.

CMakeLists.txt
```diff
@@ -354,10 +354,6 @@ target_include_directories(
     zeek_dynamic_plugin_base SYSTEM
     INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/prometheus-cpp/include>)
 
-target_include_directories(
-    zeek_dynamic_plugin_base SYSTEM
-    INTERFACE $<INSTALL_INTERFACE:include/zeek/3rdparty/expected-lite/include>)
-
```
awelzel, line 360: The commit title is misleading if you wanted to fix that up. We continue to install it (and need to as the files are included from util.h), the change here removes adding a non-existing directory to dynamic plugin's target include_directories which failed to compile.

NEWS
```diff
@@ -29,6 +29,28 @@ New Functionality
 
 - The new ``is_valid_subnet()`` function mirrors ``is_valid_ip()``, for subnets.
 
+- A new Storage framework was merged into the Zeek tree. The intention with this framework
+  is to eventually completely replace the storage functionality that Broker provides,
```
awelzel, line 33: Is the "eventually completely" needed here?

NEWS
```diff
@@ -29,6 +29,28 @@ New Functionality
...
+  ``Broker::put_unique`` as well as storage-backed tables via the ``&backend``
+  attribute. This is an initial version for testing, and will be expanded upon in the
+  future. The current state of the framework is as follows:
+
+  - Functions for opening and closing backends, and insertion, retrieval, and erasure of
+    elements are available.
+
+  - Backends can support both asynchronous mode (using ``when`` statements) and
+    synchronous mode (blocking until the operation copmletes). BIF methods were added
+    under new ``Storage::Async`` and ``Storage::Sync`` namespaces for these two modes. The
```
awelzel, line 44: namespaces -> modules.

NEWS
```diff
@@ -29,6 +29,28 @@ New Functionality
...
+    modes can be used interchangeably with the same backend handle.
+
+  - SQLite and Redis backends exist in the Zeek tree by default. We are working on a
+    backend for NATS that will be available as an external plugin, but it is not quite
+    ready yet. Both of the existing backends support usage in a cluster environment.
```
awelzel, line 49: I wouldn't mention NATS at all here. It could be more interesting to mention that alternative backends can be provided via plugins as that isn't quite mentioned.

NEWS
```diff
@@ -29,6 +29,28 @@ New Functionality
...
+
+  - Detailed documentation is forthcoming, although documentation for the script-level
```
awelzel, line 51: Also not sure this is important to mention in NEWS :-)

src/storage/Manager.cc
```diff
@@ -23,7 +23,9 @@ void detail::ExpirationTimer::Dispatch(double t, bool is_expire) {
     // in the interim.
     if ( ! expire_running.test_and_set() ) {
         DBG_LOG(DBG_STORAGE, "Starting new expiration thread");
-        storage_mgr->expiration_thread = std::jthread([]() { storage_mgr->Expire(); });
+        double current_network_time = run_state::network_time;
```
awelzel, line 26: How about using `t` that is given to the timer's Dispatch() instead? I think they end up the same anyway right now.

src/storage/Manager.h
```diff
@@ -77,8 +77,10 @@ class Manager final : public plugin::ComponentManager<Component> {
      * Runs an expire operation on all open backends. This is called by the expiration
      * timer and shouldn't be called directly otherwise, since it should only happen on a
      * separate thread.
+     *
+     * @param current_network_time The network time that the expiration started.
```
awelzel, line 81: I'd be okay calling this just `t` as for `Dispatch()` instead of `current_network_time` . The current network time is  what you'd most likely pass as an argument, but doesn't need to be.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -237,12 +237,18 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
 void SQLite::DoExpire(double current_network_time) {
     auto stmt = expire_stmt.get();
 
-    if ( auto res = CheckError(sqlite3_bind_double(stmt, 1, current_network_time)); res.code != ReturnCode::SUCCESS ) {
-        sqlite3_reset(stmt);
-        // TODO: do something with the error here?
+    int status = sqlite3_bind_double(stmt, 1, current_network_time);
+    if ( status != SQLITE_OK ) {
+        // TODO: do something with the error?
```
awelzel, line 242: I guess? This shouldn't happen and we run at a rather low frequency IIUC, so maybe `fprintf(stderr, ...` at least?

Mind putting something in the commit message what the data race actually was? Something in `Check()` or `Step()` ?

testing/btest/scripts/base/frameworks/storage/redis-sync.zeek
```diff
@@ -30,27 +30,31 @@ event zeek_init()
...
-	if ( res2$code == Storage::SUCCESS && res2?$value )
-		print "get result same as inserted", value2 == ( res2$value as string );
+	res = Storage::Sync::get(b, key);
+	print "get result", res;
+	if ( res$code == Storage::SUCCESS && res?$value )
+		print "get result same as inserted", value2 == ( res$value as string );
+
+	# Validate that $overwrite=F returns KEY_EXISTS
+	res = Storage::Sync::put(b, [ $key=key, $value=value2, $overwrite=F ]);
+	print "no overwrite put result", res;
```
awelzel, line 58: Could make sense to do another `get()` and ensure the value really wasn't overwritte (but would need something else but `$value=value2` as that just writes the same value again.).

testing/btest/scripts/base/frameworks/storage/overwriting.zeek
```diff
@@ -7,24 +7,35 @@
...
+	local key = "key1111";
 	local value = "value7890";
 
-	local open_res = Storage::Sync::open_backend(Storage::SQLITE, opts, string, string);
+	local open_res = Storage::Sync::open_backend(Storage::SQLITE, opts, str, str);
 	print "open result", open_res;
 	local b = open_res$value;
 
-	local res = Storage::Sync::put(b, [$key=key, $value=value]);
+	local res = Storage::Sync::put(b, [$key=key, $value=value, $overwrite=F]);
```
awelzel, line 24: Probably should use a different value here just in case the overwrite mistakenly happens, line 30 will succeed.

testing/btest/scripts/base/frameworks/storage/overwriting.zeek
```diff
@@ -7,24 +7,35 @@
...
+	print "put result", res;
+
+	res = Storage::Sync::get(b, key);
+	print "get result", res;
+	if ( res$code == Storage::SUCCESS && res?$value )
+		print "get result same as inserted", value == (res$value as string);
+
+	res = Storage::Sync::put(b, [$key=key, $value=value, $overwrite=T]);
```
awelzel, line 32: This is overwriting with the same value that's already stored? Would make sense to overwrite with something new and check that it actually made it through.

testing/btest/plugins/storage.zeek
```diff
@@ -10,9 +10,6 @@
 
 @load base/frameworks/storage/sync
 
-# Create a typename here that can be passed down into get().
```
awelzel, line 13: Nice!

awelzel (approved):
I'm approving, but have a few things that might be worth addressing.

PR #4298, opened 2025-03-19
Remove giant spam of CHANGES entries from storage framework merge

timwoj:
As suggested by @0xxon. I'm fine with squashing all of it down to one line, but I'm opening the PR in case anyone has other opinions.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -31,121 +31,7 @@
 
 7.2.0-dev.378 | 2025-03-18 11:43:48 -0700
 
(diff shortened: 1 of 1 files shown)


0xxon:
I think we should perhaps mention a bit what this entails. At least in the past, people used the changes file to be able to read over it to see what has changed in Zeek; thus I think if possible some short description of what this allow would be neat. Or a pointer to the NEWS file, or documentation if that's already in there.

timwoj:
Moved this to be part of https://github.com/zeek/zeek/pull/4297, along with a NEWS update.

PR #4299, opened 2025-03-19
CI: Run btests on macOS under sudo

timwoj:
Due to issues with DNS lookups on macOS, one of the Cirrus support people recommended we run our tests as root. See https://github.com/cirruslabs/cirrus-ci-docs/issues/1302 for more details.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -7,6 +7,13 @@
 result=0
 BTEST=$(pwd)/auxil/btest/btest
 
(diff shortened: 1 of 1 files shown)


timwoj:
FWIW, a build passed with this: https://cirrus-ci.com/task/5109687437230080. We've seen it pass periodically without any changes though too.

timwoj:
> I guess 🤷‍♂️

Yeah. I'm not a huge fan of this but it was suggested by our CI provider and it appears to fix the problem. At least it's limited to macOS only.

PR #4305, opened 2025-03-21
Include OUI in identifier when forwarding out of SNAP analyzer

### LINKED ISSUE #4281: SNAP: Support OUIs other than 0
The built-in SNAP analyzer silently ignores PDUs with OUIs other than 0:
https://github.com/zeek/zeek/blob/1b655836be5e988bb2019c0920633e5ade43d745/src/packet_analysis/protocol/snap/SNAP.cc#L39-L45
In Slack the [Cisco Discovery Protocol (CDP)](https://wiki.wireshark.org/CDP) came up as an example that uses a different OUI (`0x00000C`). Having support to multiplex different OUIs would be nice. At least we should leave some trace that packets with different OUIs are not further processed. They don't show up in `unknown_protocols.log`.

timwoj:
There's no reason to check the OUI before forwarding here, since it keeps unprocessed packets from showing up in unknown_protocols.log. Instead let the packet analysis registrations handle the protocols as normal.

Fixes #4281


src/packet_analysis/protocol/snap/SNAP.cc
```diff
@@ -30,17 +30,10 @@ bool SNAPAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
...
     len -= 5;
 
-    if ( oui == 0 ) {
-        // If the OUI is zero, the protocol is a standard ethertype and can be
-        // forwarded as such.
-        return ForwardPacket(len, data, packet, protocol);
-    }
-
-    return true;
+    return ForwardPacket(len, data, packet, protocol);
```
awelzel, line 38: The next packet analyzer doesn't have access to `oui` anymore without looking back, so not clear how one would install a custom analyzer for `oui=0x00000C`, for example.

Maybe that's not meant to be fixed here, but wonder if this really needs an extra SNAP_OUI "analyzer layer" to dispatch based on the `oui` - with `0` directly using the following `protocol`.

timwoj: That's the point of this PR. The SNAP header includes the protocol of the next header, so we don't need to look at the OUI to figure out where to forward it to. For example, in the CDP pcap I added the next protocol from the SNAP header is 0x2000 (which is CDP). We don't support that, but forwarding to it anyways causes the unknown-protocols code to kick in and log it correctly.

awelzel: My understanding is the 0x2000 is specifically allocated by Cisco and is only valid if `oui=0x00000C`, Nothing prevents a separate organization to allocate 0x2000 for something else - I guess, that's actually the purpose.

https://www.iana.org/assignments/ieee-802-numbers/ieee-802-numbers.xhtml#ieee-802-numbers-2

> If the OUI is zero, the protocol ID is the registered [EtherType](https://en.wikipedia.org/wiki/EtherType) value for the protocol running on top of SNAP. If the OUI is an OUI for a particular organization, the protocol ID is a value assigned by that organization to the protocol running on top of SNAP.

https://en.wikipedia.org/wiki/Subnetwork_Access_Protocol

Not sure we'll see anything in practice, but ignoring OUI seems limited.

timwoj: See if that new commit makes more sense. It changes all of the identifiers in the packet analyzers to 64-bit, and uses a combination of the OUI and protocol for forwarding out of SNAP.

The functions in `Analyzer` that have to change aren't in a detail namespace, but they're not virtual so they shouldn't be overridden by child classes. The `Dispatcher` class is in the detail namespace so those changes are fine.

awelzel: >  It changes all of the identifiers in the packet analyzers to 64-bit

I was wondering if that could be one approach. It makes sense to me.

However, it'd be just as reasonable to do a  two-layer registration where oui is the first identifier, and then do something like the following, where the "OUI analyzers" are essentially all the same, but the hierarchy is represented.

```
register_packet_analyzer(SNAP, 0x000000, SNAP_OUI_ETHER_TYPE);
register_packet_analyzer(SNAP, 0x00000C, SNAP_OUI_CISCO);

register_packet_analyzer(SNAP_OUI_CISCO, 0x2000, Analyzer::CISCO_CDP);
register_packet_analyzer(SNAP_OUI_ETHER_TYPE, 0x0800, ANALYZER_IP)

# Instead of the following:
register_packet_analyzer(SNAP, 0x0C2000, CISCO_CDP);
register_packet_analyzer(SNAP, 0x00800, ANALYZER_IP)

```

This seems a bit of a design call that'd be good to run by @J-Gras / @ckreibich or maybe the team?

Not sure the uint32 to uint64 is a big deal. The lookup tables are now twice the size, though not sure we were ever concerned about that.

J-Gras: I agree that the two-layer approach would be the "conceptually clean" solution.

timwoj: So you'd have to have second set of analyzers for handling the intermediate types? Or are you thinking of a second version of `register_packet_analyzer` that takes a numeric value for the third argument instead of a `PacketAnalyzer::Tag`?

awelzel (approved):
After chatting in a call and Christian's explanation that oui and identifer are together, extending the identifier seems just as reasonable. Approving.

It'd be good to point in NEWS that the identifier was extended and maybe provide an example somewhere how this can be used with SNAP. E.g. `register(..., 0x0C2000, ...)`

J-Gras:
Just as a site note: Going with the two-layer approach would allow to use different default analyzers for different OUIs. However, I don't have a need to do so.

PR #4306, opened 2025-03-24
Add attribute to allow omitting a value from docs generation

### LINKED ISSUE #4190: Zeekygen: Add ability to ignore a script file
This file generated by zeekygen is currently importing the entire Mozilla CA list into our documentation: https://docs.zeek.org/en/master/scripts/base/protocols/ssl/main.zeek.html#id-SSL::root_certs. That's probably not necessary, but zeekygen doesn't have any mechanism for ignoring files.

timwoj:
This PR adds a new `&docs_omit_value` attribute to allow zeekygen to skip adding a redef value to the docs. This is specifically used to avoid adding the massive values of `SSL::root_certs` and `SSL::ct_logs` to the documentation. I considered doing this with a new `@pragma` that would let you skip sections of files, but it ended up being a lot more concise and surgical to add an attribute for the individual variables.

The new docs look like this:

Screenshot 2025-03-24 at 10 54 24 AM

I'm not married to any of the naming or the format in the docs, so suggestions on those are definitely welcome. I also opted to leave the attribute out of the generated list in the docs for the variable, but if that seems weird it can be added back in.

Fixes #4190


scripts/base/protocols/ssl/main.zeek
```diff
@@ -116,7 +116,7 @@ export {
 
 	## The default root CA bundle.  By default, the mozilla-ca-list.zeek
 	## script sets this to Mozilla's root CA list.
-	const root_certs: table[string] of string &redef;
+	const root_certs: table[string] of string &redef &docs_omit_value;
```
awelzel, line 119: Given this is mostly about zeekygen and it's behavior is instructed by `##!, `##<`, or `##` comments, maybe we could support a magic comment rather than a new attribute?
```
..of string &redef ##@omit-value     # we have @no-test
..of string &redef ##@zeekygen-omit-value
..of string &redef ##@omit-value
..of string &redef ###omit-value
..of string &redef ##!omit-value
..of string &redef #&omit_value
```

@bbannier you have opinions here I'm sure - thoughts?

bbannier: My only skin in this game is that attributes always require special handling in a parser to disambiguate against binary `&`, e.g., the parser used for `zeek-format` [hardcodes all attributes](https://github.com/zeek/tree-sitter-zeek/blob/041d869f6af3fd854dec4e9952e5bdbbfb8cfb9d/grammar.js#L255-L284); for that reason introducing new attributes always introduces churn there.

I really like the magic comments idea (and was unaware of `&no-test`). That would not only not affect parsing elsewhere at all, but also avoid questions around how attributes compose which tend to crop up.

timwoj:
Replaced by #4325

PR #4307, opened 2025-03-24
Make storage events take a tag for the backend instead of a string

timwoj:


src/storage/backend/redis/Redis.cc
```diff
@@ -137,7 +137,7 @@ std::unique_lock<std::mutex> conditionally_lock(bool condition, std::mutex& mute
 
 namespace zeek::storage::backend::redis {
 
-storage::BackendPtr Redis::Instantiate(std::string_view tag) { return make_intrusive<Redis>(tag); }
+storage::BackendPtr Redis::Instantiate(zeek::Tag tag) { return make_intrusive<Redis>(tag); }
```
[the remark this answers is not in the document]
timwoj: I ended up moving the tag lookup into Backend's constructor, and the plugins can pass a string down to that.

src/storage/Backend.h
```diff
@@ -235,7 +237,8 @@ class Backend : public zeek::Obj {
     TypePtr val_type;
     RecordValPtr backend_options;
 
-    std::string tag;
+    zeek::Tag tag;
+    std::string tag_str;
```
ckreibich, line 241: Is it worth keeping the string version in addition to the tag?

src/storage/backend/redis/Redis.h
```diff
@@ -15,18 +15,18 @@ struct redisPollEvents;
...
+    static BackendPtr Instantiate();
 
     /**
      * Returns a descriptive tag representing the source for debugging.
      * This has to be overloaded for Redis because IOSource requires it.
      *
      * @return The debugging name.
      */
-    const char* Tag() override { return tag.c_str(); }
+    const char* Tag() override { return tag_str.c_str(); }
```
ckreibich, line 29: I noticed that if you declare `Tag()` virtual in Backend.h, you can skip it here. That seems nice since it's the same implementation, and then one less place where you need `tag_str`.

ckreibich (approved):
LGTM Tim. My comments are minor, feel free to handle as you see fit.

src/storage/Backend.h
timwoj: The string version is because if I use `Tag::AsString()` in `Tag()`, and then try to return the `.c_str` output, the compiler warns that it's returning the address of a temporary.

PR #4311, opened 2025-03-26
telemetry: Run callbacks at collect time

### LINKED ISSUE #4309: telemetry.log - metrics with callbacks have zero values
Enabling framworks/telemetry/log and tailing the telemetry.log indicates that the metrics for CPU, memory, open FDs aren't actually populated:
```
1742930752.983679       worker-8        gauge   zeek_dnsmgr_cache_entries       endpoint,type   worker-8,text   0.0
1742930752.983679       worker-8        gauge   zeek_dnsmgr_cache_entries       endpoint,type   worker-8,host   0.0
1742930752.983679       worker-8        counter process_cpu_user_seconds_total  endpoint        worker-8        0.0
1742930752.983679       worker-8        gauge   process_resident_memory_bytes   endpoint        worker-8        0.0
1742930752.983679       worker-8        gauge   process_open_fds        endpoint        worker-8        0.0
```

I suspect we don't run the callbacks when `collect_metrics()` is executed.

Also wondering if telemetry/log could/should filter out the endpoint label.

awelzel:
Calling collect_metrics() from a script would not invoke metric callbacks, resulting in most of the process metrics to be zero when a Zeek process isn't scraped via Prometheus.

Fixes #4309


0xxon:
Looking at the change - it seems like the new `RunCallbacks` always needs to be after `InvokeTelemetrySyncHooks` is called.

Is that the case? If yes, would it perhaps make sense to combine the two?

0xxon (approved):

0xxon:
Approved assuming tests pass :)

PR #4313, opened 2025-03-26
Topic/awelzel/4177 4178 custom event metadata

awelzel:
Draft to run the benchmarker on it.

---

What this does:

* Store generic event metadata as a std::vector rather than the network timestamp explicitly
* Don't extract network timestamp if it isn't needed.
* Deprecate EventHandler::Call() to remove no_remote and ts as these are auto-publish specific.


awelzel:
Closing again, doesn't look great :-(

PR #4314, opened 2025-03-26
ci: Bump version for opensuse-leap-15.6

awelzel:
Force rebuild of image to fix CI.


bbannier:
This seemed to have been the usual Cirrus issue where the image build job thinks the image is already cached, but the actual build cannot find it. I manually triggered a rebuild of the image from the UI today and the job seems to pass now. Do we need this for any other reason?

timwoj (approved):

awelzel:
> I manually triggered a rebuild of the image from the UI today and the job seems to pass now.

I didn't know that does that trick, too. Thanks!

PR #4315, opened 2025-03-26
Remove "experimental" from the QUIC history field's comment string

ckreibich:
We're unlikely to fundamentally change (or remove) this field at this point, and some users wondered whether we might do so, given the labeling.


awelzel (approved):
There's something in the docs as well about it being experimental: https://docs.zeek.org/en/master/logs/quic.html#example

This makes me remember Robin suggesting to pull the docs directly into the repo to make these things easier.

Issue #4177, opened 2025-01-15
Support for custom event metadata


J-Gras:
In #2929 timestamps have been added to events. I would like to add and query further custom event metadata. The mechanism could build upon a new API in the Event class:
1) Adding custom metadata can be done via C++ hooks such as `HookQueueEvent` that have access to the event object. 
2) Querying metadata in C++ can be done for arbitrary event objects. In addition, a script-level API to query the metadata of the current event would be nice.

Furthermore, one could distinguish between "static" metadata, which usually remains unchanged during a Zeek run (e.g., node name), and "dynamic" metadata, which may be different for each event (e.g., sequence numbers). For static metadata we could have a script-level API to set default values as well.

MP-Corelight:
To summarize the status: it looks like @awelzel added support for item #1 in [Add new plugin hook for publishing of cluster events #4401](https://github.com/zeek/zeek/pull/4401), while we still need to add support for item #2.

We may need to discuss the specific implementations and details for this issue to make sure we've prioritizing the right work.

awelzel:
> We may need to discuss the specific implementations and details for this issue to make sure we've prioritizing the right work.

The proposal for this work has been created here and already discussed here:

https://github.com/zeek/zeek/discussions/4334

awelzel:
I'm going to close this. The docs PR is here: https://github.com/zeek/zeek-docs/pull/327

As of:

> Furthermore, one could distinguish between "static" metadata, which usually remains unchanged during a Zeek run (e.g., node name), and "dynamic" metadata, which may be different for each event (e.g., sequence numbers). For static metadata we could have a script-level API to set default values as well.

This can be done by a plugin defining static metadata either in the C++ layer, or combined with script level logic. This feature is for advanced users. My somewhat opinionated stance is that at this point we do not need an official/supported script-level API for static metadata on events. This functionality can be developed separately in a plugin. I would have no trouble merging the plugin's code if it's becoming popular though.

PR #4316, opened 2025-03-27
Event/EventMgr cleanup

awelzel:
This is a prep for #4177, but it's subtle enough I'm opening this separately.

Instead of copying event information onto the EventMgr, this PR proposes to store a pointer to the event instead. This changes semantics of some of the bifs if invoked outside of event draining, but I doubt this really matters.

@rsmmr - would you mind looking?

### DIFF
--- a/src/Event.cc
+++ b/src/Event.cc
@@ -2,16 +2,13 @@
 
 #include "zeek/Event.h"
 
-#include "zeek/zeek-config.h"
-
 #include "zeek/Desc.h"
-#include "zeek/Func.h"
-#include "zeek/NetVar.h"
 #include "zeek/Trigger.h"
 #include "zeek/Val.h"
 #include "zeek/iosource/Manager.h"
-#include "zeek/iosource/PktSrc.h"
 #include "zeek/plugin/Manager.h"
+
+#include "event.bif.netvar_h"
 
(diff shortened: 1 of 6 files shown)


src/Event.cc
```diff
@@ -132,16 +119,13 @@ void EventMgr::Drain() {
...
         tail = nullptr;
 
         while ( current ) {
             Event* next = current->NextEvent();
 
-            current_src = current->Source();
-            current_aid = current->Analyzer();
-            current_ts = current->Time();
             current->Dispatch();
             Unref(current);
```
rsmmr, line 130: Minor: I find this usage of the non-local `current` a bit hard to read/follow. Could instead leave the iteration variable as a local (say `Event *ev`), and then call `Dispatch(ev)` to do the dispatch and delete.

rsmmr (approved):
Looks good to me.

src/Event.cc
awelzel: Thanks - yes, I'll revert. I'll check on the ASAN failure, too :sweat:

awelzel: He, this is exactly where the ASAN bug was. Thanks!

J-Gras (approved):

PR #3526, opened 2024-01-04
Support QUIC v2

awelzel:
Integration of https://github.com/zeek/spicy-quic/pull/16

PR #4317, opened 2025-03-27
Add "U" to QUIC history docstrings and expand version string docs

ckreibich:
Looks like we overlooked documenting "U" in zeek/zeek#3526 — does that look right to you @awelzel?


scripts/base/protocols/quic/main.zeek
```diff
@@ -57,6 +58,7 @@ export {
 		## R       RETRY packet
 		## C       CONNECTION_CLOSE packet
 		## S       SSL Client/Server Hello
+		## U       Unfamiliar QUIC version
```
awelzel, line 61: Unhandled? :-)

awelzel (approved):

PR #4321, opened 2025-03-28
Use one-based index in warning message

bbannier:
This was surfacing a zero-based counter to users which was confusing.


bbannier:
> Unless you're up for adding some opt-in option to enable that, go 🚢

I'll merge this as is for now, we could expand the error message to include the values in a follow-up.

PR #4322, opened 2025-03-30
Topic/awelzel/fix canon name builtin plugins

awelzel:
A user reported attempting to build in the [zeek-perf-support](https://github.com/zeek/zeek-perf-support) plugin. After fixing an issue with the plugin's CMakeFile, there was a further issue in that the new `zeek_add_plugin()` CMake function doesn't play nice with builtin plugins.

Opening to see if the improved CI job works.

---

goes with https://github.com/zeek/cmake/pull/125

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -768,24 +768,35 @@ zeekctl_debian11_task:
     upload_zeekctl_testing_artifacts:
       path: "auxil/zeekctl/testing/.tmp/**"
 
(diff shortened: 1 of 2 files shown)


ckreibich (approved):
Cool, yeah — let's do this.

I briefly wondered about `zeek-perf-support`, in case it could somehow verify whether Zeek itself is being built with frame pointers when the plugin gets built-in. But that's just a detail and we could adapt in the future.

PR #4325, opened 2025-03-31
Add comment annotation to disable copy redef value into docs

timwoj:
Replaces https://github.com/zeek/zeek/pull/4306

I opted for a leading annotation vs a trailing annotation because the trailing one was causing all sorts of parsing issues with existing redefs that had trialing comments. This annotation shouldn't be used very often (ever?) outside of these two cases anyways.


src/zeekygen/IdentifierInfo.cc
```diff
@@ -35,8 +35,17 @@ IdentifierInfo::~IdentifierInfo() {
 }
 
 void IdentifierInfo::AddRedef(const string& script, zeek::detail::InitClass ic, zeek::detail::ExprPtr init_expr,
-                              const vector<string>& comments) {
-    Redefinition* redef = new Redefinition(script, ic, std::move(init_expr), comments);
+                              vector<string> comments) {
+    bool omit_value = false;
+    for ( auto it = comments.begin(); it != comments.end(); ++it ) {
+        if ( it->find("@docs-omit-value") != std::string::npos ) {
+            it = comments.erase(it);
```
awelzel, line 42: Doesn't really matter due to the `break`, but the `++it` in the for head wouldn't work with `it = comments.erase(it)` as that would jump one element, maybe just remove the `it` assignment?

src/parse.y
```diff
@@ -296,8 +297,10 @@ static void build_global(ID* id, Type* t, InitClass ic, Expr* e, std::vector<Att
 
     add_global(id_ptr, std::move(t_ptr), ic, e_ptr, std::move(attrs_ptr), dt);
 
-    if ( dt == VAR_REDEF )
+    if ( dt == VAR_REDEF ) {
+        DBG_LOG(DBG_ZEEKYGEN, "redef of %s in %s", id->Name(), ::filename);
```
awelzel, line 301: Maybe move the logging into `Redef()` directly, so it's in `src/zeekygen` ?

src/zeekygen/IdentifierInfo.cc
```diff
@@ -35,8 +35,17 @@ IdentifierInfo::~IdentifierInfo() {
 }
 
 void IdentifierInfo::AddRedef(const string& script, zeek::detail::InitClass ic, zeek::detail::ExprPtr init_expr,
-                              const vector<string>& comments) {
-    Redefinition* redef = new Redefinition(script, ic, std::move(init_expr), comments);
+                              vector<string> comments) {
+    bool omit_value = false;
+    for ( auto it = comments.begin(); it != comments.end(); ++it ) {
+        if ( it->find("@docs-omit-value") != std::string::npos ) {
```
awelzel, line 41: I could see hardening this to `it->find(...) == 0` to force the keyword at beginning of a comment. My understanding is the whole line is dropped.

awelzel (approved):
:+1: 

Only thought is whether this could  apply more broadly rather than just `redef` (e.g. default initializers), but can do that when it comes up.

src/parse.y
timwoj: That was some left over debugging when I was tracing things. I removed it.

PR #4326, opened 2025-04-01
Bump spicy to current master [nomail]

timwoj:
This is primarily to bring in the fixes for CMake 4 on macOS. CI builds are currently broken because of macOS updating that via `brew`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,22 @@ if (WIN32)
     set(CMAKE_TOOLCHAIN_FILE ${_toolchain} CACHE STRING "Vcpkg toolchain file")
 endif ()
 
+if (APPLE AND CMAKE_VERSION VERSION_GREATER_EQUAL 4.0.0 AND NOT CMAKE_OSX_SYSROOT)
+    # Spicy needs having CMAKE_OSX_SYSROOT point to the macOS SDK
+    # path, but starting with CMake 4.0, CMAKE_OSX_SYSROOT is not set
+    # automatically anymore. So we follow the guidance from the CMake 4.0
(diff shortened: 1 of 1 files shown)


timwoj:
@rsmmr Apparently the fixes in the spicy tree weren't sufficient: https://cirrus-ci.com/task/5068827165196288

I ended up hard-coding to the command-line SDK for a local build to fix it, but that's probably not ideal:

```
index 5afb2cab..27eb2a33 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if (APPLE AND CMAKE_VERSION VERSION_GREATER_EQUAL 4.0.0)
     # The described situation ("let the compiler choose") doesn't quite match
     # ours, but this does lead to CMAKE_OSX_SYSROOT pointing to the macOS SDK
     # path when we need it.
-    set(CMAKE_OSX_SYSROOT "macosx")
+    set(CMAKE_OSX_SYSROOT "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk")
 endif ()
 
 project(spicy LANGUAGES C CXX)
```

rsmmr:
> @rsmmr Apparently the fixes in the spicy tree weren't sufficient: https://cirrus-ci.com/task/5068827165196288

I pushed a fix into this PR, plus another small Spicy tweak. If this works, I'll merge the Spicy tweak into Spicy `main`, and then update the submodule here once more.

rsmmr:
Well, actually this needs #4300 as well, so will focus on that.

rsmmr:
Superseded by #4300, which is merged.

PR #4331, opened 2025-04-02
Speed up EnumType::Lookup

### LINKED ISSUE #4327: cat(enum) does an O(N) search
I was playing around with https://github.com/janestreet/magic-trace + https://github.com/zeek/zeek-perf-support/ to see if it could be useful... (it works!) The first thing I zoomed in on was... surprising.

Image

That's this line from the smb scripts:

```
return clean(cat(Analyzer::ANALYZER_SMB, c$id$orig_h, c$id$resp_h, path_name, file_name, last_mod));
```

which is fairly expensive, but the odd part is that the `cat(Analyzer::ANALYZER_SMB)` is the slowest part.   The implementation shows why.. it's named Lookup, but it's not doing a lookup, it's doing a search:

```
const char* EnumType::Lookup(zeek_int_t value) const {
    for ( NameMap::const_iterator iter = names.begin(); iter != names.end(); ++iter )
        if ( iter->second == value )
            return iter->first.c_str();

    return nullptr;
}
```

this doesn't seem to be optimized by ZAM either.

timwoj:
This adds a second map to EnumType to reverse the lookup between val and name, avoiding the need for an O(n) search in `EnumType::Lookup`.

Fixes #4327

### DIFF
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -1617,12 +1617,10 @@ zeek_int_t EnumType::Lookup(const string& module_name, const char* name) const {
 }
 
 zeek_int_t EnumType::Lookup(const string& full_name) const {
-    auto pos = names.find(full_name.c_str());
-
-    if ( pos == names.end() )
-        return -1;
-    else
+    if ( auto pos = names.find(full_name.c_str()); pos != names.end() )
         return pos->second;
+
+    return -1;
 }
 
 const char* EnumType::Lookup(zeek_int_t value) const {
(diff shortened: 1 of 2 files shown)


ckreibich (approved):
LGTM. I like Arne's suggestion of putting the name into the vals, but the symmetry here is appealing, and memory use should be minimal — plus you've already done it.

I did a quick check that generated docs indeed have the same ordering, all good.

It looks like we could get rid of the `NameMap` and `RevNameMap` type aliases? At least the latter isn't used anywhere.

timwoj:
> LGTM. I like Arne's suggestion of putting the name into the vals, but the symmetry here is appealing, and memory use should be minimal — plus you've already done it.

I actually went down that route first and ran into a bunch of cases where we insert elements into the `vals` map without names, and then Zeek reports a bunch of warnings because it's trying to do lookups against values without names. This ended up being much more surgical.

JustinAzoff:
re: the map, I was wondering if there was a reason `rev_names` couldn't be a `vec`?  I believe the following are true:

* we only ever add names
* they are inserted in increasing order

so instead of

```
rev_names[val] = fullname;
```

we could have something like

```
if (val >= rev_names.size()) {
    rev_names.resize(val+1);
}
rev_names[val] = fullname;
```

JustinAzoff:
ah, that makes sense, I could see an enum like

```
type X: enum {
a = 1,
b = 2,
undefined = 0xffffffff,
};
```
especially in enums related to protocols.

PR #4332, opened 2025-04-03
Add pluggable serialization for storage backends

timwoj:
This adds a new component/plugin type for serialization for storage backends. It replaces the hard-coded JSON serialization in the existing backends with a new JSON serializer.

A couple of thoughts on the implementation for discussion:
- I'm not a huge fan of passing the serializer `unique_ptr` through the factory/instantiate/constructors, even though this is conceptually the correct way to do it in C++. it adds a bunch of boiler plate that backend authors must have. I considered adding a `SetSerializer` method to the base `Backend` object instead and calling that from the BIFs, but I don't know if that ends up more or less clear.
- I duplicated the `byte_buffer` and `byte_buffer_span` type aliases from the cluster framework. I could see pulling those out of both places into util.h, and potentially moving them out of the `detail` namespace.
- I'd love if there was a way to add multiple levels of namespacing to the script-level types, so that I could use `Storage::Backend::SQLITE` and `Storage::Serializer::JSON` enum names instead of everything being shoved in the top-level `Storage` module. It would help clarify what's being passed in the scripts.


scripts/base/frameworks/storage/async.zeek
```diff
@@ -25,7 +25,7 @@ export {
 	## Returns: A record containing the status of the operation, and either an error
 	##          string on failure or a value on success. The value returned here will
 	##          be an ``opaque of BackendHandle``.
-	global open_backend: function(btype: Storage::Backend,
+	global open_backend: function(btype: Storage::Backend, stype: Storage::Serializer,
```
awelzel, line 28: WDYT about putting the serializer into `BackendOptions`? 
```
type BackendOptions: record {
    serializer: Storage::Serializer &default=Storage::SERIALIZER_JSON
};
```
Or possibly event `&default=default_serializers` where the latter is `const &redef` within `Storage`.

Seems reasonable to me to have a default that works well out of the box, but optionally allow injecting something else.

src/storage/Manager.cc
```diff
@@ -48,24 +50,36 @@ void Manager::InitPostScript() {
     StartExpirationTimer();
 }
 
-zeek::expected<BackendPtr, std::string> Manager::Instantiate(const Tag& type) {
-    Component* c = Lookup(type);
-    if ( ! c ) {
+zeek::expected<BackendPtr, std::string> Manager::Instantiate(const Tag& btype, const Tag& stype) {
```
awelzel, line 53: It could make sense to split out `InstantitateSerializer(tag)` as an API so other places could get serializers. This might be a bit far out there, but that way a bif like the following becomes easy:
```
function serialize_val(tag: Storage::Serializer, v: any): string
```
Probably mostly useful for testing, but hey :-)

src/storage/Serializer.h
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include "zeek/Span.h"
+#include "zeek/Val.h"
+
+namespace zeek::storage {
+
+namespace detail {
+using byte_buffer = std::vector<std::byte>;
+using byte_buffer_span = Span<const std::byte>;
```
awelzel, line 12: Thumbs up for moving into `zeek::util` if you think these are reasonable types for that use case. @bbannier any opinions / experience if that's reasonable aliases in the realm of passing bytes around?

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -151,24 +151,27 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
     if ( ! db )
         return {ReturnCode::NOT_CONNECTED};
 
-    auto json_key = key->ToJSON();
-    auto json_value = value->ToJSON();
+    auto key_data = serializer->Serialize(key);
```
awelzel, line 154: This may be good for broader discussion somewhere, but it's also a rabbit hole so I don't have an overly strong opinion: Should we tag the stored data with the serializer?

E.g. instead of storing the plain serializer result, store `serializer->GetName() + ";" + <key_data>`. That'll be more overhead and maybe the proposed format is clunky, but it'd allow to check if the stored data is compatible with the chosen serializer. If we really wanted to, we could fall back to an older de-serializer if it's different than the configured one. External consumers looking at data directly in Redis or SQLite directly could also make use of it.

For cluster communication and IPC it's not super relevant as it's not about long term storage, but messages between nodes do contain the serialization format as part of the message. Messages serialized in the wrong format are rejected.

A million keys would all be prefixed with e.g. `JSONv1` and take ~6MB of more storage and it feels very clunky, so wonder what others think around this topic.

scripts/base/frameworks/storage/async.zeek
timwoj: Yeah, I like that more actually, as it removes the need to pass the serializer along through the constructors too.

evantypanski:
For NATS, the clients have a pretty strict idea of what can go into the keys. But, that currently requires a little hack to even get JSON working - I also added a "Strict" mode for string-values-only (utility TBD).

So question: If a backend has stricter requirements about how the data should be serialized, how should that be enforced?

timwoj:
> So question: If a backend has stricter requirements about how the data should be serialized, how should that be enforced?

One thing I can think is to write a serializer specific for that backend, but that gets messy considering you're dealing with Zeek `Val` objects. At least with the JSON serializer converting the `byte_buffer` result back into strings isn't too difficult. I did consider whether or not the `Serializer` class should have versions of `Serialize` and `Unserialize` that returned strings as well, for cases where backends didn't want bytes. That would be easy to add if it simplifies things for other backends.

I ended up switching the SQLite and Redis backends to binary storage in this PR, just to not have to deal with the conversion back to text. SQLite has the typical BLOB types, and I discovered that Redis commands can take `%b` for binary vs `%s` for strings.

src/storage/backend/sqlite/SQLite.cc
timwoj: I agree that versioning is a good idea here. I think the only thing I'd change from the above is where to keep the version. I'd store it separately from the serializer name, so that plugin versions don't need to change names if they're upgraded to a new version. Adding something like a `VersionedName()` method would be enough to pass into the backends.

I also think that allowing the serializer to handle how the version is represented in the data would be best. For example, for JSON it should be represented as part of the JSON data. That would require changes to our internal To/FromJSON methods to take a version to include though. For other serializers it might be represented differently.

timwoj:
@evantypanski What do you you think about the serializer taking a function as an argument that would provide a transformation that could be applied to the data during serialization? That would let a backend say "I want you to serialize this, but I want this formatting to happen while you're doing it".

evantypanski:
> What do you you think about the serializer taking a function as an argument that would provide a transformation that could be applied to the data during serialization?

I think that'd work well, NATS is just a find-and-replace (though it'd be one-to-many bytes). In a quick lay of the land, I found if MongoDB is used they may have a similar problem with `.` and `$`. There's probably more, but I just wanted to find one other

Since for NATS I currently just use a function to post-process the serialized string from JSON, providing a similar function would work well and be exactly what I would want

src/Val.h
```diff
@@ -234,6 +234,10 @@ class Val : public Obj {
 
     TableValPtr GetRecordFields();
 
+    // The version of the data generated by ToJSON and read by ValFromJSON. If that
+    // structure changes, this number must be increased.
+    static const int JSON_VERSION = 1;
+
```
awelzel, line 240: This doesn't feel great, would ditch the version.

Longer and IMO: We should just not change the output of `Val::ToJSON()` anymore (except for adding missing features or fixing obvious bugs). If there ever comes an idea for a drastically different JSON format, then that should be a new API like `Val::ToJSONv2()`, or implemented in a serializer outside of `Val.cc`.

src/storage/serializer/json/JSON.cc
```diff
@@ -8,15 +8,19 @@ namespace zeek::storage::serializer::json {
 
 std::unique_ptr<Serializer> JSON::Instantiate() { return std::make_unique<JSON>(); }
 
-JSON::JSON() : Serializer("JSON") {}
+JSON::JSON() : Serializer("JSON") { versioned_name = util::fmt("%sv%d", name.c_str(), Val::JSON_VERSION); }
```
awelzel, line 11: ...would just hard-code "JSONv1" instead of `fmt()`, assuming JSON_VERSION is dropped.

src/storage/serializer/json/JSON.cc
```diff
@@ -8,15 +8,19 @@ namespace zeek::storage::serializer::json {
...
 
 std::optional<util::byte_buffer> JSON::Serialize(ValPtr val) {
+    static auto byte_converter = [](u_char c) { return std::byte(c); };
+
+    std::string version = versioned_name + ";";
```
awelzel, line 16: The `;` was my initial thought. I'm now thinking whether of `\n` or `<space>` might be easier to work with (e.g. using default `split()` helpers). Mind making the call?

src/storage/Serializer.h
```diff
@@ -2,14 +2,48 @@
...
+/**
+ * Base class for a serializer used by storage backends.
+ */
 class Serializer {
 public:
+    /**
+     * A function pointer to a string transformation function that backends can use to
+     * force formatting on the serializer output.
+     */
+    using string_transform = std::function<std::string(std::string_view)>;
```
awelzel, line 21: Wouldn't it be more consistent to have this take a byte_span and return a byte_buffer?

But also, I'm not sure we should have it. If it's added as a parameter to `Serialize()`, all serializers need to implement the same logic. To me it feels more like backends that need to sanitize/munge the serializer output should do so explicitly, instead of pushing the transformer logic down into `Serialize()`:
```
// not sure this works, but something like it:
serializer->Serialize(val).and_then(string_transform).or_else(blow_up)
```

src/storage/Serializer.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace detail {
+using byte_buffer = std::vector<std::byte>;
+using byte_buffer_span = Span<const std::byte>;
+} // namespace detail
+
+class Serializer {
+public:
+    virtual ~Serializer() = default;
+    virtual std::optional<detail::byte_buffer> Serialize(ValPtr val) = 0;
```
awelzel: Over in `cluster`, the Serialize() method takes a `detail::byte_buffer&` that can be pre-allocated and re-used. With this API, the implementer definitely needs to allocate a new `detail::byte_buffer` and cannot re-use a pre-allocated one. Nothing to do, but wanted to point out why it's different elsewhere.

awelzel (commented):
Had a few comments, mostly (I think) to keep things simpler :-)

testing/btest/plugins/storage.zeek
```diff
@@ -22,6 +22,7 @@ redef record Storage::BackendOptions += {
 
 event zeek_init() {
 	local opts : Storage::BackendOptions;
+	opts$serializer = Storage::JSON;
```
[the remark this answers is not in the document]
timwoj: There's a few tests that have it set just to ensure it's tested as a fixed value somewhere. I can take this one out.

src/util.h
```diff
@@ -310,6 +312,10 @@ int setvbuf(FILE* stream, char* buf, int type, size_t size);
 
 } // namespace detail
 
+// Byte buffer types used by serialization code in storage and cluster.
+using byte_buffer = std::vector<std::byte>;
+using byte_buffer_span = Span<const std::byte>;
+
```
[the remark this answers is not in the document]
timwoj: I actually have a branch that moves `zeek::expected` plus `zeek_int_t` and `zeek_uint_t` to a separate `util-types.h` header. It's included by default in `util.h`, but there's lots of places we only use those int types that we don't need the entirety of util.h. These could move to that header as well.

The branch is based on https://github.com/zeek/zeek/pull/4340 though, so I was waiting for that to merge before I opened a PR for it.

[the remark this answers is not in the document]
timwoj: As for moving them out of `zeek::util`, that sounds good to me.

src/Val.h
timwoj: Seems reasonable to me.

src/storage/serializer/json/JSON.cc
timwoj: It doesn't matter to me. I'm not sure about using `util::split` though because we only want to split at the first instance and not potentially any other occurrences of the delimiter. `util::split` splits all of them.

src/storage/Serializer.h
timwoj: It was an experiment for sure. In the end I'm not the biggest fan of it either. I think the only problem it solves is preventing the backends from having to convert back from bytes to do their transformation. For example, the NATS backend converts any non-alnum character into byte-strings (`x00`-style). If this happens after the call to `Serialize()`, it's going to need convert back to a string from the byte_buffer result and then do the transformation.

timwoj: I'm going to just drop out that entire commit. I don't see it being useful enough to keep as a generic thing in the serializers.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -266,17 +273,16 @@ OperationResult SQLite::Step(sqlite3_stmt* stmt, bool parse_value) {
     int step_status = sqlite3_step(stmt);
     if ( step_status == SQLITE_ROW ) {
         if ( parse_value ) {
-            // Column 1 is the value
-            const char* text = (const char*)sqlite3_column_text(stmt, 0);
-            auto val = zeek::detail::ValFromJSON(text, val_type, Func::nil);
+            std::byte* blob = (std::byte*)sqlite3_column_blob(stmt, 0);
```
awelzel, line 276: Nit, c-style caset.

awelzel (approved):
Thanks for the updates!

PR #4333, opened 2025-04-04
Improve and fix enum identifiers, types and vals

awelzel:
1) Associated an "enum identifier" with its EnumVal such that it becomes possible to lookup, e.g., the "Conn::LOG" identifier and directly get its EnumVal via `GetVal()`.

2) Prevent ShallowClone() of an EnumType within `add_type()`. See the commit message, but essentially that resulted in enum identifiers not having a consistent EnumType associated.

3) Fixup broker enum disagreement (might be worth backporting, though I don't know what the effect really is)

4) Add a new plugin test for poking at the API.


scripts/base/frameworks/broker/main.zeek
```diff
@@ -207,24 +207,26 @@ export {
 		PEER_INVALID = 3,
 		## Remote peer not listening.
 		PEER_UNAVAILABLE = 4,
+                ## Remote peer disconnected during the handshake.
+	        PEER_DISCONNECT_DURING_HANDSHAKE =   5,
```
evantypanski, line 211: nit: this and above have inconsistent whitespace (so switch them to tabs)

src/Type.cc
```diff
@@ -1591,6 +1591,11 @@ void EnumType::CheckAndAddName(const string& module_name, const char* name, zeek
     if ( vals.find(val) == vals.end() )
         vals[val] = make_intrusive<EnumVal>(IntrusivePtr{NewRef{}, this}, val);
 
+    if ( ! id->HasVal() )
+        id->SetVal(vals[val]);
+    else if ( id->GetVal()->AsEnum() != val )
+        reporter->InternalError("inconsistent enum integer value");
```
evantypanski, line 1597: nit: Can this include the name it's trying to add?

```suggestion
        reporter->InternalError("inconsistent enum integer value for '%s'", fullname.c_str());
```

evantypanski (approved):
Cool, it's interesting that this caught a but in an existing enum

PR #4336, opened 2025-04-05
Pile of Coverity fixes

timwoj:
A Friday-afternoon spree of Coverity fixing:

- Lots of `std::move` additions, a couple of removals where they weren't doing anything, and one where Coverity actually flagged a use of an object after it was moved.
- A couple of potential memory leaks in the debug code. That code probably isn't used much, but Coverity flagged them anyways.
- A few minor fixes in the POP3 analyzer that clangd/flycheck in emacs were complaining about.
- Added an assert to avoid a divide-by-zero in `CardinalityCounter`
- Avoided some copies by using const-auto references. These were reported by Coverity.
- Avoided additional temporary strings with some concatenation in script_opt
- Disabled executing up any of the telemetry listener code if the ZEEKCTL_CHECK_CONFIG environment variable is set. Previously we still would create the callback object and just not start up the listener, but this lead to leaking the callback object. There's no reason to allocate it though if we're not going to use it.

### DIFF
--- a/src/Debug.cc
+++ b/src/Debug.cc
@@ -160,7 +160,7 @@ int TraceState::LogTrace(const char* fmt, ...) {
 
     if ( ! loc.filename ) {
         static constexpr const char str[] = "<no filename>";
-        loc.filename = util::copy_string(str, std::size(str) - 1);
+        loc.filename = str;
         loc.last_line = 0;
     }
 
@@ -679,7 +679,7 @@ string get_context_description(const Stmt* stmt, const Frame* frame) {
         loc = *stmt->GetLocationInfo();
     else {
         static constexpr const char str[] = "<no filename>";
-        loc.filename = util::copy_string(str, std::size(str) - 1);
+        loc.filename = str;
         loc.last_line = 0;
     }
 
--- a/src/ZeekString.cc
+++ b/src/ZeekString.cc
@@ -85,6 +85,18 @@ const String& String::operator=(const String& bs) {
     return *this;
 }
 
+String& String::operator=(String&& other) noexcept {
+    b = other.b;
+    n = other.n;
+    final_NUL = other.final_NUL;
+    use_free_to_delete = other.use_free_to_delete;
+
+    other.b = nullptr;
+    other.Reset();
+
+    return *this;
+}
+
 bool String::operator==(const String& bs) const { return Bstr_eq(this, &bs); }
 
 bool String::operator<(const String& bs) const { return Bstr_cmp(this, &bs) < 0; }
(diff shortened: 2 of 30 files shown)


src/analyzer/protocol/pop3/POP3.cc
```diff
@@ -673,6 +673,7 @@ void POP3_Analyzer::ProcessReply(int length, const char* line) {
                         masterState = detail::POP3_UPDATE;
 
                     break;
+                default: break;
```
awelzel, line 676: Is there some standard way to mark this as "unreachable" so it's clear this is to please a tool?

Same below.

src/probabilistic/CardinalityCounter.cc
```diff
@@ -134,6 +134,10 @@ void CardinalityCounter::AddElement(uint64_t hash) {
  **/
 double CardinalityCounter::Size() const {
     double answer = 0;
+
+    // Avoid a divide-by-zero here.
+    ASSERT(m > 0);
```
awelzel, line 139: The bif doc states that it returns `-1.0` if empty, so maybe that'd be okay to return here instead of ASSERT which may not produce any code?

awelzel (approved):
Had a few thoughts and am between "comment" and "approve", but I don't think this needs another review.

src/Debug.cc
```diff
@@ -158,9 +158,10 @@ int TraceState::LogTrace(const char* fmt, ...) {
         }
     }
 
+    char* loc_filename = nullptr;
     if ( ! loc.filename ) {
         static constexpr const char str[] = "<no filename>";
-        loc.filename = util::copy_string(str, std::size(str) - 1);
+        loc.filename = loc_filename = util::copy_string(str, std::size(str) - 1);
```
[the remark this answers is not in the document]
timwoj: It looks like it does. I'll fix the other one that had the same problem too.

src/ZeekString.cc
```diff
@@ -468,7 +468,7 @@ TEST_CASE("construction") {
     zeek::String s5{std::string("abcdef")};
     CHECK_EQ(s5.Len(), 6);
 
-    zeek::String s6{s5};
+    zeek::String s6{std::move(s5)};
```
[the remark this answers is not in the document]
timwoj: Good catch. I'll switch it back and annotate it. I did notice that `ZeekString` has a move constructor but is lacking a move assignment operator as well. I'll add that.

src/analyzer/protocol/pop3/POP3.cc
timwoj: There is `__builtin_unreachable()`. It's a compiler intrinsic that I know gcc and clang support. I'll need to see if it's supported by msvc though.

PR #4338, opened 2025-04-08
Support for custom event metadata

awelzel:
This is an implementation of #4334. Opening as draft for CI.

Very short description:

* zeek::Event holds a `unique_ptr<std::vector>` for metadata.
* `zeek::cluster::detail::Event` holds the same data structure.
* A new plugin hook, `HookPublishEvent()` can be used to populate a `cluster::detail::Event` with metadata
* Metadata is 1) added to the `EventMetadata::ID` enum and 2) their type registered statically using `EventMetadata::register_type()`.


awelzel:
I'm going to close this for the time being as I don't see it landing for 7.2. I'll probably separate out some of the fixes pre-work for a separate PR.

PR #4339, opened 2025-04-08

bbannier:
Probably the most relevant change for Zeek here is zeek/spicy#2012.


timwoj (approved):
The changes from clang-format are sort of surprising, but not a big deal.

PR #4340, opened 2025-04-08
#include cleanup

timwoj:
Emacs has been complaining about a lot of these at me for a while so I figured I'd go through and fix most of them in one pass. This is probably best reviewed commit-by-commit, especially the ones that aren't just a block of removals.

I'm not entirely sure about https://github.com/zeek/zeek/pull/4340/commits/8c5866a771e7fac0b83c839f103f0ad10a3361b5. The primary change is in plugin/Component.h, which technically has the possibility of breaking plugin builds.

### DIFF
--- a/src/Attr.cc
+++ b/src/Attr.cc
@@ -2,8 +2,6 @@
 
 #include "zeek/Attr.h"
 
-#include "zeek/zeek-config.h"
-
 #include "zeek/Desc.h"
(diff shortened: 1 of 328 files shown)


timwoj:
Now that 7.2 has been branched, this is ready for an actual review.

timwoj:
Splitting this up into a couple of smaller PRs to ease reviews.

PR #4343, opened 2025-04-09
Revert addition of std::unreachable

timwoj:
After further testing with the fuzzer corpus, std::unreachable isn't necessary here. It's fine to just let the default case break to fix the Coverity warning.


awelzel (approved):
I find it slightly surprising this tickled something as the case list above appears as if it could be exhaustive, but good with the revert.

PR #4344, opened 2025-04-09
Cluster Websocket support for Broker

awelzel:
#### Summary

Deprecate `Broker::listen_websocket()` and provide `Cluster::listen_websocket()` as an alternative that is becoming functional with Broker.

#### Details

For the ZeroMQ (or NATS) cluster backend, every WebSocket client has its own dedicated backend instance. This doesn't directly work with Broker due to peering requirements.

Instead, @Neverlord added a new `hub` primitive that has the required semantics for WebSocket clients. Primarily, that's visibility of `publish()` on the endpoint to `hub` instances, as well as `hub.publish()` visibility to the endpoint's subscribers, other hubs and remote endpoints.  A new `Broker::WebSocketShim` class holds such a hub and uses it for a WebSocket's publish and subscribe functionality instead of a full `Broker` backend.

There's a bit of API renaming and fixes sprinkled in that came up while working on this PR.

There's a new `wstest.py` file that's a testing library used by the new `cluster.websocket.broker` tests because there was lots of duplicated code, but didn't yet convert the existing ones due to the extra diff/churn.


src/broker/Manager.h
```diff
@@ -446,7 +447,7 @@ class Manager : public zeek::cluster::Backend, public iosource::IOSource {
     void Error(const char* format, ...) __attribute__((format(printf, 2, 3)));
 
     // Processes events from the Broker message queue.
-    void ProcessMessages();
+    void ProcessMessages(std::vector<broker::data_message> messages);
```
bbannier, line 450: Needs `#include <vector>`.

testing/btest/Files/ws/wstest.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    if url is None:
+        url = WS4_URL_V1
+
+    cc = websockets.sync.client.connect(url)
+    return TestClient(name, cc)
+
+
+def recv_until_timeout(
+    clients: list[TestClient], *, timeout: float = DEFAULT_RECV_TIMEOUT, desc=None
```
bbannier, line 103: An empty string is already falsy so suggest to just pass `desc: str = ""`.

testing/btest/cluster/websocket/zeromq/manager-worker-ping-pong.zeek
```diff
@@ -0,0 +1,150 @@
...
+#
+# @TEST-EXEC: btest-bg-wait 30
+# @TEST-EXEC: btest-diff ./manager/out
+# @TEST-EXEC: btest-diff ./manager/.stderr
+# @TEST-EXEC: btest-diff ./worker-1/out
+# @TEST-EXEC: btest-diff ./worker-1/.stderr
+# @TEST-EXEC: btest-diff ./client/out
+# @TEST-EXEC: btest-diff ./client/.stderr
+
+@TEST-START-FILE common.zeek
```
bbannier, line 30: Let's either always or never put BTest meta in comments. I personally prefer _always_.

Here and in all other tests this PR adds.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -325,6 +352,10 @@ bool ZeroMQBackend::DoUnsubscribe(const std::string& topic_prefix) {
         // Prepend 0x00 byte to indicate subscription to XSUB socket.
         // This is the XSUB API instead of setsockopt(ZMQ_SUBSCRIBE).
         std::string msg = '\0' + topic_prefix;
+
+        // Send a two message parts. The first part is a single byte tagging the
```
bbannier, line 356: Should we say "unsubscription" everywhere in the body of this function where it currently says "subscription"?

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -413,11 +444,26 @@ void ZeroMQBackend::Run() {
...
+                }
+                else if ( tag == InprocTag::Shutdown ) {
+                    if ( self_thread_stop )
+                        ZEROMQ_THREAD_PRINTF("inproc: error: duplicate shutdown message");
+
+                    self_thread_stop = true;
+                }
+                else {
+                    ZEROMQ_THREAD_PRINTF("inproc: error: unknown tag %x", static_cast<uint8_t>(tag));
+                }
```
bbannier, line 466: Let's use a `switch` instead. Since the enum is declared in this file we know all possible values, and no `default` branch should be needed.

```suggestion
                switch ( msg[0].data<InprocTag>()[0] ) {
                    case InprocTag::XSUB: xsub.send(msg[1], zmq::send_flags::none); break;
                    case InprocTag::Shutdown:
                        if ( self_thread_stop )
                            ZEROMQ_THREAD_PRINTF("inproc: error: duplicate shutdown");
                        self_thread_stop = true;
                        break;
                }
```

src/cluster/websocket/const.bif
```diff
@@ -0,0 +1,2 @@
+##! C++ access for consts from the websocket subsystem.
+const Cluster::websocket_dispatcher_queue_size: count;
```
bbannier, line 2: I do not know about the existing code, but is this a standard pattern of using bifcl to emit C++ forward declarations? I would have expected for us to just get the constant declared in script land in `init-bare.zeek` dynamically. The danger of decoupling from that is that we now need to maintain name and type in two places.

scripts/base/init-bare.zeek
```diff
@@ -5840,6 +5840,11 @@ export {
 	##
 	## This currently has no effect for backend BROKER.
 	const log_serializer = Cluster::LOG_SERIALIZER_ZEEK_BIN_V1 &redef;
+
+	## Size of the WebSocket event queue for external clients.
+	##
+	## Only change this if you know what you're doing.
```
bbannier, line 5846: This comment is really a cop-out, and we should at least say what effect this constant has. Also, I'd expect that changing most constants can have unexpected effects, so users should always validate their configurations -- there is no need to make that explicit everywhere.

Maybe something like this, with enough details:

```suggestion
	## Size of the WebSocket message queue for external clients.
	##
	## If there are more than `websocket_dispatcher_queue_size` messages to the client outstanding we will ...
```

src/cluster/Backend.cc
```diff
@@ -27,6 +27,11 @@ void detail::LocalEventHandlingStrategy::DoEnqueueLocalEvent(EventHandlerPtr h,
     zeek::event_mgr.Enqueue(h, std::move(args));
 }
 
+void detail::LocalEventHandlingStrategy::DoHandleError(std::string_view code, std::string_view message) {
+    // This should not happen often!
+    zeek::reporter->Error("Cluster backend error: %s: %s", std::string(code).c_str(), std::string(message).c_str());
```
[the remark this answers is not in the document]
awelzel: This was changed.

testing/btest/cluster/websocket/zeromq/manager-worker-ping-pong.zeek
```diff
@@ -0,0 +1,150 @@
...
+# @TEST-PORT: LOG_PULL_PORT
+# @TEST-PORT: WEBSOCKET_PORT
+#
+# @TEST-EXEC: cp $FILES/zeromq/cluster-layout-simple.zeek cluster-layout.zeek
+# @TEST-EXEC: cp $FILES/zeromq/test-bootstrap.zeek zeromq-test-bootstrap.zeek
+# @TEST-EXEC: cp $FILES/ws/wstest.py .
+#
+# @TEST-EXEC: zeek -b --parse-only manager.zeek
+# @TEST-EXEC: zeek -b --parse-only worker.zeek
+# @TEST-EXEC: python3 -m py_compile client.py
```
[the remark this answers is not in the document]
awelzel: The idea was more a `--parse-only` for Python to not start the test if client.py won't work anyhow.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -44,6 +44,17 @@ enum class DebugFlag : zeek_uint_t {
...
 
+enum class InprocTag : uint8_t {
+    XSUB,
+    Shutdown,
+};
+
+namespace {
+
+zeek::cluster::zeromq::InprocTag inproc_tag_shutdown[] = {InprocTag::Shutdown};
+zeek::cluster::zeromq::InprocTag inproc_tag_xsub[] = {InprocTag::XSUB};
```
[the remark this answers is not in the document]
awelzel: Ah, yes. I was stuck in "I need an array of size 1", but could just use pointer to a value. Thanks!

src/cluster/backend/zeromq/ZeroMQ.h
```diff
@@ -105,6 +105,8 @@ class ZeroMQBackend : public cluster::ThreadedBackend {
     zmq::socket_t log_pull;
 
     std::thread self_thread;
+    bool self_thread_shutdown_requested = false;
```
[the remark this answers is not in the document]
awelzel: > I wonder whether we really needs this or could instead just trigger shutdown, potentially repeatedly?

The problem I observed was that if the thread actually terminated, the `main_inproc.send()` will block indefinitely, so I added the helper. The error was mainly to notice if something is totally off, but happy to replace.

scripts/base/init-bare.zeek
awelzel: Fair!

src/cluster/websocket/WebSocket.cc
```diff
@@ -291,11 +290,30 @@ void WebSocketEventDispatcher::Process(const WebSocketOpen& open) {
 
     const auto& event_serializer_val = id::find_val<zeek::EnumVal>("Cluster::event_serializer");
     auto event_serializer = cluster::manager->InstantiateEventSerializer(event_serializer_val);
-    const auto& cluster_backend_val = id::find_val<zeek::EnumVal>("Cluster::backend");
+    auto cluster_backend_val = id::find_val<zeek::EnumVal>("Cluster::backend");
+
+    static const auto& cluster_backend_type = zeek::id::find_type<EnumType>("Cluster::BackendTag");
```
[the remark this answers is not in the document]
awelzel: I cleaned this up. At runtime there should not be new EnumVal or EnumType instsances be generated, so this should be safe.

src/cluster/websocket/const.bif
awelzel: Thanks - I'll move the queue size to `WebSocketServerOptions` instead. That allows to move the default into `cluster/main.zeek` and also makes it configurable on a per server.

src/broker/Manager.cc
```diff
@@ -1024,7 +1036,7 @@ zeek::RecordValPtr Manager::MakeEvent(ArgsSpan args, zeek::detail::Frame* frame)
 
 bool Manager::DoSubscribe(const string& topic_prefix, SubscribeCallback cb) {
     DBG_LOG(DBG_BROKER, "Subscribing to topic prefix %s", topic_prefix.c_str());
-    bstate->subscriber.add_topic(topic_prefix, ! run_state::detail::zeek_init_done);
+    bstate->hub.subscribe(topic_prefix, ! run_state::detail::zeek_init_done);
```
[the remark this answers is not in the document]
awelzel: I'm punting this as I'd probably just be hallucinating about real reasons here.

testing/btest/cluster/websocket/zeromq/manager-worker-ping-pong.zeek
```diff
@@ -0,0 +1,150 @@
...
+                    wstest.recv_until_timeout(clients, desc=f"tc1 - ping {i}")
+
+                    tc2.send_json(wstest.build_event_v1("/test/pings/", "ping", [msg, 200 + i]))
+                    wstest.recv_until_timeout(clients, desc=f"tc2 - ping {i}")
+
+                    tc3.send_json(wstest.build_event_v1("/test/pings/", "ping", [msg, 300 + i]))
+                    wstest.recv_until_timeout(clients, desc=f"tc3 - ping {i}")
+
+if __name__ == "__main__":
+    wstest.main(run, wstest.WS4_URL_V1)
```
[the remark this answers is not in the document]
awelzel: Thanks! Switched it over.

src/cluster/websocket/WebSocket.cc
```diff
@@ -297,12 +297,32 @@ void WebSocketEventDispatcher::Process(const WebSocketOpen& open) {
...
+    static const auto& event_serializer_val = id::find_val<zeek::EnumVal>("Cluster::event_serializer");
     auto event_serializer = cluster::manager->InstantiateEventSerializer(event_serializer_val);
-    const auto& cluster_backend_val = id::find_val<zeek::EnumVal>("Cluster::backend");
+    static const auto& cluster_backend_val = id::find_val<zeek::EnumVal>("Cluster::backend");
+    auto effective_backend_val = cluster_backend_val;
+
+    static const auto& broker_enum_val = zeek::id::find_val<EnumVal>("Cluster::CLUSTER_BACKEND_BROKER");
+    static const auto& broker_ws_shim_enum_val =
+        zeek::id::find_val<EnumVal>("Cluster::CLUSTER_BACKEND_BROKER_WEBSOCKET_SHIM");
+    if ( effective_backend_val == broker_enum_val ) {
```
ckreibich, line 303: You could in theory also just do `cluster::backend == broker_mgr` (like we do in zeek-setup.cc), but this is more future-proof — right?

testing/btest/Files/ws/wstest.py
```diff
@@ -0,0 +1,225 @@
...
+                all_timeout = False
+            except TimeoutError:
+                msgs += [{"client": tc.name, "timeout": True}]
+
+        if not all_timeout:
+            for msg in msgs:
+                print(json.dumps(msg))
+
+
+class TaggedArg:
```
ckreibich, line 137:
```suggestion
class TypedArg:
```
(totally up to you, but avoids wondering what the tagging is about)

testing/btest/cluster/websocket/zeromq/manager-worker-ping-pong.zeek
```diff
@@ -0,0 +1,148 @@
+# @TEST-DOC: Test WebSockets clients when the ZeroMQ cluster backend is enabled.
```
ckreibich, line 1: If you add a bit more detail here then I don't have to piece together from code what sort of ping/pong protocol you have implemented across three components.

testing/btest/cluster/websocket/zeromq/oneshot.zeek
```diff
@@ -0,0 +1,71 @@
+# @TEST-DOC: Connect with a WebSocket client, it sends 100 events and disconnects immediately. Ensure the manager sees all of them. Regression test for ZeroMQ discarding queued messages.
```
ckreibich, line 1: It's tempting to consider merging these two as they're so similar. But, totally a detail.

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -75,6 +75,16 @@ export {
 	## :zeek:see:`Cluster::create_store` with the *persistent* argument set true.
 	const default_persistent_backend = Broker::SQLITE &redef;
 
+	## The default queue size for WebSocket event dispatcher instances.
```
ckreibich, line 78: "default queue size" isn't right since queues grow and shrink by definition — I suggest sticking with "maximum queue size", as you do in OnLoop. So perhaps just `default_max_websocket_event_queue_size` to make it a bit shorter.

src/broker/WebSocketShim.cc
```diff
@@ -180,3 +180,247 @@ void WebSocketShim::ProcessMessage(std::string_view topic, broker::zeek::Invalid
...
+
+    auto make_message = [](int value) { return broker::list_builder{}.add(value).build(); };
+
+    auto make_data_message = [](auto topic, int value) {
+        auto msg = broker::list_builder{}.add(value).build();
+        return broker::make_data_message(broker::topic{topic}, msg);
+    };
+
+    SUBCASE("endpoint-publish") {
+        // Publishing through the endpoint is visible to hubs, but not subscribers.
```
ckreibich, line 224: Thanks for adding this here, very helpful. I wish we had this over in the Broker docs, where I spent a whole bunch of time earlier today trying to figure out [exactly this](https://github.com/zeek/broker/pull/451#discussion_r2055112675).

ckreibich (approved):
Super nice, thanks for making this so bite-sized for reviewing, it really helped. I have a few comments that you can handle as you see fit. I spent most of today on this, and also tinkered a bit locally. Great stuff. :rocket: 

I'm leaving the Broker side to you. If you're okay with it, just approve & merge. I would need way more time for a full review, and I really want to get the backpressure metrics PR out still.

src/broker/Manager.cc
```diff
@@ -1986,4 +1987,11 @@ void Manager::PrepareForwarding(const std::string& name) {
     DBG_LOG(DBG_BROKER, "Resolved table forward for data store %s", name.c_str());
 }
 
+broker::hub Manager::MakeHub(broker::filter_type ft) {
+    ++hub_count;
+    return bstate->endpoint.make_hub(std::move(ft));
+}
+
+void Manager::ReleaseHub(const broker::hub&& hub) { --hub_count; }
```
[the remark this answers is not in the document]
awelzel: Thanks to Benjmain :-)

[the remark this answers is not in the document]
bbannier: FTR, the original discussion was https://github.com/zeek/zeek/pull/4344#discussion_r2048540773.

I originally suggested passing a `const broker::hub&& hub` (no `const`) since that is the "usual" way to consume a value. Accepting an rvalue transfers ownership to the function and e.g., the caller does not need to clean it up anymore (after moving the value into the function they only hold an empty shell). With that the comment on `ReleaseHub` is slightly outdated and should probably be something like _destroy the hub_.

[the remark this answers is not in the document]
awelzel: Thanks, renamed to DestroyHub()

src/cluster/cluster.bif
```diff
@@ -201,3 +201,12 @@ function Cluster::__listen_websocket%(options: WebSocketServerOptions%): bool
 	auto result = zeek::cluster::manager->ListenWebSocket(server_options);
 	return zeek::val_mgr->Bool(result);
 	%}
+
+module Cluster::Backend;
+
+## Generated on cluster backend error.
+##
+## code: A structured error code, not further specified.
```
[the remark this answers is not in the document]
awelzel: I renamed to `tag`, but kept `code` in the WebSocket repsonse, because that's what is specified in the current Broker WebSocket protocol.

src/broker/WebSocketShim.h
```diff
@@ -0,0 +1,78 @@
...
+        throw new std::logic_error("not implemented");
+    }
+    bool DoSubscribe(const std::string& topic_prefix, SubscribeCallback cb) override;
+    bool DoUnsubscribe(const std::string& topic_prefix) override;
+    bool DoPublishLogWrites(const zeek::logging::detail::LogWriteHeader& header, const std::string& format,
+                            zeek::byte_buffer& buf) override {
+        throw new std::logic_error("not implemented");
+    }
+
+    // visit_as_message received from broker::subscriber
```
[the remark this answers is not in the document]
awelzel: I've updated the comment. This is about dispatching messages received from hub.poll() :-(

src/cluster/websocket/WebSocket.cc
awelzel: Ah, yeah, you're right. Keeping as is for explicitness (and I had dabbled with a `ZEEK_DISABLE_BROKER_MGR` the other day where broker_mgr would end up nullptr and I guess then things start to become cryptic).

testing/btest/cluster/websocket/zeromq/manager-worker-ping-pong.zeek
awelzel: Added a few notes about the sceanrio.

testing/btest/cluster/websocket/zeromq/oneshot.zeek
awelzel: Yeah. It'd be nice to parameterize these as some of the tests should just work for Broker or ZeroMQ - the difference is mostly setup/port usage.

I haven't yet tried as I don't obviously have a good pattern in mind.

To a degree, this applies to storage, too. Not for this PR, but I'd be more than happy if we could figure out something here.

scripts/base/frameworks/cluster/main.zeek
awelzel: Great, thanks!

PR #4349, opened 2025-04-10
generic metadata pre work

awelzel:
This PR contains a jumble of commits that were part of the generic metadata PR but can be included before that, too.


src/plugin/Manager.cc
```diff
@@ -962,13 +962,13 @@ void Manager::HookUnprocessedPacket(const Packet* packet) const {
 }
 
 void Manager::MetaHookPre(HookType hook, const HookArgumentList& args) const {
-    if ( hook_list* l = hooks[HOOK_CALL_FUNCTION] )
+    if ( hook_list* l = hooks[META_HOOK_PRE] )
```
timwoj, line 965: Does this change also mean we didn't have any testing for this?

src/cluster/Backend.h
```diff
@@ -269,6 +270,16 @@ class Backend {
...
 
+    /**
+     * @return This backend's implementation name.
+     */
+    const std::string& GetName() const { return name; }
+
+    /**
+     * @return This backend's implementation component tag.
+     */
+    const zeek::Tag& GetTag() const { return tag; }
```
timwoj, line 281: Most of the other functions like this are just named `Name()` and `Tag()`. See for example, `Expr::Tag()`, `Stmt::Tag()` and `Type::Tag()`.

awelzel: I've previously added `GetName()`s elsewhere to return `std::string` and deprecated `const char* Name()`s and also got inspired by `Analyzer::GetAnalyzerName()` and `GetAnalyzerTag()` :-)

I'll go for the short ones. Thanks!

timwoj (approved):

PR #4351, opened 2025-04-11
Prevent event timestamps set to future

J-Gras:
https://github.com/zeek/zeek/commit/831614f907b094b242b02d1ed370ea80818adcac introduced a subtle change: When publishing an event, Broker uses the event's timestamp instead of current network time. For scheduled events, the event timestamp is the intended timestamp. The pitfall: If we force timer expiration, the timestamp might be in the future. Today, this happens on shutdown. This change guarantees that event timestamps are never set beyond network time.

TBH, I'd slightly prefer to keep the intended timestamp locally as is, but only limit timestamps to network time on publishing of events (i.e. only `remote_event_schedule_ts` would change). With the current change we loose the information that an event was scheduled to the future. However, talking to @awelzel this would complicate things for upcoming event metadata changes. I guess we can pick this up when reviewing shutdown in general.

@awelzel: I went for an unconditional `std::min` instead of checking for `is_expire`, because (1) we would need this anyway as the event ts could be less than network time during forced expiration, and (2) we don't risk confusing semantics if we start to force expiration in other places as well.

It would be great if we could backport this to 7.1.


awelzel (approved):
> However, talking to @awelzel this would complicate things for upcoming event metadata changes. 

I've been pondering this from another angle:

Using the intended time in the past makes sense to me: network time jumped, but timers were intended to be executed earlier, so they get an earlier event timestamp. With expiration though, they were scheduled for later. At the same time, I think it's fair to say that draining makes all pending timers to be intended for "right now" (or else we shouldn't even raise them).

 
> because (1) we would need this anyway as the event ts could be less than network time during forced expiration

Yeah, that makes sense. Thanks!

PR #4355, opened 2025-04-14
Pluggable connection tuples

ckreibich:
This is preliminary since I still want to look into a few aspects of the code, but this makes Zeek's notion of flow tuples extensible via plugins. It includes one such (built-in) plugin, which makes flow tuples VLAN-aware.

I'll add details here once I've iterated a bit more.


src/IPAddr.h
```diff
@@ -53,6 +53,8 @@ class ConnKey {
     void Init(const IPAddr& src, const IPAddr& dst, uint16_t src_port, uint16_t dst_port, uint16_t proto, bool one_way);
 };
 
+using ConnKeyPtr = std::shared_ptr<ConnKey>;
```
bbannier, line 57: Complete drive-by comment: It looks like we never intentionally create shared copies of values of this type, but we always incur the overhead of updating the ref count. I believe this should be

```cpp
using ConnKeyPtr = std::shared_ptr<ConnKeyPtr>;
```

Doing that requires cleaning up some functions which currently take `const ConnKeyPtr` (an immutable value, not a ref) where they should just take full ownership by passing `ConnKeyPtr` if they consume the value, or `const ConnKeyPtr` if they just access it.

src/Conn.h
```diff
@@ -63,6 +63,8 @@ struct ConnTuple {
     bool is_one_way = false; // if true, don't canonicalize order
 };
 
+using ConnTuplePtr = std::shared_ptr<ConnTuple>;
```
bbannier, line 68: Ditto.

src/IPAddr.cc
```diff
@@ -126,6 +103,37 @@ void ConnKey::Init(const IPAddr& src, const IPAddr& dst, uint16_t src_port, uint
     transport = proto;
 }
 
+session::detail::Key ConnKey::SessionKey() const {
+    uint8_t* temp = new uint8_t[PackedSize()];
```
timwoj, line 107: `PackedSize()` never returns a different value here. I realize you mentioned storing the result of `Pack` in the future in the commit message, but removing the added calls to `PackedSize` is an easy one.

src/IPAddr.h
timwoj: > using ConnKeyPtr = std::shared_ptr<ConnKeyPtr>;

I'm confused here. Isn't that creating a shared_ptr to the same aliased type?

src/IPAddr.h
```diff
@@ -35,15 +36,14 @@ class ConnKey {
 
     ConnKey(const IPAddr& src, const IPAddr& dst, uint16_t src_port, uint16_t dst_port, uint16_t proto, bool one_way);
     ConnKey(const ConnTuple& conn);
-    ConnKey(const ConnKey& rhs) { *this = rhs; }
```
timwoj, line 38: I think you could leave this there, just remove the `memset`.

src/Conn.h
```diff
@@ -73,6 +73,7 @@ static inline int addr_port_canon_lt(const IPAddr& addr1, uint32_t p1, const IPA
 
 class Connection final : public session::Session {
 public:
+    Connection(const detail::ConnKeyPtr k, double t, const ConnTuple* id, uint32_t flow, const Packet* pkt);
     Connection(const detail::ConnKey& k, double t, const ConnTuple* id, uint32_t flow, const Packet* pkt);
```
timwoj, line 77: I agree with deprecating this one, unless you want to make `ConnKey` inherit from `std::enable_shared_from_this`.

src/Conn.cc
```diff
@@ -27,7 +27,7 @@ namespace zeek {
 uint64_t Connection::total_connections = 0;
 uint64_t Connection::current_connections = 0;
 
-Connection::Connection(const detail::ConnKey& k, double t, const ConnTuple* id, uint32_t flow, const Packet* pkt)
+Connection::Connection(const detail::ConnKeyPtr k, double t, const ConnTuple* id, uint32_t flow, const Packet* pkt)
```
timwoj, line 31: Taking the `ConnKeyPtr` as a non-reference here causes two ref-counts, one for the copy into the argument variable and one for the copy into the `key`. Arguably, this could either be a reference or you could use std::move for the assignment to `key`.

scripts/policy/protocols/conntuple/vlan.zeek
```diff
@@ -0,0 +1,14 @@
+##! This script adapts Zeek's connection tuples to include 802.1Q VLAN and
+##! Q-in-Q tags, when available. Zeek normally ignores VLAN tags in its flow
+##! lookups; this change makes it factor them in and also makes those VLAN tags
+##! part of the conn_id record.
+
+redef record conn_id += {
+	## The outer VLAN for this connection, if applicable.
+	vlan: int      &log &optional;
```
timwoj, line 8: Is there any concern about name collisions if someone accidentally loads the script for multiple tuples?

src/conntuple/vlan/Builder.cc
```diff
@@ -0,0 +1,108 @@
...
+    auto res = std::make_shared<VlanConnKey>(tuple);
+    res->vlan = vtuple.vlan;
+    res->inner_vlan = vtuple.inner_vlan;
+    return res;
+}
+
+zeek::detail::ConnKeyPtr Builder::GetKey(Val* v) { return std::make_shared<VlanConnKey>(v); }
+
+void Builder::FillConnIdVal(detail::ConnKeyPtr key, RecordValPtr& tuple) {
+    if ( tuple->NumFields() <= 5 )
```
timwoj, line 95: Should this report an error of some sort?

timwoj (changes requested):

src/IPAddr.h
bbannier: I updated the suggestion, this should have just been a `unique_ptr` if at all.

awelzel (commented):
As mentioned in direct chats, leaving some notes here.

Running this branch with two pcaps from the benchmarker setup in bare mode indicates this introduces quite some overhead (1.05x and 1.06x). Pcaps are available in `/mnt/data/test_data` on the os-perf-1 system).


|:---|---:|---:|---:|---:|
| `taskset -c 0,1,2,3 /opt/zeek-dev-prod-master/bin/zeek -b -C -r /media/awelzel/T7/pcaps/50k-tcp-conns.pcap Log::default_writer=Log::WRITER_NONE` | 12.666 ± 0.097 | 12.538 | 12.827 | 1.00 |
| `taskset -c 0,1,2,3 /opt/zeek-dev-prod/bin/zeek -b -C -r /media/awelzel/T7/pcaps/50k-tcp-conns.pcap Log::default_writer=Log::WRITER_NONE` | 13.456 ± 0.119 | 13.314 | 13.695 | 1.06 ± 0.01 |
| `taskset -c 0,1,2,3 /opt/zeek-dev-prod-master/bin/zeek -b -C -r /media/awelzel/T7/pcaps/awelzel/justin-smb/smb_many_open_files.pcap Log::default_writer=Log::WRITER_NONE` | 1.678 ± 0.018 | 1.662 | 1.717 | 1.00 |
| `taskset -c 0,1,2,3 /opt/zeek-dev-prod/bin/zeek -b -C -r /media/awelzel/T7/pcaps/awelzel/justin-smb/smb_many_open_files.pcap Log::default_writer=Log::WRITER_NONE` | 1.767 ± 0.015 | 1.746 | 1.800 | 1.05 ± 0.01 |

(zeek-dev-prod-master being when this PR was branched from master)

I haven't looked if this is due to `shared_ptr` or the `Pack()` / malloc overhead or something else. For the `shared_ptr`, if we don't need the thread-safety guarantees, can probably switch to IntrusivePtr with explicit `Ref/Unref` (rather than inheriting from Obj).

I'm also wondering if ConnKey could/should hold a `ConnTuple`, going back to an approach where we use the memory of an instance directly as key rather than re-encoding it into a separate buffer. But haven't yet tried this.

---

I'm somewhat overwhelmed by the `ConnTuple`, `ConnKey` and `session::detail::Key` indirections, when at the same time this all seems to be confined to  IPBasedAnalyzer currently. Essentially, I'm wondering whether the `ConnTuple::builder` should be specific to the IPBasedAnalyzer rather than global.

src/conntuple/Builder.h
```diff
@@ -0,0 +1,35 @@
...
+public:
+    Builder();
+    virtual ~Builder();
+
+    virtual ConnTuplePtr GetTuple(const Packet* pkt);
+
+    virtual zeek::detail::ConnKeyPtr GetKey(const ConnTuple& tuple);
+    virtual zeek::detail::ConnKeyPtr GetKey(Val* v);
+
+    virtual void FillConnIdVal(detail::ConnKeyPtr key, RecordValPtr& tuple) {};
```
awelzel, line 29: Hey @ckreibich - I've dabbled with this and tried a different design:

The following is a bit of a class diagram which I had in mind. This is essentially "abstract factory" pattern:

ConnKey Factory

The main idea is that behavior customization should be done in `ConnKey` implementations (e.g. via `ConnKey::DoInit(Packet pkt)`, rather than the logic being somewhere across `Builder::GetTuple()` and `Builder::GetKey()`. The other thought is a bit that the `ConnTuple` as it exists today is very IP specific and IMO the `ConnKey` interface probably shouldn't even know the `ConnTuple` concept.

I've also dropped the `Pack()` and `PackedSize()` methods from the API and instead moved to an API which only returns a `session::detail::Key` as a view over some bytes that represent the session key. How that memory is allocated and/or initialized should be up to the `ConnKey` implementation (within the branch I returned to an approach where it directly points into the ConnKey instance - e.g. for VLAN here: https://github.com/zeek/zeek/blob/d07a0a9da766c31de58a159bcfeba0111f14b89c/src/conntuple/vlan/Builder.cc#L53-L62


The branch is here - this isn't API compatible, only "hacked up" what seemed to look nice. The Builders are still named Builder, though Factory would be the better name in that approach.

https://github.com/zeek/zeek/tree/topic/awelzel/conntuple-dabbling

I've perf tested this a bit and while it doesn't have zero-overhead, it fares much better than the change in this PR. Staring at flamegraphs, there might be more value in removing the ip_hdr shared ptr in use :-/ 

Happy to talk this through!

JustinAzoff:
One thing I think that we need as part of this is a way to mask or truncate components out of a connection.

Take the simple-scan package for example.  Currently this hard codes the fact that the attacker is id.orig_h and the victim is id.resp_h + id.resp_p. A more flexible system would be a way to have something like

    attacker = mask_id(c$id, ID_MASK_ORIG_P | ID_MASK_RESP_H | ID_MASK_RESP_P)
    victim = mask_id(c$id, ID_MASK_ORIG_H | ID_MASK_ORIG_P)

for a normal connection this would give `attacker = orig_h` but if the connection was amended to include the vlan then `attacker = (vlan, orig_h)`.

I think this isn't 100% perfect idea because vlan could mean orig_vlan and resp_vlan, and we would only want orig_vlan in this case.

This same problem exists in scripts like known_hosts.  We assume that the known host is simply `id$orig_h` or `id$resp_h` when there may be overlapping address space and the vlan is required to uniquely identify the host.

We have seen this problem a lot with people that load https://github.com/corelight/log-add-vlan-everywhere but then are surprised when a number of logs don't contain the vlan information.

ckreibich:
Yeah Justin. I need to give some context: we kicked around different use cases for extensibility prior to this PR. We quickly hit the distinction between flow tuples and endpoints. This work only targets the definition of connections, how Zeek manages these internally, and how they manifest in scriptland via `conn_id`. It does not provide an extensible notion of endpoints. An application that currently identifies endpoints as IP addresses will need refactoring to make use of additional tuple information. (We considered ideas around extending the `addr` type to avoid this, but decided not to pursue that further.)

Zeek should  _also_ gain an extensible notion of endpoints, complementary to this work. We decided to tackle the connection part first. I could see simple-scan initially becoming aware of flow tuple context that includes VLANS, or Geneve VNIs, etc, on a case-by-case basis, and in a subsequent step move to a better notion of endpoints.

So, duly noted. :+1:

ckreibich:
I'm closing this in favor of #4579, since it's redoing the whole branch anyway. Thanks everyone for the input!

PR #4357, opened 2025-04-14
IXWebSocket: Bump to latest upstream master

awelzel:
The PR for the threading issue was merged. Still keeping the submodule to point at our own fork as I have a hunch there might be more.

---

Reference: https://github.com/machinezone/IXWebSocket/pull/540


0xxon:
The reference link seems dead :)

PR #4362, opened 2025-04-15
Replace dpd.log with analyzer_failed.log

0xxon:
This PR overhauls the logging of analyzer violations.

Specifically:

#### `dpd.log` deprecation

`dpd.log` is deprecated, and no longer produced by default. One can load `policy/frameworks/analyzer/dpd-log.zeek` to create `dpd.log`. Loading this script will raise an analyzer violation.

#### `analyzer.log` renamed to `analyzer_debug.log` and disabled by default.

The current `analyzer.log` is mostly useful for development, and less so for incident response use. Thus, the logfile was renamed to `analyzer-debug.log`. It is no longer created by default - to create it users have to load the `policy/frameworks/analyzer/analyzer-debug-log.zeek` policy script. The options to log analyzers being confirmed and removed are now enabled by defaul.

#### new `analyzer.log`

A new `analyzer.log` is created by default. This log file by default contains:
* protocol analyzers which were first confirmed and then raised a violation that caused them to be removed
* file analyzers that raised a violation
* packet analyers that raised a violation

Violations for protocol analzers are also logged if the analyzer detached by the time the violation is handled - so at the very end of a connection

We currently do not use the concept of confirmations for file and packet analyzers. At least for file analyzers it seems that it would make sense to have and use this concept - as we do not currently have a lot file analyzers this should be relatively simpe to add.

#### new analyzer_failed event

The event is raised after a protocol analyzer raised a violation _and_ was removed, or if the analyzer was already detached by the time the violation was handled.

#### handling of conn.log entries (removal of service field)

The handling of the removal of services in `conn.log` is aligned with entries in `analyzer_failed.log`. Previously there were different logic paths - and it could happen that service violations caused entries to be removed from the `service` field in `conn.log`, even if the analyzer was not disabled due to the violation.

This change ended up being much larger than I hoped - mostly due to test changes. I hope that having a more unified view and handling of violations is worthwhile - this removed a number of weird, essentially unexplainable corner-cases. I am not entirely sure that the log file naming is optimal - I am open to different ideas for that. I could also see just keeping the new-style log as `dpd.log` to reduce friction; however the naming of `dpd.log` has not really been appropriate for a long time, so...


scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -163,6 +172,23 @@ export {
 	##
 	## This set can be added to via :zeek:see:`redef`.
 	global requested_analyzers: set[AllAnalyzers::Tag] = {} &redef;
+
+	## Event that is raised when an analyzer raised a service violation and was
+	## removed.
```
awelzel, line 177: The extra event definitely helps understand the control flow!

I have a nit/naming thoughts: `analyzer_violation_info()` and `analyzer_failed()` are pretty close, but the "analyzer was removed" is a significant difference between them. So was pondering proposing `analyzer_removed(...)`. But that's not great due to removals happening for benign reasons (`SSH`, `SSL`). 

Maybe you have another idea, otherwise feel free to resolve, anything I have in mind right now ends-up being a mouthful - e.g `violating_analyzer_removed()` :-(

scripts/base/frameworks/analyzer/analyzer-failed-log.zeek
```diff
@@ -0,0 +1,116 @@
...
+module Analyzer::Logging;
+
+export {
+	## Add the analyzer logging stream identifier.
+	redef enum Log::ID += { LOG };
+
+	## The record type defining the columns to log in the analyzer-failed logging stream.
+	type Info: record {
+		## Timestamp of the violation.
+		ts: 						time        &log;
```
awelzel, line 15: Something is off here with the tabs :-)

scripts/policy/frameworks/analyzer/packet-segment-logging.zeek
```diff
@@ -0,0 +1,50 @@
...
+# stash the packet segment in the event causing the violation, so that it can be retrieved later.
+event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo) &priority=4
+	{
+	if ( ! is_protocol_analyzer(atype) && ! is_packet_analyzer(atype) )
+		return;
+
+	if ( ! info?$c || ! info?$aid )
+		return;
+
+	info$c$packet_segment = fmt("%s", sub_bytes(get_current_packet()$data, 0, packet_segment_size));
```
awelzel, line 35: Not that it matters, but think this is equivalent and feels a bit more modern :-)

```suggestion
	info$c$packet_segment = fmt("%s", get_current_packet()$data[:packet_segment_size]);
```

scripts/policy/frameworks/analyzer/packet-segment-logging.zeek
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+hook Analyzer::Logging::log_policy(rec: Analyzer::Logging::Info, id: Log::ID, filter: Log::Filter)
```
awelzel, line 38: Interesting.

What do you think about an explicit policy hook in `Analyzer::Logging` similar to what is in `Notice` to enable this? Or feel like that is weird/old-style? To me the reliance on the generic log policy stream seems a bit weird.

E.g.
```
module Analyzer::Logging;
    # Customization point for logging.
    global policy: hook(rec: Analyzer::Logging::Info, info: AnalyzerViolationInfo);
```

This way could also use `info$c` directly, rather than the lookups.

NEWS
```diff
@@ -65,6 +65,28 @@ New Functionality
 Changed Functionality
 ---------------------
 
+- Logging of failed analyzers has been overhauled. `dpd.log` was replaced with
+  a new `analyzer_failed.log` that presents a more unified and consistent view
+  of failed analyzers.
+
+  For protocol analyzers, `analyzers_failed.log` contains analyzers that
```
awelzel, line 72:
```suggestion
  For protocol analyzers, `analyzer_failed.log` contains analyzers that
```

NEWS
```diff
@@ -65,6 +65,28 @@ New Functionality
...
+  reported a protocol violation that caused Zeek to decide to remove the violating
+  analyzer from the connection.
```
awelzel, line 74: I saw it's in the PR cover letter - probably should mention the "analyzer was previously confirmed" here as well. I guess everyone except for me is expecting/assuming that anyhow. Might still be good to make explicit mention here and/or somewhere in the docs that failure and removal of non-confirmed protocol analyzers aren't part of the log because these aren't interesting (dpd.log always worked like that, but analyzer.log didn't).

awelzel (commented):
Hey, lots of nits and naming and random thoughts - I'd just "approve", but will use "comment" for now. Also, I didn't check the baselines in detail yet, will do later today for a bit.

I hope you get a second review from Christian or Robin. I have no concerns, other than some naming regularization here and there, or smallish alternative ideas that might be worth trying.

Either `analyzer.log` or `analyzer_violations.log` would get a vote.

> and it could happen that service violations caused entries to be removed from the service field in conn.log, even if the analyzer was not disabled due to the violation.

Do you have if there's something in the test base where this is/was the case?

scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -163,6 +172,23 @@ export {
...
+	##
+	## The event is also raised if the analyzer already was no longer active by
+	## the time that the violation was handled - so if it happens at the very
+	## end of a connection.
+	##
+	## Currently this event is only raised for protocol analyzers, as packet
+	## and file analyzers are never actively removed/disabled.
+	##
+	## ts: time at which the violation occurred
```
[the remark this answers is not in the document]
0xxon: I admittedly have no strong feelings one way or the other. I am tempted to leave as is, as this is easier, how things were handled before.

scripts/policy/frameworks/analyzer/analyzer-debug-log.zeek
```diff
@@ -1,11 +1,10 @@
-##! Logging analyzer confirmations and violations into analyzer.log
+##! Logging analyzer confirmations and violations into analyzer-debug.log
```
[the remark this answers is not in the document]
0xxon: I just took a peek at what else we do with our Zeek scripts that perform logging. In general, the postfix `-logging` seems commonly used, so I am more tempted by `debug-logging`.

scripts/policy/frameworks/analyzer/packet-segment-logging.zeek
0xxon: I'm a bit torn on that. Not needing a lookup for the connection seems like a plus. On the other hand - this kind of scenario seems to be exactly what the log_policy hooks should be here for.

Having to add a secondary hook for the cases where we actually want to modify the logs in a policy seems weird to me.

scripts/base/frameworks/analyzer/main.zeek
0xxon: yeah, I had problems coming up with a name that I liked.

`analyzer_removed` seems more confusing to me than the current state. If anyone else has a nice idea for a name - please let me know.

scripts/base/frameworks/analyzer/analyzer-failed-log.zeek
```diff
@@ -0,0 +1,116 @@
+##! Logging analyzer  violations into analyzer_failed.log
```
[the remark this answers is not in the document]
0xxon: Similar to below, keeping in line with other cases in the source tree, `logging.zeek` seems more natural.

testing/btest/Baseline/scripts.base.protocols.rdp.rdp-invalid-length/analyzer_debug.log
```diff
@@ -3,11 +3,13 @@
 #set_separator	,
 #empty_field	(empty)
 #unset_field	-
-#path	analyzer
+#path	analyzer_debug
 #open XXXX-XX-XX-XX-XX-XX
 #fields	ts	cause	analyzer_kind	analyzer_name	uid	fuid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	failure_reason	failure_data
 #types	time	string	string	string	string	string	addr	port	addr	port	string	string
 XXXXXXXXXX.XXXXXX	violation	protocol	RDP	CHhAvVGS1DHFjwGM9	-	10.0.0.1	45257	10.0.0.2	3389	Binpac exception: binpac exception: &enforce violation : RDP_Negotiation_Response:length	-
+XXXXXXXXXX.XXXXXX	disabled	protocol	RDP	CHhAvVGS1DHFjwGM9	-	10.0.0.1	45257	10.0.0.2	3389	-	-
```
awelzel, line 11: Just from skimming some baselines and definitely can be follow-up to avoid extra diffs: What would you think about having `aid: count &log &optional` added to `analyzer_debug.log` ? You mentioned that would also be useful for file analyzers, even if we don't have it right now.

This seems to be an example of enabling RDP 3 times on the same connection, A separate aid column would further show that it's 3 different instances of the analyzer.

scripts/policy/frameworks/analyzer/analyzer-debug-log.zeek
```diff
@@ -159,7 +146,7 @@ event analyzer_confirmation_info(atype: AllAnalyzers::Tag, info: AnalyzerConfirm
 	local rec = Info(
 		$ts=network_time(),
 		$cause="confirmation",
-		$analyzer_kind=analyzer_kind(atype),
+		$Analyzer::kind=Analyzer::kind(atype),
```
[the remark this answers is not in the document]
0xxon: indeed. I was actually kind of surprised that this worked without any complaints.

testing/btest/Baseline/scripts.base.protocols.rdp.rdp-invalid-length/analyzer_debug.log
0xxon: I would probably put that into a follow-up PR, but... I like it.

scripts/policy/frameworks/analyzer/dpd-log.zeek
```diff
@@ -0,0 +1,91 @@
+##! Creates the now deprecated dpd.logfile.
+
+@deprecated("dpd.log is deprecated; remove in 8.1")
```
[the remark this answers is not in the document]
0xxon: Updated as suggested

0xxon:
I renamed the scripts as suggested, and the logfile name is now `analyzer.log`.

testing/btest/scripts/policy/frameworks/logging-debug.zeek
```diff
@@ -0,0 +1,21 @@
+# @TEST-EXEC: zeek -r ${TRACES}/socks.trace %INPUT
```
awelzel, line 1: Mind moving this file into `frameworks/analyzer` instead of `frameworks/`? (totally randomly found while skimming)

testing/btest/scripts/policy/frameworks/dpd/packet-segment-logging.zeek
```diff
@@ -1,4 +1,4 @@
-# @TEST-DOC: IPv6 connection from external ipv6.pcap triggering FTP analyzer violation. Check dpd.log contains the right packet_segment
+# @TEST-DOC: IPv6 connection from external ipv6.pcap triggering FTP analyzer violation. Check analyzer.log contains the right packet_segment
```
awelzel, line 1: Nit, this continues to check `dpd.log`.

NEWS
```diff
@@ -36,6 +36,31 @@ New Functionality
...
 ---------------------
 
+- Logging of failed analyzers has been overhauled. `dpd.log` was replaced with
+  by a new `analyzer.log` that presents a more unified and consistent view
+  of failed analyzers. The previous `analyzer.log` was renamed to `analyzer-debug.log`;
+  see below for more details.
+
+  For protocol analyzers, `analyzer.log` contains analyzers that
+  were previously confirmed and then reported a protocol violation that caused
+  Zeek to decide to remove the violating analyzer from the connection.
```
awelzel, line 46:
```suggestion
  Zeek to remove the violating analyzer from the connection.
```

scripts/policy/frameworks/analyzer/deprecated-dpd-log.zeek
```diff
@@ -0,0 +1,89 @@
+##! Creates the now deprecated dpd.logfile.
+
```
awelzel, line 2: Could make sense to add a `# Remove in v8.1` comment here so that it doesn't fall through the cracks when Tim greps for deprecations after 8.0?

scripts/policy/frameworks/dpd/packet-segment-logging.zeek
```diff
@@ -1,9 +1,13 @@
+@deprecated("Please switch to frameworks/analyzer/packet-segment-logging, which logs to analyzer-failed.log. Remove in 8.1")
```
awelzel, line 1:
```suggestion
@deprecated("Please switch to frameworks/analyzer/packet-segment-logging, which logs to analyzer.log. Remove in 8.1")
```

testing/btest/Baseline/scripts.base.protocols.dce-rpc.ntlm-unterminated-av-pair-seq/analyzer.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	analyzer
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	analyzer_kind	analyzer_name	uid	fuid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	failure_reason	failure_data
```
awelzel, line 8: Just with a fresh look, the `analyzer_` prefix on these fields (if memory serves right) was due to the original `analyzer.log` also having `cause` and possibly other information for which `name` or `kind` could apply to (think `violation_kind`). I'm wondering if it's still relevant now that this log is strictly about analyzer violations after confirmations. There should be no confusion of what `kind` and `name` apply to.

Ignore, but found it worth to mention as it's a bit redundant.

awelzel (approved):
I skimmed through the latest version. LGTM, but would suggest getting another :heavy_check_mark:  from someone.

Thanks!

testing/btest/Baseline/scripts.base.protocols.dce-rpc.ntlm-unterminated-av-pair-seq/analyzer.log
0xxon: I honestly don't have a strong opinion one way or another. So - if there's another vote for this, I am happy to change it.

NEWS
```diff
@@ -36,6 +36,31 @@ New Functionality
...
 ---------------------
 
+- Logging of failed analyzers has been overhauled. `dpd.log` was replaced
+  by a new `analyzer.log` that presents a more unified and consistent view
+  of failed analyzers. The previous `analyzer.log` was renamed to `analyzer-debug.log`;
+  see below for more details.
+
+  For protocol analyzers, `analyzer_failed.log` contains analyzers that
+  were previously confirmed and then reported a protocol violation that caused
+  Zeek to remove the violating analyzer from the connection.
```
```suggestion
  For protocol analyzers, `analyzer.log` now reports initially confirmed analyzers that
  Zeek subsequently removed from from the connection due to a protocol violation.
```
(Similar tweaks for any mentioning of `analyzer_failed` below, since it doesn't exist — needs an edit pass)

NEWS
```diff
@@ -36,6 +36,31 @@ New Functionality
...
+  For protocol analyzers, `analyzer.log` contains analyzers that
+  were previously confirmed and then reported a protocol violation that caused
+  Zeek to remove the violating analyzer from the connection.
+
+  For file and packet analyzers, all errors will be logged to `analyzer.log`.
+
+  If you compare the output between `dpd.log` and `analyzer.log` you might
+  notice discrepancies in some corner cases. For `dpd.log`, the logic paths of
+  logging violations and removing analyzers were separate - so an entry in
+  `dpd.log` does not necessarily mean that an analyzer ware removed from a connection.
```
ckreibich, line 53: I don't follow this part — this is referring to the former logs, yes? It's confusing because above we just said that dpd.log is now analyzer.log. I suggest dropping the whole paragraph.

NEWS
```diff
@@ -36,6 +36,31 @@ New Functionality
...
+
+  As part of this work, a new `analyzer_failed` event has been introduced.
```
ckreibich, line 55: It would help to say a few words here about what that event does.

NEWS
```diff
@@ -36,6 +36,31 @@ New Functionality
...
+
+- `analyzer.log` was renamed to `analyzer_debug.log`, and is no longer created
+  by default. The log file will be created if the `frameworks/analyzer/debug-logging.zeek`
+  policy script is loaded.
+
+  Note that the namespace for options in the script changed. Furthermore the default
```
ckreibich, line 61: Spell it out?

scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -88,6 +88,15 @@ export {
...
 	global name: function(tag: Analyzer::Tag) : string;
 
+	## Translates an analyzer type to a string with the analyzer's type.
+	##
+	## Possible values are "protocol", "packet", "file", or "unknown".
+	##
+	## tag: The analyzer tag.
+	##
+	## Returns: The analyzer kind corresponding to the tag.
+	global kind: function(tag: Analyzer::Tag) : string;
```
ckreibich, line 98: Would it be cleaner to map this to an enum, also in analyzer.log? I've now repeatedly heard that it improves the logs' quality if they automatically carry as much typing/conciseness as possible, and strings looks a bit looser here than necessary.

ckreibich (changes requested):
Johanna could you rebase this so changes are self-contained? I found this hard to review commit-by-commit because I didn't initially realize that the later commits change much of the PR, and made comments only to realize that they later got addressed. I'd like to make another pass once it's in that state. The result looks great though, thank you. :+1: 

One more thing:  with `policy/frameworks/dpd/packet-segment-logging.zeek` now deprecated, there's only going to be `policy/frameworks/dpd/detect-protocols.zeek` left in that directory. There's also no `scripts/base/frameworks/dpd/`, breaking our usual symmetry. How about we move `detect-protocols.zeek` into `policy/frameworks/analyzer` as well at this time (leaving only an `@load` of the new one, and a deprecation note, in the old place)?

J-Gras (commented):
I love how this turned out! I believe this will vastly improve transparency of analyzer management by Zeek.

There is just one thing that troubles me. I really don't like calling the reworked log `analyzer.log`. This is for two reasons:
1. The name does not convey what to expect in the log (`analyzer-failures.log` does).
2. The name was previously used for a different log with different schema and, in particular, different semantics! I am convinced that reusing the name will lead to substantial confusion.

Last but not least, I'd love to have attaching of analyzers logged in the `analyzer-debug.log` as well 😇

scripts/policy/frameworks/analyzer/debug-logging.zeek
```diff
@@ -0,0 +1,207 @@
...
+		failure_reason: string            &log &optional;
+
+		## Data causing failure or violation if available. Truncated
+		## to :zeek:see:`Analyzer::DebugLogging::failure_data_max_size`.
+		failure_data:   string            &log &optional;
+	};
+
+	## Enable logging of analyzer violations and optionally confirmations
+	## when :zeek:see:`Analyzer::DebugLogging::include_confirmations` is set.
+	option enable = T;
```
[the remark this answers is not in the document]
0xxon: Same as above - this PR doesn't really touch the analyzer-debug log, besides renaming it. I'm open to fixing this, but it seems a separate issue.

testing/btest/Baseline/scripts.base.protocols.dce-rpc.ntlm-unterminated-av-pair-seq/analyzer.log
ckreibich: Sigh, I've stared at this for too long. I land on keeping long names, and the field ordering. "kind" and "name" are very generic, particularly if they remain strings, so I think it helps intuition. We have other instances of similar not-strictly-necessary-but-helpful naming, like "tunnel_type" in tunnel.log or "software_type" in software.log.

I _am_ tempted to recommend changing "analyzer_kind" to "analyzer_type" because we use "_kind" nowhere else but use "_type" a bunch. I'll leave that one to you.

ps: our average field length is 9.9 characters. Now you know! :-)

awelzel: > Sigh, I've stared at this for too long.

Naming joy.

> I am tempted to recommend changing "analyzer_kind" to "analyzer_type" because we use "_kind" nowhere else but use "_type" a bunch. I'll leave that one to you.

As background, "kind" was an attempt to find terminology to distinguish analyzers on a higher level: protocol, packet and file analyzers. I feel like "type" is used to describe "SSH" vs "HTTP" or "PE" vs "ZIP" analyzers. Not a strong opinion if that argument doesn't hold for you. The other words I dabbled with back then was "analyzer_class" or "analyzer_family" fwiw.

Maybe we need a "definitions" or "terminology" page in the docs at some point? :-)

> ps: our average field length is 9.9 characters. Now you know! :-)

[logschema](https://github.com/zeek/logschema/) package ftw? :+1:

scripts/policy/frameworks/analyzer/debug-logging.zeek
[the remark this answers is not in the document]
0xxon: The debug-logging script is not new; it essentially is exactly the script that generated the old `analyzer.log`, which comes with these options.

Thus, I am tempted to just leave things as they are.

scripts/policy/frameworks/analyzer/debug-logging.zeek
```diff
@@ -0,0 +1,207 @@
...
+		analyzer_name:  string            &log;
+		## Connection UID if available.
+		uid:            string            &log &optional;
+		## File UID if available.
+		fuid:           string            &log &optional;
+		## Connection identifier if available
+		id:             conn_id           &log &optional;
+
+		## Failure or violation reason, if available.
+		failure_reason: string            &log &optional;
```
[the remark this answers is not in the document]
0xxon: Honestly, I have no strong feelings either for or against this.

The comment is lifted like this from the existing script that generates `analyzer.log`. As such, I have no further insights on this comment.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -45,17 +28,12 @@ export {
 }
 
 redef record connection += {
-	dpd: Info &optional;
 	## The set of services (analyzers) for which Zeek has observed a
 	## violation after the same service had previously been confirmed.
```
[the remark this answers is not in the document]
0xxon: True, that is a regression. Let me fix this, so that `service_violation` has the same behavior as in the past.

scripts/base/frameworks/analyzer/main.zeek
0xxon: I am fine with this one way or another. I think this is mostly up to preference.

scripts/base/frameworks/analyzer/logging.zeek
```diff
@@ -200,19 +88,29 @@ event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationI
...
-		$analyzer_kind=analyzer_kind(atype),
-		$analyzer_name=Analyzer::name(atype),
-	);
+	if ( ! info?$c )
+		return;
 
-	populate_from_conn(rec, c);
+	# log only for previously confirmed service that did not already log violation
+	local analyzer_name = Analyzer::name(atype);
+	if ( analyzer_name !in info$c$service || analyzer_name in info$c$service_violation )
```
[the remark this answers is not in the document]
0xxon: That can indeed happen. I will amend the description.

scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -163,6 +172,23 @@ export {
...
+	##
+	## Currently this event is only raised for protocol analyzers, as packet
+	## and file analyzers are never actively removed/disabled.
+	##
+	## ts: time at which the violation occurred
+	##
+	## atype: atype: The analyzer tag, such as ``Analyzer::ANALYZER_HTTP``.
+	##
+	##info: Details about the violation. This record should include a :zeek:type:`connection`
+	global analyzer_failed: event(ts: time, atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo);
```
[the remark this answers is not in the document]
0xxon: I kind of like it like this, as it opens up more possible use-cases.

scripts/policy/frameworks/analyzer/debug-logging.zeek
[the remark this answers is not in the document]
awelzel: > Would it make sense to choose a more general reason here and data below?

Renaming makes sense to me and leaves the option open to reuse data or reason for something like analyzer_attached. E.g. reason=dpd or reason=port and maybe data could be the signature or the matching packet.

> As such, I have no further insights on this comment.

Probably me not yet knowing where this should be going.

scripts/base/frameworks/analyzer/dpd.zeek
```diff
@@ -1,31 +1,11 @@
-##! Activates port-independent protocol detection and selectively disables
-##! analyzers if protocol violations occur.
+##! Disables analyzers if protocol violations occur.
+
+## Historically, this file also handled the port-independent protocol detection.
+## This is now handled by the analyzer framework.
```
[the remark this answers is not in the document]
0xxon: Will remove.

J-Gras:
> Does this sway you at all?

@ckreibich I get where you are coming from, still I don't buy it 😄 The semantic changes unfold as:
`dpd.log` --> `analyzer.log`
`analyzer.log` --> `analyzer-debug.log`
What makes this tricky is that you might look at the `analyzer.log` and assume this is just about a schema change, when there is actually a change in the underlying logic.

> * This is just my take, but I think the nuances around enabling, violation, removal, "analyzer" vs "DPD", etc have long been lost on regular users. If you want to know something about your analyzers, you go to analyzer.log, and that should be it. Everything else is for developers.

The above is a more general thing: I agree that `analyzer.log` has been a power user feature so far. However, the way Zeek attaches analyzers is an essential feature. For reasoning about the visibility that Zeek provides, this is a step in the critical path. Hence, it is important for regular users.
I am definitely biased here, because I have been trying to explain DPD to users, which was challenging due to its unnecessary complexity. Luckily, most of this is gone now. That's why I think we should take the chance and make DPD as easily accessible to regular users as possible.
That being said, I am okay if I get outvoted. Leaves me with something to explain 😉

scripts/policy/frameworks/analyzer/debug-logging.zeek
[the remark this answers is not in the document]
0xxon: Ok, if we rename this here - do we also want to rename it in analyzer.log?

I actually started going down that road, and then noticed that:
* we do use `failure_reason` in other logs - specifically `packet_logging`.
* For `analyzer.log`, just using `reason` feels a bit weird to me - because in that case it definitely is a failure. `reason` feels a bit generic to me in that log.

For this log, it makes a lot of sense, especially if we use the reason for things besides failures.

0xxon:
Having talked to several people so far - I think we should stick with the `analyzer.log` name for the log. That's the name that would have made sense for it from the beginning.

I admit that the filename change is unfortunate - but so far no one I talked to seemed very concerned about this. I am going to continue with the current filenames, unless I get more people besides @J-Gras who are unhappy with the naming.

scripts/base/frameworks/analyzer/dpd.zeek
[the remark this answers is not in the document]
0xxon: Thinking about it, you are right - we need this field to only cintain the analyzers that were removed where this was not vetoed, for the remaining logic to work. 

I will update the description and rename the field.

In theory we could provide the old field for people that might be using this - however I would immediately mark it as deprecated. I am going to check if anyone is currently using this.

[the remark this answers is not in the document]
0xxon: Need to keep service_violation for the deprecated dpd.log. So - deprecated field it is.

scripts/policy/frameworks/analyzer/debug-logging.zeek
```diff
@@ -34,11 +34,11 @@ export {
 		id:             conn_id           &log &optional;
 
 		## Failure or violation reason, if available.
-		failure_reason: string            &log &optional;
+		reason:         string            &log &optional;
```
ckreibich: Ah — I thought in today's call we were talking about renaming reason and data in _both_ analyzer.log and this one. I don't think those field names should differ, that looks weird. Since `analyzer_debug.log` logs both confirmations and violations, isn't it better to keep `failure_reason` and `failure_data` here since they will only ever be populated with failure context (not confirmation context)? That'd be my preference.

I swear I don't mean to torture you. :-)

testing/btest/Baseline/scripts.base.protocols.dce-rpc.ntlm-unterminated-av-pair-seq/analyzer.log
ckreibich: > Naming joy.

It's so painful, seriously.

> I feel like "type" is used to describe "SSH" vs "HTTP" or "PE" vs "ZIP" analyzers.

Ah, we differ there. To me that's the analyzer's name. I'd still prefer `_type` for the field, but as Johanna would put it, not a hill I'm prepared to die on. :-)

> [logschema](https://github.com/zeek/logschema/) package ftw? 👍

Exactly. :-)

NEWS
```diff
@@ -114,6 +114,28 @@ New Functionality
 Changed Functionality
 ---------------------
 
+- Logging of failed analyzers has been overhauled. `dpd.log` was replaced
+  by a new `analyzer.log` that presents a more unified and consistent view
+  of failed analyzers. The previous `analyzer.log` was renamed to `analyzer-debug.log`;
+  see below for more details.
+
+  For protocol analyzers, `analyzer.log` now reports initially confirmed analyzers that
+  Zeek subsequently removed from from the connection due to a protocol violation.
```
```suggestion
  Zeek subsequently removed from the connection due to a protocol violation.
```

ckreibich:
>  That being said, I am okay if I get outvoted. Leaves me with something to explain 😉

@J-Gras, LOL! :-)

ckreibich (commented):
Thanks for that pass, Johanna! This is all looking good to me now except for that one field naming quibble, sorry to drag that out.

Also let me repeat this from the last review though because I think it got lost: with policy/frameworks/dpd/packet-segment-logging.zeek now deprecated, there's only going to be policy/frameworks/dpd/detect-protocols.zeek left in that directory. There's also no scripts/base/frameworks/dpd/, breaking our usual symmetry. How about we move detect-protocols.zeek into policy/frameworks/analyzer as well at this time (leaving only an @load of the new one, and a deprecation note, in the old place)?  (I could see that in a follow-up PR too, if there's agreement.)

For the `ubuntu24_zam` CI failure I think you only need to run a baseline update with the `zam` alternative.

scripts/policy/frameworks/analyzer/debug-logging.zeek
J-Gras: Ack that they shouldn't differ. But the idea is that they can be used for non-failure-related info as well. See @awelzel's comment:
> Renaming makes sense to me and leaves the option open to reuse data or reason for something like analyzer_attached. E.g. reason=dpd or reason=port and maybe data could be the signature or the matching packet.

I can see a the fields being used for disabling analyzers too.

awelzel: > See @awelzel's comment:

I'm actually very happy Christian pushed back. While it'd allow to pack more information into a single log line with as little fields as possible which I found pretty exciting for a debug log, it also means the reason field changes its meaning and interpretation depending on another field's value. And that actually is not so exciting, because it'd make documenting and explaining the fields painful / complex. (but also weird to work with for downstream consumers when querying).

So I'm now leaning towards "attached" or "enable" would come with a separate "enable_reason" and that could then be "dpd" or "port".

Thanks Christian!

PS: Has there ever been a document or presentation summarizing such discussions? Like "best practices for log fields". I come up with stuff every time anew and am wrong anew, too. A reference would help here (e.g. similar to https://prometheus.io/docs/practices/naming/)

0xxon: I like that. So - we are kind of tending towards just leaving `failure_reason` and `failure_data` as is, right?

@J-Gras - the reason that I did not also change `failure_reason` in the non-debug-log is that there it seems more appropriate, and conveys more or the meaning. There also are other logs that use the exact same field names.

0xxon:
I undid the field rename, and moved `policy/frameworks/dpd/detect-protocols.zeek` to analyzer, as requested.

scripts/policy/frameworks/analyzer/debug-logging.zeek
J-Gras: Just for the fun: Are you folks saying that we call the new log `analyzer.log` to keep it "future-proof for making adjustments down the road" instead of chosing a name that describes its purpose but stick to `failure_reason` in this log because it aligns with the log's purpose? 😜 

Seriously: Adding additional reason fields seems fine. In addition to `enable_reason`, `disable_reason` would be helpful to see whether an analyzer was automatically disabled because of a violation.
This also helps to justify having two different logs, because right now, I can still see having only a single `anlyzer.log` that is filtered/configured based on user needs.

ckreibich: > Has there ever been a document or presentation summarizing such discussions? Like "best practices for log fields". I come up with stuff every time anew and am wrong anew, too.

That's a great point. It's definitely something that comes up again and again. I don't think we have guidelines / lessons learned written down anywhere. There are presentations that come to mind, particularly Smoot's recent webinar and (iirc?) also an earlier ZeekWeek presentation.

I don't think we've ever done a a dedicated post or presentation about the historical evolution of an analyzer. That could be quite cool!

I think there's something particularly sticky about this PR because it concerns infrastructure logs that are a bit quirky, i.e. not something we do often. So perhaps that's why everybody comes at this with differing opinions.

ckreibich: @J-Gras hah 😜. Fwiw I don't have a grand worldview here. I like that `analyzer_debug.log` now has that explicit troubleshooting/tooling angle, and I'm for sure more open to tinkering with its fields than with those of analyzer.log. I just thought it looks strange for the two related logs to have differing field names here.

There's also a general question at which point this area stops being appropriate for logs and becomes more suitable for free-form debug streams. I definitely like Arne's suggestion of an analyzer history field — it's a middle ground for those two extremes, in a way.

PR #4363, opened 2025-04-15
Clear sessions when session manager is done

J-Gras:
When manually shutting down Zeek from a plugin in a step by step fashion, I got confused because the session manager keeps sessions in its map beyond the point where the session manager is considered done. I haven't found a reason to keep sessions after this point. Am I missing something?

If not, this change hopefully simplifies the shutdown control flow a tiny bit.

### DIFF
--- a/src/RunState.cc
+++ b/src/RunState.cc
@@ -385,7 +385,7 @@ void finish_run(int drain_events) {
         event_mgr.Drain();
 
         if ( session_mgr )
-            session_mgr->Done();
+            session_mgr->Clear();
     }
 
 #ifdef DEBUG
--- a/src/RunState.cc
+++ b/src/RunState.cc
@@ -106,7 +106,7 @@ RETSIGTYPE watchdog(int /* signo */) {
             }
 
             get_final_stats();
-            finish_run(0);
+            util::detail::set_processing_status("TERMINATING", "watchdog");
 
             reporter->FatalErrorWithCore("**watchdog timer expired, t = %d.%06d, start = %d.%06d, dispatched = %d",
                                          int_ct, frac_ct, int_pst, frac_pst, current_dispatched);
@@ -340,9 +340,7 @@ void run_loop() {
         }
     }
 
-    // Get the final statistics now, and not when finish_run() is
-    // called, since that might happen quite a bit in the future
-    // due to expiring pending timers, and we don't want to ding
+    // Get the final statistics now, as we don't want to ding
     // for any packets dropped beyond this point.
     get_final_stats();
 }
@@ -373,29 +371,6 @@ void get_final_stats() {
                        s.received, ps->Path().c_str(), s.dropped, dropped_pct, not_processed, unprocessed_pct,
                        filtered.c_str());
     }
-}
-
-void finish_run(int drain_events) {
-    util::detail::set_processing_status("TERMINATING", "finish_run");
-
-    if ( drain_events ) {
-        if ( session_mgr )
-            session_mgr->Drain();
-
-        event_mgr.Drain();
-
-        if ( session_mgr )
-            session_mgr->Clear();
-    }
-
-#ifdef DEBUG
-    extern int reassem_seen_bytes, reassem_copied_bytes;
-    // DEBUG_MSG("Reassembly (TCP and IP/Frag): %d bytes seen, %d bytes copied\n",
-    // 	reassem_seen_bytes, reassem_copied_bytes);
-
-    extern int num_packets_held, num_packets_cleaned;
-    // DEBUG_MSG("packets clean up: %d/%d\n", num_packets_cleaned, num_packets_held);
-#endif
 }
 
 void delete_run() {
--- a/src/RunState.h
+++ b/src/RunState.h
@@ -26,7 +26,6 @@ extern void init_run(const std::optional<std::string>& interfaces, const std::op
                      const std::optional<std::string>& pcap_output_file, bool do_watchdog);
 extern void run_loop();
 extern void get_final_stats();
-extern void finish_run(int drain_events);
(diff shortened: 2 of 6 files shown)


0xxon (changes requested):
Looking at the code, all of this seems a bit weird to me. Currently, `Manager::Done` essentially does nothing at all. According to the comment, it should drain events before destruction - which also sounds weird to me.

Your change changes things so that `Manager::Done` calls `Manager::Clear`, instead of doing nothing - and then you replace the calls to `Manager::Clear` to `Manager::Done`, and remove the Clear call in the destructor.

First - I would not remove the clear call in the destructor. Having it there seems prudent, even if there currently are no code paths that require it.

Second - instead of moving the `Clear` call into `Done` - could we instead just get rid of `Manager::Done`? It doesn't do anything, the documentation for it doesn't make any sensse, and the call gives no indication for what it does. For clear on the other hand, it is pretty obvious what it does.

0xxon:
I think session::Manager is not in a detail namespace. Hence this probably should be done via a deprecation, and removal in Zeek 8.1.

timwoj:
> I think session::Manager is not in a detail namespace. Hence this probably should be done via a deprecation, and removal in Zeek 8.1.

I would argue that this specific usage of the manager class (shutting down) isn't something external users should be doing themselves. If it was changes to some of the API methods external users would call then we should follow the deprecation path, but this is outside that.

J-Gras:
> I would argue that this specific usage of the manager class (shutting down) isn't something external users should be doing themselves. If it was changes to some of the API methods external users would call then we should follow the deprecation path, but this is outside that.

I am always doing things with Zeek I shouldn't do 😁 I guess there is no harm in going through a deprecation cycle here. Let me know, if I should remove the method straight away.

awelzel:
> I am always doing things with Zeek I shouldn't do 😁 I guess there is no harm in going through a deprecation cycle here. Let me know, if I should remove the method straight away.

Seems fine. Looking at `finish_run()` I'm always confused/annoyed why it exists. What would you think about ripping it out? Two call sites, almost doing nothing if one passes "false" and that "false" caller is followed by `FatalErrorWithCore()`. The `#ifdef DEBUG` lines reference variables that don't exist anymore, so I think this could just go.

```
diff --git a/src/RunState.cc b/src/RunState.cc
index 1074100fa6..7b8bb1898d 100644
--- a/src/RunState.cc
+++ b/src/RunState.cc
@@ -106,7 +106,7 @@ RETSIGTYPE watchdog(int /* signo */) {
             }
 
             get_final_stats();
-            finish_run(0);
+            util::detail::set_processing_status("TERMINATING", "watchdog_expired");
 
             reporter->FatalErrorWithCore("**watchdog timer expired, t = %d.%06d, start = %d.%06d, dispatched = %d",
                                          int_ct, frac_ct, int_pst, frac_pst, current_dispatched);
@@ -375,29 +375,6 @@ void get_final_stats() {
     }
 }
 
-void finish_run(int drain_events) {
-    util::detail::set_processing_status("TERMINATING", "finish_run");
-
-    if ( drain_events ) {
-        if ( session_mgr )
-            session_mgr->Drain();
-
-        event_mgr.Drain();
-
-        if ( session_mgr )
-            session_mgr->Done();
-    }
-
-#ifdef DEBUG
-    extern int reassem_seen_bytes, reassem_copied_bytes;
-    // DEBUG_MSG("Reassembly (TCP and IP/Frag): %d bytes seen, %d bytes copied\n",
-    //         reassem_seen_bytes, reassem_copied_bytes);
-
-    extern int num_packets_held, num_packets_cleaned;
-    // DEBUG_MSG("packets clean up: %d/%d\n", num_packets_cleaned, num_packets_held);
-#endif
-}
-
 void delete_run() {
     util::detail::set_processing_status("TERMINATING", "delete_run");
 
diff --git a/src/RunState.h b/src/RunState.h
index 1015aa888a..451b9cb63b 100644
--- a/src/RunState.h
+++ b/src/RunState.h
@@ -26,7 +26,6 @@ extern void init_run(const std::optional<std::string>& interfaces, const std::op
                      const std::optional<std::string>& pcap_output_file, bool do_watchdog);
 extern void run_loop();
 extern void get_final_stats();
-extern void finish_run(int drain_events);
 extern void delete_run(); // Reclaim all memory, etc.
 extern void update_network_time(double new_network_time);
 extern void dispatch_packet(zeek::Packet* pkt, zeek::iosource::PktSrc* pkt_src);
diff --git a/src/zeek-setup.cc b/src/zeek-setup.cc
index d08d21ef88..dead90bb1f 100644
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -322,7 +322,8 @@ static void done_with_network() {
     event_mgr.Drain();
     event_mgr.Drain();
 
-    run_state::detail::finish_run(1);
+    session_mgr->Drain();
+    event_mgr.Drain();
 
 #ifdef USE_PERFTOOLS_DEBUG
```
Not sure the `Clear()` is useful at that point - do you actually need that in your code?

J-Gras:
> Looking at finish_run() I'm always confused/annoyed why it exists. What would you think about ripping it out?

I am all for it! That being said, `done_with_network()` is what I actually want to call. As this doesn't seem to be accessible from plugins, I just copied it and use my version of it.

Re `Clear`: I am wondering whether the sessions are supposed to be around until the removal events are actually processed calling `event_mgr.Drain()`.

awelzel:
> Re Clear: I am wondering whether the sessions are supposed to be around until the removal events are actually processed calling event_mgr.Drain().

Yeah, that's a good point. That'd affect last minute `lookup_connection()` calls. Probably better to keep separate.

awelzel (approved):
LGTM, waiting for CI. Thanks.

PR #4366, opened 2025-04-16
Fix some problems with the cmake clang-tidy/iwyu setup

timwoj:
This PR fixes a few things about how we were adding/using clang-tidy, iwyu, and clang-format:

- IWYU and clang-tidy were only being added to the main zeek_objs target, which only covers the main source files and nothing in a plugin. Thie meant none of the analyzers, etc were being covered. This adds a new `zeek_target_add_linters` function that can be used to add the linter properties to a target. Note that it's not being added for Spicy targets at the moment because clang-tidy specifically is reporting findings in the generated code. I need to talk to @bbannier or @rsmmr before we enable those. It will need https://github.com/zeek/cmake/pull/129 to be merged too.
- Disable linting for files generated by Bison. This might miss a few findings in the raw C++ code in those files, but there are a lot of findings in the generated code that we can't fix.
- Updated the src/3rdparty module to disable clang-format. This pulls in https://github.com/zeek/zeek-3rdparty/pull/29, which should be merged at the same time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,20 @@ include(cmake/CheckCompilerArch.cmake)
 
 string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_LOWER)
 
+if (ENABLE_IWYU)
+    find_program(ZEEK_IWYU_PATH NAMES include-what-you-use iwyu)
+    if (NOT ZEEK_IWYU_PATH)
+        message(FATAL_ERROR "Could not find the program include-what-you-use")
+    endif ()
+endif ()
+
+if (ENABLE_CLANG_TIDY)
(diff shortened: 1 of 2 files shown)


bbannier:
> Note that it's not being added for Spicy targets at the moment because clang-tidy specifically is reporting findings in the generated code. I need to talk to `@bbannier` or `@rsmmr` before we enable those.

If with "Spicy targets" you mean plugins containing Spicy-generated code, I just opened https://github.com/zeek/cmake/pull/130 so we do not lint generated files. If you mean any other Spicy target (e.g., `hiltic` or `spicyc`), these IMO should be handled in the existing clang-tidy infra in https://github.com/zeek/spicy, not here.

awelzel (approved):

PR #4371, opened 2025-04-18
Update c-ares to v1.34.5 and vcpkg to a version that includes it

timwoj:
C-Ares has been reporting some warnings in our builds for a bit. This PR upgrades to the latest release, and should hopefully silence those warnings. The latest release also fixes a CVE that was introduced by the last version we updated to.

This PR also updates the vcpkg submodule to a version that includes the same version of c-ares.


awelzel (approved):
Is this also for 7.0.x?

timwoj:
> Is this also for 7.0.x?

It could be. The last upgrade we did was during the 7.0 dev cycle so it should apply cleanly.

timwoj:
> It could be. The last upgrade we did was during the 7.0 dev cycle so it should apply cleanly.

Scratch that I did the last upgrade during the 7.1 cycle.

PR #4372, opened 2025-04-18
Avoid pointless Broker re-peering attempts after backpressure overflows by leveraging peering directionality

### LINKED ISSUE #4318: Broker re-peering attempts in opposite direction of severed peering will always fail and clutter logs
This is a follow-up to #3981: when node A first `Broker::peer()`s with node B, and node B subsequently notices backpressure in its message transmissions to A, it will sever the peering with A — that's good. Not good is that it then also tries to re-peer with node A:

* Zeek uses the current connection endpoint info to re-establish the connection. This means that when re-peering in the opposite direction of TCP connection establishment, the peering attempt causes a TCP connection to the originator's ephemeral port, which doesn't listen and will always fail. The reattempts continue indefinitely and clutter up broker.log. 
* It's not necessary to re-peer in this opposite direction: when node A realizes that it's losing peer B, it will automatically re-establish the peering (since it originally created it).

We've so far used the existing connection information for the re-peering since we tried to separate Broker- and cluster-level logic. We could soften this separation to allow a `Cluster::node_id_to_node()`, looking up the other side's listening port (or find the absence of one). Or we could add a bit of logic to `Broker::peer()` (which currently just calls into the corresponding BiF) to track who's peering with whom. This could be a table mapping from `Broker::NetworkInfo` to Broker ID, say: the ID will initially be unknown (an empty string), but a subsequent `Broker::peer_added` event will provide the needed info and could update the entry. The re-peering code could then check whether the intended target is among previously peered endpoints, and only proceed if so.

We've in the past had other situations where we wondered who peered actively, and conceptually the presence of a listening port doesn't imply that any node should naturally do so. So I'm in favor of the latter option.

The `backend.broker.backpressure-overflow-slow-worker` test in the cluster testsuite demonstrates the problem.

ckreibich:
There are several ways to add this, and it looks like Broker (the library) is close to having a built-in mechanism for this, via `broker::peer_info` and `broker::peer_flags`. However, the flags aren't actually used — I only see `broker::peer_flags::remote` set, at all tiems, in the [core actor](https://github.com/zeek/broker/blob/master/libbroker/broker/internal/core_actor.cc#L325). (It's also not clear to me what the flags would mean, but that's a detail.)

We can potentially still implement these, but for now this provides one such flag of information, namely whether a peering got established locally (i.e., whether the local node is the TCP client). It then uses this to avoid unnecessary re-peerings after backpressure-induced disconnects.

Resolves #4318.


timwoj (approved):
LGTM. Just one nit about a copy.

PR #4373, opened 2025-04-21
Handle lower case HTTP keyword issue #4218

### LINKED ISSUE #4218: Review handling of lower case in HTTP
The HTTP analyzer stops processing if a server answers with a lower case reply line like the following:
```
http/1.1 200 OK
Content-type: text/plain

hello world
```
You will find a `not a http reply line` message in the corresponding logs, as parsing is case sensitive:
https://github.com/zeek/zeek/blob/da673d65772eb3687e2290453804a54fc7168581/src/analyzer/protocol/http/HTTP.cc#L1456
However, a quick test shows that common browsers don't care about casing. I guess it would make sense to review handling of lower case for request and response messages as well as for the DPD signatures.

kshitiz56:
### Summary

This pull request fixes  #4218, where Zeek's HTTP analyzer failed to correctly detect and parse HTTP messages that used lowercase or mixed-case protocol identifiers like `http/1.1`. According to [RFC 7230, Section 2.6](https://www.rfc-editor.org/rfc/rfc7230#section-2.6), the protocol name "HTTP" is case-insensitive, and the analyzer should treat variants such as `http/1.1`, `Http/1.0`, etc. as valid.

#### HTTP Signature (DPD)
- Updated the signature regex in `dpd.sig` to allow matching lowercase and mixed-case HTTP identifiers:
  ```diff
  -  payload /^HTTP\/[0-9]/
  +  payload /^[hH][tT][tT][pP]\/[0-9]/
  ```

#### Analyzer Logic
- Introduced a `bool ignore_case` parameter to `PrefixMatch` and `PrefixWordMatch` to allow case-insensitive prefix comparison.
- Updated parsing logic in `ParseRequest`, `HTTP_ReplyLine`, and related methods to pass `ignore_case = true` for HTTP version line matching.

#### BTest Coverage
- Added two new test PCAPs with lowercase HTTP protocol identifiers:
  - `http_small_case.pcap`
  - `http_small_case_nonstandard_port.pcap`
- Created a new test script `http-small-case.zeek` to:
  - Run the analyzer on these PCAPs.
  - Assert that no `not a http reply line` or similar parsing errors appear in the logs.


src/analyzer/protocol/http/HTTP.cc
```diff
@@ -1199,15 +1202,17 @@ bool HTTP_Analyzer::ParseRequest(const char* line, const char* end_of_line) {
             break;
     }
 
-    if ( end_of_uri >= end_of_line && PrefixMatch(line, end_of_line, "HTTP/") ) {
+    // HTTP name is case-insensitive https://www.rfc-editor.org/rfc/rfc7230#section-2.6
```
timwoj, line 1205: The RFC specifies that the name is case-sensitive: 

>  The version of an HTTP message is indicated by an HTTP-version field
>   in the first line of the message.  HTTP-version is case-sensitive.
>
>     HTTP-version  = HTTP-name "/" DIGIT "." DIGIT
>     HTTP-name     = %x48.54.54.50 ; "HTTP", case-sensitive

I don't disagree with supporting the case-insensitive version, but we should report a weird in the case of a non-uppercase version.

src/analyzer/protocol/http/HTTP.cc
```diff
@@ -1097,8 +1097,10 @@ void HTTP_Analyzer::GenStats() {
     }
 }
 
-const char* HTTP_Analyzer::PrefixMatch(const char* line, const char* end_of_line, const char* prefix) {
-    while ( *prefix && line < end_of_line && *prefix == *line ) {
+const char* HTTP_Analyzer::PrefixMatch(const char* line, const char* end_of_line, const char* prefix,
+                                       bool ignore_case) {
+    while ( *prefix && line < end_of_line &&
+            (*prefix == *line || (ignore_case && tolower((unsigned char)*prefix) == tolower((unsigned char)*line))) ) {
```
timwoj, line 1103: I think these two should be switched around, as it would let the `ignore_case` version fail on the boolean check before it makes a slower string comparison that may fail.

timwoj (changes requested):

src/analyzer/protocol/http/HTTP.cc
kshitiz56: Resolved in 9dbb5ed35b1d0380c7f84b28490d4cdc0a65f3ab

src/analyzer/protocol/http/HTTP.cc
kshitiz56: Resolved in 92d3ca016b07efdca83257deb1e2ea7beee0c17e

timwoj (changes requested):
It's getting there!

timwoj:
I opened https://github.com/zeek/zeek/pull/4397 to supersede this PR so we can get it wrapped up before 7.2. Thanks for the contribution!

PR #4378, opened 2025-04-23
COPYING-3rdparty: Add some missing submodules

awelzel:
I'm not 100% clear what COPYING-3rdparty is used for, but it missed a few submodules we added recently.

Also, while there I wondered: Anyone have strong opinions for IXWebSocket and cppzmq submodule not being in auxil and closer/local to the code that's actually using them? I like that a bit more particularly if their use is isolated to that directory/tree.


timwoj:
COPYING-3rdparty is there so we provide copies of the licenses for any 3rd party packages that we use.

timwoj (approved):

PR #4379, opened 2025-04-23
CI: No more Ubuntu 20.04 [skip ci]

awelzel:
Ubuntu 20.04's default Python doesn't deal well with the type annotations used in btest/Files/wstest.py. Given that it's about to be EOL, just remove it. With the upcoming 7.2 release, this seems to be also reasonable.


.cirrus.yml
```diff
@@ -357,14 +357,6 @@ ubuntu22_spicy_head_task:
   skip: $CIRRUS_BRANCH =~ 'release/.*'
   << : *BENCHMARK_ONLY_IF_TEMPLATE
 
-ubuntu20_task:
-  container:
-    # Ubuntu 20.04 EOL: April 2025
-    dockerfile: ci/ubuntu-20.04/Dockerfile
```
bbannier, line 363: Do you want to get rid of the now untested Docker setup as well?

bbannier (approved):

PR #4381, opened 2025-04-23
A couple of performance fixes from magic-trace findings

timwoj:
These came out of @JustinAzoff's tinkering with magic-trace (see https://github.com/zeek/zeek/discussions/4345).

- `ToUpper` shouldn't check whether the character is lowercase before converting it to uppercase. `std::toupper` effectively does this check already internally.
- The field offsets in the connection record were being looked up every time the connection val was updated for UDP and ICMP, even though the offsets shouldn't change. We were already statically looking these up for TCP connections.


src/packet_analysis/protocol/icmp/ICMPSessionAdapter.h
```diff
@@ -13,7 +13,7 @@ class ICMPSessionAdapter final : public IP::SessionAdapter {
 
     void AddExtraAnalyzers(Connection* conn) override;
     void UpdateConnVal(RecordVal* conn_val) override;
-    void UpdateEndpointVal(const ValPtr& endp, bool is_orig);
+    void UpdateEndpointVal(RecordVal* endp, bool is_orig);
```
ckreibich, line 16: I'm wondering here whether we should make it private while we're at it — it is in the `UDPSessionAdapter`.

ckreibich (approved):

src/packet_analysis/protocol/icmp/ICMPSessionAdapter.h
timwoj: Sounds good to me. There is a bit of inconsistency between all of these Adapter classes that could be smoothed out.

PR #4382, opened 2025-04-23
cluster/websocket: Allow clients to send no subscriptions

awelzel:
Testing with the [go-zeek-broker-ws library](https://github.com/corelight/go-zeek-broker-ws), it has tests where clients do not request any topic subscription. That would result in the client not receiving back an ack.

This PR fixes this and adds a regression test, too.


testing/btest/cluster/websocket/no-subscriptions.zeek
```diff
@@ -0,0 +1,68 @@
...
+# @TEST-EXEC: btest-diff ./manager/.stderr
+# @TEST-EXEC: btest-diff ./client/.stdout
+# @TEST-EXEC: btest-diff ./client/.stderr
+
+# @TEST-START-FILE manager.zeek
+@load ./zeromq-test-bootstrap
+redef exit_only_after_terminate = T;
+
+global expected_ping_count = 100;
+global ping_count = 0;
```
awelzel, line 34: Unused. Will drop during merge.

src/cluster/websocket/WebSocket.cc
```diff
@@ -430,6 +423,26 @@ void WebSocketEventDispatcher::HandleSubscriptions(WebSocketClientEntry& entry,
     }
 }
 
+void WebSocketEventDispatcher::HandleSubscriptionsActive(WebSocketClientEntry& entry) {
```
bbannier, line 426: Nit: suggest to pass a `const WebSocketClientEntry&` since this does not mutate the value.

src/cluster/websocket/WebSocket.h
```diff
@@ -228,6 +228,8 @@ class WebSocketEventDispatcher {
 
 
     void HandleSubscriptions(WebSocketClientEntry& entry, std::string_view buf);
+    // All subscriptions are now active, raise the websocket_client_added event and send the ack.
+    void HandleSubscriptionsActive(WebSocketClientEntry& entry);
```
bbannier, line 232: "Now" doesn't work well in a docstring since it could be run at any time.

Maybe something like

```suggestion

    // For all currently active subscriptions of the entry, raise the internally `websocket_client_added` event, and send the ack to the client.
    void HandleSubscriptionsActive(WebSocketClientEntry& entry);

```

Probably nit, but since we add a docstring I would also surround this by empty lines.

testing/btest/cluster/websocket/no-subscriptions.zeek
```diff
@@ -0,0 +1,68 @@
...
+
+# @TEST-START-FILE client.py
+import wstest
+
+def run(ws_url):
+    with wstest.connect("ws1", ws_url) as tc:
+        tc.send_json([])  # Send no subscriptions
+        ack = tc.recv_json()
+        assert "type" in ack, repr(ack)
+        assert ack["type"] == "ack"
```
bbannier, line 63: Nit: this is just 

```suggestion
        assert ack.get("type") == "ack", f"{ack}"
```

bbannier (approved):

PR #4384, opened 2025-04-23
Update Mozilla CA list and CT list to NSS 3.110

timwoj:


timwoj:
This is causing `tests.ctu-sme-11-vm-win7ad` in zeek-testing to fail. The `valid_ct_logs` and `valid_ct_operators` fields are getting set to zero for a bunch of connections in ssl.log.

0xxon (approved):

PR #4385, opened 2025-04-23
Change from_json to return an error rather than print it.

stevesmoot:
IMHO printing errors from zeek is an antipattern, and from_json already had half an exception structure.  Change enables
scripts to look at the errors to see if they can handle better than just exposing a print to the operator.


bbannier:
Reopened as #4386.

PR #4386, opened 2025-04-23
Change from_json to return an error rather than print it.

stevesmoot:
IMHO printing errors from zeek is an antipattern, and from_json already had half an exception structure.  Change enables
scripts to look at the errors to see if they can handle better than just exposing a print to the operator.


stevesmoot:
sigh, still dont seem to understand git.

src/zeek.bif
```diff
@@ -5240,7 +5240,7 @@ function to_json%(val: any, only_loggable: bool &default=F, field_escape_pattern
 ##           when keys are not valid field identifiers, or represent reserved
 ##           keywords like **port** or **type**.
 ##
-## returns: A value of type t.
+## returns: A record either contining a value of type t as field v with valid=T or valid=F and err contains the error.
```
timwoj, line 5243: I would change this to something less explicit, since people can look at the type to see the fields.

```suggestion
## returns: A record with the result of the conversion, containing either a value or an error message.
```

timwoj (changes requested):
I like it 👍🏼

timwoj:
Looking at this again, I'm a little concerned about the loss of information in some of the error messages. For example, the error message for `testing/btest/Baseline/bifs.from_json-8` used to include the pattern that failed to parse, but the error string returned in the result record doesn't have it. There's a couple of others that previously included the JSON that failed to parse but the errors now don't have that information.

stevesmoot:
Thats odd, it was literally moving the string from the print to the return.

timwoj:
> Thats odd, it was literally moving the string from the print to the return. (or at least that was the intent)

The rest of the error message probably comes from `emit_builtin_error`, which will print out the context of the error (the variables involved, the location, etc).

vpax:
> The rest of the error message probably comes from `emit_builtin_error`, which will print out the context of the error (the variables involved, the location, etc).

Is there a separate BiF for just returning that context?

timwoj:
> Is there a separate BiF for just returning that context?

Not that I'm aware of. I traced down through `emit_builtin_error` and all of that information is actually built up by `CallExpr:Error` (via `Obj::Error`) and subsequent calls into `Reporter` as its moving through building the error message. We could factor all of that into a method that gives you back the error string instead of logging it, but it doesn't exist right now.

There's also the matter of by the time you get the error message back from `from_json`, you've lost the original location (and arguments, sort of?) that caused the failure. That would have been known to `CallExpr` but if you've returned, that `CallExpr` is lost.

stevesmoot:
Yeah, I dug through it and came to the same conclusion.

I actually am sort of the opposite opinion - at least for handling the returns from from_json the error messages are too detailed.  The line/file info isnt useful, b/c your code is there already.  The type errors are many (in the btests) but really to me the failures should be grouped into some form of "this isnt JSON" and "JSON doesnt match the record".  Of the messages, the most useful ones are the "JSON parse error" like in from_json-3 because it points to exactly where in the string it diverges (and these are retained in the PR).  I could see someone using that.  The "cant compile pattern" and array messages are just super confusing - really they should be JSON parse errors at 0 IMHO.

Now, if you want to punt, my orig goal was avoiding the printing of errors.  We could do that with an optional argument to the BIF instead.  My code using this has a good idea how to bang on its invalid json to fix it, so wasnt actually using the error message return.  (That said, it could be improved with using those locations for the parse error, if the real world is even rougher than the code can currently handle.)

src/zeek.bif
```diff
@@ -5276,7 +5276,7 @@ function from_json%(s: string, t: any, key_func: string_mapper &default=from_jso
 	else
 		{
 		rval->Assign(valid_idx, false);
-		zeek::emit_builtin_error(res.error().c_str());
+                rval->Assign(result_type->FieldOffset("err"), res.error().c_str());
```
ckreibich, line 5279: I just noticed that `from_json()` is unusual in its use of `FieldOffset()`, which loops over the fields until it finds a matching name. Our other BiFs mostly use `n++` style indexing directly with `Assign()`, and since I can pretty easily see `from_record()` in hot codepaths I suggest moving the whole function to that approach (i.e. also above in lines 5249+). We can do that in a follow-up PR.

ckreibich (approved):
Thumbs up from me. Regarding the location context, one simple way of getting that back is by punting to the caller, who needs to check for the error anyway and has that context.

Issue #4302, opened 2025-03-20
`DBG_LOG` is not thread-safe


timwoj:
This came up in https://github.com/zeek/zeek/pull/4297, where ThreadSanitizer is reporting a data-race involving DBG_LOG and the thread that runs storage expiration. We should look into whether DBG_LOG could be made thread-safe instead of avoiding using it.

awelzel:
> We should look into whether DBG_LOG could be made thread-safe instead of avoiding using it.

In the ZeroMQ backend, I had avoided DBG_LOG from a thread and ended up just using `fprintf()` for debugging from a thread.

https://github.com/zeek/zeek/blob/1c598df53b3198dd9ec00fb7069c44b0bad82c54/src/cluster/backend/zeromq/ZeroMQ.cc#L559C17-L559C43

For BasicThread / MsgThread, there are separate helpers to log messages via the event loop, but for pure debugging (also from non-BasicThread threads or non-Zeek started threads) it would be nice to just DBG_LOG without needing to worry (the extra overhead in a debug build hopefully isn't a major topic).

timwoj:
Most of the TSan warnings I got were actually the same data-race with `zeek::run_state::detail::update_network_time()` and access to the global `run_state::network_time` variable. `DBG_LOG` uses it to write out the timestamp with each message. `MsgThread` gets around this by queuing reporter messages for the main thread when it can, instead of writing the directly. We could make `network_time` a `std::atomic<double>`, but there's a significant performance hit for doing so.

PR #4387, opened 2025-04-24
Add telemetry for Broker's per-peer send buffers

### LINKED ISSUE #4319: Establish visibility into Broker peer buffer fill levels
Another follow-up to #3981: the size of the send buffers (`Broker::peer_buffer_size` and `Broker::web_socket_buffer_size`) is a tunable. We can provide reasonable defaults, but sites should be able to gauge the fit. Currently you either see no backpressure issues, or you do — you have to tune sizing from backpressure overflow frequency.

We'd like an API into the buffering system to retrieve the current fill level of a given peering's send buffer, for example to expose them as a telemetry gauges. This should happen at the cluster level, not Broker's, so endpoint naming is fixed (think "broker_peerbuffer_entries_worker23_proxy02", not "broker_peerbuffer_entries_864f81a0-0c93-574c-954c-024e19e739bb_6dca6236-e666-52bc-a154-d5e400469359").

#3998 is a building-block for this.

ckreibich:
This adds three metrics for each Broker peering's send buffer: the current buffer fill level, the largest fill level seen recently, and a count of the number of buffer overflows. See the commits  for details, but the essence is that I'm tracking low-level counts in the Broker manager by extending the `event_observer`, and use a script-level `Telemetry:sync()` hook to tie this to the cluster framework's node-naming.

I've tried to look out for performance, particularly in the per-message observer callbacks. The sync hook could be optimized further, for example we could store the resulting Cluster node name in a redef'd version of the `BrokerPeeringStats` record. It felt a bit like premature optimization so I steered clear for now.

Also includes a fix for an unrelated race condition in the `cluster.broker.publish-any` test that triggered on my machine.

Once this makes it, I'm going to follow up with a tiny PR that switches the default buffer overflow policy to `drop_oldest`. I've tested this quite a bit and its behavior is much tamer — since no peerings ever get disconnected there's nothing to reconnect, and the buffer overflow count becomes a simple indicator of the number of messages lost. It also means that if there's marginal overflow, it's smaller loss than the entire send buffer plus data in flight, of a disconnected peering. 

Resolves #4319.


src/broker/Manager.cc
```diff
@@ -92,6 +92,142 @@ void print_escaped(std::string& buf, std::string_view str) {
...
+            stats_table->Assign(stats.peer_id, zeek::make_intrusive<zeek::RecordVal>(stats_record_type));
+            it = stats_map.find(peer);
+        }
+
+        auto& stats = it->second;
+        auto now = zeek::util::current_time();
+
+        if ( now - stats.last_interval > stats_interval ) {
+            stats.last_interval = now;
+            stats.max_queued_recently = 0;
```
awelzel, line 160:
```suggestion
            stats.max_queued_recently = stats.queued;
```

Not sure that matters all that much, but given we may not update `max_queued_recently` below if we're in a overflow situation, might make sense.

src/broker/Manager.cc
```diff
@@ -92,6 +92,142 @@ void print_escaped(std::string& buf, std::string_view str) {
...
+            assert(stats_v != nullptr);
+
+            // We may get here more than stats_interval after the last
+            // Observe(), in which case the max_queued_recently value is now
+            // stale. Reset if so.
+            auto now = zeek::util::current_time();
+
+            if ( now - stats.last_interval > stats_interval ) {
+                stats.last_interval = now;
+                stats.max_queued_recently = 0;
```
awelzel, line 194: I guess it matters more here. It'd be weird to observe a record with `queued > 0` but `max_queued_recently == 0,`, no?

src/broker/Manager.cc
```diff
@@ -92,6 +92,142 @@ void print_escaped(std::string& buf, std::string_view str) {
...
+        stats_table =
+            zeek::make_intrusive<zeek::TableVal>(zeek::id::find_type<zeek::TableType>("BrokerPeeringStatsTable"));
+        stats_record_type = zeek::id::find_type<zeek::RecordType>("BrokerPeeringStats");
+    }
+
+    // Update the peering's stats. This runs in Broker's execution context.
+    // Broker does not expose send-buffer/queue state explicitly, so track
+    // arrivals (a push, push_or_pull == true) and departures (pull,
+    // push_or_pull == false) as they happen.
+    void Observe(const broker::endpoint_id& peer, bool push_or_pull) {
```
awelzel, line 144: I find this `push_or_pull` bool super cryptic to read, but I guess you got used to me. To me it first read as "either it is a push or pull", but the usage is obviously much different. Maybe `is_push` like `is_orig` and comment that it's otherwise a pull?

awelzel (commented):
I think I'd be happy to approve, but leaving just a "comment" for now. Thanks for fixing the flaky cluster test!

src/broker/Manager.cc
ckreibich: Agree. Will tweak.

src/broker/Manager.cc
ckreibich: > I guess you got used to me
LOL — I did hesitate about that naming too, but I've seen similar patterns before. I'll make a pass, easy enough!

src/broker/Manager.cc
```diff
@@ -92,6 +92,142 @@ void print_escaped(std::string& buf, std::string_view str) {
...
+        std::lock_guard<std::mutex> lock(mutex);
```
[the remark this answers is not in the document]
ckreibich: I wondered too and looked at that a bit. I _think_ we're okay — there's no linearity in CAF threads to the number of peers. My understanding is that interaction is confined to a single one, the CAF scheduler. This changes when you increase `Broker::num_threads` from its default of one, but I don't think that's common, and even if one does, it's not clear it's immediately problematic.

I also wondered whether to make the whole thing optional and decided to be optimistic for now. It would be quite easy to iterate on this in a bugfix release.

timwoj:
Looks like something about shutdown is causing a TSan failure: https://api.cirrus-ci.com/v1/task/4854649158631424/logs/test.log

I thought this was part of the follow-on PR, but it reproduces reliably on my machine with this branch. There's actually two data races there. I'm not sure what's happening in the first, but the second one is because you can't use `TableVal` objects in a non-main thread. This is because `TableEntryVal` tries to access `run_state::network_time` in the constructor, and that variable is [not thread-safe](https://github.com/zeek/zeek/issues/4302).

ckreibich:
Interesting. I'll take a look.

timwoj:
It's curious that `TableEntryVal` even uses that value. We don't care about it unless the table has the `&expire_create` attribute on it, so it doesn't always need to be set. I could see adding an argument to the constructor, and possibly a field to the object itself, that avoids that for tables that don't need it.

ckreibich:
Okay, I was able to essentially keep the same approach, but shift all of the Zeek-side work into `PeerBufferState::GetPeeringStatsTable()`. Using `util::current_time()` was the other problem, just as you predicted Tim. I switched this to using Broker's notion of time (since it runs in a Broker event observer callback), which has its own problems since I think there's a chicken-and-egg problem here between registering the observer and availability of the local endpoint instance to the observer. I simply provide it once available, and take care when accessing in case the observer callbacks can trigger before then.

I've also rebased `topic/christian/broker-tuning` for #4396.

awelzel (approved):
I've tested this branch (actually the one from #4396) with [zeek-cluster-perf](https://github.com/awelzel/zeek-cluster-perf) with 56 workers. It's maybe 1-2% slower. these tests set `peer_buffer_size` to `512 * 1024` in order to avoid any drops/disconnects and are likely triggering this overflow code a lot.

Here's some entries from the `telemetry.log` file collected during `broadcast-highrate`:

```
worker-51       counter zeek_broker_peer_buffer_overflows_total endpoint,peer   worker-51,proxy-3       0.0
worker-51       gauge   zeek_broker_peer_buffer_levels  endpoint,peer   worker-51,proxy-4       11765.0
worker-51       gauge   zeek_broker_peer_buffer_levels  endpoint,peer   worker-51,proxy-1       11021.0
worker-54       gauge   zeek_broker_peer_buffer_recent_max_levels       endpoint,peer   worker-54,proxy-3       64573.0
worker-54       gauge   zeek_broker_peer_buffer_recent_max_levels       endpoint,peer   worker-54,proxy-4       64661.0
worker-56       counter zeek_broker_peer_buffer_overflows_total endpoint,peer   worker-56,proxy-1       0.0
worker-56       gauge   zeek_broker_peer_buffer_levels  endpoint,peer   worker-56,proxy-2       10612.0
worker-55       counter zeek_broker_peer_buffer_overflows_total endpoint,peer   worker-55,proxy-1       0.0
worker-55       gauge   zeek_broker_peer_buffer_levels  endpoint,peer   worker-55,proxy-2       10668.0
```

I've left one comment about naming and the prefix usage that'd be good to adapt, but :+1: from me otherwise.

scripts/base/frameworks/cluster/broker-telemetry.zeek
```diff
@@ -0,0 +1,69 @@
...
+@load base/frameworks/telemetry
+
+module Cluster;
+
+## This gauge tracks the current number of locally queued messages in each
+## Broker peering's send buffer. The "peer" label identifies the remote side of
+## the peering, containing a Zeek cluster node name.
+global broker_peer_buffer_queued_gf = Telemetry::register_gauge_family([
+    $prefix="zeek",
+    $name="broker-peer-buffer-levels",
```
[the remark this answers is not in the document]
ckreibich: Well. :-) These buffers are measured in number of messages, so I'd argue a message is a unit here. If they were in bytes you'd agree that bytes are a unit? Levels came from "fill level", and I wasn't super happy with them either. Happy to drop that.

I thought about the prefix too, but I like the distinction of `broker_*` coming directly out of Broker (the library) vs ones generated by Zeek. There's a btest that verifies whether Broker-internal metrics make it into Zeek's prometheus registry, and it works by prefix, so we'd be adding these there too.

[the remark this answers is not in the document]
awelzel: > so I'd argue a message is a unit here. 

Sorry - I was distracted by this sentence in the [naming docs](https://prometheus.io/docs/practices/naming/) - as I wasn't explicit, I like messages much more than levels.

>  For example, connections or notifications are not considered units for this rule and do not have to be at the end of the metric name.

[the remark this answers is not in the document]
ckreibich: It's all good! And thanks a bunch for doing performance tests, much appreciated. It's been running on `zeek-test` since yesterday without problems, btw.

PR #4390, opened 2025-04-24
cluster/websocket: Automatic WebSocket client topic subscription

awelzel:
Subscribe every WebSocket client to a unique topic, by default under

    zeek/cluster/websocket/client/<identifier>/

Add tests that verify that WebSocket clients receive messages on these topics even if they didn't explicitly pass them in their handshake message.

This is somewhere between feature and bug fix. It aids the ZeroMQ backend implementation: A WebSocket client that doesn't provide any subscriptions and immediately starts publishing would discard events until receiving other nodes subscriptions from the central XPUB/XSUB proxy. ZeroMQ does sender side topic filtering. When using subscriptions, the client waits until its own subscriptions are returned from the central XPUB/XSUB proxy, thereby also learning about other node's subscriptions.

Also, make the no-subscriptions.zeek test use 32 clients sequentially to trigger potential issues more quickly.


awelzel:
Discarding this in favor of a different approach.

PR #4392, opened 2025-04-24
Bump Spicy

evantypanski:


bbannier:
Thanks! It would be great to also get https://github.com/zeek/spicy/pull/2038 into rc1, but I wouldn't want to merge that one without some review. Worst case we land that in rc2 or 7.2.1.

bbannier (approved):

PR #4393, opened 2025-04-24
cluster/websocket: Add ReadyToPublishCallback API, implement for ZeroMQ

awelzel:
Provide a mechanism to allow a cluster backend report when it is ready
for publish operations. This is primarily useful for ZeroMQ which has
sender-side filtering and is only really ready for publishing when it
has learned about subscriptions from other nodes.

The ZeroMQ heuristic for "ready to publish" is to create an unique and
ephemeral subscription using the XSUB socket and observe it arrive on the
XPUB socket. At this point, visibility into other node's subscriptions
is provided.


testing/btest/cluster/websocket/broker/no-subscriptions.zeek
```diff
@@ -0,0 +1,66 @@
...
+global ping_count = 0;
+const ping_count_expected = 32;
+
+event ping(c: count) &is_used
+	{
+        ++ping_count;
+	print "got ping", c, ping_count;
+        if ( ping_count == ping_count_expected )
+	    terminate();
+	}
```
bbannier, line 50: Nit: funky space/tab mix.

```suggestion
event ping(c: count) &is_used
	{
	++ping_count;
	print "got ping", c, ping_count;
	if ( ping_count == ping_count_expected )
		terminate();
	}
```

src/cluster/Backend.h
```diff
@@ -269,6 +269,31 @@ class Backend {
...
+     *
+     * This mechanism allows backends to notify the caller that it has now
+     * determined readiness for publish operations.
+     *
+     * Callers should be prepared that \a cb is invoked immediately as that
+     * is the default implementation for DoReadyToPublishCallback().
+     *
+     * @param cb The callback to invoke when the backend is ready for publish operations.
+     */
+    void ReadyToPublishCallback(ReadyCallback cb) { DoReadyToPublishCallback(std::move(cb)); }
```
bbannier, line 295: Nit: this could pass `cb` by const ref (and downstream changes in signature of `DoReadyToPublishCallback` and captures of `scb` in `ZeroMQBackend::DoReadyToPublishCallback`).

src/cluster/websocket/WebSocket.cc
```diff
@@ -372,6 +372,30 @@ void WebSocketEventDispatcher::Process(const WebSocketSubscribeFinished& fin) {
         return;
     }
 
+    if ( ! entry.ready_to_publish ) {
```
bbannier, line 375: Just FTR, nothing to do here.

Like I wrote over Slack, it is slightly weird that websocket layer keeps track of whether a backend is ready to publish while the backend would accept publishes without error even if it is not (currently backends only know about clients when they subscribe to _something_, and while Zeek nodes typically subscribe, we don't expect all websocket clients would). If we would clean this up in the future we would probably add some way for a client to announce itself to the backend which could then keep track of its readiness.

bbannier (approved):

PR #4396, opened 2025-04-25
Broker tuning: reduce retry intervals, bump send-buffer sizes, switch default backpressure overflow policy to `drop_oldest`.

ckreibich:
We couldn't find any reason why those retry delays were so high, and some tests already lowered them to speed things up. See commits for details on the backpressure-related changes. I noticed that `congestion_queue_size` doesn't do anything during #4387, and @Neverlord also thinks this is just a leftover we never cleaned up.

This currently stacks on top of `topic/christian/broker-backpressure-metrics`, until #4387 is merged.


timwoj (approved):

timwoj:
I held off on merging this one until CI passed, but then TSan reported a data race: https://api.cirrus-ci.com/v1/task/4854649158631424/logs/test.log

PR #4401, opened 2025-04-28
Add new plugin hook for publishing of cluster events

awelzel:
This PR introduces a new plugin hook that can be used to intercept publish operations from the Zeek scripting layer (and technically also other publish operations, but they are unlikely, currently).

This is pre-work for #4177, but seems reasonable to open a stand-alone PR as well that can be discussed independently. The broader context and motivation was discussed in #4334. Plugins can use this new hook to modify the `Event` instance passed into the hook in order to attach additional metadata in the future.

I can also see this hook being interesting for monitoring purposes and playing tricks :-)


src/plugin/Manager.h
```diff
@@ -427,6 +436,22 @@ class Manager {
      */
     void HookUnprocessedPacket(const Packet* packet) const;
 
+    /**
+     * Hook for intercepting event publish operations.
```
rsmmr, line 440: Suggest to add some additional context here that this refers to cluster  communication,

src/plugin/Plugin.h
```diff
@@ -1077,6 +1122,22 @@ class Plugin {
      */
     virtual void HookUnprocessedPacket(const Packet* packet);
 
+    /**
+     * Hook for intercepting event publish operations.
```
rsmmr, line 1126: Per above, a bit more context.

src/plugin/Plugin.h
```diff
@@ -1077,6 +1122,22 @@ class Plugin {
...
+     *
+     * This hook can be used for metrics collection, metadata
+     * modifications and even redirecting events.
+     *
+     * @param backend The backend used to publish this event
```
rsmmr, line 1131: `backend used to` -> `backend publishing`?

src/broker/Manager.cc
```diff
@@ -825,6 +827,10 @@ std::vector<broker::peer_info> Manager::Peers() const {
 std::string Manager::NodeID() const { return to_string(bstate->endpoint.node_id()); }
 
 bool Manager::DoPublishEvent(const std::string& topic, cluster::detail::Event& event) {
+    bool do_publish = PLUGIN_HOOK_WITH_RESULT(HOOK_PUBLISH_EVENT, HookPublishEvent(*this, topic, event), true);
+    if ( ! do_publish )
+        return true;
+
```
rsmmr, line 833: I don't know much about the new structure: do we need this here *and*  in `Backend::DoPublishEvent`?.

PR #4407, opened 2025-04-29
Fix for incorrect ZAM optimization of expressions seen in single-statement inlined functions

vpax:
This fixes a ZAM optimizer bug where it would incorrectly use temporaries computed in single-statement inlined functions.


timwoj (approved):

timwoj:
Cherry-picked into release/7.2

PR #4410, opened 2025-04-30
A round of QUIC analyzer fixes and improvements

### LINKED ISSUE #4405: QUIC: Failure analyzing ClientHello / ServerHello fragmented over multiple initial packets
The following pcap contains a QUIC connection where the ClientHello is fragmented into CRYPTO frames over the first 3 Initial Packets. However, Zeek's QUIC analyzer logic only defragments/reassembles within a single Initial Packet.

[quic-fragmented-crypto.pcap.txt](https://github.com/user-attachments/files/19955905/quic-fragmented-crypto.pcap.txt)

This will need some restructuring of state handling for "CryptoBuffer" and "sink" (maybe attach to context) to properly forward the data to the TLS analyzer.

Image

Another pcap: [quic-fragmented-crypto-56802.pcap.txt](https://github.com/user-attachments/files/19959620/quic-fragmented-crypto-56802.pcap.txt)

---

There's also an issue with ACKPayload parsing that's fixed with the following hunk:

```
diff --git a/src/analyzer/protocol/quic/QUIC.spicy b/src/analyzer/protocol/quic/QUIC.spicy
index 0d6064d4c6..e910da1204 100644
--- a/src/analyzer/protocol/quic/QUIC.spicy
+++ b/src/analyzer/protocol/quic/QUIC.spicy
@@ -277,8 +277,8 @@ public type Frame = unit(header: LongHeaderPacket, from_client: bool, crypto_sin
 
   # TODO: add other FrameTypes as well
   switch ( self.frame_type ) {
-    FrameType::ACK1 -> a: ACKPayload;
-    FrameType::ACK2 -> b: ACKPayload;
+    FrameType::ACK1 -> a: ACKPayload(FrameType::ACK1);
+    FrameType::ACK2 -> b: ACKPayload(FrameType::ACK2);
     FrameType::CRYPTO -> c: CRYPTOPayload(from_client) {
       # Have the sink re-assemble potentially out-of-order cryptodata
       crypto_sink.write(self.c.cryptodata, self.c.offset.result_);
@@ -298,11 +298,26 @@ type CRYPTOPayload = unit(from_client: bool) {
   cryptodata: bytes &size=self.length.result_;
 };
 
-type ACKPayload = unit {
+# https://datatracker.ietf.org/doc/html/rfc9000#ack-ranges
+type ACKRange = unit {
+  gap: VariableLengthInteger;
+  ack_range_length: VariableLengthInteger;
+};
+
+type ACKECNCounts = unit {
+  ect0: VariableLengthInteger;
+  ect1: VariableLengthInteger;
+  ecn_ce: VariableLengthInteger;
+};
+
+# https://datatracker.ietf.org/doc/html/rfc9000#name-ack-frames
+type ACKPayload = unit(frame_type: FrameType) {
   latest_ack: VariableLengthInteger;
   ack_delay: VariableLengthInteger;
   ack_range_count: VariableLengthInteger;
   first_ack_range: VariableLengthInteger;
+  ack_ranges: ACKRange[self.ack_range_count.result_];
+  ecn_counts: ACKECNCounts if(frame_type == FrameType::ACK2);
 };
 
 type ConnectionClosePayload = unit(header: LongHeaderPacket) {

```

awelzel:
* Fixes #4405 
* fixes using the wrong destination connection id for decryption when the client starts using a different one in a later INITIAL packet
* Fixes ACK ranges parsing
* Fixes INITIAL packets consuming the remainder of the UDP datagram where there could've been other PACKET types.

Personally think this is 7.0 and 7.2 material, unfortunately.


src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -22,19 +22,81 @@ public function decrypt_crypto_payload(
...
+  # Can only decrypt the responder if we've seen the initial destination conn id.
+  if ( ! state.is_orig && |context.initial_destination_conn_id| == 0 )
+    return False;
+
+  # Only attempt decryption if we haven't flushed some SSL data yet.
+  return ! state.flushed;
+}
+
+# Buffer all CRYPTO frames, which might be fragmented and unordered
+# across multiple INITIAL packets into this CyptoBuffer unit.
```
bbannier, line 39: Do I understand correctly that this could buffer potentially unbounded data (well, at most uint24 bytes)? I am not sure I understand why we need to do this, and would much prefer an approach which passes data on as soon as we see it. Doing that would likely also simplify the tracking for `complete` state which currently happens in multiple places.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -22,19 +22,81 @@ public function decrypt_crypto_payload(
...
+    if ( ! self.length && |self.buffered| >= 4 ) {
+      # XXX: This looks terrible
+      self.length = (cast<uint32>(*self.buffered.at(1)) << 16) + (cast<uint32>(*self.buffered.at(2)) << 8) + (cast<uint32>(*self.buffered.at(3)));
+    }
+
+    if ( self.length && |self.buffered| >= uint64(*self.length) + 4 ) {
+      self.complete = True;
+    }
+  }
+};
```
bbannier, line 68: Checking whether we still need to parse the header for each chunk does unnecessary work at runtime, suggest to slightly restructure the code to avoid this.

```suggestion
type CryptoBuffer = unit {
  var buffered: bytes;
  var length: optional<uint32>;
  var complete: bool = False;

  # The first 4 bytes of crypto data contain the expected tag and a
  # 24bit length from the TLS HandshakeMessage. Extract the length
  # so we can determine when all CRYPTO frames have arrived.
  #
  # https://datatracker.ietf.org/doc/html/rfc8446#section-4
  #
  #  struct {
  #    HandshakeType msg_type;    /* handshake type */
  #    uint24 length;             /* remaining bytes in message */
  #    ...
  #

  # msg_type.
  : bytes &size=1 {
    self.buffered += $$;
  }

  # length.
  : bytes &size=3 {
      self.length = (cast<uint32>(*$$.at(0)) << 16) + (cast<uint32>(*$$.at(1)) << 8) + (cast<uint32>(*$$.at(2)));
      self.buffered += $$;
  }

  : bytes &chunked &eod {
    self.buffered += $$;

    if ( |self.buffered| >= uint64(*self.length) + 4 ) {
      self.complete = True;
    }
  }
};
```

Handling type and length separately here avoids having to remember the type when indexing into the parsed data.

If we ever end up parsing more uint24 values we could give the `length` field above a name and then parse length with a dedicated uint24 parser from that with `&parse-from`.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -524,36 +606,16 @@ type Packet = unit(from_client: bool, context: ConnectionIDInfo&) {
...
-
-      # Stop decryption attempts after processing the very first INITIAL
-      # INITIAL packet for which we forwarded data to the SSL analyzer.
-      if ( from_client )
-        context.client_initial_processed = True;
-      else
-        context.server_initial_processed = True;
-    }
+    if ( crypto_state_complete(self.crypto_state) )
+      crypto_state_flush(context, self.crypto_state);
```
bbannier, line 618: I'd suggest to inline `crypto_state_complete` into `crypto_state_flush` so it does nothing if we are not complete (checking for that would probably become unnecessary if we didn't buffer).

If we still end up buffering we should also add an identical handler for `on %error` so we do not loose buffered data for some down the line parse error.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -22,19 +22,81 @@ public function decrypt_crypto_payload(
...
+  flushed: bool &default=False;
+  buf: CryptoBuffer&;
+  snk: sink&;
+};
+
+function crypto_state_complete(state: CryptoState&): bool {
+  return state.buf != Null && state.buf.complete;
+}
+
+function crypto_state_flush(context: ConnectionIDInfo&, state: CryptoState&) {
```
bbannier, line 81: Any reason we need to pass `context` as explicit ref? By default this should just work. Also, instead of relying on implicit interior mutability for refs for modifying `state` we should be able to modify it by explicitly passing it as `inout`.

```suggestion
function crypto_state_flush(context: ConnectionIDInfo, inout state: CryptoState) {
```

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -298,11 +360,26 @@ type CRYPTOPayload = unit(from_client: bool) {
...
+};
+
+# https://datatracker.ietf.org/doc/html/rfc9000#name-ack-frames
+type ACKPayload = unit(frame_type: FrameType) {
   latest_ack: VariableLengthInteger;
   ack_delay: VariableLengthInteger;
   ack_range_count: VariableLengthInteger;
   first_ack_range: VariableLengthInteger;
+  ack_ranges: ACKRange[self.ack_range_count.result_];
+  ecn_counts: ACKECNCounts if(frame_type == FrameType::ACK2);
```
bbannier, line 382: Nit: for 7.2 and above we would probably write this as 
```suggestion
  if (frame_type == FrameType::ACK2) {
    ecn_counts: ACKECNCounts if(frame_type == FrameType::ACK2);
  };
```

This avoids too much "interesting" stuff happening at the very right-hand side.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -453,50 +530,55 @@ type Packet = unit(from_client: bool, context: ConnectionIDInfo&) {
...
-    self.crypto_sink.connect(self.crypto_buffer);
 
     if ( from_client ) {
       context.server_cid_len = self.long_header.dest_conn_id_len;
       context.client_cid_len = self.long_header.src_conn_id_len;
 
+      # This is the first INITIAL packet we attempt to decrypt and it is
+      # coming from the client. Use its destination connection ID for
+      # decryption purposes.
+      if ( |context.initial_destination_conn_id| == 0 ) {
```
bbannier, line 567: Nit: `initial_destination_conn_id` should really be an `optional<bytes>`, would you mind cleaning that up? With that the length checks would become just `!context.initial_destination_conn_id`, the "reset" above would assign `Null` instead, and value access would deref.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -524,36 +606,16 @@ type Packet = unit(from_client: bool, context: ConnectionIDInfo&) {
   # Depending on the type of header and whether we were able to decrypt
   # some of it, parse the remaining payload.
```
bbannier, line 607: This comment is slightly outdated now, could you update it?

bbannier (commented):
Left some comments. My biggest concern would be around the new buffering which seems to be practically unbounded, but maybe I am missing something.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: > Any reason we need to pass context as explicit ref?

I'm not Spicy-versed enough to understand what the difference is without looking at the generated C++ code. That is not my workflow - it could, but I'm playing naive. I'm usually stopping at what works and what I might have used elsewhere (`CryptoState&` within Context). If one should be strictly preferred over the other and/or the `&` is redundant, compiling the analyzer probably should result in a warning/hint that that's so.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: Nice, thanks!

I'll try to switch to `uint8` and `uint8[3]` as that should be cheaper, but I suspect appending to `buffered` might be a bit quirky.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: Yeah, good call out! We can put a limit on the number of bytes we feed into the sink and just throw if that's exceeded, but also directly forward to the SSL analyzer from within the CryptoBuffer unit. I'll dabble with that!

I'm wondering if Spicy should provide a feature/limit for data that is pending in a `sink` and throw if it's exceeded.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: Leaving as is due to 7.0 backport candidate. The trailing `if` is something Ruby has, too, so I'm not all that upset about it once in a while. Also, whenever I see an actual `if` block, my mind still goes into "this isn't parsing code" mode. That might be personal and I'm sure I'll overcome this, but for simpler lines the trailing `if` seems totally fine.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: The function is gone. I tried to remove `&` elsewhere, but it seems `inout` only works with units, not with structs, producing the following:
```
[error] /home/awelzel/corelight-oss/zeek/src/analyzer/protocol/quic/QUIC.spicy:476:39-476:69: unsupported type for unit parameter 'context': type of inout unit parameters must itself be a unit; for other parameter types, use references instead of inout
```

(that was for another unit rather than a function).

src/analyzer/protocol/quic/QUIC.spicy
awelzel: I've pushed a fixup that reworks this to avoid the buffering. Let me know what you think. Thanks!

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -272,16 +319,28 @@ public type LongHeaderPacket = unit {
...
+      # don't attempt to write more bytes into the sink. If it doesn't,
+      # use 2000 bytes as an arbitrary limit required to observe the
+      # length of the contained Client Hello or Server Hello.
+      if ( crypto.length > 0 ) {
+        if ( |crypto_sink| > crypto.length )
+          throw "too much crypto data received %s > %s" % ( |crypto_sink|, crypto.length);
+      } else {
+        if ( |crypto_sink| > 2000 )
+          throw "too much crypto data without length received %s" % |crypto_sink|;
+      }
```
bbannier, line 343: Maybe more as a FYI, `throw` raises a `ParseError` which could be recovered from (this parser currently performs no synchronization though). If we used `assert` which now can take an optional error message we'd raise a non-recoverable error.

src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -453,58 +529,63 @@ type Packet = unit(from_client: bool, context: ConnectionIDInfo&) {
...
           unset context.ssl_handle;
+          context.client_crypto = Null;
+          context.server_crypto = Null;
+          context.client_sink = Null;
+          context.server_sink = Null;
+          self.crypto = Null;
+          self.crypto_sink = Null;
+
+          # Reset crypto state!
+          context.initial_destination_conn_id = Null;
```
bbannier, line 544: Nit: we could move this to a dedicated function, and call it from a new `on %error` impl. We don't need this right now since we do not perform resynchronization, but it might still make sense since we still have all the context swapped in.

bbannier (approved):
Nice, thanks for the cleanup! 🚀

I still left a few comments, mostly FYI.

awelzel:
@timwoj - this would be nice for 7.2 and a 7.0 dot release to improve visibility.

timwoj:
I pulled it into the release/7.2 branch and added it to the list for 7.0 👍🏼

src/analyzer/protocol/quic/QUIC.spicy
awelzel: I didn't do anything here as it's highly unlikely we'll be able to synchronize an encrypted QUIC stream.

PR #4411, opened 2025-04-30
Add find_first string function

### LINKED ISSUE #4399: Add find_first bif
We have find_last and a collection of find_all BIFs, but not one that returns the first match of a pattern. We have strsep, but it searches for a string not for a pattern.

XueSongTap:
## Summary
This PR adds a new `find_first` function to the `strings.bif` file. It works similarly to the existing `find_last` function but returns the first match of a regular expression in a string.

## Changes
- Added the `find_first` function in `strings.bif` that takes a string and a pattern, and returns the first match or an empty string if no match is found.
- Updated `FuncInfo.cc` to register the new function with Zeek's internal function list.
- Added test cases to ensure that `find_first` behaves as expected with various input strings and patterns.

## Testing
- Verified the function works by adding unit tests that check various scenarios (empty strings, no matches, multiple matches, etc.).
- Ran the test suite and confirmed all tests pass.
```
yxc@yxc-MS-7B89:~/code/2504/zeek_dev$ ./build/src/zeek testing/btest/bifs/find_first.zeek 
hi
-------------------
0
```
## Related Issues
- Fixes  #4399


timwoj (changes requested):
Looks ok to me. You'll need to commit a baseline for the btest so that the test will pass on CI. Run `btest -U bifs/find_first.zeek` and commit the new file(s) added to the `Baseline` directory.

src/script_opt/FuncInfo.cc
```diff
@@ -265,6 +265,7 @@ static std::unordered_map<std::string, unsigned int> func_attrs = {
     {"find_entropy", ATTR_FOLDABLE},
     {"find_in_zeekpath", ATTR_IDEMPOTENT}, // can error
     {"find_last", ATTR_FOLDABLE},
+    {"find_first", ATTR_FOLDABLE},
```
vpax, line 268: Nit: these should be in alphabetical order, so move up one row

src/strings.bif
```diff
@@ -1107,6 +1107,30 @@ function find_last%(str: string, re: pattern%) : string
...
+## Finds the first occurrence of a pattern in a string.
+##
+## str: The string to inspect.
+##
+## re: The pattern to look for in *str*.
+##
+## Returns: The first string in *str* that matches *re*, or the empty string.
+##
+## .. zeek:see:: find_all find_all_ordered find_last strstr
+function find_first%(str: string, re: pattern%) : string
```
vpax, line 1119: I think it would be nice if there's an optional third parameter giving the offset where to start looking (and then add the same to `find_last`)

timwoj: I'll make a ticket to do this separately.

timwoj:
Thanks for the contribution!

PR #4412, opened 2025-04-30
ZAM-bif-tracking: Remove ZeroMQ dependency

awelzel:
Vern didn't have ZeroMQ installed and the test was skipped for him. Generally would recommend anyone working on core Zeek to install libzmq-dev or the equivalent for their environment, but until it is a real required dependency, loosen the requirements on the test.


vpax (approved):
LGTM, thanks!

PR #4413, opened 2025-04-30
Add control logic for Intel event groups

Mohan-Dhawan:
This PR adds control logic for Intel event groups hooking on `_insert` and `remove` functions. We keep a mapping between the item type and the item source, and enable the event group when the item is inserted, and remove the event group if there is no source corresponding to it.


awelzel:
@Mohan-Dhawan - instead of adding this into the base scripts. Would you be up to set this up as an optional feature in `scripts/policy` that can be opt'ed in? You can re-open the `Intel` module in those scripts if you need access to internal state, but also introduce new hooks/events if needed.

Designing a stand-alone script in`scripts/policy` would also allow you to develop/iterate on it without needing to have the changes in base/upstream Zeek.

scripts/policy/frameworks/intel/seen/smtp-url-extraction.zeek
```diff
@@ -21,7 +21,7 @@ event intel_mime_data(f: fa_file, data: string) &group="Intel::URL"
 		}
 	}
 
-event file_new(f: fa_file) &group="Intel::URL"
+event file_new(f: fa_file)
```
awelzel, line 24: What's the motivation for this one? Not adding the `ANALYZER_DATA_EVENT` at all seems a useful optimization.

Mohan-Dhawan: I wasn't sure if the DATA_EVENT was useful, and it was causing baseline changes for one of the tests.

Mohan-Dhawan:
> @Mohan-Dhawan - instead of adding this into the base scripts. Would you be up to set this up as an optional feature in `scripts/policy` that can be opt'ed in? You can re-open the `Intel` module in those scripts if you need access to internal state, but also introduce new hooks/events if needed.
> 
> Designing a stand-alone script in`scripts/policy` would also allow you to develop/iterate on it without needing to have the changes in base/upstream Zeek.

Sure, I can give it a try.

Mohan-Dhawan:
@awelzel Any feedback on this new version? I have not yet fixed the baseline for the failing test, as I'm not sure if that is the right change.

awelzel (changes requested):
Hey Mohan,

I see you attempted to do this via existing events, some of which are marked "internal for events for (cluster) data distribution". After some pondering, what do you think about:

This might be easier to implement if we'd in a first step add hooks in the `_insert()` and `remove()` and `remove_indicator()` functions that are reliably invoked the first time a new item is inserted and reliably called when it is removed again - regardless of whether the code runs on a manager or worker. I think these functions have enough information to figure this out.

Once these hooks exist as primitives, the actual event group management could leverage these and track item counts in a `table[Type] of count &default=0` and enable/disable once if `>0` or `== 0`.

Essentially, split the features into introducing generic primitives, then leverage these to implement the feature, rather than relying on existing events that may or may not fulfill your need.

Happy to chat or have a call if you need guidance.

Mohan-Dhawan:
> This might be easier to implement if we'd in a first step add hooks in the `_insert()` and `remove()` and `remove_indicator()` functions that are reliably invoked the first time a new item is inserted and reliably called when it is removed again

Yes, it is exactly what I had done in the previous implementation, but I had modified the base files. If adding new hooks is acceptable, then I'll go with that.

scripts/base/frameworks/intel/main.zeek
```diff
@@ -497,6 +513,9 @@ function insert_meta_data(item: Item): bool
 # indicates whether the item might be new for other nodes.
 function _insert(item: Item, first_dispatch: bool &default = T)
 	{
+	if ( ! hook insert_item(item) )
```
awelzel, line 516: Placing this here seems almost equivalent with the existing `filter_item()` hook.

It also doesn't work as you require. E.g. if one calls `Intel::insert()` with the same item twice, this hook is invoked twice, even though there's only one entry in the data_store / min_data_store.

scripts/base/frameworks/intel/main.zeek
```diff
@@ -598,6 +617,9 @@ function remove(item: Item, purge_indicator: bool)
 		return;
 		}
 
+	if ( ! hook purge_item(item) )
+		return;
+
```
awelzel, line 622: We also (or only - haven't decided yet) need to hook into `remove_indicator()` below, as that's the event that is published by the manager to remove items on workers

The flow of the events in the intel framework is mind-twisting just much too hard to follow IMO. This definitely needs a few tests, also with a live/test cluster.

awelzel (commented):
Hey @Mohan-Dhawan - let me attempt to add the hooks in a separate PR with tests. Probably easier than the forth and back.

awelzel:
Replaced by #4457

PR #4414, opened 2025-04-30
Add operation_timeout and command_timeout storage backend options

### LINKED ISSUE #4409: `Storage::BackendOptions` needs a timeout value
The storage framework lacks a way to set a timeout for server-based storage backends. One should be added to the `BackendOptions` record and used in the Redis backend.

timwoj:
Fixes #4409


0xxon:
This is a potential a bit generic question. Do we expect that most backends will support timeouts? And do you think this is something one will want to tweak for an individual backend, or just in general (e.g. "all redis backends").

I am mostly wondering if supported timeouts might be something that's rather backend specific, and if it might make sense to have backend-specific options - that way there is no guesswork around which options are supported for which backend.

evantypanski:
> Do we expect that most backends will support timeouts?

Just as a data point: The [NATS backend](https://github.com/evantypanski/nats-storage) could [use the connection timeout effectively](https://docs.nats.io/using-nats/developer/connecting/connect_timeout).

Unsure about the operation timeout, at a quick glance it's if there's one for keyval operations, just watching. But I may be wrong. That may be evidence that timeouts should be more backend-specific - it would seem weird to a user that they could use one but not the other..

scripts/policy/frameworks/storage/backend/redis/main.zeek
```diff
@@ -22,6 +22,15 @@ export {
 		# same server. Defaults to an empty string, but preferably should be set
 		# to a unique value per Redis backend opened.
 		key_prefix: string &default="";
+
+		## Timeout for connection attempts to the backend. Connection attempts
+		## that exceed this time will return
+		## :zeek:see:`Storage::CONNECTION_FAILED`.
+		connect_timeout: interval &default=5 sec;
```
awelzel, line 29:
```suggestion
		connect_timeout: interval &default=default_connect_timeout;
```

Would make sense to add a module wide default (as const &redef). That way the timeout for every Redis connection that doesn't override it explicitly be changed at once. Same for the operation_timeout.

awelzel (approved):
LGTM

PR #4415, opened 2025-04-30
Add Storage::is_connected BIF

timwoj:
I found the need to check whether a backend handle was connected while tinkering with the redis version of the known-services script. This adds a BIF to do that. It also cleans up some code duplication between the sync/async bif files.


src/storage/storage.bif
```diff
@@ -0,0 +1,18 @@
...
+
+using namespace zeek;
+using namespace zeek::storage;
+
+%%}
+
+module Storage;
+
+function Storage::is_connected%(backend: opaque of Storage::BackendHandle%) : bool
+	%{
```
0xxon, line 15: Nit: indenting of this line is off.

src/storage/storage.bif
```diff
@@ -0,0 +1,18 @@
...
+#include "zeek/storage/Backend.h"
+
+using namespace zeek;
+using namespace zeek::storage;
+
+%%}
+
+module Storage;
+
+function Storage::is_connected%(backend: opaque of Storage::BackendHandle%) : bool
```
0xxon, line 14: I know you only needed this for testing - but BIFs really should have documentation attached to them, even if we only use them for our tests.

src/storage/Backend.cc
```diff
@@ -164,4 +164,20 @@ bool detail::BackendHandleVal::DoUnserializeData(BrokerDataView) {
     return false;
 }
 
+namespace detail {
+
+zeek::expected<storage::detail::BackendHandleVal*, OperationResult> BackendHandleVal::CastFromAny(Val* handle) {
+    auto b = static_cast<storage::detail::BackendHandleVal*>(handle);
```
awelzel, line 170:
```suggestion
    auto b = dynamic_cast<storage::detail::BackendHandleVal*>(handle);
```

awelzel: You could probably also poke first if the Val's type tag is `TYPE_OPAQUE` and/or whether its type is `detail::backend_opaque`. That's probably faster than `dynamic_cast`.

Given this is an API on `BackendHandleVal` now (rather than static helpers) I find the mixing of cast and checking if it's open/connected unfortunate. It does more than just cast. But the API is in the detail namespace, so can also be cleaned up later I guess.

src/storage/storage.bif
```diff
@@ -0,0 +1,24 @@
...
+
+module Storage;
+
+## Checks whether a storage backend is connected.
+##
+## backend: A handle to the backed to check.
+##
+## Returns: T if the backend is open, F if the backend is not connected or if the handle
+##          is invalid.
+function Storage::is_connected%(backend: opaque of Storage::BackendHandle%) : bool
```
awelzel, line 20: Would you have gripes renaming this to `Storage::is_open` ? The accessor on `Backend` is also `IsOpen()`. The consistency would help reduce cognitive overhead and "connection" (to me anyway) sounds like a remote connection, which isn't necessarily true.

src/storage/Backend.cc
timwoj: Would `ConvertOpaque` be better?

src/storage/storage.bif
timwoj: Seems reasonable to me.

src/storage/Backend.cc
timwoj: The type tag is an easy check, but the name doesn't get set.

awelzel: > The type tag is an easy check, but the name doesn't get set.

Sorry for the confusion - this compiles here and should work:

```
-    if ( handle->GetType()->Tag() != TYPE_OPAQUE )
+    if ( handle->GetType() != detail::backend_opaque )
         return zeek::unexpected<OperationResult>(
             OperationResult{ReturnCode::OPERATION_FAILED, "Invalid storage handle type"});
```


> Would ConvertOpaque be better?

I was more coming from the direction that cast/convert and *also* doing the the `IsOpen()` check might not be obvious. For a mouthful: `CastAndEnsureOpen` or maybe `EnsureStorageHandleOpen(Val *any)` that happens to return a casted version or an error.

But this is nitpickery.

awelzel (approved):

PR #4416, opened 2025-04-30
Redis: Check server version when connecting

### LINKED ISSUE #4408: Redis backend requires at least version 6.2.0
The backend uses the `PXAT` option for the `SET` command, which wasn't added until Redis 6.2.0. `zeek-test` at LBL wasn't storing data to a Redis backend correctly. After some investigation, I realized it's running Redis 6.0.16. Also, the `SET` command should have returned an error to be handled and we're either not handling it or it's not being returned.

There's a couple options here:
1. Use the `INFO` command, query what version of server we have, and throw an error if it's too old.
2. Switch to use the `PX` option instead. `PXAT` takes a UNIX timestamp in milliseconds, whereas `PX` takes a time relative to current time in milliseconds.

timwoj:
I ran into this while testing the Redis version of known-services on zeek-test. The version of redis-server there is too low for the capabilities we're using. This PR adds a check during connection to ensure that we're using a new enough version of Redis and returns an error if it's too low.

On zeek-test:

> error in /.../share/zeek/policy/protocols/conn/known-services.zeek, line 123: zeekknownservices: Failed to open backend connection: Redis server version is too low: Found 6.0.16, need 6.2.0

Fixes #4408


src/storage/backend/redis/Redis.h
```diff
@@ -41,6 +41,7 @@ class Redis final : public Backend, public iosource::IOSource {
     void HandleGetResult(redisReply* reply, ResultCallback* callback);
     void HandleEraseResult(redisReply* reply, ResultCallback* callback);
     void HandleGeneric(redisReply* reply);
+    void HandleInfoResult(redisReply* reply);
```
0xxon, line 44: I know that this is probably a lost fight in this cause - but I want to point out that, in principle, it would be neat for functions in header files to contain comments, following our [style guide](https://github.com/zeek/zeek/wiki/Coding-Style-and-Conventions#commenting).

Same applies below.

0xxon (approved):
Just minor comments, in general this looks good to me.

PR #4417, opened 2025-04-30
Use std::string_view in Redis::DoExpire to avoid copies

timwoj:
This was a one-off patch I've had sitting that I finally decided to polish up and open a PR for. The curious thing is the change to use `util::fmt`. You're supposed to pass a size to `%s` when using `std::string_view`'s `data` method, since it's not guaranteed to be null-terminated. For some reason, whatever formatter hiredis uses for printf-style strings doesn't like that and hangs forever. `util::fmt` does handle it correctly.


src/storage/backend/redis/Redis.cc
```diff
@@ -388,37 +388,40 @@ void Redis::DoExpire(double current_network_time) {
...
+        elements.emplace_back(reply->element[i]->str, reply->element[i]->len);
 
     // TODO: it's possible to pass multiple keys to a DEL operation but it requires
     // building an array of the strings, building up the DEL command with entries,
     // and passing the array as a block somehow. There's no guarantee it'd be faster
     // anyways.
     for ( const auto& e : elements ) {
-        status = redisAsyncCommand(async_ctx, redisGeneric, NULL, "DEL %s:%s", key_prefix.data(), e.c_str());
+        status = redisAsyncCommand(async_ctx, redisGeneric, NULL,
+                                   util::fmt("DEL %s:%.*s", key_prefix.data(), static_cast<int>(e.size()), e.data()));
```
0xxon, line 403: would it perhaps make sense to put a comment here why util::fmt is being used? Similar to the commit message. Otherwise this might be confusing to potential future reviewers.

0xxon (approved):

PR #4418, opened 2025-05-01
Optimize software found cluster communication

JustinAzoff:
As a followup to 3bf8c8ceb6d4b5d3444e43d9e0212ec8bef35574 that added the parse cache, add a small short lived cache on the workers to effectively debounce the number of Software::new events sent up to the proxies.

User-Agents are highly repetitive, workers often see exact duplicate user-agents on the same orig_h. Worse, due to NAT, virtualization, and the proliferation of Electron based applications, variations of the same user-agent can be seen at the same time.  For example:

	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.6613.18 Safari/537.36 Zoom/6.2.0 (1855)

When these two user-agents are seen concurrently, the software framework will log each flip as a new user-agent. This can be fixed separately on the proxy side, but a reduction of Software::new events is still needed to reduce cluster communication overhead as well as the load on the proxies.

With a 10 minute cache on the workers, this should greatly reduce the number of redundant user-agents logged in the software.log


JustinAzoff:
Ah, forgot to update the baselines. there's a number of duplicate software entries that this change removes.

scripts/base/frameworks/software/main.zeek
```diff
@@ -239,7 +239,10 @@ function parse(unparsed_version: string): Description
 	return [$version=v, $unparsed_version=unparsed_version, $name=alternate_names[software_name]];
 	}
 
+# A cache for the proxies that stores the result of parsing unparsed_version.
 global parse_cache: table[string] of Description &read_expire=65secs;
+# A suppression cache for the workers to prevent sending the same information to the proxies multiple times.
+global found_cache: set[Info] &create_expire=10mins;
```
awelzel, line 245: Nit: I wanted to suggest to make the the expiration time `const-redef`, but I guess that won't even work due to parsing-time-evaluates-before-redef-in-effect? :-(

scripts/base/frameworks/software/main.zeek
```diff
@@ -523,6 +526,11 @@ function found(id: conn_id, info: Info): bool
 	if ( ! info$force_log && ! addr_matches_host(info$host, asset_tracking) )
 		return F;
 
+	# This assumes that callers do not fill in info$ts, none of the current callers do.
+	if (info in found_cache)
```
awelzel, line 530: Nit.
```suggestion
	if ( info in found_cache )
```

awelzel: I wonder a bit if you want to provide `Cluster::node_down()` and/or `Cluster::node_up()` to clear out `found_cache` if a proxy restarts. That pattern exists in the known-scripts. With the assumption that proxies stay up (and if they are down, just for a short period of time), it might be better to not do that.

awelzel: There is a `info$force_log` attribute. Not sure if that should ignore the caching mechanism, or whether for its current usage that isn't relevant.

JustinAzoff: I'm not sure if clearing out the cache is really required.. or at least, I'm not sure what it would accomplish.  The cache is short enough that a crashing proxy likely means there are bigger problems than a potentially missed software log entry.

I do want to further rework the data structure on the proxy side to better handle the version flip flops on that side as well.

force_log is only even used in one script and it doesn't look like there is a test for it.  I need to figure out why it uses it and what problem it solves.

JustinAzoff:
Oh, and I remembered the other area where this has a huge performance impact: internal http traffic.

If zeek sees requests to an internal nginx server, it'll trigger software found events for nginx for every single request.  This can easily be 10,000 times per second.  Without this suppression, every request triggers a proxy event.

awelzel:
@JustinAzoff - are you happy with this version? Mind updating the version changes baselines if so?

JustinAzoff:
@awelzel mostly happy, but I need to study the baseline changes in more detail.  The change breaks testing/btest/scripts/policy/frameworks/software/version-changes.zeek.  That is testing that the minor version changes from 0 -> 1 -> 2 -> 0.  With the suppression, the change from 2 back to 0 is lost since 0 has been seen before.

This is similar to the issue I am trying to fix where we see a browser change from "chrome" to "chrome (zoom)" and "chrome (slack)" and then back to "chrome".

I'm not sure if this is a problem in practice, but it does change the behavior slightly.

scripts/base/frameworks/software/main.zeek
ckreibich: Yeah, it's likely better to just keep this as-is — then one could always adjust via something like
```
event zeek_init()
    {
    Software::found_cache = set() &create_expire=20mins;
    }
```
if really needed.

ckreibich:
I'm finally dusting this off ... the branch rebases cleanly and the observation about the failing version-changes test still holds. It seems fine to me to change this behavior since you still see all versions — you just don't see them repeatedly. And if we use a `zeek_init()` approach for assigning `Software::found_cache`, we could actually use a redefable interval, and disable the caching when set to 0.

Thoughts?

JustinAzoff:
The other thing this doesn't fully address is that

```zeek
	type SoftwareSet: table[string] of Info;
```
should probably be


```zeek
	type SoftwareSet: table[string] of set[Info];
```

like the name hints.. this would allow the proxies to  store

```
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.6613.18 Safari/537.36 Zoom/6.2.0 (1855)
```

As 

```
"Chrome" -> { "Mozilla ... Safari/537.36", "Mozilla .. Zoom/6.2.0 (1855)" }
```

this would further reduce the amount of duplicate log entries.

ckreibich:
Makes sense — and the old-vs-new [version comparison](https://github.com/zeek/zeek/blob/topic/justin/software-framework-found-cache/scripts/base/frameworks/software/main.zeek#L515-L518) in `Software::register()` would become a membership test in that new set, yes?

It's technically an API change ... seems pretty unlikely that anyone uses `SoftwareSet`. I'm less sure about the `tracked` table, which the module also exports. We could just use a new one and deprecate use of the old.

ckreibich:
This spiraled a bit but I have better proxy-side redundancy suppression in place. What I like best about it is that it actually surfaces some versions that were previously swallowed, because the version comparison worked on the parsed version only, rather than the unparsed (more detailed) one when available. Take a look here:

https://github.com/zeek/zeek-testing/compare/master...topic/justin/software-framework-found-cache

Let's see how this does in CI, still gotta get used to the new setups. :-)

scripts/base/frameworks/software/main.zeek
```diff
@@ -73,6 +73,24 @@ export {
...
 
+	## The framework maintains a cache that maps unparsed version strings
+	## to :zeek:type:`Software::Version` instances. This is its expiration
+	## interval.
+	const parse_cache_interval = 65secs &redef;
+
+	## The framework maintains a redundancy cache that keeps the workers
+	## from reporting the same version number observations repeatedly in
+	## rapid succession. This is its expiration interval. Setting to 0secs
+	## disables this cache.
```
awelzel: This is total naming, but I think the mechanism here is just "deduplication" which should be a fairly universally understood term.


> The framework maintains a cache to deduplicate redundant publishes
> over the configured interval to reduce cluster load. Setting to 0secs disables
> the deduplication mechanism.

Could maybe make the NEWS section about the behavior a bit more concise.

testing/btest/scripts/policy/frameworks/software/version-changes.zeek
```diff
@@ -37,3 +38,15 @@ event zeek_init()
 	{
 	event new_software();
 	}
+# @TEST-END-FILE
+
+@load ./common.zeek
+
+# @TEST-START-NEXT
+# Disable the hit suppression on the workers -- this is the historical behavior.
```
awelzel, line 46:
```suggestion
# Disable publish dedupe on the workers -- this is the historical behavior.
```

scripts/base/frameworks/software/main.zeek
```diff
@@ -491,30 +548,37 @@ event Software::new(info: Info)
...
+		else
+			event Software::version_change(ss[info$name]$last, info);
+
+		add ss[info$name]$versions[software_fmt_unparsed_version(info)];
+		ss[info$name]$last = info;
+
+		# If the set just got too large, simply start over with the
+		# newest version. We currently lack a good way to track a set
+		# and its newest/oldest members.
+		if ( |ss[info$name]$versions| > max_software_cache_size )
```
awelzel: Nit. Would `max_software_cache_size == 0` imply unbounded make sense? Seems setting to `0` or `1` doesn't really make a difference with this approach.

awelzel (approved):

scripts/base/frameworks/software/main.zeek
ckreibich: Good one, yep! I've added this.

scripts/base/frameworks/software/main.zeek
ckreibich: Thanks, I've done some word wrangling. :-)

PR #4428, opened 2025-05-06
Fix shutdown hang when WebSocket clients blocked

### LINKED ISSUE #4420: Zeek-stack websocket (silent) disconnect at high message rates and 100% CPU hang on shutdown
@awelzel this follows up on @MP-Corelight's request [here](https://github.com/zeek/broker/issues/458#issuecomment-2845840586) to try reproduce the issue reported in https://github.com/zeek/broker/issues/458 with the new (non-broker) websocket implementation.

I have updated my [reproducer project](https://github.com/simeonmiteff/repro-zeek-broker-ws-disconnect) and:
- Added `test-cluster.zeek` which uses the `Cluster::listen_websocket()` API
- Switched the Go client from `github.com/gorilla/websocket` to `github.com/coder/websocket` (the latter is being maintained)

I observed the following hang:

Client
```
$ go build && ./repro-zeek-broker-ws-disconnect 
2025/05/02 10:22:48 connected to broker endpoint=e4bbc4af-b3b6-53bf-9208-066394846405-websocket-0 version=8.0.0-dev.15-debug
```

Zeek:
```
$ time zeek test-cluster.zeek 
warning in ./test-cluster.zeek, line 16: handler for non-existing event cannot be invoked (Test::evt)
peer added, endpoint=[id=e4bbc4af-b3b6-53bf-9208-066394846405-websocket-0, network=[address=127.0.0.1, bound_port=47338/tcp]], subscriptions=[]
^Creceived termination signal
Last count was, 37544
```

Note the ctrl-C. At this point it hangs, and a minute later (the timeout) the client exits:
```
2025/05/02 10:23:48 exited with error=failed to write msg: failed to write frame: context deadline exceeded and counter=78912
```

Now zeek is using 100% CPU, and continues to do so for 5 minutes, at which point I attach gdb:
```
x::SocketServer::closeTerminatedThreads (this=0x5910b4f79ea0)
    at /home/simeon/Code/Corelight/zeek/src/cluster/websocket/auxil/IXWebSocket/ixwebsocket/IXSocketServer.cpp:249
249	    {
(gdb) bt
#0  ix::SocketServer::closeTerminatedThreads (this=0x5910b4f79ea0)
    at /home/simeon/Code/Corelight/zeek/src/cluster/websocket/auxil/IXWebSocket/ixwebsocket/IXSocketServer.cpp:249
#1  0x000059109caf3669 in ix::SocketServer::runGC (this=0x5910b4f79ea0)
    at /home/simeon/Code/Corelight/zeek/src/cluster/websocket/auxil/IXWebSocket/ixwebsocket/IXSocketServer.cpp:443
#2  0x000059109caf6cd9 in std::__invoke_impl<void, void (ix::SocketServer::*)(), ix::SocketServer*> (
    __f=@0x5910b5f6de60: (void (ix::SocketServer::*)(ix::SocketServer * const)) 0x59109caf35ea <ix::SocketServer::runGC()>, 
    __t=@0x5910b5f6de58: 0x5910b4f79ea0) at /usr/include/c++/13/bits/invoke.h:74
#3  0x000059109caf6a72 in std::__invoke<void (ix::SocketServer::*)(), ix::SocketServer*> (
    __fn=@0x5910b5f6de60: (void (ix::SocketServer::*)(ix::SocketServer * const)) 0x59109caf35ea <ix::SocketServer::runGC()>)
    at /usr/include/c++/13/bits/invoke.h:96
#4  0x000059109caf68b1 in std::thread::_Invoker<std::tuple<void (ix::SocketServer::*)(), ix::SocketServer*> >::_M_invoke<0ul, 1ul> (
    this=0x5910b5f6de58) at /usr/include/c++/13/bits/std_thread.h:292
#5  0x000059109caf67da in std::thread::_Invoker<std::tuple<void (ix::SocketServer::*)(), ix::SocketServer*> >::operator() (this=0x5910b5f6de58)
    at /usr/include/c++/13/bits/std_thread.h:299
#6  0x000059109caf679a in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (ix::SocketServer::*)(), ix::SocketServer*> > >::_M_run (
    this=0x5910b5f6de50) at /usr/include/c++/13/bits/std_thread.h:244
#7  0x00007446772ecdb4 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x0000744676e9caa4 in start_thread (arg=<optimised out>) at ./nptl/pthread_create.c:447
#9  0x0000744676f29c3c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
(gdb) 
```

Stepping through the code shows it gets stuck in the for loop here: https://github.com/zeek/IXWebSocket/blob/1babe1d8e13573c073fa739fd2597dd0a6828b6b/ixwebsocket/IXSocketServer.cpp#L440-L463

If I leave zeek running instead, then it disconnects the client after ~20 seconds and (apparently) falling somewhat behind on processing events:

```
$ go build && ./repro-zeek-broker-ws-disconnect 
2025/05/02 12:21:12 connected to broker endpoint=c2e9fc30-c72e-5181-a6b9-dfa1a10ffcb1-websocket-0 version=8.0.0-dev.15-debug
2025/05/02 12:21:31 exited with error=failed to write msg: failed to write frame: failed to flush: write tcp 127.0.0.1:54078->127.0.0.1:16666: write: connection reset by peer and counter=119974
```

```
$ time zeek test-cluster.zeek 
warning in ./test-cluster.zeek, line 16: handler for non-existing event cannot be invoked (Test::evt)
peer added, endpoint=[id=c2e9fc30-c72e-5181-a6b9-dfa1a10ffcb1-websocket-0, network=[address=127.0.0.1, bound_port=54078/tcp]], subscriptions=[]
peer lost at cnt=97589, endpoint=[id=c2e9fc30-c72e-5181-a6b9-dfa1a10ffcb1-websocket-0, network=[address=127.0.0.1, bound_port=54078/tcp]]
received termination signal
Last count was, 97589

real	0m22.309s
user	0m36.324s
sys	0m1.249s
```

Here are the only two logs written by zeek (no indication about the cause of the disconnection):
```
$ cat reporter.log cluster.log
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	reporter
#open	2025-05-02-12-21-31
#fields	ts	level	message	location
#types	time	enum	string	string
1746152491.363630	Reporter::INFO	received termination signal	(empty)
#close	2025-05-02-12-21-31
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	cluster
#open	2025-05-02-12-21-12
#fields	ts	node	message
#types	time	string	string
1746152472.741933	(empty)	WebSocket client 'c2e9fc30-c72e-5181-a6b9-dfa1a10ffcb1-websocket-0' (127.0.0.1:54078) subscribed to []
1746152491.363630	(empty)	WebSocket client 'c2e9fc30-c72e-5181-a6b9-dfa1a10ffcb1-websocket-0' (127.0.0.1:54078) gone
#close	2025-05-02-12-21-31
```

awelzel:
Terminate() is called when Zeek shuts down. If WebSocket client threads
were blocked in QueueForProcessing() due to reaching queue limits, these
previously would not exit QueueForProcessing() and instead block
indefinitely, resulting in the ixwebsocket library blocking and its
garbage collection thread running at 100%. Not great.

Closing the onloop instance will unblock the WebSocket client threads
for a timely shutdown.

Thanks @simeonmiteff !

Closes #4420


testing/btest/cluster/websocket/terminate-while-queuing.zeek
```diff
@@ -0,0 +1,94 @@
...
+global ping: event(msg: string, c: count) &is_used;
+
+global clients: set[string] = set();
+
+event ping(client: string, n: count) &is_used
+	{
+        ++ping_count;
+        add clients[client];
+
+        if ( ping_count == 1000 )
```
bbannier, line 37: Is this `Cluster::Backend::ZeroMQ::log_rcvhwm` so we could get rid of the magic var? An `assert` here would make it clear for the Python code below where the value comes from.

testing/btest/cluster/websocket/terminate-while-queuing.zeek
```diff
@@ -0,0 +1,94 @@
...
+            for idx, tc in enumerate(clients, 1):
+                try:
+                    i += 1
+                    tc.send_json(wstest.build_event_v1("/test/pings/", "ping", [f"tc{idx}", i]))
+                except websockets.exceptions.ConnectionClosedOK as e:
+                    stop = True
+                    print(e)
+                    break
+
+        print("stopped - sent 1000 or more?", i >= 1000)
```
bbannier, line 90: No need to baseline this, using a straight assert could even surface the seen value.

```suggestion
        assert i >= 1000, f"expected to send at least 1000 messages, only sent {i}"
```

testing/btest/cluster/websocket/terminate-while-queuing.zeek
```diff
@@ -0,0 +1,94 @@
...
+# @TEST-EXEC: btest-bg-run client "python3 ../client.py >out"
+#
+# @TEST-EXEC: btest-bg-wait 30
+# @TEST-EXEC: sort ./manager/out > ./manager/out.sorted
+# @TEST-EXEC: btest-diff ./manager/out.sorted
+# @TEST-EXEC: btest-diff ./manager/.stderr
+# @TEST-EXEC: btest-diff ./client/out
+# @TEST-EXEC: btest-diff ./client/.stderr
+
+# # @TEST-START-FILE manager.zeek
```
bbannier, line 21: Nit: let's always put BTest meta into simple comments, also below. 🤡

```suggestion
# @TEST-START-FILE manager.zeek
```

testing/btest/cluster/websocket/terminate-while-queuing.zeek
```diff
@@ -0,0 +1,94 @@
...
+        i = 0
+
+        while not stop and i < 10000:
+            for idx, tc in enumerate(clients, 1):
+                try:
+                    i += 1
+                    tc.send_json(wstest.build_event_v1("/test/pings/", "ping", [f"tc{idx}", i]))
+                except websockets.exceptions.ConnectionClosedOK as e:
+                    stop = True
+                    print(e)
```
bbannier, line 87: I initially confused the 1001 from this with the client being able to get 1000+1 messages in, but it is in fact a protocol code. Let's instead assert here where we can add some additional context.

```suggestion
                    assert e.code == 1001  # Remote going away.
```

bbannier: Actually we should add something which checks that we see this for every client. I think I'd still prefer that to baselining.

testing/btest/cluster/websocket/terminate-while-queuing.zeek
awelzel: It's an arbitrary value that's large enough so the test takes some time and the clients can build up queued events. I'll leave as is.

testing/btest/cluster/websocket/terminate-while-queuing.zeek
awelzel: Fair.

Issue #4422, opened 2025-05-02
7.2.0-rc1 potential regression: HTTP bad request reply edge case


netantho:
[http_bad_request.pcap.zip](https://github.com/user-attachments/files/20018207/http_bad_request.pcap.zip)

Between Zeek 7.1.0 and 7.2.0-rc1, I found an edge case for the [changes to protocol identification](https://zeek.org/wp-content/uploads/2025/03/MUC-2025-Protocol-Identification.pdf). The pcap above is extracted from https://www.malware-traffic-analysis.net/2025/04/13/index.html. It's taken from a webserver.

We see some traffic going TCP port 80 and for which the web server gives a 400 Bad Request error.

In Zeek 7.1.0, it used to produce the following record in `http.log`:
`1743626093.627747     CzpZWzmiL9Ujkb284       92.255.57.58    48110   203.161.44.208  80      1       -       -       -       -       1.1     -       -       0       324     400     Bad Request     -       -       (empty) -       -       -       -       -       -       FkMmsG1NRSFU4CE9If      -       text/html`

And the following record in `files.log`:
`1743626093.627747     FkMmsG1NRSFU4CE9If      CzpZWzmiL9Ujkb284       92.255.57.58    48110   203.161.44.208  80      HTTP    0       (empty) text/html       -       0.000000        F       F       324     324     0       0       F       -       -       -       -       -       -       -`

In Zeek 7.2.0 rc1, these lines are not produced. On one hand it's indeed not a valid HTTP request, on the other hand the HTTP server replies and this reply may be interesting for Zeek users. I think I would vote towards restoring the old behavior, but definitely a hard edge case.

ckreibich:
Thanks @netantho! This will take us back to the cluster of discussions around mixed-protocol flows in #4200, #4251 — in particular on the observations [here](https://github.com/zeek/zeek/issues/4251#issuecomment-2697597031).

I noticed a few things while trying out our pcap. You're sending TLS content (at least it looks like it to Zeek) to an HTTP server, so there's also the question of how Zeek addresses the client's content. I believe what you're flagging here is essentially the same as @initconf's comment [here](https://github.com/zeek/zeek/issues/4251#issuecomment-2698500033), but where he saw HTTP requests to non-HTTP servers, you're seeing a non-HTTP to an HTTP server. The common thread is that you both value the server's actual protocol over the client's.

Here's what I'm seeing when running variations:

[...]

PR #4430, opened 2025-05-06
Fix policy/protocols/conn/failed-service-logging.zeek

0xxon:
In GH-4422 it was pointed out that the protocols/conn/failed-service-logging.zeek policy script only works when
`DPD::track_removed_services_in_connection=T` is set.

This was caused by a logic error in the script. This commit fixes this logic error and introduces an additional test that checks that failed-service-logging works even when the option is not set to true.


0xxon:
Please note that the whole logic flow in this will change, again, if we merge #4362.

#4362 considerably streamlines logging for protocol violations, making a lot of the code redundancies that exist here superfluous. #4362 also, already, does not exhibit the bug that is fixed by this PR.

awelzel (approved):

PR #4436, opened 2025-05-07
wstest/terminate-while-queueing: Patch close_socket()

awelzel:
I believe there's a bug/usage issue in the websockets library where during send() EOF is detected and stored, but the receiving side is then discarding the last received data. Avoid the bug by replacing the close_socket() implementation of the websockets library just for that test and leave detecting the EOF condition to the receiving thread only.


bbannier (approved):
LGTM and makes sense, but I haven't checked the guts of the `websockets` implementation. Thanks for adding the original source, this should help keeping this alive in the future.

If this fixes our issue we should probably file an upstream issue, could you keep this on your radar?

PR #4439, opened 2025-05-08
Bugfix: accurately track Broker buffer overflows w/ multiple peerings

ckreibich:
When a node restarts or a peering between two nodes starts over for other reasons, the internal tracking in the Broker manager resets its state (since it's per-peering), and thus the message overflow counter. The script layer was unaware of this, and threw errors when trying to reset the corresponding counter metric down to zero at sync time.

We now track past buffer overflows via a separate epoch table, using Broker peer ID comparisons to identify new peerings, and set the counter to the sum of past and current overflows.

I considered just making this a gauge, but it seems more helpful to be able to look at a counter to see whether any messages have ever been dropped over the lifetime of the node process.

As an aside, this now also avoids repeatedly creating the labels vector, re-using the same one for each metric.

Thanks to @pbcullen for identifying this one!

### DIFF
--- a/scripts/base/frameworks/cluster/broker-telemetry.zeek
+++ b/scripts/base/frameworks/cluster/broker-telemetry.zeek
@@ -44,26 +44,61 @@ global broker_peer_buffer_overflows_cf = Telemetry::register_counter_family([
     $help_text="Number of overflows in Broker's send buffers",
 ]);
 
+
+# A helper to track overflow counts over past peerings as well as the current
+# one.  The peer_id field allows us to identify when the counter has reset: a
+# Broker ID different from the one on file means it's a new peering.
+type EpochData: record {
+	peer_id: string;
+	num_overflows: count &default=0;
+	num_past_overflows: count &default=0;
+};
+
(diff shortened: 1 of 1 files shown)


timwoj:
> We now track past buffer overflows via a separate epoch table, using Broker peer ID comparisons to identify new peerings, and set the counter to the sum of past and current overflows.

Would it be useful to have both of these? The historical number of overflows plus the number of overflows for the current peering?

ckreibich:
Interesting thought — I'm trying to think of when that would be helpful? I imagine that people will usually just look at this number and go "have we lost any messages at all?", and when the answer is "yes", they need some historical perspective to make sense of it (since it'll likely have been bursts at some point in the past). I'm also a bit reluctant regarding having both because the iterative peerings aren't really an intentional event (you don't apply a config change or some such, they happen because of something weird — OOMs etc).

A counter for the current peering would be a gauge since it'll drop to zero with every new peering. For sure it seems that _only_ having that gauge would be bad because you lose the historical perspective.

ckreibich:
@timwoj I cherry-picked this over into `release/7.2`.

PR #4444, opened 2025-05-08
Downgrade broker clone FatalError to an Error

timwoj:
This PR downgrades the change in https://github.com/zeek/zeek/pull/4226 from a `FatalError` to just an `Error`. @initconf found during testing that having the workers exit was considerably more disruptive than having them just report an error. We can revisit why this error happens later.

This change needs to be pulled into the 7.2 release branch.


ckreibich (approved):
Thanks!

PR #4449, opened 2025-05-09
Adjust for renamed function flavor in Spicy development version

bbannier:
With zeek/spicy#2048 the flavor for functions was renamed from
`Standard` to `Function`. This patch adapts the code for that while
still allowing using earlier Spicy versions.

### DIFF
--- a/src/spicy/spicyz/glue-compiler.cc
+++ b/src/spicy/spicyz/glue-compiler.cc
@@ -1120,11 +1120,15 @@ bool GlueCompiler::compile() {
         preinit_body.addCall("zeek_rt::register_spicy_module_end", {});
 
     if ( ! preinit_body.empty() ) {
+#if SPICY_VERSION_NUMBER >= 11400
+        constexpr auto zeek_preinit_flavor = hilti::type::function::Flavor::Function;
+#else
+        constexpr auto zeek_preinit_flavor = hilti::type::function::Flavor::Standard;
+#endif
         auto preinit_function =
             builder()->function(hilti::ID("zeek_preinit"),
                                 builder()->qualifiedType(builder()->typeVoid(), hilti::Constness::Const), {},
-                                preinit_body.block(), hilti::type::function::Flavor::Standard,
-                                hilti::declaration::Linkage::PreInit);
+                                preinit_body.block(), zeek_preinit_flavor, hilti::declaration::Linkage::PreInit);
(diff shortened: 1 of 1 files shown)


src/spicy/spicyz/glue-compiler.cc
```diff
@@ -38,12 +38,8 @@ static std::string::size_type looking_at(const std::string& chunk, std::string::
                                          const std::string_view& token) {
     eat_spaces(chunk, &i);
 
-    for ( char j : token ) {
-        if ( i >= chunk.size() || chunk[i++] != j )
-            return 0;
-    }
-
-    return i;
+    bool token_at_position = token == std::string_view(chunk).substr(i, token.size());
```
evantypanski, line 41: I think there is a behavior change here if `i > chunk.size()` it would now throw an exception I think

Maybe add an assertion? I don't think it'd hit that case in real execution though

evantypanski (approved):
sorry I didn't catch the `Function` change! I thought about checking and forgot

src/spicy/spicyz/glue-compiler.cc
bbannier: Uff, good catch. I added a check that `i` does not exceed the chunk size `i < chunk.size()`. That makes the code slightly dense, but reproduces the previous behavior.

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -343,7 +343,7 @@ hilti::Result<std::string> GlueCompiler::getNextEvtBlock(std::istream& in, int*
     std::string chunk;
 
     // Parser need to track whether we are inside a string or a comment.
-    enum State { Default, InComment, InString } state = Default;
+    enum State : char { Default, InComment, InString } state = Default;
```
awelzel, line 342: Total drive-by, I agree this can be useful if there's lot of enum instances, but here there's exactly one, so this seems just extra noise, but I'm happy to learn if/whether we should do this more often.

awelzel (approved):

src/spicy/spicyz/glue-compiler.cc
bbannier: I came at this from the angle: my editor yells at me whenever I open this file since this lint is enabled in the default clang-tidy config. I personally do not think this is more noisy than a local suppression.

PR #4451, opened 2025-05-09
Fix for ZAM optimization when an aggregate is modified inside of a loop

vpax:
Some of our recent content testing revealed a bug in ZAM's AST optimization. The problem would arise in a sequence where (1) a temporary was used to hold a value computed from an aggregate (like extracting a field from a record), (2) that same expression occurred subsequently, making it a candidate for re-using the temporary, (3) but that appearance was inside a loop (which the original was not), _and_ (4) _later_ in the loop the aggregate was modified, which means on a later iteration through the loop the original temporary's value is stale and shouldn't be reused.

The fix is to extend the validity checking done when reusing expressions so that they run the end of any loop surrounding the second occurrence of the expression.


awelzel (approved):
Approving. I think I grasped the problem, but I don't know the rest of the code well enough. The added test helps and would be good for 7.0 and 7.2.

Issue #4448, opened 2025-05-09
`&ordered` on tables works with table constructors but not with table initializers


ckreibich:
Tagging tables with `&ordered` does not produce reliable ordering when initializing via `{ ... }`, but it works when using `table()`.

```
global tbl1: table[string] of count = {
        ["foo"] = 1,
        ["bar"] = 2,
        ["baz"] = 3,
} &ordered;

global tbl2: table[string] of count = table(
        ["foo"] = 1,
        ["bar"] = 2,
        ["baz"] = 3,
) &ordered;

event zeek_init() {
        for ( key, val in tbl1 )
                print fmt("%s %s", key, val);
        print "";
        for ( key, val in tbl2 )
                print fmt("%s %s", key, val);
}
```
```
$ zeek ./test.zeek
baz 3
bar 2
foo 1

foo 1
bar 2
baz 3
$ zeek ./test.zeek
baz 2
bar 1
foo 3

foo 1
bar 2
baz 3
$ zeek ./test.zeek
bar 2
foo 1
baz 3

foo 1
bar 2
baz 3
```

Issue #4189, opened 2025-01-23
Consider `&ordered` for initialization and redefinitions


J-Gras:
```Zeek
global ips_a: set[addr] = {1.1.1.1, 1.1.1.2} &ordered &redef;
global ips_b: set[addr] &ordered &redef;

redef ips_a += {2.3.4.5};

event zeek_init()
	{
	add ips_b[1.1.1.1];
	add ips_b[1.1.1.2];
	add ips_b[2.3.4.5];
	
	print "ips_a:";
	for ( i in ips_a )
		print i;
	
	print "ips_b:";
	for ( i in ips_b )
		print i;
	}
```
The above spits out the following for me (YMMV for `ips_a`):
```
ips_a:
2.3.4.5
1.1.1.2
1.1.1.1
ips_b:
1.1.1.1
1.1.1.2
2.3.4.5
```
My expectation would be that ordering is also considered on initialization and for redefinitions.

0xxon:
I think in these cases the attribute should propagate and will just define this as a bug :)

PR #4452, opened 2025-05-12
Add btest for SQLite in a cluster environment

timwoj:
This also renames all of the sqlite-based tests to be prefixed with `sqlite-` like the redis ones are.


timwoj:
Welp, this exposed a very rare occurrence that didn't happen at all during my local testing

```
  [code=Storage::INITIALIZATION_FAILED, error_str=Failed to open backend Storage::STORAGE_BACKEND_SQLITE: Error executing tuning pragma statement: database is locked, value=<opaque of BackendHandleVal>]
```

If two nodes try to run the `pragma` commands at the same time while starting up, they may run into locking/contention issues. I'll have to think about how to fix that one.

JustinAzoff:
Are we setting the busy timeout anywhere?

    PRAGMA busy_timeout = 5000;

timwoj:
> Are we setting the busy timeout anywhere?
> 
> ```
> PRAGMA busy_timeout = 5000;
> ```

Not by default currently, but it's easy to add.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -44,27 +46,73 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
+    while ( attempts < 5 ) {
+        int res = sqlite3_exec(db, "pragma integrity_check", NULL, NULL, &errorMsg);
+        if ( res == SQLITE_OK ) {
+            break;
+        }
+        else if ( res == SQLITE_BUSY ) {
+            // If we got back that the database is busy, it likely means that another process is trying to
+            // do their pragmas at startup too. Sleep for a little bit and try again.
+            sqlite3_free(errorMsg);
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
awelzel, line 60: Could we put that loop into a helper? E.g. `bool run_pragma(std::string_view name, std::optional<std::string_view>> arg)` so it's not duplicated here and below?

This allows for a max of 500msec for a busy database. We should make this configurable in case for some reason (large database, many workers, ...) this isn't sufficient. Telling a user they need to recompile to increase the timeout isn't great.

Would drop the "attempts" and do more of a "total timeout is 500msec by default, retrying every 1msec or 10msec" approach, so that it's a bit more responsive/faster, retrying every 100msec feels like a "very long" interval. I'd expect SQLite to be able to figure out the busy/lock state quickly/cheaply enough to be more aggressive here, bit with both intervals configurable (don't think it's needed per database for now, but at least on a module level?)

scripts/policy/frameworks/storage/backend/sqlite/main.zeek
```diff
@@ -24,6 +24,7 @@ export {
 		## database.  These must be pairs that can be passed to the ``pragma``
 		## command in sqlite.
 		tuning_params: table[string] of string &default=table(
```
awelzel, line 26: This is mostly seeing the users, wonder if we should have called this `pragma_commands` and make it a `vector of some Command type` rather than a `table[string] of string` that isn't extensible and where the execution order of the commands isn't defined.

Are we still in a position to fiddle with the API here?

Not a request to change this now unless you agree this moves us into a better position.

awelzel (changes requested):

scripts/policy/frameworks/storage/backend/sqlite/main.zeek
timwoj: Changing to some sort of enum means having to extend it any time that SQLite changes their list of valid pragmas. Leaving it as a simple table of string/string means that power users can look up the pragmas on their own and set their values, and the backend can fail on any that aren't valid. I don't think that a lot of people will change these, to be honest.

I don't think there are any restrictions on ordering of pragmas passed to SQLite either. The only thing I could find from a quick google was a third-party project for encryption that requires it, but nothing in the official docs indicates an ordering requirement. I could add the `&ordered` attribute to it to force the ordering like that.

src/storage/backend/sqlite/SQLite.cc
timwoj: I moved everything into a utility method, add a backend option, and set it to be exponential backoff (starts at 1ms, multiplies by 10 each time).

scripts/policy/frameworks/storage/backend/sqlite/main.zeek
awelzel: Sorry for the confusion - didn't mean to add any enumerations, more like something like this, but feel free to ignore:
```
export {
+       type PragmaCommand: record {
+               name: string;
+               value: string &optional;
+       };
+
        ## Options record for the built-in SQLite backend.
        type Options: record {
                ## Path to the database file on disk. Setting this to ":memory:" will tell
@@ -25,11 +30,8 @@ export {
                ## sqlite. The ``integrity_check`` pragma is run automatically and does
                ## not need to be included here. For pragmas without a second argument,
                ## set the value to an empty string.
-               pragma_commands: table[string] of string &default=table(
-                       ["busy_timeout"] = "5000",
-                       ["journal_mode"] = "WAL",
-                       ["synchronous"] = "normal",
-                       ["temp_store"] = "memory"
+               pragma_commands: vector of PragmaCommand &default=vector(
+                       PragmaCommand($name="integrity_check"),
+                       PragmaCommand($name="busy_command", $value="5000"),
+                       ...
                );
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
+    if ( value && ! value->empty() )
+        cmd += util::fmt(" = %.*s", static_cast<int>(value->size()), value->data());
+
+    while ( pragma_timeout > 0ms && time_spent < pragma_timeout ) {
+        int res = sqlite3_exec(db, cmd.c_str(), NULL, NULL, &errorMsg);
+        if ( res == SQLITE_OK ) {
+            break;
+        }
+        else if ( res == SQLITE_BUSY ) {
+            printf("busy?\n");
```
awelzel, line 32:
```suggestion
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
 #include "zeek/Func.h"
 #include "zeek/Val.h"
 #include "zeek/storage/ReturnCode.h"
 
+using namespace std::chrono_literals;
+
 namespace zeek::storage::backend::sqlite {
 
+OperationResult SQLite::RunPragma(std::string_view name, std::optional<std::string_view> value) {
+    int attempts = 0;
```
awelzel, line 17:
```suggestion
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
+            break;
+        }
+        else if ( res == SQLITE_BUSY ) {
+            printf("busy?\n");
+            // If we got back that the database is busy, it likely means that another process is trying to
+            // do their pragmas at startup too. Exponentially back off and try again after a sleep.
+            sqlite3_free(errorMsg);
+            std::this_thread::sleep_for(backoff);
+            time_spent += backoff;
+            backoff *= 10;
```
awelzel, line 38: Wouldn't bother doing the backoff. It'll try some 500 times with the default 500ms - seems okay to me. With the backoff it's only 4 or 5 times (1, 10, 100, 1000) and all nodes retrying at similar times, potentially reducing chances to actually succeed. There's also no try after the last sleep, so might block for a long time only to return an error (less of an issue with small sleeps).

Maybe this is all theoretical anyhow, but then the exponential backoff doesn't matter either and could just retry every 1ms. Still would make the 1ms configurable.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
+    int attempts = 0;
+    char* errorMsg = nullptr;
+    std::chrono::milliseconds time_spent = 0ms;
+    std::chrono::milliseconds backoff = 1ms;
+
+    std::string cmd = util::fmt("pragma %.*s", static_cast<int>(name.size()), name.data());
+    if ( value && ! value->empty() )
+        cmd += util::fmt(" = %.*s", static_cast<int>(value->size()), value->data());
+
+    while ( pragma_timeout > 0ms && time_spent < pragma_timeout ) {
```
awelzel, line 26:
```suggestion
    while ( pragma_timeout == 0ms || time_spent < pragma_timeout ) {
```

If pragma_timeout is actually `0` this won't run at all which isn't what the comment in the script layer says :-/ Mind adding a smoke test for this case?

src/storage/backend/sqlite/SQLite.cc
awelzel: There's also a can of worms that if this is an async open, it would be nice to yield instead of block, but I don't think we have an "easy" primitive to do so unless I'm missing something (no need to fix).

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
+            std::string err =
+                util::fmt("Error while executing %.*s pragma", static_cast<int>(name.size()), name.data());
+            sqlite3_free(errorMsg);
+            return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
+        }
+    }
+
+    if ( time_spent >= pragma_timeout ) {
+        std::string err =
+            util::fmt("Database was busy while executing %.*s pragma", static_cast<int>(name.size()), name.data());
```
awelzel, line 50:
```suggestion
            util::fmt("Database was busy while executing pragma '%s'", cmd.c_str());
```

Would that be okay, also above - it might have interesting information.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -43,41 +91,39 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
-        return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
-    }
 
-    if ( int res = sqlite3_exec(db, "pragma integrity_check", NULL, NULL, &errorMsg); res != SQLITE_OK ) {
-        std::string err = util::fmt("Error executing integrity check: %s", errorMsg);
-        Error(err.c_str());
-        sqlite3_free(errorMsg);
-        Close(nullptr);
-        return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
+    OperationResult pragma_res = RunPragma("integrity_check");
```
awelzel, line 96: One nice thing of the `pragma_commands` vector would be that `integrity_check` could be the first element, allowing users to skip it, too, without needing an extra option.

testing/btest/scripts/base/frameworks/storage/sqlite-cluster.zeek
```diff
@@ -0,0 +1,87 @@
...
+# @TEST-EXEC:
+
+@load base/frameworks/storage/sync
+@load base/frameworks/cluster
+@load policy/frameworks/storage/backend/sqlite
+@load policy/frameworks/cluster/experimental
+
+# @TEST-START-FILE cluster-layout.zeek
+redef Cluster::nodes = {
+	["manager-1"] = [$node_type=Cluster::MANAGER, $ip=127.0.0.1, $p=to_port(getenv("BROKER_PORT1"))],
```
awelzel, line 22: Mind making this manager rather than manager-1 to avoid churn if #4472  makes it?

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,58 @@
...
+            // If we got back that the database is busy, it likely means that another process is trying to
+            // do their pragmas at startup too. Exponentially back off and try again after a sleep.
+            sqlite3_free(errorMsg);
+            std::this_thread::sleep_for(backoff);
+            time_spent += backoff;
+            backoff *= 10;
+        }
+        else {
+            std::string err =
+                util::fmt("Error while executing %.*s pragma", static_cast<int>(name.size()), name.data());
```
awelzel, line 42: Mind including the `res` value and its stringified version from `sqlite3_errstr` into the error string - probably that's useful if there's ever something wrong (and/or the user used a wrong pragma/value in the table).

awelzel (commented):
Don't think it's quite there yet, but close! :-)

testing/btest/scripts/base/frameworks/storage/sqlite-cluster.zeek
timwoj: It's approved now it you want to merge it, and I'll adapt this test to that.

src/storage/backend/sqlite/SQLite.cc
timwoj: Sounds reasonable.

scripts/policy/frameworks/storage/backend/sqlite/main.zeek
timwoj: I'm going to leave it as-is for now. I feel like adding a record to basically store what's there already doesn't add anything usage-wise.

> Could also generalize even more if we wanted to allow users to run raw SQL after opening the database, but maybe we shouldn't.

This can definitely be something we added separately, but not part of this PR.

src/storage/backend/sqlite/SQLite.cc
timwoj: I added a `pragma_wait_on_failure` option and defaulted it to 5ms. 1ms even felt a little short. 500 runs of the btest though and none of them failed.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -2,13 +2,56 @@
...
+            time_spent += pragma_wait_on_failure;
+        }
+        else {
+            std::string err = util::fmt("Error while executing pragma '%s': %s", cmd.c_str(), errorMsg);
+            sqlite3_free(errorMsg);
+            return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
+        }
+    }
+
+    if ( time_spent >= pragma_timeout ) {
```
awelzel, line 46:
```suggestion
    if ( pragma_timeout != 0 && time_spent >= pragma_timeout ) {
```

A cluster test with timeout set to 0.0sec would probably have triggered the error once in a while.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -34,6 +77,12 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
     full_path = zeek::filesystem::path(path->ToStdString()).string();
     table_name = backend_options->GetField<StringVal>("table_name")->ToStdString();
 
+    auto pragma_timeout_val = backend_options->GetField<IntervalVal>("pragma_timeout");
+    pragma_timeout = std::chrono::milliseconds(static_cast<int64_t>(pragma_timeout_val->Get() * 1000));
+
+    auto pragma_wof_val = backend_options->GetField<IntervalVal>("pragma_wait_on_failure");
```
awelzel, line 83: Nit, take as you see fit.

```suggestion
    auto pragma_wof_val = backend_options->GetField<IntervalVal>("pragma_wait_on_busy");
```

scripts/policy/frameworks/storage/backend/sqlite/main.zeek
```diff
@@ -20,14 +20,28 @@ export {
...
+		## sqlite. The ``integrity_check`` pragma is run automatically and does
+		## not need to be included here. For pragmas without a second argument,
+		## set the value to an empty string.
+		pragma_commands: table[string] of string &ordered &default=table(
+			["integrity_check"] = "",
+			["busy_timeout"] = "5000",
 			["journal_mode"] = "WAL",
 			["synchronous"] = "normal",
 			["temp_store"] = "memory"
 		);
```
awelzel, line 34:
```suggestion
		) &ordered;
```


Without,  adding an option to a default constructed `Options` record wouldn't put an additional option at the end. E.g, test with:
```
$ zeek frameworks/storage/backend/sqlite -e 'local o = Storage::Backend::SQLite::Options($database_path="x", $table_name="y"); o$pragma_commands["XXX_last_pragma"] = "XXX_last_pragma"; print o'
[database_path=x, table_name=y, pragma_commands={
[integrity_check] = ,
[XXX_last_pragma] = XXX_last_pragma,
[temp_store] = memory,
[synchronous] = normal,
[journal_mode] = WAL,
[busy_timeout] = 5000
}, pragma_timeout=500.0 msecs, pragma_wait_on_failure=5.0 msecs]
```

That's probably due to #4448 or #4189 and hopefully already on @evantypanski's list of Zeek script gotchas :-)

awelzel (approved):
Thanks - a few more nits, but good to :ship:

PR #4453, opened 2025-05-12
Allow macOS to search for libkrb5, but disallow system version of library

timwoj:
This brings in the changes from https://github.com/zeek/cmake/pull/131.

Fixes #4360

PR #4456, opened 2025-05-13
Prefer libkrb5 from Homebrew over macOS system version

timwoj:
https://github.com/zeek/zeek/pull/4453 added the ability to search for krb5 on macOS, but by default it's always going to find the system library and fail. This changes it to force the user to pass --with-krb5 on macOS, but allow normal searching on other platforms.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,9 +38,10 @@ Changed Functionality
   published vectors with holes. A reporter error is produced at runtime when
   serialization of vectors with holes is attempted.
 
-- Using libkrb5 for Kerberos support is now allowed on macOS. Due to very old age and
-  brokenness, the system version of the library is unsupported and will fail to
-  configure. Use the version from Homebrew or another newer installation.
+- Kerberos support on macOS has been enabled. Due to incompatibilities, the system
+  provided libkrb5 is ignored, however. Only versions from homebrew are supported and
+  found/picked-up by default. Use --with-krb5 for pointing at a custom librkb5
+  installation.
 
 Removed Functionality
 ---------------------


J-Gras:
> Would it be controversial to patch `FindLibKrb5.zeek` with hints to pick up the version from homebrew by default for macOS and warn and ignore if only the system library was found.
> 
> The following just doesn't sound awesome:
> 
> > This changes it to force the user to pass --with-krb5 on macOS

Very much agree to @awelzel!

NEWS
```diff
@@ -38,9 +38,11 @@ Changed Functionality
...
   serialization of vectors with holes is attempted.
 
-- Using libkrb5 for Kerberos support is now allowed on macOS. Due to very old age and
-  brokenness, the system version of the library is unsupported and will fail to
-  configure. Use the version from Homebrew or another newer installation.
+- Using libkrb5 for Kerberos support is now allowed on macOS. The default system library
+  previded by macOS is a very old version and may cause Zeek to crash. ``configure`` will
+  attempt to find the version installed by Homebrew first, but will report a warning and
+  disable Kerberos support if the system library is found. The ``--with-krb5`` option for
+  ``configure`` can be used To override using the version from Hombrew, if desired.
```
awelzel, line 45: Thanks for the rework!

There's a few typos and I wonder if it could be shortened:

> Kerberos support on macOS has been enabled. Due to incompatibilities, the system provided libkrb5 is ignored, however. Only versions from homebrew are supported and found/picked-up by default. Use ``--with-krb5`` for pointing at a custom librkb5 installation.

awelzel (approved):

PR #4458, opened 2025-05-14
Add STLS command to POP3 DPD signature

J-Gras:
Looking at POP3 STARTTLS btest, it seemed that POP3 STARTTLS connections can be easily caught by adding `STLS` to the POP3 DPD signature.


J-Gras:
> I like this.

🥳 

> Do you randomly have a test-case that uses thos? :)

The test that I changed in this PR now uses the DPD signature instead of relying on adding port 110 :)

0xxon:
Oh, right, I am blind...

0xxon (approved):

PR #4460, opened 2025-05-15
Pull in `cmake/` change for isolated plugin dependencies

bbannier:
This is a companion change for zeek/cmake#134.


awelzel (approved):

timwoj:
I pulled this in separately.

PR #4469, opened 2025-05-16
`#include` Cleanup Part 2

timwoj:
This is the second of two PRs (started in https://github.com/zeek/zeek/pull/4434) to clean up a ton of #include usage across the entire codebase. This one covers everything where a single change didn't spread to a bunch of other files.

### DIFF
--- a/src/analyzer/protocol/krb/KRB.h
+++ b/src/analyzer/protocol/krb/KRB.h
@@ -2,15 +2,15 @@
 
 #pragma once
 
+// This is needed for USE_KRB5 below.
 #include "zeek/zeek-config.h"
-
-#include <mutex>
 
 #ifdef USE_KRB5
 #include <krb5/krb5.h>
 #endif
 
-#include "analyzer/protocol/krb/krb_pac.h"
+#include "zeek/analyzer/Analyzer.h"
+#include "zeek/analyzer/protocol/krb/krb_pac.h"
 
 namespace zeek::analyzer::krb {
 
--- a/src/analyzer/protocol/rpc/RPC.h
+++ b/src/analyzer/protocol/rpc/RPC.h
@@ -2,7 +2,6 @@
 
 #pragma once
 
-#include "zeek/NetVar.h"
 #include "zeek/analyzer/protocol/tcp/TCP.h"
 
 namespace zeek::analyzer::rpc {
--- a/src/analyzer/protocol/snmp/SNMP.cc
+++ b/src/analyzer/protocol/snmp/SNMP.cc
@@ -2,11 +2,6 @@
 
 #include "zeek/analyzer/protocol/snmp/SNMP.h"
 
-#include "zeek/Func.h"
-#include "zeek/Reporter.h"
-#include "zeek/analyzer/protocol/snmp/events.bif.h"
-#include "zeek/analyzer/protocol/snmp/types.bif.h"
-
 namespace zeek::analyzer::snmp {
 
 SNMP_Analyzer::SNMP_Analyzer(Connection* conn) : Analyzer("SNMP", conn) { interp = new binpac::SNMP::SNMP_Conn(this); }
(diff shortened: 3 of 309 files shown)


src/analyzer/protocol/krb/KRB.h
```diff
@@ -2,15 +2,15 @@
 
 #pragma once
 
+// This is needed for USE_KRB5 below.
```
awelzel, line 5: Ignore if you don't feel like, but I guess we could actually add a `-DUSE_KRB5` compile flag for the KRB source files, too?

src/analyzer/protocol/rpc/RPC.h
```diff
@@ -2,7 +2,7 @@
 
 #pragma once
 
-#include "zeek/NetVar.h"
+// #include "zeek/NetVar.h"
```
awelzel, line 5: ?

src/analyzer/protocol/snmp/SNMP.cc
```diff
@@ -2,10 +2,7 @@
 
 #include "zeek/analyzer/protocol/snmp/SNMP.h"
 
-#include "zeek/Func.h"
 #include "zeek/Reporter.h"
```
awelzel, line 5: Just because it stands out: Why is Reporter.h needed here?

src/analyzer/protocol/ssl/Plugin.cc
```diff
@@ -2,8 +2,6 @@
 
 #include "zeek/plugin/Plugin.h"
 
-#include "zeek/zeek-config.h"
-
 #ifndef ENABLE_SPICY_SSL
```
awelzel, line 5: Isn't that in zeek-config.h ?

src/analyzer/protocol/snmp/SNMP.cc
timwoj: That's a darn good question. It gets included by the `snmp_pac.h` via `SNMP.h` too. There's so much about what gets included by the pac files it's crazy.

src/analyzer/protocol/ssl/Plugin.cc
awelzel: That's fair. So seeing `zeek/plugin/Plugin.h` one can assume the `zeek/zeek-config.h` include? It's a bit surprising, but don't have a strong reaction.

timwoj: Yeah, it looks like this:

```
. /Users/tim/Desktop/projects/zeek/src/include/zeek/plugin/Plugin.h
.. /Users/tim/Desktop/projects/zeek/src/include/zeek/ZeekArgs.h
... /Users/tim/Desktop/projects/zeek/src/include/zeek/ZeekList.h
.... /Users/tim/Desktop/projects/zeek/src/include/zeek/List.h
..... /Users/tim/Desktop/projects/zeek/src/include/zeek/util.h
...... /Users/tim/Desktop/projects/zeek/build/zeek/zeek-config.h
```

timwoj: In case you're wondering, `List.h` brings in `util.h` because of `util::safe_malloc`.

bbannier: I am not sure I like depending on that magic include chain to remain stable, especially for a preprocessor macro. My hunch would have also been to keep the `#include "zeek/zeek-config.h"` for `ENABLE_SPICY_SSL`.

awelzel: Oh, I thought `plugin/Plugin.h` directly includes `zeek-config.h`, but hadn't checked. If it's like those 5 layers deep, I'm actually with @bbannier that we should have it explicitly. Or maybe also here do `-DENABLE_SPICY_SSL` on the source file explicitly.

That `-D` approach might be a discussion elsewhere - I don't have an intuition if that's a good pattern or worse, but it feels like a more localized approach.

PR #4472, opened 2025-05-18
Unify broker cluster-layout usage in btests

awelzel:
After a conversation with ChatGPT about a Python script to do this, this PR does the following:

* Use a new generic $FILES/broker/cluster-layout.zeek in most but a few tests
* Replace all "manger-1" occurrences with "manager"
* Replace all logger, proxy or worker occurrences with "logger-1" and "proxy-1" or "worker-1" when a cluster-layout is involved.
* All `TEST-PORT BROKER_PORTX` are replaced with lines that encode the node name when a cluster-layout is involved.

The script wasn't perfect, but helped enough to get started.

Is this worth it? I think so. We usually copy&paste cluster tests around and it'd be good to be a bit more consistent for further automatic changs/fixes.


testing/btest/Files/broker/cluster-layout.zeek
```diff
@@ -0,0 +1,92 @@
...
+@endif
+
+@if ( getenv("BROKER_LOGGER2_PORT") != "" )
+redef Cluster::manager_is_logger = F;
+redef Cluster::nodes += {
+	["logger-2"] = [$node_type=Cluster::LOGGER, $ip=127.0.0.1, $p=to_port(getenv("BROKER_LOGGER2_PORT")), $manager="manager"],
+};
+@endif
+
+@if ( getenv("BROKER_LOGGER10_PORT") != "" )
```
timwoj, line 49: Why `LOGGER10`? Also, this one isn't mentioned in the block in the header comment.

testing/btest/scripts/base/frameworks/config/cluster_resend.zeek
```diff
@@ -1,31 +1,25 @@
...
-# @TEST-EXEC: btest-bg-run worker-2  ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=worker-2 zeek -b %INPUT
-# @TEST-EXEC: btest-bg-wait 60
-# @TEST-EXEC: btest-diff manager-1/.stdout
+# @TEST-EXEC: cp $FILES/broker/cluster-layout.zeek .
+#
+# @TEST-EXEC: btest-bg-run manager  ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=manager  zeek -b %INPUT
+# @TEST-EXEC: btest-bg-run worker-1 ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=worker-1 zeek -b %INPUT
+# @TEST-EXEC: $SCRIPTS/wait-for-file manager/ready 30 || (btest-bg-wait -k 1 && false)
+# @TEST-EXEC: btest-bg-run worker-2 ZEEKPATH=$ZEEKPATH:.. CLUSTER_NODE=worker-2 zeek -b %INPUT
+# @TEST-EXEC: btest-bg-wait 10
```
timwoj, line 11: This changed from 60 to 10, but assuming it still works that's probably fine. Just noting it since it doesn't look like any of the other waits changed anywhere.

testing/btest/coverage/test-all-policy-cluster.test
```diff
@@ -41,6 +33,8 @@ redef Log::default_rotation_interval = 0sec;
 
 event zeek_init()
 	{
+	# Prevent connections to the manager defined in Cluster::nodes.
+	delete Cluster::nodes[Cluster::node]$manager;
```
timwoj, line 37: Why was this not done with the original `manager` node?

timwoj (commented):
I like the approach. It'll definitely be nice to not have copy around a cluster-layout.zeek everywhere. I just had some questions.

testing/btest/scripts/base/frameworks/config/cluster_resend.zeek
awelzel: Thanks! I reduced some of the waits while figuring out failures, didn't mean to actually change them.

testing/btest/coverage/test-all-policy-cluster.test
awelzel: The original cluster-layout encoded in the file didn't have the `$manager` field on any of the other nodes, so they didn't attempt to establish a connection. The template does have `$manager` set, but we still don't want the outgoing connection.

testing/btest/Files/broker/cluster-layout.zeek
awelzel: The `cluster/log_distribution.zeek` uses logger-10. If changed that test to add the entry explicitly to cluster-layout.zeek and remove it from the template. That seems more reasonable. Thanks!

timwoj (approved):
:shipit:

PR #4484, opened 2025-05-21
Custom event metadata - part 1

awelzel:
Decided to split the feature up in two separate PRs because it's getting a bit large.

This first part primarily concerns the non-cluster side:

* Replace `ts` in `zeek::Event` with a vector of generic metadata
* Deprecate Enqueue() and Event constructor that take a `ts` parameter. That's not super easy due to all the default parameters, so made the default timestamp -1.0 and added a warn once for Enqueue(). I'd be okay removing the parameter in 8.1, I'd assume that's only used internally anyhow.
* Make network timestamp metadata opt-in - see the NEWS entry for motivation
* Put script-layer infrastructure in place to register and query arbitrary event metadata

Micro-benchmarking event queueing (just recursively queueing events) shows the vector and timestamp allocation is fairly visible if `EventMetadata::add_network_timestamp=T`, but when reading from a PCAP and doing "actual work", it's barely measurable. The default is `F`.


src/Event.h
```diff
@@ -121,6 +154,22 @@ class EventMgr final : public Obj, public iosource::IOSource {
      */
     const Event* CurrentEvent() { return current; }
 
+    /**
+     * Register a script-layer metadata identifier *id* with type *type*.
+     *
+     * @param id The script-level ``EventMetadata::ID`` enum value.
+     * @param type The type to expect for the given metadata identifier.
+     */
+    bool RegisterMetadata(EnumValPtr id, TypePtr type);
```
awelzel, line 234: Argh, I think the better place for the metadata type tracking is in `EventRegistry`. Will move there after the first round of review.

src/Event.h
```diff
@@ -124,6 +195,26 @@ class EventMgr final : public Obj, public iosource::IOSource {
...
+        DeprecatedTimestamp() : d(-1.0) {}
+        [[deprecated("Use overload Enqueue(EventMetadataVectorPtr meta, ...) to pass timestamp metadata")]]
+        DeprecatedTimestamp(double d) /* implicit */
+            : d(d) {}
+
+        explicit operator double() const { return d; }
+
+    private:
+        double d;
+    };
```
bbannier, line 216: Nit: Let's put `implicit` where one could put `explicit`. Also IMO no need to call out historical context (but thanks!).

```suggestion
    /**
     * Helper class to produce a compile time warning if Enqueue() is called with an explicit timestamp.
     *
     * Remove in v8.1.
     */
    class DeprecatedTimestamp {
    public:
        DeprecatedTimestamp() : d(-1.0) {}
        [[deprecated("Use overload EventMgr::Enqueue(EventMetadataVectorPtr meta, ...) to pass timestamp metadata")]]
        /*implicit*/ DeprecatedTimestamp(double d)
            : d(d) {}

        explicit operator double() const { return d; }

    private:
        double d;
    };
```

src/EventRegistry.cc
```diff
@@ -164,4 +165,45 @@ void EventGroup::Disable() {
...
 
+bool EventRegistry::RegisterMetadata(EnumValPtr id, TypePtr type) {
+    static const auto& metadata_id_type = id::find_type<EnumType>("EventMetadata::ID");
+
+    if ( metadata_id_type != id->GetType() )
+        return false;
+
+    auto id_int = id->Get();
+    if ( id_int < 0 ) {
+        zeek::reporter->InternalError("Negative enum value %s: %" PRId64, obj_desc_short(id.get()).c_str(), id_int);
```
bbannier, line 176: `PRId64` needs `#include <cinttypes>` and fails to compile without it on my machine.

src/EventRegistry.cc
```diff
@@ -164,4 +165,45 @@ void EventGroup::Disable() {
...
+    }
+
+    zeek_uint_t id_uint = static_cast<zeek_uint_t>(id_int);
+
+    auto it = event_metadata_types.find(id_uint);
+
+    // If type contains anything with any, like table[count] of any, we should reject it.
+    //
+    //     cb = AnyTypeChecker()
+    //     type->Traverse(cb)
```
bbannier, line 187: Was this comment intended to be a TODO? If so let's mark it as such, else let's drop the whole comment since it doesn't apply.

src/EventRegistry.cc
```diff
@@ -164,4 +165,45 @@ void EventGroup::Disable() {
...
+    if ( it != event_metadata_types.end() )
+        return same_type(it->second.Type(), type);
```
bbannier, line 189: We can avoid putting `it` into the function scope.

I was briefly also wondering about logging of mismatched types here, but we seem to do that elsewhere on another layer already.

```suggestion
    if ( auto it = event_metadata_types.find(id_uint); it != event_metadata_types.end() )

        return same_type(it->second.Type(), type);
```

src/EventRegistry.cc
```diff
@@ -164,4 +165,45 @@ void EventGroup::Disable() {
...
+
+    return true;
+}
+
+const EventMetadataDescriptor* EventRegistry::LookupMetadata(zeek_uint_t id) const {
+    const auto it = event_metadata_types.find(id);
+    if ( it == event_metadata_types.end() )
+        return nullptr;
+
+    if ( it->second.Id() != id ) {
```
bbannier, line 201: This makes me wonder whether we shouldn't store these in some set instead and do lookup with `find_if`. Depending which set we'd use we'd need to implement equality and hash operators for `EventMetadataDescriptor`.

src/EventRegistry.cc
```diff
@@ -164,4 +165,45 @@ void EventGroup::Disable() {
...
+        zeek::reporter->InternalError("inconsistent metadata descriptor: %" PRIu64 " vs %" PRId64, it->second.Id(), id);
+        return nullptr; // unreached
```
bbannier, line 203: `InternalError` is `noreturn` so compilers know this is dead code.

```suggestion
```

src/EventRegistry.h
```diff
@@ -22,18 +22,55 @@ enum class EventGroupKind {
     Module,
 };
 
+class EnumVal;
 class EventGroup;
 class EventHandler;
 class EventHandlerPtr;
 class RE_Matcher;
+class RecordVal;
+class Type;
```
bbannier, line 31: Just as an observation, we seem to declare these same type aliases in multiple places into the same namespace. This seems error-prone and we should probably just declare them once and `#include` that definition (possibly hard since not all headers might have been designed for that).

src/EventRegistry.h
```diff
@@ -22,18 +22,55 @@ enum class EventGroupKind {
...
 using EventGroupPtr = std::shared_ptr<EventGroup>;
+using RecordValPtr = IntrusivePtr<RecordVal>;
+using TypePtr = IntrusivePtr<Type>;
 
 namespace detail {
 class ScriptFunc;
 using ScriptFuncPtr = zeek::IntrusivePtr<ScriptFunc>;
+
+/**
+ * Well known event metadata identifiers.
```
bbannier, line 43:
```suggestion
 * Well-known event metadata identifiers.
```

src/EventRegistry.h
```diff
@@ -22,18 +22,55 @@ enum class EventGroupKind {
...
+ * holds the metadata identifier and registered type. For the identifier,
+ * *id* is the unsigned int representation, while *id_val* holds the
+ * script-layer zeek::EnumVal.
+ */
+class EventMetadataDescriptor {
+public:
+    EventMetadataDescriptor(zeek_uint_t id, EnumValPtr id_val, TypePtr type)
+        : id(id), id_val(std::move(id_val)), type(std::move(type)) {}
+
+    const zeek_uint_t Id() const { return id; }
```
bbannier, line 64: `const` on returned value does nothing.

```suggestion
    zeek_uint_t Id() const { return id; }
```

src/EventRegistry.h
```diff
@@ -106,6 +160,9 @@ class EventRegistry final {
 
     // Map event groups identified by kind and name to their instances.
     std::map<std::pair<EventGroupKind, std::string>, std::shared_ptr<EventGroup>, std::less<>> event_groups;
+
+    // Map for event metadata identifier to their descriptors types.
+    std::map<zeek_uint_t, EventMetadataDescriptor> event_metadata_types;
```
bbannier, line 165: Like mentioned above, this could probably be a set (I'd go with `unordered_set` since we do not care about exact order), but it should be faster for our use case (low probability of hash collisions). If you keep this a map let's make it an `unordered_map` with the same efficiency argument.

src/cluster/Backend.cc
```diff
@@ -23,7 +23,11 @@ using namespace zeek::cluster;
 
 
 bool detail::LocalEventHandlingStrategy::DoProcessEvent(std::string_view topic, detail::Event e) {
-    zeek::event_mgr.Enqueue(e.Handler(), std::move(e.Args()), util::detail::SOURCE_BROKER, 0, nullptr, e.Timestamp());
+    zeek::detail::EventMetadataVectorPtr meta;
+    if ( auto ts = e.Timestamp(); ts > 0.0 )
+        meta = zeek::detail::MakeEventMetadataVector(e.Timestamp());
```
bbannier, line 28: Should we as a follow-up make these events store metadata as well and remove `Timestamp`?

src/zeek.bif
```diff
@@ -401,6 +401,70 @@ function current_event_time%(%): time
...
 
+## Register the expected Zeek type for event metadata.
+##
+## id: The event metadata identifier.
+##
+## t: A type expression or type alias.
+##
+## Returns: true if the registration was successful, false if *id* is
+##          registered with a different type already, or type is *any*
+##          or a composite type containing *any*.
```
bbannier, line 412: Nit: I'd move the requirements for `t` next to its arg docs, and then simplify the `Returns` docs, e.g.,

```suggestion
## t: A type expression or type alias. The type cannot be ``any`` or if a composite type contain ``any``.
##
## Returns: true if the registration was successful, false if *id* is
##          registered with a different type already, or type is *any*
##          or a composite type containing *any*.
##          registered with a different type already, or the passed type is
##          invalid.

```

src/zeek.bif
```diff
@@ -401,6 +401,70 @@ function current_event_time%(%): time
...
+## id: The metadata identifier, e.g. ``EventMetadata::NETWORK_TIMESTAMP``.
+##
+## Returns: A vector of values. The vector is empty if no metadata with
+##          the given identifier is attached to this event, otherwise a
+##          vector whose elements are of the type used during registration.
+##
+## .. zeek:see:: EventMetadata::register EventMetadata::current_all
+function EventMetadata::current%(id: EventMetadata::ID%): any_vec
+	%{
+	static const auto& vt = zeek::id::find_type<zeek::VectorType>("any_vec");
```
bbannier, line 442: Unused.

```suggestion
```

src/zeek.bif
```diff
@@ -401,6 +401,70 @@ function current_event_time%(%): time
...
+	static const auto& vt = zeek::id::find_type<zeek::VectorType>("event_metadata_vec");
+
+	auto result = zeek::make_intrusive<zeek::VectorVal>(vt);
+	const auto& mdv = zeek::event_mgr.CurrentEvent()->Metadata();
+
+	if ( mdv ) {
+		for ( const auto& entry : *mdv )
+			result->Append(entry.BuildVal());
+
+	}
```
bbannier, line 463: It seems `CurrentEvent` could return a `nullptr`. Let's either `assert` it does not, or handle it.

We could also slightly clean up where we introduce `mdv`.

```suggestion
	auto* event = zeek::event_mgr.CurrentEvent();
	if ( ! event )
		return result;

	if ( const auto* mdv = event->Metadata() ) {
 		for ( const auto& entry : *mdv )
 			result->Append(entry.BuildVal());
 	}
```

testing/btest/core/event-metadata/register.zeek
```diff
@@ -0,0 +1,30 @@
...
+		MY_STRING = 1000,
+		MY_COUNT = 1001,
+		MY_TABLE = 1002,
+	};
+}
+
+event zeek_init()
+	{
+	assert EventMetadata::register(MY_STRING, string);
+	assert EventMetadata::register(MY_STRING, string);  # double register is okay
```
bbannier, line 20: Nit:

```suggestion
	assert EventMetadata::register(MY_STRING, string); # double register is okay
```

testing/btest/core/event-metadata/register.zeek
```diff
@@ -0,0 +1,30 @@
...
+	assert EventMetadata::register(MY_COUNT, count);
+	assert EventMetadata::register(MY_COUNT, count);
+	assert EventMetadata::register(MY_TABLE, table[string] of count);
+	assert EventMetadata::register(MY_TABLE, table[string] of count);
+
+	# Type mismatch all return F, but no output on stderr.
+	assert ! EventMetadata::register(MY_STRING, count);
+	assert ! EventMetadata::register(MY_COUNT, string);
+	assert ! EventMetadata::register(MY_TABLE, table[count] of string);
```
bbannier, line 29: Could you add some tests for `any`, both naked and wrapped? Even though this is currently documented it seems unimplemented.

bbannier (commented):
The changes here feel like a good compromise to me since we need to integrate this into a pretty historically grown API we cannot break right-away.

I left some comments on the implementation details, mostly minor stuff. You probably still need to implement rejection of metadata involving `any` like documented, or change the docs.

src/Event.cc
```diff
@@ -2,29 +2,112 @@
...
+
+    for ( const auto& entry : *meta ) {
+        if ( entry.Id() != uintid )
+            continue;
+
+        // Sanity check the type.
+        if ( ! same_type(desc->Type(), entry.Val()->GetType()) ) {
+            zeek::reporter->InternalWarning("metadata has unexpected type %s, wanted %s",
+                                            obj_desc_short(entry.Val()->GetType().get()).c_str(),
+                                            obj_desc_short(desc->Type().get()).c_str());
```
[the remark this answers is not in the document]
awelzel: As in DM, this should be safe. [StackOverflow mach](https://stackoverflow.com/questions/10006891/is-it-safe-to-use-the-stdstringc-str-of-a-temporary-string)

src/EventRegistry.cc
awelzel: As in DM, find_if on a set would result in O(n). Will switch to unordered_map, however.

src/EventRegistry.h
awelzel: You're thinking about something like broker's [fwd.hh](https://github.com/zeek/broker/blob/master/libbroker/broker/fwd.hh) ? Maybe something like `ValFwd.h` ?

src/cluster/Backend.cc
awelzel: Yeah, that will happen in part 2. The commit mentions "intermediary", because I wanted to keep that PR confined mostly to the non-cluster side. Resolving.

src/zeek.bif
awelzel: Ouch, thanks!
```
$ zeek -e 'print EventMetadata::current_all()'
Segmentation fault (core dumped)
```

src/EventRegistry.cc
awelzel: Implemented now. Thanks.

src/Event.h
```diff
@@ -45,12 +96,14 @@ class Event final : public Obj {
     zeek::Args args;
     util::detail::SourceID src;
     analyzer::ID aid;
-    double ts;
-    Obj* obj;
+    zeek::IntrusivePtr<Obj> obj;
     Event* next_event;
+    detail::EventMetadataVectorPtr meta;
```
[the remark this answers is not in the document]
awelzel: Done as a separate follow-up commit due to conflict potential with other commits when done as fixup for squashing.

bbannier (approved):

PR #4486, opened 2025-05-22
Clang-tidy `bugprone` fixes

timwoj:
This is the first in a series of PRs to fix a large number of clang-tidy findings. This one fixes most of the `bugprone` checkers. The last commit in the PR adds some exclusions for checkers that we might want to enable later, with explanations as to why they're disabled.

This should probably be reviewed commit-by-commit since each commit is a separate checker and all of the changes are related. This also pulls in the changes from https://github.com/zeek/binpac/pull/43, which includes fixes for `bugprone-branch-clone` amongst others.


src/Stmt.cc
```diff
@@ -1226,7 +1226,7 @@ ValPtr ForStmt::DoExec(Frame* f, Val* v, StmtFlowType& flow) {
 bool ForStmt::IsPure() const { return e->IsPure() && body->IsPure(); }
 
 void ForStmt::StmtDescribe(ODesc* d) const {
-    Stmt::StmtDescribe(d);
+    Stmt::StmtDescribe(d); // NOLINT(bugprone-parent-virtual-call)
```
awelzel, line 1229: Looking at `Stmt::StmtDescribe()`, isn't this just running `AddTag(d)` due to `Tag () != STMT_EXPR`. Calling `AddTag(d)` directly here might be more obvious and avoids the eye-brow raising NOLINT :-)

src/Reporter.cc
```diff
@@ -467,7 +467,7 @@ void Reporter::Deprecation(std::string_view msg, const detail::Location* loc1, c
     if ( loc1 || loc2 )
         PushLocation(loc1, loc2);
 
-    Warning("%s", msg.data());
+    Warning("%.*s", static_cast<int>(msg.size()), msg.data());
```
awelzel, line 470: Thanks!

src/storage/backend/redis/Redis.cc
```diff
@@ -246,7 +246,8 @@ OperationResult Redis::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
     StringValPtr host = backend_options->GetField<StringVal>("server_host");
     if ( host ) {
         PortValPtr port = backend_options->GetField<PortVal>("server_port");
-        server_addr = util::fmt("%s:%d", host->ToStdStringView().data(), port->Port());
+        server_addr = util::fmt("%.*s:%d", static_cast<int>(host->ToStdStringView().size()),
+                                host->ToStdStringView().data(), port->Port());
```
awelzel, line 250: This would probably be less noisy wit `ToStdString().c_str()` and not slower.

src/Reporter.cc
```diff
@@ -537,10 +537,12 @@ void Reporter::DoLog(const char* prefix, EventHandlerPtr event, FILE* out, Conne
 
         // Enlarge buffer;
         size *= 2;
-        buffer = allocated = (char*)realloc(allocated, size);
+        buffer = (char*)realloc(allocated, size);
 
         if ( ! buffer )
             FatalError("out of memory in Reporter");
+        else
+            allocated = buffer;
```
awelzel, line 544: No need for the else here.

awelzel: Wow, looking around, I guess it might make sense to try and switch from the following four variables to a `std::string` approach and assume whatever optimization is being done here isn't needed :-)

```
    static char tmp[512];
    int size = sizeof(tmp);
    char* buffer = tmp;
    char* allocated = nullptr;

```

src/analyzer/protocol/modbus/modbus-protocol.pac
```diff
@@ -112,7 +112,7 @@ type ModbusTCP_Request(header: ModbusTCP_TransportHeader) = case header.fc of {
 #
 type ModbusTCP_Response(header: ModbusTCP_TransportHeader) = case header.fc & 0x80 of {
     0       -> normal_response      : ModbusTCP_NormalResponse(header);
-    default -> exception_response   : Exception(header);
+    default -> exception_response   : ExcResponse(header);
```
awelzel, line 115: Nit. Given the normal response is called `ModbusTCP_NormalResponse`, `ModbusTCP_ExceptionResponse` might be a candidate that pleases clang-tidy and doesn't look artificially shortened?

src/TraverseTypes.h
```diff
@@ -22,9 +22,7 @@ enum TraversalCode {
 
 #define HANDLE_TC_STMT_POST(code)                                                                                      \
     {                                                                                                                  \
-        if ( (code) == zeek::detail::TC_ABORTALL )                                                                     \
-            return (code);                                                                                             \
-        else if ( (code) == zeek::detail::TC_ABORTSTMT )                                                               \
```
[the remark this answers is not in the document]
timwoj: All of those macros could be `constexpr` methods that take and return a `TraversalCode` also.

src/Stmt.cc
timwoj: Sure. The lint note was because calling `ExprStmt::StmtDescribe` causes the output to differ for `ForStmt` that what we're currently outputting.

For example:

```
for ([HTTP::_], HTTP::c in HTTP::f$conns) { if (HTTP::c?$http) return (HTTP::build_url_http(HTTP::c$http))}return ()}
```

turns into

```
for HTTP::f$conns([HTTP::_], HTTP::c in HTTP::f$conns) { if (HTTP::c?$http) return (HTTP::build_url_http(HTTP::c$http))}return ()}
```

src/Reporter.cc
timwoj: Looking at this again, you actually don't need to repeatedly make this call either. Passing a null buffer to vsnprintf will return the size of the buffer needed, which you can then allocate one time instead of realloc'ing repeatedly.

timwoj: I made a change to only call `realloc()` once if needed. We can look at the performance differences of using std::string here separately.

src/analyzer/protocol/modbus/modbus-protocol.pac
timwoj: clang-tidy is very picky about naming things with `exception` in their name:

```
[6/41] Building CXX object src/analyzer/protocol/modbus/CMakeFiles/plugin-Zeek-Modbus.dir/modbus_pac.cc.o
/Users/tim/Desktop/projects/zeek-master/build/src/analyzer/protocol/modbus/modbus_pac.cc:585:43: warning: suspicious exception object created but not thrown; did you mean 'throw ModbusTCP_ExceptionResponse'? [bugprone-throw-keyword-missing]
```

timwoj: But it does seem happy with `ModbusTCP_ExceptResponse` :shrug:

PR #4488, opened 2025-05-22
Fix some follow-up issues with the sqlite cluster test

timwoj:
- Adds a bunch of debug logging around executing pragmas
- Use the iteration from TableVal when looping over the pragmas. The table is defined with `&ordered`. Converting it to a map first caused the keys to be reordered based on the process's map hashing, which might not have been the same order.
- Fixes a variable name typo that was keeping busy pragmas from retrying
- Reenables the btest

### DIFF
--- a/src/storage/backend/sqlite/SQLite.cc
+++ b/src/storage/backend/sqlite/SQLite.cc
@@ -5,7 +5,9 @@
 #include <thread>
 
 #include "zeek/3rdparty/sqlite3.h"
+#include "zeek/CompHash.h"
 #include "zeek/DebugLogger.h"
+#include "zeek/Dict.h"
 #include "zeek/Func.h"
 #include "zeek/Val.h"
 #include "zeek/storage/ReturnCode.h"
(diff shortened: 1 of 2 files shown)


src/storage/backend/sqlite/SQLite.cc
```diff
@@ -37,18 +39,21 @@ OperationResult SQLite::RunPragma(std::string_view name, std::optional<std::stri
...
+            DBG_LOG(DBG_STORAGE, "%s", err.c_str());
             return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
         }
     }
 
     if ( pragma_timeout != 0ms && time_spent >= pragma_timeout ) {
-        std::string err =
-            util::fmt("Database was busy while executing %.*s pragma", static_cast<int>(name.size()), name.data());
+        std::string err = util::fmt("Database was busy while executing '%s'", cmd.c_str());
+        DBG_LOG(DBG_STORAGE, "%s", err.c_str());
```
ckreibich, line 51: Trivial — but you could just do
```suggestion
        DBG_LOG(DBG_STORAGE, err.c_str());
```
for these.

ckreibich (approved):
Yep, the `sqlite-cluster` test now runs a-okay over here. :+1:

src/storage/backend/sqlite/SQLite.cc
timwoj: I could, except then the compiler/emacs complains that you shouldn't pass variables to printf-like functions without a format string:

```
[5/7] Building CXX object src/storage/backend/sqlite/...es/plugin-Zeek-Storage_Backend_SQLite.dir/SQLite.cc.o
/Users/tim/Desktop/projects/zeek/src/storage/backend/sqlite/SQLite.cc:51:30: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
   51 |         DBG_LOG(DBG_STORAGE, err.c_str());
      |                              ^~~~~~~~~~~
/Users/tim/Desktop/projects/zeek/src/include/zeek/DebugLogger.h:20:46: note: expanded from macro 'DBG_LOG'
   20 |     ::zeek::detail::debug_logger.Log(stream, __VA_ARGS__)
      |                                              ^~~~~~~~~~~
/Users/tim/Desktop/projects/zeek/src/storage/backend/sqlite/SQLite.cc:51:30: note: treat the string as an argument to avoid this
   51 |         DBG_LOG(DBG_STORAGE, err.c_str());
      |                              ^
      |                              "%s", 
/Users/tim/Desktop/projects/zeek/src/include/zeek/DebugLogger.h:20:46: note: expanded from macro 'DBG_LOG'
   20 |     ::zeek::detail::debug_logger.Log(stream, __VA_ARGS__)
      |                                              ^
1 warning generated.
```

PR #4489, opened 2025-05-22
CI: Modify task to run clang-tidy as part of builds

timwoj:
This modifies the ubuntu24_clang_libcpp task to run the build with `--enable-clang-tidy` enabled. It removes the skip-on-PR option for that task as well so that it always runs. As part of this, update the clang version installed on the ubuntu 24.04 VM to 19.x. I'd update to 20.x but there aren't packages for that version on ubuntu.

This depends on the changes to .clang-tidy that start in https://github.com/zeek/zeek/pull/4486, so that PR needs to be merged first. Otherwise it will report a lot of warnings that aren't fixed yet.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -300,8 +300,8 @@ ubuntu24_clang_libcpp_task:
   << : *CI_TEMPLATE
   << : *SKIP_TASK_ON_PR
   env:
-    CC: clang-18
-    CXX: clang++-18
+    CC: clang-19
+    CXX: clang++-19
     CXXFLAGS: -stdlib=libc++
 
 ubuntu22_task:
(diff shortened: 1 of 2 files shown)


ci/ubuntu-24.04/Dockerfile
```diff
@@ -47,6 +48,10 @@ RUN apt-get update && apt-get -y install \
 RUN pip3 install --break-system-packages websockets junit2html
 RUN gem install coveralls-lcov
 
+# Ubuntu installs clang versions with the binaries having the version number
+# appended. Create a symlink for clang-tidy so cmake finds it correctly.
+RUN if [ ! -f /usr/bin/clang-tidy ]; then ln -s /usr/bin/clang-tidy-19 /usr/bin/clang-tidy; fi
```
awelzel, line 53: Maybe update-alternatives?

```suggestion
RUN update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-19 1000
```

timwoj:
I'm debating whether this should be split out into another task that just does the clang-tidy build without any tests. https://cirrus-ci.com/task/6369481020145664 took a long time (over 30 minutes) to do the build step, and then had to wait for tests also. Maybe it doesn't matter considering the asan test takes over an hour already?

.cirrus.yml
```diff
@@ -298,11 +299,11 @@ ubuntu24_clang_libcpp_task:
     dockerfile: ci/ubuntu-24.04/Dockerfile
     << : *RESOURCES_TEMPLATE
   << : *CI_TEMPLATE
-  << : *SKIP_TASK_ON_PR
```
awelzel, line 301: I'd be totally fine if we run clang-tidy only after merging or just once a week, but I guess we don't care too much about the cycles?

awelzel (approved):

PR #4499, opened 2025-05-26
Custom event metadata - part 2

awelzel:
* Attempt to remove `ts` from cluster APIs
* Add generic metadata support to the broker serializer in the cluster subsystem
* Update Backend and broker/Manager to process all metadata, not just timestamps
* Change some behavior for missing remote network timestamp metadata
  * Keep it unset by default, but allow the old behavior of automatically using local network time
* Adds a plugin+cluster test that demonstrates using HookPublishEvent to attach metadata to events depending on the topic these events are published to.

The individual commits hopefully make sense.


src/Event.cc
```diff
@@ -165,12 +166,19 @@ void EventMgr::Enqueue(const EventHandlerPtr& h, Args vl, util::detail::SourceID
...
+    // If this is a local event and EventMetadata::add_network_timestamp is
+    // enabled, automatically set the network timestamp for this event to the
+    // current network time when it is < 0 (default of deprecated_ts is -1.0).
+    //
+    // See the other Enqueue() implementation for the local vs broker/remote
+    // motivation of want_network_timestamp.
+    bool want_network_timestamp =
+        BifConst::EventMetadata::add_network_timestamp &&
+        ((src == util::detail::SOURCE_LOCAL) ||
+         (src == util::detail::SOURCE_BROKER && BifConst::EventMetadata::add_missing_remote_network_timestamp));
```
bbannier, line 179: Was this supposed to be this? Also, the comment seems outdated.

```suggestion
    bool want_network_timestamp =
        (src == util::detail::SOURCE_LOCAL && BifConst::EventMetadata::add_network_timestamp) ||
        (src == util::detail::SOURCE_BROKER && BifConst::EventMetadata::add_missing_remote_network_timestamp);
```

src/Event.cc
```diff
@@ -189,15 +197,23 @@ void EventMgr::Enqueue(const EventHandlerPtr& h, Args vl, util::detail::SourceID
...
+    //
+    // Why so complicated? It seems less surprising behavior to keep network timestamp metadata unset
+    // if a remote event didn't have any attached. It should help to more easily figure out what's
+    // actually going on compared to setting it to the local network time. If all nodes are required to
+    // send *their* network timestamp, filling it with this node's network time seems more confusing
+    // and error prone compared to just leaving it unset and having the consumer deal with the situation.
+    bool want_network_timestamp =
+        BifConst::EventMetadata::add_network_timestamp &&
+        ((src == util::detail::SOURCE_LOCAL) ||
+         (src == util::detail::SOURCE_BROKER && BifConst::EventMetadata::add_missing_remote_network_timestamp));
```
bbannier, line 214: Ditto.

```suggestion
    bool want_network_timestamp =
        (src == util::detail::SOURCE_LOCAL && BifConst::EventMetadata::add_network_timestamp) ||
        (src == util::detail::SOURCE_BROKER && BifConst::EventMetadata::add_missing_remote_network_timestamp);
```

src/cluster/serializer/broker/Serializer.cc
```diff
@@ -20,6 +23,33 @@
...
+    auto meta = std::make_unique<zeek::detail::EventMetadataVector>();
+    meta->reserve(broker_meta.size());
+
+    for ( const auto& [id, v] : broker_meta ) {
+        const auto* desc = zeek::event_registry->LookupMetadata(id);
+        if ( ! desc )
+            continue;
+
+        auto d = v.to_data();
+        auto val = zeek::Broker::detail::data_to_val(d, desc->Type().get());
```
bbannier, line 45: `data_to_val` might move from `d`, at least if `TYPE_ANY` is requested. We could stringify it beforehand, or at least assert that we do would not make that simple move. Overall this Broker API is pretty hostile 😬.

```suggestion
        assert(desc->Type()->Tag() != TYPE_ANY);
        auto val = zeek::Broker::detail::data_to_val(d, desc->Type().get());
```

src/Event.cc
```diff
@@ -96,7 +97,7 @@ zeek::VectorValPtr Event::MetadataValues(const EnumValPtr& id) const {
 
 double Event::Time() const {
     if ( ! meta )
-        return 0.0;
+        return -1.0;
```
bbannier: Could you introduce a named constant for this for discoverability? We would use that constant also in `Event::CurrentEventTime` in `Event.h`.

src/cluster/Backend.cc
```diff
@@ -19,15 +22,46 @@
...
+                return m.Val()->AsTime();
+            }
+        }
+    }
 
-bool detail::LocalEventHandlingStrategy::DoProcessEvent(std::string_view topic, detail::Event e) {
-    zeek::detail::EventMetadataVectorPtr meta;
-    if ( auto ts = e.Timestamp(); ts >= 0.0 )
-        meta = zeek::detail::MakeEventMetadataVector(e.Timestamp());
+    return -1.0;
```
bbannier: This should use some constant.

src/Event.cc
```diff
@@ -109,7 +110,7 @@ double Event::Time() const {
             return m.Val()->AsTime();
         }
 
-    return 0.0;
+    return -1.0;
```
bbannier: Should use a named constant.

src/Event.cc
```diff
@@ -11,6 +11,7 @@
 #include "zeek/Val.h"
 #include "zeek/iosource/Manager.h"
 #include "zeek/plugin/Manager.h"
+#include "zeek/util.h"
```
bbannier, line 14: Unused?

```suggestion
```

src/cluster/Backend.h
```diff
@@ -9,6 +9,7 @@
 #include <string_view>
 #include <variant>
 
+#include "zeek/Event.h" // for generic metadata
```
bbannier:
```suggestion
#include "zeek/Event.h"
```

src/cluster/BifSupport.cc
```diff
@@ -14,32 +14,25 @@
 #include "zeek/broker/Manager.h" // For publishing to broker_mgr directly.
 #include "zeek/cluster/Backend.h"
 
+#include "const.bif.netvar_h"
```
bbannier: Unused?

```suggestion
```

src/broker/Manager.cc
```diff
@@ -849,7 +852,15 @@ bool Manager::PublishEvent(string topic, std::string name, broker::vector args,
     if ( peer_count == 0 && hub_count == 0 )
         return true;
 
-    broker::zeek::Event ev(name, args, broker::to_timestamp(ts));
+    broker::vector meta;
+    if ( BifConst::EventMetadata::add_network_timestamp ) {
+        broker::vector entry(2);
+        entry[0] = static_cast<broker::count>(zeek::detail::MetadataType::NetworkTimestamp);
+        entry[1] = broker::to_timestamp(ts);
```
bbannier: Nit: The following has fewer magic constants.

```suggestion
        broker::vector entry{static_cast<broker::count>(zeek::detail::MetadataType::NetworkTimestamp),
                             broker::to_timestamp(ts)};
```

src/broker/Manager.cc
```diff
@@ -849,7 +852,15 @@ bool Manager::PublishEvent(string topic, std::string name, broker::vector args,
...
+        meta.emplace_back(std::move(entry));
+    }
+
+    broker::zeek::Event ev(name, args, meta);
     DBG_LOG(DBG_BROKER, "Publishing event: %s", RenderEvent(topic, name, ev.args()).c_str());
```
bbannier: Nit: Not added here, but could clean up for if we ever should move `name` above.

```suggestion
    DBG_LOG(DBG_BROKER, "Publishing event: %s", RenderEvent(topic, ev.name(), ev.args()).c_str());
```

src/cluster/Backend.cc
```diff
@@ -19,15 +22,46 @@
...
+}
+
+bool detail::Event::AddMetadata(const EnumValPtr& id, zeek::ValPtr val) {
```
bbannier, line 40: Nit: We check `val` for null below, but never id. The safety guarantees around Zeek's pointer vals are never clear to me, but I'd prefer

```suggestion
bool detail::Event::AddMetadata(const EnumValPtr& id, zeek::ValPtr val) {
    if ( ! id || ! val )
        return false;

```

src/cluster/Backend.cc
```diff
@@ -19,15 +22,46 @@
...
+    const auto* desc = zeek::event_registry->LookupMetadata(id->Get());
+    if ( ! desc )
+        return false;
+
+    if ( ! val || ! same_type(val->GetType(), desc->Type()) )
```
bbannier: If you check `val` on entry no need to repeat it here.

Does it make sense to add debug output with the found types for when we exit here?

src/cluster/Backend.cc
```diff
@@ -19,15 +22,46 @@
...
+#include "zeek/3rdparty/doctest.h"
+
 using namespace zeek::cluster;
 
+double detail::Event::Timestamp() const {
+    if ( meta ) {
+        for ( const auto& m : *meta ) {
+            if ( m.Id() == static_cast<zeek_uint_t>(zeek::detail::MetadataType::NetworkTimestamp) ) {
+                return m.Val()->AsTime();
+            }
```
bbannier: Nit.

```suggestion
            if ( m.Id() == static_cast<zeek_uint_t>(zeek::detail::MetadataType::NetworkTimestamp) )
                return m.Val()->AsTime();
```

src/cluster/Backend.cc
```diff
@@ -19,15 +22,46 @@
...
+        return false;
+
+    if ( ! val || ! same_type(val->GetType(), desc->Type()) )
+        return false;
+
+    if ( ! meta )
+        meta = std::make_unique<zeek::detail::EventMetadataVector>();
+
+    // Internally stored as zeek_uint_t for serializers.
+    meta->push_back({desc->Id(), std::move(val)});
```
bbannier: Nit:

```suggestion
    meta->emplace_back(desc->Id(), std::move(val));
```

src/cluster/Backend.cc
```diff
@@ -100,21 +133,38 @@ bool Backend::Init(std::string nid) {
     return DoInit();
 }
 
-std::optional<detail::Event> Backend::MakeClusterEvent(FuncValPtr handler, ArgsSpan args, double timestamp) const {
+std::optional<detail::Event> Backend::MakeClusterEvent(FuncValPtr handler, ArgsSpan args) const {
```
bbannier, line 138: Complete nit, but now that you touch and use this in more places, I wonder why this isn't just a standalone function (or at least a `static` method). Did it need access to backend state in the past?

If this was a standalone function we wouldn't need to pass a `Backend` to `to_cluster_event` which has no business dealing with backends.

src/cluster/serializer/broker/Serializer.cc
```diff
@@ -20,6 +23,33 @@
...
+    if ( broker_meta.size() == 0 )
+        return nullptr;
+
+    auto meta = std::make_unique<zeek::detail::EventMetadataVector>();
+    meta->reserve(broker_meta.size());
+
+    for ( const auto& [id, v] : broker_meta ) {
+        const auto* desc = zeek::event_registry->LookupMetadata(id);
+        if ( ! desc )
+            continue;
```
bbannier, line 41: A debug log would be nice here.

NEWS
```diff
@@ -24,9 +24,26 @@ Breaking Changes
   a small overhead when enabled. There's not enough users of network timestamp
   metadata to justify the complexity of treating it separate.
 
+- The ``current_event_time()`` builtin function as well as ``Event::Time()``
+  and ``EventMgr::CurrentEventTime()`` now return ``-1.0`` if not timestamp
```
bbannier:
```suggestion
  and ``EventMgr::CurrentEventTime()`` now return ``-1.0`` if no timestamp
```

NEWS
```diff
@@ -24,9 +24,26 @@ Breaking Changes
...
+  network time anymore by default. This potentially hid useful debugging information
+  about another node not sending timestamp metadata. The old behavior can be
+  re-enabled as follows:
+
+	redef EventMetadata::add_missing_remote_network_timestamp = T;
+
 New Functionality
 -----------------
 
+- Generic event metadata support. A new ``EventMetadata`` module was added, allowing
```
bbannier:
```suggestion
- Generic event metadata support. A new ``EventMetadata`` module was added allowing
```

src/cluster/Backend.cc
```diff
@@ -100,21 +133,38 @@ bool Backend::Init(std::string nid) {
...
+    /**
+     * If you ever stare at this and wonder: Currently, if someone calls
+     * Cluster::publish() from within a remote event in script land, then
+     * CurrentEventTime() below will yield the network timestamp of the
+     * remote event. That means that the outgoing event from this node will
+     * have the network timestamp of the originating node, which may be
+     * different from what the local network time is.
+     *
+     * This could be confusing and another policy might be to always set
+     * the network timestamp metadata for for outgoing events to the local
```
bbannier, line 159:
```suggestion
     * the network timestamp metadata for outgoing events to the local
```

src/cluster/Backend.h
```diff
@@ -60,16 +61,40 @@ class Event {
...
+    double Timestamp() const;
+
+    /**
+     * Add metadata to this cluster event.
+     *
+     * The used metadata \a id has to be registered via the Zeek script-layer
+     * function EventMetadata::register(), or via the C++ API
+     * EventMgr::RegisterMetadata() during an InitPostScript() hook.
+     *
+     * Non-registered will not be added and false is returned.
```
bbannier:
```suggestion
     * Non-registered metadata will not be added and false is returned.
```

testing/btest/plugins/publish-event-metadata-plugin/src/Plugin.cc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void Plugin::InitPostScript() {
+    std::fprintf(stdout, "%.6f %-15s\n", zeek::run_state::network_time, "  InitPostScript");
+}
+
+bool Plugin::HookPublishEvent(zeek::cluster::Backend& backend, const std::string& topic,
+                              zeek::cluster::detail::Event& event) {
+    std::fprintf(stdout, "%.6f %-15s backend=%s topic=%s event=%s\n", zeek::run_state::network_time,
+                 "  HookPublishEvent", backend.Name().c_str(), topic.c_str(), std::string(event.HandlerName()).c_str());
```
bbannier: Nit: No need to be so particular.

```suggestion
    std::fprintf(stdout, "%.6f %s\n", zeek::run_state::network_time, "InitPostScript");
}

bool Plugin::HookPublishEvent(zeek::cluster::Backend& backend, const std::string& topic,
                              zeek::cluster::detail::Event& event) {
    std::fprintf(stdout, "%.6f %s backend=%s topic=%s event=%s\n", zeek::run_state::network_time, "HookPublishEvent",
                 backend.Name().c_str(), topic.c_str(), std::string(event.HandlerName()).c_str());
```

testing/btest/plugins/publish-event-metadata-plugin/src/Plugin.cc
```diff
@@ -0,0 +1,88 @@
...
+
+    const auto& table_type = zeek::id::find_type<zeek::TableType>("table_string_of_string");
+    const auto& string_md = zeek::id::find_val<zeek::EnumVal>("App::CUSTOM_METADATA_STRING");
+    auto count_md = zeek::id::find_val<zeek::EnumVal>("App::CUSTOM_METADATA_COUNT");
+    auto table_md = zeek::id::find_val<zeek::EnumVal>("App::CUSTOM_METADATA_TABLE");
+
```
bbannier, line 43: Nit since elsewhere this handles errors well.

In any case we should add `#include <zeek/Reporter.h>`.

```suggestion
    if ( ! count_md || ! table_md )
        zeek::reporter->FatalError("could not find required enum values");

```

testing/btest/plugins/publish-event-metadata-plugin/src/Plugin.cc
```diff
@@ -0,0 +1,88 @@
...
+    if ( topic == "topic1" ) {
+        if ( ! event.AddMetadata(string_md, zeek::make_intrusive<zeek::StringVal>("testing string metadata")) ) {
+            zeek::reporter->FatalError("Failed to add string metadata");
+        }
+    }
+    else if ( topic == "topic2" ) {
+        if ( ! event.AddMetadata(count_md, zeek::val_mgr->Count(42424242)) ) {
+            zeek::reporter->FatalError("Failed to add string metadata");
```
bbannier:
```suggestion
            zeek::reporter->FatalError("Failed to add count metadata");
```

testing/btest/plugins/publish-event-metadata-plugin/src/Plugin.cc
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+    else if ( topic == "topic3" ) {
+        auto tv = zeek::make_intrusive<zeek::TableVal>(table_type);
+        tv->Assign(zeek::make_intrusive<zeek::StringVal>("key1"), zeek::make_intrusive<zeek::StringVal>("val1"));
```
bbannier: Nit:

```suggestion
        if ( ! tv->Assign(zeek::make_intrusive<zeek::StringVal>("key1"),
                          zeek::make_intrusive<zeek::StringVal>("val1")) )
            zeek::reporter->FatalError("could not update table value");
```

testing/btest/core/event-metadata/current-event-time.zeek
```diff
@@ -0,0 +1,17 @@
+# @TEST-DOC: Check current_event_time() produces the same as event metadata, or else -1.0
+#
+# @TEST-EXEC: zeek -r $TRACES/http/get.trace %INPUT EventMetadata::add_network_timestamp=T > with_ts.out
+# @TEST-EXEC: zeek -r $TRACES/http/get.trace %INPUT EventMetadata::add_network_timestamp=F > without_ts.out
+#
+# @TEST-EXEC: TEST_DIFF_CANONIFIER= btest-diff with_ts.out
+# @TEST-EXEC: TEST_DIFF_CANONIFIER= btest-diff without_ts.out
+# @TEST-EXEC: btest-diff .stderr
+
```
bbannier: What do you think of cutting down the number of baselines? I personally find them harder to review than single files, and we do not really seem to need them. We also do not really care about the conn id below. Something like this:

```suggestion
# @TEST-EXEC: zeek -r $TRACES/http/get.trace %INPUT EventMetadata::add_network_timestamp=T >> output 2>&1
# @TEST-EXEC: zeek -r $TRACES/http/get.trace %INPUT EventMetadata::add_network_timestamp=F >> output 2>&1
#
# @TEST-EXEC: TEST_DIFF_CANONIFIER= btest-diff output
```

testing/btest/core/event-metadata/current-event-time.zeek
```diff
@@ -0,0 +1,17 @@
...
+
+event new_connection(c: connection)
+	{
+	print fmt("new_connection %s current_event_time=%s network_timestamp=%s",
+	          c$uid,
+	          current_event_time(),
+	          EventMetadata::current(EventMetadata::NETWORK_TIMESTAMP));
```
bbannier: Let's drop uninteresting conn id, but log the value of `add_network_timestamp`.

```suggestion
	print fmt("new_connection add_network_timestamp=%s current_event_time=%s network_timestamp=%s",
	    EventMetadata::add_network_timestamp, current_event_time(),
	    EventMetadata::current(EventMetadata::NETWORK_TIMESTAMP));
```

src/Event.cc
awelzel: The idea was that `add_missing_remote_network_timestamp` only has a meaning if `add_network_timestamp` is also `T` - that's in its documentation in `init-bare.zeek`. I'll leave as is for now.

It reduces the possible scenarios one needs to think about. I'd be up for adding some `InitPostScript()` to ensure that `add_missing_remote_network_timestamp` is only set if `add_network_timestamp` is also set.

Does that feel like too much confinement?

src/cluster/serializer/broker/Serializer.cc
awelzel: Ah, thanks for checking. I replaced the `broker::to_string(d)` call with `broker::to_string(v.to_data())`. That was your concern, correct?

src/Event.cc
bbannier: I see. This seems to reduce the scenarios Zeek developers need to think about, but still leaves users with two highly correlated variables of which one does nothing on its own so it IMO doesn't seem to remove much mental overhead. Ideally we'd only expose a single knob, but these are really two different behaviors.

The scenarios already seemed pretty confined to me, do you see us handling these in more places in the future?

src/cluster/serializer/broker/Serializer.cc
bbannier: Yes, much better.

src/Event.cc
awelzel: SOURCE_LOCAL and SOURCE_BROKER are in `util.h`. My IDE added it when completing these. It's already included via `Event.h`, so I'm okay dropping it here. But could also see keeping it, what's your take?

bbannier: Sorry, in that case KEEP.

src/cluster/Backend.cc
awelzel: I had a reporter error earlier, but I think users can check the return value and hopefully be able to figure out what's going on. It's true that a non-matching type might be hard to debug, but hopefully not impossible.  And only outputting the error when compiled with DEBUG doesn't really appeal to me.

testing/btest/core/event-metadata/current-event-time.zeek
awelzel: Yeah, makes sense, thanks!

src/cluster/Backend.cc
awelzel: Let me postpone this, but yes, `MakeClusterEvent` doesn't need to be on `Backend` and probably shouldn't.

src/Event.cc
awelzel: Not really, hopefully, but I don't see them being toggled individually, so I'll tighten it down.

awelzel:
> Do you have a feeling for the runtime impact of now constructing vectors as compared to the state before that change went in?

I've tested locally a bit and it's definitely visible in micro-benchmarks around pure event handling, but seemed acceptable when actually processing PCAP / traffic. I've stared quite a bit and ended up deciding that removing `Obj` inheritance from `Event` would help alleviate the performance impact more than focusing vector and metadata. But not doing this here.

Issue #4463, opened 2025-05-15
FTPS is not recognized


J-Gras:
The following PCAP contains a FTPS control connection: [ftps-control.pcap.tgz](https://github.com/user-attachments/files/20229339/ftps-control.pcap.tgz) Unfortunately, Zeek fails to analyze this as FTPS. From a quick glance at the code of the FTP analyzer, it seemed that Zeek should understand FTPS. Maybe it's limited to a different variant?

JustinAzoff:
ah, there are two different ftps mechanisms, Implicit and Explicit.  Explicit is what this client did using a `STARTTLS` like mechanism.

Implicit would look like any other TLS connection, but Explicit requires first handling the `AUTH TLS` command.  We appear to only handle `AUTH GSSAPI` (I think what GridFTP used).


```
T 127.0.0.1:40284 -> 127.0.0.1:21 [AP] #20
  AUTH TLS..                                                                                                                                                                  
#
T 127.0.0.1:21 -> 127.0.0.1:40284 [AP] #21
  234 Proceed with negotiation...
```

0xxon:
If someone implements this - can we use the trace as part of our testsuite?

J-Gras:
> If someone implements this - can we use the trace as part of our testsuite?

Yes! 🙂

PR #4503, opened 2025-05-27
Add explicit TLS support for FTP

0xxon:
This is defined in RFC 4217; TLS initialized by the client sending an `AUTH TLS` command.

Fixes GH-4463


src/analyzer/protocol/ftp/FTP.h
```diff
@@ -22,8 +22,9 @@ class FTP_Analyzer final : public analyzer::tcp::TCP_ApplicationAnalyzer {
 protected:
     analyzer::login::NVT_Analyzer* nvt_orig;
     analyzer::login::NVT_Analyzer* nvt_resp;
-    uint32_t pending_reply;     // code associated with multi-line reply, or 0
-    std::string auth_requested; // AUTH method requested
+    uint32_t pending_reply;       // code associated with multi-line reply, or 0
+    std::string auth_requested;   // AUTH method requested
+    bool switched_to_tls = false; // starttls active
```
awelzel, line 27: This is called `tls_active` in quite a few analyzers  and also `starttls` in case you'd be up for aligning.

awelzel (approved):
Nice!

Most other analyzers come with a `<protocol>_starttls(c: connection)` event in case you'd be up for `ftp_starttls()`.

src/analyzer/protocol/ftp/FTP.cc
```diff
@@ -172,6 +177,17 @@ void FTP_Analyzer::DeliverStream(int length, const u_char* data, bool orig) {
             }
         }
 
+        if ( reply_code == 234 && auth_requested.size() > 0 && auth_requested == "TLS" ) {
```
awelzel, line 180:
```suggestion
        if ( reply_code == 234 && auth_requested == "TLS" ) {
```

Sorry, I missed that, think the `.size()` is redundant with the strcmp, unless you wanted to save a few cycles here on purpose? Your call.

awelzel (approved):
Neat!

src/analyzer/protocol/ftp/FTP.cc
0xxon: It's analogous to how the GSSAPI check does things. Looks the same as the other code, if nothing else.

PR #4506, opened 2025-05-27
Silence -Wnontrivial-memcall warning in ConnKey methods

timwoj:
I noticed that Spicy [recently upgraded to LLVM 20](https://github.com/zeek/spicy/pull/2041) and fixed a few new compiler warnings, so I figured I'd do the same locally for Zeek. This fixes the one new reported warning: an instance of `-Wnontrivial-memcall`.

### DIFF
--- a/src/IPAddr.cc
+++ b/src/IPAddr.cc
@@ -33,8 +33,9 @@ ConnKey& ConnKey::operator=(const ConnKey& rhs) {
 
     // Because of padding in the object, this needs to memset to clear out
     // the extra memory used by padding. Otherwise, the session key stuff
-    // doesn't work quite right.
-    memset(this, 0, sizeof(ConnKey));
+    // doesn't work quite right. The static_cast is to silence a
+    // -Wnontrival-memcall warning from clang++ 17 and later.
+    memset(static_cast<void*>(this), 0, sizeof(ConnKey));
 
     memcpy(&ip1, &rhs.ip1, sizeof(in6_addr));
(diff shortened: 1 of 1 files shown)


src/IPAddr.cc
```diff
@@ -33,8 +33,9 @@ ConnKey& ConnKey::operator=(const ConnKey& rhs) {
 
     // Because of padding in the object, this needs to memset to clear out
     // the extra memory used by padding. Otherwise, the session key stuff
-    // doesn't work quite right.
-    memset(this, 0, sizeof(ConnKey));
+    // doesn't work quite right. The static_cast is to silence a
+    // -Wnontrival-memcall warning from clang++ 17 and later.
+    memset(static_cast<void*>(this), 0, sizeof(ConnKey));
```
awelzel, line 38: I wonder about explicitly pushing  `-Wno-nontrivial-memcall` to ignore this, mostly worried that we copy&paste the line without the comment somewhere and move on.

awelzel (approved):

src/IPAddr.cc
timwoj: Non-trivial memsets like this (without the cast) are technically UB, since the type isn't trivially copyable. The `void*` cast silences it, but I agree that we could just ignore the warning since it's doing the same thing. Ideally, we should fix those types so that they can be copied, or in this case properly reset/initialized.

awelzel: > Non-trivial memsets like this (without the cast) are technically UB, since the type isn't trivially copyable

Ah, thanks. Resolving!

PR #4507, opened 2025-05-27
Add handling of expiration contention to SQLite backend

### LINKED ISSUE #4482: Handle contention between storage backends running expiration
This is a tracking issue for the discussion at https://github.com/zeek/zeek/discussions/4374 so it can be added to the 8.0 board.

timwoj:
This adds handling of contention between individual Zeek processes when running expiration on an SQLite storage backend. It does this by creating an additional table storing expiration timestamps and checking/updating them during an expiration call. This happens under a transaction to prevent contention, allowing one process to lock out the others.

Fixes #4482 

RE: #4482, I decided that there wasn't anything to do for Redis in this space. Pcap mode effectively means it's a single process, and clusters should be using an actual server which means they're also using the built-in expiration from the server.


src/storage/backend/sqlite/SQLite.cc
```diff
@@ -279,36 +313,113 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+    status = SQLITE_BUSY;
+    while ( status == SQLITE_BUSY ) {
+        status = sqlite3_exec(db, "begin exclusive transaction", nullptr, nullptr, &errMsg);
+        if ( status == SQLITE_OK )
+            break;
+        else if ( status == SQLITE_BUSY )
+            // I honestly don't know if we need to wait here. This should fall under the
+            // 'busy_timeout' pragma timeout, which means that we've already waited for
+            // some amount of time before even getting here.
+            std::this_thread::sleep_for(5ms);
```
timwoj, line 346: I don't know if this even needs to wait and try again at all. If another process is holding the transaction, it means they're handling expiration already. It might be fine for the other processes to just return here.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -120,29 +127,62 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
          util::fmt("select value_str from %s where key_str=?", table_name.c_str()),
          util::fmt("delete from %s where key_str=?", table_name.c_str()),
          util::fmt("delete from %s where expire_time > 0 and expire_time != 0 and expire_time <= ?",
-                   table_name.c_str())};
+                   table_name.c_str()),
+         util::fmt("select last_run from zeek_expire_ctrl where ukey = '%s'", table_name.c_str()),
+         util::fmt("update zeek_expire_ctrl set last_run = ? where ukey = '%s'", table_name.c_str())};
 
-    std::array<unique_stmt_ptr, 5> stmt_ptrs;
+    std::array<unique_stmt_ptr, 7> stmt_ptrs;
```
awelzel: Does this work here?

```suggestion
    std::array<unique_stmt_ptr, statements.size()> stmt_ptrs;
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -208,42 +253,38 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
     if ( ! key_data )
         return {ReturnCode::SERIALIZATION_FAILED, "Failed to serialize key"};
 
-    sqlite3_stmt* stmt;
+    unique_stmt_ptr stmt;
     if ( ! overwrite )
-        stmt = put_stmt.get();
+        stmt = unique_stmt_ptr(put_stmt.get(), sqlite3_reset);
```
awelzel, line 294: I like the technique. "unique_ptr" to me quickly is free/delete. Would it be clearer to have the type be aliased as `auto_reset_stmt_ptr` or something?

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
 void SQLite::DoExpire(double current_network_time) {
-    auto stmt = unique_stmt_ptr(expire_stmt.get(), sqlite3_reset);
+    int status;
+    char* errMsg = nullptr;
+    unique_stmt_ptr stmt;
+
+    // Begin an exclusive transaction here to lock the database for this one process. That
+    // will ensure there isn't a TOCTOU bug in the time check below.
+    status = SQLITE_BUSY;
+    while ( status == SQLITE_BUSY ) {
```
awelzel: This runs every 15 seconds on every node in a cluster? Is it even worth retrying here. Also relates to your busy_timeout comment below. Seems okay to just return with busy if it actually happens.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+        status = sqlite3_exec(db, "begin exclusive transaction", nullptr, nullptr, &errMsg);
```
awelzel: Could this be "immediate"?

> EXCLUSIVE is similar to IMMEDIATE in that a write transaction is started immediately. EXCLUSIVE and IMMEDIATE are the same in [WAL mode](https://www.sqlite.org/wal.html), but in other journaling modes, EXCLUSIVE prevents other database connections from reading the database while the transaction is underway.

Seems okay to allow readers regardless of WAL mode or not (I guess we use that as default, so :man_shrugging: ). Most likely most of the time all entries in the table will not have expired, so there's no real reason to prevent concurrent readers.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+                        current_network_time - last_run, zeek::BifConst::Storage::expire_interval);
+
+                if ( current_network_time > 0 &&
+                     (current_network_time - last_run) < zeek::BifConst::Storage::expire_interval ) {
+                    sqlite3_exec(db, "rollback TRANSACTION", nullptr, nullptr, &errMsg);
+                    return;
+                }
+            }
+            else if ( status == SQLITE_BUSY )
+                // I honestly don't know if we need to wait here. This should fall under
```
awelzel: Can this even happen? Didn't this process just start a transaction?

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+        status = sqlite3_exec(db, "begin exclusive transaction", nullptr, nullptr, &errMsg);
+        if ( status == SQLITE_OK )
+            break;
+        else if ( status == SQLITE_BUSY )
+            // I honestly don't know if we need to wait here. This should fall under the
+            // 'busy_timeout' pragma timeout, which means that we've already waited for
+            // some amount of time before even getting here.
+            std::this_thread::sleep_for(5ms);
+        else if ( status != SQLITE_BUSY )
+            return;
```
awelzel, line 390: That error is probably interesting? How about logging it? Also think a plain `else {`  is sufficient here.

src/storage/backend/sqlite/SQLite.cc
awelzel: Hmm, also, this `db` here is the same `db` that's used by Zeek's main thread, correct? Given this runs in a separate thread, is it actually safe to use `db` here?

> Multi-thread. In this mode, SQLite can be safely used by multiple threads provided that no single [database connection](https://www.sqlite.org/c3ref/sqlite3.html) nor any object derived from database connection, such as a [prepared statement](https://www.sqlite.org/c3ref/stmt.html), is used in two or more threads at the same time.

https://www.sqlite.org/threadsafe.html

I wonder if we need a separate connection here.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+        }
+
+        status = sqlite3_step(stmt.get());
+        if ( status != SQLITE_ROW && status != SQLITE_DONE ) {
+            std::string err = util::fmt("Error updating expiration control time: %s", errMsg);
+            DBG_LOG(DBG_STORAGE, "%s", err.c_str());
+            Error(err.c_str());
+
+            sqlite3_exec(db, "rollback transaction", nullptr, nullptr, &errMsg);
```
awelzel: Would the rollback make sense in some "Deferred" construct that does it in the destructor automatically and possibly commit if it's marked specially at the end?

I had used something poor-man in ZeroMQ:

https://github.com/zeek/zeek/blob/31f51f7a871ea6ceaa770a8b0c9c8c02ec2c497d/src/cluster/backend/zeromq/ZeroMQ.cc#L570-L578

Maybe the C++ gurus can decide if something like that makes sense in util.h down the road :-)

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,18 +328,98 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+            // I honestly don't know if we need to wait here. This should fall under the
+            // 'busy_timeout' pragma timeout, which means that we've already waited for
+            // some amount of time before even getting here.
+            std::this_thread::sleep_for(5ms);
+        else if ( status != SQLITE_BUSY )
+            return;
+    }
+
+    // We don't need to check for contention if we're running under a single process.
+    if ( is_cluster ) {
```
awelzel: Does it hurt if we do anyhow? I'd rather cover this code even when running single-process and it's potential interesting debugging information, too.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -120,29 +127,56 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
+        std::string err = util::fmt("Error executing table creation statement: (%d) %s", res, errorMsg);
+        Error(err.c_str());
+        sqlite3_free(errorMsg);
+        Close(nullptr);
+        return {ReturnCode::INITIALIZATION_FAILED, std::move(err)};
+    }
+
+    // Create a table for controlling expiration contention. The ukey column here ensures that only
+    // one row exists for this backend's table.
+    cmd = util::fmt("create table if not exists zeek_expire_ctrl (ukey primary key, last_run double);");
```
awelzel: Maybe `zeek_expire_runs` rather than `ctrl` ? And maybe even `zeek_storage_expire_runs` for some more collision prevention with user tables.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -77,6 +79,11 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
     sqlite3_enable_shared_cache(1);
 #endif
 
+    // Store whether we're in a cluster. Knowing this means we can simplify other parts of the code.
+    auto local_node_name = id::find_val("Cluster::node")->AsStringVal();
+    if ( local_node_name->Len() > 0 )
+        is_cluster = true;
```
awelzel: If we don't need this one, that'd be great - see other comment about `is_cluster`.

awelzel (commented):
Hey - lots of small comments about various things. Mostly about cluster vs non-cluster, but also thread safety.

Keeping the expiration time in some table seems okay, but I'm also wondering if there'd actually be an issue if all nodes in a cluster just try to do the expiration. With an index on the expiration time, it should be cheap to figure out if anything even needs to be expired. Nodes already read the meta table in the database with the proposed approach. More, the proposal requires a write to a table even if nothing had to be expired. With otherwise zero activity, a rotating disk might need to spin-up every 15 seconds to flush the writes to disk, meaning poor battery life :-)

src/storage/backend/sqlite/SQLite.cc
timwoj: Huh, I didn't think that would actually work. I thought it had to be a constant value at compilation time. Obviously `statements.size()` would be constant since it's also a `std::array`, but I'm surprised the compiler sees it that way.

src/storage/backend/sqlite/SQLite.cc
timwoj: It's also used for the `sqlite_finalize` versions, so `auto_reset` wouldn't match.

src/storage/backend/sqlite/SQLite.cc
timwoj: That's kind of where I was leaning as well. One of the nodes is going to open the transaction, the others are going to be flagged as busy, and they can just bail out.

src/storage/backend/sqlite/SQLite.cc
timwoj: No harm in always doing it. Saves having to lookup the cluster state too (though it'd be good to have a utility method in `cluster::manager` that does that too).

src/storage/backend/sqlite/SQLite.cc
timwoj: > Maybe the C++ gurus can decide if something like that makes sense in util.h down the road

Done :)

I left the commit as a separate call.

src/storage/backend/sqlite/SQLite.cc
timwoj: I opted for `expiry` instead of `expire` and fixed other naming to match.

src/storage/backend/sqlite/SQLite.cc
awelzel: Fair. Resolving, I think it's okay to grok when working on the code in earnest.

src/util.h
```diff
@@ -662,5 +662,17 @@ inline std::vector<std::wstring_view> split(const wchar_t* s, const wchar_t* del
...
+/**
+ * Helper class that runs a function at destruction.
+ */
+class Deferred {
+public:
+    Deferred(std::function<void()> deferred) : closer(std::move(deferred)) {}
+    ~Deferred() { closer(); }
+
+private:
+    std::function<void()> closer;
```
awelzel: Ah, nit, sorry, "closer" I think comes from the "socket closer function" in ZeroMQ, maybe call that member `deferred`, too?

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -279,36 +314,109 @@ OperationResult SQLite::DoErase(ResultCallback* cb, ValPtr key) {
...
+
+            if ( current_network_time > 0 &&
+                 (current_network_time - last_run) < zeek::BifConst::Storage::expire_interval )
+                return;
+        }
+        else
+            return;
+    }
+
+    // Update the expiration control key
```
awelzel, line 434: Still not sure if we should really do this regularly. My comment about a disk spinning up was very tongue-in-cheek, but I do think it is worth considering completely idle systems (QA appliances, non-connected production appliances or standby-systems, virtual machines, maybe idle Windows systems with Zeek embedded, ...). It is quite unfortunate that having an SQLite backend open would mean a database on disk is written to every 15 seconds.

Not stubborn, but it is important enough to me to mention it one more time :-) Feel free to resolve, obviously.

awelzel (approved):

src/util.h
timwoj: Yeah, I just copied and pasted it as a whole 😛

src/storage/backend/sqlite/SQLite.cc
timwoj: Maybe make it check to see if there's anything to expire before updating? Create the transaction, check the number that need expiring, then do the rest if there's things to expire?

src/storage/backend/sqlite/SQLite.cc
timwoj: We open the connection with SQLITE_OPEN_FULLMUTEX, so it should be safe in multi-threaded contexts:

> [SQLITE_OPEN_FULLMUTEX](https://www.sqlite.org/c3ref/c_open_autoproxy.html)
>    The new database connection will use the "serialized" [threading mode](https://www.sqlite.org/threadsafe.html). This means the multiple threads can safely attempt to use the same database connection at the same time. (Mutexes will block any actual concurrency, but in this mode there is no harm in trying.)

It's possible that [SQLITE_OPEN_NOMUTEX](https://www.sqlite.org/c3ref/c_open_autoproxy.html) is faster. We use that version for all of the other sqlite connections in Zeek.

timwoj: I swapped it over to `NOMUTEX` with a second connection and it seems fine locally.

awelzel: > We open the connection with SQLITE_OPEN_FULLMUTEX, so it should be safe in multi-threaded contexts:

Ah, sorry, I missed that. Either would've worked for me :-)

awelzel (dismissed):

awelzel (approved):

PR #4509, opened 2025-05-28
Bump Spicy to pull in fix.

### LINKED ISSUE #4501: Spicy parser in`&eod` parsing raises error on teardown
I would expect a parser which extracts `bytes` with `&eod` to gracefully finish parsing. Instead an error is raised when its fiber is shut down.

```
# @TEST-EXEC: spicyz -d foo.spicy foo.evt -o foo.hlto
# @TEST-EXEC: zeek -Cr ${TRACES}/http/206_example_b.pcap foo.hlto "Spicy::enable_print=T" >output 2>&1
#
# @TEST-START-FILE foo.spicy
module foo;

public type X = unit {
    : bytes &eod;

    on %error(e: string) {
        print "ERROR: %s" % e;
    }
};

# @TEST-END-FILE

# @TEST-START-FILE foo.evt
import foo;

protocol analyzer X over TCP:
    parse with foo::X,
    port 0/tcp-65535/tcp;
# @TEST-END-FILE
```

Before https://github.com/zeek/zeek/commit/3362d44e0c178d3df3fec25a6ae3cbaf9f5f6035 (#4015) which landed for zeek-7.1.0 this would show no error, but from that commit on it shows `ERROR: std::exception` (I opened https://github.com/zeek/spicy/pull/2073 to improve that error message).

If we fix this we should backport the fix to `release/7.2`.

rsmmr:
Includes a new regression test.

Closes #4501.


testing/btest/spicy/tcp-eod-behavior-on-destroy.zeek
```diff
@@ -0,0 +1,39 @@
...
+
+public type X = unit {
+    : bytes &eod;
+
+    on %init() {
+        print "INIT", zeek::conn_id(), zeek::is_orig();
+    }
+
+    on %done() {
+        print "DONE", zeek::conn_id(), zeek::is_orig(); # should not be called for instance not reaching &eod
```
bbannier, line 22: Ups, didn't realize this PCAP had an incomplete flow when I picked it for the reproducer, I guess that is a useful feature to test, even though not every `INIT` having a `DONE` is pretty subtle given the number of frames in that file. Nothing to do.

PR #4524, opened 2025-05-30
Redis: Add support for sending AUTH commands during connection

### LINKED ISSUE #4350: Redis backend doesn't support password-protected or ACL-protected servers
The Redis storage backend needs to support a password option in `Storage::BackendOptions`, as well as pass the `AUTH` command after connection succeeds.

timwoj:
The primary change here is to support password- and ACL-protected Redis servers, and includes a test that starts one with a simple password. There's a few other bug fixes included as well.

Fixes #4350


src/storage/storage-sync.bif
```diff
@@ -33,7 +33,7 @@ function Storage::Sync::__open_backend%(btype: Storage::Backend, options: any, k
 
 	// If the backend only supports async, block until it's ready and then pull the result out of
 	// the callback.
-	if ( ! b.value()->SupportsSync() ) {
+	if ( op_result.code == ReturnCode::IN_PROGRESS && ! b.value()->SupportsSync() ) {
```
awelzel: Should this also be done for all the other operations? (close, get, put, erase)? Might be worth a local  `wait_for_result(...)` helper?

src/storage/backend/redis/Redis.cc
```diff
@@ -678,8 +683,13 @@ void Redis::OnDisconnect(int status) {
     else {
         --active_ops;
 
-        EnqueueBackendLost("Client disconnected");
-        CompleteCallback(close_cb, {ReturnCode::SUCCESS});
+        if ( disconnect_reason.empty() )
+            EnqueueBackendLost("Client disconnected");
+        else
+            EnqueueBackendLost(util::fmt("Client disconnected: %s", disconnect_reason.c_str()));
```
awelzel: Nit - if it's just about formatting, IMO there should be just a single enqueue call.

```
            std::string msg = util::fmt("Client disconnected%s%s", disconnect_reason.size() > 0 ? ": " : "", disconnect_reason.c_str());
            EnqueueBackendLost(msg);

```

src/storage/backend/redis/Redis.cc
```diff
@@ -630,8 +630,13 @@ void Redis::HandleInfoResult(redisReply* reply) {
...
     }
 
-    if ( ! connected && res.err_str.empty() )
-        res.err_str = "INFO command did not return server version";
+    if ( ! connected ) {
+        if ( res.err_str.empty() )
+            res.err_str = "INFO command did not return server version";
+
+        disconnect_reason = res.err_str;
+        redisAsyncDisconnect(async_ctx);
```
awelzel, line 659: Nit, It's a bit weird to initiate a disconnection in an if branch that says `! connected`. The `! connected` here means "connected and server determined to have a compatible version"? Maybe rename to `ready` ?

scripts/policy/frameworks/storage/backend/redis/main.zeek
```diff
@@ -39,6 +39,13 @@ export {
 		## Timeout for operation requests sent to the backend. Operations that
 		## exceed this time will return :zeek:see:`Storage::TIMEOUT`.
 		operation_timeout: interval &default=default_operation_timeout;
+
+		## A username to use for authentication the server is protected by an ACL.
+		acl_username: string &optional;
```
awelzel: `redis-cli`  only uses  `--user`, rather than something like `--acl-user`, so suggest to drop the `acl_` prefix.

scripts/policy/frameworks/storage/backend/redis/main.zeek
```diff
@@ -39,6 +39,13 @@ export {
...
+
+		## A username to use for authentication the server is protected by an ACL
+		## or by a simple password.
+		password: string &optional;
```
awelzel, line 48: Only food for thought - not sure this is worth it: Is there a use for `&hidden` or `&password` to `***` out such fields during plain prints of records?

src/storage/backend/redis/Redis.cc
```diff
@@ -642,25 +663,70 @@ void Redis::HandleInfoResult(redisReply* reply) {
...
-            return;
-        }
+        // If the username and/or password are set, send an AUTH command. Fail to
+        // connect if the authentication fails. We want to pause here while opening.
+        if ( ! password.empty() ) {
+            if ( username.empty() )
+                status = redisAsyncCommand(async_ctx, redisAUTH, NULL, "AUTH %s", password.c_str());
+            else
+                status =
+                    redisAsyncCommand(async_ctx, redisAUTH, NULL, "AUTH %s %S", username.c_str(), password.c_str());
```
awelzel: Is "%S" somehow special, or typoed?

Hmm, would this be a bit easier as something like?
```
if ( username.size() > 0 && password.size() > 0 )
   status =  redisAsyncCommand(async_ctx, redisAUTH, NULL, "AUTH %s %S", username.c_str(), password.c_str() /*use %s also for password?*/);
else if ( password.size() > 0 )
  status =  ...
else
  status = SendInfoRequest()

if ( status == REDIS_ERR ) {
    std::string msg = fmt("Connect failed: %s", async_ctx->err_str);
    CompleteCallback(open_cb, OPERATION_FAILED, ...)
} else {
  ++active_opts;
}
```

...and then adapt `redisAUTH` to handle the `SendInfo()` error instead of in `SendInfo()` itself?

src/storage/backend/redis/Redis.cc
```diff
@@ -642,25 +663,70 @@ void Redis::HandleInfoResult(redisReply* reply) {
...
 
-        ++active_ops;
+
+            if ( status == REDIS_ERR ) {
+                DBG_LOG(DBG_STORAGE, "AUTH command failed: %s err=%d", async_ctx->errstr, async_ctx->err);
+
+                CompleteCallback(open_cb, {ReturnCode::OPERATION_FAILED,
+                                           util::fmt("AUTH command failed to authenticate: %s", async_ctx->errstr)});
```
awelzel: Missing return here? This would do the increment `++active_ops` below.

src/storage/storage-sync.bif
timwoj: Yeah, probably a good idea to check. The only reason I noticed this one was I forgot to set a port in my environment before starting up, and it crashed.

timwoj: One bug here though is in the case of `DoPut`:

1. We queue a `SET` with the server
2. We're reading traces, so we attempt to queue a `ZADD`, but it fails to queue so it doesn't get queued and we return `OPERATION_FAILED`.
3. There's still the active `SET` operation result sitting waiting to be handled, because we didn't call `Poll` like we would have if we weren't checking for `IN_PROGRESS`. That wouldn't be handled until the next operation.
4. By that point, the callback for the put operation was deleted in the bif and ASan throws a use-after-free.

scripts/policy/frameworks/storage/backend/redis/main.zeek
timwoj: I based that naming on the fact that you don't set a username unless you're using an ACL. For regular password mode, you don't pass a username at all. It actually does underneath, it's just masked from the client and the default user is passed, but we don't explicitly send one.

src/storage/backend/redis/Redis.cc
timwoj: > Is "%S" somehow special, or typoed?

typo.

> Hmm, would this be a bit easier as something like?

Yeah. I was trying to be too cute and only check `password.empty` once.

src/storage/backend/redis/Redis.cc
timwoj: I'm indifferent 🤷🏼

awelzel: Then nothing to do. Resolving.

scripts/policy/frameworks/storage/backend/redis/main.zeek
awelzel: There won't ever be a non-ACL username, right? Then I'd just call it `username`, otherwise the non-symmetry with `password` is surprising.

src/storage/storage-sync.bif
awelzel: I guess for the `SET` and `ZADD` combination you might actually want to put them into a transaction then and treat the whole transaction as a single op?

https://redis.io/docs/latest/develop/interact/transactions/

src/storage/backend/redis/Redis.cc
```diff
@@ -626,32 +651,86 @@ void Redis::HandleInfoResult(redisReply* reply) {
...
+                                       util::fmt("AUTH command failed to queue: %s", async_ctx->errstr)});
             return;
         }
 
-        ++active_ops;
+        if ( made_auth_request )
+            ++active_ops;
+        else
+            // This will be called from the handler of the auth event if that succeeds.
+            SendInfoRequest();
```
awelzel, line 733: As in the orig proposal, I think I'd still have made that just the trailing `else {` branch and adapt the SendInfoRequest() implementation, but okay as is, too.

src/storage/backend/redis/Redis.cc
awelzel: >> Is "%S" somehow special, or typoed?

> typo.

Your latest push still has the uppercase `%S`. Maybe worth a test? :-)

PR #4525, opened 2025-05-30
Clang-tidy Part 3: `cppcoreguidelines` fixes

timwoj:
This is the third in a series of PRs to fix a large number of clang-tidy findings. This one fixes findings from a few of the `cppcoreguidelines` checkers. A lot of them are very noisy or are things that we simply don't/shouldn't care to do. I left a list of some that we could possibly investigate later in the last commit.

### DIFF
--- a/src/analyzer/protocol/login/NVT.cc
+++ b/src/analyzer/protocol/login/NVT.cc
@@ -11,27 +11,22 @@
 
 #define IS_3_BYTE_OPTION(c) ((c) >= 251 && (c) <= 254)
 
-#define TELNET_OPT_SB 250
-#define TELNET_OPT_SE 240
+static constexpr uint8_t TELNET_OPT_SB = 250;
+static constexpr uint8_t TELNET_OPT_SE = 240;
 
-#define TELNET_OPT_IS 0
-#define TELNET_OPT_SEND 1
+static constexpr uint8_t TELNET_OPT_IS = 0;
+static constexpr uint8_t TELNET_OPT_SEND = 1;
 
-#define TELNET_OPT_WILL 251
-#define TELNET_OPT_WONT 252
-#define TELNET_OPT_DO 253
-#define TELNET_OPT_DONT 254
+static constexpr uint8_t TELNET_OPT_WILL = 251;
+static constexpr uint8_t TELNET_OPT_WONT = 252;
+static constexpr uint8_t TELNET_OPT_DO = 253;
+static constexpr uint8_t TELNET_OPT_DONT = 254;
 
-#define TELNET_IAC 255
+static constexpr uint8_t TELNET_IAC = 255;
 
 namespace zeek::analyzer::login {
 
-TelnetOption::TelnetOption(NVT_Analyzer* arg_endp, unsigned int arg_code) {
-    endp = arg_endp;
-    code = arg_code;
-    flags = 0;
-    active = 0;
-}
+TelnetOption::TelnetOption(NVT_Analyzer* arg_endp, unsigned int arg_code) : endp(arg_endp), code(arg_code) {}
 
 void TelnetOption::RecvOption(unsigned int type) {
     TelnetOption* peer = endp->FindPeerOption(code);
@@ -114,15 +109,17 @@ void TelnetTerminalOption::RecvSubOption(u_char* data, int len) {
     endp->SetTerminal(data + 1, len - 1);
 }
 
-#define ENCRYPT_SET_ALGORITHM 0
-#define ENCRYPT_SUPPORT_ALGORITHM 1
-#define ENCRYPT_REPLY 2
-#define ENCRYPT_STARTING_TO_ENCRYPT 3
-#define ENCRYPT_NO_LONGER_ENCRYPTING 4
-#define ENCRYPT_REQUEST_START_TO_ENCRYPT 5
-#define ENCRYPT_REQUEST_NO_LONGER_ENCRYPT 6
-#define ENCRYPT_ENCRYPT_KEY 7
-#define ENCRYPT_DECRYPT_KEY 8
+enum EncryptOptions : uint8_t {
+    ENCRYPT_SET_ALGORITHM = 0,
+    ENCRYPT_SUPPORT_ALGORITHM = 1,
+    ENCRYPT_REPLY = 2,
+    ENCRYPT_STARTING_TO_ENCRYPT = 3,
+    ENCRYPT_NO_LONGER_ENCRYPTING = 4,
+    ENCRYPT_REQUEST_START_TO_ENCRYPT = 5,
+    ENCRYPT_REQUEST_NO_LONGER_ENCRYPT = 6,
+    ENCRYPT_ENCRYPT_KEY = 7,
+    ENCRYPT_DECRYPT_KEY = 8,
+};
 
 void TelnetEncryptOption::RecvSubOption(u_char* data, int len) {
     if ( ! active ) {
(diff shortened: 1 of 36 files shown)


src/analyzer/protocol/login/NVT.cc
```diff
@@ -11,18 +11,20 @@
...
-#define TELNET_OPT_DO 253
-#define TELNET_OPT_DONT 254
+    TELNET_OPT_WILL = 251,
+    TELNET_OPT_WONT = 252,
+    TELNET_OPT_DO = 253,
+    TELNET_OPT_DONT = 254,
 
-#define TELNET_IAC 255
+    TELNET_IAC = 255,
+};
```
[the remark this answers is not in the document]
timwoj: Good point. I'll switch them back over to `constexpr` constants.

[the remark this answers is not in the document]
timwoj: > Stuff like TELNET_OPT_WILL is a command, not an option

It's handled by `RecvOption`, so I assume it must be an option?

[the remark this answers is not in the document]
timwoj: There's also the `TELNET_OPTION` constants in NVT.h, which are handled by `NVT_Analyzer::FindOption`. Why the heck is all of this spread all over the place?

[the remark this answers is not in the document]
timwoj: I reworked all of the constants here a bit. See if that makes more sense.

src/analyzer/protocol/login/NVT.cc
```diff
@@ -114,15 +109,17 @@ void TelnetTerminalOption::RecvSubOption(u_char* data, int len) {
...
-#define ENCRYPT_SUPPORT_ALGORITHM 1
-#define ENCRYPT_REPLY 2
-#define ENCRYPT_STARTING_TO_ENCRYPT 3
-#define ENCRYPT_NO_LONGER_ENCRYPTING 4
-#define ENCRYPT_REQUEST_START_TO_ENCRYPT 5
-#define ENCRYPT_REQUEST_NO_LONGER_ENCRYPT 6
-#define ENCRYPT_ENCRYPT_KEY 7
-#define ENCRYPT_DECRYPT_KEY 8
+enum EncryptOptions : uint8_t {
+    ENCRYPT_SET_ALGORITHM = 0,
```
awelzel, line 113: These are spec'ed protocol values, right? Please keep explicitly numbering them. 

Only `ENCRYPT_STARTING_TO_ENCRYPT` and `ENCRYPT_REQUEST_START_TO_ENCRYPT` are used in the code below. Previously one could just read their values from the list if that was of interest, now I find it rather obfuscated due to the need to count. Also, if I know I want to handle `6`, which constant should I pick - this is unnecessary hard now :-)

Similarly for AuthOptions.

src/analyzer/protocol/login/NVT.cc
```diff
@@ -412,17 +411,16 @@ void NVT_Analyzer::DeliverChunk(int& len, const u_char*& data) {
         if ( binary_mode && c != TELNET_IAC )
             c &= 0x7f;
 
-#define EMIT_LINE                                                                                                      \
-    {                                                                                                                  \
-        buf[offset] = '\0';                                                                                            \
-        ForwardStream(offset, buf, IsOrig());                                                                          \
-        offset = 0;                                                                                                    \
-    }
+        auto EMIT_LINE = [this]() {
```
awelzel: How about open-coding that one? It's only 3 places and there's so many state variables involved, that macro made it harder before and the lambda doesn't improve anything here IMO.

Indeed, I think I'd prefer to keep the macro just due to "okay, it's just inserting some code in verbatim here", rather than, okay, we're calling this lambda that's capturing "this". Also, the stack is going to contain a huge `operator()` or `std::function` call/dispatch of that lambda here in flamegraphs or backtraces.

src/analyzer/protocol/netbios/NetbiosSSN.cc
```diff
@@ -10,14 +10,15 @@
...
 
-constexpr double netbios_ssn_session_timeout = 15.0;
-
-#define MAKE_INT16(dest, src)                                                                                          \
-    (dest) = *(src);                                                                                                   \
-    (dest) <<= 8;                                                                                                      \
-    (src)++;                                                                                                           \
-    (dest) |= *(src);                                                                                                  \
-    (src)++;
+static constexpr double netbios_ssn_session_timeout = 15.0;
```
awelzel, line 13: Would you be up for making this a script-level variable in a `const.bif` somewhere and accessing it with `BifConst::` ? I don't think that's a lot of work and might add a bit of value to someone out there ;-)

src/analyzer/protocol/netbios/NetbiosSSN.cc
```diff
@@ -10,14 +10,15 @@
...
+
+static constexpr void MAKE_INT16(uint16_t& dest, const u_char*& src) {
```
awelzel, line 15: This looks like (untested)
```
dest = ntohs(*reinterpret_cast<const uint16_t*>(src));
src += 2;
```

And I wonder if dropping `MAKE_INT16` completely and replacing the callers would be reasonable. But only do that if we have tests and you think it's okay, too.

src/analyzer/protocol/rpc/Portmap.cc
```diff
@@ -5,12 +5,14 @@
...
 #include "zeek/analyzer/protocol/rpc/events.bif.h"
 
-#define PMAPPROC_NULL 0
-#define PMAPPROC_SET 1
-#define PMAPPROC_UNSET 2
-#define PMAPPROC_GETPORT 3
-#define PMAPPROC_DUMP 4
-#define PMAPPROC_CALLIT 5
+enum PortmapperProcs : uint8_t {
+    PMAPPROC_NULL = 0,
```
awelzel, line 9: If those are protocol values, do number them (IMO).

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -52,6 +52,8 @@ enum class InprocTag : uint8_t {
 
 constexpr DebugFlag operator&(uint8_t x, DebugFlag y) { return static_cast<DebugFlag>(x & static_cast<uint8_t>(y)); }
 
+// NOLINTBEGIN(cppcoreguidelines-macro-usage)
```
awelzel, line 55: Hmm, thanks. Does that one or the ones in serializer and websocket upset you or seem weird? Or do they seem reasonable?

src/iosource/Manager.cc
```diff
@@ -19,7 +19,7 @@
 #include "zeek/iosource/PktSrc.h"
 #include "zeek/plugin/Manager.h"
 
-#define DEFAULT_PREFIX "pcap"
+static constexpr char DEFAULT_PREFIX[] = "pcap";
```
awelzel: I wonder if you could just hard-code this in the static `split_prefix()` function. It's the only user as far as I can see and this is all within `Manager.cc` anyhow. A docstring on split_prefix that it returns "pcap" seems more helpful than having this be a `static constexpr`.

awelzel (commented):
I had a few opinionated thoughts that go a bit further than just replacing. Mind checking?

src/analyzer/protocol/netbios/NetbiosSSN.cc
timwoj: I'll make a note to do it separately to this PR.

src/analyzer/protocol/netbios/NetbiosSSN.cc
timwoj: According to Coveralls, we have roughly zero test coverage on this analyzer.

src/cluster/backend/zeromq/ZeroMQ.cc
timwoj: I think they're reasonable. These ones are notable because they use varargs, so it's easier to just mask the error than it is to rework all of it.

Issue #4221, opened 2025-02-05
Add script function to list available tags in a namespace


timwoj:
This came up as part of the storage framework PR: https://github.com/zeek/zeek/pull/4163#discussion_r1936233591. It would be nice to have a script-level function that lists the available tags for a specific namespace (Analyzer, Storage, etc). Otherwise you have to run `zeek -NN` and know what lines to parse out to determine what tags are available.

bhaskarbhar:
Outputs should look like this?

set["SQLITE", "REDIS"]
@timwoj

bhaskarbhar:
@timwoj Will work on this then.

bhaskarbhar:
@timwoj I applied function to list out backend tags in storage but then i went through your issue again, seems like a generic function to me. Am I right?

bhaskarbhar:
@timwoj Can you close the issue as this has been resolved. PR : https://github.com/zeek/zeek/pull/4604

awelzel:
@bhaskarbhar - thank you for your contribution!

PR #4538, opened 2025-06-03
Update zeek.bif

bhaskarbhar:
Issue: #4221 

I did make clean then ./configure && make && make install.
Then made a new zeek file to test.

Input: Analyzer

test.zeek:
```
event zeek_init()
{
    local tags = get_tags_by_category("Analyzer");
    for ( i in tags )
        print tags[i];
}
```

Output:
image

Input: Storage Backend
test.zeek:
```
event zeek_init()
{
    local tags = get_tags_by_category("Storage Backend");
    for ( i in tags )
        print tags[i];
}
```
Output:
image

This is the best I can do for now. Thank you.

### DIFF
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -5615,3 +5615,48 @@ function find_in_zeekpath%(p: string%): string
 
 	return zeek::make_intrusive<zeek::StringVal>(resolved.c_str());
 	%}
+
+# Function: get_tags_by_category
+# Description:
+#   Runs a shell command `zeek -NN` filtered by the given category using grep,
+#   and returns a vector of strings containing matching tags.
+#
+# Parameters:
+#   category (string) - The category name to filter tags by.
+#
+# Returns:
+#   vector<string> - A vector containing all matching tag strings.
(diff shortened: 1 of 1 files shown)


timwoj:
I'm sorry, but this is the definitely not the right way to go about this. There are APIs internal to Zeek to get the list of tags without resorting to executing Zeek via a system call and scraping the output.

Look primarily at plugin/Manager.h, which keeps a list of all of the loaded plugins/components. Every loaded analyzer, etc is technically a plugin including the internal ones, so they should all be available from the plugin manager. You should be able to go from there through `Plugin` -> `Component` -> `Tag`, and filtering down to the ones you want, or from there directly to the `Component`s if you know the type.

As an aside, your `test.zeek` file would be good to add as a btest for coverage. Look in testing/btest for how that works.

Let us know if you have any other questions.

bhaskarbhar:
@timwoj Thank you. You cleared my confusion. Will now work on your suggestion. Will do a fresh PR. Closing this PR.

PR #4539, opened 2025-06-04
Fix `&ordered` attribute not preserved in table initializer assignments

### LINKED ISSUE #4189: Consider `&ordered` for initialization and redefinitions

### LINKED ISSUE #4448: `&ordered` on tables works with table constructors but not with table initializers

AmazingPP:
Fixes #4448 
Fixes #4189  

This PR ensure that variable attributes (e.g., `&ordered`) are correctly passed when constructing `AssignExpr` for global variable initialization using literal table initializers. Previously, attributes were omitted during assignment expression creation, causing `&ordered` tables initialized via `{ ... }` to not behave as expected. This change passes `id->GetAttrs()` into the `AssignExpr` to correctly propagate attributes.


awelzel (approved):
Thanks!

PR #4540, opened 2025-06-04
Parallelize coverage/bare-mode-errors

0xxon:
Currently, coverage/bare-mode-errors is one of the slowest tests in the entire test suite. This is caused by the fact that it has to repeatedly launch Zeek for every script that we ship. This is done sequentially.

This commit changes this test to use xargs to spawn 20 parallell processes.

This admittedly doesn't feel entirely optimal, as it is a hardcoded value. I might say that this is better than the status quo, assuming that this passes on cirrus :)

### DIFF
--- a/testing/btest/coverage/bare-mode-errors.test
+++ b/testing/btest/coverage/bare-mode-errors.test
@@ -9,5 +9,5 @@
 # @TEST-REQUIRES: have-zeromq
 #
 # @TEST-EXEC: test -d $DIST/scripts
-# @TEST-EXEC: for script in `find $DIST/scripts/ -name \*\.zeek`; do zeek -b --parse-only $script >>errors 2>&1; done
+# @TEST-EXEC: find $DIST/scripts/ -name "*.zeek" -print0 | xargs -0 -n1 -P 20 -- zeek -b --parse-only >>errors 2>&1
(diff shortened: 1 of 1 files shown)


testing/btest/coverage/bare-mode-errors.test
```diff
@@ -9,5 +9,5 @@
 # @TEST-REQUIRES: have-zeromq
 #
 # @TEST-EXEC: test -d $DIST/scripts
-# @TEST-EXEC: for script in `find $DIST/scripts/ -name \*\.zeek`; do zeek -b --parse-only $script >>errors 2>&1; done
+# @TEST-EXEC: sh -c 'find $DIST/scripts/ -name "*.zeek" -print0 | xargs -0 -n1 -P 20 -- zeek -b --parse-only' >>errors 2>&1
```
bbannier: Hardcoding 20 seems fine to me since this does not seem to be CPU bound or at least starts very short processes so overhead from launching dominates.

Is there a reason you start a subshell? I believe BTest already defaults to using `sh`.

```suggestion
# @TEST-EXEC: find $DIST/scripts/ -name "*.zeek" -print0 | xargs -0 -n1 -P 20 -- zeek -b --parse-only >>errors 2>&1
```

bbannier (approved):

testing/btest/coverage/bare-mode-errors.test
0xxon: For some reason I thought it didn't work without the sh. I must have had another error in there when testing that.

Issue #4521, opened 2025-05-30
x509: Certificate before/after dates in logs are TZ dependent


awelzel:
After #4494 and chatting a bit with Justin. The timestamp values in x509.log today are `TZ` sensitive. Running Zeek with `TZ` set to Berlin or Los Angeles produces different contents.

```
$ export TZ=Europe/Berlin; zeek -C -r ../../testing/btest/Traces/tls/tls12-decryption.pcap LogAscii::use_json=T; jq < x509.log > x509.log.${TZ/\//_}
$ export TZ=America/Los_Angeles; zeek -C -r ../../testing/btest/Traces/tls/tls12-decryption.pcap LogAscii::use_json=T; jq < x509.log > x509.log.${TZ/\//_}
$ diff x509.log.America_Los_Angeles x509.log.Europe_Berlin 
8,9c8,9
<   "certificate.not_valid_before": 1616445135.0,
<   "certificate.not_valid_after": 1624221135.0,
---
>   "certificate.not_valid_before": 1616416335.0,
>   "certificate.not_valid_after": 1624188735.0,
29,30c29,30
<   "certificate.not_valid_before": 1602123700.0,
<   "certificate.not_valid_after": 1632968500.0,
---
>   "certificate.not_valid_before": 1602091300.0,
>   "certificate.not_valid_after": 1632936100.0,
$ 
```

This isn't great. The timestamp in the x509.log is a UNIX epoch timestamp and so the value should not differ, regardless in which timezone Zeek runs in. This probably also is the cause of:

> the x509 log, where sometimes certs are set to expire 9999-12-31T23:59, and due to timezone normalization this turns into something like 10000-01-01:04:59 which can break elastic.

0xxon:
I was curious to see why our current tests do not catch this. The (kind of obivous) reason is that we remove essentially all timestamps from all our logs by default. This, in turn, makes a couple of specific tests that wre supposed to check timestamp output useless. To give an example, we have a test that checks a 1999 certificate for the correct output - which then is discarded due to canonification :)

I wonder if, for a (potentially significant) subset of tests, it would make sense to mostly remove the start/end timestamps in the logfiles, instead of all timestamps.

PR #4541, opened 2025-06-04
Change x509 not_before/not_after to not be based on local timezone

0xxon:
Not the not_before/not_after fields output GMT based times.

Fixes GH-4521


awelzel:
Thanks! Would you mind adding the commands from #4521 as some sort of btest that if we ever touch the timestamp parsing again at least there's a bit of isolated/specific coverage?


EDIT: Think this is worth backporting to 7.0 ?

0xxon:
I talked about that a little bit in #4521. Let me see if I can just exempt some of the btests from the typical timestamp removal.

0xxon:
Oh, and for backporting. The behaviour has been like this for years - probably since we started being able to parse certificates.

As such this is not a regression - and I think we can just pull it into the next release without backporting. Changing the behaviour at a release boundary feels nicer to me.

NEWS
```diff
@@ -156,6 +156,10 @@ Changed Functionality
 - `detect-protocol.zeek was the last non-deprecated policy script left in
   `frameworks/dpd`.  It was moved to `frameworks/analyzer/detect-protocol.zeek`.
 
+- The `not_valid_before` and `not_valid_after` times are now logged as GMT timestamps.
+  Before, the times were logged as local times; thus the output was dependent on the
+  timezone that your system is set to.
```
awelzel: Hmm, I guess we actually parsed them into the wrong timezone? That is, already the timestamp fields on the `X509::Certificate` field were off. I'm not sure we need to talk about that, but maybe there's someone out there using `Certificate$not_valid_before` directly and they'll now see a behavior change.

awelzel (approved):

NEWS
0xxon: If the local timezone was not GMT, we did; I updates the NEWS entry. If this looks good to you I will merge this to master.

NEWS
```diff
@@ -162,6 +162,12 @@ Changed Functionality
 - Running Zeek with Zeekygen for documentation extraction (-X|--zeekygen
   <cfgfile>) now implies -a, i.e., parse-only mode.
 
+- The `not_valid_before` and `not_valid_after` times of X509 certificates are
+  now logged as GMT timestamps. Before, the  were logged as local times; thus
+  the output was dependent on the timezone that your system is set to.
+  Similarly, the related events and the Zeek data structures all interpreted
+  times in X509 certificates as local-times.
```
```suggestion
  times in X509 certificates as local times.
```

NEWS
```diff
@@ -162,6 +162,12 @@ Changed Functionality
 - Running Zeek with Zeekygen for documentation extraction (-X|--zeekygen
   <cfgfile>) now implies -a, i.e., parse-only mode.
 
+- The `not_valid_before` and `not_valid_after` times of X509 certificates are
+  now logged as GMT timestamps. Before, the  were logged as local times; thus
```
```suggestion
  now logged as GMT timestamps. Before, they were logged as local times; thus
```

awelzel (approved):
:rocket:

PR #4542, opened 2025-06-04
Clang-tidy Part 4: `modernize` fixes

timwoj:
This is the fourth (and final?) in a series of PRs to fix a large number of clang-tidy findings. This one fixes findings from a majority of the `modernize` checkers. The last commit in the PR adds some exclusions for checkers that we might want to enable later, with explanations as to why they're disabled.


src/IP.cc
```diff
@@ -457,8 +457,8 @@ static inline bool isIPv6ExtHeader(uint8_t type) {
 }
 
 IPv6_Hdr_Chain::~IPv6_Hdr_Chain() {
-    for ( size_t i = 0; i < chain.size(); ++i )
-        delete chain[i];
+    for ( auto& c : chain )
+        delete c;
```
awelzel, line 461: We could make `chain` become `std::vector<IPv6_Hdr>`, to avoid some extra allocs, deletes and indirections.

This seems to compile - would you be up for that? Otherwise I might make it a PR.

[ip-chain-hdr.diff.txt](https://github.com/user-attachments/files/20606286/ip-chain-hdr.diff.txt)

src/PolicyFile.cc
```diff
@@ -121,8 +121,8 @@ bool LoadPolicyFileText(const char* policy_filename, const std::optional<std::st
         }
     }
 
-    for ( int i = 0; i < int(pf->lines.size()); ++i )
-        assert(pf->lines[i][0] != '\n');
+    for ( const auto& l : pf->lines )
+        assert(l[0] != '\n');
```
awelzel, line 125: This does nothing when compiled NDEBUG - maybe can drop it?

I'm also looking at the loop above and wonder if this isn't `util::tokenize_string(pf->filedata, '\n')`. But, `src/PolicyFile.h` seems to be for this debugger thing, so probably better to just leave and move on. I do have in the back of my mind whether it would be at  all controversial to rip out the existing debugger code and maybe look at something like https://microsoft.github.io/debug-adapter-protocol/ instead.

src/RandTest.cc
```diff
@@ -35,9 +36,7 @@ RandTest::RandTest() {
     inmont = mcount = 0;
     cexp = montex = montey = montepi = sccu0 = scclast = scct1 = scct2 = scct3 = 0.0;
 
-    for ( int i = 0; i < 256; i++ ) {
-        ccount[i] = 0;
-    }
+    memset(ccount, 0, sizeof(ccount));
```
awelzel: Or change to `ccount = {0}` in the member declaration, like already done for `monte` ?

src/RuleMatcher.cc
```diff
@@ -1183,8 +1183,8 @@ static bool val_to_maskedval(Val* v, maskedvalue_list* append_to, vector<IPPrefi
                 v->AsSubNet().Prefix().GetBytes(&n);
                 v->AsSubNetVal()->Mask().CopyIPv6(m);
 
-                for ( unsigned int i = 0; i < 4; ++i )
```
awelzel, line 1186: For some reason I resonate more with the index'ed loop (there's also one in IP that only goes to 4), as it is immediately clear that the compiler will likely just unroll it, something the ` : m` iteration hides a bit.

Nothing to do and no objection, just putting it out there after thinking it twice now.

src/Stats.cc
```diff
@@ -194,12 +194,12 @@ void ProfileLogger::Log() {
     file->Write(util::fmt("%0.6f Threads: current=%zu\n", run_state::network_time, thread_mgr->NumThreads()));
 
     const threading::Manager::msg_stats_list& thread_stats = thread_mgr->GetMsgThreadStats();
-    for ( threading::Manager::msg_stats_list::const_iterator i = thread_stats.begin(); i != thread_stats.end(); ++i ) {
-        threading::MsgThread::Stats s = i->second;
+    for ( const auto& ts : thread_stats ) {
+        threading::MsgThread::Stats s = ts.second;
```
awelzel: Switch to structured binding while at it, or that happens in a later commit? :sweat_smile:

src/Trigger.cc
```diff
@@ -457,9 +457,8 @@ void Manager::Process() {
     TriggerList tmp;
     pending = &tmp;
 
-    for ( TriggerList::iterator i = orig->begin(); i != orig->end(); ++i ) {
-        Trigger* t = *i;
-        (*i)->Eval();
+    for ( auto& t : *orig ) {
```
awelzel: `auto&` -> `auto*`

src/ZeekString.cc
```diff
@@ -352,8 +352,8 @@ String::Vec* String::VecFromPolicy(VectorVal* vec) {
 char* String::VecToString(const Vec* vec) {
     std::string result("[");
 
-    for ( String::VecCIt it = vec->begin(); it != vec->end(); ++it ) {
-        result += (*it)->CheckString();
+    for ( const auto& str : *vec ) {
```
awelzel: `const auto&` -> `const auto*`

src/ZeekString.cc
```diff
@@ -352,8 +352,8 @@ String::Vec* String::VecFromPolicy(VectorVal* vec) {
 char* String::VecToString(const Vec* vec) {
```
awelzel, line 352: Just because I looked, these three methods on String appear to never be called within the tree, should we deprecate them?
```
    // Helper functions for vectors:
    static VectorVal* VecToPolicy(Vec* vec);
    static Vec* VecFromPolicy(VectorVal* vec);
    static char* VecToString(const Vec* vec);
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1372,8 +1372,8 @@ bool DNS_Interpreter::ParseRR_A(detail::DNS_MsgInfo* msg, const u_char*& data, i
 bool DNS_Interpreter::ParseRR_AAAA(detail::DNS_MsgInfo* msg, const u_char*& data, int& len, int rdlength) {
     uint32_t addr[4];
 
-    for ( int i = 0; i < 4; ++i ) {
-        addr[i] = htonl(ExtractLong(data, len));
+    for ( uint32_t& a : addr ) {
+        a = htonl(ExtractLong(data, len));
```
awelzel: Repeat from before - I'm quite unsure if for parsing code and fixed limits the `for ( uint32_t& a : addr )` is better - a mutable reference being assigned hides things for me :-( Anyone else has an opinion?

src/analyzer/protocol/krb/krb-analyzer.pac
```diff
@@ -307,27 +306,15 @@ refine connection KRB_Conn += {
...
 						break;
-					default:
-						break;
-					}
-				}
-
-			if ( timestamp_i )
-				rv->Assign(4, GetTimeFromAsn1(${msg.safe_body.args[timestamp_i].args.timestamp}, timestamp_usecs));
-
-			for ( uint i = 0; i < ${msg.safe_body.args}->size(); ++i )
```
awelzel, line 324: Curios!

src/input/Manager.cc
```diff
@@ -1439,8 +1439,8 @@ int Manager::SendEventStreamEvent(Stream* i, EnumVal* type, const Value* const*
 
     if ( convert_error ) {
         // we have an error somewhere in our out_vals. Just delete all of them.
-        for ( list<Val*>::const_iterator it = out_vals.begin(), end = out_vals.end(); it != end; ++it )
-            Unref(*it);
+        for ( auto& val : out_vals )
```
awelzel: `auto*`

src/input/Manager.cc
```diff
@@ -1712,8 +1712,8 @@ void Manager::SendEvent(EventHandlerPtr ev, list<Val*> events) const {
     DBG_LOG(DBG_INPUT, "SendEvent with %" PRIuPTR " vals (list)", events.size());
 #endif
 
-    for ( list<Val*>::iterator i = events.begin(); i != events.end(); i++ )
-        vl.emplace_back(AdoptRef{}, *i);
+    for ( const auto& val : events )
```
awelzel: `const auto*`

src/logging/writers/ascii/Ascii.cc
```diff
@@ -255,52 +255,52 @@ bool Ascii::InitFilterOptions() {
     const WriterInfo& info = Info();
 
     // Set per-filter configuration options.
-    for ( WriterInfo::config_map::const_iterator i = info.config.begin(); i != info.config.end(); ++i ) {
-        if ( strcmp(i->first, "tsv") == 0 ) {
-            if ( strcmp(i->second, "T") == 0 )
+    for ( const auto& [key, value] : info.config ) {
```
awelzel, line 258: I think the bigger bang for the buck would be to switch `info.config` to `std::map<std::string, std::string>` at some point to ditch all the `strcmp` calls :-/

src/plugin/Manager.cc
```diff
@@ -452,10 +451,9 @@ void Manager::InitPreScript() {
     // order.
     ActivePluginsInternal()->sort(plugin_cmp);
 
-    for ( plugin_list::iterator i = Manager::ActivePluginsInternal()->begin();
-          i != Manager::ActivePluginsInternal()->end(); i++ ) {
-        (*i)->InitializeComponents();
-        (*i)->InitPreScript();
+    for ( Plugin* plugin : (*Manager::ActivePluginsInternal()) ) {
```
awelzel: No need for the extra parens also elsewhere in the file.

src/threading/Manager.cc
```diff
@@ -140,16 +140,16 @@ void Manager::Terminate() {
 
     // Signal all to stop.
 
-    for ( all_thread_list::iterator i = all_threads.begin(); i != all_threads.end(); i++ )
-        (*i)->SignalStop();
+    for ( const auto& t : all_threads )
```
awelzel: `const auto*` - here and elsewhere in this file as these are all raw pointers, the `&` is confusing.

src/zeekygen/Configuration.cc
```diff
@@ -72,18 +72,15 @@ Config::Config(const string& arg_file, const string& delim)
 }
 
 Config::~Config() {
-    for ( size_t i = 0; i < targets.size(); ++i )
-        delete targets[i];
+    std::for_each(targets.begin(), targets.end(), [](auto& t) { delete t; });
```
awelzel: We barely use `std::for_each`, so wonder we should stick to `for ( auto *t : targets ) delete t` loop, even if it's not as fancy.

src/zeekygen/Manager.cc
```diff
@@ -93,8 +93,7 @@ Manager::Manager(const string& arg_config, const string& command)
 }
 
 Manager::~Manager() {
-    for ( size_t i = 0; i < all_info.size(); ++i )
-        delete all_info[i];
+    std::for_each(all_info.begin(), all_info.end(), [](auto& info) { delete info; });
```
awelzel: Same as above - use a classic for-loop, `std::for_each()` usage increased by >200% with this PR :-)

I'm not against `std::for_each()`, just that we almost never seem to use it, so wonder why we should do it here now.

src/analyzer/Analyzer.cc
```diff
@@ -116,8 +116,8 @@ Analyzer::~Analyzer() {
     assert(finished);
     assert(new_children.empty());
 
-    LOOP_OVER_CHILDREN(i)
```
awelzel, line 119: Oh, thank you!

I see you kept the `LOOP_OVER_CHILDREN` macros in Analyzer.h.

Can we remove them, or are you worried they are part of the API? If they are part of the API, maybe we could change the expansion such that a deprecation warning is produced?

Maybe something like this? https://stackoverflow.com/a/2685696

src/script_opt/ZAM/ZInst.cc
```diff
@@ -26,7 +26,7 @@ void ZInst::Dump(FILE* f, const string& prefix, const string& id1, const string&
                  const string& id4) const {
     fprintf(f, "%s ", ZOP_name(op));
     // fprintf(f, "(%s) ", op_type_name(op_type));
-    if ( t && 0 )
+    if ( t && false )
```
awelzel, line 29: :man_shrugging: - this seems like it should be some environment variable or setting...

src/cluster/serializer/broker/Serializer.cc
```diff
@@ -85,7 +85,7 @@ std::optional<broker::zeek::Event> detail::to_broker_event(const detail::Event&
                 broker::vector entry(2);
                 entry[0] = static_cast<broker::count>(m.Id());
                 entry[1] = res.value();
-                broker_meta.push_back(std::move(entry));
+                broker_meta.emplace_back(std::move(entry));
```
awelzel, line 86: Oh oh, I might have missed a cleanup suggestion from Benjamin - mind changing this to:

```
                broker::vector entry{{static_cast<broker::count>(m.Id()), res.value()}};
                broker_meta.emplace_back(std::move(entry));
```

awelzel (approved):
I've put a few suggestions and thoughts on the modernize-loop-convert findings and random comments here and there, but otherwise :shipit:  and thanks a lot :-)

src/cluster/serializer/broker/Serializer.cc
bbannier: Or even better `broker::vector entry{static_cast<broker::count>(m.Id()), res.value()};` unless we wanted to mix uniform initialization and explicitly creating an initializer list where `std::vector` can already directly construct from an initializer list ;)

src/RandTest.cc
timwoj: It'd be better to just move all of this to the header.

src/IP.cc
timwoj: Separate PR is fine with me. I can do it after I merge this one.

src/analyzer/protocol/krb/krb-analyzer.pac
timwoj: I have no idea why there were two loops that ran over the same array.

src/analyzer/Analyzer.cc
timwoj: I'm worried they're part of the API. I'll take a look at deprecating them though. Probably part of a separate PR though.

src/script_opt/ZAM/ZInst.cc
timwoj: I asked @vpax. I'll let him update what to do with it.

src/PolicyFile.cc
timwoj: I've wondered for a long time about how used this debugger code is and whether it's a whole lot of complication for nothing.

timwoj: I opened https://github.com/zeek/zeek/issues/4553

src/script_opt/ZAM/ZInst.cc
vpax: Fine to remove

PR #4543, opened 2025-06-04
Enhancement/Added_bif_get_tags_by_category

bhaskarbhar:
Issue: #4221 
New BIF added to zeek.bif.

BIF function: get_tags_by_category

Sample usage:
```
event zeek_init()
    {
    local result = get_tags_by_category("ANALYZER");

    for (i in result)
        print result[i];
    }
```
Output:
image


bhaskarbhar:
@timwoj I will upload the test script for this once you approve the above, otherwise I have to write test script again if the above is not as per your requirement.

bbannier:
Recreated by OP as #4543.

src/script_opt/FuncInfo.cc
```diff
@@ -20,29 +20,29 @@ namespace zeek::detail {
 // to the event engine.
 
 // Does not change script-level state (though may change internal state).
-constexpr unsigned int ATTR_NO_SCRIPT_SIDE_EFFECTS = 0x1;
+#define ATTR_NO_SCRIPT_SIDE_EFFECTS 0x1
```
timwoj: Did you mean to revert these back to `#define`?

src/zeek.bif
```diff
@@ -338,6 +341,32 @@ zeek::RecordValPtr zeek::detail::build_dummy_conn_record()
...
+        {"STORAGE_SERIALIZER", zeek::plugin::component::STORAGE_SERIALIZER},
+    };
+
+    auto it = type_map.find(category);
+    if ( it != type_map.end() )
+        return it->second;
+
+    return static_cast<zeek::plugin::component::Type>(-1); // Invalid
+	}
+
```
timwoj, line 371: This would probably be better as a static method in plugin/Component.h.

src/zeek.bif
```diff
@@ -5615,3 +5644,47 @@ function find_in_zeekpath%(p: string%): string
...
+##    "FILE_ANALYZER", "IOSOURCE", "PKTSRC", "PKTDUMPER", "SESSION_ADAPTER",
+##    "CLUSTER_BACKEND", "EVENT_SERIALIZER", "LOG_SERIALIZER", "STORAGE_BACKEND",
+##    "STORAGE_SERIALIZER".
+##
+## @return
+##    A vector of strings (string_vec) with the tags of matching components.
+##    Returns null and raises an error if the category is unknown.
+function get_tags_by_category%(category: string%): string_vec
+	%{
+    std::string cat = category->CheckString();
```
timwoj, line 5656: Total nit: `.bif` files are typically indented with tabs. I know that's different from the rest of the code. It annoys me too.

src/zeek.bif
```diff
@@ -5615,3 +5644,47 @@ function find_in_zeekpath%(p: string%): string
 
 	return zeek::make_intrusive<zeek::StringVal>(resolved.c_str());
 	%}
+
+## Returns a vector of strings containing the tags of all plugin components
+## that belong to the specified category.
+##
+## @param category
```
timwoj: Docs for bifs don't follow doxygen style. Look at other methods in this file for examples of how to format it (search for `Returns:`)

src/zeek.bif
```diff
@@ -5615,3 +5644,47 @@ function find_in_zeekpath%(p: string%): string
...
+##    A string representing the category of components to filter by.
+##    Valid categories include: "READER", "WRITER", "ANALYZER", "PACKET_ANALYZER",
+##    "FILE_ANALYZER", "IOSOURCE", "PKTSRC", "PKTDUMPER", "SESSION_ADAPTER",
+##    "CLUSTER_BACKEND", "EVENT_SERIALIZER", "LOG_SERIALIZER", "STORAGE_BACKEND",
+##    "STORAGE_SERIALIZER".
+##
+## @return
+##    A vector of strings (string_vec) with the tags of matching components.
+##    Returns null and raises an error if the category is unknown.
+function get_tags_by_category%(category: string%): string_vec
```
timwoj: I'm debating whether this should return some sort of record that contains the name, tag, and enabled state instead of building up a string for each one. This is good enough for now though.

src/script_opt/FuncInfo.cc
bhaskarbhar: I see I was changing the old Function.cc, because the fork copy I was working with has #define it it, but now instead of #define it is now constexpr. I wll edit in the latest FuncInfo.cc.

src/zeek.bif
bhaskarbhar: Ok I will remove this from zeek.bif and add a static method in Component.h

src/zeek.bif
bhaskarbhar: Sure will change that.

src/zeek.bif
bhaskarbhar: Actually record will make it more usable. Working on this part too.
**Changes required till now:**

1. Edit the latest FuncInfo.cc.
2. Static Method in Component.h(Optional).
3. Change style of Comments.
4. Return as record instead of string(optional).

awelzel: Agree with @timwoj to return a vector of records. Sorry if I'm throwing in something that maybe should've been discussed/designed before - below is a sketch. Note that this certainly doesn't compile as is!
```
type Component: record {
  enabled: bool;
  name: string;
  canonical_name: string;
  tag: enum val (?)
};

type component_vec: vector of Component;

function get_components(typ: string &default="all"): component_vec
```

I'd prefer `typ` above to be some Zeek-script enum type like `Plugin::Component::Type` that mirrors `plugin::component::Type` in C++, but I guess we do not have that right now. Or maybe it's a string and we name it filter :-)

bhaskarbhar: > Agree with @timwoj to return a vector of records. Sorry if I'm throwing in something that maybe should've been discussed/designed before - below is a sketch. Note that this certainly doesn't compile as is!
> 
> ```
> type Component: record {
>   enabled: bool;
>   name: string;
>   canonical_name: string;
>   tag: enum val (?)
> };
> 
> type component_vec: vector of Component;
> 
> function get_components(typ: string &default="all"): component_vec
> ```
> 
> I'd prefer `typ` above to be some Zeek-script enum type like `Plugin::Component::Type` that mirrors `plugin::component::Type` in C++, but I guess we do not have that right now. Or maybe it's a string and we name it filter :-)

Thanks a lot, Sir! I was confused about building the record, now I got a direction to work on the improvements.

scripts/base/init-bare.zeek
```diff
@@ -3048,6 +3059,20 @@ type dns_svcb_rr: record {
 	target_name: string;	##< Target name, the hostname of the service endpoint.
 };
 
+## A NAPTR record.
```
timwoj, line 3062: Can you remove this record? It looks unrelated.

bhaskarbhar: I didnt added this record. I just updated init-bare.zeek and added my records. Also this is the comment(## A NAPTR record.) for this record dns_naptr_rr.

bhaskarbhar:
> I think this looks great now. There's some minor indentation problems and such that we can fix during the merge. Can you squash it down to a single commit?

I am having a issue with my git bash thats why I have to upload the files manually. So I guess squashing can be done by you when merging.

timwoj:
Superceded by https://github.com/zeek/zeek/pull/4604, which was a squashed and rebased version with a few formatting fixes.

PR #4544, opened 2025-06-05
docker: Add `net-tools` and `procps` dependencies

edoardomich:
Add `iproute2` and `procps` to the final.Dockerfile, to avoid warning in zeekctl commands like: "failed to find local IP addresses [...]" and the error in `zeekctl top` command caused because `top` is not installed.

### DIFF
--- a/docker/final.Dockerfile
+++ b/docker/final.Dockerfile
@@ -27,6 +27,8 @@ RUN apt-get -q update \
      libuv1 \
      libz1 \
      libzmq5 \
+     net-tools \
+     procps \
      python3-git \
      python3-minimal \
      python3-semantic-version \


bbannier:
@edoardomich this setup is somewhat minimal, and e.g., also does not include everything needed for compiling Spicy analyzers. I would suggest you install any additional dependencies in an image which derives from the image we publish.

awelzel:
> @edoardomich this setup is somewhat minimal, and e.g., also does not include everything needed for compiling Spicy analyzers. I would suggest you install any additional dependencies in an image which derives from the image we publish.

I'd be okay with adding these. There was a similar discussion with `jq` which is also part of the image. I've heard of people running `zeekctl` within containers before. I'd certainly be hesitant to add gcc, clang or dev dependencies, but being able to invoke and use `zeekctl` ro run a cluster seems reasonable. These are not huge packages.

:+1: from me

0xxon:
Same here - this seems reasonable to me.

ckreibich (changes requested):
Thanks! I'm fine with adding the tools — I think the agreement is that we ship with the tooling that our own ecosystem requires, so these should be there.

I'd like to request one change though: let's use `net-tools` (which provides `ifconfig`) instead of `iproute2` (which pulls in `ip`). The latter pulls in dependencies, whereas the former is self-contained. zeekctl works with either (see ifconfig [here](https://github.com/zeek/zeekctl/blob/master/ZeekControl/config.py#L750)).

```
# apt-get install --no-install-recommends net-tools
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following NEW packages will be installed:
  net-tools
0 upgraded, 1 newly installed, 0 to remove and 34 not upgraded.
Need to get 243 kB of archives.
After this operation, 1001 kB of additional disk space will be used.
```
```
# # apt-get install --no-install-recommends iproute2
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libbpf1 libbsd0 libcap2 libcap2-bin libelf1 libgssapi-krb5-2 libk5crypto3 libkeyutils1 libkrb5-3
  libkrb5support0 libmnl0 libssl3 libtirpc-common libtirpc3 libxtables12
Suggested packages:
  iproute2-doc python3:any krb5-doc krb5-user
Recommended packages:
  libatm1 libpam-cap krb5-locales
The following NEW packages will be installed:
  iproute2 libbpf1 libbsd0 libcap2-bin libelf1 libgssapi-krb5-2 libk5crypto3 libkeyutils1 libkrb5-3
  libkrb5support0 libmnl0 libssl3 libtirpc-common libtirpc3 libxtables12
The following packages will be upgraded:
  libcap2
1 upgraded, 15 newly installed, 0 to remove and 33 not upgraded.
Need to get 4300 kB of archives.
After this operation, 14.0 MB of additional disk space will be used.
```

edoardomich:
You are right,
I have just force-pushed the branch with the required changes. Thanks

awelzel (approved):

Issue #4465, opened 2025-05-15
ZeekJS support for Node >= 23


0xxon:
Adding here to be able to tie it to a milestone.

ZeekJS currently doesn't support node >= 23. Node 22 LTS will reach EOL during the lifetime of Zeek 8; we should support Node 24 before Zeek 8 comes out.

ZeekJS side issue: https://github.com/corelight/zeekjs/issues/94

PR #4549, opened 2025-06-05
Move to building everything with C++ 20

timwoj:
As part of https://github.com/zeek/zeek/issues/4465, we realized that building with Node v23 also requires building with C++20. This is a good opportunity to upgrade to this version of C++ across the board. This PR is the Zeek side of things, which surprisingly required very few changes. This is only the infrastructure changes and anything that was required to make everything compile and btests pass. There are no changes to use C++20 features (such as removing `zeek::Span` for `std::span`). Those will come later once this has merged.

Related PRs:
- [x] https://github.com/zeek/bifcl/pull/41
- [x] https://github.com/zeek/binpac/pull/46
- [ ] https://github.com/zeek/broker/pull/465
- [x] https://github.com/zeek/capstats/pull/4
- [x] https://github.com/zeek/cmake/pull/138
- [x] https://github.com/zeek/gen-zam/pull/19
- [x] https://github.com/zeek/paraglob/pull/25
- [x] https://github.com/zeek/pysubnettree/pull/40
- [x] https://github.com/zeek/zeek-aux/pull/66
- [x] https://github.com/zeek/zeekctl/pull/81
- [x] https://github.com/corelight/zeekjs/pull/106

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,18 +41,8 @@ set(BISON_FLAGS "--debug")
 
 if (MSVC)
     set(SIGN_COMPARE_FLAG "/wd4018")
-    if (BUILD_WITH_WERROR)
-        # TODO: This is disabled for now because there a bunch of known
-        # compiler warnings on Windows that we don't have good fixes for.
-        #set(WERROR_FLAG "/WX")
-        #set(WNOERROR_FLAG "/WX:NO")
-    endif ()
 else ()
     set(SIGN_COMPARE_FLAG "-Wno-sign-compare")
-    if (BUILD_WITH_WERROR)
-        set(WERROR_FLAG "-Werror")
-        set(WNOERROR_FLAG "-Wno-error")
-    endif ()
 endif ()
 
 # Rule parser/scanner
--- a/src/storage/Manager.h
+++ b/src/storage/Manager.h
@@ -3,9 +3,23 @@
 #pragma once
 
 #include <mutex>
-#include <thread>
 
+// Apple's clang has an implementation of std::jthread, but it's still marked
+// as experimental. Use the 3rdparty one for that platform and for any other
+// that doesn't have it. This could move to util.h if some other code needs
+// jthread in the future.
+#if defined(__APPLE__) || ! defined(__cpp_lib_jthread)
 #include "zeek/3rdparty/jthread.hpp"
+namespace zeek {
+using jthread = nonstd::jthread;
+}
+#else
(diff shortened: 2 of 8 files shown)


timwoj:
With the broker stuff mostly done and the ZeekJS changes landed in a release, I think we're just waiting on Spicy fixes for this. @bbannier says there's a PR in the works but it ended up being a bit more complicated than expected.

bbannier:
> I think we're just waiting on Spicy fixes for this. @bbannier says there's a PR in the works but it ended up being a bit more complicated than expected.

Technically you do not need to wait since the build against Spicy `main` is broken everywhere. This would still build Spicy itself with C++17 and Zeek Spicy analyzers with C++20, but this is designed to work. The only fix we needed for C++20 was https://github.com/zeek/spicy/pull/2099, but currently not Zeek Spicy analyzer uses `set` so this shouldn't matter.

The PR for switching Spicy to C++20 is https://github.com/zeek/spicy/pull/2096, but it still needs work in managing errors from GCC. The "more complicated than expected" PR is https://github.com/zeek/zeek/pull/4632 and is about getting any Zeek to build against Spicy `main`.

src/CMakeLists.txt
```diff
@@ -49,9 +49,23 @@ if (MSVC)
     endif ()
 else ()
     set(SIGN_COMPARE_FLAG "-Wno-sign-compare")
+
```
awelzel: I was a bit confused, and putting a `message(STATUS "${WERROR_FLAG}") includes the flags from the top-level CMakeLists.txt.

I haven't looked all that closely, is there a reason we need to duplicate this?

src/storage/Manager.h
```diff
@@ -99,7 +113,7 @@ class Manager final {
     friend class storage::detail::ExpirationTimer;
     void RunExpireThread();
     void StartExpirationTimer();
-    std::jthread expiration_thread;
+    zeek::jthread expiration_thread;
```
awelzel, line 116: Minor: We don't use explicit `zeek::` namespace elsewhere, but I'm actually never sure what the best pattern is to use. For that one, it probably makes sense to keep it explicit?

awelzel (approved):
:+1:  modulo setting WERROR_FLAG twice.

src/storage/Manager.h
timwoj: We did for `zeek::filesystem` in the places it's used, since we want to make sure it picks the zeek alias up instead of the other namespaces.

PR #4550, opened 2025-06-05
Follow-ups to configure output reformatting

timwoj:
This adds some post-merge suggestions from @awelzel to the changes made in https://github.com/zeek/zeek/pull/4517


CMakeLists.txt
```diff
@@ -1513,6 +1513,11 @@ endif ()
 ## string representation. This can also take an option second argument that is a
```
ckreibich:
```suggestion
## string representation. This can also take an optional second argument that is a
```

ckreibich (approved):
I really like the way this looks now, thanks for doing that.

PR #4552, opened 2025-06-06
telemetry: Rename endpoint label to node label

awelzel:
Using a label named "endpoint" is not intuitive and requires explaining to users that it's really just the Cluster::node value. Change the label to "node", so that we don't need to do the explaining.

This probably breaks some existing users of the Prometheus metrics, but after looking more at metrics recently, "endpoint" really is a thorn in my eye.


timwoj:
We really only stuck with the `endpoint` label because that's what it was in the Broker telemetry code.

timwoj (approved):

PR #4554, opened 2025-06-06
Switch IPv6_Hdr_Chain to a vector of objects instead of pointers

timwoj:
@awelzel provided this as a patch while reviewing the clang-tidy-modernize PR. This switches IPv6_Hdr_Chain from a vector of pointers to a vector of objects, to avoid the manual management of them.

### DIFF
--- a/src/IP.cc
+++ b/src/IP.cc
@@ -457,8 +457,6 @@ static inline bool isIPv6ExtHeader(uint8_t type) {
 }
 
 IPv6_Hdr_Chain::~IPv6_Hdr_Chain() {
-    for ( auto& c : chain )
-        delete c;
     delete homeAddr;
     delete finalDst;
 }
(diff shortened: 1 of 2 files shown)


src/IP.cc
```diff
@@ -481,23 +479,23 @@ void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, uint64_t total_len, bool se
...
 
-        next_type = p->NextHdr();
-        uint16_t cur_len = p->Length();
+        next_type = p.NextHdr();
+        uint16_t cur_len = p.Length();
 
         // If this header is truncated, don't add it to chain, don't go further.
         if ( cur_len > total_len ) {
-            delete p;
+            // delete p;
```
awelzel:
```suggestion
```

PR #4556, opened 2025-06-06
Change config member of WriterInfo to be a map of std::string

timwoj:
I'm putting this up just as an idea (from @awelzel's clang-tidy review), but I'm slightly concerned that this is an API change that could break external plugins. Feel free to reject this entirely.


src/logging/WriterBackend.h
```diff
@@ -50,8 +50,7 @@ class WriterBackend : public threading::MsgThread {
      * A struct passing information to the writer at initialization time.
      */
     struct WriterInfo {
-        // Structure takes ownership of these strings.
-        using config_map = std::map<const char*, const char*, util::CompareString>;
+        using config_map = std::map<std::string, std::string>;
```
awelzel, line 53: Thanks for giving this a shot!

Hmm, given that `WriterInfo` is in the signature of `DoInit()` and there are [multiple](https://github.com/SeisoLLC/zeek-kafka/blob/efb9d803625e9d43efd952d874c6f3187a009086/src/KafkaWriter.cc#L67-L75) [plugins](https://github.com/0xxon/zeek-postgresql/blob/92de3bca069d63d116a08826a280eec4a8cc1b75/src/PostgresWriter.cc#L154-L157) that use it, we can't just do it :-(

One idea would be to hold the config in a new `map<string, string> writer_config` member and populate the old `config` with the `.c_str()` version, then put a `[[deprecated]]` on `config`. That assumes that no one instantiates their own WriterInfo object (or we'll break them), but that seems less of an issue.

And getting rid of the `path` and `post_proc_func` raw pointers might be nice at the same time, too. Maybe introducing `Path()` and `PostProcFunc()` accessors for some private members after making this a class, too.

I'm a bit unclear if it's worth it, but less `util::copy_string()` calls are probably good.

0xxon: Agreed - writers are one of the kind of plugins that people commonly write; we should try to stick to our API guarantees for them :)

timwoj: That was one of my concerns with opening this in the first place 😅 

While I think this is an improvement, the chance of breaking someone somewhere is probably too high. I'm going to close this.

PR #4558, opened 2025-06-06
Move netbios_ssn_session_timeout to a script-level constant

timwoj:
This moves a hard-coded constant out of the Netbios analyzer into a constant that can be redef'd in script land. This code analyzer has almost zero test coverage, unfortunately. This PR is basically untested other than to make sure it builds and the script-land change doesn't break anything.

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -579,6 +579,10 @@ const io_poll_interval_live = 10 &redef;
 ## while testing, but should be used sparingly.
 const running_under_test: bool = F &redef;
 
+## The amount of time before a connection created by the netbios analyzer times
+## out and is removed.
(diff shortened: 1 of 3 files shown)


src/analyzer/protocol/netbios/NetbiosSSN.h
```diff
@@ -156,6 +156,7 @@ class NetbiosSSN_Analyzer final : public analyzer::tcp::TCP_ApplicationAnalyzer
     Contents_NetbiosSSN* orig_netbios;
     Contents_NetbiosSSN* resp_netbios;
     int did_session_done;
+    double session_timeout;
```
awelzel:
```suggestion
```
:man_shrugging:

awelzel (approved):
Thanks!

Issue #4559, opened 2025-06-07
Record initialization with type on LHS is 2x slower than with type on RHS


bbannier:
Zeek has [three ways](https://docs.zeek.org/en/master/script-reference/types.html#record) to initialize a record:

1. `local x: X = record($xa=1);`
2. `local x: X = [$xa=1];`
3. `local x = X($xa=1);`

I would expect all these to behave roughly the same at runtime, but it looks like for one benchmark (1) and (2) are 2x slower than (3). It seems that we prefer (2) over other variants in the implementation.

Benchmark:

```zeek
type X: record {
        xa: count;
        xb: count &optional;
};

event zeek_init()
        {
        local i = 0;
        while ( i < 5000000 )
                {
                # Uncomment lines below to run benchmarks for different cases, 
                # or leave commented out for iteration baseline.

                # local x: X = record($xa=1);  # (1)
                # local x: X = [$xa=1];        # (2)
                # local x = X($xa=1);          # (3)
                i += 1;
                }
        }
```

```console
$ hyperfine -L F 0,1,2,3 'zeek -b {F}.zeek'
Benchmark 1: zeek -b 0.zeek
  Time (mean ± σ):     569.1 ms ±  59.8 ms    [User: 540.6 ms, System: 27.7 ms]
  Range (min … max):   519.3 ms … 701.2 ms    10 runs

Benchmark 2: zeek -b 1.zeek
  Time (mean ± σ):      1.530 s ±  0.149 s    [User: 1.489 s, System: 0.038 s]
  Range (min … max):    1.432 s …  1.867 s    10 runs

Benchmark 3: zeek -b 2.zeek
  Time (mean ± σ):      1.565 s ±  0.134 s    [User: 1.526 s, System: 0.038 s]
  Range (min … max):    1.434 s …  1.807 s    10 runs

Benchmark 4: zeek -b 3.zeek
  Time (mean ± σ):     994.9 ms ±  31.1 ms    [User: 961.3 ms, System: 32.6 ms]
  Range (min … max):   954.2 ms … 1027.5 ms    10 runs

Summary
  zeek -b 0.zeek ran
    1.75 ± 0.19 times faster than zeek -b 3.zeek
    2.69 ± 0.38 times faster than zeek -b 1.zeek
    2.75 ± 0.37 times faster than zeek -b 2.zeek
```

JustinAzoff:
Interesting, see https://github.com/zeek/zeek/issues/425

vpax:
[...]

PR #4560, opened 2025-06-09
Prefer explicit construction to coercion in record initialization

bbannier:
While we support initializing records via coercion from an expression list, e.g.,

    local x: X = [$x1=1, $x2=2];

this can sometimes obscure the code to readers, e.g., when assigning to value declared and typed elsewhere. The language runtime has a similar overhead since instead of just constructing a known type it needs to check at runtime that the coercion from the expression list is valid; this can be slower than just writing the readible code in the first place, see #4559.

With this patch we use explicit construction, e.g.,

    local x = X($x1=1, $x2=2);


scripts/base/files/pe/main.zeek
```diff
@@ -60,13 +60,13 @@ const pe_mime_types = { "application/x-dosexec" };
 event zeek_init() &priority=5
 	{
 	Files::register_for_mime_types(Files::ANALYZER_PE, pe_mime_types);
-	Log::create_stream(LOG, [$columns=Info, $ev=log_pe, $path="pe", $policy=log_policy]);
+	Log::create_stream(LOG, Log::Stream($columns=Info, $ev=log_pe, $path="pe", $policy=log_policy));
```
awelzel, line 63: I understand the motivation, but IMO there's also a style/history and maybe eco-system consistency aspect to this (edit: I actually didn't understand the motivation - thanks for providing it :-) )

I cannot picture the performance characteristics of the record construction making a difference for `Log::create_stream()` and instead it looks a bit alien at first and this is changing a long established pattern. The `[...]` looked a bit like key-word args, which isn't exactly true, but looked nice to my eyes.

I can see arguments for always using the explicit version looking more consistent, but that doesn't seem worth the churn (totally IMO and hopefully someone else chimes in). The performance argument definitely doesn't work for `Log::create_stream()`.

bbannier: Completely agree on the "established pattern" part, but it is a slippery slope. It might be clear today what the type of a function argument actually is (and callers using coercion from `[$..]` are strongly coupled to it anyway, nothing dynamic here), but if it isn't figuring out the arg type requires two hops (find function declaration, find arg type), so I personally find being explicit strictly better. 

I think being explicit _everywhere_ makes it much easier to decide what to use when writing new code as we wouldn't need to think about what is "obvious".

I ended up here via hover and completion for record initializers in https://github.com/bbannier/zeek-language-server, and figuring out the target type of the coercion for `[$..]` is a lot of work in most situations, and I didn't implement it for function args 😬.

Curious how others feel about this.

J-Gras: > Curious how others feel about this.

Fwiw, I like the change to explicit notation 🙂

scripts/base/frameworks/logging/writers/ascii.zeek
```diff
@@ -7,9 +7,9 @@
 ##! names is printed out as meta information, with no "# fields" prepended; no
 ##! other meta data gets included in that mode.  Example filter using this::
 ##!
-##!    local f: Log::Filter = [$name = "my-filter",
+##!    local f = : Log::Filter($name = "my-filter",
```
evantypanski: This `:` should be removed?

```suggestion
##!    local f = Log::Filter($name = "my-filter",
```

scripts/base/frameworks/reporter/main.zeek
```diff
@@ -40,20 +40,20 @@ export {
 
 event zeek_init() &priority=5
 	{
-	Log::create_stream(Reporter::LOG, [$columns=Info, $path="reporter", $policy=log_policy]);
+	Log::create_stream(Reporter::LOG, Log::Stream($columns=Info, $path="reporter", $policy=log_policy));
```
evantypanski, line 43: The parts that use `Log::write` get rid of the `Reporter`, but not this. Is there a reason for that? Also applies elsewhere

EG later in this file:

``` diff
-	Log::write(Reporter::LOG, [$ts=t, $level=INFO, $message=msg, $location=location]);
+	Log::write(LOG, Info($ts=t, $level=WARNING, $message=msg, $location=location));
```

scripts/base/protocols/ftp/main.zeek
```diff
@@ -306,14 +306,19 @@ event ftp_request(c: connection, command: string, arg: string) &priority=5
...
+										$resp_h=data$h,
+										$resp_p=data$p));
+				}
 		else
-			{
-			# TODO: raise a notice?  does anyone care?
-			}
+				{
+				# TODO: raise a notice?  does anyone care?
+				}
```
evantypanski: This extra indentation seems wrong. There are a few places this happens, where parts are indented one further than they should, I won't comment on them all. Including `smb/files.zeek`, `http/files.zeek`, `ssl/files.zeek`, etc.

evantypanski (commented):
I'm personally a fan of this, I think if the source shows this then we should update the package templates, but I don't think that's a problem personally

I left some random comments that I found scrolling through, may need some manual style touchups in places

scripts/base/frameworks/reporter/main.zeek
bbannier: I didn't intend to remove the namespace below, will undo it.

scripts/base/protocols/ftp/main.zeek
bbannier: Urg, that's what I get for messing with the beautiful handcrafted indention. I went over all changes and made this more consistent. My primary intention was to get rid of indention with spaces by just breaking lines. I wasn't consistently doing this everywhere, and also sometimes accidentally used two tabs where one was already more than enough.

scripts/policy/frameworks/analyzer/debug-logging.zeek
```diff
@@ -69,8 +69,13 @@ export {
 
 event zeek_init() &priority=5
 	{
-	Log::create_stream(LOG, [$columns=Info, $path="analyzer_debug", $policy=log_policy,
-	                         $event_groups=set("Analyzer::DebugLogging")]);
+		Log::create_stream(
```
evantypanski: forgive me: indent here seems one further than it should

scripts/policy/frameworks/analyzer/detect-protocols.zeek
```diff
@@ -130,10 +131,14 @@ function do_notice(c: connection, a: AllAnalyzers::Tag, d: dir)
...
-				c$id$resp_p, (known ? " (update)" : "")),
-			$p=c$id$resp_p, $sub=s, $conn=c, $src=c$id$resp_h]);
+			NOTICE(Notice::Info(
+						$note=Server_Found,
+						$msg=fmt(
+							"%s: %s server on port %s%s",
+							c$id$resp_h, s,
+							c$id$resp_p,
+							(known ? " (update)" : "")),
+						$p=c$id$resp_p, $sub=s, $conn=c, $src=c$id$resp_h));
```
evantypanski: (one further indent again)

timwoj:
I'm a fan of it as well, if only for the performance side. I don't think we should break that you can do it the other way, but I do think we should encourage the method that's faster.

0xxon:
I like the change.

However, I don't really like the change in indentation/spacing. I get that it's easier to do it that way - but a lot of the scripts were using "tabs for indentation, spaces for alignment" before, which with this change just changes to "one more tab for alignment" - which in my opinion looks strictly worse.

That being said - I might be pretty alone with that opinion. We should probably just decide on one way or another.

rsmmr:
Torn. I'd be fine adding the new syntax, seems pretty natural/common and potentially useful in some contexts. However, I don't think I'd switch the style across all our scripts, that feels unnecessarily intrusive and would indeed be breaking with a style that has been in use for a long time. I'd change it only at places where there's a good reason for it--say, performance, even though, per Arne, that doesn't seem to be the case really. (I understand that wouldn't help with the LSP server ... :) )

Btw, one alternative syntax idea would be `local x = X[$x1=1, $x2=2];` Not sure if Iike that more or less.

scripts/base/frameworks/broker/log.zeek
```diff
@@ -87,10 +83,13 @@ event Broker::error(code: ErrorCode, msg: string)
 	ev = subst_string(ev, "_", "-");
 	ev = to_lower(ev);
 
-	Log::write(Broker::LOG, [$ts = network_time(),
-	           $ev = ev,
-	           $ty = ERROR,
-	           $message = msg]);
+	Log::write(
```
ckreibich: This is an example of where I'd rather go light, i.e.:
```suggestion
	Log::write(Broker::LOG, Info($ts = network_time(),
	           $ev = ev,
	           $ty = ERROR,
	           $message = msg));
```

scripts/base/frameworks/cluster/main.zeek
```diff
@@ -652,7 +652,12 @@ function create_store(name: string, persistent: bool &default=F): Cluster::Store
 
 function log(msg: string)
 	{
-	Log::write(Cluster::LOG, [$ts = network_time(), $node = node, $message = msg]);
+	Log::write(
```
ckreibich: Here too, smaller change:
```suggestion
	Log::write(Cluster::LOG, Info($ts=network_time(), $node=node, $message=msg));
```
etc.

scripts/base/frameworks/config/input.zeek
```diff
@@ -40,14 +40,15 @@ event zeek_init() &priority=5
 		return;
 
 	for ( fi in config_files )
-		Input::add_table([$reader=Input::READER_CONFIG,
+		Input::add_table(Input::TableDescription(
```
ckreibich: Yep! I think this is neat.

scripts/base/frameworks/broker/log.zeek
bbannier: My intention was to just locally and consistently tweak formatting due to the newly appearing type identifiers, but I didn't anticipate how much you all cared about the formatting particularities, and in addition to the pretty mechanical type expansions I probably spent 5x on tweaking formatting for feedback in this review.

This doesn't seem like a good use of everyone's time, so I backed out any real changes to e.g., line breaks and instead concentrated on just inserting the types and adjusting horizontal spacing to whatever local rule the touched code followed (there is no absolute consistency), and would propose to leave any further tweaks to follow-ups by people who care. I would have loved to use the rules you had encoded in `zeek-format`, but unfortunately line breaks are one area where it currently has its biggest gaps, e.g., this code would be formatted as

```zeek
	Log::write(Broker::LOG, Info($ts=network_time(), $ev=ev, $ty=ERROR,
		$message=msg));
```

I hope with that we can concentrate on the changes I was actually after.

scripts/base/frameworks/cluster/main.zeek
bbannier: Ditto.

rsmmr:
Taking another look, I like one type of changes, but not the other:

I like this:

>	\- const no_filter: Filter = [$name="<not found>"];
>	\+ const no_filter = Filter($name="<not found>");

But I don't really like:

>	\- Log::create_stream(NetControl::LOG, [$columns=Info, $ev=log_netcontrol, $path="netcontrol", $policy=log_policy]);
>	\+ Log::create_stream(NetControl::LOG, Log::Stream($columns=Info, $ev=log_netcontrol, $path="netcontrol", $policy=log_policy));

Feels like it doesn't buy much, and just makes it longer to read and type (in particular if LSP provides the type for the argument).

Matter of taste, so take it as a data point.

bbannier:
> ... significant subset that we likely all agree looks strictly better, namely this ...
> 
> ```
> const no_filter = Filter($name="<not found>");
> ```
> 
> ... instead of this:
> 
> ```
> const no_filter: Filter = [$name="<not found>"];
> ```
> 
> But here I wonder whether this trades one advantage for the language server with a disadvantage, namely that it now needs to infer the type of the variable from the thing assigned to it?

If you want I can make that

```zeek
const no_filter: Filter = Filter($name="<not found>");
```

I didn't do that since it felt redundant, and we also have many instances like the following in the code, all with `local` variables:

```zeek
local res = Management::Result($reqid = reqid, $node = Cluster::node);
local stdout = Queue::init(Queue::Settings($max_len = Management::Supervisor::output_max_lines));
local sn = Supervisor::NodeConfig($name=epi$id, $bare_mode=T, ..
local c = OpenFlow::Controller($state=OpenFlow::ControllerState($log_dpid=dpid, $log_success_event=success_event), ..
..
```

The `no_filter` snippet is the only instance I can see where we declare a `const`/`global` of record type with explicitly mentioned type, but we have instances where we already now omit it, e.g., this snippet attributed to you:

```zeek
global broker_backpressure_disconnects_cf = Telemetry::register_counter_family([...
```

In my mental model of execution this makes no real difference at runtime or might be even better; before we'd "dynamically" coerce an expression list to some typed identifier on the LHS, while with this patch we'd have an identifier who's type is dictated by the fully typed expression on the RHS.

All that said, the changes to variable declarations here where mostly done for consistency. The real benefit from explicit record construction to _readability_ should be in function calls where it is now possible to directly see argument types at the call location instead of having to inspect the function declaration as well. I can do that automatically in zeek-language-server, but that doesn't help when e.g., reviewing code where one does not have access to it.

* * *
> I like this:
> 
>     - const no_filter: Filter = [$name=""];
>     + const no_filter = Filter($name="");
> 
> But I don't really like:
> 
>     - Log::create_stream(NetControl::LOG, [$columns=Info, $ev=log_netcontrol, $path="netcontrol", $policy=log_policy]);
>     + Log::create_stream(NetControl::LOG, Log::Stream($columns=Info, $ev=log_netcontrol, $path="netcontrol", $policy=log_policy));
>
> Feels like it doesn't buy much, and just makes it longer to read and type (in particular if LSP provides the type for the argument).

Like I wrote, the second one is really the crux of this change while the first one is for consistency. I would argue that it improves readability since one can now reason locally (e.g., here _we are passing a `Log::Stream`_) while before this was just a bag of fields which magically fit the args of a function declared elsewhere.

My language server currently does not provide the type without hovering over identifiers, but I am open to patches adding it to [signature information](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#parameterInformation) which would be displayed with the call if not disabled by the user.

Screenshot 2025-07-10 at 12 11 24 PM

ckreibich:
Thanks for iterating on the indentation changes Benjamin. :+1: 

> If you want I can make that
> ```
>    const no_filter: Filter = Filter($name="<not found>");
> ```

Ah no, sorry — I meant it the other way around: I too prefer `const no_filter = Filter($name="");`. I just wasn't sure whether that actually provides the language server any benefit, which I thought was the original motivator for the PR, but which I think I misread. It looks like these make up the bulk of the changes, so I think that's great.

I'm also still in favor of the changes, overall. There clearly are cases that are debatable, but such is the nature of coding styles, I'd really rather not discuss on a case-by-case basis, and we have bigger things to work on. If we ever make to a point where a code-formatter reigns supreme, these conversations will go away, too.

For now I'm seeing a few instances that have inconsistent tab/space indentation mixes still, so I'm going to flag these in a review. Other than those, count me in.

scripts/base/frameworks/broker/log.zeek
```diff
@@ -87,10 +87,10 @@ event Broker::error(code: ErrorCode, msg: string)
 	ev = subst_string(ev, "_", "-");
 	ev = to_lower(ev);
 
-	Log::write(Broker::LOG, [$ts = network_time(),
-	           $ev = ev,
-	           $ty = ERROR,
-	           $message = msg]);
+	Log::write(Broker::LOG, Info($ts = network_time(),
```
ckreibich, line 90: Inconsistent indentation here,

scripts/base/frameworks/notice/main.zeek
```diff
@@ -381,16 +381,16 @@ function log_mailing_postprocessor(info: Log::RotationInfo): bool
...
+	Log::create_stream(Notice::ALARM_LOG, Log::Stream($columns=Notice::Info, $path="notice_alarm", $policy=log_policy_alarm));
 	# If Zeek is configured for mailing notices, set up mailing for alarms.
 	# Make sure that this alarm log is also output as text so that it can
 	# be packaged up and emailed later.
 	if ( ! reading_traces() && mail_dest != "" )
 		Log::add_filter(Notice::ALARM_LOG,
-		    [$name="alarm-mail", $path="alarm-mail", $writer=Log::WRITER_ASCII,
-		     $interv=24hrs, $postprocessor=log_mailing_postprocessor]);
+		    Log::Filter($name="alarm-mail", $path="alarm-mail", $writer=Log::WRITER_ASCII,
+			            $interv=24hrs, $postprocessor=log_mailing_postprocessor));
```
ckreibich: In these two lines it looks a bit off, too.

scripts/base/frameworks/packet-filter/netstats.zeek
```diff
@@ -24,10 +24,10 @@ event net_stats_update(last_stat: NetStats)
 		{
 		local new_recvd = ns$pkts_recvd - last_stat$pkts_recvd;
 		local new_link = ns$pkts_link - last_stat$pkts_link;
-		NOTICE([$note=Dropped_Packets,
-		        $msg=fmt("%d packets dropped after filtering, %d received%s",
-		                 new_dropped, new_recvd + new_dropped,
-		                 new_link != 0 ? fmt(", %d on link", new_link) : "")]);
+		NOTICE(Notice::Info($note=Dropped_Packets,
+		                    $msg=fmt("%d packets dropped after filtering, %d received%s",
```
ckreibich, line 28: (This actually looks right in my local checkout, seems Github's highlighter is getting confused.)

scripts/base/frameworks/signatures/main.zeek
```diff
@@ -189,16 +189,16 @@ event signature_match(state: signature_state, msg: string, data: string)
...
-		                    $uid=state$conn$uid,
-		                    $src_addr=src_addr,
-		                    $src_port=src_port,
-		                    $dst_addr=dst_addr,
-		                    $dst_port=dst_port,
-		                    $event_msg=fmt("%s: %s", src_addr, msg),
-		                    $sig_id=sig_id,
-		                    $sub_msg=data];
+		local info = Info($ts=network_time(),
+		                        $note=Sensitive_Signature,
```
ckreibich: These look off (a few spaces too far in).

scripts/base/frameworks/signatures/main.zeek
```diff
@@ -290,16 +290,16 @@ event signature_match(state: signature_state, msg: string, data: string)
...
-		            $src_addr=orig,
-		            $dst_addr=resp, $sig_id=sig_id, $sig_count=vcount,
-		            $event_msg=fmt("%s different signatures triggered", vcount),
-		            $sub_msg=vert_scan_msg]);
-
-		NOTICE([$note=Multiple_Signatures, $src=orig, $dst=resp,
-		        $msg=fmt("%s different signatures triggered", vcount),
-		        $n=vcount, $sub=vert_scan_msg]);
+		           Info($ts=network_time(),
+		               $note=Multiple_Signatures,
```
ckreibich: These are off by one.

scripts/base/protocols/ftp/main.zeek
```diff
@@ -307,8 +307,8 @@ event ftp_request(c: connection, command: string, arg: string) &priority=5
 
 		if ( data$valid )
 			{
-			add_expected_data_channel(c$ftp, [$passive=F, $orig_h=id$resp_h,
-			                                  $resp_h=data$h, $resp_p=data$p]);
+			add_expected_data_channel(c$ftp, ExpectedDataChannel($passive=F, $orig_h=id$resp_h,
+			                                                              $resp_h=data$h, $resp_p=data$p));
```
ckreibich: This one's off.

scripts/base/protocols/ftp/main.zeek
```diff
@@ -403,8 +403,8 @@ event ftp_reply(c: connection, code: count, msg: string, cont_resp: bool) &prior
 			if ( code == 229 && data$h == [::] )
 				data$h = c$id$resp_h;
 
-			add_expected_data_channel(c$ftp, [$passive=T, $orig_h=c$id$orig_h,
-			                          $resp_h=data$h, $resp_p=data$p]);
+			add_expected_data_channel(c$ftp, ExpectedDataChannel($passive=T, $orig_h=c$id$orig_h,
+			                                                              $resp_h=data$h, $resp_p=data$p));
```
ckreibich: As is this.

scripts/base/protocols/smb/const-dos-error.zeek
```diff
@@ -4,129 +4,129 @@
 module SMB;
 
 redef SMB::statuses += {
-	[0x00010001] = [$id="badfunc", $desc="Incorrect function."],
```
ckreibich: Okay — in these two files I'll make an exception and suggest we skip any changes, because it's just so redundant and bloated.

scripts/base/protocols/smtp/files.zeek
```diff
@@ -40,9 +40,9 @@ function describe_file(f: fa_file): string
 
 event zeek_init() &priority=5
 	{
-	Files::register_protocol(Analyzer::ANALYZER_SMTP,
-	                         [$get_file_handle = SMTP::get_file_handle,
-	                          $describe        = SMTP::describe_file]);
+		Files::register_protocol(Analyzer::ANALYZER_SMTP,
+		                         Files::ProtoRegistration($get_file_handle = SMTP::get_file_handle,
+		                                                       $describe        = SMTP::describe_file));
```
ckreibich: Indentation's off.

scripts/base/utils/exec.zeek
```diff
@@ -142,12 +142,12 @@ event InputRaw::process_finished(name: string, source:string, exit_code:count, s
...
 	else
 		for ( read_file in pending_files[name] )
-			Input::add_event([$source=fmt("%s", read_file),
-			                  $name=fmt("%s_%s", name, read_file),
-			                  $reader=Input::READER_RAW,
-			                  $want_record=F,
-			                  $fields=FileLine,
-			                  $ev=Exec::file_line]);
+			Input::add_event(Input::EventDescription($source=fmt("%s", read_file),
+			                                                      $name=fmt("%s_%s", name, read_file),
```
ckreibich: These are off by one (too far)

scripts/base/utils/exec.zeek
```diff
@@ -169,14 +169,14 @@ function run(cmd: Command): Result
...
-	Input::add_event([$name=cmd$uid,
-	                  $source=fmt("%s |", cmd$cmd),
-	                  $reader=Input::READER_RAW,
-	                  $mode=Input::STREAM,
-	                  $fields=Exec::OneLine,
-	                  $ev=Exec::line,
-	                  $want_record=F,
-	                  $config=config_strings]);
+	Input::add_event(Input::EventDescription($name=cmd$uid,
+	                                                      $source=fmt("%s |", cmd$cmd),
```
ckreibich: Also off.

scripts/policy/frameworks/analyzer/debug-logging.zeek
```diff
@@ -69,8 +69,8 @@ export {
 
 event zeek_init() &priority=5
 	{
-	Log::create_stream(LOG, [$columns=Info, $path="analyzer_debug", $policy=log_policy,
-	                         $event_groups=set("Analyzer::DebugLogging")]);
+	Log::create_stream(LOG, Log::Stream($columns=Info, $path="analyzer_debug", $policy=log_policy,
+	                                                $event_groups=set("Analyzer::DebugLogging")));
```
ckreibich: Indentation's off

scripts/policy/frameworks/intel/seen/dns.zeek
```diff
@@ -3,8 +3,8 @@
 
 event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qclass: count) &group="Intel::DOMAIN"
 	{
-	Intel::seen([$indicator=query,
-	             $indicator_type=Intel::DOMAIN,
-	             $conn=c,
-	             $where=DNS::IN_REQUEST]);
+	Intel::seen(Intel::Seen($indicator=query,
+	                           $indicator_type=Intel::DOMAIN,
```
ckreibich: Indentation...

scripts/policy/frameworks/intel/seen/file-names.zeek
```diff
@@ -10,10 +10,10 @@ event file_new(f: fa_file) &group="Intel::FILE_NAME"
...
 
 	if ( f?$info && f$info?$filename )
-		Intel::seen([$indicator=f$info$filename,
-		             $indicator_type=Intel::FILE_NAME,
-		             $f=f,
-		             $where=Files::IN_NAME]);
+		Intel::seen(Intel::Seen($indicator=f$info$filename,
+		                           $indicator_type=Intel::FILE_NAME,
+		                           $f=f,
+		                           $where=Files::IN_NAME));
```
ckreibich: These are all off, below too.

I'm stopping here, it's too many.

scripts/policy/frameworks/software/windows-version-detection.zeek
```diff
@@ -21,34 +21,34 @@ export {
         };
 
         const crypto_api_mapping: table[string] of Software::name_and_version = {
-                ["Microsoft-CryptoAPI/5.131.2195.6661"] = [$name="Windows", $version=[$major=5, $minor=131, $minor2=2195, $minor3=6661, $addl="2000 SP4"]],
-                ["Microsoft-CryptoAPI/5.131.2195.6824"] = [$name="Windows", $version=[$major=5, $minor=131, $minor2=2195, $minor3=6824, $addl="2000 with MS04-11"]],
-                ["Microsoft-CryptoAPI/5.131.2195.6926"] = [$name="Windows", $version=[$major=5, $minor=131, $minor2=2195, $minor3=6926, $addl="2000 with Hotfix 98830"]],
+                ["Microsoft-CryptoAPI/5.131.2195.6661"] = Software::name_and_version($name="Windows", $version=Software::Version($major=5, $minor=131, $minor2=2195, $minor3=6661, $addl="2000 SP4")),
```
ckreibich: I suggest skipping here too, it's just too clunky.

scripts/policy/frameworks/intel/seen/file-names.zeek
bbannier: Sorry for the many mismatched offsets, I confused myself.

I looked through all files again to make sure, but maybe I am not seeing the forest for the trees anymore.

scripts/base/protocols/smb/const-dos-error.zeek
bbannier: I assume the other file you meant was `scripts/base/protocols/smb/const-nt-status.zeek`, got rid of the typing there as well.

bbannier:
> Ah no, sorry — I meant it the other way around: I too prefer `const no_filter = Filter($name="");`. I just wasn't sure whether that actually provides the language server any benefit, which I thought was the original motivator for the PR, but which I think I misread.

I think the right framing for this PR is code readability in a plain editor or GH diffs since that should be the baseline we assume.

That said, a language server should in principle be able to resolve whatever Zeek can resolve, and the rest is quality of implementation. My server can currently resolve either form of variable declaration, but there are gaps in hover on constructs like `$name` (with named ctrs this is implemented, but not for coercions from expression lists). Completion I currently implement only for named ctrs as well.

ckreibich (approved):
Benjamin I made an edit pass since I still saw a few indents that were off — I think it's just faster that way. I changed the `SumStats::create()` blocks to also use tabs & spaces, just because they contain so many code blocks, which makes them somewhat easier to edit. I also rebased on top of master.

I'm approving but will leave to you for final review — thanks!

awelzel (approved):
I realize for me it's really just the `Log::create_stream` that tickles something as I suspect that was just done everywhere consistently with the `[ ... ]` style. So this actually introduces an inconsistency into the world. It'd be good to update `zeek-docs` so the logging tutorial and examples uses the new style. I'm thinking even adding a `.. note:` into the create_stream() docs or/and the tutorial about the new and old style. We recommend the new one, but there might be lots of code, tutorials, presentations and training slides using the old style. I.e. explicitly acknowledge that we've changed a long established style.

EDIT: Or exempting `create_stream()` :man_shrugging:

PR #4563, opened 2025-06-10
btest/opt/ZAM-bif-tracking: Disable by default

awelzel:
After a public discussion and also chatting with Vern directly, disable the ZAM bif tracking test to avoid an update every time new functions are added. Usually these aren't performance critical and the defaults characterization is fine. If they are performance critical, then Vern is currently best positioned to properly integrate an optimized version.

---

@vpax - does this work for you?

### DIFF
--- a/testing/btest/opt/ZAM-bif-tracking.zeek
+++ b/testing/btest/opt/ZAM-bif-tracking.zeek
@@ -2,6 +2,9 @@
 #
 # @TEST-REQUIRES: have-spicy
 #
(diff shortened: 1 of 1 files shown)


vpax:
@awelzel looks good to me. (FYI, the main use of the info isn't to replace BiFs with instructions but instead for the optimization analysis of what's safe-vs-might-change during a call to the BiF, to avoid reusing previously computed values if the call might render them stale.)

PR #4565, opened 2025-06-10
Move utility types from util.h into a separate header

timwoj:
I was kicking this around when I was doing all of the `#include` cleanups. This moves a bunch of types out of `util.h` and into a separate header. This lets you include just that header if all you need is `zeek_int_t` or something similar. The new header is still included by `util.h` so nothing should be break.


src/util-types.h
```diff
@@ -0,0 +1,66 @@
...
+using byte_buffer = std::vector<std::byte>;
+using byte_buffer_span = Span<const std::byte>;
+
+namespace util {
+namespace detail {
+
+/**
+ * Wrapper class for functions like dirname(3) or basename(3) that won't
+ * modify the path argument and may optionally abort execution on error.
+ */
```
awelzel, line 35: I just did a cursory grep and the `Safe*` classes are barely used. I wonder if we could switch to `filesystem` even if that might need some mangling of the caller. Safebasename has like 3 users, so questionable why we have a separate class today.

src/util-types.h
```diff
@@ -0,0 +1,66 @@
...
+class SafePathOp {
```
awelzel, line 36: My actual comment though: The implementation for these continues to exist in `src/util.cc`, think it'd be good to move these into `src/util-types.cc`.

awelzel (approved):

src/util-types.h
timwoj: `filesystem` is on my list for once C++20 is in place. We can probably remove the `ghc` version of it at that point too.

PR #4566, opened 2025-06-10
Bugfix: AsString() on an EnumVal will segfault

ckreibich:
If you force a component lookup via an unrelated enum, for example by saying ...
```
event zeek_init() {
        Files::__analyzer_name(udp);
}
```
... you get:
```
(gdb) run ./test.zeek
Thread 1 "zeek" received signal SIGSEGV, Segmentation fault.
0x000000000122f9c8 in zeek::String::CheckStringWithSize (this=0x2) at /home/christian/devel/zeek/zeek/src/ZeekString.cc:157
(gdb) bt
#0  0x000000000122f9c8 in zeek::String::CheckStringWithSize (this=0x2) at /home/christian/devel/zeek/zeek/src/ZeekString.cc:157
#1  0x000000000122fb28 in zeek::String::CheckString (this=0x2) at /home/christian/devel/zeek/zeek/src/ZeekString.cc:178
#2  0x000000000111438a in zeek::plugin::ComponentManager<zeek::file_analysis::Component>::GetComponentNameVal (this=0x48c9de0, val=...)
    at /home/christian/devel/zeek/zeek/src/include/zeek/plugin/ComponentManager.h:298
#3  0x00000000010db1ef in zeek::BifFunc::Files::__analyzer_name_bif (frame=0x665d110, BiF_ARGS=0x7fffffffb8a0) at /home/christian/devel/zeek/zeek/src/file_analysis/file_analysis.bif:106
```
With this tweak:
```
$ zeek ./test.zeek
internal warning: requested name of unknown component tag udp
```

This was previously in #4355 which is plenty big and about to get revamped, so I'm opening a separate PR here.

### DIFF
--- a/src/plugin/ComponentManager.h
+++ b/src/plugin/ComponentManager.h
@@ -267,7 +267,8 @@ const std::string& ComponentManager<C>::GetComponentName(EnumValPtr val) const {
     if ( C* c = Lookup(val.get()) )
         return c->CanonicalName();
 
-    reporter->InternalWarning("requested name of unknown component tag %s", val->AsString()->CheckString());
+    reporter->InternalWarning("requested name of unknown component tag %s",
+                              val->GetType()->AsEnumType()->Lookup(val->Get()));
     return error;
 }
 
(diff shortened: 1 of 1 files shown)


evantypanski (approved):
seems good to me, but I have an extra point:

The warnings in this file use `InternalWarning`, but this is definitely not internal since you can easily trigger it in scriptland - should it be warning instead? If I see "internal" I'd immediately think that Zeek is wrong, not me (the script writer)

though maybe I'm missing other parts where Zeek surfaces internal warnings/errors?

ckreibich:
Thanks, it's a great point. I wondered too but decided to hold off here, simply because it would need a closer look to adjust it consistently. Tags are a bit of a gray area in the script layer and I suspect you can hit trouble with them in other places, too.

The `GetComponentName` functions are invoked from many places, so ideally the warning would have more context about when it's an internal problem or a script-layer one. This is also yet another instance of our enum type-checking being too loose/absent.

PR #4568, opened 2025-06-11
Deprecate the --with-bifcl and --with-binpac configure options

timwoj:
Binpac was already always building the included version even if the argument was passed. We don't do individual releases of binpac or bifcl anymore anyways, so people should already be using the included version all the time.

This also removes another deprecated option that I missed with the v7.1 removals.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,10 @@ if (MSVC)
         #set(WERROR_FLAG "/WX")
         #set(WERROR_FLAG "/WX")
     endif ()
+
+    # Always build binpac in static mode if building on Windows
+    set(BUILD_STATIC_BINPAC true)
+
 else ()
     include(GNUInstallDirs)
     if (BUILD_WITH_WERROR)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1239,11 +1234,8 @@ endif ()
 include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
 execute_process(COMMAND "${CMAKE_COMMAND}" -E create_symlink "." "${CMAKE_CURRENT_BINARY_DIR}/zeek")
 
-if (BinPAC_ROOT_DIR)
-    set(ZEEK_CONFIG_BINPAC_ROOT_DIR ${BinPAC_ROOT_DIR})
-else ()
-    set(ZEEK_CONFIG_BINPAC_ROOT_DIR ${ZEEK_ROOT_DIR})
-endif ()
+set(ZEEK_CONFIG_BINPAC_ROOT_DIR ${BinPAC_ROOT_DIR})
+message(STATUS "#### ${ZEEK_CONFIG_BINPAC_ROOT_DIR}")
```
awelzel: Leftover?

PR #4570, opened 2025-06-11
Implement Redis protocol `message` and `pmessage`

### LINKED ISSUE #4504: Redis analyzer: `message` replies from server are unsupported
from https://github.com/zeek/zeek/pull/4288#discussion_r2105824076

In the `pubsub` test (and any pub-sub Redis traffic) the server sends `message` "replies" that shouldn't be associated with a command.

I'm not sure if there are other cases (I've seen `pmessage`?). I also don't know if these are muted with `CLIENT REPLY` commands. Those should be investigated.

This does technically break `CLIENT REPLY` associations, so marking it as a bug for the Redis analyzer.

evantypanski:
Closes #4504 

when you publish to a topic with Redis, it sends a `message` to each subscriber (or `pmessage` if subscribed with `PSUBSCRIBE`). This should not get correlated with a command.

I chose not to log them, especially since one is sent out for each subscriber. I think that makes things nicer.

I don't know how to generally tell if a message falls into this category - the best I got is an array of strings where the first one is `message` or `pmessage`, but I don't know if that can apply to more cases or not. Guess we'll have to wait and see!!


src/analyzer/protocol/redis/redis.spicy
```diff
@@ -383,21 +383,64 @@ function bulk_string_content(bulk: RESP::BulkString): bytes {
...
 }
 
+function can_stringify(data: RESP::Data): bool {
+    # Simple cases
+    if (data?.simple_error || data?.bulk_error || data?.simple_string || data?.bulk_string || data?.verbatim_string || data?.boolean)
+        return True;
+
+    if (data?.array) {
+        # Arrays can be stringified if all inner elements can be stringified
+        # TODO: Iterating over data.array with a for loop doesn't work??
```
timwoj: You do a `for` over `data.array` below. Does it not work here for some reason?

evantypanski: Yeah I was accidentally doing `data?.array.elements`. But then I just deleted that function in favor of something else so the commit history was a bit bad. I fixed the history to not introduce then remove a function

If there's one thing I always do, it's forget about the `?` when not in conditionals for optionals

src/analyzer/protocol/redis/redis.spicy
```diff
@@ -372,8 +372,27 @@ type ReplyData = struct {
...
 
-public function is_err(server_data: RESP::ServerData): bool {
-    return server_data.data?.simple_error || server_data.data?.bulk_error;
+public type ReplyType = enum {
+    Reply, # A response to a command
+    Error, # An error response to a command
+    Message, # A server message that is not responding to a command
+};
+
+public function classify(server_data: RESP::ServerData): ReplyType {
```
awelzel, line 381: Could you classify directly at parsing time - maybe even without a separate function?  I don't know exactly how the hooks from the `.evt` file are evaluated, but wouldn't be surprised if it'd call classify() three times now, including the potential call to `stringify()`?

You could attach `ReplyType ty` to `RESP::ServerData` or `Data` an populate it with `Reply` by default, but set to error for simple or bulk errors. And Message while parsing an array and the first element happens to be `message` or `pmessage`?

> I don't know how to generally tell if a message falls into this category - the best I got is an array of strings where the first one is message or pmessage, but I don't know if that can apply to more cases or not. Guess we'll have to wait and see!!

Something will get confused with this approach if the parser receives an array with 3 or more elements that happens to start with a legitimate `message` or `pmessage` string. It seems maybe this correlation should happen in the script layer: If a subscription has been established previously, a RESP2 client is not supposed to send new commands and any unsolicited/out-of-band array represents likely is a message/pmessage.

You could possibly also do this in the parser: Track subscribe/psubscribe commands from the client and once subscribed to anything, peek into arrays for "message" or "pmessage" and set some flag on the unit that can be used to change which event is raised.

> A client subscribed to one or more channels shouldn't issue commands, although it can [SUBSCRIBE](https://redis.io/docs/latest/commands/subscribe/) and [UNSUBSCRIBE](https://redis.io/docs/latest/commands/unsubscribe/) to and from other channels. The replies to subscription and unsubscribing operations are sent in the form of messages so that the client can just read a coherent stream of messages where the first element indicates the type of message. The commands that are allowed in the context of a subscribed RESP2 client are [...]

For RESP3 it seems push messages are prefixed with `>` and one could update the Push-case parsing to also propagate the `message` / `pmessage` information, rather than the post-parsing classification.

evantypanski: > For RESP3 it seems push messages are prefixed with >

Oops, yeah I forgot about RESP3, I just figured it didn't need handling, but it should have that extra case for push data.

> but wouldn't be surprised if it'd call classify() three times now, including the potential call to stringify()?

Yeah it calls it 3 times, I can just do that the same way as commands after requests. It didn't feel worth it before, but now it probably is.

> Something will get confused with this approach if the parser receives an array with 3 or more elements that happens to start with a legitimate message or pmessage string

I thought of that then figured the easy solution might be feasible, I didn't immediately see commands that would do that. It is rare, but possible with `HSET` and `HGETALL` - so the parser should account for subscribe mode.

The problem is that we have to care about that in subscribe mode for RESP2, which limits the commands in subscribe mode, but not for RESP3 since it doesn't (it just uses the push replies). So this is the first instance where the RESP2/3 difference actually matters, so I'll need to add support for that. With it comes the `HELLO` command, which from my experience wasn't getting sent with RESP2, but it is possible.

So overall, resolving this needs:

1) Supporting RESP2/3 and choosing which one with `HELLO`
2) Adding subscribed mode in RESP2, where `message` replies get sent in a different event
3) Only calling classify once (if at all after changes)

So my simple solution clearly wasn't enough, thanks for pointing it out

For the record, I think doing it in script layer is tough just because the events only send string data. It'd be really bad to check for messages by seeing if the reply starts with `"[message"` or something. We can't send the nested data structure since evt files don't currently support recursive data types. Something could probably be done, but it doesn't seem super feasible for the moment. In the future I'd really like to get the events to carry more than just the string data, though :)

evantypanski:
Argh, learning about how RESP3 deals with this is making it a lot tougher. There are a number of commands like `SUBSCRIBE` which don't expect a response (it becomes a "push" response) and subscribed mode matters for RESP2 but not 3. I'm going to draft this so it's clear that it's not ready. I think I have a clean solution but I want to make sure all of the edge cases are covered.

Bonus points: I think a solution can just not do request/reply correlation in subscribed mode in RESP3, and all replies are push (even though `PONG` replies technically aren't). This makes it easier and makes doing it in the script layer better - which I prefer. Just means it might take a little longer.

evantypanski:
Changes got pretty substantial so I'll close this and open a new one to keep history.

PR #4579, opened 2025-06-14
Pluggable connection tuples / keys for Zeek

ckreibich:
This is a pretty broad iteration over the prototype from #4355 (which I'm about to close since this really is a deep revamp, not just an iteration). I've tried to include the feedback folks provided there, particularly Arne's design for the more self-contained factories. A couple of things to call out:

- I made the `ConnKey` base class its own file in src/ since it seems sufficiently important.
- I deprecated the original `Connection` constructor and also the old `ConnKey` and `ConnTuple` itself. Given that `ConnTuple` was mainly a vehicle to get tuple info into `Connection`s we no longer really need this, but it may be controversial. If you dislike it, you can simply remove 6e501b47.
- There's a hiccup with `[[deprecated]]` on whole classes/structs: on FreeBSD and macOS these trigger hard errors instead of warnings — I simply turned them into comments for now.  
- I worked in the builder -> factory renaming and added test cases.
- Arne I agree with your observations regarding the `CONNECTION_KEY_TYPE` arguments to session keys. Wasn't sure whether it's worth changing here — probably better to consider in a separate iteration.

I'm handing this off to you @awelzel while I'm out — hopefully you don't see much need for deep changes at this point. Thanks for your help!


ckreibich:
I looked at that docs failure a bit. If you manually add this reST somewhere, the build is happy: 
```
Connection Tuples
=================

.. zeek:type:: ConnTuple::Tag

   :Type: :zeek:type:`enum`

      .. zeek:enum:: ConnTuple::CONNTUPLE_FIVETUPLE ConnTuple::Tag

      .. zeek:enum:: ConnTuple::CONNTUPLE_VLAN ConnTuple::Tag
```
So I think the proper way to do it would be a new `Target` class in Zeekyen that produces this. I wonder if we could change the way the component manager wrangles these tag types so one can also introduce them in the script layer, so they'd be covered automatically for documentation — then the whole thing wouldn't feel so "custom".

src/conntuple/Factory.h
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Instantiates a filled-in ConnKey derivative from a script-layer
+     * record, usually a conn_id instance. Implementations are free to
+     * implement this liberally, i.e. the input does not _have_ to be a
+     * conn_id.
+     *
+     * @param v The script-layer value providing key input.
+     * @return A unique pointer to the ConnKey instance.
+     */
+    virtual zeek::ConnKeyPtr FromVal(const zeek::ValPtr& v) = 0;
```
awelzel: Naming nit, `ConnKeyFromVal()` while longer, would be more symmetric with `NewConnKey()`.

Also, return `expected` for error reporting so that users of that API are forced to do error handling.

src/Conn.cc
```diff
@@ -38,6 +59,29 @@ Connection::Connection(const detail::ConnKey& k, double t, const ConnTuple* id,
         default: proto = TRANSPORT_UNKNOWN; break;
     }
 
+    key = std::make_unique<zeek::IPConnKey>();
```
awelzel, line 62: I was worried about deprecation / API breaks, but this is clever :-)

src/conntuple/vlan/Factory.cc
```diff
@@ -0,0 +1,114 @@
...
+        int vlan_off, inner_vlan_off;
+        GetConnIdFieldOffsets(vlan_off, inner_vlan_off);
+
+        if ( key.vlan && vlan_off >= 0 )
+            conn_id->Assign(vlan_off, key.vlan);
+        if ( key.inner_vlan && inner_vlan_off >= 0 )
+            conn_id->Assign(inner_vlan_off, key.inner_vlan);
+    };
+
+    void GetConnIdFieldOffsets(int& a_vlan_off, int& a_inner_vlan_off) {
```
awelzel: Maybe return tuple/pair instead of inout refs, then the caller could probably be.

```
auto [vlan_off, inner_vlan_off] = GetConnIdFieldOffsets();
```

src/conntuple/vlan/Factory.cc
```diff
@@ -0,0 +1,114 @@
...
+    int vlan_off, inner_vlan_off;
+
+    if ( rt == id::conn_id ) {
+        k->GetConnIdFieldOffsets(vlan_off, inner_vlan_off);
+    }
+    else {
+        // We don't know what we've been passed:
+        vlan_off = rt->FieldOffset("vlan");
+        inner_vlan_off = rt->FieldOffset("inner_vlan");
+
```
awelzel: Missing brace? The code below should run for `conn_id` and other record types, but currently only runs for other record types.

src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc
```diff
@@ -22,17 +24,39 @@ IPBasedAnalyzer::~IPBasedAnalyzer() {
...
     ConnTuple tuple;
-    if ( ! BuildConnTuple(len, data, pkt, tuple) )
+    if ( BuildConnTuple(len, data, pkt, tuple) ) {
+#pragma GCC diagnostic pop
+        key->InitTuple(tuple.src_addr, tuple.src_port, tuple.dst_addr, tuple.dst_port, pkt->proto);
+    }
+    else if ( ! InitConnKey(len, data, pkt, *key) ) {
         return false;
+    }
+
```
awelzel, line 49:
```suggestion
// Init hook for plugin provided ConnKey instances
```

src/session/Manager.cc
```diff
@@ -88,23 +89,23 @@ Manager::~Manager() {
 }
 
 Connection* Manager::FindConnection(Val* v) {
-    zeek::detail::ConnKey conn_key(v);
+    // XXX: This could in the future dispatch to different factories for
+    // different kinds of Vals. ``v`` will usually be a conn_id instance, which
+    // is IP-specific. If ``v`` is something else, maybe we'd like to use a
+    // different builder.
```
awelzel, line 95: I am okay keeping the XXX comment as is: This PR isn't about supporting non-IP connections, but instead be able to have pluggable keys/tuples for IP-based connections. Seems "natural" to discover places where we assume IP and we don't need to solve these today.

src/ConnKey.h
```diff
@@ -0,0 +1,75 @@
...
+    virtual zeek::session::detail::Key SessionKey() const = 0;
+
+    /**
+     * Get the error state of a ConnKey, if any.
+     *
+     * Instances of a ConnKey created from zeek::Val instances via
+     * Builder::FromVal() may not be valid. Calling Error() can be used to
+     * gather a description of the encountered error.
+     */
+    virtual std::optional<std::string> Error() const = 0;
```
awelzel: Probably better to change `Factory::FromVal()` to return `expected` so that callers are forced to do the error handling. That feels more robust than asking users to call `Error()` on the result.

src/conntuple/vlan/Factory.cc
```diff
@@ -0,0 +1,114 @@
...
+        // that's where we learn the tag values from. 0 indicates absence.
+        uint32_t vlan;
+        uint32_t inner_vlan;
+    } __attribute__((packed, aligned)) key;
+};
+
+zeek::ConnKeyPtr Factory::NewConnKey() { return std::make_unique<IPVlanConnKey>(); }
+
+zeek::ConnKeyPtr Factory::FromVal(const zeek::ValPtr& v) {
+    auto ck = zeek::conntuple::fivetuple::Factory::FromVal(v);
```
awelzel: This is a bit magic, but supposedly works out as `FromVal()` uses the virtual `NewConnKey()` in the fivetuple factory and so it ends up returning a `IPVlanConnKey`.

It's a low-level implementation-code-reuse-detail though, so I guess the magic here is ok.

src/ConnKey.h
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * When Zeek renders a connection into a script-layer record, it calls this
+     * method to provide custom conn_id fields unique to this ConnKey, such as
+     * VLAN fields. This only needs to fill in fields in addition to Zeek's
+     * five-tuple (i.e., complete the record, not fill all of it -- hence the
+     * name).
+     *
+     * The default implementation does nothing.
+     */
+    virtual void CompleteConnIdVal(RecordValPtr& conn_id) {};
```
awelzel: Introduce `DoCompleteConnIdVal()` for non-virtual-interface and probably also for `SessionKey` just for consistency.

src/conntuple/Factory.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Factory {
+public:
+    virtual ~Factory() = default;
+
+    /**
+     * Instantiates a clean ConnKey derivative and returns it.
+     * @return A unique pointer to the ConnKey instance.
+     */
+    virtual zeek::ConnKeyPtr NewConnKey() = 0;
```
awelzel: @bbannier - do you think it always makes sense to apply non-virtual-interface? Then I'd probably add corresponding `Do...` methods here, even if we don't directly have a use.

src/plugin/Component.h
```diff
@@ -38,6 +38,7 @@ enum Type {
     LOG_SERIALIZER,     /// A serializer for log batches, used by cluster backends.
     STORAGE_BACKEND,    /// A backend for the storage framework.
     STORAGE_SERIALIZER, /// A serializer for the storage framework.
+    CONNTUPLE,          /// A factory for connection tuples.
```
awelzel:
```suggestion
    CONNKEY,            /// A factory for connection keys.
```

The component creates ConnKey instances, so seems reasonable to name it that.

src/conntuple/Component.h
```diff
@@ -0,0 +1,45 @@
...
+
+#pragma once
+
+#include <functional>
+#include <memory>
+
+#include "zeek/Tag.h"
+#include "zeek/plugin/Component.h"
+
+namespace zeek::conntuple {
```
awelzel:
```suggestion
namespace zeek::conn_key {
```

...and will move to `src/conn_key/`.  We don't seem to use compressed names a lot and instead have `_` quite a bit: packet_analysis, file_analysis, run_state, conn_size.

src/packet_analysis/protocol/teredo/functions.bif
```diff
@@ -12,8 +13,10 @@ function remove_teredo_connection%(cid: conn_id%) : bool
 	zeek::packet_analysis::AnalyzerPtr teredo = zeek::packet_mgr->GetAnalyzer("Teredo");
 	if ( teredo )
 		{
-		zeek::detail::ConnKey conn_key(cid);
-		static_cast<zeek::packet_analysis::teredo::TeredoAnalyzer*>(teredo.get())->RemoveConnection(conn_key);
+		// XXX: Should be IP specific builder!
+		auto ck = zeek::conntuple_mgr->GetFactory().FromVal({zeek::NewRef{}, cid});
```
awelzel: Maybe add error handling.

scripts/base/init-bare.zeek
```diff
@@ -617,6 +617,19 @@ export {
 	const add_missing_remote_network_timestamp: bool = F &redef;
 }
 
+module ConnTuple;
```
awelzel: I'm going to rename `ConnTuple` to `ConnKey`. Mostly as right now the implementation is heavy on using the `ConnKey` terminology, but the Zeek module and also directory structure uses `ConnTuple`.

I'm preferring the "key" terminology and a bit of chatting with @J-Gras  and @bbannier, there's also no strong contender for "tuple". The main reason that this is about a "key" use in the session map.

This will also apply to the plugin component type and directory names.

src/conntuple/fivetuple/Factory.h
```diff
@@ -0,0 +1,31 @@
+// See the file "COPYING" in the main distribution directory for copyright.
```
awelzel: I wonder a bit if these could move from `src/conntuple` into  `src/packet_analysis/protocols/ip/conn_key/{vlan,fivetuple}/...`. That's where the `IPBasedConnKey` exists.

ckreibich:
I pondered the connkey vs conntuple naming aspect too. My take was that for users it's more intuitive that the feature this provides is configurable connection _tuples_ (a commonly agreed term), while the implementation focuses largely on a Zeek-internal implementation detail that happens to involve a thing called "keys". This leaves gray areas but is how I'd frame it overall.

src/conntuple/Factory.h
bbannier: Not really sure. In general I'd say NVI is a good idea since it allows to without API break inject additional code before concrete implementations are called, so I'd say use it where possible. At the same time it also introduces an extra indirection, so one needs to be careful when adding it on very hot code paths.

awelzel:
> I pondered the connkey vs conntuple naming aspect too. My take was that for users it's more intuitive that the feature this provides is configurable connection _tuples_ (a commonly agreed term), while the implementation focuses largely on a Zeek-internal implementation detail that happens to involve a thing called "keys". This leaves gray areas but is how I'd frame it overall.

Thanks for chiming in. I think I'm set on "conn key" to avoid the mixing of different terminology in directory structure and implementation. The tuple term will be kept and exposed to users, it's just not so prevalent in the directory structure and concrete implementations.

Concretely, the name of the default tag will change to `ConnKey::CONNKEY_FIVETUPLE`. To me, this name mixes nicely what the purpose is internally, but also includes the common and more abstract tuple terminology (I'm thinking of making the VLAN aware conn key name more explicit, too, as `CONNKEY_VLAN_FIVETUPLE`  instead of just  `CONNKEY_VLAN`). Anyhow, it's not that the "tuple" terminology disappears. Indeed, tuple comes up 3 times in `ConnTuple::CONNTUPLE_FIVETUPLE`. Having it only once in the `ConnKey::CONNKEY_FIVETUPLE` name for me even increases it's importance.

Wall of text, but hope it makes sense where this is coming from.

src/conntuple/Component.h
```diff
@@ -0,0 +1,45 @@
...
+    void Initialize() override;
+
+    /**
+     * Returns the analyzer's factory function.
+     */
+    factory_callback Factory() const { return factory; }
+
+protected:
+    /**
+     * Overridden from plugin::Component.
```
0xxon: Nit - this could still describe what it does, as it's done in several other parts of the source tree.

src/conntuple/Factory.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * ConnKey factories instantiate derivatives of ConnKeys, to provide pluggable flow hashing.
+ */
+class Factory {
+public:
+    virtual ~Factory() = default;
+
+    /**
+     * Instantiates a clean ConnKey derivative and returns it.
+     * @return A unique pointer to the ConnKey instance.
```
0xxon: Nit - missing newline before `@return`.

src/conntuple/Manager.h
```diff
@@ -0,0 +1,33 @@
+// See the file "COPYING" in the main distribution directory for copyright.
```
0xxon: This header currently has no documentation or description at all.

src/conntuple/fivetuple/Factory.h
```diff
@@ -0,0 +1,31 @@
...
+#include "zeek/ConnKey.h"
+#include "zeek/conntuple/Factory.h"
+
+namespace zeek::conntuple::fivetuple {
+
+class Factory : public zeek::conntuple::Factory {
+public:
+    /**
+     * Instantiates a clean ConnKey derivative and returns it.
+     * @return A unique pointer to the ConnKey instance.
```
0xxon: Nit - missing newline before `@return`.

Similar for the other NewConnKey descriptions in other headers.

src/packet_analysis/protocol/ip/ConnKey.h
```diff
@@ -0,0 +1,107 @@
...
+ * ConnKey instances for IP always hold a ConnTuple instance which is provided
+ * by the IPBasedAnalyzer. The InitConnTuple() method stored a normalized version
+ * in the tuple, loosing the information about orig and responder.
+ */
+class IPBasedConnKey : public zeek::ConnKey {
+public:
+    /**
+     * Initializes the key to the given 5-tuple. This canonicalizes the
+     * packed tuple storage, including potential endpoint flips for
+     * consistent connection lookups regardless of directionality.
```
0xxon: I am not sire if this matters super much - but the methods in this file miss the documentation of arguments and return values that is required according to our style guide.

src/packet_analysis/protocol/ip/ConnKey.h
```diff
@@ -0,0 +1,107 @@
...
+};
+
+using IPBasedConnKeyPtr = std::unique_ptr<IPBasedConnKey>;
+
+/**
+ * The usual 5-tuple ConnKey, fully instantiable.
+ */
+class IPConnKey : public IPBasedConnKey {
+public:
+    IPConnKey() {
```
0xxon: methods have no documentation (though pretty obvious).

src/packet_analysis/protocol/unknown_ip_transport/UnknownIPTransport.h
```diff
@@ -23,10 +23,7 @@ class UnknownIPTransportAnalyzer final : public IP::IPBasedAnalyzer {
     packet_analysis::IP::SessionAdapter* MakeSessionAdapter(Connection* conn) override;
 
 protected:
-    /**
-     * Parse the header from the packet into a ConnTuple object.
-     */
-    bool BuildConnTuple(size_t len, const uint8_t* data, Packet* packet, ConnTuple& tuple) override;
+    bool InitConnKey(size_t len, const uint8_t* data, Packet* packet, IPBasedConnKey& key) override;
```
0xxon, line 26: Perhaps keep the comment? (I know, the other functions don't have it).

src/ConnKey.h
awelzel: Changed to use expected and dropped `Error()` on ConnKey.

src/conntuple/Component.h
awelzel: I'm going to leave as is. Grepping over the source tree, this is what most other Component.h files have.

src/conntuple/Manager.h
awelzel: Added comments.

src/packet_analysis/protocol/ip/ConnKey.h
awelzel: Added them.

src/packet_analysis/protocol/unknown_ip_transport/UnknownIPTransport.h
awelzel: I've left it out. Agree on verbose comments on the API/interface, but think it's okay to be more lenient for actual implementations of a given API - it'd just be a repeat.

There's doxygen's `@copydoc` - maybe we could have a discussion of using that more often, though that'll only make sense when looking at the generated HTML and I realize we don't actually have a doxygen setup :-(

awelzel:
I've reworked the commits, adapting some names as proposed and also changed the `FromVal()` API to return `expected` instead of of a ConnKey instance with a set `Error()`.

The IP specific ConnKey implementations moved from `src/conntuple` into `src/packet_analysis/protocol/ip/conn_key/...`. Mostly because this is all IP specific and only used by the IPBasedAnalyzer that's in the same directory. The generic infrastructure lives in `src/conn_key`.

Given this has been some 3 or 4 cycles between Christian and me (and I think Tim, Jan, Johanna and Benjamin all skimmed), I'd be confident to merge by Monday or Tuesday and we do any follow-ups and polishes in separate PRs, even if there's not another in-depth review. Objections?

I ran hyperfine for pure packet processing locally and I'm unconcerned.

I'll add a plugin btest implementing a custom ConnKey factory and possibly a short developer intro for the docs.

src/conn_key/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+zeek_add_subdir_library(
+    connkey
+    INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_CURRENT_BINARY_DIR}
```
timwoj: These includes aren't necessary.

src/conn_key/Manager.h
```diff
@@ -0,0 +1,57 @@
...
+     * @return A pointer to a Factory given @arg tag.
+     */
+    FactoryPtr InstantiateFactory(const EnumValPtr& tag);
+
+    FactoryPtr factory;
+};
+
+} // namespace conn_key
+
+extern zeek::conn_key::Manager* conn_key_mgr;
```
timwoj, line 54: I feel like we need to pick a standard on this. Most of the managers are in the `zeek::` namespace and are named based on the component they're in. We also have at least one or two in the more-specific namespace, such as `zeek::cluster::manager`.

src/packet_analysis/protocol/gtpv1/functions.bif
```diff
@@ -12,8 +13,14 @@ function remove_gtpv1_connection%(cid: conn_id%) : bool
 	zeek::packet_analysis::AnalyzerPtr gtpv1 = zeek::packet_mgr->GetAnalyzer("GTPv1");
 	if ( gtpv1 )
 		{
-		zeek::detail::ConnKey conn_key(cid);
-		static_cast<zeek::packet_analysis::gtpv1::GTPv1_Analyzer*>(gtpv1.get())->RemoveConnection(conn_key);
+		// XXX: Should be IP specific builder!
+		auto r = zeek::conn_key_mgr->GetFactory().ConnKeyFromVal({zeek::NewRef{}, cid});
+		if ( ! r.has_value() )
+			return zeek::val_mgr->False();
+
```
timwoj, line 19: Extra space here.

src/packet_analysis/protocol/ip/conn_key/IPBasedConnKey.h
```diff
@@ -0,0 +1,116 @@
...
+} __attribute__((packed, aligned));
+
+} // namespace detail
+
+/**
+ * Abstract key class for IP-based connections.
+ *
+ * ConnKey instances for IP always hold a ConnTuple instance which is provided
+ * by the IPBasedAnalyzer. The InitConnTuple() method stored a normalized version
+ * in the tuple, loosing the information about orig and responder.
```
timwoj:
```suggestion
 * by the IPBasedAnalyzer. The InitConnTuple() method stores a normalized version
 * in the tuple, losing the information about orig and responder.
```

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
```diff
@@ -0,0 +1,124 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#include "zeek/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.h"
```
timwoj, line 3: Is it really a _*five*_tuple once you add more data to it?

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
```diff
@@ -0,0 +1,124 @@
...
+    auto ck = zeek::conn_key::fivetuple::Factory::DoConnKeyFromVal(v);
+
+    if ( ! ck.has_value() )
+        return ck;
+
+    auto* k = static_cast<IPVlanConnKey*>(ck.value().get());
+    auto rt = v->GetType()->AsRecordType();
+    auto vl = v->As<RecordVal*>();
+
+    int vlan_off, inner_vlan_off;
```
timwoj: Naming these `_offset` would be more clear. Them being named `_off` makes me think they're some sort of boolean flag.

src/packet_analysis/protocol/gtpv1/functions.bif
```diff
@@ -12,8 +13,14 @@ function remove_gtpv1_connection%(cid: conn_id%) : bool
 	zeek::packet_analysis::AnalyzerPtr gtpv1 = zeek::packet_mgr->GetAnalyzer("GTPv1");
 	if ( gtpv1 )
 		{
-		zeek::detail::ConnKey conn_key(cid);
-		static_cast<zeek::packet_analysis::gtpv1::GTPv1_Analyzer*>(gtpv1.get())->RemoveConnection(conn_key);
+		// XXX: Should be IP specific builder!
```
timwoj: I assume this is coming later?

src/conn_key/Manager.h
awelzel: Yeah, stared at this one and it's somewhat inconsistent.

What would be your suggested change for this one here? `connkey_mgr`? This bites a bit with the `conn_key` namespace having a `_`, but the component is `CONNKEY`. I had `zeek::conn_key::mgr` for a bit, but that'd also be a new pattern I do like that most though.

And yes,  I introduced the `cluster::manager` and it is inconsistent. I'm totally up for moving that to `zeek::cluster_mgr`, but not opposed to moving into a `zeek::cluster::mgr` world either.

And for those managers that are really only used to register and instantiate components, maybe there could be a better/separate place, too, rather than having a proliferation in zeek-setup.cc.

Anyhow, did you want to suggest a change of name, or was this mostly a fly-by?

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
awelzel: The meaning is intended to be `vlan-plus-fivetuple` or `vlan-scoped-fivetuple`, but dropping the "scoped". I think it works, but others might feel different.

src/packet_analysis/protocol/gtpv1/functions.bif
awelzel: Hmm, let me ditch the comment, also in teredo.

This was mostly a note that `conn_key_mgr->Factory()` is currently returning something that can always work with `IP`, but somewhere in the future we might have non-IP connections and `conn_key_mgr`. Then we should select a IP-conn-key-factory here, rather than assuming that `GetFactory()` returns a matching one.

But that's so far out, that when this starts as a project the `XXX` comment here won't help anyhow.

src/conn_key/Manager.h
timwoj: Mostly fly-by. I do think we could standardize on something though. I personally like the `zeek::namespace::manager` version, but in the storage framework I left it as the "old" style for consistency. Changing all of them would require a deprecation cycle.

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
timwoj: Maybe just `vlan_tuple`?

src/packet_analysis/protocol/ip/conn_key/IPBasedConnKey.h
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * This is virtual to give subclasses control over where
+     * to place the tuple within the key.
+     *
+     * @return A modifiable reference to the embedded PackedConnTuple.
+     */
+    virtual detail::PackedConnTuple& PackedTuple() = 0;
+
+    /**
+     * Return a no-modifiable reference to the embedded PackedConnTuple.
```
timwoj:
```suggestion
     * Return a non-modifiable reference to the embedded PackedConnTuple.
```

timwoj (approved):

timwoj:
One last nit, but looks good! :shipit:

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
awelzel: I'm going to leave it as `vlan_fivetuple`, but that doesn't mean I'd be opposed to discussing the names in a follow-up, when people feel like talking or showing it in demos/presentations is inconsistent. For me the current state is okay'ish.

PR #4582, opened 2025-06-16
IOSource: Remove IsPacketSource

### LINKED ISSUE #4573: Deprecate IOSource::IsPacketSource()?
From https://github.com/zeek/zeek/pull/4567#discussion_r2139639777:

> Total fly-by skimming: Could we deprecate `IsPacketSource()` below? Seems unused and also seems a bit strange that a consumer working with `IOSource`s should know whether it's a packet source and even our own packet sources do not override. So, maybe just ditch it right away? It hasn't worked for years.
> 
> Somewhere in the IO loop there's `iosource == pkt_src` and I suspect `IsPacketSource()` might be reminiscent of that.
>

awelzel:
This wasn't used in tree and even the PktSrc class doesn't override this to return true, so just remove it outright without deprecation.

Closes #4573

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -48,6 +48,9 @@ Breaking Changes
 
 	redef EventMetadata::add_missing_remote_network_timestamp = T;
 
+- The ``IsPacketSource()`` method on ``IOSource`` was removed. It was unused
+  and incorrectly returned ``false`` on all packet sources.
+
 New Functionality
 -----------------
 
--- a/src/iosource/IOSource.h
+++ b/src/iosource/IOSource.h
@@ -33,11 +33,6 @@ public:
     bool IsOpen() const { return ! closed; }
 
     /**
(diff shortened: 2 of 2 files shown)


NEWS
```diff
@@ -48,6 +48,9 @@ Breaking Changes
 
 	redef EventMetadata::add_missing_remote_network_timestamp = T;
 
+- The ``IsPacketSource()`` method on ``IOSource`` was removed. It was unused
+  and not even Zeek's ``PktSrc`` class would return true.
```
evantypanski: nit: I'd prefer if this said something like:

```suggestion
  and incorrectly returned ``false`` on all packet sources.
```

Just to stress that it was providing incorrect information. As is, you could misinterpret and think there are other cases it was accurate

awelzel: Perfect, thanks!

PR #4583, opened 2025-06-18
Add workaround for zeek/spicy#2091

bbannier:


0xxon (approved):
This looks good to me.

As a side-note. The workaround for the spicy file mostly undoes changes that were introduced in be9bbff6b9c. 

From what I remember, back then, from my memory, these changes were introduced because the syntax that is now used as a workaround was not supposed to work - and this was caught by stronger validation.

I'm a bit surprised that this now seems to just work again - and am mostly curious which syntax is considered the more correct/idiomatic in these cases.

bbannier:
> I'm a bit surprised that this now seems to just work again - and am mostly curious which syntax is considered the more correct/idiomatic in these cases.

We did some [straightening out related to references](https://github.com/zeek/spicy/pull/1712) just around the time of be9bbff6b9cf93a3e5d9da18356a4e61482ca9cd, and it almost feels to me like the changes to the Spicy SSL parser precede that.

Overall things should be pretty simple today:

- references are constructed with `new` and passed as `T&` like you already do with sinks
- units are exposed as values and passed like them, e.g, `Share`
- inout parameters to functions need to be explicitly declared with `inout`
- this is not needed for references since they can always be mutated through other bindings, there is a diagnostic for reference passing with `inout`

PR #4585, opened 2025-06-18
Default canonifier change to only remove first timestamp in line

0xxon:
In the past, we used a default canonifier, which removes everything that looks like a timestamp from log files. The goal of this is to prevent logs from changing, e.g., due to local system times ending up in log files.

This, however, also has the side-effect of removing information that is parsed from protocols which probably should be part of our tests. There is at least one test (1999 certificates) where the entire test output was essentially removed by the canonifier.

GH-4521 was similarly masked by this.

This commit changes the default canonifier, so that only the first timestamp in a line is removed. This should remove timestamps that are likely to change from logfiles while keeping timestamps that are parsed from protocol information.

A pass has been made over the tests, with some additional adjustments for cases which require the old canonifier.

There are some cases in which we probably could go further and not remove timestamps at all - that, however, seems like a follow-up project.


testing/btest/Baseline/scripts.base.frameworks.config.basic/zeek.config.log
```diff
@@ -21,7 +21,7 @@ XXXXXXXXXX.XXXXXX	testaddr	2607:f8b0:4005:801::1	2607:f8b0:4005:801::2	../config
 XXXXXXXXXX.XXXXXX	testsub	0.0.0.0/0	2607:f8b0:4001::/48	../configfile
 XXXXXXXXXX.XXXXXX	testsub	2607:f8b0:4001::/48	2607:f8b0:4002::/48	../configfile
 XXXXXXXXXX.XXXXXX	testinterval	1.0 sec	1.0 min	../configfile
-XXXXXXXXXX.XXXXXX	testtime	0.0	XXXXXXXXXX.XXXXXX	../configfile
+XXXXXXXXXX.XXXXXX	testtime	0.0	1507321987.0	../configfile
```
timwoj, line 24: 🤦🏼

testing/btest/Baseline/scripts.policy.frameworks.netcontrol.catch-and-release-forgotten/.stdout
```diff
@@ -2,4 +2,4 @@
 netcontrol debug (Debug-All): init
 netcontrol debug (Debug-All): add_rule: [ty=NetControl::DROP, target=NetControl::FORWARD, entity=[ty=NetControl::ADDRESS, conn=<uninitialized>, flow=<uninitialized>, ip=10.0.0.1/32, mac=<uninitialized>], expire=1.0 sec, priority=0, location=, out_port=<uninitialized>, mod=<uninitialized>, id=2, cid=2, _plugin_ids={\x0a\x0a}, _active_plugin_ids={\x0a\x0a}, _no_expire_plugins={\x0a\x0a}, _added=F]
 netcontrol debug (Debug-All): remove_rule (): [ty=NetControl::DROP, target=NetControl::FORWARD, entity=[ty=NetControl::ADDRESS, conn=<uninitialized>, flow=<uninitialized>, ip=10.0.0.1/32, mac=<uninitialized>], expire=1.0 sec, priority=0, location=, out_port=<uninitialized>, mod=<uninitialized>, id=2, cid=2, _plugin_ids={\x0a\x091\x0a}, _active_plugin_ids={\x0a\x091\x0a}, _no_expire_plugins={\x0a\x0a}, _added=T]
-Forgotten: , 10.0.0.1, [block_until=XXXXXXXXXX.XXXXXX, watch_until=XXXXXXXXXX.XXXXXX, num_reblocked=0, current_interval=0, current_block_id=2, location=<uninitialized>]
+Forgotten: , 10.0.0.1, [block_until=XXXXXXXXXX.XXXXXX, watch_until=1254722769.49206, num_reblocked=0, current_interval=0, current_block_id=2, location=<uninitialized>]
```
timwoj, line 5: Does this one need to be canonified at all?

src/analyzer/protocol/smb/smb-time.pac
```diff
@@ -11,7 +11,7 @@ double filetime2zeektime(uint64_t ts)
 	{
 	// Zeek can't support times back to the 1600's
 	// so we subtract a lot of seconds.
-	return (ts / 10000000.0L) - 11644473600.0L;
+	return (static_cast<double>(ts) / 10000000.0) - 11644473600.0;
```
timwoj, line 14: Some day when we have C++ 23: https://en.cppreference.com/w/cpp/types/floating-point.html

testing/btest/Baseline/scripts.policy.frameworks.netcontrol.catch-and-release-forgotten/.stdout
0xxon: No, it does not. This is one of the mentioned cases where we could forego canonification completely - however that seemed like a bit of a separate issue.

src/analyzer/protocol/smb/smb-time.pac
0xxon: Oh, neat.

PR #4589, opened 2025-06-19
Build builtin Spicy analyzers in debug mode if debug mode is enabled

### LINKED ISSUE #4587: Built-in spicy analyzers don't contain debug instrumentation, even if Zeek is built in debug mode
Currently, spicy analyzers in the Zeek tree will never contain debug instrumentation, even if Zeek itself is built in debug mode.

That obviously makes debugging them harder.

I currently use this cmake patch locally, which just unconditionally enables debug instrumentation:

```patch
--- a/BuiltInSpicyAnalyzer.cmake
+++ b/BuiltInSpicyAnalyzer.cmake
@@ -85,7 +85,7 @@ function (spicy_add_analyzer)
             COMMAND
                 ${CMAKE_COMMAND} -E env
                 "ZEEK_SPICY_LIBRARY_PATH=${PROJECT_SOURCE_DIR}/scripts/spicy"
-                ASAN_OPTIONS=$ENV{ASAN_OPTIONS}:detect_leaks=0 $<TARGET_FILE:spicyz> -L
+                ASAN_OPTIONS=$ENV{ASAN_OPTIONS}:detect_leaks=0 $<TARGET_FILE:spicyz> -d -L
                 ${spicy_SOURCE_DIR}/hilti/lib -L ${spicy_SOURCE_DIR}/spicy/lib -x
                 ${CMAKE_CURRENT_BINARY_DIR}/${NAME_LOWER} ${SPICYZ_FLAGS} ${SPICY_ANALYZER_SOURCES}
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```

bbannier:
Closes #4587.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -649,6 +649,7 @@ if (ENABLE_DEBUG)
     set(VERSION_C_IDENT "${VERSION_C_IDENT}_debug")
     target_compile_definitions(zeek_internal INTERFACE DEBUG)
     target_compile_definitions(zeek_dynamic_plugin_base INTERFACE DEBUG)
+    set(SPICYZ_FLAGS "-d" CACHE STRING "Additional flags to pass to spicyz for builtin analyzers")
 endif ()
 
 if (NOT BINARY_PACKAGING_MODE)


bbannier:
> We also set `ENABLE_DEBUG` to true via CMake if `--build-type=debug` is passed without `--enable-debug`. We should probably set this there as well.

I instead moved this from `./configure` fully into CMake so we set this whenever `ENABLE_DEBUG` is enabled (either explicitly, or for `Debug`, but not `RelWithDebInfo` builds); I also put the setting into the cache so user tweaks are preserved. This should be easier to maintain than having two copies.

timwoj (approved):

PR #4590, opened 2025-06-20
cluster/zeromq: Hook up and enable IPV6 by default

awelzel:
ZeroMQ's IPv6 support isn't enabled by default, resulting in
"No such device" errors when attempting to listen on an IPv6
address. This change adds a ipv6 option to the ZeroMQ module
and enables it by default. Further, adds a test configuring
everything to listen on IPv6 ::1 as well, and one test to provoke
the original error.

The addr_to_uri() calls weren't actually needed, but they apparently do
not hurt and the result is easier on the eyes, so use them :-)

---

I guess the `ipv6-logging.zeek` test is a bit much/duplicated from the `logging.zeek` test, but I'm not feeling to bad about it.

### DIFF
--- a/src/cluster/backend/zeromq/ZeroMQ.cc
+++ b/src/cluster/backend/zeromq/ZeroMQ.cc
@@ -373,7 +373,6 @@ bool ZeroMQBackend::DoUnsubscribe(const std::string& topic_prefix) {
 
 bool ZeroMQBackend::DoPublishLogWrites(const logging::detail::LogWriteHeader& header, const std::string& format,
                                        byte_buffer& buf) {
-    ZEROMQ_DEBUG("Publishing %zu bytes of log writes (path %s)", buf.size(), header.path.c_str());
     static std::string message_type = "log-write";
 
     // Publishing a log write is done using 4 parts
@@ -388,6 +387,14 @@ bool ZeroMQBackend::DoPublishLogWrites(const logging::detail::LogWriteHeader& he
         zmq::const_buffer{format.data(), format.size()},
         zmq::const_buffer{buf.data(), buf.size()},
     };
+
+    // If the log_push socket isn't yet initialized or has been closed, just return.
+    if ( ! log_push ) {
+        ZEROMQ_DEBUG("Skipping log write - log_push socket not open");
+        return false;
+    }
+
+    ZEROMQ_DEBUG("Publishing %zu bytes of log writes (path %s)", buf.size(), header.path.c_str());
 
     for ( size_t i = 0; i < parts.size(); i++ ) {
         zmq::send_flags flags = zmq::send_flags::dontwait;


testing/btest/Baseline/cluster.zeromq.ipv6-error/logger..stderr
```diff
@@ -0,0 +1,3 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+error: ZeroMQ: Failed to bind to PULL socket tcp:/[::1]:xxxx: No such device
```
awelzel: `tcp:/[` due to abspath canonifier. This is really `tcp://[` in the output...

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -385,6 +388,14 @@ bool ZeroMQBackend::DoPublishLogWrites(const logging::detail::LogWriteHeader& he
         zmq::const_buffer{buf.data(), buf.size()},
     };
 
+    // If the log_push socket isn't yet initialized or has been closed, just
```
timwoj:
```suggestion
    // If the log_push socket isn't yet initialized or has been closed, just return.
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -139,6 +142,17 @@ export {
 	## for more details.
 	const log_rcvbuf: int = -1 &redef;
 
+	## Set ZMQ_IPV6 option.
+	##
+	## The ZeroMQ library has IPv6 support in ZeroMQ. For Zeek it
+	## we enable it unconditionally such that listening or connecting
+	## with IPv6 just works.
```
timwoj:
```suggestion
	## The ZeroMQ library has IPv6 support in ZeroMQ. For Zeek
	## we enable it unconditionally so that listening or connecting
	## with IPv6 just works.
```

And remove the extra blank `##` line below.

src/cluster/backend/zeromq/ZeroMQ-Proxy.h
```diff
@@ -54,6 +55,7 @@ class ProxyThread {
     Args args;
     std::string xpub_endpoint;
     std::string xsub_endpoint;
+    int ipv6 = 1;
```
timwoj, line 58: Is there a reason these flags aren't booleans?

timwoj (approved):

src/cluster/backend/zeromq/ZeroMQ-Proxy.h
awelzel: The ZeroMQ API takes `int` values, so this seemed appropriate here.

timwoj: I would have opted to make these `bool` for clarity on our side and pass `flag ? 1 : 0`, but leaving it like this is fine.

Issue #3900, opened 2024-08-26
Add EML feature to the SMTP analyzer


zrobinette12:
I have a use-case from Zeek to extract SMTP DATA as a file (a.k.a EML file type). EML files are undecoded SMTP MIME DATA. My organization developed a plugin that achieves this, however it's a copy pasta of the SMTP analyzer. We do this by removing the MIME_Mail analyzer from being called by the SMTP analyzer, and then do the following:

- At `ProcessData` https://github.com/zeek/zeek/blob/master/src/analyzer/protocol/smtp/SMTP.cc#L852 we'll forward the SMTP data to the file_mgr if we've reached a configurable buffer (which I think we just keep to 100mb or something).
- At `EndData` https://github.com/zeek/zeek/blob/master/src/analyzer/protocol/smtp/SMTP.cc#L900-L918 instead of doing `mail->Done()`, we also send that to the file_mgr.

Here's a specific example
```
void EML_Analyzer::EndData() {
	if (file_buf.length() > 0) {
		file_id = zeek::file_mgr->DataIn(reinterpret_cast<const u_char*>(file_buf.c_str()),
					   file_buf.length(), GetAnalyzerTag(), Conn(),
					   orig_is_sender, file_id);
		file_buf.clear();
	}

	zeek::file_mgr->EndOfFile(GetAnalyzerTag(), Conn());
	data_len = 0;
	file_id.clear();
}
```

This sends the undecoded MIME data to the file analysis framework (receives a fuid). This then lets us extract this data as a single file.

I'm having trouble maintaining this plugin and feel it's not a good solution since I'm making Zeek perform SMTP analysis twice (since of course I have other people who want the smtp.log and just the file attachments extracted as files).

TLDR:
Have some sort of config option that allows bypassing the MIME analyzer and treat SMTP DATA as a file.

PR #4591, opened 2025-06-20
Topic/cccs jsjm/mime as file eml extraction

cccs-jsjm:
The main idea is to intercept in the data in `MIME_Mail::Deliver` and send the mail data it to a `File` so that it can be extracted (or not) independently of the attachments.

The file that contains the EML file should have a distinct file id than the first entity, because it is created before the entity counter is incremented, and after the mail counter is incremented. The latter part is ensured by (https://github.com/zeek/zeek/pull/4526).

I don't love how `MIME_Mail::MailDataIn` is used to add newlines because it seems inefficient to call `zeek::file_analysis::Manager::DataIn` for a two byte string. There are several ways this could be improved, and I'd appreciate feedback on this:
1. An additional parameter with a default value to `Manager::DataIn` that adds a boolean that indicates whether a CRLF should be appended. It's not great because it doesn't address the duplication across overloads of this method.
2. Add a new method to `Manager` that allows calling `File::DataIn` several times. There are various ways to do this, but it could maybe like this: `Manager::DataInWith(file_id, source, mime_type, [](SomethingLikeAStream s) { s.write(data, len); s.write("\r\n"); }`
3. Use the fact that when `trailing_CRLF` is true, there is usually a CRLF at `data+len`. Exploiting this fact, we could bundle the pointer, the length and the bool into a struct that allows accessing the string that contains this trailing newline. This struct would be passed to `MIME_Mail::Deliver` and down the call chain, so it's not actually a huge change that would impact the signature of `zeek::analyzer::Analyzer::DeliverStream`.
4. Accumulate all the lines and write them all at the end, similar to what `MIME_Mail::SubmitAllData` does. However `MIME_Mail::SubmitAllData`.

I think it's unfortunate that the *events.log changed, but I'm not sure whether I should have tried to avoid this by instructing btest to normalize those logs. It seems like these logs contain a lot of implementation details anyway. Thought on this?


awelzel:
Thanks! Will look at it soon. Before though, does this cover #3900 or is basically the same use-case?

awelzel:
> I think it's unfortunate that the *events.log changed, but I'm not sure whether I should have tried to avoid this by instructing btest to normalize those logs. It seems like these logs contain a lot of implementation details anyway. Thought on this?

Yeah, that's expected. We've been striving towards reducing random baseline changes in that test, but the trace used contains SMTP, so everytime the smtp scripts are touched it affects the dump-events tests.  Don't worry about it.

cccs-jsjm:
> Thanks! Will look at it soon. Before though, does this cover #3900 or is basically the same use-case?

Yes, I believe so.

cccs-jsjm:
Baselines from https://github.com/zeek/zeek-testing also need updating. I'll open a PR in that repo once an initial review of this PR is done.

awelzel:
Hello @cccs-jsjm, sorry for the delay here.

I can see putting the logic into MIME_Mail to make it general, but given this is only about SMTP and POP3, I'd prefer moving it into these analyzers explicitly. Mostly because I already find the MIME_Mail code mind-twisting so would rather skip adding more extra logic to it. And it's only about SMTP, POP3 and possibly IMAP at this point.

I tired this for SMTP in the following commit, introducing a new event to recognize a new "mail data" file and it looks okay to me, even if SMTP specific. What's your take on that? Would you be up picking this up and doing it similarly for POP3? It'd also allow treating SMTP and POP3 mail data separately. 

https://github.com/zeek/zeek/commit/72be23f379ac4726b237d961b2fe343617679acb

This can then be used as follows:
```
redef SMTP::enable_mail_data_file_analysis = T;

event smtp_mail_data_file(f: fa_file) {

        print f$id, f$source, f$info$mime_type, table_keys(f$conns);
        Files::add_analyzer(f, Files::ANALYZER_EXTRACT,
                            [$extract_filename=fmt("%s.eml", f$id)]);
}
```

cc @zrobinette12 - any input from your side on this? Would this work for you and possibly test it in the future?

awelzel:
> I don't love how MIME_Mail::MailDataIn is used to add newlines because it seems inefficient to call zeek::file_analysis::Manager::DataIn for a two byte string. There are several ways this could be improved, and I'd appreciate feedback on this:

Does it just seem inefficient due to the two calls, or have you actually profiled it in some way? I'm not overly concerned, but one could store the `File *` in the SMTP analyzer and call `DataIn()` on that directly if the lookup is actually problematic.

awelzel:
> `event smtp_mail_data_file(f: fa_file) {`

I was still forth and back whether that is a good approach, but it does match the `smtp_starttls()`, `pop3_starttls()`, `mysql_starttls()`, ... pattern, so it seems like a nice match.

zrobinette12:
@awelzel @cccs-jsjm this looks great and I can test next week. My only input is a question, does this feature retain the attachments so that a single fuid is the mime mail + attachments?

cccs-jsjm:
Thanks for the feedback.
@zrobinette12 My intent was to have a distinct fuid for each mail (headers+message body+attachments) and each attachment.
@awelzel I'll have a look at proposed changes shortly.

cccs-jsjm:
@awelzel I tested the proposed changes with the test suite I had prepared for this pull request. (Just cherry-picked the last two tests I submitted earlier. I had to do some trivial updates to make them work on top of your changes.).
1. I had it setup in such a way that the `parent_id` of the attachments was the `id` of the file. It no longer does that but it should still be possible to do so. I think logging the relationship between the mail and its attachments is useful for a person that analyzes the logs.
2. I had a test that checks that the extracted file is correct (the one named `eml-reextraction`). With the implementation you proposed, there seems to be some duplicated lines, probably between mails that have been transmitted with bdat, but I didn't look into it.
3. The reason I used `mime_begin_mail` in the first place was to have a way to reset `connection$smtp_state$mail_depth` to zero. This helped ensuring the fuid of an email was distinct from the last attachment of the previous mail and also distinct from the first attachment of the current mail. With the proposed changes, there are some extracted file missing, in particular for test `eml-extraction-multi-mail`.

cccs-jsjm:
> > I don't love how MIME_Mail::MailDataIn is used to add newlines because it seems inefficient to call zeek::file_analysis::Manager::DataIn for a two byte string. There are several ways this could be improved, and I'd appreciate feedback on this:
> 
> Does it just seem inefficient due to the two calls, or have you actually profiled it in some way? I'm not overly concerned, but one could store the `File *` in the SMTP analyzer and call `DataIn()` on that directly if the lookup is actually problematic.

I did not benchmark it, so maybe it doesn't matter.

zrobinette12:
> My intent was to have a distinct fuid for each mail (headers+message body+attachments) and each attachment.
Awesome, this is exactly what I'd hoped for.

awelzel:
> @awelzel I tested the proposed changes with the test suite I had prepared for this pull request. (Just cherry-picked the last two tests I submitted earlier. I had to do some trivial updates to make them work on top of your changes.).

Thanks for testing it. I should've done a better job doing it myself.

> 
> 1. I had it setup in such a way that the `parent_id` of the attachments was the `id` of the file. It no longer does that but it should still be possible to do so. I think logging the relationship between the mail and its attachments is useful for a person that analyzes the logs.

Yeah, I liked that. It got a bit simpler in the branch I believe.


> 2. I had a test that checks that the extracted file is correct (the one named `eml-reextraction`). With the implementation you proposed, there seems to be some duplicated lines, probably between mails that have been transmitted with bdat, but I didn't look into it.

The implementation was borked :-)

> 3. The reason I used `mime_begin_mail` in the first place was to have a way to reset `connection$smtp_state$mail_depth` to zero. This helped ensuring the fuid of an email was distinct from the last attachment of the previous mail and also distinct from the first attachment of the current mail. With the proposed changes, there are some extracted file missing, in particular for test `eml-extraction-multi-mail`.

Yeah, I realized now when actually trying. I've solved that by including the top-level `mail_fuid` into the `describe_file()` hash.


Can you check if the commits in the following branch addresses most of your concerns and whether that could work?

https://github.com/zeek/zeek/tree/topic/awelzel/eml-extraction-v2

I've picked up your extraction tests and used the GitHub ID from the commits as author to add them. Let me know if you'd rather not have it this way. I passed on the reextraction tests as they were too tricky for me to understand :-)

I'm happy to open a PR for the branch if that looks better to you, or if you want to pick it up for POP3, that'd be fine, too.

cccs-jsjm:
@awelzel Thanks for this. I had a look at your second pass and it looks like it does what I would expect it to. Too bad for the reextraction test. Now that we have known good test baselines, I guess it's not really useful.

I don't think it's an issue that you reused my commit but now the signature won't verify.

I currently don't need the POP3 support for this right now, so if you want to open a PR with the changes from your branch in its current state, that would be great. Are you aware of requests for a similar feature for POP3 from the community?

awelzel:
> I currently don't need the POP3 support for this right now, so if you want to open a PR with the changes from your branch in its current state

I'll pick it up without POP3. I was also wondering if POP3 is even still relevant.

cccs-jsjm:
Closing pull request, since the same functionality has been implemented in another pull request.

zrobinette12:
Hi @awelzel and @cccs-jsjm , I just tested this out and it works exactly as I expected it to. Thank you!

awelzel:
> Hi @awelzel and @cccs-jsjm , I just tested this out and it works exactly as I expected it to. Thank you!

Cool, thank you. I presume you tested whatever landed in master? :-)

zrobinette12:
> Cool, thank you. I presume you tested whatever landed in master? :-)

I did yes, followed the script recommendation in the `NEWS` file and it worked perfectly.

PR #4592, opened 2025-06-20
broker: Handle Broker::publish() with non Broker::Event

awelzel:


src/broker/messaging.bif
```diff
@@ -66,8 +66,16 @@ static bool publish_event_args(ArgsSpan args, const zeek::String* topic,
...
-		                                      args[0]->AsRecordVal());
+		{
+		auto* rv = args[0]->AsRecordVal();
+
+		// same_type() should be fast if it's the type pointers are the same.
+		if ( zeek::same_type(rv->GetType(), zeek::BifType::Record::Broker::Event) )
+			rval = zeek::broker_mgr->PublishEvent(topic->CheckString(), rv);
+		else
+			zeek::emit_builtin_error(zeek::util::fmt("expected Broker::Event, got %s",
+			                                         zeek::obj_desc_short(rv->GetType()).c_str()));
```
bbannier: Nit: suggest validation from normal flow. Also, the comment misses a.

```suggestion
		// Validate that the passed value type matches. `same_type()`
		// should be fast since it compares type pointers.
		if ( !zeek::same_type(rv->GetType(), zeek::BifType::Record::Broker::Event) )
			{
			zeek::emit_builtin_error(
				zeek::util::fmt(
					"expected Broker::Event, got %s",
					zeek::obj_desc_short(rv->GetType()).c_str()));

			return false;
			}

		rval = zeek::broker_mgr->PublishEvent(topic->CheckString(), rv);
```

bbannier (approved):

PR #4593, opened 2025-06-20
Remove `list` from Zeek grammar

evantypanski:
Using it would make Zeek segfault (at least in locals). It has been an error for 15 years. I wasn't even born then (kidding)

I see #3684 but I guess that never landed? Or wasn't replacing the "existing" list?


awelzel (approved):

awelzel:
If it's needed for #3684 , can resurrect it I guess.

vpax:
> If it's needed for #3684 , can resurrect it I guess.

Yep, simple enough. BTW, #3684 stalled because of the need to find decent syntax for adding/removing from front/back. There's some syntax in there, but it's not all that pretty ... and I am loathe to use BiFs (or pseudo-BiFs) for this as that goes against the language's general style.

vpax:
(It also stalled because my need for it at Corelight turned out to be addressable in another fashion, so it's no longer pressing from my end.)

PR #4595, opened 2025-06-21
DNS: Implement NAPTR RR support

awelzel:
My phone is sending NAPTR queries and we reported an unknown RR type 35
in weird.log for the response, so figured I'd just add it.


src/analyzer/protocol/dns/events.bif
```diff
@@ -467,6 +467,23 @@ event dns_CAA_reply%(c: connection, msg: dns_msg, ans: dns_answer, flags: count,
...
+## an individual event of the corresponding type is raised for each.
+##
+## c: The connection, which may be UDP or TCP depending on the type of the
+##    transport-layer session being analyzed.
+##
+## msg: The parsed DNS message header.
+##
+## ans: The type-independent part of the parsed answer record.
+##
+## naptr: Record representing the NAPTR RR
```
evantypanski: nit: other parts of this file don't seem to use RR, they spell it out

scripts/base/init-bare.zeek
```diff
@@ -3048,6 +3048,20 @@ type dns_svcb_rr: record {
...
 
+## A NAPTR record.
+##
+## See also RFC 2915 - The Naming Authority Pointer (NAPTR) DNS Resource Record.
+##
+## .. zeek:see:: dns_NAPTR_reply
+type dns_naptr_rr: record {
+	order: count;        ##< Order in which to process NAPTR records.
+	preference: count;   ##< Preference specifying processing order for *equal* :zeek:field:`dns_naptr_rr$order` fields.
+	flags: string;       ##< Flags to control rewriting. E.g. "u", "a", "s" or "p".
```
evantypanski, line 3059: Are flags case-insensitive? From [the RFC](https://www.rfc-editor.org/rfc/rfc2915):

> A <character-string> containing flags to control aspects of the rewriting and interpretation of the fields in the record.  Flags are single characters from the set [A-Z0-9].

Then later they mention upper case flags and other places seem to use upper case:

>  At this time only four flags, "S", "A", "U", and "P", are defined.

But the test below has a lower case flag, so it's confusing to me. Do you know if it's just not case sensitive?

scripts/base/protocols/dns/main.zeek
```diff
@@ -537,6 +537,20 @@ event dns_SRV_reply(c: connection, msg: dns_msg, ans: dns_answer, target: string
...
 	}
 
+event dns_NAPTR_reply(c: connection, msg: dns_msg, ans: dns_answer, naptr: dns_naptr_rr) &priority=5
+	{
+	# Just encode all the fields for NAPTR RR in the reply string.
+	local tmp = "";
+	if ( |naptr$regexp| > 0 )
+		tmp += fmt("%s", naptr$regexp);
+	if ( |naptr$replacement| > 0 )
+		tmp += fmt("%s%s", |tmp| > 0 ? " " : "", naptr$replacement);
```
evantypanski: This kind of feels unclear, I'd prefer the space addition just in a condition right before this, but that could just be me

evantypanski (approved):
Seems good to me, I have a couple minor comments

scripts/base/init-bare.zeek
awelzel: Oh wow Evan, nice sleuthing, I didn't read or question too closely.

From the follow-up RFC 3403:

       A <character-string> containing flags to control aspects of the
      rewriting and interpretation of the fields in the record.  Flags
      are single characters from the set A-Z and 0-9.  ***The case of the
      alphabetic characters is not significant. *** The field can be empty.

https://datatracker.ietf.org/doc/rfc3403/

I think we should keep the casing just to not loose that information.

evantypanski: Ah, I didn't dig into further RFCs, just checked that one. That works!

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -628,6 +631,50 @@ bool DNS_Interpreter::ParseRR_SRV(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+    }
+
+    // These all check rdlength and return nullptr if there's not enough data available.
+    auto flags = extract_char_string(analyzer, data, len, rdlength);
+    auto service = extract_char_string(analyzer, data, len, rdlength);
+    auto regexp = extract_char_string(analyzer, data, len, rdlength);
+
+    // The replacement string is a name. Compression shouldn't be used, but doesn't seem
+    // we have a helper that would allow to control this.
+    u_char replacement[513];
```
timwoj, line 652: Drive-by: Should we define this length as a constant somewhere since we reuse it a bunch of times in this file?

scripts/base/protocols/dns/main.zeek
0xxon: fwiw - I think this is ok.

awelzel: I actually removed all the `fmt()` calls which weren't all that useful and did what even suggested.

src/analyzer/protocol/dns/DNS.cc
awelzel: Maybe? I agree magic numbers are annoying, but I wonder if there's a actually a better helper that could be extracted somewhere. I haven't fully dug into the DNS.cc code though, so I'll just leave it as is.

PR #4596, opened 2025-06-21
Align WebSocket error in cluster with one in Broker

### LINKED ISSUE #4594: Error from `cluster::WebSocketClient` incompatible with Broker format
Broker [documents](https://docs.zeek.org/projects/broker/en/master/web-socket.html#error-messages) the following example error:

```json
{
  "type": "error",
  "code": "deserialization_failed",
  "context": "input #1 contained malformed JSON -> caf::pec::unexpected_character(1, 1)"
}
```

The cluster bindings on the other hand would produce serialization like this:

```json
{
  "type": "error",
  "code": "deserialization_failed",
  "message": "failed to decode JSON object"
}
```

The mismatching use of `context` and `message` makes it impossible to use the same serializer/deserializer for the API exposed by `Broker::listen_websocket` and `Cluster::listen_websocket`.

I do not recall whether we renamed `context` to `message` deliberately, but if we did it should have been called out as a breaking change in `NEWS`. That said, unless we have a very good reason to break the schema `Cluster::listen_websocket` should probably stick to the existing format.

bbannier:
The schema of cluster WebSocket error messages deviated from the existing one in Broker which breaks seamless migration from the Broker WebSocket bindings.

This patch aligns the serialization in cluster with the one in Broker. This is technically a breaking change of the cluster schema, but since it never worked like documented and is still experimental this is probably fine.

Closes #4594.


src/cluster/websocket/WebSocket.cc
```diff
@@ -144,7 +144,7 @@ WebSocketClient::SendInfo WebSocketClient::SendError(std::string_view tag, std::
     *out++ = ',';
     broker::format::json::v1::append_field("code", tag, out);
     *out++ = ',';
-    broker::format::json::v1::append_field("message", message, out);
+    broker::format::json::v1::append_field("context", message, out);
```
awelzel, line 147: Thanks! I remember being a bit annoyed by `"context"` when picking up the code and apparently changed it to `"message"`. Agree strongly on keeping `"context"`. I'll mark this for 7.2, too.

awelzel (approved):

PR #4597, opened 2025-06-23
zeek-setup: Delete managers on init errors

awelzel:
TSAN complains about storage manager leaking a thread. While we use jthread and have a comment that it the thread would be joined automatically, this only happens if we actually destruct the storage_mgr instance, do so in the init errors case.

---

Not sure I like this.

This is the TSAN error:
```
[#5] broker.publish-errors ... failed on retry #1
  % 'btest-bg-wait 30' failed unexpectedly (exit code 1)
  % cat .stderr
  <<< [395317] zeek -b ../recv.zeek
  received termination signal
  >>>
  <<< [395361] zeek -b ../send.zeek
  error in ../send.zeek, line 18: expected Broker::Event, got Cluster::Event (Broker::publish(/test/topic, evt))
  ==================
  WARNING: ThreadSanitizer: thread leak (pid=395383)
    Thread T10 (tid=395717, finished) created by main thread at:
      #0 pthread_create <null> (zeek+0x666235) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d)
      #1 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State>>, void (*)()) <null> (libstdc++.so.6+0xeceb0) (BuildId: ca77dae775ec87540acd7218fa990c40d1c94ab1)
      #2 zeek::detail::TimerMgr::DoAdvance(double, int) /zeek/build/src/../../src/Timer.cc:153:16 (zeek+0x16e114f) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d)
      #3 zeek::detail::TimerMgr::Advance(double, int) /zeek/build/src/../../src/Timer.cc:85:12 (zeek+0x16e131f) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d)
      #4 zeek::detail::TimerMgr::Process() /zeek/build/src/../../src/Timer.cc:92:9 (zeek+0x16e131f)
      #5 zeek::run_state::detail::run_loop() /zeek/build/src/../../src/RunState.cc:293:28 (zeek+0x16a6f71) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d)
      #6 main /zeek/src/main.cc:96:9 (zeek+0x6ebdad) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d)
  SUMMARY: ThreadSanitizer: thread leak (/zeek/build/src/zeek+0x666235) (BuildId: 3ab0e56d9b93d07aee839e8def65b07908c1e35d) in pthread_create
  ==================
  ThreadSanitizer: reported 1 warnings
  >>>
  >>> process 395361 failed with exitcode 66: zeek -b ../send.zeek
```

### DIFF
--- a/src/zeek-setup.cc
+++ b/src/zeek-setup.cc
@@ -344,6 +344,29 @@ static void done_with_network() {
     ZEEK_LSAN_DISABLE();
 }
 
+// Just a helper to destruct all managers without terminating.
+static void delete_managers() {
+    delete zeekygen_mgr;
+    delete packet_mgr;
+    delete analyzer_mgr;
+    delete file_mgr;
+    delete cluster::manager;
+    // broker_mgr, timer_mgr, supervisor, and dns_mgr are deleted via iosource_mgr
+    delete iosource_mgr;
+    delete event_registry;
+    delete log_mgr;
+    delete reporter;
+    delete plugin_mgr;
+    delete val_mgr;
+    delete session_mgr;
+    delete fragment_mgr;
+    delete telemetry_mgr;
+#ifdef HAVE_SPICY
+    delete spicy_mgr;
+#endif
+    delete storage_mgr;
+}
+
 static void terminate_zeek() {
     util::detail::set_processing_status("TERMINATING", "terminate_zeek");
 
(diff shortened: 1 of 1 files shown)


awelzel:
@timwoj - what's tickled by TSAN is that the storage manager starts the expiration timer at `InitPostScript()` which starts a thread, but when there's init errors, it's not joined properly. The change I did isn't sufficient as just deleting the managers without properly terminating tickles assert's in various Manager destructors.

Any thoughts what to do?

timwoj:
> @timwoj - what's tickled by TSAN is that the storage manager starts the expiration timer at `InitPostScript()` which starts a thread, but when there's init errors, it's not joined properly. The change I did isn't sufficient as just deleting the managers without properly terminating tickles assert's in various Manager destructors.

The thread doesn't get started until after the timer moves forward when the event loop starts. Otherwise its just a timer waiting for the time to move. I'm sort of surprised that it only complains about the storage thread since there are other threads that get started up too, such as the writer thread. 

I don't think where you made the change in this PR is the right place to do it, since that's for parsing/init errors. The shutdown here doesn't happen until after processing has started. Unfortunately, there really isn't a nice place to clean up since we're calling `exit()` in the test that's tripping this. `terminate()` would go through the regular shutdown process, but the exit BIF just calls the system's `exit()` method and the process ends.

awelzel:
> terminate() would go through the regular shutdown process, but the exit BIF just calls the system's exit() method and the process ends.

Lets do that then. Will open a PR momentarily.

PR #4599, opened 2025-06-24
Only pass session ticket data in ssl_session_ticket_handshake event

0xxon:
This commit fixes the parsing of the data field in the SSL analyzer. So far, this field contained two extra bytes at the beginning, which contain the length of the following data.

Now, the data passed to the event only contains the actual value of the session ticket.

The Spicy analyzer already contains the correct handling of this field, and does not need to be updated. A test that uses the event and exhibited the bug was added.


testing/btest/scripts/base/protocols/ssl/session-ticket.test
```diff
@@ -0,0 +1,13 @@
...
+# @TEST-EXEC: zeek -b -C -r $TRACES/tls/CVE-2015-3194.pcap %INPUT
+# @TEST-EXEC: echo "client-certificate.pcap"
+# @TEST-EXEC: zeek -b -C -r $TRACES/tls/client-certificate.pcap %INPUT
+# @TEST-EXEC: btest-diff .stdout
+
+@load base/protocols/ssl
+
+event ssl_session_ticket_handshake(c: connection, ticket_lifetime_hint: count, ticket: string)
+	{
+	print ticket_lifetime_hint, ticket;
```
awelzel, line 12: Totally minor, I wonder if it's sufficient to print `|ticket|` and `ticket[:8]` and `ticket[-8:]` or some such rather than the full ticket for a bit of a smaller baseline.

NEWS
```diff
@@ -51,6 +51,11 @@ Breaking Changes
 - The ``IsPacketSource()`` method on ``IOSource`` was removed. It was unused
   and incorrectly returned ``false`` on all packet sources.
 
+- The ``ssl_session_ticket_handshake`` event handling of session ticket data was fixed.
```
awelzel: Minor, you could also phrase that around being an issue of the parsing rather than the event handling.

E.g. "parsing of session ticket handshake" data was fixed. Previously the data string mistakenly included the 2 byte length field."

testing/btest/scripts/base/protocols/ssl/session-ticket.test
0xxon: True in principle. However, I like that we have the full ticket in a single test, just in case we ever mess something else up about it. Like the total length.

NEWS
0xxon: Oh, good idea, thank you.

PR #4601, opened 2025-06-24
Spicy SSL analyzer: make record layer version parsing more strict

0xxon:
The Spicy based SSL analyzer was, so far, more permissive with the record layer versions that it would accept.

This change brings the parsing of record layer versions in line with the binpac based analyzer. This behavioral difference was discovered due to a test that changed with the recent dpd log changes.

### DIFF
--- a/src/analyzer/protocol/ssl/spicy/SSL.spicy
+++ b/src/analyzer/protocol/ssl/spicy/SSL.spicy
@@ -740,7 +740,7 @@ type SSL2ClientMasterKey = unit(length: uint16, inout sh: Share) {
 # For TLS-y protocols - determine how to continue
 type TLSRecordFragmentChoice = unit(content_type: uint8, handshakesink: sink&, alertsink: sink&, inout msg: Message, inout sh: Share) {
     # content_type: uint8; # &convert=ContentType($$);
-    version: uint16;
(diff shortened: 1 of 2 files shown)


src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -740,7 +740,7 @@ type SSL2ClientMasterKey = unit(length: uint16, inout sh: Share) {
 # For TLS-y protocols - determine how to continue
 type TLSRecordFragmentChoice = unit(content_type: uint8, handshakesink: sink&, alertsink: sink&, inout msg: Message, inout sh: Share) {
     # content_type: uint8; # &convert=ContentType($$);
-    version: uint16;
+    version: uint16 &requires=($$ == SSLv3 || $$ == TLSv10 || $$ == TLSv11 || $$ == TLSv12 : "Invalid version in record layer. Version: %s" % $$);
```
bbannier, line 743: Another way to write this would be

```ruby
    version: uint16 &requires=($$ in set(SSLv3, TLSv10, TLSv11, TLSv12) : "Invalid version in record layer. Version: %s" % $$);
```

This could have different runtime behavior since the `set` would be computed each time we check a fragment version, so fine to keep as is.

PR #4607, opened 2025-06-25
Extend script coverage profiling to track @if conditionals

vpax:
This PR is in response to (the tail end of) the https://github.com/zeek/zeek/discussions/2966 discussion thread. One of the problems that that thread examines is the need when using code coverage to identify that blocks of code exist but were never parsed (and hence for sure never covered) due to residing inside `@if` conditionals (or their `@else` branches) that weren't active. Here we add to the profiling output a list of conditionals annotated by an execution "count" of `0` if the conditional was false (and thus its body skipped) and `1` if true. These can then be post-processed to inspect whether during testing each conditional was both for one run true, and for another false (the ideal state).


testing/btest/Baseline/coverage.zeek-profiler-file/step1.out
```diff
@@ -1,3 +1,4 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-1	./profiling-test1.zeek, line 2	print new conn;
-1	./profiling-test1.zeek, lines 1-2	event new_connection BODY
+1	./profiling-test1.zeek, line 2	CONDITIONAL
```
awelzel: I was definitely overthinking what you wanted.

I've pushed topic/vern/if-coverage-v2 with one additional commit to render `CONDITIONAL` to be closer to the original script:

```
2       ./profiling-test1.zeek, line 2  @if ( T )
2       ./profiling-test1.zeek, line 4  print new conn;
2       ./profiling-test1.zeek, lines 3-4       event new_connection BODY
0       ./profiling-test2.zeek, line 1  @if ( F )
0       ./profiling-test2.zeek, line 11 @ifndef ( Conn::Info )
1       ./profiling-test2.zeek, line 13 @else
1       ./profiling-test2.zeek, line 14 event zeek_init BODY
1       ./profiling-test2.zeek, line 14 print Conn::Info;
1       ./profiling-test2.zeek, line 2  @else
1       ./profiling-test2.zeek, line 4  print new conn;
1       ./profiling-test2.zeek, line 6  @ifdef ( Conn::Info )
1       ./profiling-test2.zeek, line 7  event zeek_init BODY
1       ./profiling-test2.zeek, line 7  print Conn::Info;
0       ./profiling-test2.zeek, line 8  @else
1       ./profiling-test2.zeek, lines 3-4       event new_connection BODY
```

Okay to fold that into your branch, then merge? You'll likely need to change your postprocessing, but the result seems nicer.

https://github.com/zeek/zeek/commit/d9d64ed928b3bf87926e21ca0f9487c1322e85ab

vpax: Nice! Please go ahead and fold it in (or I can if that's easier).

PR #4608, opened 2025-06-26
Cluster Telemetry Follow Up

awelzel:


src/cluster/websocket/WebSocket.cc
```diff
@@ -303,7 +303,6 @@ void WebSocketEventDispatcher::Process(const WebSocketOpen& open) {
     });
 
     if ( ! good_application_name ) {
```
bbannier, line 305: Just FTR, like discussed offline, I don't think `good_application_name` is ideal since it leaks Prometheus label restrictions into something only tangentially related (the names users want to give their websocket clients), and I'd tend more towards just sanitizing the name once it gets exported as a metric (but that could lead to mismatches between app names and metric labels).

If we keep it we should document `X-Application-Name` and its rules in the WebSocket API docs, probably as part of https://github.com/zeek/zeek-docs/issues/331.

testing/btest/cluster/telemetry/ws-invalid-app.zeek
```diff
@@ -0,0 +1,65 @@
...
+#
+# @TEST-EXEC: cp $FILES/zeromq/cluster-layout-simple.zeek cluster-layout.zeek
+# @TEST-EXEC: cp $FILES/zeromq/test-bootstrap.zeek zeromq-test-bootstrap.zeek
+# @TEST-EXEC: cp $FILES/ws/wstest.py .
+#
+# @TEST-EXEC: zeek -b --parse-only manager.zeek
+# @TEST-EXEC: python3 -m py_compile client.py
+#
+# @TEST-EXEC: btest-bg-run manager "ZEEKPATH=$ZEEKPATH:.. && CLUSTER_NODE=manager zeek -b ../manager.zeek"
+# @TEST-EXEC: btest-bg-run client python3 ../client.py
```
bbannier: No need to put this into background, we should be able to use something like

```suggestion
# @TEST-EXEC: python3 ./client.py >output 2>&1
```

src/RE.cc
```diff
@@ -189,17 +189,24 @@ std::string Specific_RE_Matcher::LookupDef(const std::string& def) {
     return {};
 }
 
-bool Specific_RE_Matcher::MatchAll(const char* s) { return MatchAll((const u_char*)(s), strlen(s)); }
+bool Specific_RE_Matcher::MatchAll(const char* s) { return MatchAll(std::string_view{s}); }
 
 bool Specific_RE_Matcher::MatchAll(const String* s) {
     // s->Len() does not include '\0'.
     return MatchAll(s->Bytes(), s->Len());
```
bbannier: We could clean this up as well

```suggestion
    return MatchSet(s->ToStdStringView(), matches);
```

src/RE.h
```diff
@@ -89,6 +90,7 @@ class Specific_RE_Matcher {
 
     bool MatchAll(const char* s);
     bool MatchAll(const String* s);
+    bool MatchAll(std::string_view sv);
```
bbannier, line 93: Sorry to pile more onto this, but could you add overloads taking `std::string_view` for `Match` and `LongestMatch` for consistency? We should then be able to reimplement their `const char*` and `const String*` overloads in terms of `std::string_view` as well.

src/cluster/Telemetry.cc
```diff
@@ -131,28 +127,41 @@ void VerboseTelemetry::OnIncomingEvent(std::string_view topic, std::string_view
 
 namespace {
 
-std::string determine_script_location() {
-    std::string result = "none";
-
-    if ( zeek::detail::call_stack.empty() )
-        return result;
+// Cache CallExpr pointers to their location.
+std::map<const zeek::detail::CallExpr*, std::string> location_cache;
```
bbannier: I'd just move this into `determine_script_location` as a `static` into the function using it which should behave identical, but would be easier to follow.

I do not know whether `CallExpr*` are equivalent to script locations, but I assume you checked.

src/cluster/Telemetry.cc
```diff
@@ -131,28 +127,41 @@ void VerboseTelemetry::OnIncomingEvent(std::string_view topic, std::string_view
...
         const auto* func = zeek::detail::call_stack[sidx].func;
         const auto* ce = zeek::detail::call_stack[sidx].call;
 
-        // without_zeekpath_component looks pretty expensive and might be
-        // better to cache the result using the ce pointer instead of computing
-        // it over and over again.
+        // Cached?
+        auto it = location_cache.find(ce);
+        if ( it != location_cache.end() )
+            return it->second;
```
bbannier: I'd try to avoid reusing `it` later and instead write

```suggestion
        if ( auto it = location_cache.find(ce); it != location_cache.end() )
            return it->second;
```

src/cluster/Telemetry.cc
```diff
@@ -131,28 +127,41 @@ void VerboseTelemetry::OnIncomingEvent(std::string_view topic, std::string_view
...
+        // }
+
         const auto* loc = ce->GetLocationInfo();
         std::string normalized_location = zeek::util::detail::without_zeekpath_component(loc->filename);
-        result = normalized_location + ":" + std::to_string(loc->first_line);
-        break;
+        normalized_location = normalized_location + ":" + std::to_string(loc->first_line);
+
+        bool inserted = false;
+        std::tie(it, inserted) = location_cache.emplace(ce, std::move(normalized_location));
```
bbannier: This then becomes

```suggestion
        auto [it, inserted] = location_cache.emplace(ce, std::move(normalized_location));
```

src/cluster/Telemetry.cc
```diff
@@ -131,28 +127,41 @@ void VerboseTelemetry::OnIncomingEvent(std::string_view topic, std::string_view
...
+        // if ( ignore func ) {
+        //     --sidx;
+        //      continue;
+        // }
+
         const auto* loc = ce->GetLocationInfo();
         std::string normalized_location = zeek::util::detail::without_zeekpath_component(loc->filename);
-        result = normalized_location + ":" + std::to_string(loc->first_line);
-        break;
+        normalized_location = normalized_location + ":" + std::to_string(loc->first_line);
```
bbannier: Nit:

```suggestion
        normalized_location.append(std::to_string(loc->first_line));
```

src/cluster/Telemetry.cc
```diff
@@ -131,28 +127,41 @@ void VerboseTelemetry::OnIncomingEvent(std::string_view topic, std::string_view
...
+            return it->second;
+
+        // Future: Ignore wrapper functions if we ever come across some.
+        // We only care about Broker::publish() and Cluster::publish() and
+        // these aren't wrapped, so currently nothing to do here.
+        //
+        // if ( ignore func ) {
+        //     --sidx;
+        //      continue;
+        // }
```
bbannier, line 152: At least to me it is not clear when something needs to happen here, should we just drop this comment?

testing/btest/cluster/telemetry/ws-invalid-app.zeek
awelzel: Perfect, thanks. I'll try to remember for the next time, you've brought this up a few times now.

src/RE.h
awelzel: I figured if someone needs this they'd just do it, but it's true, that RE.h might be used by plugins and it's an actual API. Having these for consistency makes sense. Thanks!

src/cluster/Telemetry.cc
awelzel: I'm going to keep this. Let me extend the comment to talk about `Func.cc::emit_builtin_error_common()` which does some magic in that area.

PR #4609, opened 2025-06-26
Redis analyzer: Handle RESP3 and out-of-band data

evantypanski:
Two main points:

- This adds handling for RESP3 out-of-band push data, since it's not part of the command/reply model. This also kind of applies to RESP2, where when you're in subscribed mode, everything is out of band... except `PING`, but I can ignore that. Technically `SUBSCRIBE` can be correlated, but it's out-of-band in RESP3, so I'll treat it that way with RESP2.
- Handles RESP3 attributes, which is a separate type that comes before the data it applies to. This was added after I started the analyzer and got the list of data types. 

The RESP3 pubsub pcap is kind of large, but it needs the connection start to get the optional `HELLO` with the message. Well, technically the analyzer handles RESP3 responses with the `min_protocol_version` - but I also want to test the `HELLO` handshake... (then the big part is the following `COMMANDS` command).

The analyzer is getting a bit annoying to build up with the split between Redis and RESP. I'd want to do more in the script layer, but it's actually tricky since we can't send recursive data through the evt file right now. :'(

### DIFF
--- a/scripts/base/protocols/redis/main.zeek
+++ b/scripts/base/protocols/redis/main.zeek
@@ -74,17 +74,17 @@ export {
 	option max_pending_commands = 10000;
 
 	# These commands enter subscribed mode
-	global enter_subscribed_mode = [KnownCommand_PSUBSCRIBE,
-	    KnownCommand_SSUBSCRIBE, KnownCommand_SUBSCRIBE];
+	global enter_subscribed_mode = [RedisCommand_PSUBSCRIBE,
+	    RedisCommand_SSUBSCRIBE, RedisCommand_SUBSCRIBE];
 
 	# These commands exit subscribed mode
-	global exit_subscribed_mode = [KnownCommand_RESET, KnownCommand_QUIT];
+	global exit_subscribed_mode = [RedisCommand_RESET, RedisCommand_QUIT];
 
 	# These commands don't expect a response (ever) - their replies are out of band.
-	global no_response_commands = [KnownCommand_PSUBSCRIBE,
-	    KnownCommand_PUNSUBSCRIBE, KnownCommand_SSUBSCRIBE,
-	    KnownCommand_SUBSCRIBE, KnownCommand_SUNSUBSCRIBE,
-	    KnownCommand_UNSUBSCRIBE];
+	global no_response_commands = [RedisCommand_PSUBSCRIBE,
+	    RedisCommand_PUNSUBSCRIBE, RedisCommand_SSUBSCRIBE,
+	    RedisCommand_SUBSCRIBE, RedisCommand_SUNSUBSCRIBE,
+	    RedisCommand_UNSUBSCRIBE];
 }
 
 redef record connection += {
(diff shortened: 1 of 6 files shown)


scripts/base/protocols/redis/main.zeek
```diff
@@ -139,10 +173,29 @@ event Redis::command(c: connection, cmd: Command)
...
+		{
+		if ( c$redis_state$resp_version == RESP2
+		    && cmd$known in enter_subscribed_mode )
+			{
+			c$redis_state$subscribed_mode = T;
+			}
+		if ( c$redis_state$resp_version == RESP2
+		    && cmd$known in exit_subscribed_mode )
+			{
+			c$redis_state$subscribed_mode = F;
```
timwoj: Should this report a warning or something if the command is in both sets? Also, the `resp_version` check could just wrap the two `known` checks in a single `if`.

timwoj (approved):
Looks fine to me, with one question.

scripts/base/protocols/redis/main.zeek
evantypanski: > Should this report a warning or something if the command is in both sets?

Sure, that works. It's currently not marked `&redef` but it would be worth a warning if it happened

scripts/base/protocols/redis/main.zeek
```diff
@@ -139,10 +173,30 @@ event Redis::command(c: connection, cmd: Command)
...
+		{
+		if ( c$redis_state$resp_version == RESP2 )
+			{
+			local should_enter = cmd$known in enter_subscribed_mode;
+			local should_exit = cmd$known in exit_subscribed_mode;
+			c$redis_state$subscribed_mode = should_enter && ! should_exit;
+
+			# Warn if it's in both
+			if ( should_enter && should_exit )
+				Reporter::warning(fmt("Command '%s' enters then immediately exits subscribed mode",
```
awelzel: Maybe make this a "weird" instead?

evantypanski: Is it better as a weird? Right now there should be no way to trigger since there just aren't commands that enter then exit, this will only realistically come up during development or if we allow redefing and someone adds a command in both enter and exit. Both of those I would classify more as a warning than a weird, but I'm not sure if other analyzers would go the weird route?

awelzel: > Right now there should be no way to trigger since there just aren't commands that enter then exit

If it does happen, then a weird is rate-limited automatically IIRC. A `Reporter::warning()` won't and would potentially spill logs (and stderr by default). So if it was to happen 10 times per second on 10 workers or so, that'd be quite annoying.

You sound fairly confident that it won't happen, so feel free to ignore, but if you grep though base/protocols for `Reporter`, it would be reasonable to stick with a weird for consistency.

awelzel: If it's more of an `assert`, maybe you could use `assert` :man_shrugging:

awelzel (approved):

scripts/base/protocols/redis/main.zeek
evantypanski: I'll just use a weird - it shouldn't happen now, but I want to work through providing users a way to extend commands (and, subsequently, those values). Since it'd be "weird" either way but shouldn't change anything if intentional, that seems to fit fine. I don't really have super strong opinions here, since I don't personally plan on adding something that enters then immediately exits subscribed mode :)

PR #4612, opened 2025-06-27
Fix clang-tidy readability-isolate-declaration warnings

bbannier:
I missed one of these in review so a machine is probably better at catching them.

I fixed the existing instances which where largely in code which look dated. Where possible I slightly reorganized the code so we do not have to leave values uninitialized, but did not touch up anything else.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,6 +2,7 @@ Checks: [-*,
          bugprone-*,
          performance-*,
          modernize-*,
+         readability-isolate-declaration,
 
          # Enable a very limited number of the cppcoreguidelines checkers.
          # See the notes for some of the rest of them below.
--- a/src/Base64.cc
+++ b/src/Base64.cc
@@ -217,17 +217,19 @@ String* decode_base64(const String* s, const String* a, Connection* conn) {
     }
 
     int buf_len = int((s->Len() + 3) / 4) * 3 + 1;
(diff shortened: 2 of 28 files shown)


timwoj (approved):
Thanks!

bbannier:
FTR, the code I was after was https://github.com/zeek/zeek/blob/e79ae5a1d9f996e4c98346664cf808c2e318be72/src/cluster/Telemetry.h#L178-L179

It looks like there is no lint which catches that.

timwoj:
I wonder if the split line is tripping it up. Strange that the analysis doesn't catch that though.

timwoj:
i built this locally and code generated by binpac is returning a bunch of warnings too. I'll look into that.

EDIT: Fixed the one from the generate code in the linked PR. I also cherry-picked in a few other fixes into this PR.

src/analyzer/protocol/netbios/functions.bif
```diff
@@ -27,7 +27,8 @@ function decode_netbios_name%(name: string%): string
 
 	unsigned char buf[16];
 	const u_char* s = name->Bytes();
-	int i, j;
+	int i;
+	int j;
 	int length = 0;
```
bbannier, line 30: This is really unnecessarily convoluted, what do you think about

```cpp
	int j = 0;
	for ( int i = 0; i < 16; ++i )
```

PR #4614, opened 2025-06-27
Add ctx to conn_id

awelzel:
This PR adds a new `ctx: conn_id_ctx` to the `conn_id` record type. Background is in #4605.


vpax:
Does this _again_ change the default layout of `conn.log`? From the diffs it seems like it does.

awelzel:
> Does this again change the default layout of conn.log? From the diffs it seems like it does.

`conn.log` does not change by default. The rendering of `conn_id` if one prints it does. The blast radius should be smaller.

The last time this came up I suggested that the tests that are sensitive to printing should be changed to only print the fields of `conn_id` they care about. We're absolutely not good at this in Zeek's btest itself, but particularly if downstream aims at supporting multiple Zeek versions, I still stand behind that. There's also `assert` now that can be used to test the expected value for specific fields, rather than using baselines.

Some `fuids` change due to including `cat(conn_id)` before hashing and I'm annoyed by this, but I'd rather work around that case specifically if that's a problem.

Your message sounds like frustration. I think the feature is too nice to not do it, but if it's unbearable we do not need to do it.

vpax:
My concern was in particular about `conn.log`, as the recent `ip_proto` addition to the layout has been a headache for our baselines. What you sketch sounds okay to me. I agree about fixing `conn_id` access to be explicit, and while `fuids` changing is a bit of a pain, I already think it's unwise to rely on UIDs being invariant across runs, so am looking to migrate out testing to not assume they're fixed anyway.

awelzel:
@rsmmr - mind looking this through? I was thinking of adding one preparation commit to not print `c$id`, but just the addresses and ports so the huge diff churn goes away. EDIT: But skipped on it as I'm not sure that's all that useful.

src/ConnKey.h
```diff
@@ -72,9 +80,13 @@ class ConnKey {
...
      *
-     * The default implementation does nothing.
+     * Plugins should override this method to set connection key specific fields
+     * on the conn_id_ctx and if there's a use, also on conn_id.
+     *
+     * @param conn_id The conn_id script-layer record to populate.
+     * @param ctx The conn_id$ctx script-layer record to populate.
      */
-    virtual void DoPopulateConnIdVal(RecordVal& conn_id) {}
+    virtual void DoPopulateConnIdVal(RecordVal& conn_id, RecordVal& ctx) {}
```
rsmmr, line 94: Wondering if we should just pass `ctx` here.  Is there really still a use case for modifying `conn_id` itself as well? If it's just logging, could that be done via script by copying values out of `ctx` before anything is written out?

Reason is that I think the API would become quite a bit clearer if there was no choice. One just wouldn't need to think about `conn_id`, which is usually the wrong path to go down. (Supporting evidence: The doc strings feel a bit confusing as they are talking about both `conn_id` and `ctx`.)

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
```diff
@@ -100,28 +99,22 @@ zeek::expected<zeek::ConnKeyPtr, std::string> Factory::DoConnKeyFromVal(const ze
...
-    int vlan_offset, inner_vlan_offset;
-    if ( rt == id::conn_id ) {
-        std::tie(vlan_offset, inner_vlan_offset) = k->GetConnIdFieldOffsets();
-    }
-    else {
-        // We don't know what we've been passed.
-        vlan_offset = rt->FieldOffset("vlan");
-        inner_vlan_offset = rt->FieldOffset("inner_vlan");
-    }
+    // Should be done in superclass!
```
rsmmr: Not sure what that comment means/suggests?

scripts/policy/protocols/dhcp/software.zeek
```diff
@@ -46,7 +46,8 @@ event DHCP::log_dhcp(rec: DHCP::Info)
 		                     $orig_p=rec$client_port,
 		                     $resp_h=rec$server_addr,
 		                     $resp_p=rec$server_port,
-				     $proto=17]; # DHCP is typically UDP
+				     $proto=17, # DHCP is typically UDP
+				     $ctx=conn_id_ctx()];  # TODO: Propagate ctx from original conn_id.
```
rsmmr: Is that initialization just so that the TODO gets recorded (vs. relying on `&default`)? And what's the plan for the TODO?

src/ConnKey.h
awelzel: > Wondering if we should just pass ctx here. Is there really still a use case for modifying conn_id itself as well? If it's just logging, could that be done via script by copying values out of ctx before anything is written out?

The logging was my main thought, really. And the fact that one can't go from `ctx` to `conn_id` - we don't have a parent. I guess I'm okay dropping it, it was a bit due to the initial design being `conn_id` based, but maybe that's just irrelevant now.

I'll add removal as a separate commit on top.

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
awelzel: Let me ditch it, I'll see if the follow-up to remove `conn_id` naturally leads to those changes - which I expect.

scripts/policy/protocols/dhcp/software.zeek
awelzel: Ah, that was before there was `&default`, I think I can just drop it, or leave a pointer to below discussion.

As of a plan, I've started a discussion about making base scripts "ctx" aware - that could become separate issues or one bigger issue with multiple commits: https://github.com/zeek/zeek/discussions/4613

Not for 8.0 though.

src/ConnKey.h
awelzel: After looking more, I'm heavily leaning on keeping `conn_id` in the API here: The `IPBasedConnKey` instances hold the 5-tuple's addr and port.  We don't do it currently, but `DoPopulateConnIdVal()` is the method that should fill out the conn_id's fields orig_h, resp_h, etc as well. From that perspective, having both records in the API makes some sense. It's true that specialization of `IPBasedConnKey` could hide it away, but that's independent of the `ConnKey` API. I'm also partly on the "provides more flexibility even if the interface is confusing".

I'm going to keep both parameters, but move populating the conn_id's values[ from Conn.cc into the IPBasedConnKey](https://github.com/zeek/zeek/blob/e3b0d1d2be440303d3bdad8923e8e2e9ac33eb51/src/Conn.cc#L226-L232) implementation so that there's at least an internal user for the conn_id parameter. Will also try to polish the doc strings a bit :-/

awelzel:
> Looks good to me, mostly straight forward. My main question is the API regarding the need to pass conn_id to the hook as well.

@rsmmr - would you be up for taking another look? I've kept the `ctx` as motivated in the other comment and moved populating `conn_id` from `Conn` to the `IPBasedConnKey` so now there's at least a reason why a `ConnKey` implementation would need access to the `conn_id` record value as well. I do think the freedom/flexibility is interesting, even if it makes the API less straight-forward to use. The FlipRoles() move is a bit noisy, but I think that's reasonable and the layout of a `conn_id` mostly confined to IPBasedConnKey now.

If this looks reasonable, I'd update the tutorial in the docs.

awelzel:
> rssmr - would you be up for taking another look? 

Ah, I just realized Robin is out for a few days. @0xxon / @timwoj  - Robin seemed generally okay with the approach. I opted to keep `conn_id` in the API after pondering and added a few follow-up commits after ([input/Manager: Ignore empty record types](https://github.com/zeek/zeek/pull/4614/commits/da13087cb910a95107a0585c0eb70bcd296e66af). Feel comfortable leaving a sign-off?

The benchmarker shows that the micro-benchmarks instantiating `c: connection` records got a few percent slower, but the impact on actual PCAP reading seems acceptable to non-existent to me. So I'm not overly concerned here. The benchmarker data is somewhat noisy so might have gotten a lucky run, because I'd certainly have expected a bit of more overhead across the board.

Screenshot from 2025-07-02 09-24-24

src/input/Manager.cc
```diff
@@ -881,13 +881,14 @@ bool Manager::UnrollRecordType(vector<Field*>* fields, const RecordType* rec, co
         if ( rec->GetFieldType(i)->Tag() == TYPE_RECORD ) {
             string prep = nameprepend + rec->FieldName(i) + ".";
 
-            if ( rec->FieldDecl(i)->GetAttr(zeek::detail::ATTR_OPTIONAL) ) {
```
0xxon, line 884: This is not really a problem - but I wonder what causes this to end up with an optional attribute...

Do you think we should add a separate test to make sure this case keeps working?

src/packet_analysis/protocol/ip/conn_key/fivetuple/Factory.cc
```diff
@@ -12,44 +13,22 @@ namespace zeek::conn_key::fivetuple {
...
-            return unexpected_conn_id;
-        }
-        // clang-format on
-    }
-
     if ( ! vl->HasField(orig_h) || ! vl->HasField(resp_h) || ! vl->HasField(orig_p) || ! vl->HasField(resp_p) ||
-         ! vl->HasField(proto) ) {
-        return unexpected_conn_id;
-    }
+         ! vl->HasField(proto) )
```
0xxon: Given the type definitions, this should not be able to happen under any circumstances, right?

If that's the case - should this perhaps also assert / throw an fatal error? Seems like the only way this can happen is some kind of memory corruption.

Also - independently - should this also check for the ctx field?

src/input/Manager.cc
awelzel: There is `scripts.base.frameworks.input.errors` which apparently produces the message, so I figured it's already tested. You might have something else in mind?

src/packet_analysis/protocol/ip/conn_key/fivetuple/Factory.cc
awelzel: Yeah, good call. I think you're right that this can't be an issue if we enforce `conn_id`, but let me try it...

src/input/Manager.cc
0xxon: Oh, I was just curious if we should have a separate test-case that includes an empty record - apparently this was the first time ever anyone tried this with the input framewok.

src/packet_analysis/protocol/ip/conn_key/fivetuple/Factory.cc
awelzel: > Given the type definitions, this should not be able to happen under any circumstances, right?

Hmm, we have the following in bifs/lookup_connection.zeek:
```
local my_id: conn_id;
lookup_connection(my_id);
```
And that has `orig_h`, `resp_h` missing and would run into an abort. While seemingly  unreasonable use, it shouldn't be possible to tickle aborts from Zeek scripts that easily, so I'm keeping the existing code, but adding a fixup to include `ctx` here.

Resolving. Shout if you disagree :-)

0xxon: Well, today I learned :)

awelzel:
> Could you perhaps add a bit more to NEWS? This does render some of the wording in the connection key update a bit outdated.

Pushed - let me know if that reads better now. I presume there'll be a pre-8.0 NEWS polish, but was definitely good to add a paragraph or two.

> While I like the feature, I also wonder what's the best way to explain which things in Zeek will not automatically be aware of the addotional connection key entries, and which parts won't be - but that' a bit of a secondary problem.

There's a tutorial (which I'll update once this PR is merged).

https://docs.zeek.org/en/master/devel/plugins/connkey-plugin.html

IMO, this is pretty low-level and powerful stuff and I'm thinking (and actually hoping) we'll be surprised how it's used, rather than explaining in detail how to use it.

0xxon (approved):

PR #4615, opened 2025-06-28
Fixes for cppcoreguidelines-pro-type-cstyle-cast

bbannier:


auxil/bifcl
bbannier, line 1: This pulls in https://github.com/zeek/bifcl/pull/42.

auxil/binpac
bbannier, line 1: This pulls in https://github.com/zeek/binpac/pull/48.

timwoj:
I'm torn on this one. I'd like to just say rip the bandaid off and merge it wholesale. The other part of me says maybe we should merge just the non-`reinterpret` ones to start with. Or maybe at least split them into separate commits to make reviewing easier?

bbannier:
> I'm torn on this one. I'd like to just say rip the bandaid off and merge it wholesale. The other part of me says maybe we should merge just the non-`reinterpret` ones to start with. Or maybe at least split them into separate commits to make reviewing easier?

I'd vote for the wholesale approach. While I could create multiple patches out of this there isn't really any organic progression, so I am unsure it would add much.

In general I believe it would be worthwhile using proper C++ casts since C-style casts just hide too much complexity, e.g., from the simplest of being unsure which parts the cast applies to if its operand is more complex, to the outright hiding bugs.

bbannier:
Setting this ready for review, the remaining clang-tidy diagnostic should be fixed by https://github.com/zeek/zeek/pull/4880.

src/CompHash.cc
```diff
@@ -386,7 +386,8 @@ bool CompositeHash::RecoverOneVal(const HashKey& hk, Type* t, ValPtr* pval, bool
                 hk.EnsureReadSpace(n);
             }
 
-            *pval = make_intrusive<StringVal>(new String((const byte_vec)hk.KeyAtRead(), n, true));
+            *pval =
+                make_intrusive<StringVal>(new String(reinterpret_cast<const unsigned char*>(hk.KeyAtRead()), n, true));
```
timwoj, line 390: You kept the uses of `byte_vec` later in ZeekString.cc. Any reason to remove this one?

src/analyzer/protocol/dce-rpc/dce_rpc-analyzer.pac
```diff
@@ -88,9 +88,9 @@ refine connection DCE_RPC_Conn += {
 			// Remove the null from the end of the string if it's there.
 			if ( ${bind.sec_addr}.length() > 0 &&
 			     *(${bind.sec_addr}.begin() + ${bind.sec_addr}.length()) == 0 )
-				sec_addr = zeek::make_intrusive<zeek::StringVal>(${bind.sec_addr}.length()-1, (const char*) ${bind.sec_addr}.begin());
+				sec_addr = zeek::make_intrusive<zeek::StringVal>(bind->sec_addr().length()-1, reinterpret_cast<const char*>(bind->sec_addr().begin()));
```
timwoj: This should keep the `$bind` naming for consistency. Same below, and in dhcp-options.pac.

src/analyzer/protocol/mime/MIME.cc
```diff
@@ -378,7 +378,9 @@ MIME_Multiline::~MIME_Multiline() {
     delete_strings(buffer);
 }
 
-void MIME_Multiline::append(int len, const char* data) { buffer.push_back(new String((const u_char*)data, len, true)); }
+void MIME_Multiline::append(int len, const char* data) {
+    buffer.push_back(new String(reinterpret_cast<const u_char*>(data), len, true));
```
timwoj, line 382: This feels like a really consistent pattern (typecasting to `const u_char` for a String constructor). There might be a good follow-up to add a constructor that takes `const char*` with a length and add-NULL boolean.

src/analyzer/protocol/netbios/NetbiosSSN.cc
```diff
@@ -106,11 +106,11 @@ void NetbiosSSN_Interpreter::ParseBroadcast(const u_char* data, int len, bool is
     // FIND THE NUL-TERMINATED NAME STRINGS HERE!
     // Not sure what's in them, so we don't keep them currently.
 
-    String* srcname = new String((char*)data);
+    String* srcname = new String(reinterpret_cast<const char*>(data));
```
timwoj, line 109: On second thought, maybe just a pass over everywhere we construct a `String`, since this one should just fall into the `string_view` constructor without a typecast.

src/analyzer/protocol/smb/smb1-com-transaction2-secondary.pac
```diff
@@ -16,8 +16,8 @@ refine connection SMB_Conn += {
 	args->Assign(7, ${val.data_displacement});
 	args->Assign(8, ${val.FID});
 
-	auto parameters = zeek::make_intrusive<zeek::StringVal>(${val.parameters}.length(), (const char*)${val.parameters}.data());
-	auto payload = zeek::make_intrusive<zeek::StringVal>(${val.data}.length(), (const char*)${val.data}.data());
+	auto parameters = zeek::make_intrusive<zeek::StringVal>(val->parameters().length(), reinterpret_cast<const char*>(val->parameters().data()));
```
timwoj: Same re: $-naming.

src/CompHash.cc
bbannier: `byte_vec` is a non-`const` `u_char*` while `KeyAtRead` returns a `const void*`, so AFAICT one cannot directly cast to one to the other. The situation is different for the code in `ZeekString.cc` where we cast from a non-const `char*`.

The following would work, but is a mouthful:

```suggestion
                make_intrusive<StringVal>(new String(reinterpret_cast<byte_vec>(const_cast<void*>(hk.KeyAtRead())), n, true));
```

src/analyzer/protocol/dce-rpc/dce_rpc-analyzer.pac
bbannier: Thanks, fix! That was from me adjusting the cast in generated code and blindly copying it over.

src/analyzer/protocol/netbios/NetbiosSSN.cc
bbannier: The issue is that most of the time we construct a `String` from a `u_char*`, but `u_char` is not even a standardized type. `std::string_view` is `std::basic_string_view<char>` so we cannot stuff a `u_char*` into the usual `string_view`. Trying to specialize a `string_view` over an unsigned `char` on my system gives:

```cpp
using ustring_view = std::basic_string_view<u_char>;
```

> `char_traits<unsigned char> is deprecated: char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 19, so please migrate off of it.`

timwoj (approved):

PR #4616, opened 2025-06-29
Prevent SQLite storage backend from serving expired entries

bbannier:
The SQLite storage backend implements expiration by hand and garbage collection is done in `DoExpire`. This previously relied exclusively on gets not running within `Storage::expire_interval` of the put, otherwise we would potentially serve expired entries.

With this patch we explictly check that entries are not expired before serving them so that the SQLite backend should never serve expired entries.


bbannier:
My understanding of the discussion in https://github.com/zeek/zeek/discussions/4374 was that storage backends should never serve expired entries, but I believe this is currently not the case. This seems like a bug to me since as e.g., a package developer one cannot reliably use storage expiration (the site might set an incompatible `Storage::expire_interval`).

There is still a gap here in that while users now cannot see expired entries, it seems to break put with overwrite of expired entries. This seems to be broken more generally and I filed https://github.com/zeek/zeek/issues/4617.

testing/btest/scripts/base/frameworks/storage/sqlite-expiration-implicit.zeek
```diff
@@ -0,0 +1,48 @@
...
+			$key=key,
+			$value=value,
+			$expire_time=expire));
+
+	# The entry we just put in exists.
+	local get = Storage::Sync::get(h, key);
+	print "BEFORE", get;
+
+	# Advance time.
+	sleep(expire);
```
awelzel: You could avoid the sleep by controlling network time. If you set `allow_network_time_forward` to `F`, network time will stick at 0.0 and so you can use `set_network_time()` to modify it as you wish.
```
redef allow_network_time_forward = F;
...
set_network_time(double_to_time(42.0));
...
set_network_time(double_to_time(43.0));
```

More classic would be to use a PCAP with `new_connection()` and `connection_state_remove()` or `new_packet()` with timestamps far away apart, but I like the explicit `set_network_time()` more.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -490,15 +490,21 @@ OperationResult SQLite::Step(sqlite3_stmt* stmt, bool parse_value) {
...
-            size_t blob_size = sqlite3_column_bytes(stmt, 0);
-
-            auto val = serializer->Unserialize({blob, blob_size}, val_type);
-
-            if ( val )
-                ret = {ReturnCode::SUCCESS, "", val.value()};
-            else
-                ret = {ReturnCode::OPERATION_FAILED, val.error()};
+            auto expire_time = sqlite3_column_double(stmt, 1);
+            auto is_expired = expire_time != 0 && expire_time < run_state::network_time;
```
awelzel:
```suggestion
            auto is_expired = expire_time > 0 && expire_time < run_state::network_time;
```

I could see this, just in case we ever want to give negative expiry times some special meaning. But either works.

awelzel (approved):
Makes sense to me.

timwoj (approved):
Makes sense to me too. We shouldn't never need to include expired entries. I think I had a conversation with @ckreibich when I was working on this whether it would be better to include the expiration time with `get` responses and let the caller deal with them, but I think the way you did it ends up being easier. We wouldn't be returning expired entries with backends that support native expiration.

src/storage/backend/sqlite/SQLite.cc
bbannier: This is a bit of a mess currently since `expire_time` is not declared `NOT NULL` so the way I used it here might mix `expire_time == 0.0` and `expire_time == NULL`. I fixed the patch to only look at values if they are present, and also moved this to a helper function for easier maintenance, but kept the comparison against `0.0` since this is what I also see elsewhere, e.g., in explicit expiry. I do however wonder if all this shouldn't be done in a prepared statement instead.

testing/btest/scripts/base/frameworks/storage/sqlite-expiration-implicit.zeek
bbannier: Nice, I did switch to `allow_network_time_forward = F` and moved the code from `current_time` to `network_time` where it matters. The downside of the "classic approach" of relying only on a PCAP to drive time is that it makes time pretty implicit where this is really all this kind of test is about.

testing/btest/scripts/base/frameworks/storage/sqlite-expiration-implicit.zeek
```diff
@@ -0,0 +1,62 @@
...
+	# new value in its place without specifying overwrite.
+	Storage::Sync::put(
+		h,
+		Storage::PutArgs(
+			$key=key,
+			$value=value+value));
+
+	get = Storage::Sync::get(h, key);
+	# print "OVERWRITE", get; # FIXME(bbannier): This currently behaves incorrectly due to #4617, return dummy data for now.
+	print "OVERWRITE, [code=Storage::SUCCESS, error_str=<uninitialized>, value=vv]";
```
bbannier: This is currently blocked by #4617.

awelzel: Merged.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -285,6 +290,28 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
...
+        else if ( step_status == SQLITE_BUSY || step_status == SQLITE_LOCKED )
+            // TODO: this could retry a number of times instead of just failing
+            return {ReturnCode::TIMEOUT};
+        else
+            return {ReturnCode::OPERATION_FAILED};
+    }
+
     auto key_data = serializer->Serialize(key);
     if ( ! key_data )
         return {ReturnCode::SERIALIZATION_FAILED, "Failed to serialize key"};
```
timwoj: The asan failure is because it doesn't set the column in the prepared statement.

```suggestion
    auto key_data = serializer->Serialize(key);
    if ( ! key_data )
        return {ReturnCode::SERIALIZATION_FAILED, "Failed to serialize key"};

    // If we are not already in overwrite mode check if an expired entry exists
    // in the database. Such entries would not be visible to the user, but even
    // outside of overwrite mode would need to be overwritten.
    if ( ! overwrite ) {
        auto stmt = unique_stmt_ptr(get_stmt.get(), sqlite3_reset);
        if ( auto res = CheckError(sqlite3_bind_blob(stmt.get(), 1, key_data->data(), key_data->size(), SQLITE_STATIC));
             res.code != ReturnCode::SUCCESS ) {
            return res;
        }
        
        int step_status = sqlite3_step(stmt.get());
        if ( step_status == SQLITE_ROW ) {
            // If an expired entry exists, switch to overwrite mode.
            overwrite =
                sqlite3_column_type(stmt.get(), 1) != SQLITE_NULL && is_expired(sqlite3_column_double(stmt.get(), 0));
        }
        else if ( step_status == SQLITE_DONE ) {
            // Nothing currently exists.
        }
        else if ( step_status == SQLITE_BUSY || step_status == SQLITE_LOCKED )
            // TODO: this could retry a number of times instead of just failing
            return {ReturnCode::TIMEOUT};
        else
            return {ReturnCode::OPERATION_FAILED};
    }
```

I also could see making `SQLite::Step` take a lambda for processing the result. It'd let us reuse all of the error handling there instead of duplicating a portion of it here. I can do that in a follow-up though if you don't want to tackle it here.

timwoj (changes requested):

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -290,6 +290,28 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
...
 
+    // If we are not already in overwrite mode check if an expired entry exists
+    // in the database. Such entries would not be visible to the user, but even
+    // outside of overwrite mode would need to be overwritten.
+    if ( ! overwrite ) {
+        auto stmt = unique_stmt_ptr(get_stmt.get(), sqlite3_reset);
+
+        int step_status = sqlite3_step(stmt.get());
+        if ( step_status == SQLITE_ROW ) {
+            // If an expired entry exists, switch to overwrite mode.
```
awelzel: I presume this makes sense, but I'm wondering if instead of an extra query, this could be pushed down to SQLite on CONFLICT handling without the need for an extra query. If you can come up with that query, no need for all the error handling.


Untested, but something like this to get "overwrite expired entries even if ! overwrite" ?
```
INSERT INTO ... ON CONFLICT
DO UPDATE SET value={value} expire_time={now} WHERE key={key} AND expire_time IS NOT NULL AND {now} > expire_time
```
https://sqlite.org/lang_upsert.html


EDIT: Just found your note from the other comment:

> I do however wonder if all this shouldn't be done in a prepared statement instead.

Agree :-)

src/storage/backend/sqlite/SQLite.cc
bbannier: Ah, of course, thanks for taking a look.

src/storage/backend/sqlite/SQLite.cc
bbannier: > EDIT: Just found your note from the other comment:

Yeah, I didn't try to refactor the code which is probably better left to somebody who knows it better. Currently this backend is a _very thin_ wrapper around sqlite3.

While there are helpers like `unique_stmt_ptr` I personally would push this a couple steps further and expose only dedicated classes around prepared statements which if can only be constructed from all required deserialized Zeek values and a database handle; that could help avoid stupid mistakes like [me forgetting to bind a variable when using one of the prepared statements](https://github.com/zeek/zeek/pull/4616#discussion_r2178372359). Such classes could then also handle the full lifecycle of a query instead us having a disconnected `Step` function.

awelzel: >  I personally would push this a couple steps further and expose only dedicated classes around prepared statements which if needed could only be constructed from deserialized Zeek values

Sure, that's one part. But using a single `INSERT` statement over pre-query-the-database-state-in-a-non-transactional-way even using the rawer primitives would still be applicable.

Don't let this block this PR: It's just that I once had a co-worker who'd have much more than nit-picked on doing extra SQL queries when not needed and sometimes that shows :-)

bbannier: You are right, using two queries here is actually not only less performant, but also could expose us to data races (unless this already starts a transaction/pulls a lock, didn't check).

timwoj: > Such classes could then also handle the full lifecycle of a query instead us having a disconnected Step function.

I think my only concern here is that a lot of working with the Sqlite API is repeated patterns of code, like what happens in `Step`. Now you've got this inheritance tree of classes so that they all have access to Step, or you have Step in some set of utility methods, and you're back to where we are. It feels like adding a bunch of complication for not much gain.

timwoj: > DO UPDATE SET value={value} expire_time={now} WHERE key={key} AND expire_time IS NOT NULL AND {now} > expire_time

I stared at that for a while yesterday but it was entirely too late in the day and my brain was struggling. I tested it a little bit here with the sqlite3 command line and the only thing I'm running into is what happens if you request an overwrite without caring about the time. There's two cases here:

1. The key conflicts, an expiration time exists in the table, but the time hasn't been reached yet. it won't overwrite because it doesn't meet the `now > expire_time` condition. Maybe that's fine and we don't allow resetting the expiration time, but that feels bad to me.
2. The key conflicts and there isn't an expiration time in the table for the entry. It won't expire because the first part of the `where` condition returns false.

At the risk of making this more complicated, maybe something like:

```
DO UPDATE SET ... WHERE (expire_time == 0.0) or (expire_time > 0.0 and (expire_time < now or new_expire_time > expire_time))
```

Am I stretching a bit too much? This is going to be fun to write a comprehensive btest for.

timwoj: Also, I'm completely ok with marking expire_time as NOT NULL. We always set a value in it anyways even if we set it to zero.

awelzel: > I'm running into is what happens if you request an overwrite without caring about the time. 

If one requests `overwrite=T` it seems we should always update the entry regardless of what's stored, that's why they said `overwrite=T`, right?

In other words, I don't think there's a "without caring about time" part in the API, lets assume they do care, just that they care to set it to infinite. Because they used `overwrite=T`, they should expect that we always do the update.

> Am I stretching a bit too much? 

For `overwrite=T` I think yes. Lets just overwrite it.

bbannier: > Also, I'm completely ok with marking expire_time as NOT NULL. We always set a value in it anyways even if we set it to zero.

If possible I wouldn't do this, if only so we do not have to overload the meaning of `expire_time=0`.

timwoj: Ah, I was trying to replace both `put` statements with a single one. If this is just replacing `put_stmt`, then what you have works perfectly.

bbannier: So you are saying that the current process has a lock, and nobody else can write to the DB between us checking whether there is an expired entry with a get, and potentially overwriting what we saw with put?

timwoj: I mean the version that Arne posted. That does what you want in a single statement.

bbannier: @timwoj would you mind helping me out and weave that in?

timwoj: Sure. I can pick this up.

timwoj: One problem I'm running into with this is the without the additional `ON CONFLICT` a conflicting `INSERT` statement would cause the SQLite API to return `SQLITE_CONSTRAINT`. We were using that response to detect when a key exists and return `ReturnCode::KEY_EXISTS`. With the `ON CONFLICT` bits, it returns `SQLITE_DONE` like any normal `INSERT` would even when the `WHERE` block prevents a change from happening in the table. That leaves us without a good way to detect conflicting keys.

I'm not sure if this is intentional in SQLite's API. I'll throw together a small reproducer and send a message to the SQLite project about it.

timwoj: > I'm not sure if this is intentional in SQLite's API. I'll throw together a small reproducer and send a message to the SQLite project about it.

See https://sqlite.org/forum/forumpost/2193e7374e. The response from the SQLite people is that the return value is intentional, and using a `WHERE` with `DO UPDATE` is intended to turn the update into a no-op if the `WHERE` doesn't match. I'm not sure what to do here. Doing `put` operations underneath a transaction would impose a performance penaltyI. There is the chance you'll run `put` and `expire` interleaved since `expire` runs on a separate thread. I'm not sure that matters though, as if you start a `put`, do the lookup and find expired entries, then the `expire` executes and removes them, there wouldn't be a conflict when the actual insertion happens. `expire` happens entirely under an exclusive transaction.

awelzel: Would using [sqlite3_changes()](https://www.sqlite.org/c3ref/changes.html) allow you to get the number of changed / inserted rows for an `ON CONFLICT` execution. The docs mention triggers and foreign key constraints won't be included, but would expect `ON CONFLICT` to work?

Using two different statements, for overwrite=T and overwrite=F, the latter could check if any rows were changed (updated or inserted) and if it was none, raise KEY_EXISTS. overwrite=T cannot return KEY_EXISTS by definition, that statement just INSERTS and ON CONFLICT updates the value and expiration time.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -314,20 +320,31 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
...
+        return res;
+    }
+
+    // This duplicates the above binding, but it's to overwrite the expiration time on the entry.
+    if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 5, expiration_time)); res.code != ReturnCode::SUCCESS ) {
+        return res;
+    }
+
+    if ( ! overwrite )
+        if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 6, util::current_time()));
```
awelzel:
```suggestion
        if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 6, zeek::run_state::network_time));
```

Then the known-services test passes :-)

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -497,15 +514,19 @@ OperationResult SQLite::Step(sqlite3_stmt* stmt, bool parse_value) {
...
-            auto blob = static_cast<const std::byte*>(sqlite3_column_blob(stmt, 0));
-            size_t blob_size = sqlite3_column_bytes(stmt, 0);
-
-            auto val = serializer->Unserialize({blob, blob_size}, val_type);
-
-            if ( val )
-                ret = {ReturnCode::SUCCESS, "", val.value()};
-            else
-                ret = {ReturnCode::OPERATION_FAILED, val.error()};
+            if ( sqlite3_column_type(stmt, 1) != SQLITE_NULL && is_expired(sqlite3_column_double(stmt, 1)) )
```
awelzel: This can be done via the get_stmt directly, too:

```
diff --git a/src/storage/backend/sqlite/SQLite.cc b/src/storage/backend/sqlite/SQLite.cc
index b97ff07331..4deb565fd6 100644
--- a/src/storage/backend/sqlite/SQLite.cc
+++ b/src/storage/backend/sqlite/SQLite.cc
@@ -17,11 +17,6 @@
 
 using namespace std::chrono_literals;
 
-namespace {
-// Helper to check whether a database-stored `expire_time` is considered expired.
-bool is_expired(double expire_time) { return expire_time != 0 && expire_time < zeek::run_state::network_time; }
-} // namespace
-
 namespace zeek::storage::backend::sqlite {
 
 OperationResult SQLite::RunPragma(std::string_view name, std::optional<std::string_view> value) {
@@ -196,7 +191,10 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
     static std::array<std::pair<std::string, sqlite3*>, 8> statements =
         {std::make_pair(util::fmt(put_base_cmd.c_str(), table_name.c_str()), db),
          std::make_pair(util::fmt(put_cmd.c_str(), table_name.c_str()), db),
-         std::make_pair(util::fmt("select value_str, expire_time from %s where key_str=?", table_name.c_str()), db),
+         std::make_pair(util::fmt("select value_str, expire_time from %s where key_str=? and ((expire_time > ?) OR "
+                                  "(expire_time IS NOT NULL AND expire_time == 0.0))",
+                                  table_name.c_str()),
+                        db),
          std::make_pair(util::fmt("delete from %s where key_str=?", table_name.c_str()), db),
 
          std::make_pair(
@@ -297,10 +295,10 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
         return {ReturnCode::SERIALIZATION_FAILED, "Failed to serialize key"};
 
     unique_stmt_ptr stmt;
-    if ( ! overwrite )
-        stmt = unique_stmt_ptr(put_stmt.get(), sqlite3_reset);
-    else
+    if ( overwrite )
         stmt = unique_stmt_ptr(put_update_stmt.get(), sqlite3_reset);
+    else
+        stmt = unique_stmt_ptr(put_stmt.get(), sqlite3_reset);
 
     if ( auto res = CheckError(sqlite3_bind_blob(stmt.get(), 1, key_data->data(), key_data->size(), SQLITE_STATIC));
          res.code != ReturnCode::SUCCESS ) {
@@ -331,7 +329,7 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
     }
 
     if ( ! overwrite )
-        if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 6, util::current_time()));
+        if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 6, zeek::run_state::network_time));
              res.code != ReturnCode::SUCCESS ) {
             return res;
         }
@@ -365,6 +363,11 @@ OperationResult SQLite::DoGet(ResultCallback* cb, ValPtr key) {
         return res;
     }
 
+    if ( auto res = CheckError(sqlite3_bind_double(stmt.get(), 2, zeek::run_state::network_time));
+         res.code != ReturnCode::SUCCESS ) {
+        return res;
+    }
+
     return Step(stmt.get(), true);
 }
 
@@ -514,19 +517,15 @@ OperationResult SQLite::Step(sqlite3_stmt* stmt, bool parse_value) {
     int step_status = sqlite3_step(stmt);
     if ( step_status == SQLITE_ROW ) {
         if ( parse_value ) {
-            if ( sqlite3_column_type(stmt, 1) != SQLITE_NULL && is_expired(sqlite3_column_double(stmt, 1)) )
-                ret = {ReturnCode::KEY_NOT_FOUND, ""};
-            else {
-                auto blob = static_cast<const std::byte*>(sqlite3_column_blob(stmt, 0));
-                size_t blob_size = sqlite3_column_bytes(stmt, 0);
-
-                auto val = serializer->Unserialize({blob, blob_size}, val_type);
-
-                if ( val )
-                    ret = {ReturnCode::SUCCESS, "", val.value()};
-                else
-                    ret = {ReturnCode::OPERATION_FAILED, val.error()};
-            }
+            auto blob = static_cast<const std::byte*>(sqlite3_column_blob(stmt, 0));
+            size_t blob_size = sqlite3_column_bytes(stmt, 0);
+
+            auto val = serializer->Unserialize({blob, blob_size}, val_type);
+
+            if ( val )
+                ret = {ReturnCode::SUCCESS, "", val.value()};
+            else
+                ret = {ReturnCode::OPERATION_FAILED, val.error()};
         }
         else {
             ret = {ReturnCode::OPERATION_FAILED, "sqlite3_step should not have returned a value"};

```

My main concern is about `(expire_time IS NOT NULL AND expire_time == 0.0)`, but I think as long as we store `0.0` for indefinite expiration that should be safe. I'm slowly tending towards that if we use SQL, we should offload as much as possible to SQL rather than doing things partly in SQL, but then also have some of the logic in code. I.e. reading the SQL statements alone should give an impression how the implementation behaves (and I wouldn't really mind just open-coding them completely, rather than trying to re-use part of the strings (which obfuscates this a bit).

This hunk contains the current_time -> network_time change, and flipping the `if ( ! overwrite )` to make it a bit easier to read.

EDIT: Here's a commit:
https://github.com/zeek/zeek/commit/c5397917e17eb58c29652d90ccf2186db701ed98

timwoj: > I think as long as we store 0.0 for indefinite expiration that should be safe. 

0.0 is the default if you don't pass an expiration value explicitly. The documentation could be updated for the `PutArgs::expire_time` field to indicate that passing zero will mean it doesn't expire. That's where I was going when I mentioned making that field `NOT NULL` because there shouldn't ever be case where we don't set the value. It would remove the need to check for `NULL` and we can just check the value.

timwoj: > I'm slowly tending towards that if we use SQL, we should offload as much as possible to SQL rather than doing things partly in SQL, but then also have some of the logic in code.

I agree with that sentiment.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -181,28 +181,39 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
 
     sqlite3_free(errorMsg);
 
+    // Roughly the same command is used for put and put-overwrite, just with an extra condition on put
+    // to only overwrite if the element is expired.
+    std::string put_base_cmd =
```
awelzel: Nit.

```suggestion
    std::string put_override_cmd =
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -181,28 +181,39 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
+        "insert into %s (key_str, value_str, expire_time) values(?, ?, ?) ON CONFLICT(key_str) "
+        "DO UPDATE SET value_str=?, expire_time=?";
+    std::string put_cmd = put_base_cmd + " WHERE expire_time > 0.0 AND expire_time < ?";
+
     static std::array<std::pair<std::string, sqlite3*>, 8> statements =
-        {std::make_pair(util::fmt("insert into %s (key_str, value_str, expire_time) values(?, ?, ?)",
+        {// Put with forced overwrite
+         std::make_pair(util::fmt(put_base_cmd.c_str(), table_name.c_str()), db),
```
awelzel: Hmm, never mind the variable name, maybe just open-code the stmts here now just like all the others?

PR #4618, opened 2025-06-30
cluster/WebSocket: Include X-Application-Name in cluster.log

awelzel:
A bit ad-hoc formatting for the log, but that's mostly because cluster.log only has message field and I don't think having a dedicated application_name column is worth it. That could also be added by custom scripts if it's really wanted for a given deployment.


src/cluster/websocket/WebSocket.cc
```diff
@@ -375,8 +375,9 @@ void WebSocketEventDispatcher::Process(const WebSocketClose& close) {
         // Raise Cluster::websocket_client_lost() after the backend has terminated.
         // In case any messages/events were still pending, Cluster::websocket_client_lost()
         // should be the last event related to this WebSocket client.
-        auto rec = zeek::cluster::detail::bif::make_endpoint_info(backend->NodeId(), wsc->getRemoteIp(),
-                                                                  wsc->getRemotePort(), TRANSPORT_TCP);
+        auto rec =
+            zeek::cluster::detail::bif::make_endpoint_info(backend->NodeId(), wsc->getRemoteIp(), wsc->getRemotePort(),
+                                                           TRANSPORT_TCP, it->second.application_name);
```
bbannier, line 380: Nit: We could move here since the data behind `it` is going away (actually similar for `wsc` and `backend`),

```suggestion
                                                           TRANSPORT_TCP, std::move(it->second.application_name));
```

PR #4622, opened 2025-07-01
Bump pre-commit hooks

bbannier:


scripts/base/frameworks/openflow/main.zeek
```diff
@@ -253,7 +253,7 @@ function controller_init_done(controller: Controller)
 
 # Functions that are called from cluster.zeek and non-cluster.zeek
 
-function register_controller_impl(tpe: OpenFlow::Plugin, name: string, controller: Controller)
+function register_controller_impl(type_: OpenFlow::Plugin, name: string, controller: Controller)
```
awelzel, line 256: @bbannier - I think this broke the docs build. At least locally I'm seeing this when building the docs.

```
doc/scripts/base/frameworks/openflow/main.zeek.rst:253: ERROR: Unknown target name: "type".                                                                   
doc/scripts/base/frameworks/intel/input.zeek.rst:68: ERROR: Unknown target name: "type".                                                                      
looking for now-outdated files... none found        
```

Suspect that trailing `_` has special meaning for sphinx/rst and we just pass it through verbatim during zeekygen. Okay to revert the PR or would you rather do a follow-up?

I had wanted to suggest adding `tpe` to `.typos.toml`, but figured it's likely going to be fine :sob:

bbannier: Ups, yes, please revert.

> I had wanted to suggest adding tpe to .typos.toml, but figured it's likely going to be fine 😭

I was hoping so too. I am starting to reach an age where when I see identifiers like `tpe` I briefly consider whether I had a stroke. 🤣

PR #4623, opened 2025-07-02
Use target-based approach for enabling sanitizers

bbannier:
The current approach for building with sanitizers is to build all targets in the tree with them, regardless of whether they are from Zeek or from something we vendor. While this seems interesting it also means that everything incurrs potential ASAN overhead, regardless of whether it is actually needed here or not, e.g., in a container on my machine `spicyz` invocations are very slow for some configurations (>30min runtimes for some builtin analyzers), but this seems largely unnecessary since Spicy already does its own sanitizer builds.

This patch changes that approach to instead be opt-in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,6 +239,136 @@ if (ENABLE_CLANG_TIDY)
     endif ()
 endif ()
 
+if (ZEEK_SANITIZERS)
+    string(REPLACE "," " " _sanitizer_args "${ZEEK_SANITIZERS}")
+    separate_arguments(_sanitizer_args)
+    set(ZEEK_SANITIZERS "")
+
+    foreach (_sanitizer ${_sanitizer_args})
+        if (ZEEK_SANITIZERS)
+            set(ZEEK_SANITIZERS "${ZEEK_SANITIZERS},")
+        endif ()
+
+        if (_sanitizer STREQUAL "thread")
+            set(ZEEK_TSAN true)
+        endif ()
+
+        if (NOT _sanitizer STREQUAL "undefined")
+            set(ZEEK_SANITIZERS "${ZEEK_SANITIZERS}${_sanitizer}")
+            continue()
+        endif ()
+
+        if (NOT DEFINED ZEEK_SANITIZER_UB_CHECKS)
+            if (DEFINED ENV{ZEEK_TAILORED_UB_CHECKS})
+                # list(APPEND _check_list "alignment") # TODO: fix associated errors
+                list(APPEND _check_list "bool")
+                # list(APPEND _check_list "builtin") # Not implemented in older GCCs
+                list(APPEND _check_list "bounds") # Covers both array/local bounds
+                                                  # options below
+                # list(APPEND _check_list "array-bounds") # Not implemented by GCC
+                # list(APPEND _check_list "local-bounds") # Not normally part of
+                # "undefined"
+                list(APPEND _check_list "enum")
+                list(APPEND _check_list "float-cast-overflow")
+                list(APPEND _check_list "float-divide-by-zero")
+                # list(APPEND _check_list "function") # Not implemented by GCC
+                # list(APPEND _check_list "implicit-unsigned-integer-truncation") # Not
+                # truly UB list(APPEND _check_list "implicit-signed-integer-truncation")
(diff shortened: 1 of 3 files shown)


timwoj:
I'd be alright with this. Anything we can do to speed up the sanitizer tests on CI would be a huge benefit.

CMakeLists.txt
```diff
@@ -226,6 +226,136 @@ if (ENABLE_CLANG_TIDY)
...
+# Usage:
+#
+#    zeek_target_enable_sanitizers(<target>)
+#
+function (zeek_target_enable_sanitizers target_name)
+    string(REPLACE " " ";" _flags "${_sanitizer_flags}")
+
+    foreach (flag ${_flags})
+        target_compile_options(${target_name} PRIVATE ${flag})
+        target_link_options(${target_name} PRIVATE ${flagS})
```
timwoj:
```suggestion
        target_link_options(${target_name} PRIVATE ${flag})
```

That fixes the linking failure on CI.

timwoj:
I think cmake/ZeekPluginStatic.cmake needs to be updated so that all of the internal plugins get built with the sanitizers enabled. For example, I `rm`'d all of the object files from src/analyzer/conn-size and rebuilt with `ninja -v`. It didn't have `-fsanitize` except on the final linking of the binary.

CMakeLists.txt
bbannier: 🤦

bbannier:
> I think cmake/ZeekPluginStatic.cmake needs to be updated so that all of the internal plugins get built with the sanitizers enabled. For example, I `rm`'d all of the object files from src/analyzer/conn-size and rebuilt with `ninja -v`. It didn't `-fsanitize` except on the final linking of the binary.

I am not immediately sure how to do that since 1) I am not sure where that file gets included and 2) don't see an immediate way to get the Zeek CMake config into that module.

timwoj:
We have `zeek_target_add_linters` which does basically the same thing you're doing here. The function is defined in the top-level `CMakeLists.txt` but used down in `ZeekPluginStatic.cmake`. We never include that file except during Zeek builds so the function should always exist.

cmake
```diff
@@ -1 +1 @@
-Subproject commit 53a2d51b256399813c18e07b47eeb7a73bd03eb7
+Subproject commit 09a3a5d5a42a0419ea55da3d8cfccd6fc769ad28
```
bbannier: This pulls in https://github.com/zeek/cmake/pull/140.

timwoj:
Copying some notes from a conversation I had with @bbannier a few weeks ago about the CI error:

> Tim Wojtulewicz 10:17 AM
> That feels like a "make a post on the Ilvm discord" kind of error
>
> Benjamin Bannier 10:18 AM
> No, I think this is just Spicy, e.g., us needing to add sanitizer flags to the runtime HLTO flags, before we'd just copy the exxflags of the binaries on (edited)
>
> Tim Wojtulewicz 10:19 AM
> I definitely like the idea though. You guys already run sanitizers/leak checks on spicy, so there's not really any reason for us to do it also.
>
> Benjamin Bannier 10:19 AM
> we'd still want to run it for analyzers, e.g., QUIC has custom C++
>
> Tim Wojtulewicz 10:20 AM
> Right, this is just for spicyz itself
> The generated code should still be built with the sanitizers enabled
>
> Benjamin Bannier 10:20 AM
> Which might be the part we are missing
> 
> Tim Wojtulewicz 10:21 AM
> Is "needing to add sanitizer flags to the runtime HLTO flags" something that can be done easily just to try the theory?
> 
> Benjamin Bannier 10:25 AM
> Yeah, definitely. We set the flags [here](https://github.com/zeek/spicy/blob/8e768adeffba2a658cf32a2044ca7868066338c1/hilti/toolchain/src/config.cc.in#L124-L206). If you look at the code you see some references to CMake variables like `HILTI_CONFIG_RUNTIME_CXX_FLAGS_RELEASE` and so on. For sanitizer builds we should amend sanitizer flags, e.g., after the `add_subdirectory(auxil/spicy)` call.

I have not had time to get back to test this theory.

timwoj:
I really like the idea of this but we clearly don't have the bandwidth right now to work on it. @bbannier do you care if I close this?

Issue #4602, opened 2025-06-24
PPPoE analyzer ignores length field


0xxon:
The PPPoE analyzer currently skips over the entire header, including over the length field.

This means that Zeek currently may parse beyond the end of the data that is incicated in the length field of a PPPoE packet.

While this should not happen in practice, we have an example of this in packet 33 of `testing/btest/Traces/pppoe-over-qinq.pcap`.

<img width="717" alt="Image" src="https://github.com/user-attachments/assets/63876969-545a-43c1-8cff-3ab4414519b7" />

The TCP packet inside this PPPoE session contains six zero bytes and is responsible for the discovery of #4547. However, the lengt of the PPPoE packet is shorter than the length of the TCP session - to be exact it's exactly 6 bytes shorter - which is exactly the last 6 bytes which caused the parsing errors in #4547.

I am not completely sure what we should do in cases like these. However, it feels like we should potentially respect the length value in the PPPoE header, and ignore data beyond its length field. We also should probably raise a weird if the length of the payload data is different from the length indicated in the PPPoE header.

stevesmoot:
Just wanted to comment if anyone is touching the PPPoE code would be nice to change it to pull the SESSION_ID out, as its a potential candidate for ConnKey processing (I think).  I ended up writing a trivial analyzer in spicy so I could get it.

PR #4626, opened 2025-07-03
PPPoE: respect payload size, add pppoe session ID to connection record

0xxon:
This PR contains two independent commits.

The first adds a check that the payload size of the PPPoE header matches what we forward to the next protocol analyzer. If there is more data than the header indicates, we snip it off and forward only the amount indicated in the header. This is in line with what other analyzers (e.g. UDP) do. Weirds are added.

This changes some test baselines as we have a trace exhibiting this behaviour.

The second commit adds support for parsing the pppoe session id and adding it to the connection log. This was flagged as potentially interesting by @stevesmoot in GH-4602.

It currently is not selectable as a connection key - I would be happy to put that in too if this is desired.

The second commit has a larger baseline change as the connection record is altered. The default connection log is not altered.


awelzel:
> This PR contains two independent commits.

Could we do them as independent PRs? The first one is non-controversial (and backport material). For the second, I wonder if there's more design conversations to be had. Extending a number of data structures / records with pppoe specific fields just doesn't feel quite right at this point. I know they are there for vlan, but we're moving to a different model where the pppoe session id could be part of a separate ConnKey implementation if a user opts into that. Another wild thought is whether pppoe should be treated as some sort of tunnel - given PPPoE isn't IP based, that's probably hairy though.

0xxon:
Sure, let me separate them out.

PR #4627, opened 2025-07-03
PPPoE: don't forward more bytes than header indicates

0xxon:
This changes the PPPoE parser so that it doesn't forward extra bytes that might be appended after the payload. Instead, it raises a weird if the payload size doesn't match the size indicated by the header.

This is in line with what other protocol parsers (like UDP) are doing.

Two tests needed to be updated - with this change, the traffic in pppoe-over-qinq.pcap is now valid TLS. A new trace was introduced for the confirmation-violation-info test.

Addresses GH-4602


0xxon:
As this is neither a regression, nor a security issue I don't think that it should be backported to 7.0.

src/packet_analysis/protocol/pppoe/PPPoE.cc
```diff
@@ -12,8 +12,20 @@ bool PPPoEAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packe
         return false;
     }
 
+    size_t payload_length = (data[4] << 8u) + data[5];
     // Extract protocol identifier
```
awelzel:
```suggestion
```

awelzel: :man_shrugging:  - either that, or adding a comment to the payload_length extraction, but that seems to document just what's done.

awelzel (commented):
LGTM.

Could you add a note to Traces/README where `tls-1.2-protocol-error.pcap` is coming from?

As of a 7.0 backport. I put it in the bug category worth backporting, but if we don't, that's totally fine.

0xxon:
> Could you add a note to Traces/README where `tls-1.2-protocol-error.pcap` is coming from?

I'm in principle happy to. However, the trace is just hex-edited from a trace already in the testsuite; that trace was also self-generated. We mostly seem to indicate traces that have an outside source in that file, not traces we create ourselves from scratch.

awelzel (approved):

PR #4628, opened 2025-07-03
Add generic connection packet threshold event

J-Gras:
Zeek provides the ConnThreshold framework for monitoring connections. To set up threshold monitoring, one needs to call `set_<type>_threshold` on the connection to monitor. If one wants to implement a threshold on all kinds of connections (TCP, UDP, ICMP, unknown transport), currently `new_connection` is the only available event. However, handling `new_connection` to setup threshold monitoring unnecessarily introduces significant overhead for short connections that would never reach the threshold (e.g., SYN flooding, scans, or DNS).
The idea here is to introduce a new event with simple semantics: Trigger on any connection that saw a certain number of packets. The packet threshold can be configured and does not take direction into account (i.e. also triggers on one-sided connections). Even a low threshold of 5 packets should already exclude the short connections mentioned above.

Fwiw, I also tested a plugin that implements this new event by attaching a ConnSize-like analyzer to every connection. However, benchmarking showed that adding a new analyzer to every connection already comes with non-neglegible performance impact. Given that having this event seems handy in general, I think we should just add it to the ConnSize analyzer.

Regarding naming, `conn_generic_packet_threshold_crossed` seems precise, aligns with the existing events, but is quite verbose. I considered `connection_active` or `connection_busy`, with active/busy being defined in terms of the packet threshold. I also thought about alternatives to "connection" like session, because this doesn't deal with connections in a strict sense. However, given the established naming in Zeek, I guess going with the above is least surprising. Thoughts?


scripts/base/init-bare.zeek
```diff
@@ -6369,6 +6369,18 @@ export {
...
+module ConnThreshold;
+
+export {
+	## Number of packets required to be observed on any IP-based session to
+	## trigger :zeek:id:`conn_generic_packet_threshold_crossed`. Note that the
+	## threshold refers to the total number of packets transferred in both
+	## directions.
+	##
+	## .. zeek:see:: conn_generic_packet_threshold_crossed
+	const generic_packet_threshold = 5 &redef;
```
awelzel: As in DM, if it's not adding too much state tracking in `ConnSize`, extending this to support multiple thresholds would make me go "yeah, lets just include this". Totally get the motivation to optimize away `new_connection()` handlers to setup threshold events for connections that'll then never even reach them.

```suggestion
	const fixed_packet_thresholds: vector of count = vector() &redef;
```

src/analyzer/protocol/conn-size/events.bif
```diff
@@ -46,3 +46,10 @@ event conn_packets_threshold_crossed%(c: connection, threshold: count, is_orig:
 ##               get_current_conn_bytes_threshold get_current_conn_packets_threshold
 ##               set_current_conn_duration_threshold get_current_conn_duration_threshold
 event conn_duration_threshold_crossed%(c: connection, threshold: interval, is_orig: bool%);
+
+## Generated for any IP-based session once :zeek:id:`ConnThreshold::generic_packet_threshold` packets have been
+## observed. Only one endpoint sending traffic is sufficient to trigger the event. This allows to handle new
+## connections, while short interactions, like scans consisting of only a few packets, are ignored.
+##
+## c: The connection.
+event conn_generic_packet_threshold_crossed%(c: connection%);
```
awelzel:
```suggestion
event conn_fixed_packets_threshold_crossed%(c: connection, threshold: count%);
```

I like "fixed". Maybe "static" as an alternative? "generic" is too generic :-)

scripts/base/init-bare.zeek
J-Gras: I've just pushed a commit that implements support for multiple thresholds (using a `set`, which felt more natural to me). Maybe, we can take a look at the magic Zeek benchmarks to see if there is any significant impact.

src/analyzer/protocol/conn-size/events.bif
J-Gras: I think the key aspect here is that the thresholds will apply to anything that ends up in conn.log. To me that's more "generic" than "fixed" ;) But I'll leave the naming to others to decide...

awelzel: And also generic because it's independent of orig/resp. I think I'm with you. I like it.

scripts/base/init-bare.zeek
awelzel: This looks good to me. I did some wrangling locally to move the lookup of the table to `InitPostScript()` rather than the static caching in the function.

https://github.com/zeek/zeek/commit/c5da9c351963f0cc1c4196f5711405147b15a886

Would you pick this up? It brings us a bit closer to a bif with which one could update dynamically if really wanted. Possibly squashing your two commits with this and and re-push. I don't mind the authorship, but if you'd rather keep it separate that's of course okay.

Happy to merge then :-)

J-Gras: I've squashed your commit in and added an entry to NEWS as requested via DM 🙂

awelzel (approved):
Thanks Jan, cool stuff!

PR #4630, opened 2025-07-03
Modify known-services to add storage framework support

timwoj:
This adds a flag to enable using the storage framework instead of Broker stores, plus a btest for both broker and the the storage framework.


scripts/policy/protocols/conn/known-services.zeek
```diff
@@ -39,6 +43,10 @@ export {
 	## operation.
 	const use_service_store = F &redef;
 
+	## Switches to the version of this script that uses the storage
+	## framework instead of Broker stores.
+	const use_storage_framework = F &redef;
```
0xxon, line 49: I guess we plan to switching this to default true in the future?

If yes - should we perhaps add an comment to just do this, e.g. in 8.1?

0xxon (commented):
This looks good to me, I just have one additional question.

I assume that the output when using the storage framework should be exactly the same as when using the broker store. Would it make sense to run the tests twice, with use_service_store = F, to check that the output is the same in both cases?

scripts/policy/protocols/conn/known-services.zeek
timwoj: Added a note.

timwoj:
> I assume that the output when using the storage framework should be exactly the same as when using the broker store. Would it make sense to run the tests twice, with use_service_store = F, to check that the output is the same in both cases?

The fixup commit merges the two btests I added into the main known-services btest, and adds diffing the output of the two of them as part of the test.

scripts/policy/protocols/conn/known-services.zeek
```diff
@@ -39,6 +43,11 @@ export {
 	## operation.
 	const use_service_store = F &redef;
```
awelzel, line 44: Should we've deprecated this one with a comment that users should switch over to and test  `use_storage_framework` , would be an incentive to remove the code at 8.1.

awelzel (commented):
Late drive-by about deprecating the `use_service_store`.

scripts/policy/protocols/conn/known-services.zeek
timwoj: Both flags are necessary. If `use_service_store` is false, the services are stored in a table in memory instead of being backed up to a store. `use_storage_framework` switches between the two storage modes if `use_service_store` is true.

awelzel: Hmm, maybe just make `use_service_store` mean storage framework then and drop the Broker code as a follow-up? I could see that.

timwoj: Yeah, that's the ultimate plan.

PR #4631, opened 2025-07-04
Fix line number misordering in Location objects, and remove column information

vpax:
For some types of scripting objects, the associated line numbers can be reordered; for example, they can be reported as "lines 49-48" rather than "lines 48-49". This PR fixes that problem and, in a separate commit, removes tracking of column information (which never had actual support, just vestigial hooks).

To fix this robustly, the PR makes the `Location` variables `private` so they can only be manipulated through a set of new methods. Those methods ensure that any changes made to line numbers associated with a `Location` object maintain monotonicity.

Because of introducing the new methods, the PR touches quite a few files, but only in minor ways.

### DIFF
--- a/src/Obj.h
+++ b/src/Obj.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <climits>
+#include <string>
 
 namespace zeek {
 
@@ -13,7 +14,9 @@ namespace detail {
 class Location final {
 public:
     constexpr Location(const char* fname, int line_f, int line_l, int col_f, int col_l) noexcept
-        : filename(fname), first_line(line_f), last_line(line_l), first_column(col_f), last_column(col_l) {}
+        : filename(fname), first_column(col_f), last_column(col_l) {
+        SetLines(line_f, line_l);
+    }
 
     Location() = default;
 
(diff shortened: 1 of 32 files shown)


src/Obj.h
```diff
@@ -22,9 +25,48 @@ class Location final {
     bool operator==(const Location& l) const;
     bool operator!=(const Location& l) const { return ! (*this == l); }
 
+    auto FileName() const { return filename; }
```
awelzel: Let's be explicit. An IDE will show what it is, but looking just in the GitHub PR, I wouldn't know what it is and it actually ends up being a `const char*`. Let's not hide that behind `auto`.

The `const char*` is okay I think - that's for sharing the same filename string from the parser.

src/Obj.h
```diff
@@ -22,9 +25,48 @@ class Location final {
...
+    auto FirstLine() const { return first_line; }
+    auto LastLine() const { return last_line; }
+
+    std::string FileAndLine() const { return std::string(filename) + ":" + std::to_string(first_line); }
+    std::string FileAndLastLine() const { return std::string(filename) + ":" + std::to_string(last_line); }
+    std::string FileAndLines() const {
```
awelzel: Prefer to have these not in the Location class's API. They are helpers for formatting, maybe just inline helpers further down?

```
std::string loc_file_and_line(Location& l) { ... }
```

Though `FileAndLastline()` is only used in the non-functional-never-used debug logger, so not excited having a helper just for that code even in `zeek::detail` (maybe a helper in zeek/Debug.h though?). `FileAndLines()` is used once in Trigger.cc - would just open-code it there. Last, `FileAndLine()` is mostly used in format-strings and seems that could be replaced with `"%s:%d, loc->FileName().c_str(), `loc->FirstLine()`. Seems fine to me even if a bit longer. So yeah, maybe none of those ;-)

testing/btest/language/line-numbers.zeek
```diff
@@ -0,0 +1,12 @@
+# @TEST-DOC: Test for regression in printing line numbers out-of-order
+#
+# Older versions of Zeek would sometimes print statement line numbers
+# reversed (like "lines 49-48"). The issue most noticeably occurs in
+# script profiling, so this test uses script profiling to look for any
+# instances.
+# @TEST-EXEC: ZEEK_PROFILER_FILE=test.prof zeek -b /dev/null
+# @TEST-EXEC: grep 'lines [0-9]*-[0-9]' test.prof >multi-lines
```
awelzel: Should that be more like `grep 'lines [0-9]+-[0-9]+' test.prof >multi-lines` ?

I find the test pretty opaque and the empty output is probably easy to produce when the regex isn't matching anything or the awk script doesn't do the right thing ;-)

I wonder if an `InternalError()` or `InternalWarning()` in `Location::Describe()` is sufficient when an inconsistency is found. What'd be your reaction to that?

vpax:
Thanks, @awelzel . Are you inclined to pick this one up and polish it? Your suggestions make sense to me (including just having an internal error on inconsistency, and removing the fairly opaque BTest).

awelzel:
> Thanks, @awelzel . Are you inclined to pick this one up and polish it? Your suggestions make sense to me (including just having an internal error on inconsistency, and removing the fairly opaque BTest).

Work for me. I'll pick it up. Ignore my message on Slack ;-)

src/Obj.h
awelzel: I just open-coded and removed the helpers. Seemed sensible.

testing/btest/language/line-numbers.zeek
awelzel: Done.

timwoj (approved):
I admit that I liked the look without the accessor methods more but I get that they're all inlined now and it doesn't matter. LGTM.

vpax:
> I admit that I liked the look without the accessor methods more

I guess another approach would be to make them `const` but `public` members and then manipulate them via cast-away-constness so that only the class can (reasonably) alter their values, and thus can maintain the invariant that `first_line` is always less than `last_line`.

PR #4632, opened 2025-07-04
Bump `auxil/spicy/` to latest development snapshot

bbannier:


auxil/spicy
bbannier, line 1: This pulls in https://github.com/zeek/spicy/pull/2105 which is not yet merged.

cmake
bbannier, line 1: This pulls in https://github.com/zeek/cmake/pull/139.

src/spicy/spicyz/main.cc
```diff
@@ -250,6 +273,21 @@ int main(int argc, char** argv) try {
...
 
+    // Spicy uses two special strings to generate identifiers, some of which
+    // are weak symbols which need to be unique across the whole program, in
+    // particular the HLTO scope. Spicy parsers compiled into HLTO files
+    // register themself with the Spicy runtime, but we also inject code so
+    // they register themself as analyzers, both of which happens in the
+    // context of a HLTO scope. In order to support compiling the same parsers
+    // into different analyzer HLTOs we need to make the weak HLTO scopes
+    // unique symbols, so we generate unique strings based on inputs here to
+    // support that use case.
```
bbannier: That this needs such a long comment is a strong smell that this doesn't do things exactly right. The alternative would be to push this into e.g., Spicy's `Driver::parseOptions`, but AFAICT Spicy doesn't really need this.

src/spicy/CMakeLists.txt
```diff
@@ -16,14 +16,30 @@ zeek_add_subdir_library(
     BIFS
     spicy.bif)
 
+# Since HILTI does not support preprocessor macro we do the version replacement in CMake.
+#
+# TODO(bbannier): Use a static file again once we drop support for Spicy versions before 1.14.0.
```
rsmmr: Do we need the backwards compatibility at all these days? I thought since we moved the Spicy logic code into Zeek proper (vs as an external plugin) we don't really mind anymore if it can still build against a previous, external versions of Spicy.

bbannier: We still support building against external installations of Spicy with `--with-spicy`, but I know of no one using that. We also do not test this, and e.g., the new Redis analyzer does not build against the last Spicy release.

bbannier: I pulled this out and we now require that any external Spicy version matches what we ship. This also makes the [bump of `cmake/`](https://github.com/zeek/zeek/pull/4632#discussion_r2185874241) unnecessary.

cmake
bbannier: This should not be needed anymore.

src/spicy/spicyz/main.cc
rsmmr: I'll take a look at this.

bbannier (commented):
Thanks Robin for pushing this over the finish line, this looks much more reasonable now.

rsmmr (approved):
I'll merge.

PR #4633, opened 2025-07-04
Properly finish Spicy TCP child analyzer

bbannier:
We would previously not `Finish` a Spicy TCP analyzer. This manifested in a Spicy child analyzer fed with `protocol_data_in` not receiving an EOD, so if it was waiting for that it would simply fail to finish parsing.

We now explicitly call `Finish` when receiving `EndOfData`. That function is also called from `Done`.


testing/btest/Baseline/spicy.tcp-eod-behavior/analyzer_debug-1024-no-fins.log
```diff
@@ -0,0 +1,2 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+SPICY_FOO	violation	expected 1024 bytes (136 available) (<...>/test.spicy:12:5-12:23)
```
rsmmr: Iirc, this is actually not desired: #4015 changed semantics to not report an error if there's no regular shutdown.

bbannier: Okay, then I need to think a bit more about this. I think it makes sense to have the parent analyzer remove and lifecycle the child as soon as the protocol handle is closed via `protocol_handle_close` as opposed to having them stick around to be cleaned up once the parent is `Done` (which doesn't seem to send EOF for the added test either).

rsmmr: I'll take a look at this.

bbannier (commented):
Thanks! 🚀

rsmmr (approved):
I'll merge.

PR #4635, opened 2025-07-04
Fix for error in ZAM's constant propagation logic

vpax:
Today I turned up a subtle bug in ZAM's constant propagation, where it could mis-re-use an assignment of a constant to a variable in a context where it was unsafe to do so. This PR fixes the issue, while retaining optimization of a common edge case that is very similar but in fact safe to do.


awelzel (approved):
@vpax - I'd just approve and merge based on the fact that you added a test. I wonder if there's someone on the team who feels better equipped, or if that's just an alright process at this point.

I assume this 7.0.x material?

vpax:
> I wonder if there's someone on the team who feels better equipped, or if that's just an alright process at this point.

The two who have a smidgen of acquaintance with the ZAM internals are @ckreibich and @pbcullen , if you want to sync with them. That said, I doubt that at this point they're deep enough into this facet of ZAM to evaluate the bug fix, it's pretty subtle and down-in-the-weeds.

vpax:
> I assume this 7.0.x material?

Yes, please. It's a nasty bug, as can be seen by running the BTest using an un-fixed ZAM.

PR #4636, opened 2025-07-07
script_opt/IDOptInfo: Add NOLINT(bugprone-branch-clone)

awelzel:


timwoj:
@vpax This came up in your recent changes post-merge. No worries about the finding, we expect those sometimes.

I have a question about the function itself though. Should this break after it finds the region? If it already found it, why does it continue searching?

vpax:
> I have a question about the function itself though. Should this break after it finds the region? If it already found it, why does it continue searching?

It's looking for the _latest_ such region prior to a given point, so needs to continue until it runs across one that's later than the given point. (The regions are sorted on start location, which enables this to be a simple linear search.)

vpax:
By the way, what does `bugprone-branch-clone` mean?

timwoj:
It means that two branches in an `if/else` or `switch` have the same block under them. This can be bugprone because it may indicate something should have been different. The NOLINT there tells clang-tidy that we did it on purpose, and to ignore it.

See https://clang.llvm.org/extra/clang-tidy/checks/bugprone/branch-clone.html

awelzel:
> By the way, what does bugprone-branch-clone mean?

Both branches have the same code, so the suggestion is to combine them, but given the complexity and comments in that code, that doesn't make sense here.

PR #4637, opened 2025-07-07
Add `*=` and `/=` operators for arithmetic types

timwoj:
This adds two new operators for arithmetic types: `*=` and `/=`, otherwise known as compound multiply and compound divide. It returns errors for non-arithmetic types.

My only note is that right now it has the parser calling `YYERROR` if it encounters a non-arithmetic type. This causes it to stop parsing, so if you have multiple of these errors only the first one gets reported. This is because others the handlers fall-through into the section for `expr`, which expects $$ to be set to an expression object. If it's not, it segfaults.


timwoj:
Never mind, gonna just close this PR for now. There's some weirdness I need to sort out with patterns.

PR #4638, opened 2025-07-08
Fix indentation of scripts / tests

0xxon:
No functional changes.

Random driveby noticed during reviews.

### DIFF
--- a/testing/btest/scripts/base/frameworks/input/optional.zeek
+++ b/testing/btest/scripts/base/frameworks/input/optional.zeek
@@ -37,11 +37,11 @@ global servers: table[int] of Val = table();
 
 event zeek_init()
 	{
-    outfile = open("../out");
+	outfile = open("../out");
 	# first read in the old stuff into the table...
-	Input::add_table([$source="../input.log", $name="input", $idx=Idx, $val=Val, $destination=servers, 
-				$pred(typ: Input::Event, left: Idx, right: Val) = { right$notb = !right$b; return T; }
-				]);
+	Input::add_table([$source="../input.log", $name="input", $idx=Idx, $val=Val, $destination=servers,
+	                  $pred(typ: Input::Event, left: Idx, right: Val) = { right$notb = !right$b; return T; }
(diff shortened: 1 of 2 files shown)


awelzel (approved):

testing/btest/scripts/base/frameworks/input/optional.zeek
```diff
@@ -37,11 +37,11 @@ global servers: table[int] of Val = table();
...
 	{
-    outfile = open("../out");
+	outfile = open("../out");
 	# first read in the old stuff into the table...
-	Input::add_table([$source="../input.log", $name="input", $idx=Idx, $val=Val, $destination=servers, 
-				$pred(typ: Input::Event, left: Idx, right: Val) = { right$notb = !right$b; return T; }
-				]);
+	Input::add_table([$source="../input.log", $name="input", $idx=Idx, $val=Val, $destination=servers,
+	                  $pred(typ: Input::Event, left: Idx, right: Val) = { right$notb = !right$b; return T; }
+	                 ]);
```
bbannier, line 44: Random drive-by comment not meant to hold up this PR: this is consistent with how we indent similar code elsewhere, but it now mixes tabs and spaces for indention (admittedly in a way which will not break for different widths of a tabstop). I personally would prefer just not doing that, e.g., here

```zeek
	Input::add_table([
		$source="../input.log",
		$name="input",
		$idx=Idx,
		$val=Val,
		$destination=servers,
		$pred(typ: Input::Event, left: Idx, right: Val) = { right$notb = !right$b; return T; }
	]);
```

The general pattern here is: if we need to break arguments, put each one on their own line and just indent them with tabs like the rest of Zeek. This is the same format I have been using e.g., in https://github.com/zeek/zeek/pull/4560.

0xxon: *shrug*

This is the traditional way we have done this in Zeek - use tabs for inentation and speces for alignment. I admittedly kind of like it.

PR #4641, opened 2025-07-08
Storage metrics

timwoj:
This PR adds a new block of metrics for storage. There is a set of general metrics that cover all of the backend operations plus some other statistics. There are also a few metrics specific to SQLite included. The SQLite metrics required some rework of how pragmas are run and how we parse results from SQL commands to make it more generic than just parsing `Val` objects. There aren't any Redis-specific metrics here since Redis has it's own hooks for Prometheus, and we don't need to present them a second time. I expect some quibbling on the naming of things, as is typical for metrics PRs. Also included are btests for sync and async on both backends.


src/storage/Backend.cc
```diff
@@ -46,6 +78,23 @@ void ResultCallback::Complete(OperationResult res) {
...
+
+void ResultCallback::UpdateOperationMetrics(EnumValPtr c) {
+    if ( operation_metrics ) {
+        if ( c == ReturnCode::SUCCESS )
+            operation_metrics->success->Inc();
+        else if ( c != ReturnCode::IN_PROGRESS )
+            operation_metrics->fail->Inc();
+
+        // Store the latency between start and completion in milliseconds.
+        operation_metrics->latency->Observe((util::current_time() - start_time) * 1000);
```
awelzel: Better to store in seconds:

> ...SHOULD use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See [below](https://prometheus.io/docs/practices/naming/#base-units) for a list of base units.

src/storage/Backend.cc
```diff
@@ -145,12 +233,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
     event_mgr.Enqueue(Storage::backend_lost, tag.AsVal(), backend_options, make_intrusive<StringVal>(reason));
 }
 
+void Backend::InitMetrics() {
+    if ( metrics_initialized )
+        return;
+
+    metrics_initialized = true;
+    auto success_family =
+        telemetry_mgr->CounterFamily("zeek", "storage_operation_success",
```
awelzel: Another thought would be have `zeek_storage_backend_operation_results_total{..., result="success|timeout|error"}`

That'd allow plotting the total number of operation_results by not filtering on the `result` label while with the given proposal one would need to sum them via prometheus? It feels a bit more natural to create plots with stacked sections this way, too.

src/storage/Backend.cc
```diff
@@ -145,12 +233,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
+                                     {"operation", "backend_type", "backend_config"}, "Successful Storage Operations");
+    auto fail_family =
+        telemetry_mgr->CounterFamily("zeek", "storage_operation_failed",
+                                     {"operation", "backend_type", "backend_config"}, "Failed Storage Operations");
+    auto timeout_family =
+        telemetry_mgr->CounterFamily("zeek", "storage_operation_timeout",
+                                     {"operation", "backend_type", "backend_config"}, "Timeouted Storage Operations");
+
+    double bounds[] = {0.01, 0.1, 1.0, 10.0, 100.0};
```
awelzel: the cluster metrics have this as a const redef in script land, you can probably copy-paste from there :-)

Hard-coding the bounds isn't great IMO. Maybe there's someone interested in more fine-grained data between 0.1...1.0, or 1...10, but they don't have a way to change without recompilation now.

src/storage/Backend.cc
```diff
@@ -145,12 +233,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
+
+    std::string metrics_config = GetConfigForMetrics();
+    put_metrics = new detail::OperationMetrics(success_family, fail_family, timeout_family, latency_family, "put",
+                                               Tag(), metrics_config);
+    get_metrics = new detail::OperationMetrics(success_family, fail_family, timeout_family, latency_family, "get",
+                                               Tag(), metrics_config);
+    erase_metrics = new detail::OperationMetrics(success_family, fail_family, timeout_family, latency_family, "erase",
+                                                 Tag(), metrics_config);
+
+    bytes_stored_metric = telemetry_mgr->CounterInstance("zeek", "storage_data_stored",
```
awelzel:
```suggestion
    bytes_stored_metric = telemetry_mgr->CounterInstance("zeek", "storage_data_written",
```
?

Mostly because I could see "stored" being confused with a gauge tracking the total amount of data in a backend. But this is a counter only tracking the new bytes written, correct? C.f. with retrieved (which could be renamed to "read" ? )

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -229,9 +253,47 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
+                                     "Storage sqlite backend file size", "", [this]() {
+                                         UpdateMetricStats();
+                                         return current_metric_stats.file_size;
+                                     });
+
     return {ReturnCode::SUCCESS};
 }
 
+void SQLite::UpdateMetricStats() {
+    // Don't update multiple times with every scrape.
```
awelzel: Seems it'd be okay to run the pragmas from the callback directly, but see comment about this capture above.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -275,6 +337,12 @@ OperationResult SQLite::DoClose(ResultCallback* cb) {
         expire_db = nullptr;
     }
 
+    if ( page_count_metric )
+        page_count_metric->RemoveCallback();
```
awelzel, line 338: Does that result in the metric vanishing from scraping, or will it continue to return the metric with the last stored value? Is there a way to invalidate the metric so it's not returned anymore? I suspect that's not possible with the prometheus-cpp model and maybe that's alright.

This could accumulate metrics if someone decides to rotate stores such that for every hour, a new table is created. Maybe that's their problem, not ours.

Could have a flag to disable metrics for a given store though to provide an escape hatch?

testing/btest/scripts/base/frameworks/storage/redis-async.zeek
```diff
@@ -4,16 +4,44 @@
...
 
 redef exit_only_after_terminate = T;
+global b : opaque of Storage::BackendHandle;
+
+event print_metrics_and_close()
+	{
+	print "";
+	print "Post-operation metrics:";
+	local storage_metrics = Telemetry::collect_metrics("zeek", "storage*");
+	for (i in storage_metrics)
```
awelzel:
```suggestion
	for ( _, m in storage_metrics)
```

Nit.

src/storage/Backend.cc
```diff
@@ -145,12 +233,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
+                                                         {{"backend_type", Tag()}, {"backend_config", metrics_config}},
+                                                         "Storage Data Stored", "bytes");
+    bytes_retrieved_metric =
+        telemetry_mgr->CounterInstance("zeek", "storage_data_retrieved",
```
awelzel:
```suggestion
        telemetry_mgr->CounterInstance("zeek", "storage_data_read",
```

If you agree on "written", this should probably be read.

src/storage/Backend.cc
```diff
@@ -145,12 +233,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
 }
 
+void Backend::InitMetrics() {
+    if ( metrics_initialized )
+        return;
+
+    metrics_initialized = true;
+    auto success_family =
+        telemetry_mgr->CounterFamily("zeek", "storage_operation_success",
+                                     {"operation", "backend_type", "backend_config"}, "Successful Storage Operations");
```
awelzel:
```suggestion
                                     {"operation", "type", "config"}, "Successful Storage Operations");
```

I'm not too excited about the repeated `backend_` prefix and wonder about moving it into the metric name. Here I'm not so sure about "type", but it should work :man_shrugging: 
```
storage_backend_operation_results_total{operation, type, config}
```

src/storage/Backend.h
```diff
@@ -242,6 +289,11 @@ class Backend : public zeek::Obj {
     std::string tag_str;
     std::unique_ptr<Serializer> serializer;
 
+    telemetry::CounterPtr bytes_stored_metric;
```
awelzel: These expose a bit much implementation detail for implementors of storage backends: Maybe protected helper methods and hide the counters as private members?

```
void IncBytesWritten(size_t written);
void IncBytesRead(size_t written);
void IncExpiredEntries(size_t expired);
```

These need to be called by implementors explicitly as there's no single place where deserialization is done I presume? Seems okay, just wondering.

src/storage/Backend.cc
```diff
@@ -68,6 +118,18 @@ void OpenResultCallback::Complete(OperationResult res) {
...
+
+    // The rest of the metrics are initialized after the backend opens, but this one has
+    // to be here because it's possible it gets used by the open callback before Open()
+    // fully returns.
+    backends_opened_metric =
+        telemetry_mgr->CounterInstance("zeek", "storage_backends_opened", {}, "Number of backends opened", "");
+}
+
+Backend::~Backend() {
+    delete put_metrics;
```
awelzel: Hmm, could these be unique_ptrs and then use `.get()` for passing them to the callbacks? That ends-up the same as using raw pointers and `delete` in the destructor, no?

src/storage/Backend.cc
```diff
@@ -46,6 +78,23 @@ void ResultCallback::Complete(OperationResult res) {
...
+    operation_metrics = m;
+    start_time = util::current_time();
+}
+
+void ResultCallback::UpdateOperationMetrics(EnumValPtr c) {
+    if ( operation_metrics ) {
+        if ( c == ReturnCode::SUCCESS )
+            operation_metrics->success->Inc();
+        else if ( c != ReturnCode::IN_PROGRESS )
+            operation_metrics->fail->Inc();
```
awelzel, line 85: Could we have a gauge for inflight operations? Probably mostly interesting for async backends.

src/storage/Backend.cc
```diff
@@ -46,6 +78,23 @@ void ResultCallback::Complete(OperationResult res) {
     trigger->Release();
 }
 
+void ResultCallback::InitOperationMetrics(detail::OperationMetrics* m) {
+    operation_metrics = m;
+    start_time = util::current_time();
```
awelzel:
```suggestion
    start_time = util::current_time(true);
```

I think we want real time, regardless of pseudo-realtime.

src/storage/Backend.h
```diff
@@ -234,6 +275,12 @@ class Backend : public zeek::Obj {
      */
     void CompleteCallback(ResultCallback* cb, const OperationResult& data) const;
 
+    /**
+     * Returns a string compatible with Prometheus that's used as a tag to differentiate
+     * entries of backend instances.
+     */
+    virtual std::string GetConfigForMetrics() const = 0;
```
awelzel:
```suggestion
    virtual std::string GetConfigMetricsLabel() const = 0;
```

This isn't hot, right? It's just called once at instantiation?

Given all other virtual hook methods in `Backend` are `private` and use the `Do...()` pattern, suggest to stay consistent and follow that?

src/storage/Backend.h
```diff
@@ -82,10 +94,37 @@ class ResultCallback {
 
     OperationResult Result() const { return result; }
 
+    /**
+     * Stores the collection of metrics instruments to update when the operation completes
+     * and sets the start time for an operation to be used to update the latency metric
+     * when the operation completes. This is unset for open/close callbacks.
+     */
+    void InitOperationMetrics(detail::OperationMetrics* m);
```
awelzel:
```suggestion
    void Init(detail::OperationMetrics* m);
```

Maybe will do other stuff in the future.

src/storage/backend/redis/Redis.cc
```diff
@@ -406,6 +417,8 @@ OperationResult Redis::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
     if ( connected && status == REDIS_ERR )
         return {ReturnCode::OPERATION_FAILED, util::fmt("Failed to queue put operation: %s", async_ctx->errstr)};
 
+    cb->AddDataTransferredSize(key_data->size() + val_data->size());
```
awelzel, line 421:
```suggestion
    cb->AddDataTransferredSize(key_data->size() + val_data->size());
```

I'd be fine assuming the data was written and just increasing the stored/written metric here rather than on success only.  That'd allow dropping the API for the transferred size from the callback.

src/storage/Backend.cc
```diff
@@ -109,29 +181,45 @@ OperationResult Backend::Put(ResultCallback* cb, ValPtr key, ValPtr value, bool
         return ret;
     }
 
-    return DoPut(cb, std::move(key), std::move(value), overwrite, expiration_time);
+    auto ret = DoPut(cb, std::move(key), std::move(value), overwrite, expiration_time);
+    if ( cb->IsSyncCallback() )
+        cb->UpdateOperationMetrics(ret.code);
```
awelzel, line 176: I was wondering if we could call `cb->Complete()` here instead, similar to what you did for Open()? That might allow to drop the `UpdateOperationMetrics()` API and just do it in `Complete()`, but I haven't tried

awelzel (commented):
Hey Tim, had some thoughts. Lets see what you think :+1:

src/storage/backend/sqlite/SQLite.cc
timwoj: I'm pretty sure the instrument continues to exist, it just returns the same value because the callback doesn't run anymore. I'll test that though.

src/storage/Backend.h
timwoj: Right, just at instantiation when the instruments are setup.

src/storage/backend/redis/Redis.cc
timwoj: This was the result of a conversation with @retr0h. We decided that the read/written metrics are only really useful on successful operation, especially in the light of operations to async backends. If I sent a put operation to Redis, and it failed for some reason the Redis side but didn't actually write, does it count as data written?

src/storage/Backend.h
timwoj: I was going to say that I wouldn't expect this to do anything specific in Backend that would warrant the `Do...` version, but then I see I did the same thing with Poll(), which likely will never do anything specific in Backend either. I guess moving it to that matter would be consistent.

src/storage/Backend.cc
timwoj: Yeah, that looks like it should be fine. That method can go away then.

src/storage/backend/sqlite/SQLite.cc
timwoj: I can confirm that as long as the instrument continues to exist, Zeek continues to return the last known value for it. The callback doesn't run again but since `RunCallback()` in the family objects stores the value from it into the instrument, the last value can still be returned.

timwoj: I don't see a method in the telemetry framework completely invalidate an instrument to it from the reporting. That might be useful feature, but I don't think it would get used very much.

src/storage/Backend.cc
timwoj: Looking at this again, I'm actually marginally concerned about this change. For an async-only backend, the backend should only be completed once the response comes through. It gets completed by the backend itself. This is true even if it's a sync callback. Sync callbacks are handled/completed by polling after the operation method (e.g. `Backend::Put`). Calling `CompleteCallback` here potentially breaks that (but surprisingly, didn't break the redis-sync btest...). I think just updating the metrics values is still the right way to go about this.

src/storage/Backend.h
awelzel: I was alternatively wondering if users should be able to provide a label via the configuration record: `config_label: string &optional` - and if it's unset. If that's set, use it as config label, otherwise do the mangling of the configuration pieces.

That could be a better escape hatch than to disable metrics.

src/storage/Backend.cc
awelzel: Let me resolve this, I don't think it's worth it for now.

src/storage/backend/redis/Redis.cc
awelzel: Hmm. Can see it from both sides. I'm thinking that if you send a 5MB value with `overwrite=F` and nothing is overwritten, does that mean it's not successful? And so we don't account for those 5MB? To me `written` / `read` would be more like bytes in / bytes out. Whether they actually have an effect would be separate to me.

E.g., could separately have gauges for the total number of keys or the db size.

src/storage/Backend.cc
```diff
@@ -46,6 +70,23 @@ void ResultCallback::Complete(OperationResult res) {
...
+    operation_metrics = m;
+    start_time = util::current_time(true);
+}
+
+void ResultCallback::UpdateOperationMetrics(EnumValPtr c) {
+    if ( operation_metrics ) {
+        if ( c == ReturnCode::SUCCESS )
+            operation_metrics->success->Inc();
+        else if ( c != ReturnCode::IN_PROGRESS )
+            operation_metrics->fail->Inc();
```
awelzel, line 85: One more thought: A `ReturnCode::KEY_EXISTS` or `ReturnCode::KEY_NOT_FOUND` will be tracked in the fail metric, but they don't really seem to be failures. Maybe "result" could be "hit" and "miss" for `Get()` instead of just "success". For `Put()` - I'm missing good words. Maybe "store" or "accept" and "exist" or "drop".

src/storage/backend/redis/Redis.h
```diff
@@ -27,7 +27,9 @@ class Redis final : public Backend, public iosource::IOSource {
      *
      * @return The debugging name.
      */
-    const char* Tag() override { return tag_str.c_str(); }
+    const char* Tag() override { return Backend::Tag(); }
+
+    std::string DoGetConfigMetricsLabel() const override;
```
awelzel: Nit, move to `private:` below?

src/storage/backend/sqlite/SQLite.cc
awelzel: > This could accumulate metrics if someone decides to rotate stores such that for every hour, a new table is created. Maybe that's their problem, not ours.
>
> Could have a flag to disable metrics for a given store though to provide an escape hatch?

I'm going to resolve this thread. I've started another comment suggesting to add a way to configure the label via the config record. That'd give users an escape hatch to control the carnality of created metrics themselves.

Prom libraries probably assume there's no ephemeral/short-lived metrics, because that's not really the idea anyhow.

src/storage/backend/sqlite/SQLite.h
```diff
@@ -24,7 +24,16 @@ class SQLite final : public Backend {
      */
     bool IsOpen() override { return db != nullptr; }
 
+    std::string DoGetConfigMetricsLabel() const override;
```
awelzel: Nit, move to private.

src/storage/Backend.cc
timwoj: Good point. I think it would be considered a success. Fail is more for "there was a hard error while doing this operation". How about adding a third field for key conflicts?

timwoj: Or actually, how about "success", "fail", and "error"? That would cover both put and get operations with the same naming.

src/storage/backend/redis/Redis.cc
retr0h: I think I agree with @awelzel here

timwoj:
There's some problems with how we sort metrics in btests, namely that we really only key off the labels. It's causing non-determinism in the btests right now. I need to sort that out still.

src/storage/Backend.cc
awelzel: I'm not sure about "fail" for "key exists" or "key not found". It seems either could just be part of normal operations and labeling these "failure" seems wrongly alarming.

I'd probably just put them under success unless we introduce explicit "miss" and maybe "exists" metrics? I could see someone go, oh. "I've a lot of  misses for "get" operations, even than "hits". But that seems unexpected, let me check what's going on".

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -241,6 +266,33 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
...
+                                         auto res = RunPragma("page_count", std::nullopt, double_parser);
+                                         if ( res.code == ReturnCode::SUCCESS ) {
+                                             last_page_count_value = cast_intrusive<DoubleVal>(res.value)->Get();
+                                         }
+
+                                         return last_page_count_value;
+                                     });
+
+    file_size_metric =
+        telemetry_mgr->GaugeInstance("zeek", "storage_sqlite_file_size", {{"config", GetConfigMetricsLabel()}},
```
awelzel: Could we add a unit into the name? I guess here it's `bytes`. Also replacing `file` with `database`.

```
zeek_storage_sqlite_database_size_bytes
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -241,6 +266,33 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
     get_expiry_last_run_stmt = std::move(stmt_ptrs[6]);
     update_expiry_last_run_stmt = std::move(stmt_ptrs[7]);
 
+    page_count_metric =
+        telemetry_mgr->GaugeInstance("zeek", "storage_sqlite_page_count", {{"config", GetConfigMetricsLabel()}},
```
awelzel:
```suggestion
        telemetry_mgr->GaugeInstance("zeek", "storage_sqlite_database_size_pages", {{"config", GetConfigMetricsLabel()}},
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -341,14 +399,17 @@ OperationResult SQLite::DoPut(ResultCallback* cb, ValPtr key, ValPtr value, bool
...
+    auto step_result = Step(stmt.get(), nullptr);
     if ( ! overwrite )
         if ( step_result.code == ReturnCode::SUCCESS ) {
             int changed = sqlite3_changes(db);
             if ( changed == 0 )
                 step_result.code = ReturnCode::KEY_EXISTS;
         }
 
+    if ( step_result.code == ReturnCode::SUCCESS )
+        IncBytesWrittenMetric(val_data->size());
```
awelzel: As in the other comment, would just increment after the `Step()` no matter what the outcome. Even if the key exists. If one continuously writes a huge entry, but it results in key exists, the `database_size_bytes` will stay constant, but bytes written would contiguously increase (or have a high `rate()`). If we only update on success, there's actually a lot of traffic/churn, but it won't be visible in the written metric.

src/telemetry/Manager.cc
```diff
@@ -239,9 +216,40 @@ static bool comparer(const std::optional<ZVal>& a, const std::optional<ZVal>& b,
...
+    auto b_opts = b_r->GetField<RecordVal>("opts");
+
+    auto a_name = a_opts->GetField<StringVal>("name");
+    auto b_name = b_opts->GetField<StringVal>("name");
+    if ( a_name->Len() > b_name->Len() )
+        return false;
+    if ( a_name->Len() < b_name->Len() )
+        return true;
+    if ( memcmp(a_name->Bytes(), b_name->Bytes(), a_name->Len()) < 0 )
+        return true;
```
awelzel, line 229: This seems a bit complicated, would the following do?
```
if ( auto r = strcmp(a_name->CheckString(), b_name->CheckString()); r != 0 )
    return r < 0;
```

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -118,7 +143,7 @@ OperationResult SQLite::DoOpen(OpenResultCallback* cb, RecordValPtr options) {
         auto vs = iter.value->GetVal()->AsStringVal();
         auto vs_sv = vs->ToStdStringView();
 
-        auto pragma_res = RunPragma(ks_sv, vs_sv);
+        auto pragma_res = RunPragma(ks_sv, vs_sv, nullptr);
```
awelzel:
```suggestion
        auto pragma_res = RunPragma(ks_sv, vs_sv);
```

src/storage/Backend.cc
```diff
@@ -145,12 +223,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
+
+    bytes_written_metric =
+        telemetry_mgr->CounterInstance("zeek", "storage_data_written", {{"type", Tag()}, {"config", metrics_config}},
+                                       "Storage Data Stored", "bytes");
+    bytes_read_metric =
+        telemetry_mgr->CounterInstance("zeek", "storage_data_read", {{"type", Tag()}, {"config", metrics_config}},
+                                       "Storage Data Retrieved", "bytes");
+
+    expired_entries_metric =
+        telemetry_mgr->CounterInstance("zeek", "storage_expired_entries", {{"type", Tag()}, {"config", metrics_config}},
```
awelzel:
```suggestion
        telemetry_mgr->CounterInstance("zeek", "storage_evicted_entries", {{"type", Tag()}, {"config", metrics_config}},
```

I've been pondering. Tracking expired entries is kind of best effort, actually? The recent work on #4616 made me realize that expired entries might exist and even overwritten without ever being accounted for as expired. Further, backends with native expiration won't ever update this metric (unless they go out of their way to do so - thinking triggers or expiration callbacks). So I'm wondering if we should soften the terminology here to "evicted" or "expunged" entries. IMO that'd help prevent confusion why expired entries aren't tracked reliably across backends (and also that expiration isn't tracked if backends do it natively).

Sorry for all the naming nits, but the metrics are kind of a user-visible API. I'm happy to be overruled, but still think better to put my thoughts here.

src/storage/Backend.cc
```diff
@@ -145,12 +223,57 @@ void Backend::CompleteCallback(ResultCallback* cb, const OperationResult& data)
...
+                                       "Storage expired entries", "");
+}
+
+void Backend::IncBytesWrittenMetric(size_t written) { bytes_written_metric->Inc(written); }
+void Backend::IncBytesReadMetric(size_t read) { bytes_read_metric->Inc(read); }
+void Backend::IncExpiredEntriesMetric(size_t expired) { expired_entries_metric->Inc(expired); }
```
awelzel, line 275: See comment above. `IncEvictedEntries()`  - not sure the `Metric` suffix is needed.

awelzel (commented):
I have a few more naming thoughts, but I think that's the last round.

It's a bit unfortunate that there's some code in the common `Backend.cc`, but then also in the backends themselves. If `DoGet()` or `DoPut()` returned some sort of struct containing the necessary information, it maybe could be consolidated, but I haven't fully thought it through. The async/sync variations don't make this easy either.

Next one should be a :shipit:  :crossed_fingers:

src/storage/backend/sqlite/SQLite.cc
timwoj: It's based on the `page_count` pragma value, which SQLite users would understand. Hence the naming.

timwoj: I suppose that information could be in the description instead.

src/storage/backend/sqlite/SQLite.cc
timwoj: Yeah, I was fighting the ordering stuff yesterday afternoon and never got back to this.

src/telemetry/Manager.cc
timwoj: For some reason it was running into a bunch of heap-overflow issues last night. I tried using `ToStdString` and `ToStdStringView` also. No idea why, but length checks there were the solution that worked correctly.

src/storage/Backend.cc
timwoj: Should we also then rename all of the "Expire" methods to "Evict"?

src/storage/backend/sqlite/SQLite.cc
awelzel: > It's based on the page_count pragma value, which SQLite users would understand. Hence the naming.

Gotcha, the Prometheus best practices want a unit at the end and having `count` in something that's of type `counter` is also a bit weird, but your call :-)

src/storage/Backend.cc
awelzel: As in DM: Ignore, probably better to stick with expire and explain the metric is best-effort.

PR #4646, opened 2025-07-10
Documentation: connection_SYN_packet

0xxon:
Slightly clarify documentation related to the connection_SYN_packet event.


src/analyzer/protocol/tcp/events.bif
```diff
@@ -155,7 +155,8 @@ event connection_reset%(c: connection%);
 event connection_pending%(c: connection%);
 
 ## Generated for a SYN packet. Zeek raises this event for every SYN packet seen
-## by its TCP analyzer.
+## by its TCP analyzer. This includes packets that have other flags set - like
+## it is the case for SYN-ACK packets.
```
timwoj:
```suggestion
## by its TCP analyzer. This includes packets that have other flags set - like
## in the case of SYN-ACK packets.
```

PR #4647, opened 2025-07-10
Crash fixes for constructs that expect add/delete expressions to return values

vpax:
As flagged by @evantypanski, now that `add` and `delete` are expressions, it's possible to crash Zeek by trying to use their values. This PR fixes addresses this by changing their types to be `void` rather than nil pointers, and updating ZAM's AST reduction to not treat the expressions as regular unary operations.


testing/btest/language/add-del-expr-void.zeek
```diff
@@ -0,0 +1,14 @@
+# @TEST-DOC: checks for type-checking for add/delete expressions
+#
+# @TEST-EXEC-FAIL: zeek -b %INPUT >out
+# @TEST-EXEC: btest-diff out
+# @TEST-EXEC: btest-diff .stderr
```
evantypanski: Seems like this will need to remove absolute paths to pass. Also probably best to just combine input here to avoid the empty file, then redo baselines and remove the now-orphan `.stderr`:


```suggestion
# @TEST-EXEC-FAIL: zeek -b %INPUT >out 2>&1
# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff out
```

vpax: Thanks for flagging! Now updated.

PR #4650, opened 2025-07-10
Add NEWS entry about C++ 20 [nomail] [skip ci]

timwoj:


NEWS
```diff
@@ -12,6 +12,11 @@ release.
 Breaking Changes
 ----------------
 
+- Zeek and all of its associated submodules now require C++20-capable compilers to
+  build. This will let us move forward in using more modern C++ features and replace some
+  workarounds that we have been carrying. Minimum recommended versions of compilers are
+  GCC 10, Clang 8, and MSVC 17.
```
evantypanski: Can this say Visual Studio 2022 or something? (instead of MSVC 17)

The numbers for these are actually really weird, MSVC 17 would be correlated with `_MSC_VER` 1700 - which is Visual Studio 2012.

Visual Studio 2022 is Visual Studio version code 17, with MSVC version 19. All C++20 support (to my knowledge) has been in MSVC 19 - which has been around since the big shift with Visual Studio 2015.

So I'd ignore all that and call out the year

[reference](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering)

timwoj: 👍🏼

PR #4651, opened 2025-07-10
Remove ghc::filesystem submodule, switch to std::filesystem

timwoj:
Now that C++20 support is official there's no reason we can't switch to `std::filesystem` everywhere. This removes the `auxil/filesystem` submodule and deprectates the `zeek::filesystem` alias in `util.h`. It also replaces all of the uses of `zeek::filesystem` with `std::filesystem` directly.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -21,6 +21,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <filesystem> // Remove in v8.1.
 #include <string>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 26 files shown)


src/util.h
```diff
@@ -79,18 +80,8 @@ extern "C" {
...
-namespace filesystem = std::filesystem;
-}
 inline constexpr std::string_view path_list_separator = ";";
 #else
-// Expose ghc::filesystem as zeek::filesystem until we can
-// switch to std::filesystem on all platforms.
-#include "zeek/3rdparty/ghc/filesystem.hpp"
-namespace zeek {
-namespace filesystem = ghc::filesystem;
-}
```
bbannier, line 93: This alias is not in a detail namespace and I think shouldn't just go away without a deprecation cycle (if not possible in code, calling out in `NEWS.rst` is IMO fine as well). We can have them point to `std::filesystem` now though, e.g.,

```cpp
#include <filesystem>
namespace zeek {
namespace filesystem = std::filesystem;
}
```

I think the new global alias below could just go away with that.

timwoj: I actually did that, and then forgot to force-push the branch 🤦🏼

timwoj: I'm confused. The alias still exists a couple lines below in this diff in the zeek namespace.

bbannier: You are right 🤦

awelzel (approved):
LGTM - the missing deprecation is a bit unfortunate. Maybe run with "full ci" once before merging?

timwoj:
> LGTM - the missing deprecation is a bit unfortunate. Maybe run with "full ci" once before merging?

I manually kicked off the tasks that hadn't run previously and they all passed, except the include_plugins one that has known problems.

PR #4654, opened 2025-07-11
Deprecate util::starts_with/ends_with, use std::string/string_view versions

timwoj:
C++ 20 added `starts_with` and `ends_with` to `std::string{_view}`. Deprecate our custom versions of those, and move to the `std::` versions instead. Alternatively I could see keeping the `util::` versions as wrappers that can take `const char*` arguments, since otherwise we have to do the conversion manually at the call sites.


src/Options.cc
```diff
@@ -320,8 +320,7 @@ Options parse_cmdline(int argc, char** argv) {
     else {
         if ( argc > 1 ) {
             auto endsWith = [](const std::string& str, const std::string& suffix) {
-                return str.size() >= suffix.size() &&
-                       0 == str.compare(str.size() - suffix.size(), suffix.size(), suffix);
+                return str.size() >= suffix.size() && str.ends_with(suffix);
```
awelzel:
```suggestion
                return str.ends_with(suffix);
```

Wouldn't this be implied?

awelzel (approved):

awelzel:
I was wondering the same, but skimming this PR it didn't seem such a big deal and might be an incentive to replace some `const char*`s over to `std::string` / `std::string_view`. I do admit `std::string_view{s}.starts_with(sub)` is a bit awkward.

timwoj:
How about leaving the util helpers not deprecated, but encouraging their use only when the type conversion is necessary?

PR #4662, opened 2025-07-15
SMTP: Support full mail data file analysis

### LINKED ISSUE #3900: Add EML feature to the SMTP analyzer

awelzel:
This adds a `SMTP::enable_mail_data_file_analysis` feature allowing to forward a full email as `fa_file` to the file analysis 
framework.

The motivating use-case is extracting emails as `.eml` files, including all attachments via the EXTRACT analyzer.

Essentially, anything between `DATA` and `.` or everything coming in via BDAT chunks is forwarded to the file analysis framework.

This was requested in #3900  by @zrobinette12  and initially implemented in #4591 by @cccs-jsjm and reworked by me.

Closes #3900


awelzel:
@rsmmr - I've requested a review from you as you might have a hunch/reaction for the approach taken. Looks okay to me. I've pondered if the functionality should turn on automatically when implementing `smtp_mail_data_file()`, but I don't think we do that kind of magic anywhere, so a opt-in option seemed better (and the file still comes in via `file_new()` or `file_over_new_connection()` when only enabling the option.

cc also @ckreibich - in case you'd want to skim over the approach, event and option.

initconf:
Hello Arne, Yes, I have vested interest in this. Infact I was designing some new detection stuff around full email body and feeding it to ML models for some classifications feedback !

I'd have tapped into mime_entity_data but seems like your file analysis work might be more convenient. I need to check this add out !! Very timely !!!

src/file_analysis/File.h
```diff
@@ -176,6 +176,15 @@ class File {
...
     void FileEvent(EventHandlerPtr h, Args args);
 
+    /**
+     * Drain the event queue and this file's analyzer modifications after
+     * raising a custom event via FileEvent.
+     *
+     * Using internal events (file_new, file_over_new_connection, file_sniff
+     * and file_extraction_limit) invokes this automatically.
+     */
+    void FileEventDrain();
```
rsmmr, line 186: The method name here feels misleading: it's draining all events. Plus, "FileEvent" is used differently in the other methods just above.

scripts/base/init-bare.zeek
```diff
@@ -667,6 +667,12 @@ export {
 	## is introduced and a weird is raised. Conventionally, MIME messages
 	## have a maximum line length of 1000 octets when properly encoded.
 	const bdat_max_line_length = 4096 &redef;
+
+	## Whether to send data of individual top-level RFC822 messages
+	## in SMTP transactions to the file analysis framework.
+	##
+	## :zeek:see:`smtp_mail_data_file`.
+	const enable_mail_data_file_analysis = F &redef;
```
rsmmr, line 675: Suggest to rename the option to something more descriptive, like maybe `enable_{eml,rfc822}_file_analysis` or `enable_full_mail_file_analysis`.

src/analyzer/protocol/smtp/SMTP.cc
```diff
@@ -880,6 +896,10 @@ bool SMTP_Analyzer::ProcessBdatArg(int arg_len, const char* arg, bool orig) {
     return true;
 }
 
+std::string SMTP_Analyzer::MailDataFileAnalysis(uint64_t len, const u_char* data) {
```
rsmmr, line 899: More naming: "mail data"  just doesn't sound very intuitive to me.

src/analyzer/protocol/smtp/SMTP.h
```diff
@@ -90,6 +91,7 @@ class SMTP_Analyzer final : public analyzer::tcp::TCP_ApplicationAnalyzer {
     std::unique_ptr<detail::SMTP_BDAT_Analyzer> bdat; // if set, BDAT chunk transfer active
 
     analyzer::mime::MIME_Mail* mail;
+    std::string mail_fuid; // fuid for mail data file analysis
```
rsmmr, line 94: More naming: `mail` is already used differently in the line just above, so how about naming this, say, `rfc822_fuid`?

src/analyzer/protocol/smtp/events.bif
```diff
@@ -106,3 +106,15 @@ event smtp_unexpected%(c: connection, is_orig: bool, msg: string, detail: string
...
+## of an RFC822 message in an SMTP transaction.
+##
+## This event can be leveraged to extract the full mail data to disk.
+##
+## c: The corresponding SMTP connection.
+##
+## f: The file corresponding to the mail data.
+##
+## .. zeek:see:: SMTP::enable_mail_data_file_analysis
+event smtp_mail_data_file%(f: fa_file, c: connection%);
```
rsmmr, line 120: Why do we actually need this new event? Couldn't one work with the regular file analysis framework events instead?

scripts/base/protocols/smtp/files.zeek
```diff
@@ -19,10 +23,17 @@ export {
...
+	c$smtp$mail_fuid = f$id;
+	}
+
 function get_file_handle(c: connection, is_orig: bool): string
 	{
+	# Adding mail_fuid here if set to ensure the top-level mail message
+	# and the first MIME attachment do not get the same fuid allocated.
 	return cat(Analyzer::ANALYZER_SMTP, c$start_time, c$smtp$trans_depth,
-	           c$smtp_state$mime_depth);
+	           c$smtp?$mail_fuid ? c$smtp$mail_fuid : "", c$smtp_state$mime_depth);
```
rsmmr, line 36: Not sure if this is a great idea but if `mime_depth` came as -1 for the RFC822 message, things would automartically work (also elsewhere where scripts are hooking into the file analysis for SMTP).

rsmmr (commented):
I like the functionality, and it's nice how it uses the `message/rfc822` MIME type.

My main issue is naming: the whole notion of "mail data" doesn't sound very intuitive to me. I'd pick a different term for meaning "whole mail" and then just that consistently everywhere (i.e., method/event/variable/option names). Maybe that could be "rfc822" to match the MIME type, so the option would become `enable_rfc822_file_analysis`, etc.

Plus, see the question about the new event.

awelzel:
> My main issue is naming: the whole notion of "mail data" doesn't sound very intuitive to me

I actually took "mail data" from the SMTP RFC and it became intuitive and never questioned it after getting used to it. I preferred "mail data" over "message content" for brevity and being a bit more abstract.


> [2.3.9](https://datatracker.ietf.org/doc/html/rfc5321#section-2.3.9).  Message Content and Mail Data
>
>   The terms "message content" and "mail data" are used interchangeably
>   in this document to describe the material transmitted after the DATA
>   command is accepted and before the end of data indication is
>   transmitted.  Message content includes the message header section and
>   the possibly structured message body.  The MIME specification ([RFC](https://datatracker.ietf.org/doc/html/rfc2045)
>   [2045](https://datatracker.ietf.org/doc/html/rfc2045) [[21](https://datatracker.ietf.org/doc/html/rfc5321#ref-21)])
> provides the standard mechanisms for structured message bodies.

The BDAT RFC uses "message data" instead:

> The message data is sent immediately after the trailing CRLF
>   of the BDAT command line.  Once the receiver-SMTP receives the
>   specified number of octets, it will return a 250 reply code.
>
>   The optional LAST parameter on the BDAT command indicates that this
>   is the last chunk of message data to be sent.  


Does that somehow change your stance? Otherwise, I'm happy to change to `SMTP::enable_rfc822_file_analysis` and adapt the method names to `Rfc822FileDataIn()` or so.

src/analyzer/protocol/smtp/events.bif
awelzel: I think yes, by also checking whether the option for file analysis is enabled. In that case, the first file in an SMTP transaction will always be the RFC822 message / mail data (and could assert on the mime type, too).

I do like the explicit event, but reusing the old ones will also void the `FileEventDrain()` discussion, so that's probably good :sweat_smile:

scripts/base/protocols/smtp/files.zeek
awelzel: Interesting. `mime_depth` is a `count` today and documented to track the number of MIME encoded files, so that feels a bit magic (granted, including the fuid does, too, but it's not changing existing semantics). Passing on it.

awelzel:
> Does that somehow change your stance? Otherwise, I'm happy to change to `SMTP::enable_rfc822_file_analysis` and adapt the method names to `Rfc822FileDataIn()` or so.

I've changed some variables to `rfc822_message_fuid` and the longish `SMTP::enable_rfc822_message_file_analysis` would make this consistent. Let me re-spin this as a v3 PR.

awelzel:
Opened separate PR at #4664 after renaming and dropping the extra event.

PR #4664, opened 2025-07-16
SMTP: Support full mail data file analysis - v3

### LINKED ISSUE #3900: Add EML feature to the SMTP analyzer

awelzel:
This adds a `SMTP::enable_rfc822_msg_file_analysis` feature allowing to forward a full email as `fa_file` to the file analysis 
framework.

The motivating use-case is extracting emails as `.eml` files, including all attachments via the EXTRACT analyzer.

Essentially, anything between `DATA` and `.` or everything coming in via BDAT chunks is forwarded to the file analysis framework.

This was requested in #3900  by @zrobinette12  and initially implemented in #4591 by @cccs-jsjm and reworked by me.

Closes #3900

---

This PR replaces #4662

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -153,6 +153,21 @@ New Functionality
   incoming and outgoing event metrics of a specific WebSocket application, simply
   by setting the X-Application-Name header.
 
+- The SMTP analyzer can now optionally forward the top-level RFC 822 message individual
+  SMTP transactions to the file analysis framework. This can be leveraged to extract
+  emails in form of ``.eml`` files from SMTP traffic to disk.
+
+  To enable this feature, set the ``SMTP::enable_rfc822_msg_file_analysis`` option
+  and implement an appropriate ``file_new()`` or ``file_over_new_connection()`` handler:
+
+	redef SMTP::enable_rfc822_msg_file_analysis = T;
+
+	event file_over_new_connection(f: fa_file, c: connection, is_orig: bool) {
+		if ( f$id == c$smtp$rfc822_msg_fuid )
+			Files::add_analyzer(f, Files::ANALYZER_EXTRACT, [$extract_filename="email"]);
+	}
+
+
 - Generic event metadata support. A new ``EventMetadata`` module was added allowing
   to register generic event metadata types and accessing the current event's metadata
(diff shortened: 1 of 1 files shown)


rsmmr (approved):
Yeah, this reads much better to me. 

I saw the comment in the old PR that "mail data" comes from the SMTP RFC, and I get why it makes sense there, but it still feels harder to understand to me when taken outside of that context (which I would argue is the case here, even though this is the SMTP analyzer). `rfc822_msg` works for me.

PR #4667, opened 2025-07-16
cluster/zeromq: Add DROP policy

### LINKED ISSUE #4431: cluster/zeromq: Support drop policy for overload situations
With Zeek 7.1 and Zeek 7.2, the ZeroMQ cluster backend implementation is similar to the original Broker approach of never dropping events. In contrast to Broker, the `Cluster::publish()` script layer calls will block in overload situations rather than buffering and running out of memory.

Over the past releases, Broker was changed to default to dropping messages in overload situations, providing metrics to users when this happens.

This ticket is about introducing a `const redef` variable in the `Cluster::Backend::ZeroMQ` module that allows switching to a dropping policy. IMO we should keep the blocking policy available, too, for testing. The dropping policy is difficult to work with if you never want messages to be dropped (outside of increasing buffers to be infinite) and don't care about potentially blocking script layer calls.

The current idea I'm thinking of is to adapt the places where explicit blocking is happening to instead drop and if possible propagate back the error to a `Cluster::publish()` call. Additionally, a "warn once" would be good for the first dropped message, just so that users without a full Prometheus get a signal on stderr that this has just happened to them.

Summary
* Make dropping events a policy configurable via the ZeroMQ module
* Introduce metrics for dropped events, possibly labeled with the topic
* Check whether metrics can also be updated in the central proxy in case dropping happens there

awelzel:
This adds a new `Cluster::Backend::ZeroMQ::overflow_policy` (overflow from Broker) that can be toggled to `DROP` in order to drop events instead of blocking until there's room in the local queue.

I had initially hoped that `Cluster::publish()` could return `F` if it fails, but the actual dropping happens asynchronously in a separate thread and there's no easy channel back. That might also be okay: On the Broker side this isn't propagated back and anyhow it's unclear what a user should do other than reporting an warning/error.

The stalls part of `zeek_cluster_zeromq_xpub_stalls_total` metric was renamed to be "blocks" instead of "stalls". Additionally, there's now `zeek_cluster_zeromq_xpub_drops_total` to count the dropped messages of a node due to the queue overflowing.

Closes #4431


scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -28,6 +28,45 @@
...
+	## an event can be published. If packet processing is more important than
+	## cluster communication, it is recommended to select the drop policy.
+	##
+	## For PCAP processing or cluster performance testing, where packet drops
+	## are usually no concern, the block policy might be more appropriate.
+	##
+	## Note that the return value of :zeek:see:`Cluster::publish` is currently
+	## not affected, even if events are dropped. It will return **T**
+	## even if the event is later dropped locally as the logic is running
+	## on a separate thread.
```
bbannier, line 55: Let's capture this changed contract also in the docstring of `Cluster::publish` which currently claims `T` means "sent".

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -28,6 +28,45 @@
...
+	};
+
+	## Overflow policy for ZeroMQ.
+	##
+	## When publishing an event via :zeek:see:`Cluster::publish` would fail
+	## due to a node's local XPUB socket reaching its high water mark
+	## (see :zeek:see:`Cluster::Backend::ZeroMQ::xpub_sndhwm`), Zeek may
+	## either block until there's room available to queue the event, or
+	## drop the event. The default behavior is to block. Note that this can
+	## eventually result in :zeek:see:`Cluster::publish` calls to block,
```
bbannier, line 44: The thing that personally worries me most about `BLOCK` is that it cannot only halt the current node, but via zeromq backpressure any node working on a topic blocked anywhere. Can you think of some way to call that out here?

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -28,6 +28,45 @@
...
+	## a counter labeled ``zeek_cluster_zeromq_xpub_blocks_total`` is
+	## incremented on every blocking operation. For the drop policy, the
+	## ``zeek_cluster_zeromq_xpub_drops_total`` metric is incremented for
+	## every dropped event.
+	##
+	## Increasing :zeek:see:`Cluster::Backend::ZeroMQ::xpub_sndhwm` can help
+	## to account for bursty publish behavior, but should be carefully evaluated
+	## as it may result in increased peak memory usage or even out-of-memory
+	## situations if set to 0.
+	const overflow_policy = BLOCK &redef;
```
bbannier, line 68: Like mentioned offline, I'd be in favor of using `DROP` as default since I cannot imagine a production user who would be fine with a stalled cluster just to ensure "reliable" event delivery (delivery is fire & forget anyway and the promises around this are pretty weak).

Maybe something to discuss in a bigger group, @ckreibich.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -114,10 +115,43 @@ void ZeroMQBackend::DoInitPostScript() {
...
+        static_cast<int>(zeek::id::find_val<zeek::BoolVal>("Cluster::Backend::ZeroMQ::log_immediate")->AsBool());
+    log_sndhwm = static_cast<int>(zeek::id::find_val<zeek::IntVal>("Cluster::Backend::ZeroMQ::log_sndhwm")->AsInt());
+    log_sndbuf = static_cast<int>(zeek::id::find_val<zeek::IntVal>("Cluster::Backend::ZeroMQ::log_sndbuf")->AsInt());
+    log_rcvhwm = static_cast<int>(zeek::id::find_val<zeek::IntVal>("Cluster::Backend::ZeroMQ::log_rcvhwm")->AsInt());
+    log_rcvbuf = static_cast<int>(zeek::id::find_val<zeek::IntVal>("Cluster::Backend::ZeroMQ::log_rcvbuf")->AsInt());
+
+    // The values of ZeroMQ::OverflowPolicy
+    static const auto& block_policy = zeek::id::find_val<zeek::EnumVal>("Cluster::Backend::ZeroMQ::BLOCK");
+    static const auto& drop_policy = zeek::id::find_val<zeek::EnumVal>("Cluster::Backend::ZeroMQ::DROP");
+    static const auto& selected_policy =
```
bbannier, line 135: Nit: I feel I had brought this up before, but I'd drop the `static` here since we do not use the values outside of this function and `DoInitPostScript` shouldn't be hot code (I am also not sure I fully grasp what binding a `static` ref to a returned temporary does exactly).

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -114,10 +115,43 @@ void ZeroMQBackend::DoInitPostScript() {
...
+        zeek::reporter->InternalError("Unknown ZeroMQ overflow_policy: %s", obj_desc(selected_policy).c_str());
+
+    // Metrics for blocking and dropping.
+    total_xpub_blocks =
+        zeek::telemetry_mgr->CounterInstance("zeek", "cluster_zeromq_xpub_blocks", {},
+                                             "Number of times publishing events stalled due to XPUB socket HWM.");
+
+    total_xpub_drops = zeek::telemetry_mgr->CounterInstance("zeek", "cluster_zeromq_xpub_drops", {},
+
+                                                            "Number of events dropped due to XPUB socket HWM.");
```
bbannier, line 154: Any reason these couldn't just be in `ZeroMQBackend`'s ctor?

Also, stray line inside `total_xpub_drops`'s initializer.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -499,16 +533,28 @@ void ZeroMQBackend::Run() {
...
+                        //
+                        // If the overflow policy is set to "drop", increment the metric and
+                        // skip processing all parts of this message by setting skip_msg=true.
+                        //
                         if ( ! result ) {
-                            total_xpub_stalls->Inc();
+                            if ( overflow_policy == OverflowPolicy::Drop ) {
+                                total_xpub_drops->Inc();
+                                skip_msg = true;
+                                break; // Breaks the do-while and skip_msg=true above breaks the outer for. I'm sorry.
```
bbannier, line 545: Urgs, I feel this is a justifiable use case for a `goto` to a label after the outer loop.

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -499,16 +533,28 @@ void ZeroMQBackend::Run() {
...
+                            }
+
+                            // If the block policy is enabled, re-send the message without the dontwait
+                            // flag after increasing the xpub stall metric. This way, ZeroMQ will block
+                            // in xpub.send() until there's enough room available.
+                            assert(overflow_policy == OverflowPolicy::Block);
```
bbannier, line 551: You are handling all cases, and  already mapped the Zeek enum value into a C++ enum, let's just use a `switch` here.

If you do that, you'll probably also want the `goto` to replace the `break` in the exception handler below.

testing/btest/cluster/zeromq/overflow-policy-block.zeek
```diff
@@ -0,0 +1,155 @@
...
+	terminate();
+}
+
+event zeek_done() {
+	local blocks = get_zeromq_blocks();
+	if ( blocks > 0 )
+		print "GOOD: Observed XPUB blocks";
+	else
+		print "FAIL: No XPUB blocks";
+}
```
bbannier, line 154: The "GOOD" output isn't really interesting, but forces us to diff the logs of the proxy and workers. WDYT about dropping it and instead returning with an error for FAIL so BTest can handle the rest? In that case we wouldn't need to baseline the proxy and worker logs at all.

```suggestion
event zeek_done()
	{
	if ( get_zeromq_blocks() == 0 )
		{
		print "FAIL: No XPUB blocks";
		exit(1);
		}
	}
```

Nit: You don't seem to follow Zeek's preference for hanging indent anywhere here.

testing/btest/cluster/zeromq/overflow-policy-block.zeek
```diff
@@ -0,0 +1,155 @@
...
+# @TEST-EXEC: btest-diff worker-2/out
+
+# @TEST-START-FILE common.zeek
+@load ./zeromq-test-bootstrap
+
+global tick: event() &is_used;
+global finish: event(name: string) &is_used;
+global ping: event(sender: string, c: count) &is_used;
+
+# Lower high watermarks from 1000 (default) to something much lower to provoke blocking.
```
bbannier, line 35:
```suggestion
# Lower high watermarks from 1000 (default) to something much lower than `batch`to provoke blocking.
```

You could even go so far as directly computing them as some fraction of `batch` so you need to hardcode only e.g., `batches`, but derive all other values.

testing/btest/cluster/zeromq/overflow-policy-block.zeek
```diff
@@ -0,0 +1,155 @@
...
+	while ( i > 0 ) {
+		--i;
+		++publishes;
+		Cluster::publish(Cluster::manager_topic, ping, Cluster::node, publishes);
+
+		if ( publishes >= total_publishes )
+			return;
+	}
+
+	schedule 0.01msec { tick() };
```
bbannier, line 140: What's the `0.01msec` delay for? If it is not strictly needed I'd prefer one less magic constant.


```suggestion
	schedule 0secs { tick() };
```

testing/btest/cluster/zeromq/overflow-policy-block.zeek
```diff
@@ -0,0 +1,155 @@
...
+	print fmt("zeek_done last_c %s", last_c);
+
+	local blocks = get_zeromq_blocks();
+	if ( blocks == 0 )
+		print "GOOD: Observed no XPUB blocks on manager";
+	else
+		print "FAIL: XPUB blocks on manager";
+}
+# @TEST-END-FILE
+
```
bbannier, line 119:
```suggestion
```

testing/btest/cluster/zeromq/overflow-policy-block.zeek
```diff
@@ -0,0 +1,155 @@
...
+global finish: event(name: string) &is_used;
+global ping: event(sender: string, c: count) &is_used;
+
+# Lower high watermarks from 1000 (default) to something much lower to provoke blocking.
+redef Cluster::Backend::ZeroMQ::xpub_sndhwm = 20;
+redef Cluster::Backend::ZeroMQ::xsub_rcvhwm = 20;
+
+const total_publishes = 50000;
+
+function get_zeromq_blocks(): count {
```
bbannier, line 41: Nit.


```suggestion
function zeromq_blocks(): count {
```

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+# Lower high watermarks from 1000 (default) to something much lower to provoke drops.
+redef Cluster::Backend::ZeroMQ::xpub_sndhwm = 20;
+redef Cluster::Backend::ZeroMQ::xsub_rcvhwm = 20;
+
+# Drop'em!
+redef Cluster::Backend::ZeroMQ::overflow_policy = Cluster::Backend::ZeroMQ::DROP;
+
+const total_publishes = 50000;
+
+function get_zeromq_drops(): count {
```
bbannier, line 44: Nit.


```suggestion
function zeromq_drops(): count {
```

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+event zeek_init() {
+	for ( name, _ in Cluster::nodes ) {
+		if ( name == "manager" )
+			next;
+
+		last_c[name] = 0;
+		drop_c[name] = 0;
+	}
+}
+
```
bbannier, line 93:
```suggestion
```

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+	terminate();
+}
+
+event zeek_done() {
+	local drops = get_zeromq_drops();
+	if ( drops > 0 )
+		print "GOOD: Observed XPUB drops";
+	else
+		print "FAIL: No XPUB drops";
+}
```
bbannier, line 164: Same argument here, I'd prefer a simple process exit over us diffing always.

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+event zeek_done() {
+	for ( n, dropped in drop_c )
+		print n, "dropped?", dropped > 0;
+
+	local drops = get_zeromq_drops();
+	if ( drops == 0 )
+		print "GOOD: Observed no XPUB drops on manager";
+	else
+		print "FAIL: XPUB drops on manager";
+}
```
bbannier, line 124: Ditto.

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+# @TEST-END-FILE
+
```
bbannier, line 126:
```suggestion
```

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
```diff
@@ -0,0 +1,165 @@
...
+		# if we've published enough in order for the manager to
+		# detect we're done even if some events were dropped.
+		if ( publishes >= total_publishes )
+			break;
+	}
+
+	# Relax publishing if we published enough so the manager
+	# isn't totally overloaded.
+	local s = publishes < total_publishes ? 0.01msec : 0.01sec;
+	schedule s { tick() };
```
bbannier, line 151: Interesting how we seem to need a non-zero delay here, I wonder whether that is related to your comment on `in order for the manager to detect we are done`. The constants appear finicky to get right for all scenarios to me (overprovisioned CI runners?), is there some other way to do this? I'd still prefer to see a delay of `0secs` here.

awelzel: The test publishes `total_publishes` as fast as possible. That can probably be done with 0.0sec. But once these are published (some of which are likely to be dropped), we need to continue publishing so the manager can detect the other nodes have reached a finish state to initiate the teardown. I initially had this without increasing the delay, but the test would run significantly longer as even with a low hwm, enough events would be queued and processed still.

I'll try the 0.0 and then go up to 0.1sec and add a comment.

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
awelzel: I'm forth and back really :-)

I find non-baseline cluster tests harder to debug, because usually I end up putting prints left and right to understand the event flow and delays. I can certainly drop the baselines here, but if tests end up being flaky, at least I personally end up very quickly adding them again to understand what's going on, so I actually prefer to have the "infrastructure" available, even if it ends-up being a single line baseline saying "OK". But that might just be my workflow.

bbannier: Fair enough, but if we'd `exit` here on the `FAIL` path you'd get exactly the same logs ;)

awelzel: But yeah, for the non-manager nodes seems reasonable here. Thanks!

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: Yeah, that's a good call.

Let me also add a test attempting to actually lockup the cluster with the `BLOCK` policy. I've been pondering and having all workers publish to the worker_topic that they are all subscribed to might allow to deadlock them :worried:

testing/btest/cluster/zeromq/overflow-policy-block.zeek
awelzel: There's more tests that don't follow Allman style in the `cluster/` directory. I'll adapt the style for the new ones here, it's just that it's a bit spacey for me. And for tests I'm pretty liberal, but maybe shouldn't.

bbannier: I am not a personal fan of Zeek's style either, but we have a style and should follow it.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: > The thing that personally worries me most about BLOCK is that it cannot only halt the current node, but via zeromq backpressure any node working on a topic blocked anywhere.

Okay, I've tested and stared yesterday and I'm now disliking `BLOCK` a lot. I ended up writing comments about how that can result in lockups when selected and it just doesn't feel great.

I'm planning on re-spinning a v2 PR that only has a DROP policy, ditching BLOCK completely. I could see an ABORT policy that just calls `abort()` for testing. If a user doesn't want dropping at all (e.g. for perf testing or offline pcap processing), they could set the hwm and onloop queue sizes to 0 for unlimited buffering.

@bbannier - does that seem a reasonable direction to you?

testing/btest/cluster/zeromq/overflow-policy-block.zeek
awelzel: I've updated the style in the tests.

testing/btest/cluster/zeromq/overflow-policy-drop.zeek
awelzel: > Fair enough, but if we'd exit here on the FAIL path you'd get exactly the same logs ;)

I had the `exit(1)` for a while yesterday and it's not a great workflow for me for iterating or debugging a test.

A `TEST-EXEC: btest-diff <node?/out` can be leveraged for `print` debugging. Running with `btest -d`, the diff for the baseline usually directly shows the added debug output. That's sort-of my workflow: Add a print, check the baseline diff, add more prints or change some logic, check the baseline diff again. Sometimes adding more lines to the baseline. Having the `exit(1)` breaks this because then `btest-bg-wait` (or a `TEST-EXEC` line) fails and the `btest-diff` command never runs: Now I have to go to some `.tmp/<test>/<node>/out` file rather than seeing the output directly.

This is pretty much in the weeds how I work locally, but I'm going to be a bit more baseliney for cluster tests, even if the baseline is just an "ok". Hope that makes sense.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
bbannier: Thanks for checking! Looking back `BLOCK` seems like an attempt to implement cluster-wide lockstep which is hard to implement or use correctly, and likely not worth it at all. What users might want is "never ever throw away my precious events", and unlimited buffering solves that.

I wouldn't bother with an `ABORT` policy for testing. Let's instead make sure we have logs or metrics to observe this in testing; whatever we need to add here would also be useful for production users.

awelzel:
Replaced by #4681

PR #4670, opened 2025-07-17
Update libkqueue submodule [nomail]

timwoj:
This pulls in @ckreibich's fix for the allocating too many file descriptors in Docker environments and blowing up Zeek's memory usage.


timwoj:
I'm semi-curious to see what the benchmarker thinks about this, to be honest. It should be be allocating a bit less memory in certain situations, but I don't know if that will show up in the benchmarker passes.

timwoj:
> I'm semi-curious to see what the benchmarker thinks about this, to be honest. It should be be allocating a bit less memory in certain situations, but I don't know if that will show up in the benchmarker passes.

Eh, nothing significant.

PR #4671, opened 2025-07-17
control: Use Cluster::publish() for replying

awelzel:
Switching to ZeroMQ as cluster backend and dabbling with zeekctl and WebSocket, replies didn't arrive due to the usage of Broker::publish() rather than Cluster::publish(). Additionally, add the node name to the topic on which we reply so that the receiver can figure out which node sent the reply. It could've been a separate event parameter, but the topic appears just fine.

This is needed for zeek/zeekctl#84.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -130,6 +130,25 @@ New Functionality
   implementation in the ``src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple``
   directory for an example.
 
+- Added support to ZeekControl for seamlessly switching to ZeroMQ as cluster
+  backend by adding the following settings to zeekctl.cfg:
+
+	ClusterBackend = ZeroMQ
+	UseWebSocket = 1
+
+  With the ZeroMQ cluster backend, Zeekctl requires to use Zeek's WebSocket API
+  to communicate with individual nodes for the ``print`` and ``netstats`` commands.
+  Setting the ``UseWebSocket`` option enables a WebSocket server on the manager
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -130,6 +130,24 @@ New Functionality
   implementation in the ``src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple``
   directory for an example.
 
+- Added support to ZeekControl for seamlessly switching to ZeroMQ as cluster
+  backend by adding the following settings to zeekctl.cfg:
+
+	ClusterBackend = ZeroMQ
+	UseWebSocket = 1
+
+  With the ZeroMQ cluster backend, Zeekctl will Zeek's WebSocket API to communicate
```
timwoj:
```suggestion
  With the ZeroMQ cluster backend, Zeekctl will use Zeek's WebSocket API to communicate
```

NEWS
```diff
@@ -130,6 +130,24 @@ New Functionality
...
+  with individual nodes for the ``print`` and ``netstats`` commands. Setting the
+  UseWebSocket option enables a WebSocket server on the manager node, listening on
```
timwoj:
```suggestion
  ``UseWebSocket`` option enables a WebSocket server on the manager node, listening on
```

timwoj: This one in applies in general to these option names below.

timwoj (approved):
Just some nits on the NEWS entry. I'll look at the Zeekctl part in a minute.

PR #4673, opened 2025-07-18
Switch to using c++20 constraints instead of std::enable_if

timwoj:
Some simple code modernization now that we have C++20.


timwoj:
Collected into https://github.com/zeek/zeek/pull/4678

PR #4676, opened 2025-07-20
Fix clang-tidy findings in embedded C++ from bif files

timwoj:
I'm not sure why these weren't triggering before, but this fixes clang-tidy findings in the embedded C++ code in various BIF files. It also disables the `-bugprone-suspicious-include` checker which was flagging the places we include BIF `.cc` files into other files.


src/stats.bif
```diff
@@ -89,21 +89,18 @@ function get_conn_stats%(%): ConnStats
...
-	ADD_STAT(s.num_TCP_conns);
-	ADD_STAT(s.max_TCP_conns);
-	ADD_STAT(s.cumulative_TCP_conns);
-	ADD_STAT(s.num_UDP_conns);
-	ADD_STAT(s.max_UDP_conns);
-	ADD_STAT(s.cumulative_UDP_conns);
-	ADD_STAT(s.num_ICMP_conns);
-	ADD_STAT(s.max_ICMP_conns);
-	ADD_STAT(s.cumulative_ICMP_conns);
+	r->Assign(n++, session_mgr ? static_cast<uint64_t>(s.num_packets) : 0L);
```
awelzel: In line 86 above, `session_mgr` is accessed unguarded, so think we can just drop all these checking and access `s.num_packets` directly.

Also the GetStats() call above is probably okay to do unconditionally?

.clang-tidy
```diff
@@ -37,6 +37,7 @@ Checks: [-*,
          -bugprone-pointer-arithmetic-on-polymorphic-object,
          -bugprone-empty-catch,
          -bugprone-exception-escape,
+         -bugprone-suspicious-include,
```
awelzel, line 40: Is this just for the following? Could we NOLINT them instead?
```
src/zeek.bif:#include "__all__.bif.cc"
src/zeek.bif:#include "__all__.bif.register.cc"
src/zeek.bif:   #include "__all__.bif.init.cc"
```

src/strings.bif
```diff
@@ -741,9 +741,9 @@ function string_to_ascii_hex%(s: string%): string
 	const u_char* sp = s->Bytes();
 
 	for ( int i = 0; i < s->Len(); ++i )
-		snprintf(x + i * 2, 3, "%02x", sp[i]);
+		snprintf(x + static_cast<ptrdiff_t>(i * 2), 3, "%02x", sp[i]);
```
awelzel: Haven't tried, but maybe this does it, too? `util::bytetohex(sp[i], &x[i * 2])`,   `snprintf(&x[i * 2], ...)` - not sure it's easier to read, but shorter :-)

src/strings.bif
```diff
@@ -1527,6 +1526,7 @@ function rjust%(str: string, width: count, fill: string &default=" "%) : string
 function swap_case%(str: string%) : string
 	%{
 	string s = str->ToStdString();
+	// NOLINTNEXTLINE(modernize-loop-convert)
```
awelzel: Is this modern enough?
```
--- a/src/strings.bif
+++ b/src/strings.bif
@@ -1527,12 +1527,12 @@ function rjust%(str: string, width: count, fill: string &default=" "%) : string
 function swap_case%(str: string%) : string
        %{
        string s = str->ToStdString();
-       for ( size_t i = 0; i < s.size(); i++ )
+       for ( auto& c : s )
                {
-               if ( std::islower(s[i]) )
-                       s[i] = toupper(s[i]);
-               else if ( std::isupper(s[i]) )
-                       s[i] = tolower(s[i]);
+               if ( std::islower(c) )
+                       c = toupper(c);
+               else if ( std::isupper(c) )
+                       c = tolower(c);
                }
```

awelzel (approved):

src/strings.bif
bbannier: None of this is safe if the character is not ASCII. All the values here are `char` on my platform while all these C functions expect `unsigned char`. IIRC passing them anything which is not equal `EOF` and doesn't fit into `unsigned char` triggers UB.

awelzel: Fair, but neither was the old version? This was mostly about the loop modernization.

Coincidentally, we do compile everything with `-funsigned-char`, so `char` is definitely unsigned within Zeek, but that's more a lucky chance for this one I guess?

bbannier: That it was here before doesn't mean it was good.

src/stats.bif
timwoj: I moved the check for session_mgr to wrap all of the calls to it there, and added an `else` that just increases the index.

.clang-tidy
timwoj: Just wrapping that section covers those specific includes, but the includes inside those files still get reported. I would think NOLINTBEGIN would be recursive, but it's not. Wrapping the individual includes in the `__all__` file requires updates to files in the CMake repo plus adding an additional function to add the NOLINTEND when we're done parsing all of the bifs.

src/strings.bif
timwoj: This isn't the only place we call these methods either. I'll make a note to fix them all at once.

timwoj: > Is this modern enough?

I debated whether we were ok with iterating over the characters in a string with a ranged-for. I don't know if we do that anywhere else.

src/strings.bif
timwoj: Changing to `snprintf(&x[i * 2], ...)` results in the same `bugprone-implicit-widening-of-multiplication-result` finding I was originally fixing. It doesn't like to use `int` as the index into an array, basically. Changing the loop index to a `size_t` should be sufficient to fix it.

PR #4677, opened 2025-07-21
Enable modernize-std-numbers clang-tidy checker, fix findings

timwoj:


timwoj:
Collected into https://github.com/zeek/zeek/pull/4678

PR #4678, opened 2025-07-21
C++20 Modernization

timwoj:
This does the following modernization, using C++20 features:

- Remove std::enable_if, replace with constraints
- Use std::ranges algorithms
- Use std::numbers for a few constants.

This also enables the related clang-tidy checkers.


timwoj:
I'm inclined to say leave this PR until after 8.0 is branched. It's a lot of changes to stick in right before that branch happens.

src/Func.h
```diff
@@ -71,8 +71,6 @@ class Func : public Obj {
         // If any of the groups are disabled, this body is disabled.
```
awelzel, line 72: Drive-by: Mind moving the `set` one up to put `disabled` into the padding behind priority to reduce the size from 72 to 64 bytes?  And also, never realized a `std::set` is so huge (48 bytes here). In this case a `std::forward_list` seems okay as we never modify it and iterating over the groups happens rarely.

```
--- a/src/Func.cc
+++ b/src/Func.cc
@@ -564,7 +564,7 @@ void ScriptFunc::AddBody(StmtPtr new_body, const std::vector<IDPtr>& new_inits,
 
     Body b;
     b.stmts = new_body;
-    b.groups = groups;
+    b.groups = {groups.begin(), groups.end()};
     current_body = new_body;
     current_priority = b.priority = priority;
 
diff --git a/src/Func.h b/src/Func.h
index 3b47057d98..f8eb1fb265 100644
--- a/src/Func.h
+++ b/src/Func.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <forward_list>
 #include <memory>
 #include <string>
 #include <tuple>
@@ -66,8 +67,8 @@ public:
 
     struct Body {
         detail::StmtPtr stmts;
+        std::forward_list<EventGroupPtr> groups;
         int priority;
-        std::set<EventGroupPtr> groups;
         // If any of the groups are disabled, this body is disabled.
         // The disabled field is updated from EventGroup instances.
         bool disabled = false;
```

This gets `struct Body` down to 24 bytes... from 72 :cold_sweat:

src/ZeekString.cc
```diff
@@ -296,26 +294,29 @@ String::Vec* String::Split(const String::IdxVec& indices) const {
     idx.insert(idx.end(), indices.begin(), indices.end());
 
     // Sanity checks.
-    for ( i = 0; i < idx.size(); ++i )
-        if ( idx[i] >= n || idx[i] < 0 )
-            idx[i] = 0;
+    std::ranges::transform(idx.begin(), idx.end(), idx.begin(), [this](int v) {
```
awelzel, line 297: I was wondering if you could skip the `idx.insert(idx.end(), indices.begin(), indices.end());` above and just do:
```
    std::ranges::transform(indices.begin(), indices.end(), std::next(idx.begin()), [this](int v) {
```

src/ZeekString.cc
```diff
@@ -296,26 +294,29 @@ String::Vec* String::Split(const String::IdxVec& indices) const {
...
+        return v;
+    });
 
     // Sort it:
-    sort(idx.begin(), idx.end());
+    std::ranges::sort(idx);
 
     // Shuffle vector so duplicate entries are used only once:
-    IdxVecIt end = unique(idx.begin(), idx.end());
+    auto [first, last] = std::ranges::unique(idx);
```
awelzel:
```suggestion
    auto [ret, last] = std::ranges::unique(idx);
```
That seems to be in cppreference and might lighten the `first` explanation below a bit.

src/ZeekString.cc
```diff
@@ -296,26 +294,29 @@ String::Vec* String::Split(const String::IdxVec& indices) const {
...
 
     // Each element in idx is now the start index of a new
     // substring, and we know that all indices are within [0, n].
     //
     Vec* result = new Vec();
```
awelzel, line 313:
```suggestion
    Vec* result = new Vec();
    result->reserve(std::distance(idx.begin(), ret));`
```
Does that work? :-)

src/analyzer/protocol/ssl/ssl-analyzer.pac
```diff
@@ -78,9 +78,9 @@ refine connection SSL_Conn += {
 			vector<int>* ciphers = new vector<int>();
```
awelzel: Fly-by: Depending a bit on how many ciphers there are in this vectors, seems adding a `reserve()` on `ciphers` would avoid repeated reallocations during the `copy()`? E.g. if `ciphers` has 5 elements,  the vector is resized here to sizes 1, 2, 4, and 8. Involving some 4 mallocs and 3 copies.

Also for other vectors or the VectorVals in this file.

timwoj:
> If you have like a bunch more of these in mind for post 8.0, then the backport / conflict argument doesn't matter and we can delay :-)

I wouldn't say I have a bunch. This just hits the easy ones that clang-tidy could do mostly for me, but there's more to come. Dominik's `std::span` PR is one, designated initializers are another, and probably more that I'm forgetting about off the top of my head.

src/Func.h
timwoj: It's amusing how many places we do this.

src/ZeekString.cc
timwoj: This one required entirely too much shuffling to figure out exactly how it worked in comparison to the old `std::unique`.

PR #4680, opened 2025-07-21
Coverity cleanup

timwoj:
This fixes a large number of Coverity findings, trying to get down to a limited number of open findings on there.

### DIFF
--- a/src/OpaqueVal.cc
+++ b/src/OpaqueVal.cc
@@ -759,7 +759,9 @@ ValPtr BloomFilterVal::DoClone(CloneState* state) {
     if ( bloom_filter ) {
         auto bf = make_intrusive<BloomFilterVal>(bloom_filter->Clone());
         assert(type);
-        bf->Typify(type);
+        if ( ! bf->Typify(type) )
+            reporter->InternalError("Failed to typify new bloom_filter clone, clone already had valid type");
+
         return state->NewClone(this, std::move(bf));
     }
 
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -751,7 +751,7 @@ IPAddr SubNetVal::Mask() const {
     for ( w = subnet_val->Length(); w >= 32; w -= 32 )
         *(mp++) = 0xffffffff;
 
-    *mp = ~((1 << (32 - w)) - 1);
+    *mp = ~((static_cast<uint32_t>(1) << (32 - w)) - 1);
 
     while ( ++mp < m + 4 )
         *mp = 0;
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -2398,7 +2398,7 @@ ValPtr TableVal::Remove(const Val& index, bool broker_forward, bool* iterators_i
     if ( broker_forward && ! broker_store.empty() )
         SendToStore(&index, nullptr, ELEMENT_REMOVED);
 
-    if ( change_func ) {
+    if ( change_func && k ) {
         // this is totally cheating around the fact that we need a Intrusive pointer.
         ValPtr changefunc_val = RecreateIndex(*(k.get()));
         CallChangeFunc(changefunc_val, va, ELEMENT_REMOVED);
--- a/src/analyzer/protocol/irc/IRC.cc
+++ b/src/analyzer/protocol/irc/IRC.cc
@@ -807,8 +807,8 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
     }
 
     else if ( irc_quit_message && command == "QUIT" ) {
-        string message = params;
-        if ( message[0] == ':' )
+        string_view message{params};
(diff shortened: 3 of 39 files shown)


src/main.cc
```diff
@@ -52,7 +52,14 @@ int main(int argc, char** argv) {
...
 
     auto time_start = zeek::util::current_time(true);
-    auto setup_result = zeek::detail::setup(argc, argv);
+    zeek::detail::SetupResult setup_result;
+
+    try {
+        setup_result = zeek::detail::setup(argc, argv);
+    } catch ( const zeek::InterpreterException& e ) {
+        fprintf(stderr, "Exception caught during initial setup\n");
+        exit(-1);
```
awelzel: Suggest exit(1) and wonder if abort() might even make sense to get a coredump, though not sure it's possible to get to the throwing location easily.

src/analyzer/protocol/irc/IRC.cc
```diff
@@ -823,7 +823,7 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
     }
 
     else if ( irc_nick_message && command == "NICK" ) {
-        string nick = params;
+        string& nick = params;
         if ( nick[0] == ':' )
             nick = nick.substr(1);
```
awelzel, line 828: So I stared at this one and it feels very confusing and also dangerous. This is replacing the contents of the original `params` string with `nick.substr(1)` if `nick` starts with a `:`. Maybe coverity understands that this is safe in this function today, but if anyone wants to do something with the original `params`, this seems like it could cause head-scratching confusion IMO.

Maybe switch nick `to std::string_view` and construct the StringVal with that?

Same pattern elsewhere in this file.

src/input/readers/ascii/Ascii.cc
```diff
@@ -175,7 +175,7 @@ bool Ascii::ReadHeader(bool useCached) {
         line = headerline;
 
     // construct list of field names.
-    auto ifields = util::split(line, separator[0]);
+    auto ifields = util::split(std::move(line), separator[0]);
```
awelzel, line 178: Random: There's so much substr() happening in `util::split()` on the input line (unless there's nothing to split) that I'm wondering if `util::split()` could take the line by const& ref rather than value.

But nothing to do for this PR.

src/storage/backend/sqlite/SQLite.cc
```diff
@@ -496,7 +496,7 @@ void SQLite::DoExpire(double current_network_time) {
     // Automatically rollback the transaction when this object is deleted.
     auto deferred_rollback = util::Deferred([this]() {
         char* errMsg = nullptr;
-        sqlite3_exec(expire_db, "rollback transaction", nullptr, nullptr, &errMsg);
+        [[maybe_unused]] int ret = sqlite3_exec(expire_db, "rollback transaction", nullptr, nullptr, &errMsg);
```
awelzel: Hmm, elsewhere there's: `sqlite3_exec(expire_db, "commit transaction", nullptr, nullptr, &errMsg);`

Maybe just ignore the return value  here, too? But would think that we should actually handle and check the result and log a message if something is off.

src/OpaqueVal.cc
```diff
@@ -759,7 +759,7 @@ ValPtr BloomFilterVal::DoClone(CloneState* state) {
     if ( bloom_filter ) {
         auto bf = make_intrusive<BloomFilterVal>(bloom_filter->Clone());
         assert(type);
-        bf->Typify(type);
+        [[maybe_unused]] auto n = bf->Typify(type);
```
awelzel: Hmm, can we just not store anything? It's clearly unused :-)

src/Val.cc
```diff
@@ -2398,7 +2398,7 @@ ValPtr TableVal::Remove(const Val& index, bool broker_forward, bool* iterators_i
     if ( broker_forward && ! broker_store.empty() )
         SendToStore(&index, nullptr, ELEMENT_REMOVED);
 
-    if ( change_func ) {
+    if ( change_func && k ) {
         // this is totally cheating around the fact that we need a Intrusive pointer.
         ValPtr changefunc_val = RecreateIndex(*(k.get()));
```
awelzel, line 2403: Drive-by: If `Val::Clone()` was const we could probably just do `index.Clone()` here, but nothing to do.

awelzel (changes requested):

awelzel (approved):
Approving, actually. The ones in IRC.cc would be good to revert or change to something less tricky IMO.

src/OpaqueVal.cc
timwoj: The finding was that we check the result of `Typify` in most (all?) of the other places it's used. A [[maybe_unused]] makes Coverity ignore it. We didn't care about the result here, but maybe we should?

PR #4682, opened 2025-07-22
Type/RecordType: Make table returned by GetRecordFieldsVal() ordered

awelzel:
Seems only reasonable to provide that guarantee as pointed out in #4674.


0xxon:
Sounds great to me. Do you perhaps want to add the fact that the returned table is ordered to the function documentation in Type.h and the record_fields bif documentation? :)

awelzel:
Yeah, good call. Also fixed the `spicy/` tests that I usually just `ctrl+c` out of after a spicy bump due to the time they need :-)

0xxon (approved):

PR #4685, opened 2025-07-22
Stop hardcoding the agent->controller connection address

dopheide-esnet:
The current implementation ignores if the user redefs Management::Controller::listen_address.  Unfortunately, we can't just re-use Management::Controller::network_info() because that gives us a default listen address of 0.0.0.0 when the agent needs a default address to connect to.

Without this, even if you specify the listen address (or aren't even using it because you're doing websockets), you'll get errors like the following:

`error in /usr/local/zeek/versions/f4357485d2006813fba768aa83e9e1bf8e8bb236/share/zeek/base/frameworks/broker/log.zeek, line 95: Broker error (Broker::PEER_UNAVAILABLE): (00000000-0000-0000-0000-000000000000, *127.0.0.1:2150, "unable to connect to remote peer")`

### DIFF
--- a/scripts/policy/frameworks/management/agent/config.zeek
+++ b/scripts/policy/frameworks/management/agent/config.zeek
@@ -76,9 +76,9 @@ export {
 	## like to use that mode, make sure to set
 	## :zeek:see:`Management::Agent::listen_address` and
 	## :zeek:see:`Management::Agent::listen_port` as needed.
-	const controller = Broker::NetworkInfo(
-	    $address=Management::Controller::network_info()$address,
-	    $bound_port=Management::Controller::network_info()$bound_port) &redef;
+        const controller = Broker::NetworkInfo(
+            $address=Management::Controller::connect_network_info()$address,
+            $bound_port=Management::Controller::connect_network_info()$bound_port) &redef;
 
 	## An optional working directory for the agent. Agent and controller
 	## currently only log locally, not via the Zeek cluster's logger
--- a/scripts/policy/frameworks/management/agent/config.zeek
+++ b/scripts/policy/frameworks/management/agent/config.zeek
@@ -76,7 +76,8 @@ export {
 	## like to use that mode, make sure to set
 	## :zeek:see:`Management::Agent::listen_address` and
 	## :zeek:see:`Management::Agent::listen_port` as needed.
-	const controller = Broker::NetworkInfo($address="127.0.0.1",
+	const controller = Broker::NetworkInfo(
+	    $address=Management::Controller::network_info()$address,
 	    $bound_port=Management::Controller::network_info()$bound_port) &redef;
 
 	## An optional working directory for the agent. Agent and controller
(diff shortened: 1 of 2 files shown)


ckreibich (changes requested):
Sorry for the delay Dop. Sure thing, can I just make a small change request? These info functions are the same except for the fallback address. So I'm thinking if instead we make the fallback configurable, we can keep a single function. So that'd be
```
function network_info(fallback_address: string &default="0.0.0.0"): Broker::NetworkInfo
        {
        local ni: Broker::NetworkInfo;

        if ( Management::Controller::listen_address != "" )
                ni$address = Management::Controller::listen_address;
        else if ( Management::default_address != "" )
                ni$address = Management::default_address;
        else
                ni$address = fallback_address;
```
and you'd use `network_info("127.0.0.1")`. And we do the same to `network_info_websocket()`, for consistency. What do you think?

I can do this if you're busy, just let me know.

ckreibich:
Let me take this on Dop — not a problem.

dopheide-esnet:
Hhmm, I thought I did, maybe I didn’t push it?  Either way, thanks, I’m
sure your version will be a bit more precise.

ckreibich:
I'm back to looking at this, Dop. I think my idea doesn't actually work, due to the difference between needing info for where to listen vs where to connect to. More soon.

ckreibich:
Okay, a few thoughts. The gist really is how we should configure the address of the controller that agents connect to, especially in a multi-host setting. For context: the current setup defaults to a single-host deployment (one agent, plus controller, on the same machine). The controller listens globally, since there's an assumption that one might still want to control such a cluster from another machine. The tests in the zeek-testing-cluster testsuite use that model. That's also why the agent's configuration defaults to 127.0.0.1 for the controller's address: it's what will work out of the box in a single-host deployment. I think we have three options:
* Change nothing. Say `redef Management::Agent::controller = Broker::NetworkInfo(<public IP>, <port>)` to tell the agent(s) where to connect to. (I should add a testcase for this, since there currently isn't one.)
* Move this part out of `Management::Agent` and into the controller's config, which is arguably more intuitive. The question there is whether the existing `Management::Controller::listen_address` should double as the address that agents connect to (which it currently doesn't), or ...
* We introduce additional variables, like `Management::Controller::connect_address` (which would default to 127.0.0.1). I don't think we'd need analogs to `network_info()`, since the agent never uses address and port as one value (a `Broker::NetworkInfo`). 
 
The third is probably conceptually cleanest, and the direction in which your patch was already going, Dop. Fwiw we landed on a similar that distinction (i.e. a listening config and a separate connection config) in the ZeroMQ cluster backend.

At this point, a question for you @dopheide-esnet : is the first option okay for you right now? If it is, then I'd rather start working on making the framework run on top of the ZeroMQ backend (which will simplify some of the current code, but also touch a fair bit of it), and not make further changes assuming Broker. It'll need to be in that state by 8.1 anyway.

dopheide-esnet:
Yeah, changing nothing for now works for me.  I suspect this is mostly a problem for folks just getting started with a multi-node cluster and not knowing why it's not connecting.  (I probably ran into it without realizing it before switching to websockets.)  Maybe it's worth adding a note in the documentation.

ckreibich:
Cool, thanks Dop. Yeah you were trailblazing — multi-node indeed isn't fully supported at the moment. I'll get going on the ZeroMQ-ification, and I'll check in with you once that's settling.

PR #4686, opened 2025-07-22
Add flag to storage options to force synchronous mode

### LINKED ISSUE #4625: Add option to enforce synchronous operation of Storage backends
Currently, the asynchronous API will operate synchronous if Zeek is reading a PCAP. It would be great to have an additional flag that allows to enforce this behavior. It could help with some testing scenarios that rely on reading from an interface.

timwoj:
This adds a flag to the `Storage::BackendOptions` record to force a backend into synchronous mode. This causes all of the operations run synchronously, even if the `Storage::Async` methods are called from script-land.

It also fixes a mixup of a couple of the storage metric names.

Fixes #4625


scripts/base/frameworks/storage/main.zeek
```diff
@@ -10,6 +10,11 @@ export {
 	type BackendOptions: record {
 		## The serializer used for converting Zeek data.
 		serializer: Storage::Serializer &default=Storage::STORAGE_SERIALIZER_JSON;
+
+		## Sets the backend into forced-synchronous mode. All operations will run
+		## in synchronous mode, even if the async functions are called.  This
+		## should generally only be set to ``T`` during testing.
+		forced_synchronous : bool &default=F;
```
awelzel:
```suggestion
		forced_sync : bool &default=default_forced_sync;
```

Mostly so the naming is "in sync" with the `Storage::Sync` module and the `is_forced_sync()` bif and also a bit separate from the "synchronous" pragma that SQLite has which is quite unrelated. The default to allow switching all backends used in any script into `forced_sync` mode by doing `redef Storage::default_forced_sync = T`, without needing to change each individual script that opens a storage backend. Think that's something @J-Gras  would do.

The latter makes me actually wonder if it should even be part of `BackendOptions`, but I guess that provides some flexibility.

src/storage/Backend.cc
```diff
@@ -127,6 +127,8 @@ OperationResult Backend::Open(OpenResultCallback* cb, RecordValPtr options, Type
     val_type = std::move(vt);
     backend_options = options;
 
+    forced_synchronous = options->GetField<BoolVal>("forced_synchronous")->Get();
```
awelzel:
```suggestion
    forced_sync = options->GetField<BoolVal>("forced_sync")->Get();
```

src/storage/Backend.h
```diff
@@ -216,6 +216,11 @@ class Backend : public zeek::Obj {
      */
     const RecordValPtr& Options() const { return backend_options; }
 
+    /**
+     * Returns the state of the forced_synchronous option that was passed in the options record to Open().
+     */
+    bool IsForcedSynchronous() const { return forced_synchronous; }
```
awelzel:
```suggestion
    bool IsForcedSync() const { return forced_sync; }
```

testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
```diff
@@ -0,0 +1,91 @@
...
+			{
+			print "put request timed out";
+			terminate();
+			}
+		}
+	timeout 5sec
+		{
+		print "open request timed out";
+		terminate();
+		}
```
awelzel, line 94: If you put a `print` here (and elsewhere) after the `when` statements, these should now be printed only once all the `when` bodies ran? That might make sense to have in the baseline.

testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
```diff
@@ -0,0 +1,91 @@
...
+			{
+			print "put result", put_res;
+
+			when [key, value] ( local get_res = Storage::Async::get(b, key) )
+				{
+				print "get result", get_res;
+				if ( get_res$code == Storage::SUCCESS && get_res?$value )
+					print "get result same as inserted", value == ( get_res$value as string );
+
+				schedule 100 msec { print_metrics_and_close() };
```
awelzel:
```suggestion
				schedule 0sec { print_metrics_and_close() };
```

Should this or event `event print_metrics_and_close();` work ?

testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
awelzel:
```
--- a/testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
+++ b/testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
@@ -69,23 +69,29 @@ event zeek_init()
                                if ( get_res$code == Storage::SUCCESS && get_res?$value )
                                        print "get result same as inserted", value == ( get_res$value as string );
 
-                               schedule 100 msec { print_metrics_and_close() };
+                               schedule 0sec{ print_metrics_and_close() };
                                }
                        timeout 5sec
                                {
                                print "get request timed out";
                                terminate();
                                }
+
+                       print "inner get done";
                        }
                timeout 5sec
                        {
                        print "put request timed out";
                        terminate();
                        }
+
+               print "inner put done";
                }
        timeout 5sec
                {
                print "open request timed out";
                terminate();
                }
+
+       print "outer open done";

```

New baseline:
```
   put result, [code=Storage::SUCCESS, error_str=<uninitialized>, value=<uninitialized>]
   get result, [code=Storage::SUCCESS, error_str=<uninitialized>, value=value5678]
   get result same as inserted, T
  +inner get done
  +inner put done
  +outer open done
```

Without the forced_sync flag, the baseline is as follows - code after each when is executed, so IMO seems useful to have these prints to verify execution.
```
  --- /dev/fd/63        2025-07-23 08:52:38.271112131 +0000
  +++ /dev/fd/62        2025-07-23 08:52:38.271112131 +0000
  @@ -1,5 +1,8 @@
  +outer open done
   open result, [code=Storage::SUCCESS, error_str=<uninitialized>, value=<opaque of BackendHandleVal>]
  +inner put done
   put result, [code=Storage::SUCCESS, error_str=<uninitialized>, value=<uninitialized>]
  +inner get done
   get result, [code=Storage::SUCCESS, error_str=<uninitialized>, value=value5678]
   get result same as inserted, T
```

awelzel (approved):

testing/btest/scripts/base/frameworks/storage/redis-forced-sync.zeek
timwoj: `event` should be fine. This was just copied and pasted from `redis-async.zeek`. It could probably be the same over there, for that matter, since it runs inside the result of the `get`.

PR #4687, opened 2025-07-22
CI: Add weekly task for running builds with newest compilers

### LINKED ISSUE #4537: Adding weekly CI task with latest compiler(s)
From a discussion within the team, it'd be nice to compile with the latest released clang and gcc releases without claiming support.

A weekly or bi-weekly Cirrus cron job might be sufficient here.

timwoj:
This adds two new CI tasks to run builds on opensuse-tumbleweed with the most current versions of Clang and GCC available on that distro. I'll add a weekly cron job to Cirrus to run these once they're stable in this PR.

Fixes #4537


timwoj:
Ignoring clang failing to build here, the actual setup of installing the newer compilers and the build process finding them seems to be working. I'm marking as ready for review. I opened https://github.com/zeek/broker/issues/481 for the Clang failure.

Issue #4202, opened 2025-01-30
Move `Conn::set_conn()` from `connection_state_remove` to `new_connection` event


ckreibich:
We regularly have reasons to access `c$conn` from a range of connection events, but the fact that for the common case we only establish the record [in `connection_state_remove()`](https://github.com/zeek/zeek/blob/master/scripts/base/protocols/conn/main.zeek#L313-L316) means that other event handlers need to remember to invoke `Conn::set_conn()` to ensure that record is present. For example we do this [here](https://github.com/zeek/zeek/blob/master/scripts/policy/protocols/conn/community-id-logging.zeek#L24) and [here](https://github.com/zeek/zeek/blob/master/scripts/policy/protocols/conn/ip-proto-name-logging.zeek#L16), or as per a recent [Slack conversation](https://zeekorg.slack.com/archives/CSZBXF6TH/p1738244825538259).

Moving the establishment to `new_connection` seems better — there's still a dependence on `&priority` for other such handlers, but most uses can then assume the record is present.

@awelzel, @JustinAzoff, fyi.

PR #4683, opened 2025-07-22
PPPoE: add session id logging

0xxon:
This adds a new PacketAnalyzer::PPPoE::session_id bif, which extracts the PPPoE session ID from the current packet.

Furthermore, a new policy script is added which adds the pppoe session id to the connection log.

Related to GH-4602

PR #4688, opened 2025-07-23
Move Conn::set_conn to new_connection

0xxon:
Thso moves Conn::set_conn to new_connections. The two commits are best reviewed in order, with the second being a slight optimization to the first.

Background in GH-4202


vpax:
Is it confirmed that this doesn't have any appreciable performance impact? It's a new event handler invocation for every connection ...

awelzel:
This branch in bare mode and only base/protocols/conn loaded:
```
$ hyperfine -w 1 -r 5 -L zeek zeek 'taskset -c 2,3 {zeek} -b  -C -r /media/awelzel/T7/pcaps/200k-syn-packets.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE'
Benchmark 1: taskset -c 2,3 zeek -b  -C -r /media/awelzel/T7/pcaps/200k-syn-packets.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE
  Time (mean ± σ):      7.301 s ±  0.068 s    [User: 7.099 s, System: 1.401 s]
  Range (min … max):    7.245 s …  7.419 s    5 runs

```

Base:
```
$ hyperfine -w 1 -r 5 -L zeek /opt/zeek-dev-prod/bin/zeek 'taskset -c 2,3 {zeek} -b -C -r /media/awelzel/T7/pcaps/200k-syn-packets.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE'
Benchmark 1: taskset -c 2,3 /opt/zeek-dev-prod/bin/zeek -b -C -r /media/awelzel/T7/pcaps/200k-syn-packets.pcap base/protocols/conn Log::default_writer=Log::WRITER_NONE
  Time (mean ± σ):      6.423 s ±  0.054 s    [User: 6.298 s, System: 1.344 s]
  Range (min … max):    6.364 s …  6.502 s    5 runs
```

In non-bare mode I'm only measuring 8.6sec vs 8.3sec and the `zeek-benchmarker` numbers don't show anything either (running 500k syn packets in non-bare, but probably noisy EC2 environment). It seems too much :-(

0xxon:
I did some tests on my system in non-bare-mode. Using a 500k-syn trace (which seems like the absolute workst case), results in a kind of stable 4-ish% or a bit less speed difference. Which does not feel great, but also is an upper bound for the impact.

This also matches the 8.6 vs 8.3 secs of @awelzel.

In cases where people do handle the `new_connection` event, this should make things faster. That seems quite common nowadays - the community-id package is one popular example where this is the case.

I won't strongly fight for this - but I think this might be justifyable.

ckreibich:
I just hit another reason over here in a package to add a manual `new_connection()` to pull this forward, so I'm in favor of doing this at this level of impact.

scripts/policy/protocols/conn/community-id-logging.zeek
```diff
@@ -21,6 +21,5 @@ module Conn;
 
 event new_connection(c: connection) &priority=5
```
awelzel, line 20: Possible to remove the above `module Conn` now?

scripts/policy/protocols/conn/failed-service-logging.zeek
```diff
@@ -30,8 +30,6 @@ event analyzer_failed(ts: time, atype: AllAnalyzers::Tag, info: AnalyzerViolatio
 	if ( analyzer_name !in c$service || analyzer_name in c$failed_analyzers )
 		return;
 
-	set_conn(c, F);
```
awelzel, line 33: Is dropping `module Conn` possible?

scripts/policy/protocols/conn/ip-proto-name-logging.zeek
```diff
@@ -12,8 +12,6 @@ redef record Info += {
 };
 
 event new_connection(c: connection) &priority=5 {
-	# In case we're the first access
-	Conn::set_conn(c, F);
```
awelzel, line 16: Same about the `module Conn` above :-)

vpax:
> I did some tests on my system in non-bare-mode. Using a 500k-syn trace (which seems like the absolute workst case), results in a kind of stable 4-ish% or a bit less speed difference. Which does not feel great, but also is an upper bound for the impact.

What do you see with ZAM? It might be quite a bit lower, since there's already a `new_connection` handler for the `tunnels` framework, and ZAM should be able to coalesce this handler with that one. (Be sure to factor out the initial ZAM compilation time.)

0xxon:
How does one factor out the initial compilation time? :)

Just running `hyperfine -w1 -r 5 "zeek -O ZAM -r ../500k-syns.pcap"`, the new version in this PR is ~0.6% faster than before.

That being said, that difference is so minor that it might just be an artifact.

vpax:
> How does one factor out the initial compilation time? :)

You do `hyperfine ... "zeek -O ZAM /dev/null"`.

vpax:
> the new version in this PR ... than before

Can you disambiguate that? What's the before/after?

0xxon:
Ah, I will calculate out the initial time tomorrow.

With ZAM, for me it's faster with the change in this PR in, and slightly slower with the current state of master; the difference was the mentioned 0.6%.

To be exact: 
c9e659cb:
hyperfine -w1 -r 5 "zeek -O ZAM -r ../500k-syns.pcap"
Benchmark 1: zeek -O ZAM -r ../500k-syns.pcap
  Time (mean ± σ):     33.335 s ±  0.121 s    [User: 38.398 s, System: 2.666 s]
  Range (min … max):   33.142 s … 33.442 s    5 runs
  
55cdb707:
Benchmark 1: zeek -O ZAM -r ../500k-syns.pcap
  Time (mean ± σ):     33.552 s ±  0.135 s    [User: 39.454 s, System: 2.954 s]
  Range (min … max):   33.374 s … 33.716 s    5 runs

vpax:
> Ah, I will calculate out the initial time tomorrow.

FWIW, given what you show, removing the initial compilation time won't change the picture significantly. (It's generally a couple of seconds, when using a default build w/ C++ optimization turned on.) So I think that says that for those who are performance sensitive, as long as they're willing to run with ZAM this change has no appreciable performance impact - very nice!

0xxon:
Rebased on master and adopted for the merging of #4683, which added one more `set_conn` call.

scripts/policy/protocols/conn/pppoe-session-id-logging.zeek
```diff
@@ -19,9 +19,6 @@ event new_connection(c: connection)
 	if ( session_id == 0xFFFFFFFF )
 		return;
 
-	# FIXME: remove when GH-4688 is merged
-	set_conn(c, F);
```
awelzel, line 23: Also `module Conn` is probably not required anymore.

(mostly just think it'd be good to not allow policy scripts access to module internals, if only `set_conn()` was the reason for it)

scripts/base/protocols/conn/main.zeek
```diff
@@ -312,6 +311,21 @@ event tunnel_changed(c: connection, e: EncapsulatingConnVector) &priority=5
...
+	set_conn(c, F);
+	}
+
+event connection_flipped(c: connection) &priority=5
+	{
+	# otherwise, set-conn has not been called yet. In that case we don't have to do anything
+	if ( c?$conn )
+		{
+		c$conn$local_orig=Site::is_local_addr(c$id$orig_h);
+		c$conn$local_resp=Site::is_local_addr(c$id$resp_h);
```
awelzel:
```suggestion
		c$conn$local_resp = Site::is_local_addr(c$id$resp_h);
```

Eh, seems these are the only non-spaced assignments here.

scripts/base/protocols/conn/main.zeek
```diff
@@ -312,6 +311,21 @@ event tunnel_changed(c: connection, e: EncapsulatingConnVector) &priority=5
...
+	{
+	set_conn(c, F);
+	}
+
+event connection_flipped(c: connection) &priority=5
+	{
+	# otherwise, set-conn has not been called yet. In that case we don't have to do anything
+	if ( c?$conn )
+		{
+		c$conn$local_orig=Site::is_local_addr(c$id$orig_h);
```
awelzel:
```suggestion
		c$conn$local_orig = Site::is_local_addr(c$id$orig_h);
```

scripts/base/protocols/conn/main.zeek
```diff
@@ -243,26 +243,31 @@ function conn_state(c: connection, trans: transport_proto): string
...
+			}
+
+		if ( c$id$proto != 65535 )
+			c$conn$ip_proto = c$id$proto;
 
-	if ( eoc )
+		if( |Site::local_nets| > 0 )
+			{
+			c$conn$local_orig=Site::is_local_addr(c$id$orig_h);
+			c$conn$local_resp=Site::is_local_addr(c$id$resp_h);
```
awelzel:
```suggestion
			c$conn$local_resp = Site::is_local_addr(c$id$resp_h);
```

scripts/base/protocols/conn/main.zeek
```diff
@@ -243,26 +243,31 @@ function conn_state(c: connection, trans: transport_proto): string
...
+			add c$conn$tunnel_parents[c$tunnel[|c$tunnel|-1]$uid];
+			}
+
+		if ( c$id$proto != 65535 )
+			c$conn$ip_proto = c$id$proto;
 
-	if ( eoc )
+		if( |Site::local_nets| > 0 )
+			{
+			c$conn$local_orig=Site::is_local_addr(c$id$orig_h);
```
awelzel:
```suggestion
			c$conn$local_orig = Site::is_local_addr(c$id$orig_h);
```

scripts/policy/protocols/conn/pppoe-session-id-logging.zeek
0xxon: Looking at other policy scripts, it seems that they typically are put into the same module namespace as the base-scripts, even if they don't use anything from that namespace.

I am temped to just keep this the same for these scripts - it's the module namespace that seems most appropriate without introducing a new one for each script, and there is no technical drawback for keeping it like this.

scripts/base/protocols/conn/main.zeek
0xxon: Thanks, I will fix this :)

scripts/policy/protocols/conn/pppoe-session-id-logging.zeek
awelzel: Ah, hmm. I was thinking no module at all. Which should be okay if these scripts don't have their own options.

And community-id has a dedicated module and explicitly re-opens `Conn` just to get access to `set_conn()`, so that seemed not useful anymore.

0xxon: Ah, you are definitely right about community-id. I am going to remove the Conn module from that.

I am going to keep Conn for the others - I admittedly still don't see an advantage to have them in GLOBAL, and it fits the scheme of the other policy scripts.

src/fuzzers/dns-fuzzer.cc
```diff
@@ -48,6 +48,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     auto conn = add_connection();
     auto a = add_analyzer(conn);
 
+    if ( new_connection )
```
awelzel, line 52: Maybe put the commit message as the comment here?

awelzel:
Mostly to keep a note somewhere. This change increased max-memory usage quite a bit for syn-only pcaps.

My understanding is that in this scenario, a high-priority `connection_state_remove()` handler would initialize and populate the `Info()` record, a low-priority handler would log it directly and soon after the whole `connection` record would be discarded by Zeek, including `Info()`. That's what happened before.

Now, with the `Info()` record being initialized during `new_connection()`, its memory is kept for a duration of `tcp_SYN_timeout` before then being logged. In a syn-flood scenario, this means more `Info()` records are alive at any given point, resulting in higher memory usage.

I think it's okay still to do this, but I was a bit disappointed seeing that any effect on memory usage from #4702 was pretty much nil'ed out by this :-)

PR #4690, opened 2025-07-23
RFC: dce-rpc: Rename named_pipe to sec_addr

### LINKED ISSUE #3935: Unclear name of the "named_pipe" field in DCE_RPC
It is not clear why the value from the "sec_addr" field goes to the "named_pipe" field for the case when dce/rpc over tcp ([ncacn_ip_tcp](https://learn.microsoft.com/en-us/windows/win32/midl/ncacn-ip-tcp)). The port number does not really fit the named pipe format (\pipe\Pipe Name)

image

I would like to find out why the name of the field is like this and propose to change the name of the field, for example, to "secondary_address" so that it is more consistent with the Open Group standard

awelzel:
This is in response to #3935. I'm really not sure we should do that, but it was important enough for someone to open an issue.

IMO, Zeek distributors could work around that by extending DCE_RPC::Info with a new sec_addr field and removing `&log` from the original one. We could provide a policy script to do that, but that seems little useful to make progress towards removing the old field. So outright removing and providing an option to add it back is what is proposed here.

Needs updates of external baselines. Skipping on that for now.

---

@ckreibich or others: Thoughts? If it's a NACK, I'm going to close #3935 with suggestion to either  do that change locally, or use the following:
```
redef Log::default_field_name_map += {
        ["named_pipe"] = "sec_addr",
};
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,19 @@ release.
 Breaking Changes
 ----------------
 
+- The ``named_pipe`` field in ``dce_rpc.log`` was renamed to ``sec_addr`` to better
+  reflect the generic meaning of the field in DCE-RPC (it can be a port number, too).
+
+  The ``named_pipe`` field continues to exist on the ``DCE_RPC::Info`` record, but
+  will be removed with Zeek 8.1.
+
+  This might break your data pipeline. It's possible to re-instantiate the logging
+  of the ``named_pipe`` field with the following snippet:
+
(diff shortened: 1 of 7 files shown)


0xxon:
Would it perhaps just be enough to document the field better, and to state what it contains?

To me this does not feel like quite a compelling enough reason to change the schema.

awelzel:
Replaced by #4701 - extending the docs a bit.

Issue #3730, opened 2024-05-09
Track uptime, and report it as well as PID in stats & telemetry


ckreibich:
When grepping through `stats.log` it's unclear when/whether a given node restarted, and inferring it from the numbers themselves is awkward and prone to error.  Adding the PID would make this explicit, and also help relate the node to other system-level logging. Tracking and reporting node uptime would help identify lifetime more broadly, since the point of restart may not be available in a given log file.

These could become part of telemetry too, as gauges.

@timwoj, @bbannier, fyi

timwoj:
This one is pretty simple with the new async counter mode in the new telemetry rework. Add a counter that has a callback method that sets the counter to the uptime, and Prometheus will update it whenever it needs to. Optionally you could set one of the labels on the counter to the PID, if that's something that is actually wanted.

JustinAzoff:
fwiw, it can potentially be more useful to have the metric be the 'start time' vs the 'up time'.

that's how prometheus normally does it for the node_exporter:

```
# HELP node_boot_time_seconds Node boot time, in unixtime.
# TYPE node_boot_time_seconds gauge
node_boot_time_seconds 1.702939098e+09
```

If you have that you don't necessarily need the pid, since a change in that time means the node restarted, and you can easily compute the uptime based on it.

bbannier:
> fwiw, it can potentially be more useful to have the metric be the 'start time' vs the 'up time'.

What is the benefit of uptime over boot time? In my experience I found uptime more useful since it always increases monotonically, so if it does not go up this is helpful information about e.g., the state of the metrics collection pipeline (it effectively acts as some timestamped health check). If uptime ever decreases one knows that there must have been a restart, and can even potentially recover the approximate time via extrapolation (modulo sampling jitter).

Apart from allowing exact recovery of the restart time I cannot really see any benefit of boot time over uptime, so maybe you could expand. Do you want to optimize for compression in the metrics collection backend?

I 100% agree that the pid is not very interesting. If needed I believe one could scrape that via Prometheus system metrics already anyway (as well as `process_start_time`).

awelzel:
> so if it does not go up this is helpful information about e.g., the state of the metrics collection pipeline (it effectively acts as some timestamped health check)

If you implement it by reading `/proc/self/stat` (or some such) anew instead of caching, just getting back a value seems to be useful info.

> Apart from allowing exact recovery of the restart time I cannot really see any benefit of boot time over uptime, so maybe you could expand.

[...]

PR #4693, opened 2025-07-23
Add process_start_time_seconds and process_pid metrics

timwoj:
Add two new metrics that can be used to track the uptime of a Zeek process and whether the process has restarted. These two are:

- `process_start_time_seconds`: The process start time in seconds, using current_time truncated to the second. Note that this might not be the exact start time due to it not being added until after script parsing, but it's likely close enough to not matter.
- `process_pid`: The process' PID

Both are gauges, even though they'll never change after startup, simply to keep the prometheus-cpp library from appending `_total` to the end of the names.

This results in these metrics in the Prometheus output:

```
# HELP process_start_time_seconds Process start time
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1753307631
# HELP process_pid Process PID
# TYPE process_pid gauge
process_pid 65697
```

This is based on @awelzel's comment [here](https://github.com/zeek/zeek/issues/3730#issuecomment-2239510285):

> Tested Prometheus with Zeek and was reminded of this ticket when I came across two process_start_time_seconds produced by the local node exporter process and the prometheus server itself. So if anything, for consistency with those?
>
> Also, searching for that name specifically brings up [results](https://www.robustperception.io/alerting-on-crash-loops-with-prometheus/) indicates one pattern is using [changes()](https://prometheus.io/docs/prometheus/latest/querying/functions/#changes) to detect how often a metric has changed and alert on that (to be fair, there's also resets()).

### DIFF
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -135,6 +135,10 @@ void Manager::InitPostScript() {
                               []() { return static_cast<double>(get_stats()->fds); });
 #endif
 
+    // These two metrics get set at startup and are never modified after.
+    process_start_time = GaugeInstance("process", "start_time", {}, "Process start time", "seconds");
+    process_start_time->Set(run_state::zeek_start_time);
+
     if ( ! iosource_mgr->RegisterFd(collector_flare.FD(), this) ) {
         reporter->FatalError("Failed to register telemetry collector descriptor");
     }
--- a/src/telemetry/Manager.h
+++ b/src/telemetry/Manager.h
@@ -282,6 +282,7 @@ private:
     CounterPtr cpu_user_counter;
     CounterPtr cpu_system_counter;
     GaugePtr fds_gauge;
+    GaugePtr process_start_time;
 
     std::shared_ptr<prometheus::Registry> prometheus_registry;
     std::unique_ptr<prometheus::Exposer> prometheus_exposer;
--- a/testing/btest/Baseline/scripts.base.frameworks.telemetry.process-collect/out
+++ b/testing/btest/Baseline/scripts.base.frameworks.telemetry.process-collect/out
@@ -5,9 +5,11 @@ process_cpu_system_seconds_total, [], >0.0, good
 process_resident_memory_bytes, [], >0.0, good
 process_virtual_memory_bytes, [], >0.0, good
 process_open_fds, [], >0.0, good
+process_start_time_seconds, [], >0.0, good
 zeek_done
 process_cpu_user_seconds_total, [], >0.0, good
 process_cpu_system_seconds_total, [], >0.0, good
 process_resident_memory_bytes, [], >0.0, good
 process_virtual_memory_bytes, [], >0.0, good
 process_open_fds, [], >0.0, good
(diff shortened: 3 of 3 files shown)


0xxon (approved):

src/telemetry/Manager.cc
```diff
@@ -135,6 +135,13 @@ void Manager::InitPostScript() {
                               []() { return static_cast<double>(get_stats()->fds); });
 #endif
 
+    // These two metrics get set at startup and are never modified after.
+    process_start_time = GaugeInstance("process", "start_time", {}, "Process start time", "seconds");
+    process_start_time->Set(std::trunc(util::current_time(true)));
```
awelzel: We have run_state::zeek_start_time, populated during setup() - how about using that? And I wouldn't truncate. Seem entirely possible to have two restarts in a second.

```suggestion
    process_start_time->Set(run_state::zeek_start_time);
```


FWIW: uptime can be computed roughly by using `zeek_net_timestamp_seconds - process_start_time_seconds`.

src/telemetry/Manager.cc
```diff
@@ -135,6 +135,13 @@ void Manager::InitPostScript() {
...
+
+    process_pid = GaugeInstance("process", "pid", {}, "Process PID", "");
+    process_pid->Set(getpid());
```
awelzel:
```suggestion
```

Can we skip the pid, please?

It's not sensible to compute a `rate()` on pids. Neither is `min()` or `max()` or `avg()` useful. Comparing the pid of one process with another isn't useful either (e.g. compare with CPU or memory usage). I'm not sure why we'd use a metrics database for the pid. The value is probably also irrelevant as soon as the the process restarts. It's definitely not a counter, but it doesn't feel like a gauge either due to not being able to use aggregations in a sensible way on it.

Maybe for correlating some log output with metrics, but it seems all logs should always have the node name in them (usually a supervisor would prefix or store log lines under the node name).

If one really really wants the pid, they can create a gauge within `local.zeek`: We have a `getpid()` bif, so that's feasible.

awelzel (changes requested):

src/telemetry/Manager.cc
timwoj: Sure, done.

src/telemetry/Manager.cc
```diff
@@ -135,6 +135,10 @@ void Manager::InitPostScript() {
                               []() { return static_cast<double>(get_stats()->fds); });
 #endif
 
+    // These two metrics get set at startup and are never modified after.
+    process_start_time = GaugeInstance("process", "start_time", {}, "Process start time", "seconds");
```
bbannier, line 139: Just wondering whether one would ever need the timezone of this value, but probably not. The actual value of this is not really interesting, and its only function is to be able to see it change.

awelzel: Unix timestamps don't need timezones :sweat_smile:

awelzel (approved):

src/telemetry/Manager.cc
awelzel: Thanks, sorry for the wall of text!

Issue #4656, opened 2025-07-14
EDNS analyzer incorrectly computes extended rcode


dwhitemv25:
DNS_MsgInfo::BuildEDNS_Val() incorrectly computes the EDNS extended RCODE field. The original RCODE field is 4 bits, not 8.

This bug exists since at least the 7.0.x branch, and is present in master.

The faulty code is here:

https://github.com/zeek/zeek/blob/448590270b17cbd3c237c8cd45bc0f534282894d/src/analyzer/protocol/dns/DNS.cc#L1740

It should read:

```
unsigned int return_error = (ercode << 4) | rcode;
```

This can be worked around in scriptland in event dns_EDNS_addl like so:

```
event dns_EDNS_addl(c: connection, msg: dns_msg, ans: dns_edns_additional)
{
	if ( c$dns?$rcode && ans?$extended_rcode ) { 
		local fixed_rcode = ((ans$extended_rcode & 0xff00) >> 4) | (c$dns$rcode & 0xf);
		c$dns$rcode = fixed_rcode;
		c$dns$rcode_name = DNS::base_errors[fixed_rcode];
	}
}
```

If it would help I could do a one-line PR (my github-fu is weak though, it'd take a little setup).

Thanks!

0xxon:
Hi,

thank you for finding and reporting this. I do have a local branch that fixes this and adds a testcase. Do you have a preferred name that we should use to credit you with reporting this? Is dwhitemv25 ok?

We probably also should start a discussion about our treatmend of EDNS records in general. It does not feel great that, currently, we just discard them and don't parse them. To me it feels like edns errors should generally be parsed and, e.g., show up in DNS.log as the normal rcode.

I assume you agree, as this is exactly what you are doing in your snippet - but I was wondering if you have any additional opinions about this.

dwhitemv25:
GitHub username is OK for credit.

The logged rcode should be the extended rcode if it's present. It's garbage otherwise. (you'll see a lot of YXRRSET if you don't, as BADCOOKIE is a common rcode and the bottom bits overlap YXRRSET.)

As far as EDNS treatment, yes, it's probably time to start logging its fields. In particular I'd like to see the advertised packet size, as it'd confirm a couple of cases of providers that appear to have missed DNS Flag Day 2020. Note that the value can be different for originator and responder. But all of the EDNS fields are valuable.

[...]

PR #4695, opened 2025-07-24
Fix parsing of EDNS rcode

0xxon:
The EDNS rcode was incorrectly calculated. The extended rcode is formed by taking the upper 8 bits of the extended rcode field, plus the lower 4 bits of the existing rcode.

This also adds a new trace with an extended rcode, and a testcase parsing it.

Reported by dwhitemv25.

Fixes GH-4656

Trace is self-generated.


testing/btest/scripts/base/protocols/dns/edns-rcode.zeek
```diff
@@ -0,0 +1,13 @@
+# @TEST-DOC: Tests that the correct extended rcode is returned for EDNS packets.
```
awelzel: Could put a reference the original ticket
```suggestion
# @TEST-DOC: Tests that the correct extended rcode is returned for EDNS packets.  Regression test for #4656.
```

PR #4696, opened 2025-07-24
Make conn_id$ctx field a singleton

awelzel:
For a 200k syn pcap, this reduces memory usage in bare mode with an empty context by ~22MB (1061928 vs 1040744).

### DIFF
--- a/src/Type.cc
+++ b/src/Type.cc
@@ -915,7 +915,22 @@ public:
         return {v, init_type};
     }
 
-    bool IsDeferrable() const override { return false; }
+    bool IsDeferrable() const override {
+        if ( init_expr->Tag() == EXPR_RECORD_CONSTRUCTOR ) {
+            // Special-case deferrable record construction.
+            auto rce = zeek::cast_intrusive<RecordConstructorExpr>(init_expr);
+            auto rt = rce->GetType<zeek::RecordType>();
+
+            // The empty constructor_list check here is a short-cut: If the
+            // constructor expression contained only further const expressions
+            // or only further deferrable record constructors, this could be
+            // more aggressively deferring initializations.
+            auto constructor_list = rce->Op();
+            return rt->IsDeferrable() && constructor_list->Exprs().empty();
+        }
+
+        return false;
+    }
 
     ExprPtr InitExpr() const override { return init_expr; }
 
@@ -995,6 +1010,8 @@ private:
             if ( ! ci.second->IsDeferrable() )
                 rt->creation_inits[i++] = std::move(ci);
             else {
+                // std::fprintf(stderr, "deferred %s$%s: %s\n", obj_desc_short(rt).c_str(), rt->FieldName(ci.first),
+                //             ci.second->InitExpr() ? obj_desc_short(ci.second->InitExpr()).c_str() : "<none>");
                 assert(! rt->deferred_inits[ci.first]);
                 rt->deferred_inits[ci.first].swap(ci.second);
             }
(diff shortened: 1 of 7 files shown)


src/Type.cc
```diff
@@ -915,7 +915,21 @@ class ExprFieldInit final : public FieldInit {
...
+            // Special case deferrable record construction.
+            auto rce = zeek::cast_intrusive<RecordConstructorExpr>(init_expr);
+            auto rt = rce->GetType<zeek::RecordType>();
+
+            // The empty constructor_list check here is a short cutt: If the
+            // constructor expression contained only const values or only
+            // further deferrable record constructors, this could be more aggressive.
+            auto constructor_list = rce->Op();
+            return rt->IsDeferrable() && constructor_list->Exprs().empty();
+        }
```
awelzel, line 930: @vpax - does that seem reasonable to you? I'm also wondering if `&default=table()`, `&default=set()` and `&default=vector()` could be handled here?

src/Type.cc
```diff
@@ -915,7 +915,21 @@ class ExprFieldInit final : public FieldInit {
         return {v, init_type};
     }
 
-    bool IsDeferrable() const override { return false; }
+    bool IsDeferrable() const override {
+        if ( init_expr->Tag() == EXPR_RECORD_CONSTRUCTOR ) {
+            // Special case deferrable record construction.
```
ckreibich:
```suggestion
            // Special-case deferrable record construction.
```

src/Type.cc
```diff
@@ -915,7 +915,21 @@ class ExprFieldInit final : public FieldInit {
...
+            auto rce = zeek::cast_intrusive<RecordConstructorExpr>(init_expr);
+            auto rt = rce->GetType<zeek::RecordType>();
+
+            // The empty constructor_list check here is a short cutt: If the
```
ckreibich:
```suggestion
            // The empty constructor_list check here is a short-cut: If the
```

src/Type.cc
vpax: Doing this for the common case of empty `table()`/`set()`/`vector()` constructors makes sense. I'm wondering why it's needed for an empty `record()`, i.e. when does that case actually arise?

vpax: (presumably it's for the `$ctx` field, but I'm not offhand picturing how that leads to this situation)

awelzel:
```
type conn_id: record {
        orig_h: addr &log;      ##< The originator's IP address.
        orig_p: port &log;      ##< The originator's port number.
        resp_h: addr &log;      ##< The responder's IP address.
        resp_p: port &log;      ##< The responder's port number.
        proto: count &default=65535;  ##< The transport protocol ID. Defaults to 65535 as an "unknown" value.
        ctx: conn_id_ctx &log &default=conn_id_ctx();  ##< The context in which this connection exists.
};
```

That `default=conn_id_ctx()` is evaluated eagerly at record construction when it could be delayed until first access, too. In the scope of this PR, we want to populate the field always with a singleton instance when `NumFields() == 0` (default). Having this extra logic prevents creating a short-lived instance before assigning the singleton.

vpax: And I take it the `&default` is needed because you don't want to break existing constructors? Can it be `&optional`?

awelzel: Yeah, `conn_id()` should just continue to work without needing to adapt anything.  Seems `&optional` would be okay as we could populate `ctx` within the core for every new connection and `c$id$ctx` would always work. But then, if you do `conn_id(...)` by hand, it isn't set. That feels surprising/confusing, so I find the `&default=` behavior actually better. I was mostly just surprised it didn't end-up being deferred initialized which is why I started staring :-)

vpax: Got it, thanks for the sketch. I'm fine going ahead with your proposed change

PR #4698, opened 2025-07-24
Remove zeek::Span and use std::span instead

Neverlord:
Filing as draft since it'll need a rebase after merging #4678.


awelzel:
I think we technically need to keep `Span.h` and `zeek::Span` around with a deprecation period until 8.1:
```
// Span.h
namespace zeek {
[[deprecated("Remove in v8.1: Use std::span instead")]]
using Span = std::span;
};
```

Though, I wouldn't mind keeping ``zeek::Span`` around even if it's just an alias to std. While I don't think there's many plugins that would be affected, does it hurt? (I'm very open to be convinced using `std::span` in the APIs is better, just wondering about the churn).

Most span APIs were introduced after the 7.0 release, so getting this in before 8.0 would be a good time to switch, otherwise we'll need to look to deprecate in v9.1. Are the APIs in `telemetry/` compatible with external plugins that may have used `zeek::Span`?

Neverlord:
> Though, I wouldn't mind keeping `zeek::Span` around even if it's just an alias to std. While I don't think there's many plugins that would be affected, does it hurt? (I'm very open to be convinced using `std::span` in the APIs is better, just wondering about the churn).

The way I see it, there's just no point in keeping `zeek::Span` beyond the deprecation cycle. It was intended from the get-go as a temporary placeholder until `std::span` becomes available and implemented just enough of the standard interface so we can start using a new STL feature ahead of widely available compiler support (we've only implemented dynamic extend, for example).

timwoj:
The other cpp20 PR landed, if you want to rebase this one.

Neverlord:
> The other cpp20 PR landed, if you want to rebase this one.

Done. 🙂

timwoj (approved):

PR #4699, opened 2025-07-25
Add ODesc::Size() that returns size_t, deprecate ODesc::Len()

timwoj:
This came up while working on the recent security issues, and I split it out to be applied separately. This adds a new `ODesc::Size()` that turns a `size_t` and deprecates `ODesc::Len()` which previously returned an `int`. It also updates all of the internal size-related values in `ODesc` to be `size_t`. There are various functions that changed from taking `unsigned int` to `size_t`. These should be fine to change without a deprecation cycle since both of those are unsigned values, and none of them are virtual so we don't have to worry about them being overridden.

It also changes `SerializationFormat` to use `size_t` for byte sizes as well. This is in the `zeek::detail` namespace, so it shouldn't require a deprecation cycle.


src/logging/writers/ascii/Ascii.cc
```diff
@@ -242,7 +242,7 @@ void Ascii::InitConfigOptions() {
 
     ODesc tsfmt;
     BifConst::LogAscii::json_timestamps->Describe(&tsfmt);
-    json_timestamps.assign((const char*)tsfmt.Bytes(), tsfmt.Len());
```
awelzel, line 245: Kind of separate from what this PR does, but I've started to leverage `zeek::obj_desc()` from `Desc.h` quite a bit recently and this could probably be:
```
json_timestamps = zeek::obj_desc_short(BifConst::LogAscii::json_timestamps);
```

src/strings.bif
```diff
@@ -121,7 +121,7 @@ function join_string_vec%(vec: string_vec, sep: string%): string
 		e->Describe(&d);
 		}
 
-	zeek::String* s = new zeek::String(true, d.TakeBytes(), d.Len());
+	zeek::String* s = new zeek::String(true, d.TakeBytes(), d.Size());
 	s->SetUseFreeToDelete(true);
 
 	return zeek::make_intrusive<zeek::StringVal>(s);
```
awelzel, line 127: Nothing to do, just wondering if a ref-qualified `StringValPtr TakeStringVal() &&` on `ODesc` could make sense to encapsulate the creation of a StringVal from `ODesc`. Feels slightly too often open-coded with that `s->SetUseFreeToDelete(true)` wart :-)

Issue #4694, opened 2025-07-24
Move service_violation field to deprecated-dpd-log


0xxon:
It turned out that the analyzer.log changes lead to a bit of an increase of memory use, because we add one extra field to the connection record.

We can move the deprecated `service_violation` one to the deprecated dpd logging script and should not really impact anyone with that. Details at https://zeekorg.slack.com/archives/CTGDG83B8/p1753272837186929

PR #4702, opened 2025-07-25
RecordType: Allow deferring &default=vector(), set(), table() fields

awelzel:
After a bit of a detour, this PR does three things:

* Recognizes `&default=set()`, `&default=table()` and `&default=vector()` on record fields as deferrable initializations.
* Deprecates putting `&optional` and `&default` on record fields - these have `&default` behavior and may only be confusing.
* Continue what started with 6.0 and ensure that deleting a record field with a `&default` attribute is always re-initialized on a subsequent access. [Vern's thoughts convinced me](https://github.com/zeek/zeek/pull/4702#discussion_r2232110694). `record?$field` will evaluate to true for record fields with `&default` now, too. This was already the case for `&default` with constant expressions, like `&default=42`, so this just makes it more consistent and fits the `&default` on table, too.


testing/btest/Baseline/language.delete-field-set/output
```diff
@@ -1,2 +1,6 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-[a=<uninitialized>, b=<uninitialized>, c=<uninitialized>]
```
awelzel, line 2: @vpax and anyone curios - the deferred initialization interacts funnily with deleting record fields, sigh :-)

The test deletes fields on a record value. All fields use a `&default=constructor()` and after the change in this PR, they are all deferred initialized. This, however, means that after a `delete` of a field, accessing them again simply generates fresh values.

```
# test.zeek
type R: record {
        c: count &default=42;
        t: table[count] of string &default=table();
};

event zeek_init()
        {
        local r1 = R();
        print "r1", r1;
        delete r1$c;
        delete r1$t;
        print "r1?$c", r1?$c;
        print "r1?$t", r1?$t;
        print r1;
        }
```
Running with the code of this PR:
```
$ zeek test.zeek 
r1, [c=42, t={

}]
r1?$c, T
r1?$t, T
[c=42, t={

}]
```

This changed with 6.0 already for atomic types due to the deferral logic:
```
root@9cc67c8ba097:/test# zeek --version
zeek version 5.2.2
root@9cc67c8ba097:/test# zeek test.zeek 
r1, [c=42, t={

}]
r1?$c, F
r1?$t, F
[c=<uninitialized>, t=<uninitialized>]
```

```root@f731402c5949:/test# zeek --version
zeek version 6.0.4
root@f731402c5949:/test# zeek test.zeek 
r1, [c=42, t={

}]
r1?$c, T
r1?$t, F
[c=42, t=<uninitialized>]    # c re-initialized!
```

I'd argue an explicit `delete` on a field should persist and skip executing the `&default` expression on the next access. This could be pulled of by storing a few more bits of information into the `record_val` vector:
```
@@ -1505,7 +1509,16 @@ private:
     // Low-level values of each of the fields.
     //
     // Lazily modified during GetField(), so mutable.
-    mutable std::vector<std::optional<ZVal>> record_val;
+    enum class FieldFlags : uint8_t {
+        Set = 1,
+        Deleted = 2,
+        IsManaged = 4,
+    };
+
+    struct RecordValSlot {
+        ZVal zval;
+        FieldFlags flags;
+    };
+
+    mutable std::vector<RecordValSlot> record_val;
```

The interesting observation here is that `std::optional<ZVal>` is already using 16bytes for a single entry on my system (1byte flag + 7bytes padding + 8bytes ZVal), so it appears reasonable to replace std::optional with a custom struct. This would actually leave 7 bytes for some auxilary information per record field in case you have some use for optimization related things here?

Does this sketch make you raise eye-brows, or does that appear a reasonable addition/change?

awelzel: Ok, I'm going to try this out, because zeek-benchmarker indicates this makes quite a difference both in runtime and memory usage for various pcaps, partly probably also captures #4694, but hey :-) 

Screenshot from 2025-07-25 16-03-06
Screenshot from 2025-07-25 16-02-50

vpax: Hmmm I'm mixed on whether deleting a  `&default` field should make it go away. Deletes of `&default` table entries don't make those go away, you still get the value when you access it. Tracking the deletion status of every field strikes me as peculiar, even though we can do it without any memory overhead.

FWIW, we could consider instead generating a warning on such record field deletes, since we can identify instances statically.

vpax: That said, if you want help running down the ASAN issues you're bumping into, the first question is what is the context around `ZAM-EvalDefs.h:9883` (which ZAM instruction it corresponds to), and same for `ZAM-EvalDefs.h:9347`. (Or I can try building your branch locally w/ ASAN to see if it reproduces.) Interesting that it bombed on a `vlan_fivetuple` test, makes me wonder if this is a more basic ZAM problem where somewhere it doesn't know about `ConnID` extensions.

awelzel: > Hmmm I'm mixed on whether deleting a &default field should make it go away. Deletes of &default table entries don't make those go away, you still get the value when you access it.

Yep, agree. I veered off attempting to restore Zeek 5.2 behavior, but not doing it appears better. I like the `table` `&default` comparison. This then also also allows `delete r$vec` or `delete r$tbl` to reset/empty a vector or table - not exactly the same as doing a plain `delete vec` or `delete tbl`, but it's has a pretty close effect.

> That said, if you want help running down the ASAN issues you're bumping into

So I did some additional dabbling with the `RecordVal` in `Val.h` and reduced its object size from 96 bytes to 64 bytes by removing the `RecordType rt` member (can just cast the Val's type), removing the `is_managed` vector ref member (storing the `managed` flag per slot based on ZVal::IsManagedType()) and also replacing the `std::vector` with a custom vector that uses 16 bytes instead of 24 bytes. The zeek-benchmarker indicates this helps runtime and memory usage for various pcaps and has some larger effects for some microbenchmarks (creating many connection records).

EDIT: Oh, and also removing `virtual ~Modifiable()` and calling `Modifiable::Unregister()` explicitly for the 3 classes where it matter.

Unfortunately, I broke ZAM and am unable to fix it :-( The `RawOptField()` and `RawField()` now have very different semantics (and probably should have other names like `RawSlot()` or so).

Can I somehow excite you to skim the PR and see if the `RecordValSlot` and `RecordValSlots` additions make sense to you and how or if you could see switching ZAM over? There's one commit in the branch where I tried updating and it actually compiles, but I'm not getting the ref-counting right and everything I do just feels very random due to lack of knowledge :-(

This isn't important and doesn't need to be in 8.0, but I realized there might be a few low-hanging fruits to save a bit of memory and improve performance, so maybe you find that cool :-)

src/Val.h
```diff
@@ -1114,6 +1116,94 @@ struct is_zeek_val {
...
+namespace detail {
+
+/**
+ * Representation of a single slot for a field in a RecordVal.
+ *
+ * Previously this was a std::optional<ZVal> that was using 16bytes. This has
+ * been switched to a struct to be able to have access to the wasted bytes to
+ * store some additional flags.
+ */
+class RecordValSlot {
```
awelzel: Supposedly this could be used in VectorVal, too. So maybe `ZValSlot` would be better.

testing/btest/Baseline/language.delete-field-set/output
vpax: @awelzel I've pushed fixed for script optimization to the `topic/vern/zam-defer-tweaks` branch off of your branch, see what you think.

There's also a patch required for `Gen-ZAM.cc`, not sure how to best push that, so I've attached that here.
[defer-init-gen-zam-patch.diff.txt](https://github.com/user-attachments/files/21449137/defer-init-gen-zam-patch.diff.txt)

vpax: oops and I should add: while I've done testing that there are no ASAN complaints, I haven't done testing for memory leaks

vpax: Oops #2: feel free to rename `RawOptField` to something like `RawOptRecordSlot` ... and I think `RawFieldRef` could maybe be `RawFieldMod` to convey that it's only used when the field will be modified.

awelzel: Thanks @vpax for checking it out! Doesn't look super invasive, hopefully at some point I'll be able to figure that out myself.

I've pushed everything here in a separate `topic/awelzel/record-val-slot-work` branch including your fixes. On further reflection, I'll revert this PR to just the `&default` deferral, keeping `std::vector<std::opional<ZVal>>` for 8.0. This was a nice detour, but for 8.0 seems just a bit too much churn so late in the cycle.

I'm thinking we could also put the `ZVal`'s TypeTag (currently just 1 byte) into the slot itself - that might allow to skip some `GetType()` indirections to figure out the field's type, though I'm not sure if that'd even apply to ZAM, or would more help a classic script interpretation.

Thanks!

vpax: Deferring makes sense, it is reaching deep into the guts, and I'd like to confirm that the ZAM side of things doesn't have any leaks.

By the `ZVal`'s TypeTag you mean the one for the associated record field? If so, no that's indeed not germane for ZAM.

NEWS
```diff
@@ -361,6 +365,9 @@ Deprecated Functionality
 
   The replacement script doesn't populate the ``email_body_sections`` anymore either.
 
+- Using ``&default`` and ``&optional`` together on a record field has been deprecated
+  as it would only result in ``&default`` behavior.
```
0xxon: Perhaps add that it will be an error starting in 8.1?

I know that that's implied - but it might make sense to call it out.

src/Type.cc
```diff
@@ -1007,12 +1021,34 @@ class RecordType::CreationInitsOptimizer : public detail::TraversalCallback {
...
+                }
+            }
             else {
-                // std::fprintf(stderr, "deferred %s$%s: %s\n", obj_desc_short(rt).c_str(), rt->FieldName(ci.first),
-                //             ci.second->InitExpr() ? obj_desc_short(ci.second->InitExpr()).c_str() : "<none>");
-                assert(! rt->deferred_inits[ci.first]);
+                // std::fprintf(stderr, "deferred %s$%s: %s (%s)\n", obj_desc_short(rt).c_str(),
+                // rt->FieldName(ci.first),
+                //             ci.second->InitExpr() ? obj_desc_short(ci.second->InitExpr()).c_str() : "<none>",
+                //            typeid(ci).name());
```
0xxon: I know that this doesn't really matter - but the new formatting of this coment looks kind of awkward, with the indentation being all over the place.

awelzel: Yeah, thanks. I was seeing if I could sneak in my fprintf() debugging here. I'll just ditch it completely.

Relatedly, I was pretty close to adding a new "ast" or "type" debug stream, so that something like `zeek -B ast` would work with `DBG_LOG`. I'll ask in #development if that could be interesting to Vern or others, too.

awelzel:
> Just so that I understand this right - with the fact that tables/set/vectors with an &default of table(), set(), and vector() are now initialized deferred - there is no memory advantage anymore of using optional anymore, right?
>
> Obviously there still is a use for &optional for logging, where there is a semantic difference between an empty and a non-existant set - but perhaps we should also highlight this somehow.

Yes, that's the right take. As you point out, there's a difference for log fields being empty vs not there. Also `&optional` has  "have I ever initialized the field" as a piece of information that `&default` doesn't provide.

I'll add a snippet about memory savings and ``&default`` to NEWS.

A `&default=vector(1,2,3)` would actually not be lazily initialized because the "is it deferrable" logic defaults to false for anything but the most simplest expressions. If someone feels strongly, this can be improved in follow-ups.

vpax:
In retrospect, I think it wasn't worth it to make mixing `&optional` and `&default` together a fatal error (or come to think of it, even a warning). I'm having to touch quite a bit of our older code due to that change, to no actual benefit.

awelzel:
> In retrospect, I think it wasn't worth it to make mixing &optional and &default together a fatal error

There was one occurrence in the Zeek tree. At the time I thought it makes sense as a cleanup to disallow mixing/adding attributes (`&optional` here specifically) that won't have any effect. But certainly overreached with the fatal error.

Prefer a revert or have you figured it out?

vpax:
> Prefer a revert or have you figured it out?

At this point a revert won't help since it'll take a while for it to actually be integrated into our build stuff and we need the stuff working soon, so we'll continue to pointwise fix.

PR #4703, opened 2025-07-25
Lazily allocate the map of option handlers in ID objects

timwoj:
While tinkering with fixing up https://github.com/timwoj/struct_layout, I noticed that we allocate a 48-byte `multimap` to hold the change handlers for options. We only actually add handlers 13 times in the base scripts. Instead of allocating it for all of them, lazily allocate it when needed. Local testing don't show a huge change with this, but this reduces the size of a `::zeek::detail::ID` instance from 152 bytes to 104 bytes, and fits it inside of 2 cache lines instead of 3.

This also moves the `type` field up a couple of rows in class definition for better packing. This fills a 6-byte hole after `is_export` and replaces a 1-byte hole after `infer_return_type` and a 4-byte hole after `offset` with a single 3-byte hole after `infer_return_type`.

I'm opening this as a draft to see what the benchmarks think.

### DIFF
--- a/src/ID.cc
+++ b/src/ID.cc
@@ -576,15 +576,21 @@ void ID::UpdateValID() {
 }
 #endif
 
-void ID::AddOptionHandler(FuncPtr callback, int priority) { option_handlers.emplace(priority, std::move(callback)); }
+void ID::AddOptionHandler(FuncPtr callback, int priority) {
+    if ( ! option_handlers )
+        option_handlers = std::make_unique<OptionHandlerMap>();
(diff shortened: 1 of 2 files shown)


timwoj:
Surprisingly this is a wash on memory and an increase of CPU time. It was worth a try.

timwoj:
Reopening this after talking to @awelzel about it.

awelzel (approved):
As in DM, think a sorted list would be fine, but no strong opinion.

Issue #4337, opened 2025-04-07
A custom analyzer to investigate all packets


FishyFluffer:
I'm trying to make a custom analyzer that is capable of handling every packet coming in and handled in Zeek. The current best approach for this I could find is to make use of the HookSetupAnalyzerTree functionality to add a new custom analyzer as high in the tree as possible.

The problem here is that the highest tree node possible is still one of the protocol analyzers, such as the TCPSessionAdapter. This introduces a point of possible packet loss in case that analyzer decides not to pass a packet on to it's children, such as malformed data or a mismatched checksum. As far as I can tell, there's no way for me to indicate interest in these undelivered packets as they are discarded in the root node.

I can work on a PR for a resolution to this, but wanted to open a discussion on the best path forward. There are two options that immediately come to mind:

- A new Hook that allows access to packets at a higher point in the data flow than the protocol session adapters. This could take place at the end of IPBasedAnalyzer::AnalyzePacket for example, and would provide access to every packet assigned to a connection. However, this would lose easy access to the scope given by a custom Analyzer for per-connection details and state tracking. An analyzer tree will be available per connection still, so a custom analyzer lookup under this hook could work but sounds like an expensive set of operations to perform per packet.
- A new function added to existing session adapters or analyzers: DeliverSkippedPackets. This would be an alternative function called by an Analyzer if it determines to skip further processing of packets, and could be picked up by the custom analyzer. An analyzer catching packets in either DeliverPackets or DeliverSkippedPackets should be capable for handling data for every packet passed through the parent, and should be enough for this requirement.

Of the two options given, I'm leaning more toward the second. The biggest potential worry I have with it is if it's expensive to iterate through child Analyzers and providing DeliverSkippedPackets on each, but with them being Noops and this being a relatively rare case, I don't see there being much actual impact from it. It might also be a challenge to actually track if a packet is being skipped or not for the followup call, but I have yet to investigate into the exact implementation details.

AmazingPP:
FYI, I once wrote a similar custom analyzer (for DPI integration) that also needed to handling every incoming packet in Zeek. I used a somewhat hacky approach—VMT hooking (Often used for game hacking)—to avoid making intrusive modifications to Zeek. The general steps were as follows:

1. Call `zeek::iosource_mgr->GetPktSrc()` to get the current packet source instance.
2. Hook the `PktSrc::ExtractNextPacket` virtual function and save the original virtual function pointer.
[...]

PR #4704, opened 2025-07-25
Added DeliverSkippedPacket() hook to capture skipped packets during analysis

siavashta:
This work addresses the suggestion made in [zeek/zeek#4337](https://github.com/zeek/zeek/issues/4337).

This introduces support for accessing packets that would otherwise be ignored by the current flow logic. Specifically, it adds a new hook (deliverSkippedPacket) that provides visibility into such packets, allowing plugins or scripts to process them if needed.


src/analyzer/Analyzer.h
```diff
@@ -165,6 +172,29 @@ class Analyzer {
     void NextPacket(int len, const u_char* data, bool is_orig, uint64_t seq = -1, const IP_Hdr* ip = nullptr,
                     int caplen = 0);
 
+    /**
+     * Passes a skipped packet input to the analyzer for processing. The
+     * analyzer will process the input with any support analyzers first
+     * and then forward the data to DeliverSkippedPacket(), which derived
+     * classes can override.
```
awelzel, line 183: It probably makes sense to describe or define a "skipped packet" here a bit. It's just a packet that was sufficiently valid to be associated with a connection/session and the analyzer tree, but then processing was "skipped" due to checksum or other consistency errors.

I'm also wondering if we should reduce the number of parameters, but might be reasonable for consistency.

@rsmmr - we've chatted about this one at some point. It's about implementing an API for analyzers to get notified for every single packet associated with a session, even those that don't make it through to `DeliverPacket()` - for one reason or another.

The PR looks good to me for 8.0, but assigning you to double check. The repeated analyzer house-keeping code is a bit annoying, but I don't think we need to fix that here.

rsmmr: The mechanics look fine to me, the code repetitiveness is really just due to the way that Zeek structures things.

I have a conceptual question: what exactly are the semantics for when an analyzer should signal a skipped packet? Is it "whenever I don't deliver it, I call skipped"? And for an analyzer implementing the callback, does it mean "whenever my *direct* parent could not handle it, I will see it as skipped"? The two things I'm wondering about here (1) are TCP/ICMP/UDP the only analyzers that need to be instrumented to call "skipped", and (2) what would happen with longer chains of analyzers: is it well-defined (and useful) when exactly skipped gets called and who gets to see it?

A test would be good to add as well, also because it would act as an example on how to use this.

awelzel: > And for an analyzer implementing the callback, does it mean "whenever my direct parent could not handle it, I will see it as skipped"?
> The two things I'm wondering about here (1) are TCP/ICMP/UDP the only analyzers that need to be instrumented to call "skipped"

Hmm, this is a good point. The use-case in question only considered the TCP/ICMP/UDP root analyzers/session-adapters, yes. The analyzer motivating this mostly wants to have a signal if a packet is not delivered and is only ever attached to the root analyzer ~at this point~. Could be attached further down, but unclear whether that would really work due to expecting reliable `DeliverPacket()` invocations.


> and (2) what would happen with longer chains of analyzers: is it well-defined (and useful) when exactly skipped gets called and who gets to see it?

Hmm. Hmm. I'd say it's only useful for the top-level session adapters to produce skipped packets, but that's also the only thing I really thought about.

It seems though the same question exists already for the other API methods? If a child analyzer doesn't call `ForwardPacket()` within its own `DeliverPacket()` method, the packet isn't forwarded to that analyzer's child analyzers anymore. So it's already at the discretion of individual analyzer implementations today what they forward (unless I'm missing something crucial here)? I could see that just staying that way also for skipped packets and I think is just what the PR does?

> A test would be good to add as well, also because it would act as an example on how to use this.

I can do the test, but it'd be essentially just HookSetupAnalyzer() attaching to the root and reading a pcap with checksum errors or so.

awelzel: Test is here: https://github.com/zeek/zeek/commit/ff00104b5915ce32339986b1191060e9075210ca

@siavashta - is this something you could cherry-pick into your branch? If not, I can also do it during the merge.

siavashta: Sure. Added.

timwoj:
@awelzel Is this needed anymore after https://github.com/zeek/zeek/pull/4725 was merged?

siavashta:
> @awelzel Is this needed anymore after #4725 was merged?

I dont think so. Closing the pr.

PR #4706, opened 2025-07-26
Expand coverage of IRC analyzer with more commands

timwoj:
This adds two new pcaps to expand the coverage for the IRC analyzer in testing.

- A new irc-dcc-send pcap that includes [tags from IRCv3](https://ircv3.net/specs/extensions/message-tags), plus a fix for that.
- A pcap that includes a bunch of commands that we weren't covering in the other pcaps, plus an expansion of the scripts.base.protocols.irc.events btest to catch all of them. I've had the pcap sitting on my desktop for a couple of years, and finally got around to opening a PR for it.

We're still missing support for [SASL `AUTHENTICATION`](https://ircv3.net/specs/extensions/sasl-3.2) commands amongst [some other v3 features](https://ircv3.net/irc/), but we should probably rewrite this thing in Spicy instead of expanding the existing analyzer.


src/analyzer/protocol/irc/IRC.cc
```diff
@@ -78,6 +78,20 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
...
+    // which is a timestamp for when the message was sent. They come as a semi-colon
+    // separated list, and are separated from the rest of the line by a space.
+    if ( myline.starts_with("@") ) {
+        std::size_t first_space = myline.find(' ');
+        // A set of tags doesn't necessarily need to be followed by a command. In this
+        // case just return.
+        if ( first_space == string::npos )
+            return;
+
+        // We don't do anything with tags right now. Just skip over them.
```
awelzel, line 91: Hmm, could we just raise the string without further decoding, at least that provides a bit of an escape hatch to get access to the data from script land without delving too deep into parsing and semantics of tags?

```
event irc_raw_tags(c: connection, tags: string);
```

The Copyright lines on that document make me think IRC is not something from the past :sweat_smile: 
https://ircv3.net/specs/extensions/message-tags

testing/btest/Baseline/core.conn-size-threshold/.stdout
```diff
@@ -4,27 +4,35 @@
...
+0
+0
+0 secs
+Threshold set for [orig_h=10.0.0.22, orig_p=42114/tcp, resp_h=10.0.0.7, resp_p=52349/tcp, proto=6, ctx=[]]
+3000
+2000
+63
+50
+100.0 msecs, 722.885132 usecs
+triggered duration, [orig_h=10.0.0.22, orig_p=42114/tcp, resp_h=10.0.0.7, resp_p=52349/tcp, proto=6, ctx=[]], 100.0 msecs, F, 6.0 secs 600.0 msecs 922.822952 usecs
```
awelzel: This is now missing the "triggered packets" line, is there not enough packets in the trace?

testing/btest/Baseline/scripts.base.protocols.conn.threshold/.stdout
```diff
@@ -1,13 +1,11 @@
...
+Threshold set for [orig_h=10.0.0.7, orig_p=52236/tcp, resp_h=10.0.0.22, resp_p=6667/tcp, proto=6, ctx=[]]
+triggered bytes, [orig_h=10.0.0.7, orig_p=52236/tcp, resp_h=10.0.0.22, resp_p=6667/tcp, proto=6, ctx=[]], 1, T
+triggered bytes, [orig_h=10.0.0.7, orig_p=52236/tcp, resp_h=10.0.0.22, resp_p=6667/tcp, proto=6, ctx=[]], 2000, F
+triggered duration, [orig_h=10.0.0.7, orig_p=52236/tcp, resp_h=10.0.0.22, resp_p=6667/tcp, proto=6, ctx=[]], 200.0 msecs, F
+Threshold set for [orig_h=10.0.0.22, orig_p=50102/tcp, resp_h=10.0.0.7, resp_p=52333/tcp, proto=6, ctx=[]]
+triggered bytes, [orig_h=10.0.0.22, orig_p=50102/tcp, resp_h=10.0.0.7, resp_p=52333/tcp, proto=6, ctx=[]], 1, T
+triggered bytes, [orig_h=10.0.0.22, orig_p=50102/tcp, resp_h=10.0.0.7, resp_p=52333/tcp, proto=6, ctx=[]], 2000, F
+Threshold set for [orig_h=10.0.0.22, orig_p=42114/tcp, resp_h=10.0.0.7, resp_p=52349/tcp, proto=6, ctx=[]]
+triggered bytes, [orig_h=10.0.0.22, orig_p=42114/tcp, resp_h=10.0.0.7, resp_p=52349/tcp, proto=6, ctx=[]], 1, T
+triggered duration, [orig_h=10.0.0.22, orig_p=42114/tcp, resp_h=10.0.0.7, resp_p=52349/tcp, proto=6, ctx=[]], 200.0 msecs, F
```
awelzel: Again lost a "triggered packets" line here.

awelzel (approved):
The threshold tests have a bit less coverage, so that'd be nice to have another look, but otherwise :+1:

src/analyzer/protocol/irc/IRC.cc
timwoj: IRCv3 is actually surprisingly new, considering how old IRC is. The ircd I used to generate the more-commands pcap didn't support any of the features when I made that pcap. This time around it stumped me why we weren't processing anything from the new dcc-send pcap.

testing/btest/Baseline/core.conn-size-threshold/.stdout
timwoj: I might just need to transfer a larger file. I used http/get.trace from our traces library. It might just not be big enough to hit the thresholds.

PR #4709, opened 2025-07-28
Deprecate `record_type_to_vector`

evantypanski:
After #4682, `record_fields` is ordered so this is redundant. Furthermore, it is incorrectly documented, as it says nested records are flattened.


awelzel (approved):
And it doesn't get the number of fields right :-)
```
$ zeek -e 'print |record_type_to_vector("conn_id")|'
7
$ zeek -e 'print |record_fields("conn_id")|'
6

```

ckreibich:
The docs label here is just so we don't forget to add this to the deprecations in `NEWS`.

PR #4711, opened 2025-07-29
Fixes for record access in specialized ZAM operations

vpax:
ZAM uses low-level `RecordVal` methods for efficient access to record fields. Some of these usages need to confirm that a field exists before accessing or modifying it. This PR adds in those fixes.

One minor subtlety is that the regression BTest doesn't check all combinations of just which fields are missing. This is because some of the accessing code might pick one or other (i.e., the first or second access in an expression) depending on how the C++ compiler decides to evaluate expressions. Either order is correct for our purposes, but the BTest could have incidental variation if we added tests to confirm that both accesses are checked. We could fix that with enough normalization, but it doesn't seem worth it for a test that's just for catching regressions.

@awelzel This collides somewhat with the ZAM fixes I sent you over the weekend, but should be easy enough to fix up. Also, the API you're introducing should obviate the `We make this call to ensure that the field exists` snippet.


src/script_opt/ZAM/ZBody.h
```diff
@@ -65,6 +65,20 @@ class ZBody : public Stmt {
 private:
     friend class CPPCompile;
 
+    // Helper run-time function for looking up a field in a record, checking
+    // that it exists and complaining if it does. A member here rather than
```
awelzel:
```suggestion
    // that it exists and complaining if it does not. A member here rather than
```
?

src/script_opt/ZAM/ZBody.h
```diff
@@ -65,6 +65,20 @@ class ZBody : public Stmt {
...
+    // a standalone run-time function because ZBody is a "friend" of RecordVal
+    // and can use its low-level record field accessors.
+    ZVal CheckAndLookupField(RecordVal* r, int f, std::shared_ptr<ZAMLocInfo> loc) {
```
awelzel:
```suggestion
    ZVal CheckAndLookupField(const RecordVal*r r, int f, const std::shared_ptr<ZAMLocInfo>& loc) {
```

I'm a bit worried about copying a `std::shared_ptr` parameter for a potentially hot function (is it?). Maybe inlining makes that a non-issue, but goldbolt indicates even for an inline function there might be a `lock inc`:
```
        mov     rcx, qword ptr [rip + __libc_single_threaded@GOTPCREL]
        cmp     byte ptr [rcx], 0
        jne     .LBB0_1
        lock            inc     dword ptr [rax + 8]
        mov     rax, qword ptr [rsp]
        mov     esi, dword ptr [rax]
        jmp     .LBB0_3
.LBB0_1:
        mov     dword ptr [rax + 8], 2
        mov     esi, 32
.LBB0_3:
        mov     rdi, qword ptr [rip + std::cout@GOTPCREL]
        call    std::ostream::operator<<(int)@PLT
```

Note that `mov     rcx, qword ptr [rip + __libc_single_threaded@GOTPCREL]` thing. libc goes out of its way to avoid a `lock add` when it "knows" that the process runs single threaded.

```
#include <memory>
#include <iostream>

inline bool x(std::shared_ptr<int> c) {
    std::cout << *c << std::endl;
}


int main() {
    auto c = std::make_shared<int>(32);
    x(c);
    return 0;
}
```

awelzel (commented):
Hey @vpax , could you double check on the shared_ptr usage? LGTM otherwise :+1:

src/script_opt/ZAM/ZBody.h
vpax: Fixed, thx for spotting

src/script_opt/ZAM/ZBody.h
vpax: That code actually isn't a hot-spot, but point taken. I've made the fix (though the first argument needs to be non-const since `RecordVal` access can modify the record itself due to deferred initializations)

vpax:
@awelzel I've pushed fixes for the above. In addition, I thought of a way to extend the BTest to individually check the different facets of the fix, so I've made that change as well.

awelzel:
> @awelzel I've pushed fixes for the above. 

Thanks. Merged. Is that something that should go in 7.0.x, too?

vpax:
@awelzel Yes, makes sense for 7.0.x.

PR #4714, opened 2025-07-29
cluster: Move cluster::detail::Event to cluster::Event

awelzel:
This class is a parameter of virtual methods of the Backend API for users to implement and also a parameter to the HookPublishEvent() API. Seems it shouldn't be in detail and instead we should own it.

Alternatively, could mark the cluster APIs as not-stable-yet, but I think we can move forward and make it non-detail for 8.0.

### DIFF
--- a/src/cluster/serializer/broker/Serializer.cc
+++ b/src/cluster/serializer/broker/Serializer.cc
@@ -2,7 +2,6 @@
 
 #include "zeek/cluster/serializer/broker/Serializer.h"
 
-#include <cinttypes>
 #include <optional>
 
 #include "zeek/DebugLogger.h"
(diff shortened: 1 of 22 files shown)


src/cluster/serializer/broker/Serializer.cc
```diff
@@ -96,7 +95,7 @@ std::optional<broker::zeek::Event> detail::to_broker_event(const detail::Event&
     return broker::zeek::Event(ev.HandlerName(), xs, broker_meta);
 }
 
-std::optional<detail::Event> detail::to_zeek_event(const broker::zeek::Event& ev) {
+std::optional<zeek::cluster::Event> detail::to_zeek_event(const broker::zeek::Event& ev) {
```
timwoj, line 98: Just for consistency, can you pick either `cluster::Event` or the fully-qualified version? It switches back and forth a couple times in this file.

timwoj (approved):
I like that being in it's own file 👍🏼

src/cluster/serializer/broker/Serializer.cc
awelzel: Ah, thanks. I tried to stay with `cluster::Event` here, apparently I failed :-)

PR #4716, opened 2025-07-30
cluster: Add on_subscribe() and on_unsubscribe() hooks

### LINKED ISSUE #4176: Introduce hooks for Cluster subscribe and unsubscribe
I have a use case where I want to know which topics a node is subscribed to. For this, I'd like to have hooks like `Cluster::on_subscribe(topic_prefix: string)` and `Cluster::on_unsubscribe(topic_prefix: string)` to maintain a list of currently subscribed topics.

awelzel:
Closes #4176

---

@J-Gras - I just saw this in the issue list. Does that work for you?

I was pretty close to suggesting you could do this via `HookFunctionCall()` in your plugin, too, so I'm a bit on the fence if we really need these new hooks in the script layer. Someone else on the team might have an intuition, too.


J-Gras:
Thanks for tackling this! I am wondering whether the hook should be called from the underlying C++ function just in case someone uses that API without going through scriptland.

awelzel:
> Thanks for tackling this! I am wondering whether the hook should be called from the underlying C++ function just in case someone uses that API without going through scriptland.

Makes sense. I pushed that version.

As mentioned in DM, I'd prefer if this was carried as an external patch until it has proven to be actually needed, but I'll let @ckreibich make that call.

awelzel:
>> Thanks for tackling this! I am wondering whether the hook should be called from the underlying C++ function just in case someone uses that API without going through scriptland.
>
> Makes sense. I pushed that version.

Hmm, actually. Doing it that way, `Cluster::on_subscribe()` and `Cluster::on_unsubscribe()` will also be called for WebSocket client subscriptions as they each have their own dedicated backends. With the proposed hooks, there's no possibility to distinguish between core and websocket subscriptions. WebSocket client subscriptions are already available via the `Cluster::websocket_client_added()` event today.

So this should probably go via `cluster::detail::EventHandlingStrategy` instead and noop'ed for WebSocket clients. Or go back to the script-level only approach where that question doesn't come into play.

@ckreibich - assigning you to make the call and/or merge. IMO this isn't quite ready yet :-/

ckreibich:
I'm still bullish on this. :-) A couple of thoughts:
* Couldn't you argue that it is in fact a benefit if you see all of the subscriptions, so you don't have to know about the difference to `Cluster::websocket_client_added()`?
* I'm reading the original motivation in #4176 from that angle too — i.e., that seeing everything is beneficial.
* It seems unproblematic to iterate on this down the road, since it's such an uninvasive change, and clearly outside what most regular users will encounter.

Thoughts?

J-Gras:
Regarding the original motivation, @ckreibich is on point! It's probably the behavior we are looking for. It could be beneficial to be able to distinguish between Zeek vs. a Websocket endpoint behind a Zeek instance being the subscriber. However, I agree to Christian that it would be fine to iterate on that, e.g. by extending the hook signature to provide that additional context.

awelzel:
> Thoughts?

:ship: :slightly_smiling_face:

PR #4718, opened 2025-07-31
Fix use-after-move reported by clang-tidy

timwoj:
This was introduced by 9eb94ee151ec437fd076ef16d016af372391af65 (https://github.com/zeek/zeek/pull/4702)


awelzel (approved):
Ouch, thanks!

PR #4724, opened 2025-08-01
Implement string- and container-length filtering at the log record level

### LINKED ISSUE #4705: Limiting max log record sizes produced by Zeek
### Discussed in https://github.com/zeek/zeek/discussions/2797

<sup>Originally posted by **awelzel** February 17, 2023</sup>
## About

Provide mechanism to limit/cap the size of log records produced by Zeek. The main motivation is for protection from DoS either on the Zeek side due to producing overly large log record, or downstream systems due to overload or space exhaustion.

## Background

Log records produced by Zeek are currently not size limited in any way when rendered into JSON or TSV. For log records having a  "bounded representation" (int, addr, port, enum, bool, time, double, ...) this isn't an issue. A log record containing string, set or vector types, however is subject to grow unbounded (or minimally be of unreasonable size).

Scripting errors, analyzer errors, pathological traffic patterns or simply including non-sanitized input from analyzers into log fields can result in unintentionally producing enormous log record sizes.

## Downstream consumers

Downstream consumers or storage systems may impose limits on the size of records they allow. Either on the size of the whole log record, or on individual values/fields within a log record. If these limits are exceeded, data might be discarded or truncated and with luck reported through metrics.

Without limits, processing overhead increase on the Zeek side, but also on downstream consumers. There [have been issues around this](https://github.com/zeek/zeek/pull/2302). Once downstream consumers are unable to keep up and fall behind processing incoming logs, eventually something else will run out of buffer space cause data loss. There have also been issues by downstream systems rejecting too large messages, but producers not recognizing the error code and retrying endlessly.

### Example limits

Minimal survey of a few limits out there. Happy to hear of some more.

* Humio - 1mb event size: https://library.humio.com/falcon-logscale/limits.html
* MongoDB - 16 mb max document size: https://www.mongodb.com/docs/manual/reference/limits/#bson-documents
* filebeat log input - 10mb default max: https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-log.html#_max_bytes_2
* Kafka message max bytes - default 1mb: https://kafka.apache.org/documentation/#brokerconfigs_message.max.bytes
* Loki max_label_value_length: <int> | default = 2048: https://grafana.com/docs/loki/latest/configuration/#limits_config

## Proposal

Introduce limits on the number of bytes allowed for individual string and container fields, as well as in total. Other field types are not relevant for capping the size with the assumption that they are small enough and may therefore not contribute significantly to the total size (with respect to generally large limits).

Different record types have different maximum sizes depending on their complexity, though are constrained by the "total" limits. A record type with four string log fields, for example, has twice the maximum allowed size than a record field with 4 string fields.

#### Summary
Introduce 4 new options, representing limits for individual fields and for the whole total of a record.
```
Log::limit_field_string_bytes         # Recommended default: 2KB
Log::limit_field_container_elements   # Recommended default: 50 (?)
Log::limit_total_string_bytes         # Recommended default: 250KB (? -250 KB of string data is large.
Log::limit_total_container_elements   # Recommended default 500 (?)
```

Extension/customization could be done on a per-stream/per-filter options overriding the global ones: For those use-cases where logging a 1MB string is on purpose.

For the actual processing, treat each record separately. When it's going through `Log::write()` and is converted to a `threading::Val` withing the logging Manager:
* Keep track of the total number of container elements and total number of string bytes (after capping)
* Cap string fields to `limit_field_string_bytes`
* Cap containers (vector, table, set) to `limit_field_container_elements`
* For other elements, process them, don't keep track.

* If limit_total_string_bytes is reached:
  * Render all strings as empty string "" unless within containers
  * Convert any containers holding strings to empty containers, if currently in such a container, stop processing it.
* If limit_total_container_elements is reached:
  * Convert any container of any type to empty containers
  * If currently in a container, return.

Essentially, there are two stages / configuration levels
  * Per field truncation
  * Emergency truncation once the "total" limits are hit.

I believe this could actually be implemented just with log policy hooks, too. It might not be the most efficient, however.

### Examples

The following record with the proposed default values would be bounded at 2*2KB + "few bytes for time" that aren't tracked.
```
type Info: record {
  t: time &log;
  s: string &log;
  t: string &log;
}
```

The following record may reach a total of 2KB + 50 * 2KB string bytes. The vector of enum values may reach a maximum size of 50 elements and then be capped, the enums will not be counted towards string bytes (even with large enum names, it's difficult to cause significant sizes).
```
type Info: record {
  s: string
  v: vector of string &log;
  e: vector of SomeEnum;
}
```

### Out-of-band: telemetry / metrics

We can expose per `record$field` occurrences of truncation as telemetry. Number of times a field was truncated and possibly the sum of bytes that were truncated. Further, "emergency" truncation can be tracked via per Record type metrics.

### reporter.log

For a combination of record type and field name, we could do a "warn once" or "warn once in a while" through reporter.log if truncation is happening so that it'll be visible on existing/established channels.

#### New Fields

* Adding new fields to log records to indicate truncation. Tricky, changes log schema.

#### Use N+1 limits
* From @bbannier
 
> Another option would be to let users configure some maximum length N, but then emit fields of length N+1 in case of overflow (we could terminate with a marker byte, but not strictly necessary). This would allow downstream consumers to detect the truncation without us having to change the schema.

I've been mulling over this and I'd personally pass on the N+1. The limits are intended to be set way larger than what's expected. If the output happens to hit the limit (or anywhere near 1% of the limit) it's likely already unexpectedly large. If it hits the limit, in very likely got truncated.

My take is that using strings of length N+1 will only bring up questions, require additional documentation, and I doubt downstream consumers would actually implement something clever with it (and then they could also implement the cleverness using the actual limit). On the other hand, a user testing something, and setting a limit/cap to 100 and then seeing strings of length 101 will think the implementation is buggy (I would).</div>

timwoj:
This adds four new constants to set maximum lengths for strings and containers when logging. Two new telemetry metrics track when the limits are met per log stream. If the limits are met, two new weirds are used for notification of the truncation.

Fixes #4705

This can wait to be merged until after 8.0 is branched.


timwoj:
The zeek-private ctu-sme pcap is failing to test because of some absolutely massive X509 logs. There's one log entry that has a `cert` string field that is 16584 characters and a `san.uri` vector field that has 268 entries. The current maximums for those fields are 4096 bytes for individual strings and 100 entries for individual containers. The maximum size of a log line is 16MB though, so we could set those higher if needed.

I'm thinking 32KB for string fields, 300 for individual containers, and 1500 for total container fields in a record. 32KB for strings feels really big though.

src/logging/Manager.cc
```diff
@@ -1149,7 +1181,9 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
 
         // Alright, can do the write now.
         size_t total_size = 0;
-        auto rec = RecordToLogRecord(stream, filter, columns.get(), total_size);
+        total_string_bytes = 0;
+        total_container_elements = 0;
+        auto rec = RecordToLogRecord(w->second, filter, columns.get(), total_size);
```
awelzel: How would you feel about over-engineering this a bit?

I'm thinking it'd be cool if the logic was in a separate class/classes and that the logging manager had a way to configure/set the strategy/policy for log record limiting. That would allow Zeek distributors to come up with their own policy without needing to patch base Zeek. It would also allow exploration of other forms of reporting (e.g. raise an event instead) without needing to patch/change anything in logging::Manager.

IMO doesn't need to be a full component/plugin style approach, but a simple class/interface might be good, even if it's all in detail.

I haven't thought about this all that much, but maybe something like the following:

```
enum class Action {
    Pass,  // take as is
    Truncate,  // truncate string or container
    Abort,  // stop processing any container elements, needed?
};

struct LimiterResult {
    Action action;
    uint32_t truncated_size;
};

class LogRecordLimiter {
    virtual Action AddString(const String* s) = 0;
    virtual Action AddContainerElement()  = 0;

    virtual void Reset() = 0;  // Either that, or create a new limiter instance for every record? We do so much malloc in this path, a new instance is probably okay and easier to reason about than `Reset()`.
 };

LogRecordLimiterFactory {
  unique_ptr<LogRecordLimiter> New();
}

// logging manager
class Manager {
   ...
   void SetLogLimiterFactory(LogRecordLimiterFactoryPtr factory);
};
```

Unit testing would also be much easier, but we don't do that a lot, so maybe that doesn't count :-)

timwoj:
> The zeek-private ctu-sme pcap is failing to test because of some absolutely massive X509 logs. There's one log entry that has a cert string field that is 16584 characters and a san.uri vector field that has 268 entries. 

This one is fixed by adding stream-level values and setting values specifically for x509 when that log is created. Next up is a failure in the zeek-testing-private `short` test that has a URI that's over 32KB. The RFC for HTTP/1.1 doesn't specify a maximum length for URIs, so it's technically a valid request.

src/logging/Manager.cc
timwoj: I'm inclined to say hold off on something like this for now. Let's get what's here already in and if we find that having the ability to set policies for it then we can iterate on it later.

scripts/base/files/x509/main.zeek
```diff
@@ -117,7 +117,11 @@ redef record Files::Info += {
 
 event zeek_init() &priority=5
 	{
-	Log::create_stream(X509::LOG, Log::Stream($columns=Info, $ev=log_x509, $path="x509", $policy=log_policy));
+	# x509 can have some very large certificates and very large sets of URIs. Expand the log size filters
+	# so that we're not truncating those.
+	Log::create_stream(X509::LOG, Log::Stream($columns=Info, $ev=log_x509, $path="x509", $policy=log_policy,
+	                                          $max_field_string_bytes=24576, $max_field_container_elements=500,
+	                                          $max_total_container_elements=1500));
```
awelzel: Mind adding the magic values as `const &redef` in the X509 module? Probably named the same as those in `Log` - does that feel reasonable to you?

```suggestion
	                                          $max_total_container_elements=X509::default_max_container_elements));
```

src/logging/Manager.cc
```diff
@@ -491,7 +501,15 @@ Manager::Manager()
       total_log_writer_discarded_writes_family(
           telemetry_mgr->CounterFamily("zeek", "log-writer-discarded-writes",
                                        {"writer", "module", "stream", "filter-name", "path"},
-                                       "Total number of log writes discarded due to size limitations.")) {
+                                       "Total number of log writes discarded due to size limitations.")),
+      total_log_writer_limited_string_fields_family(
+          telemetry_mgr->CounterFamily("zeek", "log-writer-limited-string-fields",
```
awelzel:
```suggestion
          telemetry_mgr->CounterFamily("zeek", "log-writer-truncated-string-fields",
```

Truncated is what's used in the doc strings of the options and sound nicer to me. What do you think? If you agree, should probably update the variable names, too.

src/logging/Manager.cc
```diff
@@ -659,9 +677,24 @@ bool Manager::CreateStream(EnumVal* id, RecordVal* sval) {
     streams[idx]->columns = columns->Ref()->AsRecordType();
     streams[idx]->max_delay_interval = sval->GetField("max_delay_interval")->AsInterval();
     streams[idx]->max_delay_queue_size = sval->GetField("max_delay_queue_size")->AsCount();
-
     streams[idx]->enable_remote = id::find_val("Log::enable_remote_logging")->AsBool();
 
+    streams[idx]->max_field_string_bytes = sval->GetField("max_field_string_bytes")->AsCount();
+    if ( streams[idx]->max_field_string_bytes == 0 )
+        streams[idx]->max_field_string_bytes = std::numeric_limits<size_t>::max();
```
awelzel: Nit/consistency territory. I wonder if we should stick with `0` here and adapt the logic to special-case it. Mostly for when you're looking at stuff with `gdb`, the `max()` value may be less obvious than a `0`.

It's also the first time I'm seeing `if ( x == 0 ) x = max()` logic. We have quite a bit of that logic elsewhere and it usually is explicit `0` testing.

src/logging/Manager.cc
```diff
@@ -491,7 +501,15 @@ Manager::Manager()
       total_log_writer_discarded_writes_family(
           telemetry_mgr->CounterFamily("zeek", "log-writer-discarded-writes",
                                        {"writer", "module", "stream", "filter-name", "path"},
-                                       "Total number of log writes discarded due to size limitations.")) {
+                                       "Total number of log writes discarded due to size limitations.")),
+      total_log_writer_limited_string_fields_family(
+          telemetry_mgr->CounterFamily("zeek", "log-writer-limited-string-fields",
+                                       {"writer", "module", "stream", "filter-name", "path"},
+                                       "Total number of logged string fields limited by length")),
+      total_log_writer_limited_containers_family(
```
awelzel:
```suggestion
      total_log_writer_limited_container_fields_family(
```

Just for consistency with string_fields_family.

src/logging/Manager.cc
```diff
@@ -1464,12 +1500,25 @@ threading::Value Manager::ValToLogVal(std::optional<ZVal>& val, Type* ty, size_t
...
         case TYPE_STRING: {
             const String* s = val->AsString()->AsString();
-            char* buf = new char[s->Len()];
-            memcpy(buf, s->Bytes(), s->Len());
+
+            size_t allowed_bytes = std::min({static_cast<size_t>(s->Len()), stream->max_field_string_bytes,
+                                             stream->max_total_string_bytes - total_string_bytes});
+
+            if ( allowed_bytes < s->Len() ) {
+                reporter->Weird("log_string_field_length_limited", util::fmt("%s", stream->name.c_str()));
```
awelzel:
```suggestion
                reporter->Weird("log_string_field_truncated", util::fmt("%s", stream->name.c_str()));
```

src/logging/Manager.cc
```diff
@@ -1508,33 +1557,61 @@ threading::Value Manager::ValToLogVal(std::optional<ZVal>& val, Type* ty, size_t
...
             bool is_managed = ZVal::IsManagedType(set_t);
 
-            zeek_int_t set_length = set->Length();
-            lval.val.set_val.vals = new threading::Value*[set_length];
+            size_t allowed_elements =
+                std::min({static_cast<size_t>(set->Length()), stream->max_field_container_elements,
+                          stream->max_total_container_elements - total_container_elements});
+
+            if ( allowed_elements < set->Length() ) {
+                reporter->Weird("log_container_field_length_limited", util::fmt("%s", stream->name.c_str()));
```
awelzel:
```suggestion
                reporter->Weird("log_container_field_truncated", util::fmt("%s", stream->name.c_str()));
```

src/logging/Manager.cc
```diff
@@ -1149,7 +1182,9 @@ bool Manager::WriteToFilters(const Manager::Stream* stream, zeek::RecordValPtr c
 
         // Alright, can do the write now.
         size_t total_size = 0;
```
awelzel: Should `total_size` now also become a member of logging::Manager and removed from the parameters?

Alternatively, capture the individual fields for tracking into a struct that's passed by reference. It's just that either all should be members or non I guess. Prefix the members with `record_` maybe, so they are a bit less ambiguous without reading doc strings.

src/logging/Manager.cc
```diff
@@ -1508,33 +1557,61 @@ threading::Value Manager::ValToLogVal(std::optional<ZVal>& val, Type* ty, size_t
...
         case TYPE_VECTOR: {
             VectorVal* vec = val->AsVector();
-            zeek_int_t vec_length = vec->Size();
-            lval.val.vector_val.vals = new threading::Value*[vec_length];
+
+            size_t allowed_elements = std::min({static_cast<size_t>(vec->Size()), stream->max_field_container_elements,
+                                                stream->max_total_container_elements - total_container_elements});
+
+            if ( allowed_elements < vec->Size() ) {
+                reporter->Weird("log_container_field_length_limited", util::fmt("%s", stream->name.c_str()));
```
awelzel:
```suggestion
                reporter->Weird("log_container_field_truncated", util::fmt("%s", stream->name.c_str()));
```

awelzel: A thought: There's enough padding in `threading::Val` to stash information about the fact that truncation happened into the `threading::Val` itself. If we do that as a public field in the threading::Val struct, the log writer/serializer could use that to annotate truncated fields or values if they deem that to be "the right thing to do" during writing. E.g. for an elastic-search exporter or postgres exporter, they could store `log_record_truncated: true` as a top-level field or extra columns, so an analyst would immediately know the record might be incomplete when focusing on it.

Something like this:
```
@@ -1505,6 +1505,7 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                                              stream->max_total_string_bytes - total_string_bytes});
 
             if ( allowed_bytes < s->Len() ) {
+                lval.truncated = true;
                 reporter->Weird("log_string_field_length_limited", util::fmt("%s", stream->name.c_str()));
                 info->total_limited_string_fields->Inc();
             }
@@ -1562,6 +1563,7 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                           stream->max_total_container_elements - total_container_elements});
 
             if ( allowed_elements < set->Length() ) {
+                lval.truncated = true;
                 reporter->Weird("log_container_field_length_limited", util::fmt("%s", stream->name.c_str()));
                 info->total_limited_containers->Inc();
             }
@@ -1592,6 +1594,7 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                                                 stream->max_total_container_elements - total_container_elements});
 
             if ( allowed_elements < vec->Size() ) {
+                lval.truncated = true;
                 reporter->Weird("log_container_field_length_limited", util::fmt("%s", stream->name.c_str()));
                 info->total_limited_containers->Inc();
             }
diff --git a/src/threading/SerialTypes.h b/src/threading/SerialTypes.h
index fb7336b9d5..066c346370 100644
--- a/src/threading/SerialTypes.h
+++ b/src/threading/SerialTypes.h
@@ -126,8 +126,10 @@ struct Field {
  * those Vals supported).
  */
 struct Value {
-    TypeTag type;    //! The type of the value.
-    TypeTag subtype; //! Inner type for sets and vectors.
+    TypeTag type;           //! The type of the value.
+    TypeTag subtype;        //! Inner type for sets and vectors.
+    bool present = false;   //! False for optional record fields that are not set.
+    bool truncated = false; //! True if the string or container field was truncated.
 
     struct set_t {
         zeek_int_t size;
@@ -186,8 +188,6 @@ struct Value {
         _val() { memset(this, 0, sizeof(_val)); }
     } val;
 
-    bool present = false; //! False for optional record fields that are not set.
-
     /**
      * Constructor.
      *
```

Instead of a bool, we could also make it a uint8_t flags field / enum if there might be more info we'd want to convey down the road, without wasting an extra 7 bits :-)

PS: If we'd be okay moving the `line_number` member around in threading::Val to come after `truncated`, this would reduce the size from 40 to 32 bytes. Given we allocate on threading::Val for every value in a log record, that might actually be worth it as a follow up. Btw, changing my stance on those clang-tidy enum size reports while looking at this stuff :+1:

testing/btest/scripts/base/frameworks/logging/field-length-limiting-stream.zeek
```diff
@@ -0,0 +1,108 @@
...
+export {
+	redef enum Log::ID += { LOG };
+
+	type Info: record {
+		strings1: vector of string &log;
+		strings2: vector of string &log;
+	};
+}
+
+redef Broker::disable_ssl = T;
```
awelzel:
```suggestion
```

testing/btest/scripts/base/frameworks/logging/field-length-limiting.zeek
```diff
@@ -0,0 +1,94 @@
...
+export {
+	redef enum Log::ID += { LOG };
+
+	type Info: record {
+		strings1: vector of string &log;
+		strings2: vector of string &log;
+	};
+}
+
+redef Broker::disable_ssl = T;
```
awelzel:
```suggestion
```

testing/btest/scripts/base/frameworks/logging/field-length-limiting.zeek
```diff
@@ -0,0 +1,94 @@
...
+
+event log_telemetry()
+	{
+	local storage_metrics = Telemetry::collect_metrics("zeek", "log_writer_limited**");
```
awelzel:
```suggestion
	local storage_metrics = Telemetry::collect_metrics("zeek", "log_writer_limited*");
```
This is just a glob, right?

testing/btest/Baseline/scripts.base.frameworks.logging.field-length-limiting-2/test.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	test
+#open XXXX-XX-XX-XX-XX-XX
+#fields	strings1	strings2
+#types	vector[string]	vector[string]
+ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ	ABCDEFGHIJ,ABCDE,(empty),(empty),(empty),(empty),(empty),(empty),(empty),(empty)
```
awelzel, line 10: This is curios. I'd have expected:
```
ABCDEFGHIJ,ABCDE,,,...
```
Would `(empty)` for an individual field in a `vector of string` not be interpreted as the literal "(empty)" string? I wonder if this is an artifact of the early return and `string_val.data` being nullptr? But I haven't looked. Should maybe store a valid empty-string pointer instead to represent truncation?

Also, maybe one set of tests with `LogAscii::use_json=T` would be good to double check how it ends-up looking there and if it looks like `strings2: ["ABCDEFGHIJ", "ABCDE", "", "", "", "", ...]`

testing/btest/Baseline/scripts.base.frameworks.logging.field-length-limiting-3/test.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	test
+#open XXXX-XX-XX-XX-XX-XX
+#fields	strings1	strings2
+#types	vector[string]	vector[string]
+ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDEFGHIJ,ABCDE,(empty)	(empty),(empty),(empty),(empty),(empty),(empty),(empty),(empty),(empty),(empty)
```
awelzel, line 10: Similarly wondering about `(empty)` vs actually `""`.

src/logging/Manager.cc
awelzel: Yeah, thanks for pushing back. Pondering, the algorithm itself is unlikely to change anyhow. If anything, maybe the reporting part could be made injectable. I.e. instead of the open-coded `Weird()` and `Inc()` calls, could pass in some callable that receives enough information, but no need in the first step, agree.

Resolving.

awelzel (commented):
I like this a lot! Had a few thoughts on a few corners, so posting those. CI is failing currently, too, but you probably saw that.

scripts/base/files/x509/main.zeek
timwoj: After talking with @ckreibich about these values a bit the other day, I have concerns about the length of these. The length of 24KB here fixes our testing, but there might be certificates that are longer in the wild. We'd end up truncating those.

src/logging/Manager.cc
timwoj: Yeah, that might be a little too cute.

scripts/base/files/x509/main.zeek
timwoj: I asked @0xxon to chime in.

testing/btest/Baseline/scripts.base.frameworks.logging.field-length-limiting-2/test.log
timwoj: Storing a single-character array with a null character in it doesn't change anything. It's written as part of the ascii formatter here: https://github.com/zeek/zeek/blob/master/src/threading/formatters/Ascii.cc#L99-L102. The actual `(empty)` string is passed down from the ascii log writer, which gets it from a script constant in the logging framework. That snippet checks for whether the length of the field is zero, so even setting the value to a valid pointer but with a null character wouldn't matter. The length is still zero and it gets written as `(empty)`.

testing/btest/Baseline/scripts.base.frameworks.logging.field-length-limiting-3/test.log
timwoj: Same as above.

src/logging/Manager.cc
timwoj: I like it 👍🏼 

> Btw, changing my stance on those clang-tidy enum size reports while looking at this stuff

It was going to happen eventually 😄

scripts/base/files/x509/main.zeek
awelzel: Might be reasonable to redef X509 string field length to `0` when the policy script that adds the full cert is loaded?
```
scripts/policy/protocols/ssl/log-certs-base64.zeek
```

Or - rather fancy - something like `cert: string &log=no_truncate &optional`, but that's more food for thought than actually thinking this would be worth it.

testing/btest/Baseline/scripts.base.frameworks.logging.field-length-limiting-2/test.log
awelzel: Curios! I should've stare at the code myself a bit. It's still surprising to me, but let me resolve the discussion.

To a degree, this is behavior specific to the ASCII writer, the truncation within the core/logging manager happens. If the writer decides to use 7 chars `(empty)` instead of none for reasons, that's the writer's decision :man_shrugging:

awelzel (approved):
Leaving an approval, as I don't have real concerns outside of the thing you're tracking already. A NEWS entry would be good, including info how to disable the truncation with setting everything to `0` and mentioning the weirds to look out for. Maybe even posting on Slack or the Forum in case someone out there has loads of PCAPs to test against?

Issue #4214, opened 2025-02-04
Nesting of analyzers may lead to stale argument values


J-Gras:
```
event connection_established(c: connection)
	{
	print c$resp;
	c = lookup_connection(c$id);
	print c$resp;
	}
```
The above code leads to a slightly unintuitive output of something like:
```
[size=0, state=4, num_pkts=0, num_bytes_ip=0, flow_label=0, l2_addr=00:0c:29:a5:45:e0]
[size=0, state=4, num_pkts=1, num_bytes_ip=60, flow_label=0, l2_addr=00:0c:29:a5:45:e0]
```
As `connection_established` event tiggers for TCP if we've seen a SYN-ACK, one would assume there is a packet (and maybe data) in the responder-side flow. For the connection that is given as parameter to `connection_established`, the values are not set. However, manually obtaining the connection shows the expected values.

The explanation is quite simple: When an analyzer triggers a parameterized event, it creates Zeek values for the corresponding arguments. The `connection_established` event is triggered by the TCP analyzer. However, the connection statistics are maintained by the ConnSize analyzer, which is a child analyzer of the TCP analyzer. Hence, when the `connection_established` event is generated, the values are not yet updated and the Zeek value that is passed to the event captures this pre-update state.

In general, I think this behavior is fine. I don't think it would make sense to think about lazy evaluation of arguments here. In particular, the ConnSize analyzer is probably kind of an edge case. However, I think it would make sense to document that behavior somewhere.

awelzel:
> However, I think it would make sense to document that behavior somewhere.

IMO we should see if there's a non-intrusive / low-overhead fix possible, as you write, this is very unintuitive except for people with your background.

cccs-graeme:
Here is a similar case.
```
event connection_SYN_packet(c: connection, pkt: SYN_packet)
{
    if(!pkt$is_orig) {
        print c$resp;
        c = lookup_connection(c$id);
        print c$resp;
    }
}
```
The above code produces surprising output similar to this:
```
[size=0, state=4, num_pkts=0, num_bytes_ip=0, flow_label=0, l2_addr=00:26:88:e7:8d:82]
[size=0, state=4, num_pkts=1, num_bytes_ip=60, flow_label=0, l2_addr=00:26:88:e7:8d:82]
```

awelzel:
Hmm, hmm, the main problem here is really just that we don't run `UpdateConnVal()` again.

[...]

PR #4725, opened 2025-08-02
Introduce TapAnalyzer for SessionAdapter

awelzel:
This is an alternative approach to #4704 that adds a dedicated TapAnalyzer concept.

This is close to @FishyFluffer 's first idea on #4337, but instead of a new hook, add a detail/internal API for adding light-weight analyzers instead to easily allow associating a session with a TapAnalyzer (see docstring).

@rsmmr - is this more or closer to the direction you'd prefer?

I've started with `TapAnalyzer` with the background that this is only for observing packets. I could see up-leveling this to a more powerful/flexible `FilterAnalyzer` and changing `DeliverPacket()` to something like `FilterPacket()`:

```
enum class PacketFilterVerdict {
    Pass,
    Drop,
}

class PacketFilterAnalyzer {
    FilterVerdict FilterPacket(const Packet& pkt);
};
```

Where users could then control if packets make it into session-analysis at all (think of "user-level shunting" decisions). It's not much additional code, so I'm happy to do that iteration if that seems useful.

I'm also wondering if the ConnSize analyzer could then become a tap or filter analyzer, but that might then need some additional plumbing with `UpdateConnVal()`, but should naturally lead to a fix for #4214 . So yeah, this feels like a better direction than the DeliverSkippedPacket idea overall. Thanks!


rsmmr:
> @rsmmr - is this more or closer to the direction you'd prefer?

Only took a quick look so far, but architecturally, yes, I like this better. +1 on this approach from me. One quick suggestion: combine`TapPacket` and `TapSkippedPacket` into one and use a second argument to distinguish the two cases (like an enum `Delivered` and `Skipped`). I imagine a tap analyzer will likely be using similar logic for both cases and then doesn't need to implement two methods separately.

> I could see up-leveling this to a more powerful/flexible FilterAnalyzer

Wouldn't go there for now unless there's a use case (and we also have `PacketFilter`).

awelzel:
Thanks @rsmmr ! I've force-pushed a somewhat consolidated version with a bit of documentation.

Would you be up for another look? I'm not really invested into the API details and have a hard time coming up with names (Verdict, Info struct, ...), so if you think this too complicated, nudge me into another direction :-)

> Wouldn't go there for now unless there's a use case (and we also have PacketFilter).

Yep, thanks! We have `SetSkip()` and that should be enough. Filtering individual packets seems too brittle to get right and anyhow there's also ebpf if someone is into that.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -2,13 +2,89 @@
...
+     */
+    virtual void UpdateConnVal(RecordVal* conn_val) {}
+
+    /**
+     * Hook for when this analyzer is about to be removed and destructed.
+     *
+     * This is invoked when the session's Done() method is invoked, just before
+     * the TapAnalyzer instance is destroyed.
+     */
+    virtual void Done() {};
```
rsmmr, line 84: If we have `Done`, also add `Init`, like we do with other analyzers?

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -2,13 +2,89 @@
...
+    virtual ~TapAnalyzer() = default;
+
+    /**
+     * Hook of a tap analyzer for receiving packet data.
+     *
+     * @param pkt The packet being processed.
+     * @param verdict Either Deliver or Skip as determined by session analyzers.
+     * @param info Additional information about tapping.
+     */
+    virtual void Packet(const Packet& pkt, TapVerdict verdict, const PacketTapInfo& info) = 0;
```
rsmmr: Would call this `TapPacket`, for consistency with the method in SessionAnalyzer. That way it's clear they belong togehter.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -94,9 +170,51 @@ class SessionAdapter : public analyzer::Analyzer {
...
+     * @param info Additional information about the packet.
+     */
+    void TapPacket(const Packet* pkt, TapVerdict verdict, const PacketTapInfo& info);
+
+    /**
+     * Helper for tapping a packet that is delivered.
+     *
+     * @param pkt The packet.
+     */
+    void TapPacket(const Packet* pkt) { return TapPacket(pkt, TapVerdict::Deliver, {}); }
```
rsmmr: Can just use default parameters on the other version of `TapPacket`.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -2,13 +2,89 @@
...
+enum class TapVerdict : uint8_t {
+    Deliver,
+    Skip,
+};
+
+/**
+ * Reason why delivery of a packet would be skipped.
+ */
+enum class SkipReason : uint8_t {
+    NoReason,
```
rsmmr: I'd add `Unknown` .

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -2,13 +2,89 @@
...
+    BadProtoHeader,
+    BadChecksum,
+};
+
+/**
+ * Additional information passed to TapAnalyzer::Packet()
+ */
+struct PacketTapInfo {
+    SkipReason skip_reason = SkipReason::NoReason;
+};
```
rsmmr: I think I'd remove this structure and just pass the reason directly. I assume the structure is for potential future extensions but I can't imagine we'd really add much further here, feels not worth it.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -94,9 +170,51 @@ class SessionAdapter : public analyzer::Analyzer {
...
+
+    /**
+     * Overridden from parent class, calling UpdateConnVal() on TapAnalyzer instances, too.
+     */
+    void UpdateConnVal(RecordVal* conn_val) override;
+
 protected:
     IPBasedAnalyzer* parent = nullptr;
     analyzer::pia::PIA* pia = nullptr;
+    std::list<TapAnalyzerPtr> tap_analyzers;
```
rsmmr: The rule of thumb I have in my head is a `vector` is almost always more efficient than a `list` if the container remains small.  I assume the `list` here is because of the remove operation? Even there I think that rule still applies. Plus, remove will be rare anyways. So I'd switch to `vector`.

awelzel: Ah, yep. We've recently replaced a `std::set` (48 bytes) with a `std::forward_list` (8 bytes) elsewhere for a bit of memory savings, but I realize `std::list` and std::vector` are both 24 bytes on my system.

And yeah, with probably three entries max or so I'm not worried about the add/remove operations.

awelzel: Ah, and `children` on `Analyzer` is also a `std::list<Analyzer*>`, I guess that was part of it as well.

src/packet_analysis/protocol/ip/SessionAdapter.h
awelzel: With the `TapPacket(pkt, verdict, skip_reason)` API, I went for `None` which fits better for `verdict=Deliver`. I'm not sure someone should ever pass `Unknown`, I'd rather they add a new enum value then :-)

src/packet_analysis/protocol/ip/SessionAdapter.h
awelzel: Agree.

src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc
```diff
@@ -105,9 +106,11 @@ bool IPBasedAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pkt
     run_state::current_timestamp = run_state::processing_start_time;
     run_state::current_pkt = pkt;
 
-    // TODO: Does this actually mean anything?
-    if ( conn->GetSessionAdapter()->Skipping() )
+    const auto* adapter = conn->GetSessionAdapter();
+    if ( adapter->Skipping() ) {
+        adapter->TapPacket(pkt, TapVerdict::Skip, SkipReason::Skipping);
```
awelzel: This is an additional "tapping place": When `skip_further_processing()` is called for a connection, TapAnalyzer instances will still see the packets with skip_reason=Skipping.

awelzel: Eh, I'll rename this to reason=SkipProcessing, but up for other ideas, too.

src/packet_analysis/protocol/ip/SessionAdapter.h
rsmmr: Think plugin: it cannot change the enum. But minor, ok to leave as is if you prefer .

awelzel: Yep, let me add `Unknown`, I read the comment as replacing `NoReason`  with `Unknown`.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -2,13 +2,90 @@
...
 class PIA;
 }
 
-namespace zeek::packet_analysis::IP {
+namespace zeek::packet_analysis {
+
+/**
+ * Indicator for a TapAnalyzer to determine what will happen to a packet.
+ */
+enum class TapVerdict : uint8_t {
```
timwoj: Drive-by comment here. How about `TapAction`?

awelzel: As in DM, going for `PacketAction`

PR #4729, opened 2025-08-04
Fix record coercion with compatible types

### LINKED ISSUE #4722: Vectors of records with an inner `int` type fail when adding elements
This is pretty confusing to me. This may be a usage error, but it feels buggy.

Given the [example vector code](https://docs.zeek.org/en/master/script-reference/types.html#id17) in the docs:

```
type MyRec: record {
    a: count &optional;
    b: count;
};

type MyVec: vector of MyRec;

global v2 = MyVec([$b=1], [$b=2], [$b=3]);
```

If we change `b` to `int`, it won't compile - even though it's basically the same program:

```
$ zeek test2.zeek
error in ./test2.zeek, lines 1-3 and ./test2.zeek, line 8: incompatible record types (MyRec and [$b=1])
error in ./test2.zeek, line 8 and ./test2.zeek, lines 1-3: type mismatch ([$b=1] and MyRec)
error in ./test2.zeek, line 8: inconsistent types in vector constructor (vector([$b=1], [$b=2], [$b=3]))
```

which... just seems wrong.

Note that we can use the exact some instruction to construct both in its own variable, then append that - so this works for both:

```
type MyRec: record {
	a: count &optional;
	b: count;
};

type MyVec: vector of MyRec;

global one: MyRec = [$b=1];
global two: MyRec = [$b=2];
global three: MyRec = [$b=3];

global v2 = MyVec(one, two, three);
```

might also be able to get simplified further

evantypanski:
Fixes #4722

Before, some cases were only allowed if the types were `same_type` - this loosens it for arithmetic. But, it does still error when needed because of later type checking (eg double->count conversions)


timwoj:
This could be backported to 7.0 as well.

vpax:
> This could be backported to 7.0 as well.

FWIW, doesn't seem worth it to me. It's fixing an annoyance, but one that (1) evidently doesn't occur a lot in practice, given it took a long time to get an issue on it, and (2) it can be worked around if need be.

evantypanski:
> (1) evidently doesn't occur a lot in practice, given it took a long time to get an issue on it

Yeah, I only found it when messing around with other stuff, not even real code.

Though, it's a really simple backport since that code hasn't been touched in 2 years, and that was just formatting. So ¯\\\_(ツ)_/¯

PR #4733, opened 2025-08-05
Remove deprecations tagged for v8.1

timwoj:
This removes all of the deprecations tagged for v8.1, as per usual after a x.0 version drops. There's nothing crazy in here, but I want to point out two specific items:

- The management framework has a lot of `Broker::` calls in it. We should probably make a concerted effort to move all of that to the cluster framework.
- The [scripts/policy/protocols/http/test-sql-injection-regex.zeek](https://github.com/zeek/zeek/compare/topic/timw/remove-8.1-deprecations?expand=1#diff-6cf478047158840425f7becd764515657c98198019b8b85f7cb5e36f17d9daf1) btest had one line in it that wasn't being detected. I moved it to the commented section in the test temporarily since the regex is very hairy and it didn't fit to work on it here. I'll open a separate issue to investigate.


src/EventHandler.cc
```diff
@@ -58,40 +56,6 @@ void EventHandler::Call(Args* vl, bool no_remote, double ts) {
...
-                        broker_mgr->PublishEvent(topic, Name(), ev_args, ts);
-                    else {
-                        broker_mgr->PublishEvent(topic, Name(), std::move(ev_args), ts);
-                        break;
-                    }
-                }
-            }
-        }
-    }
-
```
awelzel, line 94: Yay! Lets hope no one complains!

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1352,6 +1352,8 @@ bool DNS_Interpreter::ParseRR_BINDS(detail::DNS_MsgInfo* msg, const u_char*& dat
                                    msg->BuildBINDS_Val(&binds));
     }
 
+    delete completeflag;
```
awelzel: Maybe make  `complete_flag` in `BINDS_DATA` a `uint8_t` instead? Then in line 1349 we could store just the one byte and avoid passing down the `String*` instance as it has been done previously. Then could use a unique_ptr here for automatic release.

I was wondering if there was a leak, but we covertly transferred ownership of the `String*` instance into the `dns_bind_rr` before, right? :cold_sweat:

src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc
```diff
@@ -35,15 +35,7 @@ bool IPBasedAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pkt
...
-    // Deprecated: remove ConnTuple use in 8.1 and only use InitConnKey().
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-    ConnTuple tuple;
-    if ( BuildConnTuple(len, data, pkt, tuple) ) {
-        key->InitTuple(tuple.src_addr, tuple.src_port, tuple.dst_addr, tuple.dst_port, pkt->proto);
-#pragma GCC diagnostic pop
-    }
-    else if ( ! InitConnKey(len, data, pkt, *key) ) {
+    if ( ! InitConnKey(len, data, pkt, *key) ) {
```
awelzel:
```suggestion
    if ( ! InitConnKey(len, data, pkt, *key) )
```

src/scan.l
```diff
@@ -661,18 +661,6 @@ F	RET_CONST(zeek::val_mgr->False()->Ref())
 
 "0x"{HEX}+	RET_CONST(zeek::val_mgr->Count(static_cast<zeek_uint_t>(strtoull(yytext, 0, 16))).release())
 
-({H}".")+{HTLD} {
-	zeek::TableValPtr result;
-	std::string msg = zeek::util::fmt("Remove in v8.1: DNS lookup of host literal '%s' is deprecated. "
-	                                  "Replace with blocking_lookup_hostname().", yytext);
```
awelzel, line 667: Think we should keep the message, but convert to a reporter->Error(). Assuming someone missed the deprecation, the explicit message about host literals is more helpful than a generic lex error.

testing/btest/cluster/websocket/one-ipv6-deprecated.zeek
```diff
@@ -1,4 +1,4 @@
-# @TEST-DOC: Use listen_host to listen on an IPv6 address, otherwise same as one-ipv6.zeek
+# @TEST-DOC: Use listen_addr to listen on an IPv6 address, otherwise same as one-ipv6.zeek
```
awelzel: Seems okay to drop the whole test, given `one-ipv6.zeek` does the same.

testing/btest/core/hostname-literal-resolve.zeek
```diff
@@ -8,4 +8,4 @@
 # @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff err
 # @TEST-EXEC: btest-diff out
 
-print "dns.example.com", dns.example.com;
+print "dns.example.com", blocking_lookup_hostname("dns.example.com");
```
awelzel: Seems okay to drop the test named `hostname-literal-resolve.zeek`. blocking_lookup_host is tested in `testing/btest/dns_mgr/blocking_lookup_hostname.zeek`

src/analyzer/protocol/dns/DNS.cc
timwoj: I don't see any reason to involve the string at all. The field in script land is a `count`. The structure can just hold a uint8_t, read that one byte from the stream, and pass it along as that byte value.

PR #4738, opened 2025-08-06
Deprecate WriterBackend heartbeats

awelzel:
Relates to #4731.

This PR introduces a new `WriterBackend` constructor that disables threading heartbeats to the child threads. The motivation is that none of our built-in log writers need this. Sending the heartbeats show as significant overhead when reading sparse-multi-day PCAPs (think 1 packet every second sparse). The logging manager sends heartbeats to every thread every second.

As a macro tests, for the external test suite, this change results in significant reduction in user (~174s to ~161s) and system (~49s to ~27s) time. Elapsed time went down ~14sec on my system (from 107sec). Note that we load most every policy script when running `external/zeek-testing`.

```
$ /usr/bin/time taskset -c 0,1,2,3 btest -j -d
all 4 tests successful
174.13user 48.62system 1:47.52elapsed 207%CPU (0avgtext+0avgdata 164784maxresident)k
0inputs+3388016outputs (188major+2166682minor)pagefaults 0swaps
```
```
$ /usr/bin/time taskset -c 0,1,2,3 btest -j -d
all 4 tests successful
161.15user 26.67system 1:33.71elapsed 200%CPU (0avgtext+0avgdata 164392maxresident)k
0inputs+3373448outputs (174major+2166609minor)pagefaults 0swaps
```

For only the `m57-long` PCAP with default scripts, total runtime improves from 6.2s to 5.4s.
```
|:---|---:|---:|---:|---:|
| `/opt/zeek-dev-prod/bin/zeek -C -r /media/awelzel/T7/pcaps/2009-M57-day11-18.trace` | 6.214 ± 0.152 | 6.055 | 6.357 | 1.15 ± 0.03 |
| `zeek -C -r /media/awelzel/T7/pcaps/2009-M57-day11-18.trace` | 5.406 ± 0.046 | 5.353 | 5.436 | 1.00 |
```

IMO, also a 8.0 candidate given the significant perf impact for offline PCAP processing, but not too invested myself here.


src/logging/WriterBackend.h
```diff
@@ -38,9 +38,28 @@ class WriterBackend : public threading::MsgThread {
      * frontend, it's running in a different thread.
      *
      * @param name A descriptive name for writer's type (e.g., \c Ascii).
+     */
+    [[deprecated(
+        "Remove in v9.1: Use WriterBackend(NoThreadingHeartbeats{}, frontend). If you have a use for heartbeats in "
```
awelzel: Does that seem a reasonable deprecation path?

I found that [zeek-kafka](https://github.com/SeisoLLC/zeek-kafka/blob/0a0c44267b17ad274fd7a2af780739dbaa0b089f/src/KafkaWriter.cc#L304-L309) does call `producer->poll()`, but I'd think that should better from a thread started by that thread using real-time, rather than network-time.

timwoj: This feels a bit overengineered to me. What's wrong with just leaving this constructor with the second argument defaulted to `false`, and if others want the heartbeats they can just pass true to the constructor? The only problem I see is that it doesn't give a nice deprecation warning about it.

awelzel: > The only problem I see is that it doesn't give a nice deprecation warning about it.

That was the main motivation of the constructor tagging: Defaulting to `false` would change behavior possibly without a plugin author or user realizing. I'm not sure how much `zeek-kafka` really relies on the `DoHeartBeat()` call, but was too timid to simply stop invoking it by default.

I could see keeping the default as `true` and adapting Zeek's internal writers to pass `false` explicitly. That didn't force moving to a better place to me enough at the time (which the deprecation warning does), but I'm certainly okay with that if the constructor tagging seems like overdoing it.

awelzel: @timwoj - I have reworked this to just add a new constructor parameter that's defaulted to true though to not break existing writers. Mind taking another look?

awelzel:
> As a macro tests, for the external test suite, this change results in significant reduction in user (~174s to ~161s) and system (~49s to ~27s) time. Elapsed time went down ~14sec on my system (from 107sec).

Could we consider this for 8.0.2? For crunching through long/sparse PCAPs, this is a significant reduction in user and system CPU time for a not-too-unusual use-case IMO.

timwoj (approved):
Yeah, I like that version more. As for 8.0.2, add it to the list. I'm not finished with the branch prep yet anyways.

PR #4737, opened 2025-08-06
btest/plugins/tap-analyzer: Update baseline

awelzel:
This time running CI which was skipped in #4734

PR #4739, opened 2025-08-06
More TapAnalyzer updates for UpdateConnVal fixes and tests

awelzel:
One more follow-up for the TapAnalyzer feature, including a new test for the `UpdateConnVal()`.

@timwoj - mind looking at some of the code moving in the session adapters, particularly TCP. Nothing fell apart, but I wonder is there's some underlying contract with `pkt->session` I might be tickling here?

Relates to #4337 #4725 #4734 #4737


timwoj (approved):
The session stuff looks fine to me. It's a little weird that the session gets attached the packet before it's processed, but it likely doesn't matter as long as the session is valid.

PR #4741, opened 2025-08-08
Use virtualenv in docs generation/builds

timwoj:
Docs builds started failing after the changes in https://github.com/zeek/zeek-docs/pull/340 were merged due to a conflict between a package we're installing vs one that was installed via a debian package. This PR changes the docs builds to use a python virtualenv with the requirements installed.


.github/workflows/generate-docs.yml
```diff
@@ -66,16 +66,14 @@ jobs:
...
-          # Many distros adhere to PEP 394's recommendation for `python` =
-          # `python2` so this is a simple workaround until we drop Python 2
-          # support and explicitly use `python3` for all invocations.
-          sudo ln -sf /usr/bin/python3 /usr/local/bin/python
-          sudo pip3 install --break-system-packages -r doc/requirements.txt
-          sudo pip3 install --break-system-packages pre-commit
+          python3 -m venv ci-docs-venv
+          source ci-docs-venv/bin/activate
+          pip3 install --break-system-packages -r doc/requirements.txt
+          pip3 install --break-system-packages pre-commit
```
awelzel: Should allow dropping `--break-system-packages` with a virtual env.

Do you know why CI was skipped?

EDIT: Because you put the skip all label on it :-)

timwoj: Whoops, yeah that can go.

I skipped Cirrus because there's no need to run it for this PR.

EDIT: I also forgot that we disable docs builds with that label. Oops. I removed it and restarted a docs build, but cancelled the Cirrus jobs manually.

PR #4744, opened 2025-08-08
smb2/read: Parse only 1 byte for data_offset, ignore reserved1

### LINKED ISSUE #4730: SMB2 - ExceptionOutOfBound in SMB2_read_response:pad
I encountered strange behavior while parsing smb2 protocol:
`Binpac exception: binpac exception: out_of_bound: SMB2_read_response:pad: 272 > -67"`
I have pcap file 
[smb2.zip](https://github.com/user-attachments/files/21572924/smb2.zip)
An exception occurs while parsing packet 54
image

image

It seems to me that this packet may be incorrect, but it is not clear why end of data is negative, and why it goes to this place if it is initially incorrect, maybe the problem is in the type casting
I want to exclude an error in calculations in zeek
Or at least add a check for the correctness of the calculation of the position of the pointer to the end of the data with the corresponding exception

awelzel:
A user provided a SMB2 pcap with the reserved1 field of a ReadResponse set to 1 instead of 0. This confused the padding computation due to including this byte into the offset. Properly split data_offset and reserved1 into individual byte fields.

Closes #4730


awelzel:
@timwoj , putting this on 7.0.x and would be good to pick up into 8.0-rc2 IMO.

timwoj (approved):

timwoj:
I cherry-picked it into the 8.0 branch, and I see that it's on the list for a later 7.0 release as well.

Issue #4719, opened 2025-07-31
Move a number of submodules into main repository


timwoj:
A few of our existing submodules are side projects to Zeek. Historically there was probably hope that people would use these projects for things outside of Zeek, making external improvements that we could then pull back in. This hasn't played out that way, and keeping them as submodules just adds maintenance complexity when dealing with PRs, updating branches in the submodules, etc. I'm proposing that we move a number of them into the Zeek repo directly, instead of being submodules, and archive the external repositories. We've already stopped making individual releases for all of these. The list I'm considering is:

- binpac
- bifcl
- zeek-aux

I would argue against moving btest, zeekctl, and zkg, since at the very least btest and zkg have python packages built from them uploaded to pypi, and it seems like it would be easier to leave them as separate builds/projects. I would also skip broker since we definitely develop it like a separate project sometimes.

I could also see arguments for moving zeek-af_packet-plugin and zeekjs internally and archiving the external repos. I'm not sure how this would map with removing the packages from our package site.

@rsmmr also mentioned recently his desire to stop having zeek-docs be a separate repository and to move it to be part of the primary Zeek repo. That could happen as part of this work as well.

awelzel:
Would also add AF_PACKET and JavaScript. The former should be non-contentious, the latter was discussed as being forked from `github.com/corelight` to `github.com/zeek`, but I'd just merge the code into the main tree. There's ZAM and experimental C++ compilation in the main tree, too, and would argue JavaScript support has been less intrusive 🤷‍♂️

timwoj:
This would also reduce the maintenance burden with the CMake repository, since we could point them at the one that's in the main repo. We currently have to update all of the subprojects every time we make a change to the CMake repo, then flow those updated ones along to the Zeek repo.

timwoj:
> Would also add AF_PACKET and JavaScript. The former should be non-contentious, the latter was discussed as being forked from github.com/corelight to github.com/zeek, but I'd just merge the code into the main tree. 

I think the only real question about either of them is whether we continue supporting them as external plugins, remove them from the packages site, etc.

timwoj:
The `zeek-aux` directory includes a bunch of btests for the code there. I'm intending to move them to be part of the Zeek test suite, under a new `zeek-aux` directory. Binpac and bifcl don't have any integrated tests of their own.

timwoj:
[...]

PR #4746, opened 2025-08-11
Move BinPAC, bifcl, af_packet, and gen_zam submodules into main zeek repo

timwoj:
This PR moves the BinPAC, bifcl, and zeek-aux code into the main repository directly instead of using git submodules for them. This is intended to reduce the amount of commit shuffling that is required to update any of this code. It continues building all of the related code as part of Zeek's build, but uses Zeek's CMake setup and submodule to do so instead of continuing to keep extra copies of that. It also moves the btests that were part of zeek-aux into the main testing/btest directory so they run as part of Zeek's testing.

One this is merged, the repositories for these submodules will be marked as archives with a note in their READMEs. All further work needed can be done directly in the Zeek repo instead.

NOTES FOR REVIEWERS: The code hasn't changed much at all for the submodules themselves. You can probably skip over scrolling through them or just give them a cursory skim. The major changes in the directories themselves are in removing a lot of unnecessary files (git files, tooling configs, etc). The things to review are the CMakeLists.txt files in those directories and anything of the zeek-aux btests.

Part of #4719


awelzel:
High-level thoughts:

* Is reusing the same location (auxil/binpac, auxil/bifcl) clunky/rough when switching between `release-8.0` and `master` branches until 9.0 is out? Have some memories of confusion for myself and users building from different branches when the spicy-plugin / spicy submodules were moved.

* I could actually see `src/bifcl` and `src/binpac` just as well given there's `src/spicy/spicyz` which has a close relation.

* Maybe include the hashes of the submodules in the commits that import them, minimally for reference.

* Might have skipped on zeek-aux for this PR due to itself being a conglomeration of various projects, but no strong opinion.

timwoj:
> Is reusing the same location (auxil/binpac, auxil/bifcl) clunky/rough when switching between release-8.0 and master branches until 9.0 is out? Have some memories of confusion for myself and users building from different branches when the spicy-plugin / spicy submodules were moved.

It is sorta painful to switch back and forth. Moving it to another directory would potentially fix that.

> I could actually see src/bifcl and src/binpac just as well given there's src/spicy/spicyz which has a close relation

I considered that actually. I opted to leave it where it is to minimize the necessary CMake changes, but it probably makes more sense for it to be over there. Or maybe add a `tools` directory at the top-level? I think the `src/spicy` directory is different because there's other code in there beyond just `spicyz`.

> Might have skipped on zeek-aux for this PR due to itself being a conglomeration of various projects, but no strong opinion.

It's also the majority of the changes here. I'll port all of those changes back into the `zeek-aux` repo, drop it from this PR, and open a separate one to bring it into Zeek and we can discuss it there. The `tools` directory above would also be a good place for it. I could also see breaking up that repo into it's component parts inside of that tools dir.

EDIT: Let's discuss this in the ticket instead of here.

timwoj:
Later versions of this PR dropped `zeek-aux` from it and added `gen-zam` and `zeek-af_packet-plugin`. We had planned for ZeekJS to get into this PR also, but it turned out to be somewhat complicated and it was better to leave it out for now.

scripts/base/init-bare.zeek
```diff
@@ -6420,6 +6420,31 @@ export {
 	const generic_packet_thresholds: set[count] = {} &redef;
 }
 
+module AF_Packet;
```
awelzel: Nit. Maybe move closer to `module Pcap` given their close relationship.

src/iosource/af_packet/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+if (${CMAKE_SYSTEM_NAME} MATCHES Linux)
+    set(ZEEK_HAVE_AF_PACKET yes CACHE INTERNAL "")
+
+    include(CheckSymbolExists)
+
+    check_symbol_exists(TP_STATUS_CSUM_VALID linux/if_packet.h HAVE_TP_STATUS_CSUM_VALID)
```
awelzel: This was added sometime in 2015, do you think we can just ignore that?

src/iosource/af_packet/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+    if (NOT HAVE_TP_STATUS_CSUM_VALID)
+        message(STATUS "Checksum offloading to the kernel might not be fully supported.")
+    endif ()
+
+    zeek_add_plugin(
+        Zeek AF_Packet SOURCES src/Plugin.cc src/AF_Packet.cc src/RX_Ring.cc
+        BIFS src/af_packet.bif DIST_FILES zeekctl/af_packet.py SCRIPTS scripts/init.zeek)
```
awelzel: Hmm, zeekctl already ships af_packet packet since a while. Seems reasonable t o drop it from the main-tree.

src/iosource/af_packet/src/Plugin.h
```diff
@@ -0,0 +1,17 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#pragma once
```
awelzel: Do we need this file at all? I see in the storage framework you have some `Plugin.cc` only directories. I could see this here, too, but might as well deserve a broader discussion.

src/iosource/af_packet/scripts/init.zeek
```diff
@@ -0,0 +1,28 @@
+##! Packet source using AF_Packet.
```
awelzel: Can this be dropped? It's been moved to init-bare. Maybe drop all the scripts?

src/iosource/af_packet/zeekctl/af_packet.py
```diff
@@ -0,0 +1,53 @@
+import ZeekControl.plugin
```
awelzel: Drop? It's part of zeekctl anyhow.

awelzel (approved):
I'm a fan! Maybe one more commit to thin out the result a bit?

src/iosource/af_packet/src/Plugin.h
timwoj: iosource/pcap only has Plugin.cc also. I'll drop this out.

ckreibich:
Speed demons! :-) I was in mid-review when this got merged — I have two main comments:

I don't think we should lose the commit history. The reference to the commit in the other repo is helpful, but with time that will get brittle. There's a pretty established way for doing this via `filter-repo` to move the original content to a subdirectory, and then merging that into this one from a new origin. We also took care to preserve the gen-zam history when we moved it out, so we're now cutting that off. We would also pretty certainly preserve the history if/when we merge the docs into this repo, no?

Regarding AF_PACKET, it's a bit unfortunate to have the namespace and corresponding variables in `init-bare` even when the plugin isn't built in, which is the case when we're not on Linux. I'd agree that there are benefits to having the AF_PACKET tunables relatively accessible in there, but we don't have, say, Redis stuff in init-bare either.

For both of these we can iterate over what's now there, I think. No need to back anything out, but thoughts on the above?

timwoj:
I think if bringing in the commit history as the first step is desired, we should probably revert what's in `master` and rework the commits in the branch. My only concern with that is the number of commits that involves. BinPAC itself is ~450 commits. Add the other three in and it's about 800. Do we want to add all of those commits to the Zeek tree?

> it's a bit unfortunate to have the namespace and corresponding variables in init-bare even when the plugin isn't built in

I agree. The problem I ran into is that we want the BIF enums available from C++ but also available from early in script-land. The easiest way to do that is to put the enums in types.bif (which is part of your concern) and to put the script bits in init-bare.zeek (which is the other part). The only existing iosource plugin is pcap and it has variables in init-bare as well.

ckreibich:
>  Do we want to add all of those commits to the Zeek tree?

Strong yes from me: if we pull our own projects into our tree, the effect should be as if that development had happened in that tree. I think that holds even for bifcl and binpac even though they see little to no development today.

> The only existing iosource plugin is pcap and it has variables in init-bare as well.

Yeah, but the difference is that we never build without pcap. Fwiw I do think this is minor overall and easier to iterate on.

PR #4748, opened 2025-08-13
Add Debian 13.0 (trixie)

awelzel:
Eh, I forgot this one for Zeek 8.0. If we could use Debian 13.0 for the container images for Zeek 8.0, that'd be nice IMO.

I'm also wondering why we disabled Spicy on the Debian ARM task, was that resource usage trouble back then? I'm tempted to just turn it on.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -796,8 +796,8 @@ zeekctl_debian12_task:
         $CIRRUS_BRANCH =~ 'release/.*' )
     )
   container:
-    # Debian 12 (bookworm) EOL: TBD
-    dockerfile: ci/debian-12/Dockerfile
+    # Debian 13 (trixie) EOL: TBD
+    dockerfile: ci/debian-13/Dockerfile
     << : *RESOURCES_TEMPLATE
   sync_submodules_script: git submodule update --recursive --init
   always:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -142,6 +142,7 @@ ci_template: &CI_TEMPLATE
 
 env:
   CIRRUS_WORKING_DIR: /zeek
+  CIRRUS_LOG_TIMESTAMP: true
```
awelzel, line 145: Thought that'd be nice and makes it easier to skim which parts take how much time.

ci/debian-13/Dockerfile
```diff
@@ -1,32 +1,36 @@
-FROM debian:11
+FROM debian:trixie
```
timwoj: Do we want to call this `debian:trixie`? Are `debian:13`-tagged images available? I'm just looking at consistency with the previous debian Dockerfiles as well as the ubuntu ones.

timwoj (approved):
One minor question re: the base debian image, nothing that would hold up merging.

ci/debian-13/Dockerfile
awelzel: Ah, interesting. I've copied the `debian-12/Dockerfile` and that one had bookworm in it, but git now shows it as rename :-)

I left the `debian-12/Dockerfile` alone, but bumped zeekctl and builtin tests also to Debian 13.0 in the last push :crossed_fingers:

timwoj: Oh weird, it does. Using the named-version is fine then 👍🏼

awelzel: I'm going to leave the number now, because this is also in `ci/debian-13/Dockerfile`, so that looks better :-)

PR #4750, opened 2025-08-13
Changed behavior of var-extraction-uri.zeek from policy/protocol/http…

blightzero:
… to extract only the pURI parameter names. Do not include the path in the first parameter name. Only extract uri vars if parameters actually exist.

As per chat with @bbannier:

I was looking at the https://github.com/zeek/zeek/blob/master/scripts/policy/protocols/http/var-extraction-uri.zeek file because I was interested in getting HTTP URI variables extracted. However, the output of that module seemed odd. With the following URI `/ws-broker?action=PATCH&target=/users/1/settings `  what I would expect (as a JSON log output) is an `["action","target"]`  list, but the actual output was `[/ws-broker?action, target]` and for a URI that has no parameters/variables I would expect an empty [] list or a null/no value. But instead you will get the URI path again. This changes the behavior of this script to better match the expected output of actually extracting the parameter(uri var) names.


testing/btest/scripts/policy/protocols/http/var-extraction-uri.zeek
```diff
@@ -0,0 +1,5 @@
+# @TEST-EXEC: zeek -r ${TRACES}/http/http-filename.pcap %INPUT
+# @TEST-EXEC: zeek-cut uri_vars <http.log > http-reduced.log
```
awelzel: Mind including the uri also in the baseline? Content-Type and Content-Disposition are usually headers, so this is a pretty curious example. Looking at the pcap it makes sense, however:
```
GET /response-headers?Content-Type=application/octet-stream;%20charset=UTF-8&Content-Disposition=attachment;%20filename%3d%22test.json%22 HTTP/1.1
```

```suggestion
# @TEST-EXEC: zeek-cut uri uri_vars <http.log > http-reduced.log
```

testing/external/commit-hash.zeek-testing
bbannier, line 1: This needs https://github.com/zeek/zeek-testing/pull/24.

testing/external/commit-hash.zeek-testing-private
bbannier, line 1: This needs https://github.com/zeek/zeek-testing-private/pull/28.

PR #4757, opened 2025-08-15
RFC: management/controller: Add dedicated tls options

awelzel:
@ckreibich - could you see this, or would you prefer a different direction?

Previously I believe the TLS options were inherited from Broker, but with Cluster::listen_websocket(), these can now be configured per WebSocket server instance.

cc @dopheide-esnet

### DIFF
--- a/scripts/policy/frameworks/management/controller/config.zeek
+++ b/scripts/policy/frameworks/management/controller/config.zeek
@@ -61,6 +61,17 @@ export {
 	## for websocket clients.
 	const default_port_websocket = 2149/tcp &redef;
 
+	# TLS options to use for the controller's WebSocket server.
+	#
+	# The default value is Broker's cert-less TLS config. To disable
+	# TLS, redef this value to ``Cluster::WebSocketTLSOptions()``.
+	const tls_options_websocket = Cluster::WebSocketTLSOptions(
+		$cert_file="",
+		$key_file="",
+		$ca_file="NONE",
+		$ciphers="AECDH-AES256-SHA@SECLEVEL=0:AECDH-AES256-SHA:P-384",
+	) &redef;
+
 	## Whether the controller should auto-assign Broker listening ports to
 	## cluster nodes that need them and don't have them explicitly specified
 	## in cluster configurations.
--- a/scripts/policy/frameworks/management/controller/main.zeek
+++ b/scripts/policy/frameworks/management/controller/main.zeek
@@ -1646,7 +1646,8 @@ event zeek_init()
 	if ( cni$bound_port != 0/unknown )
 		{
 		local ws_opts = Cluster::WebSocketServerOptions($listen_addr=to_addr(cni$address),
-		                                                $listen_port=cni$bound_port);
+		                                                $listen_port=cni$bound_port,
+		                                                $tls_options=Management::Controller::tls_options_websocket);
 		Cluster::listen_websocket(ws_opts);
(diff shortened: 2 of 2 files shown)


ckreibich:
> Previously I believe the TLS options were inherited from Broker

That's right, with [corresponding settings](https://github.com/zeek/zeek-client/blob/master/zeekclient/config.py#L68-L90) over in zeek-client. Basically, zeek-client had to go out of its way to match Broker's default mode.

I like your approach of a redef'able TLS config and tried it out. I noticed a few things:

First, and this doesn't relate to this changeset, zeek-testing-cluster needs updates for local dev work. It still doesn't provide ZeroMQ in its container, and still used Debian 11. Easy to fix.

More importantly, since websockets are now unencrypted by default, I wonder whether it wouldn't be nicer to adopt that stance rather than force the webserver back to it. The TLS config default can just become `const tls_options_websocket = Cluster::WebSocketTLSOptions() &redef;` and corresponding change in zeek-client is also small.

Those changes makes all but one zeek-testing-cluster tests pass again — `deploy-a2c-authenticated`, which tests the "properly" TLS-protected setting. I hit a roadblock there: it looks like we (or rather, IXWebSocket) don't support keys encrypted with a passphrase, right? That test is using those. I was looking for an equivalent of `Broker::ssl_passphrase`. I think it's fine to skip this for now, and it was easy to make that test use variants of the cert files that don't use a passphrase (they are already in the repo). Do you have thoughts on this? I read somewhere that in order to support passphrases for IXWebSocket we'd basically need to do the decryption manually. Hrrm.

I'd be fine doing all this for an 8.0.1. We'd change behavior in zeek-client, and it would need a version bump. But I don't think it's a big deal at this stage, and users could still configure it to use the Broker default posture should they need that.

I put all of this together in [a branch](https://github.com/zeek/zeek/tree/topic/christian/management-framework-tls) if you want to take a look. zeek-testing-cluster runs cleanly with it over here.

awelzel:
> I'd be fine doing all this for an 8.0.1.

8.0 isn't affected as that still uses `Broker::listen_websocket()`. This is fallout from the deprecation removals including replacing `Broker::listen_websocket()` with `Cluster::listen_websocket()` for 8.1.

> I hit a roadblock there: it looks like we (or rather, IXWebSocket) don't support keys encrypted with a passphrase, right? [...]  Do you have thoughts on this? I read somewhere that in order to support passphrases for IXWebSocket we'd basically need to do the decryption manually. Hrrm.

Eh. My thoughts are mainly around whether it'd be a big deal to deprecate TLS and ask users to put nginx/envoy/caddy/haproxy in front. But I'm afraid that won't fly.

Not sure, maybe passphrase support for IXWebSocket - they support various platforms and TLS libraries outside of OpenSSL and not sure they'd be happy with a OpenSSL only PR, but it's worth a shot I guess.

ckreibich:
> 8.0 isn't affected 

And I had really wondered how I missed that, phew! :-) I think the path forward is pretty clear then. I think it's fine to drop passphrase support for the client for now, bring it back if/when it lands in IXWebSocket, and switch the client over to unencrypted-by-default at this point. I'll put up the PRs. Feel free to close this Arne.

awelzel:
> bring it back if/when it lands in IXWebSocket

There's actually already a ticket: https://github.com/machinezone/IXWebSocket/issues/519

> I'll put up the PRs. Feel free to close this Arne.

Closing. Thanks!

PR #4759, opened 2025-08-16
Report PostgreSQL login success only after ReadyForQuery

AmazingPP:
Previously, Zeek treated the receipt of `AuthenticationOk` as a successful login. However, according to the PostgreSQL Frontend/Backend Protocol, the startup phase is not complete until the server sends `ReadyForQuery`. It is still possible for the server to emit an `ErrorResponse` (e.g. ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION) after `AuthenticationOk` but before `ReadyForQuery`.

This change updates the PostgreSQL analyzer to defer reporting login success until `ReadyForQuery` is observed. This prevents false positives in cases where authentication succeeds but session startup fails.

reference:
https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-START-UP


scripts/base/protocols/postgresql/main.zeek
```diff
@@ -236,7 +234,8 @@ event PostgreSQL::ready_for_query(c: connection, transaction_status: string) {
 	if ( ! c$postgresql?$success )
 		c$postgresql$success = transaction_status == "I" || transaction_status == "T";
 
-	c$postgresql$rows = c$postgresql_state$rows;
+	if ( c$postgresql?$frontend && c$postgresql$frontend == "simple_query" )
+		c$postgresql$rows = c$postgresql_state$rows;
```
awelzel, line 241: What would you think about making `rows`  `&optional` and remove `&default` in state, then changing this here to
```
if ( c$postgresql_state?$rows )
    {
    c$postgresql$rows = c$postgresql_state$rows;
    delete c$postgresql_state$rows;
    }
```

And making sure `data_row()` initializes it to `0` if not yet set, maybe even with a `weird` if observed unset given `simple_query()` should do so.

Maybe someday there's support for more than `simple_query` and then we don't need to touch this part again.

timwoj:
I'll hold it for 8.0.1. We're already adding more than I would like at the last minute for 8.0.0.

PR #4760, opened 2025-08-16
Fixes for "if" and initialization statement line numbers

vpax:
Work I've been doing on assessing testing code coverage for Zeek packages turned up that the line numbers associated with `if` statements were bound to the if/else components but not the actual `if` statement line. It turns out that there's already code in the parser to make the latter association (and not do the former), but (presumably very old) code in the `IfStmt` constructor was overriding this. In addition, I noticed that the internal `InitStmt` statement would associate the initialization with the first initializer, which I changed to be the range from the first to the last.

In the process, I also fixed up ZAM's tracking of BiFs, as some routine maintenance.


ckreibich:
> We should probably put something on the TODOs of the release tasks to include ZAM-bif-tracking maintentance

Yeah, it's not part of my mental checklist either. I'm also sometimes not fully clear on which of the attributes is the right one (despite their good commenting) — that might just be me though.

vpax:
> I'm also sometimes not fully clear on which of the attributes is the right one (despite their good commenting) — that might just be me though.

Feel free to tag me on any associated PRs and I can double-check the choice of attributes.

timwoj:
I wonder if it's possible via Cirrus to set an environment variable based on their `changesInclude` function. We could then add a script that would force that btest to run as part of PR builds where some bif file changes.

awelzel:
> I wonder if it's possible via Cirrus to set an environment variable based on their changesInclude function. We could then add a script that would force that btest to run as part of PR builds where some bif file changes.

Hmm, I pushed for not running that test so that new contributors adding BiFs aren't exposed to the details of the bif-tracking test, adding friction and burden to things that should be "easy" contributions. Also, nothing is actually broken when the attributes aren't up-to-date as "safe defaults" are in effect.

timwoj:
Maybe part of the `master` nightly builds then? At least then it fails somewhere.

vpax:
Well it fails whenever I do maintenance, because I have `$ZEEK_ZAM_MAINTENANCE` set. I think it's fine to just have this be a checklist item prior to releases. (It's not even that important if it's overlooked for a release, just less-than-tidy.)

PR #4761, opened 2025-08-17
Clarify Event constructor deprectation message

timwoj:
A [discussion in Slack](https://zeekorg.slack.com/archives/CTGDG83B8/p1755294736531799) brought up that this message could be a bit more clear about the action the user needs to take.


timwoj:
I forgot to merge this in before releasing 8.0. I added it to the list for 8.0.1.

PR #1808, opened 2021-09-29
Initial Support to DNS SVCB/HTTPS RR

FlyingWithJerome:
This PR tries to provide an initial support to DNS SVCB/HTTPS RR. SVCB/HTTPS resource records can be used to share pre-connection information such as port numbers and ECH public key, and alias zone apexes. The RFC draft is available [here](https://www.ietf.org/archive/id/draft-ietf-dnsop-svcb-https-07.txt).

This PR is "initial support" because it leaves `SvcParams` unparsed. I plan to get this done in the next PR.

PR #4767, opened 2025-08-18
Parse SVCB/HTTPS SvcParams list

klemens-ya:
Add full support for RFC 9460's SvcParams list.

Amend the existing `dns_svcb_rr` record by a vector of new
`dns_svcb_param` records containing aptly typed SvcParamKey and
SvcParamValue pairs.  Example output:

```
@load base/protocols/dns
event dns_HTTPS( c: connection , msg: dns_msg , ans: dns_answer , https: dns_svcb_rr ) {
        for (_, param in https$svc_params)
                print to_json(param);  # filter uninitialised values
}
```

```
$ dig https cloudflare-ech.com +short | tr [:space:] \\n
1
.
alpn="h3,h2"
ipv4hint=104.18.10.118,104.18.11.118
ech=AEX+DQBBHgAgACBGL2e9TiFwjK/w1Zg9AmRm7mgXHz3PjffP0mTFNMxmDQAEAAEAAQASY2xvdWRmbGFyZS1lY2guY29tAAA=
ipv6hint=2606:4700::6812:a76,2606:4700::6812:b76
```

```
{"key":1,"alpn":["h3","h2"]}
{"key":4,"hint":["104.18.10.118","104.18.11.118"]}
{"key":5,"ech":"AEX+DQBBHgAgACBGL2e9TiFwjK/w1Zg9AmRm7mgXHz3PjffP0mTFNMxmDQAEAAEAAQASY2xvdWRmbGFyZS1lY2guY29tAAA="}
{"key":6,"hint":["2606:4700::6812:a76","2606:4700::6812:b76"]}
```

Values with malformed data or belonging to invalid/reserved keys
are passed raw bytes in network order for script-level inspection.

Follow up to "Initial Support to DNS SVCB/HTTPS RR"
https://github.com/zeek/zeek/pull/1808


scripts/base/init-bare.zeek
```diff
@@ -3077,6 +3077,8 @@ type dns_loc_rr: record {
 type dns_svcb_rr: record {
 	svc_priority: count;	##< Service priority for the current record, 0 indicates that this record is in AliasMode and cannot carry svc_params; otherwise this is in ServiceMode, and may include svc_params
 	target_name: string;	##< Target name, the hostname of the service endpoint.
+	svc_params: string;	##< Whitespace-separated list with each SvcParam consisting of a SvcParamKey=SvcParamValue pair or a standalone SvcParamKey. SvcParamValue may contain comma-separated lists.
```
awelzel: @klemens-ya - thanks so much for your contribution!

What do you think about making this more like:

```
type dns_svcb_param: record {
    key: count;
    value: string;
};

type dns_svcb_rr: record {
   ...
   svc_params: vector of dns_svcb_param;
};
```

With the single string, if a user is interested in one particular svc param, they'll need to implement the `key=value` string parsing in Zeek script. Passing the key as count also side-steps needing the `key_names` array I believe?

So, essentially, if a user want the `key=value` form, they can construct it themselves, but if someone is searching for a specific param, a more structured format makes that easier IMO.

awelzel: Looking at the params, might make more sense to promote `value` to a vector of strings that can be optional to cover the various cases.
```
type dns_svcb_param: record {
    key: count;
    value: vector of string &optional;
};
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,147 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+                if ( value_len % addr_len != 0 ) {
+                    analyzer->Weird("invalid ipv4/6hint size");
+                    goto malformed;
+                }
+
+                while ( item_len_parsed + addr_len <= value_len ) {
+                    const size_t addr_sz = INET6_ADDRSTRLEN;
+                    char addr[addr_sz];
+
+                    if ( zeek_inet_ntop(is_ipv4 ? AF_INET : AF_INET6, data, addr, addr_sz) ) {
```
awelzel: Hmm, I wonder if we should pass the raw bytes to script land instead. `raw_bytes_to_v4_addr()` or `raw_bytes_to_v6_addr()` can then be used and the `malformed` part might also move to script-land instead of the parser. What do you think?

klemens-ya: Passing unprocessed bytes seems weird given their format is well known and Zeek has a first class data type for it.

Even for malformed or unknown data, I'm inclined to either parse or drop it and resort to, e.g. Wireshark to inspect such packets.

The RFC says:
> If any RRs are malformed, the client MUST reject the entire RRset and fall back to non-SVCB connection establishment.

So unless there's precedence of passing whatever Zeek got off the wire, I'd keep it simple and pass addresses.

scripts/base/init-bare.zeek
klemens-ya: > With the single string, if a user is interested in one particular svc param, they'll need to implement the `key=value` string parsing in Zeek script.
> ...

I'll structure the data, a string was just quick and easy to do before getting feedback and investing more time.

> Looking at the params, might make more sense to promote value to a vector of strings that can be optional to cover the various cases.

That makes sense and matches what I had in mind after discarding multiple members named after the parameters for being too complicated and inflexible.

> Passing the key as count also side-steps needing the `key_names` array I believe?
>
> 
> So, essentially, if a user want the `key=value` form, they can construct it themselves, but if someone is searching for a specific param, a more structured format makes that easier IMO.

Yes, I reckon, that's what [`svcparam_keys`](https://github.com/zeek/zeek/pull/4767/files#diff-63e20c048494ce44781c0a064e9f8c9530f35ce311f4eee9ef109632c40cb467R188-R196) is for.

src/analyzer/protocol/dns/DNS.cc
awelzel: > Zeek has a first class data type for it.

A user still needs to call `to_addr()` on the string to get an actual `addr` value though.

But yeah, lets take the `zeek_inet_ntop()` version and if there's ever somene/something trying to smuggle non-4-and-non-16-byte values in these hints, we could add `raw_value` to the param.

klemens-ya: > A user still needs to call `to_addr()` on the string to get an actual `addr` value though.

They'd also have to convert any raw value into the respective type before dealing with it in Zeek, no?
 
> But yeah, lets take the `zeek_inet_ntop()` version and if there's ever somene/something trying to smuggle non-4-and-non-16-byte values in these hints, we could add `raw_value` to the param.

[Misaligned values get skipped entirely](https://github.com/zeek/zeek/pull/4767/files#diff-1376c79b7686e098e8706aafee26bd42ddde9f1d20cb4ce48def47d5d5e18ea8R1697-R1700).

And if lengths match, data is still arbitrary and may not contain valid IP addresses, so my point is to validate and reject it as soon as possible, not leaving it up to the user.

klemens-ya: I guess, ideally we'd export all data properly typed according to the key:
```
type dns_svcb_param_value: record {
    keys:  vector of count  &optional; # mandatory
    ids:   vector of string &optional; # alpn
    port:  port             &optional; # port
    hints: vector of addr   &optional; # ipv4hint or ipv6hint
};

type dns_svcb_param: record {
    key: count;
    value: dns_svcb_param_value &optional;
};
```

Would that be feasible?

I still have to look how to pass anything non-trivial like that, the current string approach is a fortunate copy of `target_name`.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,147 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+        const u_char* value_start = data;
+        svc_params_len -= 2 + 2;
+
+        if ( value_len > svc_params_len ) {
+            analyzer->Weird("SvcParamValue overflows SvcParam");
             break;
-        default: break; // unreachable. for suppressing compiler warnings.
+        }
+
+        if ( key <= size(key_names) )
```
bbannier: This check is off by one and allows indexing beyond `key_names`.


```suggestion
        if ( key < key_names.size() )
```

src/analyzer/protocol/dns/DNS.cc
awelzel: > They'd also have to convert any raw value into the respective type before dealing with it in Zeek, no?

Correct. I don't think there's a strict right or wrong here.

> Misaligned values get skipped entirely.

Yeah, with the `->Weird()` it seems fine for now as there's a signal in `weird.log` when this happened.

> not leaving it up to the user.

I'm more coming from the direction that DNS has been (mis-)used for tunneling purposes and smuggling data in the wild. It can be nice to expose as much detail and raw data as possible to the script-layer, even if it's not exactly conforming to a spec. Sometimes the non-conformance is the most interesting part from a network monitoring and analysis perspective - kind of a Zeek dilemma :-)

I'd even go so far as suggesting the `default:` case below as param for script-land instead of skipping in a raw/generic fashion. With that, it'd be possible to add support for new  parameters without needing to change the parser's C++ code itself.

Feel free to skip and leave it as you have it though - just putting down where my comments are coming from here.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Thanks, as `key_names` won't stay, I'll probably change this to bail out `if ( key > 6 )` or so as per https://datatracker.ietf.org/doc/html/rfc9460#name-initial-contents

The initial version capped `key` to 7 and had `"reserved/invalid"` as last value so as to simplify a later `key_names[key]` print.

When that got yanked, the check remained.

src/analyzer/protocol/dns/DNS.cc
awelzel: > I guess, ideally we'd export all data properly typed according to the key:

I was thinking the same, but wasn't sure how much effort you want to spend here. If you're up for this, that'd be awesome (feel free to ask in #development on Slack if you have issues with constructing the values).

Could also add the `raw_value: string` here - again, totally fine to skip if you don't like the idea.

klemens-ya: > Sometimes the non-conformance is the most interesting part

Fair point, it is out of my scope, though, so far. 

> I'd even go so far as suggesting the `default:` case below as param for script-land instead of skipping in a raw/generic fashion. With that, it'd be possible to add support for new parameters without needing to change the parser's C++ code itself.
> 
> Feel free to skip and leave it as you have it though - just putting down where my comments are coming from here.

Much appreciated, thanks.

I suggest first turning the single string into a list of integer-key/string-value tuples, the we can look at passing invalid values unaltered instead of dropping it.

When that works out and time allows, I'll gladly add proper parameter value types such that users can use them as-is without conversion.

klemens-ya: > I was thinking the same, but wasn't sure how much effort you want to spend here. If you're up for this, that'd be awesome (feel free to ask in #development on Slack if you have issues with constructing the values).

That'd be helpful, but Slack is not available where I live.

awelzel: > I suggest first turning the single string into a list of integer-key/string-value tuples, the we can look at passing invalit values unaltered instead of dropping it.

Sounds great!

> That'd be helpful, but Slack is not available where I live.

Feel free to ask here, too. Going from existing records in init-bare.zeek and searching for the place where they are instantiated in C++ code is a workable approach usually. We don't have a tutorial for these sort of things I believe maybe we should :slightly_smiling_face:

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,156 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
-            analyzer->EnqueueConnEvent(dns_SVCB, analyzer->ConnVal(), msg->BuildHdrVal(), msg->BuildAnswerVal(),
-                                       msg->BuildSVCB_Val(svcb_data));
+    // Parse SvcParams records, if any, according to
+    // https://datatracker.ietf.org/doc/html/rfc9460#presentation
+
+    // The type of a single SvcParam.
+    static auto dns_svcb_param = id::find_type<RecordType>("dns_svcb_param");
+    // An [optional] vector of SvcParam, to be assigned to dns_svcb_rr.
+    // XXX compile error: https://gist.github.com/klemens-ya/e2d13bfa9d824669cde81404961c1752
+    auto svc_params = make_intrusive<VectorVal>(dns_svcb_param);
```
klemens-ya: Here's where my mental model seems to be wrong, but I can't figure it out based on [compiler output](https://gist.github.com/klemens-ya/e2d13bfa9d824669cde81404961c1752) alone.

These types and contains should match what I defined in `init-bare.zeek`, yet there's some type confusion I don't see.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: > Feel free to ask here, too. Going from existing records in init-bare.zeek and searching for the place where they are instantiated in C++ code is a workable approach usually.

I did that, but ran into something, see https://github.com/zeek/zeek/pull/4767/files#r2300656413

> We don't have a tutorial for these sort of things I believe maybe we should 🙂

Wouldn't hurt to lay this out for newcomers, I'd certainly welcome such documentation.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Fixed in bb3c0e7a817abe21096d442a4b24f3f035f57b62.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: > I guess, ideally we'd export all data properly typed according to the key:

Anyways, I gave that a go before passing malformed data:  everything is properly typed now, code compiles and runs.

... but it's not quite there yet, see ea3c8ef1d939495a06b8188d7ccff2463efdcdf5.

Feedback and/or fresh eyes welcome, I'll come back to this tomorrow.

klemens-ya: > ... but it's not quite there yet, see [ea3c8ef](https://github.com/zeek/zeek/commit/ea3c8ef1d939495a06b8188d7ccff2463efdcdf5).

Fixed in 9dd040cb57e0b5df7bdcb63e3ee2231bdbe9ab83, but somehow I can't use the `addr_vec` type:
```
.../src/analyzer/protocol/dns/DNS.cc:1739:60: error: ‘addr_vec’ is not a member of ‘zeek::id’
 1739 |                 auto hints = make_intrusive<VectorVal>(id::addr_vec);
      |  
```

I don't see how that alias is different from, say `id::string_vec`, but it is used, albeit rarely (and differently):
```
$ git grep -w addr_vec 
scripts/base/init-bare.zeek:type addr_vec: vector of addr;
scripts/base/init-bare.zeek:    # XXX should be addr, but id::addr_vec does not work...
scripts/base/init-bare.zeek:            ip: addr_vec &optional &log;    ##< List of IP entries in SAN
scripts/policy/protocols/dhcp/msg-orig.zeek:            msg_orig: vector of addr &log &default=addr_vec();
src/analyzer/protocol/dns/DNS.cc:                // XXX error: ‘addr_vec’ is not a member of ‘zeek::id’
src/file_analysis/analyzer/x509/X509.cc:                ips = make_intrusive<VectorVal>(id::find_type<VectorType>("addr_vec"));
src/zeek.bif:function routing0_data_to_addrs%(s: string%): addr_vec
src/zeek.bif:   auto rval = zeek::make_intrusive<zeek::VectorVal>(zeek::id::find_type<zeek::VectorType>("addr_vec"));
```

klemens-ya: > I guess, ideally we'd export all data properly typed according to the key:
> 
> ```
> type dns_svcb_param_value: record {
>     keys:  vector of count  &optional; # mandatory
>     ids:   vector of string &optional; # alpn
>     port:  port             &optional; # port
>     hints: vector of addr   &optional; # ipv4hint or ipv6hint
> };
> 
> type dns_svcb_param: record {
>     key: count;
>     value: dns_svcb_param_value &optional;
> };
> ```

A more compact format seems apt:
```
type dns_svcb_param: record {
    key:        count;
    mandatory:  vector of count  &optional;
    alpn:       vector of string &optional;
    p:          count            &optional; # port
    ipv4hint:   vector of addr   &optional;
    ipv6hint:   vector of addr   &optional;
};
```
No need for the intermediate `SvcParamValue` type:  let all values be optional, which names matching their respective keys.

Only `p` cannot be named "port" as that conflicts with Zeek's type name.

klemens-ya: > A more compact format seems apt:

Done via cb7ca3aa13f02c1a2fca6f3888ef3d9e6821f477, I'm happy with that structure.

@awelzel What do you think?

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
             break;
-        default: break; // unreachable. for suppressing compiler warnings.
+        }
+
+        svc_param->Assign(0, make_intrusive<CountVal>(key));
+        int item_len_parsed = 0;
+
+        switch ( key ) {
+            case detail::mandatory: // list of keys
+            case detail::port: // port
```
awelzel: Given port is only allowed to be a single entry with 2 bytes, seems clearer to not mix `mandatory` and `port` parsing.

The current logic always allocates the `mandatory` vector, but then doesn't use it when it's really a port and three's and if/else later. It seems also possible to have port param with many bytes and it'd just use the last value instead of baling/weird'ing.

Mind moving to a separate `case`?

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
-        case detail::TYPE_SVCB:
-            analyzer->EnqueueConnEvent(dns_SVCB, analyzer->ConnVal(), msg->BuildHdrVal(), msg->BuildAnswerVal(),
-                                       msg->BuildSVCB_Val(svcb_data));
+    // Parse the list of SvcParams, if any.
+    static auto dns_svcb_param_vec = id::find_type<VectorType>("dns_svcb_param_vec");
+    auto svc_params = make_intrusive<VectorVal>(dns_svcb_param_vec);
+
+    int svc_params_len = rdlength - parsed_bytes;
+
+    // In AliasMode, SvcParams MUST be ignored.
```
awelzel: That's a rule for clients - could we instead attempt to parse svc_params anyhow to the best of our possibilities? If not, minimally trigger a weird here if `svc_params_len > 0` and `priority=0` so we have it as a signal in weird.log that it's happening.

src/analyzer/protocol/dns/DNS.cc
awelzel: > @awelzel What do you think?

I think this is shaping up nicely! I'll comment on the changes directly.

I'm wondering about collapsing ipv4 / ipv6 hints into a single `iphint`, as for Zeek the address type shouldn't matter and it's possible to check `key` what kind of hint it was.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+        static auto dns_svcb_param = id::find_type<RecordType>("dns_svcb_param");
+        auto svc_param = make_intrusive<RecordVal>(dns_svcb_param);
+
+        const uint16_t key = ExtractShort(data, len);
+        const uint16_t value_len = ExtractShort(data, len);
+        const u_char* value_start = data;
+        svc_params_len -= 2 + 2;
+
+        if ( key > detail::ipv6hint ) {
+            analyzer->Weird("Reserved or invalid SvcParamKey");
```
awelzel: Sorry for not commenting earlier on this. Weird names are structured with underscores and usually have a common prefix. They are not quite free text. Maybe review and take inspiration from the existing `Weird()` calls in the DNS analyzer.

Maybe `"DNS_SVCB_param_key_invalid"` would work here.

For this one specifically, it'd also be good to add the `key` as `addl` to the `Weird()`. If that message is observed in the wild, there's a good chance that a new key was introduced in an RFC and having it readily available in the logs helps to find the matching RFC :-) Or it helps to understand that it was completely bogus, too.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
             break;
-        case detail::TYPE_HTTPS:
-            analyzer->EnqueueConnEvent(dns_HTTPS, analyzer->ConnVal(), msg->BuildHdrVal(), msg->BuildAnswerVal(),
-                                       msg->BuildSVCB_Val(svcb_data));
+        }
+
+        if ( value_len > svc_params_len ) {
+            analyzer->Weird("SvcParamValue overflows SvcParam");
```
awelzel: `DNS_SVCB_param_value_overflow` ?

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
             break;
-        default: break; // unreachable. for suppressing compiler warnings.
+        }
+
+        svc_param->Assign(0, make_intrusive<CountVal>(key));
```
awelzel:
```suggestion
        svc_param->Assign(0, zeek::val_mgr->Count(key));
```
Avoids allocating a new Count instance and re-uses existing ones.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+
+                while ( item_len_parsed + 1 + 1 < value_len ) {
+                    // XXX no Extract*() to get a single octet
+                    uint8_t alpn_len = *data;
+                    data += 1;
+                    len -= 1;
+                    item_len_parsed += 1;
+
+                    if ( alpn_len == 0 || item_len_parsed + alpn_len > value_len ) {
+                        analyzer->Weird("invalid alpn size");
```
awelzel: `DNS_SVCB_alpn_invalid_size` ?

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+                        break;
+                    }
+
+                    alpn->Assign(alpn->Size(), zeek::make_intrusive<zeek::StringVal>(alpn_len, reinterpret_cast<const char*>(data)));
```
awelzel: Nit.

```suggestion
                    alpn->Append(zeek::make_intrusive<zeek::StringVal>(alpn_len, reinterpret_cast<const char*>(data)));
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+                break;
+            }
+
+            case detail::ipv4hint: // list of IPs
+            case detail::ipv6hint: // list of IPs
+            {
+                // XXX error: ‘addr_vec’ is not a member of ‘zeek::id’
+                //     No idea how this one is special;  other id::*_vec are
+                //     defined in init-bare.zeek and used here.
+                // XXX Use string for now.
```
awelzel: `zeek::id::addr_vec` just doesn't exist as it might not be popular enough. The following should do:
```
                static auto addr_vec = zeek::id::find_type<VectorType>("addr_vec");
                auto hint = make_intrusive<VectorVal>(addr_vec);
```

scripts/base/init-bare.zeek
```diff
@@ -3071,12 +3071,30 @@ type dns_loc_rr: record {
...
+#
+## .. zeek:see:: dns_svcb_rr
+type dns_svcb_param: record {
+	key: count;				##< SvcParamKey
+	mandatory: vector of count &optional;	##< "mandatory" SvcParamKey values
+	alpn: vector of string &optional;	##< "alpn" IDs
+	p: count &optional;			##< "port" number, TCP or UDP
+	# XXX should be addr, but id::addr_vec does not work...
+	ipv4hint: vector of string &optional;	##< "ipv4hint" IP addresses
+	ipv6hint: vector of string &optional;	##< "ipv4hint" IP addresses
```
awelzel:
```suggestion
	iphint: vector of addr &optional;	##< "ipv4hint" or "ipv6hint" addresses
```

See comment about looking up the `addr_vec` type.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+                analyzer->Weird("reserved or invalid SvcParam");
+            malformed:
+                // TODO Log or pass along malformed data instead of discarding it.
+                data += value_len;
+                len -= value_len;
+                item_len_parsed += value_len;
+                break;
+        }
+
+        svc_params->Assign(svc_params->Size(), std::move(svc_param));
```
awelzel:
```suggestion
        svc_params->Append(std::move(svc_param));
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+                }
+
+                if ( hint->Size() > 0 )
+                    svc_param->Assign(is_ipv4 ? 4 : 5, std::move(hint));
+                break;
+            }
+
+            case detail::no_default_alpn:
+                if ( value_len > 0 )
+                    analyzer->Weird("no-default-alpn has value");
```
awelzel: Maybe:
```suggestion
                    analyzer->Weird("DNS_SVCB_no_default_alpn_with_value");
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1613,29 +1617,171 @@ bool DNS_Interpreter::ParseRR_SVCB(detail::DNS_MsgInfo* msg, const u_char*& data
...
+            case detail::ech: // N/A
```
awelzel: Is this more like TODO instead of N/A?

https://datatracker.ietf.org/doc/html/draft-ietf-tls-svcb-ech-08#name-svcparam-for-ech-configurat

I think it'd be interesting enough to pass to script-land, even if it's just as an `ech: string &optional`. What do you think?

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Right, I had it split already, but merged the cases again, mostly to compact whilst focusing on the types.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: But it does exist:
https://github.com/zeek/zeek/blob/ea0ffd3ec16aaaa8e81d4502ef787e06df8be986/scripts/base/init-bare.zeek#L101-L104

Right where the other types are, thus I don't get how the compiler is happy with one, but not the other:
https://github.com/zeek/zeek/blob/ea0ffd3ec16aaaa8e81d4502ef787e06df8be986/scripts/base/init-bare.zeek#L87-L90

awelzel: Ah! Not all types from init-bare.zeek are available directly in C++. Only [some types](https://github.com/zeek/zeek/blob/ea0ffd3ec16aaaa8e81d4502ef787e06df8be986/src/ID.h#L262-L273) for which we [explicitly do extra work](https://github.com/zeek/zeek/blob/ea0ffd3ec16aaaa8e81d4502ef787e06df8be986/src/ID.cc#L83-L96) to make them available on `zeek::id::` for convenience. Also, record types in `.bif`  become available on `BifType::`, but that's separate.

If you don't mind the `zeek::id::find_type<>()` lookup, would keep as is.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Sure, now it's just an additional `Weird()`.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Thanks, that explains a lot.

I fixed the type and made ipv4/6hint use the same `hint` record field.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: I swept all my additions, but left existing `foo->Assign(foo->Size(), ...)` in DNS.cc untouched.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Thanks for clarifying, I adjusted all `Weird()` arguments.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Yes, passing every (raw) value is still a TODO.

klemens-ya:
>     * Would collapse hints into a single field (lookup of `addr_vec` type should work rather than `id::addr_vec`).
> 
>     * The weird names need a bit more structure - I've added some examples.

Both done.

>     * Should rudimentary ech handling be added rather than running into a weird? Even if the value is only in a `raw: string &optional` field, that's more useful for script land than ignoring it. No need to parse the ECH details out.
> 
>     * The `ParseRR_SVCB()` is getting pretty long and complicated. Would you be up for splitting part of the `while ( svc_params_len >= 2 + 2 )` body into a helper? Maybe parsing just the value? Something like `RecordValPtr ParseRR_SVCB_Param(key, value_start, value_len)` - not sure if that's the right signature, might need a bit of iterating.

Thought about it, either is fine with me, I'll split it out tomorrow, no problem.

>     * Do you by any chance have a few pcaps that could be added that exercise the code more, or do your testing pcaps contain private info? DNS is part of the fuzzer setup, so hopefully any bugs would be tickled with that, but a few more btest/pcap based test would be nice to understand proper function.

The idea was/is to amend tests, either with a few sample requests from public DNS or with crafted responses so as to trigger the invalid cases.

Simply haven't come around to that yet.

> If this is too much forth and back, someone from the team can probably pick it up in case you don't have any cycles/capacity left.

Thanks for the help, so far I'm good on resources, I think.

src/analyzer/protocol/dns/DNS.cc
awelzel: Yeah, that'd be an alternative for now, just pass `raw: string &optional` for those keys that aren't parsed. I thought you weren't a fan, but maybe you're getting the hang of it :-)

klemens-ya: `ech` now is a string of base64 encoded `ECHConfigList` data.

klemens-ya:
> Thought about it, either is fine with me, I'll split it out tomorrow, no problem.

Moving out all of SvcParams rather than a single SvcParam seemed easier and reads nicely, imho, so that's what I did.

> The idea was/is to amend tests, either with a few sample requests from public DNS or with crafted responses so as to trigger the invalid cases.

I added two requests for public DNS (see https://defo.ie/) in a new trace file to test ECH.

Compare against `dig https defo.ie` and `dig https cloudflare-ech.com`.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Saw that one just one;  all three occurences fixed, thanks.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1587,17 +1589,174 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+                break;
+            }
+
+            case detail::ech:  // ECHConfigList
+            {
+                const String* ech = ExtractStream(data, len, value_len);
+                item_len_parsed += value_len;
+
+                // Convert binary blob to presentation format.
+                String* b64 = zeek::detail::encode_base64(std::move(ech), nullptr, analyzer->Conn());
```
awelzel: My mental model is that a `std::move()` on a raw pointer doesn't do anything. Were these moves in hope we'd support a unique_ptr here? :-)

Would remove the two `std::move()`s, but also there seems to be a leak of `ech` (ASAN CI task agrees). You could put the `ech` pointer in a unique_ptr for auto-freeing, and pass the raw pointer from `.get()` to `encode_base64()`.

awelzel:
> 
> Moving out all of SvcParams rather than a single SvcParam seemed easier and reads nicely, imho, so that's what I did.
> [...] 
> I added two requests for public DNS (see https://defo.ie/) in a new trace file to test ECH.
> 
> Compare against `dig https defo.ie` and `dig https cloudflare-ech.com`.

Thanks! There's a leak in the ech handling currently. Otherwise, I think you can undraft this PR now :-)


EDIT: The one thing that concerns me a bit is the amount of script-level allocations that can be provoked within a single packet here? E.g. crafting a DNS response with a bunch of hints or alpn strings. But I guess that's mostly DNS?

src/analyzer/protocol/dns/DNS.h
```diff
@@ -158,6 +159,15 @@ enum SVCPARAM_Key : uint8_t {
...
 
+struct SvcParam {
+    uint16_t key;        ///< see SVCPARAM_Key
+    VectorValPtr mandatory;
+    VectorValPtr alpn;
+    uint16_t p;
+    VectorValPtr hint;
+    StringValPtr ech;
+};
+
```
awelzel: That's probably left-over from a separate refactor approach?

klemens-ya: I modeled it after the `dns_svc_params` record right from the start, assuming it'd be needed.

All gone now, thanks.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: The move()s are from a previous attempt when I didn't resort to/know about ExtractStream()/encode_base64() yet... on a friday afternoon ;)

Should be good now, pcap tests now pass with `--sanitizer=address` and `detect_odr_violation=1:detect_leaks=1`.

klemens-ya:
> Otherwise, I think you can undraft this PR now :-)

Sure, once I've rebased squashed commits onto master and updated title/description.

> EDIT: The one thing that concerns me a bit is the amount of script-level allocations that can be provoked within a single packet here? E.g. crafting a DNS response with a bunch of hints or alpn strings. But I guess that's mostly DNS?

https://datatracker.ietf.org/doc/html/rfc9460#name-rdata-wire-format

```
Clients MUST consider an RR malformed if:

- the end of the RDATA occurs within a SvcParam.
- SvcParamKeys are not in strictly increasing numeric order.
- the SvcParamValue for a SvcParamKey does not have the expected format.

Note that the second condition implies that there are no duplicate SvcParamKeys.

If any RRs are malformed, the client MUST reject the entire RRset and fall back to non-SVCB connection establishment.
```

Personally, I'd follow this strictly, but it conflicts with the "pass raw data even if malformed" approach.

If one decides to parse only one parameter of each key, they could either abort on duplicates, which won't get you remaining data, or allocate one raw string for it, which doesn't really protect against what you worry about.

klemens-ya:
> Personally, I'd follow this strictly, but it conflicts with the "pass raw data even if malformed" approach.
> 
> If one decides to parse only one parameter of each key, they could either abort on duplicates, which won't get you remaining data, or allocate one raw string for it, which doesn't really protect against what you worry about.

Since Zeek is not a real DNS client, most of the restrictions in the RFC are not implemented anyway, so I opted to expose all data it can get hold off, see 002e466d49f0694bcd9364f1d4387fa5611a5daf.

Now I'd like to see a few crafted DNS response packets in test pcaps to trigger this new code,
but I currently don't plan to do that myself anytime soon.

Either way, if malicious packets cause too much work because of this... well, tough luck, you've been warned this is the Internet ;)

But everyone's running Zeek with minimal privileges and system resource limits anyway, right?

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1587,17 +1589,160 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
     return rdlength == 0;
 }
 
+VectorValPtr DNS_Interpreter::Parse_SvcParams(const u_char*& data, int& len, int svc_params_len) {
+    static auto dns_svcb_param_vec = id::find_type<VectorType>("dns_svcb_param_vec");
+    auto svc_params = make_intrusive<VectorVal>(dns_svcb_param_vec);
+
+    while ( svc_params_len >= 2 + 2 ) {
```
timwoj:
```suggestion
    // Service params are at least 4 bytes (2 for the key, 2 for the length)
    while ( svc_params_len >= 4 ) {
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1587,17 +1589,160 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+        static auto dns_svcb_param = id::find_type<RecordType>("dns_svcb_param");
+        auto svc_param = make_intrusive<RecordVal>(dns_svcb_param);
+
+        const uint16_t key = ExtractShort(data, len);
+        uint16_t value_len = ExtractShort(data, len);
+        int item_len_parsed = 0;
+        svc_params_len -= 2 + 2;
```
timwoj:
```suggestion
        svc_params_len -= 4;
```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1587,17 +1589,160 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+            value_len = svc_params_len;
+            goto malformed;
+        }
+
+        svc_param->Assign(0, zeek::val_mgr->Count(key));
+
+        switch ( key ) {
+            case detail::mandatory: // list of keys
+            {
+                auto mandatory = make_intrusive<VectorVal>(id::index_vec);
```
timwoj: Move this below the `if` block to avoid the allocation on failure.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1587,17 +1589,160 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+
+                svc_param->Assign(3, zeek::val_mgr->Count(ExtractShort(data, len)));
+                item_len_parsed += 2;
+                break;
+
+            case detail::ipv4hint: // list of IPs
+            case detail::ipv6hint: // list of IPs
+            {
+                static auto addr_vec = id::find_type<VectorType>("addr_vec");
+                auto hint = make_intrusive<VectorVal>(addr_vec);
```
timwoj: Same comment re: avoiding the allocation.

timwoj:
> Now I'd like to see a few crafted DNS response packets in test pcaps to trigger this new code, but I currently don't plan to do that myself anytime soon.

We could certainly add some SVCB data to the fuzzer corpus and let oss-fuzz do funny things with it.

src/analyzer/protocol/dns/DNS.cc
klemens-ya: Sure, I also summed up the ALPN `1 + 1` (the case comment ellipse is fine, I think) and made this comment a proper sentence.

PR #4771, opened 2025-08-18
Move BinPAC, bifcl, af_packet, and gen_zam submodules into main zeek repo, take 2

timwoj:
This is a rework of https://github.com/zeek/zeek/pull/4746, but moves all of the commits from the original repositories into Zeek's tree. The commits were prepared on the branch `topic/timw/filter-repo-for-submodule-move` in each repository using the following `git filter-repo` command:

```
git filter-repo --path {paths} --refs refs/heads/topic/timw/filter-repo-for-submodule-move --prune-empty always --prune-degenerate auto --force --to-subdirectory-filter tools/gen-zam --message-callback 'return b"{repo name}: " + message'
```

The commit message from each repo is prepended with that repo's name. This allows us to keep track of them more easily once merged into the Zeek repo. There is a set of commits after each merge that removes the submodule from the Zeek repo, adapts the build to use the moved version, and makes any other necessary commits to make everything work correctly.

The important commits to review out of this giant pile of commits:

AF_Packet: 
- https://github.com/zeek/zeek/pull/4771/commits/362b0f7f31
- https://github.com/zeek/zeek/pull/4771/commits/6933fc29ee

Gen-ZAM:
- https://github.com/zeek/zeek/pull/4771/commits/77340a1b14
- https://github.com/zeek/zeek/pull/4771/commits/d586a8040e

Bifcl:
- https://github.com/zeek/zeek/pull/4771/commits/a3d6f0f478

BinPAC:
- https://github.com/zeek/zeek/pull/4771/commits/aa47f0122f
- https://github.com/zeek/zeek/pull/4771/commits/c8a754f6c1
- https://github.com/zeek/zeek/pull/4771/commits/56e97a5412
- https://github.com/zeek/zeek/pull/4771/commits/2f65e21921


ckreibich (approved):
Awesome ! Thanks for figuring out that epic `filter-repo` command, and the detailed links to the commits. I really like this. I built and tested locally and all went well.

I wonder if @J-Gras would like to look over the AF_PACKET changes in case anything comes to mind.

One tiny thing that caught my eye is that the configure output still ends with
```
BifCL:                    included
BinPAC:                   included
Gen-ZAM:                  included
```
which could now simply go away, and we can also ditch `--with-gen-zam=`?  (Can certainly be follow-ups, too.)

timwoj:
> which could now simply go away, and we can also ditch --with-gen-zam=? (Can certainly be follow-ups, too.)

I added a commit that removed them. I could see some follow-up work that restores our ability to cross-compile and brings all of those back again.

awelzel (approved):
Only thought is `src/iosource/{pcap,af_packet}` would be nicer in `src/pktsrc` or some such, but if there's only two packet sources for the next years, seems fine as is. Not like we're planning to add netmap or dpdk or ... (yet?).

PR #4773, opened 2025-08-19
logging/Manager: Non-null strings for empty strings

awelzel:
After #4724, empty strings would result in nullptrs being stored in the threading::Value's string_val.data field instead of a valid pointer to an empty strings. This upsets UBSAN's  nonnull check for memcpy()

    [01:29:45.807]   ../../src/SerializationFormat.cc:80:33: runtime error: null pointer passed as argument 2, which is declared to never be null
    [01:29:45.807]   /usr/include/string.h:44:28: note: nonnull attribute specified here
    [01:29:45.807]       #0 0x5b2e9c933a3f in zeek::detail::SerializationFormat::WriteData(void const*, unsigned long) /zeek/build/src/../../src/SerializationFormat.cc:80:5
    [01:29:45.807]       #1 0x5b2e9c935184 in zeek::detail::BinarySerializationFormat::Write(char const*, int, char const*) /zeek/build/src/../../src/SerializationFormat.cc:371:40

Continue to allocate the empty string for now as a fix.

### DIFF
--- a/src/logging/Manager.cc
+++ b/src/logging/Manager.cc
@@ -1511,9 +1511,6 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                 info->total_truncated_string_fields->Inc();
             }
 
-            if ( allowed_bytes == 0 )
-                return lval;
-
             char* buf = new char[allowed_bytes];
             memcpy(buf, s->Bytes(), allowed_bytes);
 
--- a/src/logging/Manager.cc
+++ b/src/logging/Manager.cc
@@ -1566,9 +1566,6 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                 info->total_truncated_containers->Inc();
             }
 
-            if ( allowed_elements == 0 )
-                return lval;
-
             lval.val.set_val.vals = new threading::Value*[allowed_elements];
 
             for ( size_t i = 0; i < allowed_elements && total_record_size < max_log_record_size; i++ ) {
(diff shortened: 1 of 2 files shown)


awelzel:
This should fix the failing UBSAN tasks in CI.

Technically I think we should do the same for empty sets and vectors, but these might be less likely to tickle something, though `set_val.vals` and `vector_val.vals` being a `nullptr` for empty sets/vectors instead of a valid pointer to an empty array is kind of a regression. Given this is part of the plugin API, should probably also remove the early return there? Thoughts?

timwoj:
I could also see doing it for containers. It just seems so weird to pass around a zero-length buffer instead of a nullptr.

awelzel:
Re-pushed to also do it for set / vector.

src/logging/Manager.cc
```diff
@@ -1600,9 +1594,6 @@ threading::Value Manager::ValToLogVal(WriterInfo* info, const Stream* stream, st
                 info->total_truncated_containers->Inc();
             }
 
-            if ( allowed_elements == 0 )
-                return lval;
-
             lval.val.vector_val.vals = new threading::Value*[allowed_elements];
 
             auto& vv = vec->RawVec();
```
timwoj, line 1599: Does it make sense to just skip all of the rest of this if `allowed_elements` is zero? I assume in the string case that `memcpy` is just a nop internally if the size is zero. In the `set` case, the for-loop will just exit.

awelzel: I mostly was going for fewer `return` statements just in case someone puts some more book-keeping further down and it's also how the previous code would flow if you passed an empty set or vector.

PR #4774, opened 2025-08-19
CI: Update CentOS 9 to Python 3.13

timwoj:
This updates the CI Dockerfile for Centos 9 to use Python 3.13. It also adds a NEWS entry about requiring Python 3.10 for Zeek. The NEWS entry isn't *entirely* true in that some subprojects will still work with 3.9, but setting the expectation is good here.

https://cirrus-ci.com/task/5195113661464576 has a build with this included. From the build log:

> [18:08:39.201] -- Found Python: /usr/bin/python3 (found suitable version "3.13.5", minimum required is "3.9") found components: Interpreter Development Development.Module Development.Embed

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,8 @@ We would like to thank ... for their contributions to this release.
 Breaking Changes
 ----------------
 
+- Python 3.10 is now required for Zeek and all of its associated subprojects.
+
 - The ``&optional`` script attribute will now error when applied to anything that's
   not a record field. Previously, this would have surprising behavior.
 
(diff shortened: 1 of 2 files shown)


ckreibich (approved):
LGTM. Did you want to leave bumping [the minimum in `CMakeLists.txt`](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L838)  a bit down the road?

timwoj:
> LGTM. Did you want to leave bumping [the minimum in `CMakeLists.txt`](https://github.com/zeek/zeek/blob/master/CMakeLists.txt#L838) a bit down the road?

Yeah. This is mostly a "fix it for this one platform now, and fix it for the individual projects as necessary" step.

PR #4778, opened 2025-08-21
testing: Make share/zeek/tests a relative symlink

### LINKED ISSUE #3266: Packages miss backwards-compatibility link of testing-related files
It seems that our binary packages, at least for zeek-6.0.0 and debian-11 miss a softlink of `<PREFIX>/share/btest/data` to `<PREFIX>/share/zeek/tests`. We have an explicit install rule for that, https://github.com/zeek/zeek/blob/4b71b7d05c7fad7bd2c2619471a836a142d8edf2/testing/CMakeLists.txt#L12-L18

This leads to packages relying on support files like `random.seed` or `Scripts/canonify-zeek-log` to be in this path to fail, e.g., a user reported that he was unable to install `zeek/spicy-ldap` in his system with Zeek from packages while our CI was ✅ (our CI uses Docker images which have these files).

We should either install this softlink in packages as well, or alternatively get rid of it completely in our CMake setup so both installations methods are more similar.

awelzel:
ZEEK_CONFIG_BTEST_TOOLS_DIR is just <ZEEK_ROOT_DIR>/share/btest and ZEEK_ROOT_DIR is just CMAKE_INSTALL_PREFIX, so this should work out.

Closes #3266

### DIFF
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -13,8 +13,8 @@ endif ()
 install(
     CODE "execute_process( \
     COMMAND ${CMAKE_COMMAND} -E create_symlink \
-    ${ZEEK_CONFIG_BTEST_TOOLS_DIR}/data \
-    ${CMAKE_INSTALL_PREFIX}/share/zeek/tests \
+    ../btest/data \
+    \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/share/zeek/tests \
     )")
 
 install(DIRECTORY scripts/spicy/ DESTINATION ${ZEEK_CONFIG_BTEST_TOOLS_DIR}/data/Scripts


awelzel:
@bbannier - think using relative paths might also make @r-rojo happier / prevent a bit of extra work for him for his relocatable escapades. WDYT?

testing/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ endif ()
 install(
     CODE "execute_process( \
     COMMAND ${CMAKE_COMMAND} -E create_symlink \
-    ${ZEEK_CONFIG_BTEST_TOOLS_DIR}/data \
+    ../btest/data \
```
bbannier, line 16: This does not fix #3266. Like [Johanna wrote](https://github.com/zeek/zeek/issues/3266#issuecomment-3205933911), the problem is that packaging uses a `DESTDIR` install. The problem is the line just below which does not install into the `DESTDIR`, but the default prefix. A fix needs to be somewhat along the lines of what I posted [here](https://github.com/zeek/zeek/issues/3266#issuecomment-3209454493).

As for relative paths, that's a different issue, but maybe worth tackling. I would however keep `ZEEK_CONFIG_BTEST_TOOLS_DIR` since it is used in a number of places and adjusting its value might break things here. If needed we could [make its path relative](https://cmake.org/cmake/help/latest/command/cmake_path.html#relative-path) to `CMAKE_INSTALL_PREFIX`.

awelzel: Yeah, I just realized and re-pushed, but actually, let me close this PR. I'd actually in favor if we just remove the symlink.

PR #4779, opened 2025-08-21
Fix installation of symlink with `DESTDIR`

### LINKED ISSUE #3266: Packages miss backwards-compatibility link of testing-related files

bbannier:
We install test data which we also make available under an alternative path for backwards compatibility. The installation of this symlink did not take `DESTDIR` installs like used by Zeek's packaging into account which caused installations from packages to behave different from installs from source.

This patch fixes the symlink to respect a possible `DESTDIR`.

Closes #3266.

### DIFF
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -14,7 +14,7 @@ install(
     CODE "execute_process( \
     COMMAND ${CMAKE_COMMAND} -E create_symlink \
     ${ZEEK_CONFIG_BTEST_TOOLS_DIR}/data \
-    ${CMAKE_INSTALL_PREFIX}/share/zeek/tests \
+    \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX}/share/zeek/tests \
     )")
 
 install(DIRECTORY scripts/spicy/ DESTINATION ${ZEEK_CONFIG_BTEST_TOOLS_DIR}/data/Scripts


awelzel (approved):

0xxon:
Cool. I am going to spin this into our binary packages as a patch, and will report back if it works ok :)

0xxon:
Sorry for the delay here. The last binary build for nightly broke due to a different problem - I will try to actually get this working in the next couple of hours.

0xxon (approved):
Thank you - this seems to work.

It currently still creates an absolute link. I wonder if we should use a relative one instead.

That being said, as a fix this seems to work, so - let's ship it.

PR #4782, opened 2025-08-21
Enable clang-tidy for binpac and bifcl

timwoj:
Now that these tools are part of the zeek repo and are being maintained alongside it, we can enable clang-tidy for them as part of the Zeek build. This PR enables it for binpac and bifcl and fixes a bunch of findings. It also reformats utility code embedded in the bison/flex files to use the clang-format style.


tools/binpac/src/pac_analyzer.cc
```diff
@@ -46,45 +46,46 @@ AnalyzerDecl::~AnalyzerDecl() {
...
-            } break;
-            case AnalyzerElement::FLOW: {
-                AnalyzerFlow* flow_elem = (AnalyzerFlow*)elem;
-                ProcessFlowElement(flow_elem);
-            } break;
-            case AnalyzerElement::DATAUNIT: {
-                AnalyzerDataUnit* dataunit_elem = (AnalyzerDataUnit*)elem;
-                ProcessDataUnitElement(dataunit_elem);
-            } break;
+    if ( elemlist )
```
awelzel: This would've avoided the extra indentation and made the diff a bit smaller, but it's a nit.

```
if ( ! elemlist )
    return;
```

awelzel: Hmm, hmm, looking at the "Remove foreach macro" commit, I'd probably opt for ditching that commit and keep the macro. Seems that was an established pattern in BinPAC and changing appears mostly churn without a real upside. Totally IMO and feel free to ignore if you strictly prefer the no-macro version.

tools/binpac/src/pac_common.h
```diff
@@ -108,13 +108,13 @@ typedef vector<StateVar*> StateVarList;
...
 
-#define delete_list(ct, pc)                                                                                            \
-    {                                                                                                                  \
-        foreach (delete_list_i, ct, pc)                                                                                \
-            delete *delete_list_i;                                                                                     \
-        delete pc;                                                                                                     \
-        pc = 0;                                                                                                        \
-    }
+template<typename T>
+constexpr void delete_list(T* container) {
```
awelzel, line 114: Fancy :-)

awelzel (approved):
Should this come with additions to `.git-blame-ignore-revs`?

tools/binpac/src/pac_analyzer.cc
timwoj: How about making a templated version that's roughly identical to https://en.cppreference.com/w/cpp/algorithm/for_each.html, but checks for pointer validity first?

timwoj: I think something like this should work:

```
template<typename T>
concept Container = requires(T t) {
    std::begin(t);
    std::end(t);
};

template<Container C, class Pred>
constexpr void ptr_for_each(C* c, Pred f) {
    if ( ! c )
        return;

    std::ranges::for_each(*c, f);
}
```

it wouldn't work in every case, such as ones where the loop returns from the caller function (such as [this one](https://github.com/zeek/zeek/pull/4782/commits/b068d870f7a1ad847be7c5f071fc23530546f6e3#diff-2576e9ef4f844a79307e475e39e9531f7b2cea72c260188fc469a05f270b8033R46-R48)), but it should work in most of the cases.

awelzel: > How about making a templated version that's roughly identical to

I think you need someone else to chime in on that :-)

It doesn't seem that BinPAC's `foreach` macro can easily escape into Zeek's code due to the separation between `src/pac_` files and `lib/binpac.h`. And then seems just a detail in BinPAC's implementation that's been like that for years and IMO would certainly justify a `// NOLINT` annotation. If it ends up something else, great, but don't have enough opinions or inspiration on that something else (the open-coding just stood out because of the churn and the `foreach` appeared pragmatic to me, but that might just be me).

0xxon: To add my two cents - I agree with Arne here, I'm not sure if there is a reason to change this and would probably just leave this as-is, leaving the foreach macro.

timwoj: Makes sense. I dropped that commit out.

PR #4785, opened 2025-08-22
IPBasedAnalyzer: Ensure a connection's ConnVal is updated

### LINKED ISSUE #4214: Nesting of analyzers may lead to stale argument values

awelzel:
One idea for the issue that a connection record's endpoint fields may be stale when events are raised before the ConnSize analyzer saw the packet and no analyzer calls GetVal() for the connection afterwards.

While this looks a bit ad hoc, I'm leaning towards a follow-up, extending GetVal(skip_update_connval=false) API and then updating BinPAC/Spicy to generate code that passes ``false``. That could avoid a number of unnecessary UpateConnVal() invocations when calling GetVal() multiple times for the same ~connection~ packet and instead just do it once when the packet has been processed.

Closes #4214


awelzel:
Requested a review from @rsmmr - maybe he has a nicer idea.

awelzel:
Wow - okay, for well-known-ports, protocol analyzers are added before the
ConnSize analyzer. Events raised by an analyzer attached via a well-known-port
also have stale orig and resp fields, it probably just never mattered.
    
This is why this change results in the dump-events diffs.

rsmmr:
Agree this is worth fixing.

Could we switch this from poll to push? The PR is current "polling" for `ConnVal` updates by regularly calling `GetVal()` after each packet (if an event was raised). Could the `ConnSize` analyzer instead trigger an update when it has new information that needs to be reflected? We could split the update logic out of `GetVal()` into its own `Connection::UpdateConnVal()`, which then in turn triggers the Analyzer's virtual methods to run (so control comes back to the `ConnSize` analyzer pretty much immediately to make the change). That `Connection::UpdateConnVal()` would first check if there is actually a `ConnVal` already, otherwise it can just be a no-op.

If we want it a bit fancier, `Connection::UpdateConnVal()` doesn't need to immediately trigger the updates, but could just set a flag at first, and then trigger when all analyzers are done with the packet, but before events process. That way we'd coalesce multiple updates into one--but not sure if that changes much, plus needs that flag.

Either way, generally the API for analyzers would become: "if you have new information that needs to be reflected in the `ConnVal`, you need to let us know, rather than hope we'll request it anyways."

> This is why this change results in the dump-events diffs.

... and this is why we have the dump-events tests. :-) They might be noisy, but they can reveal surprises that would be hard to notice otherwise.

awelzel:
> Could the ConnSize analyzer instead trigger an update when it has new information that needs to be reflected? 

I didn't do that as it's essentially true for every packet, but maybe that's too cautious. With `ZVal`s in records today, it might be fine to just update the fields in the `endpoint`  - there should be no `Val` allocation involved anymore, so the overhead might be negligible and it'd fix #4786.

~That wouldn't necessary be a general fix, but I'm also starting to wonder if `UpdateConnVal()` exists primarily/only for the  `ConnSize` analyzer and it might be easier to open-code that one directly.~

I'll ponder/test a bit with a v2. Thanks for looking!

Also, EDIT:

> ... and this is why we have the dump-events tests. :-) 

They are still there :-)

PR #4787, opened 2025-08-22
Adding disabling_analyzer() hook to scripts

mnhsrj:
Zeek’s policy-level SSL scripts (e.g. decryption.zeek and heartbleed.zeek) currently set SSL::disable_analyzer_after_detection to false, but that alone doesn’t guarantee the SSL analyzer remains active. If any other script calls disable_analyzer(), it turns the analyzer off globally for all subsequent scripts handling the same connection.
This PR adds Analyzer::disabling_analyzer() hook to these scripts to veto any external request to disable the SSL analyzer.


mnhsrj:
There's a helper script testing/scripts/disable-ssl-analyzer-after-max-count.zeek in this PR, if its out of place maybe it can be duplicated in both the tests.

scripts/policy/protocols/ssl/decryption.zeek
```diff
@@ -109,3 +105,10 @@ event ssl_change_cipher_spec(c: connection, is_client: bool)
 			set_secret(c, secrets[c$ssl$client_random]);
 		}
 	}
+
+# Do not disable analyzers after detection - otherwise we will not receive
+# encrypted packets.
+hook Analyzer::disabling_analyzer(c: connection, atype: AllAnalyzers::Tag, aid: count)
+    {
+    break;
```
awelzel, line 113: This makes sense, but it's missing one crucial part: `Analyzer::disabling_analyzer()` is invoked for every analyzer on every connection that's about to be disabled. You'd want to prevent disabling only the right SSL analyzer.

So this should be something like:
```
hook Analyzer::disabling_analyzer(c: connection, atype: AllAnalyzers::Tag, aid: count)
    {
    # Don't care about non-SSL analyzers.
    if ( atype != Analyzer::ANALYZER_SSL )
        return;

    # Only prevent disabling SSL analyzers of this connection.
    if ( c?$ssl && c$ssl?$analyzer_id && c$ssl$analyzer_id == aid )
        break;
    }
```

awelzel: I'm wondering a bit if we should keep track of SSL analyzer violations though. If they happen, allow the disabling. Otherwise, with this change, SSL analyzer violations won't disable the analyzer anymore.

This can be a follow-up though.

scripts/policy/protocols/ssl/decryption.zeek
```diff
@@ -13,10 +13,6 @@
 
 module SSL;
 
-# Do not disable analyzers after detection - otherwise we will not receive
-# encrypted packets.
-redef SSL::disable_analyzer_after_detection = F;
-
```
awelzel, line 19: Seems reasonable to keep this to not even attempt the `disable_analyzer()` call.

awelzel:
> There's a helper script testing/scripts/disable-ssl-analyzer-after-max-count.zeek in this PR, if its out of place maybe it can be duplicated in both the tests.

Not out of place, but seems okay to open-code it for the two tests that need it. You could also place it in a `@TEST-START-FILE` section with the same name in case we end-up with 5 or more tests that need it and at some point put it into a common place :-)

0xxon:
Hm.

I am honestly wondering if this really is necessary/makes sense. From my understanding, all of the policy scripts that Zeek ships will respect SSL::disable_analyzer_after_detection.

If they don't - that's a bug and we should fix this.

I am not sure if we, by default, should disallow any external scripts to disable the SSL analyzer if they wish - no matter what this setting says. We don't know what the environment is in which Zeek is used - and what the use-case of other scripts that are running is. Perhaps someone is running a script that determines that a decrypted SSL connection contains no interesting content/cannot be parsed because it doesn't contain HTTP or anither known protocol, and wants to disable the analyzer. That's an on usecase and would be prevented by this.

It also opens potential bugs like the one that Arne found above - the analyzer should still be disabled due to protocol violations.

So I'm currently tempted to say that we should keep the behaviour as it is.

awelzel:
> So I'm currently tempted to say that we should keep the behaviour as it is.

Seems fair - it should also be entirely possible to implement those hooks as shown outside of Zeek's base/policy scripts in third-party scripts.

mnhsrj:
Thanks for the review. 

If heartbleed detection is enabled by uncommenting [https://github.com/zeek/zeek/blob/master/scripts/site/local.zeek#L99](url) , this leaves the possibility that the heartbleed detection could potentially be stopped if there is an script which disables the SSL analyzer e.g. the disable-ssl-analyzer-after-max-count.zeek script provided in this PR. This can be proved by running the provided tests and leaving out the changes in scripts/policy/protocols/ssl/heartbleed.zeek and scripts/policy/protocols/ssl/decryption.zeek thereby making heartbleed detection to not work right out of the box. 
For sure, the disabling_analyzer() hook can be implemented outside as 3rd party script but I think that's putting too much responsibility on the end user ? Reading the #2328 PR, it gives a script power to veto the disable_analyzer() call from other scripts, that's what this PR is trying to achieve for the heartbleed and decryption scripts.

scripts/policy/protocols/ssl/decryption.zeek
mnhsrj: Thanks, these checks will help. I can provide this in a fixup commit.

0xxon:
Sure - you are right - if you load the heartbleed script, and then load another script that disables the SSL analyzer - the heartbleed script will stop analyzing the traffic.

But - I don't really see this as a problem. If you load a script that takes actions like disabling analyzers - it will always have side-effects if other scripts in the system want to analyze the traffic. Depending on the script that you load, this might even be desirable. As mentioned above - you might _want_ to disable the analyzer for an SSL connection, even if SSL decryption is active.

It is the responsibility of the user to ensure that the script that you load into the system does make sense. Not letting users disable the SSL analyzer when heartbleed / decrypion is enabled seems like the wrong approach to me; this limits the flexibility of Zeek given a specific use-case that's assumed.

mnhsrj:
ok, I'll close this PR.

Issue #4786, opened 2025-08-22
Stale connection orig/resp fields in scheduled events


awelzel:
This relates to #4214 where some events would observe stale orig and resp values. This ticket is basically the same, but for scheduled events with `connection` parameters, rather than connection/protocol events.

When regularly scheduling an event with a `c: connection` parameter, its `orig` and `resp`  values are out-of-date unless some other event was raised that would have invoked `UpdateConnVal()` for that same `connection`.

And I wonder what we should be doing here..  It seems like a gotcha to me that `connection` fields may be stale in a scheduled event, but maybe that's just me? Has this come up before?

It appears to be "known behavior" as for example `conn/polling.zeek` explicitly refreshes a `connection` via `lookup_connection()`. This is from an older version, today it does a lookup from `conn_id` to avoid keeping the whole `c: connection` alive for the next timer.

https://github.com/zeek/zeek/blob/18bd74454bed939ad4e05909e0af60f8985e4825/scripts/base/protocols/conn/polling.zeek#L35-L37

On the other hand, the `detect-protocol` script doesn't do that, so it might use stale data whenever the timer fires:

https://github.com/zeek/zeek/blob/3e89e6b3288453b7a0f89fb742384e213cf5cc94/scripts/policy/frameworks/analyzer/detect-protocols.zeek#L171-L177

I could see keeping the current behavior and discouraging / warning users from ever using `connection` as a direct or nested parameter for a scheduled event (should be able to detect statically). The warning could explicitly recommend using `lookup_connection(id: conn_id)` instead. This is what was done as memory optimization done for conn polling and it could be a pragmatic way to side-step this whole problem - but it's also a bit of a cop-out. Other thoughts?

---

The btest below prints many `print_connection` lines where orig/resp num_pkts is never updated even though there's just a single connection and all packets belong to it.

```
1112172470.501268, print_connection, CHhAvVGS1DHFjwGM9, orig num_pkts, 1, resp num_pkts, 0, pkts_recvd, 1
1112172487.320873, print_connection, CHhAvVGS1DHFjwGM9, orig num_pkts, 1, resp num_pkts, 0, pkts_recvd, 5
1112172558.685951, print_connection, CHhAvVGS1DHFjwGM9, orig num_pkts, 1, resp num_pkts, 0, pkts_recvd, 7
...
1112172737.733384, print_connection, CHhAvVGS1DHFjwGM9, orig num_pkts, 1, resp num_pkts, 0, pkts_recvd, 22
[...]

PR #4788, opened 2025-08-25
connection val update - take 2

awelzel:
This is a revision of #4785 - I'm measuring 4-5% perf overhead when running a `-b base/protocols/conn`, with the UpdateConnVal() method called after every packet showing up with ~4% in perf profiles, so I'm not really happy with it, opening a PR ot run the benchmarker tests.


awelzel:
Right side is with the change from this PR, so slower across the board. I'll ponder a bit or maybe someone has a nicer idea or spots something off ...

image

awelzel:
I'm going to close this one, too. Think #4214 and #4786 together might need a bit more discussion/design and experiments.

PR #4790, opened 2025-08-25
Proposal: Document UID pivoting via its type

evantypanski:
This relates to #4775 - I didn't actually know for sure that my suggestion would work, so I tried it in code and it worked fine. This is a draft as it's more of an RFC (I didn't meticulously get every case nor fix whitespace).

This changes a connection's UID to be a `conn_uid` type rather than a `string` type. This way, documentation will link to this new type, which may explain pivoting better. Then, users can rest assured that every `conn_uid` is meant to correlate with a connection's UID.

The one caveat that I don't know about (maybe @ckreibich can help inform me) is if this would play nicely with the log schema stuff, which was indeed the initial motivation. I just threw this together in 5 minutes to try it and figured a draft PR would be the easiest way :) - update, it might be collapsed so in logschema it just says it's `string` - maybe that's a deal breaker.

I personally like this more than a Zeekygen annotation for a few reasons:

1) It's in the type system, so if we want to make this more strict in the future in some way, we can
2) It's basically self-documenting, so you don't need to care about the doc comments etc.
3) It's easier to coerce the user to use the `conn_uid` type rather than keep a doc comment around

In docs this would just be a `conn_uid` type on the `uid` field:

Screenshot 2025-08-25 at 10 03 42 AM

You can click the type and get to the explanation, which will include info about "pivoting" via this type:

Screenshot 2025-08-25 at 10 05 00 AM

Then other logs are essentially self-documenting, since a `conn_uid` will link back to explain pivoting, so even DHCP makes sense:

Screenshot 2025-08-25 at 10 05 57 AM

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -230,6 +230,10 @@ type conn_id: record {
 	ctx: conn_id_ctx &log &default=conn_id_ctx();  ##< The context in which this connection exists.
 };
 
+## A connection's unique identifier. This can be used to pivot between logs
+## with the same connection (TODO expand on this?)
+type conn_uid: string;
+
 ## The identifying 4-tuple of a uni-directional flow.
 ##
 ## .. note:: It's actually a 5-tuple: the transport-layer protocol is stored as
@@ -885,7 +889,7 @@ type connection: record {
 	## creates an ID that is very likely unique across independent Zeek runs.
 	## These IDs can thus be used to tag and locate information associated
 	## with that connection.
-	uid: string;
+	uid: conn_uid;
 	## If the connection is tunneled, this field contains information about
 	## the encapsulating "connection(s)" with the outermost one starting
 	## at index zero.  It's also always the first such encapsulation seen
--- a/scripts/base/protocols/conn/main.zeek
+++ b/scripts/base/protocols/conn/main.zeek
@@ -22,7 +22,7 @@ export {
 		## This is the time of the first packet.
 		ts:           time            &log;
 		## A unique identifier of the connection.
-		uid:          string          &log;
+		uid:          conn_uid        &log;
 		## The connection's 4-tuple of endpoint addresses/ports.
 		id:           conn_id         &log;
 		## The transport layer protocol of the connection.
--- a/scripts/base/protocols/dce-rpc/main.zeek
+++ b/scripts/base/protocols/dce-rpc/main.zeek
@@ -12,7 +12,7 @@ export {
 		## Timestamp for when the event happened.
 		ts         : time     &log;
 		## Unique ID for the connection.
-		uid        : string   &log;
+		uid        : conn_uid &log;
 		## The connection's 4-tuple of endpoint addresses/ports.
 		id         : conn_id  &log;
 		## Round trip time from the request to the response.
--- a/scripts/base/protocols/dhcp/main.zeek
+++ b/scripts/base/protocols/dhcp/main.zeek
@@ -24,7 +24,7 @@ export {
 		## DHCP is occurring.  This behavior with multiple connections is
 		## unique to DHCP because of the way it uses broadcast packets
 		## on local networks.
-		uids:           set[string] &log;
+		uids:           set[conn_uid] &log;
 
 		## IP address of the client.  If a transaction
 		## is only a client sending INFORM messages then
--- a/scripts/base/protocols/dnp3/main.zeek
+++ b/scripts/base/protocols/dnp3/main.zeek
@@ -14,7 +14,7 @@ export {
 		## Time of the request.
 		ts:         time           &log;
 		## Unique identifier for the connection.
-		uid:        string         &log;
+		uid:        conn_uid       &log;
 		## Identifier for the connection.
 		id:         conn_id        &log;
 		## The name of the function message in the request.
--- a/scripts/base/protocols/dns/main.zeek
+++ b/scripts/base/protocols/dns/main.zeek
@@ -21,7 +21,7 @@ export {
 		ts:            time               &log;
 		## A unique identifier of the connection over which DNS messages
 		## are being transferred.
-		uid:           string             &log;
+		uid:           conn_uid           &log;
 		## The connection's 4-tuple of endpoint addresses/ports.
 		id:            conn_id            &log;
 		## The transport layer protocol of the connection.
--- a/scripts/base/protocols/ftp/info.zeek
+++ b/scripts/base/protocols/ftp/info.zeek
@@ -27,7 +27,7 @@ export {
 		## Time when the command was sent.
 		ts:               time        &log;
 		## Unique ID for the connection.
-		uid:              string      &log;
+		uid:              conn_uid    &log;
 		## The connection's 4-tuple of endpoint addresses/ports.
 		id:               conn_id     &log;
 		## User name for the current FTP session.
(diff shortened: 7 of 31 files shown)


timwoj:
Would this break other plugins? I grepped through my set of package clones, and there's definitely some out there using `uid` as a string. I assume that since `conn_id` is just a type alias, it should work fine, but can you add a test just to verify it? One example is https://github.com/corelight/zeek-spicy-stun/blob/master/analyzer/main.zeek#L102-L111, which stores a copy of `uid` as a string field.

awelzel:
I hadn't thought much about #4775, but wonder if you thought about `files/main.zeek`? There's also file UIDs / fuids for pivoting on files.log. E.g. `PE::Info$id` references `Files::Info$fuid`, and `Files::Info$uid` references `Conn$Info$uid` then.

The following isn't markup or a separate type, but instead a new attribute (which might be fine as it'll always about fields in records). The field ref should be valid syntax and I find it pretty explicit and it's a notch more generic and applicable in different scenarios without introducing new types. Like if someone used LDAP::MessageInfo$message_id in a different log, they could put in the other log `&ref=LDAP::MessageInfo$message_id` and it should just work without changing the base script to introduce a new `ldap_message_id` type.

```
module Files;

type Info: record {
    ...
    fuid: string &log;
    uid: string &log &optional &ref=Conn::Info$uid
}

module PE;

type Info {
    fuid: string &log &ref=Files::Info$fuid;
}
```

I like a new attribute over markup within the doc string, just because the markup would need to be parsed out from the docstring, but then is attached as separate/extra zeekygen (?) metadata to the identifier, that feels a bit awkward.

evantypanski:
> Would this break other plugins?

Nope, `conn_uid` and `string` will always get implicitly converted, from what I can tell. I think even in C++-land the `string` types will be available, but I may be wrong. In fact I'm probably missing a `conn_uid` swap or two and we wouldn't notice. It's basically about self-documenting.

> but wonder if you thought about files/main.zeek?

My intention was that this would just start the idea "if you have a concept of an ID between two different logs, you create a new type to document it" - but you don't have to. I would really like to force the user to follow the convention, but I'd rather not break existing code

> instead a new attribute

I like that, but I'm just hesitant to introduce a new attribute. Part of the reason I like this approach is it's just convention rather than actual noticeable changes for now. I would like it if we could get to `conn_uid` requiring an explicit conversion from a string - then it's in the type system. But not yet (or ever)

I see one big difference with this approach to an attribute: the `&ref` attribute would require some "base" field (in your example `Conn::Info$uid`), whereas this is a loose set of interconnected fields - none are the "base" field. I think both share generality: If you want to link two fields together for a "pivot" you just make a new type. That's also what you'd do in other programming languages, you would use a `using ConnUID = std::string` type-thing, so I think it's more intuitive from other languages

`&ref` does have the benefit of not requiring the base-script to get modified as you say, which may be useful for plugins. But that's also where I'd say "but it's another attribute :(" - it's up to preference there, though. I'm actually not sure which I prefer semantically.. this is just easier

awelzel:
> Part of the reason I like this approach is it's just convention rather than actual noticeable changes for now.

> whereas this is a loose set of interconnected fields - none are the "base" field

Thanks for putting these down.

I think I'm just far over in the be explicit and specific corner and the fact that `&ref` (or `&pivot` or anything) would only be allowed on a record fields (maybe even `&log` enforced). If the `Record$field` syntax is implemented in a language server, it could also be used for pivoting :-)

Compared to "all fields of that type are somehow connected". The `conn_uid` type can also be used for a global, local, in function signatures, etc. where pivot/referencing isn't necessarily all that meaningful - I guess?

The docstring `:zeek:pivot:` approach would work for me, the one part I don't like is that we'd need some convention that `zeek:pivot` can only appear by itself in a single line in the docstring and we strip it out from the final result, but maybe that's fair compared to introducing an attribute just for a bit of metadata that's otherwise not functional in Zeek-script.

ckreibich:
Cool ideas. I think all of these have pros/cons. Yeah indeed — the goal is to be able to produce schemas and documentation that make the field relationship clear.

Regarding the `zeek:pivot` tag, I still like its close relationship to `zeek:field`. It affords some flexibility by decoupling from the exact type (as in the `set[string]` case in dhcp.log). I don't think we'd need to be too strict about it — the main difference I see over `zeek:field` is that it would indicate something about the entity it's in the docstring of (such as the http.log `uid` field), not just the thing it's pointing to (the conn.log `uid`). Fwiw I think there's a fair amount of scope in encoding additional information in docstrings — things like OCSF contexts, short/long descriptions, the high-level purpose of a field (say network info, enrichtment, etc).

Typing would in principle be my favorite since type systems are neat, but I'm not sure it fully nails it here. It doesn't immediately help with schemas and docs — we'd still have to code up logic somewhere that pulls together the information from what probably would be the docstring on the type alias itself. I.e., use of `conn_uid` in an `Info` record somewhere doesn't automatically explain that the field is a pivot. I do like that it would disambiguate things like the `set[conn_uid]`. I'm wondering if there are other places where working with a `conn_uid` would be advantageous. Can you think of any?

As for `&ref`, I like that it's explicitly about a given log field, which we can enforce in the parser, and which is what this mainly is about. It does have the slight ambiguity issue of e.g `set[string]`. We'd need to make Zeekygen aware of the attribute (easy). It would likely convey it as a string in an additional `record_field`? I admit I'm also generally wary about adding attributes, but the case looks pretty strong here.

Right now I'm slightly in favor of the tag, but not strongly — let's see if more arguments come up.

awelzel:
> It does have the slight ambiguity issue of e.g set[string].

It didn't seem seem terribly ambiguous to me. `&ref` on a field of a container type would imply a 1:N relationship (1 dhcp.log can point at N conn.log entries). That's how I had pictured it. One could still validate that the container's value type is the same or compatible with the type of the field that `&ref` points to.

I'm swinging between "a new attribute isn't really worth it" and "making Zeekygen clever enough to parse and ignore lines with `:zeek:pivot:` in a field's docstring is weird, because it's a [reStructured Text role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html)".

Or... we put everything on its head and start talking about [Golang's tags](https://go.dev/wiki/Well-known-struct-tags) and if we could add them as a concept to Zeek:
```
type Info: record {
    # references Conn::Info$uid, renders and parses in JSON as "Uid" instead.
    uid: string &log `pivot:"Conn::Info$uid" json:"Uid"`;
}
```

The JSON tag is mostly to exemplify that tags could be used for more use-cases down the road. And yeah, tags could be a   `table[string] of string` that is accessible via `record_field`. Tags could even be a generic way for Zeek distributors to encode their own metadata on record fields in a "standard but loose" way, so I'm sort of like, maybe that?

EDIT: Maybe even on global vars or options, but would need to expose on something else than `record_field`.

evantypanski:
I'll close this; I could go on a while but my current preference is some sort of Zeekygen annotation

I don't think it's worth it right now to change the language for this. Partly because it is a lot for something we should be able to do right now, but also because it locks us in a lot more than something just based on convention. If we want pivots to be a first-class concept within Zeek, then it's harder if something else is baked into the language.

I would like this PR's solution to be good enough, but our tooling is lacking right now so we can't get everything we need easily with this approach.

So my vote is for the least obtrusive and most convention-based proposal. Not saying it doesn't have downsides, but it just seems like the way with the least in our way now and in the future. Feel free to disagree here or on the original discussion, though :)

PR #4793, opened 2025-08-26
Fix ruff check legacy alias for pre-commit

timwoj:
Pre-commit's output looks like this:

```
ruff (legacy alias)........................................................Passed
```

This is because we're calling `ruff` directly in .pre-commit-config.yaml. The correct way to call it is `ruff-check`. This causes the pre-commit output to look like:

```
ruff check.................................................................Passed
```


timwoj:
I intentionally manually cancelled the Cirrus tasks for this PR since they don't matter.

PR #4798, opened 2025-08-28
CI: Bump clang version on ubuntu 24 VMs to clang 21

timwoj:


timwoj:
Scratch that, I forgot that clang-21 is only available from the LLVM apt repository. Clang-20 is the last one available directly from Ubuntu, and we cover that in the weekly builds. This can stay 19 for stability.

PR #4801, opened 2025-08-31
More extensive use of IDPtr's rather than ID*

vpax:
When debugging a recent ASAN issue that cropped up in script optimization, I realized that to fix it I'd either have to introduce some additional memory management around `ID*` pointers, or migrate much of the internals away from using raw pointers and instead to use `IDPtr`s. This PR does that latter, with one of the driving changes being switching `IDPList` from being a Zeek list of pointers to being a `std::vector<IDPtr>`, and changing `IDSet` from `std::unordered_set<const ID*>` to `std::unordered_set<IDPtr>`. These changes force a lot of follow-on changes, introducing a bunch of memory safety in the process.

`IDPList` used to be in `zeek::` rather than `detail::`, but that doesn't make sense to me because it was defined in terms of `detail::` list elements, so there was no way to use it without being sensitive to `detail::`. Because of that, I moved the (new) definition into `detail::`.

While I migrated all of the script optimization code to focus on `IDPtr`, and much of the interpreter code too, for the most part I left the `Trigger` class alone, because it does subtle things with reference counting that I didn't want to mess with.

Some of my changes are likely not micro-optimal, such as missing opportunities for `std::move()` usage, or for `IDPtr` references rather than copies. My own view is that these don't matter for non-Zeek-execution-time code because the overhead of script optimization compilation already swamps them. Related to this, _if_ `IDPtr` could be used in `std::set` then there are a few places where using that rather than `std::unordered_set` would probably save a few cycles, but again I don't think any of those actually matter for the compilation phase.

To test the changes necessitated doing some routine script optimization maintenance as well, which is included here  too.


src/ZeekList.h
```diff
@@ -20,7 +20,6 @@ class Type;
 
 using ValPList = PList<Val>;
 using ExprPList = PList<detail::Expr>;
-using IDPList = PList<detail::ID>;
```
timwoj, line 23: I'm wavering on whether we can move this without a deprecation cycle. I agree that it's based on something that's in the detail namespace and I already checked that no external plugins use it. Anything using it would be broken badly because the API completely changed underneath.

src/script_opt/CPP/Stmts.cc
```diff
@@ -82,7 +82,7 @@ void CPPCompile::GenInitStmt(const InitStmt* init) {
         auto type_type = TypeType(t);
         auto type_ind = GenTypeName(t);
 
-        if ( locals.count(aggr.get()) == 0 ) {
+        if ( locals.count(aggr) == 0 ) {
```
timwoj, line 85:
```suggestion
        if ( ! locals.contains(aggr) ) {
```

src/script_opt/ZAM/Driver.cc
```diff
@@ -47,13 +47,11 @@ void ZAMCompiler::Init() {
 }
 
 void ZAMCompiler::InitGlobals() {
-    for ( auto g : pf->Globals() ) {
-        auto non_const_g = const_cast<ID*>(g);
-
+    for ( auto& g : pf->Globals() ) {
         GlobalInfo info;
```
timwoj:
```suggestion
        GlobalInfo info{.id = g, .slot = AddToFrame(g)};
```

And then delete the two lines below.

timwoj (changes requested):
I like it quite a bit. You're going to run into a bunch of conflicts with master since the `contains()` PR merged.

src/ZeekList.h
vpax: (well you know my view ...)

src/script_opt/ZAM/Driver.cc
vpax: Nifty. I'm guessing this is new w/ C++20. I've pushed that change, and also the rebase (thanks for the help with that!).

timwoj: That's C++20's [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html#Designated_initializers).

vpax:
@timwoj I know it'll be a headache given the size, but this would be a good one for 8.0.2 since it fixes a script optimization memory safety issue ...

PR #4802, opened 2025-08-31
Tightening up parsing of $field=X record constructor expressions

vpax:
@stevesmoot  recently discovered that a statement like `print $foo=3;` blithely ignores the reference to the (unknown) `foo` record field. This is both dangerous for masking scripting bugs, and causes downstream errors in script optimization because it expects expressions to be well-formed. This PR changes the grammar to only allow `$field=` constructor expressions when in a constructor context, and to enforce uniformity that either all of the constructor elements use them, or none do.


evantypanski:
I think we could deprecate this, even though it's *probably* just a buggy mistake if users use it. I like that more since I find Bison error messages kind of cryptic, instead this could be a nice warning or deprecation.

It seems it's only `Eval`'d in a `RecordConstructorExpr` - and that will all be guaranteed to be `FieldAssignExpr` anyway. What do you think about implementing `FieldAssignExpr::Eval` to always warn with a deprecation and eval the `Op`, then make `RecordConstructorExpr::Eval` just change like this:

``` diff
ValPtr RecordConstructorExpr::Eval(Frame* f) const {
    if ( IsError() )
        return nullptr;

    const auto& exprs = op->Exprs();
    auto rt = cast_intrusive<RecordType>(type);

    if ( ! map && exprs.length() != rt->NumFields() )
        RuntimeErrorWithCallStack("inconsistency evaluating record constructor");

    auto rv = make_intrusive<RecordVal>(rt);

    for ( int i = 0; i < exprs.length(); ++i ) {
+       auto field = exprs[i]->AsFieldAssignExpr();
+       auto v_i = field->Op()->Eval(f);
-       auto v_i = exprs[i]->Eval(f);
        int ind = map ? (*map)[i] : i;

        if ( v_i && v_i->GetType()->Tag() == TYPE_VECTOR && v_i->GetType<VectorType>()->IsUnspecifiedVector() ) {
            const auto& t_ind = rt->GetFieldType(ind);
            v_i->AsVectorVal()->Concretize(t_ind->Yield());
        }

        rv->Assign(ind, v_i);
    }

    return rv;
}

```

Seems like a smaller change, with added benefits that we can go through deprecation (maybe not necessary) and a nicer error (which I like).

vpax:
Hmmm two points of disagreement. The first is that it doesn't correspond to any valid language construction in the first place, so in that sense there's nothing to deprecate. We just tightened up some type-checking (in this case, enforced by the grammar rather than semantic checking). It's not as though anyone could currently be getting benefit from the construct, other than by incidental luck.

Second, if we deprecate it, then I wonder if there's an expectation that it works at least for the deprecation cycle. If so, then seems conceptually it should be supported by script optimization, which currently bombs out on it (which is how we found the problem in the first place). Adding that support strikes me as undertaking some work that will in reality bare zero (0) fruit.

(Note that if we deprecate, we need to back out this change entirely, since the grammar change is the full change.)

evantypanski:
> it should be supported by script optimization, which currently bombs out on it

Ah fair point, deprecation probably isn't right then. I like the error from `Eval` more but it seems wrong to have this allowed in the parser at all.

FWIW I did not want to add any extra support for this :) just a nicer error, but probably not worth it

evantypanski (approved):

vpax:
> FWIW I did not want to add any extra support for this :) just a nicer error

Agreed - I likewise find the error message a bit obtuse/jarring, but I don't see any easy way to fix that while retaining the nice property that the grammar itself keeps the weird constructs from occurring. At first I was trying to figure out how I could add code to catch all the different places where such expressions _aren't_ valid, and then had the Aha that in fact the grammar could just prohibit it.

PR #4804, opened 2025-09-01
Topic/timw/windows zeromq support

timwoj:
This adds `zeromq` and `cppzmq` to the vcpkg configuration for Windows and re-enables building it as part of the Windows CI build. It also adds a new label to the CI configuration for enabling just the Windows task as part of a PR build.

### DIFF
--- a/src/cluster/backend/CMakeLists.txt
+++ b/src/cluster/backend/CMakeLists.txt
@@ -1,7 +1,5 @@
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/zeromq/cmake")
 
-find_package(ZeroMQ)
-
 # Default to building ZeroMQ only if ZeroMQ was found.
 #
 # If a user enabled the cluster backend explicitly (-D ENABLE_CLUSTER_BACKEND_ZEROMQ:bool=ON),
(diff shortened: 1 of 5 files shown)


src/cluster/backend/CMakeLists.txt
```diff
@@ -1,14 +1,18 @@
...
 # Default to building ZeroMQ only if ZeroMQ was found.
 #
 # If a user enabled the cluster backend explicitly (-D ENABLE_CLUSTER_BACKEND_ZEROMQ:bool=ON),
 # but ZeroMQ wasn' found, hard bail.
 option(ENABLE_CLUSTER_BACKEND_ZEROMQ "Enable the ZeroMQ cluster backend" ON)
 
 if (ENABLE_CLUSTER_BACKEND_ZEROMQ)
+
+    if (MSVC)
+        find_package(ZeroMQ CONFIG REQUIRED)
```
ckreibich, line 8: Out of curiosity, why the config mode on `find_package()` in Windows?

src/cluster/backend/CMakeLists.txt
```diff
@@ -1,14 +1,18 @@
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/zeromq/cmake")
-
-find_package(ZeroMQ)
-
 # Default to building ZeroMQ only if ZeroMQ was found.
 #
 # If a user enabled the cluster backend explicitly (-D ENABLE_CLUSTER_BACKEND_ZEROMQ:bool=ON),
```
ckreibich: Nit: that comment seems a bit out of date since the ZeroMQ backend no longer needs explicit enabling but is enabled by default.

src/cluster/backend/CMakeLists.txt
timwoj: It's something about how their package is built with vcpkg. Without it, it doesn't set some of the CMake variables correctly.

PR #4807, opened 2025-09-02
Remove unnecessary peer signature from test `scripts.base.protocols.bittorrent.tracker`

bbannier:


bbannier:
> Please add some context why these are unnecessary. Not familiar with the protocol, aren't we now just not recognizing the 2nd variant of the protocol anymore?

This is a test, so it does not change any real detection. I am unsure why this test adds these signatures as it does not depend on them at all.

rsmmr:
> This is a test, so it does not change any real detection. I am unsure why this test adds these signatures as it does not depend on them at all.

Got it. I completely missed that this is a test-only change!

PR #4815, opened 2025-09-04
Switch types used in DNS analyzer to be more consistent

timwoj:
I'm starting work on the DNS Dynamic Update work and this was all bothering me about working on the DNS analyzer. This PR switches the types in the code to use the standard `uint*_t` types consistently instead of switching between them and regular `unsigned {short,int}` types. It also cleans up the init of the analyzer a little bit. There are probably some opportunities for packing the structures a bit better, but I'll leave that for later.

### DIFF
--- a/src/analyzer/protocol/dns/DNS.cc
+++ b/src/analyzer/protocol/dns/DNS.cc
@@ -52,9 +52,7 @@ namespace detail {
 // since it's similar to DNS but does some things differently.
 constexpr int NETBIOS_PORT = 137;
 
-DNS_Interpreter::DNS_Interpreter(analyzer::Analyzer* arg_analyzer) {
-    analyzer = arg_analyzer;
-    first_message = true;
+DNS_Interpreter::DNS_Interpreter(analyzer::Analyzer* arg_analyzer) : analyzer(arg_analyzer) {
     is_netbios =
         ntohs(analyzer->Conn()->OrigPort()) == NETBIOS_PORT || ntohs(analyzer->Conn()->RespPort()) == NETBIOS_PORT;
 }
@@ -2184,8 +2182,6 @@ DNS_Analyzer::DNS_Analyzer(Connection* conn) : analyzer::tcp::TCP_ApplicationAna
 }
 
 DNS_Analyzer::~DNS_Analyzer() { delete interp; }
-
-void DNS_Analyzer::Init() {}
 
 void DNS_Analyzer::Done() {
     analyzer::tcp::TCP_ApplicationAnalyzer::Done();
--- a/src/analyzer/protocol/dns/DNS.h
+++ b/src/analyzer/protocol/dns/DNS.h
@@ -393,9 +393,9 @@ protected:
     void SendReplyOrRejectEvent(detail::DNS_MsgInfo* msg, EventHandlerPtr event, const u_char*& data, int& len,
                                 String* question_name, String* original_name);
 
-    analyzer::Analyzer* analyzer;
-    bool first_message;
-    bool is_netbios;
+    analyzer::Analyzer* analyzer = nullptr;
+    bool first_message = true;
+    bool is_netbios = false;
 };
 
 enum TCP_DNS_state : uint8_t {
@@ -438,7 +438,6 @@ public:
 
     void DeliverPacket(int len, const u_char* data, bool orig, uint64_t seq, const IP_Hdr* ip, int caplen) override;
 
-    void Init() override;
     void Done() override;
     void ConnectionClosed(analyzer::tcp::TCP_Endpoint* endpoint, analyzer::tcp::TCP_Endpoint* peer,
                           bool gen_event) override;


src/analyzer/protocol/dns/DNS.cc
```diff
@@ -381,7 +382,7 @@ bool DNS_Interpreter::ExtractLabel(const u_char*& data, int& len, u_char*& name,
         return false;
 
     const u_char* orig_data = data;
-    int label_len = data[0];
+    uint16_t label_len = data[0];
```
bbannier: Maybe more of a style question, but `data` is a `const u_char*`, so typing the LHS implicitly changes the type (probably no ill effect here). WDYT about sticking to `auto` on the LHS for all changes in this patch? If needed we could still cast the RHS with e.g., an explict `static_cast` which would make clearer that type changes are intentional.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1295,12 +1296,12 @@ bool DNS_Interpreter::ParseRR_DS(detail::DNS_MsgInfo* msg, const u_char*& data,
...
-    unsigned int ds_key_tag = ExtractShort(data, len);
+    uint16_t ds_key_tag = ExtractShort(data, len);
     // split the two bytes for algorithm and digest type extraction
-    uint32_t ds_algo_dtype = ExtractShort(data, len);
-    unsigned int ds_algo = (ds_algo_dtype >> 8) & 0xff;
-    unsigned int ds_dtype = ds_algo_dtype & 0xff;
-    detail::DNSSEC_Digest ds_digest_type = detail::DNSSEC_Digest(ds_dtype);
+    uint16_t ds_algo_dtype = ExtractShort(data, len);
+    uint16_t ds_algo = (ds_algo_dtype >> 8) & 0xff;
+    uint16_t ds_dtype = ds_algo_dtype & 0xff;
```
bbannier: These two now actually narrow.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1373,9 +1374,9 @@ bool DNS_Interpreter::ParseRR_SSHFP(detail::DNS_MsgInfo* msg, const u_char*& dat
     if ( len < 2 )
         return false;
 
-    uint32_t algo_fptype = ExtractShort(data, len);
-    unsigned int algo = (algo_fptype >> 8) & 0xff;
-    unsigned int fptype = algo_fptype & 0xff;
+    uint16_t algo_fptype = ExtractShort(data, len);
+    uint16_t algo = (algo_fptype >> 8) & 0xff;
+    uint16_t fptype = algo_fptype & 0xff;
```
bbannier: Narrowing.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1400,14 +1401,14 @@ bool DNS_Interpreter::ParseRR_LOC(detail::DNS_MsgInfo* msg, const u_char*& data,
         return false;
 
     // split the two bytes for version and size extraction
-    uint32_t ver_size = ExtractShort(data, len);
-    unsigned int version = (ver_size >> 8) & 0xff;
-    unsigned int size = ver_size & 0xff;
+    uint16_t ver_size = ExtractShort(data, len);
+    uint16_t version = (ver_size >> 8) & 0xff;
+    uint16_t size = ver_size & 0xff;
```
bbannier: Narrowing.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1400,14 +1401,14 @@ bool DNS_Interpreter::ParseRR_LOC(detail::DNS_MsgInfo* msg, const u_char*& data,
...
 
     // split the two bytes for horizontal and vertical precision extraction
-    uint32_t horiz_vert = ExtractShort(data, len);
-    unsigned int horiz_pre = (horiz_vert >> 8) & 0xff;
-    unsigned int vert_pre = horiz_vert & 0xff;
+    uint16_t horiz_vert = ExtractShort(data, len);
+    uint16_t horiz_pre = (horiz_vert >> 8) & 0xff;
+    uint16_t vert_pre = horiz_vert & 0xff;
```
bbannier: Narrowing.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -1559,8 +1560,8 @@ bool DNS_Interpreter::ParseRR_CAA(detail::DNS_MsgInfo* msg, const u_char*& data,
         return true;
     }
 
-    unsigned int flags = ExtractShort(data, len);
-    unsigned int tagLen = flags & 0xff;
+    uint16_t flags = ExtractShort(data, len);
+    uint16_t tagLen = flags & 0xff;
```
bbannier: Narrowing.

src/analyzer/protocol/dns/DNS.cc
timwoj: > WDYT about sticking to `auto` on the LHS for all changes in this patch?

Would you want all of the other similar typed variables to change that weren't changed in this patch?

bbannier: Yeah, I think I'd prefer that long-term, but I suspect it might trigger more work downstream.

Since in this case you switched a couple types let's just stick to not changing types in places we touch in opaque ways, either by using `auto` for the changes, or by mirroring the actual types.

src/analyzer/protocol/dns/DNS.cc
timwoj: How so? `ds_algo_dtype` is a `uint16_t` itself. If anything, these could be `static_cast`'d down to uint8_t, since we're already cutting off the top 8-bits of the original value. Same with more than one of the others below.

bbannier: I was just going with clangd's type hints, but writing an example confirms that. It looks like already the shift operation returns an `int`, regardless of the type of `8`.

```cpp
#include <cstdint>
#include <type_traits>

int main() {
	auto x = uint16_t(0) >> 8;
	static_assert(std::is_same_v<decltype(x), int>);
}
```

I couldn't find an exact reference, but suspect this is related to implementation-defined promotion of `uint16_t`.

timwoj: That's... dumb. I considered making all of those explicit `static_cast`s to `uint8_t` in the original version of this commit, but figured just setting them as `uint16_t` would be sufficient.

timwoj: It's also sorta weird that clang-tidy doesn't complain about the narrowing. I enabled that check before my first comment in this thread because I was curious.

src/analyzer/protocol/dns/DNS.cc
timwoj: I switched all of the calls to `Extract{Long,Short}` to use `auto`, plus any of the shifts. There's prior art here, since some of the existing calls to those methods also used `auto`. I also fixed a few more of the casts to enums to use auto.

src/analyzer/protocol/dns/DNS.cc
bbannier: > That's... dumb.

That's just the rotten smell of C which C++ decided to never contain 😞

PR #4817, opened 2025-09-05
CI: Add tasks for reporting cron status to email

timwoj:
We currently have no way to know that a build started by one of the nightly or weekly cron jobs failed. This adds two new builds to the nightly/weekly builds that report status on failures via email, similar to the regular PR/master builds. I'm not a big fan of the `depends_on` section in the nightly one, but it's the only way to do it that I can see.


timwoj:
After some spelunking by @ckreibich, we determined that we definitely are getting emails already about cron builds, but they're not including the right data. For example the weekly job says that 28 tasks succeeded, but the build only actually run 2 tasks. Also, the emails could stand to have some more information in them about the build environment.

PR #4825, opened 2025-09-11
Add support for DNS Dynamic Update messages

### LINKED ISSUE #3856: DNS Dynamic Update: Produces weird and is not confirmed

timwoj:
Fixes #3856

### DIFF
--- a/scripts/base/protocols/dns/main.zeek
+++ b/scripts/base/protocols/dns/main.zeek
@@ -357,8 +357,8 @@ hook set_session(c: connection, msg: dns_msg, is_query: bool) &priority=5
 
 event dns_message(c: connection, is_orig: bool, msg: dns_msg, len: count) &priority=5
 	{
-	if ( msg$opcode != 0 && msg$opcode != 5 )
-		# Currently only standard queries are tracked.
+	if ( msg$opcode != DNS_OP_QUERY && msg$opcode != DNS_OP_DYNAMIC_UPDATE )
+		# Currently only standard queries and dynamic updates are tracked.
 		return;
 
 	hook set_session(c, msg, ! msg$QR);
@@ -366,8 +366,8 @@ event dns_message(c: connection, is_orig: bool, msg: dns_msg, len: count) &prior
 
 hook DNS::do_reply(c: connection, msg: dns_msg, ans: dns_answer, reply: string) &priority=5
 	{
-	if ( msg$opcode != 0 && msg$opcode != 5 )
-		# Currently only standard queries are tracked.
+	if ( msg$opcode != DNS_OP_QUERY && msg$opcode != DNS_OP_DYNAMIC_UPDATE )
+		# Currently only standard queries and dynamic updates are tracked.
 		return;
 
 	if ( ! msg$QR )
@@ -398,7 +398,7 @@ hook DNS::do_reply(c: connection, msg: dns_msg, ans: dns_answer, reply: string) 
 
 	if ( reply != "" )
 		{
-		if ( msg$opcode == 5 && ans$answer_type == DNS_UPDATE && ! starts_with(reply, "del:") )
+		if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ans$answer_type == DNS_UPDATE && ! starts_with(reply, "del:") )
 			reply = fmt("add: %s", reply);
 
 		if ( ! c$dns?$answers )
@@ -433,8 +433,8 @@ event dns_end(c: connection, msg: dns_msg) &priority=-5
 
 event dns_request(c: connection, msg: dns_msg, query: string, qtype: count, qclass: count) &priority=5
 	{
-	if ( msg$opcode != 0 )
-		# Currently only standard queries are tracked.
+	if ( msg$opcode != DNS_OP_QUERY && msg$opcode != DNS_OP_DYNAMIC_UPDATE )
+		# Currently only standard queries and dynamic updates are tracked.
 		return;
 
 	c$dns$RD          = msg$RD;
@@ -473,7 +473,7 @@ event dns_unknown_reply(c: connection, msg: dns_msg, ans: dns_answer) &priority=
 
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
 	{
-	if ( msg$opcode == DNS_UPDATE && ! msg$is_netbios )
+	if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
 		hook DNS::do_reply(c, msg, ans, fmt("A %s %s", ans$query, a));
 	else
 		hook DNS::do_reply(c, msg, ans, fmt("%s", a));
@@ -511,7 +511,7 @@ event dns_SPF_reply(c: connection, msg: dns_msg, ans: dns_answer, strs: string_v
 
 event dns_AAAA_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
 	{
-	if ( msg$opcode == DNS_UPDATE && ! msg$is_netbios )
+	if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
 		hook DNS::do_reply(c, msg, ans, fmt("AAAA %s %s", ans$query, a));
 	else
 		hook DNS::do_reply(c, msg, ans, fmt("%s", a));
@@ -519,7 +519,7 @@ event dns_AAAA_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &pri
 
 event dns_A6_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
 	{
-	if ( msg$opcode == DNS_UPDATE && ! msg$is_netbios )
+	if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
 		hook DNS::do_reply(c, msg, ans, fmt("A6 %s %s", ans$query, a));
 	else
 		hook DNS::do_reply(c, msg, ans, fmt("%s", a));
--- a/scripts/base/protocols/dns/consts.zeek
+++ b/scripts/base/protocols/dns/consts.zeek
@@ -195,6 +195,13 @@ export {
 		[6] = "ipv6hint",
 	} &default = function(n: count): string { return fmt("key-%d", n); };
 
+	const DNS_OP_QUERY = 0;
+	const DNS_OP_IQUERY = 1;
+	const DNS_OP_SERVER_STATUS = 2;
+	const DNS_OP_NOTIFY = 4;
+	const DNS_OP_DYNAMIC_UPDATE = 5;
+	const DNS_OP_DSO = 6;
+
 	## Mapping of DNS operation type codes to human readable string representation.
 	const opcodes = {
 		[0] = "query",


scripts/base/protocols/dns/main.zeek
```diff
@@ -343,11 +347,20 @@ hook set_session(c: connection, msg: dns_msg, is_query: bool) &priority=5
 		if ( msg$rcode != 0 && msg$num_queries == 0 )
 			c$dns$rejected = T;
 		}
+
+	c$dns$opcode = msg$opcode;
+	if ( msg$is_netbios )
+		if ( msg$opcode >= 5 )
+			c$dns$opcode_name = opcodes[msg$opcode + 0xFFFF];
```
awelzel: I wonder if a separate `netbios_opcodes` table would be easier.

This is pretty tricky to grasp and I'm actually confused if it's correct. Can the `0xFFFF5` from the table ever happen? On first sight, `5 + 0xFFFF`, but that's `'0x100fe'` and not `0xFFFF5`.

If there's separate tables, then `is_netbios` would suffice to distinguish which one to use and adding entries via `redef` wouldn't need the `0xFFFF` tricks.

testing/btest/scripts/base/protocols/dns/dynamic-update.zeek
```diff
@@ -1,6 +1,12 @@
...
+# @TEST-EXEC: btest-diff out
+# @TEST-EXEC: ! test -f weird.log
 
 @load base/frameworks/notice/weird
 @load base/protocols/dns
+
+event dns_dynamic_update(c: connection, msg: dns_msg, zname: string, zclass: count)
+	{
+	print msg, zname, zclass, DNS::classes[zclass];
+	}
```
awelzel, line 13: I'm only skimming. The juicy info I believe is in the updates section in the the test pcap - `nwin2.stratolab.org` and 192.168.1.106 as IP.


I wonder if `dns.log` is even the right place to put this, or whether a dedicated `dns_update.log` might be better. Maybe adding `dns.log` to this test would give some more ground/example to discuss, but I could certainly see `dns.log` being just about "query" and going somewhere else for "update".

image

awelzel (commented):
Not sure about `opcode` and `opcode_name` additions. Maybe that's worth a discussion how to expose the info?

testing/btest/scripts/base/protocols/dns/dynamic-update.zeek
timwoj: The problem with the logging the actual update data is that it's variable what's in there. As in your screenshot, should we log all three of those updates? Just the last one that has an IP in it? Should it be an array of some record that's the name, type, class, TTL, and whatever data is after without decoding it?

awelzel: If we go for a separate `dns_update.log`, I guess I could see starting with just a log that lists each of them
```
uid | transaction_id (top-level) | class | type | name | data | success
```

Or grouping by transaction_id as you indicate. I think this is log design and might be good to have a conversation with the team / users outside of the PR?

ekoyle: Doesn't DNS technically allow for multiple queries in a single request? How do we handle that?

I think the real problem here may be that our log format was based around common usage patterns rather than the actual protocol. I think we could probably do something similar here.

Even though it contains a lot of protocol information, the DNS log is only storing a summary, not all of the protocol information. It does try to store most of the query, but it drops nearly all information from the reply. This asymmetry is kind of unfortunate.

It looks like the TTL of 0 with no data means the record is not present or should be deleted, depending on the section. These are probably not really interesting in general.

If I'm understanding correctly for the packet in the screenshot, prerequisite is "no CNAME present for this name" to avoid accidentally clobbering something added manually, the updates are "delete A and AAAA records by this name, then set an A record" and the additional contains authentication data.

Functionally, all of these operations together are probably a minimal example to set a single A record, which I think is the only actual information worth logging from this request. The check/delete for the same name will likely always be present, so logging that is going to waste a lot of space in logs and also make the logs much more difficult to use.

I think we should treat this more like we treat a DNS response (it looks an awful lot like a bunch of RRSETs, after all). In a DNS response, a message like this would be condensed down to a list of "answers" which really only contains the data field from each of the returned records.

Although the protocol supports setting more than one name here, it also supports more than one query in a single packet -- I'd rather see that split into more than one log message than try to change the log format to support what I expect is an uncommon use-case.

ekoyle: After more discussion (and looking at some packets), I agree that these DNS messages have almost nothing in common with our dns log format :disappointed: .

timwoj: I copied the above two comments into the log-format discussion at https://github.com/zeek/zeek/discussions/4935.

timwoj:
I think we reached a solid place with the logging. My opinion is to get this in as-is, let people play with it, and we can make adjustments as needed during the 8.2 and 9.0 dev cycles.

src/analyzer/protocol/dns/DNS.h
```diff
@@ -304,27 +316,30 @@ class DNS_MsgInfo {
...
+    uint16_t rcode;                 ///< return code, see DNS_Code
+    bool QR;                        ///< query record flag
+    bool AA;                        ///< authoritative answer flag
+    bool TC;                        ///< truncated - size > 512 bytes for udp
+    bool RD;                        ///< recursion desired
+    bool RA;                        ///< recursion available
+    uint8_t Z;                      ///< 3 bit field (includes AD and CD)
+    bool AD;                        ///< authentic data
+    bool CD;                        ///< checking disabled
+    uint16_t qd_zo_count;           ///< number of questions (or zones for dynamic update)
```
awelzel, line 329: Mostly wondering if this ends up a bit easier when using them? If you disagree resolve. No strong opinion here.
```
union {
  struct q {
    uint16_t qdcount;
    uint16_t anscount;
    uint16_t nsscount;
    ...;
  } query;
  struct u {
    uint16_t zocount;  // zones
    uint16_t prcount;  // prereq
    uint16_t upcount;  // updates
  } update;
};
uint16_t arcount;
```

scripts/base/protocols/dns/main.zeek
```diff
@@ -468,7 +468,10 @@ event dns_unknown_reply(c: connection, msg: dns_msg, ans: dns_answer) &priority=
 
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
 	{
-	hook DNS::do_reply(c, msg, ans, fmt("%s", a));
+	if ( msg$opcode == 5 && ! msg$is_netbios )
```
awelzel: Also above, could we use `DNS_UPDATE` instead of magic 5?

scripts/base/protocols/dns/main.zeek
```diff
@@ -650,6 +653,12 @@ event dns_rejected(c: connection, msg: dns_msg, query: string, qtype: count, qcl
 		c$dns$rejected = T;
 	}
 
+event dns_dynamic_update_deletion(c: connection, msg: dns_msg, ans: dns_answer) &priority=5
+	{
+	local s: string = fmt("DELETION: %s %s", query_types[ans$qtype], ans$query);
+	hook DNS::do_reply(c, msg, ans, s);
+	}
+
```
awelzel, line 704: I think I like the inline encoding. Also because we already free-text these entries.

Here's a thought, can we make it a bit more structured, like:
```
update-add: ...
update-del: RRSet ANY|<type> <name>
update-del: RR <...>
# future:
update-pre: ...
```
Proposing lower-case here for the first part because as there's `<unknown type>` and the upper case seems to be used for the record types and feels nicer to not mix them. `DELETION` also screams a bit.

The `$query` field is actually optional today, so we could keep it unset and because `update-xxx` is in `answers` that should allow figuring things out. And then I wonder a bit about the opcode/opcode_name columns and whether they are needed. I'm definitely in the avoiding-new-columns-corner here and think they'd be quite easy to add via a separate policy script, rather than forcing them on everyone.

```
@@ -377,7 +378,7 @@ hook DNS::do_reply(c: connection, msg: dns_msg, ans: dns_answer, reply: string)
 
        if ( ans$answer_type == DNS_ANS || ans$answer_type == DNS_UPDATE )
                {
-               if ( ! c$dns?$query )
+               if ( ! c$dns?$query && msg$opcode == 0)
                        c$dns$query = ans$query;
```

This then results in something like this (without the rr/rrset distinction because I didn't look enough if that info is available).
```
{
  "ts": 1676968917.288586,
  "uid": "CzUuXA31xC1fnjafk2",
  "id.orig_h": "192.168.1.105",
  "id.orig_p": 62763,
  "id.resp_h": "192.168.1.108",
  "id.resp_p": 53,
  "proto": "udp",
  "trans_id": 61191,
  "rtt": 0.001146078109741211,
  "rcode": 0,
  "rcode_name": "NOERROR",
  "AA": false,
  "TC": false,
  "RD": false,
  "RA": false,
  "Z": 0,
  "answers": [
    "update-del: AAAA nwin1.stratolab.org",
    "update-del: A nwin1.stratolab.org",
    "update-add: nwin1.stratolab.org 192.168.1.105"
  ],
  "TTLs": [
    0.0,
    0.0,
    1200.0
  ],
  "rejected": false
}
```

timwoj: The `query` field is being used to hold the zone that the update is touching. It's typically a higher-level than the hosts that are in the actual update. I think that's relevant information. The only part of the data that's missing is the prereqs, but I don't think they're needed at a glance.

The `opcode` is there mostly so you can easily filter on it'. Without it, you have to know to search for `update-` in the answers in your example. It also becomes a bigger deal if we ever decide to support the other opcode types (server status, notify, dso). I could see leaving out the string version, but we already have a string field for `rcode` so it felt right to add it.

timwoj:
> I'm also wondering if dedicated events would be nicer if someone wants to actually build a dedicated dns_update.log (e.g., dns_update_add(...), dns_update_del_rr(), but I haven't fully thought this through and would first try to prototype it, or it could be done as a follow-up.

It's not a bad idea, but I agree that it could be done as a follow-up vs holding this PR for it.

scripts/base/protocols/dns/main.zeek
awelzel: > The query field is being used to hold the zone that the update is touching
Thanks!

> The opcode is there mostly so you can easily filter on it'. 

Yeah, I'm coming around and agree. It's possiblye to `redef record DNS::Info$opcode -= { &log; }` these days if someone really doesn't want them. Agree on consistency with rcode.

I think I'd still prefer short lower-case `add:`, `del:` and maybe `pre:` in the future as prefixes, but agree the `update-` thing is redundant.

IMO it'd be nice to distinguish the different delete actions (delete all rrsets, delete one rrset, delete a specific rr). Something like the following?

```
diff --git a/scripts/base/protocols/dns/main.zeek b/scripts/base/protocols/dns/main.zeek
index e8f93313bd..78cc800125 100644
--- a/scripts/base/protocols/dns/main.zeek
+++ b/scripts/base/protocols/dns/main.zeek
@@ -469,7 +469,7 @@ event dns_unknown_reply(c: connection, msg: dns_msg, ans: dns_answer) &priority=
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
        {
        if ( msg$opcode == 5 && ! msg$is_netbios )
-               hook DNS::do_reply(c, msg, ans, fmt("A %s %s", ans$query, a));
+               hook DNS::do_reply(c, msg, ans, fmt("add: %s %s %s", query_types[ans$qtype], ans$query, a));
        else
                hook DNS::do_reply(c, msg, ans, fmt("%s", a));
        }
@@ -655,7 +655,20 @@ event dns_rejected(c: connection, msg: dns_msg, query: string, qtype: count, qcl
 
 event dns_dynamic_update_deletion(c: connection, msg: dns_msg, ans: dns_answer) &priority=5
        {
-       local s: string = fmt("DELETION: %s %s", query_types[ans$qtype], ans$query);
+       print(ans);
+       local what = "";
+       if ( ans$qclass == DNS::ANY  && ans$qtype == DNS::ANY )
+               what = "RRSet *";
+       else if ( ans$qclass == DNS::NONE )
+               # When qclass is NONE, this is deleting an RR, but we
+               # do not have information about rdlength and rdata.
+               what = fmt("RR %s (?)", query_types[ans$qtype]);
+       else
+               # Delete all entries of a specific RRSet.
+               what = fmt("RRSet %s", query_types[ans$qtype]);
+
+
+       local s: string = fmt("del: %s %s", what, ans$query);
        hook DNS::do_reply(c, msg, ans, s);

```

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -250,14 +301,54 @@ bool DNS_Interpreter::ParseAnswer(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+            // 3. An RR being deleted from an RRset must have a class of NONE and TTL of zero. They're
+            //    otherwise normal RRs.
+            if ( msg->aclass == DNS_CLASS_ANY && msg->ttl == 0 && rdlength == 0 ) {
+                if ( dns_dynamic_update_deletion && ! msg->skip_event )
+                    analyzer->EnqueueConnEvent(dns_dynamic_update_deletion, analyzer->ConnVal(), msg->BuildHdrVal(),
+                                               msg->BuildAnswerVal());
+
+                // emit a reply event with the limited info we have
+                return true;
+            }
```
awelzel, line 337: So I think it'd be good to have the dedicated `dns_dynamic_update_add()` event here even if it's possibly to figure it out via other events. Just seems more consistent and less mind-bending.

Would probably also rename the `_deletion()` above to `_del()`, but that's cosmetic.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -250,14 +301,54 @@ bool DNS_Interpreter::ParseAnswer(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+            else if ( ! ((msg->aclass == DNS_CLASS_NONE && msg->ttl == 0) || (msg->aclass == msg->zclass)) ) {
+                analyzer->Weird("DNS_dynamic_update_invalid_update");
```
awelzel, line 339: Might be fine t to lift that into the script layer.

src/analyzer/protocol/dns/DNS.cc
```diff
@@ -250,14 +301,54 @@ bool DNS_Interpreter::ParseAnswer(detail::DNS_MsgInfo* msg, const u_char*& data,
...
+                return true;
+            }
+        }
+        else if ( msg->answer_type == DNS_PREREQUISITES ) {
+            // TODO: Do we need to do anything with this info? We're not logging it anywhere.
```
awelzel: As discussed elsewhere, seems totally reasonable to have a `_pre()` event.

src/analyzer/protocol/dns/DNS.cc
timwoj: The problem is that the other events add additional formatting to the output that goes into the `answer` field. For example, see the handling for TXT replies: https://github.com/zeek/zeek/blob/master/scripts/base/protocols/dns/main.zeek#L464-L477. That's why I opted to piggy-back off of the existing events for the rest.

src/analyzer/protocol/dns/DNS.cc
timwoj: This does seem reasonable. The data in prereqs aren't full RRs, so they don't depend on the other parsing functions.

src/analyzer/protocol/dns/DNS.cc
timwoj: Actually, I wonder if the change I made in `dns_A_reply` should be lifted into `do_reply`.

scripts/base/protocols/dns/main.zeek
timwoj: That would work well for the pre-req section as well.

scripts/base/protocols/dns/main.zeek
```diff
@@ -458,7 +470,10 @@ event dns_unknown_reply(c: connection, msg: dns_msg, ans: dns_answer) &priority=
 
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
 	{
-	hook DNS::do_reply(c, msg, ans, fmt("%s", a));
+	if ( msg$opcode == DNS_UPDATE && ! msg$is_netbios )
+		hook DNS::do_reply(c, msg, ans, fmt("A %s %s", ans$query, a));
```
awelzel, line 477: Would still prefix this with `add: ` for consistency. Then one can process dns.log by
1) Look at the opcode
2) if opcode is for dynamic-update, can split all entries in the array with `op, data = lsplit(":', 1)` or so.

If there's no prefix for the `add` ops, need to special case or assume "add" during processing.

PS: Ah, just saw the comment about lifting into `hook DNS::do_reply()` - you're still looking at this? I'd be happy to get this (preferably with the `add:` prefix) and see if it falls over in some testing environment :-)

timwoj: > PS: Ah, just saw the comment about lifting into hook DNS::do_reply() - you're still looking at this? I'd be happy to get this (preferably with the add: prefix) and see if it falls over in some testing environment.

I ended up not needing to lift it. I just added a check to `do_reply()` to prefix with `add:` if `del:` wasn't already there. That should cover it.

awelzel: Would you see anything wrong with this? I'm suspecting we might rework this part towards 9.0, so seems reasonable to centralize it a bit.

```
diff --git a/scripts/base/protocols/dns/main.zeek b/scripts/base/protocols/dns/main.zeek
index cb26bbb71d..fab14a983a 100644
--- a/scripts/base/protocols/dns/main.zeek
+++ b/scripts/base/protocols/dns/main.zeek
@@ -398,8 +398,8 @@ hook DNS::do_reply(c: connection, msg: dns_msg, ans: dns_answer, reply: string)
 
        if ( reply != "" )
                {
-               if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ans$answer_type == DNS_UPDATE && ! starts_with(reply, "del:") )
-                       reply = fmt("add: %s", reply);
+               if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ans$answer_type == DNS_UPDATE && ans$qclass != ANY && ans$qclass != NONE )
+                       reply = fmt("add: %s %s %s", query_types[ans$qtype], ans$query, reply);
 
                if ( ! c$dns?$answers )
                        c$dns$answers = vector();
@@ -473,10 +473,7 @@ event dns_unknown_reply(c: connection, msg: dns_msg, ans: dns_answer) &priority=
 
 event dns_A_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
        {
-       if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
-               hook DNS::do_reply(c, msg, ans, fmt("A %s %s", ans$query, a));
-       else
-               hook DNS::do_reply(c, msg, ans, fmt("%s", a));
+       hook DNS::do_reply(c, msg, ans, fmt("%s", a));
        }
 
 event dns_TXT_reply(c: connection, msg: dns_msg, ans: dns_answer, strs: string_vec) &priority=5
@@ -511,18 +508,12 @@ event dns_SPF_reply(c: connection, msg: dns_msg, ans: dns_answer, strs: string_v
 
 event dns_AAAA_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
        {
-       if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
-               hook DNS::do_reply(c, msg, ans, fmt("AAAA %s %s", ans$query, a));
-       else
-               hook DNS::do_reply(c, msg, ans, fmt("%s", a));
+       hook DNS::do_reply(c, msg, ans, fmt("%s", a));
        }
 
 event dns_A6_reply(c: connection, msg: dns_msg, ans: dns_answer, a: addr) &priority=5
        {
-       if ( msg$opcode == DNS_OP_DYNAMIC_UPDATE && ! msg$is_netbios )
-               hook DNS::do_reply(c, msg, ans, fmt("A6 %s %s", ans$query, a));
-       else
-               hook DNS::do_reply(c, msg, ans, fmt("%s", a));
+       hook DNS::do_reply(c, msg, ans, fmt("%s", a));
        }
```

awelzel (approved):
I have one more suggestion to centralize the `add: ` formatting that seems to work here. I think it makes sense. Leaving a :shipit: 

Maybe ask in the testing call whether people have Windows / DNS update traffic and could run latest master to see how things end up? :-)

scripts/base/protocols/dns/main.zeek
timwoj: The problem with this is that other RRs insert their name already also (see LOC and TXT). I'm going to merge it as-is and open a ticket to rework it all as a whole.

PR #4828, opened 2025-09-12
Mechanisms for suppressing script optimization

vpax:
To date there have been ways of targeting script optimization at specific functions or files, but no way to specify exceptions to those specifications. This PR adds two function attributes, `&no_ZAM_opt` and `&no_CPP_opt` to suppress ZAM or `-O gen-C++` script optimization for particular functions; and two command-line options, `--no-opt-files=` and `--no-opt-funcs=` . These latter override any `--optimize-files=` or `--optimize-funcs=` specifiers.


src/Attr.cc
```diff
@@ -541,6 +553,15 @@ bool Attributes::CheckAttr(Attr* a) {
                 return AttrError("&ordered only applicable to tables");
             break;
 
+        case ATTR_NO_ZAM_OPT:
+        case ATTR_NO_CPP_OPT: {
+            bool is_no_zam = a->Tag() == ATTR_NO_ZAM_OPT;
```
timwoj: Nit: the `bool` can go inside the `if` since that's the only place it's used.

src/Options.cc
```diff
@@ -124,9 +124,15 @@ void usage(const char* prog) {
     printf(
         "    -0|--optimize-files=<pat>       | enable script optimization for all "
         "functions in files with names containing the given pattern\n");
+    printf(
+        "       --no-opt-files=<pat>         | skip script optimization for all "
```
timwoj: I think these both should be `-no-optimize-*` to match the existing options.

src/Options.cc
```diff
@@ -393,6 +399,8 @@ Options parse_cmdline(int argc, char** argv) {
         {"re-level", required_argument, nullptr, 'T'},
         {"watchdog", no_argument, nullptr, 'W'},
         {"print-id", required_argument, nullptr, 'I'},
+        {"no-opt-funcs", required_argument, nullptr, '\001'},
```
timwoj: Is this passing a non-printable character to configure? I get that we've used a lot of the alphabet, but we can't have to resort to that yet, do we?

testing/btest/language/skip-opt1.zeek
```diff
@@ -0,0 +1,24 @@
+# @TEST-EXEC: zeek -b %INPUT >out 2>&1
```
timwoj, line 1: Can these three tests be combined into one that runs `zeek` three times? It's the same script code, so it seems weird to duplicate it.

testing/btest/opt/opt-skip-files.zeek
```diff
@@ -0,0 +1,16 @@
+# @TEST-REQUIRES: test "${ZEEK_USE_CPP}" != "1"
```
timwoj, line 1: Same as above re: combining these into one test.

timwoj (changes requested):

src/Options.cc
vpax: It seemed better to me to use something clearly non-mnemonic rather than try to either find something vaguely mnemonic, or grab something printable that down the road might have a better fit to something else. But I'm fine with switching this to printable if desired

testing/btest/language/skip-opt1.zeek
vpax: I didn't want to do that for these because then the `-O gen-C++` maintenance script won't be able to run on them, since it can't handle combined Zeek runs (because it uses the Zeek run in the script to drive the C++ generation).

testing/btest/opt/opt-skip-files.zeek
vpax: Here OTOH I could, since this one specifically avoids `-O gen-C++`, because it includes `-O ZAM` on the command line.

src/Attr.cc
vpax: Roger, done with next push

src/Options.cc
vpax: Wordy to type, but I do appreciate the point. Changed in pending push

src/Options.cc
vpax: That said, ok I'll go with '1' and '2' for these

vpax:
@timwoj I believe the latest push addresses the pending comments, please see what you think

src/Options.cc
```diff
@@ -124,9 +124,15 @@ void usage(const char* prog) {
     printf(
         "    -0|--optimize-files=<pat>       | enable script optimization for all "
         "functions in files with names containing the given pattern\n");
+    printf(
+        "       --no-optimize-files=<pat>    | skip script optimization for all "
```
timwoj, line 128: Should these be `-1|--no-optimize-files` and `-2|--no-optimize-funcs`?

vpax: Hmmm. Well they could be, but it seems darn weird since no one is going to remember those mappings (unlike `-0`, which is at least somewhat mnemonic), and if they _do_ use them, Future Them might be baffled at seeing references to them. This will only get worse as we add further options, given the mnemonic space is pretty much taken up. Thoughts?

timwoj: Maybe we should just remove the short-option versions for those? If it's going to be rare that people use them, do they need to exist?

vpax: Oh totally happy to, I just don't know how to do that. There's currently a switch on `getopt_long()`, which returns the short versions. If you want to change stuff to use something else for these, please do!

PR #4829, opened 2025-09-15
Removing SetType

vpax:
Internally, Zeek doesn't track sets separately from tables (i.e., both types of values are represented by `TableVal`s and there is no analogous `SetVal`). This doesn't extend to the type system, however, which has `SetType` derived from `TableType` (peculiar, as Zeek `set`s are more restrictive than `table`s). This PR removes the type and replaces instances with `TableType`.

There are two nuances. The first is that `SetType` actually has a separate "elements" member not present in `TableType`. However, that value isn't actually provided anywhere (it's always `nullptr`), and also is only used for error-checking (which will never happen, given it's always `nullptr`). The second is that `SetType` isn't in `detail::` so this requires deprecation. I've included that, with the one caveat that for functions returning `SetType*` pointers, I've changed them to `TableType*` because otherwise Zeek's own compilation throws deprecation warnings. This change should be harmless given that in practice the two types are already the same.

### DIFF
--- a/src/Type.h
+++ b/src/Type.h
@@ -214,7 +214,10 @@ public:
     const TableType* AsTableType() const;
     TableType* AsTableType();
 
+    [[deprecated("Remove in v9.1. Use AsTableType() instead.")]]
     const SetType* AsSetType() const;
+
+    [[deprecated("Remove in v9.1. Use AsTableType() instead.")]]
     SetType* AsSetType();
 
     const RecordType* AsRecordType() const;
@@ -438,7 +441,7 @@ private:
     bool reported_error = false;
 };
 
-class SetType final : public TableType {
+class [[deprecated("Remove in v9.1. Use TableType instead.")]] SetType final : public TableType {
 public:
     SetType(TypeListPtr ind, detail::ListExprPtr arg_elements);
     ~SetType() override;
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -1410,7 +1410,7 @@ TableValPtr DNS_Mgr::empty_addr_set() {
     auto addr_t = base_type(TYPE_ADDR);
     auto set_index = make_intrusive<TypeList>(addr_t);
     set_index->Append(std::move(addr_t));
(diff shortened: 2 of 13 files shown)


src/Type.h
```diff
@@ -214,8 +214,8 @@ class Type : public Obj {
     const TableType* AsTableType() const;
     TableType* AsTableType();
 
-    const SetType* AsSetType() const;
-    SetType* AsSetType();
+    const TableType* AsSetType() const;
+    TableType* AsSetType();
```
timwoj: We do have a prior art for not changing the function definitions here, especially since they're not in the detail namespace. I know this goes back a while, but Jon did the same with `Val::Type()` in https://github.com/zeek/zeek/commit/1eb723fc9d6ba8c0bf98353ba958b96d288987f3. I swear @awelzel  did the same in another PR but I can't find it quickly. He might be able to recall more easily. The thing I'm struggling with is that changing the name of the function breaks the naming used by all of the other functions.

Is it possible we could deprecate `AsSetType()` as a whole and just remove it later, instead of changing the return type?

timwoj: As we discussed out of band, I think the right way forward here is to:
- Leave `AsSetType` alone, but deprecate it with a message saying to use `AsTableType`
- Replace our own internal uses with `AsTableType`

vpax:
Updates now pushed

src/Type.h
```diff
@@ -214,8 +214,11 @@ class Type : public Obj {
     const TableType* AsTableType() const;
     TableType* AsTableType();
 
-    const TableType* AsSetType() const;
-    TableType* AsSetType();
+    [[deprecated("AsSetType is deprecated; use AsTableType() instead")]]
```
timwoj: Mention what version it should be removed in:

> Remove in v9.1. Use AsTableType() instead.

That makes it a lot easier to find them all when I go to remove deprecations later. I can fix it when I merge it.

vpax: Ah okay I'll try to remember that in the future. (Also a reminder that I added one other in the original commit.)

PR #4831, opened 2025-09-15
Add tracking of event groups for -O gen-standalone-C++

vpax:
`-O gen-standalone-C++` previously didn't have logic to associate event groups with the generated event handlers. This PR adds logic to do so. In the process, I also modified the `ScriptFunc` interface to track `Body` information about functions rather than pointwise `Stmt` and `priority` information (which can be derived from `Body`s, as can event groups).

### DIFF
--- a/src/script_opt/CPP/Driver.cc
+++ b/src/script_opt/CPP/Driver.cc
@@ -365,13 +365,19 @@ void CPPCompile::RegisterCompiledBody(const string& f) {
     ASSERT(fi != func_index.end());
     auto type_signature = casting_index[fi->second];
 
-    auto h = body_hashes[f];
-    auto p = body_priorities[f];
-    auto loc = body_locs[f];
-    auto body_info = Fmt(p) + ", " + Fmt(h) + ", \"" + loc->FileName() + " (C++)\", " + Fmt(loc->FirstLine());
+    const auto& bi = body_info[f];
+    auto body_info =
+        Fmt(bi.priority) + ", " + Fmt(bi.hash) + ", \"" + bi.loc->FileName() + " (C++)\", " + Fmt(bi.loc->FirstLine());
 
-    Emit("\tCPP_RegisterBody(\"%s\", (void*) %s, %s, %s, std::vector<std::string>(%s)),", f, f, Fmt(type_signature),
-         body_info, events);
+    string module_group = "\"" + bi.module + "\"";
+
+    string attr_groups = "{";
+    for ( const auto& g : bi.groups )
+        attr_groups += " \"" + g + "\",";
+    attr_groups += " }";
+
+    Emit("\tCPP_RegisterBody(\"%s\", (void*) %s, %s, %s, std::vector<std::string>(%s), %s, %s),", f, f,
+         Fmt(type_signature), body_info, events, module_group, attr_groups);
 }
 
 void CPPCompile::GenEpilog() {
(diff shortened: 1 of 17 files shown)


src/script_opt/CPP/Driver.cc
```diff
@@ -365,13 +365,23 @@ void CPPCompile::RegisterCompiledBody(const string& f) {
     ASSERT(fi != func_index.end());
     auto type_signature = casting_index[fi->second];
 
-    auto h = body_hashes[f];
-    auto p = body_priorities[f];
-    auto loc = body_locs[f];
+    const auto& bi = body_info[f];
+
+    auto h = bi.hash;
```
timwoj: Nit: These three temporaries seem sorta pointless, if they're just being used for the formatting below it.

src/script_opt/CPP/Emit.h
```diff
@@ -68,6 +68,14 @@ void Emit(const std::string& fmt, const std::string& arg1, const std::string& ar
     NL();
 }
 
+void Emit(const std::string& fmt, const std::string& arg1, const std::string& arg2, const std::string& arg3,
```
timwoj, line 71: This function definition feels bad but I can't think of a better way to do it and still allow passing std::strings as arguments. It'd be possible if we ever switched to `fmtlib` or `std::format` because those are a little more flexible with POD types, but not at the moment.

src/script_opt/CPP/RuntimeInitSupport.cc
```diff
@@ -46,16 +46,18 @@ void register_type__CPP(TypePtr t, const string& name) {
...
 
 static unordered_map<p_hash_type, CompiledScript> compiled_standalone_scripts;
 
 void register_standalone_body__CPP(CPPStmtPtr body, int priority, p_hash_type hash, vector<string> events,
+                                   std::string module_group, std::vector<std::string> attr_groups,
                                    void (*finish_init)()) {
     // For standalone scripts we don't actually need finish_init, but
     // we keep it for symmetry with compiled_scripts.
-    compiled_standalone_scripts[hash] = {std::move(body), priority, std::move(events), finish_init};
+    compiled_standalone_scripts[hash] = {std::move(body),        priority,   std::move(events), std::move(module_group),
```
timwoj: Something funky happened with the formatting here.

src/script_opt/CPP/GenFunc.h
```diff
@@ -51,15 +51,16 @@ std::unordered_map<std::string, std::string> compiled_simple_funcs;
...
-// Maps function names to priorities, for hooks & event handlers.
-std::unordered_map<std::string, int> body_priorities;
-
-// Maps function names to script locations, for better-than-nothing error
-// reporting.
-std::unordered_map<std::string, const Location*> body_locs;
+struct BodyInfo {
+    p_hash_type hash;
+    int priority;
+    const Location* loc;             // for better-than-nothing error reporting
```
timwoj:
```suggestion
    p_hash_type hash = 0;
    int priority = 0;
    const Location* loc = nullptr;             // for better-than-nothing error reporting
```

src/script_opt/Inline.cc
```diff
@@ -299,7 +299,11 @@ void Inliner::CoalesceEventHandlers(ScriptFuncPtr func, const std::vector<Func::
 
     PostInline(oi, inlined_func);
 
-    funcs.emplace_back(inlined_func, new_scope, merged_body, 0);
+    // We don't need to worry about event groups because the CoalescedScriptFunc
+    // wrapper checks at run-time for whether any handlers have been disabled,
+    // and if so skips coalesced execution.
+    Func::Body body{.stmts = merged_body, .priority = 0};
```
timwoj: Changing `Func::Body` in src/Func.h to initialize priority to zero means you could remove setting it here and below.

PR #4834, opened 2025-09-15
Full tracking of the characteristics of globals for -O gen-standalone-C++

vpax:
Compiling to standalone-C++ requires creating globals with all of their attributes, such as whether they are `option`s or if they represent types. A bunch of this characteristics-tracking was missing; this PR fixes that.

### DIFF
--- a/src/script_opt/CPP/InitsInfo.h
+++ b/src/script_opt/CPP/InitsInfo.h
@@ -500,6 +500,16 @@ protected:
 };
 
 // Information for initializing a Zeek global.
+
+// Tracks all of the characteristics associated with a global.
+struct GlobalCharacteristics {
+    bool is_exported = false;
+    bool is_const = false;
+    bool is_option = false;
+    bool is_enum_const = false;
+    bool is_type = false;
+};
+
 class GlobalInitInfo : public GlobalLookupInitInfo {
 public:
     GlobalInitInfo(CPPCompile* c, IDPtr g, std::string CPP_name);
@@ -511,8 +521,7 @@ protected:
     int type;
     int attrs;
     std::string val;
-    bool is_exported;
-    bool is_option;
+    GlobalCharacteristics gc;
     bool func_with_no_val = false; // needed to handle some error situations
 };
 
--- a/src/script_opt/CPP/RuntimeInits.h
+++ b/src/script_opt/CPP/RuntimeInits.h
@@ -401,17 +401,21 @@ protected:
 
 class CPP_GlobalInit : public CPP_Init<void*> {
 public:
-    CPP_GlobalInit(IDPtr& _global, const char* _name, int _type, int _attrs, int _val, bool _is_exported,
-                   bool _is_option, bool _func_with_no_val)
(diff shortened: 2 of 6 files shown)


src/script_opt/CPP/InitsInfo.h
```diff
@@ -500,6 +500,16 @@ class GlobalLookupInitInfo : public CPP_InitInfo {
 };
 
 // Information for initializing a Zeek global.
+
+// Tracks all of the characteristics associated with a global.
+struct GlobalCharacteristics {
```
timwoj, line 505: What would you think about promoting this outside of `script_opt` and adding an `install_ID` method that just takes the structure as an argument?

src/script_opt/CPP/RuntimeInits.h
```diff
@@ -401,17 +401,21 @@ class CPP_GlobalLookupInit : public CPP_Init<void*> {
 
 class CPP_GlobalInit : public CPP_Init<void*> {
 public:
-    CPP_GlobalInit(IDPtr& _global, const char* _name, int _type, int _attrs, int _val, bool _is_exported,
-                   bool _is_option, bool _func_with_no_val)
+    CPP_GlobalInit(IDPtr& _global, const char* _name, int _type, int _attrs, int _val, bool is_exported, bool is_const,
```
timwoj, line 404: Can this take a `GlobalCharacteristics` instead of a bunch of `bools`?

src/script_opt/CPP/InitsInfo.h
vpax: Good thought (same as below). I added it because having a bunch of `bool` arguments one after another sure seemed like a recipe for mistakes. I guess I might add `func_with_no_val` too, because that's also a characteristic, and give the characteristics all default to `false` that shouldn't get in anyone's way.

vpax: Hmmm actually as I look at it, `install_ID` only cares about `is_export`, which doesn't seem worth introducing a  `GlobalCharacteristics` version for.

src/script_opt/CPP/RuntimeInits.h
vpax: Turns out that that constructor is only invoked by the table-driven `-O gen-C++` initialization, so it would be harder to have it use `GlobalCharacteristics` rather than a bunch of (integers-implicitly-cast-to-)`bool`s.

src/script_opt/CPP/InitsInfo.h
timwoj: Could a `GlobalCharacteristics` version also move [this block](https://github.com/zeek/zeek/pull/4834/files#diff-44aae4b53c21bb2be93d4d05697e9a0a91b7905bdb114603ad7430efae34cceaR144-R152) inside of `install_ID`? Is that even worth it? I guess we only do that here for script_opt, so probably not.

vpax: Doesn't seem worth it given there's no natural usage outside of script_opt ...

PR #4836, opened 2025-09-16
Force SQLite to be in thread-safe mode during CMake

### LINKED ISSUE #4285: Remove checks for SQLite thread-safe mode
We check `sqlite3_threadsafe` in the sqlite3 input reader and logging writer (and soon the storage backend). Could we remove these checks, and instead build our SQLite library always in thread-safe mode? Do we support linking against an external version of SQLite that might not have it enabled?

timwoj:
This allows us to remove the need to check for thread-safe mode in the various SQLite plugins. See https://www.sqlite.org/compile.html#threadsafe for why `1` is a good choice here.

Fixes #4285

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -467,8 +467,16 @@ set(THIRD_PARTY_SRCS
     3rdparty/strsep.c
     3rdparty/zeek_inet_ntop.c)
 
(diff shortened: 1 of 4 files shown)


ckreibich (approved):
Thanks for getting back to that!

PR #4837, opened 2025-09-17
Remove the built-in script debuggger

### LINKED ISSUE #4553: Remove built-in script debugger?
@awelzel brought this up during one of the clang-tidy PR reviews, but it's also something I've wondered about for a long time. Is the built-in script debugger actually used by anyone? It is something we could remove and replace with an external library/protocol? Arne suggested https://microsoft.github.io/debug-adapter-protocol/ in his comment.

timwoj:
This has be long out-of-use and there's no testing on it to know if it even works correctly. Removing it simplifies some of the internals in various places.

Fixes #4553


evantypanski:
I think there are a few other places this is used - most noticeably `zeek --help` still [mentions it](https://github.com/zeek/zeek/blob/938073dfed2a9e4b19989a9517fe7d0938edc496/src/Options.cc#L86). Though, is that flag technically worthy of a deprecation cycle? Removing it seems to work fine

also should this get rid of [`@DEBUG`](https://github.com/zeek/zeek/blob/938073dfed2a9e4b19989a9517fe7d0938edc496/src/scan.l#L418) here and undocument it?

There are also some cases in [`Frame.h`](https://github.com/zeek/zeek/blob/938073dfed2a9e4b19989a9517fe7d0938edc496/src/Frame.h#L147-L153) that mention a debugger, are those related? they seem very useless :) but I may just be picking up on other tech debt or weirdness now

0xxon:
I admittedly know of no one really using this - but is this something for which we should go through a deprecation cycle?

vpax:
You know, I just tried it thinking it would be busted, and I have to say it's kind of nifty ...

src/Frame.h
```diff
@@ -255,15 +246,3 @@ class Frame : public Obj {
...
- * If we stopped using this and instead just made a struct of the information
- * that the debugger actually uses we could make the Frame a class a template.
- * The template argument could be <int frame_size> and doing this would allow
- * us to use an std::array under the hood rather than a c style array.
- *
- * Another way to do this might to be to have Frame inherit from a class
- * DebugFrame which provides the information that the debugger uses. See:
- * https://stackoverflow.com/a/16211097
- */
-extern std::vector<zeek::detail::Frame*> g_frame_stack;
```
awelzel, line 269: There's some profiling plugins / tools that have used this symbol to walk the Zeek script stack, so wonder if we should keep it?

No strong opinion on keeping/deprecating. Seems the existing `zeek::detail::call_stack` should suffice.

src/Stmt.cc
```diff
@@ -1408,15 +1364,8 @@ ValPtr StmtList::Exec(Frame* f, StmtFlowType& flow) {
         auto stmt = stmt_ptr.get();
 
         f->SetNextStmt(stmt);
-
-        if ( ! pre_execute_stmt(stmt, f) ) { // ### Abort or something
```
awelzel, line 1412: I remember seeing these "hooks" at times in profiles for classic script execution.


Looking at the benchmarker, this PR also speeds up runtime for some of the microbenchmarks by 5-10% (function calls, recursive function calls). I've only skimmed and didn't do any tests locally.

image

0xxon:
I'm a bit torn on this. I kind of doubt that anyone is using this - but on the other hand we don't really know. If the debugger currently works (which it seems to) I think we should announce the removal of this visibly quite long before. E.g. by throwing deprecation warnings for a release.

vpax:
FWIW I just came up with a use case that's actually quite nifty (though see below): I'm debugging a package with its scripts compiled to C++, using `lldb` to breakpoint and examine stuff. It would be hugely handy to run a parallel session with the same breakpoints on the original scripts!

However: the debugger doesn't survive calling `suspend_processing()` followed by `continue_processing()` so I'm not actually able to do this .... sigh.

ckreibich:
This has parallels to when we removed the segment profiler, which also was completely undocumented and none of us had used it in years. I think we should at least minimally document it if we keep it. It doesn't sound to me like we really know it works ... I also wonder if the whole thing could become a plugin, given `HookCallFunction`? (Edit: well, no, that doesn't offer the same granularity as stepping through statements...)

Let's talk it over in an upcoming huddle.

timwoj:
This is clearly going to require some additional discussion. Moving this PR back to a draft.

ckreibich:
For a compromise, we could make the debugger available only in debug builds — that'd get us the performance win Arne flagged while preserving the functionality (which I'm increasingly in favor of).

timwoj:
Superseded by https://github.com/zeek/zeek/pull/4852. We can discuss separately whether fully removing it makes sense.

PR #4843, opened 2025-09-22
OpaqueVal, OCSP, X509: drop outdated LibreSSL guards to fix OpenBSD build

klemensn:
Whatever is used with recent OpenSSL is also available under with latest LibreSSL
on OpenBSD 7.8-beta as of today.

Some of these hunks have been in the net/bro port for years, others I
recently added whilst gradually updating from 6.0.5 to 8.0.1.

### DIFF
--- a/src/OpaqueVal.cc
+++ b/src/OpaqueVal.cc
@@ -27,11 +27,11 @@
 #include "zeek/probabilistic/BloomFilter.h"
 #include "zeek/probabilistic/CardinalityCounter.h"
 
-#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
+#if ( OPENSSL_VERSION_NUMBER < 0x10100000L )
 inline void* EVP_MD_CTX_md_data(const EVP_MD_CTX* ctx) { return ctx->md_data; }
 #endif
 
-#if ( OPENSSL_VERSION_NUMBER < 0x30000000L ) || defined(LIBRESSL_VERSION_NUMBER)
+#if ( OPENSSL_VERSION_NUMBER < 0x30000000L )
 #include <openssl/md5.h>
 #endif
 
@@ -225,7 +225,7 @@ constexpr size_t SHA1VAL_STATE_SIZE = sizeof(SHA_CTX);
 
 constexpr size_t SHA256VAL_STATE_SIZE = sizeof(SHA256_CTX);
 
-#if ( OPENSSL_VERSION_NUMBER < 0x30000000L ) || defined(LIBRESSL_VERSION_NUMBER)
+#if ( OPENSSL_VERSION_NUMBER < 0x30000000L )
 
 // -- MD5
 
--- a/src/file_analysis/analyzer/x509/OCSP.cc
+++ b/src/file_analysis/analyzer/x509/OCSP.cc
@@ -26,7 +26,7 @@ namespace zeek::file_analysis::detail {
 static constexpr size_t OCSP_STRING_BUF_SIZE = 2048;
 
 static bool OCSP_RESPID_bio(OCSP_BASICRESP* basic_resp, BIO* bio) {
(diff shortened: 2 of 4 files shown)


timwoj:
Should these check for a LibreSSL version that implements them instead?

klemensn:
> Should these check for a LibreSSL version that implements them instead?

Perhaps for those testing for OpenSSL 3.x, but most guards are for long EOL 1.1.0.
I'd argue to rather drop these <1.1.0 completely;  Zeek's dependency list already requires >=3, anyway, no? 

https://www.libressl.org/releases.html says
> LibreSSL stable branches are updated for 1 year after their corresponding OpenBSD branch is tagged for release.

Looking at the first OpaqueVal.cc hunk for example, it uses [EVP_MD_CTX_md_data(3)](https://man.openbsd.org/EVP_MD_CTX_md_data.3#HISTORY) which says

```
HISTORY
     ...
     EVP_MD_CTX_pkey_ctx() and EVP_MD_CTX_md_data() first appeared in OpenSSL
     1.1.0 and EVP_MD_CTX_set_pkey_ctx() in OpenSSL 1.1.1.  These functions
     have been available since OpenBSD 7.1.
```

So at latest even with OpenBSD 7.4 released in Oct 2023 the corresponding LibreSSL went EOL.

klemensn:
https://github.com/zeek/actor-framework/pull/3 is another local patch OpenBSD carries for Zeek.

botovq:
Most of these checks break with LibreSSL 3.5.2 as they are, since in that version many types were made opaque to match the OpenSSL 1.1 API. However, zeek has been using the `TLS1-PRF` API in libcrypto for a long time (since 5.0.0?), which we made available only in LibreSSL 4.0.0, so it already requires a more recent version than that.

timwoj:
It looks like this missed a few other instances:

```
src% ack LIBRESSL_VERSION_NUMBER
file_analysis/analyzer/x509/X509.h
29:#if ! defined(LIBRESSL_VERSION_NUMBER) || (LIBRESSL_VERSION_NUMBER < 0x2070000fL)

file_analysis/analyzer/x509/functions.bif
1061:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)

file_analysis/analyzer/x509/X509.cc
352:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)

digest.cc
15:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
```

Are those left behind intentionally?

klemensn:
> It looks like this missed a few other instances:
> 
> ```
> src% ack LIBRESSL_VERSION_NUMBER
> file_analysis/analyzer/x509/X509.h
> 29:#if ! defined(LIBRESSL_VERSION_NUMBER) || (LIBRESSL_VERSION_NUMBER < 0x2070000fL)
> 
> file_analysis/analyzer/x509/functions.bif
> 1061:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
> 
> file_analysis/analyzer/x509/X509.cc
> 352:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
> 
> digest.cc
> 15:#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
> ```
> 
> Are those left behind intentionally?

No, surely an oversight from iterating over the build while patching out failures.

I built 8.0.1 with all those removed as well now, no change in test results.

Shall I open a new PR or do you want to take care of them when cleaning up OpenSSL defines?

timwoj:
> Shall I open a new PR or do you want to take care of them when cleaning up OpenSSL defines?

I'll take care of them in the follow-up. Thanks for checking.

PR #4844, opened 2025-09-23
utils/decompose_uri: Support URIs containing IPv6 addresses

### LINKED ISSUE #4842: decompose_uri() - bad conversion to count with ipv6
The decompose_uri() function doesn't seem to properly handle ipv6 bracketed addresses, like [::1]:80/test/a/b.exe?a=b. This results in the error message:

```
 error in /opt/zeek-dev-prod/share/zeek/base/utils/urls.zeek, line 127: bad conversion to count (to_count(parts[1], 10) and :1]:80)
[scheme=http, netlocation=[, portnum=0, path=/test/a/b.exe, file_name=b.exe, file_base=b, file_ext=exe, params={
[a] = b
}]
```

awelzel:
An URI containing a bracketed or non-bracketed IPv6 address of the form http://[::1]:42 was previously split on the first colon for port extraction, causing a subsequent to_count() call to fail. Harden this to check for a digits in the last :[0-9]+ component.

Fixes #4842


testing/btest/scripts/base/utils/decompose_uri.zeek
```diff
@@ -29,5 +30,15 @@ event zeek_init()
...
+
+	# Bracketed IPv6
+	dc("http://[::1]:8080/?foo=bar&baz=qux");
+	dc("http://[::1]/foo/bar");
+	dc("http://[::1]/foo/bar");
+	dc("[::1]:80/test/a/b.exe?a=b");
+
+	# Un-bracketed as ambiguous, but not causing errors.
+	dc("http://beeb:deed::1/test");
+	dc("http://beeb:deed::1:8080/test");
```
bbannier, line 42: There are other places where `:` may show up, in particular the path or query parameters. This seems to work, could you add baselines for that as well?

```suggestion
	dc("http://beeb:deed::1:8080/test");

	dc("https://en.wikipedia.org/wiki/Template:Welcome");
	dc("https://[::1]:8080/wiki/Template:Welcome");
	dc("https://[::1]:8080/wiki/Template:Welcome?key=:&value=:");
```

awelzel: Yeah, good call, thanks!

awelzel:
@bbannier - mind re-checking the last fixup? There is an existing test elsewhere for `"git://git.kernel.org:/pub/scm/linux/`  which assumes the `:` is stripped. I changed the regex from `:[0-9]+` to `:[0-9]*` for the digits to keep that behavior.

PR #4849, opened 2025-09-25
cluster/zeromq: Improve EINTR handling

awelzel:
When using ZeroMQ also within the Supervisor process, zmq::poll() and recv() were observed to return EINTR, handle these.


src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -650,8 +650,17 @@ void ZeroMQBackend::Run() {
...
+            try {
+                int r = zmq::poll(poll_items, std::chrono::seconds(-1));
+                ZEROMQ_DEBUG_THREAD_PRINTF(DebugFlag::POLL, "poll: r=%d", r);
+            } catch ( const zmq::error_t& err ) {
+                ZEROMQ_DEBUG_THREAD_PRINTF(DebugFlag::POLL, "poll exception: what=%s num=%d", err.what(), err.num());
+                // Retry interrupted zmq::poll() calls.
+                if ( err.num() == EINTR )
+                    continue;
+
+                throw err;
```
bbannier: Nit: Wonder if we couldn't just rethrow here with

```suggestion
                throw;
```

Same below.

bbannier (approved):

Issue #4762, opened 2025-08-18
Default to ZeroMQ cluster backend with 8.1


awelzel:
I'm not 100% sure what this involves, it might make sense to extract some of the Broker specific peering/connection logic in `setup-connections` into broker specific policy scripts (i.e. move `cluster/setup-connections.zeek` elsewhere).

awelzel:
Closing this. Zeekctl and tools/systemd-generator default to the ZeroMQ cluster backend.

Issue #4432, opened 2025-05-06
cluster/zeromq: Support encryption for multi-node deployments


awelzel:
The current ZeroMQ cluster backend uses plaintext communication.

For multi-node clusters, it'd be good to have a way to encrypt communications among nodes.

There's "zmq_curve - secure authentication and confidentiality": http://api.zeromq.org/4-1:zmq-curve

Summary:
* Check and experiment with zmq_curve
* New options for server and secret keys in the ZeroMQ module and apply to the sockets in use
* Document deployment setups

MP-Corelight:
In conversation with the team, the consensus was that we'd target this at Zeek 8.1 — recognizing that users who want to switch to the new pluggable back-ends with encrypted Zeek will have to wait for that release.

awelzel:
Closing after merge of #5227 and zeek/zeekctl#106.

PR #4851, opened 2025-09-25
ZeroMQ by default, part 1

awelzel:
This is for #4762 which is titled "Default to ZeroMQ cluster backend with 8.1". That's not exactly what's happening here though.

This PR introduces a `None` cluster backend that's becoming the default. The idea being that `policy/frameworks/cluster/backend/broker` or `policy/frameworks/cluster/backend/zeromq` is then loaded explicitly, rather than implicitly.

@ckreibich - there isn't all that much happening here so far, outside of keeping all tests functional. Before I polish this up, does the "None" approach seem somewhat reasonable to you? I believe it avoids having ZeroMQ pieces pop up within the Cluster framework itself, like it definitely happened with Broker (stores, connection logic, broker specific telemetry, ...). The drawback is a requirement to explicitly load the needed policy scripts for a given cluster backend in an actual cluster setup, but it seems that would be done by Zeekctl / Management framework / script orchestration anyhow.


timwoj:
I'm really not a fan of the idea of a `None` backend be the default. I imagine that most people installing Zeek are doing so in a cluster environment. Having to specify a cluster backend as part of setup vs everything just working out of the box feels really off to me. I'd prefer if we just pick one of them (probably ZeroMQ at this point). I don't mind the `None` backend existing for people that really don't need one, but having it as the default is probably not ideal.

testing/btest/cluster/zeromq/subscribe-before-init.zeek
```diff
@@ -0,0 +1,55 @@
+# @TEST-DOC: Regression test Cluster::subscribe() blocking if called in a high-priority zeek_init() handler
```
ckreibich, line 1: This is very terse — blocking isn't bad per se. Is the problem you're addressing in this commit that it would lock up forever?

src/cluster/backend/none/Plugin.cc
```diff
@@ -0,0 +1,57 @@
...
+
+namespace zeek::plugin::Zeek_Cluster_None {
+
+class Plugin : public zeek::plugin::Plugin {
+    zeek::plugin::Configuration Configure() override {
+        AddComponent(new cluster::BackendComponent("None", NoneBackend::Instantiate));
+
+        zeek::plugin::Configuration config;
+        config.name = "Zeek::Cluster_Backend_None";
+        config.description = "Cluster backend none";
```
ckreibich:
```suggestion
        config.description = "A no-op cluster backend";
```

scripts/policy/frameworks/cluster/backend/broker/main.zeek
```diff
@@ -0,0 +1,152 @@
+##! Broker cluster backend support.
+##!
+##! The Broker cluster backend is a peer-to-peer backend that has been
+##! in use since Bro 2.6 until Zeek 8.1. Individual Zeek cluster nodes
+##! peer with each other using a fixed connection strategy using the information
+##! stored in :zeek:see:`Cluster::nodes` populated by the cluster-layout.zeek
+##! file or via supervisor internal functionality.
```
ckreibich:
```suggestion
##! The Broker cluster backend is a peer-to-peer backend that has been
##! in use since Bro 2.6 and the default until Zeek 8.1. Cluster nodes peer
##! with each other selectively, using a fixed connection strategy based on
##! cluster node types. This information is stored in :zeek:see:`Cluster::nodes`
##! as populated by the cluster-layout.zeek file, or internally via the Supervisor
##! when in use.
```

awelzel:
> I'm really not a fan of the idea of a None backend be the default. I imagine that most people installing Zeek are doing so in a cluster environment. Having to specify a cluster backend as part of setup vs everything just working out of the box feels really off to me.

I guess I'm thinking of Zeek vs Zeek cluster a bit differently.

As mentioned by Christian, when running `zeek -r pcap <some analysis scripts>` in a terminal, the cluster backend should be rather irrelevant as long as one has access to the `Cluster::publish()`, `Cluster::subscribe()`, etc. API. Whether these functions actually do something in non-cluster mode is of little importance.

In contrast, if you want to run a Zeek cluster of just two process by hand locally:

* Decide on the names you want to give the two processes/nodes and their roles.
* Create a `cluster-layout.zeek` file, mapping names for cluster nodes to ports and listen addresses and the role of those two processes. This already  involves advanced and intricate knowledge and is dependent on the cluster backend.
* Start two terminals, create separate working directories for the Zeek processes so they don't trample over each other, ensure cluster-layout.zeek is in the PATH of both processes.
* Start the processes in each terminal with the `CLUSTER_NODE` environment variable (or `Cluster::node=...`) set.
* If it's a worker, ensure to add `-i <interface>` to the command-line.
* If you only configured a manger and logger, realize that somehow you need to `redef Cluster::manager_is_logger=F`, otherwise the manager also receives log writes, but you might have expected only the logger to produce them.
* With Broker, have knowledge which nodes peer (e.g. running two worker processes won't actually do anything). For ZeroMQ, ensure your start the XPUB/XSUB proxy somewhere (automatically in manager, or externally).
* Realize you want to have Prometheus, so now the cluster-layout.zeek needs `$metrics_port`, too.

All that to say that starting two Zeek processes that form a cluster involves so many steps and intricate knowledge already (maybe the list above should actually be in the docs). It's like starting an airplane if you want. If the list of steps additionally involves selecting the cluster backend via putting a policy script on the command line, from my perspective, that explicitness is actually an improvement for users to understand what's going on if they really want. Most others will use zeekctcl, zeek-systemd-generator (:crossed_fingers:), management framework, or whatever they have built by themselves.

Maybe the name could be `Null`, `Stub` or `Noop` instead of `None` - would that help?

awelzel:
@bbannier pointed out that it'd be nice to run WebSocket even with a single node process, e.g. with `zeek -i lo` and that's very fair. We could build a WebSocketShim for the `None` backend or re-use the `ZeroMQ` implementation.

testing/btest/cluster/zeromq/subscribe-before-init.zeek
awelzel: It would've locked up `Cluster::subscribe()` within a zeek_init() handler which essentially locks up everything. This commit already landed in master via https://github.com/zeek/zeek/pull/4850 as it was meant for 8.0.x.

awelzel:
@ckreibich - I rebased and squashed. The main changes since a month ago is the addition off `cluster/websocket/server.zeek` that can be loaded by a standalone Zeek process to run a WebSocket server. This should address the "easy testing" of WebSocket functionality (it runs ZeroMQ internally. If you preload Broker, it would use Broker hubs).

The other part were all those changes in Zeekctl. This seems to work out for a single node setup. I'm planning on tackling #4432 next and that should involve some multi-node testing. I'd not be surprised if something doesn't quite work for multi-node after merging this PR. Maybe the people running `master` + `Zeekctl` will quickly tell me. It's always possible to switch back to Broker via `ClusterBackend = Broker` in `zeekctl.cfg`.

testing/external/scripts/testing-setup.zeek
```diff
@@ -28,3 +28,9 @@ redef Intel::manage_seen_event_groups = F;
 redef Analyzer::DebugLogging::ignore_analyzers += { Analyzer::ANALYZER_IMAP };
 redef Analyzer::DebugLogging::include_confirmations = F;
 redef Analyzer::DebugLogging::include_disabling = F;
+
+# The external tests are loading test-all-policy which in turn will load
+# policy scripts for a cluster backends. Redefine back to Cluster::CLUSTER_BACKEND_NONE
+# in order to skip initialize of cluster backend. Cluster backends may keep
```
ckreibich:
```suggestion
# policy scripts for a cluster backend. Redefine back to Cluster::CLUSTER_BACKEND_NONE
# in order to skip initialization of cluster backend. Cluster backends may keep
```

scripts/test-all-policy.zeek
```diff
@@ -20,7 +20,7 @@
 @load frameworks/cluster/backend/broker/telemetry.zeek
 @ifdef ( Cluster::CLUSTER_BACKEND_ZEROMQ )
 @load frameworks/cluster/backend/zeromq/__load__.zeek
-# @load frameworks/cluster/backend/zeromq/connect.zeek
+@load frameworks/cluster/backend/zeromq/connect.zeek
```
ckreibich, line 23: I'm probably missing something — if it's deprecated, why do we still load it in our own use? To test that loading it doesn't cause any trouble?

scripts/policy/frameworks/cluster/websocket/server.zeek
```diff
@@ -0,0 +1,54 @@
+##! Script to load for running a single WebSocket server.
+##!
+##! This script is mostly meant for ad-hoc testing. In a Zeekctl environment,
+##! the ``UseWebSocket`` option should be used instead.
+##!
+##! Note that if :zeek:see:`Cluster::backend` is ``CLUSTER_BACKEND_NONE`` at the
+##! time this script is loaded, it loads the ZeroMQ cluster backend and starts a
+##! locally running XPUB/XSUB proxy thread. If you instead want to use Broker's hub
+##! functionality instead, ensure to load policy/frameworks/cluster/backend/broker
```
ckreibich:
```suggestion
##! functionality instead, load policy/frameworks/cluster/backend/broker
```

scripts/policy/frameworks/cluster/websocket/server.zeek
```diff
@@ -0,0 +1,54 @@
...
+##! before loading this script.
+##!
+##! Note also that this script will raise a fatal error if the cluster backend
+##! is :zeek:see:`Cluster::CLUSTER_BACKEND_NONE`, but :zeek:see:`Cluster::nodes`
+##! is populated with entries.
```
ckreibich, line 14: Are we sure this is the right place to make that call? The reasoning makes sense to me, but I wonder why we would error on it when the user wants to enable the webserver?

scripts/policy/frameworks/cluster/websocket/server.zeek
```diff
@@ -0,0 +1,54 @@
...
+@endif
+@endif
+
+event zeek_init() &priority=-100
+	{
+	local listen_addr = 127.0.0.1;
+	local listen_addr_env = getenv("ZEEK_WEBSOCKET_LISTEN_ADDRESS");
+	if ( |listen_addr_env| > 0 )
+		{
+		listen_addr_env = rstrip(lstrip(listen_addr_env, "["), "]");
```
ckreibich, line 40: Is this simply for robustness or are there other reasons why the address would be in `[ ... ]`? I ask because I wonder whether we should just handle that case in `to_addr`.

testing/btest/cluster/websocket/server/ipc.zeek
```diff
@@ -0,0 +1,67 @@
+# @TEST-DOC: Running a standalone Zeek using the ipc:// transport with ZeroMQ.
```
ckreibich:
```suggestion
# @TEST-DOC: Running a standalone Zeek process with a WebSocket server and ZeroMQ, using the ipc:// transport for the latter.
```

scripts/policy/frameworks/cluster/websocket/server.zeek
```diff
@@ -0,0 +1,54 @@
+##! Script to load for running a single WebSocket server.
+##!
+##! This script is mostly meant for ad-hoc testing. In a Zeekctl environment,
+##! the ``UseWebSocket`` option should be used instead.
+##!
+##! Note that if :zeek:see:`Cluster::backend` is ``CLUSTER_BACKEND_NONE`` at the
+##! time this script is loaded, it loads the ZeroMQ cluster backend and starts a
+##! locally running XPUB/XSUB proxy thread. If you instead want to use Broker's hub
```
ckreibich, line 8: By default this will demand two ports, right? Would it make sense to use the IPC mechanism for the proxy thread by default in that case, just to reduce the chance of port collisions?

NEWS
```diff
@@ -12,6 +12,29 @@ We would like to thank @chrisjlly, Klemens Nanni (@klemensn), and Klemens Nanni
...
 ----------------
 
+- A standalone ``zeek`` invocation on the command line will now default to running
+  a No-op Cluster backend (``Cluster::CLUSTER_BACKEND_NONE``). Previously this
+  defaulted to Broker without having any real effect. For a multi-process / cluster
+  deployment of Zeek, a cluster backend policy script needs to be loaded explicitly.
+
+  Zeekctl has been updated to use the ZeroMQ cluster backend by default and should
+  require no action. If you encounter issues with ZeroMQ, it is possible to re-enable
+  Broker cluster backend by adding the following entry into ``zeekctl.cfg`` and
```
```suggestion
  the Broker cluster backend by adding the following entry into ``zeekctl.cfg`` and
```

NEWS
```diff
@@ -12,6 +12,29 @@ We would like to thank @chrisjlly, Klemens Nanni (@klemensn), and Klemens Nanni
...
+  running `zeekctl deploy``:
+
+	ClusterBackend = Broker
+
+  If you manage a non-Zeekctl Zeek setup, ensure to load ``policy/frameworks/cluster/backend/zeromq``
```
```suggestion
  If you manage a non-Zeekctl Zeek setup, make sure you load ``policy/frameworks/cluster/backend/zeromq``
```
(Just because "ensure to load" feels a bit clunky — hence also my tweak in the other place. Or just say "load".)

NEWS
```diff
@@ -12,6 +12,29 @@ We would like to thank @chrisjlly, Klemens Nanni (@klemensn), and Klemens Nanni
...
+  or ``policy/frameworks/cluster/backend/broker``. If you forget doing so, Zeek
```
```suggestion
  or ``policy/frameworks/cluster/backend/broker``. If you forget to do so, Zeek
```

scripts/test-all-policy.zeek
awelzel: Hmm, somewhat. I'm going to leave it, but no strong opinion. I wanted to say to test it, but bare-mode-errors would catch anything completely off.

NEWS
awelzel: Thanks! I went with "load".

scripts/policy/frameworks/cluster/websocket/server.zeek
awelzel: The IPC mechanism will create files in the current directory, so wasn't a big fan of that.

scripts/policy/frameworks/cluster/websocket/server.zeek
awelzel: My view was that if someone explicitly populates `Cluster::nodes`, they should also have made the decision to set `Cluster::backend`. We have a slightly inverted check in `cluster/main.zeek` now: We error if `Cluster::node` is set, but `Cluster::backend` is set to NONE. So I'll leave it.

scripts/policy/frameworks/cluster/websocket/server.zeek
awelzel: Primarily robustness, because sometimes `[` and `]` is needed and sometimes it's not and here it shouldn't make a difference.

I have no reaction if `to_addr()` should magically strip, but will keep it here.

ckreibich: (And I'd do that in a separate PR anyway if we make that change, so all good  — thanks.)

Issue #724, opened 2020-01-03
3.0.2 meta issue


jsiwek:
Just collecting notable things to potentially address in a 3.0.2 release:

- [ ] backport the Cirrus CI config (for better coverage of macOS/FreeBSD): cd9fec7bdbf3b039627ac636ca01c0d5e82a1061, 52f97c7e457017d05790e01ff134ef6e00c2f027, 84e3bc7aca0234b8d5d6db7b55e246334a2da005
- [ ] use-after-free in paraglob: zeek/paraglob@d65dd0a
- [ ] malloc/delete mismatch in JSON formatting: c0d6eb9efbc133c320a963729e99251d4b6952e3
- [ ] leak in OCSP parsing when using OpenSSL 1.1: 2fbcf23f767797d321d7239f5260d16a25b16eb1
- [ ] leak in Kerberos ticket decryption: 53fadb2bb0e9a165dac3899290282236f770be68
- [ ] leak when table-based input stream overwrites old entry: 3742e5601c7390aacd1f8ce59b054e0090d7524a
- [ ] leak in packet filter functions: a961f0b4c4ad5a365d2078acbacaf6a380997013
- [ ] leak in `system_env()` BIF: 273eb19ff5780d8d823631f906fb2fcf2e81b312
- [ ] leak of `Log::Filter` "config" field: bf05add5423cdeedddad21c07b08a3a65ce98431
- [ ] leak in `Reporter::get_weird_sampling_whitelist()` BIF: 3b6a2a5f4ebb04dda32f7f35fad4ed603d226a6d
- [ ] leaks in input framework error-handling cases: 6f5f7df9705e7eff95b88203b2b4ee95d52d55b8
- [ ] leak when a plugin prevents a log write operation: 09578c6176e5be2f794ebbfc323ef517e45131e9
- [ ] leaks due to reference-counting issues in lambdas/closures: 929e85dd146855f67397112a7819680ecf5f6841

timwoj:
Should we consider including 46e730842 (the rapidjson changes) for this release?

jsiwek:
Past releases:

# Zeek 8.2.2

- [x] https://github.com/zeek/zeek/pull/5644
- [x] https://github.com/zeek/zeek/pull/5645
- [x] https://github.com/zeek/zeek/commit/f15c70e4747f81dd79617407405a273655e1d42b (Use thread-safe Fmt() in backend threads)
- [x] https://github.com/zeek/zeek/pull/5675
[...]

PR #4852, opened 2025-09-26
Enable the script debugger on debug builds only

timwoj:
After discussion in https://github.com/zeek/zeek/pull/4837, I decided to shelve the idea of removing the debugger entirely for right now. @awelzel pointed out that we could remove it for release builds though, and that would give us some fairly significant performance savings. This PR does that by only enabling it if `DEBUG` is set. It also causes the debugger and script tracing arguments to emit fatal errors if you try to use them on non-debug builds.


timwoj:
Benchmarks on a release build confirm savings:

Screenshot 2025-09-26 at 9 17 53 AM

awelzel:
Dropped the approval. Hmm, `g_frame_stack` is definitely used with some plugins at Corelight for profiling. Maybe we could keep updating it even in non-DEBUG mode? I saw you marked for backporting to 7.0.x or 8.0.x - based on #724, I think for those versions it should continue to reflect the script execution's stack. I don't mind for 8.1 so much.

awelzel:
On second thought and after a quick chat with Tim, I'd be okay to ditch `g_frame_stack` in debug mode. It's weird that `g_frame_stack` isn't. There's also `zeek::detail::call_stack`, which could maybe also start tracking the frame in addition, rather then keeping two stacks.

timwoj:
@bbannier asked whether this could be done via a runtime flag instead. It would allow people to enable it on release builds too, without needing to run a completely separate binary. There'd be a minor amount of overhead for the boolean checks, but not the same as having the whole feature enabled.

timwoj:
Actually I wonder if it would be enough to base it on the flag passed to Zeek at startup. Meaning, don't do any of this tracking unless someone explicitly wanted the debugger. The performance impact of pushing/popping to `g_frame_stack` is pretty low, so that commit can be dropped. The rest could be wrapped in checks against a new global that stores the state of the runtime argument.

@awelzel What do you think about that instead?

awelzel:
> The rest could be wrapped in checks against a new global that stores the state of the runtime argument.
>
> @awelzel What do you think about that instead?

No super strong opinion. Given we keep all the code, sounds reasonable to keep it functional also in release builds.

Looking at the `IfStmt` and `StmtList` classes, I think I'd introdcue separate subclasses that implement `Exec()` / `DoExec()` to have the debugging hooks and clean out the original ones. Then, depending on the debug flag construct `DbgStmtList`  or `StmtList`. That would unclutter `Exec()` code of the non-debug version at the expense of some if / else at AST construction time.

awelzel (commented):
TBH, not sure this is actually worth it. This is slight opposite of what I had hoped for with #4837  :-)

While it might have a slight perf improvement, now there's a new scripting layer option that at least some users (Corelight) need to set sometimes to `T` and a bunch more if conditions in various places, which, if they are wrong, results in stack underflow or memory leaks.

I'm forth and back between "don't care enough and approve" and "could we just close #4837 as not planned and move on" and am leaning to the latter.

I could see it being nice getting rid of `g_frame_stack` with a deprecation in 9.1 and improve the `call_stack` data structure such that the debugger can use it, too. For the `pre/post_execute_stmt()` calls, still leaning on specialized IfStmt and StmtList classes as being nicer.

timwoj:
Superseded by #4874

PR #4853, opened 2025-09-26
iosource/Packet: Allow ToRawPktHdrVal() for reassembled packets

### LINKED ISSUE #4845: get_current_packet_header bif does not contain ip headers for fragmented packets
Currently, the `get_current_packet_header` bif does not seem to return any ip header information, at least in some cases of fragmented packets.

Example pcap attached.

Script to run:

```
event connection_SYN_packet(c: connection, pkt: SYN_packet)
	{
	local header = get_current_packet_header();
	print header;
	}
```

Output:

```
[l2=[encap=LINK_ETHERNET, len=74, cap_len=50, src=98:ee:cb:bd:07:fe, dst=1e:0b:8b:14:f6:f9, vlan=<uninitialized>, inner_vlan=<uninitialized>, eth_type=2048, proto=L3_IPV4], ip=<uninitialized>, ip6=<uninitialized>, tcp=<uninitialized>, udp=<uninitialized>, icmp=<uninitialized>]
```

Expected output: containing at least the ip header of the packet. It would make sense for this to not contain a tcp header, as that header might not be in the fragment that's currently being analyzed.

[fragmented_syn.pcap.zip](https://github.com/user-attachments/files/22496735/fragmented_syn.pcap.zip)

awelzel:
Closes #4845


awelzel:
From the ticket:

> Expected output: containing at least the ip header of the packet. It would make sense for this to not contain a tcp header, as that header might not be in the fragment that's currently being analyzed.

Hmm, hmm. The current packet's IP header is [replaced with the reassembled version once all fragments arrived](https://github.com/zeek/zeek/blob/83f05dde34810ade0421c77ef860ab48f0258d56/src/packet_analysis/protocol/ip/IP.cc#L168-L170) and that's what is exposed via `get_current_packet_header()`. That's visible in the baselines as `MF=T` and `offset=0` which means it's using the IP header from the first fragment rather than the second one with `MF=0`.

Not sure there's a right or wrong here. Marking as draft for feedback.

0xxon:
I am happy for this to contain the information fo the reassembled packet once all fragments arrive. That's a perfectly fine stance to take, and much better than what happens at the moment.

I also don't think there is a right or wrong - but this sounds perfect as long as we add the fact that this is the behaviour to dhe documentation of the BiF :)

awelzel:
@0xxon - mind checking the docs addition and suggest rewordings?

src/zeek.bif
```diff
@@ -3951,6 +3951,16 @@ function get_current_packet%(%) : pcap_packet
 ## .. note::
 ##
 ##      See :zeek:see:`get_current_packet` for caveats.
+##
+## .. note::
+##
+##      Zeek currently does not expose individual IP datagram fragments to the
+##      script-layer. Therefore, this function either returns the original header
+##      of a non-fragmented IP datagram, or the header of a reassembled datagram.
```
0xxon: Does..

or, if the last fragment has been received, the header of a reassembled datagram.

make sense?

awelzel: Yeah, fits better, thanks!

Issue #4545, opened 2025-06-05
Examine other uses of mktime


0xxon:
After #4521, I checked for other uses of `mktime` in our source tree.

There are two. smb uses this when parsing smb negotiations (https://github.com/zeek/zeek/blob/e6755325e1d835c96c392a466707d62765489f5c/src/analyzer/protocol/smb/smb-time.pac#L27).

Furthermore, the `strptime` bif uses `mktime` - https://github.com/zeek/zeek/blob/e6755325e1d835c96c392a466707d62765489f5c/src/zeek.bif#L3652

I have a suspition that both of these occurrences should be switched to timegm - but this would change longstanding behaviour of the bif. An alternative for the bif might be to document behavior.

markoverholser:
@ekoyle and I spent some time looking at these today, and have the following thoughts.

https://github.com/zeek/zeek/blob/e6755325e1d835c96c392a466707d62765489f5c/src/analyzer/protocol/smb/smb-time.pac#L27

This code seems to have a couple potential issues:
* The `tm` struct is not initialized, and therefore things like the timezone offset aren't guaranteed to be consistent from run to run. In our experimentation, time zone offsets in an uninitialized `tm` were random (on my Mac)
* We are assuming that `tz` is also a time zone offset obtained from the NTLM message, and if so it should probably be _subtracted_ from the return of `mktime` if the goal is to get back to UTC from local time in the NTLM message, but honestly the better approach would be to just set the `tm_gmtoff` in the `tm` struct directly?
* Should probably not move to `timegm()` because it will ignore the time zone offset that is asserted in the NTLM message

https://github.com/zeek/zeek/blob/e6755325e1d835c96c392a466707d62765489f5c/src/zeek.bif#L3652

* The usage of `! localtime_r(&timeval, &t)` in the conditional statement _seems_ to be only there to initialize the `tm` structure (to avoid the indeterminate behavior issues referenced above), but it makes the code hard to read. Instead, should maybe just explicitly initialize the `tm` struct to zero with something like `struct tm t = {}`?
* This does not look like we want to switch from `mktime()` to `timegm()` because the user might pass in a format string and value string which contain TZ information and `timegm()` would discard that information when converting back to an epoch time. Instead, we should leave this as using `mktime()` and ensure that the timezone offset is set appropriately in the `tm` struct

timwoj:
[...]

PR #4856, opened 2025-09-26
Fix parsing of time in SMB_COM_NEGOTIATE messages

timwoj:
This fixes the parsing of the time in the LanManager version of SMB_COM_NEGOTIATE messages. As noted in #4545, there's some confusion as to the length of these timestamps. I opted to consider them as two 16-bit fields (against what the docs say), and parse them as a simple bit-packed format. It also adds the timezone to the `tm` structure that we weren't doing before, since it was already available in the data. Lastly, now that we're setting `tm_gmtoff`, we switch from using `mktime()` to `timegm()` to take that into account.


src/analyzer/protocol/smb/smb1-com-negotiate.pac
```diff
@@ -181,8 +181,8 @@ type SMB1_negotiate_lanman_response(header: SMB_Header) = record {
 	max_number_vcs        : uint16;
 	raw_mode              : uint16; # expanded in &let
 	session_key           : uint32;
-	server_time           : SMB_time;
-	server_date           : SMB_date;
+	server_time           : uint16;
```
awelzel, line 184: Whoa. We parsed 3x 16bit before and now just a single one. So that never really worked.

I smoke checked Wireshark and it produces the same date 💪

src/analyzer/protocol/smb/smb1-com-query-information.pac
```diff
@@ -31,7 +31,7 @@ type SMB1_query_information_request(header: SMB_Header) = record {
 type SMB1_query_information_response(header: SMB_Header) = record {
 	word_count      : uint8;
 	file_attribs    : uint16;
-	last_write_time : SMB_time;
+	last_write_time : uint32;
```
awelzel, line 34: I've found the following refs and it seems uint32 for the UTIME format is good.

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/cfa18c97-61c7-422e-ab64-cd3849864dfd

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cifs/94ef49b8-648c-47f1-9c64-def4ad0a63a0

The docs of UTIME mention little endian explicitly. I'm assuming that fields are parsed littleendian due to the outer `SMB_Message_Response` record having `&byteorder`.

awelzel (approved):

Issue #2686, opened 2023-01-13
Zeek doesn't complain about completely redefining a record


vpax:
Surprisingly, this script compiles fine:
```
type r: record {
        a: count &default=3;
        b: string &default="missing";
};

global x1: r;

type r: record {
        d: count &default=4;
        e: string &default="present";
        f: double &default=3.14;
};

global x2: r;

print x1, x2;
```
and when run prints:
```
[a=3, b=missing], [d=4, e=present, f=3.14]
```

timwoj:
Just at a quick glance, it looks like we only avoid creating duplicate IDs if we're not in the global scope. It'll happily do this otherwise.

awelzel:
It's the same for enums, actually and we might be relying on this working due to enum's being repeated for `.bif` files and in `.zeek` files:

https://github.com/zeek/zeek/pull/4063#discussion_r1858864804

vpax:
Actually, it's worse, the redeclared identifier can be a completely different type. This just bit me in some of our code, so I'm planning to fix it.

timwoj:
@vpax Is this fixed in the above merge?

vpax:
Just confirmed that it now correctly generates an error.

PR #4857, opened 2025-09-28
Fix re-declarations of type identifiers

vpax:
As discussed in GH-2686, Zeek has allowed introducing a new declaration for a type identifier that is (completely) different than its original declaration. This PR flags such instances, while allowing for (1) re-declaring empty records (needed due to some stub records in `init-bare.zeek`) and (2) re-declaring an `enum` type equivalent to its original (needed because some BiFs declare `enum` types separately from scripts that use those.


timwoj (approved):
Is this something that should be called out in NEWS?

vpax:
> Is this something that should be called out in NEWS?

To my mind, no. It's not a change in any previously known behavior, it's just a bug fix for Zeek's type-checking and the like.

PR #4858, opened 2025-09-29
RFC: Reduce RecordVal size

awelzel:
This PR reduces the size of RecordVal instances for 96bytes to 64bytes. This started in #4702 and had it stashed for a bit.

Summary:

* Remove vtable entry of `detail::Modifiable` - switch to calling Unregister() explicitly
* Remove `rt` and `is_managed` members from `RecordVal`, put information into the `record_val` vector itself, using a new ZValSlot class instead of the std::optional.
* Replace `std::vector` (24 bytes) for record_val with a custom vector using 32 bit size and capacity (16 bytes). I *think* we could switch to a raw pointer if we'd be okay getting rid of `AppendField()`

I'm not particularly attached to any of these changes, but the fact that they allow to reduce the RecordVal size makes me think they'd be useful to discuss. Any C++ or low-level gurus interested?

Opening as draft to see if the benchmarks work out.

### DIFF
--- a/src/Val.h
+++ b/src/Val.h
@@ -1484,7 +1484,7 @@ protected:
 
     Obj* origin = nullptr;
 
-    using RecordTypeValMap = std::unordered_map<RecordType*, std::vector<RecordValPtr>>;
+    using RecordTypeValMap = std::unordered_map<const RecordType*, std::vector<RecordValPtr>>;
     static RecordTypeValMap parse_time_records;
 
 private:
--- a/src/Val.h
+++ b/src/Val.h
@@ -1656,9 +1656,9 @@ protected:
      */
     void AppendField(ValPtr v, const TypePtr& t) {
         if ( v )
-            record_val.emplace_back(ZValSlot(v, t));
+            record_val.push_back(ZValSlot(v, t));
         else
-            record_val.emplace_back(ZValSlot(t));
+            record_val.push_back(ZValSlot(t));
     }
 
     // For internal use by low-level ZAM instructions and event tracing.
@@ -1710,7 +1710,7 @@ private:
     // Low-level values of each of the fields.
     //
     // Lazily modified during GetField(), so mutable.
-    mutable std::vector<ZValSlot> record_val;
+    mutable detail::vector32<ZValSlot> record_val;
 };
 
 class EnumVal final : public detail::IntValImplementation {
--- a/src/Val.h
+++ b/src/Val.h
@@ -1329,7 +1329,7 @@ public:
      */
     template<class T>
     void AssignField(const char* field_name, T&& val) {
-        int idx = rt->FieldOffset(field_name);
+        int idx = GetRecordType()->FieldOffset(field_name);
         if ( idx < 0 )
             reporter->InternalError("missing record field: %s", field_name);
         Assign(idx, std::forward<T>(val));
@@ -1351,7 +1351,7 @@ public:
         if ( record_val[field] )
             return true;
 
-        return rt->DeferredInits()[field] != nullptr;
+        return GetRecordType()->DeferredInits()[field] != nullptr;
     }
 
     /**
@@ -1361,7 +1361,7 @@ public:
      * @return  Whether there's a value for the given field name.
      */
     bool HasField(const char* field) const {
-        int idx = rt->FieldOffset(field);
+        int idx = GetRecordType()->FieldOffset(field);
         return (idx != -1) && HasField(idx);
     }
 
@@ -1373,14 +1373,14 @@ public:
     ValPtr GetField(int field) const {
         auto& fv = record_val[field];
         if ( ! fv ) {
-            const auto& fi = rt->DeferredInits()[field];
+            const auto& fi = GetRecordType()->DeferredInits()[field];
             if ( ! fi )
                 return nullptr;
 
             fv = fi->Generate();
         }
 
-        return fv.ToVal(rt->GetFieldType(field));
+        return fv.ToVal(GetRecordType()->GetFieldType(field));
     }
 
     /**
@@ -1510,7 +1510,7 @@ public:
 
     template<typename T>
     auto GetFieldAs(const char* field) const {
-        int idx = rt->FieldOffset(field);
+        int idx = GetRecordType()->FieldOffset(field);
 
         if ( idx < 0 )
             reporter->InternalError("missing record field: %s", field);
@@ -1591,7 +1591,7 @@ protected:
     ZValSlot& RawOptField(int field) {
         auto& f = record_val[field];
         if ( ! f ) {
-            const auto& fi = rt->DeferredInits()[field];
+            const auto& fi = GetRecordType()->DeferredInits()[field];
             if ( fi )
                 f = fi->Generate();
         }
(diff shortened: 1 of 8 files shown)


src/Val.h
```diff
@@ -1114,6 +1115,213 @@ struct is_zeek_val {
...
+
+    void reserve(size_t new_capacity) {
+        if ( cap > new_capacity )
+            throw std::length_error("cannot truncate");
+
+        if ( new_capacity > std::numeric_limits<decltype(cap)>::max() )
+            throw std::length_error("new_capacity too large");
+
+        if ( new_capacity > cap ) {
+            std::unique_ptr<T[]> new_data = std::make_unique<T[]>(new_capacity);
```
bbannier: This always constructs `new_capacity` elements (invokes the constructor), `std::vector` OTOH only constructs when elements are added. I suspect this might make this container more expensive which is a potential foot gun.

If we want this alternative class, we probably want it to allocate a contiguous block of memory and then only construct elements if needed.

src/Val.h
```diff
@@ -1114,6 +1115,213 @@ struct is_zeek_val {
...
+ * management without needing to keep external auxilarly information around.
+ *
+ * The is_managed flag and type_tag flags are meant to be immutable except
+ * when a ZValSlot is reassigned.
+ *
+ * A ZValSlot instance holds a reference to a managed value. Such ZValSlot
+ * instances can be copied or assigned and the reference count of the ZVal
+ * will be updated accordingly.
+ */
+class ZValSlot {
```
awelzel: If there's a better name, I'm all up for it. I had `ZValRef` in mind, but not a good intuition. In some sense, this is a helper class supporting optional and automatic ref counting of raw ZVal values.

src/Val.h
awelzel: I'm actually thinking now to just use the unique/raw array directly in RecordVal that always has length `NumFields()` - we always allocate it and don't really need the dynamic growing for a RecordVal.

src/Val.h
evantypanski: Just a fly-by, I'd prefer to avoid "slot" since it conflicts with frame "slots" which is jargon used in an extremely similar context

I wouldn't mind a simple `ZValWrapper` or something, just to avoid conflicting jargon (even ref has some notions that don't really apply here, so I went with something generic). I don't really mind much about specifics, just avoiding existing jargon

awelzel: Perfect. I'd be happy with `ZValWrapper`.

@vpax - any thoughts on the class name? This PR only uses it for the `RecordVal` fields, but I could see it being useful elsewhere as well.

vpax: FWIW I'd (slightly) prefer `ZValElement` since it better conveys that this is one-of-a-number and thus things like `is_managed` and `is_set` make sense.

src/Val.h
```diff
@@ -1114,6 +1115,145 @@ struct is_zeek_val {
...
+    ValPtr ToVal(const TypePtr& t) {
+        assert(IsSet());
+        return zval.ToVal(t);
+    }
+
+private:
+    bool is_set;
+    bool is_managed;
+    TypeTag type_tag;
+    ZVal zval;
```
vpax, line 1266: Just curious - is it clear that this order leads to optimal packing?

awelzel: My understanding is that we cannot go below 16 bytes, as a) ZVal is always 8 byte aligned and its size is 8 byte and b) a class always starts on a 8byte boundary and will always be padded to end on an 8 byte boundary on 64 bit. So the current layout is:

```
is_set: 1 byte
is_managed: 1 byte
type_tag: 1 byte
// 5 byte padding
zval: 8 bytes, 8 byte aligned
// no padding, struct ends on 8 byte bounday.
```

So, I don't think we can do better. We could do worse, causing 24 bytes:
```
is_set: 1 byte
// 7 byte padding
zval: 8 bytes, 8 byte aligned
is_managed: 1 byte
type_tag: 1 byte
// 6 byte padding to fill struct
```

I don't think we should worry about the 5 lost bytes for padding at this point. There's bigger fish to fry (looking at TableVal for example).

I see the goal of this PR twofold: Reducing the size of RecordVal, but also adding the ZValElement as an abstraction to working with raw ZVals more easily. I even wonder if they could work in `Frame`, too, but probably requires a lot of surgery.

src/Val.h
awelzel: I'm going with ZValElement. Thanks!

EDIT: Mostly because `Wrapper` is such a common fallback :-)

awelzel:
> Opening as draft to see if the benchmarks work out.

I've undrafted after reworking a bit. The benchmark shows memory reduction for some microbenchmarks, but it's not super impressive for PCAP reading. I still think it's worth it as `ZValElement` might have uses elsewhere and it reduces some of the redundant information on RecordVal.

Anyone up for looking through the commits and/or the final class/result?

awelzel:
@vpax - would actually like to know your thoughts on that ZValElement class and if could be useful elsewhere. You're probably most knowledgeable to working with the RecordVal internals and ZVal instances with your ZAM work. IMO it's nice once can do `record_val[i].Reset()` or reassign the element and the internal ZVal is unref'ed as needed - it might be useful in `vector of any` in a future PR maybe.

vpax:
@awelzel Yes those are nice properties. I'm not sure it helps with `vector of any` since the problem with those is that each element needs to carry around its type, which `ZValElement`s won't know. That said, switching over to it for low-level access would require a bunch of thought / ZAM instruction changes, so I wouldn't want to aim to take that on any time soon.

src/Val.h
```diff
@@ -1450,20 +1602,21 @@ class RecordVal final : public Val, public notifier::detail::Modifiable {
...
-            record_val.emplace_back(std::nullopt);
+            record_val[num_fields++] = ZValElement(t);
     }
 
     // For internal use by low-level ZAM instructions and event tracing.
     // Caller assumes responsibility for memory management.  The first
     // version allows manipulation of whether the field is present at all.
     // The second version ensures that the optional value is present.
-    std::optional<ZVal>& RawOptField(int field) {
+    ZValElement& RawOptField(int field) {
```
evantypanski, line 1616: I won't fight too much about names since I don't have a much better one, but this is kinda a misnomer now, since it's not returning an optional

src/Val.h
```diff
@@ -1114,6 +1115,159 @@ struct is_zeek_val {
...
+     * @param v The value.
+     * @param t The value's type.
+     */
+    ZValElement(ValPtr v, const TypePtr& t)
+        : is_set(true), is_managed(ZVal::IsManagedType(t)), tag(t->Tag()), zval(v, t) {}
+
+    /**
+     * Initialize a ZValElement using a TypeDecl.
+     */
+    ZValElement(const TypeDecl& td) : is_set(false), is_managed(td.is_managed), tag(td.type->Tag()) {}
```
evantypanski: also my `clang-tidy` complains locally about these redundant `is_set` assignments since it's explicitly set to `false` later (also later):


```suggestion
    ZValElement(const TypeDecl& td) : is_managed(td.is_managed), tag(td.type->Tag()) {}
```

src/Val.h
awelzel: Yep, me too :-)  The `ZValElement` behaves pretty much like a `std::optional` so I opted for leaving as is. I'm happy if this and the one below gets renamed eventually. Or more preferably if `RawField()` just goes away as setting the `ZValElement` might as well be done by the caller and its semantics are rather surprising IMO.

src/Val.h
awelzel: Ah, let me try to improve my local clang-tidy setup. Thanks!

awelzel:
@evantypanski - I've pushed one more commit and top to avoid `type->Tag()` calls and squashed the fixups it. Nothing fundamentally changed, but maybe you can put another :heavy_check_mark:

PR #4861, opened 2025-09-29
Use strlcpy() instead of strncpy()

timwoj:
`strlcpy()` was previously not generally available on Linux, but it was added to glibc in v2.38. This PR switches us away from `strncpy()` calls with that instead, since it's typically safer (in that it forces the null termination). It's also mostly to check if it's available on all of our platforms. It might not be on Windows, but I can add it to `libunistd` if that's the only one missing.


timwoj:
As expected, a bunch of our older platforms failed to build. For example, CentOS Stream 9 is on glibc 2.34. Oh well, it was worth a try.

PR #4862, opened 2025-09-30
Fix for standalone to enable -O gen-standalone-C++ code to extend enum types

vpax:


vpax:
Will re-do, the GH web interface interpreted something I typed as submitting the PR before I finished the description

PR #4866, opened 2025-09-30
Reworking initialization of globals for -O gen-standalone-C++ code

vpax:
Previously, `-O gen-standalone-C++` initialized Zeek global variables after building all of the types, etc. needed for the standalone code. That turns out not to work when for example a record type that's getting built has a `&default` that refers to one of the globals. This PR shifts global initialization into the main initialization framework to address such dependencies.

### DIFF
--- a/src/script_opt/CPP/Consts.h
+++ b/src/script_opt/CPP/Consts.h
@@ -15,8 +15,12 @@ public:
 // Returns the associated initialization info.  In addition, consts_offset
 // returns an offset into an initialization-time global that tracks all
 // constructed globals, providing general access to them for aggregate
-// constants.
+// constants. The second form is for when this isn't needed.
 std::shared_ptr<CPP_InitInfo> RegisterConstant(const ValPtr& vp, int& consts_offset);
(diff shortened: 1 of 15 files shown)


src/script_opt/CPP/Consts.h
```diff
@@ -15,8 +15,12 @@
 // Returns the associated initialization info.  In addition, consts_offset
 // returns an offset into an initialization-time global that tracks all
 // constructed globals, providing general access to them for aggregate
-// constants.
+// constants. The second form is for when this isn't needed.
 std::shared_ptr<CPP_InitInfo> RegisterConstant(const ValPtr& vp, int& consts_offset);
+std::shared_ptr<CPP_InitInfo> RegisterConstant(const ValPtr& vp) {
+    int consts_offset; // ignored
```
timwoj: Adding the `maybe_unused` attribute prevents the compiler from possibly emitting an unused variable warning.

```suggestion
    [[maybe_unused]] int consts_offset; // ignored
```

vpax: Neat, didn't know about that. Happy to push a fixup, or of course fine for you to add.

PR #4869, opened 2025-10-05
Integrate new Broker optimization

Neverlord:
See Broker-side PR https://github.com/zeek/broker/pull/487


ckreibich:
Just an update here after I finally got back to this late last week. Dominik and I spoke at length about this today, so this is more of an fyi to capture current status.

With this changeset the backpressure tests in the cluster testsuite fail, and I'm wondering if the Broker change does something deeper to the buffering/backpressure behavior (which we should avoid given the behavior we've established with 7.1 & 7.2).

I've put together a branch `topic/christian/core-actor-redesign` in [zeek-testing-cluster](https://github.com/zeek/zeek-testing-cluster) that logs more information to stdout to simplify comparison and adds tcpdump to the container so one can trace the local activity on loopback.

I'll focus on the `tests/backend/broker/backpressure-overflow-slow-proxy.sh` test. This test runs a single-host cluster in a single container, so all Broker activity runs over loopback. It uses the disconnect policy and locks up the receiving node (the single proxy node) with `sleep()`s in the script layer so receiving events never get drained, building up backpressure, until the sender (the single worker node) sees the overflow.

This is the behavior on current Zeek `master`:

The sender (the worker node) starts off ...
```
$ less .tmp/tests.backend.broker.backpressure-overflow-slow-proxy/controller/var/lib/nodes/worker/stdout
Backpressure policy: disconnect
Sender ping: 1771888882.11041 0 1
Sender ping: 1771888882.110416 0 2
Sender ping: 1771888882.110418 0 3
Sender ping: 1771888882.11042 0 4
...
```
... and the receiver (the proxy) waits through a few received messages before blocking at the script layer:
```
$ less .tmp/tests.backend.broker.backpressure-overflow-slow-proxy/controller/var/lib/nodes/proxy/stdout
Backpressure policy: disconnect
received ping: 1771888882.195403 0 58
received ping: 1771888882.19541 0 59
received ping: 1771888882.195412 0 60
received ping: 1771888882.195413 0 61
...
received ping: 1771888882.195432 0 74
received ping: 1771888882.195433 0 75
received ping: 1771888882.195435 0 76
1771888882.195454 WEDGING
```
At this point the worker keeps sending events until the proxy slams the receive window shut. You can see this by running
```
$ docker exec -it tests_backend_broker_backpressure_overflow_slow_proxy-controller-1 tcpdump -nn -i lo port 2201
```
... (2201 is the proxy's Broker port) which'll look something like this (note the `win 0`) and then nearly stall:
```
23:21:30.326059 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 824280, win 30, options [nop,nop,TS val 39171818 ecr 39171777], length 0
23:21:30.326087 IP 127.0.0.1.33012 > 127.0.0.1.2201: Flags [P.], seq 824280:831386, ack 2317, win 64, options [nop,nop,TS val 39171818 ecr 39171818], length 7106
23:21:30.367068 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 831386, win 24, options [nop,nop,TS val 39171859 ecr 39171818], length 0
23:21:30.367100 IP 127.0.0.1.33012 > 127.0.0.1.2201: Flags [P.], seq 831386:838834, ack 2317, win 64, options [nop,nop,TS val 39171859 ecr 39171859], length 7448
23:21:30.408086 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 838834, win 17, options [nop,nop,TS val 39171900 ecr 39171859], length 0
23:21:30.408144 IP 127.0.0.1.33012 > 127.0.0.1.2201: Flags [P.], seq 838834:853217, ack 2317, win 64, options [nop,nop,TS val 39171900 ecr 39171900], length 14383
23:21:30.449058 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 853217, win 3, options [nop,nop,TS val 39171941 ecr 39171900], length 0
23:21:30.695068 IP 127.0.0.1.33012 > 127.0.0.1.2201: Flags [P.], seq 853217:856289, ack 2317, win 64, options [nop,nop,TS val 39172187 ecr 39171941], length 3072
23:21:30.695090 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 856289, win 0, options [nop,nop,TS val 39172187 ecr 39172187], length 0
23:21:30.935090 IP 127.0.0.1.33012 > 127.0.0.1.2201: Flags [.], ack 2317, win 64, options [nop,nop,TS val 39172427 ecr 39172187], length 0
23:21:30.935111 IP 127.0.0.1.2201 > 127.0.0.1.33012: Flags [.], ack 856289, win 0, options [nop,nop,TS val 39172427 ecr 39172187], length 0
```
At this point the worker sees what's happening and initiates recovery:
```
$ less .tmp/tests.backend.broker.backpressure-overflow-slow-proxy/controller/var/lib/nodes/worker/stdout
...
Sender ping: 1771888905.698935 0 15331
Sender ping: 1771888905.698987 0 15332
Sender ping: 1771888905.699071 0 15333
Broker::peer_removed removed connection to remote peer (caf::sec::backpressure_overflow)
1771888905.705888 UNPEERED, epoch now 1
```
... at which point receiving resumes and the test succeeds.

Now, the branch version. What I'm seeing is the backpressure mechanism simply not kicking in. The sender keeps sending events to the receiver despite that fact that it locks the script layer, until the test times out. Since nothing's changed in the script layer, it's as if content is read from the socket but then discarded, so the backpressure never makes it to the sender — could that be?

It looks like this: the worker takes off ...
```
$ less .tmp/tests.backend.broker.backpressure-overflow-slow-proxy/controller/var/lib/nodes/worker/stdout
Backpressure policy: disconnect
Sender ping: 1771889560.443585 0 1
Sender ping: 1771889560.443592 0 2
Sender ping: 1771889560.443594 0 3
```
... and the proxy locks up as before:
```
$ less .tmp/tests.backend.broker.backpressure-overflow-slow-proxy/controller/var/lib/nodes/proxy/stdout
Backpressure policy: disconnect
received ping: 1771889560.487314 0 69
received ping: 1771889560.487322 0 70
received ping: 1771889560.487324 0 71
received ping: 1771889560.487325 0 72
...
ceived ping: 1771889560.528353 0 93
received ping: 1771889560.528354 0 94
received ping: 1771889560.528355 0 95
1771889560.528376 WEDGING
```
But at the tcpdump level the receiver never slows down the sender, and the `Broker::peer_removed` message never appears in the worker's stdout. The test eventually times out and fails.

I'll put together equivalent tests for the other backpressure policies since we should have those anyway, but since the behavior isn't visible at the TCP level I suspect it won't act any differently.

Neverlord:
Too late for 8.2

PR #4870, opened 2025-10-05
tools/systemd-generator: Add zeek-systemd-generator

awelzel:
Fakten schaffen :-)

I've had chats at All Systems Go about running Zeek using systemd natively and was pointed at [systemd generators](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html).

This PR adds such a generator for a single-node Zeek deployment. It also introduces the concept of a `<PREFIX>/etc/default/zeek` configuration file which defines the configuration for a systemd managed single-node cluster. Essentially, this allows the following workflow to run a single-node Zeek cluster on Linux with systemd:

```
    # Install the generator - could be done by binary packages or during ninja install, too.
    $ ln -s <prefix>/bin/zeek-systemd-generator /etc/systemd/system-generators/
    # Set the right interface in <prefix>/etc/default/zeek
    $ vim <PREFIX>/etc/default/zeek

    $ systemctl daemon-reload
    $ systemctl start zeek.target
```

The generator is written in C++ as the generator documentation suggested to do it in a fast (they mention C) language. The implementation doesn't have a dependency on Zeek or any other third party library.

I'm opening this as a PR in the zeek repo to a) show the idea and b) because I don't think an extra repository is worth it. It's not that much code (and it's just rendering files to disk). The real magic happens in systemd and its concepts of drop-in files, dependencies etc. IMO.

EDIT: This doesn't yet have support for using systemd's sandboxing features, but I'd add that in a follow-up, certainly.

Personally, it'd be nice to get this in as I'll promote its use for in an internal project, but also think other people can benefit form it. Also, this essentially covers the "easy run Zeek after installing the OBS binary packages on Linux" from my perspective. Just edit `<PREFIX>/etc/default/zeek`, put your interface at the right spot, and two systemd commands. Got a running cluster ready to go! I believe all our supported distros we build for OBS are systemd-based, so this should just work on all of them. It's definitely nothing that would work on FreeBSD, but I tried to at least split the config reading out, so if there's anyone wanting to generate some daemontools services or so for a FreeBSD box, that could reuse that code at least.

EDIT: This doesn't bother at all with multi-node, but I could maybe see adding something to the config that would allow a `<node_type>_offset` and a `cluster_layout_override` option for people that want to run multi-node to provide ways to allow that (putting the burden of worker allocations and cluster-layout generation on them).

Sample configuration:
```
# Zeek Network Security Monitor
#
# Configure the interface this Zeek deployment will be monitoring
# on the local system. Currently, a single interface is supported.
# The interface supports the following substitution variables
# for templating purposes:
#
#   worker_name:   worker-1, worker-2, etc.
#   worker_index:  1, 2, 3, ...
#   worker_index0: 0, 1, 2, ...
#   worker_cpu:    CPU from workers_cpu_list or empty if not set.
#
# Use Linux's AF_PACKET (built-in) relying on the kernel's
# default packet dissector for load balancing.
# interface = af_packet::wlp0s20f3
#
# # Use NETMAP.
# interface = netmap::eth0}${worker_index0}
#
# # Use libpcap for listening on loopback
# interface = lo
#
# # Other examples:
# interface = napatech::${worker_index0}     # Napatech streams
# interface = af_packet::if-${worker_name}   # Individual interfaces for workers.
#
# An empty interface means Zeek is disabled on the system.
interface =

# The number of Zeek worker processes to run.
workers = 4

# The list of CPUs on which to pin workers (modulo semantics).
# Pinning is disabled when this list is empty.
# TODO: Support ranges and strides.
# workers_cpu_list = 0,1,2,3

# The NUMA policy to set for all workers.
# Allowed values are currently default and local.
workers_numa_policy = local

# The number of logger processes to run.
loggers = 1

# The number of proxy processes to run.
proxies = 2

# Arguments passed to all Zeek nodes.
#
# Keep this list short. More extensive configuration should be done
# via local.zeek or the Configuration framework. Worker processes have
# the ``-i <interface>`` prepended.
args = local ignore_checksums=T frameworks/cluster/backend/zeromq/connect

# User and group to run Zeek under.
user = zeek
group = zeek

# Override Zeek's base directory.
# base_dir =

# Prepend the following paths to ZEEKPATH. You might want to use this
# with scripts in unusual locations and you want to let Zeek know about
# these. If you solely use zkg, you should not need to set anything here.
# ext_zeek_path = /path/to/extra1:/path/to/extra2

# The base port for Zeek's listener ports in a Zeek cluster.
# port = 27760
#
# The base port for the Prometheus HTTP listener for each process.
# metrics_port = 9991

# Default listening addresses to use for cluster functionality.
# address = 127.0.0.1
#
# The listening address for the Prometheus HTTP listener.
# This will be redef'ed in cluster-layout.zeek
# metrics_address = 0.0.0.0
#
#
# Whether to run the zeek-archiver process.
#
# Zeek logger processes usually rotate their completed log files into
# <PREFIX>/spool/log-queue. The zeek-archiver process then picks up
# these logs and archives them into <PREFIX>/logs/<date>. If your Zeek
# configuration is such that logs are directly sent to Kafka or another
# queueing system, set this to 0.
archiver = 1

# The directory into which zeek-archiver will archive logs.
# archiver_dir = <PREFIX>/logs

# Extra arguments passed to the zeek-archiver invocation.
archiver_args = -v -i 5

# The maximum amount of memory an individual process type may
# consume. The default unit is bytes, you may add K, M, G, T
# suffixes as supported by systemd. On Linux, this is enforced
# using cgroups.
#
# <node_type>_memory_max = 1234M
manager_memory_max =
logger_memory_max =
proxy_memory_max =
worker_memory_max =

# The nice levels to configure for individual processes types.
manager_nice =
worker_nice =
proxy_nice =
logger_nice =
```


awelzel:
Gave a demo to the team. Removing RFC / draft state. The code is so so, if it looks like it's not really there yet, put the high-level points and I'm happy to polish it.

tools/systemd-generator/README.md
```diff
@@ -0,0 +1,38 @@
...
+This directory contains code for a [systemd generator](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html)
+to deploy a single-node Zeek deployment on modern Linux distributions.
+
+## Installation
+
+Zeek's default build on Linux contains a ``zeek-systemd-generator`` executable
+in its bin directory. To enable the generator, link it into one of systemd's
+generator directories:
+
+    $ /usr/local/zeek/bin/zeek-systemd-generator /etc/systemd/system-generators/
```
evantypanski: should this actually link it instead?


```suggestion
    $ ln -s /usr/local/zeek/bin/zeek-systemd-generator /etc/systemd/system-generators/
```

tools/systemd-generator/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+After that, modify ``<PREFIX>/etc/default/zeek`` as needed. Particularly, set
+the ``interface`` key::
```
evantypanski: sorry I can't help myself

```suggestion
the ``interface`` key:
```

evantypanski (commented):
I tried to play with this a bit this evening and had some comments in the instructions; not an actual review :)

By the way I like this being opinionated

tools/systemd-generator/README.md
awelzel: Thanks!

tools/systemd-generator/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+project(systemd-zeek-generator)
+
+if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    set(_default_enable_systemd_generator on)
+endif ()
+
+option(ENABLE_ZEEK_SYSTEMD_GENERATOR "Enable building and installing zeek-systemd-generator"
+       ${_default_enable_systemd_generator})
+
+if (ENABLE_ZEEK_SYSTEMD_GENERATOR)
```
bbannier, line 24: Can we always build this target? It has no system dependencies and the test should pass on e.g., macos as well.

I.e., drop `_default_enable_systemd_generator` and the `if` here. If you really wanted to you could make the `install` below conditional, but TBH I wouldn't mind having this present on e.g., macos or freebsd (the name makes the purpose pretty clear).

tools/systemd-generator/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    # Separate supporting libraries that aren't systemd specific.
+    add_library(systemd-generator STATIC src/systemd-unit.cc)
+    add_library(zeek-cluster-config STATIC src/zeek-cluster-config.cc)
```
bbannier: Unsure why these can't just be part of the executable, if anything lets make them `OBJECT` libraries so nobody external links against them ever.


```suggestion
    add_library(systemd-generator OBJECT src/systemd-unit.cc)
    add_library(zeek-cluster-config OBJECT src/zeek-cluster-config.cc)
```

tools/systemd-generator/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+    add_executable(zeek-systemd-generator src/zeek-systemd-generator.cc)
+    target_compile_definitions(
+        zeek-systemd-generator
+        PRIVATE "DEFAULT_CONFIG_FILE=\"${ZEEK_ETC_INSTALL_DIR}/default/zeek\"")
+    target_compile_definitions(zeek-systemd-generator
+                               PRIVATE "ZEEK_BASE_DIR=\"${CMAKE_INSTALL_PREFIX}\"")
```
bbannier: Nit.

```diff
     target_compile_definitions(
         zeek-systemd-generator
-        PRIVATE "DEFAULT_CONFIG_FILE=\"${ZEEK_ETC_INSTALL_DIR}/default/zeek\"")
-    target_compile_definitions(zeek-systemd-generator
-                               PRIVATE "ZEEK_BASE_DIR=\"${CMAKE_INSTALL_PREFIX}\"")
+        PRIVATE "DEFAULT_CONFIG_FILE=\"${ZEEK_ETC_INSTALL_DIR}/default/zeek\""
+                "ZEEK_BASE_DIR=\"${CMAKE_INSTALL_PREFIX}\"")
```

tools/systemd-generator/src/zeek-systemd-generator.cc
```diff
@@ -0,0 +1,233 @@
...
+    manager_unit.Write();
+    logger_unit.Write();
+    proxy_unit.Write();
+    worker_unit.Write();
+}
+
+} // namespace
+
+int main(int argc, const char* argv[]) {
+    const char* normal_dir;
```
bbannier: Nit.


```suggestion
    std::string_view normal_dir;
```

tools/systemd-generator/src/zeek-systemd-generator.cc
```diff
@@ -0,0 +1,233 @@
...
+    }
+    else {
+        std::cerr << "Usage: " << argv[0] << " normal-dir [early-dir] [late-dir]\n";
+        std::exit(1);
+    }
+
+    // Default configurations to attempt to load.
+    std::vector<std::filesystem::path> config_files = {
+        DEFAULT_CONFIG_FILE,
+        "/etc/default/zeek",
```
bbannier: Leftover? Seems to duplicate `DEFAULT_CONFIG_FILE`.

tools/systemd-generator/src/zeek-systemd-generator.cc
```diff
@@ -0,0 +1,233 @@
...
+    };
+
+    // CONFIG_FILE environment overrides default search path.
+    const char* env_config_file = std::getenv("CONFIG_FILE");
+    if ( env_config_file && env_config_file[0] ) {
```
bbannier: Nit. That's an unusual way to check that the string is not empty, I'd just go with

```suggestion
    if ( env_config_file && std::strlen(env_config_file) ) {
```

tools/systemd-generator/src/zeek-cluster-config.cc
```diff
@@ -0,0 +1,224 @@
...
+        }
+        else if ( key == "workers_cpu_list" ) {
+            config.workers_cpu_list = CpuList(entry.value);
+        }
+        else if ( key == "port" ) {
+            config.port = std::atoi(entry.value.c_str());
+        }
+        else if ( key == "address" ) {
+            config.address = entry.value;
+        }
```
bbannier, line 359: clang-tidy points out that this duplicates the `listen_address` branch, do we need both?

tools/systemd-generator/src/zeek-cluster-config.cc
```diff
@@ -0,0 +1,224 @@
...
+
+    std::size_t pos = 0;
+
+    while ( pos < content.size() ) {
+        auto nl_pos = content.find('\n', pos);
+
+        std::string orig_line = content.substr(pos, nl_pos - pos);
+        std::string line = orig_line;
+
+        trim(line);
```
bbannier, line 64: This is to support whitespace around the `=`. With that it is e.g., impossible to have values start with significant whitespace. I am not sure that is useful now, but I could see it becoming handy for e.g., hacking something into `args`.

WDYT about either making whitespace significant, at least on the value side? Alternative we could allow some form of quoting on the value side.

Mostly probably fine to leave for a later cleanup, but it has some implications for what format we bake into the API.

tools/systemd-generator/src/zeek-cluster-config.cc
awelzel: Ah, thanks, dropped.

tools/systemd-generator/src/zeek-systemd-generator.cc
awelzel: `DEFAULT_CONFIG_FILE` is actually `<PREFIX>/etc/default/zeek` and the extra one is for looking into `/etc/default/zeek` as well.

How about `/etc/zeek/zeek.conf` like `/etc/redis/redis.conf` though. The ones in `default` look like environment variables quite a bit, so separating from there might be good.

tools/systemd-generator/src/zeek-cluster-config.cc
awelzel: As discussed offline, will continue to trim whitespace on both sides of the value.

ekoyle:
I really like the idea. This should make zeek behave more like any other package (at least on Linux), which I am highly in favor of.

One minor note: `/etc/default` is a somewhat Debian-centric approach for managing the behavior of init scripts (traditionally these files were sourced by the init script of the same name so that you don't have to edit the script to influence command-line arguments). I think its use with systemd is frowned upon.

It might be cleaner to use a config file in a sensible zeek location by default.

awelzel:
> It might be cleaner to use a config file in a sensible zeek location by default.

Yep, thanks. I had already changed this locally to `<PREFIX>/etc/zeek/zeek.conf` (https://github.com/zeek/zeek/pull/4870#discussion_r2417390150)

awelzel:
I've squashed everything together after:
* Adding zeek-archiver support
* Adding configurable memory_max and nice value per node type
* Implemented interface variable substitution (e.g. `netmap::eth1}${worker_index0}` or `af_packet::${worker_name}` if one has interface names according to zeek worker names.)
* Addressed Benjamin's comments. This now builds and installs also on FreeBSD and OSX. No strong opinion from my side if it should or not. It can be disabled via CMake by packagers.

I've updated the example configuration in the PR description.

[Strides and ranges](https://docs.kernel.org/admin-guide/kernel-parameters.html#cpu-lists) for the `workers_cpu_list` would be nice and also sandboxing, but I think these can come as follow up.

tools/systemd-generator/etc/zeek/zeek.conf
```diff
@@ -0,0 +1,111 @@
...
+
+# The number of proxy processes to run.
+proxies = 2
+
+# Arguments passed to all Zeek nodes.
+#
+# Keep this list short. More extensive configuration should be done
+# via local.zeek or the Configuration framework. Worker processes have
+# the ``-i <interface>`` prepended.
+args = local ignore_checksums=T frameworks/cluster/backend/zeromq/connect
```
timwoj: Should this be left commented out by default?

tools/systemd-generator/src/zeek-cluster-config.cc
```diff
@@ -242,6 +205,72 @@ std::optional<std::string> ZeekClusterConfig::SubstituteVars(const std::string&
     return result;
 }
 
+// More grumble.
+CpuList::CpuList(const std::string& list) {
+    size_t pos = 0;
+    while ( pos < list.size() ) {
```
timwoj: How about https://www.en.cppreference.com/w/cpp/ranges/split_view.html?

scripts/policy/misc/systemd-generator.zeek
```diff
@@ -0,0 +1,18 @@
+# Opinonated configuration for a cluster configured with the zeek-systemd-generator.
```
timwoj:
```suggestion
# Opinionated configuration for a cluster configured with the zeek-systemd-generator.
```

And one other below.

tools/systemd-generator/README.md
```diff
@@ -0,0 +1,42 @@
+# zeek-systemd-generator
+
+This directory contains code for a [systemd generator](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html)
+to run a single-node Zeek deployment on modern Linux distributions.
+
+## Installation
+
+Zeek's default build on Linux contains a ``zeek-systemd-generator`` executable
```
timwoj:
```suggestion
Zeek's default installation on Linux contains a ``zeek-systemd-generator`` executable
```

tools/systemd-generator/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+To verify the ``zeek-systemd-generator`` produces the expected unit files for
+a given configuration, invoke it with a single testing directory:
+
+    $ mkdir test-directory
+    $ <PREFIX>/bin/zeek-systemd-generator test-directory
+
+``zeek-systemd-generator`` will attempt to read configuration files at
+``<PREFIX>/etc/default/zeek`` and ``/etc/default/zeek``. It's possible
+to set the environment variable ``CONFIG_FILE`` to use a different path.
```
timwoj: > environment variable ``CONFIG_FILE`

This feels like a really generic environment variable name, i.e. probably not something you'd want set permanently in your environment. Are you just intending for to be set via the command-line when you call the generator? Could it be an argument to the generator instead?

timwoj (changes requested):

tools/systemd-generator/README.md
awelzel: `CONFIG_FILE` was primarily to point the generator to a custom config, but `--config` works for me, `getenv()` was just quicker to implement :-)

tools/systemd-generator/etc/zeek/zeek.conf
awelzel: I'm not quite sure yet, but I think yes.

I am thinking of adding `cluster_backend_args` as separate configuration value, defaulting to zeromq, also commented out.

cluster_backend_args

ekoyle: I think we should consider changing the zeek default for `ignore_checksums` to true. I know it's been like this forever, but the current default:

  * Causes confusion due to missing data when processing pcaps collected from an interface with checksum offloading enabled (i.e. most modern systems)
  * Silently hides serious network problems which would otherwise be easily detectable via conn history

(edit: kind of off topic, sorry)

awelzel: > I think we should consider changing the zeek default for ignore_checksums to true. I know it's been like this forever, but the current default:

> (edit: kind of off topic, sorry)

All good - I could maybe see it as a default in `local.zeek`? Maybe? Feel free to open a discussion.

timwoj (approved):

PR #4872, opened 2025-10-08
QUIC: Skip packets with fixed_bit 0

### LINKED ISSUE #4847: QUIC violation: expected 3 bytes (0 available) from quic/QUIC.spicy:466:5-466:50
Relates to #4461 - observing these QUIC violations my local WLAN network and ran a capture for a bit to grab these.

[quic-39264-rand.pcap.txt](https://github.com/user-attachments/files/22519558/quic-39264-rand.pcap.txt)
[quic-38677-rand.pcap.txt](https://github.com/user-attachments/files/22519559/quic-38677-rand.pcap.txt)

```
#fields ts      analyzer_kind   analyzer_name   uid     fuid    id.orig_h       id.orig_p       id.resp_h       id.resp_p       proto   failure_reason  failure_data
#types  time    string  string  string  string  addr    port    addr    port    enum    string  string
1758728767.168016       protocol        QUIC    C95jX14kZxOxIPu3J8      -       765b:b77b:dad1:7e73:56f9:5a0f:52a3:f4f  39264   725f:1f71:525f:a3b:525f:3673:525f:3673  443     udp     expected 3 bytes (0 available) (/home/awelzel/corelight-oss/zeek/src/analyzer/protocol/quic/QUIC.spicy:466:5-466:50)  \xcf\x00\x00\x00\x01\x08\xce\xefy\x90\xf5\xbb@q\x03\x1em\xc7@F\x00\xe6\x88N\xea\xe9(\xc4\xa08\xde\xf6e8N\x18\xb6\x0aE5
```
Wireshark reports malformed packet, but doesn't completely fail on it. Wonder if it's really just this trailing `\x00`, but haven't looked in earnest.

Image

awelzel:
The RFC specifies that QUIC packets with unset fixed_bit need to be discarded. Do so.

Fixes #4847


src/analyzer/protocol/quic/QUIC.spicy
```diff
@@ -539,7 +556,9 @@ type Packet = unit(from_client: bool, context: Context&) {
                 self.crypto_sink = Null;
             }
         }
-    };
+
+        # Only parse the packet if fixed_bit == 1
+    } if(self.first_byte.fixed_bit);
```
bbannier: It isn't really clear to me from the code whether we really "discard the package" if `fixed_bit` is unset, e.g., I would need to check that `packet_data` or `frames` have no hooks attached, and that the value of computed fields like `decrypted_data` makes sense.

Is there a reason we cannot simply wrap parsing we wouldn't want to do in a conditional block? This would also allow us to keep the checking in a single place instead of having to sprinkle it in field attributes far away.

src/analyzer/protocol/quic/QUIC.spicy
bbannier: FYI, I fixed spicy-format to correctly indent the block and released 0.27.0.

src/analyzer/protocol/quic/QUIC.spicy
awelzel: > Is there a reason we cannot simply wrap parsing we wouldn't want to do in a conditional block?

Just didn't like the extra indentation and I think the `if()s` do the right thing, but I agree it's spaghetti.

Let me do the huge switch indent.

awelzel: @timwoj - this should go into 8.0.2. Sorry for the immense diff.

PR #4873, opened 2025-10-08
Introduce QUIC::discarded_packet() event and include as X in QUIC history

awelzel:
Addition to #4872, but not for 8.0.2.

Requires: https://github.com/zeek/zeek-testing/pull/25


0xxon:
I think you forgot the test baseline :)

PR #4874, opened 2025-10-08
Replace g_frame_stack with call_stack used by ScriptProfiler

timwoj:
This is a third try at reducing the CPU time spent in debugger code when not actually use the debugger. @awelzel noted in the last PR that we're currently storing two copies of the call stack: one for the debugger, and one for the script profiler. This PR combines the uses of the former into the latter, and only keeps that copy instead of both.

it also makes a change to how we handle a couple of the script-level statements, abstracting away the code the would be used in the debugger so that it's not used during normal run-times.

### DIFF
--- a/src/Frame.h
+++ b/src/Frame.h
@@ -20,11 +20,11 @@ using ValPtr = IntrusivePtr<Val>;
 
 class BrokerListView;
 class BrokerData;
+class Func;
 
 namespace detail {
 
 class CallExpr;
-class ScriptFunc;
 using IDPtr = IntrusivePtr<ID>;
 
 namespace trigger {
@@ -47,7 +47,7 @@ public:
      * @param func the function that is creating this frame
      * @param fn_args the arguments being passed to that function.
      */
-    Frame(int size, const ScriptFunc* func, const zeek::Args* fn_args);
+    Frame(int size, const Func* func, const zeek::Args* fn_args);
 
     /**
      * Returns the size of the frame.
@@ -117,7 +117,7 @@ public:
     /**
      * @return the function that the frame is associated with.
      */
-    const ScriptFunc* GetFunction() const { return function; }
+    const Func* GetFunction() const { return function; }
 
     /**
      * @return the arguments passed to the function that this frame
@@ -130,7 +130,7 @@ public:
      *
      * @param func the function for the frame to be associated with.
      */
-    void SetFunction(ScriptFunc* func) { function = func; }
+    void SetFunction(Func* func) { function = func; }
 
     /**
      * Sets the next statement to be executed in the context of the frame.
@@ -239,7 +239,7 @@ private:
     const OffsetMap* captures_offset_map = nullptr;
 
     /** The function this frame is associated with. */
-    const ScriptFunc* function = nullptr;
+    const Func* function = nullptr;
 
     // The following is only needed for the debugger.
     /** The arguments to the function that this Frame is associated with. */
@@ -266,4 +266,4 @@ private:
  * DebugFrame which provides the information that the debugger uses. See:
  * https://stackoverflow.com/a/16211097
  */
-extern std::vector<zeek::detail::Frame*> g_frame_stack;
+// extern std::vector<zeek::detail::Frame*> g_frame_stack;
--- a/src/Func.cc
+++ b/src/Func.cc
@@ -93,10 +93,11 @@ std::string render_call_stack() {
         if ( lvl > 0 )
             rval += " | ";
 
-        const auto& name = ci.func->GetName();
+        const auto& name = ci.frame->GetFunction()->GetName();
         std::string arg_desc;
 
-        for ( const auto& arg : ci.args ) {
+        const auto& args = ci.frame->GetFuncArgs();
+        for ( const auto& arg : *args ) {
             ODesc d;
             d.SetShort();
             arg->Describe(&d);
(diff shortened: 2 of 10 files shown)


src/Frame.h
```diff
@@ -266,4 +267,4 @@ class Frame : public Obj {
  * DebugFrame which provides the information that the debugger uses. See:
  * https://stackoverflow.com/a/16211097
  */
-extern std::vector<zeek::detail::Frame*> g_frame_stack;
+// extern std::vector<zeek::detail::Frame*> g_frame_stack;
```
timwoj, line 269: I left this in commented out, solely because of the comment above it. There may be more we can do here, but it would require some actual cycles spent on the debugger.

timwoj:
This doesn't result in nearly the same savings as the previous PR. The first image below is from #4852, the second one is from this PR. There are still some savings in a few of the benchmarks, but it's actually a loss in others that were gains in the first version.

Screenshot 2025-10-08 at 11-11-58 PR-Metrics - Dashboards - Grafana

Screenshot 2025-10-08 at 11-11-04 PR-Metrics - Dashboards - Grafana

awelzel:
> This doesn't result in nearly the same savings as the previous PR. 

I saw it too :-( My understanding is that BIF function calling got significantly slower with this change (bar in the middle), while the `StmtList` / `IfStmt` changes from a351a73e06237c999533ecec865f51f862dd14c9 probably help everywhere else. The logging tests all go through `Log::write()` so if BifFuncs get slower, they all get slow.

image


Maybe a separate PR for just that StmtList/IfSmt commit? That seems reasonable to me. For the frame / function calls improvements, probably needs some low-level hacking. I know ZAM as specialized e.g. `Log::write()` to not go through ScriptFunc / BifFunc at all, so a bit unsure if it's really worth looking more. At the same time, it'd be nice if function calls were much less expensive and optimizing them not required.

Heh, I just realized I looked at this before - see #4182 (removing Obj inheritance from Frame and using placement this for frames to allocate frame and slots at once). I think it helped, but I left the same comment about ZAM there before :-)

timwoj:
> I saw it too :-( My understanding is that BIF function calling got significantly slower with this change (bar in the middle)

Turns out I had left an errant debugging `printf` in a spot where only the BIF code was hitting it. I also caught a few places where I could be using references instead of copying the frame pointers. It should be much faster now. I'll wait for benchmark results. It might be because it's storing a `FramePtr` in `CallInfo` now, and so it's having to do the lookup against that instead of looking at the `Func*` and `Args` directly. On top of that, it might be hitting some vtable lookups since Frame stores a `Func*` instead of a `ScriptFunc*`.

timwoj:
Yep, most of the hit is from the Frame allocation and deallocation in `BuiltinFunc::Invoke`. Top is master, bottom is branch:

Screenshot 2025-10-10 at 4 22 08 PM
Screenshot 2025-10-10 at 4 22 00 PM

src/Frame.h
```diff
@@ -117,7 +118,7 @@ class Frame : public Obj {
     /**
      * @return the function that the frame is associated with.
      */
-    const ScriptFunc* GetFunction() const { return function; }
+    const Func* GetFunction() const { return function; }
```
awelzel, line 120: Is it worth templatizing this one defaulting to  `<T = Func>`, but allow passing BuiltinFunc or ScriptFunc classnames? Could then put a debug-only assert if function kind isn't matching the class we want to cast to. And it avoids the `static_cast` on the call side.

src/Func.cc
```diff
@@ -331,17 +332,16 @@ ValPtr ScriptFunc::Invoke(zeek::Args* args, Frame* parent) const {
         return Flavor() == FUNC_FLAVOR_HOOK ? val_mgr->True() : nullptr;
     }
 
-    auto f = make_intrusive<Frame>(frame_size, this, args);
+    Frame f{static_cast<int>(frame_size), this, args};
```
awelzel, line 335: I didn't feel brave enough to do that, but I like it!  This allocates the Frame object on the native C++ stack instead of the heap, so its lifetime is now bound to the native C++ stack. The elements/slot array will still be on the heap, but that's separate (and automatically freed when the frame is destructed). This now also implies that `Frame` should not to be a `Obj` or ref-counted anymore (indeed `Unref()`  on a stack allocated `Obj` is very dangerous).

So 👍👍 from me and some day someone follows up with removing `Obj` inheritance from `Frame`? :-)

src/Func.h
```diff
@@ -339,21 +340,16 @@ class BuiltinFunc final : public Func {
...
 };
 
 extern bool check_built_in_call(BuiltinFunc* f, CallExpr* call);
 
 struct CallInfo {
-    const CallExpr* call;
-    const Func* func;
-    const zeek::Args& args;
+    const CallExpr* call = nullptr;
+    Frame* frame = nullptr;
```
awelzel, line 351: This is great. I had wondered if `Frame` isn't anyhow tracking all that info. In fact, I wonder if `call` (parent call expr) could be part of `Frame` and suddenly `call_stack` is just frames.

Nothing to be done here, but could be a follow-up.

src/Func.h
```diff
@@ -42,6 +42,7 @@ class Frame;
 using ScopePtr = IntrusivePtr<Scope>;
 using IDPtr = IntrusivePtr<ID>;
 using StmtPtr = IntrusivePtr<Stmt>;
+using FramePtr = IntrusivePtr<Frame>;
```
awelzel: Leftover?

src/Frame.h
timwoj: There's a couple of ways I could do that assert, but both of them would require the full type information instead of just forward declarations. I was trying not to have to include Func.h here.

awelzel: The assert was mostly a nice to have.

timwoj:
I realized late that `g_frame_stack` isn't in a namespace and should technically go through a deprecation cycle. I'm opting to skip it for the following reasons:
- The likelihood of external use of this variable is probably very low, and those uses can be adapted. We only know of one external use at the moment.
- The only way to deprecate it involves either keeping the variable around entirely (defeating the point of this PR) or creating some sort of wrapper to the frame stack that mimics `std::vector` but looks at `call_stack` internally, which is complicated enough that it's not worth it.

I added a note to NEWS in the Breaking Changes about the removal and the lack of deprecation.

PR #4876, opened 2025-10-09
Add TCP::raw_option_list BiF

0xxon:
This BiF extracts the full list of options with their raw, unparsed data from the current TCP packet.


src/analyzer/protocol/tcp/functions.bif
```diff
@@ -138,3 +139,76 @@ function get_contents_file%(cid: conn_id, direction: count%): file
...
+	static auto tcp_option_list = zeek::id::find_type<zeek::VectorType>("TCP::RawOptionList");
+	auto result = zeek::make_intrusive<zeek::VectorVal>(tcp_option_list);
+
+	static const auto& analyzer = zeek::packet_mgr->GetAnalyzer("TCP");
+	auto spans = zeek::packet_mgr->GetAnalyzerData(analyzer);
+
+	if ( spans.empty() || spans[0].size() < sizeof(struct tcphdr) )
+		return result;
+
+	const auto* tcp_hdr = reinterpret_cast<const struct tcphdr*>(spans[0].data());
```
awelzel, line 163: Neat feature!

One high-level comment: For Geneve options the parsing and `Val` construction, I had introduced a callback version.

https://github.com/zeek/zeek/blob/ba6f34d8497c2a7a6dfc9b67ee3a70d7fa6726e6/src/packet_analysis/protocol/geneve/Geneve.h#L24-L30

https://github.com/zeek/zeek/blob/ba6f34d8497c2a7a6dfc9b67ee3a70d7fa6726e6/src/packet_analysis/protocol/geneve/functions.bif#L28-L39


Think Jan eyed at the `Geneve::parse_options()` code to extract just a single option outside of the script-layer in a plugin without needing RecordVal instances etc. It looks like this pattern could work here, too? Or does that seem like it's overdoing it?

testing/btest/bifs/tcp_raw_option_list.zeek
```diff
@@ -0,0 +1,27 @@
...
+# @TEST-EXEC: zeek -C -r $TRACES/tls/tls1_1.pcap %INPUT
+# Ensure correct results when run against non-TCP-traffic
+# @TEST-EXEC: zeek -C -r $TRACES/tls/dtls1_0.pcap %INPUT
+# @TEST-EXEC: btest-diff .stdout
+
+global packet_number: count = 0;
+
+event connection_SYN_packet(c: connection, pkt: SYN_packet)
+	{
+	print c$id, TCP::raw_option_list();
```
awelzel: How would just `TCP::raw_options()` sound? Are you particularly attached to `list` ?

0xxon: Nope, don't care at all. So - sure

src/analyzer/protocol/tcp/functions.bif
0xxon: For me personally that feels like it's overdoing things - at the moment this is a single, self-contained BiF - which is pretty easy to argue about and is something that I have seen several people need (there is one workaround in existance that I am aware of, and I needed this functionality for a script).

A callback-based architecture seems like a separate project to integrate into the TCP option parsing of the TCP analyzer.

0xxon:
Renamed it as suggested; will merge if tests pass.

PR #4877, opened 2025-10-09
Make UIDs returned by unique_id BIFs consistent with other UIDs

### LINKED ISSUE #4513: unique_id and unique_id_from make IDs shorter than the new standard UID length
I was making a connection from scriptland, and noticed its UID was short.
The functions seem to have the old UID size baked in.  (effectively ~12 bytes instead of ~18)
There are a few options to resolve:

- Unilaterally change to use bits_per_uid (plus: matches others, minus - suddenly they get longer)
- add an optional param to specify the desired size (plus: backwards compat, more flexability, minus: does anyone really care? may be adding options to no good value)

I'd be happy to do either, or a third better suggestion, but wondered what others thought was the "right" course.  (for me, the first one, tho I started thinking the second, so am certainly not wedded to it!)

timwoj:
This changes the `unique_id` and `unique_id_from` BIFs to use the same `UID` class that we use for Conn IDs. This causes the length of those UIDs to be consistent with the others, as well as removing an extra code path for how we generate UIDs.

Fixes #4513


src/NetVar.cc
```diff
@@ -181,7 +181,7 @@ int dpd_ignore_ports;
 
 int record_all_packets;
 
-zeek_uint_t bits_per_uid;
+zeek_uint_t bits_per_uid = 0;
```
timwoj, line 184: Part of me really wants to set default values for all of these variables, but I left it alone for this PR and only set the one that I cared about. I'll open another PR that sets the rest of them.

PR #4878, opened 2025-10-10
Skip setting tm_gmtoff on Windows

timwoj:
tm_gmtoff is a POSIX extension to the tm struct that Windows doesn't support it. It doesn't appear to change the output from the btest, so we can just skip it.

### DIFF
--- a/src/analyzer/protocol/smb/smb-time.pac
+++ b/src/analyzer/protocol/smb/smb-time.pac
@@ -33,9 +33,12 @@ double time_from_lanman(uint32_t smb_time, uint32_t smb_date, uint16_t tz)
 	lTime.tm_year = ((smb_date >> 9) & 0x7f) + 80;
 	lTime.tm_isdst = -1;
 
+#ifndef _MSC_VER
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/smb/smb-time.pac
```diff
@@ -33,9 +33,12 @@ double time_from_lanman(uint32_t smb_time, uint32_t smb_date, uint16_t tz)
 	lTime.tm_year = ((smb_date >> 9) & 0x7f) + 80;
 	lTime.tm_isdst = -1;
 
+#ifndef _MSC_VER
 	// The timezone passed in the data is the number of minutes from UTC, while
-	// tm_gmtoff is the number of seconds east of UTC. Adjust for that.
+	// tm_gmtoff is the number of seconds east of UTC. Adjust for that. This fueld
```
awelzel:
```suggestion
	// tm_gmtoff is the number of seconds east of UTC. Adjust for that. This field
```

timwoj: Clearly I don't have pre-commit enabled on Windows.

PR #4881, opened 2025-10-10
Add versions of IfStmt and StmtList for use in debugger

timwoj:
This splits some of the changes out of https://github.com/zeek/zeek/pull/4874 into a separate PR. Specifically, this moves some code that is needed by the script debugger out into specialized classes that are only used by the debugger, avoiding calling the code when the debugger isn't active.

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -163,6 +163,10 @@ static int func_hdr_cond_epoch = 0;
 EnumType* cur_enum_type = nullptr;
 static ID* cur_decl_type_id = nullptr;
 
+namespace zeek::detail {
+extern bool g_policy_debug;
+}
+
 static void parse_new_enum(void) {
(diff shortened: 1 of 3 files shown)


src/parse.y
```diff
@@ -163,6 +163,10 @@ static int func_hdr_cond_epoch = 0;
 EnumType* cur_enum_type = nullptr;
 static ID* cur_decl_type_id = nullptr;
 
+namespace zeek::detail {
+    extern bool g_policy_debug;
```
awelzel: Nit. Don't think we usually indent in namespaces.

timwoj:
Benchmark numbers look pretty good

Screenshot 2025-10-10 at 10-43-52 PR-Metrics - Dashboards - Grafana

PR #4884, opened 2025-10-11
ZAM: Fix memory leak in Construct-Known-Record-With-Inits

awelzel:
Commit b5d17f7f49c32e2b87856da01b51cf512c52e0e9 mistakenly removed Unref() for init_vals pre-generated, but then potentially replaced by values from a record. Re-instantiate it and add a test that tickled the leak.

This also replaced some autos with the concrete types, even if they are a mouthful. Mostly to make my poor man's grep-based workflow a bit easier. Also, given not everything is in .cc files, using an IDE to figure out types is semi successful (works for the generated code only).


vpax:
@awelzel happy to review, but FYI I'm out on PTO for the next week, so it'll be a while before I circle back to this

awelzel:
> @awelzel happy to review, but FYI I'm out on PTO for the next week, so it'll be a while before I circle back to this

Thanks! If it goes in before maybe you can review what landed.

I made a mistake in the referenced commit, so this is partly a revert of https://github.com/zeek/zeek/commit/b5d17f7f49c32e2b87856da01b51cf512c52e0e9 which at least makes sense to me :-)

src/script_opt/ZAM/OPs/constructors.op
```diff
@@ -76,11 +76,20 @@ macro	AssignFromRec(rhs)
...
 		{
-		auto rhs_i = FieldValWithCheck(rhs, rhs_map[i]);
-		auto& init_i = init_vals[lhs_map[i]];
+		ZVal rhs_i = FieldValWithCheck(rhs, rhs_map[i]);
+		std::optional<ZVal>& init_i = init_vals[lhs_map[i]];
+		/*
+		 * ConstructRecordPost() used after AssignFromRec() calls the
+		 * optimization specific RecordVal(t, init_vals) constructor
+		 * which adpots managed ZVal references, so need to 1) ref any
+		 * managed fields from rhs and 2) unref managed managed ZVal
```
evantypanski:
```suggestion
		 * managed fields from rhs and 2) unref managed ZVal
```

src/script_opt/ZAM/OPs/constructors.op
```diff
@@ -76,11 +76,20 @@ macro	AssignFromRec(rhs)
...
+		 * values existing in init_vals created during GenInits().
+		 */
 		if ( is_managed[i] )
 			{
 			zeek::Ref(rhs_i.ManagedVal());
+			if ( init_i.has_value() )
+				zeek::Unref(init_i->ManagedVal());
```
evantypanski, line 92: If this were more common I'd consider some `zeek::Move`, but I couldn't find that actually being useful

src/script_opt/ZAM/ZInstAux.h
```diff
@@ -175,7 +175,7 @@ class ZInstAux {
     // Same, but using the "map" to determine where to place the values.
     // Returns a non-const value because in this situation other updates
     // may be coming to the vector, too.
-    auto& ToZValVecWithMap(const ZVal* frame) {
+    std::vector<std::optional<ZVal>>& ToZValVecWithMap(const ZVal* frame) {
```
evantypanski, line 178: The `auto` changes are one hint that these won't auto-unref, but it requires everyone remembering that one type won't auto-unref and the other will. Maybe update the comments on these and say that these are indeed not `ZValElement` so you need to explicitly call `Unref`?

it could also just be on `zvec` but that seems far away

awelzel: > everyone remembering that one type won't auto-unref and the other will

Yeah, `std::optional<ZVal>` implies manual memory management but only if it actually contains a "managed" value (so it's pretty tricky). I'm pretty sure that's just assumed to be known :-/ I actually don't know what to comment here - I really just wanted a non-auto type on the signature, because even looking at the implementation there's just `zvec` and one still doesn't know the type :-)

evantypanski: Fair enough, there are a few places that use a similar pattern with `std::optional<ZVal>`. It's probably fine as-is

it still feels a bit icky but it's already ingrained in the code I think so fine for this

vpax: Keep in mind that the only caller to that is ZAM instructions. If you're tinkering with those, you'd darn well better understand the low-level specifics. So I wouldn't worry about adding a warning, though go ahead if you want.

PR #4886, opened 2025-10-13
Add callback based record fields, deprecate Analyzer::UpdateConnVal()

### LINKED ISSUE #4214: Nesting of analyzers may lead to stale argument values

awelzel:
Attempt to fix #4214 and #4786.

Instead of propagating up-to-date information to RecordVal instances, the proposal here allows to replace record fields with callbacks that are invoked only when script land needs access the corresponding field.

Inspiration taken from Google's V8 engine where one can set callbacks for JavaScript properties on objects.

Oh, and this introduces `&volatile` to mark record fields that the core may change behind the scenes. It doesn't actually do anything, but was thinking assigning a callback for a record field could be guarded by a `&volatile` check. I also think `&volatile` is nice for documentation purposes and ZAM can then use that to check if an individual record field needs a callback check or not. Not too attached to the name, but volatile seemed fitting.

### DIFF
--- a/src/script_opt/ZAM/OPs/macros.op
+++ b/src/script_opt/ZAM/OPs/macros.op
@@ -103,7 +103,7 @@ macro LoadFieldWithDefault(target, r, f) {
 		}
 	else if ( zvel.HoldsFieldCallback() )
 		{
-		target = zvel.FieldCallback()->Invoke(*r, f);
+		target = zvel.FieldCallback()->Invoke(*r, RecordField(r, f));
 		}
 	else
 		{
@@ -130,7 +130,7 @@ macro LoadField(target, r, f) {
 		}
 	else if ( zvel.HoldsFieldCallback() )
 		{
-		target = zvel.FieldCallback()->Invoke(*r, f);
+		target = zvel.FieldCallback()->Invoke(*r, RecordField(r, f));
 		}
 	else
 		{


awelzel:
Okay. This isn't ready for merging and benchmarking indicates it's a bit expensive currently, but I have some ideas for making it faster.

First, however, wanted to see if people think this is totally crazy :-) I actually think it's a pretty powerful concept. As mentioned, that's what V8 allows with JS objects, too.

In summary, this PR adds support to RecordVals to hold callbacks for individual fields. These callbacks then allow a RecordVal to be backed by a separate object within the core (e.g. ConnSize analyzer or Connection). Whenever a script access to a record field happens, e.g., `c$history`, the callback is invoked and the implementor can lazily produce a value (or in the case of un-managed/atomic types, wrap the actual value in a `ZVal`).

I think the main slow-down I observe comes from `GetField()` not being inlined anymore. I haven't really stared. The other is the new `is_callback` flag - maybe `is_set` / `is_callback` could become a single byte enum. For now, more curious about the conceptual high-level idea here, rather than low-level optimizations.

The ZValElement is pretty convoluted now and I'm not very happy with the names. Would think we should put this anyhow into `zeek::detail` so we can revisit in the future without worrying about API breaks.

@rsmmr - can I excite you to take a look? The main driver is for stale `num_pkts` and `num_bytes_ip` in `endpoint` data from #4214, but also enables a way to have other analyzers (or any plugin really) expose up-to-date data directly on record fields on access. Such fields must be annotated with `&volatile`.

src/Conn.h
```diff
@@ -57,6 +58,63 @@ static inline int addr_port_canon_lt(const IPAddr& addr1, uint32_t p1, const IPA
...
+ * Object backing the script-layer connection record data type for the fields
+ *
+ * - duration
+ * - history
+ *
+ * An instance of this class is embededd into the Connection itself.
+ */
+class ConnectionRecordValCallback : public detail::RecordFieldCallback {
+public:
+    ConnectionRecordValCallback() = default;
```
timwoj, line 75: It'd be good for this to have a overridden destructor, just for future sanity.

src/Val.h
```diff
@@ -1163,18 +1173,42 @@ class ZValElement {
...
+          tag(tag),
+          data(zval.value_or(ZVal())) {}
+
     /**
      * Copy constructor.
      */
-    ZValElement(const ZValElement& o) : is_set(o.is_set), is_managed(o.is_managed), tag(o.tag), zval(o.zval) {
-        if ( is_set && is_managed )
-            Ref(zval.ManagedVal());
+    ZValElement(const ZValElement& o) : state(o.state), is_managed(o.is_managed), tag(o.tag) {
```
timwoj, line 1192: This and the assignment operator different pretty wildly. Is there a reason for that?

src/Val.h
```diff
@@ -1207,17 +1244,21 @@ class ZValElement {
         if ( this == &o )
```
timwoj, line 1248: If there's a move operator there probably should be a move constructor too.

src/Val.h
```diff
@@ -1249,34 +1290,71 @@ class ZValElement {
...
+        return *this;
+    }
+
+
+    /**
+     * Operator bool only checks for HoldsZVal(). It should
+     * be used to protect -> or * operators. These don't
+     * work with callbacks!
+     */
+    operator bool() const noexcept { return HoldsZVal(); }
```
timwoj: Is this necessary if you already have the assert() checking for callbacks in the other two operators?

src/Val.h
```diff
@@ -1317,6 +1409,31 @@ class RecordVal final : public Val, public notifier::detail::Modifiable {
...
     void Assign(int field, ValPtr new_val);
 
+    /**
+     * Assign a callback to a record field.
+     *
+     * @param field The Field index to assign
+     * @param cb The ZValCallback instance to assign.
+     */
+    void AssignCallback(int field, detail::RecordFieldCallback* cb) {
+#ifdef DEBUG
```
timwoj: Is this debug-only temporarily? I feel like that check should be live all the time.

src/RecordFieldCallback.h
```diff
@@ -0,0 +1,65 @@
+// See the file "COPYING" in the main distribution directory for copyright.
```
timwoj, line 1: Nit: I'd love to try to keep files named the same as the class, if the file only defines a single class.

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -89,6 +89,56 @@ using TapAnalyzerPtr = std::unique_ptr<TapAnalyzer>;
...
+ * Call AssignCallbacks() the first time UpdateConnVal() is invoked on
+ * a SessionAdapter. Use RemoveCallbacks() during Done().
+ *
+ * SessonAdapter implementations are required to implement
+ * GetEndpointSize(bool is_orig) and GetEndpointState(bool is_orig) to
+ * query the most recent values.
+ */
+class EndpointRecordValCallback : zeek::detail::RecordFieldCallback {
+public:
+    EndpointRecordValCallback() = default;
```
timwoj, line 93: Same re: destructor.

rsmmr:
I took another high-level look and have to say I personally don't like this approach that much. It's adding too much complexity for my taste just to address a couple of special-cases (important special cases, I know, but still). I'd be more excited if we could generalize this to add arbitrary computed fields to records, including from script-land. I realize that's quite a large ask, and I don't even know if that's possible (and/or: would solve the original problem, which needs the connection access). But as it stands, I think this is more of a one-off, which, for me, doesn't seem to justify the complexity it's adding.

awelzel:
> I think this is more of a one-off, which, for me, doesn't seem to justify the complexity it's adding.

I'd argue pretty much the same way about `UpdateConnVal()` being a one-off (i.e. it doesn't translate to file or packet analyzers or anything non-analyzer'ish. With record field callbacks we open a few possibilities). I do agree callback based record fields are a bit mind-twisting, but as discussed other scripting language implementations provide them them, too. Further, one upside might be slightly improve performance as fields aren't eagerly populated anymore. Optimally, if `c$history` is never accessed, the StringVal should never be allocated. Today, we allocated a new `c$history` StringVal whenever an event for the connection in question is raised and the history has changed. Regardless of whether anyone reads `c$history` (`http_request()` or other protocol events). For conn.log logging, only `connection_state_remove()` needs history IIRC.

I have multiple times considered stopping while working on this, but I don't see a better solution for #4214 and particularly for #4786. But I'm obviously okay to close this, but someone else might need to stare at those bugs (I might be stuck in a corner, too).

I reached out to Vern and will chat with Tim later to get their thoughts.

src/Val.h
awelzel: Thanks! Updated.

src/Val.h
awelzel: Added.

src/Val.h
awelzel: For the user if they treat it like an optional, but your'e right this maybe should better go. I'll stare at ZAM and then decide.

awelzel: I've dropped it now. Thanks!

NEWS
```diff
@@ -162,6 +166,27 @@ New Functionality
 Changed Functionality
 ---------------------
 
+- Fields representing state information about connections on ``connection`` and
```
vpax: on -> in

NEWS
```diff
@@ -162,6 +166,27 @@ New Functionality
...
+  ``endpoint`` record values now always yield the most up-to-date value.
+
+  In previously versions of Zeek, fields like ``connection$history``, ``endpoint$num_pkts``
```
vpax: previously -> previous

NEWS
```diff
@@ -162,6 +166,27 @@ New Functionality
...
+  or ``endpoint$num_bytes_ip`` could potentially hold stale values. The most prominent case
+  is that in a ``new_connection(c: connection)`` event handler, zero packets were reported
```
vpax: is -> was

NEWS
```diff
@@ -162,6 +166,27 @@ New Functionality
...
+  for either endpoint, while clearly there had to have been a packet for the connection to
+  be created in the first place. Previous workarounds for this issue included explicitly
+  calling ``lookup_connection()`` which would refresh the record's fields.
+
+  In a similar vain, stashing ``connection`` objects into a vector or container and
```
vpax: vain -> vein

NEWS
```diff
@@ -162,6 +166,27 @@ New Functionality
...
+  regularly reading ``c$duration`` or ``c$history`` of these connections within a
+  scheduled event would likely yield stale values, even if Zeek had observed further
```
vpax, line 181: I don't quite get this. Since the objects are shallow copies, wouldn't they get updated?

NEWS
```diff
@@ -202,6 +227,32 @@ Removed Functionality
...
+  This is a pretty advanced use-case. Usually, any data from analyzers should
+  either be queried via custom bif functions, or as parameters of analyzer
+  specific events.
+
+  The motivating reason for this is that ``Analyzer::UpdateConnVal()`` was push-based:
+  For any event that has a connection parameter in its signature, analyzers conventionally
+  call ``GetVal()`` on a ``Connection`` instance to receive its script-level connection value.
+  Every ``GetVal()`` call recursively invoked ``UpdateConnVal()`` on every analyzer
+  in the connection's analyzer tree. This allowed analyzers to "push" their latest values
+  to the connection or endpoint records. However, there may have never been a need
```
vpax: The "However ..." sentence construction is confusing, in particular "never been a need".

doc/script-reference/attributes.rst
```diff
@@ -789,3 +789,22 @@ groups are enabled.
...
+
+.. zeek:attr:: &volatile
+
+&volatile
+---------
+
+.. versionadded:: 8.1
+
+Record fields with this attribute reflect state that is held within the core.
+These fields may change their value without a user assigning a value.
```
vpax: "assigning a value" -> "assigning to them"

doc/script-reference/attributes.rst
```diff
@@ -789,3 +789,22 @@ groups are enabled.
...
+
+The current use is to annotate fields of :zeek:see:`connection` or
+:zeek:see:`endpoint` that have always behaved this way.
+
+Assigning a value to a volatile field stores the provided value and decouples
+the field from being updated by the core. This can be useful when instantiating
+records that aren't backed by real connections. It also avoids surprises where
+after assigning a field, Zeek's core would re-assign to its internal notion as
+has happened in Zeek 8.0 and before.
```
vpax: has happened -> happened

src/Conn.cc
```diff
@@ -234,21 +319,30 @@ const RecordValPtr& Connection::GetVal() {
...
+        // the session adapter know about it, once.
+        //
+        // AddChildAnalyzer() will do it for new child
+        // analyzers added when conn_val exists.
+        //
+        // If the adapter isn't set yet, SetSessionAdapter()
+        // will do the job, too.
+        if ( adapter ) {
+            analyzer::for_each(*adapter, [this](analyzer::Analyzer& a) {
+                // Call InitConnVal() on all analyzer in this connection's analyzer tree.
```
vpax: all analyzer -> all analyzers
OTOH, this comment seems obvious so you could just remove it

src/Conn.h
```diff
@@ -57,6 +58,61 @@ static inline int addr_port_canon_lt(const IPAddr& addr1, uint32_t p1, const IPA
...
+    ~ConnectionRecordValCallback() override = default;
+
+    // Avoid copying or assigning instances.
+    ConnectionRecordValCallback(ConnectionRecordValCallback&& o) = delete;
+    ConnectionRecordValCallback(ConnectionRecordValCallback& o) = delete;
+    ConnectionRecordValCallback& operator=(ConnectionRecordValCallback& o) = delete;
+
+    /**
+     * Called when the Connection's record_val is creatd for the first time
```
vpax: creatd -> created

src/Conn.h
```diff
@@ -138,6 +194,12 @@ class Connection final : public session::Session {
      */
     const RecordValPtr& GetVal() override;
 
+    /**
+     * Returns a raw pointer to this connection's record, or a nullptr if
+     * the \a conn_val member hasn't been initialized yet.
```
vpax, line 199: It would be helpful to mention why this is needed

NEWS
awelzel: The crux here is that the respective script-layer connection record only gets updated if something in the core calls `GetVal()` on the underlying connection. That mostly only happens if there's an event to be raised for a given connection by an analyzer.

As a basic demonstration of it not working currently, the script attached tracks a single connection, printing its packet rate per second via a scheduled event. Running this against a pcap with a singe HTTP download that took about 15 seconds, one gets the following output with Zeek 8.0. The packet numbers on the endpoint are stale. While the HTTP response is delivered, the HTTP analyzer (nor any other analyzer) raises an event for this connection and so the corresponding record isn't ever actually refreshed via  `GetConnVal() -> UpdateConnVal()`.
```
$ zeek -C -r 100mb-39650.pcap LogAscii::use_json=T ./packets-per-second.zeek 
12:22:08, 1.01s 4.9 pkts/s 16.9 avgpkts/s
12:22:10, 1.02s 0.0 pkts/s 16.9 avgpkts/s
12:22:11, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:12, 1.02s 0.0 pkts/s 16.9 avgpkts/s
12:22:13, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:14, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:15, 1.11s 0.0 pkts/s 16.9 avgpkts/s
12:22:16, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:17, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:18, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:19, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:20, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:21, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:22, 1.00s 0.0 pkts/s 16.9 avgpkts/s
12:22:22, 0.80s 107953.3 pkts/s 5796.3 avgpkts/s
```

With the changes in this branch instead one gets the following output which should need no explanation - it's just what one would expect I guess.
```
$ taskset -c 0,1,2,3 /usr/bin/time zeek -C -r 100mb-39650.pcap LogAscii::use_json=T ./packets-per-second.zeek
12:22:08, 1.01s 262.2 pkts/s 263.2 avgpkts/s
12:22:10, 1.02s 525.9 pkts/s 395.2 avgpkts/s
12:22:11, 1.00s 3408.6 pkts/s 1386.7 avgpkts/s
12:22:12, 1.02s 3031.3 pkts/s 1800.0 avgpkts/s
12:22:13, 1.00s 4559.0 pkts/s 2345.4 avgpkts/s
12:22:14, 1.00s 4129.6 pkts/s 2639.8 avgpkts/s
12:22:15, 1.11s 3376.1 pkts/s 2753.4 avgpkts/s
12:22:16, 1.00s 5349.7 pkts/s 3071.6 avgpkts/s
12:22:17, 1.00s 9370.6 pkts/s 3758.9 avgpkts/s
12:22:18, 1.00s 8562.5 pkts/s 4232.8 avgpkts/s
12:22:19, 1.00s 10360.8 pkts/s 4782.9 avgpkts/s
12:22:20, 1.00s 8399.6 pkts/s 5080.0 avgpkts/s
12:22:21, 1.00s 9033.9 pkts/s 5380.9 avgpkts/s
12:22:22, 1.00s 9062.9 pkts/s 5640.7 avgpkts/s
12:22:22, 0.80s 7634.0 pkts/s 5746.9 avgpkts/s
```

I haven't tracked down the final avg number difference in detail, but `5746.9` agrees with what you get when using the conn.log packet counts and duration. Zeek 8.0 supposedly reads a stale `$duration` value (which is also fixed in this branch).

Hope that helps.

EDIT:
> 12:22:08, 1.01s 262.2 pkts/s 263.2 avgpkts/s

I realized the first entry is a bit strange. The script excludes the syn-packet from the first interval, so that's why it's "off-by-one" compared to computing the avg.

---

```
global the_conn: connection;
global last_ts: time;
global last_pkts: count = 0;

event tick()
        {
        local td = interval_to_double(network_time() - last_ts);
        local pkts = the_conn$orig$num_pkts + the_conn$resp$num_pkts;
        print strftime("%H:%M:%S", network_time()), fmt("%.2fs %.1f pkts/s %.1f avgpkts/s", td, (pkts - last_pkts) / td, pkts / interval_to_double(the_conn$duration));
        last_pkts = pkts;
        last_ts = network_time();

        schedule 1sec { tick() };
        }

        {
        the_conn = c;
        last_ts = network_time();
        last_pkts = c$orig$num_pkts + c$resp$num_pkts;
        schedule 1sec { tick() };
        }

```

NEWS
awelzel: ouch.

NEWS
awelzel: Rewrote it a bit

> Analyzers did so unconditionally, regardless
>  of whether any event handlers read these values before the next update.

src/Conn.cc
awelzel: will drop.

src/Conn.h
awelzel: Added more context. This is useful for determining whether to "tell" a newly added child analyzer about the connection value. The existing `GetVal()` always allocates one and that shouldn't happen when "only" adding a new analyzer.

NEWS
vpax: Okay. I had read the text as saying something different, but upon re-reading I don't think my initial interpretation is implied by anything there, so I'm closing this

src/RecordFieldCallback.h
```diff
@@ -0,0 +1,66 @@
...
+     * @param rv The RecordVal instance for which the callback is invoked. The value should not be modified.
+     * @param field The field index for which the callback is invoked.
+     *
+     * @return Implementations are required to return a ZVal instance that matches the expected type. The caller owns
+     * the returned reference if this is a managed ZVal.
+     *
+     */
+    virtual ZVal Invoke(const RecordVal& rv, int field) const = 0;
+
+    // Food for thought: How about setter support? This could allow calling
```
vpax, line 58: Interesting thought. I wonder if this could allow some consolidation of various Event-engine-control globals. I guess the key is whether there are controls that need to happen dynamically rather than at start-up.
(Not a request for changes, so feel free to close.)

src/Val.cc
```diff
@@ -2984,20 +2985,46 @@ void RecordVal::Assign(int field, ValPtr new_val) {
...
+    auto& fv = record_val[field];
+    bool was_set = fv.HoldsZVal() || fv.HoldsFieldCallback();
+
+    fv.Reset();
 
     if ( was_set )
         Modified();
 }
 
+ValPtr RecordVal::GetFieldSlow(const RecordType& rt, ZValElement& fv, int field) const {
```
vpax, line 2997: This name seems a bit strange (the "slow" part is non-obvious). Off the top of my head, the best alternative that comes to mind is `GetFieldFullyResolved`.

src/Val.h
```diff
@@ -1135,6 +1136,16 @@ inline constexpr bool is_zeek_val_v = is_zeek_val<T>::value;
  */
 class ZValElement {
 public:
+    /**
+     * The state of a ZValElement.
+     */
+    enum class State : uint8_t {
+        NoInit,        // Not yet initialized. Used to catch usage errors.
+        Unset,         // Holds nothing but tag and is_managed are valid.
+        ZVal,          // Holds a ZVal in data.
```
vpax, line 1145: Might be good to use something like `HoldsZVal` here, just so there isn't visual confusion between this meaning of `ZVal` and the more common meaning

src/Val.h
```diff
@@ -1245,37 +1299,77 @@ class ZValElement {
      * Initialize a ZValElement using a TypeDecl assignment.
      *
      * This is used at record construction time to set the is_managed
-     * and tag fields properly.
+     * and tag fields properly and get the ZValElement into the unset
```
vpax, line 1302: Micro-nit: structures like "foo and bar and bletch" read better as "foo, bar and bletch". (I will also be okay with an Oxford comma there ;-).

src/Val.h
```diff
@@ -1135,6 +1136,16 @@ inline constexpr bool is_zeek_val_v = is_zeek_val<T>::value;
  */
 class ZValElement {
 public:
+    /**
+     * The state of a ZValElement.
+     */
+    enum class State : uint8_t {
+        NoInit,        // Not yet initialized. Used to catch usage errors.
```
vpax, line 1143: I couldn't quite tell from the following whether this state will ever actually occur in practice, or is just for debugging. It would be useful to explain here either way.

src/Val.h
```diff
@@ -1286,17 +1380,31 @@ class ZValElement {
...
-    bool is_set = false;
+    State state = State::NoInit;
     bool is_managed = false;
     TypeTag tag = TYPE_ERROR;
     // 5 bytes of padding here.
-    ZVal zval;
+    union {
+        ZVal zval;
+        detail::RecordFieldCallback* field_callback;
+    } data = {};
```
vpax, line 1407: What does `= {}` do for a union member variable? How does it know which element to pick? I would think it could just be left off.

src/ZVal.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "zeek/IntrusivePtr.h"
+#include "zeek/util-types.h"
```
vpax, line 8: Why is this needed given there are no other changes to the file?

src/analyzer/Analyzer.cc
```diff
@@ -360,9 +370,19 @@ bool Analyzer::AddChildAnalyzer(Analyzer* analyzer, bool init) {
...
-    if ( init )
+    if ( init ) {
         analyzer->Init();
 
+        // If the connection already has an adapter and a conn_val instance,
+        // also call InitConnVal() directly on the new analyzer. If either
+        // of these isn't yet the case, this will happen during GetVal()
+        // or SetSessionAdapter()
+        if ( Conn()->GetSessionAdapter() != nullptr ) {
+            if ( auto* conn_val = Conn()->RawVal(); conn_val != nullptr )
```
vpax: I didn't realize C++ supported this sort of convoluted `if` expression ... not sure it adds any clarity to just doing the elements step-by-step

src/analyzer/Analyzer.h
```diff
@@ -600,8 +613,27 @@ class Analyzer {
...
      */
+    [[deprecated("Remove in v9.1: Replaced with InitConnVal()")]]
     virtual void UpdateConnVal(RecordVal* conn_val);
 
+    /**
+     * Called once on an analyzer when either a connection's conn_val is
+     * instantiated for the first time, or right after the analyzer is
+     * initialized if the Connection's conn_val is available at that time.
+     *
+     * This is only interesting for analyzers that expose data on the
```
vpax: micro-nit: I think "expose data for" would be a bit more idiomatic than "expose data on"

src/analyzer/Analyzer.h
```diff
@@ -600,8 +613,27 @@ class Analyzer {
...
+     * connection or any nested records.
+     *
+     * The \a conn_val will live at least until the analyzer's Done()
+     * method is called. If an analyzer installs record field callbacks,
+     * these need to be cleared during Done(). Usually by keeping a pointer
```
vpax: Nit, there's a sentence fragment here. One fix: "... Done(). Usually" -> "... Done(), usually"

src/analyzer/protocol/conn-size/ConnSize.h
```diff
@@ -2,10 +2,71 @@
...
+    ZVal Invoke(const RecordVal& val, int field) const override;
+
+    static void InitPostScript();
+
+private:
+    ConnSize_Analyzer* analyzer = nullptr;
+
+    // Pointers to the endpoint records within connection.
+    RecordVal* orig_endp = nullptr;
+    RecordVal* resp_endp = nullptr;
```
vpax, line 63: Given that these are run-time pointers, it would be helpful to add a comment explaining why they're guaranteed to be valid if non-nil.

src/logging/Manager.cc
```diff
@@ -1655,15 +1655,20 @@ detail::LogRecord Manager::RecordToLogRecord(WriterInfo* info, Filter* filter, c
             auto* vr = val->AsRecord();
             const auto& field = vr->RawOptField(index);
 
-            if ( ! field.IsSet() ) {
+            if ( field.HoldsZVal() ) {
+                val = *field;
+                vt = cast_intrusive<RecordType>(vr->GetType())->GetFieldType(index).get();
```
vpax: Can't this be factored out to come after the if-else sequence

src/packet_analysis/protocol/icmp/ICMPSessionAdapter.cc
```diff
@@ -74,3 +46,13 @@ void ICMPSessionAdapter::Done() {
     SessionAdapter::Done();
     matcher_state.FinishEndpointMatcher();
 }
+
+zeek_uint_t ICMPSessionAdapter::GetEndpointSize(bool is_orig) const {
+    zeek_int_t size = is_orig ? request_len : reply_len;
+    return size < 0 ? 0 : size;
```
vpax, line 53: Worth a comment here explaining how the size could be negative

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -89,6 +78,50 @@ using TapAnalyzerPtr = std::unique_ptr<TapAnalyzer>;
...
+     */
+    void FlipRoles() { std::swap(orig_endp, resp_endp); }
+
+    static void InitPostScript();
+
+private:
+    const SessionAdapter* adapter = nullptr;
+    // Pointers to the endpoint records within connection.
+    RecordVal* orig_endp = nullptr;
+    RecordVal* resp_endp = nullptr;
```
vpax, line 121: (same meta-comment here)

src/packet_analysis/protocol/ip/SessionAdapter.h
```diff
@@ -202,14 +235,32 @@ class SessionAdapter : public analyzer::Analyzer {
                    SkipReason skip_reason = SkipReason::None) const;
 
     /**
-     * Overridden from parent class, calling UpdateConnVal() on TapAnalyzer instances, too.
+     * Override to install endpoint callbacks.
```
vpax: I find this comment confusing because the statement itself is an `override`. Does the comment mean "if you derive from this, then you should override the following"? (in which case clarify if this instance of the method should still be called by the overridden version) Or is it saying "the `override` here is so we can install endpoint callbacks?

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -1035,23 +1035,26 @@ void TCPSessionAdapter::FlipRoles() {
...
     // Call children's UpdateConnVal
     SessionAdapter::UpdateConnVal(conn_val);
 
     // Have to do packet_children ourselves.
     for ( Analyzer* a : packet_children )
         a->UpdateConnVal(conn_val);
+#pragma GCC diagnostic pop
+}
+
+// TCPSessionAdapter has its own set of packet children that aren't
```
vpax: "packet children" is a strange phrase, I'm not immediately sure what it means here

src/script_opt/ZAM/OPs/macros.op
```diff
@@ -70,9 +70,75 @@ macro TableIter(slot) (*tiv_ptr)[slot]
...
+	const auto& zvel = DirectOptField(r, f);
+	if ( zvel.HoldsZVal() )
+		{
+		target = *zvel;
+		if ( zvel.IsManaged() )
+			zeek::Ref(target.ManagedVal());
+		}
+	else if ( zvel.HoldsFieldCallback() )
+		{
+		target = zvel.FieldCallback()->Invoke(*r, f);
```
vpax: I think `f` here needs to be `RecordField(r, f)`

src/script_opt/ZAM/OPs/macros.op
```diff
@@ -70,9 +70,75 @@ macro TableIter(slot) (*tiv_ptr)[slot]
...
```
vpax: Same comment here re `RecordField`

src/script_opt/ZAM/ZBody.cc
```diff
@@ -2,6 +2,8 @@
 
 #include "zeek/script_opt/ZAM/ZBody.h"
 
+#include <type_traits>
```
vpax, line 5: Just for my own understanding - why is this needed, given there are no other code changes?

testing/btest/plugins/tap-analyzer-conn-val-plugin/src/Plugin.cc
vpax, line 1: Why is this (and companion files) deleted?

tools/gen-zam/Gen-ZAM.cc
```diff
@@ -1108,14 +1108,34 @@ void ZAM_OpTemplate::GenAssignOpValCore(const OCVec& oc, const string& orig_eval
...
-        eval += "zeek::Ref(rhs);\n";
-        eval += "Unref($$.ManagedVal());\n";
-        eval += "$$ = ZVal(rhs);\n";
+    if ( AssignValIsZVal() ) {
+        // If assign-zval was used, v owns the reference and we
+        // can just assign it directly, but need to ensure any
+        // already set managed value is properly unref'ed.
+        //
+        // Add some C++ code to statically assert it's actually a ZVal at
+        // compile time.
```
vpax, line 1117: Why is this needed?

vpax: Ah I think you answer that below. Useful to hoist the comment to here

tools/gen-zam/Gen-ZAM.cc
vpax, line 1: I've lost track of whether you flagged that you confirmed things work for `btest -a zam` ... ?

tools/gen-zam/Gen-ZAM.h
```diff
@@ -945,6 +955,7 @@ class ZAMGen {
 
     void IndentUp() { ++indent_level; }
     void IndentDown() { --indent_level; }
+    std::string IndentString() { return std::string(indent_level, '\t'); }
```
vpax, line 958: If you care about this sort of tidiness in the generated code (which, as you can no doubt tell, I only sort of do), then you could consider using this other places too beyond just where you introduced it

NEWS
vpax: Hmmm I only half get this, but am okay with believing you ;-)

vpax: Hmmm#2 I keep banging on `Resolve conversation` but it's not showing it as resolved. The UI briefly switches to `Resolving conversation ...` but then doesn't show it as such.

vpax:
I'm done with my review, LGTM other than a bunch of clarity nits and a couple of low-level questions about the new ZAM macros (and asking for confirmation that this has been tested using `-a zam`). 

Two related general thoughts. At one point you framed that perhaps `&volatile` could be extended to apply to record fields that might change during event draining (such as by allowing script functions to serve as the backing for a field). For ZAM optimization, it would be valuable to distinguish those sorts of `&volatile` - which need to always be evaluate via the calls, no reuse of previous values allowed - versus the current more restricted usage, which has the guarantee that values don't change during event draining, so it's fine to reuse a previously accessed value via the temporary that holds it. That difference could be directly specified by using two attributes rather than one, with one attribute meaning guaranteed-not-to-change-during-event-draining and the other being could-change-at-any-time. Alternatively, this could be dealt with in a spirit similar to `src/script_opt/FuncInfo.cc`, where we maintain side information about which `&volatile` fields are known to be invariant during event draining.

The second related thought regards the comment _Record fields with this attribute reflect state that is held within the core_, which should be broadened if down the road we might want to allow the more general notion of record field volatility (that said, not sure we should though!).

awelzel:
> I'm done with my review, LGTM other than a bunch of clarity nits and a couple of low-level questions about the new ZAM macros (and asking for confirmation that this has been tested using -a zam).

Thanks. Yes, this was tested with ASAN+ZAM locally as well as in CI.

I have already fixed some of the comments / sentence structures you pointed out earlier, but haven't seen any comments regarding ZAM in email or in the UI here. @vpax  Is it possible the review is still queued on your end waiting for a click on "submit"? :-)

> versus the current more restricted usage, which has the guarantee that values don't change during event draining, so it's fine to reuse a previously accessed value via the temporary that holds it

The current implementation actually assumes `&volatile` may change at any time, but realistically the callbacks installed by  ConnSize/Conn will yield the same value during one event loop iteration and only yield a different value after the next packet passed through the analyzer tree. For now, I'd just keep the model that `&volatile` may change at any time, even if that means invoking the callback unnecessarily. Maybe a new `&packet-volatile` that more strictly indicates a field needs refreshing only after the next packet has been processed (name/semantics and whether that's reasonable tbd, definitely).

At the same time, the first (and probably less involved?) optimization I could see for ZAM is that a `non-&volatile` field (which is static type info) does not need to check for the field callback - e.g. the new `LoadField()` macros check for the callback even when there's no `&volatile` on it. That means optimizing for non-&volatile fields (of which there are many more) rather than focus on caching `&volatile`. But that should also be profiling guided - there might just be bigger fish :-)

> general notion of record field volatility (that said, not sure we should though!).

Me neither. I see this push/change currently more of ensuring to effectively fix the stale connection record fields in event handlers (which do upset me, because it's a fixable quirk in an area where there should be no surprises). It does put a few things in place in the core that may allow for more interesting and crazy constructs, but also staying conscious that there's  sharp edges and we probably shouldn't overdo it :-)

vpax:
> haven't seen any comments regarding ZAM in email or in the UI here

See for example line 106 for src/script_opt/ZAM/OPs/macros.op (and I think one more in that file). I entered it via the Web UI and see it now when I revisit.

vpax:
> The current implementation actually assumes `&volatile` may change at any time

Yes I realize that, but I'd like optimization to be able to leverage won't-change-during-this-event-drain. And now that you mention it, I don't believe you added anything to the Common Subexpression Elimination logic to rule out reusing a field reference to a `&volatile` field. (BTW, I would rather put the burden of a clunky attribute name on can-change-at-any-time rather than can-change-between-packets, so maybe `&fully-volatile` for the latter - especially since we don't have any of them now.)

> ... does not need to check for the field callback

Actually, better would be to have ZAM instructions specifically for callback fields, and the logic to select those over the regular ones only when needed. If you're up for adding that distinction, I can help guide you through where stuff should be changed.

vpax (commented):
High-level comments are in the PR thread

ckreibich (commented):
I have finally made it through the whole PR. I apologize for taking this long, because in doing so I allowed it to get this large without us capturing our thinking somewhere first (unless I'm overlooking it?).

I think this PR brings up great questions around how to handle in-core vs scriptland state, and my immediate reaction is that I like the concepts — the notion of field access callbacks, the idea of `&volatile` — but I'm not sure I like their current realization.

The approach strikes me as fairly specific to low-level analyzer and `ConnSize` requirements and as C++-heavy overall (e.g. by requiring `SessionAdapter` to implement `GetEndpointSize()` and `GetEndpointState()`, or via the `RecordFieldCallback` class approach). Without `&volatile` it would be completely invisible to the script layer.

Perhaps that's a good thing, but if the main goal of the PR is to ensure state is up to date when it gets accessed in scriptland, couldn't we have the notion of callbacks directly in scriptland on records, or record fields, and ensure state is up to date via such callbacks (that could then call into the core as needed via BIFs or whatever, , and possibly enable other high-level uses too)?

For the `&volatile` concept I wondered about
> assigning a value to a volatile field stores the provided value and decouples the field from being updated by the core.

That's pretty unusual, no? Wouldn't it be better if a volatile field (if it means it's managed by the core) _cannot_ be assigned to in the script layer, for example?

I made a push on this PR this weekend to see whether this might make 8.1 still, but I think that would be rushed — let's give this a bit more time to bake.

vpax:
> couldn't we have the notion of callbacks directly in scriptland on records, or record fields, and ensure state is up to date via such callbacks

Along these lines, if you haven't yet, see what you think of my discussion of this notion and why it would then be useful to have two different flavors of callback, one that allows Common Subexpression Elimination for repeated accesses to the same field, and the other that prohibits it.

(Also, regarding potentially targeting 8.1, I do think this needs more work for script optimization to do well with it.)

awelzel:
> I have finally made it through the whole PR. I apologize for taking this long, because in doing so I allowed it to get this large without us capturing our thinking somewhere first (unless I'm overlooking it?).

Fair. I'll update the `connection` and `endpoint` documentation to mention the bug and leave bread-crumbs and references to the prior issues as well as prior PRs and this PR for background.

> the notion of field access callbacks, the idea of &volatile — but I'm not sure I like their current realization.

My view on this PR is it's just a non-trivial bug fix that can be iterated over on in the next releases if it didn't quite hit the right spot / abstractions. But I might be too naive how others think about this or maybe this development approach makes others too nervous. I mentioned various times that nothing should be set in stone with this change here (towards 9.0 at least), that it fixes bugs that I find pretty rough and important and it seems to perform okay. I'm not too attached to the implementation in the end.

> couldn't we have the notion of callbacks directly in scriptland on records, or record fields, and ensure state is up to date via such callbacks (that could then call into the core as needed via BIFs or whatever, , and possibly enable other high-level uses too)?

Sure. Again, to me the scope is a bug fix that pushes ideas for future features. Not implementing those features.

I'd also prefer we look at what other script interpreters do here rather than coming up with our own ideas.

>  let's give this a bit more time to bake.

Yep, all good :+1:

src/ZVal.h
awelzel: util-types.h declares `zeek_int_t` and `zeek_unit_t`. Likely some other header will have pulled these in, but using `ZVal.h` standalone fails because these types are unknown.

src/analyzer/Analyzer.cc
awelzel: No strong opinion here. I mostly picked this up from Spicy's use here and there and like its compactness, but expanding is fine for me, too.

src/Val.h
awelzel: It picks the first one AFAICT. If left off it produces this:

```
zeek/Val.h:1404:11: note: ‘zeek::ZValElement::<unnamed union>::<constructor>()’ is implicitly deleted because the default definition would be ill-formed:                                                                                                                                                                                         1404 |     union {                                                                                                                                                                            
      |           ^                                                                                                                                                                                                                                                                                                                                                                           zeek/Val.h:1405:14: error: union member ‘zeek::ZValElement::<unnamed union>::zval’ with non-trivial ‘zeek::ZVal::ZVal()’                                                                                                                                                                                                                          1405 |         ZVal zval;                   
```

I had `ZVal zval = {}` within the union, but IIRC that failed with some systems in CI, so this is just what I ended up with.

src/packet_analysis/protocol/ip/SessionAdapter.h
awelzel: The latter, thanks. Changed.

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
awelzel: Hmm, TCPSessionAdapter tracks "packet children" separately from the generic analyzer tree and has a dedicated API related to these. So this is just referring to an existing member name / terminology. I'm neutral on the name...

https://github.com/zeek/zeek/blob/11a2eeb7d053430e4f7ac8aba27004332c944a8e/src/packet_analysis/protocol/tcp/TCPSessionAdapter.h#L35-L37

This is because once reassembly is done, a TCP analyzer tree doesn't receive `DeliverPacket()` anymore and "packet children" is the API to still get access to per-packet data:

https://github.com/zeek/zeek/blob/11a2eeb7d053430e4f7ac8aba27004332c944a8e/src/packet_analysis/protocol/tcp/TCP.cc#L137-L140

https://github.com/zeek/zeek/blob/11a2eeb7d053430e4f7ac8aba27004332c944a8e/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc#L988-L995

src/script_opt/ZAM/ZBody.cc
awelzel: For the `std::is_same_v` template added to ZAM-Eval-Defs.h via `assign-val` / `assign-zval`.

testing/btest/plugins/tap-analyzer-conn-val-plugin/src/Plugin.cc
awelzel: I proposed to remove the `UpdateConnVal()` API from `TapAnalyzer` in a separate commit and the test became irrelevant.

src/script_opt/ZAM/OPs/macros.op
awelzel: Yeah, it appears that just ends up being `f` after expansion, but its used elsewhere that way.

src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
vpax: My concern is simply the phrase "packet children". Changing it to "packet analyzer children" would resolve it.

src/script_opt/ZAM/OPs/macros.op
vpax: But it won't in the future when the instructions are reused in other contexts. That was the point in introducing the macros.

src/analyzer/Analyzer.cc
bbannier: The more common way to write this in C++ would probably be to leave off the redundant condition since a `nullptr` is already falsy.

```suggestion
            if ( auto* conn_val = Conn()->RawVal() )
```

awelzel:
I'm closing this PR due to feedback about the complexity it introduces. I'm not opposed to re-opening, but it might be better to have someone else take a separate stab at https://github.com/zeek/zeek/issues/4214 and maybe https://github.com/zeek/zeek/issues/4786 for simpler solutions.

PR #4887, opened 2025-10-13
Logging: Add BiF to flush all logs

J-Gras:
Let me know if you prefer going with the `Log::__func` pattern.


timwoj:
The pattern you have is fine. We typically only use the `__` pattern for BIFs that have an associated script-level function, where the BIF should only be called from that function. It's more of a "this is a private method" indicator.

PR #4924, opened 2025-10-14
Deprecate DNS_Mgr T_* constants, replace with standard ns_t_* constants

### LINKED ISSUE #4575: Remove use of `T_*` constants from DNS_Mgr, switch to more-standard `ns_t_*` constants
From https://github.com/zeek/zeek/pull/4567#discussion_r2143383729:

> It's not really a c-ares header, it's just where the constants are defined since we're using it. On a lot of other systems they could include arpa/nameser.h, but some platforms (such as macOS) they also have to set an environment variable to get arpa/nameser_compat.h included that defines all of the T_* constants. At some point, the world moved over to ns_t_* naming for all of those constants. I don't know when though, and which of our platforms support/don't support it. Looking at what I have available here (macOS, ubuntu 24, FreeBSD 14), the all support the ns_t_* naming via arpa/nameser.h. Maybe the better solution is just to switch to that as well. Granted that would break anyone using the constant at the moment, but we could deprecate it and tell people to use the other one.

timwoj:
This requires https://github.com/zeek/libunistd/pull/9 as well.

Fixes #4575

### DIFF
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -30,7 +30,6 @@ using ztd::out_ptr::out_ptr;
 
 #include <ares.h>
 #include <ares_dns.h>
-#include <ares_nameser.h>
 
 #include "zeek/DNS_Mapping.h"
 #include "zeek/Event.h"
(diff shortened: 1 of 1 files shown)


src/DNS_Mgr.cc
```diff
@@ -60,63 +60,62 @@ constexpr int MAX_UDP_BUFFER_SIZE = 4096;
 // This unfortunately doesn't exist in c-ares, even though it seems rather useful.
 static const char* request_type_string(int request_type) {
```
awelzel: This function has two callers, one of them constructing a fake result for testing, the other results in a FatalError(). Seems we could do with just the numeric value in both cases and remove the switch case?

Or... we have the same list here and I wouldn't be all upset looking at that, but that might not be available in bare mode:
https://github.com/zeek/zeek/blob/cdea7512948ba5b07585ef4220690cb57b26e438/scripts/base/protocols/dns/consts.zeek#L13

Anyway, I think I'd be actually fine just rendering the numeric value for the two cases where it matters and ditching the switch? Not too strong, but it'd drop a few lines.

PR #4846, opened 2025-09-23
Move the zeek-docs repo into the main repo

timwoj:
This PR moves all of the docs and docs-generation configuration into the main Zeek repository. I've been working with the people at ReadTheDocs to iron out a couple of wrinkles in the build process there, and there is a test build that finally completed at https://app.readthedocs.org/projects/zeek-main-repo/builds/29695350/. The built docs are at https://zeek-main-repo.readthedocs.io/en/topic-timw-move-zeek-docs-to-main-repo/.

We should probably discuss if all of Github issues in the zeek-docs repo should move over into the main repo when this lands.

PR #4927, opened 2025-10-14
Migrate docs for the release/8.0 branch

timwoj:
This is the equivalent of https://github.com/zeek/zeek/pull/4846 for the release/8.0 branch.


doc/devel/spicy/autogen/tftp-no-accept.spicy
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) 2021 by the Zeek Project. See LICENSE for details.
```
evantypanski: was this file supposed to be removed in the original PR?

evantypanski (approved):

doc/devel/spicy/autogen/tftp-no-accept.spicy
timwoj: Yeah it was. It should be removed here too.

PR #4929, opened 2025-10-14
Send correct value for `auth` from RFB analyzer

ynadji:
The RFB analyzer raises the `rfb_auth_result` event, which then sets `c$rfb$auth` to its negation. This means `c$rfb$auth` has the correct value according to [the docs](https://docs.zeek.org/en/master/scripts/base/bif/plugins/Zeek_RFB.events.bif.zeek.html#id-rfb_auth_result) while the event will always have `result` be inverted from what is claimed by the docs. This changes `proc_handle_security_result` to send the correct value to the event in the first place.

### DIFF
--- a/src/analyzer/protocol/rfb/events.bif
+++ b/src/analyzer/protocol/rfb/events.bif
@@ -9,8 +9,17 @@ event rfb_authentication_type%(c: connection, authtype: count%);
 ##
 ## c: The connection record for the underlying transport-layer session/flow.
 ##
+## result: whether or not authentication was successful (false means success, true means failure)
+##
+## .. zeek:see:: rfb_authentication_result
+event rfb_auth_result%(c: connection, result: bool%) &deprecated="Remove in v9.1. Use rfb_authentication_result which has the correct value for result.";
+
+## Generated for RFB event authentication result message
+##
+## c: The connection record for the underlying transport-layer session/flow.
+##
 ## result: whether or not authentication was successful
-event rfb_auth_result%(c: connection, result: bool%);
+event rfb_authentication_result%(c: connection, result: bool%);
 
 ## Generated for RFB event share flag messages
 ##
--- a/src/analyzer/protocol/rfb/rfb-analyzer.pac
+++ b/src/analyzer/protocol/rfb/rfb-analyzer.pac
@@ -61,8 +61,12 @@ refine flow RFB_Flow += {
 
 	function proc_handle_security_result(result : uint32) : bool
 		%{
+		// Remove in v9.1: rfb_auth_result should have been removed.
 		if ( rfb_auth_result )
 			zeek::BifEvent::enqueue_rfb_auth_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), result);
+
+		if ( rfb_authentication_result )
+			zeek::BifEvent::enqueue_rfb_authentication_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), result == 0);
 		return true;
 		%}
 };
--- a/scripts/base/protocols/rfb/main.zeek
+++ b/scripts/base/protocols/rfb/main.zeek
@@ -149,9 +149,9 @@ event rfb_server_parameters(c: connection, name: string, width: count, height: c
 	write_log(c);
 	}
 
-event rfb_auth_result(c: connection, result: bool) &priority=5
+event rfb_authentication_result(c: connection, result: bool) &priority=5
 	{
-	c$rfb$auth = !result;
+	c$rfb$auth = result;
 	}
 
 event rfb_share_flag(c: connection, flag: bool) &priority=5


src/analyzer/protocol/rfb/rfb-analyzer.pac
```diff
@@ -62,7 +62,7 @@ refine flow RFB_Flow += {
 	function proc_handle_security_result(result : uint32) : bool
 		%{
 		if ( rfb_auth_result )
-			zeek::BifEvent::enqueue_rfb_auth_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), result);
```
awelzel, line 65: I think I'd be okay if we note it in the NEWS. The log stays consistent, it's a fix for the event parameter. So approving, but would be good to have a second approval. Thanks!

0xxon:
Ouch, this one is tricky. If someone already uses the event downstream, it's really easy to miss that we changed this.

I'm not really opposed - but I am wondering a bit if it might make sense to keep the current behaviour and clarify the documentation of the event instead.

awelzel:
> Ouch, this one is tricky. If someone already uses the event downstream, it's really easy to miss that we changed this.

Yes, but also, if they wrote code without testing based solely on the documentation `result: whether or not authentication was successful`, the code is doing the wrong thing and changing would maybe fix it. 

Alternatively, introducing a fixed `rfb_authentication_result()` and deprecating the existing one? There is `rfb_authentication_type()` already. A true/false bool that has the wrong value even if documented is a bit embarrassing :-) @ynadji - would you be up for that?

0xxon:
Introducing a new event certainly feels like the cleanest alternative. But is a bit more work.

ynadji:
Okay that sounds good then. I'll update the MR in the next couple of days.

ynadji:
Updated! I guessed at the removal version for the `&deprecated` message on `rfb_auth_result`. I also wasn't sure how I should flag the eventual removal of the first `enqueue` call in [rfb-analyzer.pac](https://github.com/zeek/zeek/pull/4929/files#diff-8f6d90ef53da8d34a54f8c2a8df92e9f726ad37106e35c95dd5991dd3b26f7fd) so any specific feedback there would be helpful. Do I need to manually update `./doc/scripts/base/bif/plugins/Zeek_RFB.events.bif.zeek.rst` as well or is that something that's handled by a script? I tried `$ make doc` on a whim but it failed for me.

ah, will have to fix the tests too.

src/analyzer/protocol/rfb/rfb-analyzer.pac
```diff
@@ -62,7 +62,10 @@ refine flow RFB_Flow += {
 	function proc_handle_security_result(result : uint32) : bool
 		%{
 		if ( rfb_auth_result )
+			{
 			zeek::BifEvent::enqueue_rfb_auth_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), result);
+			zeek::BifEvent::enqueue_rfb_authentication_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), ! result);
```
awelzel: There's a some tests failing with the latest change because line 64 is checking for any implemented/provided `rfb_auth_result` handlers - which there are none anymore. So the new `enqueue` call needs its own `if`. Something like:

```
// Remove in v9.1: rfb_auth_result should have been gone.
if ( rfb_auth_result )
        zeek::BifEvent::enqueue_rfb_auth_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), result);

if ( rfb_authentication_result )
        zeek::BifEvent::enqueue_rfb_authentication_result(connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(), ! result);
```

From skimming where the `result` value is parsed., maybe change `! result` to `result == 0`, as from the  protocol `0` means success and this is a bit more explicit and elsewhere there is:
```
271         function handle_security_result(msg: RFBSecurityResult) : bool          
272                 %{                                                              
273                 if ( ${msg.result} == 0 )                                       
274                         // OK                                                   
275                         server_state = SERVER_INIT;     
```

src/analyzer/protocol/rfb/events.bif
```diff
@@ -10,7 +10,14 @@ event rfb_authentication_type%(c: connection, authtype: count%);
 ## c: The connection record for the underlying transport-layer session/flow.
 ##
 ## result: whether or not authentication was successful
-event rfb_auth_result%(c: connection, result: bool%);
+event rfb_auth_result%(c: connection, result: bool%) &deprecated="Remove in v8.2. Instead use rfb_authentication_result which has the correct value for result.";
```
awelzel:
```suggestion
event rfb_auth_result%(c: connection, result: bool%) &deprecated="Remove in v9.1. Use rfb_authentication_result which has the correct value for result.";
```

src/analyzer/protocol/rfb/events.bif
```diff
@@ -10,7 +10,14 @@ event rfb_authentication_type%(c: connection, authtype: count%);
 ## c: The connection record for the underlying transport-layer session/flow.
 ##
 ## result: whether or not authentication was successful
```
awelzel:
```suggestion
## result: whether or not authentication was successful (false means success, true means failure)
##
## .. zeek:see:: rfb_authentication_result
```

src/analyzer/protocol/rfb/rfb-analyzer.pac
ynadji: fixed!

PR #4930, opened 2025-10-15
Backport zeek-client & zeek-testing-cluster fixes into 8.0.x

ckreibich:
This backports #4799 and #4812 to make zeek-client work by default (using TLS-less communication) and fix a zeek-testing-cluster testcase.


timwoj (approved):
Ignore the docs failure. That'll get fixed when I merge the docs move.

PR #4931, opened 2025-10-15
broker/WebSocketShim: Fix WebSocket clients seeing peering errors

awelzel:
The WebSocketShim subscribed to the "local errors topic" of Broker. That meant that whenever the node hosting the WebSocket server failed to peer to another node using `Broker::peer()`, every attached WebSocket client would be notified about these errors as messages that looked like:

```
{"type": "error", "code": "broker_error", "context": "data(<$>/local/data/errors, (error, peer_unavailable, ((nil, 127.0.0.1, 21/tcp, 1), unable to connect to remote peer)))"}
```

This PR fixes it by not subscribing to the local errors topic and also removes the code to handle error and status topics as neither is subscribed to anymore.

Probably 8.0.3 material.


src/broker/Manager.h
```diff
@@ -455,7 +455,7 @@ class Manager : public zeek::cluster::Backend, public iosource::IOSource {
     friend class WebSocketState;
 
     // Create a hub for WebSocket clients.
-    broker::hub MakeHub(broker::filter_type ft);
+    broker::hub MakeHub(broker::filter_type ft = {});
```
bbannier: Nit: We never seem to pass any filters, should we just drop the arg and instead hardcode `{}` in the impl?

testing/btest/cluster/websocket/broker/spurious-peer-error.zeek
```diff
@@ -0,0 +1,86 @@
...
+# @TEST-PORT: WEBSOCKET_PORT
+#
+# @TEST-EXEC: cp $FILES/broker/cluster-layout.zeek .
+# @TEST-EXEC: cp $FILES/ws/wstest.py .
+#
+# @TEST-EXEC: zeek -b --parse-only manager.zeek
+# @TEST-EXEC: python3 -m py_compile client.py
+#
+# @TEST-EXEC: btest-bg-run manager "ZEEKPATH=$ZEEKPATH:.. && CLUSTER_NODE=manager zeek -b ../manager.zeek"
+# @TEST-EXEC: btest-bg-run client "python3 ../client.py"
```
bbannier: Nit: use `client.py` to drive the test.


```suggestion
# @TEST-EXEC: python3 client.py
```

Needs baseline update below.

awelzel: Again. Thanks :-)

ckreibich:
For reference, this was what I stumbled over in zeek-client in zeek/zeek-client#54.

PR #4934, opened 2025-10-15
Audit links in docs via sphinx-build's linkcheck module

### LINKED ISSUE #4911: Fix broken links
When fixing links I broke, Tim [pointed out](https://github.com/zeek/zeek-docs/pull/316#pullrequestreview-2855756804) the link checker has broken links.

Should also look at running the checker automatically

```
sphinx-build -M linkcheck . <outputdir>
```

timwoj:
This adds `sphinx-build -M linkcheck` to our docs generation, while also fixing all of the findings from it in the existing documentation. I ended up having to ignore a bunch of links. See the "ignore rules" commit for details.

Fixes #4911

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -93,6 +93,20 @@ jobs:
       - name: Check Spicy docs
         run: cd doc && make check-spicy-docs
 
+      - name: Run sphinx linkcheck
+        run: |
+          source ci-docs-venv/bin/activate
+          cd doc
+          set -o pipefail
+          make linkcheck > make.out 2>&1
+          make_status=$?
+          echo "*** Sphinx Linkcheck Output ***"
+          cat make.out
+          test ${make_status} -ne 0 && exit 1
+          echo "Check for Linkcheck Warnings ***"
+          grep -q WARNING make.out && exit 1
+          rm make.out
+
       # Cache pre-commit environment for reuse.
       - uses: actions/cache@v4
         with:
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -18,6 +18,10 @@ html: builddir
 livehtml: builddir
 	sphinx-autobuild --ignore "*.git/*" --ignore "*.lock" --ignore "*.pyc" --ignore "*.swp" --ignore "*.swpx" --ignore "*.swx" -b html $(SPHINXOPTS) . ./build/html
 
+linkcheck:
+	mkdir -p build/linkcheck
+	sphinx-build -v -b linkcheck . ./build/linkcheck
+
 commit:
 	git add * && git commit -m 'Update generated docs'
 


doc/Makefile
```diff
@@ -18,6 +18,10 @@ html: builddir
 livehtml: builddir
 	sphinx-autobuild --ignore "*.git/*" --ignore "*.lock" --ignore "*.pyc" --ignore "*.swp" --ignore "*.swpx" --ignore "*.swx" -b html $(SPHINXOPTS) . ./build/html
 
+linkcheck:
+	mkdir -p build/linkcheck
+	sphinx-build -M linkcheck . ./build/linkcheck
```
bbannier: Maybe a good idea to add more output here with `-v`.

```suggestion
	sphinx-build -v -M linkcheck . ./build/linkcheck
```

bbannier (approved):
We tried running linkcheck in CI for Spicy, but it ended up failing spuriously too often so we took it out again and now only run it manually.

Left a few pointers to make issues easier to debug.

timwoj:
> it ended up failing spuriously too often so we took it out again and now only run it manually.

The github-anchor thing alone is so aggravating, not to mention getting rate limited and having the build run 100x slower. There's some check that causes https://www.gnu.org/software/bison/ to get rate-limited on occasion here locally too that I've yet to figure out. It eventually succeeds but it takes a while. The bison one is especially weird because we only link to that URL once in all of the docs, so it's not like we're making a ton of repeated requests for it.

.github/workflows/generate-docs.yml
```diff
@@ -93,6 +93,19 @@ jobs:
...
         run: cd doc && make check-spicy-docs
 
+      - name: Run sphinx linkcheck
+        run: |
+          source ci-docs-venv/bin/activate
+          cd doc
+          make linkcheck > make.out 2>&1
+          make_status=$?
+          echo "*** Sphinx Linkcheck Output ***"
+          cat make.out
```
bbannier, line 104: linkcheck can run very long if remotes let it wait, or just force retries with 503s which will cause linkcheck to go into a backoff loop. Maybe something like this instead?

```suggestion
set -o pipefail

echo "*** Sphinx Linkcheck Output ***"
make linkcheck | tee make.out
make_status=$?
```

timwoj: It might be worth setting the `linkcheck_timeout` and `linkcheck_rate_limit_timeout` to something smaller than the defaults (30s and 300s respectively).

doc/Makefile
timwoj: This ends up having to be `-v -b linkcheck`, which is better anyways because it makes it consistent with the `-b html` argument we use for the other Makefile step.

bbannier: Sorry for freestyling that without checking, I actually copied the `-v` from Spicy which uses exactly these flags, https://github.com/zeek/spicy/blob/main/doc/Makefile#L56

PR #4937, opened 2025-10-16
Follow-ups for recent docs work

timwoj:
This makes a couple minor changes related to the recent docs move.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -142,7 +142,7 @@ jobs:
           git status
           # git commit errors when there's nothing to commit, so guard it
           # with a check that detects whether there's anything staged.
-          git diff-index --cached --quiet HEAD || { git commit -m "Generate docs" && git push; }
+          git diff-index --cached --quiet HEAD || { git commit -m "Docs: Regenerated via GitHub workflow" && git push; }
 
       - name: Send email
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-docs.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           git status
           # git commit errors when there's nothing to commit, so guard it
           # with a check that detects whether there's anything staged.
-          git diff-index --cached --quiet HEAD || { git commit -m "Generate docs" && git push; }
+          git diff-index --cached --quiet HEAD || { git commit -m "Docs: Regenerated via GitHub workflow" && git push; }
```
awelzel, line 145: Feeling nit picky.

```suggestion
          git diff-index --cached --quiet HEAD || { git commit -m "doc: Updated Zeekygen docs via GitHub workflow" && git push; }
```

The script is called `./ci-update-zeekygen-docs.sh`, so that aligns a bit. I usually use the directory name or capitalized class names when using the `<prefix>: ` style for commit message subjects, but that's just very personal. Might be good to chat about at some point. I can adapt to other styles, I just like that as it helps me to quickly recall/guide which subsystem a commit is a about without running `git diff --stat` on it.

I'll approve - I don't mind it too much so feel free to ignore the can of worms above :-)

awelzel (approved):

PR #4938, opened 2025-10-16
Move docs link checking to a weekly workflow

timwoj:
There's no reason to run the link checking with every pull request build, since it's unlikely that the links change that frequently. Additionally, a known issue with some URL shouldn't cause every docs build to fail, such as the recent DNS failure of icir.org.

It also fixes those noted icir.org links so that the workflow will succeed.

### DIFF
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -138,23 +138,21 @@ in a low-cost manner.
 History
 =======
 
-Zeek has a rich history stretching back to the 1990s. `Vern Paxson
-<https://www.icir.org/vern/>`_ designed and implemented the initial version in
-1995 as a researcher at the `Lawrence Berkeley National Laboratory (LBNL)
-<https://www.lbl.gov/>`_. The original software was called “Bro,” as an
-“Orwellian reminder that monitoring comes hand in hand with the potential
-for privacy violations”.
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-link-check.yml
```diff
@@ -0,0 +1,62 @@
+name: Documentation Link Check
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
```
awelzel: Suggest some random hour / minutes: It probably doesn't matter, but if everyone puts `0 0` for everything everywhere might cause small spikes.

.github/workflows/docs-link-check.yml
```diff
@@ -0,0 +1,62 @@
...
+          pip3 install pre-commit
+
+      - name: Run sphinx linkcheck
+        run: |
+          source ci-docs-venv/bin/activate
+
+          cd doc
+
+          echo "*** Running Sphinx Link Check ***"
+          make linkcheck | tee make.out 2>&1
```
awelzel: I think you want this?
```
make linkcheck 2>&1 | tee make.out
```

But I probably would ditch the `make.out` thing completely.

awelzel (approved):

PR #4939, opened 2025-10-17
Topic/awelzel/systemd generator follow up

awelzel:
* Switch away from ranges for older distros
* Add StartLimitIntervalSec=0 to never stop restarting services.

### DIFF
--- a/tools/systemd-generator/CMakeLists.txt
+++ b/tools/systemd-generator/CMakeLists.txt
@@ -18,9 +18,15 @@ else ()
     set(_default_base_dir "${CMAKE_INSTALL_PREFIX}")
 endif ()
 
+set(_default ON)
+if (MSVC)
+    set(_default OFF)
+endif ()
+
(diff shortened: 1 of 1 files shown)


tools/systemd-generator/src/zeek-cluster-config.cc
```diff
@@ -35,10 +34,23 @@ void tolower(std::string& s) {
...
+        result.emplace_back(v.substr(pos, end - pos));
+        if ( end == std::string_view::npos )
+            break;
+
+        pos = end + 1;
+
+        // Trailing delimiter? Add empty entry.
+        if ( pos >= v.size() )
+            result.emplace_back(v.substr(pos, 0));
+    } while ( pos < v.size() );
```
timwoj, line 53: Dang, that's a bummer. It looks like that form for `ranges::split_view` didn't go in until GCC 12.

timwoj (approved):
I'd say just skip building this on Windows. The only reason you'd want it there is to play with building the config files.

PR #4941, opened 2025-10-17
frameworks/notice: Implement suppression batching

awelzel:
It has been observed that during a cluster restart in a live environment, workers will have lost all their suppression state and in the first few seconds may send up to 1000s of events per second out (depending on traffic and configured notices). Add functionality to batch suppression propagation before publishing it to other nodes in the cluster.

Two new options allow tweaking this behavior.

With 4 workers publishing 100.000 unique notices each at ~15.000 notices/s to 4 proxies via the manager, I'm observing the following result with `master` locally:

Master (dropped events with both backends - no tuning):
```
broker:
manager: SUMMARY [success=F, message=Received 620676, expected 1600000] duration=15.73s total_user_system=92.50s
zeromq:
manager: SUMMARY [success=F, message=Received 1595792, expected 1600000] duration=15.99s total_user_system=144.43s (note that ZeroMQ delivered a lot more notices, so more CPU time here is actually okay).
```

With this change, no more message drops and considerably less CPU time use.
```
broker:
manager: SUMMARY [success=T, message=Received 1600000, expected 1600000] duration=9.85s total_user_system=58.36s

zeromq:
manager: SUMMARY [success=T, message=Received 1600000, expected 1600000] duration=9.16s total_user_system=57.08s
```
Note that the publish-through-manager isn't great for ZeroMQ, because it results in an extra round-trip through the XPUB/XSUB broker, while for broker that's not needed.


scripts/base/frameworks/notice/main.zeek
```diff
@@ -231,6 +231,18 @@ export {
 		cuid: string  &optional; ##< Connection UID over which file is sent.
 	};
 
+	## How long to batch suppression information locally before sending
+	## it out within the cluster.
+	##
+	## Previously, workers immediately sent suppression information for every
+	## individual notice. However, at high notice and suppression rates, this
+	## causes a flood of remote events in the cluster. Batching these up for
+	## a short period of time is more efficient.
```
timwoj:
```suggestion
	## Setting this prevents workers from sending suppression information for every
	## individual notice. At high notice and suppression rates, sending individual events
	## causes a flood of remote events in the cluster. Batching these up for
	## a short period of time is more efficient. Setting this to `0msec` disables the batching.
```

scripts/base/frameworks/notice/main.zeek
```diff
@@ -231,6 +231,18 @@ export {
...
+	option suppression_batch_period = 10msec;
+
+	## Maximum number of suppression infos to batch up within batch_period.
```
timwoj:
```suggestion
	## Maximum number of suppression infos to batch up within batch_period. Setting this to `0` disables the batching.
```

scripts/base/frameworks/notice/main.zeek
```diff
@@ -356,6 +368,59 @@ function per_notice_suppression_interval(t: table[Notice::Type, string] of time,
...
+# Code for batching suppressions on worker nodes before sending them
+# out for distribution.
+type SuppressionInfo: record {
+	ts: time;
+	suppress_for: interval;
+	note: Type;
+	identifier: string;
+};
+
+# A collection of and when they were published.
```
timwoj:
```suggestion
# A collection of info and when they were published.
```

ckreibich:
Looks great to me too, thanks Arne. I have a question about
> Note that the publish-through-manager isn't great for ZeroMQ

— you're mentioning this because with ZeroMQ we can ideally rework the propagation logic to a proper broadcast, not route-via-manger, i.e. the comment this removed [here](https://github.com/zeek/zeek/pull/4941/files#diff-e388ed1969cff8bbebdc7ac623f00cddb81958ad66976ac7175bdb402373d225L527-L533), yes?

awelzel:
> — you're mentioning this because with ZeroMQ we can ideally rework the propagation logic to a proper broadcast, not route-via-manger, i.e. the comment this removed [here](https://github.com/zeek/zeek/pull/4941/files#diff-e388ed1969cff8bbebdc7ac623f00cddb81958ad66976ac7175bdb402373d225L527-L533), yes?

Correct. With global pubsub visibility, there's no reason for the manager to be involved in this anymore. Workers could just send their batches directly to other workers and proxies and the XPUB/XSUB broker will do the right thing. No need to re-publish from the manager.

I had moved that comment to the event that handles the batches:
https://github.com/zeek/zeek/pull/4941/files#diff-e388ed1969cff8bbebdc7ac623f00cddb81958ad66976ac7175bdb402373d225R632-R633

PR #4944, opened 2025-10-17
Reduce nightly container builds

timwoj:
We're currently always building/uploading new containers to the registry during the nightly jobs, even if there weren't any new changes to the branch. This adds `skip` conditions to two tasks that modify the registry to skip the task if the SHA for the last green build is the same as the current SHA.

My only concern with this is that the documentation for `CIRRUS_LAST_GREEN_CHANGE` is somewhat vague about what build it's associated with. If it's any build on the master branch, including non-nightly jobs, we potentially miss building new containers overnight if the build for the last merge to master was green. The SHAs would match and it would skip doing the build. It also means that if `master` is currently failing, the tasks would always run because the SHA would always be different.

It's unfortunate that you can't run a script as part of a `skip` condition, as we could do something fancy with GraphQL instead by looking up the last successful nightly cron build.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -732,6 +732,7 @@ amd64_container_image_docker_builder:
 container_image_manifest_docker_builder:
   cpu: 1
   << : *ONLY_IF_RELEASE_TAG_NIGHTLY
+  skip: $CIRRUS_CHANGE_IN_REPO == $CIRRUS_LAST_GREEN_CHANGE
   env:
     DOCKER_USERNAME: ENCRYPTED[!505b3dee552a395730a7e79e6aab280ffbe1b84ec62ae7616774dfefe104e34f896d2e20ce3ad701f338987c13c33533!]
     DOCKER_PASSWORD: ENCRYPTED[!6c4b2f6f0e5379ef1091719cc5d2d74c90cfd2665ac786942033d6d924597ffb95dbbc1df45a30cc9ddeec76c07ac620!]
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -732,6 +732,7 @@ amd64_container_image_docker_builder:
 container_image_manifest_docker_builder:
   cpu: 1
   << : *ONLY_IF_RELEASE_TAG_NIGHTLY
+  skip: $CIRRUS_CHANGE_IN_REPO == $CIRRUS_LAST_GREEN_CHANGE
```
awelzel, line 735: I'm wondering if this could just be `skip: !changesInclude('*')` or `skip: !changesInclude('**/*)`

The intention being to skip if no files changed. The docs sound like it might work and handle the failure / never-passed scenario a bit better? But what you have seems fine. Maybe a comment what this is about above?

https://cirrus-ci.org/guide/writing-tasks/#supported-functions

> Supported Functions[¶](https://cirrus-ci.org/guide/writing-tasks/#supported-functions)
> Currently two functions are supported in the only_if and skip expressions:

> changesInclude function allows to check which files were changed
> changesIncludeOnly is a more strict version of changesInclude, i.e. it won't evaluate to true if there are > changed files other than the ones covered by patterns
> These two functions behave differently for PR builds and regular builds:

> For PR builds, functions check the list of files affected by the PR.
> For regular builds, the CIRRUS_LAST_GREEN_CHANGE [environment variable](https://cirrus-ci.org/guide/writing-tasks/#environment-variables) will be used to determine list of affected files between CIRRUS_LAST_GREEN_CHANGE and CIRRUS_CHANGE_IN_REPO. In case CIRRUS_LAST_GREEN_CHANGE is not available (either it's a new branch or there were no passing builds before), list of files affected by a commit associated with CIRRUS_CHANGE_IN_REPO environment variable will be used instead.

timwoj: Worst that can happen is I merge it and it misses a build, and I have to fix it. These ones are somewhat hard to test.

PR #4945, opened 2025-10-19
QUIC: Add short packets as oO in binary logarithmic fashion to history

awelzel:
This should allow to visually distinguish QUIC connections based on upload/download or ping-pong characteristics. It also help to understand if short packets happen early in the handshake phase.

Also, it ends up looking rather intriguing :-)


timwoj (approved):

NEWS
```diff
@@ -135,6 +135,16 @@ New Functionality
   present, the data is added to the new ``user_security_parameters`` field of the ``SNMP::HeaderV3``
   record.
 
+- The QUIC analyzer now raises ``QUIC::short_packets_threshold()`` for every
```
timwoj:
```suggestion
- The QUIC analyzer now raises ``QUIC::short_packets_threshold_crossed()`` for every
```

PR #4946, opened 2025-10-20
ZAM fix for recursive functions that iterate over tables

vpax:
For script functions that iterate over tables, ZAM tracks state associated with the iteration(s). If the function is recursive, then this state needs to live on the stack, otherwise it's a member of the `ZBody` object. During execution, ZAM uses a pointer that points to either the dynamic state or the member variable state. That pointer used to be a member variable, which fails if the recursion occurs inside a table iteration. This PR moves the pointer to also be on the stack.


src/script_opt/ZAM/ZBody.cc
```diff
@@ -404,6 +404,11 @@ ValPtr ZBody::Exec(Frame* f, StmtFlowType& flow) {
     std::unique_ptr<TableIterVec> local_table_iters;
     std::vector<StepIterInfo> step_iters(num_step_iters);
 
+    // Points to the TableIterVec used to manage iteration over tables.
+    // For non-recursive functions, we just use the static one, but
+    // for recursive ones this points to the local stack variable.
+    TableIterVec* tiv_ptr = &table_iters;
+
```
awelzel, line 411: IIUC, this happens when doing something like this:
```
function f(...) {
    for ( loop over something ) {
        f(...);  # looping again.
    } 
```

Is it easy to add a test in `language/` so that this is tested with and without ZAM in the future?

vpax: Good thought, just added a test. (I put it in `opt/` rather than `language/`, as the former is where other optimization regression tests live).

PR #4948, opened 2025-10-21
Update Spicy code for HILTI's new tuple and optional types.

rsmmr:
This goes with https://github.com/zeek/spicy/pull/2183.

### DIFF
--- a/src/spicy/cookie.h
+++ b/src/spicy/cookie.h
@@ -15,6 +15,7 @@
 #include <vector>
 
 #include <hilti/rt/backtrace.h>
+#include <hilti/rt/types/optional.h>
 
 #include "zeek/Reporter.h"
 #include "zeek/Val.h"
@@ -30,8 +31,8 @@ namespace cookie {
 /** State representing analysis of one file. */
 struct FileState {
     FileState(std::string fid) : fid(std::move(fid)) {}
-    std::string fid;                      /**< unique Zeek-side file ID */
-    std::optional<std::string> mime_type; /**< MIME type, if explicitly set */
+    std::string fid;                            /**< unique Zeek-side file ID */
+    hilti::rt::Optional<std::string> mime_type; /**< MIME type, if explicitly set */
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/spicy/cookie.h
```diff
@@ -30,8 +31,8 @@ namespace cookie {
 /** State representing analysis of one file. */
 struct FileState {
     FileState(std::string fid) : fid(std::move(fid)) {}
-    std::string fid;                      /**< unique Zeek-side file ID */
-    std::optional<std::string> mime_type; /**< MIME type, if explicitly set */
+    std::string fid;                            /**< unique Zeek-side file ID */
+    hilti::rt::Optional<std::string> mime_type; /**< MIME type, if explicitly set */
```
bbannier, line 35: This isn't directly exposed to HILTI code, and I wonder whether instead of using runtime types here we shouldn't provide a `operator std::optional<T>` in `Optional` (we definitely wouldn't want to make the base class public).

That would help us avoid having to switch to runtime types in places like this, and also remove error potential in writing the (trivial) conversion inline.

src/spicy/protocol-analyzer.h
```diff
@@ -118,7 +118,7 @@ class ProtocolAnalyzer {
 private:
     EndpointState _originator; /**< Originator-side state. */
     EndpointState _responder;  /**< Responder-side state. */
-    std::optional<::spicy::rt::UnitContext> _context;
+    hilti::rt::Optional<::spicy::rt::UnitContext> _context;
```
bbannier, line 121: Same here.

src/spicy/runtime-support.cc
```diff
@@ -938,8 +936,8 @@ void rt::cookie::FileStateStack::remove(const std::string& fid) {
     }
 }
 
-static void _data_in(const char* data, uint64_t len, std::optional<uint64_t> offset,
-                     const std::optional<std::string>& fid) {
+static void _data_in(const char* data, uint64_t len, hilti::rt::Optional<uint64_t> offset,
```
bbannier: This should have passed `offset` by ref since it is not small.

```suggestion
static void _data_in(const char* data, uint64_t len, const hilti::rt::Optional<uint64_t>& offset,
```

bbannier (approved):

src/spicy/cookie.h
rsmmr: Hmm ... This is coming in as a runtime Optional, so storing it as that seems fine to me. Not sure I like the implicit cast, I'd rather keep the different optionals separate and conversion explicit.

PR #4950, opened 2025-10-22
Change ZAM Profiling to use per-thread CPU timing 

### LINKED ISSUE #3659: Change ZAM Profiling to use per-thread CPU timing
> I've brought this up quite a while ago and still holding onto it: This should use CPU time of the main/current thread rather than CPU time of the full process (as curr_CPU_time() currently seems to do). Zeek is multi-threaded, so this includes CPU time consumed by logger or other threads running independently and in parallel to script execution. It seems when pinning a `zeek -r` invocation to a single CPU this could skew and jitter things around.
> 
> Last time this came up with script profiling you mentioned "logging/caf" activity may be related to script activity and it would make sense to include in the profile (which I'd still challenge :-) ). However, this is now about individual statements and there's logic to to correct for measurement overhead. Seems to me excluding noise/cpu time used by parallel running threads would be more than reasonable for more consistent measurements first.

_Originally posted by @awelzel in https://github.com/zeek/zeek/pull/3644#discussion_r1523554382_

The thread linked above has a bunch more context/ideas in it.

ghost:
### Changes:
1. Added new namespace `zeek::utils::this_thread` having function `get_cpu_time` which measures CPU time consumed by this thread
2. In ZAM Profiling, replaced the current `curr_CPU_time` (which measures CPU time consumed by the whole process) with `zeek::utils::this_thread::get_cpu_time`.

resolves #3659

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -405,8 +405,15 @@ extern int time_compare(struct timeval* tv_a, struct timeval* tv_b);
 // Returns the CPU time consumed to date.
 extern double curr_CPU_time();
 
+namespace this_thread {
 // Returns the thread CPU time consumed to date.
-extern double curr_THREAD_time();
+inline double get_cpu_time() {
+    struct timespec ts;
+    clock_gettime(CLOCK_THREAD_CPUTIME_ID, &ts);
+    return double(ts.tv_sec) + double(ts.tv_nsec) / 1e9;
+}
+} // namespace this_thread
+
 
 // Returns an integer that's very likely to be unique, even across Zeek
 // instances. The integer can be drawn from different pools, which is helpful
--- a/src/script_opt/ZAM/ZBody.cc
+++ b/src/script_opt/ZAM/ZBody.cc
@@ -40,18 +40,18 @@ static double mem_prof_overhead = 0.0;
 // minimum has proven robust in practice.
 //
 // Note that the measurement itself has some overhead from calling
-// util::curr_THREAD_time(), though this becomes quite minor as long as "navg"
+// this_thread::get_cpu_time(), though this becomes quite minor as long as "navg"
 // isn't too small / "f" is sufficiently heavyweight.
 
 static double est_min_overhead(void (*f)(), int n, int navg) {
-    double last_t = util::curr_THREAD_time();
+    double last_t = zeek::util::this_thread::get_cpu_time();
     double min_dt = -1.0;
     int ncall = 0;
 
     for ( int i = 0; i < n; ++i ) {
         (*f)();
         if ( ++ncall % navg == 0 ) {
-            double new_t = util::curr_THREAD_time();
+            double new_t = zeek::util::this_thread::get_cpu_time();
             double dt = new_t - last_t;
             if ( min_dt >= 0.0 )
                 min_dt = std::min(min_dt, dt);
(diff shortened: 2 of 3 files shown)


src/util.h
```diff
@@ -405,6 +405,9 @@ extern int time_compare(struct timeval* tv_a, struct timeval* tv_b);
 // Returns the CPU time consumed to date.
 extern double curr_CPU_time();
 
+// Returns the thread CPU time consumed to date.
+extern double curr_THREAD_time();
```
awelzel: Would you be up for renaming and implementing it in a custom `this_thread` namespace, so that it ends up like this?

```
namespace this_thread {
  // Returns the CPU time in seconds (user+system) consumed by the calling thread.
  double get_cpu_time();
}
```

That aligns slightly with `std::this_thread` namespace which provides accessors that related to the calling/current thread, and the caller can then do:
```
double x = zeek::detail::this_thread::get_cpu_time();
```
And it essentially doesn't need a comment IMO.

If `curr_CPU_time()` ends up not being used afterwards, can also deprecate it.

ghost: Should I define `this_thread` namespace at the end of `util.cc` or someplace else ?

awelzel: Inline implementing it in `util.h` would be fine. So where currently it says `extern double curr_THREAD_time();`, you should be able to open `namespace this_thread { ... }`.

ghost: But then it will come under `zeek::util` namespace, whereas in first comment you mentioned to define `this_thread ` under `zeek::detail` namespace.

awelzel: Ah, I was confused. `zeek::util::this_thread` seems fine.

awelzel (approved):
I've tested it with `ZEEK_PROFILE=1 zeek -OZAM  -r /media/awelzel/T7/pcaps/2009-M57-day11-18.trace` and the zprof.out file created that looks like so:

```
Profile sampled every 100 instructions; all calls profiled
Profiling overhead = 164 nsec/instruction, memory profiling overhead = 287 nsec/call
any-vector-elem-assign-VVV      36      0.000012
table-elem-assign-VV    115     0.000134
table-elem-assign-VC    2       0.000004
add-VVC-U       304     0.000044
add-VVCi-field-S        8       0.000006
...
smb1_message 60 5 0.000001 goto-b 65 // SMB1::smb1_message;SMB1::smb1_message:62-67;SMB1::smb1_message:64-65;SMB1::smb1_message:65
smb1_message 61 9 0.000001 field-RVi-R 0 (<internal>::#42), 0 (SMB1::c), 49 // SMB1::smb1_message;SMB1::smb1_message:62-67;SMB1::smb1_message:67
smb1_message 62 5 0.000001 field-RVi-T 0 (<internal>::#43), 0 (<internal>::#42), 3 // SMB1::smb1_message;SMB1::smb1_message:62-67;SMB1::smb1_message:67
smb1_message 63 6 0.000001 field-RVi-U 2 (<internal>::#44), 1 (SMB1::hdr), 7 // SMB1::smb1_message;SMB1::smb1_message:62-67;SMB1::smb1_message:67
smb1_message 64 4 0.000004 deltable-VO 0 (<internal>::#43) // SMB1::smb1_message;SMB1::smb1_message:62-67;SMB1::smb1_message:67
smb1_read_andx_request did not execute
smb1_write_andx_request did not execute
smb2_message did not execute
module FileHash sampled CPU time 0.000010, 41 sampled instructions
module SMB sampled CPU time 0.000041, 184 sampled instructions
module KRB sampled CPU time 0.000039, 206 sampled instructions
```

I'll merge it. @vpax , ping that the ZAM profiling has changed from process CPU time to thread CPU time such that CPU time spent by broker or logging threads isn't included in the ZAM profiling anymore.

vpax:
@awelzel Thanks for the ping. That's probably more useful than the process time.

PR #4952, opened 2025-10-22
Escape specific characters in LDAP filter strings

### LINKED ISSUE #3955: LDAP search filter escaping
Hi team,

I have a question. We are using the LDAP parser to log LDAP search filters in the ldap_search logs. We postprocess these queries using different tooling and we notice that some of the written LDAP queries get rejected by other parsers.

After some investigation I have a hypothesis on why this is so. According to https://www.rfc-editor.org/rfc/rfc4515#section-3, when writing a string representation of a search query, certain special characters in attribute values need to be escaped.

For example:
```
(&(name=Test(Admin!)))
```

Should be escaped as
```
(&(name=Test\28Admin!\29))
```

While the first representation is easier to read for a human, it might get rejected by other libraries for parsing.

I wonder if it would be an idea to consider escaping the search string according to aforementioned RFC and have it either be configurable (escape yes/no) or write an escaped and human readable version in two different fields.

I have added a PCAP with a few different queries that may or may not require escaping for testing purposes in the future.

What is your take on this?

[ldap_escaping_test.pcapng.txt](https://github.com/user-attachments/files/17163980/ldap_escaping_test.pcapng.txt)

timwoj:
According to https://www.rfc-editor.org/rfc/rfc4515#section-3, LDAP filter strings should have certain characters escaped as their hex values instead of putting the actual character in the output.

Fixes #3955


src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -670,6 +670,27 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
     return ret;
 }
 
+public function escape_filter_string(bts: bytes): string {
+    local ret = "";
+    local cnt = 0;
+
+    while (cnt < |bts|) {
```
bbannier: We could use a `for` loop here to avoid manual counter management, and having to get the value out of `bts`.

```ruby
    for (bt in bts) { ... }
```

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -670,6 +670,27 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
...
+        local bt: uint8 = *bts.at(cnt);
+
+        if (bt in [0, 27, 40, 41, 42]) {
```
bbannier: I would suggest hoisting the `set` creation out of the loop.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -670,6 +670,27 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
...
+            ret += "\\%02x" % bt;
+        } else {
+            ret += "%c" % bt;
+        }
+
+        cnt += 1;
+    }
+
+    print ret;
```
bbannier: This looks like stray debug output.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -670,6 +670,27 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
     return ret;
 }
 
+public function escape_filter_string(bts: bytes): string {
```
bbannier, line 673: With all the other comments this could become

```ruby
public function escape_filter_string(bts: bytes): string {
    local ret = "";
    local replace: set<uint8> = [0, 27, 40, 41, 42];

    for (bt in bts) {
        if (bt in replace) {
            ret += "\\%02x" % bt;
        } else {
            ret += "%c" % bt;
        }
    }

    return ret;
}
```

The general motivation would be to minimize Spicy code we need to execute since this might see pretty heavy traffic. It might also make sense to add an early return of `bts.decode()` if nothing in the input needs escaping.

I was also wondering whether we should have a function to replace individual bits in a `bytes`, right now we could `match` them, but don't gain much in constructing the new string. We can probably add that if it becomes an issue.

src/analyzer/protocol/ldap/ldap.spicy
timwoj: A lot of this was following the existing patterns from the methods above this one. I'll open a separate PR to fix those once I merge this one.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -670,6 +670,21 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
...
+    local ret = "";
+    local chars_to_escape: set<uint8> = [0, 27, 40, 41, 42];
+
+    for (bt in bts) {
+        if (bt in chars_to_escape) {
+            ret += "\\%02x" % bt;
+        } else {
+            ret += "%c" % bt;
```
awelzel, line 681: Hmm, this breaks on search strings containing multi-byte UTF-8 characters, no?

If someone has a LDAP search string with an umlaut or so, maybe that would be a good test case :-)

> [3](https://www.rfc-editor.org/rfc/rfc4515#section-3).  String Search Filter Definition
>   The string representation of an LDAP search filter is a string of
?   UTF-8 [[RFC3629](https://www.rfc-editor.org/rfc/rfc3629)] encoded Unicode characters [[Unicode]?(https://www.rfc-editor.org/rfc/rfc4515#ref-Unicode)] that is defined
   by the following grammar, [...]

bbannier: Hmm yeah, this seems to miss [parts of the range](https://www.rfc-editor.org/rfc/rfc4515#section-3) and doesn't deal with UTF BOMs:

>       normal         = UTF1SUBSET / UTFMB
>       escaped        = ESC HEX HEX
>       UTF1SUBSET     = %x01-27 / %x2B-5B / %x5D-7F
>                           ; UTF1SUBSET excludes 0x00 (NUL), LPAREN,
>                           ; RPAREN, ASTERISK, and ESC.

The spec claims this is sufficient to produce a valid encoding.

> The <valueencoding> rule ensures that the entire filter string is a valid UTF-8 string and provides that the octets that represent the ASCII characters "*" (ASCII 0x2a), "(" (ASCII 0x28), ")" (ASCII 0x29), "\" (ASCII 0x5c), and NUL (ASCII 0x00) are represented as a backslash "\" (ASCII 0x5c) followed by the two hexadecimal digits representing the value of the encoded octet.
>
> This simple escaping mechanism eliminates filter-parsing ambiguities and allows any filter that can be represented in LDAP to be represented as a NUL-terminated string.  Other octets that are part of the <normal> set may be escaped using this mechanism, for example, non-printing ASCII characters.

Unfortunately we need to accept non-unicode data.

> Implementations SHOULD accept as input strings that are not valid UTF-8 strings.  This is necessary because RFC 2254 did not clearly define the term "string representation" (and in particular did not mention that the string representation of an LDAP search filter is a string of UTF-8-encoded Unicode characters).

awelzel: > The spec claims this is sufficient to produce a valid encoding.

I *think* if instead of the `"%c" % bt` we'd have an intermediary `bytes` that gets just `bt` appended, that might work out. That would likely behave better. As is, I'm pretty sure an UTF-8 encoded Ü in a filter string would result in looking really bad.

```
$ cat Üü.spicy 
module Uu;

# UTF-8 encoded Üü
global vec = vector<uin8>(0xc3, 0x9c, 0xc3, 0xbc);

for (v in vec) {
    print "%x" % v;
    print "%c" % v;
}

$ spicyc -j Üü.spicy 
c3
<illegal UTF8 sequence>
9c
<illegal UTF8 sequence>
c3
<illegal UTF8 sequence>
bc
<illegal UTF8 sequence>
```

timwoj: It seems fine with a UTF character in the filter string. I built a really simple packet in scapy with just a single equality in it:

```
>>> pkt2
<IP  version=4 ihl=5 tos=0x0 len=97 id=1 flags= frag=0 ttl=64 proto=tcp chksum=0xf6d7 src=192.168.1.100 dst=192.168.1.10 |<TCP  sport=54321 dport=ldap seq=1001 ack=2001 dataofs=5 reserved=0 flags=PA window=8192 chksum=0x5c23 urgptr=0 |<LDAP  messageID=0x1 <ASN1_INTEGER[1]> protocolOp=<LDAP_SearchRequest  baseObject=<ASN1_STRING[b'dc=example,dc=com']> scope='wholeSubtree' 0x2 <ASN1_ENUMERATED[2]> derefAliases='neverDerefAliases' 0x0 <ASN1_ENUMERATED[0]> sizeLimit=0x0 <ASN1_INTEGER[0]> timeLimit=0x0 <ASN1_INTEGER[0]> attrsOnly=False <ASN1_BOOLEAN[0]> filter=<LDAP_Filter  filter=<LDAP_FilterEqual  attributeType=<ASN1_STRING[b'name']> attributeValue=<ASN1_STRING[b'\xc3\x9c\xc3\xbc']> |> |> |> Controls=None |>>>
```

ldap_search.log with that packet:

```
1761342480.275434       CwIKUlV09gUVikfz4       192.168.1.100   54321   192.168.1.10    389     1       tree          never   dc=example,dc=com       0       -       -       (name=Üü)       -
```

Strangely though, wireshark seems to think that's part of a MAC address in the value:

Screenshot 2025-10-24 at 2 48 41 PM

awelzel: Interesting. Thanks for checking - seems concatenating the `"%c" % v` results end up valid utf-8 again. Very curious and might be worth adding that pcap as test case, seems an area that could change surprisingly.

```
module Uu;

# UTF-8 encoded Üü
global vec = vector<uin8>(0xc3, 0x9c, 0xc3, 0xbc);
global res: string;

for (v in vec) {
    print "%x" % v;
    print "%c" % v;
    res += "%c" % v;
}

print "res", res;


$ spicyc -j Üü.spicy 
c3
<illegal UTF8 sequence>
9c
<illegal UTF8 sequence>
c3
<illegal UTF8 sequence>
bc
<illegal UTF8 sequence>
res, Üü
```

PR #4957, opened 2025-10-22
af_packet: Warn for usage of loopback interface

timwoj:
Reimplementation of https://github.com/zeek/zeek-af_packet-plugin/pull/59 in the Zeek repo, since the af-packet repo is archived.

### DIFF
--- a/src/iosource/af_packet/AF_Packet.cc
+++ b/src/iosource/af_packet/AF_Packet.cc
@@ -2,6 +2,7 @@
 
 #include "zeek/iosource/af_packet/AF_Packet.h"
 
+#include "zeek/Reporter.h"
 #include "zeek/iosource/af_packet/RX_Ring.h"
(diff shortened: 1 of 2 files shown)


src/iosource/af_packet/AF_Packet.cc
```diff
@@ -60,6 +61,18 @@ void AF_PacketSource::Open() {
         return;
     }
 
+    if ( info.IsLoopback() ) {
+        zeek::reporter->Warning(
+            "AF_Packet: %s is a loopback interface! Using the native AF_PACKET"
+            "AF_PACKET packet source with a loopback interface is unsupported. "
```
awelzel, line 67:
```suggestion
            " packet source with a loopback interface is unsupported. "
```

timwoj: I'm going to edit that message to have some linebreaks in it too instead of just wrapping.

PR #4958, opened 2025-10-22
Clean up LDAP spicy functions

timwoj:
This follows some recommendations from https://github.com/zeek/zeek/pull/4952 to clean up a few of the other utility functions in the LDAP analyzer.

### DIFF
--- a/src/analyzer/protocol/ldap/ldap.spicy
+++ b/src/analyzer/protocol/ldap/ldap.spicy
@@ -641,13 +641,15 @@ public function utf16_guid_to_hex_repr(bts: bytes): string {
         return "GUID_FORMAT_FAILED";
     }
 
+    local guid_breaks: set<uint64> = [0, 4, 6, 9];
+
     local ret = "";
     for (i in [[3, 2, 1, 0], [5, 4], [7, 6], [8, 9], [10, 11, 12, 13, 14, 15]]) {
         for (j in i) {
             local bt: uint8 = *bts.at(j);
-            ret = ret + "%02x" % bt;
-            if (j in [0, 4, 6, 9]) {
-                ret = ret + "-";
+            ret += "%02x" % bt;
+            if (j in guid_breaks) {
+                ret += "-";
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -641,13 +641,16 @@ public function utf16_guid_to_hex_repr(bts: bytes): string {
         return "GUID_FORMAT_FAILED";
     }
 
+    local guid_parts: vector<vector<uint8>> = [[3, 2, 1, 0], [5, 4], [7, 6], [8, 9], [10, 11, 12, 13, 14, 15]];
+    local guid_breaks: set<uint8> = [0, 4, 6, 9];
+
     local ret = "";
-    for (i in [[3, 2, 1, 0], [5, 4], [7, 6], [8, 9], [10, 11, 12, 13, 14, 15]]) {
+    for (i in guid_parts) {
```
bbannier: The temporary we previously created here already lived for the duration of the function, so this should behave identically but could be less clear; similar for `sid_parts` and `reversed` below. My hunch would be to keep the previous forms.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -658,12 +661,11 @@ public function bytes_sid_to_hex_repr(bts: bytes): string {
     local ret = "";
     local cnt = 0;
 
-    while (cnt < |bts|) {
-        local bt: uint8 = *bts.at(cnt);
-        ret = ret + "%02x" % bt;
+    for (bt in bts) {
```
bbannier, line 663: Just as a note, since we get also use enumerate the elements via `cnt` this seems redundant, but is actually probably still better since we avoid indexing into `bts` (via `at` and dereferencing, or similar with `bts[...]`) which would be checked at runtime.

src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -641,13 +641,16 @@ public function utf16_guid_to_hex_repr(bts: bytes): string {
...
+
     local ret = "";
-    for (i in [[3, 2, 1, 0], [5, 4], [7, 6], [8, 9], [10, 11, 12, 13, 14, 15]]) {
+    for (i in guid_parts) {
         for (j in i) {
             local bt: uint8 = *bts.at(j);
-            ret = ret + "%02x" % bt;
-            if (j in [0, 4, 6, 9]) {
-                ret = ret + "-";
+            ret += "%02x" % bt;
```
bbannier, line 650: These I would expect to have the biggest perf impact since we avoid guaranteed heap allocations.

Issue #4951, opened 2025-10-22
`zeek -u` False positives on `when` captures


evantypanski:
Zeek has a [`-u` flag](https://docs.zeek.org/en/master/scripting/usage.html) (that I don't know many people know about) which detects "usage errors" for variables. But, given basic captures in `when` statements, it detects any uses of the capture as used without definition. Here's the output from the documentation's [`when` example](https://docs.zeek.org/en/master/script-reference/statements.html#when):

```
# when.zeek
 
type r: record { x: int; y: int; };
global g = r($x=100, $y=100);

event zeek_init()
    {
    local l = r($x=1, $y=2);
    local l2 = r($x=3, $y=4);

    when [l, copy l2] ( g$x < 0 )
        {
        print l, l2;
        }

    l$x = 10;
    l2$x = 20;
    }

event zeek_init() &priority=-10
    {
    g$x = -999;
    }
```

```
$ zeek -ub when.zeek

warning in ./when.zeek, line 13: used without definition (l)
warning in ./when.zeek, line 13: used without definition (l2)
[x=10, y=2], [x=3, y=4]
```

Unfortunately, this causes a large amount of spam when used without bare mode. From my local install:

```
$ zeek -u when.zeek
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/exec.zeek, line 181: used without definition (Exec::cmd)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 101: used without definition (ActiveHTTP::bodyfile)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 101: used without definition (ActiveHTTP::headersfile)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 101: used without definition (ActiveHTTP::cmd)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 101: used without definition (ActiveHTTP::stdin_data)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 104: used without definition (ActiveHTTP::result)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 106: used without definition (ActiveHTTP::req)
warning in /Users/etyp/.local/zeek/share/zeek/base/utils/active-http.zeek, line 107: used without definition (ActiveHTTP::resp)
[...]

PR #4960, opened 2025-10-23
Usage Analyzer Fixes

vpax:
This PR addresses https://github.com/zeek/zeek/issues/4951 by (1) adapting the analysis to the lambda-based `when` implementation we now use, and (2) avoiding reduction-level `hook` optimization when doing usage analysis. I updated the associated BTest to now not use `-b` and to catch a few more instances of unused assignments, and removed a couple of unused assignments from the base scripts.


src/parse.y
```diff
@@ -736,8 +736,10 @@ expr:
 			set_location(@2, @4);
 			if ( ! locals_at_this_scope.empty() )
 				locals_at_this_scope.back().insert($2);
-			$$ = add_and_assign_local({AdoptRef{}, $2}, {AdoptRef{}, $4},
+			IDPtr local_id_ptr{AdoptRef{}, $2};
+			$$ = add_and_assign_local(local_id_ptr, {AdoptRef{}, $4},
 			                                        val_mgr->True()).release();
+			local_id_ptr->AddAttr(make_intrusive<Attr>(ATTR_IS_ASSIGNED));
```
evantypanski: Could this logic just go in `add_and_assign_local`, then revert this `local_id_ptr` change? This part made my "use after move" instincts jump until I realized that function copies `local_id_ptr`.

I think anything that would use that function would want this anyway

testing/btest/Baseline/language.unused-assignment/out
```diff
@@ -1,2 +1,5 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
-warning: please_warn assignment unused: please_warn = test; <...>/unused-assignment.zeek, line 10
+warning: reassign assignment unused: reassign = complain here too; <...>/unused-assignment.zeek, line 17
+warning: reassign assignment unused: reassign = original; <...>/unused-assignment.zeek, line 16
+warning: please_warn assignment unused: please_warn = test; <...>/unused-assignment.zeek, line 13
```
evantypanski, line 4: just a note, the warnings for assignment unused have the source info *last*, but the use-without-definition has it first:

```
warning in /Users/etyp/.local/zeek/share/zeek/base/frameworks/notice/actions/pp-alarms.zeek, line 245: used without definition (Notice::h2name_)
warning: Notice::resp_p assignment unused: Notice::resp_p = ; /Users/etyp/.local/zeek/share/zeek/base/frameworks/notice/actions/pp-alarms.zeek, line 174
```

Just kind of annoying since they're part of the same output. Could fix here or can fix it later, just pointing it out

src/parse.y
vpax: Makes sense. Are you game to make that tweak (and also adjust the warning output as you frame below)?

evantypanski: I can grab them tomorrow

vpax: Thanks!

testing/btest/Baseline/language.unused-assignment/out
evantypanski: I'll do this in a followup PR in case anyone has extra comments, but it's a 1-liner

PR #4962, opened 2025-10-23
TCP::raw_options fix ubsan error

0xxon:
Now we special-case the empty-data case.

This should fix the ubsan error. I was a bit torn between assigning the empty string, or allowing this field to be optional. I currently opted for the empty string, but I am open to feedback.

### DIFF
--- a/src/analyzer/protocol/tcp/functions.bif
+++ b/src/analyzer/protocol/tcp/functions.bif
@@ -202,7 +202,10 @@ function raw_options%(stop_at_eol: bool &default=T%): TCP::RawOptionList
 
 		auto r = make_intrusive<RecordVal>(tcp_option);
 		r->Assign(0, opt);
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/tcp/functions.bif
```diff
@@ -202,7 +202,10 @@ function raw_options%(stop_at_eol: bool &default=T%): TCP::RawOptionList
 
 		auto r = make_intrusive<RecordVal>(tcp_option);
 		r->Assign(0, opt);
-		r->Assign(1, zeek::make_intrusive<zeek::StringVal>(current_option_data.size(), reinterpret_cast<const char*>(current_option_data.data())));
+		if ( current_option_data.size() > 0 )
```
timwoj, line 205:
```suggestion
		if ( ! current_option_data.empty() )
```

timwoj (approved):
I think leaving it as an empty string is fine.

src/analyzer/protocol/tcp/functions.bif
timwoj: I took care of it during the merge.

PR #4966, opened 2025-10-24
Make unused assignment warning more consistent

evantypanski:
followup for https://github.com/zeek/zeek/pull/4960#discussion_r2461031119 to make the warning more consistent, now the outputs from `zeek -u` are the same.

New PR in case there are extra comments here ¯\\\_(ツ)_/¯


timwoj:
Can you fix the output for the `language.unused-assignment` btest under ZAM here as well?

evantypanski:
> Can you fix the output for the language.unused-assignment btest under ZAM here as well?

🤦  done

timwoj:
I poked the ubuntu24_04_zam task on CI, so you might wait for that one to pass. It's the one that was showing a failure after the original merge.

evantypanski:
I'll merge since the test passed, but job failed, there's a separate fix for that in the works #4962

PR #4971, opened 2025-10-29
Add SHA224 and SHA384 BiFs

0xxon:
Both of these do apparently see real-world use, e.g. in intelligence signatures.


timwoj (approved):
Approving because it's basically just duplicating the existing model, but this code is really really repetitive and feels like there should be ways to template some of it.

0xxon:
Yeah, I am aware that it's quite repetetive - if I was redoing all of this completely, I would do it differently.

Issue #4838, opened 2025-09-18
Ethernet IEEE 802 length field.


arm7mm:
Hello.

Is there any reason not to add the len = protocol line to the Ethernet and VLAN analyzer code for case protocol <= 1500?

J-Gras:
Adding where?

arm7mm:
https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/ethernet/Ethernet.cc#L52

        if ( len < protocol || protocol < 2 ) {
            Weird("truncated_ethernet_frame", packet);
            return false;
        }   
        len = protocol;

and
https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/vlan/VLAN.cc#L37

        if ( len < protocol || protocol < 2 ) {
            Weird("truncated_vlan_frame", packet);
            return false;
        }
        len = protocol;

timwoj:
What would this accomplish? Setting it to something smaller than what's actually available in the packet would cause problems later down the processing tree.

0xxon:
On the flipside, parsing more data than is indicated in the packet header can also lead to problems further down the pipeline.

For other protocols, we tend to do exactly what is proposed here. If there is more data available than indicated in the packet header, we tend to cut the length to the length that the packet header indicates. This is, e.g., done for TCP and UDP if I am not completely mistaken.

What makes this a bit weird is that Ethernet 2 doesn't indicate packet length in the header, wheread 802.3 frames do indicate packet lengths in the header. Simultaneously, 802.3 frames are not exactly used a lot.

As we are doing this with the other protocols, I am tempted to add the proposed checks - perhaps with a weird, and test them on some deployments to see if this driggers.

arm7mm:
The task of reassembling fragmented S7COMM frames prompted me to create this question.
The following encapsulation is considered: IEEE 802.3 Ethernet => LLC => CLNP Inactive subset (in wireshark terms)
=> COTP => S7COMM.

IEEE 802.3 Ethernet frames may contain padding bytes to at least meet the minimum length requirement (64 bytes). During the process of reassembling an S7COMM frame from COTP segments, padding bytes must be skipped as they are not part of the payload.

Setting len to protocol in the places mentioned seems to help with this.

To be fair, I should note that such cases have not yet been encountered, but the theoretical possibility exists...

I couldn't find any way to get information about the number of padding bytes during segments reassembly.

arm7mm:
Thank you for your participation.

[...]

PR #4973, opened 2025-10-29
Enforce length for 802.3 frames. Accept truncated 802.3 frames.

0xxon:
Non-ethernet-2 frames provide the length in one of their fields.

This commit switches to using the provided length during parsing. This means that trailing bytes will now be ignored, whereas previously they would have been passed on to downstream protocol analyzers.

This change aligns our parsing of 802.3 packets to the parsing of other packet-types, where do enforce length fields (e.g. tcp/udp).

Furthermore, this commit now accepts trucated 802.3 frames; before they were discarded. This similarly aligns us with other protocol parsers, where this already was done; this also already was done with 802.3 frames wrapped in VLANs.

Thirdly, this commit changes the weird value `truncated_vlan_frame` to `truncated_VLAN_header`. The weird was not updated when the check was changed from a truncated frame check to a truncated header check. This aligns the weird with another weird earlier in the file, and also makes the handling the same as for normal ethernet packets in Ethernet.cc.

Addresses GH-4838

### DIFF
--- a/src/packet_analysis/protocol/vlan/VLAN.cc
+++ b/src/packet_analysis/protocol/vlan/VLAN.cc
@@ -35,9 +35,11 @@ bool VLANAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
 
         // Need at least two bytes to check the packet types below.
         if ( len < 2 ) {
-            Weird("truncated_vlan_frame", packet);
+            Weird("truncated_VLAN_header", packet);
             return false;
         }
+        if ( len > protocol )
+            len = protocol; // use 802.3/802.2 length field and remove trailing bytes
 
         if ( data[0] == 0xAA && data[1] == 0xAA )
             // IEEE 802.2 SNAP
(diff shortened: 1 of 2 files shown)


src/packet_analysis/protocol/vlan/VLAN.cc
```diff
@@ -35,9 +35,11 @@ bool VLANAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
 
         // Need at least two bytes to check the packet types below.
         if ( len < 2 ) {
-            Weird("truncated_vlan_frame", packet);
+            Weird("truncated_VLAN_header", packet);
```
timwoj, line 38: Do we need to call out that this weird changed names in NEWS?

timwoj (approved):

src/packet_analysis/protocol/vlan/VLAN.cc
0xxon: I'm happy to do that.

PR #4974, opened 2025-10-29
Deprecate return values in hooks

evantypanski:
Prompted from me being confused about why I could return values from hooks, and curiosity about why [this post](https://zeek.org/2021/06/functions-hooks-and-events-oh-my/) says they can return booleans. Turns out, it does nothing! That seems confusing, so I'd rather others not think that returning `F` would make the `hook my_hook()` call return `F` as well.

I think it was this way since it was easier to appease the type system, but I don't mind an edge case to avoid confusion.


evantypanski:
Just to put in the PR some discussion from [slack](https://zeekorg.slack.com/archives/CTGDG83B8/p1761748593531589): It's pretty unintuitive that you would use `break` to exit from executing more hooks, and *that* makes it return `F` when called. So we could also just make returning `T` mean continue, then `F` means stop. An empty return would probably just be implicit `T`. Then, deprecate `break` in hooks.

I'm normally pretty happy with changing the language to be more intuitive, but that feels like it's punishing users for doing it the correct way. If you've been good and don't ever `return F;` in your hooks (which does nothing, remember) you now have to go in and change all of the `break;` with a `return F;` - just so newer users can have a bit easier time. This has been around for *decades* (I think idk I didn't check). That seems pretty user-hostile to me, even if I like the end result more

The blast radius of this PR is just deprecating the unintuitive thing - returning `T` or `F`. We could even mention `break` in the error/deprecation warning to point users the right way. I think that's nicer, but happy to hear other opinions. Just putting it in the PR directly for history keeping :)

doc/script-reference/types.rst
```diff
@@ -822,11 +822,10 @@ will not be visible to remaining ``hook`` handlers, but it's still possible
...
 :zeek:type:`record`, :zeek:type:`set`, or :zeek:type:`table`.
 
-The return value of a hook call is an implicit :zeek:type:`bool`
-value with ``T`` meaning that all handlers for the hook were
-executed and ``F`` meaning that only some of the handlers may have
-executed due to one handler body exiting as a result of a ``break``
-statement.
+Hooks cannot return values, but the return value of a hook call is an
+implicit :zeek:type:`bool` value. True (``T``) means that all handlers
+for the hook were executed. False (``F``) means that one handler halted
```
awelzel:
```suggestion
for the hook were executed and none encountered/executed a``break`` statement. False (``F``) means that one handler halted
```

Not sure that's needed, but if the last hook breaks, technically all hooks were executed?

src/Stmt.cc
```diff
@@ -1374,6 +1374,11 @@ ReturnStmt::ReturnStmt(ExprPtr arg_e) : ExprStmt(STMT_RETURN, std::move(arg_e))
     }
 
     else {
+        // Hooks returning a value does nothing, but it's necessary to allow
+        // since they yield a boolean. This should be an error.
+        if ( ft->Flavor() == FUNC_FLAVOR_HOOK )
+            Warn("Remove in v9.1: Hooks returning values is deprecated. Use an empty return instead.");
```
awelzel:
```suggestion
            Warn("Remove in v9.1: Hooks returning values is deprecated. Use an empty return instead, or a break if the condition should inhibit execution of lower-priority hooks.");
```
I'm unable to word this properly, but if someone has a return, maybe it'd be a good nudge for them to reconsider semantics? Or is that only confusing?

NEWS
```diff
@@ -128,6 +123,9 @@ Removed Functionality
 Deprecated Functionality
 ------------------------
 
+- Hooks returning boolean values is now deprecated. Instead, return without
+  a value. Previously, the value was entirely ignored.
```
```suggestion
Return values from a hooks were previously silently discarded. This behavior was found surprising in relation with the ``break`` statement available in hooks. In Zeek 9.1, returning a value from a hook will become an error.
```

src/Stmt.cc
evantypanski: I reworded a bit more to mention empty returns second and add "if necessary"

NEWS
evantypanski: Took this suggestion but put the part that explains what is deprecated first for clarity

evantypanski:
I'm going to leave this open till Monday and merge then, since there was a little bit of contention around if we wanted to deprecate this or change the behavior, still happy to hear other's opinions

PR #4979, opened 2025-10-30
Latest round of Coverity fixes

timwoj:
Most of these are in the new systemd generator. Minor, yes, but easy to close the Coverity issues for them.


evantypanski (approved):
Man, gotta love C++ :)

Issue #4972, opened 2025-10-29
Removal of libpcap-dev and libssl-dev dependencies for plugin builds


awelzel:
A bit of a nice to have and low priority, but could we remove dependencies on libpcap-dev and libssl-dev for plugins that do not require them (think the rot13 example) or the following trivial BiF to make `top` show user configurable process names.
```
%%{
#include <zeek/util.h>
#include <zeek/Val.h>
%%}

function setproctitle%(title: string%): bool
    %{
    zeek::util::detail::set_thread_name(title->CheckString());
    return zeek::val_mgr->True();
    %}
```

Compiling this in `zeek/zeek-dev` requires to install the following packages to get started:

1) cmake
2) g++
3) make

So far so good. Then, this runs into:
```
CMake Error at /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY
  OPENSSL_INCLUDE_DIR)
Call Stack (most recent call first):
```
This is already surprising, as the plugin doesn't require OpenSSL, so why should the dependency be required? I thought we might also be at a point where Zeek's headers are clean of any OpenSSL includes so maybe can just remove (and instruct users to call `find_package(OpenSSL)` and pass `DEPENDENCIES` to `zeek_add_plugin()` or expose the required paths in some CMake config that the build ships (not exactly sure - but I remember there being a requirement to link against the same libssl used by the Zeek build).

After that the cmake configuration works out, but buliding the plugin fails as follows:
```
root@c4349067f5db:/src/build# make
[...]
In file included from /usr/local/zeek/include/zeek/Conn.h:18,
                 from /usr/local/zeek/include/zeek/zeek-bif.h:6,
                 from /src/build/test.bif.h:9,
                 from /src/build/test.bif.cc:4:
/usr/local/zeek/include/zeek/iosource/Packet.h:18:10: fatal error: pcap.h: No such file or directory
   18 | #include <pcap.h> // For DLT_ constants
      |          ^~~~~~~~
compilation terminated.
```
And that's also sad. There's a `libpcap-dev` dependency even though the plugin doesn't care. I understand `iosource/Packet.h` needs `libpcap-dev`, but it'd be nice if BiFs wouldn't automatically get that include.

timwoj:
[...]

PR #4980, opened 2025-10-30
Remove pcap.h dependency from iosource/Packet.h

timwoj:
I thought to try to add `#include <net/bpf.h>` instead to Packet.h, since that should always exist, but that results in some other compiler failures due to duplicate definitions when we later include `pcap.h`. Interestingly, removing this dependency also breaks a bunch of files from building because they were depending on `inttypes.h` to have been included by it. Those are fixed in this PR also, but could possibly cause external plugins to break. I can add something in NEWS about it if that's a big enough concern.

Towards #4972

### DIFF
--- a/src/iosource/Packet.h
+++ b/src/iosource/Packet.h
@@ -7,6 +7,7 @@
 #include <string>
 
 #if defined(__OpenBSD__)
+
 #include <net/bpf.h>
 using pkt_timeval = bpf_timeval;
 #else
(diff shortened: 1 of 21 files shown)


src/iosource/Packet.h
```diff
@@ -7,15 +7,24 @@
 #include <string>
 
 #if defined(__OpenBSD__)
+
 #include <net/bpf.h>
 using pkt_timeval = bpf_timeval;
+#ifndef DLT_RAW
+#define DLT_RAW 14
```
awelzel: That made me wonder if it's such a good idea after all.

This is in `pcap/dlt.h`, so I guess with a comment it'd be fine to just do the same? And we already had the `ifdef` 🤷‍♂️ 
```
  85 #ifdef __OpenBSD__                                                     
  86 #define DLT_RAW         14      /* raw IP */                         
  87 #else                                                                
  88 #define DLT_RAW         12      /* raw IP */                           
  89 #endif             
```

src/iosource/Packet.h
```diff
@@ -7,15 +7,24 @@
...
+
 using pkt_timeval = struct timeval;
 #include <sys/socket.h>
 #include <sys/time.h>
+#ifndef DLT_RAW
+// nolintnextline(cppcoreguidelines-macro-usage)
+#define DLT_RAW 12
 #endif
 
-#include <pcap.h> // For DLT_ constants
```
awelzel, line 18: I guess the NEWS entry would be good, but also, if someone needs to use `DLT_` constants and didn't have the `<pcap.h>` include, oh well...

PR #4986, opened 2025-11-01
Conditional CI docker builds

### LINKED ISSUE #4794: CI Rework Follow-up: Build nightly containers only after changes
This is a follow-up to https://github.com/zeek/zeek/pull/4735. We should only rebuild/push the containers as part of the nightly builds if a change happened on the master branch. There's no reason to rebuild them otherwise.

timwoj:
This changes how we build/tag/export the docker images on CI to conditionally check whether the Git SHA has changed since we last built the images, and skip various steps if that condition is true. This will still attempt to push the image upstream, but at least won't go through the effort of building everything if nothing has changed.

Fixes #4794

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -728,40 +728,78 @@ docker_build_template: &DOCKER_BUILD_TEMPLATE
   prepare_builder_script: |
     set -x
     mkdir -p ${BUILDER_IMAGE_CACHE_DIR} ${ZEEK_IMAGE_CACHE_DIR}
-    if [ -f ${BUILDER_IMAGE_CACHE_DIR}/builder.zst ]; then zstd -d < ${BUILDER_IMAGE_CACHE_DIR}/builder.zst | docker load; fi
-    if [ -f ${BUILDER_IMAGE_CACHE_DIR}/final.zst ]; then zstd -d < ${BUILDER_IMAGE_CACHE_DIR}/final.zst | docker load; fi
-    cd docker && docker build \
+    if [ -f ${BUILDER_IMAGE_CACHE_DIR}/builder.zst ]; then
+      zstd -d <${BUILDER_IMAGE_CACHE_DIR}/builder.zst | docker load
+      LAST_BUILDER_IMAGE_ID=$(docker inspect -f "{{ .Id }}" zeek-builder:latest 2>/dev/null)
+    fi
+
+    if [ -f ${BUILDER_IMAGE_CACHE_DIR}/final.zst ]; then
+      zstd -d <${BUILDER_IMAGE_CACHE_DIR}/final.zst | docker load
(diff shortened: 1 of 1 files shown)


docker/final.Dockerfile
```diff
@@ -7,6 +7,12 @@ FROM debian:13-slim
 # failures more easily.
 SHELL ["/bin/sh", "-x", "-c"]
 
+# Used to store the SHA for the git commit that generated this image into
+# the image itself. This allows using `docker inspect` to determine what SHA
+# is related to the image.
+ARG GIT_COMMIT
+LABEL git_commit=$GIT_COMMIT
```
awelzel: Should we use an OCI suggested label here instead?

https://www.docker.com/blog/docker-best-practices-using-tags-and-labels-to-manage-docker-image-sprawl/

`org.opencontainers.image.revision`

.cirrus.yml
```diff
@@ -744,7 +778,6 @@ amd64_container_image_docker_builder:
 container_image_manifest_docker_builder:
   cpu: 1
   << : *ONLY_IF_RELEASE_TAG_NIGHTLY
-  skip: $CIRRUS_CHANGE_IN_REPO == $CIRRUS_LAST_GREEN_CHANGE
```
awelzel: Still wondering if `changesInclude()` here would do.

timwoj: As discussed separately, the docs for `changesInclude` say:

> changesInclude function can be very useful for skipping some tasks when no changes to sources have been made since the last successful Cirrus CI build.

This is effectively what I did in https://github.com/zeek/zeek/pull/4944 that we later reverted. `changesInclude` n nightly builds wouldn't return any changes after a regular `master` build because the last build for that SHA succeeded.

timwoj:
Avoiding rebuilding the images here might not be necessary after some further exploration. We cache container images using Cirrus' caching every time the `{amd64/arm64}_container_image` jobs run. This means that if `master` runs, the `amd64_container_image` task ran and cached the image, then it won't rebuild it again during the nightly. The amd64 image is the same from one nightly build to the next, assuming nothing changed in between them.

Unfortunately because we use the same key for the container image caches for every type of build, we run the chance of having to rebuild the one from master if we ran a PR build between the last merge to master and the next nightly build. That's assuming the PR build had the "CI: Cluster Testing" tag on it. Maybe the cache should use one key for `master` and one for PRs?

awelzel:
> Maybe the cache should use one key for master and one for PRs?

My understanding is that PR runs [do not clobber the cache ](https://cirrus-ci.org/guide/writing-tasks/#cache-instruction )of the main/master branch:

> Caching for Pull Requests
> Tasks for PRs upload caches to a separate caching namespace to not interfere with caches used by other tasks. But such  > PR tasks can read all caches even from the main caching namespace for a repository.


The current idea of this branch is to check cached images in the Cirrus cache with the SHA1 of head? I think that'd work. Given that this is only about `zeek/zeek-dev`, it might make sense to fetch/inspect the last pushed image that is in Docker's registry and key rebuilding off of that, but that can certainly be a next step once we push images including the Git information.

PR #4987, opened 2025-11-02
Enable clang-tidy performance-unnecessary-value-param checker

timwoj:
This enables clang-tidy's [performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-value-param.html) checker and fixes a lot of the findings from it. I explicitly disabled anything from `zeek::plugin::HookArgument` because all of them would be API breaks. I also disabled anything from `zeek::*Ptr` because there's a TON of those. We can revisit them later. This PR catches the rest.

### DIFF
--- a/src/cluster/Telemetry.cc
+++ b/src/cluster/Telemetry.cc
@@ -61,7 +61,7 @@ std::vector<std::string_view> to_label_names_vec(const LabelList& static_label_l
 
 InfoTelemetry::InfoTelemetry(std::string_view name, const LabelList& static_labels, std::string_view prefix) {
     if ( name != "core" && name != "websocket" )
-        zeek::reporter->FatalError("name can only be backend or websocket, got '%s'", std::string(name).c_str());
+        zeek::reporter->FatalError("name can only be core or websocket, got '%s'", std::string(name).c_str());
 
     std::string out_name = util::fmt("cluster_%s_outgoing_events", std::string(name).c_str());
     std::string in_name = util::fmt("cluster_%s_incoming_events", std::string(name).c_str());
@@ -236,7 +236,7 @@ void DebugTelemetry::OnIncomingEvent(std::string_view topic, std::string_view ha
 // the given backend with it.
 void configure_backend_telemetry(Backend& backend, std::string_view name, const LabelList& static_labels) {
     if ( name != "core" && name != "websocket" )
-        zeek::reporter->FatalError("name can only be backend or websocket, got '%s'", std::string(name).c_str());
+        zeek::reporter->FatalError("name can only be core or websocket, got '%s'", std::string(name).c_str());
 
     static const auto& info = zeek::id::find_val<zeek::EnumVal>("Cluster::Telemetry::INFO");
     static const auto& verbose = zeek::id::find_val<zeek::EnumVal>("Cluster::Telemetry::VERBOSE");


src/analyzer/Analyzer.cc
```diff
@@ -466,7 +466,7 @@ Analyzer* Analyzer::FindChild(ID arg_id) {
     return nullptr;
 }
 
-Analyzer* Analyzer::FindChild(zeek::Tag arg_tag) {
+Analyzer* Analyzer::FindChild(zeek::Tag arg_tag) { // NOLINT(performance-unnecessary-value-param)
```
awelzel: Maybe leave a comment here (or in Analyzer.h) that we NOLINT to not break the API. I see the SessionAdapters implement `FindChild()` and trying to go to `const zeek::Tag& tag` breaks them.

I'd probably be okay breaking this one (and AnalyzerConfirmation(), AnalyzerViolation()) - are we expecting an external analyzer to override FindChild() ? - but that's a longer discussion. I did see `zeek::Tag` in microbenchmarks and was thinking of using refs before, but was a bit too timid to even propose back then.

src/cluster/Telemetry.cc
```diff
@@ -59,7 +59,7 @@ std::vector<std::string_view> to_label_names_vec(const LabelList& static_label_l
 
 } // namespace
 
-InfoTelemetry::InfoTelemetry(std::string_view name, LabelList static_labels, std::string_view prefix) {
+InfoTelemetry::InfoTelemetry(std::string_view name, const LabelList& static_labels, std::string_view prefix) {
     if ( name != "core" && name != "websocket" )
         zeek::reporter->FatalError("name can only be backend or websocket, got '%s'", std::string(name).c_str());
```
awelzel: Fly-by, sorry ;-)

```suggestion
        zeek::reporter->FatalError("name can only be core or websocket, got '%s'", std::string(name).c_str());
```

src/cluster/Telemetry.cc
```diff
@@ -234,7 +234,7 @@ void DebugTelemetry::OnIncomingEvent(std::string_view topic, std::string_view ha
 
 // Reads Cluster::Telemetry consts, instantiates and appropriate Telemetry instance and configures
 // the given backend with it.
-void configure_backend_telemetry(Backend& backend, std::string_view name, LabelList static_labels) {
+void configure_backend_telemetry(Backend& backend, std::string_view name, const LabelList& static_labels) {
     if ( name != "core" && name != "websocket" )
         zeek::reporter->FatalError("name can only be backend or websocket, got '%s'", std::string(name).c_str());
```
awelzel:
```suggestion
        zeek::reporter->FatalError("name can only be core or websocket, got '%s'", std::string(name).c_str());
```

src/EventTrace.cc
```diff
@@ -689,13 +689,13 @@ void EventTrace::Generate(FILE* f, ValTraceMgr& vtm, const DeltaGenVec& dvec, st
     else {
         auto tm = vtm.TimeConstant(nt);
         fprintf(f, "\tset_network_time(%s);\n", tm.c_str());
-        fprintf(f, "\tevent __EventTrace::%s();\n", successor.c_str());
+        fprintf(f, "\tevent __EventTrace::%.*s();\n", static_cast<int>(successor.size()), successor.data());
```
awelzel: Would probably have reached for `const std::string&` to avoid the static cast and `%.*s`, but mostly nit.

src/Trigger.h
```diff
@@ -65,7 +65,7 @@ class Trigger final : public Obj, public notifier::detail::Receiver {
     // in where the "timeout" parameter appears, and in making the "loc"
     // parameter optional. ("loc" is only used for internal logging when
     // debugging triggers.)
-    Trigger(std::shared_ptr<WhenInfo> wi, const IDSet& globals, std::vector<ValPtr> local_aggrs, double timeout,
+    Trigger(const std::shared_ptr<WhenInfo>& wi, const IDSet& globals, std::vector<ValPtr> local_aggrs, double timeout,
```
awelzel, line 68: Hmm, hmm. Looking at the Trigger constructor's body, could probably deprecate the shared_ptr version altogether and go for `Trigger(const WhenInfo& wi, ...)`. But no need to do that here.

tools/binpac/src/pac_output.h
```diff
@@ -11,7 +11,7 @@ using namespace std;
 
 class OutputException {
 public:
-    OutputException(const char* arg_msg);
+    OutputException(std::string_view arg_msg);
```
awelzel: Nit: I'm blinking at `std::string_view` here and `const std::string&` below. Would it make sense to use the same in either one? Seems OutputException is not anything we need to keep stable. Feel free to ignore, just me blinking ;-)

PR #4988, opened 2025-11-02
 Enable clang-tidy modernize-use-equals-default checker

timwoj:
This enables clang-tidy's modernize-use-equals-default checker which searches for empty and extremely trivlal constructors, destructors, and copy operators and recommends replacing them with `= default` statements instead. I started to fix the code generated by binpac to not generate empty constructors/destructors, but decided in the end to just tag them with NOLINT annotations.

### DIFF
--- a/src/analyzer/protocol/login/Telnet.h
+++ b/src/analyzer/protocol/login/Telnet.h
@@ -9,7 +9,6 @@ namespace zeek::analyzer::login {
 class Telnet_Analyzer : public Login_Analyzer {
 public:
     explicit Telnet_Analyzer(Connection* conn);
-    ~Telnet_Analyzer() override = default;
 
     static analyzer::Analyzer* Instantiate(Connection* conn) { return new Telnet_Analyzer(conn); }
 };
--- a/src/Attr.h
+++ b/src/Attr.h
@@ -67,8 +67,6 @@ public:
     Attr(AttrTag t, ExprPtr e);
     explicit Attr(AttrTag t);
 
-    ~Attr() override = default;
-
     AttrTag Tag() const { return tag; }
 
     const ExprPtr& GetExpr() const { return expr; }
@@ -118,8 +116,6 @@ public:
     Attributes(std::vector<AttrPtr> a, TypePtr t, bool in_record, bool is_global)
         : Attributes(std::move(a), std::move(t), in_record, is_global, false) {}
 
-
-    ~Attributes() override = default;
 
     void AddAttr(AttrPtr a, bool is_redef = false);
 
--- a/src/Expr.h
+++ b/src/Expr.h
@@ -1342,7 +1342,6 @@ extern RecordValPtr coerce_to_record(RecordTypePtr rt, Val* v, const std::vector
 class TableCoerceExpr final : public UnaryExpr {
 public:
     TableCoerceExpr(ExprPtr op, TableTypePtr r, bool type_check = true);
-    ~TableCoerceExpr() override = default;
 
     // Optimization-related:
     ExprPtr Duplicate() override;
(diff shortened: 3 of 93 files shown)


src/analyzer/protocol/login/Telnet.h
```diff
@@ -9,7 +9,7 @@ namespace zeek::analyzer::login {
 class Telnet_Analyzer : public Login_Analyzer {
 public:
     explicit Telnet_Analyzer(Connection* conn);
-    ~Telnet_Analyzer() override {}
+    ~Telnet_Analyzer() override = default;
```
bbannier: Am I missing anything, or are these really completely redundant now (we "override" a base dtr with nothing, i.e., all our dtr does is invoke the base class dtr)?

If possible I think it would be good to get rid of all these which would remove another 180 loc.

.clang-tidy
```diff
@@ -17,7 +17,7 @@ Checks: [-*,
          -bugprone-narrowing-conversions,
          -bugprone-unchecked-optional-access,
          -performance-unnecessary-value-param,
-         -modernize-use-equals-default,
+         modernize-use-equals-default,
```
awelzel: Could you drop that given the `modernize-*` select above?

tools/binpac/src/pac_common.h
```diff
@@ -30,7 +30,7 @@ class Object {
         location = strfmt("%s:%d", filename.c_str(), line_number);
     }
 
-    ~Object() {}
+    virtual ~Object() = default;
```
bbannier, line 33: Ups 😅. Actually surprised this didn't trigger any leak warnings.

src/TunnelEncapsulation.h
```diff
@@ -62,16 +62,7 @@ class EncapsulatingConn {
     /**
      * Copy constructor.
      */
-    EncapsulatingConn(const EncapsulatingConn& other)
```
bbannier, line 65: Could we default the assignment operator below as well?

src/plugin/Plugin.h
```diff
@@ -159,15 +159,7 @@ class Configuration {
...
-    Configuration& operator=(const Configuration& c) {
-        zeek_version = c.zeek_version;
-
-        name = c.name;
-        description = c.description;
-        version = c.version;
-
-        return *this;
-    }
+    Configuration& operator=(const Configuration& c) = default;
```
bbannier, line 139: Let's default the ctrs and move assignment operator above as well.

src/Type.cc
```diff
@@ -499,7 +499,7 @@ TableType::TableType(TypeListPtr ind, TypePtr yield) : IndexType(TYPE_TABLE, std
         RegenerateHash();
 }
 
-TableType::~TableType() {}
+TableType::~TableType() = default;
```
bbannier, line 502: Do you follow a pattern for whether you default in the header, or the impl like here? In the earlier binpac patch and also later here you opted for defaulting in the header which seemed fine to me.

bbannier (approved):

src/Type.cc
timwoj: There were a couple where defaulting in the header caused build failures. When I looked into them, it's because they were forward-delcaring IntrusivePtrs, then using those IntrustivePtrs as member variables. When it went to destroy them in the default destructor, it didn't have the full definition of IntrusivePtr, didn't understand how to Unref, and failed to build. Moving the defaulting to the cc file fixes it because the cc file includes IntrusivePtr.h.

tools/binpac/src/pac_common.h
timwoj: Yeah, that was definitely surprising.

PR #4993, opened 2025-11-04
Enable clang-tidy readability-container-size-empty checker, fix findings

timwoj:
This is something I comment on a lot of PRs, so I figure we should enable the checker so CI can complain about it instead. This enables the [readability-container-size-empty checker](https://clang.llvm.org/extra/clang-tidy/checks/readability/container-size-empty.html) from clang-tidy and fixes the findings from it.

### DIFF
--- a/src/iosource/Manager.cc
+++ b/src/iosource/Manager.cc
@@ -220,7 +220,7 @@ void Manager::Poll(ReadySources* ready, double timeout, IOSource* timeout_src) {
 
             // If we added a source that is the same as the passed timeout_src, take
             // note as to avoid adding it twice.
-            timeout_src_added |= ! ready->empty() ? ready->back().src == timeout_src : false;
+            timeout_src_added |= ready->empty() ? false : ready->back().src == timeout_src;
         }
 
         // A timeout_src with a zero timeout can be considered ready.
--- a/src/Expr.cc
+++ b/src/Expr.cc
@@ -1,5 +1,4 @@
 // See the file "COPYING" in the main distribution directory for copyright.
-
 #include "zeek/Expr.h"
 
 #include "zeek/DebugLogger.h"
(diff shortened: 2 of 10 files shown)


src/iosource/Manager.cc
```diff
@@ -220,7 +220,7 @@ void Manager::Poll(ReadySources* ready, double timeout, IOSource* timeout_src) {
 
             // If we added a source that is the same as the passed timeout_src, take
             // note as to avoid adding it twice.
-            timeout_src_added |= ready->size() > 0 ? ready->back().src == timeout_src : false;
+            timeout_src_added |= ! ready->empty() ? ready->back().src == timeout_src : false;
```
bbannier: Nit: probably easier to read, I'd argue even in general.

```suggestion
            timeout_src_added |= ready->empty() ? false : ready->back().src == timeout_src;
```

There are some more instances below.

tools/bifcl/builtin-func.y
```diff
@@ -686,7 +686,11 @@ body_start:	TOK_LPB c_code_begin
 
 			if ( ! var_arg )
 				{
-				fprintf(fp_func_def, "\tif ( %s->size() != %d )\n", arg_list_name, argc);
+                if ( argc == 0 )
+                    fprintf(fp_func_def, "\tif ( ! %s->empty() )\n", arg_list_name);
+                else
+                    fprintf(fp_func_def, "\tif ( %s->size() != %d )\n", arg_list_name, argc);
```
bbannier: IMO totally fine to suppress this one since this makes the code more complicated, not easier.

bbannier (approved):
Thanks, big fan of this one.

I left a comment on ternaries below. There's a similar pattern `if (! empty) .. else ..` which one could rewrite to `if (empty) .. else ..`, but like the ternaries this could have a slight performance impact (either direction), depending on which branch was hot.

PR #5000, opened 2025-11-07
Enable a couple of clang-tidy checkers without findings

timwoj:
- `bugprone-undefined-memory-manipulation` and `bugprone-suspicious-include` were both explicitly disabled in .clang-tidy, except neither of them report any findings when enabled. 
- `modernize-type-traits` was disabled by fc56eeaf390d31049ef8db0dd2a531d6680609dd by mistake. 
- `bugprone-empty-catch` got a better comment about why it's disabled.
- `bugprone-switch-missing-default-case` is not because of bifcl but because of binpac.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,8 +6,6 @@ Checks: [-*,
          readability-container-contains,
          readability-container-size-empty,
 
-         -modernize-type-traits,
-
          # Enable a very limited number of the cppcoreguidelines checkers.
          # See the notes for some of the rest of them below.
          cppcoreguidelines-macro-usage,


.clang-tidy
```diff
@@ -31,17 +29,15 @@ Checks: [-*,
          # double-underscores in the name.
          -bugprone-reserved-identifier,
 
-         # bifcl generates almost every switch statement without a default case
-         # and so this one generates a lot of warnings.
+         # binpac generates switch statements without default cases if case
+         # blocks in the pac files don't have default statements. This one
+         # generates a lot of warnings.
```
evantypanski, line 34: I personally like the `bugprone-switch-missing-default-case` check and think we should explore enabling some of these checks that are noisy on generated code, but nice to have around. I ran it in the Zeek tree and there are a few TPs that would be nice to fix and keep fixed

maybe we can use nested `.clang-tidy` configs (like in `src/`) or find a filter of some kind to do this? I haven't experimented with it but it seems possible, albeit not super high value

doesn't have to be here, just something I noticed

Issue #4961, opened 2025-10-23
ASAN CI build sporadically gets OOMKilled


timwoj:
Filing this because i've seen it more than once. The ASAN build on Cirrus will get OOM killed sporadically. https://cirrus-ci.com/task/6369774600454144 is one example. Unfortunately when this happens it doesn't save the log, so we don't have any indication what it was doing at the time.

awelzel:
I've seen it quite a bit, too. I wonder a if the reliance on `calloc()`'s lazy memory initialization within libkqueue applies to ASAN, or whether we could try to reduce open files limit and see if this helps.

timwoj:
I sent an email to Cirrus asking whether the OOM'd task can save the logs.

timwoj:
This is happening a lot more recently. The team at Cirrus fixed it so it at least logs the last few minutes before the process gets OOMKilled. It's not terribly useful though: https://cirrus-ci.com/task/4833787088994304?logs=test#L0

awelzel:
> I've seen it quite a bit, too. I wonder a if the reliance on calloc()'s lazy memory initialization within libkqueue applies to ASAN, or whether we could try to reduce open files limit and see if this helps.

I've opened a PR to force open files to 256. Think we could try? The interactive console in the Cirrus Web UI showed 1 mio open files default.

timwoj:
It looks like #5001 stabilized it. There haven't been any OOMs since that was merged. We can either close this or leave it open to explore whether we can fix libkqueue better.

timwoj:
Scratch that, https://cirrus-ci.com/task/5285899841306624 got OOMKilled.

PR #5001, opened 2025-11-10
ci: Force open files limit to 256

awelzel:
References #4961.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -235,6 +235,13 @@ env:
   # a solution for the mtime pruning above.
   ZEEK_CCACHE_EPOCH: 2
 
+  # Set ulimit open files limit when running ./ci/test.sh to this value.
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -235,6 +235,13 @@ env:
   # a solution for the mtime pruning above.
   ZEEK_CCACHE_EPOCH: 2
 
+  # Set ulimit open files limit when running ./ci/test.sh to this value.
+  # In November 2025, the default open files limit on Cirrus is 1 mio and
+  # the assumption here is that libkqueue's calloc optimization for large
+  # open file limits doesn't apply to ASAN runtimes, resulting in regular
+  # but sporadic OOM failures.
+  ZEEK_TEST_ULIMIT_OPEN_FILES: 256
```
awelzel, line 243: I actually think we should fix libkqueue to use a less memory hungry data structure, but lets first see if this is reducing the number of OOMs.

PR #5002, opened 2025-11-10
x509: Implement deduplication without Broker stores

awelzel:
This uses explicit Cluster::publish() through the manager and a Cluster::node_up() handler to pre-seed a worker with anything that the manager knows should be de-duplicated.

Relates to #4762


testing/btest/scripts/base/files/x509/log-caching-cluster.test
```diff
@@ -5,6 +5,7 @@
 # @TEST-PORT: BROKER_WORKER2_PORT
 #
 # @TEST-EXEC: cp $FILES/broker/cluster-layout.zeek .
+# @TEST-EXEC: zeek --parse-only %INPUT
```
0xxon: Is this here for a reason, or is this a leftover from debugging?

testing/btest/scripts/base/files/x509/log-caching-cluster-zeromq.test
```diff
@@ -0,0 +1,76 @@
+# Test that certificate log deduplication works in clustered mode for ZeroMQ.
+#
+# @TEST-PORT: XPUB_PORT
+# @TEST-PORT: XSUB_PORT
+# @TEST-PORT: LOG_PULL_PORT
+#
+# @TEST-EXEC: cp $FILES/zeromq/cluster-layout-no-logger.zeek cluster-layout.zeek
+# @TEST-EXEC: cp $FILES/zeromq/test-bootstrap.zeek zeromq-test-bootstrap.zeek
+#
+# @TEST-EXEC: zeek --parse-only %INPUT
```
0xxon: Is this here for a reason, or is this a leftover from debugging? (Question repeated below, I noticed it first there)

scripts/base/files/x509/main.zeek
```diff
@@ -101,7 +101,17 @@ export {
 	## to work cluster wide, but come at a slightly higher cost of memory and inter-node-communication.
 	##
 	## This setting is ignored if Zeek is run in standalone mode.
```
0xxon: Should we perhaps add a note here that this is ignored if known_log_certs_enable_publish is set to true?

testing/btest/scripts/base/files/x509/log-caching-cluster-zeromq.test
```diff
@@ -0,0 +1,76 @@
+# Test that certificate log deduplication works in clustered mode for ZeroMQ.
```
0xxon: Should this be prefixed with @TEST-DOC:?

testing/btest/scripts/base/files/x509/log-caching-cluster.test
awelzel: I've started to put them in a few cluster tests because the `btest-bg-run` / `btest-bg-wait` sequence doesn't fail quickly on syntax errors while working on the tests (or later when debugging a flaky test).

I'll drop them. It's kind of my own approach to working with the tests.

scripts/base/files/x509/main.zeek
awelzel: Yeah, good call. I added a small note.

awelzel:
> This in principle looks good to me 

Thanks! I've had a few more thoughts after creating the PR:

1) Wondering if the known_log_certs `&create_expire` should actually be `&write_expire` such that the expiry time is updated on all nodes. There's already drift in expiry times, particularly in the face of node restarts. Happy to address this one subsequently, but maybe you have a reaction / knowledge how that worked with Broker stores? Did they end up using the same create expiry time behind the scenes?

2) Batching as done for the notice framework recently, but would only go there if it's actually needed and also might want to think about a framework/helper that does the batching in a more generic fashion.

0xxon:
> > This in principle looks good to me
> 
> Thanks! I've had a few more thoughts after creating the PR:
> 
>     1. Wondering if the known_log_certs `&create_expire` should actually be `&write_expire` such that the expiry time is updated on all nodes. There's already drift in expiry times, particularly in the face of node restarts. Happy to address this one subsequently, but maybe you have a reaction / knowledge how that worked with Broker stores? Did they end up using the same create expiry time behind the scenes?

I am not sure how this works behind the scenes. Write expire seems to make more sense - however it probably is not worth it if it comes at an additional communication overhead to synchronize the expiry times. I have no clue anymore how this works behind the scenes :)

>     2. Batching as done for the notice framework recently, but would only go there if it's actually needed and also might want to think about a framework/helper that does the batching in a more generic fashion.

Sounds good to me.

PR #5005, opened 2025-11-10
CI: Make cluster_test and amd64_container_image have the same conditions

timwoj:
Cirrus has been complaining about this recently:

```
task "cluster_testing" depends on task "amd64_container_image", but their only_if conditions are different
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -842,10 +842,6 @@ public_ecr_cleanup_docker_builder:
 cluster_testing_docker_builder:
   cpu: *CPUS
   memory: *MEMORY
-  only_if: >
-    ( $CIRRUS_CRON == '' ) &&
-    ( ( $CIRRUS_PR != '' && $CIRRUS_BRANCH !=~ 'dependabot/.*' ) ||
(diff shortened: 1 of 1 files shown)


0xxon (approved):

.cirrus.yml
```diff
@@ -869,7 +869,7 @@ cluster_testing_docker_builder:
       path: "testing/external/zeek-testing-cluster/.tmp/**"
   depends_on:
     - amd64_container_image
-  << : *ONLY_IF_PR_RELEASE_AND_NIGHTLY
+  << : *ONLY_IF_PR_MASTER_RELEASE_NIGHTLY
```
awelzel, line 868: Did this already override the explicit `only_if:` on line 845? If yes, maybe ditch the one above? 🤷‍♂️

timwoj: Yes, the existing only_if can go away.

timwoj:
I merged an older version of this in 850f12ff3827969ec21fb0a2a41e9fb41cacab2e. I'll fix the other minor change that was in the force-push in a separate commit.

PR #5011, opened 2025-11-11
Fix findings from newer versions of clang-tidy

timwoj:
I have clang-tidy 22.0.0 on a local machine that was reporting clang-tidy findings we haven't fixed yet. Some of these are from new checkers that weren't in previous versions, and some are findings for existing checkers we had enabled but weren't being caught by older versions.

This also fixes the current breakage on master.

### DIFF
--- a/src/IntrusivePtr.h
+++ b/src/IntrusivePtr.h
@@ -300,6 +300,7 @@ auto operator!=(const zeek::IntrusivePtr<T>& x, const zeek::IntrusivePtr<U>& y) 
 
 namespace std {
 template<class T>
+// NOLINTNEXTLINE(bugprone-std-namespace-modification)
 struct hash<zeek::IntrusivePtr<T>> {
     // Hash of intrusive pointer is the same as hash of the raw pointer it holds.
     size_t operator()(const zeek::IntrusivePtr<T>& v) const noexcept { return std::hash<T*>{}(v.get()); }
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -1298,7 +1298,7 @@ function paraglob_init%(v: any%) : opaque of paraglob
 
 	try
 		{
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -489,10 +489,17 @@ function (zeek_create_bif_autogen_stubs)
...
     endforeach ()
 endfunction ()
 
+function (zeek_finalize_bif_autogen_stubs)
+    foreach (fpath ${ARGN})
+        file(APPEND "${CMAKE_BINARY_DIR}/${fpath}" "// NOLINTEND(bugprone-suspicious-include)\n")
+    endforeach ()
+endfunction ()
+
 # Note: these files are filled from BifCl.cmake via `file(APPEND ...)`.
```
ckreibich, line 503: Is that comment still correct given the function above?

CMakeLists.txt
```diff
@@ -1385,6 +1392,9 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake_templates/zeek-config-paths.h.i
                ${CMAKE_CURRENT_BINARY_DIR}/zeek-config-paths.h)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/zeek-config-paths.h DESTINATION include/zeek)
 
+zeek_finalize_bif_autogen_stubs(src/__all__.bif.cc src/__all__.bif.init.cc
```
ckreibich, line 1397: Maybe add a comment here why this is the right place to call this? I couldn't immediately tell which of the above make any modifications so the finalization can't come earlier.

CMakeLists.txt
ckreibich: Ah, I get it now — filled _after_ this call. Let me close this comment, all good.

PR #5012, opened 2025-11-11
Allow name mismatches between Spicy and Zeek

### LINKED ISSUE #5008: Spicy Zeek integration: records that contain elements named `signed` and `unsigned` don't get mapped correctly.
I stumbled across this while playing around with SNMP and spicy. One of the records for the SNMP protocol has two elements that are called `signed` and `unsigned`.

Creating a struct on the spicy side, and trying to send an event containing the struct to Zeek (which should convert to the record) results in the following error:

```
1367411051.972852 analyzer error in /Users/johanna/zeek/spicy-bug/test.spicy, line 14: Event parameter mismatch, mismatch in field name: expected 'signed_', found 'signed'
```

Looking at the generated code, the `signed` member is indeed converted to `signed_` by spicy - probably to prevent errors with reserved names. However, it seems like this is something that should be undone when mapping to Zeek events.

Full sources to reproduce:

Zeek file:

```
type laType: record {
	a: int;
	signed: int &optional;
	unsigned: count &optional;
};

event test_event(la: laType)
	{
	print la;
	}
```

spicy file:
```
module Test;

import spicy;
import zeek;

type TestValue = struct {
	a: int64;
	signed: optional<int64> &default=Null;
	unsigned: optional<uint64> &default=Null;
};

public function make_test_value(p: Packet): TestValue {
	local test: TestValue = [$a=5];
	return test;
}

public type Packet = unit {
};
```

evt file:
```
protocol analyzer spicy::Test over UDP:
    parse with Test::Packet,
    port 69/udp;

import Test;

on Test::Packet -> event test_event(Test::make_test_value(self));
```

Run with any pcap that contains traffic on port 69 udp.

evantypanski:
Fixes #5008 

The main motivating factor here is because reserved keywords get their names changed, but overall the restriction that Zeek and Spicy records must exactly align doesn't seem necessary

@rsmmr any thoughts on this? It's a bit heavy-handed, but also think name mismatches are kind of just a "nice to have". It could be a bit more targeted too by ignoring trailing underscores when doing the name comparison.


rsmmr:
Oh, sorry, I saw this PR only after I created https://github.com/zeek/spicy/pull/2198. However, I think that is still the better fix because it's really a more general Spicy-side issue. And I'd be reluctant to remove the Zeek-side name validation as getting names wrong/misaligned somehow seems like the more common error.

But I'd take the test case here when we merge the fix.

evantypanski:
I'll close this then, feel free to steal the test when bumping spicy. I like that solution more too

PR #5013, opened 2025-11-12
quic,postgresql,websocket: Set zeek_init() with Log::create_stream() call to priority=5

awelzel:
Reported by Zach Robinette on Slack. Oversight by me following the common priority rules for Log::create_stream()


evantypanski:
quick grep reveals a couple other cases, like postgresql and websocket, too

awelzel:
> quick grep reveals a couple other cases, like postgresql and websocket, too

Probably all copy-paste-by-the-same-guy. Going to include them in this PR. I'm forth and back if this is worth a linter :-)

PR #5014, opened 2025-11-13
Allow disabling of escaping non-printable UTF-8 characters in JSON output

### LINKED ISSUE #3948: Discrepancy between JSON and TSV logs for non-printable bytes
Currently, Zeek uses hex escapes (\x00) to escape non-printable characters (and possibly others) in both TSV and JSON logs.
For TSV, this is fine. However, because JSON requires strict UTF-8 adherence, all backslashes are escaped again to prevent errors (`\x` -> `\\x`). Unfortunately, this is a destructive transformation that makes it impossible to reliably distinguish between text and a byte representation in edge cases.

Take for example terminal control character `07` (BEL). When we include this character in a (clearly invalid) HTTP URI `/non_printable_[byte 07]`, Zeek would log the following:

TSV: `/non_printable_\x07`
JSON: `/non_printable_\\x07`

Now include the (literal) text `\x07` instead of the control character in the HTTP URI: `/non_printable_literal_\x07`. Zeek logs the following:

TSV: `/non_printable_literal_\\x07`
JSON: `/non_printable_literal_\\x07`

This results in identical output for both cases in JSON logs and means that we are unable to distinguish between the byte `\x07` and text `\x07`. I can for example no longer reliably transform the escapes back to their original bytes.

This issue is caused by the fact that `\\x07` is already valid JSON and does not need to be escaped again (case 2), as opposed to `\x07` which has to become `\\x07` (case 1).

One solution would be to "over-escape" all backslashes regardless of context, so `\x07` becomes `\\x07` and `\\x07` becomes `\\\\x07`. I'm not familiar enough with Zeek internals to propose a proper solution :)

I've used HTTP for the examples because it was straightforward to create a sample PCAP, but this issue applies to every string type field in every Zeek log.

Attached are:
  - sample PCAP with four HTTP requests that highlight the discrepancy
  - JSON and TSV logs from Zeek 6 and 7 (identical besides the UIDs)

[pcap.zip](https://github.com/user-attachments/files/17084149/pcap.zip)
[zeek6_http.json](https://github.com/user-attachments/files/17084151/zeek6_http.json)
[zeek6_http.tsv](https://github.com/user-attachments/files/17084153/zeek6_http.tsv.txt)
[zeek7_http.json](https://github.com/user-attachments/files/17084150/zeek7_http.json)
[zeek7_http.tsv](https://github.com/user-attachments/files/17084152/zeek7_http.tsv.txt)

If memory serves, this change was introduced in Zeek 3 when the log encoding changed from ASCII to UTF-8 (#458). #596 might also be related.

timwoj:
This PR changes the escaping methods we use for UTF8 data (notably for JSON) to allow callers to disable the escaping of non-printable ASCII control characters. In JSON output, this results in those characters being inserted as straight Unicode characters. The default for JSON output is to not escape them, but for all other callers the default is for them to continue being escaped. Invalid UTF-8 sequences continue to be escaped as they were before.

It also fixes a bug with the existing `escape_printable_controls` argument, which was doing the opposite of what was intuitive based on the name of the argument. It does this by deprecating the existing `json_escape_utf8` method and replacing it with a new `escape_utf8` method that does the right thing.

Fixes #3948


src/util.h
```diff
@@ -532,18 +532,38 @@ void zeek_strerror_r(int zeek_errno, char* buf, size_t buflen);
...
  * @param val_size the length of the character data
+ * @param escape_printable_controls flag for whether printable control characters
+ * (e.g. newlines, tabs, etc) should be escaped in the output. Defaults to false.
+ * @param escape_other_controls flag for whether non-printable control characters
+ * (e.g. nulls, bells, etc) should be escaped in the output.
  * @return the escaped string
  */
-std::string json_escape_utf8(const char* val, size_t val_size, bool escape_printable_controls = true);
+std::string escape_utf8(const char* val, size_t val_size, bool escape_printable_controls = false,
+                        bool escape_other_controls = true);
```
bbannier, line 543: Any reason we cannot have just a single function?

```cpp
std::string escape_utf8(std::string_view val, bool escape_printable_controls = false,
                        bool escape_other_controls = true);
```

timwoj: I tested that but it was having problems with cases where strings had embedded nulls. I'll look again to see if there's a good way around it other than explicitly creating `string_view`s from the data in that case.

bbannier: Something like this seemed to work for me, at least with the test suite.

```diff
diff --git a/src/Desc.cc b/src/Desc.cc
index 521ee1df03..3b0bb448e5 100644
--- a/src/Desc.cc
+++ b/src/Desc.cc
@@ -249,7 +249,8 @@ void ODesc::AddBytes(const void* bytes, size_t n) {
 
         if ( esc_start != nullptr ) {
             if ( utf8 ) {
-                std::string result = util::escape_utf8(s, esc_start - s, true);
+                assert(esc_start >= s);
+                std::string result = util::escape_utf8({s, static_cast<size_t>(esc_start - s)}, true);
                 AddBytesRaw(result.c_str(), result.size());
             }
             else
@@ -260,7 +261,8 @@ void ODesc::AddBytes(const void* bytes, size_t n) {
         }
         else {
             if ( utf8 ) {
-                std::string result = util::escape_utf8(s, e - s, true);
+                assert(e >= s);
+                std::string result = util::escape_utf8({s, static_cast<size_t>(e - s)}, true);
                 AddBytesRaw(result.c_str(), result.size());
             }
             else
diff --git a/src/threading/formatters/JSON.cc b/src/threading/formatters/JSON.cc
index 34ea7de729..757ca6d977 100644
--- a/src/threading/formatters/JSON.cc
+++ b/src/threading/formatters/JSON.cc
@@ -144,7 +144,9 @@ void JSON::BuildJSON(zeek::json::detail::NullDoubleWriter& writer, Value* val, c
         case TYPE_STRING:
         case TYPE_FILE:
         case TYPE_FUNC: {
-            writer.String(util::escape_utf8(val->val.string_val.data, val->val.string_val.length, false, false));
+            assert(val->val.string_val.length >= 0);
+            writer.String(util::escape_utf8({val->val.string_val.data, static_cast<size_t>(val->val.string_val.length)},
+                                            false, false));
             break;
         }
 
diff --git a/src/util.cc b/src/util.cc
index a4b7452e33..2d0b61040e 100644
--- a/src/util.cc
+++ b/src/util.cc
@@ -2355,12 +2355,12 @@ static bool check_ok_utf8(const unsigned char* start, const unsigned char* end)
     return true;
 }
 
-string escape_utf8(const string& val, bool escape_printable_controls, bool escape_other_controls) {
-    return escape_utf8(val.c_str(), val.size(), escape_printable_controls, escape_other_controls);
-}
+string escape_utf8(string_view val_, bool escape_printable_controls, bool escape_other_controls) {
+    const auto* val = val_.data();
+    auto val_size = val_.size();
+    const auto* val_data = reinterpret_cast<const unsigned char*>(val);
 
-string escape_utf8(const char* val, size_t val_size, bool escape_printable_controls, bool escape_other_controls) {
-    auto val_data = reinterpret_cast<const unsigned char*>(val);
+    // FIXME: Clean up mixed uses of `val` and `val_data` below.
 
     // Reserve at least the size of the existing string to avoid resizing the string in the
     // best-case scenario where we don't have any multi-byte characters. We keep two versions of
diff --git a/src/util.h b/src/util.h
index 7d57715f12..f409893f22 100644
--- a/src/util.h
+++ b/src/util.h
@@ -85,6 +85,8 @@ inline constexpr std::string_view path_list_separator = ";";
 inline constexpr std::string_view path_list_separator = ":";
 #endif
 
+#include <string_view>
+
 #include "zeek/UID.h"
 #include "zeek/util-types.h"
 
@@ -538,21 +540,7 @@ void zeek_strerror_r(int zeek_errno, char* buf, size_t buflen);
  * (e.g. nulls, bells, etc) should be escaped in the output.
  * @return the escaped string
  */
-std::string escape_utf8(const std::string& val, bool escape_printable_controls = false,
-                        bool escape_other_controls = true);
-
-/**
- * Escapes bytes in a string that are not valid UTF8 characters with \xYY format. Used
- * by the JSON writer and BIF methods.
- * @param val the character data to be escaped
- * @param val_size the length of the character data
- * @param escape_printable_controls flag for whether printable control characters
- * (e.g. newlines, tabs, etc) should be escaped in the output. Defaults to false.
- * @param escape_other_controls flag for whether non-printable control characters
- * (e.g. nulls, bells, etc) should be escaped in the output.
- * @return the escaped string
- */
-std::string escape_utf8(const char* val, size_t val_size, bool escape_printable_controls = false,
+std::string escape_utf8(std::string_view val, bool escape_printable_controls = false,
                         bool escape_other_controls = true);
 
 [[deprecated("Remove in v9.1. Use escape_utf8 instead.")]]

```

timwoj: Looks like that works, thanks. The one test I was concerned about was from the unit tests:

```
    CHECK(escape_utf8({"\f\n\r\t\x00\x15", 6}) == "\f\n\r\t\\x00\\x15");
```

It does indeed fail until the input is wrapped in `{}` and then it's fine. I pushed a fixup with that.

ckreibich:
I wanted to get a sense of the kinds of changes this will produce in our logs, so as an experiment I ran our external testsuites, switched their logging to JSON, and compared the results. They're encouraging:

#### zeek-testing

Of 603,479 lines 417 (0.06%) changed. These changes are concentrated in dns.log (only in the CTU trace), analyzer.log, and ntp.log (only in m57-long). Here are examples of each change:

analyzer.log: the `packet_segment` field.

```diff
- {"ts":XXXXXXXXXX.XXXXXX,"analyzer_kind":"protocol","analyzer_name":"FTP","uid":"C5kXV01CdG9vMwIfbi","id.orig_h":"2001:470:1f05:17a6:d69a:20ff:fefd:6b88","id.orig_p":55257,"id.resp_h":"2001:6a8:a40::21","id.resp_p":21,"proto":"tcp","failure_reason":"non-numeric reply code","failure_data":"SSH-2.0-mod_sftp/0.9.7","packet_segment":"\\xd4\\x9a \\xfdk\\x88\\x00\\x80\\xc8\\xb9\\xc2\\x06\\x86\\xdd`\\x00\\x00\\x00\\x00t\\x067 \\x01\\x06\\xa8\n@\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00! \\x01\\x04p\\x1f\\x05\\x17\\xa6\\xd6\\x9a \\xff\\xfe\\xfdk\\x88\\x00\\x15\\xd7\\xd9\\x1e\\xb2l\\x8b]\\x14\\xc8\\x83P\\x18\\x00\t\\xbb\\x12<...>/n\\x00\\x00\\x00D\b\\x01\\x00\\x00\\x00\f\\x00\\x00\\x00)Maximum connections for host/user reached\\x00\\x00\\x00\\x05en-USa\\xc5\\xbbM\\x1e\\xe6\\x87\\x89"}
+ {"ts":XXXXXXXXXX.XXXXXX,"analyzer_kind":"protocol","analyzer_name":"FTP","uid":"C5kXV01CdG9vMwIfbi","id.orig_h":"2001:470:1f05:17a6:d69a:20ff:fefd:6b88","id.orig_p":55257,"id.resp_h":"2001:6a8:a40::21","id.resp_p":21,"proto":"tcp","failure_reason":"non-numeric reply code","failure_data":"SSH-2.0-mod_sftp/0.9.7","packet_segment":"\\xd4\\x9a \\xfdk\\x88\u0000\\x80\\xc8\\xb9\\xc2\u0006<...>/u00067 \u0001\u0006<...>/u0000! \u0001\u0004p\u001F\u0005\u0017\\xa6\\xd6\\x9a \\xff\\xfe<...>/u0015\\xd7\\xd9\u001E\\xb2l\\x8b]\u0014<...>/t<...>/u0000)Maximum connections for host/user reached\u0000\u0000\u0000\u0005en-USa\\xc5\\xbbM\u001E\\xe6\\x87\\x89"}
```
(Annoyingly, a diff canonifier kicked in there for some reason.)


dns.log: NetBIOS nameservice names

```diff
- {"ts":XXXXXXXXXX.XXXXXX,"uid":"C0HBSa4fOmvxbjnzY5","id.orig_h":"192.168.1.101","id.orig_p":137,"id.resp_h":"192.168.1.255","id.resp_p":137,"proto":"udp","trans_id":18939,"query":"\\x01\\x02__MSBROWSE__\\x02","qclass":1,"qclass_name":"C_INTERNET","qtype":32,"qtype_name":"NIMLOC","AA":false,"TC":false,"RD":true,"RA":false,"Z":1,"rejected":false,"opcode":0,"opcode_name":"netbios-query","original_query":"ABACFPFPENFDECFCEPFHFDEFFPFPACAB"}
+ {"ts":XXXXXXXXXX.XXXXXX,"uid":"C0HBSa4fOmvxbjnzY5","id.orig_h":"192.168.1.101","id.orig_p":137,"id.resp_h":"192.168.1.255","id.resp_p":137,"proto":"udp","trans_id":18939,"query":"\u0001\u0002__MSBROWSE__\u0002","qclass":1,"qclass_name":"C_INTERNET","qtype":32,"qtype_name":"NIMLOC","AA":false,"TC":false,"RD":true,"RA":false,"Z":1,"rejected":false,"opcode":0,"opcode_name":"netbios-query","original_query":"ABACFPFPENFDECFCEPFHFDEFFPFPACAB"}
```

ntp.log: raw bytes in the `ref_id` field:

```diff
- {"ts":XXXXXXXXXX.XXXXXX,"uid":"C4U7msmJdZuUt4Nr5","id.orig_h":"192.168.1.103","id.orig_p":123,"id.resp_h":"91.189.94.4","id.resp_p":123,"version":4,"mode":3,"stratum":0,"poll":16.0,"precision":0.015625,"root_delay":1.0,"root_disp":1.0,"ref_id":"\\x00\\x00\\x00\\x00","ref_time":0.0,"org_time":0.0,"rec_time":0.0,"xmt_time":XXXXXXXXXX.XXXXXX,"num_exts":0}
+ {"ts":XXXXXXXXXX.XXXXXX,"uid":"C4U7msmJdZuUt4Nr5","id.orig_h":"192.168.1.103","id.orig_p":123,"id.resp_h":"91.189.94.4","id.resp_p":123,"version":4,"mode":3,"stratum":0,"poll":16.0,"precision":0.015625,"root_delay":1.0,"root_disp":1.0,"ref_id":"\u0000\u0000\u0000\u0000","ref_time":0.0,"org_time":0.0,"rec_time":0.0,"xmt_time":XXXXXXXXXX.XXXXXX,"num_exts":0}
```

The `&binary` tag idea would resolve this nicely for analyzer.log and ntp.log, given the binary nature of those fields.

#### Our private testsuite

Of 207,267 lines 61 (0.02%) changed. Except for three similar analyzer.log changes this is all http.log in the short test, in the `uri` field (e.g. `/scripts/..\\xc1\\x1c../winnt/system32` -> `/scripts/..\\xc1\u001C../winnt/system32`).

NEWS
```diff
@@ -172,8 +172,12 @@ Changed Functionality
 - The Notice framework's suppression implementation was changed to batch suppression
   information for up up to 10msec before distributing it to other nodes in a Zeek cluster.
   This reduces cluster communication overhead when there's a storm of notices
-  generated. Two new options, ``Notice::suppression_batch_period`` and ``Notice::suppression_batch_max_size``
-  can be used to tune this behavior. See their documentation for details.
+  generated. Two new options, ``Notice::suppression_batch_period`` and
+  ``Notice::suppression_batch_max_size`` can be used to tune this behavior. See their
+  documentation for details.
+
+- Non-printable ASCII control characters are no longer escaped with an ``\\x`` prefix in
```
ckreibich: Maybe turn the wording around to strengthen that all JSON logging is affected by this one, possibly with an example?

timwoj: Done, and I added the test too.

.pre-commit-config.yaml
```diff
@@ -26,7 +26,7 @@ repos:
       - "c"
       - "c++"
       - "json"
-    exclude: '^src/3rdparty/.*'
+    exclude: '^src/3rdparty/.*|testing/btest/Baseline/*'
```
ckreibich: A nit:
```suggestion
    exclude: '^src/3rdparty/.*|^testing/btest/Baseline'
```
(Ideally we match all of the Baseline directories?)

src/util.h
```diff
@@ -531,19 +531,26 @@ void zeek_strerror_r(int zeek_errno, char* buf, size_t buflen);
...
+ * null characters in the middle of the data.
+ * @param escape_printable_controls flag for whether printable control characters
+ * (e.g. newlines, tabs, etc) should be escaped in the output. Defaults to false.
+ * @param escape_other_controls flag for whether non-printable control characters
+ * (e.g. nulls, bells, etc) should be escaped in the output.
  * @return the escaped string
  */
-std::string json_escape_utf8(const std::string& val, bool escape_printable_controls = true);
+std::string escape_utf8(std::string_view val, bool escape_printable_controls = false,
+                        bool escape_other_controls = true);
```
awelzel, line 543: I'm just skimming and nothing to hold this off, but leaving a comment here anyhow.

IMO we should avoid adding more APIs with trailing `bool` parameters and even more if they have defaults. Or, enforce the awkward `/*escape_printable_controls=*/false` style at call sites.

If you're new to Zeek's code base, these trailing `false, false` or `true` are the small quests that keep you on your toes:


>        case TYPE_FUNC: {
>            writer.String(util::escape_utf8({val->val.string_val.data, static_cast<size_t>(val->val.string_val.length)},
>                                            false, false));

>                 std::string result = util::escape_utf8({s, static_cast<size_t>(e - s)}, true);

timwoj: I'll fix this in a follow-up.

PR #5017, opened 2025-11-13
Add VLAN PCP/DEI fields to packets

### LINKED ISSUE #2971: Review exposing full VLAN TCI
As mentioned in https://github.com/zeek/zeek-af_packet-plugin/issues/60#issuecomment-1518194023, it might make sense to expose the complete VLAN TCI including PCP and DEI. Currently, this information is dropped at the analyzer-level.

evantypanski:
Closes #2971

I was looking at vlans for xdp shunting, noticed PCP/DEI aren't exposed in Zeek. Then there's an issue for it. Figured I'd give it a shot.

But, most importantly: do we want this?


timwoj:
I think the only thing wrong with adding it is that it expands the size of a `conn` record, but not significantly. Two questions (without really reviewing):

1. Should we expand the vlan_fivetuple conn_key plugin to include the data. 
2. It looks like the openflow framework code includes `vlan_pcp` and `dl_vlan_pcp` fields. Can those be merged with this?

evantypanski:
> Should we expand the vlan_fivetuple conn_key plugin to include the data.

Probably not since the PCP/DEI aren't used to identify the connection (... to my knowledge :) )

> It looks like the openflow framework code includes vlan_pcp and dl_vlan_pcp fields. Can those be merged with this?

I'm not really sure, I've looked a bit and don't see how it'd be merged with the zeek-netcontrol vlan stuff. It looks like the switch mostly handles that? But I can't say I'm qualified to give an answer without diving in a fair bit. If anyone has experience with that and can give a better answer I'm happy to hear it

> I think the only thing wrong with adding it is that it expands the size of a conn record

We could just optionally include the pcp/dei if we want? I wouldn't personally think it's a big deal, but also could see not wanting to add a small field every time we want it just to avoid a slow bloat over time.

Technically since it's all within the TCI we could just store the TCI and give mechanisms to grab the ID/PCP/DEI from that, then it'd remain the same size, but that'd definitely require more work (like deprecation of the old `vlan`, but we can split into `vlan` and `vlan_tci` then deprecate/remove `vlan` later).

awelzel:
~Hmm, hmm, as an alternative we could treat this more as information you can get lazily from script land if you want it rather than always adding it to `connection`. C.f. with `Geneve::get_options()` or `PPPoe::session_id()` for which we tried a different pattern of lazily extracting packet level info. I guess AF_PACKET will be annoying here because it eats the VLAN info, but could special case and stash the info away elsewhere.~

EDIT:

>> I think the only thing wrong with adding it is that it expands the size of a conn record

>  land if you want it rather than always adding it to connection.

This doesn't seem to actually happen. The Packet class gets bigger though, so seems okay to me! :+1:

src/iosource/Packet.h
```diff
@@ -167,11 +167,31 @@ class Packet {
      */
     uint32_t vlan = 0;
 
+    /**
+     * (Outermost) VLAN PCP if any, else 0.
+     */
+    uint32_t vlan_pcp = 0;
+
+    /**
+     * (Outermost) VLAN DEI if any, else false.
```
0xxon: As above, I don't like the documentation of this. This indicates the VLAN DEI if there is a vlan header - not if there is a VLAN DEI. Reading this the first time prompted me to re-lookup how vlan headers work, and if the DEI/PCP are optional or not - and I assume I won't be the only person that happens to.

scripts/base/init-bare.zeek
```diff
@@ -2295,7 +2295,11 @@ type l2_hdr: record {
 	src: string &optional;	##< L2 source (if Ethernet).
 	dst: string &optional;	##< L2 destination (if Ethernet).
 	vlan: count &optional;	##< Outermost VLAN tag if any (and Ethernet).
+	vlan_pcp: count &optional;	##< Outermost VLAN PCP if any (and Ethernet).
```
0xxon: I would refine the documentation for these cases to something like

```
##< Outermost VLAN PCP, if vlan header is present
```

Or 
```
###< Outermost VLAN PCP (if vlan and Ethernet).
```

The current formulation sounds a bit unclear to me, as it doesn't make it clear that those fields always are set when a vlan header is available, and are not optional in case a vlan header is set.

evantypanski:
Great, I updated the comments to be clear that they rely on whether the corresponding `vlan` or `inner_vlan` field is set, I think it's clearest that way. also added a NEWS entry

0xxon:
Great, thank you :)

PR #5021, opened 2025-11-14
ZAM built-in for incrementing a Telemetry counter

vpax:
I'd like to make telemetry as cheap to use as possible. My initial use case concerns counters, so this PR adds a custom ZAM instruction for those, rather than enduring the overhead of calling the BiF. Down the road I may add others if they wind up being hot-spots too.

### DIFF
--- a/src/script_opt/ZAM/OPs/ZBI.op
+++ b/src/script_opt/ZAM/OPs/ZBI.op
@@ -490,6 +490,39 @@ op-types I S U
 side-effects OP_FILES_SET_REASSEMBLY_BUFFER_Vi OP_VV_I2
 eval	$$ = ZAM::file_mgr_set_reassembly_buffer($1, $2);
 
+macro TelemetryCounterInc(result, family, increment)
+	{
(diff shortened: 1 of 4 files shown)


src/script_opt/ZAM/OPs/ZBI.op
```diff
@@ -490,6 +490,39 @@ op-types I S U
 side-effects OP_FILES_SET_REASSEMBLY_BUFFER_Vi OP_VV_I2
 eval	$$ = ZAM::file_mgr_set_reassembly_buffer($1, $2);
 
+macro TelemetryCounterInc(result, family, increment)
+	{
+	if ( auto ptr = dynamic_cast<CounterMetricVal*>(family) )
```
awelzel, line 495: That `dynamic_cast` might be slow, but you'll probably see it in flamegraphs if it actually is.

vpax: Please tell me a better way to do it! I just lifted it from the BiF.

bbannier: https://github.com/zeek/zeek/pull/5025

PR #5024, opened 2025-11-17
websocket: Fix spurious if not

awelzel:
Thanks @vpax

### DIFF
--- a/scripts/base/protocols/websocket/main.zeek
+++ b/scripts/base/protocols/websocket/main.zeek
@@ -155,9 +155,13 @@ event http_request(c: connection, method: string, original_URI: string,
                    unescaped_URI: string, version: string)
 	{
 	# If we see a http_request and have websocket state, wipe it as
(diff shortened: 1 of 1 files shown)


scripts/base/protocols/websocket/main.zeek
```diff
@@ -155,9 +155,11 @@ event http_request(c: connection, method: string, original_URI: string,
                    unescaped_URI: string, version: string)
 	{
 	# If we see a http_request and have websocket state, wipe it as
-	# we should've seen a websocket_established even on success and
-	# likely no more http events.
-	if ( ! c?$websocket )
+	# there might be a nested WebSocket upgrade coming, but it's the
+	# same connection record.
+	#
+	# Dragons here.
```
awelzel: Mostly because I wonder if WebSocket connection should have a dedicated inner `connection` instance. That's not just for WebSocket, also HTTP CONNECT or other "upgrades" suffer from this.

PR #5028, opened 2025-11-17
snmp: Fix crash with PCAP extracted from benchmarker

awelzel:
@0xxon - without the proposed patch to snmp-analyzer.pac, this would cause a segfault because usm_security_parameters yields a nullptr.

I haven't looked further, but seems `v3UsmSecurityParameters` maybe isn't properly parsed. Do you have enough context swapped in still?


awelzel:
> Yup, that's an oversight of me. This happens when the protocol indicates that there are security parameters, but they are not valid.

Wireshark says they are encrypted, so I was wondering if we wanted to have some encrypted flag in snmp.log (maybe that's useful for compliance use-cases or so). I'm happy to merge as is.

0xxon:
They are encrypted - but that's independent of this problem. That header is independent of encryption, and in this case just contains a somewhat broken payload.

PR #5032, opened 2025-11-19
Bump Spicy to new optimizer implementation.

rsmmr:
This pulls in https://github.com/zeek/spicy/pull/2201.

This disables the new, expensive Spicy-side debug checks when built as part of Zeek. While that was the plan for Zeek anyways, I was originally hoping to run them through the Zeek-side CI at least once. However, Zeek's ASAN check times out when they are enabled. (FWIW, the other default Zeek CI tasks did pass with them on).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -996,6 +996,8 @@ else ()
 endif ()
 
 if (NOT DISABLE_SPICY)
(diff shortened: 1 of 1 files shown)


src/spicy/spicyz/glue-compiler.cc
bbannier, line 1: None of the changes in this file should be needed when working against a recent Zeek `master` branch.

rsmmr: Ah, missed that!

Issue #2505, opened 2022-10-26
Error installing packages when FIPS is enabled


clopmz:
Host: RedHat Enterprise Linux 9
Zeek release: 5.0.2
OpenSSL version: 3.0.1

When you try to install af_packet package under a RHEL9 host with FIPS enabled returns the following error:

```console
$ more testing/zeek-af_packet-plugin/clones/zeek-af_packet-plugin/zkg.test_command.stderr
scripts.show-plugin ... failed
  % 'btest-diff output' failed unexpectedly (exit code 1)
  % cat .diag
  == File ===============================
  == Diff ===============================
  --- /tmp/test-diff.57126.output.baseline.tmp	2022-10-26 09:01:57.467293567 +0000
  +++ /tmp/test-diff.57126.output.tmp	2022-10-26 09:01:57.460293516 +0000
  @@ -1,11 +0,0 @@
  -Zeek::AF_Packet - Packet acquisition via AF_Packet (dynamic, version)
  -    [Packet Source] AF_PacketReader (interface prefix "af_packet"; supports live input)
  -    [Type] AF_Packet::FanoutMode
  -    [Constant] AF_Packet::buffer_size
  -    [Constant] AF_Packet::enable_hw_timestamping
  -    [Constant] AF_Packet::enable_fanout
  -    [Constant] AF_Packet::enable_defrag
  -    [Constant] AF_Packet::fanout_mode
  -    [Constant] AF_Packet::fanout_id
  -    [Constant] AF_Packet::link_type
  -
  =======================================

  % cat .stderr
  internal error: EVP_DigestInit failed

1 of 1 test failed
```

Testing against openssl framework confirms some type of problem when FIPS is enabled:
```console
$ zeek -e 'print md5_hash("zeek"); print sha1_hash("zeek"); print sha256_hash("zeek")'
internal error: EVP_DigestInit failed
Aborted (core dumped)
```

Disbling FIPS framework, all it is working ok.

0xxon:
For anyone who wants to reproduce this - the magic commands you have to use to make libraries like `libpcap-devel` available on RHEL9 are (at least for me):

```
subscription-manager config --rhsm.manage_repos=1
subscription-manager repos --enable codeready-builder-for-rhel-9-$(arch)-rpms
```

This is the backtrace of the command line mentioned in the ticket, after enabling FIPS-mode with `fips-mode-setup --enable`:

```
[...]

PR #5033, opened 2025-11-19
Swap internal uses of MD5 with SHA256

timwoj:
This is part of the work to support FIPS mode in #2505. This PR swaps out some of our default internal uses of MD5 with SHA256, namely:

- KRB5 ticket hashing
- MIME content hashing
- SSH fingerprint hashing

These are all technically breaking changes, due to changes in what gets logged breaking backwards compatibility. After this, the only default use is in the `md5` field in files.log, which might be a trickier problem.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -63,6 +63,27 @@ Breaking Changes
       ${OPENSSL_LIBRARIES}
     )
 
+- We are revisiting removing support for MD5 hashing across the Zeek code base. In support
+  of this, the following changes were made:
+
+  - The ``host_key`` field in ssh.log was removed and replaced with a new
+    ``host_key_fingerprint`` field. The old field held an MD5 hash of the host key. The
+    new field holds a base64-encoded version of the hash of the host key, in the same
+    format that the ``ssh-keygen -l`` command would return. It currently uses SHA256 for
+    the hashing, so the output is of the format ``SHA256:<fingerprint>``. A new
+    ``policy/protocols/ssh/md5-host-key-logging.zeek`` script was added to restore the old
+    field if desired.
+
+  - The ``auth_ticket`` and ``new_ticket`` fields in krb.log were removed and replaced
+    with new ``auth_ticket_sha256`` and ``new_ticket_sha256`` fields. The old fields held
+    MD5 hashes of the tickets and the new fields hold SHA256 hashes of the same. A new
+    ``policy/protocols/krb/md5-ticket-logging.zeek`` script was added to restore the old
+    fields if desired.
+
+  - A new ``mime_content_hash_sha256`` event was added to send a SHA256 version of the
+    content in MIME messages. This is similar to the existing ``mime_content_hash`` event
+    that sends MD5 versions.
+
 New Functionality
 -----------------
 


testing/btest/Baseline/scripts.base.protocols.ssh.basic/ssh.log
```diff
@@ -7,26 +7,26 @@
...
-XXXXXXXXXX.XXXXXX	CLNN1k2QMum1aexUK7	192.168.1.32	58641	131.103.20.168	22	2	F	1	OUTBOUND	SSH-2.0-OpenSSH_6.7	SSH-2.0-OpenSSH_5.3	aes128-ctr	umac-64@openssh.com	none	diffie-hellman-group-exchange-sha256	ssh-rsa	97:8c:1b:f2:6f:14:6b:5c:3b:ec:aa:46:46:74:7c:40
-XXXXXXXXXX.XXXXXX	CBA8792iHmnhPLksKa	192.168.1.32	58646	131.103.20.168	22	2	T	1	OUTBOUND	SSH-2.0-OpenSSH_6.7	SSH-2.0-OpenSSH_5.3	aes128-ctr	umac-64@openssh.com	none	diffie-hellman-group-exchange-sha256	ssh-rsa	97:8c:1b:f2:6f:14:6b:5c:3b:ec:aa:46:46:74:7c:40
-XXXXXXXXXX.XXXXXX	CGLPPc35OzDQij1XX8	192.168.1.32	58649	131.103.20.168	22	2	T	1	OUTBOUND	SSH-2.0-OpenSSH_6.7	SSH-2.0-OpenSSH_5.3	aes128-ctr	umac-64@openssh.com	none	diffie-hellman-group-exchange-sha256	ssh-rsa	97:8c:1b:f2:6f:14:6b:5c:3b:ec:aa:46:46:74:7c:40
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.79	51880	131.159.21.1	22	2	-	0	OUTBOUND	SSH-2.0-OpenSSH_5.9	SSH-2.0-OpenSSH_5.8	aes128-ctr	hmac-md5	zlib@openssh.com	ecdh-sha2-nistp256	ecdsa-sha2-nistp256	ee:3c:6f:2c:23:39:eb:f3:b1:f7:15:2b:a1:28:e1:bb:99:2e:a9:62:15:03:d0:f3:ed:43:5e:59:9e:3e:50:40
```
awelzel: This isn't the format that `ssh`  would show you or that is used to document host keys. It's specific to MD5. Could we do this instead so that values in the log can be used for referencing public information?

> These are GitHub's public key fingerprints:

> SHA256:uNiVztksCsDhcc0u9e8BujQXVUpKZIDTMczCvj3tD2s (RSA)
> SHA256:br9IjFspm1vxR3iA35FWE+4VTyz1hYVLIE2t1/CeyWQ (DSA - closing down)
> SHA256:p2QAMXNIC1TJYWeIOttrVc98/R1BUFWu3/LiyKgUfQM (ECDSA)
> SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU (Ed25519)

https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints

src/Hash.cc
```diff
@@ -53,6 +51,19 @@ void KeyedHash::InitializeSeeds(const std::array<uint32_t, SEED_INIT_SIZE>& seed
     seeds_initialized = true;
 }
 
+void KeyedHash::InitializeHmacMd5Seed() {
+    if ( ! seeds_initialized )
+        assert("InitializeSeeds() must be called before InitializeMD5Seed()");
```
awelzel: FatalError()? `assert()` will be removed in prod builds.

src/analyzer/protocol/ssh/ssh-analyzer.pac
```diff
@@ -211,12 +215,14 @@ refine flow SSH_Flow += {
 		%{
 		if ( ssh_server_host_key )
```
awelzel, line 231: Mostly just an idea, but should we continue to support md5 with the `ssh_server_host_key` event (not used or disabled via a const &redef and introduce a new `ssh_server_host_sha256_key` that's used by default in the scripts?

That should provide a way to users to revert to md5 by implementing the other event and overwriting the sha256 value (or some other pattern). We may even want to add a policy script like `portocol/ssh/use-deprecated-md5-hostkey` or some such?

I'm just very very timid we change the format of a column in ssh.log without an easy way to revert to the prior format.

src/analyzer/protocol/mime/MIME.cc
```diff
@@ -1256,7 +1256,7 @@ void MIME_Mail::SubmitData(int len, const char* buf) {
 
     if ( compute_content_hash ) {
```
awelzel, line 1271: I'd probably also here consider adding `compute_content_hash_sha256` and maybe deprecating the existing one instead of changing the actual value.

If someone uses this event to lookup information in some MD5-only database, we outright break their use-case.

ckreibich: I was mainly thinking of the corresponding `mime_content_hash` event — it seems little-used (in fact I can't find _any_ use), so I'd lean toward introducing a `mime_content_hash_sha256`. As you're mentioning for the SSH changes, it also provides a way to keep existing activity alive by providing the `mime_content_hash` event in other ways in the future.

src/analyzer/protocol/ssh/ssh-analyzer.pac
ckreibich: That would also mirror the same approach for the `mime_content_hash` event, so yep, it gets my vote.

testing/btest/Baseline/scripts.base.protocols.ssh.basic/ssh.log
ckreibich: Heh, I also noticed that the `ssh_server_host_key` event's docstring is outdated in that regard. I'd be on board with this, particularly if the original motivation was indeed to be compatible with SSH's output — the fingerprint used to be MD5 in the colon-separated format until 2015, so that looks likely. The new one is apparently just base64 of the sha256.

Btw the `host_key` field name to me suggests it's the key, literally — not its fingerprint. So a column name transition to something like `host_key_fingerprint` would resolve that and provide a transition window.

src/Hash.h
```diff
@@ -183,6 +183,12 @@ class KeyedHash {
      */
     static void InitializeSeeds(const std::array<uint32_t, SEED_INIT_SIZE>& seed_data);
 
+    /**
+     * Initialize the (typically process-specific) seed for hmac_md5(). This is lazily called
+     * the first time hmac_md5() is called.
+     */
+    static void InitializeHmacMd5Seed();
```
ckreibich: What's the motivation for the lazy-init Tim?

ckreibich (commented):
Just a few comments  for now as we're still asking around for thoughts regarding this. I certainly like that the PR is not very big!

src/Hash.h
timwoj: The failure reported in the first post of the ticket is that we call `EVP_DigestInit_ex` for MD5 at startup. We do that as part of `KeyedHash::IniitializeSeeds` during startup to initialize the seeds for `hmac_md5`. That's only used by the MD5-based IP anonymizers and the `md5_hmac` BIF (which isn't called by any of our default scripts). In order to avoid doing that init at startup, it can get lazily init'd either.

Thinking about it further though, this just pushes the error until later in runtime, when it might be surprising that it occurs. Re: the MD5 anonymizers, I'm still of the opinion that we could simply deprecate those, replace them with SHA256 ones, and remove them later on.

src/analyzer/protocol/mime/MIME.cc
timwoj: I like this idea the most so far. It presents a new name that has new data in it and lets us deprecate the old one.

awelzel: > I'd lean toward introducing a mime_content_hash_sha256. As you're mentioning for the SSH

I think that's what I meant, but didn't really look how MIME works in detail. :+1:

testing/btest/Baseline/scripts.base.protocols.ssh.basic/ssh.log
timwoj: I added another commit that switches the SSH changes to add a `host_key_fingerprint` field that contains a b64-encoded sha256 version of the host key. The output matches the format that `ssh-keygen -l` would output. It also adds a `ssh_server_host_key_fingerprint` for setting the value in the field. I also marked the existing field deprecated.

src/analyzer/protocol/ssh/ssh-analyzer.pac
timwoj: See https://github.com/zeek/zeek/pull/5033#discussion_r2551147081. I'm pretty indifferent if we should add a flag that turns the existing field on or off. I'd be fine with just leaving it in, adding a deprecation note to NEWS, and eventually just removing it entirely in 9.1.

timwoj:
I pushed a change to the kerberbos bit that adds two new columns for sha256 instead of replacing the existing columns. It also deprecates the old ones.

src/analyzer/protocol/ssh/ssh-analyzer.pac
```diff
@@ -211,12 +230,17 @@ refine flow SSH_Flow += {
...
-			unsigned char digest[ZEEK_MD5_DIGEST_LENGTH];
-			zeek::detail::internal_md5(${key}.data(), ${key}.length(), digest);
+			unsigned char digest[ZEEK_SHA256_DIGEST_LENGTH];
+			auto ctx = zeek::detail::hash_init(zeek::detail::Hash_SHA256);
+			zeek::detail::hash_update(ctx, ${key}.data(), ${key}.length());
+			zeek::detail::hash_final(ctx, digest);
 
 			zeek::BifEvent::enqueue_ssh_server_host_key(connection()->zeek_analyzer(),
 				connection()->zeek_analyzer()->Conn(),
 				zeek::make_intrusive<zeek::StringVal>(fingerprint_md5(digest)));
```
awelzel, line 240: I'm a bit confused: Are we using the first few bytes of the SHA256 digest and format it in the `XX:XX...` MD5 format? I'd think that's not great :-)

Think this also needs a separate `if ( ssh_server_host_key_fingerprint )` block, rather than being within `ssh_server_host_key`. If someone uses *only* the new `ssh_server_host_key_fingerprint`, this code won't run.

If possible and  doesn't hurt maintainability: Is there a real pressing reason to completely remove MD5 fingerprinting. Maybe some users would like to continue having it (for whatever reason)?

What do you guys think about deprecating included by default, but with an option to opt-in again, rather than deprecating with the idea of completely removing it. It seems we're taking something away wholesale just because in certain (special?) environments it's not allowed to be used (and this one specifically is more difficult to get back via a plugin compared to MD5 file hashing).

src/Hash.h
```diff
@@ -207,7 +213,9 @@ class KeyedHash {
     alignas(16) static unsigned long long shared_siphash_key[2];
     // This key changes each start (unless a seed is specified)
     inline static uint8_t shared_hmac_md5_key[16];
+    inline static std::array<uint32_t, SEED_INIT_SIZE> seed_data;
```
awelzel: Wouldn't mind a comment for that one, given the above is fairly verbose. And sticking with a C-style array might be a bit more consistent, but no strong opinion.

scripts/policy/protocols/krb/ticket-logging.zeek
```diff
@@ -6,9 +6,13 @@ module KRB;
 
 redef record Info += {
 	## Hash of ticket used to authorize request/transaction
-	auth_ticket: string &log &optional;
+	auth_ticket: string &log &optional &deprecated="Remove in v9.1. Use auth_ticket_sha256";
```
awelzel: The `&deprecated` isn't visible for log consumers, so need a pretty loud NEWS entry.

What we did for `files.log` and `tx_hosts` / `rx_hosts` was to outright remove them, but provide a policy script to get the fields back. Maybe we could do this here, too: Remove the fields, then have a way to get them back via:
```
@load policy/protocol/krb/md5-new-ticket-auth-ticket-fields
```

That script would add `new_ticket` and `auth_ticket` and have separate event handlers to set them, I don't think anything needs the `&deprecated` then.

That way it'll be in the `NEWS`, it'll break users, and we have a way to easily get the fields back for those that really want md5. Similar as for the SSH md5 host key, I wouldn't just outright remove them completely, but maybe remove them from the default configuration. They've been around for pretty long.

scripts/base/protocols/ssh/main.zeek
```diff
@@ -53,7 +53,9 @@ export {
 		## The server host key's algorithm
 		host_key_alg:    string       &log &optional;
 		## The server's key fingerprint
-		host_key:        string       &log &optional;
+		host_key:        string       &log &optional &deprecated="Remove in 9.1. Use host_key_fingerprint.";
+		## The server's key fingerprint, in the format that `ssh-keygen -l` would output.
```
awelzel, line 55: Maybe add an example? I'm forth and back if the  `SSH256:<base64>` format is nice, but I think it works.

src/analyzer/protocol/ssh/ssh-analyzer.pac
awelzel: Similarly as for krb, but more work, a policy script to add back the MD5 fingerprints would be really nice, but I might also be overthinking this.

src/analyzer/protocol/ssh/ssh-analyzer.pac
```diff
@@ -48,17 +50,30 @@ zeek::VectorValPtr name_list_to_vector(const bytestring& nl)
...
+	char* outbuf = nullptr;
+	int outlen = 0;
+	zeek::detail::Base64Converter enc(nullptr, "");
+	enc.Encode(ZEEK_SHA256_DIGEST_LENGTH, d, &outlen, &outbuf);
+
+	if ( outlen == 0 )
+		return {};
+
+	std::string ret = "SHA256:";
+	ret.append(outbuf, outlen-1);
```
awelzel: Could probably rewrite this as something like the following to avoid a copy? But not sure it's worth it.
```
std::string ret = "SHA256:";
ret.resize(...);
enc.Encode(ZEEK_SHA256_DIGEST_LENGTH, d, &outlen, &ret[7]);
```

testing/btest/Baseline/scripts.base.protocols.ssh.basic/ssh.log
```diff
@@ -5,28 +5,28 @@
...
-XXXXXXXXXX.XXXXXX	CGLPPc35OzDQij1XX8	192.168.1.32	58649	131.103.20.168	22	2	T	1	OUTBOUND	SSH-2.0-OpenSSH_6.7	SSH-2.0-OpenSSH_5.3	aes128-ctr	umac-64@openssh.com	none	diffie-hellman-group-exchange-sha256	ssh-rsa	cf:35:d0:39:74:91:04:48:94:b6:e1:3c:02:29:09:60:ac:1b:1d:ac:e1:49:cd:28:8d:ec:fd:61:76:86:a7:50
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	version	auth_success	auth_attempts	direction	client	server	cipher_alg	mac_alg	compression_alg	kex_alg	host_key_alg	host_key	host_key_fingerprint
+#types	time	string	addr	port	addr	port	count	bool	count	enum	string	string	string	string	string	string	string	string	string
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	192.168.1.79	51880	131.159.21.1	22	2	-	0	OUTBOUND	SSH-2.0-OpenSSH_5.9	SSH-2.0-OpenSSH_5.8	aes128-ctr	hmac-md5	zlib@openssh.com	ecdh-sha2-nistp256	ecdsa-sha2-nistp256	ee:3c:6f:2c:23:39:eb:f3:b1:f7:15:2b:a1:28:e1:bb	SHA256:7jxvLCM56/Ox9xUroSjhu5kuqWLYA9Dz7UNeWZ4+UEA
```
awelzel: The `ee` here is the first byte of the SHA256 fingerprint, should probably stay MD5 here ;-)

src/Hash.h
timwoj: It's mostly a std::array because that's what the argument to `InitializeSeeds` is, and this just copies the data from that.

scripts/policy/protocols/krb/ticket-logging.zeek
timwoj: How about adding a script to add them back in, loading it by default, but with a deprecation warning for the script that it won't be loaded by default after 9.1?

awelzel: > How about adding a script to add them back in, loading it by default, but with a deprecation warning for the script that it won't be loaded by default after 9.1?

We went through this [discussion with files.log](https://docs.google.com/document/d/1M2QNbofW__MeDNHbFTqpy8CdpWXZrjB4D4OiUB-fJ3U/edit?tab=t.0#heading=h.me63rkckoayz) (see Deprecation Approach - it mentions a deprecation warning by default, but Robin was fairly strongly against that). In the end, tx_hosts, rx_hosts was removed by default with a script to add them back.

I haven't heard many or strong complaints that this didn't work, and it's been a bit by now. IMO it's a good balance of raising awareness (remove by default), but also an easy way to get back to a working setup in the short term.

src/Hash.h
ckreibich: Makes sense Tim, agree.

timwoj: I'm going to drop this bit from this PR and do it in a separate one.

scripts/policy/protocols/krb/md5-new-auth-ticket-fields.zeek
```diff
@@ -0,0 +1,33 @@
...
+	## MD5 hash of ticket used to authorize request/transaction
+	auth_ticket: string &log &optional;
+	## MD5 hash of ticket returned by the KDC
+	new_ticket:  string &log &optional;
+};
+
+event krb_ap_request(c: connection, ticket: KRB::Ticket, opts: KRB::AP_Options)
+	{
+	# Will be overwritten when request is a TGS
+	c$krb$request_type = "AP";
```
awelzel:
```suggestion
```

think this can go.

testing/btest/Baseline.zam/coverage.test-all-policy-cluster/.stderr
```diff
@@ -0,0 +1,9 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+warning in <no location>: deprecated (SSH::Info$host_key): Remove in 9.1. Use host_key_fingerprint. (SSH::c$ssh$host_key)
+received termination signal
+warning in <no location>: deprecated (SSH::Info$host_key): Remove in 9.1. Use host_key_fingerprint. (SSH::c$ssh$host_key)
+received termination signal
+warning in <no location>: deprecated (SSH::Info$host_key): Remove in 9.1. Use host_key_fingerprint. (SSH::c$ssh$host_key)
+received termination signal
+warning in <no location>: deprecated (SSH::Info$host_key): Remove in 9.1. Use host_key_fingerprint. (SSH::c$ssh$host_key)
+received termination signal
```
awelzel, line 5: This seems stale? There's no "Remove in 9.1" anymore in these changes.

scripts/policy/protocols/krb/md5-ticket-logging.zeek
```diff
@@ -0,0 +1,33 @@
+##! Add MD5 Kerberos ticket hashes to the krb.log
```
awelzel, line 1: Sorry. Maybe a better name for the file is `md5-ticket-logging.zeek` to align a bit with the `-logging.zeek` pattern used elsewhere.

scripts/policy/protocols/ssh/md5-host-key-logging.zeek
```diff
@@ -0,0 +1,20 @@
+##! Adds MD5 host_key field to ssh.log
```
awelzel, line 1: Similarly here, maybe `md5-host-key-logging.zeek` is enough.

awelzel (approved):
Seems okay to me. I'm definitely a fan of not changing the semantics of existing log columns and instead removing them with an easy way to get them back.

Definitely needs a big entry in NEWS and detailed instructions which fields are affected and how to get these back. It might not hurt to post on community.zeek.org that this is happening.

I'm stepping a bit back, and the motivation here is to stop using MD5 in a default configuration. The underlying reason is that in various environments MD5 is phased out and might cause runtime errors. If users know that MD5 is fine in their environment and want these columns, they can easily load the scripts again.

EDIT: Some stale baselines need updates.

PR #5035, opened 2025-11-21
Remove Spicy DNS from CI

evantypanski:
It probably just isn't worth it to keep up with it for now.

This caused a CI failure in master because of https://github.com/zeek/spicy-dns/issues/19


evantypanski:
well spicy_head fails because Ubuntu is currently failing in CI over there, argh

EDIT: Nevermind, it's just because spicy-head needs a Spicy bump to change `transform`

timwoj (approved):

PR #5036, opened 2025-11-22
Support JSON output in `policy/misc/dump-events.zeek`

ckreibich:
I use this script somewhat regularly and find the ad-hoc format inconvenient. Setting `DumpEvents::use_json=T` creates one line per event, with a JSON object that looks like this (pretty-printed):
```
{
  "ts":  <network time>,
  "event": <event name>,
  "args": {
    <name>: <val>,
    ...
}
```
It does not currently include the toplevel argument type info since I've not had a need. I made args an object because it seems convenient, e.g.
```
$ cat all-events.json | jq 'select(.event=="smtp_reply").args.c.id.orig_h'
"10.10.1.4"
"10.10.1.4"
...
```
(edit: corrected a cut'n'paste error in the above)
The underlying tables are ordered for determinism. See baselines for examples.

I noticed almost by accident that the existing `dump-events.zeek` tests don't actually seem to need timestamp canonicalization, so I'm trying CI without it, both for the original test and the JSON equivalent — consider it an experiment.

I was tempted to expand `tuning/use-json` so it enables JSON for this script too when available, but held off for now.

I also noticed a bug in that the `call_argument$value` field actually is set even when the argument used a `&default` value — otherwise the existing code would have crashed a long time ago. I'm not looking further into this right now.

### DIFF
--- a/scripts/policy/misc/dump-events.zeek
+++ b/scripts/policy/misc/dump-events.zeek
@@ -71,8 +71,8 @@ event new_event(name: string, args: call_argument_vector)
 
 		if ( a?$value )
 			print fmt("                  [%d] %-18s = %s", i, proto, a$value);
-		else
-			print fmt("                  | %-18s = %s [default]", proto, a$value);
+		else if ( a?$default_val )
+			print fmt("                  | %-18s = %s [default]", proto, a$default_val);
 		}
 
 	print "";


ckreibich:
Ah right, thanks Tim — will check.

PR #5038, opened 2025-11-22
Bump `auxil/spicy`

bbannier:


timwoj:
Just leaving a note to update https://github.com/zeek/zeek/wiki/Zeek-Operating-System-Support-Matrix when this lands with a note about requiring GCC 12 on Ubuntu 22 and CentOS 9.

ci/centos-stream-9/Dockerfile
```diff
@@ -44,6 +43,10 @@ RUN dnf -y --nobest install \
     zlib-devel \
   && dnf clean all && rm -rf /var/cache/dnf
 
+# Set gcc-12 as default compiler.
```
timwoj: Nit: Can you move this down next to the other `alternatives` runs?

bbannier: Done. This is now also more explicit about needed tools by using `alternatives` instead of tweaking `PATH`. I am a bit surprised this seems to work without modifying `LD_LIBRARY_PATH` 🤷.

ci/centos-stream-9/Dockerfile
```diff
@@ -25,8 +25,7 @@ RUN dnf -y --nobest install \
     cppzmq-devel \
     diffutils \
     flex \
-    gcc \
-    gcc-c++ \
+    gcc-toolset-12-gcc-c++ \
```
bbannier, line 28: https://github.com/zeek/zeek/pull/5038#issuecomment-3572381521

PR #5040, opened 2025-11-24
RFC: cluster: Add Cluster::Table::publish_element_new

awelzel:
This is something that @bbannier and me discussed as an alternative to &backend / &broker_backend specifically for the use-case of cluster-wide caching tables.

The usage is essentially

    global tbl: string[string] of count &on_change=Cluster::Table::publish_element_new &write_expire=5min;

Crucially, `publish_element_new()` means only *new* elements are published to other nodes. Anything else is out of scope. This maps nicely to the &on_change's TABLE_ELEMENT_NEW. Additionally, there's no re-sync of a table when a node restarts and re-joins the cluster. This is solvable via a `Cluster::node_up()` handler publishing the full table contents from, e.g. the manager, to the node that joined the cluster if the use-case requires it.

I'm a fan of the idea as it's convenient, captures a number of use-cases within Corelight and it's easy to explain. I'm not a huge fan of the implementation (wrangling with any and Broker::Data wrappers...), but all in all there's nothing really surprising or magic happening.

---

This adds a bif that can be used as an &on_change handler for distributing
every new element in a table across a Zeek cluster.

The implementation is based on a Cluster::Table::elements_new_internal()
event which is sent from within the change handler.

We should probably consider batching with a timer to make this more
efficient at the expense of slight delay.

### DIFF
--- a/scripts/base/frameworks/cluster/table.zeek
+++ b/scripts/base/frameworks/cluster/table.zeek
@@ -3,11 +3,11 @@
 ##! This module explores the idea of using the ``&on_change`` attribute
 ##! on tables for state propagation in a Zeek cluster.
 ##!
-##! The :zeek:see:`Cluster::Table::publish_new_element()` builtin function
+##! The :zeek:see:`Cluster::Table::publish_new_element` builtin function
 ##! can be used as the ``&on_change`` function and will internally publish
 ##! any new elements of a table to a table specific topic. For example,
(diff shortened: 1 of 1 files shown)


awelzel:
Oh and except for the pre-work for `&on_changed` and `bifcl` this could totally be an external plugin. No need to be in-tree, though it makes things easier.

awelzel:
Extracted pre-work into #5157, closing this one.

PR #5043, opened 2025-11-24
CI: Require gcc-12 on Ubuntu 22.04 and CentOS 9

timwoj:
This updates the CI Dockerfiles for Ubuntu 22.04 and Centos Stream 9 to require gcc-12. See https://github.com/zeek/spicy/issues/2204 for details. I'll update the support matrix once this is merged.


bbannier:
I already opened https://github.com/zeek/zeek/pull/5038, but would be happy to close it if your simpler approach on centosstream-9 works.

timwoj:
Ah, I missed that it was part of that. It ends up being the same, so I'll close this one.

PR #5047, opened 2025-11-25
Fix wrong packet/byte accounting for flipped connections

awelzel:
Splitting this out from #4886 because that feels more tricky than what #4886 actually wants to do, I mostly stumbled over it by accident.

In short, flipped connections (specifically those where Zeek flips on the second packet) result in stale `is_orig` values being passed to the ConnSize analyzer. Also, the ConnSize analyzer was never flipped because `packet_children` were simply ignored.

The three connections affected in the external suites have been extracted and added into `btest/Traces`. I've stared with Wireshark mostly and the updated values are correct. Specifically the `one-sided-2745.pcap` was very confused. It produces these results with Zeek 8.0:
```
 zeek-cut -m ts uid id.orig_h id.orig_p id.resp_h id.resp_p history orig_pkts orig_ip_bytes resp_pkts resp_ip_bytes < conn.log < conn.log 
ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       history orig_pkts       orig_ip_bytes   resp_pkts       resp_ip_bytes
1079418989.488904       CxSlFhMLKeOd8qcOk       247.180.123.139 2745    245.181.165.133 1433    ^haf    2       100     1       40
1079436603.911860       CTlL4s278VA28ABjUg      91.189.148.149  2745    245.181.165.131 281     ^haf    2       100     1       40
```

But if you look into the PCAP, only one endpoint (the responder) ever sent packets:

image

After this change, the result is as follows:
```
ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       history orig_pkts       orig_ip_bytes   resp_pkts       resp_ip_bytes
1079418989.488904       CvIbvstAFDM1m2IB6       247.180.123.139 2745    245.181.165.133 1433    ^haf    0       0       3       140
1079436603.911860       CotmJy3aRfV3oi52ah      91.189.148.149  2745    245.181.165.131 281     ^haf    0       0       3       140
```

I've also added a section to `doc/devel` with some thoughts about this, because this feels like a strange fix.

Goes with:
https://github.com/zeek/zeek-testing-private/pull/31
https://github.com/zeek/zeek-testing/pull/28


awelzel:
@rsmmr - I've ~assigned~ requested a review from you in the hope this is a corner you remember. Feel free to add any thoughts to the few paragraphs, I might have missed things or used unusual terminology.~

src/packet_analysis/protocol/tcp/TCP.cc
```diff
@@ -110,8 +110,27 @@ void TCPAnalyzer::DeliverPacket(Connection* c, double t, bool is_orig, int remai
...
+    // See the docs/devel/conn.rst for more details and ideas how to
+    // improve here. This pretty tricky and fragile.
+    const auto* originator = adapter->Orig();
+
     adapter->Process(is_orig, tp, len, ip, data, remaining);
 
+    // If the connection flipped during adapter->Process(), update the local is_orig
+    // to reflect that so that child analyzers observe the right value.
+    if ( originator != adapter->Orig() )
+        is_orig = ! is_orig;
```
rsmmr, line 132: I'm wondering if UDP has the same issue in its `DeliverPacket()` method?

doc/devel/conn.rst
```diff
@@ -0,0 +1,51 @@
...
+
+===================
+Connection Handling
+===================
+
+Flipping Connections
+====================
+
+Zeek works with a concept of originator and responder for a connection. This
+is visible in the Zeek scripting layer as the ``is_orig: bool`` event parameter,,
```
rsmmr, line 11:
```suggestion
is visible in the Zeek scripting layer as the ``is_orig: bool`` event parameter,
```

doc/devel/conn.rst
```diff
@@ -0,0 +1,51 @@
...
+Additionally, the Analyzer API offers a virtual ``FlipRoles()`` method that
+is executed recursively on the analyzer tree when endpoint flipping happens.
+All analyzers have to update their internal state upon such an event.
+Consider the ``ConnSize_Analyzer`` analyzer: It tracks packet and byte counts
+transferred by originator and responder endpoints. When the notion of these
+endpoints changes, a ``ConnSize_Analyzer`` instance needs to update its own
+internal state, as for any following ``DeliverPacket()`` calls, the meaning
+of ``is_orig`` is inverted.
+
+Luckily, this flipping usually happened before the first packet of connection
```
rsmmr, line 33:
```suggestion
Luckily, this flipping usually happens before the first packet of connection
```

rsmmr (approved):

src/packet_analysis/protocol/tcp/TCP.cc
awelzel: I was also wondering that, but I think we're "lucky" that for UDP we only flip on the first packet as of today and I'd just leave it for now. I was pondering a weird/warning in FlipRoles() on the second (UDP), respective third packet (TCP), but it's not actually straight-forward to get the number of packets processed for a connection (need to lookup ConnSize_Analyzer, ...).

rsmmr: Why not just do the same check/flip of `is_orig` for UDP as well, even if strictly speaking it's not needed right now? Would avoid missing it in the future as  technically direction flipping is allowed to happen anytime.

awelzel: Assuming this line:

https://github.com/zeek/zeek/blob/4445ffe737b0294fbb12990163e786ae5190e48b/src/packet_analysis/protocol/udp/UDP.cc#L83

I had looked a bit, and this `DeliverPacket()` call IIUC doesn't actually do anything for UDP (only executes Analyzer::DeliverPacket()), so this cannot result in a flip today. I had a stashed comment here why the call is even there - but probably for `-B dpd` logging). Adding the check might be more confusing.


Something that can flip the connection is the `ForwardPacket()` call at the end of that function, but that just returns directly anyhow.

https://github.com/zeek/zeek/blob/4445ffe737b0294fbb12990163e786ae5190e48b/src/packet_analysis/protocol/udp/UDP.cc#L200-L204

---

Of course, where it probably makes sense to check is while forwarding to child analyzers within the Analyzer implementation:

https://github.com/zeek/zeek/blob/4445ffe737b0294fbb12990163e786ae5190e48b/src/analyzer/Analyzer.cc#L250-L256

https://github.com/zeek/zeek/blob/4445ffe737b0294fbb12990163e786ae5190e48b/src/analyzer/Analyzer.cc#L270-L275

But I didn't want to go there for now - mostly as it feels like a can of worms that's waiting to be opened. Is that fair?

`FlipRoles()` is definitely on my list of "this might take some time" corners now.

rsmmr: > Assuming this line:

Yeah, that's the line I had been looking at too. 

Sounds fair then. :-)

PR #5050, opened 2025-11-25
Add flags for controlling utf8 escaping instead of bools

timwoj:
This came up in the review of https://github.com/zeek/zeek/pull/5014, but I opted to push it off into a follow-up. This changes the flags for controlling what characters are escaped for UTF8 output to an `enum` of flags instead of a set of `bool`s. This should make it more explicit about what is happening.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -528,28 +528,34 @@ std::string canonify_name(const std::string& name);
  */
 void zeek_strerror_r(int zeek_errno, char* buf, size_t buflen);
 
+enum UTF8EscapingFlags : uint8_t {
+    ESCAPE_NONE = 0x00,
+    // Escape printable ASCII control characters (e.g. `\n` is converted to `\x0a`)
+    ESCAPE_PRINTABLE_CONTROLS = 0x01,
+    // Escape unprintable ASCII control characters
+    ESCAPE_UNPRINTABLE_CONTROLS = 0x02,
+};
+
 /**
  * Escapes bytes in a string that are not valid UTF8 characters with \xYY format. Used
  * by the JSON writer and BIF methods.
  * @param val the character data to be escaped. this is a string_view, but it can hold
  * null characters in the middle of the data.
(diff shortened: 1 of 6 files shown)


src/util.h
```diff
@@ -528,28 +528,34 @@ std::string canonify_name(const std::string& name);
  */
 void zeek_strerror_r(int zeek_errno, char* buf, size_t buflen);
 
+enum UTF8EscapingFlags : uint8_t {
+    ESCAPE_NONE = 0x00,
+    // Escape printable ASCII control characters
```
awelzel:
```suggestion
    // Escape printable ASCII control characters (e.g. `\n` is converted to `\x0a`)
```

src/util.h
```diff
@@ -528,28 +528,34 @@ std::string canonify_name(const std::string& name);
...
+    ESCAPE_PRINTABLE_CONTROLS = 0x01,
+    // Escape unprintable ASCII control characters
```
awelzel, line 535: I just remembered that this all exist in Spicy land, too (in one way or another, haven't looked in detail):

https://github.com/zeek/spicy/blob/3a8e48d166f583a0e0cd432a751515d1cad1c1cc/hilti/runtime/include/util.h#L281

I'll just approve what you did here, but I'm wondering if we should just depend on Spicy starting with 8.2 / 9.0 and re-use some of that code?

awelzel (approved):
Approving, but note comment about this existing in Spicy, too.

src/util.h
timwoj: I agree that it'd be nice, but how does that work if someone passes `--disable-spicy` to `configure`?

awelzel: From the earlier message:

> I'm wondering if we should just depend on Spicy starting with 8.2 / 9.0 and re-use some of that code?

I didn't look in detail and it's probably not be a drop-in replacement.

PR #5053, opened 2025-11-26
Completely rewrite the scripting basics section

evantypanski:
This is a PR on the doc-tutorial branch, not master

The content here is meant to start from network concepts, then explain just enough scripting to be dangerous. It's partly informed by my poll in the Zeek slack, people seem to start with Zeek at a lot of different points with understanding programming (except mostly not absolute beginners it seems). The only connecting thread is probably "I want to do something on my network" so I leaned into that a bit more than the current basics section does (like.. I explain `addr` before `int`)

I rewrote it because:

1) I really don't like how long the original section is. After the rewrite it's ~4500 words with code, before it was ~13500
2) I can make a consistent story, we are trying to get to a point to understand 1 script
3) A lot was repeated with script reference, so I tried to make the basics section "show not tell" then link to the reference section for details (solidified by @ckreibich [mentioning it](https://github.com/zeek/zeek/pull/4928#discussion_r2544081366))

I think this is much nicer, I'd actually be happy to link this as a good way to get started Zeek scripting now.

### DIFF
--- a/doc/script-reference/operators.rst
+++ b/doc/script-reference/operators.rst
@@ -271,7 +271,7 @@ field name must be in the declaration of the record type.
   * - Field value existence test
     - ``a?$b``
     - Evaluates to type :zeek:type:`bool`.  True if the specified field has
-      been assigned a value, or if not.
+      been assigned a value, or false if not.
 
 .. _pattern-operators:
 


doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
+We have already seen the main *output* from Zeek: logs. But, Zeek has a
+whole layer designed to define the logic that creates those logs (and
+more!). That is Zeek's scripting language.
 
-   Zeek's notions of originator and responder aim to capture the natural
-   roles of connection endpoints given the protocol information
-   observed. They differ from the packet-level concepts of source and
-   destination, as well as from higher-level abstractions such as client
-   and server.
+We have already seen parts of Zeek script. The :ref:`Providing Script
```
awelzel: Total nit, the previous paragraph starts with the same 4 words.

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
-   will flip the endpoint roles, making the sender of this packet the
-   connection's responder.
+Zeek scripting is also the core of *detection* logic. You can use Zeek's
+scripting language to react to network events, store state about those
+events, and then inform incident responders.
 
-Zeek defines events for the primary parts of the connection life-cycle,
-such as the following:
+We will demonstrate this with a high level example. For this, we will
+check if the network traffic contains any malware from the `Team Cymru
```
awelzel: slight wording: "contains any known malware using the Team Cymru"

doc/tutorial/scripting/basics.rst
ckreibich, line 17: This had me wondering whether we should just start calling it Zeekscript. We always dance around this with writing in "Zeek's scripting language", "in Zeek", etc.

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
-variable. If ``EXPRESSION`` is the same type as ``TYPE``, the two forms
-should be equivalent. The decision as to which type of declaration to
-use is dictated by personal preference and readability.
-
-.. literalinclude:: basics/data_type_declaration.zeek
-   :caption:
-   :language: zeek
-   :linenos:
-   :tab-width: 4
+The body of our event does two things:
```
ckreibich:
```suggestion
The body of our event handler does two things:
```
In the previous version you called out the difference between event and event handler. It'd be good to keep those notions separate, even this early.

doc/tutorial/scripting/basics/types_time.zeek
```diff
@@ -0,0 +1,20 @@
+event zeek_init()
+	{
+	# Setup some variables. This uses current_time() since time constants don't
+	# exist in Zeek.
+	local time_spotted: time = current_time();
+	sleep(1sec); # Waits for a second to put time between two current_time calls
```
ckreibich: Just an idea to avoid using `sleep()`, since it's so unusual in Zeek — turn it around and show computation on time:
```
    local t0: time = current_time();
    local t1: time = t0 + 1 sec;
```
or some such.

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
+For more information, see :zeek:see:`vector`, :zeek:see:`set`, and
+:zeek:see:`table`.
 
-.. note::
+Record Types
+============
 
-   You may choose to declare some sets with the ``&ordered`` attribute
-   in order to keep the elements ordered.
+Records are just collections of named values---like a ``struct`` in C.
```
ckreibich, line 154: Thank you for the `---` :-)

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
+:zeek:see:`zeek_done` event:
+
+.. literalinclude:: basics/scope_global.zeek
    :caption:
    :language: zeek
    :linenos:
    :tab-width: 4
 
-.. code-block:: console
+For more information, see :zeek:see:`local` and :zeek:see:`global`.
```
ckreibich, line 286: Two thoughts came to mind here, (1) whether you'd like to touch declaration vs initialization, (2) whether we should say that loop variables (`for ( x, y in my_vector ) ...`) implicitly define locals, which can trip you up if you want to say `local x ...` later on, after that block.

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
-a 20 bit subnet mask.
-
-.. literalinclude:: basics/data_type_subnets.zeek
+With this change, nothing will print---all addresses were allowed.
+
+Redef
+=====
+
+We can change more than just native variables. In fact, you can use
+:zeek:see:`redef` to do far more. Here, we will use it to demonstrate one
```
ckreibich: We should mention the common thread to redefs, whether as a statement or attribute: to make initialization-time changes that are set in stone (and cannot be done) at runtime. "Redefinition" actually captures that notion pretty well.

doc/tutorial/scripting/basics.rst
```diff
@@ -1,1463 +1,609 @@
...
+**********************
+
+Events
+======
+
+Throughout this section, we have used events such as :zeek:see:`zeek_init`
+and :zeek:see:`new_connection`. Zeek itself executes through a series of
+events in a queue. For example, when Zeek sees an HTTP request in some
+network traffic, it triggers the :zeek:see:`http_request` event. The HTTP
+analyzer may then use that event in order to gather data, correlate a
```
ckreibich: I suggest not using "HTTP analyzer" here, since we generally think of the HTTP analyzer as creating that event in the first place. Perhaps "Script-layer HTTP analysis may then" or some such.

Btw — I think it would be super helpful to add a quick diagram of Zeek's basic processing loop, in essence (1) process packet, (2) build up events, (3) drain event queue, (4) repeat. (Needn't be in this PR!) We've never really had that — the closest I can think of is architecture.png which captures the overall design. This would immediately convey that your scripts execute in a tight processing loop, not somehow concurrently etc — especially since below you say they execute "later".

evantypanski:
> We're not covering basic statements and control flow. We should probably at least refer out to https://docs.zeek.org/en/master/script-reference/statements.html to say that common control-flow constructs exist in Zeek?

Sure, I can link out, but that is actually very intentional. It's problem-first: do you need to test if two things are equal then do something? Here's an `if` statement.

One reason it's not a separate section is because Zeekscript has multiple different "ways" of looping over something - each set, vector, and table have a different way (ie `for ( i in my_container )` will have different ideas of what `i` will be, is it an index? or a pair? the element? who knows!). I don't think it's even useful to teach people "here's a for loop" - instead, I teach "here is how to go through elements of a vector" and it happens to be a for loop.

that said, I can link out for people who want to know more.

> We're not touching the concept of attributes.

That's also intentional to some extent. Attributes don't do just one thing. Attributes are more organized in a "syntax" way to the user. I'm trying to avoid exactly that kind of teaching ("here is the syntax to loop") by introducing them as solutions to problems rather than a separate thing

Maybe we can say more words on what attributes are in a note or something, but I want it to be separate from the "main flow" of the tutorial, personally.

> The notion of state-keeping.

I think this would be a good addition. I didn't add it because it's not going towards the goal of understanding the detect-MHR script. But, I'm not opposed to one or two sections after that teach useful/common high-level problems and their solutions in Zeek, that's exactly what I'd want to enable

though: the next section of the scripting tutorial *does* do state keeping, but it doesn't touch on expiration

doc/tutorial/scripting/basics.rst
evantypanski: Maybe we have to change the name of https://github.com/zeek/zeekscript first?

doc/tutorial/scripting/basics.rst
evantypanski: > whether you'd like to touch declaration vs initialization

Maybe we can find a way to sneak in a variable that is declared but not initialized. I would rather not introduce that jargon, but the concept would be ok. Maybe as a followup we can think through what important concepts are missing that we can add not-so-subtle hints about.

> whether we should say that loop variables (for ( x, y in my_vector ) ...) implicitly define locals

Sure, I'll also use that to mention that local variables can't be "redeclared"

PR #5055, opened 2025-11-28
Allow port redefs for built-in analyzers

J-Gras:
A second try on https://github.com/zeek/zeek/pull/4244.

### DIFF
--- a/testing/btest/Baseline/plugins.hooks/output
+++ b/testing/btest/Baseline/plugins.hooks/output
@@ -17,7 +17,7 @@
 0.000000   MetaHookPost  CallFunction(Analyzer::register_for_port, <frame>, (Analyzer::ANALYZER_HTTP, 8080/tcp)) -> <no result>
 0.000000   MetaHookPost  CallFunction(Analyzer::register_for_port, <frame>, (Analyzer::ANALYZER_HTTP, 81/tcp)) -> <no result>
 0.000000   MetaHookPost  CallFunction(Analyzer::register_for_port, <frame>, (Analyzer::ANALYZER_HTTP, 8888/tcp)) -> <no result>
-0.000000   MetaHookPost  CallFunction(Analyzer::register_for_ports, <frame>, (Analyzer::ANALYZER_HTTP, {80<...>/tcp})) -> <no result>
+0.000000   MetaHookPost  CallFunction(Analyzer::register_for_ports, <frame>, (Analyzer::ANALYZER_HTTP, {80<...>/tcp}, {})) -> <no result>
 0.000000   MetaHookPost  CallFunction(Broker::__subscribe, <frame>, (zeek/supervisor)) -> <no result>
 0.000000   MetaHookPost  CallFunction(Broker::subscribe, <frame>, (zeek/supervisor)) -> <no result>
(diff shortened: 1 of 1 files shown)


J-Gras:
Looks like the external tests fail due to a DoQ connection that gets flipped now, because QUIC ports are automatically added to the `likely_server_ports`.

scripts/base/frameworks/analyzer/main.zeek
```diff
@@ -232,16 +237,25 @@ function disable_analyzer(tag: AllAnalyzers::Tag) : bool
...
 		}
 
+	for ( p in non_server_ports )
+		{
+		if ( ! register_for_port(tag, p) )
+			rc = F;
+		}
+
+	# Automatically update likely_server_ports with server_ports
+	likely_server_ports += server_ports;
```
awelzel, line 257: Interesting, I think I never realized that `const &redef` would allow you to do that, but i guess that makes some sense. It breaks my mental model about anything `const &redef` not changing at runtime though, but I might have been too naive :-)

Technically this needs a new BiF to invalidate the likely_server_ports cache in the core, or better transport pass the full likely_server_ports "into" the core.

https://github.com/zeek/zeek/blob/b2488808ffd1eca98412e96a7fcabdf5e71a8ef2/src/packet_analysis/protocol/ip/IPBasedAnalyzer.cc#L144-L155

At the same time, unclear if `register_for_ports()` at runtime has an actual use - there's no corresponding `unregister_...` .

scripts/base/protocols/dns/main.zeek
```diff
@@ -163,8 +163,7 @@ redef record connection += {
 	dns_state: State &optional;
 };
 
-const ports = { 53/udp, 53/tcp, 137/udp, 5353/udp, 5355/udp };
-redef likely_server_ports += { ports };
+const ports = { 53/udp, 53/tcp, 137/udp, 5353/udp, 5355/udp } &redef;
```
awelzel, line 166: Sorry for not looking earlier. These would be nicer in the `export` section like it's done for `Finger` below.

As is, the user cannot do:
```
redef DNS::ports += { 1153/udp };
```
and instead needs to do:
```
module DNS;
redef ports += { 1153/udp };
```

awelzel (commented):
Hey @J-Gras - sorry this was ignored for the past weeks.

I'm okay merging as is (will update the baseline) - with QUIC becoming pretty popular, having 443/udp in likely_server_ports seems okay. The mechanism is broad, so 🤷‍♂️ 

I have two comments that are totally fine as follow-ups or towards 8.2. One about the likely_server_ports caching in the core. Just a technical detail. If register only ever happens during zeek_init() it's fine.

Moving the individual port variables into an `export` section would feel nicer from a UX perspective (don't need to reopen the module), or were you consciously "hiding" them a bit?
```
# This doesn't work.
redef SSL::ports += { 11443/tcp };

# This works.
module SSL;
redef ports += { 11443/tcp };
```
The Finger analyzer has it in the export section, so might make sense to be consistent.

scripts/base/protocols/dns/main.zeek
awelzel: I see you're on PTO. I'll do the follow up for this one.

awelzel:
Rebased and included my feedback in a new PR #5093 - closing this one.

PR #5057, opened 2025-12-01
Fix error when table type cannot be determined (ie missing `[]` around key)

### LINKED ISSUE #5048: Table initialization behavior without `[]` around keys is unintuitive
I was working on a new scripting tutorial and accidentally forgot square brackets around the keys, like:

```
	local asset_names: table[addr] of string = table(192.168.1.1 = "Router", 8.8.8.8 = "Google DNS");
```

This parses fine, but if you ever try to use `asset_names`, it's apparently uninitialized!

```
expression error in ./types_table.zeek, line 6: value used but not set (asset_names)
```

Here's another way to see it: the first print doesn't even print *anything*:

```
event zeek_init()
	{
	# Prints nothing. Not even an empty table.
	print table("SSH" = 22/tcp, "HTTPS" = 443/tcp);
	# Prints the table.
	print table(["SSH"] = 22/tcp, ["HTTPS"] = 443/tcp);
	}
```

```
$ zeek table-init.zeek
{
[SSH] = 22/tcp,
[HTTPS] = 443/tcp
}
```

This is very unintuitive for users. I consider this a bug, we should either produce some sort of error or allow the keys to not be wrapped in square brackets. Silently failing is just asking for people to get annoyed.

I would prefer we don't force square brackets around single-type keys.

evantypanski:
Fixes #5048

Since this is apparently a bug (ie it errored, just didn't have a message) I'll leave further work out for later. In the future, we could allow no square brackets around the keys and just promote each to a list. I guess this was "just a bug"

I added a little hint about what could've gone wrong, but can reword or remove that if it seems wrong too.


timwoj:
We didn't have any test coverage on this before:

Screenshot 2025-12-01 at 4 03 42 PM

evantypanski:
oh, I think that will still be untested. The reason, though, is I think that condition is unreachable because of the check in this change. I can dive a bit further into why, I didn't look much at why `init_type(op)` was resulting in a falsey type

timwoj:
> oh, I think that will still be untested. The reason, though, is I think that condition is unreachable because of the check in this change. I can dive a bit further into why, I didn't look much at why `init_type(op)` was resulting in a falsey type

Ah, you're right. I linked to another place that we don't set an error message 😬

timwoj (approved):

PR #5059, opened 2025-12-02
Use ranges versions of all_of, any_of, etc.

Neverlord:
Modernizing low-hanging fruits to use less-noisy C++20 versions. Not pulling in `<ranges>` yet, only sticking to new overloads in the `<algorithm>` header.

### DIFF
--- a/src/DNS_Mapping.cc
+++ b/src/DNS_Mapping.cc
@@ -161,8 +161,8 @@ void DNS_Mapping::Save(FILE* f) const {
 }
 
 void DNS_Mapping::Merge(const DNS_MappingPtr& other) {
-    std::copy(other->names.begin(), other->names.end(), std::back_inserter(names));
-    std::copy(other->addrs.begin(), other->addrs.end(), std::back_inserter(addrs));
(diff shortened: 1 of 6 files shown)


timwoj:
I did include `<ranges>` a bit in other places, so feel free to do it here too.

timwoj:
I'm also sort of surprised that clang-tidy's modernize-use-ranges didn't catch any of these.

Neverlord:
> I did include `<ranges>` a bit in other places, so feel free to do it here too.

I didn't mean it in a sense of "avoiding `<ranges>`". That was simply me poking a bit around for low hanging fruits where we can use the same algorithm without the `x.begin(), x.end()` overhead. The `<ranges>` header has more to offer like views and actions that we could use to rewrite things in a more declarative style but that would not be a trivial replacement anymore. 🙂

> I'm also sort of surprised that clang-tidy's modernize-use-ranges didn't catch any of these.

Yeah, they're all straightforward. 🤷‍♂️

timwoj (approved):

PR #5062, opened 2025-12-03
Fix missing `Packet` object member resets in `Packet::Init()`

ckreibich:
The `Packet::Init()` method didn't reset all packet fields, which could lead to side effects across packet processing iterations. I'm also adding a unit test (which isn't super great since it doesn't take into account future fields — but it does fail prior to the fix), and I'm reordering the initializations to better match the class layout.

@cccs-jsjm reported the missing reset for the `session` field — thank you!

I think we should backport this into 8.0.x for the applicable fields in that branch.

### DIFF
--- a/src/iosource/Packet.cc
+++ b/src/iosource/Packet.cc
@@ -43,28 +43,32 @@ void Packet::Init(int arg_link_type, pkt_timeval* arg_ts, uint32_t arg_caplen, u
     else
         data = arg_data;
 
-    dump_packet = false;
-    dump_size = 0;
-
     time = ts.tv_sec + double(ts.tv_usec) / 1e6;
     eth_type = 0;
+
     vlan = 0;
     vlan_pcp = 0;
     vlan_dei = false;
+
     inner_vlan = 0;
     inner_vlan_pcp = 0;
     inner_vlan_dei = false;
 
+    l3_proto = L3_UNKNOWN;
+
     is_orig = false;
+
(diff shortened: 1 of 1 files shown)


awelzel (approved):

src/iosource/Packet.cc
```diff
@@ -197,3 +208,75 @@ ValPtr Packet::FmtEUI48(const u_char* mac) const {
...
+        p.processed = true;
+        p.dump_packet = true;
+        p.dump_size = 1;
+        p.encap = std::make_shared<zeek::EncapsulationStack>();
+        p.ip_hdr = std::make_shared<zeek::IP_Hdr>(nullptr, false);
+        p.proto = 1;
+        p.tunnel_type = BifEnum::Tunnel::IP;
+        p.gre_version = 1;
+        p.gre_link_type = DLT_EN10MB;
+        p.session = reinterpret_cast<zeek::session::Session*>(1);
```
awelzel, line 246: 🤞 about what UBSAN will think :-)

ckreibich: I was nervous too :-) — and admittedly a bit too lazy to make up a full `Connection` instance. Let me do a full CI run.

ckreibich:
Well, the sanitizers turn out to be happy but Windows is not. I'll try a few things, and rebase while I'm at it to get rid of the unrelated failures.

PR #5068, opened 2025-12-05
Analyzer confirmation callbacks

vpax:
A common pattern in Zeek scripting packages is to have a `analyzer_confirmation_info` event handler with the actual logic conditioned on the confirmation being for only 1 or 2 different types of analyzers. This PR adds a mechanism for scripts to provide per-analyzer-type (tag) callbacks, allowing them to forgo such handlers (which are invoked a _lot_). The callbacks execute before the corresponding events, allowing them to influence subsequent event handling if needed.

The commits (1) add the functionality, (2) add BTests to confirm that it's working, (3) update some base scripts to use callbacks rather than event handlers, (3) update BTests reflecting these changes (and also ZAM's knowledge of the built-in).

FYI I estimate that 90-95% of the changes in this PR were generated by Claude Code (while I was off doing other things), using a not-large number of prompts on my end (let me know if you're interested in seeing the list). Its changes had some shortcomings and style issues, some of which I got it to fix directly with more prompting, and some of which I addressed by hand (I've gone through all of the code to vet it, and correct as needed).


awelzel:
I do like the idea a lot!

The synchronous function calls to Zeek script while we're still doing packet processing - i.e. we're not yet in the event draining phase - is making me pause. I'd prefer we'd keep Zeek script execution to the event draining phase and not introduce direct script function callbacks into the packet processing / analyzer path.

So I'm wondering what you'd think about:

1) Go from `Analyzer::register_confirmation_callback` to `Analyzer::register_confirmation_event` and pass a `arg: event(tag, info)` instead of a function. Then *enqueue* the registered event at confirmation time rather than invoking it synchronously. As long as it is enqueued before the generic `analyzer_confirmation_info()` event, the execution order should be as wanted and it feels more natural. This would also allow users to implement their own `event rdp_analyzer_confirmation_info(...)` handlers, rather than registering a function via a bif and internally we'd only need a single EventHandlerPtr (?) per analyzer tag.

Using events also voids the synchronous call concern and again, just feels more like using existing patterns, rather than the callback registration, of which we don't have many.

2) Implement the callback registry in Zeek script and execute the logic from a generic  `analyzer_confirmation_info()` handler:
```
global confirmation_callbacks: table[AllAnalyzers::Tag] of vector of AnalyzerCallback &default_insert=vector();

event analyzer_confirmation_info(tag, info) &priority=1000
    {
    if ( tag in confirmation_callbacks )
        {
        for ( _, cb in confirmation_callbacks[tag] )
             cb(tag, info);
        }
    }

Analyzer::register_confirmation_callback(tag, f)
   {
   confirmation_callbacks[tag] += f;
   }
```
Even here, using hooks is probably  the more Zeeky way for "multi-body-functions" instead of registering multiple functions (and break would have no effect or should by convention not be used ).

I certainly like 1) more due keeping the confirmation event an actual event. I suspect the one quirk here is that we'd need to use `ev: any` for the register bif, as IIRC we cannot use `ev: event(tag: AllAnalyzers::Tag, info: AnalyzerConfirmationInfo)` as a parameter :-(

Approach 2) could be implemented out-of-tree as well, but I'm actually very sympathetic to the idea of analyzer specific confirmation events/hooks in-tree. If we do this, we should probably update the package-template to create hooks/events when creating new zkg packages.

> using a not-large number of prompts on my end (let me know if you're interested in seeing the list)

I'm curious. Does it also run btests on your system autonomously while developing? I'm using AI tools as a very powerful Google, but haven't gotten towards coding for me yet. Any chance you might give a demo either internally or potentially in a community call about that workflow?

vpax:
You make good points about function callbacks being a bit weird. Doing it all in scripting seems pretty good to me, actually (I'm curious why you prefer approach 1 to it). That said, registering events rather than functions seems equally fine too.

I've pushed changes for the latter. For the interested, the development process was the following, with prompts to Claude in italics:

1. _Change the interface so that instead of using function callbacks, instead scripts register event handlers, and the confirmation is done by raising those events. Be sure to (1) re-read the source because some edits have been made since you last modified it, (2) update comments to reflect the change to events, (3) update the scripts you modified, (4) update the new BTests, which are bifs/disable_analyzer-callback.zeek and bifs/register_confirmation_callback.zeek_
2. Scrutinize the diffs, spot only one issue:
3. _For registering event handlers, you used a string to identifer the handler.  Instead, this should be directly typed. Introduce a named type for doing so similar to how the code used to have `AnalyzerConfirmationCallback` defined in init-bare.zeek_
4. Scrutinize the diffs, looks good. Whoops though it crashes ...
5. ... due to a bug I introduced yesterday trying to address the signed/unsigned issue flagged by CI. Fix that
6. Update ZAM function-tracking (forgot to include that in my prompt above)
7. Look into BTest differences, one of them took some code-diving to convince myself it was correct
8. Trim some overly cautious error-checking
9. Send in the PR

The Claude steps all chugged along while I was on Zoom meetings :-)

vpax:
@awelzel Thinking about it, your approach 2 seems like a no-brainer. Soooo much simpler! Wish I had thought of it originally.

awelzel:
> @awelzel Thinking about it, your approach 2 seems like a no-brainer. Soooo much simpler! Wish I had thought of it originally.

I sort of assumed you were trying to squeeze cycles and that's why you went for vectors and C++. I'm certainly not against function callbacks and just doing the table lookup in script land. We can get the int value for a tag in script land, right? Then we could also use vectors for slightly more efficient lookup.

I think this is a nice idea and should be discussed with a few others from the team. If function callbacks over events make more sense, maybe we should just go with approach 2), it just seemed that "<protocol>_analyzer_confirmation()" is an event and events give "multi-body" functions for free (oh and we actually can use event types as bif parameters, nice!)

scripts/base/protocols/rdp/main.zeek
```diff
@@ -277,13 +270,20 @@ event file_over_new_connection(f: fa_file, c: connection, is_orig: bool) &priori
 		}
 	}
 
-event analyzer_confirmation_info(atype: AllAnalyzers::Tag, info: AnalyzerConfirmationInfo) &priority=5
+event rdp_analyzer_confirmation(tag: Analyzer::Tag,
```
awelzel, line 273: If the handler doesn't check the `tag`, we should probably just drop it from the API.

vpax: We could consider that. An advantage of including it is if you want to use the same handler for multiple protocols, and it needs to make some minor changes depending on the specific protocol. That said, that's probably a rare case, and can be dealt with by using "feeder" handlers, one per protocol, that supply the protocol to the common handler.

vpax:
> I sort of assumed you were trying to squeeze cycles and that's why you went for vectors and C++

That wasn't the actual motivation for that approach, I think instead I'd (1) not absorbed how connection_state_remove handlers were implemented, (2) was beguiled into wondering what else I could get Claude to do on my behalf.

Regarding the latter, after your comments I asked it to do the script-based approach instead. It fared much less well, which was a bit surprising (much simpler task, and I pointed it at the connection_state_remove handler approach as a pattern) ... and not (it simply doesn't really grok much Zeek).

> I think this is a nice idea and should be discussed with a few others from the team.

Please wrangle them up! (In what some googling shows me is the North American English notion of "wrangle", to whit: "round up, herd, or take charge of (livestock)" and um you can ignore the "livestock" part ;-).

awelzel:
> Please wrangle them up!

I had placed it on the list for discussion, but it didn't quite grab much excitement.

> Thinking about it, your approach 2 seems like a no-brainer. Soooo much simpler! Wish I had thought of it originally.

I'm starting to think the same. A script-only `Analyzer::register_confirmation_callback(tag, callback)` might be more reasonable: It doesn't require a new convention for the analyzer ecosystem and there's no need to update any existing internal/external analyzers and no "risk" of diverging conventions. It's also a much less involved change.

Should we just do the callback registry instead?

vpax:
> Should we just do the callback registry instead?

Yep! I think it's for sure the best solution. Will you be adding that to your queue? In any case, I'm closing this PR.

PR #5070, opened 2025-12-05
Deprecate the MD5 IP anonymizers, add SHA256 ones

timwoj:
This PR is part of the work for https://github.com/zeek/zeek/issues/2505. This deprecates the MD5 IP anonymizer classes and script-related functionality. It adds new SHA256 versions as replacements. It also deprecates the HMAC-MD5 functionality and moves the initialization of the keys for that functionality to happen lazily instead of at startup. Lastly, it adds some btests around the anonymization BIFs since we were missing that coverage previously.

I can see removing the lazy initialization bit, since the uses of the key data are all deprecated now. Once those uses are removed in v9.1, then we can remove the initialization as a whole as well.


src/Anon.cc
```diff
@@ -329,16 +340,67 @@ AnonymizeIPAddr_A50::Node* AnonymizeIPAddr_A50::find_node(ipaddr32_t a) {
...
+
+        // HK(PAD(x_0 ... x_{i-1})).
+        util::detail::hmac_sha256(sizeof(prefix), reinterpret_cast<u_char*>(&prefix), digest);
+
+        // f_{i-1} = LSB(HK(PAD(x_0 ... x_{i-1}))).
+        ipaddr32_t bit_mask = (digest[0] & 1) << (31 - i);
+
+        // x_i' = x_i ^ f_{i-1}.
+        output ^= bit_mask;
+    }
```
awelzel, line 379: I wondered a bit if a template that takes the hmac function would make sense, or are you anyhow thinking "well, in 9.1 I'll rip out MD5 code anyhow"?

scripts/base/init-bare.zeek
```diff
@@ -1430,9 +1430,11 @@ type PcapFilterID: enum { None };
 type IPAddrAnonymization: enum {
 	KEEP_ORIG_ADDR,
 	SEQUENTIALLY_NUMBERED,
-	RANDOM_MD5,
 	PREFIX_PRESERVING_A50,
-	PREFIX_PRESERVING_MD5,
+	RANDOM_MD5 &deprecated="Remove in v9.1. Use the A50 or SHA256 anonymizers instead.",
```
awelzel, line 1434:
```suggestion
	RANDOM_MD5 &deprecated="Remove in v9.1: Use the A50 or SHA256 anonymizers instead.",
```

Hmm, we should come up with a convention here. Seems you're using the `.` consistently. So, feel free to ignore.

src/Hash.cc
```diff
@@ -38,21 +38,46 @@ void KeyedHash::InitializeSeeds(const std::array<uint32_t, SEED_INIT_SIZE>& seed
...
-                 shared_hmac_md5_key); // The last 128 bits of buf are for siphash
-    // yes, we use the same buffer twice to initialize two different keys. This should not really be
-    // a security problem of any kind: hmac-md5 is not really used anymore - and even if it was, the
-    // hashes should not reveal any information about their initialization vector.
+    // Save off the seed data so it can be used to lazily calculate the hmac_md5 key if
+    // needed.
+    memcpy(KeyedHash::seed_data, seed_data.data(), sizeof(seed_data));
+
+    // From the 80-byte seed:
+    // - Bottom 64 bytes are used for hmac_md5 and shared_highwayhash_key.
```
awelzel: I'd probably say "first 64 bytes (bytes 0-63), but then we probably won't look at this all that much.

I do wonder also about a `static_assert(sizeof(seed_data) == 80)` or on the `SEED_INIT_SIZE` so that this comment doesn't become stale without a compile error, but might not be worth it.

testing/btest/bifs/anonymize_addr.zeek
```diff
@@ -0,0 +1,63 @@
...
+event zeek_init()
+	{
+	local a: addr = 1.2.3.4;
+	local b: addr = 1.2.3.5;
+
+	print anonymize_addr(a, ORIG_ADDR);
+	print anonymize_addr(b, ORIG_ADDR);
+	print anonymize_addr(a, RESP_ADDR);
+	print anonymize_addr(b, RESP_ADDR);
+	}
```
awelzel, line 65: Would it make sense to add a test where the same addr is passed twice to anonymize_addr() ? Just to ensure it yields the same anonymization.

testing/btest/bifs/anonymize_addr.zeek
```diff
@@ -0,0 +1,63 @@
...
+
+global orig_addr_anonymization: IPAddrAnonymization = PREFIX_PRESERVING_A50;
+
+event zeek_init()
+	{
+	local a: addr = 1.2.3.4;
+	local b: subnet = 1.2.3.0/16;
+
+	preserve_subnet(b);
+	print anonymize_addr(a, ORIG_ADDR);
```
awelzel, line 36: Might make sense to add one addr here that doesn't match the prefix, like `1.3` ?

awelzel (approved):

src/Anon.cc
timwoj: Definitely the latter.

testing/btest/bifs/anonymize_addr.zeek
timwoj: Interestingly, that's broken. I'll fix it in a follow-up.

PR #5071, opened 2025-12-07
Build Javascript bindings in CI on macos

bbannier:


awelzel:
> Thanks @awelzel, with the zeekjs bump this seems to work now. :shipit:

I'll merge. Thanks for doing this! :rocket:

Issue #5078, opened 2025-12-09
ssl.log will miss certificate fingerprints when a protocol error occurs


0xxon:
`ssl.log` currently will not have any certificate fingerprints, when a protocol error occurs after the certificates have been encountered on the connection, but before the connection is established. This is due to the necessary hooks not being called in this case.

PR #5079, opened 2025-12-09
Fix ssl certificate logging when encountering protocol errors

0xxon:
This change fixes a problem with ssl certificate logging when a protocol error is encountered after certificates are sent over a SSL/TLS connection, but before the connection is logged. Currently, in these cases, `ssl.log` will not contain the fingerprints of the certifiates contained in the connection. This is due to the necessary hook not being called.

Test pcap is hex-edited frome existing pcap.

Fixes GH-5078


0xxon:
I think we should consider pulling this one into 8.0. Taking a bit of a liberal view, this is a regression caused by the `ssl.log` changes in Zeek 4.1.

bbannier:
This change has broken the [build where we enable the in-tree Spicy SSL analyzer](https://cirrus-ci.com/task/6391860563804160).

PR #5080, opened 2025-12-09
Error on &broker_store and &backend when a non-Broker cluster backend is selected

awelzel:


src/cluster/Manager.cc
```diff
@@ -2,13 +2,45 @@
...
+    const auto& globals = zeek::detail::global_scope()->Vars();
+    std::string x509_known_log_certs_with_broker = "X509::known_log_certs_with_broker";
+
+    for ( const auto& [name, id] : globals ) {
+        if ( ! id->HasVal() )
+            continue;
+
+        // Remove in v9.1: This one is only used when a deprecated option is
+        // set that will only work with Broker anyhow. Not overthinking this.
+        if ( id->Name() == x509_known_log_certs_with_broker )
```
0xxon, line 31: lol :)

src/cluster/Manager.h
```diff
@@ -12,6 +12,18 @@
...
 namespace zeek::cluster {
 
+namespace detail {
+
+/**
+ * Iterate over all global variables and report these with
+ * &broker_store or &backend attributes as non functional.
+ *
+ * Only call this if Cluster::backend is not Broker and not None.
+ */
```
0xxon, line 24: should this have a comment that the entire function can be removed in 9.1, because the attributes go away then too?

src/cluster/Manager.cc
```diff
@@ -2,13 +2,45 @@
...
+#include "zeek/Scope.h"
 #include "zeek/cluster/Serializer.h"
 #include "zeek/cluster/websocket/WebSocket.h"
 #include "zeek/util.h"
 
+#include "Desc.h"
+
 using namespace zeek::cluster;
 
+void detail::report_non_functional_broker_tables(const zeek::EnumValPtr& cluster_backend_val) {
```
0xxon, line 15: this is probably a bit overcautious - but should this have an assert checking that the backend is != broker?

awelzel: Yeah, I can double check that one.

PR #5083, opened 2025-12-09
Fix FuncInfo flag for ZAM for anonymize_addr BIF

timwoj:
Running these tests under ZAM results in this error:

```
  internal error: IP anonymizer not initialized
```


timwoj:
@vpax Does it make sense that this wouldn't be initialized when running under ZAM?

vpax:
> Does it make sense that this wouldn't be initialized when running under ZAM?

Not in general. What could be happening is that there's a BiF that ZAM is calling at compile-time because of incorrect attributes assigned to it in `script_opt/FuncInfo.cc`. Let me know if you'd like me to look further

ckreibich (approved):

Issue #5077, opened 2025-12-09
Proposal: enable SHA256 file hashing in hash-all-files


0xxon:
In https://zeekorg.slack.com/archives/CSZBXF6TH/p1765294211015129?thread_ts=1765293511.460229&cid=CSZBXF6TH, @mmguero points out that SHA256 seems to be the standard for file hash lookups.

We should consider enabling it by default in `hash-all-files.zeek`. This also feels somewhat related to GH-2505.

mmguero:
as far as sha256 being commonly used, one example is Google Threat Intelligence:

<img width="2560" height="1441" alt="Image" src="https://github.com/user-attachments/assets/7ce06781-dfa9-4dd5-a09b-2c60973099c3" />

Files are identified first by their sha256 sum. You can get the sha1 and md5 sums but they're buried deeper in each file IOCs page. I believe from what I've seen pulling IOCs with the Google threat intelligence API, sha256 sums seem to be what I'm getting.

PR #5086, opened 2025-12-10
Add SHA256 to hash-all-files policy script

0xxon:
SHA256 is now commonly used for IOCs. See GH-5077.

Requires update of external test-suites.


mmguero:
Thanks @0xxon !

PR #5087, opened 2025-12-10
cirrus: Try double star matching for changesIncludeOnly()

awelzel:
In the original discussion [1], they used 'doc/**', but the docs now show 'doc/*'. Try the former and if it works I'll update their documentation.

[1] https://github.com/cirruslabs/cirrus-ci-docs/issues/873#issuecomment-882889185


timwoj (approved):
I think they use traditional globbing for their paths here, so this is probably correct.

PR #5092, opened 2025-12-11
Add packet source that uses LightPcapNg to be able to read pcapng files

timwoj:
This is a packet source for reading data from pcapng files (see https://datatracker.ietf.org/doc/html/draft-ietf-opsawg-pcapng). It supports reading packets, and handling the interface and section header data from the files. It uses the https://github.com/Technica-Engineering/LightPcapNg library (MIT licensed) to read the blocks from the files. 

The only packet source feature not currently supported is BPF filtering with Zeek's `-f` argument. It still needs a btest as well. I have been testing with the file from https://weberblog.net/the-ultimate-pcap/. It's capable of fully reading that file and generates all of the expected logs from it.


src/iosource/pcapng/Source.h
```diff
@@ -0,0 +1,127 @@
...
+        std::optional<uint64_t> packet_id;
+        std::optional<uint32_t> queue;
+        std::optional<std::vector<std::string>> verdicts;
+        std::optional<std::string> processid_threadid;
+    };
+
+    void ParseSectionHeaderBlock(light_block block);
+    void ParseInterfaceBlock(light_block block);
+    PacketBlock ParseEnhancedPacketBlock(light_block block);
+    PacketBlock ParseSimplePacketBlock(light_block block);
```
awelzel: I'd suggest moving the parsing code into a separate class, possibly even a separate file (including making the enums somewhere accessible in a non-packet-source header (though maybe in the packet source directory for now). They seems kind of useful? (EDIT: Maybe in lightpcapng?)

I haven't thought much how exactly that would look like, but maybe something like that:

```
class PcapngParser {
    NextBlock(light_block* block);  // dispatch and raise events
}
```
...and then delegate in ExtractNextPacket() to an embedded instance if an instance of that or so.

The `Source.cc` file and `pivate:` list is pretty overwhelming. This looks more like a protocol parser than a packet source when opening it (indeed, if we hit the right abstraction, maybe a pcapng file analyzer would just fall into existence (though it should likely raise different events, or have some indication that it's not from the packet source - nothing to do now though).

Splitting the code would also make unit testing of just the parsing easier if we ever wanted to do that.

src/iosource/pcapng/Source.h
```diff
@@ -0,0 +1,127 @@
...
+#pragma once
+
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "zeek/iosource/PktSrc.h"
+
+struct light_file_t;
+using light_file = struct light_file_t*;
```
awelzel, line 12: Is there a specific reason you're doing this? This is also non-namespaced, so I'm a bit confused.

I might be old-school, but IMO typedefs should never hide `*`. 

This PR contains the following:

>         light_block sh_block = nullptr;
>    light_block current_block = nullptr;


and seems totally fine to use the verbose `struct light_block_t*` or a typedef if that adds something, but they are definitely missing a `*`, or is that just me?

src/event.bif
```diff
@@ -906,3 +906,21 @@ event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: str
 ##
 ## pkt: Data for the unprocessed packet
 event packet_not_processed%(pkt: pcap_packet%);
+
+## Generated for each new interface in a pcapng file.
+##
+## intf: A record containing data about the interface.
+event pcapng_new_interface%(intf: Pcapng::Interface%);
```
awelzel:
```suggestion
event pcapng_new_interface%(interface: Pcapng::Interface%);
```
It already says the full "interface" twice, so three times a charm, or just `i`? :-)

src/event.bif
```diff
@@ -906,3 +906,21 @@ event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: str
...
+
+## Generated when a section header block is found in a pcapng file. This contains various
+## fields about the file.
+##
+## info: A record containing the fields from the section header block.
+event pcapng_file_info%(info: Pcapng::FileInfo%);
```
awelzel: With Spicy parsers we started moving to a model where events are also module scoped. E.g. `Pcapng::new_interface()`. But I like the proposed `pcapng_<event>` more here. Mostly wanted to mention it if we ever should settle on a "proper" way for the future.

scripts/base/init-bare.zeek
```diff
@@ -6503,6 +6503,99 @@ export {
...
+		# The link type for the interface.
+		link_type: count;
+
+		snaplen: count;
+		# The name of the device used to capture data.
+		name: string &optional;
+		# A description of the device used to capture data.
+		description: string &optional;
+		# A list of IPv4 addresses assigned to the interface.
+		ipv4_addrs: vector of string &optional;
```
awelzel: Drive-by. Not sure if easy..

```suggestion
		addrs: vector of addr &optional;
```

src/event.bif
```diff
@@ -906,3 +906,21 @@ event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: str
 ##
 ## pkt: Data for the unprocessed packet
 event packet_not_processed%(pkt: pcap_packet%);
+
+## Generated for each new interface in a pcapng file.
```
awelzel: Mostly food for thought and related to comments elsewhere: This reads like an event description for a file analyzer. Maybe the de-conflicting is something we should think about a bit?

awelzel (commented):
Just some high level comments.

This is really exciting. I wonder a bit about a cool test or POC demo like adding the interface name stored in the pcapng file into `conn.log` or more fancy, logging in a per-interface directory or per-interface conn.log. I think we have enough hooks in place to pull this off :-)

scripts/base/init-bare.zeek
timwoj: This one is string because it's not just an address. It's an address and a netmask. Same with the ipv6, where it's an address and a prefix length.

src/iosource/pcapng/Source.h
timwoj: This is following the model that LightPcapNg uses internally. All of their functions take a version of the type with the `*` hidden.

awelzel: Ah, thanks. Hmm. I guess if it makes it consistent with the library... I'll try to ignore :-)

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,566 @@
...
+        }
+
+        if ( block->type == LIGHT_ENHANCED_PACKET_BLOCK )
+            break;
+        else if ( block->type == LIGHT_SIMPLE_PACKET_BLOCK ) {
+            // Report a weird for simple packet blocks instead of dropping them, for
+            // documentation reasons. We don't want to parse simple packet blocks because
+            // they don't contain timestamp information. Without it, there's not much we
+            // can do with them.
+            reporter->Weird("pcapng_simple_packet_block");
```
ckreibich: These will get rate-limited automatically, right? I ask because they will likely affect every packet if somebody recorded a trace that way.

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,566 @@
...
+        if ( pb.verdicts ) {
+            auto vec = make_intrusive<VectorVal>(id::string_vec);
+            for ( const auto& verdict : pb.verdicts.value() )
+                vec->Append(make_intrusive<StringVal>(verdict));
+            rec->Assign(6, vec);
+        }
+        if ( pb.processid_threadid )
+            rec->Assign(7, make_intrusive<StringVal>(pb.processid_threadid.value()));
+
+        event_mgr.Enqueue(pcapng_packet_options,
```
ckreibich: This is dangerous, no? Potential per-packet eventing?

ckreibich (commented):
Super cool! This is not a full review — I just wanted to flag a few high-level things. 

The `Pcap::file_done` event should probably fire for this packet source too? (I haven't checked for other pcap-related events, but for most uses the user likely doesn't care which specific file type they're dealing with.)

I'm unsure about handling any of the metadata in the pcap via eventing, due to the disconnect of events from the current packet. Have you given this any thought? For example, if we started to support custom blocks, I could see a need to alter some internal state/processing so that each subsequently processed packet is handled somehow differently (as opposed to doing something only when the event is processed). I don't mean to overthink it, but if we're not sure we immediately need those new events, we might not want to add them just yet.

src/iosource/pcapng/Source.cc
timwoj: It is potentially per-packet eventing, but this is also only reading from files so that's probably fine.

src/iosource/pcapng/Source.cc
timwoj: Yes, weirds get rate limited. I wanted something there to indicate that you're dropping packet data on the floor instead of processing it.

timwoj:
> I'm unsure about handling any of the metadata in the pcap via eventing, due to the disconnect of events from the current packet. Have you given this any thought? For example, if we started to support custom blocks, I could see a need to alter some internal state/processing so that each subsequently processed packet is handled somehow differently (as opposed to doing something only when the event is processed). I don't mean to overthink it, but if we're not sure we immediately need those new events, we might not want to add them just yet.

@awelzel brought up the idea of moving the eventing and whatnot to a file analyzer. If you want events and information about your pcapng file, you could run it through that, and leave the packet source just about parsing out packet data.

src/iosource/pcapng/Source.h
timwoj: I opened https://github.com/Technica-Engineering/LightPcapNg/pull/44 to push at least the constants upstream.

kayoub5: light pcapng have a low level api for reading the internal pcapng blocks, and a high level api for those interested only in packets `light_pcapng_ext.h`

typically I would recommend using the high level api, if there is an option not exposed in `light_pcapng_ext.h` api, PRs are welcome to add it.

timwoj: Part of the reason we're using the lower level API is that we're a bit performance-dependent, and looping over the entire array of options repeatedly via the internal calls to `light_get_option` isn't necessary. Another part is that we want some additional control over what happens with invalid option values for error reporting.

That said, I did find it semi-weird that the high-level API didn't just provide all of the option values from the calls.

src/iosource/pcapng/pcapng.bif
```diff
@@ -0,0 +1,15 @@
+%%{
+#include "zeek/iosource/Manager.h"
+#include "zeek/iosource/pcapng/Source.h"
+%%}
+
+function pcapng_current_interface_name%(%): string
```
awelzel, line 6: Related to the comments on event.bif, if we store interface description blocks in a global vector, then this could become:

```
pcapng_get_current_interface_id(): count


event new_connection(c: connection) {
  local name = Pcapng::current_section_interfaces[pcapng_get_current_interface_id()]$name;
  c$conn$interface_name = name;
}
```

I'm not a huge fan of tracking the interface names in the C++ layer. If someone doesn't want to use this feature, they shouldn't take the hit for tracking them (while extracting the interface id from a EPB or SPB is fast and easy). Were you just all head-down in C++ or did Zeek script not feel quite right? I think I might like it.

src/event.bif
```diff
@@ -906,3 +906,21 @@ event unknown_protocol%(analyzer_name: string, protocol: count, first_bytes: str
 ##
 ## pkt: Data for the unprocessed packet
 event packet_not_processed%(pkt: pcap_packet%);
+
```
awelzel: Commenting again on `event.bif` after looking at the RFC in a bit more detail.


In section 3.3 there is this:

```
3.3.  Logical Block Hierarchy

   The blocks build a logical hierarchy as they refer to each other.
   Figure 2 shows the logical hierarchy of the currently defined blocks
   in the form of a "tree view":

   Section Header
   |
   +- Interface Description
   |  +- Simple Packet
   |  +- Enhanced Packet
   |  +- Interface Statistics
```

And it would be nice if the event names would reflect exactly these names so that when you read the RFC that terminology directly applies to Zeek's events. It's also documented that any packet blocks follow a Interface Description Block and refer to it (EPB via an interface ID, SPB via implicit 0 ID).
So I'm thinking if the info from an interface description blocks could be appended to a global vector in script land that is reset whenever there's a new section_header_block. I haven't looked closely, but I believe you did something like that in C++. The Interface ID just increments with every new IDB implicitly, or that's how I understand it (Section 4.2), so the vector should be appropriate.

Further I'm thinking that the options (section 3.5) might be nicer in raw form. It seems pcapng supports custom and private options and exposing them all unparsed would give a way to expose them. A user could add their own options during capturing and they would be accessible to Zeek scripts through the event, even if the parser internally knows nothing about them.

```
type PcapngOption: record {
  opt_type: count;
  raw_val: string;
}
```

This comes at the expense of getting at individual options directly a bit harder (the comment, os_desc, hw_desc and all the interface fields), but we could munge those in script land, honestly. It seems `opt_comment` is allowed to come multiple times in a block and using rawer vectors would provide for that (the user could decide what they want to do).

This is much different from your current implementation where much of the option parsing is implemented in C++ (I only skimmed a bit). The idea here would be to expose the raw option data to the script layer that then interprets it. I like that more as more logic is in the script-layer and possibly more easily to iterate or improve when interpreting new options (no recompile cycle, no C++ intimidation for newcomers and it's basically a file analyzer). It'd be pretty big rework, so maybe should have a separate discussion session. I do think it'd make the C++ side much leaner though.




```
event pcapng_section_header_block(major: count, minor: count, options: vector of PcapngOption);
event pcapng_interface_description_block(link_type: count snaplen: count, options: vector of PcapngOption);
event pcapng_interface_statistics_block(interace_id: count, timestamp: time, options: vector of PcapngOption)
event pcapng_decryption_secrets_block(...);
event pcapng_name_resolution_block(...?);
event pcapng_unknown_block(block_type: count, data: string);


# Accessor to get the interface ID of the current packet's block.
function pcapng_get_current_interface_id(): count;
```

Also I think decryption_secrets_block could be interesting, even if we can't make use of the secrets at this point.

awelzel (commented):
Another long comment with thoughts only about the event signatures. I know this is a PR, but I think it's worth discussing how to expose the information in a bit more detail. I probably should've done that the first time around.

src/event.bif
timwoj: I feel like we've spent so much time spinning around on the events on this PR, when the point of the packet source should be about being able to read the packets and process them. At this point I'm inclined to just strip out all of the events so it can move forward.

awelzel: I was yesterday thinking the same about splitting this up in parts, so +1 from me.

timwoj:
Superseded by https://github.com/zeek/zeek/pull/5156.

ckreibich:
I'd like to add here that the point of this PR was never any pcapng-specific eventing (which clearly has a bunch of subtleties due to the various places in which custom content can live in a pcapng, and also due to its similarity to the classic how-to-flag-per-packet-stuff-into-the-scriptlayer problem) — it was just to enable Zeek to reliably process the traffic in a pcapng file.

So yeah — let's do that minimal bit first, and iterate on the rest later. I still hold that we should not normally be hitting design/scope conversations in mid-PR. It should be clear what to build (or at least what we _try_ to build!) when we open a PR.

PR #5093, opened 2025-12-12
Allow port redefs for built-in analyzers - take 3

awelzel:
Replaces #5055.

This adds the suggestions made on #5055, moving all port variables into export sections, trying to keep the naming a bit consistent. Excludes DoQ ports from likely server ports for now to avoid the baseline diff in the private testing suite and sudden flipped connections for anything 853/udp.

@bbannier / @timwoj - I was okay approving Jan's work to in #5055, could you check if my follow-ups make sense and then I'd merge for 8.1 :crossed_fingers:

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,19 @@ Breaking Changes
 New Functionality
 -----------------
 
+- Well-known ports of built-in analyzers can now be configured via redef. For example:
+
+	redef HTTP::ports += { 8181/tcp };
+
+  For protocol analyzers, the variable name is generally ``Analyzer_Module::ports``,
+  but in certain cases (SSL, QUIC, RDP, ...) the variable name includes the protocol
+  name to distinguish different internal analyzers. ``SSL::ssl_ports`` and ``SSL::dtls_ports``
+  as a concrete example. Check the analyzer module's ``export`` section or online documentation.
+  For packet analyzers, the analyzer name is included in the variable name for historical
+  reasons. Example:
+
+	redef PacketAnalyzer::VXLAN::vxlan_ports += { 4711/udp };
+
 Changed Functionality
 ---------------------
 


scripts/base/packet-protocols/gtpv1/main.zeek
```diff
@@ -15,10 +15,11 @@ module PacketAnalyzer::GTPV1;
 export {
         ## Default analyzer
         const default_analyzer: PacketAnalyzer::Tag = PacketAnalyzer::ANALYZER_IP &redef;
+
+	## The set of UDP ports used for GTPV1 tunnels.
```
0xxon, line 19: Nit - indentation is off.

scripts/base/packet-protocols/teredo/main.zeek
```diff
@@ -15,10 +15,11 @@ module PacketAnalyzer::TEREDO;
 export {
         ## Default analyzer
         const default_analyzer: PacketAnalyzer::Tag = PacketAnalyzer::ANALYZER_IP &redef;
+
+	## The set of UDP ports used for Teredo tunnels.
```
0xxon, line 19: Nit - indentation for both lines is off

scripts/base/protocols/quic/main.zeek
```diff
@@ -10,6 +10,19 @@ module QUIC;
...
 	redef enum Log::ID += { LOG };
 
+	## Well-known ports for QUIC.
+	const quic_ports = {
+		443/udp, # HTTP3-over-QUIC
+	} &redef;
+
+	## Well-known ports for DNS-over-QUIC.
+	##
+	## Currently not added to likely_server_ports for historical reasons.
```
0xxon: Out of curiosity - is there a reason to not add them, or is that just a past oversight that we re replicating here?

scripts/base/packet-protocols/teredo/main.zeek
awelzel: Thanks. it's actually the original lines above using spaces here...

scripts/base/protocols/quic/main.zeek
awelzel: It results in a baseline diff on private testing and it seemed okay for this point as of now to keep them out of likely_server_ports. Let me add that part to the comment and also include how to redef likely_server_ports for a give environment.

I'm not really sure if or what the right answer here. If you think it should be part of likely_server_ports

PR #5095, opened 2025-12-12
Import IGMP analyzer from external contributor

timwoj:
This PR imports the Spicy IGMP analyzer written by François De Keersmaeker (@fdekeers) at https://github.com/zeek-plugins/igmp into the main Zeek tree. I emailed him about doing this and he approved of doing so:

> I'm delighted that you consider adding my package to the Zeek codebase !

The analyzer is a packet analyzer, since it's easier to work into coming right after IP vs doing it as a protocol analyzer. It is not considered a transport analyzer. This means it does not generate conn.log entries. It does generate a new igmp.log and a set of events when various data is seen. I've made almost zero changes outside of formatting, fixing some warnings/errors from the compiler, fixing the naming of the plugin, and fixing a few minor script issues around registering the plugin. Any issues with the Spicy code itself existed in the original version, so this is basically brand new code.

I also didn't bring over the README from the existing repository, but it's pretty informative so it might be worth importing into the analyzer's source path.

Looking at the changes from https://github.com/zeek/zeek-testing/commit/036825c748dc8fbacd5a53066e8148b09ec77d96, I wonder if it would be worth de-duping the log messages somehow. I get that all of those are unique entries technically (since the timestamps are different in the actual log) but it ends up being a lot of duplication:

```
tests.ctu-sme-11-vm-win7ad% cat igmp.log | grep "XXXXXXXX" | sort | uniq -c
1444 XXXXXXXXXX.XXXXXX	192.168.1.105	224.0.0.22	IGMP::MEMBERSHIP_REPORT_V3
1445 XXXXXXXXXX.XXXXXX	192.168.1.106	224.0.0.22	IGMP::MEMBERSHIP_REPORT_V3
1449 XXXXXXXXXX.XXXXXX	192.168.1.107	224.0.0.22	IGMP::MEMBERSHIP_REPORT_V3
```

I'm fine with this missing the 8.1 branching, but it'd be cool to have in the final release if we can get it in before then. Otherwise, it can wait for 8.2.


timwoj:
I'm not sure what about the docs build is causing it to eat the Redis enums. The Redis analyzer is built and available in my local build, but zeekygen isn't picking it up.

evantypanski:
one immediate point: Can we make sure to co-author (or author) the original contributor in the commit that brings in the analyzer?

0xxon:
I am not sure I would try to still target this for 8.1 - it seems like a better idea to add this early in 8.2 to have opportunity to iterate over things :)

timwoj:
> one immediate point: Can we make sure to co-author (or author) the original contributor in the commit that brings in the analyzer?

Yep, that I can definitely do. I didn't think about it at all.

> I am not sure I would try to still target this for 8.1 - it seems like a better idea to add this early in 8.2 to have opportunity to iterate over things :)

Agreed. It's broken right now anyways and I won't have time to fix it before the holidays.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,53 @@
...
+##! Generates the IGMP.log file.
+
+module IGMP;
+
+@load base/frameworks/logging
+
+export {
+	redef enum Log::ID += { LOG };
+
+	type IgmpLog: record {
```
0xxon: This is missing documentation

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,53 @@
...
+		timestamp: time &log;  # Timestamp
```
0xxon: I think these are not in a format that will show up in our documentation

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,53 @@
...
+	## to the logging framework.
+	global log_igmp: event(rec: IgmpLog);
+}
+
+
+########## EVENTS ##########
+
+# Triggered when the module is loaded.
+# 1. Registers the IGMP packet analyzer.
+# 2. Creates the IGMP log stream.
```
0xxon: These comments seem superfluous - same the ones in zeek init.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,53 @@
...
+event zeek_init() &priority=5
+	{
+	# Register the IGMP packet analyzer.
+	if ( ! PacketAnalyzer::try_register_packet_analyzer_by_name("IP", 0x02, "IGMP") )
+		{
+		Reporter::error("Failed to register IGMP Spicy analyzer.");
+		}
+
+	# Create the IGMP log stream.
+	Log::create_stream(IGMP::LOG, [$columns=IgmpLog, $ev=log_igmp, $path="igmp"]);
```
0xxon: This probably should have a policy hook too

scripts/base/packet-protocols/igmp/types.zeek
```diff
@@ -0,0 +1,37 @@
+# Types used by the IGMP packet analyzer plugin
+
+module IGMP;
+
+export {
+
+    # IGMP message types
```
0xxon: These probably all should have documentation that shows up in our docs.

src/packet_analysis/protocol/igmp/igmp.evt
```diff
@@ -0,0 +1,51 @@
...
+packet analyzer IGMP:
+    parse with IGMP::Message;
+
+
+########## EVENTS ##########
+
+# Generated for every processed IGMP message.
+# :param $packet: raw packet header
+# :param self.msg_type: IGMP message type
+# :param self.checksum: transmitted message checksum
```
0xxon: all of the events should probably also be added to the Zeek scripts, so that we have documentation for them

src/packet_analysis/protocol/igmp/igmp.spicy
```diff
@@ -0,0 +1,136 @@
...
+
+    # Add remaining bytes to accumulator
+    on remaining_bytes foreach {
+        self.computed_checksum += $$;
+        if (self.computed_checksum >= uint32(0x10000)) {
+            # If accumulator exceeds 0x10000, subtract 0xffff
+            self.computed_checksum -= uint32(0xffff);
+        }
+    }
+
```
0xxon: This way of calculating the checksum first, and then re-parsing everything is quite inventive :)

I wonder what @bbannier thinks about that. I don't see any obvious problems with it - I just wonder if there is another way of accomplishing this.

0xxon (commented):
I just gave this a bit of a pass and added some random comments.

In addition to this:
* I think we should extend the testcases a bit. From what I can see, the test pcap currently has a couple of IGMPV2 and 3 messages - we should probably add one with IGMPv1 too. I think we also are missing any router messages - at least I didn't see any membership queries
* As mentioned - it seems like deduplicating the current log is useful. There typically will be repeated messages - as, e.g., the IGMPv2 RFCs encourage implementations to send messages repeatedly within short time periods. 
*  I also tried considering if it would make sense to have some kind of non transaction-based logfile that tracks state. But I am not really sure it's wort the effort.

src/packet_analysis/protocol/igmp/igmp.spicy
bbannier: The goal here is to only trigger the parsing logic if the transmitted checksum passes, and the way it is done here is right for that. The use of `set_input` causes the `unit` to hold all its data in memory; this is fine for a UDP or package analyzer like here, but could be problematic with TCP and message size not constrained in the parser.

If parsing input with unmatched checksums is acceptable an alternative would be to build up or verify the checksum during normal parsing; this could be done intrusively, i.e., injected into all parsing steps, but a cleaner way would be to use a [filter](https://docs.zeek.org/projects/spicy/en/latest/programming/parsing.html#filters) which passes data through, but also computes a checksum.

Unrelated and probably not a goal for this iterastion, the code here could probably be cleaned up to not expose that much state, e.g.,

```ruby
# Generic IGMP message
public type Message = unit {
    ### FIRST STEP: verify message checksum

    var is_checksum_ok: bool = False; # Indicates if the checksum verification is OK or not

    # FIXME(bbannier): Check if we need this field at all, and couldn't instead
    # reduce `transmitted_checksum` to a known value instead of building it up
    # here and comparing.
    #
    # Add first 2 bytes to accumulator.
    computed_checksum: uint16 &convert=uint32($$);

    transmitted_checksum: uint16; # Transmitted checksum value

    # All remaining bytes until the end, grouped by 2 bytes
    : uint16[] &eod foreach {
        # Add remaining bytes to accumulator
        self.computed_checksum += $$;
        if (self.computed_checksum >= 0x10000) {
            # If accumulator exceeds 0x10000, subtract 0xffff
            self.computed_checksum -= 0xffff;
        }
    } {
        # Compare checksum values
        self.computed_checksum = ~cast<uint16>(self.computed_checksum);
        self.is_checksum_ok = self.transmitted_checksum == self.computed_checksum;

        # Backtrack to beginning of message
        self.set_input(self.input());
    }

    # TODO: add error handling if message does not contain an even number of bytes

    ### SECOND STEP: Parse message if checksum verification passed

    msg_type: uint8 &convert=Type($$);

    # Determine message type
    if (self.is_checksum_ok) {
        switch (self.msg_type) {
            Type::MEMBERSHIP_QUERY -> mq: MembershipQuery;
            Type::MEMBERSHIP_REPORT_V1 -> mrv1: MembershipReportV1;
            Type::MEMBERSHIP_REPORT_V2 -> mrv2: MessageV2(True);
            Type::LEAVE_GROUP -> lg: MessageV2(False);
            Type::MEMBERSHIP_REPORT_V3 -> mrv3: MembershipReportV3;
        };
    };
};
```

timwoj:
@rsmmr is looking into the strange documentation issue. This PR's on hold until we get that sorted out.

rsmmr:
> @rsmmr is looking into the strange documentation issue. This PR's on hold until we get that sorted out.

Ok, I figured this out. It took me a bit to track it down but I'm happy to report it's not a bug, it's a feature. :-)

Turns out some of our built-in Spicy analyzers are missing `%doc-id` and `%doc-description` tags in their EVT file. Those tags are meant specifically for built-in analyzers to get their pieces included into the documentation: `%doc-id` turns into the title and anchor for the doc section, and `%doc-description` turns into the section's lead. Ideally, analyzers that are lacking these wouldn't appear in the docs at all, but it looks like what happens in practice is that it's rather undefined if and where they do. That's why the Redis analyzer ended up being in the docs before, even though it doesn't have the tags; but then disappeared once IGMP was added (which likewise doesn't have them). 

I'm going to push a commit to this branch that adds the missing tags. @timwoj, not sure if you want to want to keep the fixes for the existing analyzers in this PR as well, otherwise feel free to split them out. I'm also fixing a couple of other issues with IGMP, see commit message.

I've been looking at `spicyz`  to see if we can make this all more deterministic (and/or warn about missing tags) but I'm a bit reluctant to risk breaking something as I don't quite recall if changes here would have an impact on external analyzers well (the built-in analyzers are a special case for `spicyz`; external analyzers are the standard case).

timwoj:
This is ready for review now.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,49 @@
+##! Implements base functionality for IGMP analysis.
+##! Registers the IGMP packet analyzer.
+##! Generates the IGMP.log file.
```
0xxon: Nit - that 3-line description feels a bit awkward.

scripts/base/packet-protocols/igmp/types.zeek
```diff
@@ -0,0 +1,45 @@
+## Types used by the IGMP packet analyzer plugin
```
0xxon: Should this be ##! ?

src/packet_analysis/protocol/igmp/igmp.spicy
```diff
@@ -0,0 +1,136 @@
...
+    ### FIRST STEP: verify message checksum
+
+    var computed_checksum: uint32 = uint32(0); # Checksum verification accumulator
+    var is_checksum_ok: bool = False; # Indicates if the checksum verification is OK or not
+    first_2_bytes: uint16 {
+        self.computed_checksum = self.first_2_bytes;
+    } # Add first 2 bytes to accumulator
+    transmitted_checksum: uint16; # Transmitted checksum value
+    remaining_bytes: uint16[] &eod; # All remaining bytes until the end, grouped by 2 bytes
+    # TODO: add error handling if message does not contain an even number of bytes
```
0xxon: Should we add a minimal check for this before merging this in? That seems relatively easy.

0xxon (commented):
After looking at this again - in general I think this all looks pretty good, besides a couple of small nits that you can ignore.

I am curious if we should ask someone in networking about `igmp.log`, specifically if the current set of information in there is useful or not.

Specifically I wonder if we should aggregate some queries (I _think_ they can happen quite often). And, on the other end, I wonder if the data contained in the v3 membership reports might be interesting to log :)

src/packet_analysis/protocol/igmp/igmp.spicy
timwoj: Wouldn't that line automatically fail to parse if it wasn't an even number of bytes? It's parsing out an array of 16-bit values.

0xxon: Yup, you are right. On the flipside - that means the todo can be removed, right?

timwoj:
> I am curious if we should ask someone in networking about igmp.log, specifically if the current set of information in there is useful or not.

I implemented rate limiting for the logs. Some other suggestions were to:

- add history tracking for a specific source/group pair like TCP connection states. Since IGMP isn't a connection and the requests can come in whenever, there's no good way to collect them together into a history to be logged.
- sessionize an IGMP group, so that you can see when a new group is added via a source joining it and when a group is destroyed after the last source leaves it. I opted to skip this for the first pass. It would probably end up in a separate log.

I also moved the bad checksum reporting from a discrete IGMP event to a weird, since that can be done directly from Spicy.

doc/scripts/base/packet-protocols/igmp/index.rst
```diff
@@ -0,0 +1,18 @@
...
+
+
+:doc:`/scripts/base/packet-protocols/igmp/types.zeek`
+
+
+:doc:`/scripts/base/packet-protocols/igmp/main.zeek`
+
+   Implements base functionality for IGMP analysis.
+   Registers the IGMP packet analyzer.
+   Generates the IGMP.log file.
```
ckreibich: I see the first line in main.zeek, but I'm a bit confused where these two lines come from (here and in the other two instances). This is generated documentation, no? It's just a bit, but this should say `igmp.log`.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+
+	## A generic enum to map the v1/v2/v3 state changes to something common.
+	type GroupAction: enum {
+		JOIN = 0,
+		LEAVE = 1,
+	};
+
+	## The record type which contains the column fields of the IGMP log.
+	type Info: record {
+		## Timestamp
```
ckreibich: It's worth being more verbose on these comments because they end up in the log schemas, so they get consumed by all manner of post-processors that infer meaning from them. Could we say what this is the timestamp of?

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+		timestamp: time &log;
+		## Source IP address
+		src:  addr &log;
+		## Destination IP address
```
ckreibich: How about
```suggestion
		## Destination group address, as per the
		## `IANA Multicast Address Registry <https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml>`_ 
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+		## Message type
+		action:  GroupAction &log;
+	};
+
+	## Event that can be handled to access the IGMP record as it is sent on
+	## to the logging framework.
+	global log_igmp: event(rec: Info);
+
+	## The number of repeats of the same action that are allowed before a log
+	## bypasses the rate limit.
```
ckreibich: "bypass" suggests the rate limit does not apply, but should describe the opposite, no? (See below, I might be confused about how the rate limiting is supposed to work.)

```suggestion
	## The number of repeats of the same action that are allowed before Zeek
	## suppresses such messages.
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+	const rate_limit_repeats = 10 &redef;
+
+	## The amount of time between the last two reports of the same action that
+	## is allowed before a log bypasses the rate limit.
```
ckreibich: I found this a bit hard to follow — why two reports? Is this just this:
```suggestion
	## The amount of time for which repeat messages remain suppressed once
	## rate-limiting applies.
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+## A record to keep track of the last state action seen for a specific src/group address
+## pair. Stored in the active_states map.
+type State: record {
+	last_seen: time;
+	repeats: count;
+	last_action: GroupAction;
+};
+
+## The currently active igmp actions. This is used to manage rate-limiting.
+global active_states: table[addr, addr] of State;
```
ckreibich: I see nothing that removes entries from that table? We should likely have a `&create_expire` or some such.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+		else if ( state$repeats > rate_limit_repeats ) {
+			state$repeats = 0;
+			do_log = T;
+		}
+		else if ( network_time() - state$last_seen > rate_limit_duration ) {
+			do_log = T;
+		}
+
+		state$last_seen = network_time();
+		active_states[src, group] = state;
```
ckreibich: You don't need to re-insert since you're already working on a state instance retrieved from the table?

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+		local state: State = active_states[src, group];
+
+		state$repeats += 1;
+
+		if ( state$last_action != action ) {
+			do_log = T;
+			state$last_action = action;
+			state$repeats = 0;
+		}
+		else if ( state$repeats > rate_limit_repeats ) {
```
ckreibich: Could you explain somewhere how the rate-limiting is supposed to work? I assumed the idea is that at first we _do_ log repeat messages, up until the threshold of repeats, and then don't, for the configured interval? This seems to implement that we log nothing until it's happened too often.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,152 @@
...
+	if ( should_log(packet$ip$src, group_addr, LEAVE) ) {
+		Log::write(IGMP::LOG, Info(
+			$timestamp = network_time(),
+			$src = packet$ip$src,
+			$group = group_addr,
+			$action = LEAVE));
+	}
+}
+
+event IGMP::membership_report_v3(packet: raw_pkt_hdr, groups: vector of IGMP::Group) {
```
ckreibich: Just a nit: would be nice if that's grouped with the other membership reports.

src/packet_analysis/protocol/igmp/igmp.spicy
```diff
@@ -0,0 +1,132 @@
...
+    MODE_IS_EXCLUDE = 2,
+    CHANGE_TO_INCLUDE_MODE = 3,
+    CHANGE_TO_EXCLUDE_MODE = 4,
+    ALLOW_NEW_SOURCES = 5,
+    BLOCK_OLD_SOURCES = 6,
+};
+
+# Version 3 Membership Report Group Record
+type Group = unit {
+    group_type: uint8 &convert=GroupType($$);
```
ckreibich, line 35: What happens if this fails?

src/packet_analysis/protocol/igmp/igmp.spicy
```diff
@@ -0,0 +1,132 @@
...
+# Generic IGMP message
+public type Message = unit {
+
+    # Iterator, to parse message after checksum has been verified
+    var start: iterator<stream>;
+    on %init {
+        self.start = self.input();
+    }
+
+    ### FIRST STEP: verify message checksum
```
ckreibich: In the ICMP analyzer, we [bypass checksumming](https://github.com/zeek/zeek/blob/master/src/packet_analysis/protocol/icmp/ICMP.cc#L70) if `ignore_checksums` is true. Would be nice if that held here too ... is there a way to do that?

scripts/base/packet-protocols/igmp/spicy-events.zeek
```diff
@@ -0,0 +1,45 @@
+##! Events generated by the IGMP analyzer.
+##!
+##! See :rfc:`3376`.
+
+## Generated for every processed IGMP message.
+##
+## packet: Raw packet header.
+##
+## msg_type: The type of message received.
+global IGMP::message: event(packet: raw_pkt_hdr, msg_type: IGMP::MessageType);
```
ckreibich, line 10: I was curious here if we really need `raw_pkt_hdr`, which is a bit unusual for our analyzers. Could we do something like we have in `icmp_info` or `icmp_context`, or just use the addresses?

ckreibich (commented):
I finally looked at this — very sorry for the delay Tim. I just have a handful of questions right now. I haven't looked closely at the tests yet.

doc/scripts/base/packet-protocols/igmp/index.rst
timwoj: The docs just need to be regenerated.

scripts/base/packet-protocols/igmp/main.zeek
timwoj: Should that just be `rate_limit_duration`?

scripts/base/packet-protocols/igmp/main.zeek
timwoj: If the table has a `create_expire` duration, doesn't reinserting reset that?

src/packet_analysis/protocol/igmp/igmp.spicy
timwoj: According to https://docs.zeek.org/projects/spicy/en/latest/programming/language/types.html#enum, it would be converted to something like `unknown-#`. I don't know what Zeek would do with it when it gets passed through an event.

src/packet_analysis/protocol/igmp/igmp.spicy
timwoj: Sure, the spicy runtime library has `zeek::get_value` for looking up script-land values.

scripts/base/packet-protocols/igmp/spicy-events.zeek
timwoj: That was in the original plugin from upstream, so I just left it in. I can see how it's not terribly useful to have all of that information though. The only field we're using is the source IP address, so I'll just pass that alone.

src/packet_analysis/protocol/igmp/igmp.spicy
bbannier: If a Spicy `enum` value gets exported to Zeek (e.g., by passing a Spicy `enum` value in a Zeek `event` directly or as part of an exported `unit`) we [create a matching Zeek `enum`](https://github.com/zeek/zeek/blob/cb2cea7a7e253f651d6c61010a68d4d1071bef8e/src/spicy/runtime-support.cc#L123-L144), e.g., for

```ruby
module Ana;

public type E = enum {
    A,
    B,
};
```

we would generate a Zeek `enum` `Ana::E` with namespaced names `Ana::E_A`, `Ana::E_B`, and `Ana::E_Undef`.

0xxon: That's a good catch. If, in principle, it's possible that implementations stick other values in there, it seems like just passing the uint through, and converting it on the Zeek side is nicer.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,153 @@
...
+	last_action: GroupAction;
+};
+
+## The currently active igmp actions. This is used to manage rate-limiting.
+global active_states: table[addr, addr] of State &create_expire=rate_limit_duration;
+
+function should_log(src: addr, group: addr, action: GroupAction) : bool {
+	local do_log: bool = F;
+
+	if ( [ src, group] in active_states ) {
```
0xxon: This is a total nit - there's a space missing after group (or one too much before src) for consistency.

scripts/base/packet-protocols/igmp/main.zeek
0xxon: So - this actually raises a great point. I am not an expert on IGMP, but I think it's common to see repeated messages. I'm not sure there's much value in logging these by default - so I could completely be convinced to just by-default not log them. But for that we should probably change the explanation, and (perhaps) offer an opt-out?

0xxon:
I like the log much more than the previous one - thank you very much. Also - thank you for flagging it as experimental - that gives us the leeway to change it, if people report issues using this.

I personally think this looks pretty good now - the main remaining question seems to be rhe rate limiting.

timwoj:
I ran into a fun issue with some traffic in the m57-long test in zeek-testing. The IGMP packets there have trailing data in the Ethernet frames (6 bytes of padding plus 4 bytes of FCS). The packet analyzers flow this data all the way down through to the IGMP analyzer, which then tries to use that data when calculating the checksums. This causes extra weirds that shouldn't happen since that data isn't part of the actual IGMP packet.

Assuming Spicy can tell me the byte position after parsing, I wonder if it would be better to let it parse the data and then verify the checksum once it knows how long the IGMP part is.

src/packet_analysis/protocol/igmp/igmp.spicy
timwoj: I don't see anything in the RFC that says implementations may add other types. It does say that unknown types should be silently ignored, so perhaps that should result in a weird on the Spicy side?

scripts/base/packet-protocols/igmp/main.zeek
timwoj: The way I understand how we do weird rate limiting is that we log every X'th duplicate weird or log a duplicate if some amount of time has passed since we last logged one. The code here in the analyzer does the following:

> If the src/group pair is in the existing states, increase the number of repeats and then:
> 
> 1. If the action is different, reset the action in the state and repeats and allow logging
> 2. if the number of repeats is over the limited number, reset the counter to zero and allow logging
> 3. If the amount of time since the last seen is over the duration, reset the counter to zero and allow logging
> 
> Update the entry in the table every time to force the create_expire timer to reset and to reset the network_time value.
> 
> If the src/group pair is not in the existing states:
> 
> Allow logging and add the new entry to the existing states.

timwoj: Now that the table has `&create_expire` it might be possible to remove (3) above and let that part be handled through the table expiration.

0xxon:
Ooooh, interesting.

That's a really interesting issue to come up. FWIW - for TCP and UDP, the TCP and UDP analyzers any extra data after the packet  before passing it on to the downstream analyzers. It kind of feels like the same should happen for Ethernet before the data is passed into the packet analyzer. So this might be a Zeek issue more than an issue in the analyzer.

timwoj:
The original plugin has an event just for the message itself (`igmp_message`). I left it in at the time, but does anyone see any reason to keep it? We send the relevant data as part of the typed events, so it's really just sending an extra event for every message?

scripts/base/packet-protocols/igmp/main.zeek
ckreibich: Yep! To my credit, the table didn't have `&create_expire` at the time. ;-)

scripts/base/packet-protocols/igmp/main.zeek
ckreibich: Yeah, it looks good to me now. Thanks!

scripts/base/packet-protocols/igmp/main.zeek
ckreibich: Yeah, agree Tim. And thanks for the explanation, I get it now. (I completely missed the else-part when I wrote the question.)

ckreibich:
Sorry to sidetrack this a bit, but I poked through the IGMP RFCs in more detail and am a bit worried.

It looks to me that while the concepts behind IGMP are pretty basic, the reality on the wire is not. A node may send a membership report in response to a router query or when its internal state changes. That state change can be many things, not just join/leave. The membership reports, and in particular the rules around when nodes send these, are [really complex](https://datatracker.ietf.org/doc/html/rfc3376#autoid-37), at least in v3: lots of logic around delays, repeats, what-was-just-sent, etc.

I bring this up because it has me wondering whether we're getting ahead of ourselves with the rate-limiting. Our log is trying to track endpoints joining and leaving multicast groups. My main question then is this: do we know that we can reliably identify whether a node is joining and leaving a group? I'm nervous for two reasons:
* [This code](https://github.com/zeek/zeek/blob/topic/timw/import-igmp-analyzer/scripts/base/packet-protocols/igmp/main.zeek#L124-L127) makes that determination for IGMPv3 reports based on the group type alone, but as far as I can tell that just indicates the source-filtering mode is changing(?).
* The log generation does not distinguish why a node sends a report — via internal logic in a group member, or based on prompting by a router. 

If we can make this join/leave determination reliably, then I am actually not too worried about the log rate, because I don't think nodes really join and leave groups much. They might be _queried_ to report their memberships much more often than they change them, though.

If we can't make that determination reliably, then I think we'd first need to figure out how to track group memberships in Zeek from any reports coming through, as well as leave-messages. I.e., the act of a node joining a group might not be an explicit message on the wire, but we would determine it based on a _change_ in a node's reported groups — or because we hear from a node for the first time. In a nutshell, we'd implement the same logic running in a multicast router. We're not doing that right now.

So my suggestion for next steps is that we first make sure that we understand the state machine. I'm really no IGMP expert — so I could be totally off in the above and it's all easy.

If it's reliable, I am actually tempted to not do any rate-limiting at all in a first version — I'd hope we can still get enough feedback from the testing group before 8.2 to figure that out. (Fwiw, I'm not seeing a lot of IGMP traffic in any of the sites I checked, but that's obviously not definitive.)

ckreibich:
I've read through the RFCs. The messages indeed report current state more than action, but it looks like we can get what we need for the intended log. I propose the following:
* Let's change the `JOIN` / `LEAVE` action terminology to a state,  perhaps `JOINED` / `LEAVING`, because we will have a hard time figuring out when exactly a node is joining. (Leaving is somewhat easier.)
* For v1/v2 the semantics are straightforward: the membership reports tell us the groups a node is in, and the leave-group message indicates departure.
* For v3, the group record types basically all indicate active memberships, so we can collect them from membership reports. There's one exception: `CHANGE_TO_INCLUDE_MODE` with an empty source list is effectively a leave-group. This is buried in [7.3.2](https://www.ietf.org/rfc/rfc9776.html#name-in-the-presence-of-older-versi) of the IGMP v3 RFC.
* The rate-limiting semantics then become "only one `JOINED` or `LEAVING` message per source, group, and configured time interval".
* There are some theoretical corner cases: what if a node says that a specific list of sources is now no longer allowed, and only the router knows that the result is now the empty set? When a node we have on file as a joined member does not respond to a query — does that mean it has left? I think we can ignore these.
* The default router query interval is 125 seconds, so I think the choice of 5 minutes is good.

ckreibich:
Two ideas for things to do later:
* We could codify the multicast group address definitions from https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml to give the group addresses some meaning, in an additional column. Perhaps in a policy script.
* We may want ignore-lists for groups. If somebody finds that all of the machines on a network only ever talk about their mDNS membership, perhaps it's time to ignore 224.0.0.251, etc.

timwoj:
I know that `IS_EXCLUDE` with an empty source list is equivalent to an IGMPv2 join, but I wonder if `CHANGE_TO_EXCLUDE` with an empty source list is the same. You're swapping from some other state to `IS_EXCLUDE` and the empty source list should indicate a join. I don't see anything in the docs that indicates that's correct though. Maybe because you would have already been joined when switching states? Claude thinks they're semantically equivalent, fwiw.

timwoj:
> The rate-limiting semantics then become "only one JOINED or LEAVING message per source, group, and configured time interval".

Isn't that what the semantics are already? It's keyed on action, source, and group, and it's limited by the number of times it's seen that within a specific window.

doc/scripts/base/packet-protocols/igmp/main.zeek.rst
```diff
@@ -0,0 +1,137 @@
...
+   suppresses such messages.
+
+Types
+#####
+.. zeek:type:: IGMP::GroupAction
+   :source-code: base/packet-protocols/igmp/main.zeek 14 18
+
+   :Type: :zeek:type:`enum`
+
+      .. zeek:enum:: IGMP::JOIN IGMP::GroupAction
```
ckreibich: In `main.zeek` below, these are `JOINED` and `LEAVING`. So this probably needs a docs refresh?

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,154 @@
...
+@load base/frameworks/logging
+
+export {
+	redef enum Log::ID += { LOG };
+
+	## A default logging policy hook for the stream.
+	global log_policy: Log::PolicyHook;
+
+	## A generic enum to map the v1/v2/v3 state changes to something common.
+	type GroupAction: enum {
```
ckreibich: This is nitpicky Tim, sorry — but since we're technically not seeing the act of joining or leaving, only the state that the node is in when it transmits this, `GroupState` might be better. I looked at the RFCs again and their use of "action" refers to routers reacting to the updates.

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,154 @@
...
+	## The record type which contains the column fields of the IGMP log.
+	type Info: record {
+		## The network time when the message was received.
+		ts: time &log;
+		## Source IP address of the message.
+		src:  addr &log;
+		## Destination group address, as per the
+		## `IANA Multicast Address Registry <https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml>`_
+		group:  addr &log;
+		## IGMP action requested in the message.
```
ckreibich:
```suggestion
		## The membership status this multicast receiver reports for the group.
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,154 @@
...
+	type GroupAction: enum {
+		JOINED = 0,
+		LEAVING = 1,
+	};
+
+	## The record type which contains the column fields of the IGMP log.
+	type Info: record {
+		## The network time when the message was received.
+		ts: time &log;
+		## Source IP address of the message.
```
ckreibich: We need to say somewhere around here that this log captures state updates from multicast group members, as opposed to logging all IGMP activity (which "message" implies). If we do this somewhere here in the record, it makes that information available in log schemas, in contrast to explanation in NEWS. So perhaps:

```suggestion
		## IP address of the multicast receiver sending this update.
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,154 @@
...
+
+## A record to keep track of the last state action seen for a specific src/group address
+## pair. Stored in the active_states map.
+type State: record {
+	last_seen: time;
+	repeats: count;
+	last_action: GroupAction;
+};
+
+## The currently active igmp actions. This is used to manage rate-limiting.
```
ckreibich:
```suggestion
## The currently active multicast receivers and their group statuses. This is used to manage rate-limiting.
```

scripts/base/packet-protocols/igmp/main.zeek
```diff
@@ -0,0 +1,154 @@
...
+		## IGMP action requested in the message.
+		action:  GroupAction &log;
+	};
+
+	## Event that can be handled to access the IGMP record as it is sent on
+	## to the logging framework.
+	global log_igmp: event(rec: Info);
+
+	## The number of repeats of the same action that are allowed before Zeek
+	## suppresses such messages.
```
ckreibich: I think this comment was the source of my confusion around how the rate-limiting works. This says messages are allowed before Zeek suppresses, but the algorithm below suppresses messages (after the first one) until we've seen _more_ than this many (and then logs one).

NEWS
```diff
@@ -86,12 +75,19 @@ New Functionality
...
+- The Zeek documentation has received a major refresh. A new tutorial walks newcomers
+  through setup, basic Zeek invocations, the use of Zeek packages, ZeekControl for cluster
+  maintenance, the basics of Zeek's logs, as well as scripting in Zeek's own language as
+  well as JavaScript. A new reference section houses the introduction to common logs, the
+  Zeek scripting language, Zeek's frameworks, and the scripting index.
+
+- A new analyzer for IGMP was imported from an external project, originally written by
+  François De Keersmaeker (@fdekeers). It adds a new ``igmp.log`` which currently logs
+  state changes for individual hosts joining and leaving multicast groups. These state
+  changes are rate-limited, similar to how weirds are. The rate-limiting is controlled by
```
```suggestion
  changes are rate-limited, similarly to weirds. The rate-limiting is controlled by
```

ckreibich (commented):
Almost there Tim — I only have nits regarding wording in the code.

I still worry about the rate-limiting, though. You asked about the semantics and I mean something slightly different from what's currently there. An example is probably best: I grabbed a pcap on my home network. IGMP looks something like this:
```
1775183467.832394 IP 192.168.0.1 > 224.0.0.1: igmp query v3
1775183472.208723 IP 192.168.0.109 > 224.0.0.22: igmp v3 report, 2 group record(s)
1775183473.388293 IP 192.168.0.26 > 224.0.0.22: igmp v3 report, 1 group record(s)
1775183592.934560 IP 192.168.0.1 > 224.0.0.1: igmp query v3
1775183598.390263 IP 192.168.0.109 > 224.0.0.22: igmp v3 report, 2 group record(s)
1775183602.028291 IP 192.168.0.26 > 224.0.0.22: igmp v3 report, 1 group record(s)
1775183718.044684 IP 192.168.0.1 > 224.0.0.1: igmp query v3
1775183723.068293 IP 192.168.0.26 > 224.0.0.22: igmp v3 report, 1 group record(s)
1775183723.853915 IP 192.168.0.109 > 224.0.0.22: igmp v3 report, 2 group record(s)
...
```
So basically a router asking every couple of minutes, and two machines reporting memberships. With the defaults, this turns into the following log over the course of an hour:
```
1775183472.208723       192.168.0.109   239.255.255.250 IGMP::JOINED
1775183472.208723       192.168.0.109   224.0.0.251     IGMP::JOINED
1775183473.388293       192.168.0.26    224.0.0.251     IGMP::JOINED
1775183850.348297       192.168.0.26    224.0.0.251     IGMP::JOINED
1775183852.852470       192.168.0.109   239.255.255.250 IGMP::JOINED
1775183852.852470       192.168.0.109   224.0.0.251     IGMP::JOINED
1775184223.565097       192.168.0.109   239.255.255.250 IGMP::JOINED
1775184223.565097       192.168.0.109   224.0.0.251     IGMP::JOINED
1775184227.948294       192.168.0.26    224.0.0.251     IGMP::JOINED
1775184596.529029       192.168.0.109   239.255.255.250 IGMP::JOINED
1775184596.529029       192.168.0.109   224.0.0.251     IGMP::JOINED
1775184599.788288       192.168.0.26    224.0.0.251     IGMP::JOINED
1775184972.668292       192.168.0.26    224.0.0.251     IGMP::JOINED
1775184975.178626       192.168.0.109   239.255.255.250 IGMP::JOINED
1775184975.178626       192.168.0.109   224.0.0.251     IGMP::JOINED
1775185345.328746       192.168.0.109   239.255.255.250 IGMP::JOINED
1775185345.328746       192.168.0.109   224.0.0.251     IGMP::JOINED
1775185349.068280       192.168.0.26    224.0.0.251     IGMP::JOINED
1775185719.869343       192.168.0.109   239.255.255.250 IGMP::JOINED
1775185719.869343       192.168.0.109   224.0.0.251     IGMP::JOINED
1775185726.828289       192.168.0.26    224.0.0.251     IGMP::JOINED
1775186096.748282       192.168.0.26    224.0.0.251     IGMP::JOINED
1775186102.829926       192.168.0.109   239.255.255.250 IGMP::JOINED
1775186102.829926       192.168.0.109   224.0.0.251     IGMP::JOINED
1775186471.234439       192.168.0.109   239.255.255.250 IGMP::JOINED
1775186471.234439       192.168.0.109   224.0.0.251     IGMP::JOINED
1775186475.388258       192.168.0.26    224.0.0.251     IGMP::JOINED
```
There's actually little going on here — the two nodes remain members of their groups (UPnP and mDNS) throughout. So this log is too verbose. When I run it with `IGMP::rate_limit_duration=60min` `IGMP::rate_limit_repeats=10000` it cuts out the repetitions:
```
1775183472.208723       192.168.0.109   239.255.255.250 IGMP::JOINED
1775183472.208723       192.168.0.109   224.0.0.251     IGMP::JOINED
1775183473.388293       192.168.0.26    224.0.0.251     IGMP::JOINED
```
So I'm thinking that we should up `rate_limit_duration` significantly, to something like 30 minutes or even an hour. And I wonder whether we should simply drop `rate_limit_repeats`, since in the above I used a large number to mean "never report the same thing repeatedly in the duration window". If you want more frequent reporting, you can simply reduce the interval, no?

timwoj:
> And I wonder whether we should simply drop rate_limit_repeats, since in the above I used a large number to mean "never report the same thing repeatedly in the duration window". If you want more frequent reporting, you can simply reduce the interval, no?

Couldn't the same argument be made for weirds though? I'm inclined to leave it as-is just because it's following an established pattern.

ckreibich (approved):
Okay thanks Tim! Let's give this a go. 🚀

scripts/base/packet-protocols/__load__.zeek
```diff
@@ -21,6 +21,7 @@
 @load base/packet-protocols/udp
 @load base/packet-protocols/tcp
 @load base/packet-protocols/icmp
+@load base/packet-protocols/igmp
```
klemensn, line 24: This code requires Spicy, but Zeek may be built without it, as is done on OpenBSD:
```
-- Warning: Disabling analyzer for IGMP because Spicy is not available
```

Yet this analyzer is loaded unconditionally and breaks thusly starting with 8.2.0 release:
```
# rcctl start zeek    
zeek(failed)

# zeekctl diag
...
warning in /usr/local/share/zeek/base/packet-protocols/igmp/spicy-events.zeek, line 38: handler for non-existing event cannot be invoked (IGMP::leave_group)
...

```

Other analyzers are equally disabled due to lack of Spicy, but they cause no problems.

One workaround is to `@unload` IGMP in `local.zeek`.

I'll commit a local patch to OpenBSD's `net/bro` now removing this `@load` line.

PR #5097, opened 2025-12-14
Restrict more templates using concepts

Neverlord:


awelzel:
> Restrict more templates using concepts #5097

I have stance if this is useful, but if that was directed / found by clang-tidy or some other static tooling, updating `.clang-tidy` would be nice so that the average C++ user (me...) is told they should pick a more modern way.

Neverlord:
> > Restrict more templates using concepts #5097
> 
> I have stance if this is useful, but if that was directed / found by clang-tidy or some other static tooling, updating `.clang-tidy` would be nice so that the average C++ user (me...) is told they should pick a more modern way.

I'm not aware of any `clang-tidy` or `cppcheck` checks that would capture something like this. There is [`modernize-use-constraints`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-constraints.html), but that's just for using concepts instead of SFINAE.

I think a check for unconstrained templates that could be constrained using a concept from the STL would be really nice to have, but it's not as simple as "use feature X instead of feature Y" (or just "don't do X") as most of the existing `clang-tidy` checks. It would basically have to look at each unconstrained template and then try some heuristic to figure out if there's a concept from the standard library that might match based on how the template parameter is used.

timwoj (approved):

PR #5098, opened 2025-12-15
Added offset to find first and last BIFs

### LINKED ISSUE #4424: Add start offset for `find_first`/`find_last` BIFs
From @vpax in https://github.com/zeek/zeek/pull/4411#pullrequestreview-2808518835:

> I think it would be nice if there's an optional third parameter giving the offset where to start looking (and then add the same to find_last)

Ruddxxy:
Fixes #4424 
Added `offset: count &default=0` to the BIF signatures. This ensures full backward compatibility with existing Zeek scripts and uses pointer arithmetic which adjusts length parameters passed to the regex engine. This avoids `substring` allocation, ensuring zero-copy overhead for the search operation. Added explicit bounds checking. 

Testing
Added new BTest suite: `testing/btest/bifs/find_first_offset.zeek`. to ensure proof of logic. Verified that original `find_first` tests pass to ensure no regression.
 Please have a look.

### DIFF
--- a/src/strings.bif
+++ b/src/strings.bif
@@ -1102,7 +1102,7 @@ function find_last%(str: string, re: pattern, offset: count &default=0%) : strin
 	const u_char* s = str->Bytes();
 	const u_char* e = s + str->Len();
 
-	if ( offset >= str->Len() )
+	if ( offset >= (zeek_uint_t)str->Len() )
 			return zeek::val_mgr->EmptyString();
 			
 	//Scans backwards from end of string, but stops at the offset.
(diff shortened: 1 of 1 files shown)


testing/btest/Baseline/bifs.find_first_offset/output
```diff
@@ -0,0 +1,15 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
timwoj, line 1: Could you add some tests that search starting at an offset but fail to find something?

timwoj (changes requested):
This looks overall fine to me. I had one suggestion about the testing.

testing/btest/Baseline/bifs.find_first_offset/output
Ruddxxy: Sure, I will get it done.

evantypanski: Hey @Ruddxxy, any update on this?

we can add that too, just let us know. I have a couple nits about the C-style casts that should get into this but overall looks okay to me.

Ruddxxy: Yes please.

evantypanski:
I opened #5191 to add the review feedback directly in the branch so that it's easier. The commits are still authored by you though :) thanks for the contribution!

PR #5099, opened 2025-12-15
init-bare: Document stale connection and endpoint field behavior

awelzel:
As Jan suggested initially on #4214, lets at least document the surprising behavior.

References #4214

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -848,6 +848,41 @@ type event_metadata_vec: vector of EventMetadata::Entry;
 
 ## Statistics about a :zeek:type:`connection` endpoint.
 ##
+## .. note::
+##
+##     As of version 8.1, Zeek does not reliably keep the state, num_pkts and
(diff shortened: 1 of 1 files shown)


scripts/base/init-bare.zeek
```diff
@@ -848,6 +848,41 @@ type event_metadata_vec: vector of EventMetadata::Entry;
...
+##     these fields may be stale and not reflect the most recent values as stored
+##     within Zeek's core after processing a packet.
+##
+##     Most notably, the packet and byte counts in a ``new_connection()`` handler
+##     are zero, rather than non-zero for the originator endpoint of a connection.
+##     The same applies to protocol analyzer events: The packet counts may not
+##     include the packet that raised the event.
+##
+##     If you use packet or byte counts in event handlers, the workaround to
+##     get the most recent values on the endpoint records is to execute
```
timwoj, line 865:
```suggestion
##     get the most recent values on the endpoint records is to execute the
```

Issue #5103, opened 2025-12-15
global_container_footprints() for table[pattern] of X slower and slower


awelzel:
Reported by @pbcullen .

Calling `global_container_footprints()` in regular intervals with a `global table[pattern] of count` with 10000 entries gets significantly slower for every invocation. Haven't looked at the details. This makes `global_container_footprints()` unusable when `table[pattern] of X` are in play.

```
$ cat slow.zeek 
redef exit_only_after_terminate = T;

global tbl: table[pattern] of count;
function set_footprints()
        {
        local footprints = global_container_footprints();
        print |footprints|, "footprints";
        }

event tick()
        {
        local t0 = current_time();
        set_footprints();
        local t1 = current_time();
        print fmt("took %.3fs", t1 - t0);
        schedule 0.01sec { tick() };
        }

event zeek_init() &priority=-100000
        {
        # Populate table with patterns.
        local i = 10000;
        while ( i > 0 )
                {
                tbl[string_to_pattern("/a" + cat(i) + "/", T)] = i;
                --i;
                }
        event tick();
        }
```

```
$ zeek slow.zeek 
308, footprints
took 0.244s
308, footprints
took 0.335s
308, footprints
took 0.409s
308, footprints
took 0.630s
308, footprints
took 0.642s
308, footprints
took 0.719s
308, footprints
took 0.812s
308, footprints
took 1.622s
308, footprints
took 1.840s
308, footprints
took 2.143s
308, footprints
took 2.648s
308, footprints
took 2.757s
308, footprints
took 3.087s
308, footprints
took 3.408s
```

vpax:
Interesting! It's not as dramatic on my laptop, but definitely increasing. I'll take a quick look given this is a BiF I introduced.

vpax:
Hmmm it has nothing to do with `global_container_footprints()` (_phew_). You can replace the body of `set_footprints()` with:
```
        local sum = 0;
        for ( t in tbl )   
                ++sum;   
```
and get the same effect. Note, if instead the loop is `for ( _, t in tbl )` then it zips along, so the issue is recovering the patterns.

vpax:
In particular, best guess is the issue is [this line](https://github.com/zeek/zeek/blob/0a880a7416ee22b01c27bb9c0fcaf36c1416a06c/src/CompHash.cc#L257) in `CompositeHash::RecoverOneVal`.

vpax:
[...]

PR #5104, opened 2025-12-15
Fix for NFA construction state growth

vpax:
NFA construction uses a static data structure for performance reasons. Ironically, when constructing large numbers of patterns, this structure continues to grow in size due to monotonically increasing NFA state identifiers, and becomes increasingly slow. This PR resets the identifiers to start afresh for each new NFA (which works fine, as they only need to be distinct inside their own matcher).

I also took the opportunity to update the baseline for a ZAM-related BTest for tracking BiFs.

Closes https://github.com/zeek/zeek/issues/5103.


ckreibich (approved):
Thanks for the quick turnaround, LGTM. It'd be helpful to put the explanation into the git comment in such cases, since it'll help locate reasoning faster and the NFA/DFA code is tricky!

PR #5108, opened 2025-12-17
Bump zeekjs to v0.22.0

### LINKED ISSUE #5106: "internal error: bad reference count" when using ZeekJS
I stumbled across a strange error when using ZeekJS, tried to recreate it with a reduced script and ended up with the following:

```zeek
module Test;

global h: hook(s: string);

function f(s: string) {
    hook h(s);
}

@load test.js

event Test::e() {
    schedule 1usec { Test::e() };
    f(cat("test"));
}

event zeek_init() {
    event Test::e();
}
```

`test.js`:

```js
zeek.hook('Test::h', (s) => {
    let t = s + "test";
});
```

Sooner or later Zeek will crash with the error message `internal error: bad reference count [2]` (though I've also seen `[0]`). When calling `f("test");` instead of `f(cat("test"));` Zeek does not crash but I don't understand why this would make a difference. When originally noticing this crash I wasn't using `cat()` at all but had a hook handler in JS that would write entries to a global Zeek table.

awelzel:
7706df3 Executor: Drop args completely
    0d5c87d Nodejs: Fix args passed as copy for hook handler
    98e65df tests: Add test from zeek/zeek#5106 to stress hook invocations
    f329b94 Executor: Do not run task under lock

Closes #5106


timwoj:
Should I pull this into 8.1?

awelzel:
> Should I pull this into 8.1?

Yeah, that'd be great. I put it in for 8.0.5 as well.

PR #5110, opened 2025-12-18
Fix VLAN tags of 0

### LINKED ISSUE #5105: Zeek does not handle VLAN ID 0 correctly
Zeek doesn't handle the difference between absence of VLAN tags and presence with VLAN ID value 0 correctly. The latter is a real use case in 802.1Q, for priority-tagged frames.

I'm attaching two pcaps, one with a single tag and one with Q-in-Q tags, all with ID 0:

```
$ tshark -r http_vlan.pcap -T fields -e vlan.id | uniq
0
```
```
$ tshark -r http_qinq.pcap -T fields -e vlan.id | uniq
0,0
```

The behavior is wrong for both `vlan-logging.zeek` (last two columns are `-`) ...
```
$ zeek -r http_vlan.pcap protocols/conn/vlan-logging
$ cat conn.log
...
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       proto   service duration        orig_bytes       resp_bytes      conn_state      local_orig      local_resp      missed_bytes    history orig_pkts       orig_ip_bytes    resp_pkts       resp_ip_bytes   tunnel_parents  ip_proto        vlan    inner_vlan
#types  time    string  addr    port    addr    port    enum    string  interval        count   count   string  bool    bool     count   string  count   count   count   count   set[string]     count   int     int
1362692526.869344       CKEOCOLbpm3xgKdGe       141.142.228.5   59856   192.150.187.43  80      tcp     http    0.211484136      5007    SF      F       F       0       ShADadFf        7       512     7       5379    -       6       -       -
```
```
$ zeek -r http_qinq.pcap protocols/conn/vlan-logging
$ cat conn.log
...
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       proto   service duration        orig_bytes      resp_bytes      conn_state      local_orig      local_resp      missed_bytes    history orig_pkts       orig_ip_bytes   resp_pkts       resp_ip_bytes      tunnel_parents  ip_proto        vlan    inner_vlan
#types  time    string  addr    port    addr    port    enum    string  interval        count   count   string  bool    bool    count   string  count   count   count   count   set[string]     count   int     int
1362692526.869344       CCXUVH2rJFcYFu8hY9      141.142.228.5   59856   192.150.187.43  80      tcp     http    0.211484        136     5007    SF      F       F       0       ShADadFf        7       512     7       5379    -       6       -       -
```
... and `frameworks/conn_key/vlan_fivetuple` (`id.ctx` isn't populated):
```
$ zeek -r http_vlan.pcap frameworks/conn_key/vlan_fivetuple
$ cat conn.log
...
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       id.ctx.vlan     id.ctx.inner_vlan       proto   service duration        orig_bytes      resp_bytes      conn_state      local_orig      local_resp      missed_bytes    history orig_pkts       orig_ip_bytes      resp_pkts       resp_ip_bytes   tunnel_parents  ip_proto
#types  time    string  addr    port    addr    port    int     int     enum    string  interval        count   count   string  bool    bool    count   string  count   count   count   count   set[string]     count
1362692526.869344       CiVy884inAe1MBpsW       141.142.228.5   59856   192.150.187.43  80      -       -       tcp     http    0.211484        136     5007    SF      F       F       0       ShADadFf        7       512     7       5379    -       6
```
```
$ zeek -r http_qinq.pcap frameworks/conn_key/vlan_fivetuple
$ cat conn.log
...
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       id.ctx.vlan     id.ctx.inner_vlan       proto   service duration        orig_bytes      resp_bytes      conn_state      local_orig      local_resp      missed_bytes    history orig_pkts       orig_ip_bytes      resp_pkts       resp_ip_bytes   tunnel_parents  ip_proto
#types  time    string  addr    port    addr    port    int     int     enum    string  interval        count   count   string  bool    bool    count   string  count   count   count   count   set[string]     count
1362692526.869344       CRVbRx1NMmGgdoNXV       141.142.228.5   59856   192.150.187.43  80      -       -       tcp     http    0.211484        136     5007    SF      F       F       0       ShADadFf        7       512     7       5379    -       6
```

That's because we use `if ( vlan )` (thus skipping when vlan == 0) or similars in the code — at least in the `Packet`, `Connection`, and `IPVlanConnKey` classes.

The main question is likely how to adapt the current `uint32_t` variables to indicate absence/presence, since this will affect non-`detail` APIs.

[vlan-0.zip](https://github.com/user-attachments/files/24179664/vlan-0.zip)

evantypanski:
Fixes #5105 

This started as experiments but seemed to work, so I took it to its natural conclusion.

The first commit implements it completely, but without optionals. The next 2 add `std::optional` in different capacities, happy to drop the `Packet` one if it seems too much. They were suggested by @awelzel [here](https://github.com/zeek/zeek/commit/ccdfe31d6ff648ab43e3de17993b92eb0ac05bf2#r173099343) and wanted to see how it felt

~We could also use a `VlanInfo` struct that includes DEI/PCP. But it does make the most common case (just grabbing vlan tags) a little more annoying. I'm pretty firmly in the "I don't care" camp beyond this. Though, I kind of like the `std::optional` accessors in `Packet`!~

I added another commit that uses a `VlanTag` to store the ID, PCP, and DEI. If we decide it's good I'll squash all 4 of those, they're just separate in case one part isn't agreeable.

The last one uses this opportunity to change the types to what you'd expect. Again, can drop if that's too far...

There's one caveat. If someone manually sets `vlan` when it wasn't set before, they're going to have weird/wrong behavior. This will get fixed by using the deprecated methods, but since they never set the `vlan_present` flag(s), Zeek's internals won't catch it. We can fix this by checking like `vlan_present || vlan` or something, but then we won't fix this bug until the field-accesses are deprecated. I like it how it is though.


src/iosource/Packet.h
```diff
@@ -148,48 +160,142 @@ class Packet {
...
-     * If layer 2 is Ethernet, innermost ethertype field.
+     * (Outermost) VLAN tag if present, otherwise empty
      */
-    uint32_t eth_type;
+    std::optional<VlanTag> GetVlanTag() const {
+        if ( ! vlan_present )
+            return {};
+
+        if ( ! std::in_range<uint16_t>(vlan) )
+            reporter->InternalError("VLAN ID value %u exceeds 16 bits", vlan);
```
awelzel: Would expect the checks to be done in the Set* methods rather than the Get* methods? Any motivation for delaying the check to Get*?

testing/btest/scripts/base/protocols/vlan/vlan-0-qinq.zeek
```diff
@@ -0,0 +1,8 @@
+# @TEST-DOC: Verify VLAN of 0 gets logged with the two ways to log it
```
awelzel:
```suggestion
# @TEST-DOC: Verify VLAN of 0 gets logged with the two ways to log it with double tagged packets (QinQ).
```

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
```diff
@@ -42,26 +42,31 @@ class IPVlanConnKey : public zeek::IPBasedConnKey {
...
         static int inner_vlan_offset = -2;
 
         if ( vlan_offset == -2 && inner_vlan_offset == -2 ) {
             vlan_offset = id::conn_id_ctx->FieldOffset("vlan");
             if ( vlan_offset < 0 || id::conn_id_ctx->GetFieldType(vlan_offset)->Tag() != TYPE_INT )
                 vlan_offset = -1;
 
+            vlan_present = id::conn_id_ctx->FieldOffset("vlan_present");
+            if ( vlan_offset < 0 || id::conn_id_ctx->GetFieldType(vlan_offset)->Tag() != TYPE_INT )
+                vlan_offset = -1;
```
awelzel: This seems superfluous. vlan_present doesn't exist and the code doesn't check its presence. Just delete this hunk and the new static variable?

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
```diff
@@ -83,9 +90,11 @@ class IPVlanConnKey : public zeek::IPBasedConnKey {
         struct detail::PackedConnTuple tuple;
         // Add 802.1Q vlan tags to connection tuples. The tag representation
         // here is as in the Packet class (where it's oddly 32-bit), since
-        // that's where we learn the tag values from. 0 indicates absence.
+        // that's where we learn the tag values from.
         uint32_t vlan;
         uint32_t inner_vlan;
+        bool vlan_present;
+        bool inner_vlan_present;
```
awelzel: I guess that makes sense. We could encode 0xFF000000 as not set, too, as in the tutorial for VXLAN VNI:

> We override DoPopulateConnIdVal() to set the vxlan_vni field of the conn_id_ctx record value to the extracted VXLAN VNI. A small trick employed is that we default the most significant byte of key.vxlan_vni to 0xFF. As a VNI has only 24 bits, this allows us to determine if a VNI was actually extracted, or whether it remained unset.

https://docs.zeek.org/en/v8.0.2/devel/plugins/connkey-plugin.html
If that makes you shudder, leave it as is :-)

src/packet_analysis/protocol/vlan/VLAN.cc
```diff
@@ -17,19 +17,13 @@ bool VLANAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
...
-    }
-    else {
-        packet->inner_vlan = vlan_id;
-        packet->inner_vlan_pcp = vlan_pcp;
-        packet->inner_vlan_dei = vlan_dei;
-    }
+    if ( ! packet->GetVlanTag() )
+        packet->SetVlanTag({.id = vlan_id, .pcp = vlan_pcp, .dei = vlan_dei});
+    else
+        packet->SetInnerVlanTag({.id = vlan_id, .pcp = vlan_pcp, .dei = vlan_dei});
```
awelzel: Definitely existing, I wonder if we should check for InnerVlanTag() being already set (no need to do it in this PR)
```
if ( GetInnerVlanTag() )
    reporter->Weird("triple_tagged_vlan_qinqinq_unsupported");
```

src/iosource/Packet.h
```diff
@@ -148,48 +160,142 @@ class Packet {
...
+    uint32_t link_type;           /// pcap link_type (DLT_EN10MB, DLT_RAW, etc)
 
     /**
-     * (Innermost) VLAN DEI if inner_vlan is set, otherwise invalid
+     * If layer 2 is Ethernet, innermost ethertype field.
      */
+    uint32_t eth_type;
+
+    // These should be made combined into 2 private, optional VlanTag instances
+    [[deprecated("Remove in v9.1. Use GetVlan() instead.")]]
```
awelzel: Would probably useful to mention the `SetVlanTag()`, too.

src/iosource/Packet.cc
```diff
@@ -46,13 +46,8 @@ void Packet::Init(int arg_link_type, pkt_timeval* arg_ts, uint32_t arg_caplen, u
...
 
-    vlan = 0;
-    vlan_pcp = 0;
-    vlan_dei = false;
-
-    inner_vlan = 0;
-    inner_vlan_pcp = 0;
-    inner_vlan_dei = false;
+    ClearVlanTag();
+    ClearInnerVlanTag();
```
awelzel: Maybe fold them together into `ClearVlanTags()` ?

src/iosource/Packet.h
```diff
@@ -148,48 +160,142 @@ class Packet {
...
+        vlan_dei = false;
+        vlan_present = false;
+    }
 
     /**
-     * (Outermost) VLAN DEI if vlan is set, otherwise invalid
+     * (Innermost) VLAN ID if present, otherwise empty
      */
-    bool vlan_dei = false;
+    std::optional<uint32_t> GetInnerVlan() const {
```
awelzel: May or may not skip on that one. A user could do
```
if ( auto tag = GetInnerVlanTag() ) {
  id = tag.id;
}

```

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
evantypanski: oops, that's a bug! It was leftover from an experiment, my bad. Good eye

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
evantypanski: I didn't in order to avoid an API break (then any code that relied on `vlan == 0` meaning not set is broken). With this we guarantee that `vlan == 0` can still mean that it's not set, but also internally we can just rely on `vlan_present`.

Though, I would prefer your solution here

src/packet_analysis/protocol/vlan/VLAN.cc
evantypanski: maybe, I mean really you can have arbitrarily nested vlan tags, right? So `triple` would be a misnomer in there, and really Zeek *should* handle that. Not sure why we only handle the outermost/innermost, but it's pretty clear looking at Zeek's documentation that that's the expected behavior.

I would instead favor changing this to a vector or something to handle arbitrary numbers of VLAN tags.

src/iosource/Packet.h
evantypanski: My thought process was that internally, we already hold these guarantees when using the `Set*` methods because the `VlanTag` struct's types are proper. But, a user can then go in and change `vlan` to be `> UINT16_MAX` (via direct field access). Then, Zeek's core may try to get the tag, and that should be an internal error. 

When we deprecate direct access to `vlan` and `inner_vlan`, then these checks are not necessary since they will always use the `Set*` method, which takes in the correct types.

With that being said, this change could do with a "deprecation guide" since there are a couple moving parts to clean up

stevesmoot:
Broadening vlan support beyond two has come up before…

evantypanski:
If we're already deprecating these and want to broaden VLAN support to more than just inner/outermost, perhaps we should expand this so that it doesn't go through multiple cycles.

I could see a function to return a vector of VLAN tags, then one to grab the number of vlan tags and the Nth vlan tag. Then `vlan` is `NthVlanTag(0)` and `inner_vlan` is `NthVlanTag(NumVlanTags()-1)` or something.

Doing these now would be more convenient for users, rather than two sets of deprecations. Any opinions here?

The logs might need some consideration, but we can keep them as-is for 9.0 and change around when the inner/outer vlan logic gets removed. Internally we'll just give access to each tag until then.

src/packet_analysis/protocol/vlan/VLAN.cc
0xxon: I think technically you can have arbitrarily nested vlan tags. However, as @awelzel indicated, the only standard for this that I am really aware of is IEEE 802.1ad (QinQ) - which allows exactly 2 vlans. I am not aware of (and have never seen) more than 2 vlans being used/supported by anything. So - the easy path seems to just be a Weird, as @awelzel proposes, until someone shows that this actually is used anywhere :)

0xxon:
> Broadening vlan support beyond two has come up before…

And since I just commented above - do you know where/how that's used? :)

src/packet_analysis/protocol/vlan/VLAN.cc
0xxon: And, just noticing the comment by @smoot below - I might be wrong about this being uncommon.

stevesmoot: Thats @stevesmoot (-;  Anyway, the "easy" one is real Q-in-Q traffic that's grabbed and forwarded to Zeek, so gets a Geneve around it, thus 3 layers.  I asked my team if they had other examples, will add it furitful.

evantypanski: Hm, I would hope that case would currently be supported actually but cannot say with certainty

I did look up Suricata and found a [counter for `counter_vlan_qinqinq`](https://github.com/OISF/suricata/blob/539e4ee6654f1642f45f1bd35dbef85e2728b7e3/src/decode-vlan.c#L66), wonder how often it is used!

It definitely is hard to find cases where more than 2 tags are even mentioned online though. I guess in this case, a weird would give more information about this case than we currently do without the downsides of a vector for 99.999% of people. I just want people to have some way of getting the information if it is important rather than silently discarding it

0xxon: QinQ over another encapsulation is not a problem, as there still only will be two vlan tags; other encapsulations are parsed separately. I created a test-pcap - and TCP over QinQ over Geneve is decoded/logged as expected. Even TCP over QionQ over Geneve over QinQ is not a problem and logged correctly.

With that, I think I stand by my earlier opinion, and just do as suggested by @awelzel and output a weird if we encounter more than 2 vlan tags. I don't think it is currently worthwhile to add support for more 2 tags at the moment - as mentioned, I don't know of anyone supporting this.

Test pcaps:
[geneve_qinq_http.pcap.txt](https://github.com/user-attachments/files/25391086/geneve_qinq_http.pcap.txt)
[double_qinq_geneve.pcap.txt](https://github.com/user-attachments/files/25391131/double_qinq_geneve.pcap.txt)

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
0xxon: Just to chime in here - I get that we want to not have an API breakage, but encoding this as @awelzel suggested still looks nicer & cleaner to me.

evantypanski: I like it more too, if the API break is fine then cool!

I'll fix this and add the weird for triple-VLAN tags when I'm not neck deep in some XDP stuff :)

0xxon: FWIW - I would be ok with the API break; perhaps we should ask on Slack if anyone depends on this / look through packages.zeek.org if we see anything that might depend on it.

detection-labs:
In modern Service Provider networks, it’s possible thru IETF spec (and I have seen) to have a mix of `802.1ad` and `802.1Q` 'tags'.

`IEEE 802.1ah` (aka PBB / MAC-in-MAC)
https://documentation.nokia.com/sas/23-9/books/mts-services-overview/ieee-802-1ah-provider-backbone-bridg-pbb.html

Look out for Ethernet frame stack starting with Ethertype 0x88a8....

Scenario: Service Provider carrying a customer’s QinQ traffic over their backbone, you'd end up with:

```
+-----------------------------------------------------------+
|      Ethernet Header (B-MAC: Backbone Destination/Source) |
+-----------------------------------------------------------+
|  B-TAG (802.1ad/Q) | TPID: 0x88a8 | VID: 1  | S-VLAN      | <--- Tag 1
+-----------------------------------------------------------+
|  I-TAG (802.1ah)   | TPID: 0x88e7 | ISID: 10| PBB Instance| <--- Tag 2
+-----------------------------------------------------------+
|      Encapsulated Customer Ethernet Header (C-MAC)        |
+-----------------------------------------------------------+
|  C-TAG (802.1Q)    | TPID: 0x8100 | VID: 2  | Customer V-1| <--- Tag 3
+-----------------------------------------------------------+
|  C-TAG (802.1Q)    | TPID: 0x8100 | VID: 3  | Customer V-2| <--- Tag 4
+-----------------------------------------------------------+
|              IP Payload (IPv4/IPv6/ICMP)                  |
+-----------------------------------------------------------+
```

image


_This raised an interesting question for me on Zeek's handling.. running a example through Zeek 8.1.1 shows_
{
  "ts": 1771459472.901108,
  "uid": "CGDkTE26TkuyYu7JM8",
  "id.orig_h": "10.25.123.123",
  "id.orig_p": 8,
  "id.resp_h": "10.25.222.222",
  "id.resp_p": 0,
  "proto": "icmp",
  "conn_state": "OTH",
  "local_orig": true,
  "local_resp": true,
  "missed_bytes": 0,
  "orig_pkts": 1,
  "orig_ip_bytes": 28,
  "resp_pkts": 0,
  "resp_ip_bytes": 0,
  "ip_proto": 1,
  "vlan": 1,
  "inner_vlan": 3,
  "orig_l2_addr": "00:00:00:00:00:01",
  "resp_l2_addr": "00:00:00:00:00:02"
}

So looks like Zeek 8.1.1 handling of this packet structure is not correct, as it's recording the initial B-TAG (802.1ad) via vlan, I am guessing there isn't the structure to pull out B-TAG 802.1ad / I-TAG 802.1ah ISID.

My mind also leads to think about L2/MAC too... how to represent the true PHY MAc's
> Ethernet II, Src: Dell_89:ba:fa (d4:ae:52:89:ba:fa), Dst: Dell_33:4b:9f (78:2b:cb:33:4b:9f)


I recall PBB handling was recently added, found it in notes "6.0.0-dev.75".

src/packet_analysis/protocol/ip/conn_key/vlan_fivetuple/Factory.cc
evantypanski: I moved to `0xFF000000` for these. I searched the package browser and found no one using `vlan` in C++ code, then asked in Zeek's development channel and no one responded.

It seems like a relatively safe break since the script level stuff remains the same (except now `0` is possible)

src/iosource/Packet.h
```diff
@@ -148,6 +162,95 @@ class Packet {
...
+#elif defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#endif
+    /**
+     * (Outermost) VLAN tag if present, otherwise empty
+     */
+    std::optional<VlanTag> GetVlanTag() const {
+        if ( vlan == vlan_unset_val )
+            return {};
```
awelzel:
```suggestion
            return std::nullopt;
```

Nit to align with the inner version.

testing/btest/scripts/base/protocols/vlan/vlan-qinqinq.zeek
```diff
@@ -0,0 +1,6 @@
+# @TEST-DOC: Verify triple-tagged QinQinQ traffic produces a weird
```
awelzel, line 1: Mostly nit.

It seems the IP part is corrupted (according to wireshark and the truncated_IP_len wired). I think that's okay, but if you created the PCAP with scapy or so, maybe can keep it intact in case the PCAP is reused if we ever add support?

src/packet_analysis/protocol/vlan/VLAN.cc
```diff
@@ -22,8 +22,13 @@ bool VLANAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
     bool vlan_dei = (tci & 0x1000) != 0;
     if ( ! packet->GetVlanTag() )
         packet->SetVlanTag({.id = vlan_id, .pcp = vlan_pcp, .dei = vlan_dei});
-    else
+    else {
+        if ( auto inner = packet->GetInnerVlanTag() )
+            Weird("triple_tagged_vlan_unsupported", packet,
+                  util::fmt("overwrote inner vlan ID of %d with %d", inner->id, vlan_id));
```
awelzel: I'm a bit afraid this might end up pretty expensive in environments where it affects each and every packet. The `fmt()` isn't super cheap and the `Weird()` with a packet does quite a bit of work, too.

Don't have a good idea outside of `static bool raised_weird = false` and only raise it once. We don't have a session yet, so cannot record per session.

Maybe a `enable_too_many_vlan_tags_weird: bool = T` in a const.bif, so the code path can be turned off in the worst case?

Not sure TBH. Might be overthinking. Minimally a comment for those running `perf record`  / `perf report` that end up here with the idea of a toggle could help a bit.

src/iosource/Packet.h
```diff
@@ -148,6 +162,95 @@ class Packet {
      */
     static constexpr const u_char L2_EMPTY_ADDR[L2_ADDR_LEN] = {0};
 
+    struct VlanTag {
+        uint16_t id = 0; // This struct is used optionally so no need to encode as 0xFF000000
+        uint8_t pcp = 0;
+        bool dei = false;
+
+        auto operator<=>(const VlanTag&) const = default;
```
awelzel: I got a bit hung up on this (but learned something, too :-) ). We only need this for the tests, and `operator==` would suffice. Seems okay (maybe better) if `tag1 > tag2` doesn't actually work. Would you prefer sticking with `<=>` or ack on reducing to `operator==` ?

src/iosource/Packet.h
```diff
@@ -162,34 +265,18 @@ class Packet {
      */
     uint32_t eth_type;
 
-    /**
-     * (Outermost) VLAN tag if any, else 0.
-     */
+    // These should be made combined into 2 private, optional VlanTag instances
+    [[deprecated("Remove in v9.1. Use GetVlanTag() and SetVlanTag(tag) instead.")]]
     uint32_t vlan = 0;
```
awelzel: this will always go through ClearVlanTags(), so maybe assign it `vlan_unset_val` directly to avoid confusion (same for the inner).

testing/btest/scripts/base/protocols/vlan/vlan-qinqinq.zeek
evantypanski: yeah I think it was scapy. I will replace it with `http/get.trace` but with 3 vlans

src/packet_analysis/protocol/vlan/VLAN.cc
evantypanski: So I got rid of the `util::fmt` call, even though that gets rid of info I think we should keep. You're right that if each packet is triple tagged that'll be huge.

Not as worried about the `Weird` call overhead, since there are mechanisms there for repeat messages. I kept it as a separate commit and will keep this unresolved just in case, though

awelzel: wfm, resolving.

evantypanski:
> One main concern is about Packet.h vlan / inner_vlan: Users that accessed this directly can now observe 0xFF000000 instead of 0, correct?

Yeah, that's why I was worried about the API break above. Though I searched through anything relevant on packages.zeek.org and asked in Zeek's #development channel, neither of which seemed to have any uses of the C++ version of those. The script ones are unchanged, other than `0` is now possible.

So I'm not too worried, even the XDP shunter only ever used the script version.

With that being said, it may be reasonable to just make them private anyway, since apparently from my research they're unused. That would mean we could get rid of bounds checks for like `GetVlanTag()`. That breaks deprecations even more though.

awelzel:
> With that being said, it may be reasonable to just make them private anyway

As in person, I think that's reasonable. For anyone accessing `vlan` / `inner_vlan` on Packet directly from C++ it's likely better if we break them rather than surprisingly returning `0xff000000` in some cases now. With a proper NEWS entry I think it's okay to break the API.

Feel free to keep the approval, or re-request a review.

evantypanski:
Chatted with @awelzel and just keeping it public and using `std::optional` seems better to me, actually. Kind of went full circle since I dismissed it as not possible because of the API break. This feels like the least surprising impl to me.

Note that I kept the connkey as using the `0xFF000000`. I worried a little about an optional in something so integrated with scripts, and it seems really easy to leave that as an implementation detail for that particular part. Happy to change it too. Funnily enough lowering it to a `uint16_t` with `0xFFFF` as unset had some log ordering implications

NEWS
```diff
@@ -34,6 +34,16 @@ Breaking Changes
 - The packet captures in the btest Traces library have standardized on .pcap and .pcapng
   file suffixes. This may cause breakage in plugins that were using the old .trace suffix.
 
+- The public ``vlan`` and ``inner_vlan`` members within C++ ``Packet`` has changed to an
+  optional type. This does not change scripts. In order to access the ID as before, check
```
```suggestion
...has changed from uint32_t to an optional VlanTag struct. The corresponding `dei` and `pcp` members have been removed and merged into the VlanTag struct. This does not change scripts. In order to access the ID as before, check
```

NEWS
```diff
@@ -34,6 +34,16 @@ Breaking Changes
...
+  the optional first:
+
+    if ( packet.vlan )
+      my_vlan = packet.vlan->id;
+
+  This allows Zeek to properly handle VLAN tags of 0, which before were handled as if
```
```suggestion
  This allows Zeek to properly handle VLAN IDs of 0, which before were handled as if
```

src/iosource/Packet.h
```diff
@@ -148,6 +148,14 @@ class Packet {
      */
     static constexpr const u_char L2_EMPTY_ADDR[L2_ADDR_LEN] = {0};
 
+    struct VlanTag {
+        uint16_t id = 0;
+        uint8_t pcp = 0;
+        bool dei = false;
```
awelzel, line 156:
```suggestion
        // Drop Eligible Indicator
        bool dei = false;
```

src/iosource/Packet.h
```diff
@@ -148,6 +148,14 @@ class Packet {
      */
     static constexpr const u_char L2_EMPTY_ADDR[L2_ADDR_LEN] = {0};
 
+    struct VlanTag {
+        uint16_t id = 0;
+        uint8_t pcp = 0;
```
awelzel, line 154:
```suggestion
        // Priority Code Point
        uint8_t pcp = 0;
```

NEWS
awelzel: You could also put an entry under new/changed functionality about support for VLANs with ID 0 :-)

PR #5111, opened 2025-12-18
Replace use of modp methods with std::to_chars

timwoj:
This PR replaces the uses of `modp_dtoa{2,3}` and `modp_{ul,l}toa10` with `std::to_chars`. In a synthetic benchmark, `to_chars` is considerably faster for the double conversions, but we might as well replace all of the `modp` methods. In local testing, all of the btests still pass except for the `language.double` btest which had output changes to very small fractional numbers.

We should eventually remove `src/3rdparty/modp_dtoa.{h,c}` as well, but I didn't do it as part of this PR.

Benchmark results:
Screenshot 2026-01-03 at 6 05 50 PM

### DIFF
--- a/src/Desc.cc
+++ b/src/Desc.cc
@@ -146,11 +146,14 @@ void ODesc::Add(double d, bool no_exp) {
     if ( IsBinary() )
         AddBytes(&d, sizeof(d));
     else {
-        auto res = util::double_to_str(d, IsReadable() ? 6 : 8, no_exp);
+        char tmp[350];
+        auto res = util::double_to_str(d, tmp, sizeof(tmp), IsReadable() ? 6 : 8, no_exp);
+        if ( res == 0 )
+            return;
 
-        AddBytes(res.data(), res.size());
+        AddBytes(tmp, res);
 
-        if ( util::approx_equal(d, nearbyint(d), 1e-9) && std::isfinite(d) && (res.find('e') == std::string::npos) )
+        if ( util::approx_equal(d, nearbyint(d), 1e-9) && std::isfinite(d) && (strchr(tmp, 'e') == nullptr) )
             // disambiguate from integer
             Add(".0");
     }
--- a/src/util.cc
+++ b/src/util.cc
@@ -1988,56 +1988,55 @@ string escape_utf8(string_view val, int flags) {
  */
 bool approx_equal(double a, double b, double tolerance) { return std::abs(a - b) < std::abs(tolerance); }
 
-std::string double_to_str(double d, int precision, bool no_exp) {
+size_t double_to_str(double d, char* buf, size_t buf_size, int precision, bool no_exp) {
     // Short-circuit check for NaN, which should always return the same string
-    if ( std::isnan(d) )
-        return "nan";
+    if ( std::isnan(d) ) {
+        strcpy(buf, "nan");
+        buf[3] = '\0';
+        return 3;
+    }
 
-    // work with positive values and add the negative sign at the end if needed. Also special
-    // case negative zero so it ends up positive in the end.
-    bool neg = false;
+    char* start = buf;
+    char* end = buf + buf_size;
+
     if ( d < 0 ) {
-        neg = true;
+        buf[0] = '-';
+        start += 1;
         d = -d;
     }
     else if ( d == -0.0 )
         d = 0.0;
 
-    // Buffer needs enough chars to store max. possible "double" value
-    // of 1.79e308 without using scientific notation.
-    char buf[350];
-    memset(buf, 0, sizeof(buf));
-
-    std::string res;
     std::to_chars_result result;
     if ( ! no_exp && d > static_cast<double>(std::numeric_limits<int>::max()) ) {
-        result = std::to_chars(buf, buf + sizeof(buf), d, std::chars_format::scientific);
-        res = std::string{buf, static_cast<size_t>(result.ptr - buf)};
+        result = std::to_chars(start, end, d, std::chars_format::scientific);
+        if ( result.ec == std::errc::value_too_large )
+            return 0;
+
+        *result.ptr = '\0';
+        return result.ptr - buf;
     }
     else if ( ! no_exp && d < pow(10.0, precision * -1.0) ) {
-        result = std::to_chars(buf, buf + sizeof(buf), d, std::chars_format::general, precision);
-        res = std::string{buf, static_cast<size_t>(result.ptr - buf)};
-    }
-    else {
-        result = std::to_chars(buf, buf + sizeof(buf), d, std::chars_format::fixed, precision);
-        res = std::string{buf, static_cast<size_t>(result.ptr - buf)};
+        result = std::to_chars(start, end, d, std::chars_format::general, precision);
+        if ( result.ec == std::errc::value_too_large )
+            return 0;
 
-        // We require `std::reverse_iterator::base` here which in e.g., gcc-10.2.1
-        // is not implemented for the range equivalent of the code
-        // (`borrowed_iterator_t` over a `reverse_view`). Stick to the non-ranges
-        // version for now.
-        if ( res.ends_with('0') ) {
-            // NOLINTNEXTLINE(modernize-use-ranges)
-            res.erase(std::find_if(res.rbegin(), res.rend(), [](char c) { return c != '0'; }).base(), res.end());
-            if ( res.back() == '.' )
-                res.pop_back();
-        }
+        *result.ptr = '\0';
+        return result.ptr - buf;
     }
 
-    if ( neg )
-        res.insert(0, 1, '-');
+    result = std::to_chars(start, end, d, std::chars_format::fixed, precision);
+    if ( result.ec == std::errc::value_too_large )
+        return 0;
 
-    return res;
+    char* last = result.ptr - 1;
+    while ( *last == '0' )
+        last--;
+    if ( *last != '.' )
+        last++;
+
+    *last = '\0';
+    return last - buf;
 }
 
 
--- a/src/util.h
+++ b/src/util.h
@@ -648,7 +648,7 @@ inline std::vector<std::wstring_view> split(const wchar_t* s, const wchar_t* del
 /**
  * Returns a string version of a double value following all of the rules from the modp_dtoa methods.
  */
-std::string double_to_str(double v, int precision, bool no_exp);
+size_t double_to_str(double v, char* buf, size_t buf_size, int precision, bool no_exp);
 
 } // namespace util
 } // namespace zeek
--- a/src/script_opt/ZAM/BuiltInSupport.cc
+++ b/src/script_opt/ZAM/BuiltInSupport.cc
@@ -70,9 +70,11 @@ void FixedCatArg::RenderInto(const ZVal& z, char*& res, char* res_end) {
         case TYPE_DOUBLE:
         case TYPE_TIME: {
             auto d = z.AsDouble();
-            str = util::double_to_str(d, 6, false);
-            strncpy(res, str.c_str(), str.size());
-            res += str.size();
+            auto chars_output = util::double_to_str(d, res, res_end - res, 6, false);
+            if ( chars_output == 0 )
+                break;
+
+            res += chars_output;
 
             if ( util::approx_equal(d, nearbyint(d), 1e-9) && std::isfinite(d) ) {
                 // disambiguate from integer


src/util.cc
```diff
@@ -1993,6 +1994,59 @@ string escape_utf8(string_view val, int flags) {
...
+        neg = true;
+        d = -d;
+    }
+    else if ( d == -0.0 )
+        d = 0.0;
+
+    // Buffer needs enough chars to store max. possible "double" value
+    // of 1.79e308 without using scientific notation.
+    char buf[350];
+    memset(buf, 0, sizeof(buf));
```
awelzel: This actually clears 350 bytes on the stack everytime?

I'm wondering if this could be done via `std::string::data()` directly (e.g. resize just under the SSO limit) and if to_chars() returns `value_too_large`, resize the string to something larger? It seems that for example timestamps shouldn't ever need the full 350 bytes.

src/util.cc
```diff
@@ -1993,6 +1994,59 @@ string escape_utf8(string_view val, int flags) {
...
+        if ( res.ends_with('0') ) {
+            // NOLINTNEXTLINE(modernize-use-ranges)
+            res.erase(std::find_if(res.rbegin(), res.rend(), [](char c) { return c != '0'; }).base(), res.end());
+            if ( res.back() == '.' )
+                res.pop_back();
+        }
+    }
+
+    if ( neg )
+        res.insert(0, 1, '-');
```
awelzel: This needs to move all chars around, no? Could this be done up front and then use `string.data() + 1` for `std::to_chars()` ?

src/IPAddr.cc
```diff
@@ -224,12 +224,13 @@ IPPrefix::IPPrefix(const IPAddr& addr, uint8_t length, bool len_is_v6_relative)
 std::string IPPrefix::AsString() const {
     char l[16];
 
+    std::to_chars_result res;
```
awelzel, line 233: The maximum expected prefix length is `128`? Seems you could directly render into a 3 char std::string and use that for concat later instead of going through the intermediary std::string and `l`  Same above.

timwoj:
The changes to make it directly render into std::string made it a win across the board in the benchmarker:

Screenshot 2026-01-09 at 9 24 03 AM

src/IPAddr.cc
```diff
@@ -222,14 +222,21 @@ IPPrefix::IPPrefix(const IPAddr& addr, uint8_t length, bool len_is_v6_relative)
 }
 
 std::string IPPrefix::AsString() const {
-    char l[16];
+    std::string str = prefix.AsString() + "/";
+    size_t prefix_len = str.size();
+    str.reserve(prefix_len + 16);
```
awelzel: Hmm, I realize this `reserve()` here is actually UB and we'd need to go for `resize()`.

This is what I ended up with:

```
std::string IPPrefix::AsString() const {
    std::string str = prefix.AsString() + "/";
    size_t prefix_len = str.size();
    str.resize(prefix_len + 3);

    assert(length >= 0 && length <= 128); // length should only need 3 digits!

    char* start = str.data() + prefix_len;
    std::to_chars_result res;
    if ( prefix.GetFamily() == IPv4 )
        res = std::to_chars(start, start + 3, length - 96);
    else
        res = std::to_chars(start, start + 3, length);

    if ( res.ec == std::errc::value_too_large )
        zeek::reporter->InternalError("uhhh\n");

    // Resize the string according to the characters used.
    str.resize(res.ptr - str.data());
    return str;
}
```

I've added a microbenchmark setup (will post in summary) and that version results in:
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
IPPrefix4_AsString        228 ns          228 ns      3174374
IPPrefix6_AsString        360 ns          360 ns      1920521
```

where the original version was:

```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
IPPrefix4_AsString        231 ns          231 ns      3004343
IPPrefix6_AsString        382 ns          382 ns      1823783
```

The version in this PR is as follows:
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
IPPrefix4_AsString        254 ns          254 ns      2755203
IPPrefix6_AsString        364 ns          364 ns      1915737
```

src/Desc.cc
```diff
@@ -129,18 +146,11 @@ void ODesc::Add(double d, bool no_exp) {
     if ( IsBinary() )
         AddBytes(&d, sizeof(d));
     else {
-        // Buffer needs enough chars to store max. possible "double" value
-        // of 1.79e308 without using scientific notation.
-        char tmp[350];
+        auto res = util::double_to_str(d, IsReadable() ? 6 : 8, no_exp);
```
awelzel: With a microbenchmark, `ODesc::Add()` has become a factor of 2-4 slower with this change. I think that's because of `util::double_to_str()`, so would skip this one.

Current master version:
```
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
ODesc_Add_double/0                49.0 ns         49.0 ns     13325812
ODesc_Add_double/1                48.8 ns         48.8 ns     11803954
ODesc_Add_double/128              50.8 ns         50.8 ns     13749893
ODesc_Add_double/16384            54.2 ns         54.2 ns     12910759
ODesc_Add_double/2097152          56.5 ns         56.5 ns     12359533
ODesc_Add_double/268435456        60.9 ns         60.8 ns     11562970
ODesc_Add_double/1768397792       62.8 ns         62.8 ns     11180516
```

New version in this PR:
```
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
ODesc_Add_double/0                 107 ns          107 ns      6522701
ODesc_Add_double/1                 130 ns          130 ns      5419160
ODesc_Add_double/128               128 ns          128 ns      5470111
ODesc_Add_double/16384             132 ns          132 ns      5168054
ODesc_Add_double/2097152           134 ns          134 ns      5200221
ODesc_Add_double/268435456         264 ns          264 ns      2636543
ODesc_Add_double/1768397792        263 ns          263 ns      2662913
```

Think the jump from 134 to 264 is due to an extra memory allocation. But it seems pretty clear that we should stick to the on stack buffer :-)

src/Desc.cc
```diff
@@ -120,7 +133,11 @@ void ODesc::Add(uint64_t u) {
         AddBytes(&u, sizeof(u));
     else {
         char tmp[256];
-        modp_ulitoa10(u, tmp);
+        auto res = std::to_chars(tmp, tmp + sizeof(tmp), u);
```
awelzel, line 136: Haven't tested this one, but std::to_chars() for unsigned / signed int seems faster (`int-count-to-str`).

src/threading/Formatter.cc
```diff
@@ -79,20 +80,21 @@ threading::Value::addr_t Formatter::ParseAddr(const std::string& s) const {
 std::string Formatter::Render(const threading::Value::subnet_t& subnet) {
```
awelzel, line 80: This is terrible to standalone perf or unit test because `Formatter` is abstract and also wants a `MsgThread`, maybe we could extract the actual rendering into free functions and then add a benchmark, but I believe the same findings as for `IPPrefix::AsString()` apply.

src/threading/Formatter.cc
```diff
@@ -79,20 +80,21 @@ threading::Value::addr_t Formatter::ParseAddr(const std::string& s) const {
...
+    std::string s = Render(subnet.prefix) + "/" + std::string{l, static_cast<size_t>(res.ptr - l)};
 
     return s;
 }
 
 std::string Formatter::Render(double d) {
     char buf[256];
-    modp_dtoa(d, buf, 6);
-    return buf;
+    auto res = std::to_chars(buf, buf + sizeof(buf), d, std::chars_format::fixed, 6);
```
awelzel, line 96: On my system this is slower than `modp_dtoa()` - do you see something else?

```
BM_double_modp_dtoa/0                             19.8 ns         19.7 ns     35815677
BM_double_modp_dtoa/1                             21.7 ns         21.6 ns     34083126
BM_double_modp_dtoa/128                           25.6 ns         25.6 ns     29773496
BM_double_modp_dtoa/16384                         27.5 ns         27.5 ns     23998932
BM_double_modp_dtoa/2097152                       32.8 ns         32.8 ns     21418607
BM_double_modp_dtoa/268435456                     38.8 ns         38.8 ns     18493759
BM_double_modp_dtoa/1768397792                    39.9 ns         39.9 ns     17962474

BM_double_raw_std_to_chars_fixed/0                44.2 ns         44.2 ns     15833893
BM_double_raw_std_to_chars_fixed/1                64.9 ns         64.9 ns     10708733
BM_double_raw_std_to_chars_fixed/128              63.4 ns         63.3 ns     11065531
BM_double_raw_std_to_chars_fixed/16384            66.8 ns         66.7 ns     10391774
BM_double_raw_std_to_chars_fixed/2097152          67.8 ns         67.8 ns     10325235
BM_double_raw_std_to_chars_fixed/268435456        68.0 ns         68.0 ns     10319039
BM_double_raw_std_to_chars_fixed/1768397792       67.4 ns         67.4 ns     10351328
```

This should also be drop-in replaceable with `double_to_str()`, though likely that's slower.

timwoj: My initial benchmarking comes from a forked version of dtoa-benchmark (https://github.com/timwoj/dtoa-benchmark) which now includes the modp_numtoa code. The output from that looks like:

```
Benchmarking randomdigit double-conversion    ... [ 285.635ns,  456.108ns]
Benchmarking randomdigit dragonbox            ... [ 137.553ns,  149.328ns]
Benchmarking randomdigit fmt                  ... [ 125.644ns,  143.558ns]
Benchmarking randomdigit modp                 ... [ 419.422ns,  499.329ns]
Benchmarking randomdigit null                 ... [   1.463ns,    1.645ns]
Benchmarking randomdigit ostringstream        ... [ 544.173ns,  626.444ns]
Benchmarking randomdigit ryu                  ... [ 110.836ns,  156.462ns]
Benchmarking randomdigit schubfach            ... [  71.999ns,  100.544ns]
Benchmarking randomdigit sprintf              ... [ 397.810ns,  472.770ns]
Benchmarking randomdigit to_chars             ... [  30.015ns,   42.207ns]
Benchmarking randomdigit xjb64                ... [ 116.111ns,  119.948ns]
Benchmarking randomdigit yy                   ... [  87.661ns,  101.979ns]
Benchmarking randomdigit zmij                 ... [ 224.860ns,  232.629ns]
```

There is a slight variation in the modp_numtoa code used there in that I had to extend the number of characters that it can output up to 16. If you run the benchmark locally, you'll get a lot of warnings about roundtrip errors with `modp` because we truncate to smaller strings than it's expecting to use, but that truncation shouldn't affect the time outputs. The allocation of the buffer used for conversion isn't included as part of the benchmark time, it's only the actual conversions. The benchmark works from 1 digit to `std::numeric_limits<double>::max_digits10` (17 on my local system), getting a random value for each range, and runs 10 trials of each range.

Based on the numbers I pasted above, I'm kind of surprised you're seeing such a wide variance between the old modp-based code and the new `to_chars` code. `to_chars` should be so much faster, so I wonder if it's somewhere else in my changes that causes that variance.

awelzel: > Based on the numbers I pasted above, I'm kind of surprised you're seeing such a wide variance between the old modp-based code and the new to_chars code. to_chars should be so much faster, so I wonder if it's somewhere else in my changes that causes that variance.

Thanks! Interesting, yes! modp being slower reproducers here with this benchmark.

> I wonder if it's somewhere else in my changes that causes that variance.

So the above benchmark creates random numbers. When restricting the range of values to what we'd expect to see in logs - for example conn.log's `ts` and `duration` fields (or TTLs in dns.log) - `modp_dtoa()` is faster by ~2x for me. In the initial testing I limited to the current unix timestamp as maximum double value (I sort of think that's fair - I'm blanking on where we'd actually render double values otherwise outside of timestamps and intervals).

It's only with timestamps 13 years in the future that modp_dtoa() becomes like 12x slower (seems like a perf bug in `modp_dtoa()` - maybe?).

I've also added rapidjson's `dtoa()` into the mix, because that's what we'll be using when logging to JSON anyhow.

Does that reproduce on your end? Maybe my CPU is weird - it's kind of a weak one, but I pinned the processes and disabled turbo boost, the numbers should be sensible relative to each other.

```
--------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations
--------------------------------------------------------------------------------------
BM_double_modp_dtoa/0                             20.8 ns         20.8 ns     33838347
BM_double_modp_dtoa/3                             20.4 ns         20.4 ns     34423051
BM_double_modp_dtoa/36                            21.8 ns         21.8 ns     32275618
BM_double_modp_dtoa/360                           24.3 ns         24.3 ns     28879886
BM_double_modp_dtoa/3600                          26.2 ns         26.2 ns     26719899
BM_double_modp_dtoa/1768553641                    39.7 ns         39.7 ns     17597796
BM_double_modp_dtoa/1800089641                    39.5 ns         39.5 ns     17675990
BM_double_modp_dtoa/1926233641                    39.5 ns         39.5 ns     17634237
BM_double_modp_dtoa/2083913641                    39.6 ns         39.6 ns     17693077
BM_double_modp_dtoa/2178521641                     469 ns          469 ns      1499300
BM_double_modp_dtoa/2399273641                     469 ns          469 ns      1485324
BM_double_raw_std_to_chars_fixed/0                44.3 ns         44.3 ns     15824199
BM_double_raw_std_to_chars_fixed/3                62.7 ns         62.7 ns     11169177
BM_double_raw_std_to_chars_fixed/36               62.8 ns         62.8 ns     11153737
BM_double_raw_std_to_chars_fixed/360              64.1 ns         64.1 ns     11086588
BM_double_raw_std_to_chars_fixed/3600             63.7 ns         63.7 ns     10954466
BM_double_raw_std_to_chars_fixed/1768553641       69.8 ns         69.8 ns     10024765
BM_double_raw_std_to_chars_fixed/1800089641       69.9 ns         69.9 ns     10025266
BM_double_raw_std_to_chars_fixed/1926233641       73.3 ns         73.3 ns      9991972
BM_double_raw_std_to_chars_fixed/2083913641       69.8 ns         69.8 ns      9421473
BM_double_raw_std_to_chars_fixed/2178521641       67.4 ns         67.4 ns     10359489
BM_double_raw_std_to_chars_fixed/2399273641       67.6 ns         67.6 ns     10378198
BM_rapidjson_internal_dtoa/0                      48.5 ns         48.5 ns     14491270
BM_rapidjson_internal_dtoa/3                      50.2 ns         50.2 ns     14002569
BM_rapidjson_internal_dtoa/36                     54.1 ns         54.1 ns     12890721
BM_rapidjson_internal_dtoa/360                    58.3 ns         58.3 ns     12108986
BM_rapidjson_internal_dtoa/3600                   68.3 ns         68.2 ns     10228498
BM_rapidjson_internal_dtoa/1768553641             76.5 ns         76.5 ns      9320716
BM_rapidjson_internal_dtoa/1800089641             75.8 ns         75.8 ns      9201416
BM_rapidjson_internal_dtoa/1926233641             75.1 ns         75.1 ns      9319118
BM_rapidjson_internal_dtoa/2083913641             80.2 ns         80.2 ns      8815704
BM_rapidjson_internal_dtoa/2178521641             80.8 ns         80.8 ns      8378917
BM_rapidjson_internal_dtoa/2399273641             87.7 ns         87.7 ns      8287796

```

A micro benchmark writing very many ntp.log entries (which has a ton of `time` and `interval` fields) and looking a bit more on the macro level could be useful here? We'd only expect a change for TSV, as JSON goes through rapidjson anyhow.

---

```
// See the file "COPYING" in the main distribution directory for copyright.

#include <benchmark/benchmark.h>
#include <rapidjson/internal/dtoa.h>
#include <zeek/util.h>
#include <charconv>
#include <cstdio>

// So we have a few digits for precision.
constexpr double wiggle = 0.47114242;

static void BM_double_modp_dtoa(benchmark::State& state) {
    double value = static_cast<double>(state.range(0)) + wiggle;
    char buf[64] = {};

    for ( auto _ : state ) {
        size_t result = modp_dtoa(value, buf, 6);
        if ( result > sizeof(buf) )
            std::exit(1);
    };
};

static void BM_double_raw_std_to_chars_fixed(benchmark::State& state) {
    double value = static_cast<double>(state.range(0)) + wiggle;
    char buf[64] = {};

    for ( auto _ : state ) {
        auto result = std::to_chars(&buf[0], &buf[sizeof(buf) - 1], value, std::chars_format::fixed, 6);
        if ( result.ec == std::errc::value_too_large )
            std::exit(1);
    };
};

static void BM_rapidjson_internal_dtoa(benchmark::State& state) {
    double value = static_cast<double>(state.range(0)) + wiggle;
    char buf[64] = {};

    for ( auto _ : state ) {
        const auto* result = rapidjson::internal::dtoa(value, &buf[0], 6);
        benchmark::DoNotOptimize(result);
    };
};

static void CustomArguments(benchmark::internal::Benchmark* b) {
    // Typical connection durations.
    b->Args({0L});
    b->Args({3L});
    b->Args({36L});
    b->Args({360L});
    b->Args({3600L});

    // Timestamps starting today (2025-01-16)
    b->Args({1768553641L});
    b->Args({1768553641L + 1L * 365 * 24 * 3600});  // A year from now
    b->Args({1768553641L + 5L * 365 * 24 * 3600});  // Five years from now.
    b->Args({1768553641L + 10L * 365 * 24 * 3600}); // 10 years from now.
    b->Args({1768553641L + 13L * 365 * 24 * 3600}); // 13 years from now (slow, 12 years is still fast...
    b->Args({1768553641L + 20L * 365 * 24 * 3600}); // 20 years from now.
}


BENCHMARK(BM_double_modp_dtoa)->Apply(CustomArguments);
BENCHMARK(BM_double_raw_std_to_chars_fixed)->Apply(CustomArguments);
BENCHMARK(BM_rapidjson_internal_dtoa)->Apply(CustomArguments);
BENCHMARK_MAIN();
```

src/util.h
```diff
@@ -645,5 +645,10 @@ inline std::vector<std::wstring_view> split(const wchar_t* s, const wchar_t* del
     return split(std::wstring_view(s), std::wstring_view(delim));
 }
 
+/**
+ * Returns a string version of a double value following all of the rules from the modp_dtoa methods.
+ */
+size_t double_to_str(double v, char* buf, size_t buf_size, int precision, bool no_exp);
```
awelzel, line 651:
```suggestion
size_t double_to_chars(double v, char* first, char *last, int precision, bool no_exp);
```

Just because it feels more consistent with the change to the raw pointers and reads better next to `std::to_chars()`:

And the last pointer so it matches the usage pattern for std::to_chars signature.

src/IPAddr.cc
awelzel: That'd probably still useful to pick up.

awelzel (approved):
After a bit of DM with Tim, I like the last update a lot :-) One more naming/signature nit, but I think that's worth it. And one comment for the prefix rending.

:shipit:

src/util.h
awelzel: Actually, maybe that's not so great because of the null terminator behavior.

🤷‍♂️ - your call.

timwoj:
I ran one more CI build to let os-perf make a benchmark pass at it, and now it's a loss in bunch of cases 🤦🏼 
Screenshot 2026-01-16 at 3 02 11 PM

PR #5112, opened 2025-12-19
Support for dynamic Config::config_files additions

vpax:
`Config::config_files` is a list of files to read to set `option`s. Currently it's a `const` variable so can't be changed dynamically. This PR changes it to `global` so that in a (high-priority) `zeek_init()` handler, a script can adjust its contents dynamically.

### DIFF
--- a/scripts/base/frameworks/config/input.zeek
+++ b/scripts/base/frameworks/config/input.zeek
@@ -12,7 +12,11 @@ export {
 	##
 	## If the same configuration option is defined in several files with
 	## different values, behavior is unspecified.
-	const config_files: set[string] = {} &redef;
+	##
+	## If you change the variable at run-time rather than via &redef,
+	## you need to do so in a zeek_init() event handler with priority
+	## higher than 5 in order to have the change take effect.
+	global config_files: set[string] = {} &redef;
 
 	## Read specified configuration file and apply values; updates to file
 	## are not tracked.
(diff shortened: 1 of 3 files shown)


0xxon:
Hm. On a first glance, I am not entirely sure I like this - from a functionality point of view this should be a const at the moment.

It would probably be possible to change it to an option, but the change handlers would be a bit of a pain to write. 

May I ask what the usecase for this is? You obviously want to be able to change this in `zeek_init` - but what situation do you encounter where you can evoke a high priority zeek init, but can't just redef it?

vpax:
The use-case is an existing package that has a `redef`-able value that it uses to initialize `config_files`. I want, separately (e.g. in a BTest), to change that value and have it change the config file. If the package could use a `zeek_init` handler to modify `config_files` instead of having to use a `redef`, then making the external change is easy. As it is, I need to use
```
redef Config::config_files -= old_value;
redef Config::config_files += new_value;
```
which isn't pretty since it means the external needs to know what to specify for `old_value`.

awelzel:
> The use-case is an existing package that has a redef-able value that it uses to initialize config_files.

Isn't that always bit quirky? Having a `&redef` variable in a package which then uses this variable to `redef` something in Zeek base doesn't really work out right (also [came up here](https://github.com/zeek/zeek/pull/4244#issuecomment-2684967932)).


Is it common to override the config via redef, or is that just a testing vehicle? Because the below works and in the btest you could just set `MY_PACKAGE_CONFIG_FILE` in the environment.

```
module MyPackage;

export {
        const config = |getenv("MY_PACKAGE_CONFIG_FILE")| > 0 ? getenv("MYPACKAGE_CONFIG_FILE") : "default.config";
}


redef Config::config_files += { config };

event zeek_init()
        {
        print "x", config;
        }
```

vpax:
The override will be germane to change in different production environments. That said, your approach with `getenv()` strikes me as a reasonable way to do this, so I'll close this.

PR #5128, opened 2026-01-05
-O gen-standalone-C++ support for creating modules

vpax:
One form of `@ifdef` conditional checks for a module name rather than a script identifier. Scripts compiled to standalone C++ thus need to still provide the module names internal to the scripts, which the PR implements. I also took the opportunity to do some minor factoring of the conditional code.

### DIFF
--- a/src/script_opt/CPP/Driver.cc
+++ b/src/script_opt/CPP/Driver.cc
@@ -52,6 +52,10 @@ void CPPCompile::Compile(bool report_uncompilable) {
         if ( had_to_skip )
             reporter->FatalError("aborting standalone compilation to C++ due to having to skip some functions");
 
+        // Used to find modules only present in the analyzed code.
+        unordered_set<string> analyzed_modules;
+        unordered_set<string> non_analyzed_modules;
+
         for ( auto& g : global_scope()->OrderedVars() ) {
             bool compiled_global = obj_matches_opt_files(g) == AnalyzeDecision::SHOULD;
 
@@ -62,8 +66,16 @@ void CPPCompile::Compile(bool report_uncompilable) {
                         break;
                     }
 
-            if ( ! compiled_global )
+            if ( ! compiled_global ) {
+                if ( g->GetLocationInfo()->FirstLine() != 0 )
+                    // Make sure we're dealing with a global actually used
+                    // elsewhere (for example, weed out some spicy events
+                    // that don't have matching Zeek script).
+                    non_analyzed_modules.insert(g->ModuleName());
                 continue;
+            }
+
+            analyzed_modules.insert(g->ModuleName());
 
             // We will need to generate this global's definition, including
             // its initialization. Make sure we're tracking it and its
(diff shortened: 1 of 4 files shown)


src/scan.l
```diff
@@ -947,20 +947,7 @@ void do_atifdef(const char* id) {
...
-
-    const auto& i = zeek::detail::lookup_ID(id, zeek::detail::current_module.c_str());
-
-    // If the ID didn't exist, look to see if this is a module name instead.
-    bool found = (i != nullptr);
-    if ( ! found )
-        found = (zeek::detail::module_names().count(id) != 0);
-
-    set_ignoring(found, "@ifndef", id);
+    set_ignoring(found != if_true, "@ifdef", id);
```
awelzel: There's one baseline change for the `zeek-profiler-file` test that should've been `@ifndef` instead of `@ifdef`, this should fix it, but see below.

```suggestion
    set_ignoring(found != if_true, if_true ? "@ifdef" : "@ifndef", id);
```

I'd probably keep the explicit functions instead of a bool parameter - they read nicer than the `if_true` construct for me, even if slightly repeated. Maybe a refactor to extract the id lookup, so they can re-use that part?

```
diff --git a/src/scan.l b/src/scan.l
index 0c0d204bdd..ac2c33bb71 100644
--- a/src/scan.l
+++ b/src/scan.l
@@ -937,30 +937,23 @@ void do_atif(zeek::detail::Expr* expr) {
     set_ignoring(! val->AsBool(), "@if", zeek::obj_desc_short(expr));
 }
 
-void do_atifdef(const char* id) {
-    start_conditional();
-
+static bool id_exists(const char *id) {
     const auto& i = zeek::detail::lookup_ID(id, zeek::detail::current_module.c_str());
+    if ( i )
+        return true;
 
     // If the ID didn't exist, look to see if this is a module name instead.
-    bool found = (i != nullptr);
-    if ( ! found )
-        found = (zeek::detail::module_names().count(id) != 0);
+    return zeek::detail::module_names().count(id) != 0;
+}
 
-    set_ignoring(! found, "@ifdef", id);
+void do_atifdef(const char* id) {
+    start_conditional();
+    set_ignoring(! id_exists(id), "@ifdef", id);
 }
 
 void do_atifndef(const char* id) {
     start_conditional();
-
-    const auto& i = zeek::detail::lookup_ID(id, zeek::detail::current_module.c_str());
-
-    // If the ID didn't exist, look to see if this is a module name instead.
-    bool found = (i != nullptr);
-    if ( ! found )
-        found = (zeek::detail::module_names().count(id) != 0);
-
-    set_ignoring(found, "@ifndef", id);
+    set_ignoring(id_exists(id), "@ifndef", id);
 }
```

vpax: Feel free to pick this up, left to my druthers I'll just delete that commit as I don't care strongly.

awelzel: I'll amend and do the merge :+1:

PR #5130, opened 2026-01-06
Docs trailing spaces fix

evantypanski:
@timwoj pointed out [doc8](https://doc8.readthedocs.io/readme.html), I used it to find a bunch of trailing whitespace and fix it. Mostly in Zeekygen for stuff like blank comments that get indented keeping the indent, or tables having trailing whitespace.

A few cases of trailing whitespace in bifs, and one in the docs quickstart RST from... someone :)

there are other categories that will need fixing:

```
Detailed error counts:
    - doc8.checks.CheckCarriageReturn = 0
    - doc8.checks.CheckIndentationNoTab = 26
    - doc8.checks.CheckMaxLineLength = 5612
    - doc8.checks.CheckNewlineEndOfFile = 0
    - doc8.checks.CheckTrailingWhitespace = 0
    - doc8.checks.CheckValidity = 17
```


src/zeekygen/ReStructuredTextTable.cc
```diff
@@ -41,17 +41,36 @@ string ReStructuredTextTable::AsString(char border) const {
...
                 size_t whitespace = longest - last + 1;
-                rval += string(whitespace, ' ');
+                row_str += string(whitespace, ' ');
             }
 
-            rval += row[col];
+            row_str += row[col];
         }
 
+        // Pop off trailing spaces
```
timwoj, line 71: From `util::strstrip`:

```
    auto notspace = [](unsigned char c) { return ! std::isspace(c); };
    s.erase(std::find_if(s.rbegin(), s.rend(), notspace).base(), s.end());
```

I'd suggest using `strstrip` directly, except that it removes leading spaces too which is probably not desired.

timwoj (changes requested):

src/zeekygen/ReStructuredTextTable.cc
evantypanski: Removing leading spaces here breaks the RST as it ends up putting elements in different table cells, like:

``` diff
 :zeek:id:`bloomfilter_counting_init`: :zeek:type:`function`  Creates a counting Bloom filter.
 :zeek:id:`bloomfilter_decrement`: :zeek:type:`function`      Decrements the counter for an element that was added to a counting bloom filter in the past.
 :zeek:id:`bloomfilter_internal_state`: :zeek:type:`function` Returns a string with a representation of a Bloom filter's internal
-                                                             state.
+state.
 :zeek:id:`bloomfilter_intersect`: :zeek:type:`function`      Intersects two Bloom filters.
 :zeek:id:`bloomfilter_lookup`: :zeek:type:`function`         Retrieves the counter for a given element in a Bloom filter.
 :zeek:id:`bloomfilter_merge`: :zeek:type:`function`          Merges two Bloom filters.
```

Screenshot 2026-01-06 at 9 21 42 AM

timwoj: Yep, that's what I figured.

evantypanski: done, I used `std::ranges` though:

```
        row_str.erase(std::ranges::find_if(std::ranges::reverse_view(row_str), notspace).base(), row_str.end());
```

timwoj: Unfortunately there's this comment in util.cc (which I admittedly didn't copy/paste):

```
    // We require `std::reverse_iterator::base` here which in e.g., gcc-10.2.1
    // is not implemented for the range equivalent of the code
    // (`borrowed_iterator_t` over a `reverse_view`). Stick to the non-ranges
    // version for now.
    // NOLINTNEXTLINE(modernize-use-ranges)
```

evantypanski: aw shucks, okay

evantypanski: seems we don't need that anymore so I'll go back to ranges again tomorrow morning, doesn't help I broke CI in the process too :)

evantypanski:
I went back to ranges and fixed the no-lint in `strstrip` in the last commit

PR #5131, opened 2026-01-06
IPAddr: Align IPPrefix::ConvertString() parsing to Zeek script semantics

### LINKED ISSUE #5118: `to_subnet` does not handle IPv6 addresses correctly
With the same prefix size I would expect any IPv6 network to be massively bigger than a IPv4 network with the same prefix size. This holds when specifying literal `subnet` values, but not when obtaining them from `to_subnet`.

```zeek
event zeek_init()
	{
	# WORKS.
	local v6_1 = [::ffff:0.0.0.0];
	local v6_2 = [::];
	local v4 = 0.0.0.0;

	print "from value", v4, v6_1, v6_2;

	assert v4 == v6_1; # Should be equivalent.
	assert v6_1 != v6_2; # Not equivalent.
	}


event zeek_init()
	{
	# WORKS.
	local v6_1 = to_addr("::ffff:0.0.0.0");
	local v6_2 = to_addr("::");
	local v4 = to_addr("0.0.0.0");

	print "from string", v4, v6_1, v6_2;

	assert v4 == v6_1; # Should be equivalent.
	assert v6_1 != v6_2; # Not equivalent.
	}

event zeek_init() 
	{
	# WORKS.
	local v6_1 = [::ffff:0.0.0.0]/1;
	local v6_2 = [::]/1;
	local v4 = 0.0.0.0/1;

	print "from value", v4, v6_1, v6_2;

	assert v6_1 == v6_2; # Should be equivalent.
	assert v4 != v6_1; # Not equivalent, v4 network should be massively smaller.
	}

event zeek_init()
	{
	# BROKEN.
	local v6_1 = to_subnet("::ffff:0.0.0.0/1");
	local v6_2 = to_subnet("::/1");
	local v4 = to_subnet("0.0.0.0/1");

	print "from string", v4, v6_1, v6_2;

	# HERE: BOTH OF THESE FAIL.
	assert v6_1 == v6_2; # Should be equivalent.
	assert v4 != v6_1; # Not equivalent, v4 network should be massively smaller.
	}
```

```console
$ zeek net.zeek
from value, 0.0.0.0, 0.0.0.0, ::
from string, 0.0.0.0, 0.0.0.0, ::
from value, 0.0.0.0/1, ::/1, ::/1
from string, 0.0.0.0/1, 0.0.0.0/1, ::/1
error in net.zeek, line 47: assertion failure: v6_1 == v6_2
fatal error: errors occurred while initializing
```

This affects both script code like in above example, but also the C++ API via `IPPrefix::ConvertString` where I ran into it while doing property tests on an external integration.

awelzel:
If in Zeek script an IPv6 net is parsed, we pass len_is_v6_relative=true into the IPPrefix constructor to account for that, do the same in IPPrefix::ConvertString() to align its and to_subnet() with the behavior of the script parser.

Fixes #5118


src/IPAddr.cc
bbannier, line 247: Drive-by comment: It is weird and probably tech debt how we implement parsing both here and in `scan.l`. Is there any reason why the scanner cannot reuse the code here? Then we probably also wouldn't have to maintain and test `extract_ip_and_len`.

awelzel: Drive-by answer after trying: `to_subnet()` and the C++ API does not support `[` or `]` to be included while it's required within Zeek scripts. One could argue the former should support that, too, but not for this PR / change I'd say.

src/IPAddr.cc
```diff
@@ -259,12 +259,16 @@ bool IPPrefix::ConvertString(const char* text, IPPrefix* result) {
     if ( ! IPAddr::ConvertString(ip_str.data(), &tmp) )
         return false;
 
+    // Set len_is_v6_relative=true if this looks like an IPv6 address to behave
+    // the same as when parsing subnet literals in the scripting language.
+    bool len_is_v6_relative = s.find(':') != std::string::npos;
```
bbannier, line 264: I was wondering whether this heuristic introduces any new edge cases, but we are probably fine. I checked e.g. , `to_subnet("1.2.3.4:8080")` and `to_subnet("1.2.3.4:8080/2")`. These fail similar as before, and are probably not worth capturing in tests.

PR #5132, opened 2026-01-06
Add policy script to disable the opcode and opcode_name fields from dns.log

timwoj:
This needs to go into 8.1 also.

### DIFF
--- a/scripts/policy/protocols/dns/disable-opcode-log-fields.zeek
+++ b/scripts/policy/protocols/dns/disable-opcode-log-fields.zeek
@@ -1,0 +1,6 @@
+##! This script filters the opcode and opcode_name fields from dns.log.
+
+@load base/protocols/dns
+
+redef record DNS::Info$opcode -= { &log };
+redef record DNS::Info$opcode_name -= { &log };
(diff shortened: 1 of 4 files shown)


scripts/policy/protocols/dns/disable-opcode-log-fields.zeek
```diff
@@ -0,0 +1,6 @@
+#! This script filters the opcode and opcode_name fields from dns.log.
```
evantypanski:
```suggestion
##! This script filters the opcode and opcode_name fields from dns.log.
```

right?

ckreibich:
I agree with that @evantypanski — my thought was that by standardizing the naming of such log compatibility scripts we could get the deprecation/removal essentially for free, and simplify its use. For example, if we put this in something like `policy/tuning/previous-release-logs.zeek`, and update this on a per-release basis, we'd make it simpler for people to know how to undo changes, and we'd automatically ensure that this concept only applies to the previous release.

There's some stuff to think through there, like whether to do this on LTS granularity too, and the fact that it might not always be feasible to undo them fully.

I suggest we use Tim's approach right now to not delay RC2, but then start thinking about how we do this going forward.

ckreibich (approved):
LGTM Tim but I suggest we call it out in `NEWS` in the entry where we cover the field additions.

PR #5134, opened 2026-01-07
Bump Spicy.

rsmmr:
Includes a fix for spicyz's internal name spacing, which was still using
the reserved double-underscore (which led to trouble now).


bbannier:
@timwoj this also needs to go into `release/8.1`.

PR #5137, opened 2026-01-08
testing/scripts: Add btest-diff-cut

awelzel:
This is a reimplementation of Benjamin's btest-diff-cut in a few
lines of bash instead of Python to include in the Zeek tree.

This PR also adapts a few of the probably 100ish cases for which it is applicable as a demonstration. I'd probably let AI do the rest, but this is here as an RFC first to show why I think it's useful and visually more appealing to have this kind helper readily available.


testing/scripts/btest-diff-cut
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -eu
+
+[ "$#" -lt 2 ] && echo "Usage: $0 [options] filename" >&2 && exit 1
```
ckreibich: This is a detail but if the options really are optional and I only give a filename, this will error out — is that intended?

testing/scripts/btest-diff-cut
```diff
@@ -0,0 +1,10 @@
...
+
+args=("${@:1:$#-1}") # bash array magic slicing everything but the last argument
+file=${!#}
+tmpfile=$(mktemp --suffix="$file.cut")
+zeek-cut "${args[@]}" <"$file" >"$tmpfile" && mv "$tmpfile" "$file"
```
ckreibich: Should we do a `trap cleanup EXIT` or some such? I don't recall when/whether `zeek-cut` can fail, but it would leave the tempfile around.

ckreibich (approved):
LGTM Arne, thanks! Agree that this is just cleaner all around. I'm just flagging some nits in the script that I'll happily leave to you.

testing/scripts/btest-diff-cut
awelzel: Ah, yeah, can make them optional.

testing/scripts/btest-diff-cut
awelzel: I was sort of thinking that btest sets TMPDIR anyhow to `.tmp`, but have added it.

PR #5138, opened 2026-01-08
Describe current as default docs version

evantypanski:
I think it makes more sense to make the current release the default on docs, not master, so this makes it clear in the documentation. I won't merge this until I do the swap in RTD, probably with 8.1.

So you don't need to build, it looks like this:

Screenshot 2026-01-09 at 9 07 19 AM

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -47,24 +47,23 @@ Documentation Versioning
 
 .. attention::
 
-  The Zeek codebase has three primary branches of interest to users so this
-  document is also maintained as three different versions, one associated with
-  each branch of Zeek.  The default version of `docs.zeek.org
-  <https://docs.zeek.org>`_ tracks Zeek's latest Git development:
+  Zeek publishes both *feature* and *long-term support* releases. By default,
+  the Zeek documentation at `docs.zeek.org <https://docs.zeek.org>`_ points
+  to whichever release is the most recent (or *current*). In the current
+  documentation, you may also find a dropdown menu in the banner, which lets
+  you select the documentation version. For your convenience, the most used
+  versions are:
 
-    * Git *master* branch: https://docs.zeek.org/en/master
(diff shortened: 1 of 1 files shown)


doc/index.rst
```diff
@@ -47,24 +47,23 @@ Documentation Versioning
...
-
-    * Long-Term Support Release: https://docs.zeek.org/en/lts
-    * Current Feature Release: https://docs.zeek.org/en/current
-
-  To help clarify which release you are using, the version numbering
-  scheme for the two release branches is described in the `Release
-  Cadence <https://github.com/zeek/zeek/wiki/Release-Cadence>`_ policy.
-
-  Documentation for older Zeek releases remains available for approximately one
-  full major-version release cycle, i.e., about a year. You can browse recent
```
evantypanski, line 68: We currently have back to 4.0.9, so I don't think the docs are available for "approximately one full major-version release cycle." Instead, I just said they aren't *guaranteed* to stick around. Happy to reword, though

timwoj: I typically keep the last version of every prior lts/feature release. The current releases I keep all of them until they're superseded.

evantypanski: So can we just tell users to check for their version in the dropdown menu, not mentioning they'll get removed?

timwoj: I think you could add something similar to what I said in the last paragraph:

> We typically keep the last version from each release cycle available. The current release cycle(s) (LTS and/or feature) will have all versions available, but some maybe hidden simply for aesthetic reasons the ReadTheDocs UI.

evantypanski: That works, "typically" answers my concerns of possibly having an ever-growing list of versions without cleaning them up :)

timwoj:
I added this to the backport list for 8.0.6 and 8.1.1.

PR #5143, opened 2026-01-09
Topic/timw/use std format for numbers

timwoj:
This is a follow-up to #5111, with a recommendation from @bbannier to try std::format instead of std::to_chars. This works for the integer calls, but the double calls get sticky around scientific notation and the number of digits present, etc. This is a draft PR just to see what the benchmarker thinks of it.

### DIFF
--- a/src/Desc.cc
+++ b/src/Desc.cc
@@ -7,6 +7,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <cstring>
+#include <format>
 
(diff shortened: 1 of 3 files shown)


timwoj:
This is a downgrade in comparison the other PR, so I'm closing this. It was worth a try.

Other PR:
Screenshot 2026-01-12 at 12 36 21 PM

This PR:
Screenshot 2026-01-12 at 12 36 30 PM

Issue #1820, opened 2021-10-18
Events that take EnumTypes as arguments allow other types


timwoj:
I discovered a strange effect of using `EnumType` as arguments for events. Apparently our type checking allows something like this to succeed:

```
event test(t : PacketAnalyzer::Tag) {
	print "3";
}

event zeek_init() {
	schedule 1 secs { test(Files::ANALYZER_X509) };
}
```

I traced it down to a call to `same_type` from `check_and_promote_exprs`. For `TYPE_ENUM`, we have this block:

```
		case TYPE_ENUM:
			// We should probably check to see whether all of the
			// enumerations are present and in the same location.
			// FIXME: Yes, but perhaps we should better return
			// true per default?
			return true;
```

We have the type data in the `Type` arguments passed into `same_type`. Shouldn't it at least compare that the to type IDs are the same? Would this break something else somewhere?

duffy-ocraven:
Also a long-standing situation in enums for Action:: and Weird:: https://github.com/zeek/zeek/issues/1250

vpax:
Time for a blast-from-the-past: this issue is creating a hard-to-fix-separately problem for ZAM. First, this code:
```
type Color: enum { RED, BLUE };
type Size: enum { LITTLE, BIG };

global c = RED;

event zeek_init()
	{
	print c, type_name(c);
	c = BIG;
	print c, type_name(c);
	}
```
prints
```
RED, Color
BIG, Size
```
with no complaint that the variable `c` has changed (enum) types. Worse, with ZAM it prints:
```
RED, Color
BLUE, Color
```
The reason is that ZAM pretty much has to rely on a variable's type being correct (since for low-level values, by using `ZVal`s it does away with them carrying their own types). It's darn hard to see how to fix this for ZAM in isolation. So it might be time to bite off fixing this in general.

The main headache of course is the various flavors of `Analyzer`, and the fact that it in essence has sub-types.

Thoughts on how to fix this, either for ZAM, or as a near-term kludge (e.g. special knowledge of `Analyzer`), or in a general, sound way? I played a bit with changing script-level to use `any` (thinking we could add run-time type-checking), but that fails for various sets/tables, which can't have `any` values as indices.

bbannier:
This seems to be a general shortcoming of user-defined types, and we also accept the following

```zeek
# foo.zeek
type X: record {
	foo: string &default="x";
};

type Y: record {
	foo: string &default="y";
};

event zeek_init()
	{
	local x = X();
[...]

PR #5149, opened 2026-01-14
Debugging info for ZAM instruction auxiliary information

vpax:
Some ZAM instructions include "auxiliary" information, a grab-bag of disparate values that don't fit into the general `ZInst` instruction design. When looking into what ultimately led to https://github.com/zeek/zeek/issues/1820, I found that I needed to see the particulars of these, so I extended the dumping of instructions to include this info.

### DIFF
--- a/src/script_opt/ZAM/ZInstAux.cc
+++ b/src/script_opt/ZAM/ZInstAux.cc
@@ -1,0 +1,162 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#include "zeek/script_opt/ZAM/ZInstAux.h"
+
+#include "zeek/module_util.h"
+
+namespace zeek::detail {
+
+void ZInstAux::Dump(FILE* f) const {
+    if ( id_val )
+        printf(" id=%s", id_val->Name());
+
+    if ( call_expr )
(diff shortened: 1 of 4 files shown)


src/script_opt/ZAM/ZInstAux.cc
```diff
@@ -0,0 +1,162 @@
...
+
+#include "zeek/script_opt/ZAM/ZInstAux.h"
+
+#include "zeek/module_util.h"
+
+namespace zeek::detail {
+
+void ZInstAux::Dump(FILE* f) const {
+    if ( id_val )
+        printf(" id=%s", id_val->Name());
```
awelzel, line 11:
```suggestion
        fprintf(f, " id=%s", id_val->Name());
```

awelzel: Happy to fix this during the merge.

Not sure how I saw that :-)

awelzel (approved):

src/script_opt/ZAM/ZInstAux.cc
vpax: Whoops! Good find, thanks! Yes please fix during merge.

Issue #5151, opened 2026-01-14
Recent paraglob changes resulting in search failures on large pattern sets


timwoj:
This was reported by Ben Reardon at Corelight. Our recent change in libraries used for Paraglob introduced a bug where very large pattern sets would fail searching. The following script reproduces it:

```
event zeek_init()
	{
	const pattern_vec = vector("amber=*", "bridge=*", "cushion=*", "dazzle=*", "effort=*",
		"falcon=*", "glimpse=*", "harvest=*", "ignite=*", "jovial=*", "kitchen=*",
		"lantern=*", "mosaic=*", "nurture=*", "octopus=*", "prism=*", "quarrel=*",
		"ribbon=*", "shadow=*", "timber=*", "uplift=*", "violet=*", "whisper=*",
		"zenith=*", "anchor=*", "breeze=*", "copper=*", "dolphin=*", "ember=*",
		"forest=*", "garden=*", "harbor=*", "island=*", "jungle=*", "kindle=*",
		"meadow=*", "nebula=*", "origin=*", "puzzle=*", "quartz=*", "rustic=*",
		"spiral=*", "temple=*", "unique=*", "voyage=*", "wonder=*", "yellow=*",
		"archive=*", "blossom=*", "crystal=*"
	);

	const pattern_glob = paraglob_init(pattern_vec);

	print paraglob_match(pattern_glob, "amber=");
	print paraglob_match(pattern_glob, "voyage=");
	}
```

There's two problems here. First, the paraglob library isn't reporting when `paraglob_init` fails to insert a pattern into the set of patterns (or, in detail, the tree used by the aho corasick algorithm). Second, the limitation of the number of nodes in the tree is set to much too low of a value. The script above fails to properly add all of the patterns to the tree, but because it fails silently, the lines for matching continue. The first search string is found ok, but the second one fails because that entry was never added to the tree.

PR #5152, opened 2026-01-14
Fix issues with paraglob size limtations

timwoj:
Fixes https://github.com/zeek/zeek/issues/5151

This adds a new argument to `paraglob_init` that allows the caller to set the maximum size of the paraglob tree. This wasn't a configurable value and our previous default was too small (see https://github.com/zeek/paraglob/pull/28). The new default should be large enough for extremely large sets of patterns, but making it user-configurable is better. It also adds a btest for checking that the size limitation returns an error message now.


src/zeek.bif
```diff
@@ -1296,10 +1296,14 @@ function sha512_hash_finish%(handle: opaque of sha512%): string
 ##
 ## v: Vector of patterns to initialize the paraglob with.
 ##
+## max_tree_size: The maximum size of the paraglob tree structure when adding new
+##                elements. paraglob_init will return an error if this limit is reached.
+##                Defaults to 2048 nodes in the tree.
```
bbannier: Given that the only way to set up a paraglob from script land is `paraglob_init` which already receives all the patterns upfront (I couldn't find a function to add to the already compiled glob), isn't there a way we could compute the tree size for users, e.g., from the number of passed patterns? I am not sure users would have a better idea than us.

If we want users to come up with a `max_tree_sizes` we will also have to explain how "number of patterns" translates to "tree size". Overall I'd prefer a low-level paraglob API which allows thinking in number of patterns, and then have us derive a good default value wherever we can.

In the C++ API there is also the `Paraglob(size_t max_tree_size = 2048)` ctr (which contrary to its docs doesn't invoke `compile`), and users `add` an unknown number of patterns to only later. That's a harder to use API and if we cannot resize automatically it might be acceptable to expect users to come up with their own size estimate, but even that would be much, much easier to use if it was formulated in number of patterns instead of tree size.

timwoj: I dug around in the libaca code a bit further, and there's a comment that the limit should be roughly equivalent to the total length of all of the strings being inserted. That should be easy enough to calculate when building the tree prior to calling `aca_init`, and it would remove the need for this additional variable.

src/zeek.bif
```diff
@@ -1310,16 +1310,22 @@ function paraglob_init%(v: any%) : opaque of paraglob
 		}
 
 	std::vector<std::string> patterns;
+	size_t tree_size = 0;
 	VectorVal* vv = v->AsVectorVal();
 	for ( unsigned int i = 0; i < vv->Size(); ++i )
 		{
 		auto s = vv->StringAt(i);
 		patterns.emplace_back(reinterpret_cast<const char*>(s->Bytes()), s->Len());
+		tree_size += s->Len();
```
bbannier: Paranoia nit:


```suggestion
		tree_size += std::abs(s->Len());
```

Alternatively we could just hope this function never returns a negative value.

PR #5155, opened 2026-01-15
Type: Nominal (name-based) type checking for enums

### LINKED ISSUE #1820: Events that take EnumTypes as arguments allow other types

awelzel:
This PR adds nominal type checking for enum types. It's a language change, but as discussed in #1820 might be reasonable. The main hurdle I came across was that pure ListVals created from sets stored base type pointers only and cause runtime type check errors. This has been fixed by storing the concrete TypePtr instead.

Update: At this point this logs a deprecation warning at parse time that will become an error with 9.1.

Closes #1820

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -1408,9 +1408,6 @@ ListVal::ListVal(TypePtr t) : Val(make_intrusive<TypeList>(t ? (t->Tag() == TYPE
     tag = t ? t->Tag() : TYPE_ANY;
 }
 
-// Delegate to ListVal(TypePtr t) using base_type(t).
-ListVal::ListVal(TypeTag t) : ListVal(base_type(t)) {}
-
 ListVal::ListVal(TypeListPtr tl, std::vector<ValPtr> _vals) : Val(std::move(tl)) {
     tag = TYPE_ANY;
     vals = std::move(_vals);
--- a/src/Val.h
+++ b/src/Val.h
@@ -657,7 +657,7 @@ public:
     // Constructor used to build up a homogeneous list of values;
     // or, if 't' is TYPE_ANY, then a heterogeneous one whose type
     // is built up as values are appended.
-    explicit ListVal(TypeTag t);
+    explicit ListVal(TypeTag t) : ListVal(base_type(t)) {}
 
     // Constructor used to build up a homogeneous list of values;
     // or, if 't' is nil or base_type(TYPE_ANY), then a heterogeneous


src/CompHash.cc
```diff
@@ -356,18 +358,22 @@ bool CompositeHash::RecoverOneVal(const HashKey& hk, Type* t, ValPtr* pval, bool
                 case TYPE_LIST: {
                     int n;
                     hk.Read("list-size", n);
-                    auto tl = t->AsTypeList();
-                    auto lv = make_intrusive<ListVal>(TYPE_ANY);
+                    auto tl = IntrusivePtr{NewRef{}, t->AsTypeList()};
```
vpax: FWIW, these days I usually use `<cast_intrusive>` for things like this

src/CompHash.cc
```diff
@@ -105,22 +105,24 @@ std::unique_ptr<HashKey> CompositeHash::MakeHashKey(const Val& argv, bool type_c
 }
 
 ListValPtr CompositeHash::RecoverVals(const HashKey& hk) const {
-    auto l = make_intrusive<ListVal>(TYPE_ANY);
     const auto& tl = type->GetTypes();
 
     hk.ResetRead();
 
+    std::vector<ValPtr> vals;
```
vpax: While this is cleaner, it sounded from your comments elsewhere that it's needed, not just nice-to-have. I'm not seeing offhand what makes it now required, give the list is of `TYPE_ANY` ... ?

src/Type.cc
```diff
@@ -1948,12 +1949,30 @@ bool same_type(const Type& arg_t1, const Type& arg_t2, bool is_init, bool match_
...
+            // a .bif file and a .zeek file. Supervisor::ClusterRole and
+            // Broker::BackendType are examples.
+            //
+            // It'd be nice if enum types were singletons instead.
+            if ( t1->GetName() == t2->GetName() )
+                return true;
+
+            // If an EnumType has a parent and the type to compare
+            // against is that parent, they are compatible...
+            const auto* t1p = t1->AsEnumType()->GetParentType().get();
```
vpax, line 1964: Nit: do you need `.get()` here? I'd think you can just directly compare the intrusive pointers

src/Val.cc
```diff
@@ -1400,7 +1400,9 @@ ValPtr PatternVal::DoClone(CloneState* state) {
     return state->NewClone(this, make_intrusive<PatternVal>(re));
 }
 
-ListVal::ListVal(TypeTag t) : Val(make_intrusive<TypeList>(t == TYPE_ANY ? nullptr : base_type(t))) { tag = t; }
+ListVal::ListVal(TypePtr t) : Val(make_intrusive<TypeList>(t)) { tag = t ? t->Tag() : TYPE_ANY; }
```
vpax: This confused me a bit. It appears to be introducing the notion that `TYPE_ANY` lists are represented now by having a nil type as opposed to a type-tag of TYPE_ANY. But what if I create one of these with `t` set to `base_type(TYPE_ANY)`? Does that introduce to different ways of representing the same underlying notion?

src/CompHash.cc
awelzel: It shouldn't be needed. I'll ditch the commit so the change is more obvious/streamlined. Thanks.

src/CompHash.cc
awelzel: I'll look again, but hmm, hmm, `t` is a raw pointer and I cannot use `cast_intrusive<>`. But I'll probably just skip this for now.

src/Type.cc
awelzel: `t1` and `t2` are `Type*` and `GetParentType()` yields `const EnumValPtr&` and there's no matching `operator==` for these, but it works for the raw pointers. Probably missing some template magic in IntrusivePtr, so I'll keep as is (given we anyhow fiddle with the raw `t1` / `t2` further up.

src/Type.cc
```diff
@@ -1948,12 +1949,30 @@ bool same_type(const Type& arg_t1, const Type& arg_t2, bool is_init, bool match_
...
+            const auto* t2p = t2->AsEnumType()->GetParentType().get();
+
+            if ( t1p && t1p == t2 )
+                return true;
+
+            if ( t2p && t2p == t1 )
+                return true;
+
+            // ...otherwise not.
```
awelzel: We can actually do `zeek::reporter->DeprecationWarning()` here and return true. This way we won't break scripts with 8.2 and instead produce a warning for users to fix up any wrong enum usages and then make it an error with 9.1.

src/Val.cc
awelzel: Nice catch. This was solely based on the existing `t == TYPE_ANY ? nullptr : base_type(t)` logic, but I agree using the any base type would be better. I shall revisit.

awelzel: I've reworked it. Not sure I like the result more. It's the TypeList that wants a `nullptr` for non-pure / heterogeneous lists. This was easier when it was only about the tag.

NEWS
```diff
@@ -14,6 +14,15 @@ Breaking Changes
 - The minimum version requirement for ZeroMQ is now 4.3.0. This version has
   been released at the end of 2018 - a good 7 years ago.
 
+- Enums are now nominally typed. Zeek will produce warnings of the following form
+  if you have scripts where enum types are mixed.
+
+	Remove in v9.1. Mixing incompatible enum types color and city will become an error.
+
+  With version 9.1, these warnings will become errors. Previously, all enum types
+  were compatible with each other, resulting in quirks and making certain optimization
```
evantypanski: nit

```suggestion
  were compatible with each other, resulting in quirks and making certain optimizations
```

evantypanski (approved):
Well I'm a fan of this. The stronger checks seem like only a good thing to me.

The only weird part is the "parent enum" type, but I agree that it's probably the best way to do it without breaking previous behavior too much.

the ternaries in `ListVal(TypePtr t)` are certainly hard to follow, but I'll hold my nose for that one :)

src/Val.h
```diff
@@ -946,7 +951,11 @@ class TableVal final : public Val, public notifier::detail::Modifiable {
     ValPtr Remove(const detail::HashKey& k, bool* iterators_invalidated = nullptr);
 
     // Returns a ListVal representation of the table (which must be a set).
-    ListValPtr ToListVal(TypeTag t = TYPE_ANY) const;
+    [[deprecated("Remove in v9.1. Pass a TypePtr instead, using Type::nil for TYPE_ANY")]]
+    ListValPtr ToListVal(TypeTag t) const;
+
+    // Returns a ListVal representation of the table (which must be a set).
+    ListValPtr ToListVal(TypePtr t = Type::nil) const;
```
vpax: Can you remind me why this is `Type::nil` as opposed to just `nullptr`?

awelzel: No reason, but also no good intuition if it makes a difference. I'll switch to nullptr.

src/Val.h
```diff
@@ -657,7 +657,12 @@ class ListVal final : public Val {
     // Constructor used to build up a homogeneous list of values;
     // or, if 't' is TYPE_ANY, then a heterogeneous one whose type
     // is built up as values are appended.
-    explicit ListVal(TypeTag t);
+    explicit ListVal(TypeTag t) : ListVal(base_type(t)) {}
```
awelzel, line 660: FYI - I've pushed on more commit to inline this constructor as it just delegates anyhow.

PR #5156, opened 2026-01-15
Add packet source that uses LightPcapNg to be able to read pcapng files

timwoj:
This is a packet source for reading data from pcapng files (see https://datatracker.ietf.org/doc/html/draft-ietf-opsawg-pcapng). It supports reading packets, and handling the interface data from the files. It uses the https://github.com/Technica-Engineering/LightPcapNg library (MIT licensed) to read the blocks from the files. 

The only packet source feature not currently supported is BPF filtering with Zeek's `-f` argument. I have been testing with the file from https://weberblog.net/the-ultimate-pcap/. It's capable of fully reading that file and generates all of the expected logs from it. A simple btest is included.


J-Gras:
For a multi-interface PCAP, the current interface is not known to script-land, correct? How about making this info available?

timwoj:
> For a multi-interface PCAP, the current interface is not known to script-land, correct? How about making this info available?

That was in the prior version and was stripped out with all of the eventing. I can do it in a follow-up if desired, but I'd like to get this part landed with just the packet processing.

src/iosource/pcapng/Plugin.cc
```diff
@@ -0,0 +1,23 @@
...
+
+#include "zeek/iosource/Component.h"
+#include "zeek/iosource/pcapng/Source.h"
+
+namespace zeek::plugin::detail::Zeek_Pcapng {
+
+class Plugin : public plugin::Plugin {
+public:
+    plugin::Configuration Configure() override {
+        AddComponent(new iosource::PktSrcComponent("PcapngReader", "pcapng", iosource::PktSrcComponent::BOTH,
```
awelzel:
```suggestion
        AddComponent(new iosource::PktSrcComponent("PcapngReader", "pcapng", iosource::PktSrcComponent::TRACE,
```

Cannot do live. Currently this works but does nothing:
```
# touch lo
# zeek -i pcapng::lo

```

src/iosource/pcapng/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(LIGHT_USE_ZSTD OFF CACHE BOOL "" FORCE)
+set(LIGHT_USE_ZLIB OFF CACHE BOOL "" FORCE)
+
+add_subdirectory(auxil/LightPcapNg)
+
+zeek_add_plugin(
+    Zeek PcapNg
```
awelzel:
```suggestion
    Zeek Pcapng
```

Just because that's what's used for the plugin namespace and plugin name.

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,209 @@
...
+    return val;
+}
+
+static uint64_t pcapng_extract_uint64(uint8_t* opt_data, uint16_t opt_length) {
+    if ( ! validate_option_length(8, opt_length) )
+        return 0;
+
+    uint64_t val = pcapng_extract_uint32(opt_data, opt_length);
+    val << 32;
+    val |= pcapng_extract_uint32(opt_data + 2, opt_length);
```
awelzel: Should this be +4? Reading two times uint32.

```suggestion
    val |= pcapng_extract_uint32(opt_data + 4, opt_length);
```

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,209 @@
...
+}
+
+static uint16_t pcapng_extract_uint16(uint8_t* opt_data, uint16_t opt_length) {
+    if ( ! validate_option_length(2, opt_length) )
+        return 0;
+
+    uint16_t val = opt_data[0];
+    val << 8;
+    val |= opt_data[1];
+    return val;
```
awelzel, line 32: Do we technically need to take `endian_swap` into account here? Otherwise, for an endian swapped pcap we might read the bytes in the wrong order? This is reading in big endian / network order now, right? Or does LightPcapng swap them or am I off track?

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,209 @@
...
+        light_read_block(pd, &block, &endian_swap);
+        if ( ! block ) {
+            // If we get a nullptr back, we've run out of blocks and the file is done.
+            Close();
+            return false;
+        }
+
+        if ( block->type == LIGHT_INTERFACE_BLOCK )
+            ParseInterfaceBlock(block);
+        else if ( block->type == LIGHT_ENHANCED_PACKET_BLOCK ) {
```
awelzel, line 113: Seems "simple packet block" would be a good follow-up here, too, with implicit interface = 0.

awelzel: Another follow up would be to handle the SHB and call interfaces.clear().

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,209 @@
...
+
+    uint64_t ts = lepb->timestamp_high;
+    ts <<= 32;
+    ts += lepb->timestamp_low;
+
+    // The timestamp is the number of "units" of time since unix epoch. The units are based on the
+    // timestamp resolution from the interface.
+    uint32_t ts_res = 1e6;
+    if ( pb.interface < interfaces.size() )
+        ts_res = interfaces[pb.interface].ts_resolution;
```
awelzel, line 203: else weird?

src/iosource/Component.cc
```diff
@@ -55,6 +55,28 @@ void PktSrcComponent::DoDescribe(ODesc* d) const {
 
     d->Add(" ");
     d->Add(prefs);
+
+    std::string magics;
```
awelzel: This isn't actually used, is it?

src/iosource/pcapng/Source.cc
```diff
@@ -0,0 +1,209 @@
...
+            if ( current_pkt_block.interface >= interfaces.size() ) {
+                reporter->Weird("pcapng_invalid_interface_number", util::fmt("%d", current_pkt_block.interface));
+                light_free_block(block);
+                return false;
+            }
+
+            current_block = block;
+
+            break;
+        }
```
awelzel, line 131: Maybe a else if wit something like
```
DBG_LOG(DBG_PKTIO, "pcapng: ignoring block %x", ...)
```

awelzel (approved):
I think as a first stab this works. Can always iterate and polish.

I've got a few thoughts here and there, but 👍

src/iosource/pcapng/Source.cc
timwoj: This is handled in `ExtractNextPacket`.

src/iosource/pcapng/Source.cc
timwoj: > Seems "simple packet block" would be a good follow-up here, too, with implicit interface = 0.

It's also missing the original capture length (to set `origlen`), as well as the drop count option. We could set origlen to the captured packet length though.

src/iosource/pcapng/Source.cc
timwoj: LightPcapng will swap them if needed. The variable passed into `light_read_block` returns whether or not it was swapped.

src/iosource/pcapng/Source.cc
timwoj: I went ahead and added the SHB bit.

src/iosource/pcapng/Source.cc
awelzel: Great, thanks!

PR #5159, opened 2026-01-18
Fixes for event groups in standalone compilation

vpax:
It turned out that event groups were completely broken for compiling scripts to standalone-C++. In fixing that, I found that the `Func::AddBody` interfaces had grown baroque and somewhat confusing, so I streamlined them by making the primary interface in terms a of `Func::Body` object rather than disparate parameters. A further simplification was to change the `ScriptFunc` constructor used for script optimization to take a vector of `Func::Body`s rather than separate vectors.


vpax:
The CI failures are because I didn't (used to) have Node available in my environment. Here's a patch to `zeekjs` to fix the issue:
```
diff --git a/src/Plugin.cc b/src/Plugin.cc
index f582ed3..4bc4f93 100644
--- a/src/Plugin.cc
+++ b/src/Plugin.cc
@@ -309,10 +309,9 @@ bool Plugin::RegisterAsScriptFuncBody(zeek::EventHandlerPtr zeek_eh,
   zeek::FuncPtr func = zeek_eh->GetFunc();
   zeek::detail::StmtPtr stmt =
       zeek::make_intrusive<InvokeJsEventHandlerStmt>(zeek_eh.Ptr(), js_eh);
-  std::vector<zeek::detail::IDPtr> inits;  // ? What are inits?
 
   // 0 framesize
-  func->AddBody(stmt, inits, 0, priority);
+  func->AddBody({.stmts = std::move(stmt)}, {}, 0);
 
   return true;
 }
@@ -371,8 +370,7 @@ bool Plugin::RegisterJsHookHandler(const std::string& name,
 #endif
 
   zeek::detail::StmtPtr stmt = zeek::make_intrusive<InvokeJsHookHandlerStmt>(js_hh);
-  std::vector<zeek::detail::IDPtr> inits;  // ? What are inits?
-  func->AddBody(stmt, inits, 0, priority);
+  func->AddBody({.stmts = std::move(stmt)}, {}, 0);
 
 #if ZEEK_VERSION_NUMBER < 80000
   PLUGIN_DBG_LOG(plugin, "Added body to %s", func->Name());
```

vpax:
(and because I guess CI doesn't tolerate deprecations)

I forget the Git submodule foo needed to incorporate this change, so if Some Hero wants to do it, please proceed!

vpax:
I pushed some fixing for building with GCC on one of our Linux systems.

src/Func.cc
```diff
@@ -539,6 +540,30 @@ void ScriptFunc::SetCaptures(Frame* f) {
...
 }
 
+void ScriptFunc::AddBody(Func::Body new_body, const std::vector<IDPtr>& new_inits, size_t new_frame_size) {
+    if ( new_frame_size > frame_size )
+        frame_size = new_frame_size;
+
+    auto num_args = static_cast<size_t>(GetType()->Params()->NumFields());
+
+    if ( num_args > frame_size )
+        frame_size = num_args;
```
timwoj:
```suggestion
    auto num_args = static_cast<size_t>(GetType()->Params()->NumFields());
    frame_size = std::max({frame_size, new_frame_size, num_args});
```

vpax:
> This would have been easier to review if the deprecations/rework had been in a separate commit from the functional changes. Just a note for next time.

Thx will keep that in mind.

PR #5166, opened 2026-01-21
Correctly handle vector self assign-append

### LINKED ISSUE #5164: Vector self-append crashes Zeek
This script never terminates:
```
event zeek_init()
        {
        local v = vector(1, 3, 5);
        v += v;
        print v;
        }
```

I'd hope this can be fixed by special-casing both sides being the same object, it'd be a pity to address by always copying the RHS value first given the common-case efficiency loss.

bbannier:
When iterating over the source vector during vector `+=` we previously would not cache the size which lead to an infinite loop for self assign-append. By now caching the size we only append the elements originally present.

Closes #5164.

### DIFF
--- a/src/Val.cc
+++ b/src/Val.cc
@@ -3442,7 +3442,9 @@ bool VectorVal::AddTo(Val* val, bool /* is_first_init */) const {
 
     auto last_idx = v->Size();
 
-    for ( auto i = 0u; i < Size(); ++i )
+    const auto num_elements = Size();
+
+    for ( auto i = 0u; i < num_elements; ++i )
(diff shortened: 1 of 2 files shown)


bbannier:
> The `0u` made me realize we have `int TableVal::Size()` and `unsigned in VectorVal::Size()` - joy 🤷‍♂️

And to top it all off, neither has a fixed size so depending on how Zeek was compiled you might get a `[u]int32_t` or `[u]int64_t`, but there isn't really a way to query that from e.g., `zeek-config` (maybe some header Zeek provides? dunno).

PR #5173, opened 2026-01-23
Fix memory leak in TableIterInfo::BeginLoop

siavashta:
# Fix memory leak in TableIterInfo::BeginLoop

## Summary

Fixes a memory leak in `zeek::detail::TableIterInfo::BeginLoop()` where the `loop_vars` vector accumulates elements across multiple iterations, causing unbounded memory growth in production workloads.

## Problem

When `TableIterInfo` objects are reused (which ZAM does for performance), `BeginLoop()` is called multiple times on the same object. The `loop_vars` member vector was never cleared, causing each call to `push_back()` to append to an existing vector instead of starting fresh.

### Evidence from Heap Profiling

Heap profiling using jemalloc revealed significant memory growth attributed to this function. The following analysis compares two heap dumps taken during production:

**Command:**
```bash
jeprof --text --cum --base=heap.0002.+57MB.616MB.heap /opt/corelight/bin/zeek heap.0057.+173MB.733MB.heap | grep -v prof_backtrace
```

**Heap Dump Analysis:**
```
Total: 108.5 MB
     0.0   0.0%   0.0%    192.0 177.0% zeek::detail::TableIterInfo::BeginLoop
     0.0   0.0%   0.0%    191.5 176.6% std::vector::push_back (inline)
     0.0   0.0%   0.0%    183.0 168.7% std::_Vector_base::_M_allocate (inline)
     0.0   0.0%   0.0%    182.0 167.8% std::vector::emplace_back (inline)
     0.0   0.0%   0.0%    181.5 167.3% std::vector::_M_realloc_insert (inline)
     0.0   0.0%   0.0%    177.0 163.2% std::__new_allocator::allocate (inline)
     0.0   0.0%   0.0%    177.0 163.2% std::allocator::allocate (inline)
     0.0   0.0%   0.0%    177.0 163.2% std::allocator_traits::allocate (inline)
     0.0   0.0%   0.0%    176.4 162.7% zeek::Event::Dispatch
     0.0   0.0%   0.0%    176.4 162.7% zeek::EventHandler::Call
     0.0   0.0%   0.0%    176.4 162.7% zeek::EventMgr::Drain
     0.0   0.0%   0.0%    176.4 162.7% zeek::detail::CoalescedScriptFunc::Invoke
     0.0   0.0%   0.0%    176.4 162.7% zeek::detail::ScriptFunc::Invoke
     0.0   0.0%   0.0%    176.4 162.7% zeek::detail::ZBody::Exec
```

**Key Findings:**
- **192 MB** of memory growth was attributed to `BeginLoop()` → `push_back()` → vector reallocation
- This represents **177% of total growth** (108.5 MB net growth, but 192 MB cumulative from this function)
- The call stack shows: `ZBody::Exec` → `TableIterInfo::BeginLoop` → `push_back` → vector reallocation
- **99.7%** of `BeginLoop`'s allocations come from `push_back`, confirming unbounded vector growth

### Root Cause

The `loop_vars` vector is a member variable (`std::vector<ZVal*>`) that persists across `BeginLoop()` calls. When the same `TableIterInfo` object is reused (which ZAM does for performance in non-recursive functions):

**Without the fix:**
1. First iteration: `loop_vars = [ptr1, ptr2, ptr3]` (3 elements, ~24 bytes)
2. Second iteration: `loop_vars = [ptr1, ptr2, ptr3, ptr4, ptr5, ptr6]` (6 elements, ~48 bytes)
3. Third iteration: `loop_vars = [ptr1, ..., ptr9]` (9 elements, ~72 bytes)
4. After 1 million iterations: `loop_vars` contains 3 million pointers → **~24 MB per TableIterInfo object**
5. With multiple workers and frequent table iterations: **192 MB+ total leak**

**Why this happens:**
- `TableIterInfo` objects are reused for performance (see comment in code: "Helpful for supporting recursive functions")
- `BeginLoop()` uses `push_back()` which **appends** to the existing vector
- `Clear()` method resets `tbl_iter` and `tbl_end` but **forgets** to clear `loop_vars`
- The other `BeginLoop()` overload correctly uses `std::move()` to replace the vector, but this one doesn't

## Solution

Add `loop_vars.clear()` at the start of `BeginLoop()` to ensure the vector starts fresh for each iteration:

```cpp
void BeginLoop(TableValPtr _tv, ZVal* frame, ZInstAux* aux) {
    tv = std::move(_tv);

    // Clear loop_vars to prevent unbounded growth when TableIterInfo is reused
    loop_vars.clear();

    for ( auto lv : aux->loop_vars )
        // ... rest unchanged
}
```

This ensures each `BeginLoop()` call starts with an empty vector, preventing accumulation.

## Impact

- **Severity**: High - causes unbounded memory growth in production
- **Affected**: Any Zeek script that iterates over tables frequently 
- **Memory Growth Rate**: 
  - **~100 MB per hour per worker** in production deployments
  - **192 MB+ cumulative growth** observed in heap profiling analysis
  - Over 6 hours: workers grew from 32 GiB to 48 GiB (16 GiB total growth)
- **Cluster Impact**: With 56 workers, this translates to **~5.6 GB/hour** cluster-wide memory leak
- **Performance**: Minimal - `clear()` is O(1) and preserves vector capacity

## Testing

The fix was verified by:

1. **Heap Profiling:**
   - Before: `jeprof` analysis showed 192 MB growth attributed to `BeginLoop`
   - After: No memory growth from `BeginLoop` in subsequent heap dumps

2. **Production Deployment:**
   - Before fix: Workers showed steady growth (~100 MB/hour), reaching 48 GiB after 6 hours
   - After fix: Memory usage stabilized, workers maintain constant footprint
   - Extended monitoring (24+ hours) confirms no memory accumulation

3. **Performance:**
   - No performance regression observed
   - `clear()` is O(1) operation that preserves vector capacity
   - Overhead is negligible compared to the table iteration work

## Production Context

This bug was discovered while investigating memory leaks in production Zeek deployments running internal plugins. The plugins frequently iterate over connection tables using Zeek's `for (key in table)` syntax, which triggers ZAM table iteration code paths.

**Symptoms:**
- Worker processes showing **~100 MB memory growth per hour** per worker
- Memory growth correlated with table iteration frequency
- Heap profiling consistently pointing to `TableIterInfo::BeginLoop`
- Over a 6-hour period, workers accumulated **~16 GiB** of additional memory (from 32 GiB to 48 GiB)

**Production Memory Graphs:**

**Before Fix:**
- Workers showed steady, unbounded memory growth
- Memory increased from ~32 GiB to ~48 GiB over 6 hours
- Growth rate: approximately **100 MB per hour per worker**
- Multiple workers in a cluster compound the issue (e.g., 56 workers × 100 MB/hour = 5.6 GB/hour cluster-wide)
Screenshot 2026-01-23 at 12 35 14 PM

**After Fix:**
- Memory usage stabilized and remained constant
- Workers maintain steady memory footprint without unbounded growth
- No memory accumulation observed over extended periods
Screenshot 2026-01-23 at 12 35 58 PM

**Workaround:**
None - requires code fix. The leak affects any Zeek script that iterates over tables in loops, which is a common pattern.

## Related

This bug was discovered while investigating memory leaks in internal plugins that frequently iterate over connection tables. The fix has been verified in production deployments showing stable memory usage after the patch.

### DIFF
--- a/src/script_opt/ZAM/IterInfo.h
+++ b/src/script_opt/ZAM/IterInfo.h
@@ -46,7 +46,8 @@ public:
     void BeginLoop(TableValPtr _tv, ZVal* frame, ZInstAux* aux) {
         tv = std::move(_tv);
 
-        assert(loop_vars.empty());
+        // Clear loop_vars to prevent unbounded growth when TableIterInfo is reused
+        loop_vars.clear();
 
         for ( auto lv : aux->loop_vars )
             if ( lv < 0 )
@@ -108,7 +109,6 @@ public:
     void Clear() {
         tbl_iter = std::nullopt;
         tbl_end = std::nullopt;
-        loop_vars.clear();
     }
 
 private:


src/script_opt/ZAM/IterInfo.h
```diff
@@ -46,6 +46,9 @@ class TableIterInfo {
     void BeginLoop(TableValPtr _tv, ZVal* frame, ZInstAux* aux) {
         tv = std::move(_tv);
 
+        // Clear loop_vars to prevent unbounded growth when TableIterInfo is reused
+        loop_vars.clear();
```
evantypanski: this should probably live in `Clear()` instead, then add an `assert(loop_vars.empty())` here

timwoj:
I added this to the backport list for both 8.0.6 and 8.1.1, which are both targeted for mid next-week. We can delay them if this needs some extra time.

timwoj:
Also, can you rebase this on top of zeek/zeek's master? I'm not sure what your fork is based on, but some of the CI builds are failing because they're versions of operating systems that are unsupported by our CI host (FreeBSD 14.2 and macOS Sequoia)

src/script_opt/ZAM/IterInfo.h
siavashta: isn't safer to leave the clear() here? are we sure EndIter -> Clear is guaranteed after BeginLoop ?

evantypanski: If there's a case that `loop_vars` is not empty, we care, because that means it was not properly destructed. Do you witness a case that clearing in `Clear()` does not leave it as empty here?

I care because having the assertion means that we can test this in CI, essentially. Also, `Clear()` does other stuff, it could also end up as a logic bug if `Clear()` is not called.

Do you witness a scenario where the iterator is reused without `Clear()` being called? I would classify that as a bug.

siavashta: ok. i 've updated the mr with your approach.
> you witness a case that clearing in Clear() does not leave it as empty here?
> Do you witness a scenario where the iterator is reused without Clear() being called?

not yet. I’m going to bake this change in my environment for a few hours to see if any bad memory behaviour shows up, and I’ll let you know what I find.

evantypanski: awesome thanks!

I started the ZAM tasks in CI, let me know if anything pops up with the memory usage. Locally the tests all seemed fine with the assertion

evantypanski:
Welp, I'm not sure where it gets reused in the failing ZAM tests, but seems like `Clear()` isn't called for the `scripts.policy.frameworks.telemetry.prometheus` test

I think for the patch release just sticking with the initial commit would be good (my bad!) just to get this in without much more hassle.

In the future I'm worried because the `tbl_iter` and `tbl_end` could be sticking around after iteration is done, then there's this comment:

```
// Class for iterating over the elements of a table.  Requires some care
// because the dictionary iterators need to be destructed when done.
```

That would likely be a question for @vpax later, though. It might be nothing.

vpax:
I'll look at this when I'm back from PTO later next week.

PR #5174, opened 2026-01-23
Fixing findings from new clang-tidy modernize-avoid-c-style-cast checker

timwoj:
clang-tidy recently added a `modernize-avoid-c-style-cast` checker, which is basically a much higher-powered version of the `cppcoreguidelines-pro-type-cstyle-cast` checker. It finds a lot more that weren't covered when we fixed those (https://github.com/zeek/zeek/pull/4615). This PR fixes what should be all of the rest of the C-style casts across the code base. It includes fixes for generated code from both binpac and gen-zam, as well as an additional suppression in bifcl. There's a few casts that were removed outright because they weren't necessary.

If it's easier to break this up in to multiple parts to make it not so painful to review, let me know. I realize it's a lot of changes in a lot of files and most of them are very repetitive.

### DIFF
--- a/src/Hash.cc
+++ b/src/Hash.cc
@@ -276,7 +276,7 @@ void HashKey::Set(zeek_int_t bi) {
 }
 
 void HashKey::Set(zeek_uint_t bu) {
-    key_u.bi = static_cast<zeek_int_t>(bu);
+    key_u.bu = static_cast<zeek_int_t>(bu);
     key = reinterpret_cast<char*>(&key_u);
     size = write_size = sizeof(bu);
 }
--- a/src/Hash.h
+++ b/src/Hash.h
@@ -377,6 +377,7 @@ protected:
         bool b;
         int i;
         zeek_int_t bi;
+        zeek_uint_t bu;
         uint32_t u32;
         double d;
         const void* p;


tools/binpac/lib/binpac_bytestring.h
```diff
@@ -38,7 +38,8 @@ class const_datastring {
     bool operator==(const_datastring<T> const& s) {
         if ( length() != s.length() )
             return false;
-        return memcmp((const void*)begin(), (const void*)s.begin(), sizeof(T) * length()) == 0;
+        return memcmp(reinterpret_cast<const void*>(begin()), reinterpret_cast<const void*>(s.begin()),
+                      sizeof(T) * length()) == 0;
```
bbannier: Shouldn't the cast to `const void*` always be possible?

```suggestion
        return memcmp(begin(), s.begin(), sizeof(T) * length()) == 0;
```

tools/binpac/src/pac_regex.cc
```diff
@@ -13,9 +13,9 @@ const char* RegEx::kMatchPrefix = "MatchPrefix";
 string escape_char(const string& s) {
     char* buf = new char[s.length() * 2 + 1];
     int j = 0;
```
bbannier, line 15: Yikes, what a nasty function

src/Attr.cc
```diff
@@ -46,8 +46,8 @@ const char* attr_name(AttrTag t) {
 	};
     // clang-format on
 
-    if ( int(t) >= 0 && int(t) < NUM_ATTRS )
-        return attr_names[int(t)];
+    if ( static_cast<int>(t) >= 0 && static_cast<int>(t) < NUM_ATTRS )
+        return attr_names[static_cast<int>(t)];
     else
```
bbannier, line 51: I think the code was unnecessary and possibly wrong before:

- `NUM_ATTRS` is already from an enum over `uint8_t` so it is safe to use as an integer
- casting `t` to an `int` and then comparing against the `uint8_t` `NUM_ATTRS` mixes signed and unsigned types and the comparison will be done in `unsigned anyway` due to integer promotion.

I think all the cast here should just be removed.

There are more files touched where this pattern is used.

src/Debug.cc
```diff
@@ -442,15 +442,15 @@ void tokenize(const char* cstr, string& operation, vector<string>& arguments) {
     char delim = '\0';
     const string str(cstr);
 
-    for ( int i = 0; i < (signed int)str.length(); ++i ) {
-        while ( isspace((unsigned char)str[i]) )
+    for ( int i = 0; i < static_cast<int>(str.length()); ++i ) {
```
bbannier: Can we make `i` and `start` below `size_t`?

src/Debug.cc
```diff
@@ -442,15 +442,15 @@ void tokenize(const char* cstr, string& operation, vector<string>& arguments) {
...
+        while ( isspace(static_cast<unsigned char>(str[i])) )
```
bbannier: `isspace` takes an `int`, I believe this cast can go away.

src/Debug.cc
```diff
@@ -556,13 +556,13 @@ int dbg_execute_command(const char* cmd) {
          * http://tiswww.case.edu/php/chet/readline/history.html
          * suggests that it's safe to assume it's really const char*.
          */
-        add_history((char*)cmd);
+        add_history(const_cast<char*>(cmd));
```
bbannier: In my version of readline `add_history` takes a `const char*` and this cast shouldn't be needed.

src/Debug.cc
```diff
@@ -556,13 +556,13 @@ int dbg_execute_command(const char* cmd) {
...
         HISTORY_STATE* state = history_get_history_state();
-        state->entries[state->length - 1]->data = (histdata_t*)get_debug_cmd_info(cmd_code);
+        state->entries[state->length - 1]->data = reinterpret_cast<histdata_t>(get_debug_cmd_info(cmd_code));
```
bbannier: With my readline `histdata_t` is `void*` and the cast should work implicitly.

src/DebugCmds.cc
```diff
@@ -570,7 +571,9 @@ int dbg_cmd_list(DebugCmd cmd, const vector<string>& args) {
         pre_offset = 0;
     }
 
-    if ( (int)pre_offset + (int)g_debugger_state.last_loc.FirstLine() - (int)CENTER_IDX < 0 )
+    if ( static_cast<int>(pre_offset) + static_cast<int>(g_debugger_state.last_loc.FirstLine()) -
```
bbannier: The first two cast are from `int` to `int` 🤷

src/DebugLogger.cc
```diff
@@ -158,7 +158,7 @@ bool DebugLogger::CheckStreams(const std::set<std::string>& plugin_names) {
 }
 
 void DebugLogger::Log(DebugStream stream, const char* fmt, ...) {
-    Stream* g = &streams[int(stream)];
+    Stream* g = &streams[static_cast<int>(stream)];
```
bbannier: The `enum` value is over an `uint8_t` and should convert to a proper integer for indexing implicitly.

src/DebugLogger.h
```diff
@@ -78,19 +78,19 @@ class DebugLogger {
...
 
-    void PushIndent(DebugStream stream) { ++streams[int(stream)].indent; }
-    void PopIndent(DebugStream stream) { --streams[int(stream)].indent; }
+    void PushIndent(DebugStream stream) { ++streams[static_cast<int>(stream)].indent; }
+    void PopIndent(DebugStream stream) { --streams[static_cast<int>(stream)].indent; }
 
-    void EnableStream(DebugStream stream) { streams[int(stream)].enabled = true; }
-    void DisableStream(DebugStream stream) { streams[int(stream)].enabled = false; }
+    void EnableStream(DebugStream stream) { streams[static_cast<int>(stream)].enabled = true; }
+    void DisableStream(DebugStream stream) { streams[static_cast<int>(stream)].enabled = false; }
```
bbannier: Same for all of these.

src/DebugLogger.h
```diff
@@ -78,19 +78,19 @@ class DebugLogger {
...
 
     // Takes comma-separated list of stream prefixes.
     void EnableStreams(const char* streams);
 
     // Check the enabled streams for invalid ones.
     bool CheckStreams(const std::set<std::string>& plugin_names);
 
-    bool IsEnabled(DebugStream stream) const { return streams[int(stream)].enabled; }
+    bool IsEnabled(DebugStream stream) const { return streams[static_cast<int>(stream)].enabled; }
```
bbannier: And here.

src/Desc.cc
```diff
@@ -316,7 +316,7 @@ void ODesc::AddBytesRaw(const void* bytes, size_t n) {
         // The following casting contortions are necessary because
         // simply using &base[offset] generates complaints about
         // using a void* for pointer arithmetic.
-        memcpy((void*)&(reinterpret_cast<char*>(base))[offset], bytes, n);
+        memcpy(reinterpret_cast<void*>(&(reinterpret_cast<char*>(base))[offset]), bytes, n);
```
bbannier: Wonder if that comment still applies to today's compilers like it did when it was written in 2010. The following seems to work just fine for me:

```suggestion
        memcpy(&(reinterpret_cast<char*>(base))[offset], bytes, n);
```

src/Desc.h
```diff
@@ -141,7 +141,7 @@ class ODesc {
         // Don't clear offset, as we want to still support
         // subsequent calls to Len().
 
-        return byte_vec(t);
+        return reinterpret_cast<byte_vec>(const_cast<void*>(t));
```
bbannier, line 144: Yikes, should we introduce `const_byte_vec` to avoid code like this?

src/Dict.h
```diff
@@ -877,9 +877,9 @@ class Dictionary {
                     printf("%'10d \n", i);
                 else
                     printf("%'10d %1s %'10d %4d %4d 0x%08x 0x%016" PRIx64 "(%3d) %2d\n", i, (i <= remap_end ? "*" : ""),
-                           BucketByPosition(i), (int)table[i].distance, OffsetInClusterByPosition(i),
-                           uint(table[i].hash), FibHash(table[i].hash), (int)FibHash(table[i].hash) & 0xFF,
-                           (int)table[i].key_size);
+                           BucketByPosition(i), static_cast<int>(table[i].distance), OffsetInClusterByPosition(i),
+                           uint(table[i].hash), FibHash(table[i].hash), static_cast<int>(FibHash(table[i].hash) & 0xFF),
+                           static_cast<int>(table[i].key_size));
         }
```
bbannier: Could we please just adjust the format string (if at all needed) instead of doing all this casting?

src/Dict.h
```diff
@@ -934,12 +934,12 @@ class Dictionary {
...
         else {
-            char key = char(random() % 26) + 'A';
+            char key = static_cast<char>(random() % 26) + 'A';
             snprintf(key_file, 100, "%d.%d-%d.ckey", Length(), max_distance, key);
             std::ofstream f(key_file, std::ios::out | std::ios::trunc);
             for ( int idx = 0; idx < Capacity(); idx++ )
                 if ( ! table[idx].Empty() ) {
-                    std::string s((char*)table[idx].GetKey(), table[idx].key_size);
+                    std::string s(reinterpret_cast<char*>(table[idx].GetKey()), table[idx].key_size);
```
bbannier: Unneeded cast? A `string` can construct from the `const char*` returned by `GetKey` just fine.

src/Expr.cc
```diff
@@ -4160,7 +4160,7 @@ ValPtr CallExpr::Eval(Frame* f) const {
     // Check for that.
     if ( f ) {
         if ( trigger::Trigger* trigger = f->GetTrigger() ) {
-            if ( Val* v = trigger->Lookup((void*)this) ) {
+            if ( Val* v = trigger->Lookup(reinterpret_cast<const void*>(this)) ) {
```
bbannier: Seems the cast to `void*` should work implicitly.

src/Frag.cc
```diff
@@ -36,7 +36,7 @@ FragReassembler::FragReassembler(session::Manager* arg_s, const std::shared_ptr<
         proto_hdr_len = ip->HdrLen();
         proto_hdr = new u_char[64]; // max IP header + slop
         // Don't do a structure copy - need to pick up options, too.
-        memcpy((void*)proto_hdr, (const void*)ip4, proto_hdr_len);
+        memcpy(reinterpret_cast<void*>(proto_hdr), reinterpret_cast<const void*>(ip4), proto_hdr_len);
```
bbannier: Same for all changed casts in this file.

src/Frame.h
```diff
@@ -190,7 +190,7 @@ class Frame : public Obj {
 
     void SetCall(const CallExpr* arg_call) {
         call = arg_call;
-        SetTriggerAssoc((void*)call);
+        SetTriggerAssoc(reinterpret_cast<const void*>(call));
```
bbannier: Same.

src/Hash.cc
```diff
@@ -276,7 +276,7 @@ void HashKey::Set(zeek_int_t bi) {
 }
 
 void HashKey::Set(zeek_uint_t bu) {
-    key_u.bi = zeek_int_t(bu);
+    key_u.bi = static_cast<zeek_int_t>(bu);
```
bbannier: Pretty sure this always triggered and still triggers UB due to signed integer overflow for huge integer dict keys.

src/IP.cc
```diff
@@ -474,7 +474,7 @@ void IPv6_Hdr_Chain::Init(const struct ip6_hdr* ip6, uint64_t total_len, bool se
     uint8_t next_type = IPPROTO_IPV6;
     const u_char* hdrs = reinterpret_cast<const u_char*>(ip6);
 
-    if ( total_len < (int)sizeof(struct ip6_hdr) ) {
+    if ( total_len < static_cast<int>(sizeof(struct ip6_hdr)) ) {
```
bbannier: `total_len` is an `uint64_t` so this cast isn't only unnecessary, but also wrong.

src/IPAddr.cc
```diff
@@ -19,14 +19,14 @@ const IPAddr IPAddr::v6_unspecified = IPAddr();
 IPAddr::IPAddr(const String& s) { Init(s.CheckString()); }
 
 std::unique_ptr<detail::HashKey> IPAddr::MakeHashKey() const {
-    return std::make_unique<detail::HashKey>((void*)in6.s6_addr, sizeof(in6.s6_addr));
+    return std::make_unique<detail::HashKey>(reinterpret_cast<const void*>(in6.s6_addr), sizeof(in6.s6_addr));
```
bbannier: Should work implicitly.

src/IPAddr.cc
```diff
@@ -19,14 +19,14 @@ const IPAddr IPAddr::v6_unspecified = IPAddr();
...
 }
 
 static inline uint32_t bit_mask32(int bottom_bits) {
     if ( bottom_bits >= 32 )
         return 0xffffffff;
 
-    return (((uint32_t)1) << bottom_bits) - 1;
+    return ((static_cast<uint32_t>(1)) << bottom_bits) - 1;
```
bbannier: Nit: I'd suggest to perform the computation in `int` to make this easier on the eye:

```suggestion
    return static_cast<uint32_t>((1 << bottom_bits) - 1);
```

src/IPAddr.cc
```diff
@@ -133,12 +133,13 @@ std::string IPAddr::AsHexString() const {
...
+        snprintf(buf, sizeof(buf), "%08x", static_cast<uint32_t>(ntohl(*p)));
     }
     else {
         const uint32_t* p = reinterpret_cast<const uint32_t*>(in6.s6_addr);
-        snprintf(buf, sizeof(buf), "%08x%08x%08x%08x", (uint32_t)ntohl(p[0]), (uint32_t)ntohl(p[1]),
-                 (uint32_t)ntohl(p[2]), (uint32_t)ntohl(p[3]));
+        snprintf(buf, sizeof(buf), "%08x%08x%08x%08x", static_cast<uint32_t>(ntohl(p[0])),
+                 static_cast<uint32_t>(ntohl(p[1])), static_cast<uint32_t>(ntohl(p[2])),
+                 static_cast<uint32_t>(ntohl(p[3])));
     }
```
bbannier: Does this do some magic truncation? If not I'd suggest to tweak the format string if needed instead of casting.

src/NetVar.cc
```diff
@@ -269,17 +269,18 @@ void init_net_var() {
...
+    tcp_content_deliver_all_orig = static_cast<bool>(id::find_val("tcp_content_deliver_all_orig")->AsBool());
+    tcp_content_deliver_all_resp = static_cast<bool>(id::find_val("tcp_content_deliver_all_resp")->AsBool());
 
-    udp_content_deliver_all_orig = bool(id::find_val("udp_content_deliver_all_orig")->AsBool());
-    udp_content_deliver_all_resp = bool(id::find_val("udp_content_deliver_all_resp")->AsBool());
-    udp_content_delivery_ports_use_resp = bool(id::find_val("udp_content_delivery_ports_use_resp")->AsBool());
+    udp_content_deliver_all_orig = static_cast<bool>(id::find_val("udp_content_deliver_all_orig")->AsBool());
+    udp_content_deliver_all_resp = static_cast<bool>(id::find_val("udp_content_deliver_all_resp")->AsBool());
+    udp_content_delivery_ports_use_resp =
+        static_cast<bool>(id::find_val("udp_content_delivery_ports_use_resp")->AsBool());
```
bbannier: At least for me the cast is not needed since `ValPtr` can convert to `bool` (`explicit`, but should work here since that's the target type).

src/NetVar.cc
```diff
@@ -269,17 +269,18 @@ void init_net_var() {
...
 
     dns_session_timeout = id::find_val("dns_session_timeout")->AsInterval();
     rpc_timeout = id::find_val("rpc_timeout")->AsInterval();
 
-    watchdog_interval = int(id::find_val("watchdog_interval")->AsInterval());
+    watchdog_interval = static_cast<int>(id::find_val("watchdog_interval")->AsInterval());
```
bbannier, line 282: We probably should document to users that we use an interval of `0sec` for anything less than a second. Alternatively we could change `watchdog_interval` (in C++) to be a `double`.

src/PolicyFile.cc
```diff
@@ -157,7 +157,8 @@ bool PrintLines(const char* policy_filename, unsigned int start_line, unsigned i
         start_line = 1;
 
     if ( start_line > pf->lines.size() ) {
-        debug_msg("Line number %d out of range; %s has %d lines\n", start_line, policy_filename, int(pf->lines.size()));
+        debug_msg("Line number %d out of range; %s has %d lines\n", start_line, policy_filename,
+                  static_cast<int>(pf->lines.size()));
```
bbannier:
```suggestion
        debug_msg("Line number %d out of range; %s has %zu lines\n", start_line, policy_filename, pf->lines.size());
```

src/PrefixTable.cc
```diff
@@ -165,10 +165,10 @@ void* PrefixTable::GetNext(iterator* i) {
             i->Xrn = *(--i->Xsp);
 
         else
-            i->Xrn = (patricia_node_t*)nullptr;
+            i->Xrn = nullptr;
 
         if ( i->Xnode->prefix )
-            return (void*)i->Xnode->data;
+            return reinterpret_cast<void*>(i->Xnode->data);
```
bbannier: Implicit cast should be fine.

src/RE.cc
```diff
@@ -135,7 +135,7 @@ bool Specific_RE_Matcher::Compile(bool lazy) {
 }
 
 bool Specific_RE_Matcher::CompileSet(const string_list& set, const int_list& idx) {
-    if ( (size_t)set.length() != idx.size() )
+    if ( static_cast<size_t>(set.length()) != idx.size() )
```
bbannier: Nit: Preexisting, but I'd be easier with

```suggestion
    if ( set.length() != static_cast<int>(idx.size()) )
```

I worry more that somebody implements `string_list` `length=-1` with special meaning than us holding too many values in `idx`.

src/Reporter.cc
```diff
@@ -580,7 +580,8 @@ void Reporter::DoLog(const char* prefix, EventHandlerPtr event, FILE* out, Conne
     }
 
     if ( raise_event && event && via_events && ! in_error_handler ) {
-        auto vl_size = 1 + (bool)time + (bool)location + (bool)conn + (addl ? addl->length() : 0);
+        auto vl_size = 1 + static_cast<bool>(time) + static_cast<bool>(location) + static_cast<bool>(conn) +
+                       (addl ? addl->length() : 0);
```
bbannier: No need to cast values to their existing type.

```suggestion
        auto vl_size = 1 + time + location + static_cast<bool>(conn) + (addl ? addl->length() : 0);
```

src/SerializationFormat.cc
```diff
@@ -241,7 +241,7 @@ bool BinarySerializationFormat::Read(IPPrefix* prefix, const char* tag) {
 }
 
 bool BinarySerializationFormat::Read(struct in_addr* addr, const char* tag) {
-    uint32_t* bytes = (uint32_t*)&addr->s_addr;
+    uint32_t* bytes = reinterpret_cast<uint32_t*>(&addr->s_addr);
```
bbannier, line 245: I wonder whether this violates strict aliasing if `in_addr_t` is ever defined to anything other than an `uint32_t`. Maybe we should prefix this with

```cpp
    static_assert(sizeof(in_addr_t) == sizeof(uint32_t));
```

src/SmithWaterman.cc
```diff
@@ -12,7 +12,7 @@
 
 namespace zeek::detail {
 
-Substring::Substring(const Substring& bst) : String((const String&)bst), _num(), _new(bst._new) {
+Substring::Substring(const Substring& bst) : String(static_cast<const String&>(bst)), _num(), _new(bst._new) {
```
bbannier: `Substring` is a `String`, and I cannot see why this needs to manually cast.


```suggestion
Substring::Substring(const Substring& bst) : String(bst), _num(), _new(bst._new) {
```

src/Stmt.cc
```diff
@@ -24,7 +24,7 @@
 namespace zeek::detail {
 
 const char* stmt_name(StmtTag t) {
-    static const char* stmt_names[int(NUM_STMTS)] = {
+    static const char* stmt_names[static_cast<int>(NUM_STMTS)] = {
```
bbannier: None of the casts touched here should be needed.

src/Stmt.cc
```diff
@@ -202,7 +202,7 @@ void Stmt::DecrBPCount() {
 
 void Stmt::AddTag(ODesc* d) const {
     if ( d->IsBinary() )
-        d->Add(int(Tag()));
+        d->Add(static_cast<int>(Tag()));
```
bbannier: Also not this one.

src/Attr.cc
```diff
@@ -15,7 +15,7 @@ namespace zeek::detail {
 const char* attr_name(AttrTag t) {
     // Do not collapse the list.
     // clang-format off
-	static const char* attr_names[int(NUM_ATTRS)] = {
+	static const char* attr_names[static_cast<int>(NUM_ATTRS)] = {
```
bbannier:
```suggestion
	static const char* attr_names[NUM_ATTRS] = {
```

src/Attr.cc
```diff
@@ -46,8 +46,8 @@ const char* attr_name(AttrTag t) {
 	};
     // clang-format on
 
-    if ( int(t) >= 0 && int(t) < NUM_ATTRS )
-        return attr_names[int(t)];
+    if ( t >= 0 && t < NUM_ATTRS )
+        return attr_names[static_cast<int>(t)];
```
bbannier:
```suggestion
        return attr_names[t];
```

src/Frag.cc
```diff
@@ -159,7 +159,7 @@ void FragReassembler::Weird(const char* name) const {
 }
 
 void FragReassembler::Overlap(const u_char* b1, const u_char* b2, uint64_t n) {
-    if ( memcmp((const void*)b1, (const void*)b2, n) != 0 )
+    if ( memcmp(reinterpret_cast<const void*>(b1), reinterpret_cast<const void*>(b2), n) != 0 )
```
bbannier:
```suggestion
    if ( memcmp(b1, b2, n) != 0 )
```

src/NetVar.cc
```diff
@@ -269,17 +269,18 @@ void init_net_var() {
...
+    tcp_content_deliver_all_orig = id::find_val("tcp_content_deliver_all_orig")->AsBool();
+    tcp_content_deliver_all_resp = id::find_val("tcp_content_deliver_all_resp")->AsBool();
 
-    udp_content_deliver_all_orig = bool(id::find_val("udp_content_deliver_all_orig")->AsBool());
-    udp_content_deliver_all_resp = bool(id::find_val("udp_content_deliver_all_resp")->AsBool());
-    udp_content_delivery_ports_use_resp = bool(id::find_val("udp_content_delivery_ports_use_resp")->AsBool());
+    udp_content_deliver_all_orig = id::find_val("udp_content_deliver_all_orig")->AsBool();
+    udp_content_deliver_all_resp = id::find_val("udp_content_deliver_all_resp")->AsBool();
+    udp_content_delivery_ports_use_resp =
+        static_cast<bool>(id::find_val("udp_content_delivery_ports_use_resp")->AsBool());
```
bbannier:
```suggestion
    udp_content_delivery_ports_use_resp = id::find_val("udp_content_delivery_ports_use_resp")->AsBool();
```

src/zeek.bif
```diff
@@ -6126,7 +6126,7 @@ function modulo%(a: int, b: int%): int
 	%{
 	if ( b == 0 ) {
 		zeek::reporter->Error("modulo by zero");
-		return 0;
```
bbannier, line 6129: Is this change correct? We should return a `ValPtr` and since `ValPtr` doesn't have a ctr from e.g., `int` `0` just means `nullptr`. If you want to return an int `0` we'd need `return zeek::val_mgr->Int(i: 0);`

timwoj: RIght. The existing code returns a null, because there isn't an explicit conversion. `nullptr` is also used as an error case in a lot of the BIFs, since we're throwing an `Error` string immediately before. I guess the question here is whether modulo should return a zero on failure or an error.

bbannier: Sorry, I read the diff wrong, you were actually fixing the conversion from an `int` to a ptr 🤦. I think this is exactly how it was intended, resolving.

Emitting a non-fatal error is probably wrong since we then need to hope that the user never inspects the returned value (otherwise they might get a SEGFAULT). We do this in a lot of places though, so this needs a broader solution, maybe even a changed contract with bifcl so it could abort on nullptr return values.

src/Frag.cc
timwoj: I grepped and fixed a few others of these while I was at it.

PR #5175, opened 2026-01-23
Docs: render the version on the index page

ckreibich:
With the switch to showing the `current`-tagged docs by default we noticed that it's hard to identify which version one's looking at. It's already in the page title, but that's quite tucked away. This adds the version, already accessible from our VERSION parsing in conf.py, to the index page.

We should backport this into `release/8.1`, since that branch will receive the next `current` release.


timwoj:
I like it 👍🏼

PR #5179, opened 2026-01-26
misc: downgrade mkdir() failure from error to warning

### LINKED ISSUE #3595: mkdir() BiF generates errors on failure rather than warnings

mchennai:
## Summary
Downgrades mkdir() failure reporting from error to warning, since directory
creation failures are non-fatal for most scripts.

## Changes
- Replaced error reporting with warning in mkdir() BiF
- Added btest coverage

Fixes #3595

### DIFF
--- a/testing/btest/scripts/misc-mkdir-warning.zeek
+++ b/testing/btest/scripts/misc-mkdir-warning.zeek
@@ -1,0 +1,8 @@
+# @TEST-EXEC: zeek
+
+event zeek_init()
+    {
+    local ok = mkdir("/root/should_fail");
+    print ok;
+    }
+


evantypanski:
Hi, the issue was already closed a year ago. Was this just to add the test? I believe the test will not work either since it just invokes `zeek` without the input file. I'm just not sure if this was supposed to be more but was lost somewhere?

timwoj:
I'm going to close this. The original issue was already fixed, and the test is invalid anyways.

PR #5180, opened 2026-01-27
PktSrc: Remove spurious newline after listening on line

awelzel:
I've been looking a bit at the process output in a cluster, and this extra newline looks kind of out of place.

### DIFF
--- a/src/iosource/PktSrc.cc
+++ b/src/iosource/PktSrc.cc
@@ -59,7 +59,7 @@ void PktSrc::Opened(const Properties& arg_props) {
     }
 
     if ( props.is_live )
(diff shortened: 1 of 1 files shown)


awelzel:
> I almost like it standalone, though, especially since when it terminates it has a line before termination, like:

Oh, I was mostly coming from an "empty lines are strange in a world of log forwarding / log shipping", but on second thought, I'm also conflicted: Who knows if there's a script that matches on exactly this to detect startup or so. I'll close rather than merge. Sorry for the noise.

evantypanski:
[This comic](https://xkcd.com/1172/) comes to mind a lot lately! but reasonable enough

Issue #3347, opened 2023-10-01
Consider adding convenience workaround so sanitizer builds of Zeek can be run on macOS


bbannier:
The default clang compiler which ships with macOS does not support leak sanitizer. This means that users who want to run a sanitizer build of Zeek need to explicitly set `ASAN_OPTIONS=detect_leaks=0` in their environment.

We could consider making this more streamlined. For developers running out of the build directory we could inject above `ASAN_OPTIONS` if the developer environment doesn't already define `ASAN_OPTIONS`. For this to work for installations we would need to embed changed ASAN defaults into all binaries we install (since `-fsanitze=address` is viral), e.g., add the following to all binaries we ship once (ODR etc.):

```cpp
#ifdef __cplusplus
extern "C"
#endif
const char* __asan_default_options() { return "detect_leaks=0"; }
```

Until this is ship, anyone wishing to run binaries build with ASAN on macOS with a compiler which does not support LSAN should set in their environment (probably statically since the default behavior of clang is just broken there)

```console
export ASAN_OPTIONS=detect_leaks=0
```

awelzel:
> For this to work for installed binaries we would need to embed changed ASAN defaults into all binaries we install (since -fsanitze=address is viral), e.g., [add the following](https://github.com/google/sanitizers/wiki/AddressSanitizerFlags#run-time-flags) to all binaries we ship exactly once (ODR etc.):

But only for OSX, right?


> The default clang compiler which ships with macOS does not support leak sanitizer. This means that users who want to run a sanitizer build of Zeek need to explicitly set ASAN_OPTIONS=detect_leaks=0 in their environment.

I'm not fully following #3026. It reads as if @vpax  gets errors about unsupported LSAN when running `zeek` when building with address sanitizer and Spicy enabled, but then the errors go away when building with `--disable-spicy`. Or did I miss something here?

---

Mentioning the following bifcl/binpac hunks below as semi-related in that for those executables LSAN is disabled explicitly, otherwise they'd likely cause non-functional builds on OSX today, too. Background in #699.

The `__asan_default_options()` approach seems much nicer, though.

https://github.com/zeek/bifcl/blob/c7bf54c587439d3bcb16d53b0d77a702e48d2526/builtin-func.l#L319-L325

https://github.com/zeek/binpac/blob/84b730fdcc5b983c65c6226ec092aee66c486680/src/pac_main.cc#L219-L232

timwoj:
[...]

PR #5183, opened 2026-01-28
Disable LeakSanitizer on macOS by default

timwoj:
This is part of the discussion in #3347. It extends the existing disabling of LeakSanitizer on FreeBSD to include macOS by default, since LSan isn't implemented there and we're required to add `detect_leaks=0` to `ASAN_OPTIONS`. Instead of requiring that, just disable it entirely.

### DIFF
--- a/tools/bifcl/builtin-func.l
+++ b/tools/bifcl/builtin-func.l
@@ -307,8 +307,8 @@ void finish_alternative_mode() {
 #endif
 #endif
 
-// FreeBSD doesn't support LeakSanitizer
-#if defined(USING_ASAN) && ! defined(__FreeBSD__)
+// FreeBSD and macOS don't support LeakSanitizer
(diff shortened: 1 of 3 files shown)


tools/bifcl/builtin-func.l
```diff
@@ -308,7 +308,7 @@ void finish_alternative_mode() {
 #endif
 
 // FreeBSD doesn't support LeakSanitizer
```
0xxon: Perhaps update the comment here too :)

tools/binpac/src/pac_main.cc
```diff
@@ -200,7 +200,7 @@ void usage() {
 #endif
 
 // FreeBSD doesn't support LeakSanitizer
```
0xxon: Similar here, the comment should be updated

PR #5189, opened 2026-01-30
cluster/experimental: Add non-Broker backend support

awelzel:
There are users and packages (logschema ...) that rely on Experimental::cluster_started() and will not function otherwise.
Implement support for ZeroMQ with semantics of all nodes have seen Cluster::node_up() from all other nodes, report to the manager which then broadcasts to all other nodes.

This switches to use a single cluster_started() specific topic rather than going through broadcast_topics, because that's the better mechanism to use with global pub/sub visibility (and it works for Broker, too).

Technically there's no connection between nodes for pub/sub in ZeroMQ anymore, so the terminology is all a bit awkward, but I don't think this needs more structure/cleanup: That would come naturally if either we support global pub/sub for Broker, or remove support for it. I'm inclined to say that any other cluster backends is going to provide global pub/sub. Seems pretty unreasonable not to. The NATS backend supports it, for example.


J-Gras:
> This switches to use a single cluster_started() specific topic rather than going through broadcast_topics, because that's the better mechanism to use with global pub/sub visibility (and it works for Broker, too).

FYI (https://github.com/zeek/zeek/pull/2823/changes/c7626d797f26ebc10537c507d14a949844d59d9c):
https://github.com/zeek/zeek/blob/c7626d797f26ebc10537c507d14a949844d59d9c/scripts/base/frameworks/cluster/main.zeek#L47-L52

awelzel:
> because enabling implicit Broker forwarding would 
> cause a routing loop for this topic.

This happens only in combination with `Broker::forward_messages=T` and the explicit `Broker::forward("topic")` calls IIUC? I'm happy to put a `Reproter::fatal()` into the new `zeek_init()` if this is enabled. We should start calling things old cruft and "don't use it" when it prevents us from doing something better/easier.

EDIT: I was actually tempted to put `&deprecated` on `broadcast_topics`. It's not a good pattern.

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -41,8 +41,13 @@ event zeek_init() &priority=-15
 	{
 	# Make sure the manager recognizes itself as ready if no
 	# connections have to be initiated.
+	#
+	# This is only needed for Broker.
+	if ( Cluster::backend != Cluster::CLUSTER_BACKEND_BROKER )
```
J-Gras, line 48: How about using `@if` directives? Imho, would be easier to read.

Also, why is that only required for broker? Because we assume full mesh connectivity for other pub-sub-backends?

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -51,6 +56,10 @@ event Cluster::node_up(name: string, id: string)
 	# report successful setup. As they do not establish connections
 	# we can consider this case here.
 
+	# This is only needed for Broker.
```
J-Gras: Same as above ☝️

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -48,13 +60,28 @@ global is_cluster_started = F;
...
+	# works for Broker because the it is the manager that sends out the
+	# cluster_started() event, so all other nodes will receive it.
+	Cluster::subscribe(cluster_started_topic);
+	}
+
 event Cluster::node_up(name: string, id: string) &priority=-10
 	{
+	# Do not do this for non-Broker backends. This is all Broker specific
+	# and the code won't be easier trying to combine them.
+	if ( Cluster::backend != Cluster::CLUSTER_BACKEND_BROKER )
```
J-Gras, line 84: Same here, can't we use `@if` directives for readability?

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -84,4 +111,53 @@ event Cluster::Experimental::cluster_started()
...
+# among nodes anymore, they all connect to a central XPUB/XSUB broker.
+#
+# The manager code handling node_fully_connected() is in nodes-experimental/manager.zeek.
+#
+# When any node in the cluster restarts and a new Cluster::node_Up() is raised,
+# all nodes send node_fully_connected() to the manager again. They set the resending
+# flag to ``T`` if they've previously observed cluster_started(). The manager will
+# not raise another cluster_started() event when it observes the resending
+# flag as ``T``.
+global node_ups_pending: set[string];
```
J-Gras: Why don't we use `connectees_pending`? Exact same purpose...

testing/btest/scripts/policy/frameworks/cluster/cluster_started.zeek
```diff
@@ -13,19 +13,19 @@
...
+# @TEST-EXEC: TEST_DIFF_CANONIFIER=sort btest-diff manager/.stdout
 
 @load policy/frameworks/cluster/experimental
 
 redef Log::default_rotation_interval = 0secs;
 
 event Cluster::Experimental::node_fully_connected(name: string, id: string, resending: bool)
 	{
-	print "node fully connected";
+	print "A node fully connected", name, resending;
```
J-Gras: Suggest to use `A)`, because `A` in `A node` reads like an article.

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
awelzel: > How about using @if directives? Imho, would be easier to read.

The main reason is that this doesn't work if this script is loaded before the `redef` before `Cluster::backend`. I don't think we have a good pattern to deal with this. The dynamic `Cluster::backend` check is also used in a few other places, so it's a somewhat established pattern. It's also not relevant to performance at all.

> Because we assume full mesh connectivity for other pub-sub-backends?

I'd more say "we can assume global pub/sub visibility". Script writers should not concern themselves with connectivity aspects.  That moves discussions on a higher-level.

scripts/policy/frameworks/cluster/experimental.zeek
awelzel: > Why don't we use connectees_pending?

It's not a good name. There's no concept of "connectees" for ZeroMQ or NATS.

Main motivation is also that all the lines before this part are Broker-specific and could be yanked in the future.

awelzel (commented):
@J-Gras - I believe we have somewhat different opinions how this code should work going forward. I'm going to do the suggested test update, but otherwise feel free to pick this up.

testing/btest/scripts/policy/frameworks/cluster/cluster_started.zeek
awelzel: Thanks, makes sense.

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
J-Gras: >The main reason is that this doesn't work if this script is loaded before the redef before Cluster::backend.

Ah, I see, `Cluster::backend` is "too dynamic" to go with `@if`s.

> I'd more say "we can assume global pub/sub visibility".

To me the fact that each node is able to reach each other node is a key property, the updated logic builds upon. Imho, the global reachability requirement should be explicitly mentioned (e.g., in the comment blob).

scripts/policy/frameworks/cluster/experimental.zeek
J-Gras: I don't buy the naming argument. In that case `node_fully_connected` should be renamed too (keeping the old for backwards-compatibility).

J-Gras:
> @J-Gras - I believe we have somewhat different opinions how this code should work going forward. I'm going to do the suggested test update, but otherwise feel free to pick this up.

@awelzel, no difference wrt. how that should work. Just a matter of documentation imho.

I'd like too see this landing asap and also being backported to 8.0.

scripts/policy/frameworks/cluster/experimental.zeek
awelzel: Fair. I'll reuse it. And keep the event name.

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
awelzel: > the fact that each node is able to reach each other node is a key property

Okay, I made some more comments here and there around this

In guess in my little box, this is just how it should've always worked as it's much easier to think and reason about when writing scripts. I basically don't bother to mention it. Broker not having this property feels more like the special case.

scripts/policy/frameworks/cluster/experimental.zeek
```diff
@@ -112,44 +114,49 @@ event Cluster::Experimental::cluster_started()
...
+# reasons. Connections between nodes in a cluster are an implementation detail, however.
+# Indeed, with ZeroMQ there's no direct connection between nodes for pub/sub functionality
+# anymore (there is direct push/pull connections for logging, but that's separate).
+# Instead, they all connect to a central XPUB/XSUB broker.
 #
 # The manager code handling node_fully_connected() is in nodes-experimental/manager.zeek.
 #
-# When any node in the cluster restarts and a new Cluster::node_Up() is raised,
-# all nodes send node_fully_connected() to the manager again. They set the resending
+# When any node in the cluster restarts and Cluster::node_up() events raised, all
```
J-Gras:
```suggestion
# When any node in the cluster restarts and Cluster::node_up() events are raised, all
```

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -42,7 +42,9 @@ event zeek_init() &priority=-15
 	# Make sure the manager recognizes itself as ready if no
 	# connections have to be initiated.
 	#
-	# This is only needed for Broker.
+	# This is only needed for Broker. With non-Broker backends, nodes
+	# see Cluster::node_up() from all other nodes in a cluster due to
+	# global pub/sub visibility.
```
J-Gras:
```suggestion
	# global pub/sub reachability.
```

scripts/policy/frameworks/cluster/nodes-experimental/manager.zeek
```diff
@@ -56,7 +58,9 @@ event Cluster::node_up(name: string, id: string)
 	# report successful setup. As they do not establish connections
 	# we can consider this case here.
 
-	# This is only needed for Broker.
+	# This is only needed for Broker. With non-Broker backends, nodes
+	# see Cluster::node_up() from all other nodes in a cluster due to
+	# global pub/sub visibility.
```
J-Gras:
```suggestion
	# global pub/sub reachability.
```

J-Gras:
Can we get this backported to Zeek 8.0.X? Thanks!

PR #5190, opened 2026-02-01
Docs: add script profiling (--profile-scripts) to troubleshooting guide

### LINKED ISSUE #4897: Script profiling is undocumented
Other than a short paragraph about `--profile-scripts` in NEWS I don't see docs anywhere, not even a wiki page. Would be great as part of a much broader performance troubleshooting guide.

LDeinkumo:
Problem
Script profiling is currently only mentioned briefly in release notes and lacks user-facing documentation.

What changed
Added a concise “Script profiling (--profile-scripts)” subsection to the troubleshooting documentation describing what the feature does, how to enable it, and the purpose of the output.

Scope
This PR intentionally keeps the change minimal and does not attempt to introduce a broader performance troubleshooting guide.

Related issue
Fixes #4897


evantypanski (changes requested):
Hey, thanks for the contribution.

I don't like this, though. It's very surface level and just explains what a user would intuit from the one-line explanation from `zeek --help`.

Furthermore, it's kind of incorrect and vague. `zeek --profile-scripts` does not automatically dump to a file, you must give the filename. That filename is TSV format with particular fields. Giving a brief overview of those fields would be useful to someone looking for a way to profile their scripts, so they know if that option will help them. This does not provide that. I imagine that's what the original issue had in mind, even without a general troubleshooting guide.

There is also no mention of `--profile-script-call-stacks`.

I wouldn't want a broader performance and troubleshooting guide, but I also don't think this accurately documents the feature.

Minimally, I would want what is found in the `NEWS` file mentioned in the issue, preferably with examples.

https://github.com/zeek/zeek/blob/37933d53b70fbbc36ca18dc4470cb22849575928/NEWS#L3157-L3168

Do you use this feature yourself?

LDeinkumo:
Thanks for the detailed feedback.

It’s very helpful, and you’re right on the points you raised.

My initial draft was intentionally conservative, but I agree it ended up
being too surface-level and missed some important details of how the
feature actually behaves. In particular, you’re correct that
--profile-scripts requires an explicit output filename and that the output
is a TSV with defined fields, which I should document more concretely.

I also missed --profile-script-call-stacks, which should clearly be
mentioned alongside the primary option.

I don’t regularly use this feature yet, but I’m happy to revise the
documentation to be more accurate and useful. My proposed update would:

     - explicitly document the required output filename,

     - describe the TSV output at a high level (including the main fields
and what they represent), and

     - mention --profile-script-call-stacks and how it complements script
profiling.

I’ll keep the scope limited to accurately documenting the feature rather
than expanding into a broader performance guide.

If that approach sounds reasonable, I can update the PR accordingly.

Regards.
@LDeinkumo

evantypanski:
Sure, but if you will continue, please run `zeek` with the option and see how it works, along with call stacks. Examples would be worthwhile here. It is also necessary to use the feature in order to teach how to use the feature.

And in case it doesn't go without saying, if you are using AI, please disclose that usage. I don't think that Zeek would do well if documentation in particular has obvious AI artifacts, as that erodes trust greatly. We absolutely do not need hallucinations in the documentation.

LDeinkumo:
I understand your point regarding actually running it and observing the
output, including --profile-script-call-stacks, rather than inferring
behaviour from flags or source references. I’ll work on running Zeek
locally and base the documentation on the actual TSV output and observed
fields, with examples.

Regarding AI use, I sometimes use AI tools solely for assistive purposes. I
will base any proposed changes to the documentation on hands-on tests and
direct observations of Zeek's behaviour. I understand your concerns about
trust and the unacceptable hallucinatory behaviour you mentioned.

I’ll follow up with an updated PR once I’ve validated the feature
end-to-end and can document it accurately.

evantypanski:
I am going to close this since it has been a couple of weeks, if you come back with new content please open a new PR

PR #5192, opened 2026-02-02
Make README a symlink to README.md

evantypanski:
It's weird that we would maintain two different READMEs. I just made it a symlink to README.md

~(draft first to see if this breaks something weird, I had to symlink it to get the build to work since it expected its existence, which also happens in plugins)~


awelzel:
Thank you!

PR #5194, opened 2026-02-02

bbannier:
We also fix a bug in how we declared the C++ enum for the `AnalyzerType`
HILTI enum, and expanded tests for its behavior.

### DIFF
--- a/testing/btest/spicy/has-analyzer.spicy
+++ b/testing/btest/spicy/has-analyzer.spicy
@@ -13,7 +13,10 @@ assert ! zeek::has_analyzer("XXX");
 assert zeek::analyzer_type("HTTP") == zeek::AnalyzerType::Protocol;
 assert zeek::analyzer_type("SHA1") == zeek::AnalyzerType::File;
 assert zeek::analyzer_type("VLAN") == zeek::AnalyzerType::Packet;
+
+assert zeek::analyzer_type("XXX") == zeek::AnalyzerType::Undef;
 assert ! zeek::analyzer_type("XXX");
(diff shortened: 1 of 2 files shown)


testing/btest/spicy/has-analyzer.spicy
```diff
@@ -13,7 +13,7 @@ assert ! zeek::has_analyzer("XXX");
 assert zeek::analyzer_type("HTTP") == zeek::AnalyzerType::Protocol;
 assert zeek::analyzer_type("SHA1") == zeek::AnalyzerType::File;
 assert zeek::analyzer_type("VLAN") == zeek::AnalyzerType::Packet;
-assert ! zeek::analyzer_type("XXX");
+assert zeek::analyzer_type("XXX") == zeek::AnalyzerType::Undef;
```
bbannier, line 17: This seems wrong, and I suspect the issue is in how `AnalyzerType` is declared

```cpp
HILTI_RT_ENUM(AnalyzerType, Undef, File, Packet, Protocol);
```

`HILTI_RT_ENUM` should already inject the `Undef` value, so doing it twice seems to break things here (in the new impl the user-provided `Undef` would be a valid label).

I would have expected this to be

```cpp
HILTI_RT_ENUM(AnalyzerType, File, Packet, Protocol);
```

which fails with an error.

bbannier: This is fixed now, and I added additional tests.

auxil/spicy
bbannier, line 1: This currently points to the branch of zeek/spicy#2262, and needs to be updated once that one was merged.

PR #5196, opened 2026-02-02
Overall TOC reorg on top of the new tutorial

ckreibich:
@evantypanski and I have been discussing additional reconfiguration of the docs around the tutorial. The main goal here is a more manageable, less overwhelming overall TOC. To that end, this
* Introduces a single top-level reference section that features the logs, Zeekscript, its frameworks, and the auto-generated scripting reference
* Combines the previous "About Zeek" section with the monitoring one, in a new subdirectory
* Cleans up the grab-bag section currently in the new docs's advanced section.

You can check out the result of this revamp [here](https://docs.zeek.org/en/topic-christian-doc-toc-reorg/).

### DIFF
--- a/doc/frameworks/index.rst
+++ b/doc/frameworks/index.rst
@@ -3,7 +3,7 @@
  Scripting Frameworks
 ======================
 
-Zeek includes several software frameworks that provide commonly used
+Zeek includes several frameworks that provide commonly used
 functionality to the scripting layer. Among other things, these frameworks
 enhance Zeek’s ability to ingest data, structure and filter its outputs, adapt
 settings at runtime, and interact with other components in your network. Most


doc/frameworks/index.rst
```diff
@@ -1,7 +1,7 @@
 
-==========
-Frameworks
-==========
+======================
+ Scripting Frameworks
```
evantypanski, line 3: What's the purpose for renaming this to "Scripting frameworks"? (genuine question, not saying it should not, just curious about the reasoning here)

This page then refers to it in three ways: "Scripting frameworks," "software frameworks," and "frameworks." I think consolidating those into one "scripting frameworks" would be nice here if it gets renamed.

It also probably makes the first sentence make more sense, since now it'd read:

> Zeek includes several scripting frameworks that provide commonly used
> functionality to the scripting layer.

doc/reference/zeekscript/packages.rst
```diff
@@ -1,14 +1,14 @@
...
-==================
+============
+Module Index
+============
 
-Zeek has the following script packages (e.g. collections of related scripts in
-a common directory).  If the package directory contains a ``__load__.zeek``
-script, it supports being loaded in mass as a whole directory for convenience.
+Zeek ships with the following script modules, i.e. collections of related scripts in
+a common directory.
```
evantypanski, line 8: we need a Zeek glossary at this point. I say this only half jokingly

doc/reference/zeekscript/index.rst
```diff
@@ -0,0 +1,15 @@
+=====================
+ Zeekscript Language
+=====================
+
+The nuts and bolts of the Zeekscript language.
```
evantypanski: note that this rename is "problematic" (?) for a couple reasons

1) The [zeekscript](https://github.com/zeek/zeekscript) repo is now confusing, because it does not implement zeekscript the language
2) Other parts of the docs refer to the "Zeek scripting language" ([1](https://github.com/zeek/zeek/blob/12bd26e749cd3e040e66ab6b12a35b2331ed3ff3/doc/reference/zeekscript/statements.rst?plain=1#L4), [2](https://github.com/zeek/zeek/blob/12bd26e749cd3e040e66ab6b12a35b2331ed3ff3/doc/reference/zeekscript/directives.rst?plain=1#L4), [3](https://github.com/zeek/zeek/blob/12bd26e749cd3e040e66ab6b12a35b2331ed3ff3/doc/reference/zeekscript/types.rst?plain=1#L4), [4](https://github.com/zeek/zeek/blob/12bd26e749cd3e040e66ab6b12a35b2331ed3ff3/doc/reference/zeekscript/operators.rst?plain=1#L4), [5](https://github.com/zeek/zeek/blob/12bd26e749cd3e040e66ab6b12a35b2331ed3ff3/doc/reference/zeekscript/attributes.rst?plain=1#L6))

Not saying we shouldn't, or that it has to be fixed here, but figured I would point those out. Since we almost all colloquially refer to it as "zeekscript" it is probably best to just rebrand it a little

ckreibich: Definitely good catch re. the zeekscript repo, and I should have called this one out in the PR description — so thank you for flagging this one!  I'll stick to "Zeek scripting language" for the time being.

doc/frameworks/index.rst
ckreibich: Ha! I definitely dislike "software frameworks". I was simply looking for something that's a bit more descriptive than "frameworks", and since the unifying theme is that they all are effectively script-layer APIs, settled for "scripting frameworks". I don't feel strongly here at all.

doc/reference/zeekscript/packages.rst
ckreibich: It's totally true!

PR #5202, opened 2026-02-03
Add better documentation for prefix anonymization

timwoj:
This came out of a comment here: https://github.com/zeek/zeek/pull/5070#discussion_r2604256145, where I thought the anonymizer wasn't working correctly. It turns out it is working, it just wasn't documented well enough.


src/zeek.bif
awelzel: Do you think this can be dropped?

PR #5203, opened 2026-02-04
Restore running pre-commit during 'Generate Docs' workflow

timwoj:
https://github.com/zeek/zeek/commit/bcede2a1351d8e448eb91a23d3ea9fe0e3be68df removed running pre-commit as part of the "Generate Docs" workflow. I'm not entirely sure why I did that. The workflow is now complaining about not being able to cache the pre-commit setup because it never actually gets initialized. Restore running pre-commit so that the typos job gets run as part of the workflow at least.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -110,6 +110,9 @@ jobs:
 
           cd doc
 
+          echo "*** Running pre-commit ***"
+          pre-commit run -a --show-diff-on-failure --color=always
+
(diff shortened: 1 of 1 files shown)


timwoj:
Actually, thinking about this now that the workflow is running, maybe the better thing would be just to remove any reference to pre-commit from this workflow. The pre-commit workflow will always get run alongside it, and anything that would get caught in the docs workflow would get caught by that one instead...

timwoj:
Will replace with a version that removes pre-commit from the docs workflow.

PR #5209, opened 2026-02-06
Avoid ABI-breaking use of `DEBUG`

bbannier:


src/Dict.h
```diff
@@ -136,7 +132,6 @@ class DictEntry {
         key = nullptr;
         value = nullptr;
         key_size = 0;
-        bucket = 0;
```
bbannier, line 139: This ifdef block still feels weird, but I left it as is. This could still change runtime behavior, but at least the ABI should be well-defined now.

src/threading/MsgThread.cc
```diff
@@ -371,14 +369,10 @@ void MsgThread::InternalError(const char* msg) {
     abort();
 }
 
-#ifdef DEBUG
-
 void MsgThread::Debug(DebugStream stream, const char* msg) {
     SendOut(new detail::DebugMessage(stream, this, BuildMsgWithLocation(msg)));
```
timwoj, line 373: The implementation of this could be blocked out with the `ifdef`, leaving the function definition in the header safe to include all of the time. Then `DebugMessage` could also still be `ifdef`'d. I don't know how much this gets used though.

timwoj (approved):
One minor idea. Feel free to ignore if it doesn't make sense.

src/threading/MsgThread.cc
bbannier: I was briefly wondering about that as well just from transforming the code, but felt that was just asking for more problems e.g., during linking down the line. What would you want to achieve here? Avoiding increased binary size?

PR #5212, opened 2026-02-10
Fixes for initializations when compiling scripts to C++

vpax:
It turns out that the approach used in https://github.com/zeek/zeek/pull/5165 exposed some other issues related to initializing scripts that have been compiled to C++ via `-O gen-C++` or `-O gen-standalone-C++`:

- Needing to track functions instantiated from lambdas appearing in the initializers of globals (these concrete functions become "aliases" of the lambdas for script optimization purposes)
- A fix for initializing pattern tables
- A fix for globals whose _final_ initial value doesn't match their _initial_ value (from their own initializer) due to changes made to the global by _other_ initializations.

### DIFF
--- a/src/Var.cc
+++ b/src/Var.cc
@@ -112,24 +112,24 @@ static ExprPtr initialize_var(const IDPtr& id, InitClass c, ExprPtr init) {
             if ( ! IsAggr(t) )
                 return nullptr;
 
-            ValPtr init_val;
+            auto empty_list = make_intrusive<ListExpr>();
 
             if ( t->Tag() == TYPE_RECORD ) {
-                try {
-                    init_val = make_intrusive<RecordVal>(cast_intrusive<RecordType>(t));
-                } catch ( InterpreterException& ) {
-                    id->Error("initialization failed");
-                    return nullptr;
-                }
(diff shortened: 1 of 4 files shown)


src/script_opt/CPP/InitsInfo.h
```diff
@@ -399,12 +399,15 @@ class PatternConstInfo : public CPP_InitInfo {
...
+        ivs.emplace_back(std::to_string(exact_pat));
+        ivs.emplace_back(std::to_string(any_pat));
         ivs.emplace_back(std::to_string(is_case_insensitive));
         ivs.emplace_back(std::to_string(is_single_line));
     }
 
 private:
+    int exact_pat;           // string representation of "exact" pattern
+    int any_pat;             // same but for "any" pattern
     int pattern;             // index into string representation of pattern
```
timwoj, line 411: Is this variable unused now?

timwoj (changes requested):

src/script_opt/CPP/InitsInfo.h
vpax: Ah yes you are right. If you're game, plz remove when merging.

PR #5213, opened 2026-02-12
Update Spicy integration for new struct field optimization.

rsmmr:
This goes with https://github.com/zeek/spicy/pull/2249. 

Generally Spicy's new struct field optimization is safe in terms of
only removing fields that arent' used anywhere, including inside the
generated Zeek glue code. There's one exception however: when we pass
a struct instance to Zeek through `to_val()`, the Spicy optimizer
can't tell that all the fields are required on the receiving side
(because the Zeek-side layout is statically determined, either from
the Spicy-side type (if using EVT `export`), or by the Zeek user
creating the corresponding `record` manually). In other words, we need
to tell Spicy not to mess with structs that are passed wholesale
through `to_val()`. This change adds corresponding HILTI-side `export`
declarations for all types that go, directly or indirectly, through
`to_val()`. Internally that needs some care because we can determine
the relevant types only once our glue code has been created and fully
resolved.


rsmmr:
@bbannier With https://github.com/zeek/spicy/pull/2249 got to merge, this is ready for review as well.

Issue #4148, opened 2025-01-03
Coredump: 7.0.5 -  Broker::create_clone  [ zeek-7.0.5/auxil/broker/libbroker/broker/detail/flare.cc:107 ] 


initconf:
I can replicate in my environment but Still need to figure out a PoC. Once I do that I'll paste it here too.

Meanwhile - I see a coredump in `7.0.5` ( and `7.0.2`) - Here is the backtrace:

```
Core was generated by `/home/bro/zeek-7.0.5/bin/zeek -i myri0 -U .status -p zeekctl -p zeekctl-live -p'.
Program terminated with signal SIGSEGV, Segmentation fault.
Address not mapped to object.
#0  broker::detail::flare::fd (this=0x1d8) at /home/bro/install/zeek-7.0.5/auxil/broker/libbroker/broker/detail/flare.cc:107
107       return fds_[0];
[Current thread is 1 (LWP 102689)]
(gdb) bt
#0  broker::detail::flare::fd (this=0x1d8) at /home/bro/install/zeek-7.0.5/auxil/broker/libbroker/broker/detail/flare.cc:107
#1  0x0000000801ad9ed4 in broker::internal::flare_actor::descriptor (this=<optimized out>)
    at /home/bro/install/zeek-7.0.5/auxil/broker/libbroker/broker/internal/flare_actor.hh:47
#2  broker::mailbox::descriptor (this=<optimized out>) at /home/bro/install/zeek-7.0.5/auxil/broker/libbroker/broker/mailbox.cc:16
#3  0x0000000000ce2aa6 in zeek::Broker::Manager::MakeClone (this=0x803a2ea80, name="___sync_store_Scan::whitelist_subnet", resync_interval=<optimized out>,
    stale_interval=<optimized out>, mutation_buffer_interval=<optimized out>) at /home/bro/install/zeek-7.0.5/src/broker/Manager.cc:1655
#4  0x0000000000fbfb0d in zeek::BifFunc::Broker::__create_clone_bif (frame=<optimized out>, BiF_ARGS=<optimized out>)
    at /home/bro/install/zeek-7.0.5/src/broker/store.bif:65
#5  0x0000000000fe27ef in zeek::detail::BuiltinFunc::Invoke (this=0x8050f5400, args=<optimized out>, parent=<optimized out>)
    at /home/bro/install/zeek-7.0.5/src/Func.cc:739
[...]

PR #5214, opened 2026-02-13
Deprecate Broker store entry APIs

awelzel:
This PR deprecates Broker::create_master(), Broker::create_clone() and Cluster::create_store() for removal in v9.1. All of which are based on Broker stores. We don't have a really good story what to replace them with, but all reports I'm aware of using these at scale either cause severe performance issues, or [spurious crashes](https://github.com/zeek/zeek/issues/4148) that we have never been able to fully tracked down (fixed by reporting an error instead of crashing, but never solved the root cause IIUC).

Also, hard-error if these APIs are used, but the selected backend isn't Broker as to prevent silent non-functional behavior.

In addition, this PR removes the "use_x_store" options from the `known-*` scripts that have been defaulted to `F` since 2023 because performance issues when set to `T`. We probably should've deprecated these back then already.

@ckreibich - mind signing off on this, or would you prefer another way here?


timwoj:
At one point I added support for the storage framework to `known-services` as an initial usage example. Would it makes sense to do the same for the other `known` policies if we're going to remove the Broker APIs?

awelzel:
> At one point I added support for the storage framework to known-services as an initial usage example. Would it makes sense to do the same for the other known policies if we're going to remove the Broker APIs?

Maybe? I've not really worked on the known scripts, but my understanding is they never dealt with actual persistence across restarts and always used ephemeral (memory) Broker stores. Seems like a separate feature to me.

0xxon:
Should the deprecation warnings perhaps indicate what to use instead? So - perhaps point to the storage framework.

awelzel:
I'm closing this for now until there's a new primitive in the tree that can be linked to. Will keep the branch, because deprecating the entry points to broker stores seems reasonable to do pre-9.0.

PR #5215, opened 2026-02-13
Use \u00xx escaping for all non-UTF8 bytes in JSON logs

awelzel:
...and add a section about the \u00xx meaning in the JSON logs and why this might be better than the old \xXX encoding.

---

@timwoj / @ckreibich - this is what I had in mind. I find it controversial, but I also find it better what we have today. See the docs entry for details. Needs a bit of a NEWS entry if we do it.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -2019,8 +2019,8 @@ std::string escape_as_u00xx_utf8_json_string(std::string_view sv) {
             idx += 1;
         }
         else {
-            // High-bit set!
-            assert(ch & 0x80);
+            // ASCII-control character or high-bit 0x80-0xff bytes
+            // are all encoded as \u00xx.
             unsigned int char_size = getNumBytesForUTF8(ch);
 
             if ( idx + char_size > sv.size() || ! check_ok_utf8(sv_data + idx, sv_data + idx + char_size) ) {


0xxon:
After looking at the json spec - I don't think I like this.

As the NEWS article already shows - our definition of what "\u00XX" means will be different than the interpretation of typical JSON libraries, which will interpret it to be a specific unicode code point. I kind of feel like we should just fix the escaping of our \x encoding, to fix the disambiguity problem. Doing the "\u00XX" solution also means that, in the future, we cannot use the standard notation to encode unicode code points, if we so desire. 

If we do go this way, I think we should have a toggle for this, as it is a major change.

awelzel:
>  I kind of feel like we should just fix the escaping of our \x encoding

Do you have a better fix for this that's not lossy? I'd be okay explaining the `\u00xx` - the proposed paragraph is for the docs, not NEWS.

> If we do go this way, I think we should have a toggle for this, as it is a major change.

[Agree with the policy toggle](https://github.com/zeek/zeek/issues/3948#issuecomment-3318634158), but it should also include (or we revert) the `\x07` to `\u0007` behavior that we have in `master` now.

> I guess whatever we do, it'd probably be good to opt-in or have some sort of policy on the LogAscii module. Keeping what we do today as a fallback available.

0xxon:
Looking through GH-3984 for more context - I think I like the approach that @timwoj proposed in https://github.com/zeek/zeek/issues/3948#issuecomment-3115898482 best.

So - just encoding these cases as (invalid) half-surrogate pairs, with `\udcxx`. That way we don't paint ourselves into a corner where we might want to use `\u00xx` in the future to encode actual unicode code points.

awelzel:
> So - just encoding these cases as (invalid) half-surrogate pairs, with \udcxx.

I think you mentioned this, but `serde_json` and possibly other parser will bail on this. So need to prepend it with a high-surrogate to make it valid, e.g `\udb80` and make sure to map into private range. E.g. `\udb80\udcxx`.

EDIT: `\udb80\udcxx` still needs documentation, pretty custom code to get the original byte value and the raw logs look less nice. So yeah, technically more correct, but :man_shrugging:

I'm closing this. After a team chat seems most do not like the `\u00xx` encoding as an option :wave:

PR #5219, opened 2026-02-17
analyzer/detect-protocols: Honor analyzer_failed()

### LINKED ISSUE #5204: DPD incorrectly flagging HTTP ( eg . when the traffic SSH )
Both `ProtocolDetector::Server_Found` and `ProtocolDetector::Protocol_Found` are incorrectly flagging HTTP when the traffic is going to server running SSH.

```
1770126426.041811       Cq3yM13DnU3vya4Fda      136.216.42.40   44338   130.74.164.231  22      -       -       -       tcp     ProtocolDetector::Protocol_Found        136.216.42.40:44338 > 130.74.164.231:22 HTTP on port 22/tcp     HTTP    136.216.42.40   130.74.164.231  22      -       -       Notice::ACTION_LOG      (empty) 3600.000000     -       -       -       -       -
1770126426.041811       Cq3yM13DnU3vya4Fda      136.216.42.40   44338   130.74.164.231  22      -       -       -       tcp     ProtocolDetector::Server_Found  130.74.164.231: HTTP server on port 22/tcp      HTTP    130.74.164.231  130.74.164.231  22      -       -       Notice::ACTION_LOG      (empty) 3600.000000     -       -       -       -       -
```

```
$ cat dpd.log  
1770126425.898127       Cq3yM13DnU3vya4Fda      136.216.42.40   44338   130.74.164.231  22      tcp     HTTP    not a http reply line
```

`ProtocolDetector::Server_Found  130.74.164.231: HTTP server on port 22/tcp` when in-fact there is no HTTP running on 22/tcp on 130.74.164.231

Rather here is ssh logs :

```
1770126426.269476       COjWXe3laZTK75Pm2e      139.59.137.158  44350   128.3.50.246    22      -       -       0       -       -       SSH-2.0-OpenSSH_8.7     -       -       -       -       -       -
1770162948.551554       C5vJex3RrlWY4BL811      73.235.162.175  63590   128.3.50.246    22      2       -       0       -       SSH-2.0-OpenSSH_10.0    SSH-2.0-OpenSSH_8.7     aes128-gcm@openssh.com  hmac-sha2-256-etm@openssh.com   none    ecdh-sha2-nistp256      ssh-ed25519     e2:5c:0f:3d:3f:1c:c5:3f:68:b4:84:9a:be:30:33:f8
```

So we'd like to make sure that both `Server_Found` and `Protocol_Found` don't have obvious false positives. Infact wireshark flags this correctly as SSH. 

Here is a pcap and how to replicate the issue.

```

[incorrect-dpd.pcap.zip](https://github.com/user-attachments/files/25059786/incorrect-dpd.pcap.zip)

$ zeek -C -r incorrect-dpd.pcap policy/frameworks/dpd/detect-protocols.zeek
```

awelzel:
The detect-protocols script tracks confirmed protocols analyzers for connections and raises notices for detected protocols and servers. It doesn't honor analyzer violation events, however, and so in the case where a HTTP client connects to an SSH server, the HTTP request results in an analyzer confirmation, but the SSH identification string from the server tickles a violation. However, previously Zeek raised notices about an HTTP server running on port 22 despite the analyzer clearly indicating a violation.

Closes #5204

---

Baseline updates:

https://github.com/zeek/zeek-testing-private/pull/32

https://github.com/zeek/zeek-testing/pull/29


0xxon:
If I understand this correctly, with this change, any protocol violation will cause the protocol to no longer be tracked for the connection. This feels non-ideal in the case where a protocol violation appears quite late in a connection. In this case, the connection definitely did speak the protocol - there just was an error at a ltage stage.

I think we maybe should do this differently, and unify the behavior with how the logging of services currently works in `conn.log`, where `DPD::ignore_violations_after` is used as a threshold after which protocol violation events are ignored for the sake of protocol tracking.

0xxon:
Also - this seems to be another of these cases that shows that it might make sense to track the protocols for each direction of a connection separately. I think that idea popped up several times in the past - I kind of wonder if we should revisit at at some point of time.

awelzel:
> If I understand this correctly, with this change, any protocol violation will cause the protocol to no longer be tracked for the connection. 

Found protocols are logged after 30 seconds or when a connection reaches 4000 bytes by default. Whichever comes first and only if both sides had traffic:

https://github.com/zeek/zeek/blob/8e22b0491bd420af3f03fedf0183e1c7cf533901/scripts/policy/frameworks/analyzer/detect-protocols.zeek#L179-L180

https://github.com/zeek/zeek/blob/8e22b0491bd420af3f03fedf0183e1c7cf533901/scripts/policy/frameworks/analyzer/detect-protocols.zeek#L52-L57

So it's a different set of knobs, but I wasn't too concerned about the behavior change here. Do you disagree? (I do agree some more unification would be nice, but a bit outside of the concrete issue).

I'll look at the external test failures and extract some pcaps.

Also agree on the confirmation per side topic: The client definitely spoke HTTP and the server likely spoke SSH, just that they couldn't speak properly with each other.

0xxon:
Thanks for the pointer. You are right - that is somewhat less concerning. However, following the logic, it seems like the thresholds are only checked every `mimimum_duration` (so 30 seconds).

This means if you transfer 20mb or so in the first 20 seconds, and have a protocol violation then, this would  now not result in a log message, even though it's quite a bit over `minimum_volume`. So, perhaps the violation check should also check for `minimum_volume` (or the dpd equivalent, but that would probably be more confusing).

awelzel:
> So, perhaps the violation check should also check for minimum_volume 

That makes sense. Thanks! It's checked every 5 seconds IIUC, but that check seems reasonable - if a violation happens after the minimum_volume, report the detected protocol anyhow.

0xxon:
I'm a bit confused by one of the changes in the external test suites. It seems to me like this removed a bunch of lines in `Baseline/tests.ctu-sme-11-vm-win7ad/notice.log` - which is expected.

But it also seems that this change adds a new line 

```
XXXXXXXXXX.XXXXXX   XXXXXXXXXXX 192.168.1.108   62520   87.240.132.72   80  -   -   -   tcp ProtocolDetector::Server_Found  87.240.132.72: SSL server on port 80/tcp    SSL 87.240.132.72   87.240.132.72   80  -   -   Notice::ACTION_LOG (empty) 3600.000000 -   -   -   -   -   F   1:aExrHHEgQyaQnLvpOy4pcJP9HjU="
```

Did you look into this these connections? I am just curious what's happening here. Is that host legitimately serving TLS on port 80, and the earlier detections now get removed for some reason, or is this still reported incorrectly sometimes?

awelzel:
I had looked and put this comment in the commit/description:

> The one that moved (1:aExrHHEgQyaQnLvpOy4pcJP9HjU=, ShADTgadtfF) has a gap and so the analyzer violation of the responder side isn't raised and Zeek continues to detect SSL.

The client side confirms SSL for the client hello and because of a gap on the responder side the SSL analyzer violation never happens. In other words, we do not reliably remove spurious SSL detections.

This seemed okay to me - do you feel differently? Maybe detect-procols could ignore any connections with gaps, but that seems a bit orthogonal.

0xxon (approved):
Well, I am blind once again. Thanks for the context, that seems fine to me.

PR #5220, opened 2026-02-18
Add a bit of documentation for running btests on Windows

timwoj:
I spent some time fighting getting `btest` to run on Windows again, only to realize I was just holding it wrong. This adds:

- Documentation for how to run it successfully so I don't forget again.
- A script to allow skipping tests on Windows via `@TEST-REQUIRES`. There is a handful of tests that hang indefinitely on Windows and it's ideal to document that they're being skipped on purpose.


testing/btest/core/network_time/set-network-time.zeek
```diff
@@ -1,4 +1,5 @@
 # @TEST-DOC: Ensure setting network time is reflected in following events.
+# @TEST-REQUIRES: ! is-windows
```
awelzel, line 3: It'd probably be good to put a small comment above with a short explanation. Some of them are pretty fundamental and should probably work, so this is mostly temporary until figured out?

awelzel (approved):
Approving, but would be good to comment the `! is-windows` reason into each test - even if the same.

PR #5223, opened 2026-02-19
Add a hook to prevent timeouts

evantypanski:
This came about when chatting with @ckreibich about how to unshunt with the XDP-based shunter. The idea is that we align the shunted timeout with Zeek's, then Zeek remains in charge. I'm curious what others think about this, it was a super simple change, works great, and could be generally useful.

-------------------------

For shunting, there is a problem when finding out how to unshunt. We want to use timeouts, but then clusters have issues determining who unshunts what. We do not want to stop the world to clean up the shunt map. I can add lots of information about why other unshunting solutions fall short for XDP if necessary

Instead, this adds a new feature to Zeek: the ability to prevent a timeout based on external criteria. Then, shunting can rely entirely on Zeek's timeout timers, just hooking in to say "no don't timeout this connection yet, the BPF map says it's still getting shunted packets."

This could be generally useful, like when sharing data between Zeek and some other monitoring system. Those are theoretical, though.


scripts/base/init-bare.zeek
```diff
@@ -1763,6 +1763,14 @@ const icmp_inactivity_timeout = 1 min &redef;
 ## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout icmp_inactivity_timeout set_inactivity_timeout
 const unknown_ip_inactivity_timeout = 1 min &redef;
 
+## When this flow is about to be timed out due to inactivity, this hook will get
+## executed. If you wish to force the connection to stay alive (not get timed out),
+## then break from the hook. The connection will then wait another timeout interval
+## until attempting another timeout.
+##
+## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout icmp_inactivity_timeout unknown_ip_inactivity_timeout set_inactivity_timeout
+type suppress_timeout_hook: hook(c: connection);
```
awelzel: Neat idea.

I'd reach for `hook connection_timing_out(c: connection)` or `session_timing_out(c: connection)` to align with `Analyzer::disabling_analyzer()` and maybe `ssl_finishing()`. The `ing` pattern works for me, but others should speak up if that looks off. Or `hook session_timeout_policy(c: connection)` - could definitely also see that.

So name it more about where/when the hook is invoked rather than what you can do with it?

evantypanski: `connection_timing_out` feels right to me!

scripts/base/init-bare.zeek
```diff
@@ -1763,6 +1763,15 @@ const icmp_inactivity_timeout = 1 min &redef;
 ## .. zeek:see:: tcp_inactivity_timeout udp_inactivity_timeout icmp_inactivity_timeout set_inactivity_timeout
 const unknown_ip_inactivity_timeout = 1 min &redef;
 
+## A hook that can be used to veto timing out a connection via :zeek:see:`break`.
+## The connection will then wait another timeout interval until attempting
```
awelzel:
```suggestion
## The connection will then wait another inactivity interval until attempting
```

src/session/Session.cc
```diff
@@ -155,13 +156,25 @@ void Session::AddTimer(timer_func timer, double t, bool do_expire, zeek::detail:
 void Session::RemoveTimer(zeek::detail::Timer* t) { timers.remove(t); }
 
 void Session::InactivityTimer(double t) {
+    auto new_timeout = inactivity_timeout;
+
     if ( last_time + inactivity_timeout <= t ) {
-        Event(session_timeout_event, nullptr);
-        session_mgr->Remove(this);
-        ++zeek::detail::killed_by_inactivity;
+        static const auto suppress_timeout_hook = zeek::id::find_func("connection_timing_out");
```
awelzel: Nit.

```suggestion
        static const auto timing_out_hook = zeek::id::find_func("connection_timing_out");
```

NEWS
```diff
@@ -61,6 +61,9 @@ New Functionality
 - Added the ``modulo`` BiF, which behaves differently from Zeek's ``%`` operator
   for negative numbers.
 
+- Added the ``connection_timing_out`` hook which, if ``break`` is used within, will
+  prevent Zeek from timing out a particular connection.
```
awelzel: You could add your shunting use case. I wouldn't mind and it's always nice to have some context.

PR #5226, opened 2026-02-23
windows compatibility

mamaorha:


vcpkg-triplets/x64-windows-static-17.cmake
```diff
@@ -0,0 +1,11 @@
+# Custom triplet for MSVC compiler version 17
+
+set(VCPKG_TARGET_ARCHITECTURE x64)
+set(VCPKG_CRT_LINKAGE static)
+set(VCPKG_LIBRARY_LINKAGE static)
+
+set(VCPKG_CXX_FLAGS "${VCPKG_CXX_FLAGS} /guard:cf /Z7")
```
matan-x, line 7: why is needed?

vcpkg-triplets/x64-windows-static-17.cmake
```diff
@@ -0,0 +1,11 @@
...
+set(VCPKG_C_FLAGS "${VCPKG_C_FLAGS} /guard:cf /Z7")
+set(VCPKG_LINKER_FLAGS "${VCPKG_LINKER_FLAGS} /debug:full")
```
matan-x, line 9: same

vcpkg-triplets/x64-windows-static-17.cmake
mamaorha: You can see the comments in the cmakelist (that point to the triplet)
"Set custom vcpkg triplet for Windows to use static libraries with proper PCAP support"

vcpkg-triplets/x64-windows-static-17.cmake
mamaorha: You can see the comments in the cmakelist (that point to the triplet)
"Set custom vcpkg triplet for Windows to use static libraries with proper PCAP support"

vcpkg-triplets/x64-windows-static-17.cmake
timwoj, line 7: Oh man, `/z7` looks nice to not have the pdb files around anymore.

CMakePresets.json
```diff
@@ -0,0 +1,108 @@
...
+      "name": "base",
+      "hidden": true,
+      "description": "Base preset",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/out/build/${presetName}",
+      "installDir": "${sourceDir}/out/install/${presetName}",
+      "cacheVariables": {
+        "CMAKE_C_COMPILER": "cl.exe",
+        "CMAKE_CXX_COMPILER": "cl.exe",
+        "VCPKG_TARGET_TRIPLET": "x64-windows-static-17",
```
timwoj, line 14: Can this be determined automatically? For example, if someone opts to use VS 2026?

vcpkg-triplets/x64-windows-static-17.cmake
```diff
@@ -0,0 +1,11 @@
+# Custom triplet for MSVC compiler version 17
```
timwoj, line 1: This file should be identical for VS 2026 as well, correct? There isn't anything funky about it that changes?

testing/btest/README
```diff
@@ -94,8 +94,15 @@ From within a ``bash`` shell, create a Python virtual environment, install the
     python -m venv /path/to/venv
     /path/to/venv/Scripts/python -m pip install multiprocess
     cd <repo>/testing/btest
+	export MSYS=disable_pcon
     /path/to/venv/Scripts/python ../../auxil/btest/btest <btest options>
 
+### Note: 
+compiling using the presets will make the output go to "out/build/${presetName}" instead of "build" so need to pass the build_dir, for example:
```
timwoj, line 101: The need to change this option is mentioned at the bottom of this file. Maybe it could be combined with that? Either set the value in the file or pass the argument?

testing/btest/bifs/create_file.zeek
```diff
@@ -7,6 +7,13 @@
 
 event zeek_init()
 	{
+	local crlf = 0;
+
+	if ( getenv("OS") == "Windows_NT" )
```
timwoj, line 12: This should be handled as part of the `file_size` BIF instead of as a one-off in a btest (see src/zeek.bif).

scripts/base/utils/paths.zeek
```diff
@@ -5,14 +5,23 @@ const absolute_path_pat = /(\/|[A-Za-z]:[\\\/]).*/;
...
 ## input: a string that may contain an absolute path.
 ##
 ## Returns: the first absolute path found in input string, else an empty string.
 function extract_path(input: string): string
 	{
+	const quoted_dir_pattern = /\"(\/|[A-Za-z]:[\\\/])[^\"]*\"/;
 	const dir_pattern = /(\/|[A-Za-z]:[\\\/])([^\"\ ]|(\\\ ))*/;
+	local quoted_parts = split_string_all(input, quoted_dir_pattern);
+
+	if ( |quoted_parts| >= 3 )
```
timwoj, line 17: Add a comment about what this is doing and why.

testing/btest/scripts/base/utils/paths.test
```diff
@@ -13,9 +13,27 @@ function test_extract(str: string, expect: string)
...
+			str = "";
+		}
+
+	return str;
+	}
+	
 function test_compress(str: string, expect: string)
 	{
 	local result = compress_path(str);
+	result = normalize_path(result);
```
timwoj, line 36: I'm torn on whether this should happen in the `compress_path` BIF or just as part of the test. It being in the BIF means that we'd always output UNIX-style paths from that BIF which might be confusing on Windows. It being part of the test means we have to repeat this kind of thing everywhere it gets called though.

@ckreibich Thoughts?

timwoj (changes requested):
Like I mentioned in Slack, I think it'd be good to split the build-system changes out into a separate PR from the test changes. Getting those in early means follow-up PRs can just be about the functional changes.

testing/btest/scripts/base/utils/paths.test
mamaorha: i had the same thoughts on where to patch it, eventually i rather doing it on the test since it doesnt really have "impact" as windows can handle both "/" and "\\" and i rather avoid the "production runtime" checks and substitution as it will unnecessary slow the function. wdyt?

CMakePresets.json
mamaorha: adjusting to be generic (creating a separated pr and abandoning this one - separating the build changes from tests as u suggested)

vcpkg-triplets/x64-windows-static-17.cmake
mamaorha: adjusting to be generic (creating a separated pr and abandoning this one - separating the build changes from tests as u suggested)

testing/btest/bifs/create_file.zeek
mamaorha: the file_size funciton behave ok, it returns the real number. the problem is the test that is expecting specific "size". in windows when you write "\n" it translated to "\r\n" due to crlf - so either we change the "writing" mechanism or we adjust the test. wdyt?

mamaorha:
splitted this pr into 2 parts:
https://github.com/zeek/zeek/pull/5231
https://github.com/mamaorha/zeek/pull/1 (this one relies on the pr above, once its merged i will rebase and create another pr to zeek master)

PR #5227, opened 2026-02-23
Add optional ZeroMQ CURVE encryption and ZAP handler

awelzel:
This PR adds four new script-level options to the ZeroMQ backend module for public and secret keys for ZeroMQ CURVE server and client roles. These can also be set via environment variable of the form:
```
ZEEK_CLUSTER_BACKEND_ZEROMQ_CLIENT_PUBLICKEY
ZEEK_CLUSTER_BACKEND_ZEROMQ_CLIENT_SECRETKEY
...
```

The central XPUB/XSUB sockets and all logger's PULL sockets act as CURVE servers and use the server keys. Zeek processes connecting to these sockets act as clients, using the client keys.

Anyone with the server public key and the client public and secret key will be able to interact with the pub/sub layer.

There's the potential to support multiple client keys down the road for granular authentication, but seems more useful do to this on the WebSocket API layer. So while this PR adds a ZAP handler, all it does is to verify that the public key from the client is the one that's configured via `curve_client_publickey`.

The zeekctl bump updates ZeekControl to a version that automatically enables encryption when a multi-node cluster is detected:
https://github.com/zeek/zeekctl/pull/106

This PR also introduces a new event `Cluster::Backend::ZeroMQ::monitoring_event()` that allows visibility into socket events produced by ZeroMQ. This can be used to detect a non-encrypted socket connecting to an encrypted one or when authentication fails. These errors result in a `Reporter::fatal()`.  See also https://libzmq.readthedocs.io/en/latest/zmq_socket_monitor.html.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -79,6 +79,28 @@ New Functionality
 - The ``dns.log`` now also contains information about DNS notifications (RFC 1996).
   The ``opcode`` is set to 4 and ``opcode_code`` name will be "notify".
 
+- ZeroMQ cluster communication can now be encrypted using the CURVE mechanism built
+  into ZeroMQ. This requires a pair of keys, one for the server role and one for
+  the client role. The central XPUB/XSUB sockets and LOG pull sockets are configured
+  as CURVE servers, all other sockets (the connecting Zeek nodes) as CURVE clients.
+
+  You can generate Z85-encoded keypair as follows with Zeek:
+
+	$ zeek -e 'print to_json(Cluster::Backend::ZeroMQ::generate_keypair())' | jq
+	{
+	  "public": "}}6zuTySXq(C(6dF#D3Kvt@A5D}LcFlOgVav5{%)",
+	  "secret": "h.<[iyv]O#cWdO%2yuLiEJQ}6tC)@HxN6Hp?![td"
+	}
+
+  When using ZeekControl, keys are automatically generated and configured when a multi-node
+  cluster based on node IP addresses is detected. To force encryption to be always on, set
+  the following ZeroMQ specific setting in zeekctl.cfg (default is "auto"):
+
+	cluster_backend_zeromq.use_curve_encryption = 1
+
+  More details and background is available in the
+  ``policy/frameworks/cluster/backend/zeromq/main.zeek`` script documentation.
+
 - The number of messages and bytes processed by the central ZeroMQ XPUB/XSUB
   cluster component is now exposed as metrics of the form:
 


rsmmr:
> I'm mostly wondering if there's concerns putting the keys as `const &redef` variables and keeping them around in memory? They are easily accessible via `zeekctl print`, but if someone can do that, they can probably already do a lot more things.

Commenting on this part: I think having the redefable variables as one option is fine, we've been doing credentials like that historically as well (say for SSL) and, as you say, having access to `zeekctl print` is a pretty strong position to begin with. But I think it would be good to have an alternative way as well, both for security if somebody doesn't feel it's safe enough, and for convenience if the secrets are coming dynamically from some different place. Otherwise would need to either auto-generate an additional Zeek script in that case or put the values onto the Zeek command-line (and *that* would not be very safe). Most flexible as a second option would probably be environment variables that are examined if the script variables aren't set.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,44 @@
...
+##! The server variables are used for the XPUB/XSUB sockets created by the proxy thread.
+##! If you're hosting the proxy thread elsewhere or using a non-Zeek process, make sure
+##! to configure it with the proper secret key. The logger's PULL sockets use the same
+##! secret key at this point. This means the secret key is required by logger nodes
+##! even if a central XPUB/XSUB component is used externally.
+##!
+##! You may generate the keys as follows. Index 0 is the public key, index 1 the secret key.
+##!
+##!     $ zeek -e 'print Cluster::Backend::ZeroMQ::generate_keypair()'
+##!     [ZE%G/b%A#RP&i(J!#b8CDT^x]XRTMSjdWhAbN$>x, >bxKhbk*%5Wu{V[&-]:1*X5A2%gL3!!=TRK[tx4z]
```
ckreibich: Using Zeek for this directly is cute. :-) Looking at these strings I wondered about their encoding and found https://rfc.zeromq.org/spec/32/, which explains it — maybe link to that?

I also wondered whether it might help to offer a convenience wrapper that prints these directly?
```
redef Cluster::Backend::ZeroMQ::curve_server_publickey = "....";
redef Cluster::Backend::ZeroMQ::curve_server_secretkey = "....";
redef Cluster::Backend::ZeroMQ::curve_client_publickey = "....";
redef Cluster::Backend::ZeroMQ::curve_client_secretkey = "....";
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,44 @@
...
+##!
+##! All Zeek processes share and have access to the same credentials. Note that while
+##! the underlying protocol uses asymmetric cryptographic primitives, we leverage this
+##! more like shared symmetric encryption. Any external process with the server's public
+##! key can and the client secret and public key may connect to the central XPUB/XSUB
+##! socket.
+##!
+##! ZeroMQ supports ZAP to do per-client authentication. This is lightly implemented
```
ckreibich: I wondered here what ZAP is and found it online, so perhaps
```suggestion
##! ZeroMQ's ZAP protocol supports per-client authentication. We implement this lightly
```
(since with "is implemented" I wondered whether you're talking about us or ZeroMQ.)

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,44 @@
...
 ##! :zeek:see:`Cluster::Telemetry::websocket_metrics` for ways to get a better
 ##! understanding about the events published and received.
+##!
+##! Encryption using the CURVE mechanism
+##!
+##!   http://api.zeromq.org/4-2:zmq-curve
+##!
+##! When a Zeek cluster spans multiple systems on a shared and untrusted network,
+##! it's strongly recommended to encrypt the network traffic between individual
+##! Zeek systems. To enable ZeroMQ CURVE encryption, generate keypairs for the
```
ckreibich: Maybe a little bit more context:
```suggestion
##! Zeek systems. ZeroMQ features built-in support for elliptic public-key encryption. To enable it, generate keypairs for the
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -572,3 +640,26 @@ event Cluster::Backend::ZeroMQ::unsubscription(topic: string)
 	else
 		Reporter::warning(fmt("unsubscription of unknown node with id '%s'", gone_node_id));
 	}
+
+event Cluster::Backend::ZeroMQ::monitoring_event(number: count, value: count, address: string)
+	{
+	# Anytime we see a handshake failed error (e.g. wrong CURVE keys or CURVE vs non-CURVE sockets)
+	# speaking, terminate the Zeek process. Preferably we'd do this with
```
ckreibich: Just guessing. :-)
```suggestion
	# terminate the Zeek process. Preferably we'd do this with an error code.
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -572,3 +640,26 @@ event Cluster::Backend::ZeroMQ::unsubscription(topic: string)
...
+	#
+	# From zmq.h
+	# Unspecified system errors during handshake. Event value is an errno.
```
ckreibich, line 676: I wondered as I saw this one whether we want to report these too — could they be useful signals of other connectivity/config trouble?

src/cluster/backend/zeromq/ZeroMQ-Proxy.cc
```diff
@@ -62,6 +64,19 @@ bool ProxyThread::Start() {
 
     xpub.set(zmq::sockopt::xpub_nodrop, xpub_nodrop);
 
+    // Setup curve encryption on the two sockets if enabled. The central XPUB/XSUB
+    // sockets act as curve servers and connecting Zeek processe as curve clients.
+    //
+    // Additionally, start a separate thread with a REP socket in the same context
+    // on the well known ZAP inproc endpoint.
```
ckreibich: A little bit of reasoning why we need to do this would help, either here or in ZeroMQ-ZAP.h.

src/cluster/backend/zeromq/ZeroMQ-ZAP.cc
```diff
@@ -0,0 +1,74 @@
...
+
+            if ( r && *r >= 7 ) {
+                auto mech = request_frames[5].to_string_view();
+                if ( mech == "CURVE" ) {
+                    std::string client_publickey = {request_frames[6].data<const char>(), request_frames[6].size()};
+
+                    if ( zap_args->allowed_publickeys.contains(client_publickey) ) {
+                        status_code = "200";
+                        status_text = "OK";
+                        user_id = "0";
```
ckreibich, line 44: Is the reason we need to do this simply so something confirms acceptance of the public key to a ZAP-speaking client?

src/cluster/backend/zeromq/ZeroMQ.cc
```diff
@@ -323,6 +379,50 @@ bool ZeroMQBackend::DoInit() {
...
     xsub.set(zmq::sockopt::rcvbuf, xsub_rcvbuf);
 
+    if ( curve_config.isClientEnabled() ) {
+        ZEROMQ_DEBUG("Enabling encryption on client XPUB and XSUB sockets");
+        curve_config.configureClientCurveSockOpts(xpub);
+        curve_config.configureClientCurveSockOpts(xsub);
+    }
+
+    // Create monitoring sockets for xpub, xsub and log_push sockets. For now,
+    // there'll only be three client sockets.
```
ckreibich, line 433: Here too I wondered why we need to do this — so we can see whether the handshake errors out due to crypto problems, yes?

testing/btest/cluster/zeromq/logging-encryption.zeek
```diff
@@ -0,0 +1,146 @@
...
+fi
+
+if [ -f DONE ]; then
+	exit 0
+fi
+
+# Remove hostname and pid from node id in message.
+zeek-cut node message < $CLUSTER_LOG | sed -r 's/_[^_]+_[0-9]+_/_<hostname>_<pid>_/g' | sort > cluster.log.tmp
+
+# 4 times 5
```
ckreibich, line 138: :-)

testing/btest/cluster/zeromq/logging-encryption.zeek
```diff
@@ -0,0 +1,146 @@
+# @TEST-DOC: Startup a ZeroMQ cluster by hand, testing basic logging and node_up and node_down events. This is the same as logging.zeek, but also enables encryption.
```
ckreibich, line 1: Would it make sense to add a `Cluster::Backend::ZeroMQ::monitoring_event` to these tests to confirm that encryption is in fact active? I may well have missed it, but I wondered whether these tests would pass happily if encryption were somehow configured but not actually active.

ckreibich (commented):
Looks great to me Arne, just a couple of questions/suggestions from me for now. I think this is the same as when we started out: you know the details about the ZeroMQ API, I don't — so at times a bit of extra context would help me.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: Yeah, will add. Thanks!

testing/btest/cluster/zeromq/logging-encryption.zeek
awelzel: Hmm, the same events are raised whether encryption is on or not. It's true the test would pass without encryption, too. There's tests that verify encrypted/non-encrypted nodes cannot connect and fail fast and the redef variables work and tcpdump'ing cluster comms seems like overdoing it here.

src/cluster/backend/zeromq/ZeroMQ.cc
awelzel: Yeah, the main motivation for the monitoring sockets was to fail fast Otherwise ZeroMQ will just continue reconnections failing with handshakes over and over again.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: Hmm, actually. These would be raised for connectivity issues during the handshake apparently (that's what the tests indicate) and for those seems okay to just have ZeroMQ retry again and instead only hard-exit if the handshake actually failed.

src/cluster/backend/zeromq/ZeroMQ-ZAP.cc
awelzel: Yeah, that's correct. And the ZAP-speaking-client is actually within libzmq itself:

https://github.com/zeromq/libzmq/blob/master/src/zap_client.hpp


Everytime there's a connection to a XSUB or XPUB socket, libzmq itself sens a ZAP formatted message and we reply with a response.

I'm a bit conflicted if we really need this all - encryption works without this. But it does pave the way a bit if anyone ever would like to add per-process client key auth or some such.

doc/advanced/devel/cluster/backend/zeromq.rst
```diff
@@ -116,3 +121,18 @@ types in a Zeek cluster looks something like the following.
 
 .. figure:: /images/cluster/backend/zeromq/zeromq-cluster.png
 
+
+Encryption
+==========
+
+.. versionadded:: 8.2
+
+When using ZeroMQ as a cluster backend, communication between different Zeek nodes
```
ckreibich:
```suggestion
When using ZeroMQ as a cluster backend, communication between Zeek nodes
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,63 @@
...
+##! To avoid confusion, either script-level or environment variable configuration
+##! should be used. Mixing the approaches will result in a fatal error at startup.
+##!
+##! The central XPUB/XSUB sockets created by the proxy thread act as CURVE server.
+##! If you're hosting the XPUB/XSUB sockets elsewhere or using a non-Zeek process,
+##! make sure to configure it with the proper secret key and provide the public key
+##! to the clients. The logger's PULL socket uses the same secret key as the XPUB/XSUB
+##! sockets at this point. This means logger nodes require access to the secret key
+##! even if the XPUB/XSUB component is external to Zeek. Additionally, the current
+##! implementation uses the client's public key for authentication.
```
ckreibich, line 99: You could mention higher up that this setup aims to provide both confidentiality and authentication — it's cool that it does both. (No biggie.)

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,63 @@
...
+##! more like shared symmetric encryption. Any external process with access to the
+##! server public key and the client secret and public key can connect to the central
+##! XPUB/XSUB sockets or any logger PULL socket.
+##!
+##! ZeroMQ's ZAP protocol supports per-client authentication. We implement this lightly
+##! such that any client needs to possess the configured client secret and public key.
+##! Today, this means every node holds the client keys as by default the manager not
+##! only hosts the central XPUB/XSUB sockets, but also connects to them.
+##! It's not clear if anything more is really useful. Advanced authentication or
+##! authorization concepts should probably be added to the WebSocket API instead.
```
ckreibich, line 122: I think you could consider dropping this paragraph. I'm not sure it adds much for folks who aren't ZAP gurus, and the code below has relevant details for those who are.

NEWS
```diff
@@ -50,6 +50,28 @@ New Functionality
 
 	redef PacketAnalyzer::VXLAN::vxlan_ports += { 4711/udp };
 
+- ZeroMQ cluster communication can now be encrypted using the CURVE mechanism built
+  into ZeroMQ. Today, this requires two keys. One for the server role and one for
```
```suggestion
  into ZeroMQ. This requires a pair of keys, one for the server role and one for
```

NEWS
```diff
@@ -50,6 +50,28 @@ New Functionality
...
+  the client role. The central XPUB/XSUB sockets and LOG pull sockets are configured
+  as CURVE servers, all other sockets (the connecting ones) as CURVE clients.
```
```suggestion
  as CURVE servers, all other sockets (the connecting Zeek nodes) as CURVE clients.
```

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -61,6 +61,63 @@
...
+##!   * ``ZEEK_CLUSTER_BACKEND_ZEROMQ_CURVE_SERVER_SECRETKEY``
+##!
+##! To avoid confusion, either script-level or environment variable configuration
+##! should be used. Mixing the approaches will result in a fatal error at startup.
+##!
+##! The central XPUB/XSUB sockets created by the proxy thread act as CURVE server.
+##! If you're hosting the XPUB/XSUB sockets elsewhere or using a non-Zeek process,
+##! make sure to configure it with the proper secret key and provide the public key
+##! to the clients. The logger's PULL socket uses the same secret key as the XPUB/XSUB
+##! sockets at this point. This means logger nodes require access to the secret key
```
ckreibich:
```suggestion
##! sockets. This means logger nodes require access to the secret key
```

ckreibich (approved):
This still looks great to me. Thanks for the detailed comments. I haven't spent much time actually tinkering with it, but it worked in a quick zeekctl experiment. 👍

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: I'm prefixing it with "Implementation Note" just in case anyone looks at this in a few years this might still be a good hint.

PR #5230, opened 2026-02-24
Remove trivial uses for unistd.h

Neverlord:
The only potentially controversial change, I think, is in `File::Size`: using `fseek` / `ftell` instead of `fstat` to get the file size. Those two may give different results.

- Replace `sleep` and `usleep` with `sleep_for`
- Replace `ssize_t` with `ptrdiff_t`
- Replace `pid_t` with `int`
- Replace uses of `cwd` & co with `std::filesystem`
- Remove `unistd.h` includes where they aren't actually needed

### DIFF
--- a/src/File.cc
+++ b/src/File.cc
@@ -30,7 +30,11 @@
 
 namespace zeek {
 
-std::list<std::pair<std::string, File*>> File::open_files;
+namespace {
+
+std::list<std::pair<std::string, File*>> open_files;
+
+} // namespace
 
 // Maximizes the number of open file descriptors.
 static void maximize_num_fds() {
--- a/src/Anon.cc
+++ b/src/Anon.cc
@@ -3,7 +3,6 @@
 #include "zeek/Anon.h"
 
 #include <sys/time.h>
-#include <unistd.h>
 #include <cassert>
 #include <cstdlib>
 
--- a/src/DNS_Mgr.cc
+++ b/src/DNS_Mgr.cc
@@ -10,10 +10,11 @@
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <unistd.h>
 #include <algorithm>
 #include <cerrno>
+#include <chrono>
 #include <cstdlib>
+#include <thread>
 #include <vector>
 
 #ifdef TIME_WITH_SYS_TIME
@@ -45,6 +46,8 @@ using ztd::out_ptr::out_ptr;
 #include "zeek/telemetry/Manager.h"
 
 #include "zeek/3rdparty/doctest.h"
+
+using namespace std::literals;
 
 // Number of seconds we'll wait for a reply.
 constexpr int DNS_TIMEOUT = 5;
@@ -1538,7 +1541,7 @@ TEST_CASE("dns_mgr async host" * doctest::skip(true)) {
     int count = 0;
     while ( ! cb.done && (count < DNS_TIMEOUT + 1) ) {
         mgr.Process();
-        sleep(1);
+        std::this_thread::sleep_for(1s);
         if ( ! cb.timeout )
             count++;
     }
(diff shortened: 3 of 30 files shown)


src/File.cc
```diff
@@ -302,13 +305,23 @@ void File::RaiseOpenEvent() {
...
+        reporter->Error("can't get file position: %s", strerror(errno));
+        return 0;
+    }
+    if ( fseek(f, 0, SEEK_END) < 0 ) {
+        reporter->Error("can't seek to the file's end: %s", strerror(errno));
+        return 0;
+    }
+    const auto sz = ftell(f);
+    if ( fseek(f, pos, SEEK_SET) < 0 ) {
+        reporter->Error("can't restore file position: %s", strerror(errno));
```
timwoj: Could this use `std::filesystem::file_size`?

timwoj (commented):
I like it 👍🏼 

I started up the Windows build. I'm also curious if there's any benchmarker differences but I highly doubt there will be.

timwoj:
```
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(127): error C3646: '__attribute__': unknown override specifier
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(127): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(127): error C2059: syntax error: '('
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(127): error C2059: syntax error: '('
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(128): error C2143: syntax error: missing ')' before ';'
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(128): error C2059: syntax error: ')'
[20:34:05.748] C:\zeek\src\include\zeek/threading/BasicThread.h(128): error C2238: unexpected token(s) preceding ';'
```

`__attribute__` doesn't exist on Windows. In libunistd, it's [defined away](https://github.com/robinrowe/libunistd/blob/3f269dfc49249dd218cf229439b787164d1e3e03/unistd/sys/posix_stdio.h#L151). Note that this is currently broken because `__has_attribute` is defined for some reason, and so the `#ifdef` never gets parsed. I have it fixed on the upgrade branch in our fork to just remove the `__has_attribute` check. Either way, we could just define that somewhere (util.h?).

timwoj:
@Neverlord You'll also need to check `tools/bifcl` and `tools/binpac`. Both have inclusions of `unistd.h`.

Note that this removal is just a start: It's not just `unistd.h` but a whole raft of other headers that libunistd provides (`sys/types.h`, etc). For example,  paraglob depends on `fnmatch()` from `fnmatch.h`.

EDIT: What do you think about a longer-term plan of making a `tools/posix_utils` path/library/set of headers that can be used across all of the included projects? Things like `src/3rdparty/bsd-getopt-long.{c,h}` could move there as well. This discussion might be a better handed in a ticket.

src/File.cc
Neverlord: That's been my first idea as well. The problem is that this function needs a path to the file, not a `FILE*` handle. And we have this constructor in `File`:

```c++
File::File(FILE* arg_f) {
    Init();
    f = arg_f;
    name = access = nullptr;
    t = base_type(TYPE_STRING);
    is_open = (f != nullptr);
}
```

Since `name` can be `null`, we can't reliably produce a filesystem path for `file_size` here. Also, we do things like `print_stdout = new File(stdout);` (in `Stmt.cc`). So the `File` might actually not have a path to begin with.

Neverlord:
> This discussion might be a better handed in a ticket.

Full ACK. I've seen you've opened a ticket, so let's discuss the details over there.

src/File.cc
Neverlord: OK, I've played a bit around with it. `fstat(fileno(stdout))` will report the file size as 0.

If you do `ftell(stdout)`, it will produce an error. We could:
- ignore the error and just assume the file isn't on disk (return 0)
- call `std::filesystem::file_size` if `name != nullptr` and return 0 otherwise
- just leave this one as-is and for now and deal with it via the new `posix_utils` later

@timwoj preference?

Neverlord: I've reverted it for now, can be a topic for another day.

timwoj: That works for me. I think the `file_size` checking for null is fine also, but the PR as a whole is a lot of cleanup and missing this one instance is ok.

PR #5234, opened 2026-02-25
[windows] fixing btest failures on windows

mamaorha:
this PR fixes issues that were raised btests (down from 495 to 70 failing tests)

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -553,6 +553,14 @@ string normalize_path(std::string_view path) {
     if ( had_dot_prefix && ! result.starts_with("./") && ! result.starts_with("../") )
         result = "./" + result;
 
+    if ( result.size() > 1 && result.back() == '/' ) {
+        bool is_drive_root = result.size() == 3 && std::isalpha(static_cast<unsigned char>(result[0])) &&
+                             result[1] == ':' && result[2] == '/';
+
+        if ( ! is_drive_root )
+            result.pop_back();
+    }
+
     return result;
 #else
     if ( path.find("/.") == std::string_view::npos && path.find("//") == std::string_view::npos ) {


src/zeek.bif
```diff
@@ -4582,81 +4582,97 @@ function haversine_distance%(lat1: double, long1: double, lat2: double, long2: d
 ##          ``rw[xsS]rw[xsS]rw[xtT]``.
 function file_mode%(mode: count%): string
 	%{
+	// Use standard POSIX bit values directly so this function works
+	// correctly on all platforms, including Windows where the system
+	// defines map group/other bits to user bits.
+	constexpr unsigned int PM_IRUSR = 0400;
```
timwoj, line 4648: I would rather map these to the header values on Linux, even if we have to have a separate block to define them by hand on Windows. I realize they're unlikely to ever change in POSIX, but it's safer.

src/util.cc
```diff
@@ -841,10 +849,17 @@ int setvbuf(FILE* stream, char* buf, int type, size_t size) {
...
+    // Use _IONBF instead so data is flushed promptly on each write, which is
+    // closer to line-buffering semantics than full buffering.
+    if ( type == _IOLBF )
+        return ::setvbuf(stream, NULL, _IONBF, 0);
+
+    // For _IOFBF, allocate a default-sized buffer when none is provided.
+    if ( type == _IOFBF && buf == nullptr && size == 0 )
+        size = BUFSIZ;
+
+    return ::setvbuf(stream, buf, type, size);
```
timwoj, line 894: Nice! Glad to have that TODO cleaned up.

src/zeek.bif
```diff
@@ -604,7 +604,13 @@ static bool prepare_environment(zeek::TableVal* tbl, bool set)
 
 static int do_system(const char* s)
 	{
+#ifdef _MSC_VER
+	// system() on Windows uses cmd.exe which doesn't handle Unix shell
+	// syntax.  Wrap the command in bash for consistent behavior.
+	const char* system_fmt = "bash -c \"(%s) 1>&2 &\"";
```
timwoj: I have concerns with this if the command has quotes in it itself.

src/zeek.bif
mamaorha: as you said - those wont change as a lot of things rely on it but i will adjust if you wish, just to be clear - you want me to add a "IFDEF" block that check if we are on windows and "unset&set" those values again? (this feels risky as im not sure what other usages of those will pop) - or i missunderstood and you wish to have 2 functions and route based on OS?

src/zeek.bif
mamaorha: great catch!, adding a test to verify + will fix issues if raised. (this happen twice in that file, will fix both)

src/zeek.bif
timwoj: I mean something like:

```
#ifndef _WIN32
const unsigned int PM_IRUSR = S_IRUSR;
...
#else
constexpr unsigned int PM_IRUSR = 0400;
...
#endif

File: src/zeek.bif (hunk 3 above)
Comment (mamaorha) [reply src/zeek.bif]: @timwoj can i resolve this? (see the changes i have made here https://github.com/zeek/zeek/pull/5234/changes/72dd16691b90478de9342904a1f88d28c965e68c)

File: src/zeek.bif (hunk 1 above)
Comment (mamaorha) [reply src/zeek.bif:4648]: @timwoj  adjusted here https://github.com/zeek/zeek/pull/5234/changes/da696f35666d321ed2e0749eb1bbd5b6bf49490d (feels a bit odd tho, i dont believe those values will ever change) - kindly resolve if that was what you meant, otherwise lmk what i missed

File: src/zeek.bif (hunk 3 above)
Comment (timwoj) [reply src/zeek.bif]: I definitely feel better about that now. Thanks.

**Comment by timwoj** [comment]
It looks like something here broke the unit tests for `normalize_path`:

```
[17:44:14.110] C:\zeek\src\util.cc(2134): ERROR: CHECK( detail::normalize_path("../zeek/testing/..") == "../zeek" ) is NOT correct!
[17:44:14.110]   values: CHECK( ../zeek/ == ../zeek )
```

You can run the unit tests with `zeek --test`. There's not a lot of them, but they cover a bunch of the functions in `util.h`.

PR #5239, opened 2026-02-27
Move zeek-cut from auxil to tools

timwoj:
This PR moves the `zeek-cut` utility out of zeek-aux repo into the tools directory of the main repo. It also fixes some issues needed to make it build on Windows. The first commit moves the files as-is with no other changes. The following two commits adapt the code to work both on Windows and run the btests on all platforms. I also verified that `zeek-cut` is still installed as part of the regular Zeek installation.


ckreibich (changes requested):
Just one request Tim: since the tools folder is growing pretty steadily, could you make `testing/btest/zeek-cut` -> `testing/btest/tools/zeek-cut`, for symmetry with systemd-generator (and same for Baselines)?

I also wonder whether `testing/btest/Logs` should be `testing/btest/Files/logs` since it's arguably not that special, but I don't feel strongly here.

timwoj:
I moved the log file to Files/zeek-cut-logs and moved all of the tests/baselines under tools.

PR #5241, opened 2026-03-02
Fix issue with timeouts rescheduling too much

evantypanski:
If we always base the timeout on the last packet, it will be incorrect after the second time it tries to timeout and spin forever without waiting. This makes the rescheduling of timeouts after connection_timing_out breaks to be based on the timer instead. This was essentially the intention of the initial feature.

Doing so I changed the test a bit with an artificial pcap


src/session/Session.cc
```diff
@@ -169,12 +169,14 @@ void Session::InactivityTimer(double t) {
             return;
         }
 
-        // The hooked stopped us! Add the inactivity timeout so that we don't
-        // immediately run again.
-        new_timeout += inactivity_timeout;
+        // The hooked stopped us! Wait inactivity_timeout time until
+        // checking again from this time, not from the last packet.
+        // This stops it from infinitely spinning.
```
awelzel: Puh, nice catch.

Mostly nit for the comment.

```suggestion
        // Schedule the next timer based on this timer's expiry rather than the Session's
        // last_time (last packet timestamp).
        // Otherwise, the InactivityTimer expires right away again, causing in infinite spinning.
```

testing/btest/core/stop-timeout.zeek
```diff
@@ -1,29 +1,33 @@
...
-		print fmt("Suppressed the timeout for connection %s %d time(s)", c$uid, current_suppressions + 1);
-		current_suppressions += 1;
-		break;
-		}
+	print fmt("Suppressed the timeout for connection %s %d time(s)", c$uid,
+	    current_suppressions + 1);
+
+	# Ensure each one waits at least tcp_inactivity_timeout
+	if ( last_timeout_attempt != 0 )
+		assert last_timeout_attempt + tcp_inactivity_timeout <= network_time();
```
awelzel, line 21: You could consider printing `network_time()` to the baseline. That can be a bit easier to understand when something goes off compared to an assert (and anyhow there's a baseline already). Particular if it happens only in CI on some system and its not immediately reproducible locally.

Also nit.

awelzel (approved):

PR #5244, opened 2026-03-03
ci/freebsd: Make pkg use latest over quarterly ABI, bump 13.5 to 15.0

awelzel:
Currently, the FreeBSD builds are failing because the dnsmasq package isn't available in the quarterly ABI/repo (?). Edit /etc/pkg/FreeBSD.conf to use the "latest" ABI/repo where the dnsmasq package is again available.

Also bump FreeBSD 13.5 to 15.0. The latter was released early December 2025.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -532,6 +532,19 @@ macos_sequoia_task:
   << : *MACOS_ENVIRONMENT
 
 # FreeBSD EOL timelines: https://www.freebsd.org/security/#sup
+freebsd15_task:
+  freebsd_instance:
+    # FreeBSD 15 EOL: Dec 29 2029
+    image_family: freebsd-15-0-amd64-zfs
+    << : *FREEBSD_RESOURCES_TEMPLATE
+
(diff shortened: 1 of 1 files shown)


0xxon:
I was curious about what's going on here with dnsmasq in the quarterly pkg repo - as that's the recommended (stable) one to use.

It turns out that the build was broken, and then unbroken a couple of days later in https://github.com/freebsd/freebsd-ports/commit/31a7d2a01bc7bdcbb05e7c32aa4ef7dd514dbd08

This, however, has obviously not quite propagated yet - as I also don't see binary packages available.

awelzel:
> It turns out that the build was broken, and then unbroken a couple of days later in https://github.com/freebsd/freebsd-ports/commit/31a7d2a01bc7bdcbb05e7c32aa4ef7dd514dbd08

Sorry for not leaving a bread crumb. I also got there.

Would you prefer sticking with quarterly instead? I figured for CI either would be fine, but no real opinion, so I'm okay we wait it out until dnsmasq becomes available in quarterly again, too.

PR #5245, opened 2026-03-03
Spicy: Fix event parameter mismatches due to too aggressive optimizations.

rsmmr:
We now declare units as exported not only when they are passed
directly as event parameters, but also if they are nested into other
aggregates (e.g., a vector of units).


timwoj:
I can confirm this fixes the weird that I was seeing from this IGMP analyzer. Thanks!

PR #5246, opened 2026-03-03
[windows] enhance windows combatibility, fixing various btests

mamaorha:


.gitattributes
```diff
@@ -2,5 +2,5 @@
 *.pac linguist-detectable=false
 *.bif linguist-language=C++
 *.l linguist-language=Lex
-testing/btest/Baseline/** linguist-detectable=false
+testing/btest/Baseline/** linguist-detectable=false -text
```
timwoj: Can you explain exactly what this is doing? Would it be better to put `text eol=lf` like the file below, if we're just after making sure that the line endings are enforced? I can't think of a test baseline file that isn't text.

src/input/readers/config/Config.h
```diff
@@ -39,7 +40,7 @@ class Config : public ReaderBackend {
 
     std::ifstream file;
     time_t mtime;
-    ino_t ino;
+    uint64_t ino;
```
timwoj: I would rather define `ino_t` in libunistd than change the type here.

timwoj (changes requested):

.gitattributes
mamaorha: we can probably revert this after the change we merged into btest to remove the "\\r" - i will verify

src/input/readers/config/Config.h
mamaorha: i dont think we have any "ino_t" in libunistd due to the fact its more posix style. maybe we can add ifdef on the config.h to use ino_t on non windows and uint64_t on windows?

what do you think the best approach here will be?

src/util.cc
```diff
@@ -522,8 +522,17 @@ string flatten_script_name(const string& name, const string& prefix) {
     if ( ! rval.empty() )
         rval.append(".");
 
-    if ( is_package_loader(name) )
-        rval.append(SafeDirname(name).result);
+    if ( is_package_loader(name) ) {
+        string dir = SafeDirname(name).result;
+
+#ifdef _MSC_VER
+        // Windows dirname() may leave a trailing separator.
```
timwoj: Doesn't `SafeDirName()` do this for you already (the change directly above this one)?

testing/btest/core/reporter-paths.zeek
```diff
@@ -1,7 +1,7 @@
 # This test verifies Zeek's file path normalization.
 #
 # Absolute paths are preserved but normalized:
-# @TEST-EXEC: zeek -b $PWD/././test.zeek 2>&1 | sed "s|$PWD|/...|" >output
+# @TEST-EXEC: zeek -b $PWD/././test.zeek 2>&1 | sed "s|$(cygpath -m "$PWD" 2>/dev/null || echo "$PWD")|/...|;s|$PWD|/...|" >output
```
timwoj, line 7: Can you add a comment about why it's calling `cygpath`?

src/input/readers/config/Config.h
timwoj: Right, but definitions for other things like `pid_t`, `gid_t`, etc exist. See `unistd/sys/sys_types.h`.

src/util.cc
mamaorha: when giving dir like "test" it was resolving to "test/".
edit: oh i see your point, already handled that in another commit

testing/btest/core/reporter-paths.zeek
mamaorha: @timwoj added a comment in the file, resolve if its clear / lmk if you need it requires a change

src/input/readers/config/Config.h
mamaorha: there's a blocker. The MSVC CRT defines _ino_t as unsigned short (16-bit) and struct stat.st_ino uses it directly:

   // MSVC ucrt/sys/types.h
   typedef unsigned short _ino_t;  // guarded by _INO_T_DEFINED
   typedef _ino_t ino_t;

   // MSVC ucrt/sys/stat.h
   struct _stat64 {
       _ino_t st_ino;    // always 0 on Windows anyway
       ...
   };

  If we redefine ino_t to uint64_t in libunistd, the struct stat layout would break — every stat() call would corrupt memory since the CRT fills in a 2-byte field but we'd read 8
  bytes.
  
  
  i will wrap it in the places i used it so it will only hit on windows

testing/btest/Baseline/scripts.base.frameworks.telemetry.sync-collect/out
```diff
@@ -1,13 +1,14 @@
 ### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+### NOTE: This file has been sorted with diff-sort.
+btest_connections_total, [tcp], 1.0
+btest_connections_total, [tcp], 1.0
+btest_connections_total, [tcp], 1.0
+btest_connections_total, [tcp], 1.0
+btest_connections_total, [udp], 1.0
+btest_connections_total, [udp], 2.0
+btest_connections_total, [udp], 2.0
```
timwoj: I don't think this one should be sorted, as the output is specifically what's happening between each sync point.

testing/btest/scripts/base/frameworks/intel/path-prefix/input-intel-absolute-prefixes.zeek
```diff
@@ -5,8 +5,8 @@
 # /foo/bar/intel).
 #
 # @TEST-EXEC: mkdir -p intel
-# @TEST-EXEC: cat %INPUT | sed "s|@path_prefix@|$PWD/intel|" >input.zeek
-# @TEST-EXEC: ZEEKPATH=$ZEEKPATH:$TEST_BASE/scripts/base/frameworks/intel/path-prefix zeek -b input.zeek >output
+# @TEST-EXEC: cat %INPUT | sed "s|@path_prefix@|$(cygpath -m "$PWD" 2>/dev/null || echo "$PWD")/intel|" >input.zeek
+# @TEST-EXEC: ZEEKPATH=$ZEEKPATH:$(cd "$TEST_BASE/scripts/base/frameworks/intel/path-prefix" && pwd) zeek -b input.zeek >output
```
timwoj, line 9: This is weird to me. Is this something with the variable not being set correctly when zeek runs?

src/input/InputFile.cc
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    FILE* fp = _fdopen(fd, mode);
+    if ( ! fp )
+        _close(fd); // also closes the underlying handle
+
+    return fp;
+}
+
+uint64_t reliable_inode(const char* path, uint64_t stat_ino) {
```
timwoj: Should this return `file_ino_t`? If that definition needs to move somewhere more common, `util-types.h` is a good place for it.

src/input/InputFile.cc
```diff
@@ -0,0 +1,144 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#ifdef _WIN32
+
+#include "zeek/input/InputFile.h"
```
timwoj, line 3: This include can be lifted outside the #ifdef block

src/input/InputFile.h
```diff
@@ -0,0 +1,82 @@
...
+#include <cstdio>
+#include <fstream>
+#include <string>
+
+#ifdef _WIN32
+
+#include <istream>
+#include <memory>
+
+namespace zeek::input::reader::detail {
```
timwoj, line 17: I'd rather the namespace wrap the whole #ifdef block here instead of repeatedly opening and closing it. You'll need one extra #ifdef for the _WIN32 includes, but I think it'll look cleaner in the end.

testing/btest/scripts/base/frameworks/input/path-prefix/absolute-prefix.zeek
```diff
@@ -2,14 +2,14 @@
...
-# @TEST-EXEC: cat %INPUT | sed "s|@path_prefix@|$PWD/subdir|" >input.zeek
+# @TEST-EXEC: cat %INPUT | sed "s|@path_prefix@|$(cygpath -m "$PWD" 2>/dev/null || echo "$PWD")/subdir|" >input.zeek
 # @TEST-EXEC: mkdir -p subdir
 #
 # Note, in the following we'd ideally use %DIR to express the
 # additional path, but there's currently a problem in btest with using
 # %DIR after TEST-START-NEXT.
 #
-# @TEST-EXEC: ZEEKPATH=$ZEEKPATH:$TEST_BASE/scripts/base/frameworks/input/path-prefix zeek -b input.zeek >output
+# @TEST-EXEC: ZEEKPATH=$ZEEKPATH:$(cd "$TEST_BASE/scripts/base/frameworks/input/path-prefix" && pwd) zeek -b input.zeek >output
```
timwoj, line 12: Same comment re: why calling the `pwd` here as before.

timwoj (changes requested):

src/input/InputFile.cc
mamaorha: yeah i guess i missed that (had this one duplicated and later extracted into this shared class... not sure about util-types.h since this type is only used by the InputFile.cc wdyt?

mamaorha: edit: ah odd i see its already fine in the branch - not sure why its showing old code in this window

src/input/InputFile.h
mamaorha: adjusted. resolve if it is what you had in mind otherwise lmk

src/input/InputFile.cc
mamaorha: adjusted. resolve if it is what you had in mind otherwise lmk

src/input/InputFile.cc
mamaorha: still relevant or can we resolve?

testing/btest/Baseline/scripts.base.frameworks.telemetry.sync-collect/out
mamaorha: since std::hash is implementation-defined, MSVC and GCC/libstdc++ produce different bucket orderings — Linux happens to iterate [udp] before [tcp], Windows iterates [tcp] before [udp].

i saw we have "running_under_test" which should help with deterministic sorting, do you want me to try that?

testing/btest/scripts/base/frameworks/intel/path-prefix/input-intel-absolute-prefixes.zeek
mamaorha: we have "relative paths", im getting "fatal error in ./input.zeek, line 13: can't find path-prefix-common.zeek".
by "cd-ing" into the folder and running pwd im getting full path

testing/btest/scripts/base/frameworks/input/path-prefix/absolute-prefix.zeek
mamaorha: we have "relative paths", im getting "fatal error in ./input.zeek, line 13: can't find path-prefix-common.zeek".
by "cd-ing" into the folder and running pwd im getting full path

testing/btest/Baseline/scripts.base.frameworks.telemetry.sync-collect/out
timwoj: Yes, please. The test ordering here does matter.

testing/btest/scripts/base/frameworks/intel/path-prefix/input-intel-absolute-prefixes.zeek
timwoj: A comment here would help. I wonder if we should document these specific things somewhere so that if someone runs into something similar when adding a new test, they're not completely lost. Maybe just keep track of the workarounds for now and I can write something up one you're all done.

src/input/InputFile.cc
timwoj: Ah, I was reviewing commit-by-commit so I might have been looking at something old. Resolving.

src/input/InputFile.cc
timwoj: That looks better now 👍🏼

testing/btest/scripts/base/frameworks/intel/path-prefix/input-intel-absolute-prefixes.zeek
mamaorha: we have this in multiple files - you want me to add comment to each?

testing/btest/Baseline/scripts.base.frameworks.telemetry.sync-collect/out
mamaorha: https://github.com/zeek/zeek/pull/5246/changes/566ca51efd61f634baac81173aeec26f826f2c51

testing/btest/scripts/base/frameworks/intel/path-prefix/input-intel-absolute-prefixes.zeek
timwoj: I guess not. Just make a note somewhere that you had to do it and we'll collect them together once everything is done.

src/input/readers/raw/Raw.cc
```diff
@@ -113,8 +151,141 @@ std::unique_lock<std::mutex> Raw::AcquireForkMutex() {
...
+        return false;
+    }
+    SetHandleInformation(hStdoutRead, HANDLE_FLAG_INHERIT, 0);
+
+    // Create stderr pipe if needed
+    HANDLE hStderrRead = INVALID_HANDLE_VALUE, hStderrWrite = INVALID_HANDLE_VALUE;
+    if ( use_stderr ) {
+        if ( ! CreatePipe(&hStderrRead, &hStderrWrite, &sa, 0) ) {
+            Error(Fmt("Failed to create stderr pipe: %lu", GetLastError()));
+            CloseHandle(hStdoutRead);
```
timwoj: Would it be possible to use cleanup tricks with `unique_ptr` to handle the calls to `CloseHandle`? It feels like it's really easy to miss something when closing whatever's been opened at any point. I'd be ok with doing this as a follow-up to this PR. Maybe just using `UniqueHandle`?

src/input/InputFile.cc
```diff
@@ -0,0 +1,175 @@
...
+UniqueHandle open_shared(const char* path, DWORD access) {
+    return UniqueHandle(CreateFileA(path, access, share_all, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr));
+}
+
+} // namespace
+
+// A streambuf backed by a Windows HANDLE opened with FILE_SHARE_DELETE,
+// allowing external renames while the file is open (matching POSIX semantics).
+class WinShareDeleteBuf : public std::streambuf {
+    UniqueHandle handle_;
```
timwoj: Style nit: We typically do public/protected/private, with member variables at the end of the appropriate section.

src/util.cc
```diff
@@ -522,8 +522,11 @@ string flatten_script_name(const string& name, const string& prefix) {
     if ( ! rval.empty() )
         rval.append(".");
 
-    if ( is_package_loader(name) )
-        rval.append(SafeDirname(name).result);
+    if ( is_package_loader(name) ) {
+        string dir = SafeDirname(name).result;
```
timwoj: This temporary seems unnecessary. Was this change a leftover from something else?

src/zeek.bif
```diff
@@ -605,13 +609,79 @@ static bool prepare_environment(zeek::TableVal* tbl, bool set)
...
 	{
 #ifdef _MSC_VER
-	// Pipe the command to bash's stdin to avoid quoting issues with
-	// cmd.exe that system() uses on Windows.
-	FILE* f = popen("bash", "wb");
-	if ( ! f )
+	// Use CreateProcess with an explicit pipe instead of _popen.
+	// _popen internally uses _dup/_dup2 on the CRT file descriptor
+	// table which is not thread-safe and can corrupt FDs held by the
+	// log-writer thread (observed as "safe_write error: 9").
```
timwoj, line 615: Oh, that's a good catch!

src/supervisor/Supervisor.cc
```diff
@@ -513,19 +604,23 @@ void Supervisor::HandleChildSignal() {
         auto msg = make_create_message(node.config);
         write_msg(stem_pipe, msg, &msg_buffer);
     }
+#endif
 }
 
 void Supervisor::InitPostScript() {
     node_status = event_registry->Register("Supervisor::node_status");
 
+#ifndef _MSC_VER
```
timwoj, line 610: Why don't these hooks fire for Windows? We opened the stdout/stderr pipes, so it seems like they should.

src/supervisor/Supervisor.cc
```diff
@@ -819,6 +966,112 @@ std::optional<SupervisedNode> Stem::Revive() {
...
+        LogError("failed to create stdout pipe for node '%s'", node->Name().data());
+        return false;
+    }
+
+    SetHandleInformation(child_stdout_read, HANDLE_FLAG_INHERIT, 0);
+
+    HANDLE child_stderr_read, child_stderr_write;
+
+    if ( ! CreatePipe(&child_stderr_read, &child_stderr_write, &sa, 0) ) {
+        CloseHandle(child_stdout_read);
```
timwoj: Same comment here about `UniqueHandle` and cleaning these up.

timwoj (changes requested):
A couple more comments related to the Supervisor code.

src/supervisor/Supervisor.cc
mamaorha: the part where you opened stdout/stderr is now wrapped in "#ifndef _MSC_VER", the stem runs in a thread in the same procecss (we dont have fork, cant share resources and there wasnt really a need to separate to a diff process) - while we kept the process opening on "spawn" as we need the resiliency

src/input/readers/raw/Raw.cc
```diff
@@ -35,6 +35,19 @@ extern "C" {
 using zeek::threading::Field;
 using zeek::threading::Value;
 
+#ifdef _MSC_VER
+namespace {
+struct WinHandleDeleter {
```
timwoj: This could move into src/util-types.h. There's already two uses of it, and I can imagine there being others.

timwoj (changes requested):
One more thing, then :shipit:

src/input/readers/raw/Raw.cc
mamaorha: util-types is widely used and it will require me to add "windows.h" in it and the order is fragile due to unistd.h ordering constraint. i will extract it into a winHandle.h - is that ok?

timwoj: Oh, good point. Extracting it is fine.

PR #5248, opened 2026-03-03
Switch default check pragma for sqlite to quick_check

### LINKED ISSUE #5247: storage/sqlite: Don't run PRAGMA integrity_check on open
We've observed the [integrity check](https://sqlite.org/pragma.html#pragma_integrity_check) when opening a database can take minutes depending on the size of the database. There's also a [quick version](https://sqlite.org/pragma.html#pragma_quick_check).

Probably should backport to 8.1/8.0.

timwoj:
This PR changes the default check pragma used when opening SQLite databases from `integrity_check` to `quick_check`. We observed on a system with 40+ Zeek processes with each process opening their own separate SQLite file that `integrity_check` can cause enough disk thrashing to completely bog down all of the processes. `quick_check` does most of the same checks, but excludes a few purely for performance reasons. Docs with the differences are linked below.

[1] https://sqlite.org/pragma.html#pragma_integrity_check
[2] https://sqlite.org/pragma.html#pragma_quick_check

This PR also moves all of the storage btests into directories based on backend instead of prefix-naming each btest. This will be more important as we add more backends.

Fixes #5247


timwoj:
At the very least 8.0, for sure.

PR #5252, opened 2026-03-05
Add is_valid_base64() BiF

### LINKED ISSUE #5242: is_valid_base64() utility function
The `decode_base64{,_conn}` BiFs will generate reporter errors if the string passed in isn't valid Base64. It would be helpful to provide a utility predicate so callers can avoid such errors if they want to use different logic for invalid strings.

martincmr:
Returns T/F instead of generating reporter errors. Fixes #5242

### DIFF
--- a/src/Base64.cc
+++ b/src/Base64.cc
@@ -260,11 +260,10 @@ bool is_valid_base64(const String* s, const String* a) {
         return false;
 
     int buf_len = static_cast<int>((s->Len() + 3) / 4) * 3 + 1;
-    int rlen = buf_len;
-    char* rbuf = new char[rlen];
+    char* rbuf = new char[buf_len];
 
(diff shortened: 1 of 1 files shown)


src/Base64.cc
```diff
@@ -253,4 +255,20 @@ String* encode_base64(const String* s, const String* a, Connection* conn) {
     return new String(true, reinterpret_cast<u_char*>(outbuf), outlen);
 }
 
+bool is_valid_base64(const String* s, const String* a) {
+    if ( a && a->Len() != 0 && a->Len() != 64 )
+        return false;
+
+    int buf_len = static_cast<int>((s->Len() + 3) / 4) * 3 + 1;
+    int rlen = buf_len;
```
timwoj: This variable isn't really necessary. Just use `buf_len` directly.

timwoj (changes requested):
Just one minor thing. Looks ok otherwise.

PR #5253, opened 2026-03-05
Rename all .trace files to .pcap in our Traces library

timwoj:
This has always bothered me a little bit. This renames all .trace files to .pcap, and also renames all pcapng files to .pcapng. The latter is based on what `file` reports for the files on macOS.


0xxon:
So - to prefix this - I agree, this has always annoyed me.

But - we do install/expose our pcap libraries for packages to use - so this could cause some downstream breakage.

timwoj:
> But - we do install/expose our pcap libraries for packages to use - so this could cause some downstream breakage.

@ckreibich and I both mentioned this when we were talking about it, but decided it was probably not a huge thing. I figured I'd open the PR anyways.

ckreibich:
Yeah, I think this worth doing regardless of that risk, but let's make sure to call it out in the release notes. I've tagged the PR. If you want to add a NEWS snippet at this time Tim that's of course fine too, up to you.

timwoj:
Renamed that one file and added a snippet to NEWS. I also rebased since there were some conflicts caused by the recent Windows changes.

PR #5255, opened 2026-03-05
Disable zeekygen warnings by default

pbcullen:
## Summary

This PR modifies Zeek to disable zeekygen warnings by default, with an environment variable to enable them when needed. This improves the developer experience by reducing noise during normal script development while still allowing documentation authors to enable warnings when working on zeekygen documentation.

## Motivation

Zeekygen warnings can be noisy during normal script development and are typically not useful unless actively working on documentation. Users have reported that these warnings clutter their output and make it harder to see relevant information.

## Changes

- Add `ZEEK_ENABLE_ZEEKYGEN_WARNINGS` environment variable to control warning output
- Warnings are now **disabled by default**
- When `ZEEK_ENABLE_ZEEKYGEN_WARNINGS` is set (to any value), warnings are enabled
- Converted `DbgAndWarn()` and `WarnMissingScript()` from static functions to member functions for proper access control
- Added test coverage in `testing/btest/doc/zeekygen/warning-control.zeek` for both enabled and disabled states

## Behavior

**Before this change:**
- Zeekygen warnings always displayed (noisy)

**After this change:**
- Default: No warnings (quiet, better developer experience)
- With `ZEEK_ENABLE_ZEEKYGEN_WARNINGS=1`: Warnings enabled (for documentation work)

## Testing

New test case `doc/zeekygen/warning-control.zeek` validates:
- Default behavior produces no warnings
- Setting `ZEEK_ENABLE_ZEEKYGEN_WARNINGS` enables warnings as expected

## Backwards Compatibility

This change is backwards compatible. Users who want the old behavior (warnings enabled) can simply set the environment variable.


ckreibich:
I agree — thanks Peter. We should probably set that environment variable in `update-zeekygen-docs.sh` then?

ckreibich:
@pbcullen I've merged this with a small tweak to our docs generator to set `ZEEK_ENABLE_ZEEKYGEN_WARNINGS=1`. Thanks!

PR #5256, opened 2026-03-07
cluster/backend/zeromq: Set ipv6=F by default

### LINKED ISSUE #5251: Cluster fails to start: Zeek sets ZeroMQ IPv6 option on IPv4 socket
On systems like OpenBSD lacking dual-family sockets where `127.0.0.1` results in `AF_INET` sockets without `AF_INET6` support, `zeekctl deploy` is broken by default as
https://github.com/zeek/zeek/blob/372b7ff5dc0b217072b6d70aa09bb383b6584c10/src/cluster/backend/zeromq/ZeroMQ-Proxy.cc#L51
causes
https://github.com/zeek/zeek/blob/372b7ff5dc0b217072b6d70aa09bb383b6584c10/src/cluster/backend/zeromq/ZeroMQ-Proxy.cc#L66
to fail:

```
error: ZeroMQ: Failed to bind xpub socket tcp://127.0.0.1:27760: Operation not supported by device (19)
fatal error in /usr/local/share/zeek/policy/frameworks/cluster/backend/zeromq/main.zeek, line 441: Failed to spawn ZeroMQ proxy thread
```

With `*sub_endpoint` set to proper IPv6 addresses for which ZeroMQ's option actually applies the cluster does start:
https://github.com/zeek/zeek/blob/0d9aa5f321129f27fab9d7c9b9a5c5ebde584f20/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek#L73
but that prevents users from using (default) IPv4 localhost.

Rebuilding with L51 dropped makes it work for both address families without any other changes.

I have not looked any further yet;  this was noticed with the 8.1.1 release on OpenBSD 7.8-current through its binary package, for which I'll now add a patch dropping L51 as interim fix to unbreak cluster mode.

awelzel:
On systems without support for dual-stack sockets, the default of ipv6=T
results in a broken default as binding the ZeroMQ XPUB/XSUB and PULL sockets
fails. The ZeekControl ZeroMQ plugin is adapted separately to set the option
true if any of the node addresses is an IPv6 address.

This was reported by @klemensn while testing on OpenBSD. Thanks!

Closes #5251

### DIFF
--- a/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
+++ b/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
@@ -227,14 +227,17 @@ export {
 	## for more details.
 	const log_rcvbuf: int = -1 &redef;
 
-	## Set ZMQ_IPV6 option.
+	## Sets the ZMQ_IPV6 option on ZeroMQ contexts created by Zeek.
 	##
-	## The ZeroMQ library has IPv6 support in ZeroMQ. For Zeek we enable it
-	## unconditionally such that listening or connecting  with IPv6 just works.
+	## By default, IPv6 support on the ZeroMQ context is disabled. If you're
+	## using IPv6 addresses for any of the endpoint options, or you're using
+	## Zeek in a dual-stack environment and want to listen on IPv6 and IPv4
+	## addresses at the same time, set this option to ``T``. If you're
+	## deploying Zeek using ZeekControl, this will happen automatically.
 	##
-	## See ZeroMQ's `ZMQ_IPV6 documentation <http://api.zeromq.org/4-2:zmq-setsockopt#toc23>`_
+	## See ZeroMQ's `ZMQ_IPV6 documentation <http://api.zeromq.org/4-2:zmq-ctx-set#toc9>`_
 	## for more details.
-	const ipv6 = T &redef;
+	const ipv6 = F &redef;
 
 	## Do not silently drop messages if high-water-mark is reached.
 	##
--- a/testing/btest/cluster/zeromq/ipv6-error.zeek
+++ b/testing/btest/cluster/zeromq/ipv6-error.zeek
@@ -20,8 +20,8 @@
 # @TEST-EXEC: TEST_DIFF_CANONIFIER="sed -E 's/line [0-9]+/line xxx/g' | sed -E 's,^error: ZeroMQ: Failed to bind ([^ ]+) socket tcp://\[::1\]:[0-9]+:.*$,error: ZeroMQ: Failed to bind \1 socket...,g' | $SCRIPTS/diff-remove-abspath" btest-diff logger/.stderr
 
 # @TEST-START-FILE common.zeek
-@load frameworks/cluster/backend/zeromq
-# Explicitly disable ipv6 support to provoke errors.
-redef Cluster::Backend::ZeroMQ::ipv6 = F;
 @load ./zeromq-test-bootstrap
+
+# Explicitly disable IPv6 support to provoke errors.
+redef Cluster::Backend::ZeroMQ::ipv6 = F;
 # @TEST-END-FILE
--- a/testing/btest/cluster/zeromq/ipv6-logging.zeek
+++ b/testing/btest/cluster/zeromq/ipv6-logging.zeek
@@ -32,6 +32,9 @@
 # @TEST-START-FILE common.zeek
 @load ./zeromq-test-bootstrap
 
+# Enable IPv6, disabled by default.
+redef Cluster::Backend::ZeroMQ::ipv6 = T;
+
 redef Log::default_rotation_interval = 0sec;
 redef Log::flush_interval = 0.01sec;
 


klemensn:
> @klemensn reports that enabling IPv6 by default doesn't work on OpenBSD. There's also no real reason to enable it by default given we use IPv4 addresses for a default or standalone configuration.

Thanks for fixing things, but I don't quite agree with this wording as it sounds like IPv6 per-se on OpenBSD is the problem.

Switching endpoint addresses from `127.0.0.1` to `::1` on OpenBSD does enable IPv6, makes ZeroMQ's option being set without problems and works without problems.

The problem is the assumption that all systems Zeek runs on do support dual-socket families, i.e. that `ZMQ_IPV6` on a socket created through an IPv4 literal address (however conflicting that may sound) successfully enables IPv6 for besaid socket.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -229,12 +229,14 @@ export {
...
-	## unconditionally such that listening or connecting  with IPv6 just works.
+	## The ZeroMQ library has IPv6 support in ZeroMQ. By default, IPv6 support
+	## on the ZeroMQ sockets is disabled. If you're using IPv6 addresses for
+	## any of the endpoint options you'll likely want to set this option
+	## to ``T``.
 	##
 	## See ZeroMQ's `ZMQ_IPV6 documentation <http://api.zeromq.org/4-2:zmq-setsockopt#toc23>`_
 	## for more details.
-	const ipv6 = T &redef;
+	const ipv6 = F &redef;
```
klemensn, line 240: But if you do use `*pub_endpoint = "[::1]:..."` and do have `ipv6 = F` Zeek will still use IPv6 just fine.

Or does a cluster really not come up on Linux if you do that?

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
```diff
@@ -229,12 +229,14 @@ export {
...
 	## Set ZMQ_IPV6 option.
 	##
-	## The ZeroMQ library has IPv6 support in ZeroMQ. For Zeek we enable it
-	## unconditionally such that listening or connecting  with IPv6 just works.
+	## The ZeroMQ library has IPv6 support in ZeroMQ. By default, IPv6 support
+	## on the ZeroMQ sockets is disabled. If you're using IPv6 addresses for
+	## any of the endpoint options you'll likely want to set this option
+	## to ``T``.
 	##
 	## See ZeroMQ's `ZMQ_IPV6 documentation <http://api.zeromq.org/4-2:zmq-setsockopt#toc23>`_
```
klemensn: Again, to quote from https://libzmq.readthedocs.io/en/zeromq4-x/zmq_setsockopt.html (`api.zeromq.org` is geo-blocking me):

> ZMQ_IPV6: Enable IPv6 on socket
>
> Set the IPv6 option for the socket. A value of 1 means IPv6 is enabled on the socket, while 0 means the socket will use only IPv4. When IPv6 is enabled the socket will connect to, or accept connections from, both IPv4 and IPv6 hosts.

I think Zeek's code would benefit from clearer wording wrt. what this knob actually does:  control dual-socket functionality, which appears to be default-off at least for IPv4 sockets on systems supporting it.

---

Is there any real benefit to it besides the convenience of having both `127.0.0.1` and `::1` be reachable?

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: > But if you do use *pub_endpoint = "[::1]:..." and do have ipv6 = F Zeek will still use IPv6 just fine.

Oh, interesting. At least the binding / listening part doesn't work on my end with IPv6 endpoints if this option is `F` and it needs setting to `T`. There's actually also a [test for this behavior](https://github.com/zeek/zeek/blob/master/testing/btest/cluster/zeromq/ipv6-error.zeek).

```
==== stderr.log
error: ZeroMQ: Failed to bind xpub socket tcp://[::1]:27760: No such device (19)
fatal error in /opt/zeek-dev-prod/share/zeek/policy/frameworks/cluster/backend/zeromq/main.zeek, line 443: Failed to spawn ZeroMQ proxy thread
```

So I'd stick with the comment of needing to turn this to `T` if the endpoints contain IPv6. If you want to dig deeper, we'd certainly be open to clarify the behavior.

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
awelzel: > I think Zeek's code would benefit from clearer wording wrt. what this knob actually does: control dual-socket functionality, which appears to be default-off at least for IPv4 sockets on systems supporting it.

I'm okay just pointing at the ZeroMQ docs here, actually. As a user using ZeekControl, you shouldn't worry all that much (and if you do, the ZeroMQ docs will likely be the better reference).

I also just realized we're setting this option on the ZeroMQ context, not the socket (though internally I suppose that'll just result in the sockets using to option automatically), so I'll fix the wording.

https://libzmq.readthedocs.io/en/zeromq4-x/zmq_ctx_set.html

awelzel:
> Thanks for fixing things, but I don't quite agree with this wording as it sounds like IPv6 per-se on OpenBSD is the problem.
> 

Good point. Thanks. I didn't want to make this specifically about OpenBSD. Let me attempt to rewrite the commit message and PR description.

> Switching endpoint addresses from `127.0.0.1` to `::1` on OpenBSD does enable IPv6, makes ZeroMQ's option being set without problems and works without problems.
> 
> The problem is the assumption that all systems Zeek runs on do support dual-socket families, i.e. that `ZMQ_IPV6` on a socket created through an IPv4 literal address (whoever conflicting that may sound) successfully enables IPv6 for besaid socket.

I'll use that as in the description and mention that you have observed this on OpenBSD. Sounds good?

scripts/policy/frameworks/cluster/backend/zeromq/main.zeek
klemensn: What about defaulting both addresses and option to IPv6/`F`?

IIUC, this should work just fine on Linux whilst providing backward IPv4 connectivity by default due to the option enabling dual-socket functionality.

On OpenBSD, it'll just switch to plain IPv6 on ::1, which, by default, works equally well, so there'd be no breaking change, imho.

We need more IPv6 and less IPv4, not the other way around ;-)

awelzel: > What about defaulting both addresses and option to IPv6/F?

I presume you meant `T` - I like the idea, but would conservatively stay with `ipv6=F`. Also, in retrospect, that's the default in `libzmq` and so makes more sense than the `T`.

> We need more IPv6 and less IPv4, not the other way around ;-)

Ha! Ambitious :-) With ZeroMQ on a single system, I'd actually much rather use `ipc://` for the ZeroMQ sockets and avoid the whole IPv4/IPv6/TCP/firewall when everything is on a single system anyhow. Sticking with `tcp://` here was mostly to not be completely deviate here compared to Broker, and sticking with IPv4 to be conservative.

You're obviously free to run this in IPv6-only mode in your environment :+1:

PR #5257, opened 2026-03-08
cmake: use absolute source paths in flex_target() calls to fix DWARF debug info

dogukancagatay:
The five flex_target() calls in `src/CMakeLists.txt`, `tools/bifcl/CMakeLists.txt`, and `tools/binpac/src/CMakeLists.txt` previously passed the `.l/.ll` input file as a bare filename. When CMake invokes flex from an out-of-source build directory, flex writes `#line 1 "rule-scan.l"` (a relative path) into the generated C++ output. GCC records this as `<build-dir>/rule-scan.l` in DWARF debug information, a path that does not exist in the source tree.

This silently breaks the zeek-debugsource RPM on Fedora/RHEL (and equivalent dbgsym packages on Debian-based distributions): cpio cannot stat the non-existent paths and omits the .l grammar files from the debugsource package, making source-level debugging of scanner code impossible.

Changing the input to `${CMAKE_CURRENT_SOURCE_DIR}/rule-scan.l` (absolute path) causes flex to emit the full path in #line directives. GCC records the correct source path in DWARF, debugedit can rewrite it, and the debugsource package is complete.

This change has no effect on the generated scanner logic, the Zeek API, or build performance. It only affects the string written into #line directives and the resulting DWARF debug info.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,8 +55,8 @@ replace_yy_prefix_target(${CMAKE_CURRENT_BINARY_DIR}/rup.cc
                          ${CMAKE_CURRENT_BINARY_DIR}/rule-parse.cc rules_ rules_)
 replace_yy_prefix_target(${CMAKE_CURRENT_BINARY_DIR}/rup.h ${CMAKE_CURRENT_BINARY_DIR}/rule-parse.h
                          rules_ rules_)
-flex_target(RuleScanner rule-scan.l ${CMAKE_CURRENT_BINARY_DIR}/rule-scan.cc
-            COMPILE_FLAGS "-Prules_")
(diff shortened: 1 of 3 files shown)


awelzel:
@dogukancagatay - lgtm. Mind running pre-commit / cmake-format and repush? Otherwise can do during the merge.

awelzel (approved):
Doesn't seem controversial to me, so will merge.

@dogukancagatay - there's three more `flex_target()` invocations in the github.com/zeek/spicy project that's included as a submodule. Care to look at those, too, or are those not an issue in your env?

dogukancagatay:
I came across the issue when I was trying to update the Fedora package. I'll take a look at spicy. @awelzel

PR #5258, opened 2026-03-09
Update basics.rst to include 'string' data type

AbduSami-bK:
Add 'string' to the list of data types in the basics documentation.

### DIFF
--- a/doc/scripting/basics.rst
+++ b/doc/scripting/basics.rst
@@ -683,6 +683,9 @@ reinventing the wheel.
   * - :zeek:see:`pattern`
     - regular expression
 
+  * - :zeek:see:`string`
+    - text or binary data
+
 Sets
 ~~~~
 


doc/scripting/basics.rst
awelzel, line 643: Thanks for catching!

Would you mind moving string first in the table and update this to "first five", then the counting works again :-)

There's also https://github.com/zeek/zeek/pull/4928 pending - i wonder if you should target that branch instead. @evantypanski , @ckreibich ?

evantypanski: I don't know if this change is particularly necessary, since the new tutorial (in that PR) mentions Strings as it does other "basic" types [here](https://docs.zeek.org/en/new-tutorial/tutorial/scripting/basics.html#standard-types)

this section is about to be completely rewritten

awelzel: Hmm, with that I'd suggest to close this PR. @AbduSami-bK okay with you?

ckreibich: I agree, especially now that the tutorial has landed. But thank you, @AbduSami-bK! You can check out the new docs layout [here](https://docs.zeek.org/en/master/)... — feedback very welcome.

PR #5261, opened 2026-03-09
ssl: Fix pre_shared_key logic in tls-handshake-analyzer

awelzel:
It seems there was some confusion about client and server. The OfferedPsks are parsed from a ClientHello record, so seems reasonable to rename from proc_pre_shared_key_server_hello to proc_pre_shared_key_client_hello and vice-versa to align with the events that are actually enqueued. Also add a test.

Reported by @cccs-graeme on Slack.


ckreibich (approved):
Nice catch. Might be good to double-check the new test with the Spicy-based TLS analyzer, though you probably already have. :-)

awelzel:
> double-check the new test with the Spicy-based TLS analyzer, though you probably already have. :-)

Hadn't, but started the CI job and that was green, so the Spicy version didn't have the same issue.

0xxon:
Thank you - that is a great catch :)

PR #5263, opened 2026-03-09
Add implementation and version fields to plugin/analyzer introspection

MP-Corelight:
Extend PluginComponent record with implementation (C++/Spicy) and version fields.

Add Analyzer::implementation() BiF and wrapper to query an analyzer's implementation type by tag, using the same three-manager lookup pattern as Analyzer::name().

These changes provide some additional information that's helpful when testing plugin/analyzer deployments.

### DIFF
--- a/src/zeek.bif
+++ b/src/zeek.bif
@@ -6118,9 +6118,7 @@ function get_plugin_components%(category: string%): plugin_component_vec
 		auto ver = plugin->Version();
 		std::string version_str;
 		if ( ver )
-			version_str = std::to_string(ver.major) + "." +
-			              std::to_string(ver.minor) + "." +
-			              std::to_string(ver.patch);
+			version_str = util::fmt("%d.%d.%d", ver.major, ver.minor, ver.patch);
 
 		for ( const auto* component : plugin->Components() )
 			{


src/zeek.bif
```diff
@@ -6112,19 +6113,43 @@ function get_plugin_components%(category: string%): plugin_component_vec
...
-	for ( auto* component : all_components )
+	for ( const auto* plugin : plugin_mgr->ActivePlugins() )
 		{
-		if ( component->Type() == type )
+		auto ver = plugin->Version();
+		std::string version_str;
+		if ( ver )
+			version_str = std::to_string(ver.major) + "." +
+			              std::to_string(ver.minor) + "." +
+			              std::to_string(ver.patch);
```
timwoj:
```suggestion
version_str = util::fmt("%d.%d.%d", ver.major, ver.minor, ver.patch);
```

src/zeek.bif
```diff
@@ -6112,19 +6113,43 @@ function get_plugin_components%(category: string%): plugin_component_vec
...
+				{
+				for ( const auto* c : zeek::spicy_mgr->Components() )
+					{
+					if ( c == component )
+						{
+						impl = "Spicy";
+						break;
+						}
+					}
+				}
```
timwoj: Could this be extracted somewhere to avoid the duplication here and in the `__implementation` BIF? Maybe into `Component` itself?

src/zeek.bif
MP-Corelight: Done.

src/zeek.bif
MP-Corelight: Done.

src/plugin/Component.h
```diff
@@ -132,6 +132,13 @@ class Component {
      */
     bool Enabled() const { return enabled; }
 
+    /**
+     * Returns the implementation type of this component as a string.
+     * Currently returns "Spicy" for Spicy-based components and "C++"
+     * for all others.
+     */
+    std::string Implementation() const;
+
```
awelzel, line 141: Hmm, hmm. I'm a bit wondering if this should've become a member on the to-level Component class and have a new parameter to the component constructor defaulting to `nullopt` and/or a setter. Then changing the three analyzer components to default to "C++" or within `src/spicy`, passing/setting "Spicy" explicitly, so we could avoid interrogating the `spicy_mgr`.

That would also allow external plugins to inject their own string if we ever do something with "Rust" or "
"Kaitai Struct" analyzers out of tree.

cc @rsmmr for the plugin/component infrastructure.

rsmmr: My 2c:

- I would have wanted to learn more about use case here. From an infrastructure perspective, I'd actually have been reluctant to add this at all; it shouldn't matter to script-land how the analyzer is  implemented. I'm afraid this will lead to script code like `if ( Analyzer::implementation() == "Spicy"  ) do-something-different`.

- I would say that the question about plugins extending the set of options further, supports that gut feeling of "we shouldn't need this". What does it matter if it's Rust?

- If we need to have it, I'd use enums and I'd also distinguish between "pure C++" and "BinPAC".

bbannier: The default is wrong for an existing component implemented in 80% Rust and 20% C++ ("C++" instead of e.g., "Rust"), and the semantics of what I should return in a custom `Implementation` aren't clear to me (the comment just documents the defaults). Should I also call out that I include 50 lines of CMake? 

IMO unstructured information should go into `DoDescribe`, and I'd put implementation language into that bucket as well. I'd be in favor of removing this from the component. If there is a use case we could expose a API to query `spicy_mgr`.

ckreibich:
I agree that this isn't the way to provide the implementation, but I'm sympathetic to being able to query additional detail about analyzers & plugins from scriptland. It's handy to have for complex installations.

The downside of the `DoDescribe` idea is that it requires all components to feature similar implementation-level detail, which I'm afraid is unlikely to happen.

I'm with Robin regarding better understanding the use-case — perhaps we can figure out the right way to do this from there.

PR #5264, opened 2026-03-09
Add postgres storage plugin

timwoj:
The initial bits of this were generated by Claude, based on the existing SQLite backend. It's gone through a bunch of manual cleanup, including a lot of code de-duplication and error handling changes. The btests were all added manually based on the existing Redis tests. I added postgres to a few of the CI hosts for test coverage.

I'm not the biggest fan of having completely rebuild the database files on each invocation of the btests. Unfortunately the database files are version-specific so they can't be created and stored as part of the repo.


timwoj:
I had to cancel the CI runs because postgres won't run as root in the CI environment. I'll figure out a workaround for that.

awelzel:
Could we consider keeping this externally as a plugin like `zeek/zeek-backend-storage-postgres`  or so.

I think by now we have sqlite, postgres, redis and nats with three of them builtin we might be overdoing it.

timwoj:
Closing this and creating a package out of it eventually

ckreibich:
I'm super late here — but fwiw, if we have a reasonable, optional, modest-scale feature this close to going in, I think we should go ahead. Right now the worst happened, namely the functionality is still unavailable to users.

Best way to avoid these conversations is clarity among us (via issues/discussion) before the PR!

PR #5268, opened 2026-03-10
[windows] enable spicy and fix build errors

mamaorha:


src/CMakeLists.txt
```diff
@@ -604,7 +610,45 @@ if (TARGET zeek_exe)
     set_target_properties(zeek_exe PROPERTIES ENABLE_EXPORTS TRUE)
 
     if (MSVC)
-        set(WINDOWS_EXPORT_ALL_SYMBOLS ON)
+        # WINDOWS_EXPORT_ALL_SYMBOLS exceeds the COFF 65535 symbol limit.
+        # Instead, generate a targeted DEF file exporting HILTI/Spicy runtime
+        # symbols plus transitive dependencies resolved from zeek_objs.
```
ckreibich, line 615: Given that this block solves a pretty specific problem it seems doubtful that we'll use `generate-jit-export-def.py` elsewhere, no? Our cmake repo has no standalone tooling of that kind — it has shell scripts, but they're used by cmake scripts that live in that repo. So I don't think we should "park" the script over there (via zeek/cmake#145).

One similar script, used only by this `CMakeLists.txt`, is `make_dbg_constants.py`, which lives right here in `src`. I think I'd prefer that ... what do you think @bbannier? We could potentially have a folder `src/buildtools` or some such down the road for such things.

bbannier: Agree it probably should be here. Spicy currently does not pull in zeek/cmake and looking at the windows PR over zeek/spicy we didn't seem to have needed this script there either.

mamaorha: moved the file into zeek src similar to make_dbg_constants.py. (updated the cmake pr & submodule commit)

src/cluster/PublishOnChangeState.cc
```diff
@@ -24,7 +24,6 @@
 #include "zeek/ZeekString.h"
 #include "zeek/broker/Data.h" // for data_to_val()
 #include "zeek/cluster/Backend.h"
-#include "zeek/types.bif.netvar_h"
```
mamaorha, line 27: already defined in the h file

src/cluster/PublishOnChangeState.h
```diff
@@ -12,7 +12,7 @@
 
 #include "zeek/EventHandler.h"
 #include "zeek/IntrusivePtr.h"
-#include "zeek/types.bif.netvar_h"
+#include "types.bif.netvar_h"
```
mamaorha, line 16: On MSVC, #pragma once does not deduplicate headers reached through symlinks, so use the unprefixed path (types.bif.netvar_h is also pulled in via NetVar.h without the zeek/ prefix).

src/spicy/spicyz/glue-compiler.h
```diff
@@ -308,8 +308,8 @@ template<>
 struct hash<zeek::spicy::glue::Event> {
     std::size_t operator()(const zeek::spicy::glue::Event& e) {
         // We only hash enough information here to unique identify the event.
-        return hilti::rt::hashCombine(std::hash<std::string>()(e.file), std::hash<std::string>()(e.name),
-                                      std::hash<std::string>()(e.path), std::hash<std::string>()(e.location));
+        return hilti::rt::hashCombine(std::hash<std::string>()(e.file.generic_string()), std::hash<hilti::ID>()(e.name),
```
mamaorha, line 311: The old code relied on implicit conversions to std::string that don't exist on MSVC

src/File.cc
```diff
@@ -26,7 +26,7 @@
 #include "zeek/Reporter.h"
 #include "zeek/RunState.h"
 #include "zeek/Type.h"
-#include "zeek/event.bif.netvar_h"
+#include "event.bif.netvar_h"
```
mamaorha, line 30: On MSVC, #pragma once does not deduplicate headers reached through symlinks, so use the unprefixed path (types.bif.netvar_h is also pulled in via NetVar.h without the zeek/ prefix).

src/spicy/runtime-support.cc
```diff
@@ -1227,320 +1256,339 @@ inline void setRecordField(RecordVal* rval, const IntrusivePtr<RecordType>& rtyp
 }
 
 ValPtr rt::detail::to_val(const hilti::rt::type_info::Value& value, const TypePtr& target) {
```
mamaorha, line 1258: this one looks like huge change but its mostly "formatting" due to the wrapping of the function with try-catch on msvc

src/File.cc
timwoj: I fixed this with https://github.com/zeek/zeek/pull/5322. The includes in NetVar.h should have had the full path all along, we just missed it.

PR #5270, opened 2026-03-11
Introduce &publish_on_change attribute

awelzel:
This is a result of #5020, #5040 and direct conversations with @vpax, @rsmmr and others.

It probably makes most sense to review the documentation commit first for the idea.

As a minimal example, the following makes Zeek publish all newly inserted entries in the `endpoints` set to all nodes in a cluster via topic `zeek/table/endpoints`. Other nodes will transparently "apply" changes published by other nodes. Using more elaborate `$changes` selection is possible (REMOVED, CHANGED), but comes with consistency caveats.

```
global endpoints: set[addr, addr] &write_expire=60sec &publish_on_change=[
    $changes=set(TABLE_ELEMENT_NEW),
];
```

Essentially, this presents an alternative to the Broker-specific `&backend` and `&broker_store` attributes that works for any cluster backend by disseminating table changes as Zeek remote events. This also enables any external WebSocket clients to observe table changes by subscribing to the appropriate topic.

### DIFF
--- a/doc/reference/zeekscript/attributes.rst
+++ b/doc/reference/zeekscript/attributes.rst
@@ -722,6 +722,8 @@ named ``ip``, ``srcp``, ``proto``, and ``msg``:
 &backend
 --------
 
+.. deprecated:: 8.1 This attribute is Broker specific, use :zeek:attr:`&publish_on_change` or the :ref:`Storage Framework <framework-storage>` instead.
+
 Used for persisting tables/sets and/or synchronizing them over a cluster.
 
 This attribute binds a table to a Broker store. Changes to the table
@@ -743,6 +745,8 @@ example, to bind a table to a memory-backed Broker store, use:
 
 &broker_store
 -------------
+
+.. deprecated:: 8.1 This attribute is Broker specific, use :zeek:attr:`&publish_on_change` or the :ref:`Storage Framework <framework-storage>` instead.
 
 This attribute is similar to :zeek:attr:`&backend` in allowing a Zeek table to
 bind to a Broker store. It differs from :zeek:attr:`&backend` as this attribute


awelzel:
I've re-pushed after a bit more docs cleanup and marked it ready for review.

Link to the commit with only the docs about this attribute: https://github.com/zeek/zeek/pull/5270/changes/6befcd55c152b7fbeaff157297a5470647c05d46

src/cluster/PublishOnChangeState.cc
```diff
@@ -0,0 +1,664 @@
...
+        tci_index_offset = field_offset_or_fatal(table_change_info, "index");
+        tci_value_offset = field_offset_or_fatal(table_change_info, "value");
+        tci_previous_value_offset = field_offset_or_fatal(table_change_info, "previous_value");
+
+        table_change_infos = zeek::id::find_type<zeek::VectorType>("Cluster::TableChangeInfos");
+        if ( ! table_change_infos )
+            zeek::reporter->FatalError("Cluster::TableChangeInfos not found");
+    }
+
+    zeek::StringValPtr BuildString(const std::string& s) const { return zeek::make_intrusive<zeek::StringVal>(s); }
```
timwoj: This function feels unnecessary. Just call `make_intrusive`.

src/cluster/PublishOnChangeState.cc
```diff
@@ -0,0 +1,664 @@
...
+// Storage for static variables.
+std::optional<std::string> PublishOnChangeState::forward_topic;
+EventHandlerPtr PublishOnChangeState::eh_table_change_infos;
+EventHandlerPtr PublishOnChangeState::eh_forward_table_change_infos;
+StringValPtr PublishOnChangeState::local_node_id;
+
+PublishOnChangeState::PublishOnChangeState(StringValPtr identifier, TableVal* table_val, uint8_t change_mask,
+                                           std::optional<std::string> topic, FuncPtr topic_func, size_t max_batch_size,
+                                           double max_batch_delay)
+    : identifier(identifier),
```
timwoj: A move here would avoid an extra ref/deref.

src/cluster/PublishOnChangeState.cc
```diff
@@ -0,0 +1,763 @@
...
+    const auto& topic_effective = forward_topic.has_value() ? forward_topic.value() : topic;
+
+    debug("publish event=%s topic=%s topic_effective=%s table_change_infos=%u", eh_effective->Name(), topic.c_str(),
+          topic_effective.c_str(), tcinfos->Size());
+
+    // The arguments for the Cluster::table_change_infos() event.
+    Args args{std::move(tcheader), std::move(tcinfos)};
+
+    // By convention, the forwarding event has the topic topic as the last parameter,
+    // so we can just append it here has a new StringVal.
```
timwoj:
```suggestion
    // By convention, the forwarding event has the topic as the last parameter,
    // so we can just append it here as a new StringVal.
```

src/cluster/PublishOnChangeState.cc
```diff
@@ -0,0 +1,763 @@
...
+            reporter->InternalError("topic set, but topic_changes not empty");
+
+        if ( changes )
+            PublishQueuedChanges(now, *topic, tcheader, changes);
+        else
+            reporter->InternalWarning("PublishOnChangeState: Weird: PublishQueuedChanges() without queud changes");
+
+        changes = helper.BuildTableChangeInfos();
+    }
+    else if ( topic_changes.size() > 0 ) {
```
timwoj:
```suggestion
    else if ( ! topic_changes.empty() ) {
```

src/cluster/PublishOnChangeState.h
```diff
@@ -0,0 +1,291 @@
...
+    // topic_changes members below are unused.
+    std::optional<std::string> topic;
+    VectorValPtr changes;
+
+    // If topic_func is not nil, for every change a new topic is
+    // determined by calling topic_func. Changes are queued per
+    // topic in the topic_changes map. The topic and changes
+    // members above are unused.
+    FuncPtr topic_func;
+    std::map<std::string, VectorValPtr, std::less<>> topic_changes;
```
timwoj, line 258: This name confused me for a bit trying to map it to what it actually was. Maybe reversing the name helps a little?

```suggestion
    std::map<std::string, VectorValPtr, std::less<>> change_topics;
```

timwoj (changes requested):
I just had a couple of nits and quibbling on naming that confused me.

src/cluster/PublishOnChangeState.cc
awelzel: :+1:

src/cluster/PublishOnChangeState.h
awelzel: No overly strong opinion, but `topic_changes` works better for me:
In my mind it was a map indexed by *topic* holding pending *changes*.

I'll leave as is. It's private and not part of the API and its use fairly confined. There's a bit of an explanation above it, too. If someone else is confused down the road or we find a bug because of the name, happy to revisit the name, not attached to it :+1:

timwoj (approved):
🚀

PR #5271, opened 2026-03-11
ci: Skip MacOS Sequoia for PRs

awelzel:


timwoj:
Is this just because we don't need two of the same platform? if that's the case, should we also skip freebsd14?

awelzel:
> Is this just because we don't need two of the same platform? if that's the case, should we also skip freebsd14?

Yeah, I was surprised that we had two MacOS tasks running.  Let me follow-up with FreeBSD 14. I missed that.

PR #5274, opened 2026-03-12
Redirect maintenance to reflect tutorial merge (#4928)

ckreibich:
This is based on the attached output from `doc-redirects scan`, plus a few additional cleanups.

A good way to test these in the future would be to set up a duplicate of the docs project in RTD.

[new-redirects.yml.txt](https://github.com/user-attachments/files/25924152/new-redirects.yml.txt)

### DIFF
--- a/doc/redirects.yml
+++ b/doc/redirects.yml
@@ -152,3 +152,5 @@
   to_url: /reference/zeekscript/event-semantics.html#tracing-events
 - from_url: /scripting/usage.html
   to_url: /advanced/scripting/usage.html
+- from_url: /troubleshooting.html
+  to_url: /advanced/troubleshooting.html


doc/redirects.yml
```diff
@@ -124,15 +118,39 @@
...
+- from_url: /script-reference/scripts.html
+  to_url: /reference/zeekscript/included.html
+- from_url: /script-reference/*
+  to_url: /reference/zeekscript/:splat
+- from_url: /scripting/basics.html
+  to_url: /tutorial/scripting/basics.html
+- from_url: /scripting/conn-id-ctx.html
+  to_url: /advanced/scripting/conn-id-ctx.html
+- from_url: /scripting/event-groups.html
+  to_url: /reference/zeekscript/event-semantics.html
```
evantypanski: technically this would be better as `/reference/zeekscript/event-semantics.html#event-groups`, but it's currently the first in that section so not a big deal

ckreibich: Thanks for reading closely! Fixed.

ckreibich:
These are now live on RTD.

PR #5278, opened 2026-03-13
postgresql: Fix violation on empty ParameterStatus value

awelzel:
This was found while monitoring the PostgreSQL traffic produced by the Greenhouse example app from the Loki fundamentals tutorial:

https://github.com/grafana/loki-fundamentals?tab=readme-ov-file#deploying-the-sample-application

The app uses an empty application name and that resulted in an analyzer violation:

    error while parsing PostgreSQL: failed to match regular expression (<...>/protocol/postgresql/postgresql.spicy:278:12-278:25)


testing/btest/Baseline/scripts.base.protocols.postgresql.greenhouse-app/postgresql.log
bbannier, line 1: That's a lot of baseline when we primarily here want to test that we can see an empty parameter status value.

WDYT about instead adding a handler for `PostgreSQL::parameter_status` which prints is inputs and baseline that?

```zeek
event PostgreSQL::parameter_status(c: connection, name: string, value: string)
	{
	print fmt("id: %s, name: '%s', value: '%s'", c$id, name, value);
	}
```
This still prints 22 lines where only a single one is interesting, but I'd argue that's easier to understand and maintain than the current full baseline of `postgresql.log`.

awelzel: > That's a lot of baseline when we primarily here want to test that we can see an empty parameter status value.

It's large, but was mostly also to have more integration/smoke testing coverage than just testing that single issue. It's not supposed to change, or if it does, we probably want to know about it.

Indeed, I was actually thinking if it'd be better for users if we filtered the BEGIN/COMMIT/ROLLBACK lines by default.

I'm on the side of keeping it.

src/analyzer/protocol/postgresql/postgresql.spicy
```diff
@@ -275,7 +275,7 @@ type BackendMessage = unit {
 type ParameterStatus = unit {
     name: /[-_\/A-Za-z0-9]+/ &requires=(|$$| > 0);
```
evantypanski, line 276: is there a world where we allow empty names too? I see there's a list of "hard-wired" names it can be, but it doesn't seem like it should be a protocol violation, more of just a weird. We know that it's null terminated anyway.

awelzel: Hmm, hmm, I'll leave as is. These come from the server and I don't think it should send zero-length keys. Not technically a violation, but might be a good indication that it's not the right protocol.

PR #5281, opened 2026-03-13
GH-5273: Pass the payload length from IP analyzer if available

### LINKED ISSUE #5273: Length passed from IP analyzer can be too large if Ethernet has a footer
I ran into this while working on the IGMP analyzer in #5095. I have some IGMPv3 packets where the Ethernet frame contains 10 extra bytes at the end for the footer (6 padding bytes, 4 FCS bytes). The full size of the "remaining data" after the Ethernet header gets passed down to the IP analyzer. Currently, the IP analyzer simply subtracts the size of the IP header from that and passes the remaining length along to the IGMP analyzer. This causes problems with the checksum calculation there because it thinks the IGMP part is 10 bytes larger than it should be. I worked around it by doing the checksum calculation later, but it's not ideal. The IP analyzer does know the correct size of the payload (via the IP header) and is available as `packet->ip_hdr->PayloadLen()`, but that's not what we're passing along. Should we be?

timwoj:
Payload length in an IP packet can be shorter than the actual packet, if one of the earlier protocols (e.g. Ethernet) has trailing data after the IP parts. The payload length in the header can also be zero, so this makes sure to check for that. The two tests that change are because of truncated data. For example, the outer packet in core.tunnels.ip-in-ip-version reports a payload length of 44, but next inner packet is only 40 bytes. Instead of reporting that the outer packet is wrong, we report a truncation of the inner packet instead.

Fixes #5273

### DIFF
--- a/src/packet_analysis/protocol/ip/IP.cc
+++ b/src/packet_analysis/protocol/ip/IP.cc
@@ -3,6 +3,7 @@
 #include "zeek/packet_analysis/protocol/ip/IP.h"
 
 #include <netinet/in.h>
+#include <algorithm>
 
 #include "zeek/Discard.h"
 #include "zeek/Event.h"
@@ -222,6 +223,9 @@ bool IPAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet) 
     // pointing at the reassembled data for a fragmented packet.
     data = packet->ip_hdr->Payload();
     len -= ip_hdr_len;
+
+    if ( packet->ip_hdr->PayloadLen() != 0 )
+        len = std::min(len, static_cast<size_t>(packet->ip_hdr->PayloadLen()));
 
     // Session analysis assumes that the header size stored in the packet does not include the IP
     // header size. There are two reasons for this: 1) Packet::ToRawPktHdrVal() wants to look at the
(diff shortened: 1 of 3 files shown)


src/packet_analysis/protocol/ip/IP.cc
```diff
@@ -223,6 +223,9 @@ bool IPAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet)
     data = packet->ip_hdr->Payload();
     len -= ip_hdr_len;
 
+    if ( packet->ip_hdr->PayloadLen() != 0 && len > packet->ip_hdr->PayloadLen() )
+        len = packet->ip_hdr->PayloadLen();
```
evantypanski: The condition made my brain jump around a little, I find this a lot more readable. Especially in the future when the context isn't here (so a comment wouldn't hurt either)

```suggestion
    if ( packet->ip_hdr->PayloadLen() != 0 )
        len = std::min(len, static_cast<std::size_t>(packet->ip_hdr->PayloadLen()));
```

or we can do `std::min` with `ForwardPacket`, like with the UDP analyzer, but that adds an extra complexity with the `0` check:

https://github.com/zeek/zeek/blob/7ec486942899c6904ed527d0a9d8f4c3648c2eac/src/packet_analysis/protocol/udp/UDP.cc#L201

PR #5291, opened 2026-03-23
Add SMTPS port to well-known ports

J-Gras:
See https://datatracker.ietf.org/doc/html/rfc8314#section-3.3


awelzel:
@J-Gras - this seems to miss an update to the `core.print-bpf-filters` baseline. Not sure that test is super useful for this PR, but it tickles a failure :man_shrugging:

PR #5295, opened 2026-03-26
Ignore checksums by default

timwoj:
This swaps Zeek's behavior to ignore checksums by default. The only thing here that I have qualms about is that I swapped the semantics of `-C` to enable checksumming instead of disabling it. If that's a problem to someone, I can move enabling it to a different argument and sunset the `-C` argument.

### DIFF
--- a/scripts/base/init-bare.zeek
+++ b/scripts/base/init-bare.zeek
@@ -1684,7 +1684,7 @@ const UDP_ACTIVE = 1;	##< Endpoint has sent something.
 ## feature is enabled.
 ## Note that the ``-C`` command-line option overrides the setting of this
 ## variable.
-const ignore_checksums = F &redef;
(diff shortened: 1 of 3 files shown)


vpax:
I think this is a bit trickier. In his original issue, @markoverholser noted that seasoned Zeek users could still see that there were checksum issues by the presence of `C`s in the history field. However, the logic in the analyzers doesn't even try to validate the checksum if the flag is set, so you won't get any history indications. There's also a related (but slightly distinct, as it's not about Principle Of Least Surprise) issue that not validating checksums shaves off some non-trivial processing that has to touch every payload byte, which in some situations might be a good tradeoff to make - so we might need a separate option meaning _No, really, don't validate the checksums_ ...

timwoj:
This was a pretty naive attempt. I'm going to close this and rework it separately.

PR #5298, opened 2026-03-27
Spicy: Update integration for new HILTI string implementation.

rsmmr:
This is a bit messy due to the need to explicitly convert between
string types. General rule of thumb is to (1) use `hilti::rt::String`
inside `src/spicy/` wherever strings come in from HILTI, (2) convert
them to `std::string` at the places where we call Zeek-side APIs that
take strings, and (3) occasionally use `std::string_view` in `src/spicy`
where it's safe and makes passing parameters easier.


timwoj:
I started up the `ubuntu24_04_spicy_head` build on CI since it's not part of the default set, and since that is the build currently failing on master.

rsmmr:
> I started up the `ubuntu24_04_spicy_head` build on CI 

Good call. :-)

Issue #426, opened 2019-06-20
Explore replacing Zeek's regex engine 


rsmmr:
Zeek is still using a custom regex engine that comes with some limitations. Let's revisit whether we could replace it with a more standard, external library. The answer is not clear unfortunately because we have some requirements that historically other engines had trouble meeting. Here's a stab at collecting our requirements for choosing a new engine:

- DFA-based (for performace)
- Support for extracting capture groups
- Stream API: Feed input in arbitrary chunks, with each result indicating whether regexp has (1) already matched; (2) has not yet matched, but might still match with more input; or (3) will never match anymore.
- Parallel matching for a set of regexes, with the result telling us which one matched. Parallel matching must scale in terms of memory and CPU to at least a similar size than our current engine (note that that one builds the DFAs incrementally on-demand to save memory).
- Option to prefer the earliest match over the longest (strictly speaking this is not a Bro requirement right now, but Spicy needs it; and I'd rather use the same library for both).

We have multiple use cases for regexes in Zeek, and need these features in various combinations (e.g., streaming set matching preferring the earliest match).

When I looked around years ago, I didn't see any library doing all this; which is why I ended up writing a [custom one](https://github.com/rsmmr/hilti/tree/master/libhilti/justrx) that [HILTI/Spicy](https://www.icir.org/hilti) is using. However, that one remains prototypish, and with its own challenges; wouldn't recommend using that for Zeek (nor eventually for Spicy).

0xxon:
https://rust-leipzig.github.io/regex/2017/03/28/comparison-of-regex-engines/ might be helpful (even if a bit dated).

timwoj:
Currently Hyperscan is winning in the hunt for a library, supporting everything that @rsmmr requested above plus being able to build regexes at compile time. I'm building a table of supported features at https://docs.google.com/spreadsheets/d/1B5kc1PgvOHF621AtrrbQ1LUPiaVBBWjx1Dc8_leTcmk/edit?usp=sharing

timwoj:
@rsmmr I have a couple of questions in the list above. For capture groups, do you want the ability to back-reference or just extract the captures from the resulting match? For earliest match vs longest match, are you just referring to lazy/non-greedy mode?

rsmmr:
> For capture groups, do you want the ability to back-reference or
> just extract the captures from the resulting match?

Just extract.

> For earliest match vs longest match, are you just referring to lazy/non-greedy mode?

Yes, I think so. 

[...]

PR #5300, opened 2026-03-28
Rust regex backend

sethhall:
After dinner the other night, I spent some time and swapped out Zeek's regular expression support with Rust.  This branch is obviously super aggressive.  It completely removes Zeek's regular expression engine and it now requires Rust to build Zeek.  

You can refer to some of the docs in this branch to see how some of the somewhat odd behaviors in Zeek's regex engine were brought into this implementation (including global matcher DFA state caching!).  We may want to have a discussion if some of these features are still wanted/needed in Zeek?  There is some weird thing that really stuck out to me about quoted strings in regexes that I thought was odd.

As you all can guess, this was AI written but I spent quite a bit of time on it.

### DIFF
--- a/rust/CMakeLists.txt
+++ b/rust/CMakeLists.txt
@@ -1,6 +1,9 @@
 # TODO(bbannier): We usually vendor CMake dependencies.
 include(FetchContent)
 
+# `FetchContent_MakeAvailable` requires internet access which might be disabled. Explicitly allow it.
+set(FETCHCONTENT_FULLY_DISCONNECTED OFF)
+
 FetchContent_Declare(
     Corrosion
     GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git


bbannier (commented):
Thanks for adding this, I was always curious whether it would be possible to reimplement Zeek's regex engine on top of [`regex-automata`](https://crates.io/crates/regex-automata)!

That said, in its current form I don't think this is something we would like to merge, and I am not even sure it is in a form where one could assess whether this is a "better" approach.

- even though this PR does not add a lot of tests it seems to have more code than the current impl
- the new impl contains zero comments for argument semantics, and has zero crate docs. I'd expect LLM-generated code to at least include docstrings and doctests
- test coverage is pretty light which is disappointing for a LLM-generated patch
- the current impl (even outside the FFI interface in `lib.rs`) feels way too complicated for me, and I feel it would be helpful to come up with some abstractions
- the integration into `RE.cc` feels hacky, wrapping the FFI interface into some C++ abstractions would probably be helpful

I briefly looked into the implementation itself, and I do not see enough "human in the loop" here. This not only makes this problematic to accept for license reasons, but likely also means that there are instances of unneeded fluff, bugs, or oddities which give a human pause. Since LLMs (supposedly) save so much time, I don't think expecting a human to spend 2-3x of the time spent to generate the code on code review before submitting it is asking for too much. As an example, I looked into `compat::parse_hex_digit` and it was pretty obvious that it would panic on invalid input; there are more instances of panics in the code which takes away from the benefit of introducing Rust as a Zeek dependency.

I pushed a couple commits to your branch which clean up trivial issues, but IMO this would need much more work. From an absolute Rust fanboy, this currently would be a 👎 from me.

timwoj:
Having not looked at the code at all, my primary concern is suddenly requiring a Rust compiler to be installed in order to build Zeek by default. Perhaps that's not a big deal, but it's something to consider here.

bbannier:
> Having not looked at the code at all, my primary concern is suddenly requiring a Rust compiler to be installed in order to build Zeek by default. Perhaps that's not a big deal, but it's something to consider here.

That's definitely something we should consider very carefully, but ideally we'd be weighing a succinct, and easy to maintain and contribute to Rust implementation against more complicated C++ code. Like I wrote, I don't think this PR currently builds a good case for that.

awelzel:
Benjamin asked for the benchmarker results where we do some large PCAP processing with `zeek -r`:

image

There's quite a bit of perf regression, so may need a close look with focus on perf, too.

ckreibich:
I'll go ahead and close this at this point — it's clearly not in a state that we would consider as-is, or iterate toward something we'd merge. But it is impressive that it's a workable substitute at all, given the challenges we've faced in finding a replacement for the current engine (https://github.com/zeek/zeek/issues/426 / https://github.com/zeek/zeek/discussions/2342).

PR #5309, opened 2026-04-01
Use debian unstable for weekly edge compiler builds

timwoj:
OpenSUSE Tumbleweed still doesn't have clang-22 packages even though it's been out for a while. In interest of using updated compiler versions, switch the weekly builds over to use debian-unstable, which has packages for both clang-22 and gcc-16.

We could also use the LLVM apt repository to get clang-23, but I'm not sure that's worth it just yet. gcc-16 is nearing a release, but clang-23 was only recently branched.

### DIFF
--- a/src/3rdparty/patricia.c
+++ b/src/3rdparty/patricia.c
@@ -227,7 +227,7 @@ prefix_t *New_Prefix2(int family, void *dest, int bitlen, prefix_t *prefix)
     if (family == AF_INET) {
       if (prefix == NULL) {
 #ifndef NT
-        prefix = calloc(1, sizeof(prefix4_t));
+        prefix = calloc(1, sizeof(prefix_t));
         if (prefix == NULL)
           out_of_memory("patricia/new_prefix2: unable to allocate memory");
 #else


timwoj:
Test builds:
gcc-16: https://cirrus-ci.com/task/6740820415479808
clang-22: https://cirrus-ci.com/task/4515408880861184

bbannier:
> I wonder a bit if there's a trick to bust the image cache, so that Cirrus rebuilds the image every week, but that's separate from `gcc` / `clang` testing.

The standard approach in Cirrus seems to be to use a `fingerprint_script` to return a cache key. If we made that return the week, e.g.`date +%Y-%V` which is `2026-14` for today, we'd get Cirrus to automatically refresh the image every week. Combined with `apt-get dist-upgrade` this should get us up-to-date setups while still avoiding pulling too many changes.

ci/debian-unstable/prepare-weekly.sh
```diff
@@ -1,24 +1,24 @@
 #!/bin/sh
 
-zypper refresh
-zypper patch -y --with-update --with-optional
+apt update
+apt upgrade -y
```
awelzel: As suggested by @bbannier:


```suggestion
apt dist-upgrade -y
```

timwoj:
> The standard approach in Cirrus seems to be to use a `fingerprint_script` to return a cache key. If we made that return the week, e.g.`date +%Y-%V` which is `2026-14` for today, we'd get Cirrus to automatically refresh the image every week. Combined with `apt-get dist-upgrade` this should get us up-to-date setups while still avoiding pulling too many changes.

Does a `fingerprint_script` cover the Docker container too? Looking at the docs, I think it's only for other caches.

timwoj:
> Does a `fingerprint_script` cover the Docker container too? Looking at the docs, I think it's only for other caches.

It looks like I can trick it with `docker_arguments` and passing some value that would be different every time. https://github.com/cirruslabs/cirrus-ci-docs/issues/692 was the inspiration for this (thanks @bbannier for the 5-year-old ticket 😛 ).

PR #5310, opened 2026-04-02
Enabling parsing for DNS opcode 4 (NOTIFY)

fatemabw:
Zeek already has parsing support for the DNS opcode 0, which is for standard query. DNS Notify is a DNS message type used by the DNS primaries to signal DNS stealth secondaries to initiate a zone transfer as the zone data has been updated. 

As the RFC 1996 notes:
NOTIFY uses the DNS Message Format, although it uses only a subset of the available fields.. NOTIFY is similar to QUERY in that it has a request message with the header QR flag "clear" (0, same as for standard Query) and a response message with QR "set" (1, same as for standard Response).  The response message contains no useful information, but its reception by the master is an indication that the slave has received the NOTIFY and that the master can remove the slave from any retry queue for this NOTIFY event.

And since Zeek already supports standard query/response (opcode 0) parsing in DNS message, NOTIFY messages can be treated indifferently to be supported built-in without any additional parsing code, if we enable opcode 4 and not exclude it from the packets parsing src code in DNS.cc file.

The purpose of this PR is to enable standard QR parsing for DNS NOTIFY messages and avoid the flood of "DNS_unknown_opcode (4)" in the zeek weird.log file.


testing/btest/Baseline/scripts.base.protocols.dns.notify/dns.log
```diff
@@ -0,0 +1,13 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
awelzel, line 1: The CI pipeline is failing seemingly because the baseline name is off.

If you rename the `dns_notify.zeek` test file to just `notify.zeek` it should align, otherwise rename the baseline to `testing/btest/Baseline/scripts.base.protocols.dns.dns_notify/dns.log‎`

fatemabw: Oops, I see now the miss-alignment of the test file name and the Baseline directory. Sorry about that. I have changed the dns_notify.zeek to notify.zeek to align with the dir name. Thanks Arne!

awelzel: Hey @fatemabw - CI is still failing because the baseline is outdated. I'll fix this up during the merge and add a NEWS entry.

fatemabw: Thanks Arne! :)

PR #5312, opened 2026-04-02
Support loading of .so files

awelzel:
This allows to load `.so` files directly from scripts using `@load ./plugin.so`. See the added tests.

There might be different opinions how to exactly do this and the implementation is fairly ad-hoc, but having this in 8.2 would allow to experiment with self-contained package+plugin directories and can fix up any hurdles on the way to 9.0. See the added test how this would look like.

Opening if draft and curious what people think.

Relates to #5167


src/plugin/Manager.cc
```diff
@@ -155,6 +155,71 @@ void Manager::SearchDynamicPlugins(const std::string& dir) {
...
+
+    current_plugin = nullptr;
+    current_sopath = path.c_str();
+    void* hdl = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
+    current_sopath = nullptr;
+
+    if ( ! hdl ) {
+        const char* err = dlerror();
+        std::string error = util::fmt("cannot load plugin library %s: %s", path.c_str(), err ? err : "<unknown error>");
+        return zeek::unexpected<std::string>(error);
```
bbannier: I'd suspect Coverity would flag this.

```suggestion
        return zeek::unexpected<std::string>(std::move(error));
```

src/plugin/Manager.cc
```diff
@@ -155,6 +155,71 @@ void Manager::SearchDynamicPlugins(const std::string& dir) {
...
+    }
+
+    if ( ! current_plugin ) {
+        std::string error = util::fmt("load plugin library %s did not instantiate a plugin", path.c_str());
+        dlclose(hdl);
+        return zeek::unexpected<std::string>(error);
```
bbannier:
```suggestion
        return zeek::unexpected<std::string>(std::move(error));

File: src/plugin/Manager.cc [hunk 3]
```diff
@@ -155,6 +155,71 @@ void Manager::SearchDynamicPlugins(const std::string& dir) {
...
+    auto* plugin = current_plugin;
+    current_plugin = nullptr;
+
+    // This is a bit quirky: If we go through ActivateDynamicPluginInternal(),
+    // its logic sets current_dir to a classic plugin's top-level directory,
+    // also called base_dir. Concretely, Plugin::Plugin() -> Manager::RegisterPlugin()
+    // -> Plugin::SetPluginPath() populates the plugin's base_dir and sopath members.
+    //
+    // If a plugin is loaded via @load ./plugin.so, there's no classic base_dir.
+    // Manager::RegisterPlugin() will skip setting the paths on the plugin We
```
bbannier:
```suggestion
    // Manager::RegisterPlugin() will skip setting the paths on the plugin. We
```

testing/btest/plugins/plugin-load-so/src/Plugin.cc
```diff
@@ -0,0 +1,33 @@
+#include "Plugin.h"
+
+#include <iostream>
```
bbannier:
```suggestion
#include <cstdio>
```

testing/btest/plugins/plugin-load-so/src/Plugin.cc
```diff
@@ -0,0 +1,33 @@
...
+
+using namespace plugin::Demo_Foo;
+
+zeek::plugin::Configuration Plugin::Configure() {
+    zeek::plugin::Configuration config;
+    config.name = "Demo::Foo";
+    config.description = "<Insert description>";
+    config.version.major = 0;
+    config.version.minor = 1;
+    config.version.patch = 0;
```
bbannier: Not needed, suggest to drop it.


```suggestion
```

testing/btest/plugins/plugin-load-so/src/Plugin.h
```diff
@@ -0,0 +1,20 @@
+
```
bbannier:
```suggestion
```

bbannier (approved):
LGTM. Thanks for refactoring the the current dylib loading so this doesn't introduce too much duplication.

PR #5313, opened 2026-04-02
XDP Shunter

evantypanski:
This adds the XDP shunter as an optional component. I'm worried about it building the BPF file and loading/distributing it on all Linux systems, so I'm not necessarily keen on making it on-by-default (yet).

There are a number of code issues I want to fix, but for this initial part I would prefer it get in then do development. We can mark it "experimental"

~(draft for CI and stuff, don't look too closely at the glue yet!)~


evantypanski:
just marking this ready because it's basically there and it's an optional component anyway, so mainly care about build integration stuff. I would discourage nitpicking here for that reason, there are a number of things I want better but it has been a lot nicer to deal with this in-tree already honestly

and I guess this will introduce Rust into Zeek :) sorry

Last 4 commits have substance (kind of), the ton before that are an unedited import. Can squash them if that seems better but I like the experiments being in the history.

src/xdp_shunt/bpf/vmlinux.h
```diff
@@ -0,0 +1,216 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+// This file is a generated file that was later stripped down to necessary
```
evantypanski, line 3: I learned how to reign this file in from 160k lines! It takes some manual work but you can make a BTF file from the object file (which shoves out any relocations) like:

```
bpftool gen min_core_btf /sys/kernel/btf/vmlinux vmlinux_min.btf build/src/xdp_shunt/bpf/filter.o
```

then dump it into a C file:

```
bpftool btf dump file vmlinux_min.btf format c > vmlinux_min.h
```

then I manually added some stuff (`icmphdr`, `xdp_action`, `bpf_map_type`).

I didn't document it because I'm hoping we can just add to this file as necessary, and there ends up being a lot of nuance ([for your consideration](https://lkml.org/lkml/2024/12/13/1296)). Though I can add some basic steps if that seems nice

scripts/policy/frameworks/xdp-shunt/__load__.zeek
```diff
@@ -0,0 +1,11 @@
+# Only load if the shunter plugin provided the necessary BiFs
+@ifdef ( XDP::_start_shunt )
```
awelzel: Not a biggie, but is there a reason for the `@ifdef`s here and elsewhere in all the files?

If someone explicitly loads `policy/frameworks/xdp-shunt`, seems okay to blow up if the plugin isn't available rather than not doing anything.

Or is that because of making some coverage tests easier?

scripts/policy/frameworks/xdp-shunt/bulk.zeek
```diff
@@ -0,0 +1,39 @@
+##! Shunt large flows (aka elephant flows).
+
+@ifdef ( XDP::_start_shunt )
+
+@load base/protocols/conn
```
awelzel:
```suggestion
@load base/protocols/conn/thresholds
```

I think that' should be enough and would prevent enabling the `conn.log` when loading `frameworks/xdp-shunt`.

scripts/policy/frameworks/xdp-shunt/connect.zeek
```diff
@@ -0,0 +1,70 @@
...
+module XDP;
+
+@load ./main
+
+export {
+	## Set this to true in order to tell Zeek to load the shunting
+	## XDP program itself. This is generally only useful for
+	## standalone instances, testing, or some unique deployment
+	## scenarios.
+	option start_new_xdp: bool = F;
```
awelzel: Nit. None of the options here are runtime configurable (whether that's even reasonable I'm skeptical). Using `const &redef` rather than `option` would make this explicit.

scripts/policy/frameworks/xdp-shunt/log_conn_id.zeek
```diff
@@ -0,0 +1,46 @@
+##! Adds logging to the connection shunter.
+
+@ifdef ( XDP::_start_shunt )
+@load base/protocols/conn
```
awelzel:
```suggestion
```

Not seeing why this needs this? Also spuriously enables `conn.log` logging.

src/xdp_shunt/shunter.bif
```diff
@@ -0,0 +1,128 @@
...
+    };
+    auto ps = zeek::iosource_mgr->GetPktSrc();
+    if ( ! ps ) {
+        zeek::reporter->Error("Packet source invalid!");
+        return zeek::make_intrusive<XDPProgramVal>();
+    }
+
+    auto ifindex = if_nametoindex(ps->Path().c_str());
+    if ( ifindex == 0 ) {
+        zeek::reporter->Error("Invalid device");
```
awelzel: Could be interesting  to include `ps->Path()` and the errno string here.

scripts/policy/frameworks/xdp-shunt/connect.zeek
```diff
@@ -0,0 +1,70 @@
...
+
+	vlans_included = ( ! force_no_vlans ) && should_load_with_vlan();
+
+	local opts: XDP::ShuntOptions = [ $attach_mode=attach_mode,
+	    $conn_id_map_max_size=conn_id_map_max_size,
+	    $ip_pair_map_max_size=ip_pair_map_max_size,
+	    $include_vlan=vlans_included, $pin_path=pin_path,  ];
+
+	if ( start_new_xdp )
+		start_shunt(opts);
```
awelzel: Nit territory. I'm getting hung up on the `start_shunt()` name and how it compares to "start shunting of a connection". An XDP program also runs for each packet rather than being "started"?

Wonder if this could be `XDP::load_bpf_program()` and `XDP::unload_bpf_program()` which I guess is what they actual do (and would align with `should_load`).

tools/zeek-xdp-loader/src/main.rs
```diff
@@ -0,0 +1,292 @@
+mod maps;
```
awelzel, line 1: Don't mean to be anti-rs, bould this be `.cc` and delegate to `load_and_attach()` in `UserXDP.cc` instead?

I played around a bit, and there's `bpftool map create ...` that allows to pre-create BPF maps of certain sizes and wanted to suggest using that, but actually doing so resulted in the BPF verifier failed because the maps were created without BTF ID and some spinlock didn't work and so didn't look further.

My understanding: Primary use for this too is to set the map sizes which vanilla xdp-loader doesn't support.

The `UserXDP.cc` program has this `load_and_attach()` helper that should be easy enough to wrap with a small C++ loader executable?

I think I gather there's more debugging and pretty printing functionality, timestamps, etc., but that could be an external/contrib project?

It's just that there's now 3 ways to load the XDP program: Zeek C++/BIF, zeek-xdp-loader, xdp-loader (with default map sizes) and it could be done with Python/[bcc](https://github.com/iovisor/bcc/blob/master/examples/networking/xdp/xdp_drop_count.py).

scripts/policy/frameworks/xdp-shunt/conn-id-logging.zeek
```diff
@@ -0,0 +1,46 @@
+##! Adds logging to the connection shunter.
```
awelzel, line 1: Would rename to `conn-id-logging.zeek` - we have a few more scripts ending with `-logging.zeek`. You could also call it `debug-logging.zeek` similar to the one in `frameworks/analyzer/`.

src/xdp_shunt/bpf/filter.c
```diff
@@ -0,0 +1,286 @@
...
+        tuple.port2 = port_source;
+    }
+
+    struct shunt_val* val = bpf_map_lookup_elem(&filter_map, &tuple);
+    if ( val ) {
+        update_value(val, ctx, from_ip1);
+        return XDP_DROP;
+    }
+
+    // Check IP pairs
```
awelzel, line 265: So I' wondering. Support for ip_pair filtering requires a whole separate map and a separate key and  various separate code paths. And then it's somewhat optional (or not used in the scripts at this point).

Have you thought of using a single canonical_tuple that discriminates between the matching style via an extra byte and then zero out (port1, port2 and protocol). Do the same when starting to shunt a ip_pair: use zero ports and protocol, but set the kind/style byte appropriately.


```
struct canonical_tuple {
    struct in6_addr ip1; // The lower IP
    struct in6_addr ip2; // The higher IP
    __u16 port1;         // The port corresponding with ip1
    __u16 port2;         // The port corresponding with ip2
    __u8 protocol;      // The protocol for this connection.
    # new:
    __u8 key_attributes; // 1 << 0: ippair, 1 << 1 vlans-included, ...
    __u16 outer_vlan_id; // The outer vlan, or 0 if none
    __u16 inner_vlan_id; // The inner vlan, or 0 if none
};
```

The protocol is only a single byte, so there's some room left.

Additionally, currently vlan_id = 0 vs vlan not set cannot be discriminated? So could also reserve a single bit that indicates "vlan-matching on" and "ip-pair only matching" as part of the key.


Does that make sense at all? Somewhere up in line 198 could first create the tuple for ip_pair matching, then fill the remaining parts for full matching. Doing the lookup in the same map.

On the flip side, treating them separate may be better, setting different limits and individual accounting, but I'm not sure how interesting that really is, or whether you'd mirror that information elsewhere, too.

src/xdp_shunt/bpf/UserXDP.cc
```diff
@@ -0,0 +1,210 @@
...
+        case XDP_MODE_UNSPEC: break;
+        case XDP_MODE_NATIVE: flags |= (1U << 2); break;
+        case XDP_MODE_SKB: flags |= (1U << 1); break;
+        case XDP_MODE_HW: flags |= (1U << 3); break;
+    }
+
+    return flags;
+}
+
+std::optional<std::string> reconnect(struct filter** skel, xdp_options opts) {
```
awelzel: I'm getting a bit hung up on reconnect(), mostly because connection is usually a network connection and not sure about connecting to BPF program / map.

Maybe "reuse" instead to align with  `bpf__map_reuse_fd()`?

scripts/policy/frameworks/xdp-shunt/main.zeek
```diff
@@ -0,0 +1,80 @@
...
+	global xdp_prog: opaque of XDP::Program;
+
+	## Whether or not vlans should be included. This is necessary to
+	## construct the correct "canonical" tuple for the XDP program.
+	global vlans_included: bool;
+}
+
+function reconnect(options: XDP::ShuntOptions): bool
+	{
+	xdp_prog = _reconnect_shunt(options);
```
awelzel: Very much nit, usually we use double underscores for private BiFs.

src/xdp_shunt/conn_id_shunter.bif
```diff
@@ -0,0 +1,155 @@
...
+#include "zeek/xdp_shunt/bpf/UserXDP.h"
+#include "zeek/xdp_shunt/bpf/filter_common.h"
+
+using namespace zeek::plugin::detail::Zeek_XDP_Shunter;
+%%}
+
+module XDP::Shunt::ConnID;
+
+## Event raised whenever a connection is shunted.
+event shunted_conn%(cid: conn_id%);
```
awelzel: I haven't thought about this much, but it'd be cool if the name included that conneciton shunting is now in progress and be a bit more generic.

The most verbose ones I can think of is global:
```
connection_shunting_started(cid: conn_id)
connection_shunted(c: connection)
```

and preferably with `c: connection` as the latter demonstrates so it'd mirror the `shunt(c: connection)` API.


With `connection_timing_out()` and `connection_state_remove()`, this would fit nicely?

src/xdp_shunt/conn_id_shunter.bif
```diff
@@ -0,0 +1,155 @@
...
+
+    auto shunt_map = get_map<canonical_tuple>(get_canonical_id_map(prog));
+    for ( const auto& [key, val] : shunt_map ) {
+        // Skip if we are still within time_since_last_packet
+        if ( time_since_last_packet &&
+             (time_since_last_packet + monoToWall(val.timestamp) > zeek::util::current_time(true)) )
+            continue;
+
+        auto zeek_key = zeek::make_intrusive<zeek::RecordVal>(canonical_id);
+        if ( IN6_IS_ADDR_V4MAPPED(&key.ip1) )
```
awelzel: I don't think you have to do the if/else: The AddrVal constructor will store the mapped address and become IPv4. Pretty sure :-)

scripts/policy/frameworks/xdp-shunt/shunt-conn-id.zeek
```diff
@@ -0,0 +1,131 @@
...
+	option shunted_connection_timeout: interval = 10sec;
+
+	## If connections should always unshunt when the connection is removed
+	## from Zeek.
+	option unshunt_on_connection_remove: bool = T;
+
+	global finalize_shunt: Conn::RemovalHook;
+}
+
+function get_map(time_since_last_packet: interval &default=0sec)
```
awelzel, line 73: What is this actually used for? Reaping from Zeek scripts? There's no caller in the current scripts unless I missed them.

src/xdp_shunt/ip_pair_shunter.bif
```diff
@@ -0,0 +1,143 @@
...
+##
+## .. zeek:see:: shunt shunt_stats
+function _unshunt%(xdp_prog: opaque of XDP::Program, pair: XDP::ip_pair%) : XDP::ShuntedStats %{
+    auto zeek_prog = XDPProgramVal::CastFromAny(xdp_prog);
+    if ( ! zeek_prog ) {
+        zeek::reporter->Error("%s", zeek_prog.error().c_str());
+        return makeEmptyShuntedStats();
+    }
+
+    if ( ! (*zeek_prog)->is_set ) {
```
awelzel, line 115: Not a biggie, just wondering if the code here and in the conn_id_shunter.bif could/should move into XDPProgramVal members to consolidate some of the duplicated error handling.

```
auto prog = CastFromAny(xdp_prog);
prog->ShuntIPPair(pair);
```

```
class XDPProgramVal {
   bool Shunt(const RecordVal& conn_id)
    RecordValPtr Unshunt(const RecordVal& conn_id);

    bool ShuntIPPair(const ReordVal& ip_pair);
    RecordValPtr UnshuntIPPair(const ReordVal& ip_pair);
}
```

It'd move more `Val` things into `XDPProgramVal`, but it's already an OpaqueVal, so 🤷‍♂️ 

Feel free to bounce, too :-)

src/xdp_shunt/ShuntUtil.cc
```diff
@@ -0,0 +1,199 @@
...
+zeek::RecordValPtr makeEmptyShuntedStats() {
+    static auto shunt_stats_type = zeek::id::find_type<zeek::RecordType>("XDP::ShuntedStats");
+
+    auto stats = zeek::make_intrusive<zeek::RecordVal>(shunt_stats_type);
+    stats->Assign(0, zeek::val_mgr->Count(0)); // packets_from_1
+    stats->Assign(1, zeek::val_mgr->Count(0)); // bytes_from_1
+    stats->Assign(2, zeek::val_mgr->Count(0)); // packets_from_2
+    stats->Assign(3, zeek::val_mgr->Count(0)); // bytes_from_2
+    // Timestamp is optional
+    stats->Assign(5, zeek::val_mgr->Bool(false)); // present
```
awelzel: Nit perf optimization: Use `Assign(5, false)`, `Assign(0, 0)` or so, no need to go through `val_mgr`.

src/xdp_shunt/ShuntUtil.cc
```diff
@@ -0,0 +1,199 @@
...
+
+    pair.ip1 = ip2_higher ? ip1 : ip2;
+    pair.ip2 = ip2_higher ? ip2 : ip1;
+    pair.outer_vlan_id = outer_vlan_id;
+    pair.inner_vlan_id = inner_vlan_id;
+
+    return pair;
+}
+
+zeek::AddrValPtr makeAddr(struct in6_addr ip) {
```
awelzel: Unused (lucky find).

src/xdp_shunt/Plugin.cc
```diff
@@ -0,0 +1,25 @@
...
+    zeek::plugin::Configuration config;
+    config.name = "Zeek::XDP_Shunter";
+    config.description = "Shunts traffic via an XDP program";
+    config.version.major = 0;
+    config.version.minor = 1;
+    config.version.patch = 0;
+    return config;
+}
+
+void Plugin::InitPostScript() { program_opaque = zeek::make_intrusive<zeek::OpaqueType>("xdp::XDPProgram"); }
```
awelzel:
```suggestion
void Plugin::InitPostScript() { program_opaque = zeek::make_intrusive<zeek::OpaqueType>("XDP::Program"); }
```

Does that work? To align with `XDP::Program` in the scripts.

src/xdp_shunt/shunter.bif
```diff
@@ -0,0 +1,128 @@
...
+    if ( err ) {
+        zeek::reporter->Error("Failed to load BPF skeleton: %s", (*err).c_str());
+        return zeek::make_intrusive<XDPProgramVal>();
+    }
+
+    auto xdp_prog = zeek::make_intrusive<XDPProgramVal>(skel, opts);
+    return xdp_prog;
+%}
+
+function _end_shunt%(xdp_prog: opaque of XDP::Program%) : bool %{
```
awelzel: I think detach here is more specific and reuses BPF terminology - related to the other comment about the `_start` function.

src/xdp_shunt/ShuntUtil.cc
```diff
@@ -0,0 +1,199 @@
...
+        stats->Assign(1, zeek::val_mgr->Count(val->bytes_from_1));
+        stats->Assign(2, zeek::val_mgr->Count(val->packets_from_2));
+        stats->Assign(3, zeek::val_mgr->Count(val->bytes_from_2));
+    }
+    else {
+        stats->Assign(0, zeek::val_mgr->Count(val->packets_from_2));
+        stats->Assign(1, zeek::val_mgr->Count(val->bytes_from_2));
+        stats->Assign(2, zeek::val_mgr->Count(val->packets_from_1));
+        stats->Assign(3, zeek::val_mgr->Count(val->bytes_from_1));
+    }
```
awelzel, line 111: No need for `val_mgr->Count()` as elsewhere.

src/xdp_shunt/ShuntUtil.cc
```diff
@@ -0,0 +1,199 @@
...
+
+    if ( val->timestamp != 0 ) {
+        double packet_wall_time = monoToWall(val->timestamp);
+        stats->Assign(4, zeek::make_intrusive<zeek::TimeVal>(packet_wall_time));
```
awelzel: there is  a more efficient `->AssignTime()`.

tools/zeek-xdp-loader/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[target."cfg(all())"]
+runner = "sudo -E"
```
awelzel, line 2: Is this important or can we expect dev/users to invoke `sudo` themselves?

tools/zeek-xdp-loader/src/main.rs
```diff
@@ -0,0 +1,292 @@
...
+            let _ = map.remove(&key);
+        }
+
+        eprintln!("Purged {count} entries.");
+    }
+
+    Ok(())
+}
+
+#[tokio::main]
```
awelzel, line 214: Why is `[tokio::main]` needed, why the dependency? Seems this compiles just fine without this an removing the dependency from Cargo.toml.

awelzel: Removing it make the release binary from 2.7M to 2.3M :-)

tools/zeek-xdp-loader/src/main.rs
awelzel, line 1: Given there's `load_and_attach()` in UserXDP.cc, I'm not much seeing why this needs Rust here, could wrap the existing C++ code minimally for loading the program and setting the map sizes appropriately.

Or, on second thought, you could probably do the following from zeekctl as a one-shot "external loader" for the time being (half-tested) given you implemented support in Zeek script for loading/attaching XDP programs.

```
# ./bin/zeek -i enp7s0 -e 'XDP::_start_shunt([ $attach_mode=XDP::UNSPEC, $conn_id_map_max_size=10000, $ip_pair_map_max_size=1000, $include_vlan=T, $pin_path="/sys/fs/bpf/zeek/"])'
<command line>, line 3: listening on enp7s0

libbpf: couldn't reuse pinned map at '/sys/fs/bpf/zeek//filter_map': parameter mismatch
libbpf: map 'filter_map': error reusing pinned map
libbpf: map 'filter_map': failed to create: Invalid argument(-22)
libbpf: failed to load object 'filter'
libbpf: failed to load BPF skeleton 'filter': -22
error in <command line>, line 3: Failed to load BPF skeleton: Could not find BPF program
fatal error in <command line>, line 3: errors occurred while initializing
```
If instead of using the active packet source in C++ code, an additional `$interface: string` field is added, that provides a half decent CLI IMO and wouldn't need the `-i enp7s0` which is a bit annoying.

And supposedly the purging could be done with a similar Zeek invocation, using the otherwise not yet used `get_map()` script functions?

I guess I'm just far more on the frugal side and seems an initial version of the XDP shunter would work reasonably well without introducing a a new language and you already implemented most pieces in C++/Zeek script anyhow.

No overly strong feelings and if others feel this is the right moment to add Rust, sounds good, but slight eyebrow raising from me :-)

awelzel (commented):
I've ran it a bit in a VM and observed the BPF maps populated and also "truncated" histories in conn.log.

Comments are mostly thoughts / suggestions an opinions. Given this isn't enabled by default, I could see landing it as is (modulo a bit the Rust-based loaded - I'v left my thoughts there).

src/xdp_shunt/shunter.bif
awelzel: As mentioned later, Might make more sense to pass `$interface=` into in the record and populate it via the `packet_source()$path` in script land:

```
# ./bin/zeek -i enp7s0 -e 'print packet_source()$path'
```

scripts/policy/frameworks/xdp-shunt/connect.zeek
awelzel: Maybe `load_and_attach`...

tools/zeek-xdp-loader/src/main.rs
awelzel: Ignore this one, I thought I deleted the comment.

scripts/policy/frameworks/xdp-shunt/shunt-conn-id.zeek
evantypanski: I used it a couple times for debugging and saw utility, it was pretty early on and I just haven't removed it. It could just disappear too.

Initially it was for unshunting timeouts, so once every minute or so walk through each element in the map and unshunt anything whose last packet is over 30 seconds ago or something. But that has since been made obsolete.

src/xdp_shunt/ShuntUtil.cc
evantypanski: I got rid of like 6 right before marking as ready ;) guess I missed one

tools/zeek-xdp-loader/.cargo/config.toml
evantypanski: it was in the Aya template and I didn't clean it since it seemed reasonable enough

tools/zeek-xdp-loader/src/main.rs
evantypanski: leftover from Aya template, oops

tools/zeek-xdp-loader/src/main.rs
evantypanski: So here is my reasoning:

1) XDP loaders almost always run as root (at least, I haven't figured out how to consistently get around it, it may be possible without?)
2) If we must run as root, then that should be a separate binary (not Zeek)
3) We need a way to purge old keys if a worker crashes
4) If there is a large environment, then we should not require Zeek to purge the map (it is large)

Therefore:
5) We should use a separate, periodic run to purge values
6) It should be separate from Zeek

Then I just experimented with Aya and found it far easier and nicer than libxdp etc. so I stuck with it.

There are other considerations too: in the future, we will want more XDP support (at least, I do). We can fully control a loader and make it more pleasant to write that code. There could be multiple Zeek processes sharing an interface, so each one cannot start its own XDP. At that point using Zeek as the XDP loader feels incredibly odd - why would we run Zeek first to load the XDP program when we could separate that into a purpose-built executable?

It's also just a nice separation of concerns. I would much rather rip out any XDP knowledge from Zeek, then it just knows how to communicate with these BPF maps. Then we have a separate binary that knows about XDP.

Regarding Rust: It's not a big deal to me. I can just rewrite it in C++ if that's really preferable. I'm pretty opposed to Zeek loading the XDP program. But I don't see a better time to introduce Rust than with a library (Aya) that I find nicer than libxdp, as a separate executable, where not many people will have to maintain it (since it's just a little bonus feature). I already want to implement features in the XDP program that only target a bit newer kernels for large MTUs. I don't see why a different language would be wrong for a separate binary. It fit the job better here.

FWIW I actually don't think this should stop from merging here, since from a user perspective this would all magically work anyway (other than what dependencies they need in the meantime, but it's optional). Only power users would know (if they're loading the XDP programs themselves) and I actually intended to give some nice features for them.

evantypanski:
I'll work through the nits, I had a number of them myself, but I don't think much should stop it from merging. Unless we have serious concerns with XDP in general with Zeek, that is :)

If we can agree on a separate binary used for the loader (which I personally lean towards), then it being Rust is just an implementation detail, we could technically change post-merge with no meaningful change to users.

Also it'll need the [corresponding zeekctl PR](https://github.com/zeek/zeekctl/pull/107) reviewed

scripts/policy/frameworks/xdp-shunt/__load__.zeek
evantypanski: Yeah it's because of the `scripts.bare-mode-errors` test. Is there a better way? I really dislike my solution!

I guess I can introduce a `have-xdp-shunter` script or something too, for that test..

tools/zeek-xdp-loader/.cargo/config.toml
bbannier: This seems unusual to me since it e.g., forces `cargo test` to run under `sudo` for no reason (there are currently no tests, but still ...). This only affects people working with the project sources as it doesn't translate to installed binaries, and I'd argue we shouldn't hide `sudo` deep in the guts in that case either.

tools/zeek-xdp-loader/Cargo.toml
bbannier, line 1: Since best case the next tool we'll add would just work from this as a template, it would be great to establish more aggressive linting rules. I got good value from the following, but might even want to get rid of these allows in a shared codebase.

```toml
[lints.clippy]
pedantic = { level = "deny", priority = -1 }
missing_errors_doc = "allow"
missing_panics_doc = "allow"
```

tools/zeek-xdp-loader/.cargo/config.toml
evantypanski: my impression was that it's for `cargo run`, which running as `sudo` is a little odd (ie if you only have `cargo` installed for the user it may not be in the `PATH`, which is very reasonable)

(happy to remove, it doesn't matter here, just clarifying)

src/xdp_shunt/bpf/filter.c
evantypanski: I actually thought of IP pairs as an alternative to BPF filters: so I don't even have unshunting for them!

The vlans part was just to lower the number of map lookups. I haven't measured the perf impact, but for something that should improve the performance, I want it to be as lightweight as possible. If we add an attribute for vlans, then we will need two map lookups for flows and two for IP pairs, rather than one each. I figured you either care about vlans or not.

Alternatively we can just always care about vlans, but that's not what Zeek does by default for flows.

tools/zeek-xdp-loader/.cargo/config.toml
bbannier: Yeah, I think it is for running the binary, tests and benchmarks, but e.g., build scripts should still run as normal user. I personally am not a fan of this and would be totally fine with `cargo build && sudo ./target/debug/...` (maybe hidden in a `Makefile` 🤷), but won't force this.

src/xdp_shunt/bpf/filter.c
awelzel: Got mostly hung up how you discriminate between vlan not set and vlan set with ID 0. Encoding a "has vlan bit" in the key itself seemed to make it possible to discriminate these cases and do just a single lookup.

```
192.168.0.1:4321          192.168.0.2:443 vlan 0       - enable shunting
192.168.0.1:4321          192.168.0.2:443 untagged     - also shunted.
```
Maybe I misread the code right and also realize we haven't had proper vlan support in a while and this is a total detail and special case. Feel free to just ignore.

> I actually thought of IP pairs as an alternative to BPF filters: so I don't even have unshunting for them!

Yep. And I'm not sure I like bringing in all the types and infrastructure for it, but I gather you'd revisit this post-merge.

evantypanski: Oh, no it actually does not discriminate between VLAN 0 and no VLAN set, it is on my list!

I was more worried about integration with Zeek where Zeek does not have VLAN tags (because out of the box it does not) but on the wire the flow does have VLAN tags. Then the shunter would continuously fail to shunt it, since it set the vlan tags and Zeek could not. We could account for this by two map lookups: one where it checks with the vlan tags, and one where it checks without *plus* that extra bit in the key.

the extra `romemory` bit to include vlans is just asking whether Zeek knows enough about vlans to "properly" set the tags

evantypanski:
(seems like consensus is this won't be in 8.2, just putting that here for visibility)

PR #5314, opened 2026-04-02
Add storage-framework options to known-certs and known-hosts

timwoj:
This adds the same setup for the storage framework to known-certs and known-hosts that already exists for known-services. These three scripts are the only places we call `Cluster::create_store` in the base scripts.


scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -30,23 +34,48 @@ export {
 	## operation.
 	const use_host_store = F &redef;
 
+	## Switches to the version of this script that uses the storage
+	## framework instead of Broker stores.
+	const hosts_use_storage_framework = F &redef;
```
awelzel: To actually turn on the storage framework, users needs to do two things:
```
redef use_host_store = T;
redef hosts_use_storage_framework = T;
```

And then, if they don't want to memory store backend per worker, but instead something centralized `redef host_store_backend_options = [... $redis=[ ...]]`, or a central file based SQLite database?

These scripts are now also supporting three different variants how they can be configured (or more if you account for different storage backends). I wonder a bit if we should take a step back and just support a single configuration here.

But I don't have an overly strong opinion. Deferring to @ckreibich . Might also be okay to just drop the `use_x_store` logic completely instead.

timwoj: I agree that it's very verbose. The extra bool would go away once the Broker storage stuff goes away. We can't remove `use_x_store` because these scripts don't use the data store by default (historically for performance reasons).

awelzel: I was looking at this a bit more and wonder if we could have just a single variable:

```
redef Known::enable_persistence = T;
```

Then, if that is true and `use_x_store` not otherwise configured, the storage framework is used. So instead of adding an option `Known::service_use_storage_framework`, have one global option for persistence that implies usage of the storage framework unless the old and deprecate `use_x_store` is set to `T`.

And it'd also be good to not default to SQLite `:memory:`. Leveraging the known scripts, we'd want to have a) persistence and b) centralization. The `:memory` backend provides neither (indeed, I think there's no real usecase in general for `:memory:` other than testing/perf testing: Wouldn't a Zeek table almost always be the better alternative?).

So if we default to a file backed SQLite store by default, we could put it under:
```
`Cluster::default_store_dir` / known / hosts.sqlite
`Cluster::default_store_dir` / known / services.sqlite
```

On a single-node cluster, that should end up in `${SpoolDir}/stores/known/hosts.sqlite`. In a multi-node cluster, every system would have their own directory, but they should then switch to Redis or Postgres.

For running Zeek with the known scripts in your home network on a single machine, that seems like a nice behavior, no?

The `redef Known::enable_persistence = T` is all a user would do to store "known stuff" in a central SQLite databases. Operators of larger clusters might want to look into using a non-SQLite storage.

I'm a bit wary of concurrent accesses to the SQLite databases and may want to add a tiny cache to avoid put/insert of the very same values into the databases, but that could come later.

---

Here's the sketch for known-services - untested. Would that seem reasonable?
```
diff --git a/scripts/policy/protocols/conn/known-services.zeek b/scripts/policy/protocols/conn/known-services.zeek
index 8d2e8204e9..f74996112e 100644
--- a/scripts/policy/protocols/conn/known-services.zeek
+++ b/scripts/policy/protocols/conn/known-services.zeek
@@ -37,6 +37,9 @@ export {
 		service:        set[string]     &log;
 	};
 
+	# Should move into policy/frameworks/known/main.zeek ?
+	const enable_persistence = F &redef;
+
 	## Toggles between different implementations of this script.
 	## When true, use a Broker data store, else use a regular Zeek set
 	## with keys uniformly distributed over proxy nodes in cluster
@@ -141,11 +144,17 @@ function check(info: ServicesInfo) : bool
 event zeek_init()
 	{
-	if ( ! Known::use_service_store )
+	if ( ! Known::use_service_store && ! Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
 
-	if ( Known::services_use_storage_framework )
+@pragma push ignore-deprecations
+	if ( Known::use_service_store )
+		{
+		Known::service_broker_store = Cluster::create_store(Known::service_store_name);
+@pragma pop ignore-deprecations
+		}
+	else
 		{
 		local res = Storage::Sync::open_backend(Known::service_store_backend_type, Known::service_store_backend_options, Known::AddrPortServTriplet, bool);
 		if ( res$code == Storage::SUCCESS )
@@ -153,18 +162,12 @@ event zeek_init()
 		else
 			Reporter::error(fmt("%s: Failed to open backend connection: %s", Known::service_store_prefix, res$error_str));
 		}
-	else
-		{
-@pragma push ignore-deprecations
-		Known::service_broker_store = Cluster::create_store(Known::service_store_name);
-@pragma pop ignore-deprecations
-		}
 	}
 
 event service_info_commit(info: ServicesInfo)
 	{
-	if ( ! Known::use_service_store )
+	if ( ! Known::use_service_store && ! Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
 
@@ -174,19 +177,23 @@ event service_info_commit(info: ServicesInfo)
 		{
 		local key = AddrPortServTriplet($host = info$host, $p = info$port_num, $serv = s);
 
-		if ( Known::services_use_storage_framework )
+@pragma push ignore-deprecations
+		if ( Known::use_service_store )
+@pragma pop ignore-deprecations
 			{
-			when [info, s, key] ( local put_res = Storage::Async::put(Known::service_store_backend, [$key=key, $value=T, $overwrite=F,
-			                                                    $expire_time=Known::service_store_expiry]) )
+			when [info, s, key] ( local r = Broker::put_unique(Known::service_broker_store$store, key,
+			                                    T, Known::service_store_expiry) )
 				{
-				if ( put_res$code == Storage::SUCCESS )
+				if ( r$status == Broker::SUCCESS )
 					{
-					info$service = set(s);	# log one service at the time if multiservice
-					Log::write(Known::SERVICES_LOG, info);
+					if ( r$result as bool ) {
+						info$service = set(s);	# log one service at the time if multiservice
+						Log::write(Known::SERVICES_LOG, info);
+						}
 					}
-				else if ( put_res$code != Storage::KEY_EXISTS )
-					Reporter::error(fmt("%s: data store put_unique failure: %s",
-					                    Known::service_store_name, put_res$error_str));
+				else
+					Reporter::error(fmt("%s: data store put_unique failure",
+					                    Known::service_store_name));
 				}
 			timeout Known::service_store_timeout
 				{
@@ -195,19 +202,17 @@ event service_info_commit(info: ServicesInfo)
 			}
 		else
 			{
-			when [info, s, key] ( local r = Broker::put_unique(Known::service_broker_store$store, key,
-			                                    T, Known::service_store_expiry) )
+			when [info, s, key] ( local put_res = Storage::Async::put(Known::service_store_backend, [$key=key, $value=T, $overwrite=F,
+			                                                    $expire_time=Known::service_store_expiry]) )
 				{
-				if ( r$status == Broker::SUCCESS )
+				if ( put_res$code == Storage::SUCCESS )
 					{
-					if ( r$result as bool ) {
-						info$service = set(s);	# log one service at the time if multiservice
-						Log::write(Known::SERVICES_LOG, info);
-						}
+					info$service = set(s);	# log one service at the time if multiservice
+					Log::write(Known::SERVICES_LOG, info);
 					}
-				else
-					Reporter::error(fmt("%s: data store put_unique failure",
-					                    Known::service_store_name));
+				else if ( put_res$code != Storage::KEY_EXISTS )
+					Reporter::error(fmt("%s: data store put_unique failure: %s",
+					                    Known::service_store_name, put_res$error_str));
 				}
 			timeout Known::service_store_timeout
 				{
@@ -220,7 +225,7 @@ event service_info_commit(info: ServicesInfo)
 event known_service_add(info: ServicesInfo)
 	{
-	if ( Known::use_service_store )
+	if ( Known::use_service_store || Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
 
@@ -255,7 +260,7 @@ event known_service_add(info: ServicesInfo)
 event Cluster::node_up(name: string, id: string)
 	{
-	if ( Known::use_service_store )
+	if ( Known::use_service_store || Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
 
@@ -269,7 +274,7 @@ event Cluster::node_up(name: string, id: string)
 event Cluster::node_down(name: string, id: string)
 	{
-	if ( Known::use_service_store )
+	if ( Known::use_service_store || Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
 
@@ -283,7 +288,7 @@ event Cluster::node_down(name: string, id: string)
 event service_info_commit(info: ServicesInfo)
 	{
-	if ( Known::use_service_store )
+	if ( Known::use_service_store || Known::enable_persistence )
 		return;
 @pragma pop ignore-deprecations
```

timwoj: I like this a bit because it separates the broker parts from the storage framework parts, which enables removing the broker part more easily. I'm not sure about the `persistence` name since something like memcached isn't really persistent across runs, but maybe that's fine and I'm being picky.

timwoj: I also don't think the flag should move to an upstream point. The original scripts let you turn stores on and off individually so they should probably stay that way.

testing/btest/scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -18,6 +18,40 @@
...
+# @TEST-EXEC: mv known_hosts.log knownhosts-broker-store.log
+# @TEST-EXEC: btest-diff knownhosts-broker-store.log
+
+# @TEST-EXEC: zeek -r $TRACES/wikipedia.pcap %INPUT storage-framework-config.zeek
+# @TEST-EXEC: mv known_hosts.log knownhosts-storage-framework.log
+# @TEST-EXEC: btest-diff knownhosts-storage-framework.log
+
+# @TEST-EXEC: cat knownhosts-broker-store.log | $SCRIPTS/diff-remove-timestamps > broker-store.log
+# @TEST-EXEC: cat knownhosts-storage-framework.log | $SCRIPTS/diff-remove-timestamps > storage-framework.log
+# @TEST-EXEC: diff broker-store.log storage-framework.log > logs-diff.txt
```
awelzel: This should fail and print something reasonable on stderr and then there's no need for empty baselines.

```suggestion
# @TEST-EXEC: diff -u broker-store.log storage-framework.log
```

scripts/policy/protocols/ssl/known-certs.zeek
```diff
@@ -77,45 +112,80 @@ export {
...
 		}
-	timeout Known::cert_store_timeout
+	else
 		{
-		# Can't really tell if master store ended up inserting a key.
-		Log::write(Known::CERTS_LOG, info);
+		when [info, key] ( local put_res = Storage::Async::put(Known::cert_store_backend, [$key=key, $value=T, $overwrite=F,
+		                                                    $expire_time=Known::cert_store_expiry]) )
+			{
+			print(put_res$code);
```
awelzel:
```suggestion
```

scripts/policy/protocols/ssl/known-certs.zeek
```diff
@@ -44,20 +52,47 @@ export {
...
+	## The name to use for :zeek:see:`Known::cert_store_backend`. This will be used
+	## by the backends to differentiate tables/keys. This should be alphanumeric so
+	## that it can be used as the table name for the storage framework.
+	const cert_store_prefix = "zeekknowncerts" &redef;
+
+	## The type of storage backend to open.
+	const cert_store_backend_type : Storage::Backend = Storage::STORAGE_BACKEND_SQLITE &redef;
+
+	## The options for the cert store. This should be redef'd in local.zeek to set
+	## connection information for the backend. The options default to a memory store.
```
awelzel: Here and elsewhere.

```suggestion
	## connection information for the backend. The options default to a central persistent sqlite database.
```

scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -24,6 +28,10 @@ export {
 		host:    addr &log;
 	};
 
+	## Use the storage framework to enable persistence of the stored
+	## hosts between runs.
+	const enable_hosts_persistence = F &redef;
```
awelzel: I am tempted to suggest un-pluralizing this one and the others given other names aren't in plural either.

scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -67,43 +97,77 @@ export {
...
+			{
+			if ( put_res$code == Storage::SUCCESS )
+				Log::write(Known::HOSTS_LOG, info);
+			else if ( put_res$code != Storage::KEY_EXISTS )
+				Reporter::error(fmt("%s: data store put_unique failure: %s",
+				                    Known::host_store_name, put_res$error_str));
+			}
+		timeout Known::host_store_timeout
+			{
+			Log::write(Known::HOSTS_LOG, info);
```
awelzel, line 164: Seems adding `Reporter::error()` or warning here for visibility would be useful? Would ignore for the Broker stores though.

scripts/policy/protocols/ssl/known-certs.zeek
```diff
@@ -44,20 +52,47 @@ export {
...
+
+	## The expiry interval of new entries in :zeek:see:`Known::cert_broker_store` and
+	## :zeek:see:`Known::cert_store_backend`. This also changes the interval at which
+	## certs get logged.
 	option cert_store_expiry = 1day;
 
 	## The timeout interval to use for operations against
-	## :zeek:see:`Known::cert_store`.
+	## :zeek:see:`Known::cert_broker_store` and
+	## :zeek:see:`Known::cert_store_backend`. .
```
awelzel:
```suggestion
	## :zeek:see:`Known::cert_store_backend`.
```

testing/btest/scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -18,6 +18,40 @@
...
+# @TEST-EXEC: btest-diff logs-diff.txt
```
awelzel:
```suggestion
```

scripts/policy/protocols/ssl/known-certs.zeek
```diff
@@ -44,20 +52,47 @@ export {
...
+	const cert_store_prefix = "zeekknowncerts" &redef;
+
+	## The type of storage backend to open.
+	const cert_store_backend_type : Storage::Backend = Storage::STORAGE_BACKEND_SQLITE &redef;
+
+	## The options for the cert store. This should be redef'd in local.zeek to set
+	## connection information for the backend. The options default to a memory store.
+	const cert_store_backend_options : Storage::BackendOptions = [ $sqlite = [
+		$database_path=fmt("%s/known/certs.sqlite", Cluster::default_store_dir),
+		$table_name=Known::cert_store_name ]] &redef;
```
awelzel:
```suggestion
		$table_name=Known::cert_store_prefix]] &redef;
```

I deployed this and it bailed because table names with `/` aren't working, so that's a bogus default. Also applies to the other scripts.

testing/btest/scripts/policy/protocols/conn/known-hosts.zeek
```diff
@@ -18,6 +18,40 @@
...
+# @TEST-START-FILE storage-framework-config.zeek
+
+@load policy/frameworks/storage/backend/sqlite
+
+redef Known::use_host_store=T;
+redef Known::enable_hosts_persistence=T;
+
+redef Known::host_store_backend_type = Storage::STORAGE_BACKEND_SQLITE;
+redef Known::host_store_backend_options = [ $sqlite = [
+    $database_path="test.sqlite", $table_name=Known::host_store_prefix ]];
```
awelzel: I'd ditch the explicit redefs  and instead only redef `Cluster::default_store_dir = getenv("TESTBASE")` or so before loading the main known-hosts script. There's some issues with the current defaults.

awelzel (commented):
I had mostly nits, wanted to approve, then tried test deploying and found two snags with the defaults that need fixups.

I had used the following in `local.zeek`, but that caused the processes to error on startup:
```
redef Cluster::default_store_dir = "/opt/zeek-dev-prod/var/lib/zeek/stores/";

# Scripts that do asset tracking.
@load protocols/conn/known-hosts
@load protocols/conn/known-services
@load protocols/ssl/known-certs

redef Known::enable_hosts_persistence = T;
redef Known::enable_certs_persistence = T;
redef Known::enable_services_persistence = T;

redef Known::service_tracking = ALL_HOSTS;
redef Known::host_tracking = ALL_HOSTS;
redef Known::cert_tracking = ALL_HOSTS;
```


* The default backend options use the store name that contains `/` and SQLite doesn't like that, I think that should be the `prefix` option (that otherwise isn't used).
* The `known/` subdirectory may not exist and the `open_backend()` call fails.

For the it seems semi-fair to run `mkdir(fmt(%s/known", Cluster::default_store_dir))` just before or after the `Log::create()` calls in each script.

testing/btest/scripts/policy/protocols/conn/known-hosts.zeek
timwoj: Sounds reasonable. I had it that way originally because it was redefining the entire store configuration, but with the move to using file-based stores by default it's not as meaningful. I also turned on ALL_HOSTS for the known-hosts tests.

awelzel (approved):
:shipit:

A NEWS entry about the new option and the default database locations for visiblity would be cool.

PR #5318, opened 2026-04-07
Deprecate Broker store entry APIs, take two

awelzel:
Redo #5214 now that `&publish_on_change` has been merged.

This deprecates the entry points: `Broker::create_master()`, `Broker::create_clone()` and `Cluster::create_store()`. Also adds a AST traversal to reject invocation of these functions when a non-Broker backend is selected. Seems okay to remove all the API taking 

Deprecate the `use_x_store` options for the `known-*` scripts, too: These have been defaulted to `F` since a long time due to performance concerns and we probably should've deprecated them quite a while ago.


awelzel:
@timwoj / @ckreibich - would one of you be up to :heavy_check_mark: this one?

PR #5334, opened 2026-04-10
Fix state comparison in SMTP_Analyzer::Undelivered

cccs-will:
The condition `state == detail::SMTP_IN_DATA || state == detail::SMTP_CMD_BDAT` mistakenly compares `state` to `SMTP_Cmd::SMTP_CMD_BDAT` instead of `SMTP_State::SMTP_IN_BDAT`.


awelzel:
@cccs-will - thanks! Any chance you have a PCAP available with a gap during a BDAT transfer that could be readily added to the tests?

I'm also wondering if the `enable_rfc822_msg_file_analysis` / `Rfc822MsgDataIn()` code paths should be extended to deal with `Undelivered()`, but that's a separate topic, though the PCAP would be useful :-)

awelzel (approved):
Thanks a lot!

awelzel:
@timwoj - any room for 8.0.x candidates left? :-)

PR #5335, opened 2026-04-10
enable windows tests in all prs

mamaorha:


ci/windows/test.cmd
```diff
@@ -1,11 +1,18 @@
 :: See build.cmd for documentation on this call.
 call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
 
+:: Install btest's Windows dependency.
+pip install multiprocess
```
timwoj, line 5: This should probably move into the Dockerfile so we don't have to install it everytime. We can do that as a follow-up though to avoid more churn on this PR.

timwoj (approved):

PR #5339, opened 2026-04-11
Fix up pattern matching in ZAM's `table[pattern]` logic

ckreibich:
Follow-up to #5331. This fixes the `cluster.publish-on-change.broker-set-element-new-worker-only` test, and expands the regression test to cover the negative case.

(@vpax, fyi)


vpax:
That's more efficient so a clear win. That said, I'm not seeing why it would make a difference in terms of correct operation ... ?

awelzel (approved):

awelzel:
> That said, I'm not seeing why it would make a difference in terms of correct operation ... ?

`LookupPattern() != nullptr` is always true. The returned vector is empty if there was no match.

PR #5340, opened 2026-04-13
Switch docs link-check workflow to run monthly

timwoj:
The link-checking workflow doesn't need to run every week, since it's not something that should change often enough to require it. Switch it to run on the first of the month every month instead.

### DIFF
--- a/.github/workflows/docs-link-check.yml
+++ b/.github/workflows/docs-link-check.yml
@@ -2,7 +2,7 @@ name: Documentation Link Check
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-link-check.yml
ckreibich: Not related to your PR Tim, but I think this is a typo, right? It's not the generate-docs workflow. Might be good to fix while you're in here.

ckreibich (approved):
Wow, I'm not in the habit of checking those reports. I'll start doing so. Looks like it's been failing for a while.

Issue #5320, opened 2026-04-07
ML-DSA Signature Algorithms Mislabeled as RSA-PSS


timwoj:
Thank you to @pmcao for reporting this.

Zeek's TLS and X.509 analyzers contain two exploitable logic errors that cause incorrect or missing logging of Post-Quantum Cryptography (PQC) algorithms in TLS traffic. An attacker using NIST-standardized PQC signature algorithms (ML-DSA per FIPS 204) will appear in Zeek's logs as using classical RSA-PSS, defeating PQC-specific security monitoring rules and masquerading as a different algorithm entirely.

Zeek maps TLS signature algorithm codepoints to human-readable names using a single-byte lookup table that only considers the **low byte** of the 2-byte codepoint. IANA-assigned ML-DSA codepoints (draft-ietf-tls-mldsa) share their low byte with legacy RSA-PSS codepoints:

| PQC Codepoint | Low Byte | Zeek Label (Problem) | Correct Label |
|--------------|----------|--------------------|---------------|
| 0x0904 (ML-DSA-44) | 0x04 | `rsa_pss_sha256` | `ML-DSA-44` |
| 0x0905 (ML-DSA-65) | 0x05 | `rsa_pss_sha384` | `ML-DSA-65` |
| 0x0906 (ML-DSA-87) | 0x06 | `rsa_pss_sha512` | `ML-DSA-87` |

**File:** `scripts/base/protocols/ssl/consts.zeek`, lines 94šC110

Reproducer: [pqc_mislabel.pcap.zip](https://github.com/user-attachments/files/26548590/pqc_mislabel.pcap.zip)

0xxon:
Is it ok to include the reproducer in a testcase for Zeek?

pmcao:
Yes!

0xxon:
It doesn't really matter much for reproduction - but out if curiosity, may I ask how that pcap file was generated?

It has some rather odd artifacts in it - and I wonder if there is a software that does this, or if these are results of hand-editing.

0xxon:
Sorry for the piecewise reply.

After looking at this, I don't think we are doing anything incorrect at the moment. Zeek itself never logs a text-form of signature algorithms. The only logging of signature algorithms in the scripts that are shipped with Zeek is in the `ssl-log-ext` policy script - which logs the numeric values.

[...]

PR #5342, opened 2026-04-14
Update documentation of ssl_extension_signature_algorithm for TLS 1.3

0xxon:
TLS 1.2 and TLS 1.3 use different interpretations of the data available in the event. This PR adds documentation pointing out this difference.

Related to #5320

As noted in #5320, we could also choose to deprecate the hash_algorithms and signature_algorithms tables in ssl consts. I did not do this in this PR, as this is a bit more invasive - but I would be happy to add it :)


timwoj (approved):
> As noted in https://github.com/zeek/zeek/issues/5320, we could also choose to deprecate the hash_algorithms and signature_algorithms tables in ssl consts. I did not do this in this PR, as this is a bit more invasive - but I would be happy to add it :)

How about doing this after 8.2 is out?

0xxon:
Sounds great.

PR #5349, opened 2026-04-14
Use Redis RCE for tutorial so Zeek maintains it

evantypanski:
The tutorial was using an unmaintained Corelight CVE since it was really simple, but the tests weren't even passing. I'd rather we own it, and this is a super simple RCE that I plan on using in a variety of places. So it adds a sense of continuity.

the only downside is that I intentionally force explicitly loading the scripts, since there are two that do similar things. Maybe we can revisit that and pick one or something.

Needs https://github.com/zeek/zeek-training/pull/6 and https://github.com/zeek/packages/pull/272 first

### DIFF
--- a/doc/tutorial/using-packages.rst
+++ b/doc/tutorial/using-packages.rst
@@ -1,4 +1,4 @@
-.. _cve-2022-26809: https://msrc.microsoft.com/update-guide/vulnerability/CVE-2022-26809
+.. _redis_rce: https://github.com/LoRexxar/redis-rogue-server
 .. _zeek package browser: https://packages.zeek.org/
 .. _zkg_docs: https://docs.zeek.org/projects/package-manager/en/stable/index.html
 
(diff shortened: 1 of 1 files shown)


timwoj (approved):

doc/tutorial/using-packages.rst
```diff
@@ -18,16 +18,16 @@ explain how to use ``zkg`` in order to get the packages you need.
  Finding and Installing Packages
 *********************************
 
-Say you're particularly worried about CVE-2022-26809_ (there is nothing
-special about this CVE, it's simply an exploit that we can use for
-demonstration). Zeek does not offer any detection for this
+Say you're particularly worried about `a Redis remote code execution
+<redis_rce_>`_ (there is nothing special about this exploit, it's simply
```
evantypanski, line 22: forgot the underscore here initially, oops

PR #5350, opened 2026-04-14
Fixups after Broker store deprecations

awelzel:
This should fix the zeekctl, clang-tidy and ZAM related tasks for which baselines changed.

I saw the TSAN task failing, too, but will look/address that separately.


timwoj (approved):
That looks a lot better 👍🏼

PR #5355, opened 2026-04-15
Document `zeek` invocation behavior without args

evantypanski:
@ckreibich mentioned this when we were talking

Zeek would read from stdin without arguments, but that behavior is unintuitive, especially because this is stated exactly no where (to my knowledge). So, this documents it.

### DIFF
--- a/doc/tutorial/invoking-zeek.rst
+++ b/doc/tutorial/invoking-zeek.rst
@@ -54,7 +54,7 @@ any packets with checksum errors---due to checksum offloading, this may
 be all packets in a particular direction!
 
 .. _providing_script_values:
- 
+
 *************************
  Providing Script Values
 *************************
@@ -90,6 +90,14 @@ segments:
    # zeek -e "print \"hello\"; print\"there\";"
    hello
    there
+
+Similarly, executing ``zeek`` without arguments causes it to read
+script code from standard input:
+
+.. code:: console
+
+   # echo "print \"Hello, world\!\";" | zeek
+   Hello, world!
 
 As you begin to make more complex adjustments, it quickly becomes
 easier to write your own Zeek scripts. More on this shortly!
--- a/src/Options.cc
+++ b/src/Options.cc
@@ -80,7 +80,7 @@ void usage(const char* prog) {
 
     printf("usage: %s [options] [file ...]\n", prog);
     printf("usage: %s --test [doctest-options] -- [options] [file ...]\n", prog);
-    printf("    <file>                          | Zeek script file, or read stdin\n");
+    printf("    <file>                          | Zeek script file, or read stdin (-)\n");
     printf("    -a|--parse-only                 | exit immediately after parsing scripts\n");
     printf("    -b|--bare-mode                  | don't load scripts from the base/ directory\n");
     printf("    -c|--capture-unprocessed <file> | write unprocessed packets to a tcpdump file\n");


vpax:
> but that behavior is unintuitive

I don't think it is. It's usually given a main input on the command line, and like many Unix utilities, if that's absent, it reads from _stdin_ (without requiring `-`). I do this all the time for quick tests. In addition, there may be scripts out there that build up Zeek snippets and pipe them to zeek which this would break.

evantypanski:
Fair enough, there are a number of programs that take in stdin without a prompt with no arguments.

I think the reason I find it unintuitive is that `zeek -i eth0` will *also* stall similarly, though it does print "listening on eth0" which makes it better. @awelzel also brought up that executing `ruby` reads from stdin, but I actually see that as a bit different because it's "just" a programming language.

Really that brings me to a pretty big gripe, it's ambiguous what Zeek is reading from stdin. I actually initially thought it was a raw pcap stream, since we already have `-e` (and `-`, which I only knew about from looking in code). Even `tcpdump` will read from a network interface without arguments (... which is surprising)! Since Zeek won't print anything as it processes traffic, it's ambiguous.

Also, many Unix utilities will print usage output without arguments:

```
$ awk
usage: awk [-F fs] [-v var=value] [-f progfile | 'prog'] [file ...]
$ echo $?
1
```

I don't particularly see the explicit `-` as a burden. But this PR could just document that empty or `-` mean stdin - the main issue is it's just not documented

vpax:
Not sure this addresses your concern, but I'm fine changing the behavior if `-i` is specified, I don't think that's likely to bite anyone.

awelzel:
> Since Zeek won't print anything as it processes traffic, it's ambiguous.

It's late on Friday, so I might be missing something. What exactly is ambiguous? If you pass, `-e`,  `-r` or `-i`, Zeek will *not* read scripts from stdin unless you pass an explicit `-` as "script file". I don't remember getting too hung up on this. Just because Zeek isn't printing anything doesn't make it ambiguous I'd argue.

We do not read packet data from stdin unless via explicit `-r -` IIRC. And if you use that, will not read a script from stdin. I do see `zeek -r - -` as pretty quirky and a non-functional possibility, but why would you do that :-)

So I'd put a vote out for just keeping things as they are. I've definitely used `echo ... | zeek` or `zeek<\n> <type type> CTRL+D` to run a bit of code and liked the symmetry we have with `ruby`:

```
$ ruby -e 'puts 1'
1
```

```
$ ruby
puts 1
<CTRL+D>
1
```

```
$ ruby -
puts 1
<CTRL+D>
1
```

This we do differently (attempting to read every `-` in the arguments as script file), but I think mainly because we do not have a concept of `argv`.
```
$ ruby -e 'puts ARGV' - 1 2
-
1
2
```

And also this is curious.
```
$ ruby - - -
puts ARGV
<CTRL+D>
-
-
```

Maybe this would be interesting to align a bit more, but that seems fairly cosmetic and as we have no real `ARGV` support anyhow, not sure it's worth it and just a curious detour.

evantypanski:
Okay, I just pushed a docs change, I want this to be mentioned somewhere so I am not really comfortable just doing nothing here.

But, I still think it's a good idea to just print the help output from a user perspective. We can just document the behavior, but here's why:

--------------------

The first thing a user does in the quickstart is read a pcap to get logs:

```
zeek -r quickstart.pcap LogAscii::use_json=T
```

Then, they read from live traffic, which makes Zeek process traffic, rather than terminating:

```
zeek -i en0 -C
```

Then, they run a script on pcap, then live traffic.

We could also have them run just a script (without traffic), that's kind of irrelevant.

At this point, if a user types `zeek` in the command line, what do they think it's doing?

1) Waiting for a pcap file (since they've seen pcap files as input and Zeek monitors network traffic)
2) Reading live traffic on a default interface (this is what `tcpdump` does)
3) Waiting for a script file (since programming languages wait for script files)

To me, the very fact that it's not immediately obvious means that we should make the choice explicit. Of course it's obvious to us now. But, when I was first learning, I thought it was waiting for a pcap file. If I cared I would've dove into documentation to find *exactly nothing* about what it's doing. The only way to find out would be to try inputting a few types of data into it and see what it does.

Even if we say multiple tools take data as stdin, I still think it is ambiguous if `zeek` expects a pcap or a script (at least immediately).

I really don't think I'm out of line to say that others have done the same thing and been confused.

Not to mention: Zeek has multiple times that it executes until stopped. If you run `zeek` without arguments, I don't see how it's clear that "no, this time is different, here it's executing until you give it a script, then it'll stop." It's just inconsistent.

> and liked the symmetry we have with ruby

But Ruby has exactly one thing that you would expect when executed without command line arguments. Actually I'd say I would expect a REPL, but Ruby has separate REPLs

Zeek is not just a programming language, so I don't think it's a fair comparison at all.

The least surprising thing is to just *show you how to correct it*, and that's what printing usage output does. For users who want the old behavior, it's literally a one character fix.

I think we could do some TTY detection to see if `zeek` is being piped into or something, but I don't want to dig into that right now.

awelzel:
>> So I'd put a vote out for just keeping things as they are.
>
> [...] I am not really comfortable just doing nothing here. [...]

Thanks a lot for elaborating. I guess the current behavior just doesn't bother me, but I realize my opinion isn't super strong.

---

It's a separate topic, but a bit in the same corner: I'd find it a lot more exciting if we moved `-i` and `-r` processing and termination policy into script land completely:

```
module PacketSource;

// same thing as what `packet_source()` gives today
global current_pktsrc: PktSrc = Empty();

event zeek_init()
    {
    # If we find -r or -i in args, instantiate a packet source to read packets from.
    for ( _, arg in zeek_args() )
        if ( arg$opt == "-r" )
            current_pktsrc = PacketSource::open(OFFLINE, arg.val);
        else if ( arg$opt == "-i" )
            current_pktsrc = PacketSource::open(LIVE, arg.val)
    }

event PacketSource::opened(pktsrc:: PktSrc)
    {
    }
    
event PacketSource::closed(pktsrc:: PktSrc)
    {
    # Allow other scripts to install the next packet source.
    if ( hook PacketSource::closing(pktsrc) )
        return;

    terminate();
    }
```

If we can get the script layer API right, then this should allow to process a directory of pcaps just by writing an appropriate Zeek or Javascript file that ~installs~ opens the next pcap in a directory once the previous one ran out of packets. That'd would cover the use case of someone placing PCAP files into a spool directory fairly neatly IMO.

awelzel (approved):
for the docs / usage update.

PR #5357, opened 2026-04-15
dns: validate LOC rdlength and enforce RR-local parsing

urvalkheni:
## Summary

Fixes incorrect handling of DNS LOC RDATA length in `ParseRR_LOC()`.

The parser previously performed a minimum-length check against the outer message buffer and could proceed even when the LOC RR `rdlength` was invalid. This could lead to inconsistent parsing behavior and potential reads beyond the RR boundary.

## Changes

- Validate `rdlength == 16` at the start of `ParseRR_LOC()`
- Reject malformed LOC records early, independent of `dns_LOC` or `msg->skip_event`
- Parse LOC RDATA using an RR-local cursor (`rr_data`) and byte counter (`remaining`)
- Ensure no reads occur beyond the RR boundary
- Return failure if any bytes remain unparsed

## Why this is needed

LOC RDATA has a fixed size of 16 bytes. Accepting shorter lengths or relying on outer buffer size can:

- Cause the parser to read into subsequent RRs
- Produce inconsistent or partially parsed values
- Diverge from behavior of other RR handlers (e.g., `ParseRR_A`)

This change enforces strict RR-local parsing and consistent validation.

## Testing

Added a regression test:

- `testing/btest/scripts/base/protocols/dns/loc-invalid-length.zeek`
- `testing/btest/Traces/dns/loc-invalid-length.pcap`

The test verifies:

- Malformed LOC RRs (`rdlength != 16`) are rejected
- `dns_invalid_loc_length` is raised exactly once
- No LOC event output is produced

## Notes

This change is localized to LOC parsing and does not affect other DNS RR handlers.


timwoj:
Thanks for the patch! Was this (and your other two PRs) generated by an LLM? If so, could you let us know what model/version you used? We'll handle it as any other patch, but we're talking about requiring such disclosure in the future and so we're trying to keep track of when people are using such tools to submit patches.

urvalkheni:
@timwoj Thanks for the question!

I used AI tools (ChatGPT) primarily to explore the codebase more quickly and identify potential issues. All fixes were then manually reviewed, refined, and tested by me before submission.

For #5357 (LOC) and #5358 (ExtractName), I researched DNS behavior and implemented the fixes independently, validating them with custom test cases and Zeek builds.

For #5359 (EDNS), I initially implemented a solution myself but wasn’t able to fully resolve the issue. I then used AI again to help identify the problem in my approach, after which I refined and validated the final fix through testing.

Across all three PRs, I made sure to understand the changes (especially DNS parsing and boundary handling) and verified behavior using Zeek builds and btests. I also spent time over the past few weeks exploring the Zeek codebase before preparing these patches.

Happy to clarify any part of the implementation if helpful.

timwoj (approved):

PR #5359, opened 2026-04-15
dns: enforce RR-local bounds for EDNS option parsing

urvalkheni:
## Summary

This fixes an issue in EDNS option parsing where we were not properly
respecting the RR-local `rdlength`.

Right now the parser uses the outer DNS message cursor (`data`, `len`).
If an OPT record has a bad or inconsistent `rdlength`, parsing can run
past the end of the RR and into the next one, which is not correct.

## What changed

The parsing is now strictly limited to the current RR.

- added a local view of the RR:
  - `rr_data`
  - `remaining = rdlength`
- EDNS options are parsed only within that buffer
- option length is checked against remaining bytes:
  - reject if `option_len > remaining - 4`
- switched option parsing to use:
  - `option_data`
  - `option_remaining`
- all option handlers now operate on local buffers only
- outer `data/len` is advanced once, after finishing the RR

If something is off:

- option length goes past available bytes, or
- there are leftover bytes after parsing

→ we raise `EDNS_truncated_option`

## Why

EDNS parsing should stay within the OPT RR’s `rdlength`.

Using the outer buffer could:

- read into the next RR
- break parsing state
- generate misleading events

This just makes EDNS handling consistent with the rest of the DNS parser.

## Testing

Added a small btest:

- `testing/btest/scripts/base/protocols/dns/edns-bad-length.zeek`
- `testing/btest/Traces/dns/dns-edns-bad-length.pcap`
- `testing/btest/Baseline/scripts.base.protocols.dns.edns-bad-length/weird.log`

Checks:

- malformed option length is rejected
- `EDNS_truncated_option` shows up exactly once
- parsing doesn’t continue into the next RR


timwoj:
It looks like your btest on this one is missing one of the files (`weird.log`). Can you update it? Also, can you rebase this on master since I merged your other two PRs?

urvalkheni:
> It looks like your btest on this one is missing one of the files (`weird.log`). Can you update it? Also, can you rebase this on master since I merged your other two PRs?

Thanks! I’ve updated the btest to include weird.log validation, rebased onto the latest master, and squashed the changes into a single commit.

timwoj (approved):

timwoj (approved):

PR #5360, opened 2026-04-16
Docs: add section on security considerations

ckreibich:
Just some general perspective and initial suggestions, in the tutorial.

### DIFF
--- a/doc/security-considerations.rst
+++ b/doc/security-considerations.rst
@@ -1,0 +1,55 @@
+
+.. _security-considerations:
+
+=========================
+ Security Considerations
+=========================
+
+When running Zeek, you should be aware of the security implications involved,
+and consider available options to protect your systems from harm.
+
+Zeek is a complex application. As a network monitor, it processes fundamentally
+untrusted input in every packet it inspects, regardless of whether that's live
+traffic or recorded into a pcap file. By transmitting carefully crafted packets,
+attackers can try to evade Zeek's analysis, or---worse---try to exploit Zeek's
+protocol parsers for nefarious purposes.
+
+A Zeek cluster consists of a set of communicating system processes, possibly on
+multiple machines (we'll say more about this :ref:`later
+<devel-cluster-architectures>`), so as a distributed system Zeek faces
+additional threats from eavesdropping, denial of service, and untrusted
+input. The cluster's internal communication is structured around messages sent
+by `publishing` to various `topics` and received by `subscribing` to such
+topics. That is, all members of the cluster communicate as equals. Add to this
+the fact that Zeek often runs as root, and caution is clearly warranted.
+
+Hardening your setup against these threats is an advanced topic, and you don't
+need to figure out all of this right away. For now, just keep in mind that Zeek
+is security-critical infrastructure and you'll need to treat it as such.
+
+Protective Measures
+===================
+
+The following lists steps you can take to protect your Zeek cluster.
+
+* Isolate your Zeek cluster from the outside world. On physical networks, place
+  Zeek in a dedicated, access-controlled management network, firewalled from
+  production networks. In containerized or virtual networks, apply similar
+  isolation primitives. Double-check external port reachability.
+
+* Configure encrypted cluster communication to protect against message
+  eavesdropping and to authenticate endpoints, particularly if you run a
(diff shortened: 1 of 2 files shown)


doc/tutorial/invoking-zeekctl.rst
```diff
@@ -272,3 +272,53 @@ capable Zeek cluster.
 In the next section, we will go over analyzing logs. If you just want to
 see how logs function when using ZeekControl, see the
 :ref:`zeekcontrol_logs` section.
+
+.. _cluster-security-considerations:
+
+=========================
```
bbannier: This feels a bit buried to me in Tutorial > ZeekControl > Security Considerations; I also have the feeling that it applies to everyone deploying Zeek, not just ZeekControl users or people going through the tutorial. WDYT about instead moving this to Get Started > Quick Start Guide?

0xxon: I concur - the security considerations shold be something that we highlight in a place that people will probably read early :)

ckreibich: No problem. I struggled a bit also with the fact that the actual security advice immediately becomes advanced stuff. How about we give the general security implications context in the quick start guide (I like that idea), and defer the concrete technical steps to a new section in the advanced topics? Then we can expand the latter as needed in the future, too.

evantypanski: Quick start seems fine-ish to me right now, but it only kind of fits. It really should be as small as possible and sell a minimal part of Zeek to show its capabilities. This fits in the "Get Started > Zeek Cluster Setup" section *perfectly* IMO, the problem is I don't think that section is where it should be right now. 

So for now quick start works.. but I would like this to move relatively soon, when the cluster setup section is changed.

ckreibich: How would you feel about a separate section _between_ Quick Start Guide and Zeek Cluster Setup?

ckreibich: (I gave this a try — let me know what you think...)

doc/security-considerations.rst
```diff
@@ -0,0 +1,30 @@
...
+and consider available options to protect your systems from harm.
+
+Zeek is a complex application. As a network monitor, it processes fundamentally
+untrusted input in every packet it inspects, regardless of whether that's live
+traffic or recorded into a pcap file. By transmitting carefully crafted packets,
+attackers can try to evade Zeek's analysis, or---worse---try to exploit Zeek's
+protocol parsers for nefarious purposes.
+
+A Zeek cluster consists of a set of communicating system processes, possibly on
+multiple machines (we'll say more about this :ref:`later
```
evantypanski, line 18: "later" feels wrong here, I don't think people should go reading the advanced section sequentially, but that's nitpicky!

evantypanski (commented):
I'm happier with this than the quickstart. But: it feels less jumbled to just put the advanced stuff into the first section. There can be a separate header for it so it's linkable, then people can see it and go "oh I don't care" then move on. Seems more contained. Both sections are small enough that I think they fit better smooshed together. Otherwise, I'm happy with this!

doc/tutorial/invoking-zeekctl.rst
bbannier: _Advanced_ doesn't feel that awesome to me. This is a topic we'd want everybody deploying Zeek to read, and the advanced section currently holds mostly stuff people could punt on. Most of this is not your fault, so maybe nothing to do. Like Evan suggested below, I think it would be fine to keep this in a single place.

I think we discussed before how we could flatten the tree by getting rid of _Advanced_. Doing that would allow us to make most topics there top-level sections, and one of them should definitely be "security considerations".

ckreibich:
As long as the hardening guide is this short, I can live with that, even though in Get Started people won't know what a cluster backend is, not to mention ZeroMQ or Broker. But I trust people's ability to skip :-)

Take a look at what's there now.

evantypanski (approved):
maybe next in the docs bucket is decide how to introduce clusters and deployments better (ie touch up `cluster-setup` section)

ckreibich:
Agreed! It'd be nice to have an image that doesn't look like the user is hanging themselves. :stuck_out_tongue_winking_eye:

PR #5362, opened 2026-04-16
No more \u00xx escaping in JSON strings

awelzel:
From the NEWS:

Reverted the change that went into Zeek 8.1 where ASCII control characters < 32
were encoded as \u00XX in JSON strings to the prior \xXX style. The escaping style
will become configurable to allow for reliably representing the contained bytes
in strings with Zeek 9.0.

See the discussion at https://github.com/zeek/zeek/discussions/5240 for more background
and feel free the chime in with feedback.


testing/btest/Baseline/scripts.base.frameworks.logging.json-non-printable-escaping/http.log.json
```diff
@@ -1,5 +1,5 @@
...
-{"ts":XXXXXXXXXX.XXXXXX,"uid":"C4J4Th3PJpwUYZZ6gc","id.orig_h":"172.24.0.2","id.orig_p":55142,"id.resp_h":"172.24.0.3","id.resp_p":80,"trans_depth":1,"method":"GET","host":"172.24.0.3","uri":"/non_printable_\u0007","version":"1.0","user_agent":"Python-urllib/3.12","request_body_len":0,"response_body_len":335,"status_code":404,"status_msg":"File not found","tags":[],"resp_fuids":["FLQcdMkvd8G7536U2"],"resp_mime_types":["text/html"]}
+{"ts":XXXXXXXXXX.XXXXXX,"uid":"C4J4Th3PJpwUYZZ6gc","id.orig_h":"172.24.0.2","id.orig_p":55142,"id.resp_h":"172.24.0.3","id.resp_p":80,"trans_depth":1,"method":"GET","host":"172.24.0.3","uri":"/non_printable_\\x07","version":"1.0","user_agent":"Python-urllib/3.12","request_body_len":0,"response_body_len":335,"status_code":404,"status_msg":"File not found","tags":[],"resp_fuids":["FLQcdMkvd8G7536U2"],"resp_mime_types":["text/html"]}
```
awelzel, line 4: This is the one from #3948 which breaks now again, but we had re-opened the issue.

ckreibich (approved):
Wow, I see this turned out to be more subtle than a mere revert. I was looking at the pre-#5014 version and got confused until I saw the other commit in that PR, ecf33fb0a675bd556fa72daf20203fb129ba81c8, that updated the `escape_printable_controls` argument's meaning. :-)

LGTM!

PR #5370, opened 2026-04-17
scan.l: Reject strings with embedded NUL bytes

### LINKED ISSUE #5302: core dump after error
Not a big deal I expect, but looks like zeek fails to recover...the script:
```
module junk;
const a=" ";

event zeek_init() {
        local c=gsub(a,/ /,"^@");
        print c;
        }
```

results in:
internal error in ./main.zeek, line 5: string scanning confused
Aborted (core dumped)

awelzel:
We didn't support this and would run into a InternalError(). Use yyleng for iteration rather than strlen(), check for NUL bytes explicitly and report them as errors.

Initial idea was to support this, but one cannot readily use such strings with, for example, fmt(), so seems better to reject them for now.

Fixes #5302


evantypanski (commented):
can we add `string-nul.zeek` to `.gitattributes` as text, so it doesn't say it's a binary file?

```
testing/btest/language/string-nul.zeek text
```

I think

awelzel:
> I think

Hmm, GitHub doesn't show it, but after putting `diff` as an attribute, I get at least this locally for git show:

```
diff --git a/testing/btest/language/string-nul.zeek b/testing/btest/language/string-nul.zeek
new file mode 100644
index 0000000000..ce2bd4d088
--- /dev/null
+++ b/testing/btest/language/string-nul.zeek
@@ -0,0 +1,10 @@
+# @TEST-EXEC-FAIL: zeek -b %INPUT >out
+# @TEST-EXEC: TEST_DIFF_CANONIFIER=$SCRIPTS/diff-remove-abspath btest-diff .stderr
+
+global s = "a^@b"; # Literal nul in string.
+
+# @TEST-START-NEXT
+event zeek_init()
+       {
+       print fmt("^@%s", "a");
+       }
```

evantypanski (approved):
I think it's just cached, good enough for me :)

PR #5372, opened 2026-04-17
bifs/to_port: Fix calling streq() past input string

awelzel:
Commit eb34c9e984889013e5ce15f3a94c3570c582246d exposed a buffer overread in to_port(). If strtol() stopped on the final NUL character of the input string, we'd previously compare some bytes behind it against the proto.


evantypanski (approved):
funky

PR #5378, opened 2026-04-20
CI: Request 16GB of RAM for Windows builds

timwoj:
I'm hoping this speeds up Windows builds a little. There was a reason we only used 8GB back when we added Windows builds, but I don't remember why it was.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -659,7 +659,7 @@ windows_task:
     os_version: 2019
     cpu: 8
     # Not allowed to request less than 8GB for an 8 CPU Windows VM.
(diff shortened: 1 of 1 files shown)


timwoj:
Cirrus is fine with us requesting this much RAM but it doesn't seem to affect the time it takes to do a Windows build. The theory was that the 8GB level was causing a lot of thrashing. Apparently not. One thing about the Windows build is that it doesn't use ccache. That would likely be a bigger bang for the buck.

No real reason not to merge this but it's not a big help.

0xxon (approved):

PR #5380, opened 2026-04-20
RFC: Adding multi-interface support for zeek.conf

awelzel:
This change adds the possibility to use multiple interfaces with the zeek-systemd-generator and zeek.conf. Because `zeek.conf` is just a flat configuration file, the rule is that every `interface =` line introduces a new interface configuration. Only worker/interface specific options are allowed to follow an `interface =` line (except for the first `interface =` for backwards compat). No interface specific options are inherited from prior configurations.

Maybe sounds complicated, but it essentially allows a `zeek.conf` to look like this:

    loggers   = 3
    proxies   = 7

    # eth1 configuration
    interface            = afpacket::eth1
    workers              = 8
    workers_cpu_list     = 1-8
    worker_args          = AF_Packet::fanout_id=42

    # eth2 configuration
    interface            = afpacket::eth2
    workers              = 8
    workers_cpu_list     = 17-25
    worker_args          = AF_Packet::fanout_id=4711

    # Also listen for VXLAN mirrored packets, not pinning processes.
    interface            = udp::192.168.0.1:4789:vxlan
    workers              = 2

... and it should start 18 workers, 8 listening on eth1 and 8 on eth2 and two for VXLAN encapsulated packets. The eth1 and eth2 workers are pinned and have the AF_Packet fanout_id set to avoid the conflicts.

Didn't polish the implementation all that much and the worker unit files aren't `zeek-worker@1.service` anymore and instead `zeek-interface-1-worker@1.service` (see the example in the docs for a full output), because every interface gets its own template unit and the naming needs to align, otherwise systemd complains. It's not as nice and short as before, but still workable. The `1` from `interface-1` could be replaced with a user option to make it nicer (e.g. so it's `interface-uplink` or so), but for the time being would leave it auto-generated. The worker name are still `worker-1` to `worker-n`. I'd rather steer clear of encoding the interface name or so into worker names, as I don't think names should be all that important and information about the interface can 1) come from `packet_source()` and 2) can be published/discovered dynamically (#5208).

The other format I've been contemplating for way too long has been horizontally placing interfaces, but that's just not pretty.

    interface = af_packet::eth1 ; af_packet::eth2
    workers = 8 ; 8
    workers_cpu_list = 1-8 ; 17-25


awelzel:
I've had a conversation with @bbannier. Also @ckreibich provided some feedback.

Will pivot/steer towards something with a bit more ini-like structure and then open v2.

PR #5385, opened 2026-04-21
Cap all parsed Redis fields to 250 bytes

evantypanski:
Some commands (like `FULLRESYNC`) will have huge outputs, we should not output all of it. It's unfortunate this is not configurable, but it limits copies and the size of output logs.

There is a max bytes limit in the parser, but it's really high. It shouldn't stop parsing, just shouldn't forward it to Zeek, I think.


awelzel:
> It's unfortunate this is not configurable, but it limits copies and the size of output logs.

There is `MQTT::max_payload_size` that seems related. So there's precursor to having such capping configurable.

https://github.com/zeek/zeek/blob/f09461abd3047088487186275da73604e0403613/src/analyzer/protocol/mqtt/commands/publish.pac#L33-L41

On the QUIC side, there's `QUIC::max_discarded_packet_events` as a `const &redef` to influence behavior.

Could this be some `Redis::max_value_size`? I'm not sure I like the `QUIC` pattern all that much with the optional, but it's workable.

```
global _max_value_size: optional<int64>;

function max_value_size(): int64 {
    if ( ! _max_value_size )
        _max_value_size = zeek::as_int(zeek::get_value("Redis::max_value_size"));
    return *_max_value_size;
}
```

evantypanski:
I was actually opposed to it since I don't want a dependency on Zeek. It's really useful to run this parser with `spicy-driver` etc. with no fuss. So, I made a workaround with `@if HAVE_ZEEK` after messing around with options this morning. I think we should stick to this pattern when using Zeek integration with Spicy (or something similar)

What do you think @awelzel ?

src/analyzer/protocol/redis/redis.spicy
```diff
@@ -6,6 +6,21 @@ module Redis;
 
 import RESP;
 
+# Max size we will stringify into. This can limit logs output size and copies.
+# Allow configuring max size via Zeek, but don't include if building separate.
+@if HAVE_ZEEK
+import zeek;
+global _max_value_size: optional<uint64>; function max_value_size(): uint64 {
```
awelzel, line 13: LGTM. Not sure about being strict about having the `HAVE_ZEEK` guard. Personally, I never use `spicy-driver` for parsers within Zeek and only go via PCAPs.

Also, elsewhere `0` means unlimited (including the `max_discarded_packet_events`, but also some limits used in SMB or HTTP). Could see using UINT64_MAX in that case, then a user can set `Redis::max_value_size=0` to not cap.

awelzel (approved):
LGTM.

src/analyzer/protocol/redis/redis.spicy
evantypanski: good idea, I capped it at `RESP::MAX_SIZE` since that's a hardcoded limit anyway

Probably don't have to be strict about `HAVE_ZEEK`, for some parsers it just doesn't make as much sense. This one is a simple text protocol so it probably makes more sense here

awelzel (approved):

PR #5386, opened 2026-04-21
Fix spicy.preprocessor btest

timwoj:
The spicy.preprocessor btest used version 9.0 as a "far enough in the future" version to check that the `@if` directive works. That's clearly no longer in the future, so extend it out to version 20.0 instead.


awelzel (approved):
Fun.

I hadn't fully realized we've just started working towards 9.0 🎉

PR #5387, opened 2026-04-21
Add log policy that tracks multicast connections to sources from IGMP

timwoj:
This adds a policy script that adds a new log file to track multicast connections where we know more information about the participants from IGMP. See the btest baseline for an example of the output. The only real question I have about it is whether the membership tracking makes sense to keep in the policy, or if it could be moved up into the main scripts but controlled by a flag.


scripts/policy/protocols/igmp/igmp-conn-log.zeek
```diff
@@ -0,0 +1,111 @@
...
+@load base/packet-protocols/igmp/types
+
+module IGMP;
+
+export {
+	redef enum Log::ID += { IGMP_CONN_LOG };
+
+	global log_policy_igmp_conn: Log::PolicyHook;
+
+	type IGMPConnInfo: record {
```
ckreibich: I suggest naming this just `Info`, and the log ID `LOG`, for consistency with other modules / log streams that just have a single log.

`##`-comments on the fields would be great too — that way we get explanation into log schemas and docs.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
```diff
@@ -0,0 +1,111 @@
...
+		ts: time &log;
+		cid: string &log;
+		orig_h: addr &log;
+		resp_h: addr &log;
```
ckreibich: I'm struggling with 'resp_h" here because it's the multicast address, which doesn't respond. I could see something like `multicast_addr` and `multicast_p`?

scripts/policy/protocols/igmp/igmp-conn-log.zeek
```diff
@@ -0,0 +1,111 @@
...
+		resp_p: port &log;
+		igmp_srcs: set[addr] &log;
```
ckreibich: Could I talk you into `receivers` or `members`, here and below where you use "sources"`? It suggests it's sending something, but in terms of multicast it isn't. Is your intuition that they're sources of IGMP membership reports?

scripts/policy/protocols/igmp/igmp-conn-log.zeek
```diff
@@ -0,0 +1,111 @@
...
+# Map connections to multicast group address and port number, but separately.
+# Multiple ports can be used on the same multicast group address.
+global igmp_conns: table[addr] of set[conn_id];
+
+# Map multicast group addresses to the addresses of the members.
+# TODO: This feels like it should be in the main IGMP script instead of here.
+global igmp_sources: table[addr] of set[addr];
+
+event zeek_init() &priority=5 {
+	Log::create_stream(IGMP::IGMP_CONN_LOG, Log::Stream($columns=IGMPConnInfo, $ev=log_igmp_conn, $path="igmp_conn", $policy=log_policy_igmp_conn));
```
ckreibich: I wonder whether we should just call it multicast.log.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
```diff
@@ -0,0 +1,111 @@
...
+		add igmp_conns[c$id$resp_h][c$id];
+
+		if ( c$id$resp_h in igmp_sources )
+			c$igmp_srcs = copy(igmp_sources[c$id$resp_h]);
+		else
+			c$igmp_srcs = set();
+	}
+}
+
+event connection_state_remove(c: connection) {
```
ckreibich: Since this handler will only be productive for the (hopefully few) multicast connections it'd probably be better to run this via `Conn::register_removal_hook()`.

(I wonder whether this whole topic warrants a new `new_multicast_connection()` event — not immediately sure about that one, but it would be nicely surgical.)

scripts/policy/protocols/conn/multicast-participants.zeek
```diff
@@ -0,0 +1,111 @@
...
+	igmp_srcs: set[addr] &optional;
+};
+
+# Map connections to multicast group address and port number, but separately.
+# Multiple ports can be used on the same multicast group address.
+global igmp_conns: table[addr] of set[conn_id];
+
+# Map multicast group addresses to the addresses of the members.
+# TODO: This feels like it should be in the main IGMP script instead of here.
+global igmp_sources: table[addr] of set[addr];
```
ckreibich, line 50: I wondered here about a `&create_expire` on the set, to be safe, but I don't see a good value we'd use...

scripts/policy/protocols/conn/multicast-participants.zeek
```diff
@@ -0,0 +1,111 @@
...
+
+function add_igmp_source(source: addr, group: addr ) {
+	if ( group !in igmp_sources )
+		igmp_sources[group] = set();
+
+	add igmp_sources[group][source];
+
+	if ( group in igmp_conns ) {
+		for ( cid in igmp_conns[group] ) {
+			local c = lookup_connection(cid);
```
ckreibich, line 106: This can fail, in which case it returns a dummy connection that won't have `c$igmp_srcs`, so checking for its existence is a good safeguard (also in general).

testing/btest/Traces/README
```diff
@@ -102,3 +102,6 @@ Trace Index/Sources:
 - ldap/ldap-max-recursion.pcap
   Generated by Phuong Cao (@pmcao) as part of a security report. Used to test
   limitations on recursion when parsing LDAP search filters.
+- igmp/home-multicast-short.pcap
+  A snippet of a larger pcap generated by Christian Kreibich (@ckreibich) from his home
+  network.
```
ckreibich: :-)

ckreibich (changes requested):
Thanks Tim. I mostly have nits around naming right now, take a look.

I noticed a complication — given that we're indexing data structures with plain `addr`s, this is technically a use-case for our context-aware `Endpoint` idea. This is complicated by the fact that the IGMP events won't have such context, since they're not connections. The multicast connections do, though. Seems okay to ignore for now.

And sorry to nag, but it'd be better to indent the scripts in our usual style...

scripts/policy/protocols/igmp/igmp-conn-log.zeek
timwoj: I'd think for policies we'd want them semi-uniquely named in case someone adds a default log to IGMP and conflicts.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
timwoj: > Is your intuition that they're sources of IGMP membership reports?

Yep, that's exactly it. `receivers` might be ok. `particpants` would be ok too, but seems long for a log field.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
timwoj: It's not all multicast though. It's just the ones that we have IGMP information for the participants.

scripts/policy/protocols/conn/multicast-participants.zeek
timwoj: I guess this is a wait-and-see kind of thing. I don't think that set would grow very large on a well-behaved IGMP environment, since you're only going to be adding (and removing) the sources that are reported, but I'd need to see it on a big network.

scripts/policy/protocols/conn/multicast-participants.zeek
timwoj: Ah, good catch. I thought i had hit all of those cases already but I missed this one.

scripts/policy/protocols/conn/multicast-participants.zeek
ckreibich: I agree, yeah.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
ckreibich: That's true. And we probably shouldn't park it in `policy/protocols/igmp` in that case either. On the other hand if we added MLD support, say, would we make that a different log? Probably not. I mainly dislike `igmp_conn` because it sounds like IGMP connections.

`multicast_igmp` as a compromise?

scripts/policy/protocols/igmp/igmp-conn-log.zeek
timwoj: I changed it to `participants`. I feel like that matches well what's happening here.

scripts/policy/protocols/igmp/igmp-conn-log.zeek
timwoj: Maybe move it into scripts/policy/protocols/multicast? `multicast_igmp` doesn't feel good either if we're talking about maybe supporting MLD also.

timwoj: How about moving it to `scripts/policy/protocols/conn`, renaming it to `multicast.zeek`, and naming the log `multicast.log` but with a disclaimer that it only includes the connections where we have IGMP/MLD data?

ckreibich: Nice — I like that!

scripts/policy/protocols/conn/multicast-participants.zeek
```diff
@@ -0,0 +1,133 @@
+# Logs hosts known to take part in multicast conversations based on IGMP data. This is not
+# all multicast connections, but just the ones we know the participants.
```
ckreibich:
```suggestion
# all multicast connections, but just the ones for which Zeek knows active recipients.
```

scripts/policy/protocols/conn/multicast-participants.zeek
```diff
@@ -0,0 +1,133 @@
...
+	{
+	# Multicast addresses only show up in the resp fields.
+	if ( c$id$resp_h in 224.0.0.0/4 )
+		{
+		if ( c?$multicast_srcs && |c$multicast_srcs| > 0 )
+			Log::write(MULTICAST_PARTICIPANTS_LOG, MulticastParticipantsInfo(
+			    $ts=c$start_time, $cid=c$uid, $orig_h=c$id$orig_h,
+			    $group_addr=c$id$resp_h,
+			    $group_p=c$id$resp_p,
+			    $participants=c$multicast_srcs));
```
ckreibich, line 69: I had a thought here Tim ... the participants are currently an unordered set, so the ordering in the logs depends on Zeek's seeds. So I'm wondering whether we should make participants a `vector of addr` and sort them by address upon log write. (In JSON, sets become arrays anyway.)

It's not a must — I looked through our log schemas and there's a mix of sets and vectors with ordering implied by sequencing in protocol interaction. But it seems to me that even short lists of addresses are more easily scanned if there's some logic to the ordering — what do you think?

testing/btest/scripts/policy/protocols/conn/multicast-participants.zeek
```diff
@@ -0,0 +1,6 @@
+# @TEST-DOC: Tests the output from the multicast-participants policy script
+
+# @TEST-EXEC: zeek -C -r $TRACES/igmp/home-multicast-short.pcap %INPUT
```
ckreibich:
```suggestion
# @TEST-EXEC: zeek -r $TRACES/igmp/home-multicast-short.pcap %INPUT
```
I noticed that my pcap had checksum errors ... here's one that has those fixed.

[home-multicast-short.pcap.gz](https://github.com/user-attachments/files/27186292/home-multicast-short.pcap.gz)

ckreibich (commented):
Okay Tim, I'm basically ready to approve (again :-). I have a question/suggestion regarding the `participants` fields and ordering — take a look? 

Minor stuff: we still need `##` comments on the log fields to explain them. (And another docs refresh, fwiw — just flagging to make sure.)

scripts/policy/protocols/conn/multicast-participants.zeek
timwoj: I disagree. The order of when the participants joined the group shouldn't matter, just whether they received the packets.

ckreibich (approved):
Thanks Tim!

PR #5388, opened 2026-04-21
Import zeek/spicy-zip analyzer for ZIP archives

timwoj:
This was brought up in #4519 about whether it would be useful to have the zip extraction from https://github.com/zeek/spicy-zip in-tree. I agree that it would be and migrated the code. I haven't made any changes to the existing plugin code here outside of fixing CMakeLists.txt to make it build and link correctly, and to fix the btests so they run correctly.


testing/btest/scripts/base/files/zip/basic.zeek
```diff
@@ -0,0 +1,22 @@
+# @TEST-EXEC: zeek -r ${TRACES}/zip/test.pcap frameworks/files/extract-all-files frameworks/files/hash-all-files %INPUT
+# @TEST-EXEC: for i in extract_files/*; do (printf "$i "; wc -c "$i" | awk '{print $1}'); done | sort >extracted.log
+# @TEST-EXEC: for i in files.log extracted.log .stdout; do cat $i | sed 's#\(extract-[^-]*\)-[^-]*-#\1-xxx-#g' | sed 's#F[A-Za-z0-9]\{16,17\}#XXXXXXXXXXXXXXXXX#g' >$i.tmp && mv $i.tmp $i; done
+#
+# @TEST-EXEC: TEST_DIFF_CANONIFIER='zeek-cut uid mime_type source filename extracted extracted_size ftime' btest-diff files.log
+#
+# @TEST-EXEC: btest-diff .stdout
+# @TEST-EXEC: btest-diff extracted.log
+#
+# @TEST-DOC: Test ZIP analyzer with a download of a simple ZIP file that contains an entry with data descriptor (i.e., its compressed size is not known in advance).
```
awelzel: Move TEST-DOC more towards teh top?

testing/btest/scripts/base/files/zip/basic.zeek
```diff
@@ -0,0 +1,22 @@
+# @TEST-EXEC: zeek -r ${TRACES}/zip/test.pcap frameworks/files/extract-all-files frameworks/files/hash-all-files %INPUT
+# @TEST-EXEC: for i in extract_files/*; do (printf "$i "; wc -c "$i" | awk '{print $1}'); done | sort >extracted.log
+# @TEST-EXEC: for i in files.log extracted.log .stdout; do cat $i | sed 's#\(extract-[^-]*\)-[^-]*-#\1-xxx-#g' | sed 's#F[A-Za-z0-9]\{16,17\}#XXXXXXXXXXXXXXXXX#g' >$i.tmp && mv $i.tmp $i; done
+#
+# @TEST-EXEC: TEST_DIFF_CANONIFIER='zeek-cut uid mime_type source filename extracted extracted_size ftime' btest-diff files.log
```
awelzel: Nit. This could be a bit streamlined:

```suggestion
# @TEST-EXEC: btest-diff-cut -m uid mime_type source filename extracted extracted_size ftime files.log
```

src/file_analysis/analyzer/zip/zip.evt
```diff
@@ -0,0 +1,11 @@
+# See the file "COPYING" in the main distribution directory for copyright.
+
+file analyzer spicy::ZIP:
```
awelzel:
```suggestion
file analyzer ZIP:
```

This should make `SPICY_ZIP` become `ZIP` in the logs. We stopped adding the `spicy::` prefix in most places.

scripts/base/files/zip/main.zeek
```diff
@@ -0,0 +1,44 @@
...
+	};
+}
+
+redef record Files::Info += {
+	## File timestamp
+	ftime: time &optional &log;
+};
+
+# Maps FIDs to their meta data.
+global archives: table[string] of ZIP::File;
```
awelzel: As this table isn't exported, I'm wondering if we instead could attach the "meta" information to the `fa_file`, then it can also be found/used by other event handlers that only receive `fa_file`.

```
redef record fa_file += {
    zip_meta: ZIP::File &optional
}

event ZIP::file(f: fa_file, meta: ZIP::File)
    {
	f$zip_meta = meta;
	}

event file_state_remove(f: fa_file)
    {
    if ( ! f?$zip_meta )
        return;

    f$info$filename = f$zip_meta$filename;
	f$info$ftime = meta$time_;
    }

**Review by awelzel — COMMENTED** [review]
Thanks! I've left some small nits, but I'd happy to approve.

As long as we have an entry in the NEWS how to disable it, seems a good moment to add this early in the cycle.

File: scripts/base/files/zip/main.zeek (hunk 4 above)
Comment (timwoj) [reply scripts/base/files/zip/main.zeek]: Ah yeah, that seems cleaner.

**Comment by timwoj** [comment]
@awelzel I fixed up your comments as well as some other btest baselines that were failing, if you want to take a full pass at this. It's broken on Windows still though.

> As long as we have an entry in the NEWS how to disable it

I added a NEWS entry. It can be disabled like any other file analyzer with `Files::disable_analyzer(Files::ANALYZER_ZIP);`. Is there a reason to document that in NEWS? I don't see anywhere else that we did that previously.

File: scripts/base/files/zip/main.zeek [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+		## Comment associated with file.
+		comment: string;
+		## Compression type
+		compression: ZIP::CompressionMethod;
+		## True if encrypted
+		encrypted: bool;
+	};
+}
+
+redef record Files::Info += {
```
awelzel: Think we shouldn't do this one by default. It adds a column to files.log by default that's only populated from ZIPs.

I don't know a better way, I'd just skip for now. If someone wants this, they can write their own policy script / handler.

scripts/base/files/zip/main.zeek
```diff
@@ -0,0 +1,42 @@
...
+	## File timestamp
+	ftime: time &optional &log;
+};
+
+redef record fa_file += {
+	zip_meta: ZIP::File &optional;
```
awelzel: On second thought - just so it aligns with the `ZIP::File` type. I could see meta being `zip_meta` also for the `ZIP::file()` handler. as meta isn't used in the parser, so it's a new term that could probably be avoided.

```suggestion
	zip_file: ZIP::File &optional;
```

testing/btest/Baseline/coverage.record-fields/out.default
```diff
@@ -262,6 +262,7 @@ connection {
               * extracted_cutoff: bool, log=T, optional=T
               * extracted_size: count, log=T, optional=T
               * filename: string, log=T, optional=T
+              * ftime: time, log=T, optional=T
```
awelzel: As mentioned, this I'd skip. Lets not change `files.log` for an experimental zip analyzer.

src/file_analysis/analyzer/zip/zeek_zip.spicy
```diff
@@ -0,0 +1,26 @@
+# See the file "COPYING" in the main distribution directory for copyright.
+
+module Zeek_ZIP;
+
+import ZIP;
+
+import zeek;
+import zeek_file;
+
+on ZIP::File::compressed_size {
```
awelzel: I don't know if this is needed, but `compressed_size` in `zip.spicy` also has a hook attached and this one should probably run after it. Might not be needed, but this makes it more explicit which one runs first and shouldn't cost anything.

```suggestion
on ZIP::File::compressed_size &priority=-5 {
```

src/file_analysis/analyzer/zip/zip.spicy
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    uncompressed_size: uint32;
+    filename_length: uint16;
+    extra_field_length: uint16;
+    filename: bytes &size=self.filename_length;
+    extra_field: bytes &size=self.extra_field_length;
+
+    switch (self.flags.has_descriptor) {
+        0 -> : bytes &size=self.compressed_size &chunked -> self.content;
```
awelzel, line 131: I was a bit scared of recursive ZIP files, but there's `Spicy::max_file_depth` with a default of 5, so should be okay.

testing/btest/scripts/base/files/zip/basic.zeek
```diff
@@ -0,0 +1,21 @@
+# @TEST-DOC: Test ZIP analyzer with a download of a simple ZIP file that contains an entry with data descriptor (i.e., its compressed size is not known in advance).
+#
+# @TEST-EXEC: zeek -r ${TRACES}/zip/test.pcap frameworks/files/extract-all-files frameworks/files/hash-all-files %INPUT
+# @TEST-EXEC: for i in extract_files/*; do (printf "$i "; wc -c "$i" | awk '{print $1}'); done | sort >extracted.log
+# @TEST-EXEC: for i in files.log extracted.log .stdout; do cat $i | sed 's#\(extract-[^-]*\)-[^-]*-#\1-xxx-#g' | sed 's#F[A-Za-z0-9]\{16,17\}#XXXXXXXXXXXXXXXXX#g' >$i.tmp && mv $i.tmp $i; done
+#
+# @TEST-EXEC: btest-diff-cut -m uid mime_type source filename extracted extracted_size ftime files.log
```
awelzel: I was surprised the baselines had the same uid over and over again. Maybe include fuid, too?
```suggestion
# @TEST-EXEC: btest-diff-cut -m uid fuid mime_type source filename extracted extracted_size ftime files.log
```

testing/btest/scripts/base/files/zip/basic.zeek
```diff
@@ -0,0 +1,21 @@
...
+# @TEST-EXEC: btest-diff .stdout
+# @TEST-EXEC: btest-diff extracted.log
+
+@load base/files/zip
+
+event ZIP::file(f: fa_file, meta: ZIP::File)
```
awelzel:
```suggestion
event ZIP::file(f: fa_file, zip_file: ZIP::File)
```

testing/btest/scripts/base/files/zip/descriptor.zeek
```diff
@@ -0,0 +1,26 @@
...
+# @TEST-EXEC: for i in extract_files/*; do (printf "$i "; wc -c "$i" | awk '{print $1}'); done | sort >extracted.log
+# @TEST-EXEC: for i in files.log extracted.log .stdout; do cat $i | sed 's#\(extract-[^-]*\)-[^-]*-#\1-xxx-#g' | sed 's#F[A-Za-z0-9]\{16,17\}#XXXXXXXXXXXXXXXXX#g' >$i.tmp && mv $i.tmp $i; done
+#
+# @TEST-EXEC: btest-diff-cut -m  uid mime_type source filename extracted extracted_size ftime files.log
+# @TEST-EXEC: btest-diff .stdout
+# @TEST-EXEC: btest-diff extracted.log
+
+@load base/files/zip
+
+event ZIP::file(f: fa_file, meta: ZIP::File)
```
awelzel:
```suggestion
event ZIP::file(f: fa_file, zip_file: ZIP::File)
```

testing/btest/scripts/base/files/zip/nested.zeek
```diff
@@ -0,0 +1,21 @@
...
+# @TEST-EXEC: for i in extract_files/*; do (printf "$i "; wc -c "$i" | awk '{print $1}'); done | sort >extracted.log
+# @TEST-EXEC: for i in files.log extracted.log .stdout; do cat $i | sed 's#\(extract-[^-]*\)-[^-]*-#\1-xxx-#g' | sed 's#F[A-Za-z0-9]\{16,17\}#XXXXXXXXXXXXXXXXX#g' >$i.tmp && mv $i.tmp $i; done
+#
+# @TEST-EXEC: btest-diff-cut -m uid mime_type source filename extracted extracted_size ftime files.log
+# @TEST-EXEC: btest-diff .stdout
+# @TEST-EXEC: btest-diff extracted.log
+
+@load base/files/zip
+
+event ZIP::file(f: fa_file, meta: ZIP::File)
```
awelzel: Sorry, just don't see where "meta" is coming from :-)

```suggestion
event ZIP::file(f: fa_file, zip_file: ZIP::File)
```

NEWS
```diff
@@ -14,6 +14,10 @@ Breaking Changes
 New Functionality
 -----------------
 
+- Imported the zeek/spicy-zip file analyzer into the main tree. This analyzer outputs
+  information about ZIP files found in other protocols, including information about the
+  contents of such archives.
+
```
awelzel, line 49: This is what we had for LDAP:

```
  If you observe unusually high CPU consumption or other issues due to this
  analyzer being enabled by default, the easiest way to disable it is via the
  ``Analyzer::disabled_analyzers`` const as follows:

        redef Analyzer::disabled_analyzers += {
                Analyzer::ANALYZER_LDAP_UDP,
                Analyzer::ANALYZER_LDAP_TCP,
        };

```

And this for QUIC:

```
  To disable the analyzer in case of issues, use the following snippet:

        redef Analyzer::disabled_analyzers += {
                Analyzer::ANALYZER_QUIC,
        };
```

This should work also with `Files::FILES_ANALYZER_ZIP`.

NEWS
```diff
@@ -14,6 +14,10 @@ Breaking Changes
 New Functionality
 -----------------
 
+- Imported the zeek/spicy-zip file analyzer into the main tree. This analyzer outputs
```
awelzel, line 39: Might make sense to mention the recursive behavior as a feature, too. For example, the ZIP [from  issue #11 in spicy-zip](https://github.com/zeek/spicy-zip/issues/11) has a zip containing an exe file and there's actually a pe.log created of that inner PE file. I find this pretty cool! `Spicy::max_file_depth` is the cutoff.

The issue above persist, so we should probably move it into the Zeek repo.

awelzel (approved):
> Is there a reason to document that in NEWS? I don't see anywhere else that we did that previously.

Left some notes in the comments. Otherwise many nits, but I think it's cool :-)

scripts/base/files/zip/main.zeek
timwoj: I'd definitely be fine skipping it because it adds a lot of churn to unrelated btest baselines too.

NEWS
timwoj: Huh, I completely missed that. I grepped for `disable_analyzer`, since I assumed it'd be called directly.

NEWS
timwoj: > The issue above persist, so we should probably move it into the Zeek repo.

Yep, I was going to move it over once this is merged plus add a note to that repo that it won't be needed after Zeek 9.0.

PR #5390, opened 2026-04-22
Remove stale btest baselines

J-Gras:


evantypanski:
Did you do this manually, or with a script? I remember I ran [this script](https://gist.github.com/evantypanski/f9fc099f32fad47ca5c0eae97e966ad1) a while ago (#4649) but I knew it missed some cases. Having something run in CI would be nice, I think.

(when CI is back!!!)

bbannier:
> Having something run in CI would be nice, I think.

We do this with a [pre-commit hook](https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/.pre-commit-config.yaml#L30-L37) via a [script](https://github.com/zeek/spicy/blob/9903fc6c26434319484950a3c3aaadc10ff1314a/tests/Scripts/stray_baselines.py) over in Spicy. That script is in Python (already a Spicy dependency) and uses `btest -l` managed by pre-commit.

J-Gras:
> Did you do this manually, or with a script?

I found one instance and then asked AI to list all stale ones. Actually, it came up with a script. However, I didn't review the script. Just verified the result by running the tests.

testing/btest/Baseline/scripts.base.protocols.http.http-11-request-then-cruft/analyzer.log
```diff
@@ -1,11 +0,0 @@
-### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
ckreibich, line 1: Something funny is going on here Jan: there is currently a test `testing/btest/scripts/base/protocols/http/http-11-request-then-cruft.pcap` — but that's a test, not a pcap. It runs when invoked on the command line but gets skipped in bulk runs. The test passes and leaves analyzer.log, http.log, and weird.log. So the fix is to rename the test, not remove these baselines.

I'm guessing this slipped in during the recent bulk-renaming.

ckreibich (changes requested):
Thank you!

testing/btest/Baseline/scripts.base.protocols.http.http-11-request-then-cruft/analyzer.log
J-Gras: Haha, that's a sneaky one...

timwoj: Whoops, that was probably my bad.

timwoj (approved):

PR #5391, opened 2026-04-22
Include zeek-packet-source-udp as builtin plugin

awelzel:
I chatted with Christian a while ago and he wasn't completely opposed. The integration is straightforward and IMO this is a nice feature for the cloudier environments.

---

@ckreibich or others - what do you think?


ckreibich:
Yes, strong ack from me. Looking it over now.

ckreibich (approved):
LGTM. My only request would be to mention it in the docs, but that can come later. I could see a new "popular customizations" section for packet sources.

PR #5392, opened 2026-04-23
Make LDAP recursion tests more robust

bbannier:
We previously would baseline the full exerpts from `debug.log` when all we cared was that a dedicated violation occurred, so this patch uses a looser approach to look for the violation.

AFAICT these tests never worked on my machine, so for it to pass in CI I suspect in CI we never ran in a setup which satisfied the test requirements.


testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
```diff
@@ -4,5 +4,4 @@
 # @TEST-REQUIRES: test "$($BUILD/zeek-config --build_type)" = "debug"
 
 # @TEST-EXEC: zeek -B spicy -r ${TRACES}/ldap/ldap-max-recursion.pcap
-# @TEST-EXEC: grep "max filter recursion depth" debug.log > debug-filtered.log
-# @TEST-EXEC: TEST_DIFF_CANONIFIER="$SCRIPTS/diff-remove-abspath | $SCRIPTS/diff-remove-timestamps" btest-diff debug-filtered.log
+# @TEST-EXEC: grep "error during parsing, triggering analyzer violation: LDAP exceeded max filter recursion depth" debug.log
```
awelzel: I'd think the more usual approach is to implement `analyzer_violation_info()` in a zeek script and btest-diff stdout. Also works in non-debug builds.

bbannier: Good idea. The ldap-specific PCAP triggers a lot of recursion-related violations when I think we'd care only about seeing one, but that's a separate issue.

testing/btest/scripts/base/protocols/ldap/asn1_max_recursion.zeek
```diff
@@ -1,8 +1,12 @@
...
-# @TEST-EXEC: zeek -B spicy -r ${TRACES}/ldap/asn1-max-recursion.pcap
-# @TEST-EXEC: grep "exceeded max recursion depth" debug.log > debug-filtered.log
-# @TEST-EXEC: TEST_DIFF_CANONIFIER="$SCRIPTS/diff-remove-abspath | $SCRIPTS/diff-remove-timestamps" btest-diff debug-filtered.log
+# @TEST-EXEC: zeek -r ${TRACES}/ldap/asn1-max-recursion.pcap %INPUT >output
+# @TEST-EXEC: TEST_DIFF_CANONIFIER="$SCRIPTS/diff-remove-abspath" btest-diff output
+
+event analyzer_violation_info(atype: AllAnalyzers::Tag,
+    info: AnalyzerViolationInfo)
+	{
+	print atype, info$reason;
```
awelzel:
```suggestion
	print info$c$uid, atype, info$reason;
```

testing/btest/Baseline/scripts.base.protocols.ldap.ldap_max_recursion/output
```diff
@@ -0,0 +1,101 @@
...
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+Analyzer::ANALYZER_LDAP_TCP, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
```
awelzel: That is strange, but including `info$c$uid` clears it: These are all individual connections rather than a single connection. So that makes sense, but the pcap could probably be smaller'ish.

testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
```diff
@@ -1,8 +1,12 @@
...
-# @TEST-REQUIRES: test "$($BUILD/zeek-config --build_type)" = "debug"
 
-# @TEST-EXEC: zeek -B spicy -r ${TRACES}/ldap/ldap-max-recursion.pcap
-# @TEST-EXEC: grep "max filter recursion depth" debug.log > debug-filtered.log
-# @TEST-EXEC: TEST_DIFF_CANONIFIER="$SCRIPTS/diff-remove-abspath | $SCRIPTS/diff-remove-timestamps" btest-diff debug-filtered.log
+# @TEST-EXEC: zeek -B spicy -r ${TRACES}/ldap/ldap-max-recursion.pcap %INPUT >output
+# @TEST-EXEC: TEST_DIFF_CANONIFIER="$SCRIPTS/diff-remove-abspath" btest-diff output
+
+event analyzer_violation_info(atype: AllAnalyzers::Tag,
+    info: AnalyzerViolationInfo)
```
awelzel, line 9: Total nit, if you use a formatter, can you bump it to 120 cols or so. This line break is much worse than the 4 chars or so it'd overrun the 80 cols limt.

awelzel (approved):

testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
```diff
@@ -1,8 +1,12 @@
...
+	{
+	print atype, info$reason;
```
awelzel:
```suggestion
	print info$c$uid, atype, info$reason;
```

bbannier: The conn ID is only useful since the PCAP we work with raises more than one violation. In an ideal world we wouldn't need to baseline it, but I'll add it here.

testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
bbannier: I agree, but this is straight `zeek-format`. I feel it should just format this as

```zeek
event analyzer_violation_info(
    atype: AllAnalyzers::Tag,
    info: AnalyzerViolationInfo)
```

Leaving as is.

awelzel (approved):

PR #5395, opened 2026-04-23
support script loading from a zip

mamaorha:
adding the ability to load zeek scripts directly from in-memory zip, as we would like to avoid "extracting the zip" and be open to tampering


src/scan.l
```diff
@@ -816,37 +828,67 @@ static int load_files(const char* orig_file) {
...
-            if ( ! f )
-                zeek::reporter->FatalError("can't open %s", file_path.c_str());
-
-            // open_file() / open_package() modify file_path, so can only
-            // do this check after opening the files.
-            zeek::detail::ScannedFile sf(file_stack.length(), file_path);
-            if ( sf.AlreadyScanned() ) {
-                fclose(f);
-                return 0;
+            // Try to load from ZIP archive first.
```
awelzel: @mamaorha - have you considered going via the `HookLoadFileExtended()` plugin hook? See a few lines further up where it's called.

This hook allows to intercept `@load` directives and return the strings to be parsed.  Maybe you could leverage this to return whatever is stored in the ZIP instead?

If that doesn't work, it might be better to improve the hooks. There have been ideas/prototypes to do `@load` from an sqlite database as well and conceptually that's not so different from a zip file.

mamaorha: i will add a commit on top the one i commited earlier trying to switch to plugin approach, thanks for the redirect

mamaorha: @awelzel i ammended the changes to my prev commit, kindly lmk if that was your intention

awelzel: hey - yeah, that looks like a good start. My hope was that implementing `HookLoadFileExtended()` alone would suffice - i.e. no changes in `src/scan.l` or the utils.cc such that this code could in theory be completely external.

Are you running into snags keeping the changes completely outside of Zeek?

Does the ZIP provide additional/separate content, or must it act more like an "overlay", so that you can replace base scripts, too?

E.g. you could make `@load content.zip` work instead of adding `--load-zip`, but maybe that's not what you're looking for.

mamaorha: basically we are loading everything (Base included) from in-memory zip to make sure we dont get tampered, we run with high privilege and we cant risk somebody exploiting the scripts to push malicious code.

this is the bare-minimum i could come up with - i wonder if its something you are willing to merge or we should keep it as fork / if you have other approach we can use

awelzel: >  i wonder if its something you are willing to merge or we should keep it as fork / if you have other approach we can use

I chatted with @ckreibich and @timwoj a bit. It looks like having some sort of VFS abstraction accessible via `zeek::vfs` would be nice. You essentially have the beginning of an API sketched in `class ZipScriptProvider` and probably found most places that need some looking. Historically we've just assumed to directly access files on disk.

```
# src/vfs/VFS.h
class VFSFile {
  std::string identifier;  // for "already-loaded detection / same file" ??
  std::optional<std::string> content; // ?
  virtual std::string_view Content(); // ?
  ... type? 
};

class VFS {
   virtual HasFile();
   virtual VFSFile ReadFile(const path& path);
   virtual HasDir(const path& path);
   ...
};

// Global or some instance() helper if better.
std::unique_ptr<VFS> vfs;
```

Then, `can_read()` could become, etc.
```
static bool can_read(const string& path) {
   return zeek::vfs->HasFile(path);
}
```


 If the `vfs` instance is controllable by plugins, this allows to have things like a `CompositeVFS` where a file is first looked up from one or more  `ZipVFS` instances and fallback to a `FilesystemVFS`.

This needs a bit of design. How / when to allow changing the `zeek::vfs` instance, how to handle `ZEEKPATH` and track the current path so relative loads work out. How does tracking of already loaded files work cleanly? Also not sure if plugins fit well: `dlopen()` takes an actual path on disk, so packing `.so` files into a ZIP might not easily work. But this can be done incremental if you don't plan to have `.so` files in ZIP archives.

A good fist step would be creating a issue or a discussion with the use case and some requirements and sketch of some initial ideas.

Does that seem reasonable to you @mamaorha ? Would you be up for starting a discussion and close this PR? I gather you're comfortable keeping the functionality out of tree for the time being and then you can reference the issue/discussion for the future, too.

mamaorha: i have created the following issue: https://github.com/zeek/zeek/issues/5398 and i will try to adjust this pr (diff commit) to move to a VFS approach, we need this core functionality and we rather avoid patches as much as possible (we will for now but we rather avoid diverging from zeek sources again)

awelzel: Okay :+1:

If possible, structure the changes such that there's one commit introducing the API/interfaces, one that changes the existing code paths to use the new abstraction, etc. This makes piece-wise review easier.

> we need this core functionality and we rather avoid patches as much as possible 

Happy you're up for exploring! Hope you don't mind a bit of iteration and some discussions with others from the team.

ci/build.sh
```diff
@@ -6,6 +6,24 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
 set -e
 set -x
 
+# Ensure libzip is available (required for ZipVFS).
```
mamaorha: this wont hurt but its redundant, i only put it now as i couldnt affect the zeek images that need to be rebuilt with this package

src/scan.l
mamaorha: ah already pushed it as 1 commit (its fairly small tho, most of the change is adding the ziplib to the ci), i will split it once i see its green

ckreibich:
Maor we first need to agree on how/whether to introduce a file system abstraction before we can productively move forward with this PR. You're welcome to do so for your own exploration/learning, but I'm going to move your PR into draft state for the time being — this basically just means we won't be looking closely while you're churning through code iterations.

To be clear, we're all in agreement that this is a neat feature and Zeek should enable you to build it cleanly and easily with a plugin, without patching Zeek. So thanks for creating #5398, let's iterate there next.

PR #5401, opened 2026-04-24
Fixes for handling files from Spicy file analyzers

### LINKED ISSUE #5397: Files from spicy file analyzer get `<error>` inserted for source
In `data_in` in src/spicy/runtime_support.cc, we check to see if `cookie->protocol` is set and use the tag from that protocol as the source for the file. If it's not set, we pass the default result from `Tag()` to use for the source. This results in `<error>` being inserted into the filename. This is fine on Linux, but it's invalid on Windows because filenames shouldn't have `<` or `>` in them (or any of a set of other invalid characters). This results in `fopen()` later throwing an error when it tries to extract the file.

Is `cookie->protocol` the right thing to check here? We do use that value for other places for files, but we also use `cookie->file` in various places. I'm looking at this in light of the ZIP analyzer from https://github.com/zeek/zeek/pull/5388. Would it not be better to pull the tag from the file analyzer (in `cookie->file`) and use that for the source, which would result in `ZIP` being used for the source string?

cc @rsmmr, who probably has the most context here.

EDIT: Ah, there's a further complication here. While attempting to try the above, I realized that `file_analysis::Manager::GetFile` calls `analyzer_mgr->GetComponentName()` on the tag. That won't work for file analyzer tags and will still return `<error>`. It needs to call `file_mgr->GetComponentName()` for file analyzer tags, but that might break something else that was expecting a protocol analyzer. Ugh.

timwoj:
The changes made in this PR were suggested by Claude, with some additional massaging of the change and updates to the btests by me.

This came up after the ZIP analyzer landed and the btests broke on Windows. The extracted files were only looking at network protocol tags when being named, and so they were getting `<error>` inserted as their source. See the linked issue for more details.

The primary change in this PR fixes the problem by instead of passing the tag down to the file manager, we call the version that takes the source name directly. This fixes the lookup problem and results in the source being set in the filename to match what file analyzer the files came from.

Two secondary changes fix the same pattern in other places in spicy/runtime-support.cc where we were only looking up the protocol analyzer from the cookie instead of also checking the file analyzer. In these cases we don't actually need to look at the cookie at all because we already have a file ID and lookup the data from the file instead.

Fixes #5397


awelzel (approved):
Hand-wavy approval as it does remove the "error" entries from the issue here: https://github.com/zeek/spicy-zip/issues/11

(thought there's more follow-up needed as we're using compressed size instead of uncompressed size for inner files)

timwoj:
@awelzel Given we're pulling in https://github.com/zeek/zeek/pull/5401 into 8.2, I'm going to pull this one in too. What do you think?

awelzel:
> @awelzel Given we're pulling in https://github.com/zeek/zeek/pull/5401 into 8.2, I'm going to pull this one in too. What do you think?

Yeah :+1: I'm not overly worried about either of these in 8.2 as it seems they triggered bugs mainly via spicy-zip, but probably good to have.

PR #5403, opened 2026-04-26
Fix ZIP file extraction and btests on Windows

mamaorha:
When a file analyzer (like ZIP) creates sub-files via file_begin(), the file was initially created with source '<error>' because the correct source name was only set after the file_new event fired. On Linux this worked since <> are valid in filenames, but on Windows they caused EINVAL when the extract analyzer tried to create files with '<error>' in the path.

Fix by using the string-based DataIn overload in file_begin() when running in a file analyzer context, passing the correct source name (e.g. 'ZIP') so the file is created with the right source before file_new fires.

Also fix the ZIP btests to:
- Sort extracted.log after anonymization (not before) so sort order is stable regardless of the source string in filenames
- Use 'cat >file && rm' instead of 'mv' for in-place file replacement, avoiding 'Device or resource busy' errors on Windows when .stdout is held open by btest


timwoj:
Fixed in a different way by https://github.com/zeek/zeek/pull/5401. I'm going to close this one. If the other part is still an issue once that's merged feel free to reopen.

timwoj:
I lifted the bits about `.stdout` to my PR.

PR #5415, opened 2026-04-29

bbannier:


bbannier:
@timwoj setting this as Ready For Review so it is at least on your radar. For Zeek releases we still prefer dependency releases instead of untagged commits, but if we would merge it now we'd at least make sure that we do not miss the latest backported fixes in Spicy. Alternatively we could keep this PR open for testing and a reminder that it should not exist one we put out the next zeek-8.0.x (I don't see a _Zeek 8.0_ board where this could also end up).

timwoj (approved):

PR #5419, opened 2026-04-29
Remove static from sqlite3 prepared statements

timwoj:
Thanks @awelzel for discovering this one.

Because of the `static` here, all sqlite databases get the same set of prepared statements pointing at the first database that was opened. This results in operations running against a single database file no matter how many you open. Removing the static ensures all backends get their own set of statements.

This PR also switches over to persistent mode for the prepared statements, since we keep them around for the entire run of Zeek. The standard mode is intended for short-lived statements. See https://www.sqlite.org/c3ref/c_prepare_dont_log.html#sqlitepreparepersistent.


awelzel (changes requested):
Ah, thanks for finding this one and ouch..

I'll be pedantic: Grepping through the existing tests indicates we never open two SQLite backends in a single Zeek instance. Seems we should have this, so using "Request changes".

Two `open_backend()`, two `put()` calls or so, either with different or the same key/value pairs, then using `sqlite` to dump the raw contents to a baseline. Something like that should show the bug? Or, inserting in `db1` and looking up via `db2`.

awelzel (approved):
Thanks for the follow-up!

PR #5421, opened 2026-04-30
Fix assignment typo in `zeek::threading::Value`'s move constructor

ckreibich:
This is a minor Claude-provided finding reported by Martin Arlitt. This originally came via #3893, a perf improvement.


awelzel (approved):
Thanks, not sure that's a typo.

PR #5423, opened 2026-04-30
spicy/runtime-support: Fix file lifecycle

bbannier:
Feeding anything, including empty bytes, maybe cause an existing file to
be removed due to lifecycle operations being triggered via DataIn().
Protect by first looking up the file and skipping the DataIn() calls if
it already exists.

This was tickled when an outer ZIP archive is created twice with the
same fuid and the inner file outlives the outer archive.


src/spicy/file-analyzer.cc
```diff
@@ -30,9 +30,10 @@ static auto create_file_state(FileAnalyzer* analyzer) {
             depth = f->depth + 1;
     }
 
-    cookie::FileAnalyzer cookie{.analyzer = analyzer,
-                                .depth = depth,
-                                .fstate = cookie::FileStateStack(analyzer->GetFile()->GetID())};
+    auto id = hilti::rt::fmt("%s:%s", analyzer->GetID(), analyzer->GetFile()->GetID());
```
awelzel, line 33: I think I preferred the tag here to keep the produced `_analyzer_id` constant even if it's a new analyzer instance, but also not sure there's a right or wrong here.

src/spicy/protocol-analyzer.cc
```diff
@@ -26,10 +26,10 @@ static auto create_endpoint(bool is_orig, analyzer::Analyzer* analyzer, ::spicy:
 
     cookie::ProtocolAnalyzer cookie{.analyzer = analyzer,
                                     .is_orig = is_orig,
-                                    .fstate_orig =
-                                        cookie::FileStateStack(hilti::rt::fmt("%" PRIx64 ".orig", id_counter)),
-                                    .fstate_resp =
-                                        cookie::FileStateStack(hilti::rt::fmt("%" PRIx64 ".resp", id_counter))};
+                                    .fstate_orig = cookie::FileStateStack(
+                                        hilti::rt::fmt("%s:%" PRIx64 ".orig", analyzer->GetID(), id_counter)),
```
awelzel: That seems superflous given `id_counter` is `static` ?

src/spicy/file-analyzer.cc
bbannier: Picked this since it seemed like simplest to implement. That said, there's a ton of subtlety going on here 😬.

awelzel: Resolving this comment. Seems fine and as in our earlier conversation we should probably write down how File and FileAnalyzer's interact and maybe then things look weird conceptually and we can change 👍

src/spicy/cookie.h
```diff
@@ -47,7 +47,7 @@ class FileStateStack {
     /**
      * Constructor.
      *
-     * @param analyzer_id unique ID string representing parent connection/file analyzer
+     * @param uid unique ID string representing parent connection/file uid
```
awelzel:
```suggestion
     * @param analyzer_id unique ID string representing parent protocol/file analyzer
```

With the `GetID()` calls, I think we should talk about analyzer again here.

src/spicy/runtime-support.cc
```diff
@@ -1066,7 +1066,16 @@ hilti::rt::String rt::file_begin(const hilti::rt::Optional<hilti::rt::String>& m
     _data_in("", 0, {}, {});
 
     auto file = file_mgr->LookupFile(fstate->fid);
-    assert(file); // passing in empty data ensures that this is now available
+    if ( ! file ) {
+        // Only feed an empty chunk into file analysis to force creating
+        // the file state inside Zeek when the file does not yet exist.
+        //
+        // If the file exists, calling DataIn() for a file can cause side
```
awelzel:
```suggestion
        // If the file already exists, calling DataIn() for it can cause side
```

awelzel (approved):

awelzel:
FWIW, this test tickles the original assert/crash. Two separate parsers opening a new file with the same fuid. Might be worth adding?

```
$ cat file-analysis-file-begin.zeek
# @TEST-REQUIRES: have-spicy
#
# @TEST-EXEC: spicyz -d -o test.hlto ssh.spicy ./ssh-cond.evt
# @TEST-EXEC: zeek -r ${TRACES}/ssh/single-conn.pcap test.hlto %INPUT Spicy::enable_print=T >out
# @TEST-EXEC: btest-diff out

module SSH;

event zeek_init()
        {
        print "zeek_init";
        Analyzer::register_for_port(Analyzer::ANALYZER_SPICY_SSH, 22/tcp);
        Analyzer::register_for_port(Analyzer::ANALYZER_SPICY_SSHX, 22/tcp);
        }

# @TEST-START-FILE ssh.spicy
module SSH;

import zeek;

public type Banner = unit {
    on %init {
      print "fun!";
      self.file_id = zeek::file_begin("text/plain", "F1234567");
      zeek::file_set_size(0, self.file_id);
      print "no fun";
    }

    var file_id: string;

};
# @TEST-END-FILE

# @TEST-START-FILE ssh-cond.evt
protocol analyzer spicy::SSH over TCP:
    parse with SSH::Banner,
    replaces SSH;

protocol analyzer spicy::SSHX over TCP:
    parse with SSH::Banner;
```

```
$ btest -t -v -s build_dir=fuzz-build -d file-analysis-file-begin.zeek 2>&1 | ../../scripts/diff-remove-abspath 
spicy.file-analysis-file-begin ...
  > have-spicy
  > spicyz -d -o test.hlto ssh.spicy ./ssh-cond.evt
  > zeek -r ${TRACES}<...>/single-conn.pcap test.hlto %INPUT Spicy::enable_print=T >out
... spicy.file-analysis-file-begin failed
  % 'zeek -r ${TRACES}<...>/single-conn.pcap test.hlto <...>/file-analysis-file-begin.zeek Spicy::enable_print=T >out' failed unexpectedly (exit code 134)
  % cat .stderr
  zeek: <...>/runtime-support.cc:1069: hilti::rt::String zeek::spicy::rt::file_begin(const hilti::rt::Optional<hilti::rt::String> &, const hilti::rt::Optional<hilti::rt::String> &): Assertion `file' failed.
  Aborted

1 of 1 test failed
```

Issue #5422, opened 2026-04-30
rdp-x509 test & parsing broken


0xxon:
The rdp x509 test that was introduced in 9441dc68ec51ef55fac12c5640c1f9abddf65735 was broken in 2e0edd741608ba73c5f0e6aca9b2bc4a2277ad17 (in 2018), when the test baseline was patched out because it did no longer work in OpenSSL 1.1.

This should not have happened - instead we should have fixed the bug that was correctly exposed by the test failure.

PR #5426, opened 2026-04-30
Make Zeek compile against OpenSSL 4.0

### LINKED ISSUE #5363: Try OpenSSL 4.0
We should check that Zeek compiles against OpenSSL 4.0 :)

0xxon:
The bulk of the changes are function signature changes that move to more const-ness. This is slightly annoying as we use these in our function signatures too.

We also move away from the deprecated asn1 time comparison function.

I used typedefs to keep the changes contained. If anyone has a better idea, please let me know.

This commit is based on top of the commit for #5422 and has to be rebased when/if that commit is accepted.

This commit was partially authored by Claude Sonnet 4.6. I reviewed and fully understand all changes that I did not author myself.

### DIFF
--- a/src/file_analysis/analyzer/x509/OCSP.cc
+++ b/src/file_analysis/analyzer/x509/OCSP.cc
@@ -499,7 +499,7 @@ void OCSP::ParseResponse(OCSP_RESPONSE* resp) {
 
         num_ext = OCSP_SINGLERESP_get_ext_count(single_resp);
         for ( int k = 0; k < num_ext; ++k ) {
-            X509_EXTENSION* ex = OCSP_SINGLERESP_get_ext(single_resp, k);
+            auto* ex = OCSP_SINGLERESP_get_ext(single_resp, k);
             if ( ! ex )
                 continue;
 
@@ -537,7 +537,7 @@ void OCSP::ParseResponse(OCSP_RESPONSE* resp) {
     // ok, now that we are done with the actual certificate - let's parse extensions :)
     num_ext = OCSP_BASICRESP_get_ext_count(basic_resp);
     for ( int k = 0; k < num_ext; ++k ) {
-        X509_EXTENSION* ex = OCSP_BASICRESP_get_ext(basic_resp, k);
+        auto* ex = OCSP_BASICRESP_get_ext(basic_resp, k);
         if ( ! ex )
             continue;
 
@@ -550,7 +550,7 @@ clean_up:
     BIO_free(bio);
 }
 
-void OCSP::ParseExtensionsSpecific(X509_EXTENSION* ex, bool global, ASN1_OBJECT* ext_asn, const char* oid) {
+void OCSP::ParseExtensionsSpecific(openssl_x509_ext_t* ex, bool global, openssl_asn1_obj_t* ext_asn, const char* oid) {
     // In OpenSSL 1.0.2+, we can get the extension by using NID_ct_cert_scts.
     // In OpenSSL <= 1.0.1, this is not yet defined yet, so we have to manually
     // look it up by performing a string comparison on the oid.
--- a/src/file_analysis/analyzer/x509/X509.cc
+++ b/src/file_analysis/analyzer/x509/X509.cc
@@ -84,7 +84,7 @@ bool X509::EndOfFile() {
 
     int num_ext = X509_get_ext_count(ssl_cert);
     for ( int k = 0; k < num_ext; ++k ) {
-        X509_EXTENSION* ex = X509_get_ext(ssl_cert, k);
+        auto* ex = X509_get_ext(ssl_cert, k);
         if ( ! ex )
             continue;
 
@@ -124,7 +124,7 @@ RecordValPtr X509::ParseCertificate(X509Val* cert_val, file_analysis::File* f) {
     pX509Cert->Assign(2, make_intrusive<StringVal>(len, buf));
     BIO_reset(bio);
 
-    X509_NAME* subject_name = X509_get_subject_name(ssl_cert);
+    auto* subject_name = X509_get_subject_name(ssl_cert);
     // extract the most specific (last) common name from the subject
     int namepos = -1;
     for ( ;; ) {
@@ -273,7 +273,7 @@ void X509::FreeRootStore() {
         X509_STORE_free(e.second);
 }
 
-void X509::ParseBasicConstraints(X509_EXTENSION* ex) {
+void X509::ParseBasicConstraints(openssl_x509_ext_t* ex) {
     assert(OBJ_obj2nid(X509_EXTENSION_get_object(ex)) == NID_basic_constraints);
 
     BASIC_CONSTRAINTS* constr = reinterpret_cast<BASIC_CONSTRAINTS*>(X509V3_EXT_d2i(ex));
@@ -296,7 +296,7 @@ void X509::ParseBasicConstraints(X509_EXTENSION* ex) {
         reporter->Weird(GetFile(), "x509_invalid_basic_constraint");
 }
 
-void X509::ParseExtensionsSpecific(X509_EXTENSION* ex, bool global, ASN1_OBJECT* ext_asn, const char* oid) {
+void X509::ParseExtensionsSpecific(openssl_x509_ext_t* ex, bool global, openssl_asn1_obj_t* ext_asn, const char* oid) {
     // look if we have a specialized handler for this event...
     if ( OBJ_obj2nid(ext_asn) == NID_basic_constraints )
         ParseBasicConstraints(ex);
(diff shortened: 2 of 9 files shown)


src/file_analysis/analyzer/x509/X509Common.h
```diff
@@ -25,6 +26,19 @@ class File;
 
 namespace detail {
 
+// X509_get_ext(), X509_EXTENSION_get_object() and related functions return const
+// pointers in OpenSSL 4.0+, but non-const in earlier versions. This is awkward, as
+// we have the signatures in some functions. We use type aliases in these cases now,
+// and adjust them to be the same as the OpenSSL API. Not pretty, but I don't have a
+// better idea.
+#if OPENSSL_VERSION_NUMBER >= 0x40000000L
```
bbannier: Even though these symbols are in a `detail` namespace directly using `OPENSSL_VERSION_NUMBER` is still potentially dangerous since this is in a header. It forces anybody including this header directly or indirectly to have consistent versions between the one that compiled Zeek (baked into the symbols from `OSCP.cc` or `X509.cc`) and the one using these headers in an installation when compiling a plugin. If that ever becomes inconsistent this could lead to best case noisy linker failures. The underlying issue is that we do not control the value of `OPENSSL_VERSION_NUMBER` even though we ship binary code depending on it.

One fix would be to instead define our own `ZEEK_OPENSSL_VERSION_NUMBER` from the value of `OPENSSL_VERSION_NUMBER` detected at compile time, ship it in some defs header, and use it here.

0xxon: This is true, and I honestly did not consider this, and am happy to change this as suggested.

However, if someone compiles a plugin, they have to use the same OpenSSL version as we are using in Zeek, otherwise kind of everything breaks. There cannot be several different versions of OpenSSL that are dynamically or statically linked into the same process space.

bbannier: I think a plugin wouldn't need to actually link a library from OpenSSL for this to become an issue, already including an incompatible header like `openssl/opensslv.h` could be enough to possibly run into e.g., ODR violations (which trigger UB).

src/file_analysis/analyzer/x509/functions.bif
```diff
@@ -813,8 +817,8 @@ zeek::StringValPtr x509_entity_hash(zeek::file_analysis::detail::X509Val *cert_h
 		return nullptr;
 		}
 
-	X509_NAME *subject_name = X509_get_subject_name(cert_x509);
-	X509_NAME *issuer_name = X509_get_issuer_name(cert_x509);
+	auto subject_name = X509_get_subject_name(cert_x509);
```
awelzel: Mostly nit, but it's nice to keep the `*` out of auto. I'll just mark the other places without comments.


```suggestion
	auto* subject_name = X509_get_subject_name(cert_x509);
```

src/file_analysis/analyzer/x509/functions.bif
```diff
@@ -813,8 +817,8 @@ zeek::StringValPtr x509_entity_hash(zeek::file_analysis::detail::X509Val *cert_h
...
+	auto issuer_name = X509_get_issuer_name(cert_x509);
```
awelzel:
```suggestion
	auto* issuer_name = X509_get_issuer_name(cert_x509);
```

src/file_analysis/analyzer/x509/OCSP.cc
```diff
@@ -499,7 +499,7 @@ void OCSP::ParseResponse(OCSP_RESPONSE* resp) {
 
         num_ext = OCSP_SINGLERESP_get_ext_count(single_resp);
         for ( int k = 0; k < num_ext; ++k ) {
-            X509_EXTENSION* ex = OCSP_SINGLERESP_get_ext(single_resp, k);
+            auto ex = OCSP_SINGLERESP_get_ext(single_resp, k);
```
awelzel:
```suggestion
            auto* ex = OCSP_SINGLERESP_get_ext(single_resp, k);
```

src/file_analysis/analyzer/x509/OCSP.cc
```diff
@@ -537,7 +537,7 @@ void OCSP::ParseResponse(OCSP_RESPONSE* resp) {
     // ok, now that we are done with the actual certificate - let's parse extensions :)
     num_ext = OCSP_BASICRESP_get_ext_count(basic_resp);
     for ( int k = 0; k < num_ext; ++k ) {
-        X509_EXTENSION* ex = OCSP_BASICRESP_get_ext(basic_resp, k);
+        auto ex = OCSP_BASICRESP_get_ext(basic_resp, k);
```
awelzel:
```suggestion
        auto* ex = OCSP_BASICRESP_get_ext(basic_resp, k);
```

src/file_analysis/analyzer/x509/X509.cc
```diff
@@ -84,7 +84,7 @@ bool X509::EndOfFile() {
 
     int num_ext = X509_get_ext_count(ssl_cert);
     for ( int k = 0; k < num_ext; ++k ) {
-        X509_EXTENSION* ex = X509_get_ext(ssl_cert, k);
+        auto ex = X509_get_ext(ssl_cert, k);
```
awelzel:
```suggestion
        auto* ex = X509_get_ext(ssl_cert, k);
```

src/file_analysis/analyzer/x509/X509.cc
```diff
@@ -124,7 +124,7 @@ RecordValPtr X509::ParseCertificate(X509Val* cert_val, file_analysis::File* f) {
     pX509Cert->Assign(2, make_intrusive<StringVal>(len, buf));
     BIO_reset(bio);
 
-    X509_NAME* subject_name = X509_get_subject_name(ssl_cert);
+    auto subject_name = X509_get_subject_name(ssl_cert);
```
awelzel:
```suggestion
    auto* subject_name = X509_get_subject_name(ssl_cert);
```

src/file_analysis/analyzer/x509/X509Common.cc
```diff
@@ -166,22 +166,23 @@ double X509Common::GetTimeFromAsn1(const ASN1_TIME* atime, file_analysis::File*
...
 
-void X509Common::ParseSignedCertificateTimestamps(X509_EXTENSION* ext) {
+void X509Common::ParseSignedCertificateTimestamps(openssl_x509_ext_t* ext) {
     // Ok, signed certificate timestamps are a bit of an odd case out; we don't
     // want to use the (basically nonexistent) OpenSSL functionality to parse them.
     // Instead we have our own, self-written binpac parser to parse just them,
     // which we will initialize here and tear down immediately again.
 
-    ASN1_OCTET_STRING* ext_val = X509_EXTENSION_get_data(ext);
+    auto ext_val = X509_EXTENSION_get_data(ext);
```
awelzel:
```suggestion
    auto* ext_val = X509_EXTENSION_get_data(ext);
```

src/file_analysis/analyzer/x509/X509Common.cc
```diff
@@ -207,11 +209,11 @@ void X509Common::ParseSignedCertificateTimestamps(X509_EXTENSION* ext) {
     delete conn;
 }
 
-void X509Common::ParseExtension(X509_EXTENSION* ex, const EventHandlerPtr& h, bool global) {
+void X509Common::ParseExtension(openssl_x509_ext_t* ex, const EventHandlerPtr& h, bool global) {
     char name[256];
     char oid[256];
 
-    ASN1_OBJECT* ext_asn = X509_EXTENSION_get_object(ex);
+    auto ext_asn = X509_EXTENSION_get_object(ex);
```
awelzel:
```suggestion
    auto* ext_asn = X509_EXTENSION_get_object(ex);
```

awelzel (approved):
Only had some `auto*` nits to keep the `*` explicit. Grepping for `auto*`, it's a common pattern.

Otherwise seems fine - I don't think the `const` difference in that header specifically will cause breakage when running/linking with the wrong OpenSSL library, but we probably should keep the OPENSSL_MAJOR version somewhere and refuse to compile plugins or start Zeek when it disagrees to what's available. Maybe we already do that.

> From 3.0.0 and above, the OpenSSL versioning scheme uses the format: MAJOR.MINOR.PATCH
>
> With this format, API/ABI compatibility will be guaranteed for the same MAJOR version number.

https://openssl-library.org/policies/releasestrat/

0xxon:
I adopted a mix of your suggestions. Now cmake keeps track of the major version, and the header file has a static assert to check that it matches the version that Zeek was configured with.

@bbannier - does that look ok to you?

src/file_analysis/analyzer/x509/X509Common.h
bbannier: With the new `static_assert` this should now be safe, thanks!

Small nits:

- in the preproc `#if` I'd branch on `ZEEK_OPENSSL_VERSION_MAJOR` instead of `OPENSSL_VERSION_MAJOR` since semantically this is about what Zeek was built with, not what the user pull in (they are still the same value though 🤷)
- I'd pull `static_assert` just before the preproc branch since they are intimately connected

bbannier (approved):
LGTM wrt the include file issue, but I have not validated the OpenSSL 4.0-related changes.

I briefly tried to figure out why we install `X509Common.h` in the first place since it doesn't seem needed in the public interface, but couldn't find a place doing it explicitly. I guess that is some glob-related automatism somewhere, but that would be an unrelated cleanup.

src/file_analysis/analyzer/x509/X509Common.h
0xxon: Thank you - I also folded both of these in.

PR #5427, opened 2026-05-01
BiFs for profiling ZAM code in live execution

vpax:
This PR reworks ZAM's profiling in a number of ways:

- The number of ZAM body calls, and counts per ZAM instruction of how often they execute, are now always tracked.
- The new `ZAMProf::measure_module()` BiF allows turning on more expensive CPU+memory profiling on a module-by-module basis.
- The new `ZAMProf::get_module_profile()` BiF retrieves a running profile of a given module's execution, including (optional) CPU+memory profiling.
- The new `ZAMProf::meas_overhead()` returns an estimate of the cost of measuring (estimating) CPU and memory usage.

In addition, the PR (1) removes an `assert` in printing out ZAM instructions that was actually incorrect, (2) updates the BTest used to ensure that ZAM's knowledge of BiFs stays up to date (and updates that knowledge).

The existing `-O profile-ZAM` output remains the same.


src/script_opt/ZAM/prof.bif
```diff
@@ -0,0 +1,71 @@
...
+##
+## Returns: The estimate of the total overhead, as a non-negative interval.
+##
+## .. note::
+##
+##    This value is fixed per Zeek run. Repeated calls to it from within
+##    that run will return the same value.
+##
+## .. zeek:see:: ZAMProf::mem_meas_overhead ZAMProf::get_module_CPU_mem
+function ZAMProf::meas_overhead%(%): interval
```
awelzel: I think this one is fine be longer rather than short and include the estimated part in the name. The `zeek:see:` line above looks outdated or like a prior attempt?

```suggestion
function ZAMProf::estimated_profiling_overhead%(%): interval
```

src/script_opt/ZAM/prof.bif
```diff
@@ -0,0 +1,71 @@
...
+##    activation follows the most recent call for any of those modules.
+##
+## .. note::
+##
+##    A return value of 0 means the module name doesn't correspond to
+##    any compiled bodies. This will always be the case if Zeek is not
+##    running with some form of -O ZAM enabled.
+##
+## .. zeek:see:: ZAMProf::get_module_profile
+function ZAMProf::measure_module%(mod: string, active: bool%): count
```
awelzel: Definitely naming, but got hung up. The comment starts with:

> Activates/deactivates CPU+memory profiling for all ZAM-compiled bodies
> that include the given module.

and internally this `calls ->SetProfilingCalls()`. The "measure" terminology just seems new/surprising while looking through. Could we go with `profile_module()` or `set_profile_module()` given the active toggle instead? Or did "measure" feel better to you?

src/script_opt/ZAM/prof.bif
```diff
@@ -0,0 +1,71 @@
+##! Functions to enable script-level control & querying of ZAM profiling.
+
+module ZAMProf;
```
awelzel: Maybe a nested `Prof` module?

```suggestion
module ZAM::Prof;
```

src/script_opt/ZAM/CMakeLists.txt
```diff
@@ -18,3 +18,5 @@ zeek_add_subdir_library(
     ZInst.cc
     ZInstAux.cc
     ZOp.cc)
+
+bif_target(prof.bif)
```
awelzel: You can add a `BIFS` section above (src/cluster/CMakeLists.txt does that).

```
    ZOp.cc
    BIFS
    prof.bif)
```

Also suggesting `zam-prof.bif` just because that ends up as a top-level .bif.zeek file in `scripts/base/bif`, so being a bit more specific would help. And a `-` because it seems that's the convention (and I didn't follow with the publish_on_change bif :-( )

src/script_opt/ZAM/zam-prof.bif
```diff
@@ -0,0 +1,71 @@
...
+##    The return value will often indicate fewer bodies than present in
+##    the module due to inlining. On the flip side, it can indicate more
+##    bodies than present in the module due to ZAM generates bodies for
+##    both standalone and "coalesced" versions of event handlers.
+##
+## .. note::
+##
+##    Due to event handler coalescence (enabled by default), some ZAM-compiled
+##    bodies will correspond to multiple modules. For these bodies, sampling
+##    activation follows the most recent call for any of those modules.
```
awelzel, line 33: From reading the code and this comment: If one enables profiling for two modules and then disables one, the result is a potentially partly enabled module?

Seems quirky and easy to get confused, but in practice one would probably not really do this kind of thing?

src/script_opt/ScriptOpt.cc
```diff
@@ -725,6 +732,61 @@ void analyze_scripts(bool no_unused_warnings) {
...
+        is_profiling = is_profiling || b->IsProfilingCalls();
+
+        ncalls += b->NumBodyCalls();
+
+        auto tot_inst = b->NumBodyInsts();
+        if ( tot_inst > 0 ) {
+            auto mod_inst = b->NumModuleInsts(mod);
+            double ratio = double(mod_inst) / double(tot_inst);
+
+            ninst += int(ratio * tot_inst + 0.5);
```
awelzel: Clang-tidy suggests using `lround()` from cmath.

awelzel (commented):
Hey Vern - I have mainly "measure" vs "profiling" naming question/feedback and a module overlap question. Otherwise the details of ZBody.h is sort of at your governance, I was mostly wondering if at some point a profiling hook could make sense, but then again `ZBody::Exec()` isn't all that big yet, but it might make it easier to follow.

src/script_opt/ZAM/prof.bif
vpax: Cool, that's better, now in the latest push. (Hopefully I've also fixed the `zeek:see:`s.)

src/script_opt/ZAM/prof.bif
vpax: I've now regularized this

src/script_opt/ZAM/prof.bif
vpax: Sounds good, done

src/script_opt/ZAM/CMakeLists.txt
vpax: Good thought, done

src/script_opt/ZAM/zam-prof.bif
vpax: Yeah I do think it's not likely. It's also hard to avoid. The issue is that event handler coalescence - which can be a nice win for popular event handlers - fundamentally intermingles code from different modules. Inlining can do the same thing. The logic for deciding to do measurement is done on a per-ZAM-body basis; anything more selective would likely have considerable overhead, which I'm trying to avoid.

vpax:
> I was mostly wondering if at some point a profiling hook could make sense

I remain very stingy with how much computation to allow for profiling, which leads me to want to avoid hooks if possible. (Indeed, before adding the always-on instruction counting, I timed what that would cost, but it turns out to be negligible ...)

WIth luck, the latest push addresses all the pending issues.

awelzel (approved):
I'll fix up the `prof.bif` vs `zeek-prof.bif` baseline difference during the merge.

Thanks for the follow-ups.

PR #5429, opened 2026-05-04
Split cluster logic for ZeroMQ into separate script

J-Gras:
I tried running the ZeroMQ proxy in the supervisor process and run into a couple of ugly warnings like the following:
```
warning in .../zeek/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek, line 549: Expired hello from nodeid zeromq__System.local_86622_NwrZBf1q3FkxySjt2a

warning in .../zeek/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek, line 543: Expired subscription from nodeid zeromq_logger_System.local_86659_NwrZBf1q3FkxySjt2a 
warning in .../zeek/scripts/policy/frameworks/cluster/backend/zeromq/main.zeek, line 543: Expired subscription from nodeid zeromq_proxy_System.local_86656_NwrZBf1q3FkxySjt2a
...
```
They are caused by the cluster logic for ZeroMQ that makes the supervisor try to participate in the cluster like a proper cluster node. However, the supervisor is not part of the cluster, i.e. it doesn't have a node name and is not listed in the cluster layout. As a consequence, subscriptions and hellos expire.

The issue was not surfaced by the existing btest as it didn't run into the timeout. The first commit adapts the btest to trigger the warnings. The following splits out a `cluster.zeek` for the ZeroMQ-related cluster logic. This way the supervisor can run the ZeroMQ proxy without trying to join the cluster by loading `main.zeek` only.


awelzel:
As in DM. Suggesting to split into `options.zeek` and `main.zeek` rather than `main.zeek` and `cluster.zeek`. Happy to do so before the merge.

Thanks for looking at this!

J-Gras:
> Suggesting to split into `options.zeek` and `main.zeek` rather than `main.zeek` and `cluster.zeek`. Happy to do so before the merge.

Just changed the names while fixing some btests. Not sure why CI still fails for Windows and MacOS. I'd like to give it a second try but I am unable to re-run the tests.

scripts/policy/frameworks/cluster/backend/zeromq/options.zeek
```diff
@@ -0,0 +1,454 @@
...
+	## Zeek creates a short-lived subscription for a auto-generated
+	## topic name with this prefix and waits for it to be confirmed
+	## on its XPUB socket. Once this happens, the XPUB socket should've
+	## also received all other active subscriptions of other nodes in a
+	## cluster from the central XPUB/XSUB proxy and therefore can be
+	## deemed ready for publish operations.
+	const internal_topic_prefix = "zeek.zeromq.internal." &redef;
+}
+
+redef Cluster::backend = Cluster::CLUSTER_BACKEND_ZEROMQ;
```
awelzel, line 426: That redef and the handler below should rather live in `main.zeek` such that `options.zeek` is only about `options` rather than changing the backend, I'll fixup during the merge.

scripts/base/utils/addrs.zeek
```diff
@@ -1,73 +1,108 @@
 ##! Functions for parsing and manipulating IP and MAC addresses.
 
-# Regular expressions for matching IP addresses in strings.
+module GLOBAL;
```
awelzel, line 3: Phew, nice find. There's more issues in the `scripts/base/utils`, will look as a follow-up.

awelzel (approved):

scripts/policy/frameworks/cluster/backend/zeromq/options.zeek
J-Gras: I'd prefer to keep this because of https://github.com/zeek/zeek/blob/7ad2f54414f5d9d1d2d4159b03d5fc5d986aef56/src/cluster/backend/zeromq/cluster_backend_zeromq.bif#L11-L22

J-Gras: Oh, and if that doesn't feel right because of the file naming, I'd suggest to go with `main.zeek` and `cluster.zeek` 😜

awelzel: >  Slight changes during merge, keep options.zeek just for options, move supervisor
    logic into test rather than the bootstrap script so that the required steps to
    run the proxy in a supervisor are visible in a single @if section in the test.

I have this in the merge commit. I agree this isn't super great/clean but it works and keeps non-supervisor beahvior okay'sih IMO.

PR #5431, opened 2026-05-05

bbannier:


bbannier:
The same bump will also happen on `release/8.2` since today's Spicy `main` is still at a version targeted for zeek-8.2.

timwoj (approved):

PR #5432, opened 2026-05-05
cluster/backend/zeromq: Move redefs into options.zeek

awelzel:
This change allows to load options.zeek by a Zeek supervisor process and
connect with the ZeroMQ XPUB/XSUB socket without partaking in the Zeek
cluster eventing (Cluster::nodes, Cluster::node_up, etc). The ZeroMQ specific
events remains in main.zeek, which shouldn't be loaded in the supervisor as
it is not part of Cluster::nodes.

This provides a channel for the supervisor process to send and receive
events to supervised nodes. With the Broker backend, was previously possible
with explicit Broker::listen() calls in the supervisor process and Broker::peer()
calls from the supervised nodes.


J-Gras:
The resulting functionality looks great! Just a small naming nit - of course: `redefs.zeek` is not a very telling naming. In particular, as we also have `options.zeek`. My favorite would be `redefs.zeek` --> `backend-init.zeek` and `main.zeek` --> `cluster-init.zeek`. This pattern might apply to other backends as well.

awelzel:
> Just a small naming nit - of course: redefs.zeek is not a very telling naming.

Yeah, I wasn't very inspired and mostly wanted to see if it does work out at all. I'm looking at it now and am thinking those redefs could just move into `options.zeek` the way you had it in your original PR initially, too. Would skip the `-init.zeek` variants as it doesn't seem the separation is actually required right now. Will ponder and rework tomorrow again.

testing/btest/cluster/zeromq/supervisor-runs-zmq-proxy-and-connects.zeek
```diff
@@ -0,0 +1,143 @@
...
+
+# Supervisor gets we_are_all_up() from each of the nodes when they see
+# cluster_started(). The supervisor publishes you_are_all_up() to
+# each of the node topics, then await a second round of we_are_all_up()
+event we_are_all_up(from: string)
+	{
+	add nodes_up[from];
+	print "we_are_all_up", "round", round, "nodes_up", |nodes_up|;
+
+
```
ckreibich: (extra whitespace)

ckreibich (approved):
LGTM, yep!

Is there scope for a `policy/frameworks/supervisor/zeromq.zeek` or some such to package up the endpoint/thread tweaks in reusable form?

awelzel:
> Is there scope for a policy/frameworks/supervisor/zeromq.zeek or some such to package up the endpoint/thread tweaks in reusable form?

I'd bounce that for now until we have a settled/revised supervisor story.

Thanks for the review.

Issue #5090, opened 2025-12-11
Confusions around `is`/`as` operators


vpax:
The `is` operator is defined as returning `T` if the given value can be cast (via the `as` casting operator) to the given type. This makes sense as a way to manipulate `any` types, which at run-time have a concrete type that scripts need to understand/access. However because it's defined in terms of can-be-cast-to instead of actually-is-under-the-hood, we get peculiarities. Here's one:
```
print set(5) is set[count], set(5) is vector of count;
```
prints `T, T` because the `as` operator can be used to cast a `set[X]` to a `vector of X` (and vice versa). Similarly, this script:
```
function foo(v: any)
	{
	switch v {
	case type vector of count: print "vector-of-count"; break;
	case type set[count]: print "set[count]"; break;
	}
	}

event zeek_init()
	{
	foo(vector(1, 3, 5));
	foo(set(2, 4, 6));
	}
```
prints
```
vector-of-count
vector-of-count
```
which is ... surprising.

I'm thinking what we want are two predicate operators, `is` and `converts_to` (or something like that). The former operator is only germane for `any` types, and is about an exact comparison with the concrete type the value has. (This would be a breaking change, though seems likely a low-touch one.) The latter returns `T` if an `as` operator will succeed in _converting_ the value to the given type. We can then extend `as` to support all of reasonable conversions, which today are mostly - but not entirely - available via one-off BiFs.

The basic motivation behind this is that I continually find the BiFs unwieldy, especially since the set is somewhat incomplete (e.g. there's `int_to_double()` and `double_to_interval()` but no `int_to_interval()`).

awelzel:
> The former operator is only germane for any types, and is about an exact comparison with the concrete type the value has.

Make sense to me.

> The basic motivation behind this is that I continually find the BiFs unwieldy, especially since the set is somewhat incomplete (e.g. there's int_to_double() and double_to_interval() but no int_to_interval()).

👍 

So the following might end up working?
```
local ts = 4711.0 as time;
local delta = (current_time() - network_time()) as double;
```

Do we need `converts_to` in Zeek script? Or would that be evaluated statically anyhow?

vpax:
Yes, both of those examples would work, as would `4711 as time`. My thinking is that we can be fairly lenient on supported conversions (providing they're safe - they can generate run-time errors, such as for `(-1) as count`), given the script writer has to go out of their way to specify them.

> Do we need `converts_to` in Zeek script? Or would that be evaluated statically anyhow?

[...]

PR #5433, opened 2026-05-06
Expand `as` operator and add `?as` operator

vpax:
This PR addresses the issues discussed in https://github.com/zeek/zeek/issues/5090 (and closes that issue). It includes the following:

- Factoring out the type conversion logic from the existing BiFs into its own source files (`val-convert.{h,cc}`)
- Extending the `as` operator to support all of the type conversions currently available via BiFs, plus some additional ones that are natural companions
- Adding a `?as` predicate that evaluates to `T` if its lefthand operand can be safely converted to its righthand type
- Supporting both of these for `-O ZAM` and `-O gen-C++`
- Converting all of the standard scripts to use `as` instead of conversion BiFs
- Converting all of the BTest inputs (other than those for conversion BiFs) to use `as`
- Fixing an AST optimizer bug that testing of these changes revealed: constant propagation could lead to unsafe folding of an expression (for example, an invalid type conversion) that generated compile-time errors even though guards would prevent the error condition from arising during actual execution

I used Claude Opus 4.5 to factor out the conversion logic into `val-convert`, to generate of the conversion extensions, and to translate the standard scripts. I have fully vetted the code it generated in this regard. I did not use it for the introducing the `?as` operator, the script optimization support, converting the BTests, nor fixing the optimizer bug.

The PR is large but much of that is mechanical or otherwise simple changes.


src/val-convert.h
```diff
@@ -0,0 +1,50 @@
+// See the file "COPYING" in the main distribution directory for copyright.
+
+#pragma once
+
+#include "zeek/IPAddr.h"
+#include "zeek/Val.h"
+
+namespace zeek {
```
awelzel: Would put these into detail for flexibility changing the API/names/etc, e.g. to go to ZVal-like instead of Val.

```suggestion
namespace zeek::detail {
```

If it's in detail we can just outright change / rework this without needing to go through deprecations.

src/Val.cc
```diff
@@ -3995,7 +4002,8 @@ ValPtr cast_value_to_type(Val* v, Type* t) {
 
         return ret;
     }
-    else if ( IsVector(v->GetType()->Tag()) && t->IsSet() ) {
+
+    if ( IsVector(v->GetType()->Tag()) && t->IsSet() ) {
```
awelzel, line 4006: Mostly-fly-by and pre-existing: Shouldn't this check  `t's` indices being `size == 1` like above?

awelzel: Or maybe we could check if `Assign()` worked out which I think fails if the index doesn't match.

```
if ( ! ret->Assign(std::move(ve), nullptr) )
    return nullptr;
```

src/val-convert.h
```diff
@@ -0,0 +1,50 @@
...
+
+extern ValPtr convert_addr_to_subnet(const IPAddr& addr);
+extern ValPtr convert_addr_to_counts(const IPAddr& addr);
+
+extern ValPtr convert_subnet_to_addr(const IPPrefix& sn);
+extern ValPtr convert_subnet_to_count(const IPPrefix& sn);
+
+extern ValPtr convert_port_to_count(uint32_t port);
+
+extern ValPtr convert_counts_to_addr(const VectorVal* vv, std::string& err);
```
awelzel, line 48: This one seems a bit strange.

I eagerly tried the following, but clearly that's not how one is supposed to use it.
```
$ zeek -e 'print vector(192, 168, 0, 1) as addr'
0:c0:0:a8::1
```

and there's slight overlap with
```
print 192 as addr
```

So this only works with vectors of size 1 or size 4. The explicit/special `counts_to_addr()` appears to clearer for what seems to be a bit of a special conversation. We don't have any callers for those bifs in the base scripts. Do you think the `vector of count as addr` conversion is important enough to support via operators?

awelzel: Oh, and total lucky, can tickle a crash :-)
```
$ zeek -e 'print vector(1,2) ?as addr'
Segmentation fault (core dumped)
```

src/Val.cc
```diff
@@ -4010,7 +4018,157 @@ ValPtr cast_value_to_type(Val* v, Type* t) {
...
+    ValPtr result;
+
+    switch ( s_tag ) {
+        case TYPE_STRING: {
+            auto sv = v->AsStringVal();
+
+            switch ( t_tag ) {
+                case TYPE_DOUBLE: result = convert_string_to_double(sv, err); break;
+                case TYPE_TIME: result = convert_string_to_time(sv, err); break;
+                case TYPE_INTERVAL: result = convert_string_to_interval(sv, err); break;
```
awelzel: Think the string to time/interval ones should be skipped. These aren't existing in BIFS from what I can tell and they just do `convert_string_to_double()` internally and that might easier to let the user do via

```
"42.0" as double as interval
```

Avoids confusion whether the following should work (I do not think that either of them should work).
```
"2026-05-06T16:31:54+02:00" as time
"42.0 msec" as interval
```

There's one change in the PR that already does the double first:

```
				local mhr_first_detected = (MHR_answer[0] as double) as time;
```

src/Val.cc
```diff
@@ -4010,7 +4018,157 @@ ValPtr cast_value_to_type(Val* v, Type* t) {
...
+
+        case TYPE_COUNT: {
+            auto cv = v->AsCount();
+
+            switch ( t_tag ) {
+                case TYPE_INT:
+                    if ( cv > INT64_MAX )
+                        err = "overflow converting count to int";
+                    else
+                        result = make_intrusive<IntVal>(zeek_int_t(cv));
```
awelzel: val_mgr has a Int() helper to reuse IntVal instances, and a non-C-style cast suggestion.

```suggestion
                        result = val_mgr->Int(static_cast<zeek_int_t>(cv));
```

doc/reference/zeekscript/operators.rst
```diff
@@ -329,24 +331,54 @@ or user-defined).
...
-  corresponding Zeek script types.
+- All values can be cast to their declared types (i.e., this is a no-op).
 - A value of declared type :zeek:type:`any` can be cast to its actual
   underlying type.
-- All values can be cast to their declared types (i.e., this is a no-op).
+- :zeek:type:`string` types can be cast to :zeek:type:`double`, :zeek:type:`time`, :zeek:type:`interval`, :zeek:type:`int`, :zeek:type:`count`, :zeek:type:`addr`, :zeek:type:`subnet`, or :zeek:type:`port`.
+- :zeek:type:`int` can be cast to :zeek:type:`double` or :zeek:type:`count`.
+- :zeek:type:`count` can be cast to :zeek:type:`double`, :zeek:type:`int`, or :zeek:type:`addr`.
+- :zeek:type:`port` can be cast to :zeek:type:`count`.
+- :zeek:type:`double` can be cast to :zeek:type:`time`, :zeek:type:`interval`, :zeek:type:`int`, or :zeek:type:`count`.
```
awelzel: The rounding behavior might be surprising? Maybe should leave a comment.
```
$ zeek -e 'print "42.7" as double as count'
43
```

awelzel (commented):
I kind of like this.

I'd step back from the proposed `string to time/interval` conversion and the `vector of count to addr` I find also to niche/quirky to have available, there's some crash that can be tickled and would also suggest to move the new conversion helpers and the new API functions into `zeek::detail::` so we have an easier time iterating after 9.0 is released without feeling bad about breaking a `zeek::` public API.

Hope these make sense, seems fine to land to me on the whole.

src/Val.cc
```diff
@@ -4037,10 +4195,59 @@ static bool can_cast_set_and_vector(const Type* t1, const Type* t2) {
     return false;
 }
 
-bool can_cast_value_to_type(const Val* v, Type* t) {
```
awelzel, line 4040: That's in the `zeek::` namespace, but I guess it's such an internal interpreter detail okay to remove without deprecation.

(reposting a day after - 🤞 GitHub allows me).

src/Val.h
```diff
@@ -1967,22 +1967,37 @@ extern void delete_vals(ValPList* vals);
...
+// usual use is for "any" values (hence the name), for backwards compatibility
+// it also returns true for casting a value to its own type.
+extern bool can_cast_any_to_type(const Val* v, Type* t);
 
 // Returns true if values of type s may support casting to type t. This is
 // purely static check to weed out cases early on that will never succeed.
 // However, even this function returns true, casting may still fail for a
 // specific instance later.
-extern bool can_cast_value_to_type(const Type* s, Type* t);
+extern bool can_cast_type_to_type(const Type* s, Type* t);
```
awelzel, line 2000: I'm tempted to suggest to move all these new helpers into `zeek::detail` as well so there's more freedom to iterate on them after 9.0 is released without going through a deprecation cycle.

(E.g. thinking/straightening whether we'd want to have the API be `TypePtr` or `Type*` or `Type&` and the same for `Val`, but that's not for figuring out in this PR here).

EDIT: Delayed post because GitHub errors prevented from doing so yesterday)

src/val-convert.h
vpax: I've gone ahead with this change.

src/Val.cc
vpax: Makes sense to tighten the original check, rather than letting it get to the `Assign()`.

src/val-convert.h
vpax: Yeah, that is surprising. You've convinced me to remove it (and also to fix that crash, which could happen in other contexts, too).

src/Val.cc
vpax: Likewise, I buy your points and have now removed this.

src/Val.cc
vpax: Done.

doc/reference/zeekscript/operators.rst
vpax: Added to the docs. I also discovered that it's round-to-nearest (which I had guessed, as I know it's the preferred technique, but didn't actually know was the case), and added that.

vpax:
@awelzel I think everything you flagged has now been addressed.

src/Val.cc
vpax: (Seems that way to me, too)

src/Val.h
vpax: I didn't do this one, but certainly wouldn't mind if you think it's the way to go ...

awelzel:
> @awelzel I think everything you flagged has now been addressed.

Thanks! CI isn't super happy, there's a build issue and some failing test for ASAN. Mind taking a look once more?

@ckreibich - mind checking through the `docs` update? It's a bit of a language change, but seems pretty fine.

doc/reference/zeekscript/operators.rst
```diff
@@ -329,24 +331,53 @@ or user-defined).
...
-The function in this example tries to cast a value to a string:
+The function in this example converts a set to a vector:
+
+.. code-block:: zeek
+
+    function example()
+        {
+	local s = set(1, 2, 3);
+	local v = s as vector of count;
+        }
```
evantypanski, line 377: just noting indentation here is off

evantypanski (commented):
just to give some thoughts, I didn't look at the implementation though:

I really like the new `as`! Feels good and looks good. I really don't like those conversion bifs.

`?as` I'm a little bit iffy on, but fine if it gets in. The crux is that if you're checking if a type converts at runtime, a lot of the time you will also want to give a reason why. For example, from the issue:

```
local s = "192.168.0.1";
if ( s ?as addr ) {
  local a = s as addr;
}
```

Wouldn't most cases you use something like that also want to report? Eg

```
local s = "192.168.0.1";
if ( s ?as addr ) {
  local a = s as addr;
} else {
  print "input is not an address!";
}
```

Or if you're getting an environment variable and it doesn't convert, you want to know why, perhaps?

I think it's a shame that we give no justification for why it does not convert. My mind jumped to bifs at first (eg users call a bif that says if it converts, that bif returns a reason why not) but that doesn't seem particularly clean.

Otherwise we could also do a result or error type - where `?as` returns an optional error string if it failed to convert, saying why. Then it's easier to debug and reason about it if it fails for some reason, and users don't need an extra branch:

```
local s = "192.168.0.1";
local err = s ?as addr;
if ( err ) {
  print err;
}

local a = s as addr;
```

If there are cases that `?as` is useful to check many different cases and apply different logic, I get why a simple `bool` would be better. Are there cases like that? eg:

```
local x = "<unknown>";

if ( x ?as port ) {
  # ...
} else if ( x ?as addr ) {
  # ...
} 
# ...
```

Regardless, fine if it gets in as-is since I think both `as` expansion and `?as` are good additions, but I think there's room to expand this and be more in-line with how people would likely use it. Though, I don't know how common that'd be, so it may not be that big of a deal.

EDIT: Oh, and if both are useful, they can certainly coexist

doc/reference/zeekscript/operators.rst
vpax: whoops - fixed in the upcoming push

vpax:
> The crux is that if you're checking if a type converts at runtime, a lot of the time you will also want to give a reason why

I see the utility of that. Zeek doesn't have a notion of optional strings, so `?as` needs to just return a `bool`. But it's reasonable for it to squirrel away the error message, and then a call to a BiF can return it. I don't see a cleaner way than that ... ?

awelzel:
I'm going to merge this after a rebase and squashing one fixup.

For the `?as` error reporting, it seems in a UX/UI you can always `Reporter::error(fmt("not able to convert to address %s", value))`. For per-packet/per-event errors, you'd probably reach for `Reporter::conn_weird()` and pass the original value, rather than the reason? I'd think the original value is anyhow more interesting and logging that more useful than logging a reason (IMO).

evantypanski:
> I don't see a cleaner way than that ... ?

Yeah I don't really right now, I keep hitting cases that having an easy result type would be really useful, but alas haven't gotten to formalizing that opinion.

If we had it I would've been pushed more to get better errors/reasons, but we do not. Just one more data point that I think would make it useful to have good result handling (and more formal optional handling, for that matter)

ckreibich:
I added a NEWS update in #5519. With this and the fact that the PR included docs I think we're done here regarding documentation.

PR #5435, opened 2026-05-06
Remove support for OpenSSL versions < 3.0

0xxon:
All supported distributions have at least OpenSSL 3.0, which also is the oldest supported LTS release of OpenSSL. Removing it allows us to get rid of quite a bit of code that's no longer exercised by any of our ci tests.

This commit was partially authored by Claude Opus 4.6. I reviewed and fully understand all changes that I did not author myself.


awelzel (approved):
A NEWS entry about the minimum requirement, but otherwise, cool!

I wonder a bit if having this in 8.2 would've been better for testers/distributors, but oh well, we don't have these long term plans :-)

awelzel:
Oh, and the tests.ctu-sme-11-vm-win7ad  pe.log appears to have been perturbed a bit, but not sure if this is you or just a side-effect of the zip/file analysis changes that had  been in master when you branched.

0xxon:
Thanks for the review - let me take a peek at the external test baselines. I admittedly didn't check those :)

0xxon:
I rebased this to master, and then checked; there are no external testline changes on my system.

awelzel (approved):

PR #5436, opened 2026-05-06
Update pe.log baseline for CTU-SME external test

timwoj:
This was apparently an errant change from the recent IGMP log work. I have no idea how it even made that change 🤷🏼


ckreibich (approved):
I'm guessing update-changes. :-)

PR #5444, opened 2026-05-10
Warn when expiration interval is a dynamic value (e.g. function argument)Warn when expiration interval is a dynamic value

### LINKED ISSUE #5207: Using a function argument for &create_expire appears to have no effect
I was dabbling with a script where the `&create_expire` time of a table can be provided as an argument to a make / create function.

This doesn't appear to work: Only using a globals or a literals seem to have an effect. The function `make_cache_table_dynamic()` always returns a table value where no expiration is in effect. The `make_cache_table_global()` is also interesting in that the `global_ce` value applies to both tables even though `global_ce` is changed to 30days (meaning no expiration for the rotation.trace) after creating `tbl_global1` . So it seems the 30day create_expire appears also applies to `tbl_global1`. Very quirky.

```
$ zeek -Cr rotation.trace ./dynamic-create-expire.zeek 
tbl_literal, 1
tbl_global1, 20
tbl_global2, 20
tbl_dynamic, 20
```
More expected would be that `tbl_global1` and `tbl_dynamic` should behave as if they received `&create_expire` of 5.0sec. If this is hard or not expected to work, we should probably err or warn at least on the dynamic construct.

```
module DynamicCreateExpire;

global global_ce = 5.0sec;

function make_cache_table_literal(ce: interval): table[string] of string
        {
        return table() &create_expire=5.0sec;
        }

function make_cache_table_global(ce: interval): table[string] of string
        {
        return table() &create_expire=global_ce;
        }

function make_cache_table_dynamic(ce: interval): table[string] of string
        {
        return table() &create_expire=ce;
        }


global tbl_literal: table[string] of string;
global tbl_global1: table[string] of string;
global tbl_global2: table[string] of string;
global tbl_dynamic: table[string] of string;

        {
        # Insert uid into cache.
        tbl_literal[c$uid] = c$uid;
        tbl_global1[c$uid] = c$uid;
        tbl_global2[c$uid] = c$uid;
        tbl_dynamic[c$uid] = c$uid;
        }

event zeek_init()
        {
        tbl_literal = make_cache_table_literal(5.0sec);

        tbl_global1 = make_cache_table_global(5.0sec);
        # This applies also to tbl_global1 ??
        global_ce = 30days;
        tbl_global2 = make_cache_table_global(5.0sec);

        tbl_dynamic = make_cache_table_dynamic(5.0sec);
        }

event zeek_done()
        {
        print "tbl_literal", |tbl_literal|;
        print "tbl_global1", |tbl_global1|;
        print "tbl_global2", |tbl_global2|;
        print "tbl_dynamic", |tbl_dynamic|;
        }
```

Monisha325:
Fixes #5207

## Problem
When &create_expire, &read_expire, or &write_expire is given a dynamic 
value such as a function argument, Zeek silently ignores it and no 
expiration takes effect. Only literals and globals work correctly.

## Fix
Added a warning in Attr.cc when the expiration expression is not a 
constant (EXPR_CONST) or global variable (EXPR_NAME), so users are 
informed instead of experiencing silent incorrect behavior.

### DIFF
--- a/src/Attr.cc
+++ b/src/Attr.cc
@@ -390,6 +390,18 @@ bool Attributes::CheckAttr(Attr* a, const TypePtr& attrs_t) {
                 return AttrError(
                     "set/table can only have one of &read_expire, &write_expire, "
                     "&create_expire");
+
+            // Warn if the expiration expression is not a constant or global
+            // variable. Dynamic values such as function arguments are evaluated
+            // in isolation and will not reflect the intended runtime value,
+            // leading to silently incorrect expiration behavior.
+            if ( a->GetExpr() &&
+                 a->GetExpr()->Tag() != EXPR_CONST &&
+                 a->GetExpr()->Tag() != EXPR_NAME )
+                reporter->Warning(
+                    "expiration interval is not a constant or global variable; "
+                    "dynamic values such as function arguments will not have "
+                    "the intended effect");
         }
 
 #if 0


evantypanski:
thanks for the contribution :)

Looking at CI and testing this a bit, this doesn't seem to fix the issue, though. The test case provided in #5207 does not warn (but it should) and this case does warn (but it should not):

https://github.com/zeek/zeek/blob/8dd5deff3f82ae62d1a773d480a8f10f64f1936e/scripts/policy/frameworks/notice/extend-email/hostnames.zeek#L14

From some quick testing, at least the addition seems to work fine. So I think this may need a bit more thought: exactly which values would behave surprisingly?

It seems like values not known "statically" (ie when loading the program) will mess up. But, we can know the state of a `redef` variable with addition (I think?). So it may be reasonable to just decline non-constant names. I don't quite know, but it'd take me doing it fully myself to know for sure ;)

Would you be up for another shot? If so, be sure to run the tests and add a new test for this case. You can use the [`rotation.pcap`](https://github.com/zeek/zeek/blob/master/testing/btest/Traces/rotation.pcap) as the example trace from the issue, and you should be able to replicate its results.

Also: Did you use AI in the making of this PR? I think our AI policy isn't distributed here yet, but we will require disclosure of any coding agents.

evantypanski:
hi @Monisha325 

were you planning on returning to this PR? If not, or I get no response, I'll close it soon. Thanks.

Let me know if you want to work on it and need help!

evantypanski:
Closing this since it's becoming stale and does not fix the issue. Feel free to open another PR if you come back to this.

PR #5447, opened 2026-05-11
ci: fix macos build

bbannier:
zeekjs builds against neither `node` (aka today `node@26`) nor `node@25` as shipped by Homebrew, so this patch pins CI to build against `node@24`.


timwoj (approved):

timwoj:
I'm going to cherry-pick this into the release branches as well to keep those running smoothly.

PR #5449, opened 2026-05-11
btest/cluster: Attempt to stabilize overload drop tests

### LINKED ISSUE #5377: Btests `cluster.zeromq.overload-drop` and `cluster.zeromq.overload-worker-proxy-topic-drop` are flaky
I'm seeing this on both my local machine and frequently on CI.

awelzel:
Increase high-water-mark (HWM) to avoid dropping events during normal cluster setup before the actual publishing starts. During the test, publish 1000 events 100 times (100.000 events in total) and expect some of these getting dropped by 1) the sender and 2) the receiver as the HWMs should still be low enough (500 and 333) to not be able to buffer this.

Lets see.

Closes #5377


awelzel:
Looking at two asan-build runs, this should be an improvement, otherwise I'm tempted to go with skipping on ASAN.

There's still failures on Windows, but I'm even more inclined to skip the tests there. Seems not worth the effort.

awelzel:
> There's still failures on Windows, but I'm even more inclined to skip the tests there. Seems not worth the effort.

There's already been https://github.com/zeek/zeek/commit/df7500149f8ca938a97032911216676801de2786 (testing: skip flaky tests in Windows CI), so I've extended on that a bit.

ckreibich (approved):

Issue #5410, opened 2026-04-28
ZIP analyzer and files.log expansion


awelzel:
With the inclusion of the ZIP analyzer, we now log `files.log` entries for every contained file in a ZIP archive.

There's no de-duplication mechanism in place that I'm aware of. I wonder if that's more noise and overhead than useful and we should disable that behavior by default and/or use a separate `zip.log` for the detailed contents, rather than `files.log`?

Maybe the concrete `files.log` entry for a individual ZIP archive could instead have `contained_files: count` and `contained_mime_types: set[string]` or some such.

I'm mostly worried that if someone transferred solely 50 zip files per day, each containing 1000 archived files, their `files.log` is now 1000x larger. And similarly for file extraction, does it actually make sense to extract the ZIP and its contained files?

timwoj:
Oof, that's a good point. Would expanding the schema for files.log be controversial? Maybe this is a good question for the community Slack.

0xxon:
I am partial to the idea of just having a `zip.log` - or, if we see ourselves adding more compression formats - a `compressed_files.log` and putting the file information in there.

awelzel:
cc @ckreibich

bbannier:
I would argue that having contained files in `files.log` is the right thing, but it might increase how much data we emit.

The zip analyzer feeds files from analyzed zip files into Zeek' file analysis framework which means extracted files could end up in other analyzer logs or in extracted files. Since `files.log` is the main log to correlate files to connections or `parent_fuids` (similar to `conn.log` for protocol analyzers), not including contained files in `files.log` could break user workflows.

timwoj:
@awelzel, @ckreibich, and I talked about this for a bit and this is our current thinking of a path forward:

- Disable recursive parsing by default, add a flag to do it if you want but with a big warning about it in the docs about it causing potential DoS of your Zeek.
- Parse dictionary of file names at the time level, write file info to `archives.log` with a parent FUID linking back to the entry in files.log. This is because the non-extracted files won't have actual FUIDs themselves.

In the mean time, until the above is done:

- Disable the analyzer by default, with a big warning about enabling it causing potential DoS of your Zeek.

bbannier:
@timwoj @awelzel @ckreibich, could you capture here

- what is the DoS scenario when enabling this analyzer
- work which would be needed to enable the ZIP analyzer unconditionally

[...]

PR #5450, opened 2026-05-11
Disable ZIP analyzer by default, add policy script

timwoj:
This is related to the discussion in #5410. This PR unregisters the default MIME type for the ZIP analyzer, instead adding a policy script that can be loaded to register it.


scripts/policy/files/zip/enable-analyzer.zeek
0xxon, line 1: I assume we just want one of these files? Having both register and enable-analyzer do the same seems a bit weird :)

scripts/policy/files/zip/register.zeek
```diff
@@ -0,0 +1,6 @@
+## Registers the ZIP analyzer to handle the application/zip mime type
```
0xxon:
```suggestion
##! Registers the ZIP analyzer to handle the application/zip mime type
```

scripts/policy/files/zip/enable-analyzer.zeek
awelzel: I like reading `zip/enable-analyzer.zeek` a lot more from a newbie perspective than `zip/register.zeek`, but that might be just me :-)

timwoj: The only reason I didn't name it that was because it's not calling `enable_file_analyzer`. I'm indifferent though.

awelzel: Sorry for the forth and back here. I could see putting ZIP also into `Analyzer::disabled_analyzers` by default for explicitness "we actively disable ZIP because we know it's sharp" and then doing the enable and register steps in this scripts.

If anyone is looking at what the script actually does, they learn something, too (toggling analyzers *and* registering the mime type). So that might be a nice side-effect.

timwoj: I'll admit I don't like the idea of the analyzer being enabled but not doing anything. It shows up in `zeek -NN` but a user won't know they need to load the policy to make it do something. That's similar to the POP3 analyzer though, so perhaps people are used to it.

That said, I'd rather the script only do one thing. Either it enables the analyzer that's disabled by default, or it registers for the mime type.

0xxon: FWIW - analyzers being enabled without doing anything has a long history with us, so I personally am not too worried about that :)

I also am not sure how many people know about `zeek -NN` - that's a rather technical level.

awelzel: I'll resolve the conversation and approve. Can revisit down the road if/when there's more file analyzers that we'd want to disable by default.

I was pondering/comparing the mime type registration with the default port registration of protocol analyzers. It seems it'd be nice to align the approaches conceptually (here base scripts register mime types by default), but can do so down the road.

awelzel (approved):

ckreibich:
Since Tim is out for a bit and I'd like to get the analyzer out of the critical path I went ahead and merged this.

I'm okay with "register.zeek". I think it'd be good though to be clearer about what the analyzer, when enabled/registered, actually does — "outputs information" is a bit vague. If we put that in the docstring it'll appear in the docs via that. I'm going to tag this as needing docs in the meantime.

PR #5452, opened 2026-05-12
cluster/websocket: Bump ixwebsocket

awelzel:
Fixes the ASAN+IPV6 stack-overflow report from the websocket/one-ipv6.zeek
test and adds support for WebSocket send timeout. Also, support for LibreSSL,
but not sure that's interesting.


awelzel:
This is also part of #5449, but seems better to have it go in as a separate PR and then rebase #5449.

bbannier (approved):

awelzel:
Bumped with #5449 already, actually.

PR #5456, opened 2026-05-13
cluster/zeromq: Fix CMake error when ZeroMQ_INCLUDE_DIR not found

awelzel:
In set_zmq_version() there's some FILE commands that fail if ZeroMQ_INCLUDE_DIR isn't set that hide the better message:

    Could NOT find ZeroMQ (missing: ZeroMQ_LIBRARY ZeroMQ_INCLUDE_DIR)

### DIFF
--- a/src/cluster/backend/zeromq/cmake/FindZeroMQ.cmake
+++ b/src/cluster/backend/zeromq/cmake/FindZeroMQ.cmake
@@ -38,7 +38,9 @@ set(AUXIL_CPPZMQ_DIR ${CMAKE_CURRENT_LIST_DIR}/../auxil/cppzmq)
 find_library(ZeroMQ_LIBRARY NAMES zmq HINTS ${ZeroMQ_ROOT_DIR}/lib)
 find_path(ZeroMQ_INCLUDE_DIR NAMES zmq.h HINTS ${ZeroMQ_ROOT_DIR}/include)
 
-set_zmq_version()
+if (ZeroMQ_INCLUDE_DIR)
+    set_zmq_version()
+endif ()
 
 find_path(ZeroMQ_CPP_INCLUDE_DIR NAMES zmq.hpp HINTS ${ZeroMQ_ROOT_DIR}/include)
 


src/cluster/backend/zeromq/cmake/FindZeroMQ.cmake
```diff
@@ -38,7 +38,9 @@ set(AUXIL_CPPZMQ_DIR ${CMAKE_CURRENT_LIST_DIR}/../auxil/cppzmq)
 find_library(ZeroMQ_LIBRARY NAMES zmq HINTS ${ZeroMQ_ROOT_DIR}/lib)
 find_path(ZeroMQ_INCLUDE_DIR NAMES zmq.h HINTS ${ZeroMQ_ROOT_DIR}/include)
```
bbannier, line 39: Is there a reason we cannot just mark at least these two as `REQUIRED` to simplify the flow? Then we could keep running the following `set_zmq_version` unconditionally.

```suggestion
find_library(ZeroMQ_LIBRARY NAMES zmq HINTS ${ZeroMQ_ROOT_DIR}/lib REQUIRED)
find_path(ZeroMQ_INCLUDE_DIR NAMES zmq.h HINTS ${ZeroMQ_ROOT_DIR}/include REQUIRED)
```

The code around `ZeroMQ_CPP_VERSION` below is a bit messier since it implements a fallback, but I think here this should just work.

awelzel: The message of the proposed change is as follows:
```
--   Building in plugin: zeek-packet-source-udp (/src/auxil/zeek-packet-source-udp)
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
  Could NOT find ZeroMQ (missing: ZeroMQ_LIBRARY ZeroMQ_INCLUDE_DIR)
  (Required is at least version "4.3.0")                                                       
Call Stack (most recent call first):                                                           
  /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:654 (_FPHSA_FAILURE_MESSAGE)
  src/cluster/backend/zeromq/cmake/FindZeroMQ.cmake:61 (find_package_handle_standard_args)
  src/cluster/backend/CMakeLists.txt:14 (find_package)
```

And with your proposal:
```
--   Building in plugin: zeek-packet-source-udp (/src/auxil/zeek-packet-source-udp)
CMake Error at src/cluster/backend/zeromq/cmake/FindZeroMQ.cmake:38 (find_library):
  Could not find ZeroMQ_LIBRARY using the following names: zmq
Call Stack (most recent call first):
  src/cluster/backend/CMakeLists.txt:14 (find_package)
```

So I find the current one a bit nicer as it makes the caller fail:
```
find_package(ZeroMQ 4.3.0 REQUIRED)
```

ckreibich (approved):
Thanks for looking into this Arne. For the record, the less helpful message was
```
file STRINGS file
  "/home/christian/devel/zeek/zeek/src/cluster/backend/ZeroMQ_INCLUDE_DIR-NOTFOUND/zmq.h"
  cannot be read.
```
So both of the suggestions are much better.

PR #5458, opened 2026-05-13
Mark SSL::hash_algorithms and SSL::signature_algorithms as deprecated

### LINKED ISSUE #5320: ML-DSA Signature Algorithms Mislabeled as RSA-PSS

0xxon:
We don't use those tables in Zeek ourselves, and they are (kind of) outdated with TLS 1.3. Users that need these values are probably better served by creating their own tables and keeping them up to date.

Closes GH-5320

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -69,6 +69,11 @@ Changed Functionality
 Deprecated Functionality
 ------------------------
 
+- The ``SSL::hash_algorithms`` and ``SSL::signature_algorithms`` tables have been deprecated.
+  They are outdated with TLS 1.3, and not used in Zeek. If you currently use these tables,
+  you should switch to your own copy of the IANA tables and consider also using the
+  TLS SignatureScheme IANA table.
+
 Zeek 8.2.0
 ==========
 
--- a/scripts/base/protocols/ssl/consts.zeek
+++ b/scripts/base/protocols/ssl/consts.zeek
@@ -92,7 +92,7 @@ export {
 		[5] = "sha384",
 		[6] = "sha512",
 		[8] = "Intrinsic",
-	} &default=function(i: count):string { return fmt("unknown-%d", i); };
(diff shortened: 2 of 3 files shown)


timwoj (approved):
This needs a NEWS entry about the deprecation.

0xxon:
I also missed that this necessitates a change in the one test that uses these tables. I amended the commit with the NEWS entry and the test change.

timwoj (approved):

PR #5459, opened 2026-05-13
Add support for signature_algorithms_cert TLS extension. Fix bug raising signature_algorithm extension

### LINKED ISSUE #5354: ssl_extension_signature_algorithm event incorrectly raised by CertificateRequest message
The SSL analyzer (both spicy and binpac) has a logic error which causes the `ssl_extension_signature_algorithm` to be incorrectly raised when a `CertificateRequest` Handshake message is encountered.

0xxon:
This PR creates a new event for the (new in TLS 1.3) signature_algorithms_cert extension.

This PR also fixes GH-5354 - which was discovered while developing this feature.

The event is added to both the spicy and the binpac implementations.

The only part of this PR authored with AI is a java program that helped with creating the test pcap. The java source code is attached to the commit that adds the pcap.


src/analyzer/protocol/ssl/events.bif
```diff
@@ -187,9 +187,33 @@ event ssl_extension_ec_point_formats%(c: connection, is_client: bool, point_form
...
+## handshake. It gives the list of signature schemes supported by the
+## client for Certificates.
+##
+## c: The connection.
+##
+## is_client: True if event is raised for the client side of the connection
+##            (the side that sends the client hello). This is typically equivalent
+##            with the originator, but does not have to be in all circumstances.
+##
+## signature_schemes: List of supported signature schemes.
```
awelzel, line 204: There's the following in the RFC (https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.3)

>    Note: This enum is named "SignatureScheme" because there is already a
   "SignatureAlgorithm" type in TLS 1.2, which this replaces.  We use
   the term "signature algorithm" throughout the text.

And so I'm wondering if renaming this to `signature_algorithms` is a fair suggestion and hide the "scheme" terminology. Wireshark also uses algorithms, so that might make sense to align:

Image

src/analyzer/protocol/ssl/tls-handshake-analyzer.pac
```diff
@@ -246,25 +246,43 @@ refine connection Handshake_Conn += {
...
+
+			zeek::BifEvent::enqueue_ssl_extension_signature_algorithm(zeek_analyzer(), zeek_analyzer()->Conn(), ${rec.is_orig} ^ flipped_, std::move(slist));
 			}
+		else if ( type == 50 ) // signature_algorithms_cert
+			{
+			if ( ! ssl_extension_signature_algorithms_cert )
+				return true;
 
-		zeek::BifEvent::enqueue_ssl_extension_signature_algorithm(zeek_analyzer(), zeek_analyzer()->Conn(), ${rec.is_orig} ^ flipped_, std::move(slist));
+			auto slist = zeek::make_intrusive<zeek::VectorVal>(zeek::id::find_type<zeek::VectorType>("index_vec"));
```
awelzel: Nit, this one exists, no need for the lookup:

```suggestion
			auto slist = zeek::make_intrusive<zeek::VectorVal>(zeek::id::index_vec));
```

awelzel: Could also `->Reserve()` to avoid some extra allocations during the Assign() / Append() below.

src/analyzer/protocol/ssl/tls-handshake-analyzer.pac
```diff
@@ -246,25 +246,43 @@ refine connection Handshake_Conn += {
 		return true;
 		%}
 
-	function proc_signature_algorithm(rec: HandshakeRecord, supported_signature_algorithms: SignatureAndHashAlgorithm[]) : bool
+	function proc_signature_algorithm(rec: HandshakeRecord, type: uint16, supported_signature_algorithms: uint16[]) : bool
 		%{
-		if ( ! ssl_extension_signature_algorithm )
-			return true;
+		if ( type == 13 ) // signature_algorithm
```
awelzel: Would it be possible to use the enum value here and below for the `== 50` instead of the comment?

awelzel (approved):
Just nits, lgtm.

src/analyzer/protocol/ssl/events.bif
0xxon: I actually thought about this when writing up the documentation - and I think it should be signature schemes, for easier findability, and to differentiate from signature algorithms. Especially when you are not keeping super up to date with what happens you might misunderstand how to look them up afterwards. 

I think wireshark is in the wrong here - and I also think the wireshark presentation is a bit weird because they let you expand the signature algorithm into its two parts :)

In any case - I completely accept that this is something where one can have differing opinions - but as they are called Signature Schemes on the IANA side, this makes most sense to me :)

awelzel: Thanks! Yeah, fair. The iana page as "Scheme" in the header (could actually see linking it).

https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-signaturescheme

Let me resolve.

PR #5464, opened 2026-05-17
Remove obsolete Broker WebSocket server references

Neverlord:
Mostly documentation cleanup, but the Broker manager also had some code paths that are no longer needed.


src/broker/Manager.h
```diff
@@ -125,7 +119,7 @@ class Manager : public zeek::cluster::Backend, public iosource::IOSource {
      * return value equals *port* on success. If *port* equals 0, then the
      * return values represents the bound port as chosen by the OS.
      */
-    uint16_t Listen(const std::string& addr, uint16_t port, BrokerProtocol protocol = BrokerProtocol::Native);
```
0xxon, line 128: If I understand this correctly, this is not in a detail namespace. I think this means that this, in theory, needs a deprecation cycle...

ckreibich: Yeah, I agree ... since the Broker manager is always there, it's better that way. We can just introduce the new version alongside the old, remove the default argument from the old to avoid ambiguity, and deprecate it.

scripts/base/frameworks/broker/main.zeek
```diff
@@ -7,14 +7,6 @@ export {
 	## otherwise, this is the port to connect to and listen on.
 	const default_port = 9999/tcp &redef;
 
-	## Default port for Broker WebSocket communication. Where not specified
-	## otherwise, this is the port to connect to and listen on for
-	## WebSocket connections.
-	##
-	## See the Broker documentation for a specification of the message
-	## format over WebSocket connections.
-	const default_port_websocket = 9997/tcp &redef;
```
ckreibich, line 16: I'm thinking we should deprecate these first too. The Broker framework is still always loaded (from `base/init-frameworks-and-bifs`), and it's relatively likely that people still touch these in some of their scripts. We could remove the docstrings at this point, though.

Would you agree @0xxon? @awelzel?

src/broker/Manager.h
ckreibich: That implies leaving the `BrokerProtocol` enum intact for now, too. I think that's okay — we can trigger a runtime error if somebody tries to listen via websockets, and point them at `Cluster::listen_websocket()`. Even though that's at the script layer, I think it's the right place people should consider in the unlikely case that this affects them. Sound good?

scripts/base/frameworks/broker/main.zeek
ckreibich: Actually — no. `Broker::listen_websocket` is already gone, so it's fair to ask people to remove these at the same time they transition to `Cluster::listen_websocket`. Let me resolve this.

ckreibich:
I rebased this onto master, bumped `auxil/broker` to pull in the Broker-side WebSocket stack removal, updated one test that noticed the internal telemetry changing accordingly, and expanded the `NEWS` wording to convey that this expands on the earlier removal of `Broker::listen_websocket()`.

awelzel (approved):

PR #5465, opened 2026-05-18
github/workflows: Add cifuzz.yml

awelzel:
Instructions from there:

    https://google.github.io/oss-fuzz/getting-started/continuous-integration/

There's way to limit when to run this workflow (we would not need to do it for doc only changes), but for now see if it works at all...

    https://google.github.io/oss-fuzz/getting-started/continuous-integration/#branches-and-paths

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,37 @@
+name: CIFuzz
+
+on:
+  pull_request:
+
+permissions: {}
+
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zeek'
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'zeek'
+        fuzz-seconds: 600
+        output-sarif: true
+    - name: Upload Crash
+      uses: actions/upload-artifact@v4
+      if: failure() && steps.build.outcome == 'success'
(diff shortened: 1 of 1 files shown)


awelzel:
Ah, well, the new job fails, but that's good. The job added in this PR is for testing whether the OSS-Fuzz fuzzer builds work. And it fails because of the base-builder image doesn't have OpenSSL 3.0 or later:

```
-- Looking for pcap_dump_open_append
-- Looking for pcap_dump_open_append - found
CMake Error at /usr/local/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR: Found unsuitable version "1.1.1f", but
  required is at least "3.0" (found /usr/lib/x86_64-linux-gnu/libcrypto.so, )
Call Stack (most recent call first):
  /usr/local/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:598 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/share/cmake-3.29/Modules/FindOpenSSL.cmake:686 (find_package_handle_standard_args)
  CMakeLists.txt:885 (find_package)
```

And indeed, this also happens on OSS-Fuzz:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#zeek
https://oss-fuzz-build-logs.storage.googleapis.com/log-ca7ce470-ec6c-40a0-95c8-963e9b2ec6c9.txt

```
$ docker run --rm -it --entrypoint=/bin/bash gcr.io/oss-fuzz-base/base-builder
root@d3cf7cfa2423:/src# gcc --version
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
```

I think this "just" needs bumping to `gcr.io/oss-fuzz-base/base-builder:ubuntu-24-04`.

OSS-Fuzz PR: https://github.com/google/oss-fuzz/pull/15522

timwoj (approved):

PR #5469, opened 2026-05-20
encrypted_client_hello TLS 1.3 extension

0xxon:
This adds the encrypted client hello extension to both the binpac and the spicy implementation.

The second commit (binpac implementation) was mostly authored by Claude Opus 4.7. I have reviewed and understand all of the changes in the commit. 

The spicy implementation was hand written.


src/analyzer/protocol/ssl/tls-handshake-protocol.pac
```diff
@@ -832,6 +833,27 @@ type ConnectionId(rec: HandshakeRecord) = record {
...
 };
 
+# The encrypted_client_hello extension carries an type; only
+# the outer form (0) carries further data. The second variant
+# should not occur in unencrypted packets.
+type EncryptedClientHelloSwitch(rec: HandshakeRecord) = record {
+	tpe: uint8;
+	ech: case tpe of {
+		0 -> outer: EncryptedClientHello(rec);
+		default -> inner: bytestring &restofdata &transient;
```
timwoj, line 844: Should this report a weird since it shouldn't ever happen?

src/analyzer/protocol/ssl/spicy/SSL.spicy
```diff
@@ -1124,6 +1126,21 @@ type Extension = unit(inout sh: Share, client_hello: bool) {
     }
 };
 
+type EncryptedClientHelloSwitch = unit(sh: Share) {
+    tpe: uint8; # 0 = outer, 1 = inner - we should never ever see 1 as a type, but who knows, right...
```
timwoj, line 1130: Same question about a weird for inner?

NEWS
```diff
@@ -57,6 +57,9 @@ New Functionality
 - The event ``ssl_extension_signature_algorithms_cert`` was added. It is raised for the
   signature_algorithms_cert TLS extension.
 
+- The event ``ssl_extension_encrypted_client_hello`` was added. It contains information about
+  the encrypted_client_hello TLS extension.
+
```
awelzel, line 66: Would it make sense to have a policy script that adds an `ech` bool to ssl.log and the kdf/aead/config ids? Like `policy/protocols/ssl/ech.zeek` ? Maybe useful for compliance use cases and potentially makes for a more interesting NEWS entry and with that new event should be fairly straight-forward?

```
ech:          bool             &log &optional; 
ech_kdf_id:          count             &log &optional; 
ech_aead_id:          count             &log &optional; 
ech_config_id:          count             &log &optional; 
```

src/analyzer/protocol/ssl/tls-handshake-protocol.pac
0xxon: *shrug*
Potentially. There are so many things in this protocol that should never happen that we don't raise a weird for.
If one is motivated, one can read this out of the raw extension data. I am also happy to raise a weird for it.

src/analyzer/protocol/ssl/spicy/SSL.spicy
0xxon: Same answer. We could, but there are many many protocol weirdnesses we don't check/raise for.

NEWS
0xxon: Not really / a bit unlikely. The extension is specifically made in a way where it's not supposed to expose if it was used in the end. So it only gives away if the client has the potential capability to use ech - and I don't think many people care.
I would be happy to add the kdf id/aead id to the ssl_log_ext - those seem the most interesting.

awelzel: > So it only gives away if the client has the potential capability to use ech - and I don't think many people care.

Makes sense. And those that do care can do so themselves.

> I would be happy to add the kdf id/aead id to the ssl_log_ext - those seem the most interesting.

From a bit of skimming, seems config_id might rotate/change, so thought it could be interesting to some, but yeah, maybe that's then just an external script. I'll resolve and let you decide :-)

0xxon:
I added both of the requested weirds. I am going to open an additional pr for the change of ssl_log_ext.

awelzel (approved):

PR #5471, opened 2026-05-20
Fix segfault when `global_ids` called before types

### LINKED ISSUE #5197: Accessing global_ids() in a global initialization seg-faults
The following seg-faults:
```
global gi = global_ids();
```
The problem is that not all variables have types associated with them during parse-time (e.g. `LDAP::bind_request`, which is the one that the debugger points up as crashing for me).

evantypanski:
Fixes #5197


src/zeek.bif
```diff
@@ -2755,6 +2755,11 @@ function global_ids%(%): id_table
 
 	for ( const auto& [_, id] : zeek::detail::global_scope()->Vars() )
 		{
+		if ( ! id->GetType() ) {
+			zeek::reporter->Warning("skipping global ID without type %s", id->Name());
+			continue;
```
awelzel: Hmm.. if the top-level `global_ids()` at parsing time is a valid use, the warning isn't appropriate.

Would either suggest to either drop the warning, or disallow `global_ids()` during parsing completely, probably with slight preference for the latter, possibly depending what the original motivation was to use it top-level?

```
if ( run_state::is_parsing )
    {
    reporter->Error("top-level global_ids() call unsupported")
    return ids;
    }
```

evantypanski: Yeah that's probably better

my main gripe is that it's tough for the caller to tell something bad happened other than `ids` is empty. It'd be cool if we had some sort of type to return the result of an operation or something... ;)

awelzel: In my mind, `reporter->Error()` when `run_state::is_parsing == true` makes Zeek not even start, so in this case that seems fine. But yeah, result, optional, expected story here :-)

awelzel (approved):

PR #5472, opened 2026-05-20
doc: Add websockets pip --target trick

awelzel:
Thanks @0xxon!

### DIFF
--- a/doc/building-from-source.rst
+++ b/doc/building-from-source.rst
@@ -207,6 +207,18 @@ The `ZeekControl`_ management shell has an optional dependency on the Python
 This package is available on PyPI under the name ``websockets``. On Ubuntu
 and Fedora, the distribution package is called ``python3-websockets``.
 
+.. note::
+
+   Some distributions do not yet have websockets version 11.0 or later packaged.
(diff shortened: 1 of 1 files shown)


doc/building-from-source.rst
```diff
@@ -207,6 +207,17 @@ The `ZeekControl`_ management shell has an optional dependency on the Python
 This package is available on PyPI under the name ``websockets``. On Ubuntu
 and Fedora, the distribution package is called ``python3-websockets``.
 
+.. note::
+
+   Some distributions do not yet have websockets version 11.0 or later packaged.
+   As a workaround, we recommend installing a more recent websockets version into
```
evantypanski:
```suggestion
   If you encounter troubles when using the distribution package, we recommend installing a more recent websockets version into
```
I think this makes it clearer that you may hit into troubles if it doesn't work, and it applies to the distro package. Since above we also mention the PyPI package. Just my preference.

awelzel: Yeah, much better, thanks!

evantypanski (approved):

PR #5473, opened 2026-05-20
Tune ZeroMQ drop tests to overflow the XPUB watermark more easily

ckreibich:
These two tests consistently fail on my machines at the previous XPUB high watermark. Even with a fraction of 50 they're still unreliable. With 100, and me keeping the machines otherwise idle, it passes hundreds of times without error. Since the threshold shouldn't matter as much as us verifying that the overflow is detected relibly, I think this is okay.

I experimented with making the ping event consume additional buffer space by adding large strings to its argument. This prolongs the tests considerably (for example, a 10K buffer makes the overload-drop test take nearly twice as long), but has no impact on reaching the high water mark, so (perhaps unsurprisingly) the overhead all seems to be in per-message processing.

### DIFF
--- a/testing/btest/cluster/zeromq/overload-drop.zeek
+++ b/testing/btest/cluster/zeromq/overload-drop.zeek
@@ -48,8 +48,8 @@ const batch = 1000;
 # seems to reliably pass with an ASAN build, nothing more. They
 # interact with the number of publishes and chosen batch. If this
 # becomes unmaintainalbe going forward, easier to skip on ASAN IMO.
-redef Cluster::Backend::ZeroMQ::xpub_sndhwm = batch / 5;
-redef Cluster::Backend::ZeroMQ::onloop_queue_hwm = batch / 6;
+redef Cluster::Backend::ZeroMQ::xpub_sndhwm = batch / 100;
+redef Cluster::Backend::ZeroMQ::onloop_queue_hwm = batch / 20;
 
 global test_nodes = set( "proxy", "worker-1", "worker-2" ) &ordered;
 # @TEST-END-FILE
--- a/testing/btest/cluster/zeromq/overload-worker-proxy-topic-drop.zeek
+++ b/testing/btest/cluster/zeromq/overload-worker-proxy-topic-drop.zeek
@@ -119,7 +119,7 @@ event zeek_done()
 # seems to reliably pass with an ASAN build, nothing more. They
 # interact with the number of publishes and chosen batch. If this
 # becomes unmaintainalbe going forward, easier to skip on ASAN IMO.
-redef Cluster::Backend::ZeroMQ::xpub_sndhwm = batch / 5;
+redef Cluster::Backend::ZeroMQ::xpub_sndhwm = batch / 100;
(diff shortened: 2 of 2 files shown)


ckreibich:
I didn't show the failure — it's this for each of the senders:
```
$ cat .tmp/cluster.zeromq.overload-drop/.diag
had xpub_drops?, F
had onloop_drops?, F
--- /dev/fd/63  2026-05-20 20:02:21.148506903 +0000
+++ /dev/fd/62  2026-05-20 20:02:21.148506903 +0000
@@ -1,2 +1,2 @@
-had xpub_drops?, T
+had xpub_drops?, F
 had onloop_drops?, F
```

ckreibich:
This initially stumbled over MacOS in the manager's results (see [here](https://cirrus-ci.com/task/4850314903289856?logs=test#L2589)). I explored this a bit via Cirrus's SSH access and it didn't like the original fractions (5, 6), but was also not happy with the introduction of 100 for the xpub one alone. I updated the change to also lower the watermark for onloop (with a fraction of 20), which seems to work everywhere.

I'm definitely uneasy about this, because I noticed that if I set both fractions to 100 the test again fails on my systems, with
```
$ cat .tmp/cluster.zeromq.overload-drop/.diag
had xpub_drops?, T
had onloop_drops?, T
--- /dev/fd/63  2026-05-20 22:39:30.984387090 +0000
+++ /dev/fd/62  2026-05-20 22:39:30.985387097 +0000
@@ -1,2 +1,2 @@
 had xpub_drops?, T
-had onloop_drops?, F
+had onloop_drops?, T
```
seemingly always in worker-2.

I wonder if we could test the two watermarks in isolation, or perhaps forcibly disable any processing so we're guaranteed to hit watermarks simply based on the number of sent messages.

awelzel:
> These two tests consistently fail on my machines at the previous XPUB high watermark. 

Do they fail with `btest -j` runs only? Or also when run individually? With ASAN or without?

Many tests in `cluster/` run `btest-bg-run` with multiple processes and so this can easily spawn many more than `nproc` processes at a time and starve available CPU. I'm actually pretty surprised there isn't more tests that fall over every now and then.

> I wonder if we could test the two watermarks in isolation, or perhaps forcibly disable any processing so we're guaranteed to hit watermarks simply based on the number of sent messages.

Feel free to experiment, but I'd actually be pretty okay to skip these tests completely (or just on ASAN), or have a separate test suite / run where tests run serialized to ensure there's a reasonable amount of CPU available for them. Making them robust in the face of CPU time starvation, not sure is worth the effort.

ckreibich:
> Do they fail with btest -j runs only? Or also when run individually? With ASAN or without?

I should have added more context: they fail consistently when run in isolation, no other tests, regardless of ASAN or not. I was pretty surprised.

>  I'm actually pretty surprised there isn't more tests that fall over every now and then.

Me too. :-) The reruns definitely mask failures. In the MacOS run I linked there's one for `cluster.websocket.metadata` for example, that worked out during the re-run.

ckreibich:
Closing this in favor of #5487.

PR #5474, opened 2026-05-20
CI: Rework Windows Dockerfile to use newer version of Windows

timwoj:
This updates the Dockerfile used for Windows CI builds to the lstc2025 image, which is much newer than the old 2019 image. It also updates to Visual Studio 2026 (or "latest", since they don't offer versioned installers anymore past 2022).

This requires merging https://github.com/zeek/libunistd/pull/15 as well.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -647,7 +647,7 @@ windows_task:
   timeout_in: 120m
   windows_container:
     dockerfile: ci/windows/Dockerfile
-    os_version: 2019
+    os_version: 2025
     cpu: 8
     # Not allowed to request less than 8GB for an 8 CPU Windows VM.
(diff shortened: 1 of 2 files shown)


timwoj:
This will apparently have to wait until we move off Cirrus:

> Failed to start an instance: FAILED_PRECONDITION: Unsupported OS version 2025! Only 2019 are supported!

timwoj:
@bbannier Do you want to pull this into your Github branch? I'm not using it at all for the CircleCI config and it doesn't work for Cirrus, so it's probably best to live there. I can close this PR then.

timwoj:
Benjamin cherry-picked this into his GH CI actions branch, so I'm closing this PR.

PR #5476, opened 2026-05-21
Add Github Actions CI

bbannier:


.github/workflows/check.yml
```diff
@@ -0,0 +1,569 @@
...
+  HILTI_CXX_COMPILER_LAUNCHER: "ccache"
+
+  ZEEK_CI_BTEST_RETRIES: 2
+  ZEEK_CI_SKIP_EXTERNAL_BTESTS: 1 # TODO: enable this eventually
+
+  # Prevent Docker build steps from posting to the summary.
+  DOCKER_BUILD_SUMMARY: false
+
+  # TODO: Migrate scripts which use this legacy variable.
+  CIRRUS_WORKING_DIR: ${{ github.workspace }}
```
timwoj: This should change to `ZEEK_CI_WORKING_DIR` after the other script changes.

.github/workflows/check.yml
```diff
@@ -0,0 +1,569 @@
...
+  push:
+    branches: [master]
+    tags: ["*"]
+  pull_request:
+  workflow_dispatch:
+
+env:
+  GIT_CONFIG_PARAMETERS: "'safe.directory=*'"
+
+  GITHUB_WORKSPACE: "${{ github.workspace }}"
```
timwoj: Can this be replaced with the `ZEEK_CI_WORKING_DIR` variable instead of duplicating it?

.github/workflows/check.yml
```diff
@@ -0,0 +1,569 @@
...
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      with:
+        submodules: recursive
+        persist-credentials: false
+
+    - name: Set up MSVC Dev Cmd
+      uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
+      with:
+        sdk: "10.0.26100.0"
```
timwoj, line 445: I love how this is technically a Windows 11 SDK, but is still 10.x 🤦🏼. I actually had to go look.

.github/workflows/check.yml
```diff
@@ -0,0 +1,569 @@
...
+
+    - name: Install dependencies
+      run: |
+        choco install -y --no-progress winflexbison3 ccache
```
timwoj: I'm surprised this doesn't need `sed`.

.github/workflows/check.yml
```diff
@@ -0,0 +1,569 @@
...
+
+    - name: Build
+      run: |
+        ./ci/windows/build.cmd
+
+    - name: Test
+      run: |
+        ./ci/windows/test.cmd
+
+  freebsd:
```
timwoj: With FreeBSD not being officially supported anymore, I think this entire section can just be removed. Or were you thinking we should include it for now and remove it later?

ci/windows/build.cmd
```diff
@@ -2,10 +2,17 @@
 :: cmd current shell. This path is hard coded to the one on the CI image, but
 :: can be adjusted if running builds locally. Unfortunately, the initial path
 :: isn't in the environment so we have to hardcode the whole path.
-call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
+call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
```
timwoj: I wonder if we should make this path an environment variable. The reason I ask is that the path is different on CircleCI (it's the other path you changed from).

timwoj (changes requested):

.github/workflows/check.yml
bbannier: I added `sed` just in case 🤷

ci/windows/build.cmd
bbannier: Yeah, that would make a lot of sense. All this does is set up some env vars, and since I use `ilammy/msvc-dev-cmd` I actually do not need it at all and just need to keep it consistent with what it installed. Hardcoding this all over the place also makes it much harder to test different msvc versions.

One var that script sets is `VSINSTALLDIR` which in this case would correspond to `C:\Program Files\Microsoft Visual Studio\2022\Enterprise`. I switched the scripts to require and use that var now, can you work with that? If you don't need it either we could also just remove all these calls of `vcvarsall.bat`.

.github/workflows/check.yml
bbannier: Yeah, I was wondering whether it was worth leaving as documentation, but removed it now.

.github/workflows/check.yml
bbannier: I believe I just copied this version, is it correct?

.github/workflows/check.yml
```diff
@@ -0,0 +1,530 @@
+name: Check
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [master]
```
timwoj: You'll want to catch pushes to `release/*` also, even if we're not doing release docker builds from GH.

.github/workflows/check.yml
```diff
@@ -0,0 +1,530 @@
...
+            # For now, just check if the external plugins are available.
+            zeek -N Zeek::PerfSupport
+            zeek -N Zeek::MoreHashes
+            zeek -N Zeek::Cluster_Backend_NATS
+            zeek -N Seiso::Kafka
+
+    name: ${{ matrix.id }}
+
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
```
timwoj: Just for consistency with the others, what version does this point to?

.github/workflows/check.yml
timwoj: Yeah, it's correct. I'm just musing on Microsoft's versioning being stupid.

timwoj (approved):
I just had a couple of minor comments. I can fix them during the merge.

timwoj:
I also ported your additional script-level changes over into my script-changes branch so I'll merge that first, then rebase this one on top of it.

.github/workflows/check.yml
```diff
@@ -0,0 +1,532 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - id: fedora-43
+          file: ci/fedora-43/Dockerfile
+          env:
+            ZEEK_CI_CONFIGURE_FLAGS: '--build-type=debug --disable-broker-tests --prefix=$ZEEK_CI_WORKING_DIR/install --ccache -D FETCHCONTENT_FULLY_DISCONNECTED:BOOL=ON'
```
ckreibich, line 44: I'm trying to map these configurations to what we had in Cirrus and I'm wondering what motivated this one. It is closest to our clang-tidy config. Help me out?

.github/workflows/check.yml
```diff
@@ -0,0 +1,532 @@
...
+          run: |
+            ./ci/pre-build.sh
+            ./ci/build.sh
+            ./ci/test.sh
+
+        - id: fedora-44
+          file: ci/fedora-44/Dockerfile
+          env:
+            ZEEK_CI_CONFIGURE_FLAGS: '--prefix=$ZEEK_CI_WORKING_DIR/install --libdir=$ZEEK_CI_WORKING_DIR/install/lib --binary-package --enable-static-broker --enable-static-binpac --disable-broker-tests --build-type=Release --ccache --enable-werror'
```
ckreibich, line 53: This is basically `binary_config`, got it. So `fedora-44-binary` might be a better name, to mirror `debian-13-binary` below.

.github/workflows/check.yml
```diff
@@ -0,0 +1,532 @@
...
+    name: ${{ matrix.id }}
+
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+      with:
+        submodules: recursive
+        persist-credentials: false
+
+    - name: Cache Docker image
+      # uses: ScribeMD/docker-cache@0.5.0
```
ckreibich, line 372: What prompted the choice of the `AndreKurait` version? I'm also curious how this actually works. :-) It must be running the `docker build` because I'm not seeing an explicit build anywhere.

.github/workflows/check.yml
```diff
@@ -0,0 +1,532 @@
...
+    #   uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+
+    - name: Run CI
+      env:
+          # We convert the environment variable dictionaries to JSON to work
+          # around potential issues with shell quoting.
+          _GHA_ENV: ${{ toJSON(env) }}
+          _MATRIX_ENV: ${{ toJSON(matrix.env) }}
+      run: |
+        # Assembly environment variables for the script. We support both global
```
ckreibich:
```suggestion
        # Assemble environment variables for the script. We support both global
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,532 @@
...
+  HILTI_CXX_COMPILER_LAUNCHER: "ccache"
+
+  ZEEK_CI_WORKING_DIR: "${{ github.workspace }}"
+  ZEEK_CI_CONFIGURE_FLAGS: '--build-type=release --disable-broker-tests --prefix=$ZEEK_CI_WORKING_DIR/install --ccache --enable-werror -D FETCHCONTENT_FULLY_DISCONNECTED:BOOL=ON'
+
+  ZEEK_CI_BTEST_RETRIES: 2
+  ZEEK_CI_SKIP_EXTERNAL_BTESTS: 1 # TODO: enable this eventually
+
+  # Prevent Docker build steps from posting to the summary.
+  DOCKER_BUILD_SUMMARY: false
```
ckreibich, line 31: Is this for `AndreKurait/docker-cache`?

ckreibich (approved):
Looks great to me. I only put in questions below. I think it'd be nice to centralize the configure flags again, like we had in Cirrus. But certainly not critical.

The build-up of `container.env` is cool 👍

timwoj:
I talked to @ckreibich and he's alright with me merging this even with his questions above. If there's any follow-ups from them they can be fixed later.

PR #5477, opened 2026-05-21
docs: clarify note about package manager versions in Quickstart Guide

sujalavnelavai:
This update adds a brief clarification to the Quickstart Guide noting that some Linux distributions may provide older versions of Zeek through their system package managers. This helps new users understand why their installed version may not match the documentation or available features, and encourages them to consider building from source when they need the latest release.

The change improves accuracy and reduces potential confusion for first‑time users following the installation instructions.

### DIFF
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -15,7 +15,6 @@ systems and does not require custom hardware. See :doc:`install` in order to
 install from pre-built binary packages, or :doc:`building-from-source` in order
 to build Zeek from source.
 
-
 We will first analyze previously captured network traffic from a ``pcap`` file -
 :download:`quickstart.pcap <traces/quickstart.pcap>`. Later, we will use Zeek to
 monitor live traffic. Each section builds on the previous section.
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -12,7 +12,7 @@ Quick Start Guide
 
 Zeek is a network traffic analyzer. Zeek works on most modern Unix-based
 systems and does not require custom hardware. See :doc:`install` in order to
-install from pre-built binary packages (note that some distributions may provide older versions),  or :doc:`building-from-source` in order
+install from official Zeek binary packages (see https://docs.zeek.org/en/master/install.html#binary-packages),  or :doc:`building-from-source` in order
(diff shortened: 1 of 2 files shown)


doc/quickstart.rst
```diff
@@ -12,7 +12,7 @@ Quick Start Guide
 
 Zeek is a network traffic analyzer. Zeek works on most modern Unix-based
 systems and does not require custom hardware. See :doc:`install` in order to
-install from pre-built binary packages, or :doc:`building-from-source` in order
+install from pre-built binary packages (note that some distributions may provide older versions),  or :doc:`building-from-source` in order
```
bbannier: I believe this was talking about the binary packages provided by the Zeek project, maybe instead linking to https://docs.zeek.org/en/master/install.html#binary-packages would make this clearer.

evantypanski: yeah, I was about to suggest putting it in the install section instead, namely where the distro packages are discussed. I think here would be the clearest, but honestly could be anywhere in there:

https://github.com/zeek/zeek/blob/cb80b6ac8d6f101f459d141529ae8854448aeb98/doc/install.rst?plain=1#L102-L103

sujalavnelavai: Thanks for the suggestion!
I will update the wording to clarify that this refers to the Zeek‑provided binary packages and link to the binary package section.

sujalavnelavai: Thanks for the additional suggestion!
I can move this clarification into the install.rst section where the distro packages are discussed.
I will update the PR accordingly.

sujalavnelavai:
Thanks for the feedback!
I have updated the PR by restoring the original Quickstart wording and moving the clarification about binary package versions into the install.rst section, as suggested. The note now appears where distribution package details are discussed and links to the official Zeek binary packages page.

Let me know if any further adjustments are needed.

evantypanski:
thanks, I just pushed the touchup

(forgot to approve but approved, oops)

sujalavnelavai:
Thank you for reviewing and merging my PR. I appreciate the clear feedback during the process - it helped me understand the documentation structure better. I am looking forward to contributing more to Zeek.

PR #5479, opened 2026-05-21
Adapt CI shell scripts to run on other CI providers

timwoj:
I'm pulling these changes out of my CircleCI branch to reach a common consensus with @bbannier's GitHub CI branch before we get too much further along. This adapts the various shell scripts used during the builds to not be Cirrus-specific. A few notes about the changes:

- I didn't touch `benchmark.sh` or `upload-coverage.sh` yet, because I haven't gotten to running those scripts yet on CircleCI.
- `init-external-repos.sh` still has some `CIRRUS` variables in it because I have questions about how to even handle that block. CircleCI doesn't really have read/write permissions like that. The SSH key is generally available to all runs via the environment (I can explain separately) though so we don't need to be worried about it getting logged via the script. This block might be able to just go away.
- `collect-repo-info.py` has a usage of `CIRRUS_CI` but the script works fine on CircleCI so that check might be able to go away.
- I split the `build.cmd` script for Windows into `configure.cmd` and `build.cmd` so that I could run them as separate steps similar to the Linux and macOS jobs. I'm not sure why we did them together like that on Cirrus in the first place.
- There's a few TODOs sprinkled around where I had questions about something but couldn't quickly resolve it.


timwoj:
This is obviously going to fail to build on Cirrus because I'm not going to go through and adapt that configuration to match.

ci/windows/build.cmd
```diff
@@ -4,8 +4,4 @@
 :: isn't in the environment so we have to hardcode the whole path.
 call "c:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
 
-mkdir build
-cd build
-
-cmake.exe .. -DCMAKE_BUILD_TYPE=release -DVCPKG_TARGET_TRIPLET="x64-windows-static" -DENABLE_ZEEK_UNIT_TESTS=yes -DENABLE_CCACHE=yes -G Ninja
-cmake.exe --build .
+cmake.exe --build build -j 4
```
bbannier: Could we not tie parallelism to a particular runner's hardware with `-j4` here? Previously it was pretty clear that we were using `Ninja` which build in parallel by default, with the new `configure.cmd` this is less clear. Maybe we could go with something like `cmake.exe --build build -j $(nproc)` here 🤷

ci/windows/run-btests.sh
```diff
@@ -63,4 +63,10 @@ if [ -d .tmp ]; then
     tar -czf tmp.tar.gz .tmp 2>/dev/null || true
 fi
 
+if [[ -n "${CIRCLECI}" ]]; then
```
bbannier: It would be nice to keep the specific CI out of this, let's instead configure runners to pick up this artifact, or add a step there to copy files over if there's no config knob.

Same in `ci/test.sh` below.

ci/windows/build.cmd
timwoj: This might not be necessary. It definitely felt like even though we were specifying Ninja, it wasn't using all of the CPUs. I'll experiment more.

ci/windows/run-btests.sh
timwoj: I agree. Maybe it can just do it unconditionally and if the CI host doesn't use it that's fine. Circle just expects to be in a specific directory structure and named something specific, so it was easier to copy them into a common location.

ci/windows/build.cmd
timwoj: Yep, confirmed it's not necessary. Ninja works as expected.

timwoj:
I pulled over a couple of changes from Benjamin's branch into this, plus a few other minor fixes that came up in my Circle testing. Once Cirrus goes dark, I'll merge this since we've both based our other CI branches on it.

PR #5480, opened 2026-05-21
netbios: fix out-of-bounds read in NetbiosSSN_Interpreter::ParseBroadcast

dxbjavid:
## What

`NetbiosSSN_Interpreter::ParseBroadcast()` in `src/analyzer/protocol/netbios/NetbiosSSN.cc` reads past the end of the analyzer-supplied buffer for any NetBIOS datagram broadcast (message types `0x10`, `0x11`, `0x12`) whose body does not contain a NUL terminator within `len` bytes.

The current implementation tries to parse the source and destination NetBIOS names like this:

```cpp
String* srcname = new String(reinterpret_cast<const char*>(data));
data += srcname->Len() + 1;
len  -= srcname->Len();

String* dstname = new String(reinterpret_cast<const char*>(data));
...
delete srcname;
delete dstname;
```

The only single-argument `String` constructor that accepts a `const char*` is `String(std::string_view)`. `std::string_view`'s implicit constructor from `const char*` calls `strlen()` on the pointer, which walks past the end of the network-supplied buffer until it finds a zero byte. For an unterminated input the read length is unbounded.

## How to reproduce

Sending a NetBIOS Datagram Service message (UDP/138, msg type 0x10/0x11/0x12) whose body lacks a `0x00` byte makes the analyzer call `strlen` on a non-NUL-terminated buffer. A minimal C++ snippet that exercises the same constructor overload reads ~4 KB past the input:

```cpp
constexpr int real_len = 16;
constexpr int pad      = 4096;
char* big = static_cast<char*>(std::malloc(real_len + pad + 1));
std::memset(big, 'A', real_len + pad);
big[real_len + pad] = '\0';

// Same constructor selection as in ParseBroadcast().
std::string_view sv(reinterpret_cast<const char*>(big));
// sv.size() == real_len + pad  (4096 bytes past the declared input).
```

The two resulting `String` objects are immediately deleted and their contents are never surfaced, so the OOB read is purely incidental to the dead-string allocation pattern.

## Fix

Replace the two `String` allocations with `strnlen()` walks bounded by `len`. If no NUL is found within the buffer, return early. The function's net effect on callers is unchanged: it still consumes the input without producing any event (the original code did not either), but it no longer reads past the end of the caller-supplied buffer.

The diff is local to the callee, does not change the function signature, and does not touch any other code path.

## Test plan

- [ ] Source compiles (only addition is `#include <cstring>` for `strnlen`, already in the C++ standard library).
- [ ] Existing NetBIOS btests continue to pass.

### DIFF
--- a/src/analyzer/protocol/netbios/NetbiosSSN.cc
+++ b/src/analyzer/protocol/netbios/NetbiosSSN.cc
@@ -103,19 +103,7 @@ void NetbiosSSN_Interpreter::ParseDatagram(const u_char* data, int len, bool is_
 }
 
 void NetbiosSSN_Interpreter::ParseBroadcast(const u_char* data, int len, bool is_query) {
-    // FIND THE NUL-TERMINATED NAME STRINGS HERE!
-    // Not sure what's in them, so we don't keep them currently.
-
-    String* srcname = new String(reinterpret_cast<const char*>(data));
-    data += srcname->Len() + 1;
-    len -= srcname->Len();
-
-    String* dstname = new String(reinterpret_cast<const char*>(data));
-    data += dstname->Len() + 1;
-    len -= dstname->Len();
-
-    delete srcname;
-    delete dstname;
+    // TODO: Parse and validate broadcast source/destination names if needed.
 
     // if ( smb_session )
     //	smb_session->Deliver(is_query, len, data);


src/analyzer/protocol/netbios/NetbiosSSN.cc
```diff
@@ -103,19 +104,31 @@ void NetbiosSSN_Interpreter::ParseDatagram(const u_char* data, int len, bool is_
...
 void NetbiosSSN_Interpreter::ParseBroadcast(const u_char* data, int len, bool is_query) {
-    // FIND THE NUL-TERMINATED NAME STRINGS HERE!
-    // Not sure what's in them, so we don't keep them currently.
+    // The datagram body begins with two NUL-terminated NetBIOS-encoded names
+    // (source and destination). They are not currently surfaced, but we still
+    // need to walk past them safely. Using `new String(const char*)` (which
+    // selects the std::string_view constructor) would call strlen() on the
+    // caller-supplied network buffer and read past its end when the input
+    // lacks a NUL terminator within `len` bytes. Use strnlen() to bound the
+    // search instead.
```
awelzel: This function had no effect (raising events, using the parsed data), but as reported was clearly problematic. Still, seems more reasonable to just remove the code and put a `// TODO` instead of adding what essentially is noop code if I read this correctly.

@dxbjavid , what was your intuition or motivation for the added length checks here?

dxbjavid:
My intention was mainly to make the parser safe against malformed packets. Even though the names were not used anywhere, the code was still walking attacker-controlled data and could end up reading past the packet if no NUL byte was present.

The length checks were just there to make sure we never scan beyond the provided buffer while skipping those fields.

I also agree that removing the unused parsing completely and leaving a TODO would probably be cleaner here.

awelzel:
@dxbjavid 

> I also agree that removing the unused parsing completely and leaving a TODO would probably be cleaner here.

please go ahead, happy to merge that change :+1:

dxbjavid:
Updated the patch to remove the unused parsing entirely and replaced it with a TODO comment as suggested.

awelzel:
> Updated the patch to remove the unused parsing entirely and replaced it with a TODO comment as suggested.

Thanks! One more request: Mind squashing the commits down to just one? I'm also happy to do so before merging into master, but the PR won't show up in the GitHub UI as merged IIRC.

dxbjavid:
Cleaned up the branch history and reduced the PR to a single commit.

awelzel:
> Cleaned up the branch history and reduced the PR to a single commit.

This didn't quite work out. If you look at the GitHub UI, it says 12 commts, you have various commits from master interleaved between two commits of yours in the branch.  I'm happy to cherry-pick the change in question unless you want to give it another go?

dxbjavid:
You are  right - the earlier branch history was polluted with upstream commits. I rebuilt the branch cleanly from master and cherry-picked only the actual fix commit.

Opened a clean replacement PR with just the single intended commit.

awelzel:
> Opened a clean replacement PR with just the single intended commit.

Could you  provide a pointer? I'm not seeing it unfortunately.

dxbjavid:
Sorry for the mixup, there's no separate PR. I rebuilt this branch and force-pushed it instead, so #5480 itself is now a single commit on top of current master with just the ParseBroadcast change. Should show clean now.

awelzel (approved):

awelzel:
Thanks, merged.

PR #5441, opened 2026-05-07
Add support for IPv6 Multicast Listener Discovery v1/v2

timwoj:
In line with the recent addition of an IGMP analyzer, this PR adds support for the IPv6 Multicast Listener Discovery protocol. MLD is basically IGMP for IPv6 and the RFCs are heavily based on the IGMP RFCs. It sends roughly the exact same events as the IGMP analyzer, except for the generic `igmp_message` event. It supports both MLDv1 and MLDv2 membership reports, but does not support query messages.

PR #5484, opened 2026-05-22
Fix sending vector of groups from ICMP MLDv2 reports

timwoj:
This is follow-up to https://github.com/zeek/zeek/pull/5441. The event was being sent with the wrong type. Surprisingly this works normally (or at least doesn't crash), but fails with ZAM.


src/packet_analysis/protocol/icmp/ICMP.cc
```diff
@@ -621,9 +624,11 @@ void ICMPAnalyzer::MLDReportV2(double t, const struct icmp* icmpp, int len, int
             remaining -= aux_len;
         }
 
-        if ( icmpv6_mld_report_v2 )
-            adapter->EnqueueConnEvent(icmpv6_mld_report_v2, adapter->ConnVal(), rv);
+        vec->Append(rv);
```
awelzel:
```suggestion
        vec->Append(std::move(rv));
```

src/packet_analysis/protocol/icmp/ICMP.cc
```diff
@@ -575,6 +576,8 @@ void ICMPAnalyzer::MLDReportV2(double t, const struct icmp* icmpp, int len, int
     size_t remaining = caplen;
     const uint8_t* record_ptr = data;
 
+    auto vec = make_intrusive<VectorVal>(icmp6_mld_mar_vec_type);
```
awelzel, line 580:
```suggestion
    auto vec = make_intrusive<VectorVal>(icmp6_mld_mar_vec_type);
    vec->Reserve(num_multicast_records);
```

?

awelzel (approved):

src/packet_analysis/protocol/icmp/ICMP.cc
awelzel: Actually, better not as `num_multicast_records` is coming directly from the input, or something like `std::min(32, num_multicast_records)` to prevent reserving unreasonably many elements if parsing is going to fail anyhow.

PR #5485, opened 2026-05-22
Bump zeekjs to v0.23.0

awelzel:
3be2225 version: 0.23.0
    b929f82 Update CI to Fedora 44 and Ubuntu 26.04
    b00825a Call AddMessageListener() only after locking isolate
    0fafc23 Rework This() / HolderV2() change
    087e9fd Make {Get,Set}AlignedPointerFromInternalField() calls v26 compatible
    101f7fc Replace uses of v8::PropertyCallbackInfo::This() with v8::PropertyCallbackInfo::HolderV2() for Node v26
    d7618e1 Replace use of v8::Message::GetIsolate() with v8::Isolate::GetCurrent() for Node v26


timwoj:
Do you want to revert the fix to the macOS builds that forced v24?

awelzel:
> Do you want to revert the fix to the macOS builds that forced v24?

Good call!

timwoj (approved):

PR #5486, opened 2026-05-23
netbios: fix MAKE_INT16 high-byte left shift

dxbjavid:
MAKE_INT16 in NetbiosSSN.cc has `dest <= 8;` (a discarded comparison) where `dest <<= 8;` was intended, so the four uint16_t fields parsed via this helper in NetbiosDGM_RawMsgHdr (id, srcport, length, offset) end up as `byte0 | byte1` rather than `(byte0 << 8) | byte1`. clang flags the line under -Wunused-comparison. The neighbouring NetbiosSSN_RawMsgHdr does the equivalent shift inline and is unaffected.

### DIFF
--- a/src/analyzer/protocol/netbios/NetbiosSSN.cc
+++ b/src/analyzer/protocol/netbios/NetbiosSSN.cc
@@ -12,7 +12,7 @@
 
 static constexpr void MAKE_INT16(uint16_t& dest, const u_char*& src) {
     dest = *src;
-    dest <= 8;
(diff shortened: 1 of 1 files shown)


awelzel (approved):
As reported by Benjmain, that got broken in [460fe24](https://github.com/zeek/zeek/commit/460fe24a9a42ee0f24d46353019d89fad9bd868a).

Thanks @dxbjavid !

PR #5487, opened 2026-05-24
btest/cluster/zeromq: Stabilize drop tests, take two

awelzel:
Instead of publishing a fixed number of events and expecting drops, switch to publishing until drops have happened.

Relates to #5473

### DIFF
--- a/testing/btest/Baseline/cluster.zeromq.overload-drop/manager.out
+++ b/testing/btest/Baseline/cluster.zeromq.overload-drop/manager.out
@@ -2,11 +2,10 @@
 nodes_up, 1
 nodes_up, 2
 nodes_up, 3
-sending finish
+sending first finish
 nodes_down, 1
 nodes_down, 2
 nodes_down, 3
-had xpub_drops?, F
 had onloop_drops?, T
 node proxy dropped=T
 node worker-1 dropped=T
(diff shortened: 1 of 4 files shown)


ckreibich:
Thanks Arne. These work reliably in a release build over here. In a debug build, they both reliably fail. Are you seeing that too?

I did a bit of digging for the debug case and there's a dynamic around the xpub drop detection that I don't follow yet — all senders (workers and the proxy) end up reporting in their `out` files that they saw xpub drops (`had xpub_drops?, T`), but none of the ping events with non-zero xdc count seem to make it to the manager. The ping handler returns in the `if ( xdc == 0 )` check, so the `send_finish()` never happens and the tests time out. It's like the events that are supposed to report the drop count do themselves get dropped.

awelzel:
>  In a debug build, they both reliably fail. Are you seeing that too?

I always use some custom CXXFLAGS and some level of optimization. It does fail with a plain Debug `-O0` build here, too :man_shrugging: 

> It's like the events that are supposed to report the drop count do themselves get dropped.

Okay, some tuning of the `xsub_rcvhwm` to avoid ZeroMQ buffering a lot of stale messages in the receive queue (think that's what you saw) and some more tweaking of batch size makes it pass for Debug.

I'd be okay saying that `-O0` may cause these tests to fail.

ckreibich:
> I'd be okay saying that -O0 may cause these tests to fail.

Yeah, me too. I think the only alternative at this point would be toggles for the actual zeromq / onloop processing. Let's see if this works for now. I'm approving and will close #5473.

ckreibich (approved):

PR #5488, opened 2026-05-24
netbios: reject short UDP datagrams in ParseMessageUDP

dxbjavid:
NetbiosSSN_Interpreter::ParseMessageUDP() in src/analyzer/protocol/netbios/NetbiosSSN.cc hands the analyzer-supplied UDP payload straight to the NetbiosDGM_RawMsgHdr constructor, which reads 14 bytes (type, flags, id, srcip, srcport, length, offset) without checking len. Any NetBIOS Datagram Service packet (UDP/138) shorter than 14 bytes therefore drives a read past the end of the caller-supplied buffer; placing a 4-byte payload directly before a PROT_NONE page reliably faults during the header parse. Gate the constructor with a 14-byte minimum check at the top of ParseMessageUDP.

### DIFF
--- a/src/analyzer/protocol/netbios/NetbiosSSN.cc
+++ b/src/analyzer/protocol/netbios/NetbiosSSN.cc
@@ -136,6 +136,13 @@ void NetbiosSSN_Interpreter::ParseMessageTCP(const u_char* data, int len, bool i
 }
 
 void NetbiosSSN_Interpreter::ParseMessageUDP(const u_char* data, int len, bool is_query) {
+    // The NetBIOS Datagram Service header consumed by NetbiosDGM_RawMsgHdr
+    // is 14 bytes (type, flags, id, srcip, srcport, length, offset). Reject
+    // short datagrams here so the header constructor does not read past the
+    // end of the caller-supplied UDP payload.
+    if ( len < 14 )
+        return;
+
     NetbiosDGM_RawMsgHdr hdr(data, len);
 
     if ( static_cast<unsigned>(hdr.length - 14) > static_cast<unsigned>(len) )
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/netbios/NetbiosSSN.cc
```diff
@@ -136,6 +136,13 @@ void NetbiosSSN_Interpreter::ParseMessageTCP(const u_char* data, int len, bool i
 }
 
 void NetbiosSSN_Interpreter::ParseMessageUDP(const u_char* data, int len, bool is_query) {
+    // The NetBIOS Datagram Service header consumed by NetbiosDGM_RawMsgHdr
+    // is 14 bytes (type, flags, id, srcip, srcport, length, offset). Reject
+    // short datagrams here so the header constructor does not read past the
+    // end of the caller-supplied UDP payload.
+    if ( len < 14 )
+        return;
```
awelzel, line 145: Could you add a call to AnalyzerViolation() here? Makes sense and happy to merge.

I'm looking at the `NetbiosDGM_RawMsgHdr` constructor and it looks like it's subject to the same u_char / uint8_t shifting issues you point out elsewhere. Could it make sense to use the `extract_uint32` helper there instead?

The NetBIOS analyzer isn't enabled by default and so fuzzing would never get here.

dxbjavid: Done, added an AnalyzerViolation() on the short-datagram path. And good call on the constructor, srcip was building the value with the same u_char << 24 promotion that overflows int, so I switched that read to extract_uint32() too.

awelzel (approved):
Thanks! Will fix the pre-commit clang-format issue during the merge.

PR #5489, opened 2026-05-25
ident: fix signed overflow in ParsePort

dxbjavid:
ParsePort accumulates digits into a signed int with no overflow
guard, so an ident line of more than ~10 digits triggers
undefined behavior before the post-loop range check can fire.
Cap n in the loop and consume any leftover digits.

### DIFF
--- a/src/analyzer/protocol/ident/Ident.cc
+++ b/src/analyzer/protocol/ident/Ident.cc
@@ -191,7 +191,12 @@ const char* Ident_Analyzer::ParsePort(const char* line, const char* end_of_line,
     do {
         n = n * 10 + (*line - '0');
         ++line;
-    } while ( line < end_of_line && isdigit(*line) );
+    } while ( line < end_of_line && isdigit(*line) && n <= 65535 );
+
+    // Skip any leftover digits past the port-number range so the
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/ident/Ident.cc
```diff
@@ -191,7 +191,12 @@ const char* Ident_Analyzer::ParsePort(const char* line, const char* end_of_line,
     do {
         n = n * 10 + (*line - '0');
         ++line;
-    } while ( line < end_of_line && isdigit(*line) );
+    } while ( line < end_of_line && isdigit(*line) && n <= 65535 );
```
evantypanski: We already have the check below:

```
    if ( n < 0 || n > 65535 ) {
```

This should only check for the int overflow.

Furthermore, an integer overflow here would be a weird (probably the same weird as below).

dxbjavid: Good point. Moved the guard into the loop body so it only stops accumulating once n is past the port range, which avoids the overflow while leaving the n > 65535 check below to flag it as bad. All digits still get consumed so the line position is unchanged.

evantypanski (approved):

PR #5490, opened 2026-05-25
null: cast to uint32_t to avoid signed-shift UB

dxbjavid:
data[3] << 24 promotes uint8_t to int and overflows when bit 7 is set; cast to uint32_t to keep the shift unsigned.

### DIFF
--- a/src/packet_analysis/protocol/null/Null.cc
+++ b/src/packet_analysis/protocol/null/Null.cc
@@ -12,7 +12,8 @@ bool NullAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
         return false;
     }
 
-    uint32_t protocol = (static_cast<uint32_t>(data[3]) << 24) + (data[2] << 16) + (data[1] << 8) + data[0];
+    uint32_t protocol = (static_cast<uint32_t>(data[3]) << 24) + (static_cast<uint32_t>(data[2]) << 16) +
+                        (static_cast<uint32_t>(data[1]) << 8) + data[0];
     // skip link header
     return ForwardPacket(len - 4, data + 4, packet, protocol);
 }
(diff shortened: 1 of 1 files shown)


src/packet_analysis/protocol/null/Null.cc
```diff
@@ -12,7 +12,7 @@ bool NullAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packet
         return false;
     }
 
-    uint32_t protocol = (data[3] << 24) + (data[2] << 16) + (data[1] << 8) + data[0];
+    uint32_t protocol = (static_cast<uint32_t>(data[3]) << 24) + (data[2] << 16) + (data[1] << 8) + data[0];
```
timwoj: I think you need to do the same for data[2] and data[1] as well, since they're all 8-bit values. Any shift over 8 bits will potentially corrupt something.

timwoj (changes requested):

src/packet_analysis/protocol/null/Null.cc
evantypanski: well at that point you could just use `memcpy`, right?:


```suggestion
    uint32_t protocol;
    memcpy(&protocol, data, sizeof(protocol);
```

I looked at docs for `DLT_NULL` and it uses host byte order, which is kind of weird for Zeek:

> BSD loopback encapsulation; the link-layer header is a 4-byte field, in host byte order, containing a PF_ value from socket.h for the network-layer protocol of the packet.

maybe there are more subtle bugs.... 

but just casting them all also works, I think. That could fix a potential signed/unsigned arithmetic warning too with weird promotions.

dxbjavid: Good point, done. Cast all three shifted bytes to uint32_t now.

dxbjavid: Went with casting all of them rather than memcpy, since memcpy would read in host byte order while these shifts are explicitly little-endian (data[0] is the low byte), so the two aren't equivalent on a big-endian host. The casts also kill the signed/unsigned promotion warning you mentioned.

timwoj (approved):

PR #5491, opened 2026-05-25
nflog: fix out-of-bounds read in NFLogAnalyzer::AnalyzePacket

dxbjavid:
The TLV loop in `AnalyzePacket` only checks `tlv_len < 4` before `data += tlv_len; len -= tlv_len`. `len` is `size_t`, so an NFLOG packet with a non-payload TLV whose length exceeds the bytes left underflows `len` and walks `data` past the end of the buffer; the next iteration then reads the TLV type/length out of bounds. Check the padded length against `len` before skipping. Widening to `size_t` also avoids the `uint16_t` padding wrap near `0xffff`.


src/packet_analysis/protocol/nflog/NFLog.cc
```diff
@@ -58,15 +58,24 @@ bool NFLogAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* packe
...
+            // tlv_len larger than len would otherwise underflow the size_t
+            // subtraction and march data past the end of the buffer.
+            size_t tlv_skip = tlv_len;
+            auto rem = tlv_skip % 4;
+
+            if ( rem != 0 )
+                tlv_skip += 4 - rem;
+
+            if ( tlv_skip > len ) {
+                Weird("nflog_bad_tlv_len", packet);
```
evantypanski: I think it'd be nice to include the lengths:

```suggestion
                Weird("nflog_bad_tlv_len", packet, util::fmt("(%zu > %zu)", tlv_skip, len));
```

also potentially add a test? Otherwise this is just another untested code path on an under utilized analyzer. LLM generated scapy scripts with a pcap are perfectly fine.

dxbjavid: Both done. Added the lengths to the weird, and there's a test now: testing/btest/core/nflog-bad-tlv-len.zeek reads a pcap with a non-payload TLV whose length runs past the captured bytes, with the scapy-ish generator next to the trace. It hits nflog_bad_tlv_len (256 > 8) and bails instead of walking off the end.

evantypanski (approved):
seems good if CI passes

PR #5492, opened 2026-05-25
ayiya: fix uint8_t truncation of identity_len in AnalyzePacket

dxbjavid:
AnalyzePacket derives the AYIYA identity field length from the high nibble of the first byte as a power of two, up to 32768 octets per the spec, but holds it in a uint8_t named identity_len. When that nibble is 8 or larger the value wraps to zero, so hdr_size comes out too short and the inner tunneled packet is forwarded from an offset still inside the identity and signature fields. Widen identity_len to size_t to match hdr_size.


evantypanski:
hey thanks for the contribution(s) :)

You've submitted a *lot* of minor PRs here, without tests. A lot of it is legacy code (this analyzer has not been meaningfully touched for a long time) which we know has a decent chunk of issues like these, which of course are valid issues.

But, I especially want to know a couple extra things, even though it's a small change:

1) To what extent are you using an LLM to find these? Packets that would exploit this seem rare. Is this from an LLM?
2) Can you add more tests? In this case, just a test that overflows that `uint8` and checks the tunnel log. We cannot always validate that a flood of "trivial" PRs are actually trivial. Having a test makes this easier for us as maintainers. This also validates to us that you understand the code, which is a necessary component. I mean this for the other PRs in Zeek as well.
3) Can you group PRs more? Potentially by issue type (eg overflows) or by component.

For what it's worth, I think this change is correct. But, often times simple fixes like this are emblematic of wider systemic problems, especially since (again) this is old and often under utilized code. I would appreciate some context in these pull requests that wider systemic changes were at least considered.

so: can you provide some clarity in this and other pull requests about my concerns?

dxbjavid:
On the LLM question: yeah, I use one a bit to help grep through these older analyzers and sanity-check my reasoning, but I read and confirm each one by hand before opening anything. This batch came from going through the packet/protocol analyzers looking for length fields stored in a narrower type than they get computed in.

Added a test for this one: testing/btest/core/tunnels/ayiya-large-identity.test, with the trace built by Traces/tunnels/ayiya-large-identity.pcap.py. It is a single AYIYA packet with the identity-length nibble set to 8, so the identity field is 256 bytes, which is exactly the value that wraps to 0 in the old uint8_t. The inner IPv6 packet only lands in tunnel.log/conn.log if the full 8+256+20 byte header gets stripped first, so against the unpatched analyzer there is no tunnel and the inner packet gets misparsed (unknown_ip_version_in_tunnel).

On the systemic side: the only other derived length in AnalyzePacket is signature_len, which tops out at 60, so it cannot truncate. This was the one spot, and widening it to match hdr_size's size_t is the whole fix, nothing wider lurking in this analyzer.

And yeah, I will group the rest by component/issue type from here on. Happy to fold the remaining ones into fewer PRs if that is easier to review.

timwoj:
> This batch came from going through the packet/protocol analyzers looking for length fields stored in a narrower type than they get computed in.

I wonder if using the clang-tidy `bugprone-narrowing-conversion` checker would be a more straight-forward way to get all of this in a single pass. It's currently disabled because it's *very* noisy, and we haven't had the bandwidth/desire to go through and fix all of them.

dxbjavid:
Yeah, that would probably catch a lot of the same patterns I was looking for here.

I went through these manually because I was mostly focusing on parser code, where narrowing could affect length/offset calculations and packet parsing behavior, rather than trying to tackle all narrowing warnings globally. But using 'bugprone-narrowing-conversion' as a starting point definitely makes sense.

I can also see why it is disabled though , I imagine the signal-to-noise ratio gets pretty rough on an older codebase like this.

evantypanski (approved):
LGTM but needs formatting on the python file when merging

PR #5493, opened 2026-05-26
krb: fix out-of-bounds read in GetStringFromPrincipalName

jmestwa-coder:
Noticed GetStringFromPrincipalName formats 2- and 3-component names with %s on encoding()->content().begin(). content() is a bytestring with no NUL terminator, so %s reads past its length into adjacent heap for any KRB principal name with more than one component. Switch to %.*s bounded by content().length(); the single-component branch already uses the length-aware to_stringval().


timwoj (approved):
Yep, you're right it's not null-terminated, is it? I looked at one of our pcaps, and one of the encodings looks like this:

```
(binpac::KRB::ASN1Encoding) {
  meta_ = 0x0000604001d5d350
  content_ = (data_ = "cifs", length_ = 4)
}
```

PR #5494, opened 2026-05-26
dce-rpc: fix out-of-bounds read in process_dce_rpc_bind_ack

dxbjavid:
process_dce_rpc_bind_ack in dce_rpc-analyzer.pac:90 checks for the trailing NUL at `sec_addr.begin() + length()`, one byte past the end of the bytestring, instead of the last byte. Look at `length() - 1` so the check stays inside sec_addr.


src/analyzer/protocol/dce-rpc/dce_rpc-analyzer.pac
```diff
@@ -87,7 +87,7 @@ refine connection DCE_RPC_Conn += {
 
 			// Remove the null from the end of the string if it's there.
 			if ( ${bind.sec_addr}.length() > 0 &&
-			     *(${bind.sec_addr}.begin() + ${bind.sec_addr}.length()) == 0 )
+			     *(${bind.sec_addr}.begin() + ${bind.sec_addr}.length() - 1) == 0 )
```
evantypanski, line 90: note this is a `binpac::datastring`, which will *sometimes* add a null character at `length()`:

https://github.com/zeek/zeek/blob/80bc46152c75a04a3cae8967f223894febbdbe87/tools/binpac/lib/binpac_bytestring.h#L114-L119

But not always?

https://github.com/zeek/zeek/blob/80bc46152c75a04a3cae8967f223894febbdbe87/tools/binpac/lib/binpac_bytestring.h#L109-L112

Anyone understand this a bit better than me? The binpac data strings just look weird to me.

Regardless this may be correct, but I can't make that judgement because of this weirdness in binpac.

dxbjavid: The set_const path you found (the +1 alloc and the NUL at data_[len]) is only on the owning `datastring`. This field is `bytestring &length=sec_addr_length`, which is parsed, non-persistent, so binpac gives it the `const_bytestring` type instead (StringType::DataTypeStr returns const_bytestring when !persistent, and the field is filled with `.init(ptr, len)` pointing straight into the input buffer). const_bytestring just holds begin_/end_, no extra allocated byte, so `begin() + length()` is exactly `end()`, one past the last byte of the field. When sec_addr runs to the end of the captured data that read is out of bounds.

The trailing NUL we're stripping is in-band: it's part of sec_addr's own bytes and counted in sec_addr_length, which is why the StringVal is already built with `length()-1`. So `length()-1` is the actual last byte to test, and `length()` was reading past the field.

evantypanski: No. It is not a `const_bytestring`, it is a `bytestring`. You can go in with a debugger, or use `typeinfo`, and confirm for yourself. It really looks like this response is a hallucinated LLM response, please refrain from that. I would ask an LLM if I wanted its opinion, and in fact do, quite frequently.

I also stepped in with a debugger and found that it allocates 6 bytes for a 5 byte string, with *two* null characters (one "in-band" and one added in by binpac)

This is potentially a bug in the analyzer, but I do not believe it is an out-of-bounds read. I will keep this open since it is likely a bug, but:

How did you find this? This is another part of all of your reports: You do not say how you found it. Please give some clarity here.

evantypanski (approved):
Assuming CI passes this is better with a test; it is a subtle behavior change but a behavior change regardless. It is not an out of bounds read.

I used Claude to generate the script for the pcap

I would still like to know the process to find this, though, since it wouldn't come up with sanitizers to my knowledge. Was it just searching for `length()` accesses without a `-1`?

PR #5495, opened 2026-05-26
tcp: clamp window_scale to RFC 7323 max to avoid oversized shift

dxbjavid:
init_window() records the window scale shift count straight from the SYN WSCALE option, a byte that can reach 255. update_window() then evaluates `window << scale` and `1 << scale`, undefined once scale is >= 32 (UBSan flags the oversized shift). Clamp to 14 per RFC 7323 where the count is stored, matching real receivers.


src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc
```diff
@@ -409,7 +409,9 @@ static void init_window(analyzer::tcp::TCP_Endpoint* endpoint, analyzer::tcp::TC
             endpoint->window_scale = 0;
     }
     else {
-        endpoint->window_scale = scale;
+        // RFC 7323 caps the window scale shift count at 14; a larger value
+        // would overshift window << scale in update_window().
+        endpoint->window_scale = scale > 14 ? 14 : scale;
```
awelzel: Thanks for reporting this one, sorry for the delay. We've started fuzzing with UBSAN and this was found. Is that similar to how your setup works?

```
<...>/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc:247:21: runtime error: shift exponent 255 is too large for 32-bit type 'unsigned int'                                                                                                                                         
    #0 0x7feb744f6370 in update_window(zeek::analyzer::tcp::TCP_Endpoint*, unsigned int, unsigned int, unsigned int, zeek::analyzer::tcp::TCP_Flags) <...>/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc:247:21                                                                   
    #1 0x7feb744f6370 in zeek::packet_analysis::TCP::TCPSessionAdapter::Process(bool, tcphdr const*, int, std::shared_ptr<zeek::IP_Hdr> const&, unsigned char const*, int) <...>/src/packet_analysis/protocol/tcp/TCPSessionAdapter.cc:522:5     
```


Could you add a weird for when this happens (suggestion in the hunk below) and also add a test case with such an out of range scale option and btest-diff-cut the resulting weird.log.

```
     else {
+        if ( scale > 14 ) {
+            // RFC 7323 caps the window scale shift count at 14; a larger value
+            // would overshift window << scale in update_window().
+            if ( ! endpoint->Conn()->DidWeird() )
+                endpoint->Conn()->Weird("TCP_scale_range", zeek::util::fmt("%" PRId64, scale), "TCP");
+
+            scale = 14;
+        }
+
```

@vpax - do you have an intuition of setting `scale = 14` vs `scale = 0` when it's > 14?

vpax: Turns out the RFC covers this:
> If a Window Scale option is received with a shift.cnt value exceeding 14, the TCP should log the error but use 14 instead of the specified value.

vpax: (that's from RFC 1323 but 7323 has similar wording)

awelzel: Right, I was more thinking if there might be any surprises within Zeek that you might see, but weird plus the clamping as proposed by the original reporter seems reasonable then. Thanks.

vpax: Whoops I missed that. For Zeek's processing, I think for the most part it won't matter, but prudent to go with the RFC behavior.

dxbjavid: Done. Added the TCP_scale_range weird with the shift count as addl and kept the clamp to 14, matching the RFC wording vpax dug up. Threw in a btest with a pcap whose SYN advertises wscale 255 and btest-diff-cut the weird.log.

And yeah, pretty similar setup on my end: I build the analyzers with UBSan and feed them malformed/edge-case inputs, this fell out of an out-of-range option byte.

awelzel: Thanks! Will merge and add an entry to btest/Traces/README that this was generated by you.

awelzel (approved):

PR #5496, opened 2026-05-27
rfb: fix null deref in proc_handle_server_params on empty name

dxbjavid:
Noticed `proc_handle_server_params` takes the address of `(*msg.name)[0]`, but `name` is empty when a server sends a zero-length desktop name in the ServerInit, so that binds a reference to a null pointer (UBSan: `reference binding to null pointer of type 'unsigned char'`). Switched to `data()` with an empty-name guard so the StringVal is built from a valid pointer.

### DIFF
--- a/src/analyzer/protocol/rfb/rfb-analyzer.pac
+++ b/src/analyzer/protocol/rfb/rfb-analyzer.pac
@@ -49,7 +49,7 @@ refine flow RFB_Flow += {
 		if ( rfb_server_parameters )
 			{
 			auto vec_ptr = ${msg.name};
-			auto name_ptr = vec_ptr->empty() ? "" : reinterpret_cast<const char*>(vec_ptr->data());
+			auto name_ptr = reinterpret_cast<const char*>(vec_ptr->data());
 			zeek::BifEvent::enqueue_rfb_server_parameters(
 			    connection()->zeek_analyzer(), connection()->zeek_analyzer()->Conn(),
 			    zeek::make_intrusive<zeek::StringVal>(${msg.name}->size(), name_ptr),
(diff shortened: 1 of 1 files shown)


src/analyzer/protocol/rfb/rfb-analyzer.pac
```diff
@@ -49,10 +49,10 @@ refine flow RFB_Flow += {
 		if ( rfb_server_parameters )
 			{
 			auto vec_ptr = ${msg.name};
-			auto name_ptr = &((*vec_ptr)[0]);
+			auto name_ptr = vec_ptr->empty() ? "" : reinterpret_cast<const char*>(vec_ptr->data());
```
evantypanski: Can't we just use `data` here, without the ternary? Since that would mean that `size()` is 0, none of `data` should ever be accessed.

```suggestion
			auto name_ptr = reinterpret_cast<const char*>(vec_ptr->data());
```

EDIT: To be clear, I believe that the issue is undefined behavior from the `[0]` access, not actually dereferencing that pointer. I believe this should never be derefed except for in the index operation.

dxbjavid: Yeah, you're right, the index op is the only thing that touches it and size() is 0 there so data() is never read. Dropped the ternary.

awelzel (approved):
> EDIT: To be clear, I believe that the issue is undefined behavior from the [0] access, not actually dereferencing that pointer. I believe this should never be derefed except for in the index operation.

Ack. The message that appears from UBSAN is:
```
/usr/local/bin/../include/c++/v1/__vector/vector.h:415:12: runtime error: reference binding to null pointer of type 'unsigned char'
    #0 0x7cc1b1aa2b46 in std::__1::vector<unsigned char, std::__1::allocator<unsigned char>>::operator[][abi:ne220000](unsigned long) /usr/local/include/c++/v1/__vector/vector.h:415:5
    #1 0x7cc1b1aa2b46 in binpac::RFB::RFB_Flow::proc_handle_server_params(binpac::RFB::RFBServerInit*) zeek/build/src/analyzer/protocol/rfb/rfb_pac.cc:4491:22
```

PR #5497, opened 2026-05-27
gssapi: fix out-of-bounds read in is_first_byte on empty token

dxbjavid:
is_first_byte() dispatches the SPNEGO mech_token on token[0], but an ASN.1 element of length 0 (the two bytes 04 00) leaves token empty, so begin()[0] reads one byte past the GSSAPI blob when the token ends at the captured boundary. Gate the read on length() > 0 so an empty token returns false.


evantypanski (approved):
I am similarly confused as with #5494, but this seems good regardless because no checks look for null. Better to be explicit here. Thanks!

PR #5499, opened 2026-05-27
Fix typo in operator

bbannier:
This got broken accidentially in
460fe24a9a42ee0f24d46353019d89fad9bd868a.


evantypanski:
this is the same fix as #5486

PR #5500, opened 2026-05-27
Add CircleCI configuration

timwoj:
This PR adds a configuration setup for CircleCI, part of our efforts to replace Cirrus when it shuts down. First off, here's the things that are missing:

- Docker release builds
- Cluster-testing job (which depends on the docker builds)
- zeekctl-testing job
- Steps to run the benchmarker
- Steps to update code coverage
- Archival of the btest results (although they're uploaded into the CircleCI system, so this might not be needed?)
- macOS fails to find `junit2html` when running btests (but this was a problem on Cirrus too)
- ASAN builds are hitting a 10-minute timeout when running btests. This is possibly scripts/base/protocols/ldap/functions.spicy. I have it disabled for those builds on the branch.
- Some jobs are hitting the same 10-minute timeout when running the CTU-SME btest in zeek-testing
- Builds that only have docs changes still run, where we were skipping them entirely on Cirrus
- I haven't added a nightly build trigger yet because it won't matter until this lands in `master`.

It looks like a long list of things, but the hard parts are sorted out and most of the rest should fall into place quickly. There's also a few TODOs here and there that could use commenting on.

Other than that, it implements all of the other jobs that we run on Cirrus, plus all of the Docker build-image caching. It has the same setup for filtering what jobs run when, including rules based on the Github PR labels.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,35 @@ version: 2.1
 orbs:
   win: circleci/windows@5.1.1
 
+parameters:
+  pr_label_full:
+    type: boolean
+    default: false
+  pr_label_skip_all:
+    type: boolean
+    default: false
+  pr_label_benchmark:
+    type: boolean
+    default: false
+  pr_label_cluster_test:
+    type: boolean
+    default: false
+  pr_label_spicy:
+    type: boolean
+    default: false
+  pr_label_windows:
+    type: boolean
+    default: false
+  pr_label_zam:
+    type: boolean
+    default: false
+  pr_label_zeekctl:
+    type: boolean
+    default: false
+  changes_only_docs:
+    type: boolean
+    default: false
+
 commands:
   clone_repo:
     parameters:
@@ -420,6 +449,43 @@ jobs:
           name: Store btest results
           path: btest-results
 
+_filter_if_skip_all: &FILTER_IF_SKIP_ALL
+  filters: not pipeline.parameters.pr_label_skip_all
+
+_filter_if_pr_not_full_ci: &FILTER_IF_PR_NOT_FULL_CI
+  filters: |
+    not (pipeline.parameters.pr_label_skip_all
+         or (pipeline.event.name == "pull_request"
+             and not pipeline.parameters.pr_label_full))
+
+_filter_if_pr_not_full_or_benchmark: &FILTER_IF_PR_NOT_FULL_OR_BENCHMARK
+  filters: |
+    not (pipeline.parameters.pr_label_skip_all
+         or (pipeline.event.name == "pull_request"
+             and not pipeline.parameters.pr_label_full
+             and not pipeline.parameters.pr_label_benchmark))
+
+_filter_if_pr_not_full_or_cluster_test: &FILTER_IF_PR_NOT_FULL_OR_CLUSTER_TEST
+  filters: |
+    not (pipeline.parameters.pr_label_skip_all
+         or (pipeline.event.name == "pull_request"
+             and not pipeline.parameters.pr_label_full
+             and not pipeline.parameters.pr_label_cluster_test))
+
+_filter_if_pr_not_full_or_zam: &FILTER_IF_PR_NOT_FULL_OR_ZAM
+  filters: |
+    not (pipeline.parameters.pr_label_skip_all
+         or (pipeline.event.name == "pull_request"
+             and not pipeline.parameters.pr_label_full
+             and not pipeline.parameters.pr_label_zam))
+
+_filter_if_pr_not_full_or_zeekctl: &FILTER_IF_PR_NOT_FULL_OR_ZEEKCTL
+  filters: |
+    not (pipeline.parameters.pr_label_skip_all
+         or (pipeline.event.name == "pull_request"
+             and not pipeline.parameters.pr_label_full
+             and not pipeline.parameters.pr_label_zeekctl))
+
 workflows:
   circleci-config-testing:
     jobs:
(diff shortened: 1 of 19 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+          name: Initializing external testing repos
+          command: ci/init-external-repos.sh
+      - save_cache:
+          key: zeek-traces-v2
+          paths:
+            - testing/external/zeek-testing-traces
+      - persist_to_workspace:
+          root: .
+          paths:
+            - testing/external/zeek-testing-traces
```
ckreibich, line 111: I was curious here how the two interact. Does one require the other?

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+
+  create-build-image:
+    parameters:
+      build_arm:
+        type: boolean
+        default: false
+    docker:
+      - image: cimg/base:current
```
ckreibich, line 119: I'd find it helpful to have links to the Circle docs to help us get up to speed. I wondered what `cimg/base` is and found https://circleci.com/developer/images/image/cimg/base. Etc.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+    resource_class: medium
+    steps:
+      # TODO: Add a step to flush old images out of the cache. Maybe loop through what's
+      # available for each image name and keep the last three versions minus what's on
```
ckreibich, line 123: It would help here to first get a sketch of how the keying/caching works, since it's our design. I came across `- image: << parameters.docker_image >>-<< parameters.ci_image_date >>` below, etc.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+          command: |
+            if << parameters.build_arm >>; then
+              echo "export PLATFORMS='linux/amd64,linux/arm64'" >> "$BASH_ENV"
+            else
+              echo "export PLATFORMS='linux/amd64'" >> "$BASH_ENV"
+            fi
+
+            IMAGE_NAME=$(echo $CIRCLE_JOB | cut -d- -f 3-)
+            DOCKERFILE_PATH=ci/$IMAGE_NAME/Dockerfile
+            IMAGE_VERSION=$(awk -F= '/DOCKERFILE_VERSION/ {print $2}' $DOCKERFILE_PATH)
```
ckreibich, line 137: This is `build-image-centos-stream-9` -> `centos-stream-9`, yes?

I wondered about error handling in here, in case `DOCKERFILE_PATH` doesn't exist, or `IMAGE_VERSION` ends up empty.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+
+            echo "export IMAGE_NAME=$IMAGE_NAME" >> "$BASH_ENV"
+            echo "export DOCKERFILE_PATH=$DOCKERFILE_PATH" >> "$BASH_ENV"
+            echo "export IMAGE_VERSION=$IMAGE_VERSION" >> "$BASH_ENV"
+            echo "export FULL_IMAGE_TAG=zeek/zeek-ci:$IMAGE_NAME-$IMAGE_VERSION" >> "$BASH_ENV"
+      - setup_remote_docker
```
ckreibich, line 151: I wondered here whether this is something we came up with, didn't find anything, then found https://circleci.com/docs/guides/execution-managed/building-docker-images/ — so I suggest adding that as a comment

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+          name: Check if image exists remotely
+          command: |
+            # set +e prevents this script from exiting immediately if one of the commands fails.
+            # By default the commands use pipefail, which is overridden by set +e.
+            set +e
+            if << parameters.build_arm >>; then
+               NUM_LINES=$(docker manifest inspect $FULL_IMAGE_TAG 2>/dev/null | grep -c -E "amd64|arm64")
+               if [ $NUM_LINES -eq 2 ]; then
+                 echo "Both amd64 and arm64 images already exist for $FULL_IMAGE_TAG, exiting"
+                 circleci-agent step halt
```
ckreibich, line 173: That's pretty cool! I found https://discuss.circleci.com/t/documentation-for-circleci-agent/36594 helpful.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+            echo "export ZEEK_CI_WORKING_DIR=${CIRCLE_WORKING_DIRECTORY}" >> "$BASH_ENV"
+            IFS=$'\n'
+            for line in ${ZEEK_CIRCLE_EXTRA_ENV}; do
+              echo "export ${line}" >> "$BASH_ENV"
+            done
+      - clone_repo
+      - restore_cache:
+          name: Restore ccache directory
+          keys:
+            - ccache-v1-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ .Branch }}-{{ .Revision }}
```
ckreibich, line 261: Could you explain the cascade of keys?

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+  circleci-config-testing:
+    jobs:
+      - fetch-test-traces
+
+      # The job name is split up and used to map to the Dockerfile and final image
+      # name. Everything after "build-image-" should map to a directory in the ci
+      # directory in the repo.
+      - create-build-image:
+          name: build-image-alpine
+          << : *FILTER_IF_PR_NOT_FULL_CI
```
ckreibich, line 499: I'm curious here about the structure of image builds relative to the linux builds/tests below. Are there pros/cons to having the build-image creations as separate, required jobs vs running as steps invoking commands in the linux-build jobs?

If the image build were part of the linux-build job, it might be a little more concise — for example you would not need to duplicate the `FILTER_IF_...` conditions and image names.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+      - linux-build:
+          name: debian-13-arm
+          docker_image: zeek/zeek-ci:debian-13
+          resource_class: arm.large
+          requires: [fetch-test-traces, build-image-debian-13]
+          << : *FILTER_IF_PR_NOT_FULL_CI
+      - linux-build:
+          name: debian-13-binary
+          docker_image: zeek/zeek-ci:debian-13
+          configure_flags:  --prefix=$ZEEK_CI_WORKING_DIR/install --libdir=$ZEEK_CI_WORKING_DIR/install/lib --binary-package --enable-static-broker --enable-static-binpac --disable-broker-tests --build-type=Release --ccache --enable-werror
```
ckreibich, line 579: Might just be me, but just like in #5476 I think it'd be nice to have these flag combos centralized somehow.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+          << : *FILTER_IF_PR_NOT_FULL_CI
+      - linux-build:
+          name: fedora-44
+          docker_image: zeek/zeek-ci:fedora-44
+          configure_flags:  --prefix=$ZEEK_CI_WORKING_DIR/install --libdir=$ZEEK_CI_WORKING_DIR/install/lib --binary-package --enable-static-broker --enable-static-binpac --disable-broker-tests --build-type=Release --ccache --enable-werror
+          requires: [fetch-test-traces, build-image-fedora-44]
+          << : *FILTER_IF_SKIP_ALL
+      - linux-build:
+          name: opensuse-leap-16.0
+          docker_image: zeek/zeek-ci:opensuse-leap-16.0
```
ckreibich, line 601: It'd be cool here if the linux-build job definition could reuse the `name` value for a default `docker_image` value.

.circleci/config.yml
```diff
@@ -0,0 +1,745 @@
...
+          paths:
+            - ~/.ccache
+      - run:
+          name: Test
+          command: ci/test.sh
+      - store_test_results:
+          name: Store btest results
+          path: btest-results
+
+_filter_if_skip_all: &FILTER_IF_SKIP_ALL
```
ckreibich, line 452: Nice that this remains so familiar.

ckreibich (approved):
Really interesting. Kudos for getting the labels to work. I like the overall feel too. I do wonder whether we can cut down on redundant mentioning of the build instances (like `alpine`, `centos-stream-10` etc) a bit more via different job/step structure, a matrix, etc.

Let's get it running!

.circleci/config.yml
timwoj: `save_cache` saves the output for later runs, so we don't have to redownload the files for every full build (workflow) pass. `persist_to_workspace` is within the individual workflow runs, so that the individual jobs get the data without having to restore it from cache themselves. It also keeps each job from downloading/resetting the cache individually if something changes.

.circleci/config.yml
timwoj: Sure, I can add those.

.circleci/config.yml
timwoj: It might be easier to add documentation to the parameter definitions, so you can just look back at what they mean. I was going to add that but it's mostly intended for orb definitions, which we're not doing here, but I can see how it would be useful to have anyways.

.circleci/config.yml
timwoj: Good point. It should fail to build if that doesn't exist.

.circleci/config.yml
timwoj: https://circleci.com/docs/reference/configuration-reference/#ending-a-job-from-within-a-step is the more direct documentation.

.circleci/config.yml
timwoj: I added it to https://github.com/zeek/zeek/issues/5512

.circleci/config.yml
timwoj: That's definitely possible. I kind of liked it being explicit, but I suppose it could check whether the parameter is set before basing it on the name. It matters for the various debian jobs, for example.

.circleci/config.yml
timwoj: It's also a lot smaller. The branch/tag/cron/pull-request filtering happens in the setup step, so we don't have to re-check those things for every job like we were on Cirrus.

.circleci/config.yml
timwoj: It has to be separate because I have to tell the Linux build/test jobs what docker image they're based on via the `docker:` directive. i can't do that unless the image exists beforehand, which means the Linux image-build jobs need to happen before. Adding the FILTER directives to them avoids the need to check/build the builder images if they're not going to be used anyways.

.circleci/config.yml
timwoj: Actually, I take that back, it can't be that way. The `docker` directive can either be fully parameterized or fixed. I can't script setting it to something.

.circleci/config.yml
timwoj: I added it to https://github.com/zeek/zeek/issues/5512.

PR #5502, opened 2026-05-28
dns: cast to uint32_t in ExtractLong to avoid signed-shift overflow

dxbjavid:
DNS_Interpreter::ExtractLong in src/analyzer/protocol/dns/DNS.cc builds the 4-byte field with `val = data[0] << 24`. data[0] is u_char, integer-promoted to int, so when the high bit is set the result of the shift exceeds INT_MAX, which is signed-integer overflow. The field is used for TTL on every DNS RR (line 312) as well as SOA serial/refresh/retry/expire/minimum, RRSIG inception/expiration and TSIG sign_time_sec, all reached from any DNS response on the default-enabled analyzer. Cast data[0] to uint32_t before the shift, matching the well-defined-behavior idiom from `938fd42d Avoid integer overflow`. The remaining `data[N] << 16` and `data[N] << 8` shifts fit in int and are unchanged.

### DIFF
--- a/src/analyzer/protocol/dns/DNS.cc
+++ b/src/analyzer/protocol/dns/DNS.cc
@@ -636,8 +636,8 @@ uint32_t DNS_Interpreter::ExtractLong(const u_char*& data, int& len) {
     uint32_t val;
 
     val = static_cast<uint32_t>(data[0]) << 24;
-    val |= data[1] << 16;
-    val |= data[2] << 8;
+    val |= static_cast<uint32_t>(data[1]) << 16;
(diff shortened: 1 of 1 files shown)


timwoj (changes requested):
Same comment here as in https://github.com/zeek/zeek/pull/5490 and https://github.com/zeek/zeek/pull/5504

dxbjavid:
Good point, done. Cast data[1] and data[2] as well, since those shifts can overflow the same way clang-tidy flags.

evantypanski (approved):

timwoj (approved):

PR #5504, opened 2026-05-28
net_util: cast to uint32_t in extract_uint32 to avoid signed-shift overflow

dxbjavid:
zeek::extract_uint32 in src/net_util.cc builds the 4-byte network-order field with `val = data[0] << 24`. data[0] is u_char, integer-promoted to int, so when the high bit is set the result of the shift exceeds INT_MAX, which is signed-integer overflow. The function is called on the IPv4 octets of every APP_IPADDRESS_TAG (network-address) ASN.1 value in SNMP via network_address_to_val(), reached from any SNMP packet on the default-enabled UDP/161 analyzer, and on the wire-order bytes of every <IPv4,port> tuple in a BitTorrent tracker peers list. Cast data[0] to uint32_t before the shift, matching the well-defined-behavior idiom from `938fd42d Avoid integer overflow`. The remaining data[N] << 16 and data[N] << 8 shifts fit in int and are unchanged.

### DIFF
--- a/src/net_util.cc
+++ b/src/net_util.cc
@@ -229,8 +229,8 @@ uint32_t extract_uint32(const u_char* data) {
     uint32_t val;
 
     val = static_cast<uint32_t>(data[0]) << 24;
-    val |= data[1] << 16;
-    val |= data[2] << 8;
+    val |= static_cast<uint32_t>(data[1]) << 16;
+    val |= static_cast<uint32_t>(data[2]) << 8;
     val |= data[3];
 
     return val;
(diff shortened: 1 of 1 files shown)


timwoj (changes requested):
I think the same thing here as https://github.com/zeek/zeek/pull/5490. clang-tidy says this change should happen for the others as well. Any shift of 8-bits or more may overflow.

dxbjavid:
Good point, same as #5490. Cast data[1] and data[2] too now, since those shifts can corrupt the high bits as well. data[3] has no shift so I left it alone.

evantypanski (approved):

PR #5507, opened 2026-05-29
irc: guard empty params before indexing nick in DeliverStream

dxbjavid:
A NICK command with no parameters ("NICK\r\n", or only trailing whitespace) reaches the handler with an empty params string. `string_view nick{params}; nick[0]` then indexes a zero-length view, which is UB under [string.view.access] and aborts under hardened libstdc++/libc++. Guard on non-empty before the index, matching the WHO and INVITE branches just below.

### DIFF
--- a/src/analyzer/protocol/irc/IRC.cc
+++ b/src/analyzer/protocol/irc/IRC.cc
@@ -838,7 +838,7 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
 
     else if ( irc_nick_message && command == "NICK" ) {
         string_view nick{params};
-        if ( nick[0] == ':' )
+        if ( nick.starts_with(':') )
             nick = nick.substr(1);
 
         EnqueueConnEvent(irc_nick_message, ConnVal(), val_mgr->Bool(orig), make_intrusive<StringVal>(prefix.c_str()),


src/analyzer/protocol/irc/IRC.cc
```diff
@@ -838,7 +838,7 @@ void IRC_Analyzer::DeliverStream(int length, const u_char* line, bool orig) {
 
     else if ( irc_nick_message && command == "NICK" ) {
         string_view nick{params};
-        if ( nick[0] == ':' )
+        if ( ! nick.empty() && nick[0] == ':' )
```
awelzel: Nice find.

What do you think about aligning with the "QUIT" command above instead?
```
if ( nick.starts_with(":")
    nick = nick.substr(1)
```

That said, looking more, those two are the only ones using string_view, while elsewhere, `params[0] == ":"` or a derived `message[0] == ":"` test is assumed to work (because string[0] is defined to be `\0` even for an empty string, so it's all a bit sad.

An isolated `starts_with(":")` fix for NICK seems reasonable as a fix. Mind changing towards that, will merge then.

dxbjavid: Good call, switched it to starts_with(":") to match the QUIT branch. Agreed the string vs string_view split is a bit error-prone, but kept this one scoped to the NICK fix.

awelzel (approved):

PR #5508, opened 2026-05-29
logging: Make Log::set_buf() sticky for new writers

awelzel:
@J-Gras reports that WriterFrontends created dynamically at runtime after a Log::set_buf() call for the stream during zeek_init() will not inherit the buffering state. Cache ``buf`` also on the stream and apply it to writers created dynamically.

The test isn't 100%, but it fails fairly regularly when removing the Log::set_buf() calls in zeek_init() or the change in this commit is reverted. The logging infrastructure is pretty opaque, so...

Added:

Also make `Log::flush_interval = 0.0` disable the flushing time. Separate commit and adapt the test to use it to make it reliable.


awelzel:
https://zeekorg.slack.com/archives/CTGDG83B8/p1779992377252229

scripts/base/init-bare.zeek
```diff
@@ -3879,7 +3879,13 @@ export {
 	## In earlier Zeek releases this was governed by :zeek:see:`Threading::heartbeat_interval`.
 	## For Broker, see also :zeek:see:`Broker::log_batch_interval`.
 	##
+	## Setting this value to a zero interval (0.0sec), disables the global
+	## log flushing timer. Note that it is then your responsible to call
```
timwoj, line 3883:
```suggestion
	## log flushing timer. Note that it is then your responsibility to call
```

timwoj (approved):
Just a minor grammatical nit

J-Gras:
Thanks for merging! Fwiw, I can confirm I was able to change buffering as intended now.

PR #5509, opened 2026-05-29
Fix threading value constructor to copy sizes

J-Gras:
The constructor not setting the corresponding size fields looks like an oversight.


ckreibich (approved):
Oops, indeed. Thanks Jan. 👍

ckreibich:
Btw the test failures were due to us accidentally pulling in the wrong baseline for `zeek-testing` — Tim fixed that [here](https://github.com/zeek/zeek/commit/ba9b8593d4b4892b39b97b87e69aeb9982488cf3). I ran locally and they passed fine.

PR #5510, opened 2026-05-29
dtls: bound handshake fragment sequence shift in proc_handshake

dxbjavid:
proc_handshake shifts message_sequence_seen by the attacker-controlled DTLS sequence delta, which is UB once it exceeds 31; reject fragments outside the 32-bit reassembly window before the shift.


0xxon:
I added a pcap that raises the notice (created with Claude Opus 4.7). I also opened an additional issue, as the current parsing code is just incorrect (see #5532). Apart from that, this looks good to me - but as I added to it someone else should probably take a short look.

The fix itself looks good to me - and this is something that we should fix before reworking the DTLS logic.

dxbjavid:
Appreciate you adding the pcap and btest. Agreed on #5532, the sequence numbers should really only drive replay detection; weaving them into the reassembly state is what makes this path so fragile. Landing this first makes sense so the shift UB is gone before that rework. Happy to wait on a second look.

ckreibich (approved):

ckreibich:
I dug into the [RFC](https://www.rfc-editor.org/info/rfc6347/#section-4.1.2.6) regarding the bitfield window and [found](https://www.rfc-editor.org/info/rfc6347/#section-4.1.2.6)
> A minimum window size of 32 MUST be supported, but a window size of 64 is preferred and SHOULD be employed as the default. Another window size (larger than the minimum) MAY be chosen by the receiver.

— fwiw.

PR #5515, opened 2026-06-01
ftp: cast to uint32_t in parse_port to avoid signed-shift overflow

dxbjavid:
A PORT command like `200,0,0,1,4,31` makes parse_port evaluate `bytes[0] << 24` with bytes[0] in 128..255, which shifts a set bit into the sign of int (signed overflow). The value comes straight off the wire through parse_ftp_port, so it is reachable on the default FTP path. Cast bytes[0] to uint32_t before the shift, matching the data[0] casts already used in the dns and null analyzers.

### DIFF
--- a/src/analyzer/protocol/ftp/functions.bif
+++ b/src/analyzer/protocol/ftp/functions.bif
@@ -30,7 +30,7 @@ static zeek::RecordValPtr parse_port(const std::string& line)
 
 		if ( good )
 			{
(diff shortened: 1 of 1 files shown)


timwoj (changes requested):
Same comment about casting all of the values.

dxbjavid:
Done, cast the rest of the shifted bytes too (matching the null/dns cleanups).

evantypanski (approved):

PR #5517, opened 2026-06-01
ipaddr: cast first octet to uint32_t in ConvertString

dxbjavid:
Noticed the IPv4 path in `IPAddr::ConvertString` builds the address with `a[0] << 24`. `a[0]` is an int validated to 0-255, so for any first octet >= 128 the shift pushes a set bit into the sign bit of int, which is signed overflow. Reached from `to_addr()` on a dotted-quad with a high first octet, e.g. `to_addr("200.0.0.1")`. Cast it to uint32_t so the shift runs on an unsigned type, matching the idiom in the recent DNS/null/net_util fixes.

### DIFF
--- a/src/IPAddr.cc
+++ b/src/IPAddr.cc
@@ -95,7 +95,7 @@ bool IPAddr::ConvertString(const char* s, in6_addr* result) {
         if ( num < 0 || num > 255 )
             return false;
 
-    uint32_t addr = (a[0] << 24) | (a[1] << 16) | (a[2] << 8) | a[3];
+    uint32_t addr = (static_cast<uint32_t>(a[0]) << 24) | (a[1] << 16) | (a[2] << 8) | a[3];
     addr = htonl(addr);
     memcpy(result->s6_addr, v4_mapped_prefix, sizeof(v4_mapped_prefix));
     memcpy(&result->s6_addr[12], &addr, sizeof(uint32_t));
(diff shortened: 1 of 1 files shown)


evantypanski (commented):
Hi, from what I can tell, the same comments from #5490, #5504, and #5502 apply here.

It is not respectful to completely ignore feedback on previous PRs on future PRs. You also continue to not add tests unless prompted, and all of these are extremely small changes in separate PRs when they could be grouped.

If you continue without addressing the feedback from previous PRs in future contributions, we will not allow any more pull requests from you. Please respect the Zeek maintainers' time.

Also, the project's AI policy was just merged. From now on, please attribute *each pull request* with LLM assistance to the LLM. It seems reasonable to go on previous responses that these pull requests were assisted (or created) by an LLM, but let me know if that is wrong.

Thanks.

dxbjavid:
Hi, and sorry you're right, and I should have addressed the earlier feedback before opening more PRs. I got too focused on pushing small fixes quickly and didn't slow down to read and apply your comments or add tests. That's on me.

Going forward I'll group related changes together, include tests, and follow the AI policy and disclose any LLM assistance on each PR. Thanks for your patience, and sorry for the wasted time.

evantypanski (approved):
LGTM with the extra shifts I added; just making sure CI is ok in case something silly happened

PR #5521, opened 2026-06-02
Add note about Broker WebSocketShim abort() to NEWS

awelzel:
@ckreibich - mind looking this through?


0xxon:
This looks good to me. We should also note this in https://docs.zeek.org/en/master/advanced/devel/websocket-api.html

0xxon (approved):

PR #5525, opened 2026-06-02
MIME: Fix mime_content_hash_sha256 and remove unused variable

awelzel:
Local fuzzing with ubsan tickled an integer overflow for content_hash_length_sha256 due to the variable never being initialized in the first place.

It also hadn't actually been used and there was a bug for mime_content_hash_sha256 receiving content_hash_length, but that one only being updated when mime_content_hash was in use.

So, just use the one single content_hash_length for both and add a test for the different combinations of event usages.


0xxon (approved):
This looks much saner than before. Thank you.

PR #5526, opened 2026-06-03
Extend bifcl to generate C++ native-callable functions

vpax:
Currently, `bifcl` produces C++ functions that take a `Frame` pointer and an `Args` vector of `ValPtr`s, and return a `ValPtr`. This PR modifies the compiler to generate a _shim_ function with that calling sequence, and a _native_ function that takes and returns C++ values. The _shim_ does standard checking on the `Args` vector, converts its `ValPtr` objects to C++ values (as before), calls the _native_ function, and converts its return value back to a `ValPtr`.

This split provides two benefits, one minor and one potentially more significant:

- `.bif` functions no longer need to convert their values to `ValPtr` objects, which makes them a tad simpler to write
- Script optimization can potentially directly call the _native_ functions without going through the _shim_. This saves a significant amount of `Val` construction/destruction and value conversions.

Because of the change for BiF return types from `ValPtr` to native C++ values, this change is not backward-compatible by itself. To address that issue, `bifcl` only makes the split for source files that include a `%gen-native` directive.

Some BiFs do not have simple native C++ calling sequences. This includes variadic functions, and functions whose bodies either access the argument vector directly via `@ARG@`/`@ARGS@`/`@ARGC@`, or those that access the `Frame` pointer. For all of these (which are rare), `bifcl` still generates a _shim_/_native_ split (to enable those functions to return native C++ values), but the calling sequence for the _native_ function retains the `Frame` pointer and the `Args` vector. In addition, to unify the recognition by the compiler of such functions, access to the `Frame` pointer is now done via a `@FRAME@` variable akin to those for the arguments.

The groups of changes/commits include:

- Adding low-level type conversion utility functions that return native types rather than `ValPtr`s
- Converting `bifcl` to a two-pass structure, where the first pass accumulates BiF function bodies in a string to enable identification of `@ARG@` etc. usage.
- Converting all of the standard BiFs to leverage `%gen-native`
- Trimming some unnecessary/detrimental `return std::move(....)` in the standard BiFs

I used Claude Opus 4.7 for the 2nd and 3rd of these. I reviewed all of the code it generated, iterating with it for style and design choices, and have a full understanding of the changes.


src/val-convert.cc
```diff
@@ -8,15 +8,60 @@
...
+    return i;
+}
+
+std::optional<zeek_uint_t> convert_string_to_native_count(const StringVal* sv, std::string& err, int base) {
+    const char* s = sv->CheckString();
+    char* end_s;
+    uint64_t u = static_cast<uint64_t>(strtoull(s, &end_s, base));
+    if ( s[0] == '\0' || std::any_of(static_cast<const char*>(end_s), s + ::strlen(s),
+                                     [](char c) { return ! (c == '\0' || ::isspace(c)); }) ) {
+        err = "bad conversion to count";
```
ckreibich, line 40: Minor: I wondered here whether we should take this opportunity to be more detailed in those errors, perhaps "bad conversion of %s to count" etc, since those error messages are mostly used as-is.

src/val-convert.cc
```diff
@@ -8,15 +8,60 @@
...
+        return std::nullopt;
     }
-    return make_intrusive<DoubleVal>(d);
+    return d;
+}
+
+std::optional<zeek_int_t> convert_string_to_native_int(const StringVal* sv, std::string& err, int base) {
+    const char* s = sv->CheckString();
+    char* end_s;
+    zeek_int_t i = strtoll(s, &end_s, base);
```
ckreibich, line 25: I realize you're just moving the meat of these over, but I'm surprised by these conversions. They don't check errno for overflows and just return LLONG_MAX etc on out-of-bounds values:
```
$ zeek -e 'print to_int("99999999999999999999999")'
9223372036854775807
```
Seems worth a closer look, though likely not in this PR.

src/val-convert.h
```diff
@@ -11,6 +11,16 @@ namespace zeek::detail {
...
 // on failure.
 
+// These return native C++ types, or, if the optional value is missing,
+// an error occurred and has been populated in `err`.
+extern std::optional<double> convert_string_to_native_double(const StringVal* sv, std::string& err);
+extern std::optional<zeek_int_t> convert_string_to_native_int(const StringVal* sv, std::string& err, int base);
+extern std::optional<zeek_uint_t> convert_string_to_native_count(const StringVal* sv, std::string& err, int base);
+extern std::optional<zeek_uint_t> convert_int_to_native_count(zeek_int_t i, std::string& err);
+extern std::optional<double> convert_int_to_native_double(zeek_int_t i);
+extern std::optional<double> convert_count_to_native_double(zeek_uint_t c);
```
ckreibich: These are declared but not defined.

tools/bifcl/bif_arg.cc
```diff
@@ -35,49 +51,88 @@ BuiltinFuncArg::BuiltinFuncArg(const char* arg_name, int arg_type) {
...
+void BuiltinFuncArg::PrintCImplParam(std::string& out) { appendf(out, "%s %s", bif_types[type].c_type, name); }
+
+void BuiltinFuncArg::PrintCImplCallArg(std::string& out) { appendf(out, "%s", name); }
+
+const char* BuiltinFuncArg::NativeReturnType() const { return bif_types[type].native_return_type; }
+const char* BuiltinFuncArg::NativeToVal() const { return bif_types[type].native_to_val; }
+
+// FILE* convenience overloads: format into a std::string, then fputs.
+// to_file resolves the std::string& overload at the lambda call site,
+// avoiding the ambiguity that a direct &PrintZeek member-pointer would have.
```
ckreibich: I found this kind of confusing, and it doesn't just apply to `PrintZeek` but all of the functions below. Might just drop that part.

tools/bifcl/builtin-func.l
```diff
@@ -163,18 +167,18 @@ FILE* fp_netvar_h = nullptr;
 FILE* fp_netvar_def = nullptr;
 FILE* fp_netvar_init = nullptr;
 
-void remove_file(const char *surfix);
+void remove_file(const char *suffix);
```
ckreibich, line 170: Wow, we were pretty consistent about that one :-)

tools/bifcl/builtin-func.y
```diff
@@ -152,35 +154,230 @@ const char* arg_list_name = "BiF_ARGS";
...
+
+// Set when the input file contains a top-level "%gen-native" directive.
+// When false (the default), every BiF gets the legacy single-function
+// emission, preserving backward compatibility for .bif files that haven't
+// been adapted to return native types.
+bool gen_native = false;
+
+// The current BiF's return type (its BuiltinFuncArg representation) or
+// nullptr for void. Captured at return_type, consumed at body_end so
+// the _native can use a native primitive return type when applicable.
```
ckreibich: That's pretty terse. I get that this is referring to the grammar's symbols as opposed to functions or whatever, but what is `_native`?

tools/bifcl/builtin-func.y
```diff
@@ -152,35 +154,230 @@ const char* arg_list_name = "BiF_ARGS";
...
+std::unique_ptr<BuiltinFuncArg> return_type_arg;
+
+// While parsing a func body we accumulate captured C++ into body_buf:
+// the body_end action picks among natively-callable native+shim,
+// not-natively-callable native+shim, and the legacy single-function form
+// based on gen_native plus var_arg / uses_frame / uses_args_token.
```
ckreibich: I would find it amazing if this PR provided a full example of what a BiF turns into in C++, in these three flavors. `bifcl` has never been adequately documented, so it might help us work toward that too.

src/analyzer/protocol/ftp/functions.bif
```diff
@@ -226,7 +227,7 @@ function fmt_ftp_port%(a: addr, p: port%): string
 	else
 		{
 		zeek::emit_builtin_error("conversion of non-IPv4 address in fmt_ftp_port",
```
ckreibich, line 229:
```suggestion
                zeek::emit_builtin_error("conversion of non-IPv4 address in fmt_ftp_port", a);
```

ckreibich (changes requested):
Neat, I like the way the new BiFs look.

I didn't trace the details of all the new generator code in builtin-func.y, but I studied examples of the generated code. The sanitizers, plus the fact that the tests pass, should hopefully have these covered.

I used Gemini to cross-check the bulk changes and look for security issues.

My main mechanical request: we need to get rid of the merge of master into your branch — those make for messy merge graphs over time. Could you ditch those and just rebase the 4 commits onto master?

src/val-convert.cc
vpax: Deferring as this will result in baseline changes unrelated to this PR's functionality.

src/val-convert.cc
vpax: Likewise deferring.

src/val-convert.h
vpax: Good catch, fixed in next push.

tools/bifcl/bif_arg.cc
vpax: Good point, fixed in the upcoming.

tools/bifcl/builtin-func.l
vpax: Heh!

tools/bifcl/builtin-func.y
vpax: Likewise fixed in upcoming

tools/bifcl/builtin-func.y
vpax: Cool, I'll append that as a new comment shortly.

src/analyzer/protocol/ftp/functions.bif
vpax: (deferring as this is separate from this effort)

vpax:
Here's an example illustrating the changes. For this input:
```
## Returns whether a string starts with a substring.
function starts_with%(str: string, sub: string%) : bool
        %{
        if ( sub->Len() > str->Len() )      
                return zeek::val_mgr->Bool(false);

        auto sub_s = sub->ToStdStringView();
        auto s = str->ToStdStringView();
        return zeek::val_mgr->Bool(s.starts_with(sub_s));
        %}
```
`bifcl` generates .cc that looks like this (with editing to trim out distractions that aren't actually semantic), same as before:
```
// This file was automatically generated by bifcl from original-demo.bif.

ValPtr BifFunc::starts_with_bif(Frame* frame, const Args* BiF_ARGS)
	{
	if ( BiF_ARGS->size() != 2 )
		{
		emit_builtin_error(util::fmt("starts_with() takes exactly 2 argument(s), got %lu", BiF_ARGS->size()));
		return nullptr;
		}

	StringVal* str = ((*BiF_ARGS)[0].get()->AsStringVal());
	StringVal* sub = ((*BiF_ARGS)[1].get()->AsStringVal());

	if ( sub->Len() > str->Len() )      
		return val_mgr->Bool(false);

	auto sub_s = sub->ToStdStringView();
	auto s = str->ToStdStringView();
	return val_mgr->Bool(s.starts_with(sub_s));
	} // end of BifFunc::starts_with
```
If we update the input to the following (changes flagged with `# <---`):
```
%gen-native # <---

## Returns whether a string starts with a substring.
function starts_with%(str: string, sub: string%): bool
	%{
	if ( sub->Len() > str->Len() )
		return false; # <---

	auto sub_s = sub->ToStdStringView();
	auto s = str->ToStdStringView();
	return s.starts_with(sub); # <---
	%}
```
then `bifcl` generates the following:
```
// This file was automatically generated by bifcl from new-demo.bif.

bool BifFunc::starts_with_native(StringVal* str, StringVal* sub)
	{
	if ( sub->Len() > str->Len() )
		return false;

	auto sub_s = sub->ToStdStringView();
	auto s = str->ToStdStringView();
	return s.starts_with(sub);
	} // end of BifFunc::starts_with

ValPtr BifFunc::starts_with_bif(Frame* frame, const Args* BiF_ARGS)
	{
	if ( BiF_ARGS->size() != 2 )
		{
		emit_builtin_error(util::fmt("starts_with() takes exactly 2 argument(s), got %lu", BiF_ARGS->size()));
		return nullptr;
		}

	StringVal* str = ((*BiF_ARGS)[0].get()->AsStringVal());
	StringVal* sub = ((*BiF_ARGS)[1].get()->AsStringVal());
	return val_mgr->Bool(BifFunc::starts_with_native(str, sub));
	} // end of shim for BifFunc::starts_with
```

ckreibich (approved):
Thanks for those updates, Vern — I've pushed the rebase. Let's see how it fares in CI.

PR #5530, opened 2026-06-03
Add encrypted_client_hello HPKE KDF and AEAD id to ssl-log-ext

0xxon:
This commit adds two new cryptographic values from the encrypted client hello to the ssl-log-ext policy script.

This was (very partially) authored by Claude Opus 4.7. I understand and verified every part of the commit that I didn't write by hand.

Merging this also requires merging the external test baselines.


awelzel (approved):
Thanks!

PR #5531, opened 2026-06-03
Automatically cancel exisiting GH CI workflows on PR update

bbannier:
Previously we would run all GH actions CI workflows to completion, even if the PR was updated and another run was started. Since we only have a limited number of workers available this causes contention across the whole GH org. It is also inconsistent with how Cirrus CI behaved, so people are not at all concious about cancelling outdated jobs.

This PR changes the configuration of all workflows which trigger for pull requests to automatically cancel existing jobs on updates, mirroring the behavior of Cirrus CI.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,6 +12,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


0xxon (approved):

0xxon:
That looks good to me. Thank you - and sorry, I was totally unaware :)

timwoj (approved):

Issue #5512, opened 2026-05-29
Remainder of CircleCI work


timwoj:
This is a follow-up ticket for https://github.com/zeek/zeek/pull/5500 with a list of the tasks remaining for the config to be complete:

- [x] Builds that only have docs changes still run, where we were skipping them entirely on Cirrus
- [x] zeekctl-testing job
- [x] include-plugins job
- [x] Steps to update code coverage
- [x] Steps to run the fuzzer tests
- [x] Add parameter for ZEEK_CI_CPUS to set it nearby where the arch is. See https://github.com/zeek/zeek/pull/5534#discussion_r3379434660.
- [x] Docker release builds
- [x] ASAN builds are hitting a 10-minute timeout when running btests. This is possibly scripts/base/protocols/ldap/functions.spicy. I have it disabled for those builds on the branch.
- [x] Some jobs are hitting the same 10-minute timeout when running the CTU-SME btest in zeek-testing
- [x] Add documentation to the job parameters even if we're not going to make orbs out of this code. It makes things easier to follow.
- [x] Make yaml anchors for the various extra `configure` arguments
- [x] macOS fails to find junit2html when running btests (but this was a problem on Cirrus too)
- [x] Archival of the btest results (although they're uploaded into the CircleCI system, so this might not be needed?)
- [x] A nightly build trigger
- [x] Cluster-testing job (which depends on the docker builds)
- [x] Steps to run the benchmarker
- [x] Bring back Broker CI. See zeek/broker#507
- [ ] The build-image building job needs something that periodically cleans up old images (see TODO in config)

awelzel:
> Archival of the btest results (although they're uploaded into the CircleCI system, so this might not be needed?)

The uploading failed for [this job](https://app.circleci.com/pipelines/circleci/JFr6soTf1AmP5bL3bFPRKn/uUFkHbvQy8B6MBNzPh61b/453/workflows/1b1a389d-55f4-4f54-93c0-5be79722d507/jobs/2240):

```
Archiving the following test results
  * /root/project/btest-results/btest/results.xml
  * /root/project/btest-results/zeek-testing/results.xml


failed uploading test results: File: root/project/btest-results/btest/results.xml had the following problems:
  * ws-invalid-app, invalid testcase element: unstable
  * invalid top level end element: testcase
  * ws-app, invalid testcase element: unstable
  * invalid top level end element: testcase
  * invalid top level end element: testsuite
```

Seems the btest junit format isn't fully compatible.

[...]

PR #5534, opened 2026-06-03
Add support for building Docker images to CircleCI config

timwoj:
This PR migrates the jobs from Cirrus for building and pushing Docker images to DockerHub and AWS ECR. It maintains the same handling of not rebuilding the same image twice. For AWS ECR, it uses [Circle's builtin support for OIDC](https://circleci.com/docs/guides/permissions-authentication/openid-connect-tokens/) instead of using a username/password for access.

Related to https://github.com/zeek/zeek/issues/5512

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   win: circleci/windows@5.1.1
+  aws-cli: circleci/aws-cli@5.4.1
 
 # These parameters are set via setup.yml. The PR ones are based on the labels for PRs
 # from Github. changes_only_docs is for PRs that only change docs.
@@ -31,6 +32,10 @@ parameters:
     type: boolean
     default: false
   changes_only_docs:
+    type: boolean
+    default: false
+  has_release_tag:
+    description: True if running for a tag of the form v1.2.3 or v1.2.3-rc1
     type: boolean
     default: false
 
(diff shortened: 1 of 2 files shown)


timwoj:
Kicking this over to draft since my conditional for running the jobs is wrong and Circle's complaining about it.

timwoj:
This is ready for review now.

.circleci/generate-pr-params.py
```diff
@@ -50,5 +52,8 @@
     if not params:
         print("No GitHub labels found on PR")
 
+if GIT_TAG and re.match(r"/^v\d+\.\d+\.\d+$", GIT_TAG):
```
awelzel:
```suggestion
if GIT_TAG and re.match(r"^v\d+\.\d+\.\d+(-rc[0-9]+)?$", GIT_TAG):
```

Fronting `/` seems off and allowing trailing `-rc[0-9]+` for "RC releases". does that seem reasonable to you?

.circleci/config.yml
```diff
@@ -31,6 +32,9 @@ parameters:
   changes_only_docs:
     type: boolean
     default: false
+  has_release_tag:
+    type: boolean
```
awelzel, line 39:
```suggestion
    type: boolean
    description: True if running for a tag of the form v1.2.3 or v1.2.3-rc1.
```

or some such?

.circleci/config.yml
```diff
@@ -456,6 +460,231 @@ jobs:
...
+      arch:
+        type: string
+      resource_class:
+        type: string
+    environment:
+      ZEEK_CONFIGURE_FLAGS: --ccache --generator=Ninja --build-type=Release --disable-btest-pcaps --disable-cpp-tests --disable-broker-tests
+      BUILDKIT_PROGRESS: plain
+      IMAGE_TAG: zeek/zeek-multiarch:<< parameters.arch >>
+      # This is based on the medium resource class passed into the job.
+      ZEEK_CI_CPUS: 4
```
awelzel, line 503: Nit. Is it easy to inject this via parameters? If we change the resource class, would be nice if the magic number and the class are next to each other (or using different classes for arm vs x86).

.circleci/config.yml
```diff
@@ -456,6 +460,231 @@ jobs:
...
+            echo "Building final image for ${ZEEK_VERSION} at ${CREATED_DATE}"
+
+            (cd docker && docker build \
+                                 --build-arg ZEEK_VERSION=${ZEEK_VERSION} \
+                                 --build-arg GIT_COMMIT=${CIRCLE_SHA1} \
+                                 --build-arg CREATED_DATE=${CREATED_DATE} \
+                                 -t ${IMAGE_TAG} \
+                                 -f final.Dockerfile .)
+
+            mkdir -p /tmp/zeek_images/<< parameters.arch >>
```
awelzel: Nit. Setting `IMAGE_ARCH=<< parameters.arch >>` in the environment section would make this look a bit less alien and easier to copy/paste into a shell script. Or maybe change to use `$IMAGE_TAG` as the filename for `docker save` (certainly needs `mkdir -p` here too and not sure it's better).

.circleci/config.yml
```diff
@@ -456,6 +460,231 @@ jobs:
...
+
+  container-image-manifest-build:
+    environment:
+      BUILDKIT_PROGRESS: plain
+      AWS_REGION: us-east-1
+    docker:
+      - image: cimg/aws:2025.01
+    steps:
+      - run:
+          name: Output docker/builx versions
```
awelzel:
```suggestion
          name: Output docker/buildx versions
```

.circleci/config.yml
```diff
@@ -456,6 +460,231 @@ jobs:
...
+            # to allow the branch in the above only_if attribute of this task.
+            # elif [ "${CIRCLE_BRANCH}" = "topic/timw/circleci-docker-image-builds" ]; then
+            #   echo "IMAGE_NAME=zeek-next"  >> $BASH_ENV
+            #   echo "IMAGE_TAG=latest" >> $BASH_ENV
+            else
+              echo "Bad tag/branch for container_image_manifest"
+              env
+              exit 1
+            fi
+      # Clone the repo to get the CI scripts but don't include any of the extras
```
awelzel: Hmm, hmm, does this need to come before "Set image tag"? The "Set image tag" logic accesses `VERSION` so that already needs a clone?

awelzel (commented):
Had a few comments, but nothing major. Thanks for porting this over.

I'd be okay moving this in and then testing via tagging an 8.0.9-rc1 or some such to see if the pipeline works out okay.

---

Aside: After a bit forth and back with Claude, GitHub + Docker have nice Multi-platform container image support today, building on separate native runners and creating the manifest:

https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

What we're doing today was introduced [early 2023](https://github.com/zeek/zeek/commit/5510b2496a51fe5cb2d6f7c76fd7c63eb2f9bb1d) when GitHub didn't have native ARM runners and QEMU builds for arm on x86 were timing out. Arm runners are [freely available since early 2025](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), so there might be the topic of moving this to GitHub actions, or maybe CircleCI has ORBs that could do that.

.circleci/generate-pr-params.py
timwoj: Good call.

.circleci/config.yml
timwoj: Sure. I actually opened a support question with the CircleCi guys about whether they could expose those sizes in some way automatically. Having a bunch of magic numbers when they have fixed resource classes seems silly. He said he'd look into it.

timwoj: I'll do this one in a follow-up PR, since it's going to touch some other parts of the config too, not just this one job. I added a note to https://github.com/zeek/zeek/issues/5512.

.circleci/config.yml
timwoj: Ah, good catch. I looked for uses of any of the `ci` scripts, but didn't note the use of `VERSION`.

timwoj:
> maybe CircleCI has ORBs that could do that

There might be. I didn't look too hard at what was available. I figure I can port over what we have (with some cleanup) and then explore replacing bits of it with orbs.

.circleci/config.yml
```diff
@@ -483,6 +488,232 @@ jobs:
...
+                -v /tmp/ccache:/tmp/ccache \
+                -w /src/zeek zeek-builder:latest \
+                bash -c "./configure $ZEEK_CONFIGURE_FLAGS && ninja -C build install"
+
+            # docker cp ${CIRCLE_WORKING_DIRECTORY}/. zeek-builder-container:/src/zeek
+            # if [ -d /tmp/ccache ]; then
+            #   docker cp /tmp/ccache zeek-builder-container:/tmp/ccache
+            # fi
+            # docker start --attach zeek-builder-container
+            # docker cp zeek-builder-container:/tmp/ccache /tmp/ccache
```
awelzel: Maybe drop those? /tmp/ccache is bind-mounted and cached below, so not sure if these are useful to have? Suspect that was trying t stash the cache into the builder or so?

.circleci/config.yml
```diff
@@ -483,6 +488,232 @@ jobs:
...
+            ./ci/container-images-tag-and-push.sh
+            REGISTRY_PREFIX=public.ecr.aws/ ./ci/container-images-tag-and-push.sh
+
+  public-ecr-cleanup:
+    environment:
+      AWS_REGION: us-east-1
+    docker:
+      - image: cimg/base:current
+    steps:
+      - clone_repo
```
awelzel:
```suggestion
      - clone_repo:
            include_submodules: false
            include_external_tests: false
```
Unless they are default? But used explicitly above, so maybe here too?

awelzel (approved):
Lets see :-)

.circleci/config.yml
timwoj: They're true by default.

.circleci/config.yml
timwoj: Ah yah, those were leftovers from when I was trying to figure out why the directories weren't getting mounted. That entire commented block can go.

PR #5538, opened 2026-06-04
Geneve: fix parsing of encapsulated ipv6

0xxon:
This fixes a type in the
PacketAnalyzer::register_packet_analyzer call that prevented encapsulated ipv6 from being parsed.

The majority of this commit was written by Claude Opus 4.7. The issue was found with the help of Claude Opus 4.7 while exploring logic errors in the Zeek codebase.

I either wrote or verified all Zeek codechanges. I didn't look at the python script to generate the pcap at all - but I did examine and verify the pcap that was generated.


testing/btest/Traces/tunnels/geneve-ipv6.pcap.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def main():
+    pkts = [
+        geneve_ipv6_packet(
+            "10.0.0.1", "10.0.0.2", 12345,
+            "fd00::1", "fd00::2", 40000, 40001, b"hello",
+        ),
+        geneve_ipv6_packet(
+            "10.0.0.2", "10.0.0.1", 12345,
```
awelzel:
```suggestion
            "10.0.0.1", "10.0.0.2", 12345,
```

This is a nit and not related to the issue. Making this a uni-directional tunnel would align a bit more with using geneve for packet mirroring, rather than having two independent tunnels. The two udp geneve connections caught me by surprise.

Untested, but think this should then result in a single GENVE tunnel only.

awelzel (approved):
Nice find!

Could update Traces/README/

testing/btest/Traces/tunnels/geneve-ipv6.pcap.py
0xxon: Thank you for this - I applied as suggested.

PR #5540, opened 2026-06-04
Management framework: fix regression in metrics-port collision avoidance

0xxon:
A bug was introduced in 0c0769b1b2, which added a statement before the opening braces of the for loop.

This means that most of the ports were never recorded in instance_ports_set, which can lead to collisions.

Found with the help of Claude Opus 4.7.


awelzel (approved):
Ouch.

PR #5541, opened 2026-06-04
Fix off-by-one error in base/utils/thresholds

0xxon:
Base-utils-thresholds will try to read a non-existing entry from the thresholds vector, when crossing the last threshold.

As far as I can tell this error has been present since the original SVN import.

This problem was found using Claude Opus 4.7.


timwoj (approved):

ckreibich:
Johanna's out for a bit, so I went ahead and merged.

PR #5542, opened 2026-06-04
conn_key: allow factories to customize fragment reassembly keys

### LINKED ISSUE #4871: Add virtual DoFragmentKey() API to ConnKey factory
Feedback on the pluggable (programmable?) ConnKey / ConnTuple feature mentioned IP fragments are reassembled using a hard-coded key constructed here:

https://github.com/zeek/zeek/blob/050035e33eed88cc1248c5d857508da60b86ee3a/src/Frag.cc#L300-L302

I'm thinking given that fragments are IP specific, we could add a virtual method `DoFragmentKey(ip, Packet)` to `zeek::conn_key::fivetuple::Factory` that by default does what is currently hard-coded in Frag.cc, but if a user needs to, they can override the hook. The ``std::tuple`` likely needs to be replaced with something different that is more buffer like.

This will allow users to provide a custom IP fragment key for reassembly. Users only really can rely on L2 and IP header being available, but that seems reasonable.

arnavahire19:
## Summary

  This adds fragment-key customization to conn-key factories so custom connection-key schemes can also
  control fragment reassembly keys.

  ## Changes

  - add a fragment-key factory API
  - use factory-provided keys in fragment reassembly lookup
  - add VLAN-aware fragment-key handling for the VLAN five-tuple factory
  - add a regression test plugin/btest for fragment-key customization

  Fixes #4871.

### DIFF
--- a/src/Frag.cc
+++ b/src/Frag.cc
@@ -6,6 +6,7 @@
 #include "zeek/NetVar.h"
 #include "zeek/Reporter.h"
 #include "zeek/RunState.h"
+#include "zeek/conn_key/Manager.h"
 #include "zeek/session/Manager.h"
 
 constexpr uint32_t MIN_ACCEPTABLE_FRAG_SIZE = 64;
@@ -26,10 +27,9 @@ void FragTimer::Dispatch(double t, bool /* is_expire */) {
 }
 
 FragReassembler::FragReassembler(session::Manager* arg_s, const std::shared_ptr<IP_Hdr>& ip, const u_char* pkt,
-                                 const FragReassemblerKey& k, double t)
-    : Reassembler(0, REASSEM_FRAG) {
+                                 FragReassemblerKey k, double t)
+    : Reassembler(0, REASSEM_FRAG), key(std::move(k)) {
     s = arg_s;
-    key = k;
 
     const struct ip* ip4 = ip->IP4_Hdr();
     if ( ip4 ) {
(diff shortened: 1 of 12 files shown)


arnavahire19:
The Windows CI failures split into two groups:

  1. `plugins.connkey-fragment` segfaults with exit code 139, which is the
  likely regression from this PR and needs direct debugging.
  2. Telemetry/Broker failures (`Broker::PEER_UNAVAILABLE`, failed local
  requests) may be unrelated platform-specific CI issues unless they can be
  reproduced as part of this change.

  I’m treating the fragment-test crash as the primary issue and would avoid
  changing Broker error handling based on the current evidence.

evantypanski:
Hi, thanks for the PR! We'll look at it but we have a lot going on right now.

In the meantime, did you use an LLM/AI? We merged an [AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md) so disclosing usage is necessary. Not meant to discourage you, but we would like to know the extent :)

And I looked at the test, it doesn't look like you updated the baseline. That would be through a command like `btest -U plugins/connkey-fragment.zeek` in the `testing/btest` directory. This could be a bit tricky to test without making a custom plugin for this, I would look at a number of the other tests and try to get a bit closer, I see it looks like [this test](https://github.com/zeek/zeek/blob/master/testing/btest/plugins/connkey.zeek) (which uses [this plugin](https://github.com/zeek/zeek/tree/master/testing/btest/plugins/connkey-plugin), hence those commands at the top). But, in order to test this particular feature you probably need a different plugin. It's a bit of a pain, but perhaps necessary for this.

Let me know if you need help; I'm probably not the best person for this particular feature but we can try to help. Keep in mind we are a bit busy right now.

arnavahire19:
Thanks, that makes sense.

  I agree the current test is too indirect and the baseline update is missing.
  I’ll rework this around a custom test plugin that changes fragment-key
  behavior in an observable way, then regenerate the btest baseline rather than
  relying on the current hook-style verification.

awelzel:
Hey @arnavahire19 ,

thanks for looking at that issue and tackling it :-)

> I agree the current test is too indirect and the baseline update is missing.
> I’ll rework this around a custom test plugin that changes fragment-key

What you could do here is to create a pcap with TCP connections where some of the data packets are fragmented. Put 3 of these connections in a single pcap (all sharing the same five tuple), but with different vlan taggings (none, one vlan tag, two vlan tags). Take `Traces/vlan-collisions.pcap` as inspiration. The crux/spin for the fragment reassembler is that some of the packets should be fragmented. Zeek should associate them with the correct connection when the `vlan_fivetuple` plugin is loaded.

`testing/btest/scripts/policy/frameworks/conn_key/vlan_fivetuple.zeek` should help as an example, too.

arnavahire19:
@awelzel 

  Thanks, that makes sense.

  I’ll rework the test around a trace-based `vlan_fivetuple` scenario instead of
  the current indirect plugin test. I’ll create a pcap with three TCP
  connections sharing the same five-tuple but differing by VLAN tagging (none,
  single, double), with fragmented data packets mixed in, and then verify Zeek
  associates the fragments with the correct connection when the `vlan_fivetuple`
  plugin is loaded.

  I’ll also regenerate the btest baseline as part of that update.

awelzel:
Hello @arnavahire19 - are you still interested in continuing with this PR? Do you need more guidance or did it end up too large a task?

ckreibich:
Another nudge here @arnavahire19, thanks.

ckreibich:
I'm closing this out until @arnavahire19 revisits this or we beat him to it. :-)

PR #5545, opened 2026-06-04
serialization: don't abort() on short read in binary format

uwezkhan:
ReadData() calls abort() when there aren't enough bytes left to satisfy a
read. The function is already written to return false on that condition and
all callers handle it, so the abort() just turns a recoverable parse error
into a hard process exit. Since the binary format is fed by cluster
log-write messages, a single malformed message can take a node down.

Drop the abort() and let the existing `return false` do its job. While here,
rewrite the bounds check as `count > input_len - input_pos` instead of
`input_pos + count > input_len`, which can wrap for a large count and let an
oversized read slip through.


ckreibich:

PR #5546, opened 2026-06-04
threading: validate set/vector sizes when reading binary values

uwezkhan:
Value::Read() reads the element count for sets and vectors as a signed 64-bit
value straight off the wire and hands it to new Value*[size] with no checks.

Two problems with that:
  - a negative count becomes a huge size_t in new[], which throws
    bad_array_new_length and takes the process down.
  - new Value*[size] leaves the pointers uninitialized. We fill them in a
    loop, but if one of the element reads fails partway through we return with
    size still set to the full count. The destructor then deletes every entry
    up to size, including the slots we never populated, so it frees garbage
    pointers.

Reject a negative size up front, and value-initialize the array so the unused
slots are null and safe to delete. Valid input is unaffected since the loop
overwrites every slot anyway.

These values are reachable from the cluster log-write path, where the value
type and count come from a remote peer.

### DIFF
--- a/src/threading/SerialTypes.cc
+++ b/src/threading/SerialTypes.cc
@@ -295,7 +295,22 @@ bool Value::Read(detail::SerializationFormat* fmt) {
             if ( ! fmt->Read(&val.set_val.size, "set_size") )
                 return false;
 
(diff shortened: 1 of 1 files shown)


ckreibich:

PR #5549, opened 2026-06-05
Fix parsing of NTP extensions to match changes in RFC7822

timwoj:
Claude Opus 4.6 guided the fix here, with some additional edits by me. I fully understand what it's doing, and why.

The handling of NTP extensions changed in RFC 7822 (in 2016) to where the extensions are now a minimum length of 16 bytes and also have a header fields for the extension length. OSS-Fuzz generated a packet with a trailing Message Authentication Code, but our handling of the extensions was eating it, throwing a binpac exception, and disabling the analyzer for the connection. OSS-Fuzz caught this through the undefined behavior sanitizer because the exception handler causes an integer overflow while trying to log.

This PR changes the code to first check to see if we have valid lengths to be parsing extensions before moving on to parse the MAC. This seems strange to handle it like this all the time since RFC 7822 is NTPv4, but the packet generated in NTPv2 and it works fine. I'm not the biggest fan of the weird `trailing` variable in `NTP_std_msg` but it's like that to be able to pass the data into the next record/function.


awelzel:
>  I'm not the biggest fan of the weird trailing variable in NTP_std_msg but it's like that to be able to pass the data into the next record/function.

I also find this pretty strange and it doesn't fix the signed integer overflow OSS-Fuzz reported. It probably fixes the reproducer instance, but not the actual issue. Following pcap still triggers it:

[extension-value-len-overflow.pcap.txt](https://github.com/user-attachments/files/28700029/extension-value-len-overflow.pcap.txt)
[gen-ntp-extension-overflow.py](https://github.com/user-attachments/files/28700030/gen-ntp-extension-overflow.py)

```
$ zeek -Cr testing/btest/Traces/ntp/extension-value-len-overflow.pcap 
/home/awelzel/corelight-oss/zeek-security/fuzz-build/src/analyzer/protocol/ntp/ntp_pac.cc:933:15: runtime error: signed integer overflow: 18 + 2147483647 cannot be represented in type 'int'
    #0 0x56396b86910d in binpac::NTP::Extension_Field::Parse(unsigned char const*, unsigned char const*, int) /home/awelzel/corelight-oss/zeek-security/fuzz-build/src/analyzer/protocol/ntp/ntp_pac.cc:933:15
    #1 0x56396b86868e in binpac::NTP::NTP_Ext_and_MAC::Parse(unsigned char const*, unsigned char const*) /home/awelzel/corelight-oss/zeek-security/fuzz-build/src/analyzer/protocol/ntp/ntp_pac.cc:566:55
    #2 0x56396b866024 in binpac::NTP::NTP_std_msg::Parse(unsigned char const*, unsigned char const*) /home/awelzel/corelight-oss/zeek-security/fuzz-build/src/analyzer/protocol/ntp/ntp_pac.cc:476:41
```


I had the minimal hunk below while fuzzing locally that limits `value_len` and `sig_len` to `len` and ensures `len >= 16`. These are used for `&length`. Without the capping, when they end up at around `INT_MAX` they tickle the signed integer overflow. Seems protocol violations are in order here. I also find the NTP extension parsing pretty strange as it always assumes RFC 5906 extensions rather than actually verifying that it's these AutoKey extensions a different extension. Minimally there is [RFC 8915](https://datatracker.ietf.org/doc/html/rfc8915) Network Time Security for the Network Time Protocol from 2020 with a [different format](https://datatracker.ietf.org/doc/html/rfc8915#name-the-nts-authenticator-and-e). But that seems a separate topic/feature.


```
diff --git a/src/analyzer/protocol/ntp/ntp-protocol.pac b/src/analyzer/protocol/ntp/ntp-protocol.pac
index dac60ebfc1..263420a9a2 100644
--- a/src/analyzer/protocol/ntp/ntp-protocol.pac
+++ b/src/analyzer/protocol/ntp/ntp-protocol.pac
@@ -114,13 +114,13 @@ type NTP_CONTROL_MAC = record {
 type Extension_Field = record {
        first_byte_ext: uint8;
        field_type:     uint8;
-       len:            uint16;
+       len:            uint16 &enforce(len >= 16);
        association_id: uint16;
        timestamp:      uint32;
        filestamp:      uint32;
-       value_len:      uint32;
+       value_len:      uint32 &enforce(value_len <= (len - 18);
        value:          bytestring &length=value_len;
-       sig_len:        uint32;
+       sig_len:        uint32 &enforce(value_len <= (len - 22));
        signature:      bytestring &length=sig_len;
        pad:            padding to (len - offsetof(first_byte_ext));
 } &let {
```

timwoj:
I'm sure there's other ways to tickle overflows in the NTP analyzer. This was mostly about fixing the way that we're parsing extensions to match the standard lengths for extensions that was causing it to fall down the wrong code path. The pcap I added doesn't have any extensions in it, while our analyzer thought it did.

src/analyzer/protocol/ntp/ntp-protocol.pac
```diff
@@ -114,13 +128,13 @@ type NTP_CONTROL_MAC = record {
...
-	len:            uint16;
+	len:            uint16 &enforce(len >= 16);
 	association_id: uint16;
 	timestamp:      uint32;
 	filestamp:      uint32;
-	value_len:      uint32;
-	value:          bytestring &length=value_len;
-	sig_len:        uint32;
+	value_len:      uint32 &enforce(value_len <= (len - 18));
+	value:          bytestring &length=value_len ;
```
awelzel, line 136:
```suggestion
	value:          bytestring &length=value_len;
```

src/analyzer/protocol/ntp/ntp-protocol.pac
```diff
@@ -52,20 +65,27 @@ type NTP_std_msg = record {
...
+	trailing:        bytestring &restofdata;
+} &let {
+	# Pass the rest of the data to be processed as extension and/or MAC
+	ext_and_mac: NTP_Ext_and_MAC(trailing) withinput trailing;
+} &byteorder=bigendian;
+
+type NTP_Ext_and_MAC(data: bytestring) = record {
 	extensions: case ( has_exts ) of {
-		true  -> exts: Extension_Field[] &until($input.length() <= 24);
+		true  -> exts: Extension_Field[] &until($input.length() < 16);
```
awelzel, line 76: After skimming the RFCs and staring, I wanted to propose the following as that seems a bit more robust (poking the buffer if there's a valid extension to be parsed for every element):

```suggestion
		true  -> exts: Extension_Field[] &until(! ntp_has_extensions($input));
```

And that compiles, but the generated code has no traces of the condition (nor does the previous version show anything about the $input.length() < 16) and the test results are the same. There's something strange with binpac as if it silently swallows this `&until` condition. I'm also not convinced that `< 16` check is all that great as even an MD5 mac may take up 20 bytes.

So, I don't feel good about this one as I don't understand what's going. Maybe someone with more BinPAC/NTP knowledge? Otherwise could see pulling the enforce changes alone for the UBSAN fix and revisit with #5551 ? (or have a discussion about moving this to Spicy - seems a good one TBH).

awelzel (commented):
> [@timwoj](https://github.com/timwoj) requested your review on: [zeek/zeek#5549](https://github.com/zeek/zeek/pull/5549) Fix parsing of NTP extensions to match changes in RFC7822.

I had actually looked at this a few days ago, but never submitted the comments. Sorry.

src/analyzer/protocol/ntp/ntp-protocol.pac
timwoj: Weird. That's definitely a bug in binpac. It's parsing the `&until`, but it's not adding the code path to check it. If i lift that line out of a case statement to the top level, then it does what it's supposed to:

```
type NTP_Ext_and_MAC(data: bytestring) = record {
    exts: Extension_Field[] &until(! ntp_has_extensions(data));

	mac_fields: case ( has_mac ) of {
		true -> mac: NTP_MAC_var(mac_len);
		false -> nil2: empty;
	} &requires(has_mac);
} &let {
```

```
    for (; /* forever */; ++t_exts__elem__it) {
        // Check &until( ( ! ntp_has_extensions(data) ) )
        if (  ( ! ntp_has_extensions(data()) )  ) {
            delete exts__elem_;
            exts__elem_ = nullptr;
            goto end_of_exts;
        }
```

All of the tests continue to pass. I looked through the pcaps we have though, and none of them have any extensions in them, so I can't be sure it's actually parsing them correctly. The new btest only really tests for the absence. I generated the following with Claude, and it throws a binpac exception while trying to parse it.

[gen_ntp_ext.py](https://github.com/user-attachments/files/28903110/gen_ntp_ext.py)
[ntp_with_extension.pcap.zip](https://github.com/user-attachments/files/28903114/ntp_with_extension.pcap.zip)

awelzel: > type NTP_Ext_and_MAC(data: bytestring) = record {
>     exts: Extension_Field[] &until(! ntp_has_extensions(data));

But that doesn't do the right thing as the data bytestring isn't moved forward with extensions being parsed.

I think we should expedite the length clamping to fix the UBSAN issue (and maybe then the fuzzer will make further progress) and delay tweaking the actual extension parsing to #5551 or a Spicy re-implementation of the NTP analyzer?

timwoj: > But that doesn't do the right thing as the data bytestring isn't moved forward with extensions being parsed.

Right, I forgot to put that part in there. It's generating the code correctly, but the code still isn't working. I'll go ahead with just lifting the length checks into another PR and close this one.

timwoj: I'm really curious about why binpac doesn't generate that code correctly, but binpac's codegen is just opaque enough that it's tough to investigate.

PR #5554, opened 2026-06-08
General CircleCI fixups

timwoj:
This is a collection of fixups for various CircleCI things, mostly pulled out of https://github.com/zeek/zeek/pull/5534 to make that PR more targeted on the docker changes alone.

- Add a trigger ID check for the master branch during setup to avoid master being built from more than one trigger. This keeps us from doing two full builds for every merge to master.
- Rename the default builds workflow from the one that I gave it when I was doing the original work.
- Rework clone_repos a little bit to allow skipping submodules, plus making submodule updates a bit faster.


ckreibich (approved):
LGTM, thanks for the explanation in person regarding the trigger UUID etc.

PR #5555, opened 2026-06-08
Some signed bitshift fixes

evantypanski:
This just has a few bitshift fixes that I found just from grepping through the codebase, not using clang tidy. We can find more with clang-tidy but it might be a bit noisy. Figured I'd just get a few in at once.

Added benefit of switching `+` to `|`, mainly because I think it's clearer to read.

The base64 one was a little funky and I used Gemini to help understand the code there. I don't think we should keep that math around, it made me really nervous with the `i++` and ternaries then later relying on `i` going beyond the buffer... just odd stuff to my eyes.

### DIFF
--- a/src/packet_analysis/protocol/geneve/Geneve.cc
+++ b/src/packet_analysis/protocol/geneve/Geneve.cc
@@ -95,10 +95,10 @@ bool GeneveAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pack
 
     // Get the next header. This will probably be Ethernet (0x6558), but get it
     // anyways so that the forwarding can do its thing.
-    auto next_header = (data[2] << 8) + data[3];
+    auto next_header = (static_cast<uint32_t>(data[2]) << 8u) | data[3];
 
     // Grab the VNI out of the data before advancing the data pointer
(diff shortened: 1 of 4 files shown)


src/packet_analysis/protocol/geneve/Geneve.cc
```diff
@@ -95,10 +95,10 @@ bool GeneveAnalyzer::AnalyzePacket(size_t len, const uint8_t* data, Packet* pack
 
     // Get the next header. This will probably be Ethernet (0x6558), but get it
     // anyways so that the forwarding can do its thing.
-    auto next_header = (data[2] << 8) + data[3];
+    auto next_header = (static_cast<uint32_t>(data[2]) << 8) | data[3];
```
timwoj: A fix that would make clang-tidy not complain anymore would also change the constants to be unsigned. Up to you on whether to add that.

```suggestion
    auto next_header = (static_cast<uint32_t>(data[2]) << 8u) | data[3];
```

evantypanski: you're right, I thought since the operator becomes unsigned `clang-tidy` would be happy, but apparently not! I'll just add the `u`'s. I guess it's only fixing a few but that's okay until we enable it

timwoj (approved):

PR #5556, opened 2026-06-08
A pile of random changes for CircleCI

timwoj:
Continuing chipping away at the list in https://github.com/zeek/zeek/issues/5512, this fixes a pile of random issues:

- `junit2html` was not in the PATH for macOS so it always reported an error
- BTest artifacts are now stored with the build so they can be downloaded later
- The various additional  `configure` flags are now defined as YAML anchors
- The various job parameters now have documentation
- The CTU-SME test is disabled for asan builds on CircleCI due to its really really long runtime causing job timeouts


.circleci/config.yml
```diff
@@ -421,6 +421,7 @@ jobs:
       - run:
           name: Set extra environment variables
           command: |
+            echo "export PATH=${PATH}:${HOME}/.local/bin" >> "$BASH_ENV"
```
awelzel: Fun commit title - want to replace `/Users/tim` with `$HOME` maybe? :-)

testing/external/commit-hash.zeek-testing
```diff
@@ -1 +1 @@
-b728810e3760411c2147e28b91749220f673c95e
+54b142ab89acb908d34e117edd20410d4d51a14b
```
awelzel:
```
# Running this test with asan enabled on CircleCI takes long enough that CI times
# out. Disable for the time being in that environment.
# @TEST-REQUIRES: ! is_asan && test "${CIRCLECI}" != "true"
```

This disables it for all of CircleCI because of the `&& CIRCLECI != true`?  (or do I need more coffee?). Could be fine as Christian also pointed out that the CTU pcap takes too long, but the comment reads as if only ASAN was intended?

awelzel (approved):

testing/external/commit-hash.zeek-testing
timwoj: Looking at it again, it should actually be `! (is_asan && test "${CIRCLECI}" = "true")`. It shouldn't run if it's asan and CircleCI.

PR #5558, opened 2026-06-10
logging: Stash vector of Field* into LogWriteHeader, use for plugin hooks

### LINKED ISSUE #5548: Use-after-free in CreateWriter: fields passed to HOOK_LOG_INIT invalidated by WriterFrontend::Init()
## Summary

`Manager::CreateWriter()` passes the `fields` pointer array to `WriterFrontend::Init()`, which now takes ownership and destroys it. It then passes the same dangling `fields` pointer to `PLUGIN_HOOK_VOID(HOOK_LOG_INIT, ...)`, resulting in a use-after-free.

This causes a segfault in any plugin that implements `HookLogInit` and accesses the `fields` parameter.

## Regression

This is a regression introduced between Zeek 7.x/8.0 and 8.2. The change is in `WriterFrontend::Init()`:

**8.0 behavior** — stores the pointer without taking ownership; makes a separate copy for the backend:

```cpp
void WriterFrontend::Init(int arg_num_fields, const Field* const* arg_fields) {
    num_fields = arg_num_fields;
    fields = arg_fields;  // Just stores the pointer
    initialized = true;

    if ( backend ) {
        auto fs = new Field*[num_fields];
        for ( auto i = 0; i < num_fields; ++i )
            fs[i] = new Field(*fields[i]);  // Copy for backend
        backend->SendIn(new InitMessage(backend, arg_num_fields, fs));
    }
}
```

**8.2 behavior** — takes ownership and destroys the input array:

```cpp
void WriterFrontend::Init(int arg_num_fields, const Field* const* arg_fields) {
    initialized = true;
    header.fields.reserve(arg_num_fields);
    for ( int i = 0; i < arg_num_fields; i++ )
        header.fields.emplace_back(*arg_fields[i]);

    if ( remote )
        broker_mgr->PublishLogCreate(..., arg_num_fields, arg_fields);

    if ( backend )
        backend->SendIn(new InitMessage(backend, arg_num_fields, arg_fields));  // Takes ownership
    else {
        for ( int i = 0; i < arg_num_fields; i++ )
            delete arg_fields[i];  // Deletes fields!
        delete[] arg_fields;       // Deletes array!
    }
}
```

## The bug in CreateWriter

In `src/logging/Manager.cc`, `CreateWriter()` calls:

```cpp
winfo->writer->Init(num_fields, fields);  // Destroys fields in 8.2

if ( ! from_remote ) {
    winfo->hook_initialized = true;
    PLUGIN_HOOK_VOID(HOOK_LOG_INIT, HookLogInit(writer_name, instantiating_filter,
                                                local, remote, *winfo->info,
                                                num_fields, fields));  // Use-after-free!
}
```

After `Init()` returns, `fields` is either freed or owned by a background thread. The subsequent `HookLogInit` dereferences dangling pointers.

## Reproduction

The crash is triggered whenever a plugin implements `HOOK_LOG_INIT` and `CreateWriter` is called with `from_remote=false`. In practice this is most easily triggered in cluster mode when a sender node creates a writer for a log stream, but it can also occur for any local writer creation.

### GDB output

```
#0  std::vector<zeek::threading::Field, ...>::_M_realloc_append<...>
#1  std::vector<...>::push_back (__x=<error reading variable: Cannot access memory at address 0x55500f369358>)
#2  my_hook_log_init (path="broker", ...) at hook_handler.cc:99
#3  plugin::MyPlugin::HookLogInit(...)
#4  zeek::plugin::Manager::HookLogInit(...) const
#5  zeek::logging::Manager::CreateWriter(...)
#6  zeek::logging::Manager::CreateWriterForFilter(...)
#7  zeek::logging::Manager::WriteToFilters(...)
```

Memory at the `fields` array shows freed/reused heap:

```
(gdb) x/8gx fields_ptr
0x55555a578fe0: 0x000055500f369358  0x1c37526cc0a4b5c8   # Corrupted/freed pointers
0x55555a578ff0: 0x0000555557be95e0  0x0000555557bf1320
```

## Suggested fix

Move the `HOOK_LOG_INIT` invocation to **before** `WriterFrontend::Init()`, or pass a separate copy of the fields to the hook. For example:

```cpp
if ( ! from_remote ) {
    winfo->hook_initialized = true;
    PLUGIN_HOOK_VOID(HOOK_LOG_INIT, HookLogInit(writer_name, instantiating_filter,
                                                local, remote, *winfo->info,
                                                num_fields, fields));
}

winfo->writer->Init(num_fields, fields);
```

This preserves the original semantics where the hook sees valid field data before ownership is transferred.

## Environment

- Zeek 8.2.0
- Any plugin implementing `HOOK_LOG_INIT` that accesses the `fields` parameter
- Triggered in cluster mode (sender creating writers for remote log streams)

---

*This issue was identified and debugged using [Claude Code](https://claude.ai/code) (model: claude-opus-4-6-v1, Anthropic).*

awelzel:
@pbcullen reports that WriterFrontend::Init() frees the fields array when local logging is disabled, but then logging::Manager happily passes on that pointer to HookLogInit(), resulting in use-after-free accesses.

Keep a separate vector of threading::Field* pointers in the LogWriteHeader held by each WriterFrontend and use its .data() attribute for HookLogInit() and HookLogWrite().

Extend the logging-hooks-plugin to verify that the fields pointer is the same for either hook invocation and also add a logging-hooks-no-local test disabling local logging as that would have tickled the issue.

Closes #5548


src/logging/Manager.cc
```diff
@@ -1785,11 +1786,13 @@ WriterFrontend* Manager::CreateWriter(EnumVal* id, EnumVal* writer, WriterBacken
...
     winfo->writer = new WriterFrontend(*winfo->info, id, writer, local, remote);
     winfo->writer->Init(num_fields, fields);
+    fields = nullptr; // freed or owned by WriterBackend, cannot use.
 
     if ( ! from_remote ) {
         winfo->hook_initialized = true;
         PLUGIN_HOOK_VOID(HOOK_LOG_INIT, HookLogInit(writer_name, instantiating_filter, local, remote, *winfo->info,
-                                                    num_fields, fields));
+                                                    winfo->writer->header.field_pointers.size(),
+                                                    winfo->writer->header.field_pointers.data()));
```
bbannier, line 1798: The only reason we materialize a `field_pointers` is so we can pass the same value on init and on write, and using a member caches the computation. Passing the same information twice seems redundant to me, and I wonder whether we couldn't change the semantics of `HookLogWrite` to allow passing an unset `fields` value (`nullptr`) so we could eventually drop the argument later; I am not entirely sure how much of a breaking change that would be since we never documented anything about `HookLogWrite`. Writers which need the value should instead cache it from `HookLogInit`. If we then never pass `fields` on `HookLogWrite` there can be no inconsistency.

In this particular instance this would allow simplifying the code by computing temporary `Field*` values.

```suggestion
        auto ptrs_ = winfo->writer->header.fields | std::views::transform([](const auto& f) {return &f;} );
        auto ptrs = std::vector(ptrs_.begin(), ptrs_.end());
        PLUGIN_HOOK_VOID(HOOK_LOG_INIT, HookLogInit(writer_name, instantiating_filter, local, remote, *winfo->info,
                                                    ptrs.size(),
                                                    ptrs.data()));
```

awelzel: > seems redundant to me, I wonder whether we couldn't change the semantics of HookLogWrite 

Agree, but not going there. I have a pretty stubborn opinion on hooks in plugin/Plugin.h to never change unless they are completely broken, which I don't think this one is. It's historically a bit weird with the raw pointers, but it's also not a lot of effort or overhead to keep that working.

I had a temporary std::vector  as a fix first, but then remembered that we did have one fixup in the past to make sure the `*winfo->info` ref is stable between HookLogInit() and HookLogWrite() (which is even tested today). At least one implementation relied on that (map lookup using &info ptr between Init() and Write()), so that's where it's coming from.

With all the other historic things in `logging/`, the proposed fix seemed okayish to me.

0xxon (approved):
Looks like a good way to deal with this to me.

PR #5559, opened 2026-06-10
rpc/nfs/mount: Fix invalid enum initializers

awelzel:
Before casting to an enum, ensure it's a valid and known one, otherwise leave the corresponding fields unset.

Also, skip the NFSACL program to avoid producing weirds. Adding a PCAP of a NFS server running in a VM indicated NFSACL is in effect by default.

This was triggered by UBSAN fuzzing. Example:

        rpc/MOUNT.cc:173:66: runtime error: load of value 48, which is not a valid value for type 'BifEnum::MOUNT3::auth_flavor_t'

The testing pcaps were created using the scramble.py over real traces rather than having LLMs come up with a full trace. Not sure this was really worth it, but hey :man_shrugging:


src/analyzer/protocol/rpc/NFS.cc
```diff
@@ -351,15 +370,26 @@ RecordValPtr NFS_Interp::nfs3_fattr(const u_char*& buf, int& n) {
 }
 
 EnumValPtr NFS_Interp::nfs3_time_how(const u_char*& buf, int& n) {
-    BifEnum::NFS3::time_how_t t = static_cast<BifEnum::NFS3::time_how_t>(extract_XDR_uint32(buf, n));
-    auto rval = BifType::Enum::NFS3::time_how_t->GetEnumVal(t);
-    return rval;
+    uint32_t raw_time_how = extract_XDR_uint32(buf, n);
+    // time_how_t in src/types.bif goes to SET_TO_CLIENT_TIME = 2.
+    if ( raw_time_how > zeek::BifEnum::NFS3::time_how_t::SET_TO_CLIENT_TIME ) {
```
bbannier: Suggest to use `Lookup` like above since this is prone to becoming stale.

src/analyzer/protocol/rpc/NFS.cc
```diff
@@ -351,15 +370,26 @@ RecordValPtr NFS_Interp::nfs3_fattr(const u_char*& buf, int& n) {
...
+    return BifType::Enum::NFS3::time_how_t->GetEnumVal(static_cast<zeek_int_t>(raw_time_how));
 }
 
 EnumValPtr NFS_Interp::nfs3_ftype(const u_char*& buf, int& n) {
-    BifEnum::NFS3::file_type_t t = static_cast<BifEnum::NFS3::file_type_t>(extract_XDR_uint32(buf, n));
-    auto rval = BifType::Enum::NFS3::file_type_t->GetEnumVal(t);
-    return rval;
+    uint32_t raw_file_type = extract_XDR_uint32(buf, n);
+
+    // file_type_t in src/types.bif goes to FTYPE_FIFO = 7.
```
bbannier: Ditto.

src/analyzer/protocol/rpc/NFS.cc
```diff
@@ -438,9 +468,15 @@ RecordValPtr NFS_Interp::nfs3_pre_op_attr(const u_char*& buf, int& n) {
 }
 
 EnumValPtr NFS_Interp::nfs3_stable_how(const u_char*& buf, int& n) {
-    BifEnum::NFS3::stable_how_t stable = static_cast<BifEnum::NFS3::stable_how_t>(extract_XDR_uint32(buf, n));
-    auto rval = BifType::Enum::NFS3::stable_how_t->GetEnumVal(stable);
-    return rval;
+    auto raw_stable_how = extract_XDR_uint32(buf, n);
+
+    // stable_how_t in src/types.bif goes to FILE_SYNC = 2.
```
bbannier: Ditto.

src/analyzer/protocol/rpc/MOUNT.cc
```diff
@@ -169,9 +176,15 @@ Args MOUNT_Interp::event_common_vl(RPC_CallInfo* c, BifEnum::rpc_status rpc_stat
 }
 
 EnumValPtr MOUNT_Interp::mount3_auth_flavor(const u_char*& buf, int& n) {
-    BifEnum::MOUNT3::auth_flavor_t t = static_cast<BifEnum::MOUNT3::auth_flavor_t>(extract_XDR_uint32(buf, n));
-    auto rval = BifType::Enum::MOUNT3::auth_flavor_t->GetEnumVal(t);
-    return rval;
+    uint32_t raw_auth_flavor = extract_XDR_uint32(buf, n);
+
+    // auth_flavor_t in src/types.bif goes to AUTH_DES = 3.
+    if ( raw_auth_flavor > zeek::BifEnum::MOUNT3::auth_flavor_t::AUTH_DES ) {
```
bbannier: Ditto.

src/analyzer/protocol/rpc/MOUNT.cc
```diff
@@ -218,8 +231,12 @@ RecordValPtr MOUNT_Interp::mount3_mnt_reply(const u_char*& buf, int& n, BifEnum:
         auto enum_vector = make_intrusive<VectorType>(BifType::Enum::MOUNT3::auth_flavor_t);
         auto auth_flavors = make_intrusive<VectorVal>(std::move(enum_vector));
 
-        for ( auto i = 0u; i < auth_flavors_count; ++i )
-            auth_flavors->Assign(auth_flavors->Size(), mount3_auth_flavor(buf, n));
+        for ( auto i = 0u; i < auth_flavors_count; ++i ) {
+            auto af = mount3_auth_flavor(buf, n);
+
+            if ( af )
+                auth_flavors->Append(std::move(af));
```
bbannier:
```suggestion
        for ( auto i = 0u; i < auth_flavors_count; ++i ) {
            if ( auto af = mount3_auth_flavor(buf, n) )
                auth_flavors->Append(std::move(af));
        }
```

src/analyzer/protocol/rpc/NFS.cc
awelzel: Let me push back once - that was pretty pragmatic: I don't actually like the Lookup() pattern, but the MOUNT3 and NFS3 status enums are sparse, so there was no easy `!=` or `>` possible.

Also, most of these literally haven't changed in ~8 or ~15 years and the staleness is now easier detected via the weird that's produced.

src/analyzer/protocol/rpc/MOUNT.cc
awelzel: Certainly subjective :-) I'll pick it up.

src/analyzer/protocol/rpc/NFS.cc
bbannier: Even if they are seldomly updated, IMO the point still stands. This is literally magic constants all over the place, and that you were able to write it quickly doesn't mean it is maintainable. The `Lookup` pattern exists exactly to solve that, and refusing to use it doesn't improve things; what might help would be e.g., to fix `GetEnumVal` so it can safely work with unsigned values, and also ensure that it never returns unmapped enum values (could e.g., return an empty `ValPtr`).

awelzel: Now using Lookup().

timwoj (approved):

PR #5561, opened 2026-06-10
Use target-based approach for enabling sanitizers

timwoj:
The current approach for building with sanitizers is to build all targets in the tree with them, regardless of whether they are from Zeek or from something we vendor. While this seems interesting it also means that everything incurrs potential ASAN overhead, regardless of whether it is actually needed here or not, e.g., in a container on my machine `spicyz` invocations are very slow for some configurations (>30min runtimes for some builtin analyzers), but this seems largely unnecessary since Spicy already does its own sanitizer builds.

This patch changes that approach to instead be opt-in.

This is remake of https://github.com/zeek/zeek/pull/4623 after letting Claude have a go at fixing what was wrong with the branch. The new changes were generated entirely by Claude Opus 4.6, and are at line 1100 in CMakeLists.txt. It fixes the way that the symbols are linked in for hilti and spicy-rt so that they're avallable correctly at runtime. There were also some additional changes to make sure that the symbols were linked into the Broker libraries, otherwise some of the broker tests complained as well.

A very naive test shows a big performance improvement when running btests:

Master:
```
tim@nuc btest% time btest -j 14 spicy
104 tests successful, 1 skipped
btest -j 14 spicy  1373.58s user 122.01s system 953% cpu 2:36.83 total
```

Branch:
```
tim@nuc btest% time btest -j 14 spicy
104 tests successful, 1 skipped
btest -j 14 spicy  500.22s user 85.10s system 1101% cpu 53.119 total
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,6 +268,137 @@ if (ENABLE_CLANG_TIDY)
     endif ()
 endif ()
 
+if (ZEEK_SANITIZERS)
+    string(REPLACE "," " " _sanitizer_args "${ZEEK_SANITIZERS}")
+    separate_arguments(_sanitizer_args)
+    set(ZEEK_SANITIZERS "")
+
+    foreach (_sanitizer ${_sanitizer_args})
+        if (ZEEK_SANITIZERS)
+            set(ZEEK_SANITIZERS "${ZEEK_SANITIZERS},")
+        endif ()
+
+        if (_sanitizer STREQUAL "thread")
+            set(ZEEK_TSAN true)
+        endif ()
+
+        if (NOT _sanitizer STREQUAL "undefined")
+            set(ZEEK_SANITIZERS "${ZEEK_SANITIZERS}${_sanitizer}")
+            continue()
+        endif ()
+
+        if (NOT DEFINED ZEEK_SANITIZER_UB_CHECKS)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1013,9 +1025,17 @@ else ()
         set(ENABLE_STATIC_ONLY true)
     endif ()
 
+    # Set linker flags so broker executables/shared libs get the ASan runtime.
+    if (_sanitizer_flags)
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=${ZEEK_SANITIZERS}")
+        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=${ZEEK_SANITIZERS}")
```
bbannier: Could we pass `_sanizier_flags` here instead? Currently this seems to be the only place outside of that variable where we set `-fsanitize`.

bbannier (approved):
Thanks for looking into this again and sorry that I missed it! These numbers look great and exactly in line with what I was hoping for 😉.

I left only one comment below, but I pushed two more fixup commits (with additional context), and a merge of `master`.

CMakeLists.txt
timwoj: Definitely 👍🏼

PR #5562, opened 2026-06-10
Tune the `testing/external/scripts/testing-setup.zeek` script

ckreibich:
I was looking into where all the time goes when processing the CTU pcap in the `zeek-testing` testsuite (as part of #5512). I found that since that pcap is relatively sparse but spans a week, we spend almost half of our time collecting telemetry (via `policy/frameworks/telemetry/log.zeek`, which collects once per minute). We don't actually baseline telemetry.log, so the number of times we collect doesn't matter for the test.

Screenshot_20260610_141219

Dialing this way down provides a healthy speedup for all pcaps, for example (with a debug build):
```
$ time btest ./tests/m57-long.zeek
all 1 tests successful

real    0m42.132s
user    0m51.141s
sys     0m3.900s
```
... vs after:
```
$ time btest ./tests/m57-long.zeek
all 1 tests successful

real    0m26.908s
user    0m30.361s
sys     0m3.881s
```

We may still want to look into whether there's anything obvious to improve in the collection — 60 * 24 * 7 is still "only" 10K metrics collections.

I'm also dialing down the profiling frequency since it likewise doesn't contribute to the baselines. It mainly saves on generated log volume, not overhead.

The other changes are just minor local tweaks.

### DIFF
--- a/testing/external/scripts/testing-setup.zeek
+++ b/testing/external/scripts/testing-setup.zeek
@@ -17,6 +17,16 @@ redef DPD::track_removed_services_in_connection=T;
 	redef LogAscii::use_json = F;
 @endif
 
+@ifdef ( Telemetry::log_interval )
+	# We're reading pcaps, so the frequency of metrics collection is not
+	# significant. Frequent collection can dominate the runtime of a test,
+	# so dial this down considerably.
+	redef Telemetry::log_interval = 24hr;
+@endif
+
+# Similar to the above, dial down the profiling frequency.
+redef profiling_interval = 24hr;
+
 # The tests don't load intel data and so all Intel event groups are disabled
 # due to intel/seen/manage-event-groups being loaded by default. Disable that
 # functionality by default to cover execution in the intel/seen scripts.
--- a/testing/external/scripts/testing-setup.zeek
+++ b/testing/external/scripts/testing-setup.zeek
@@ -1,7 +1,7 @@
 # Sets some testing specific options.
 
-@load external-ca-list
-@load external-ct-list
+@load ./external-ca-list
+@load ./external-ct-list
 
 @load protocols/conn/failed-service-logging
 redef DPD::track_removed_services_in_connection=T;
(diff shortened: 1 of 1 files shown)


ckreibich:
I didn't explain the flamegraph — it came from `-O profile-ZAM` (per [this](https://github.com/zeek/zeek/tree/master/src/script_opt/ZAM#zam-profiling)). That's why it's script-layer only.

awelzel:
> I found that since that pcap is relatively sparse but spans a week, we spend almost half of our time collecting telemetry (via policy/frameworks/telemetry/log.zeek, which collects once per minute). We don't actually baseline telemetry.log, so the number of times we collect doesn't matter for the test.

Oh, nice!

With a release build and `perf` and `zeek-perf-support` enabled can also see that the CPU time spent by `zk.telemetry` thread is significant. Also, the memory stats with glibc via profile logger show up significantly for me (not showing when using jemalloc).

image


That's via:
```
$ zcat CTU-SME-11-Experiment-VM-Microsoft-Windows7AD-1-malicious-filtered.pcap.gz | ZEEKPERFSUPPORT=1 /usr/bin/time taskset -c 0,1,2,3 perf record -g /opt/zeek-dev-prod/bin/zeek -r - test-all-policy ../../scripts/testing-setup.zeek
$ perf script | grep -v _Zeek_Perf > ctu.stack
$ inferno-collapse-perf ./ctu.stack | inferno-flamegraph > ctu.svg
```

testing/external/scripts/testing-setup.zeek
```diff
@@ -1,7 +1,7 @@
 # Sets some testing specific options.
 
-@load external-ca-list
-@load external-ct-list
+@load ./external-ca-list
+@load ./external-ct-list
```
awelzel, line 4: Thanks :-)

awelzel (approved):
🎉

timwoj:
It's wild how much faster this makes that test run.

PR #5564, opened 2026-06-11
Cache docker image directly with official Docker actions

bbannier:


bbannier:
Note that since we exceed the cache size coming with the free account, even with these changes we will likely hit cache misses.

timwoj:
In light of the comment on https://github.com/zeek/zeek/pull/5563, should this PR remove uses of that action?

bbannier:
> In light of the comment on #5563, should this PR remove uses of that action?

🤦, fixed now.

timwoj (approved):

PR #5567, opened 2026-06-11
utils/atoi_n: Prevent overflows and fix digit parsing for base < 10

awelzel:
Fuzzing with UBSAN triggered the n = n * base + d line as overflowing, incorporate the atoi_n changes ~to~ from Spicy also in the Zeek code and add some tests for the edge cases.

Reference zeek/spicy#2382 zeek/spicy#2397

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -25,6 +25,7 @@
 #include <cstdarg>
 #include <cstdio>
 #include <cstdlib>
+#include <limits>
 
 #if defined(HAVE_MALLINFO) || defined(HAVE_MALLINFO2)
 #include <malloc.h>
@@ -1138,8 +1139,10 @@ char* strcasestr(const char* s, const char* find) {
 
 template<std::integral T>
 int atoi_n(int len, const char* s, const char** end, int base, T& result) {
-    T n = 0;
     int neg = 0;
+
+    if ( base < 2 || base > 36 )
+        return 0;
 
     if ( len > 0 && *s == '-' ) {
         neg = 1;
@@ -1147,11 +1150,20 @@ int atoi_n(int len, const char* s, const char** end, int base, T& result) {
         ++s;
     }
 
+    using Unsigned = std::make_unsigned_t<T>;
+    constexpr T min = std::numeric_limits<T>::min();
+    constexpr T max = std::numeric_limits<T>::max();
+    constexpr auto max_value = std::is_signed_v<T> ? static_cast<Unsigned>(min) : static_cast<Unsigned>(max);
+    static_assert(max_value <= std::numeric_limits<uint64_t>::max());
+
+    const auto base_u = static_cast<Unsigned>(base); // base unsigned
+    uint64_t n = 0;
+
     int i;
     for ( i = 0; i < len; ++i ) {
         unsigned int d;
 
-        if ( isdigit(s[i]) )
+        if ( s[i] >= '0' && s[i] < '0' + base )
             d = s[i] - '0';
 
         else if ( s[i] >= 'a' && s[i] < 'a' - 10 + base )
@@ -1164,6 +1176,11 @@ int atoi_n(int len, const char* s, const char** end, int base, T& result) {
             break;
 
         else
+            return 0;
+
+        // Can we multiply n without overflowing? And then
+        // also add the digit just parsed? Else, fail.
+        if ( n > (max_value / base_u) || d > (max_value - n * base_u) )
             return 0;
 
         n = n * base + d;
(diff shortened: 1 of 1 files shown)


src/util.cc
```diff
@@ -1138,20 +1139,31 @@ char* strcasestr(const char* s, const char* find) {
...
     if ( len > 0 && *s == '-' ) {
         neg = 1;
         --len;
         ++s;
     }
 
+    using Unsigned = std::make_unsigned_t<T>;
+    constexpr T min = std::numeric_limits<T>::min();
+    constexpr T max = std::numeric_limits<T>::max();
+    constexpr auto max_value = std::is_signed_v<T> ? static_cast<Unsigned>(min) : static_cast<Unsigned>(max);
```
timwoj: Can you explain this line? Why is the max value for a signed type the min value for the unsigned type? Wouldn't that generally be always zero?

awelzel: It's the min value casted to unsigned. For a signed int8_t, the range is `-128...127` and so the max_value we want to parse for digits ends-up as `128`. But, there's an issue that if `!neg`, the upper limit is actually `127` so this is wrong and the limit needs to be dynamic depending on what `neg` is.

I've added a test for parsing `128` as `int_t`. It succeeds and ends up with `-127` and that's unwanted behavior IMO.

Pondering, the behavior of parsing `-1` or `-2` into into `uint8_t` succeeding and producing `255` or `254` is also not what's wanted. E.g. parsing some supposedly small numbers into an `uint8_t` and someone injecting `-1` should err, not produce `255`.

Let me try to switch this wholesale to `std::from_chars`.

Thanks for the nudge.

awelzel: > Let me try to switch this wholesale to std::from_chars.

Done, this worked out well.

@bbannier - could you also take a look? You mentioned the `-1` to `255` behavior and not breaking the API, but I do think parsing `-1` into an unsigned int is an error.  Maybe could switch rt/util.h also to std::from_chars?

src/util.cc
```diff
@@ -1138,44 +1139,23 @@ char* strcasestr(const char* s, const char* find) {
...
+    if ( base < 2 || base > 36 )
+        return 0;
 
-        else if ( i > 0 )
-            break;
+    const char* last = s + len;
 
-        else
-            return 0;
+    auto r = std::from_chars(s, last, result, base);
```
bbannier, line 1149: I also tried that over in zeek/spicy#2403 and noticed that `std::from_chars` was much slower for very short inputs (by about 2x), and only became marginally faster by about 10% for the very longest inputs. This code probably predates `std::from_chars` so this is no deliberate unrolling, but I found this interesting. One can probably argue that this function is not on the critical path for data processing in Zeek 🤷

src/util.cc
bbannier: The reason I didn't want to change the behavior parsing `"-1"` into an unsigned was that I wanted to backport the ubsan fix, but didn't want to change behavior in a patch release. I explored this in https://github.com/zeek/spicy/pull/2403.

timwoj: I did the same for some of our other number parsing methods in https://github.com/zeek/zeek/pull/5111.

src/util.cc
```diff
@@ -2300,21 +2280,230 @@ TEST_SUITE("util") {
...
 
+        const char* hex0 = "00ff";
+        CHECK(atoi_n(strlen(hex0), hex0, nullptr, 16, val) == 1);
+        CHECK(val == 0xff);
+
         const char* fail = "XYZ";
         CHECK(atoi_n(strlen(fail), fail, nullptr, 10, val) == 0);
     }
 
+    // Test that parsing -1 into an uint8_t is rejected. Also check that 255 is
```
timwoj: I think you can delete this comment. There's enough commenting below to understand what's happening, plus it doesn't appear complete.

src/util.cc
```diff
@@ -2300,21 +2280,230 @@ TEST_SUITE("util") {
...
+        std::string int64t_min_overflow = int64t_min + "9";
+        std::string int64t_max_overflow = int64t_max + "9";
+        r = atoi_n(int64t_min_overflow.size(), int64t_min_overflow.c_str(), nullptr, 10, result);
+        CHECK_EQ(r, 0);
+        r = atoi_n(int64t_max_overflow.size(), int64t_max_overflow.c_str(), nullptr, 10, result);
+        CHECK_EQ(r, 0);
+    }
+
+    TEST_CASE("atoi_n uint64_t") {
+        std::string uint64t_min = "0";
```
timwoj: The `uint32_t` test uses `std::numeric_limits<uint32_t>::min()`. Should this too for consistency?

timwoj (approved):
Just a couple of nits. Looks good to me.

PR #5572, opened 2026-06-12
Skip GH CI when running in private repos

bbannier:
The matrix in our GH CI is huge so it requires a lot of compute. This is
fine for public repos which have no compute limit, but an issue for
private versions where we would run out of credits in days or hours.

With this patch we now only run this CI job if we are in a public repo.
An alternative approach would have been to hardcode the repo name, but
that would have meant that users couldn't run CI in their forks at all,
while now we at least gatekeep on public forks.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,8 +35,25 @@ env:
   DOCKER_BUILD_SUMMARY: false
 
 jobs:
+  # A job which only succeeds if we are running in a public repository.
+  #
+  # We use this below to skip all actual CI jobs in private repositories since
+  # our CI matrix is huge and would quickly burn through the free credits
+  # available for private repositories. To implement this jobs depend on this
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -35,8 +35,25 @@ env:
   DOCKER_BUILD_SUMMARY: false
 
 jobs:
+  # A job which only succeeds if we are running in a public repository.
+  #
+  # We use this below to skip all actual CI jobs in private repositories since
+  # our CI matrix is huge and would quickly burn through the free credits
+  # available for private repositories. To implement this jobs depend on this
+  # on, but this one only ever runs for public repositories.
```
awelzel:
```suggestion
  # one, but this one only ever runs for public repositories.
```

awelzel (approved):
LGTM, thanks.

PR #5573, opened 2026-06-12
Fix some -pedantic warnings

awelzel:
Mostly just tried `-pedantic` in the compile flags and seeing the stray `;` was annoying. Then did the remaining warnings within Zeek/binpac.

libkqueue and LightPcapNG light up a lot with `-pedantic` and I'm not sure it's generally useful to compile that way, so this is mostly a fly-by.


timwoj (approved):
The semicolons = 🤦🏼

PR #5560, opened 2026-06-10
Add fuzzer build/test and code coverage to CircleCi config

timwoj:
This is based on https://github.com/zeek/zeek/pull/5534, so that PR should be merged first. This PR adds support for coverage and building/testing the fuzzers to the CircleCI configuration. It adds a few couple new parameters to control this to the Linux builds. The new variable in the CI scripts is part of the `zeek` context in CircleCI.

PR #5576, opened 2026-06-12
CI: Add enable_build_artifacts parameter to run 'make install' during the build

timwoj:
Running `make install` and storing the artifacts was previously a side effect of running the benchmarks. In https://github.com/zeek/zeek/pull/5560, I removed passing `enable_benchmarks` for the spicy jobs which broke `zkg` being in the path and broke the ability to install the spicy analyzers.

This PR lifts installing the build and storing the artifacts to be part of a separate parameter. I'm not entirely sure why we ran the benchmarks on the Spicy jobs. I can definitely see storing the artifacts being an unnecessary side-effect of this new parameter as well, since that build artifact is only used for benchmarks. I could see reverting this and adding a parameter that simply causes `make install` to run. I'm curious about others' opinions on what makes the most sense here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -242,7 +242,11 @@ jobs:
         type: boolean
         default: false
       enable_benchmarks:
-        description: Flag for whether the benchmarks should run after building
+        description: Flag for whether the benchmarks should run after building. If this is true, the store_install_tarball parameter must be true as well.
+        type: boolean
+        default: false
+      store_install_tarball:
+        description: Flag for whether the build/install artifacts are stored. This flag also causes the build to install Zeek.
         type: boolean
         default: false
       enable_coverage:
@@ -250,7 +254,7 @@ jobs:
         type: boolean
         default: false
       install_spicy_analyzers:
-        description: Flag for whether to install the Spicy analyzers used for testing
+        description: Flag for whether to install the Spicy analyzers used for testing. If this is true, the store_install_tarball parameter must be true as well.
         type: boolean
         default: false
       # The command run by this parameter requires the ZEEK_CI_COMPILER environment variable
@@ -272,7 +276,6 @@ jobs:
       ZEEK_CI_RUNNER_OS: linux
       ZEEK_TEST_ULIMIT_OPEN_FILES: 256
       ZEEK_CIRCLE_EXTRA_ENV: << parameters.extra_env >>
-      ZEEK_CI_CREATE_ARTIFACT: << parameters.enable_benchmarks >>
       ZEEK_CI_CCACHE_PRUNE_SIZE: 700M
     docker:
       - image: << parameters.docker_image >>-<< parameters.ci_image_date >>
@@ -285,6 +288,16 @@ jobs:
           name: Set extra environment variables
           command: |
             echo "export ZEEK_CI_WORKING_DIR=${CIRCLE_WORKING_DIRECTORY}" >> "$BASH_ENV"
+
+            # This could be done in the environment block of the job definition, but
+            # we pass 0/1 for consistency with other environment variables in the rest
+            # of the config.
+            if [[ "<< parameters.store_install_tarball >>" == "true" ]]; then
+              echo "export ZEEK_CI_CREATE_INSTALL_TARBALL=1" >> "$BASH_ENV"
+            else
+              echo "export ZEEK_CI_CREATE_INSTALL_TARBALL=0" >> "$BASH_ENV"
+            fi
+
             IFS=$'\n'
             for line in ${ZEEK_CIRCLE_EXTRA_ENV}; do
               echo "export ${line}" >> "$BASH_ENV"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -272,7 +276,7 @@ jobs:
       ZEEK_CI_RUNNER_OS: linux
       ZEEK_TEST_ULIMIT_OPEN_FILES: 256
       ZEEK_CIRCLE_EXTRA_ENV: << parameters.extra_env >>
-      ZEEK_CI_CREATE_ARTIFACT: << parameters.enable_benchmarks >>
+      ZEEK_CI_CREATE_ARTIFACT: << parameters.enable_build_artifacts >>
```
awelzel: Might be an opportunity to straighten out the name.

IIUC, `tar -czf ${ZEEK_CI_WORKING_DIR}/build.tgz ${ZEEK_CI_WORKING_DIR}/install` is the crux.

So maybe tweak the env as `ZEEK_CI_CREATE_ARTIFACT_INSTALL_TARBALL` and `parameters.enable_install_tarball` or `parameters.store_install_tarball`.

(sorry, I'm getting a bit hung up on the name not actually saying what the artifact is and the other names not making it obvious that it's a tar of the installation directory rather than the build directory).

timwoj: > (sorry, I'm getting a bit hung up on the name not actually saying what the artifact is and the other names not making it obvious that it's a tar of the installation directory rather than the build directory).

No, I think that's totally fair. `store_install_tarball` is a good name, tbh.

ci/build.sh
awelzel:
```suggestion
    tar -czf ${ZEEK_CI_WORKING_DIR}/install.tgz ${ZEEK_CI_WORKING_DIR}/install
```

Should we just also do this and update the downstream consumer (only the benchmarker, right?). Otherwise I'm still hung up on naming :-D

timwoj: The `make install` part of that is needed by the spicy jobs. They don't need the tarball, but they need `zkg` in the path.

timwoj: But yeah, `install.tgz` is fine with me.

timwoj:
The latest version now uses `store_install_tarball` for the parameter name. It also:

- Renames the tarball from `build.tgz` to `install.tgz`
- Renames the env variable to from `ZEEK_CI_CREATE_ARTIFACT` to `ZEEK_CI_CREATE_INSTALL_TARBALL`
- Tars just the `install` directory instead of the entire path, since before it would have tarred up `/root/project/install`.

https://app.circleci.com/pipelines/circleci/JFr6soTf1AmP5bL3bFPRKn/uUFkHbvQy8B6MBNzPh61b/899/workflows/349195fd-e53c-4253-8219-edf260727058/jobs/7450/artifacts has an example of the tarball.

.circleci/config.yml
awelzel, line 1100: You added the store_install_tarball to the `-spicy` and `-spicy-head` tasks, but think you missed the default ubuntu-24.04 that build?

awelzel: I was confused not seeing any output about the changes in the Circle UI.

awelzel (approved):
Except for store_install_tarball on plain Ubuntu 24.04 LGTM. Thanks for the cleanup.

.circleci/config.yml
timwoj: It was missing because it's only needed on that build by the benchmarker which I haven't implemented yet. I'd add it when I get to that.

bbannier:
@timwoj This was merged even though it broke `Check/ubuntu-24.04-spicy`, see [here](https://github.com/zeek/zeek/actions/runs/27636903788/job/81726491422?pr=5576). I see this PR touched both `ci/build.sh` as well as the circle config; I suspect it needs a similar tweak for the GH actions workflow, could you please follow this up with a fix there?

Currently only the job `asan-sanitizer-zam` is known to fail pretty reliable due to timing issues, and I recall you have a PR somewhere to make it work better; all other jobs are expected to pass. This makes me wonder why you merged this with failures in both `weekly-current-gcc` (looks like non-determinism) and `ubuntu-24.04-spicy` (setup not adjusted).

timwoj:
The `zkg` pathing was already broken because the install was failing. It was like that on CircleCI also. I'll open a PR to fix that for both. The change I made in this PR for `ci/build.sh` was actually just reverting something CircleCI-specific (checking a variable for `true`, when it could just be `1` for both CI systems).

> Currently only the job asan-sanitizer-zam is known to fail pretty reliable due to timing issues, and I recall you have a PR somewhere to make it work better; all other jobs are expected to pass.

Yep, https://github.com/zeek/zeek/pull/5585 just merged for that. Hopefully that fixes asan ZAM.

> weekly-current-gcc

This was because of a newer version of sqlite3 on that image. https://github.com/zeek/zeek/pull/5587 should have fixed that.

timwoj:
> The change I made in this PR for `ci/build.sh` was actually just reverting something CircleCI-specific (checking a variable for `true`, when it could just be `1` for both CI systems).

Actually, never mind, you're right. I forgot I renamed that variable. I'll push a PR that fixes GH CI.

PR #5578, opened 2026-06-13
add libressl version check to lts 8.0.x zeek

ke5C2Fin:
When building LTS branch of Zeek using LibreSSL the build fails with current versions of LibreSSL.
This patch changes all checks for:
```
#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
#if ( OPENSSL_VERSION_NUMBER < 0x1010000fL ) || defined(LIBRESSL_VERSION_NUMBER)
#if ( OPENSSL_VERSION_NUMBER < 0x10002000L ) || defined(LIBRESSL_VERSION_NUMBER)
```
and changes the OR condition to:
```
|| (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x30900000L)
```
This will keep original checks for LibreSSL but also allow it to build with current LibreSSL.
I know these checks are already removed for branch 8.1 and 8.2 but I am currently needed to build the LTS version.
This enables for instance building LTS Zeek on FreeBSD 15.0 with LibreSSL.


src/file_analysis/analyzer/x509/functions.bif
```diff
@@ -65,7 +65,7 @@ X509* x509_get_ocsp_signer(const STACK_OF(X509)* certs,
 	const ASN1_OCTET_STRING* key  = nullptr;
 	const X509_NAME*         name = nullptr;
 
-#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || defined(LIBRESSL_VERSION_NUMBER)
+#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x30900000L)
```
timwoj: This can be done following the OPENSSL_VERSION_NUMBER version without the extra existence check. Same with the rest of the changes.

```suggestion
#if ( OPENSSL_VERSION_NUMBER < 0x10100000L ) || ( LIBRESSL_VERSION_NUMBER < 0x30900000L )
```

timwoj (changes requested):
I'll admit that we don't get PRs that go directly into release branches very often (or, ever, really), so this feels really strange to me. That said, I understand this one can't go into master first. I'll add it to the list for the next release just so it's tracked together, but I'm ok with merging this directly into the branch.

ke5C2Fin:
I applied your suggestion, @timwoj.
It built and ran fine.

timwoj (approved):
Thanks!

leres:
> and changes the OR condition to:
> 
> ```
> || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x30900000L)
> ```

Oops! The `defined(LIBRESSL_VERSION_NUMBER) &&` check is missing from the actual changes, this breaks building on FreeBSD with DEFAULT_VERSIONS=ssl=openssl35  (for example).

ke5C2Fin:
That was how I originally had it, but the zeek maintainer recommended the
change.
However when I build on FreeBSD with DEFAULT_VERSIONS=ssl=openssl35 the
original changes fail as we..
Both "|| (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER <
0x30900000L)" and "|| ( LIBRESSL_VERSION_NUMBER < 0x30900000L )" fail for
me using openssl35.

leres:
Here's what builds for me with  poudriere and DEFAULT_VERSIONS=ssl=openssl35:

[patch-src_file__analysis_analyzer_x509_OCSP.cc.txt](https://github.com/user-attachments/files/29420286/patch-src_file__analysis_analyzer_x509_OCSP.cc.txt)
[patch-src_file__analysis_analyzer_x509_X509.cc.txt](https://github.com/user-attachments/files/29420287/patch-src_file__analysis_analyzer_x509_X509.cc.txt)
[patch-src_file__analysis_analyzer_x509_X509.h.txt](https://github.com/user-attachments/files/29420288/patch-src_file__analysis_analyzer_x509_X509.h.txt)
[patch-src_file__analysis_analyzer_x509_functions.bif.txt](https://github.com/user-attachments/files/29420289/patch-src_file__analysis_analyzer_x509_functions.bif.txt)

leres:
When LIBRESSL_VERSION_NUMBER is undefined, `LIBRESSL_VERSION_NUMBER < 0x30900000L` is always true...

For example:
```
#include <stdio.h>
#include <stdlib.h>

int
main(int argc, char **argv)
{
#if TVERSION < 5
        puts("TVERSION < 5");
#endif
        exit(0);
}

```
produces:
```
zinc 100 % cc -Wall c.c && ./a.out
TVERSION < 5
```

ke5C2Fin:
I had a typo when I retested || instead of &&.
I put back my original changes and it builds with both libressl and openssl35.
I will create a new MR.

timwoj:
Did 8e05b013612607eea0ae6a86e00017f2d19d2f87 not already fix the missing checks?

leres:
> Did [8e05b01](https://github.com/zeek/zeek/commit/8e05b013612607eea0ae6a86e00017f2d19d2f87) not already fix the missing checks?

My mistake; I did not notice 8e05b01 when I tested the changes from #5578 causing @ke5C2Fin to create #5623.

I'm able to get what I need from f38fa4d and 8e05b01 (and indeed #5623 is not needed).

PR #5579, opened 2026-06-13
ftp: log login-only sessions missing from ftp.log

### LINKED ISSUE #5570: Missing Detection: FTP login-only sessions (USER/PASS/QUIT) absent from ftp.log
This was opened in the security repo via a report to security@zeek.org. After triage by the team we've decided that this is a missing feature and isn't a security issue. The ticket is being moved here to the public repo instead.

--------------

From @thehybridtechnician at Corelight:

> Hiya,
> 
> Zeek version affected: Zeek 8.2.0 (commit 04a76e91)
> 
> FTP sessions consisting only of USER, PASS, and QUIT commands — the typical pattern for a brute-force login attempt — produce no ftp.log entries. Zeek correctly identifies and tracks the connection (conn.log shows service=ftp), but logs nothing about which credentials were attempted or whether authentication succeeded.
> 
> Root cause
> scripts/base/protocols/ftp/main.zeek gatekeeps log writes with a logged_commands option:
> 
> option logged_commands = {
>     "APPE", "DELE", "RETR", "STOR", "STOU", "ACCT", "PORT", "PASV", "EPRT", "EPSV"
> };
> 
> The Log::write call fires only when s$command in logged_commands. Because USER, PASS, and QUIT are not in this set, a login-only session produces zero ftp.log rows.
> 
> Reproduce
> Attached is evidence.pcap. A complete FTP login session using USER/PASS/QUIT with no data transfer commands.
> What currently happens: Conn.log present.  No ftp.log is generated.
> What I expect to happen: ftp.log should be generated with commands recorded.
> 
> -- 
> Justice Renée Bovee
> Security Test Automation Engineer | Corelight

[evidence.pcap.gz](https://github.com/user-attachments/files/28292813/evidence.pcap.gz)

achill06:
FTP sessions consisting only of USER/PASS/QUIT commands produced no ftp.log entries. This affected visibility of brute-force login attempts which would be completely invisible to defenders.

Fix: add a logged_command_seen flag to FTP::Info. When finalize_ftp runs, if USER was seen but no logged_commands were issued, write a summary log entry with command set to '<session>' so the session is visible in ftp.log.

Fixes #5570


scripts/base/protocols/ftp/info.zeek
```diff
@@ -72,6 +72,8 @@ export {
 
 		## Determines if the password will be captured for this request.
 		capture_password:   bool &default=default_capture_password;
+        ## Used to detect login-only sessions that would otherwise produce no ftp.log entries.
```
evantypanski: nit: I would take out the `login-only` part. This actually applies to all sessions: Essentially it would now always log *something* even if there were no loggable commands, which I feel like is a more correct approach here. So:


```suggestion
		## Used to detect sessions that would otherwise produce no ftp.log entries.
```

(also in the suggestion swaps it to tabs for consistency)

scripts/base/protocols/ftp/main.zeek
```diff
@@ -485,4 +488,16 @@ hook finalize_ftp(c: connection)
...
 		}
+
+	if ( c$ftp?$user && ! c$ftp$logged_command_seen )
+		{
+		local s: Info = c$ftp;
+		s$ts = network_time();
+		s$command = "<session>";
+
+		if ( s?$password && should_hide_password(s) )
+			s$password = "<hidden>";
```
evantypanski: with the reframing I propose this wouldn't be necessary, I would say this should be removed.

evantypanski (commented):
Thanks for the contribution, I have a few general comments:

1) I actually think this is pretty cool as "nothing was logged, but a user was noted" kind of log. So this is more general than the login brute force attempt, instead this `<session>` moniker is used when any session is detected, but otherwise nothing is logged. With that being said, I think it makes sense to allow disabling this behavior with a new option: like an exported `option report_unlogged_sessions = T;` or something. Then check that before logging `<session>`.
2) This change needs a test. You can use the pcap in the original issue (just put it in `testing/btest/Traces/ftp`) and follow similar tests. With point 1 adding an option, it would be nice to test that when `report_unlogged_sessions=F` then no FTP log exists, while default behavior means it does exist. Something like (untested):

```
# @TEST-EXEC: zeek -br $TRACES/ftp/only-login.pcap %INPUT FTP::report_unlogged_sessions=F
# @TEST-EXEC: test ! -e ftp.log
# @TEST-EXEC: zeek -br $TRACES/ftp/only-login.pcap %INPUT
# @TEST-EXEC: btest-diff ftp.log
```

Alternatively you can craft your own pcap and test as well. You probably need to `@load` the FTP scripts since I included `-b` in the Zeek flags.

3) We merged an [AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md) recently: just for our sake, to what extent was the code change here generated by an LLM?

achill06:
The fix was developed with AI assistance for guidance on the codebase structure and approach, but I wrote and tested every change myself locally, verified it against real pcap captures.

achill06:
Regarding the password field in the `<session>` log entry, should it rely on the existing capture_password behavior (current behavior, shows `<hidden>` by default), or should it be omitted entirely from the session summary?

achill06:
I wasn't able to download the evidence.pcap.gz from the original issue; the link returns 'Not Found'. Would it be acceptable to use the existing testing/btest/Traces/ftp/bruteforce.pcap, which captures the same scenario? I can add the btest once confirmed.

evantypanski:
> Would it be acceptable to use the existing testing/btest/Traces/ftp/bruteforce.pcap, which captures the same scenario?

sure, that pcap works. I actually didn't see we indeed have a detection for this, so you could make a combined test using that pcap. That pcap is actually completely unused, so this would help our test coverage a bit, not sure why it's there in the first place.

Basically just a test in `testing/btest/scripts/policy/protocols/ftp/bruteforce.zeek` which loads that bruteforce policy script (`protocols/ftp/detect-bruteforcing`) and baselines both ftp.log (which will have a bunch of these new log entries) and `notice.log` (from the detection). That feels right to me, though others can yell if they disagree ;)

BTW you'll need to update baselines for some other tests, namely `coverage.record-fields` which will print the new `logged_command_seen`.

> Regarding the password field in the <session> log entry, should it rely on the existing capture_password behavior (current behavior, shows <hidden> by default), or should it be omitted entirely from the session summary?

Yeah just let it do what's there already, there's no need to omit or otherwise change it in my opinion

--------------------

I like the combination of this change, which will always log to ftp.log in an FTP session, and that detection script.

Then the line in `info.zeek` still needs swapped to tabs, not spaces

achill06:
Updated; fixed tab issue in info.zeek, added btest at testing/btest/scripts/policy/protocols/ftp/bruteforce.zeek using bruteforce.pcap, and updated the coverage baseline. Couldn't generate ftp.log and notice.log baselines locally since I'm on binary packages. Can CI generate them, or can you point me to the right approach?

scripts/base/protocols/ftp/main.zeek
```diff
@@ -485,4 +492,12 @@ hook finalize_ftp(c: connection)
 		c$ftp$cmdarg = cmdarg;
 		ftp_message(c);
 		}
+
+	if ( c$ftp?$user && ! c$ftp$logged_command_seen && report_unlogged_sessions )
+		{
+		local s: Info = c$ftp;
+		s$ts = network_time();
```
evantypanski: I think this should be removed, it'd already be set if we saw at least one command, right? Putting `network_time()` in the finalize hook feels like the wrong place

evantypanski (commented):
The baselines are updated with `btest` which should be shipped alongside Zeek. So if you build from source it'll be in `<install-dir>/bin/`. Then you can use `btest -U` to update baselines (eg in `testing/btest` you would do `btest -U scripts/policy/protocols/ftp/bruteforce.zeek` and see new files in `Baseline`). We're adding documentation on that in #5583 - if you look through the changes there on testing (in `btest.rst`) that'll give you a sense for how to update baselines and whatnot.

I can just do this manually too, but it's necessary for a complete contribution so I wouldn't want to unless I had to

Feel free to ask questions or go on Slack in the development channel, others could chime in too.

lczllx:
Apologies for the cross-reference here. My PR description mistakenly included a reference to the same issue (#5580), which triggered this mention. It's been fixed now. Sorry for any confusion.

achill06:
All done
* removed `network_time()`
*  added the btest baselines

achill06:
Hi, are there any issues now?

evantypanski:
> Hi, are there any issues now?

sorry I was busy today, I'll get back to this in the next day or two :) seems like you addressed my concerns so it should be okay but I'll do one last check soon

scripts/base/protocols/ftp/main.zeek
```diff
@@ -485,4 +492,11 @@ hook finalize_ftp(c: connection)
 		c$ftp$cmdarg = cmdarg;
 		ftp_message(c);
 		}
+
+	if ( c$ftp?$user && ! c$ftp$logged_command_seen && report_unlogged_sessions )
+		{
+		local s: Info = c$ftp;
+		s$command = "<session>";
```
awelzel: Seems reasonable. I wonder if we could log the last observed command here instead of the synthetic `<session>` one? That'd be `QUIT` for the example and probably a defend-able notion given the `reply_code` and `reply_msg` columns?  This is mostly about having some signal in the ftp.log, correct?

Maybe the NEWS entry could be something like "Previously, Zeek wouldn't log FTP sessions that only had USER/PASS/QUIT commands in `ftp.log` at all. This has been changed to log at least one command (the last one) when no other command was logged for the session, even if that command isn't part of `FTP::logged_commands`.  You can revert to the previous behavior by setting `FTP::log_at_least_one_command=F` (might need a better name - not sure about "report_unlogged_sessions", but naming, so..., hmm... maybe `FTP::force_session_log` or something with "ensure"?).

No overly strong opinion on this though, just thought I'd chime in.

evantypanski: I opted against suggesting the last command for a couple reasons:

1) It's probably going to be `QUIT` which itself holds basically 0 value IMO
2) It seems a bit confusing to log commands even though the user has not added them to `logged_commands`. Feels like that could cause a goose chase to look for those commands if they don't see this new logic, but `<session>` looks different enough that people would look for that particular string

... With that being said, I think it makes describing the feature easier and it's more in line with the logs, so I'd be inclined to agree. That would mean:

a) Don't change the `s$command` var here
b) Rename the variable to `log_at_least_one_command` rather than `report_unlogged_sessions`

Those two together make sense to me. I actually like `log_at_least_one_command` because it's extremely, extremely clear.

@achill06 would you want to do those changes? Perfectly fine if not, I can just do it in a followup if not, you've done plenty here.

achill06: Well, I have done the suggested changes, but I am confused about where in the NEWS I should add it.

evantypanski: The `NEWS` entry I think was directed at whoever merges this, you don't need to worry about it.

Assuming CI passes it looks good to me & I'll merge, thanks!

achill06: Thank you for the help

evantypanski (approved):
thanks!

Issue #5407, opened 2026-04-27
Zeek values can cycle & loop indefinitely


evantypanski:
I found this while challenging some assumptions with claude for #5338 

There are a couple of ways that you can make cycles in Zeek, and both result in a segfault from stack overflow.

The first is with `any`, like this:

```
type Node: record {
    val:  count;
    data: any &optional;
};

global a: Node;
global b: Node;

event zeek_init() {
	a = Node($val=1);
	b = Node($val=2);
	a$data = b;
	b$data = a;
	print a;
}
```

And the second is with `redef`:

```
type A: record { val: count; };
type B: record { a: A &optional; };
redef record A += { b: B &optional; };

event zeek_init() {
    local a: A = [$val=1];
    local b: B = [$a=a];
    a$b = b;
    print a;
}
```

Both of these segfault.

There are actually a couple of solutions here. The one I like more is to detect cycles and reject them. I would like that statically, but probably not possible with `any`. The second is pretty targeted: in `ValDescribe` & friends, just include a `visited` set if we've seen it before, then print `cycle` or something if so.

Rejecting this also makes it more straightforward to refcount for any future work on the interpreter (wink wink)

(extra note, I tried to make a resource leak by making a self-referential closure like that, but they [unconditionally unref managed captures](https://github.com/zeek/zeek/blob/40a481823c6035003c3875fdd8fc96df53eec01c/src/Func.cc#L303-L309))

0xxon:
Even shorter reproducer with just one type:

```
type A: record { val: count; };
redef record A += { a: A &optional; };

event zeek_init() {
    local a: A = [$val=1];
    a$a = a;
    print a;
}
```

or

```
type A: record { val: any &optional; };

event zeek_init() {
    local a : A;
    a$val = a;
    print a;
}
```

Fun - thank you for finding this :)

PR #5210, opened 2026-02-08
Add CLAUDE.md for AI-assisted development guidance

isesandja:
Provides comprehensive guidance for Claude Code when working with this repository, including build instructions, architecture overview, testing workflow, and development practices. The documentation emphasizes:

Ethical and authorized use of Zeek for network visibility

Threat modeling from an attacker’s perspective to improve detection

Common classes of network‑level abuse (evasion, malformed traffic, resource exhaustion)

Defensive analysis techniques rather than exploitation

Zeek is frequently used in security research, blue‑team operations, and controlled red‑team exercises. Clear guidance helps ensure:

Responsible and legal usage

More accurate interpretation of network telemetry

Better detection engineering and threat awareness

Reduced risk of misuse or misunderstanding of Zeek’s capabilities

Scope

Documentation only

No changes to analyzers, scripts, or core logic

No offensive exploitation guidance

Security, Network Analysis, and Ethical Use

Zeek is designed for deep network visibility and defensive security analysis. Users should approach protocol analysis with an understanding that all observed network traffic may be attacker‑controlled and intentionally malformed.

Effective analysis includes:

Identifying evasion techniques that exploit protocol edge cases

Recognizing malformed or ambiguous traffic intended to bypass detection

Monitoring for abnormal resource usage patterns that may indicate abuse or denial‑of‑service attempts

Zeek should be used only in environments where monitoring is authorized and legal. Analysis and testing should focus on improving detection, resilience, and operational security rather than exploiting live systems.          name: Security Analysis Build

on:
  pull_request:
  push:

jobs:
  security-analysis:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4

      - name: Install dependencies
        run: |
          sudo apt-get update
          sudo apt-get install -y cmake make g++ libpcap-dev

      - name: Build Zeek
        run: |
          mkdir build
          cd build
          cmake ..
          make -j$(nproc)

      - name: Defensive network analysis test
        run: |
          ./build/src/zeek -r test/pcaps/malformed.pcap || true                                     ### Security-Focused Builds

Zeek can be built and tested with an emphasis on defensive network analysis by validating analyzer behavior against malformed or adversarial traffic captures.

Recommended workflow:
1. Build Zeek with default settings
2. Replay known edge-case or malformed pcaps
3. Verify analyzers fail safely without crashes or silent state corruption

PR #5583, opened 2026-06-15
Add AGENTS.md

awelzel:
Opening as draft. Mostly to start a discussion if we want to do this.

---

LLMs and agents are here to stay. After chatting a bit with Evan, here's a proposal to steer these systems into a direction we'd like them to go
 for adding new btests etc, rather than repeating ourselves on PRs.

It's not exhaustive and kind of lets write it once and see what others think. I've used this for a few days locally and the creating PCAPs section is helpful, because otherwise you need to instruct the LLM over and over again about the conventions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,42 @@
+# AGENTS.md
+
+File with instructions for AI coding agents contributing to Zeek.
+
+## Building and Running Zeek
+
+Read ./doc/advanced/devel/hacking.rst first for instructions how to
+configure, build and execute Zeek for development. You can find the
+required build dependencies in ./doc/building-from-source.rst.
+
+## Testing
+
+New or changed functionality must come with tests.
+
+You MUST read ./doc/advanced/devel/btest.rst and understand
+testing conventions thoroughly.
+
+## Security Reporting
+
+You MUST take the contents of ./doc/security-considerations.rst into
+account when submitting security issues.
+
+Accompany reports for security issues in analyzers with reproducers in the
+form of test cases and PCAPs as outlined under Testing. Always validate
+the test case reproduces in a clean environment before submitting a report.
+
+## Pull Request (PR) Submissions
+
+PR titles and descriptions MUST be kept to a bare minimum. Provide only a
+one-sentence summary, then immediately leave the remainder to the user, like:
+
+    Fix segfault when recursively printing values.
+
+    (This was AI-generated; remainder to be filled in by user)
+
+Do NOT write out reproducer snippets, file-change summaries, or any detailed
+explanation in PR descriptions. Leave all of that for the user to fill in.
+Ensure the user understands every step. Always disclose AI usage.
+
+## AI Disclosure Requirement
+
+AI usage of any form MUST be disclosed as outlined in ./AI_POLICY.md.
--- a/doc/advanced/devel/btest.rst
+++ b/doc/advanced/devel/btest.rst
@@ -1,0 +1,180 @@
+.. _devel-btest:
+
+
+Testing with BTest
+==================
+
+Zeek's main regression tests are `btest <https://github.com/zeek/btest>`_-based.
+
+Testing protocol parsers (also called analyzers) and scripting behavior is
+usually done by having Zeek read packet traces (PCAP files) to produce logs.
+These logs are then baselined. Baselines are stored in the
+testing/btest/Baseline directory and checked into the repository.
+
+There's also a Baseline.zam directory for alternative baselines when
+running Zeek with :ref:`ZAM <zam>` enabled.
+
+All btests run in CI on various platforms with and without ZAM enabled.
+
+
+Packet Traces (PCAP Files)
+--------------------------
+
+We store packet traces in testing/btest/Traces/. The README file represents an
+index where PCAPs came from or how they were created to keep a bit of lineage
+available. PCAP filenames always end with ``.pcap`` or ``.pcapng``. Usually PCAP
+files are stored uncompressed, except for some larger but highly compressible examples.
+
+There are generally two approaches to create new packet traces in isolation
+if you cannot share a packet capture from a production network.
+Either install and run the software yourself and capture the traffic
+with ``tcpdump`` in a lab or virtual environment, or
+create a Python script that produces packet traces using
+`Scapy <https://github.com/secdev/scapy>`_. In 2026, LLM agents are very
+effective for the latter. Including packet traces from real software or
+actual production networks is more realistic and if available, preferred
+over Scapy-generated traces. For edge case testing of parsers, Scapy-generated,
+Scapy-edited, or hex-edited capture files are all fair game. Keep a note in
+the README what was done to create a certain trace.
+
+The testing/btest/Traces directory is mostly structured by protocol. When
+adding a Scapy-generated trace, ``<name>.pcap``, put a ``<name>.pcap.py``
+file next to it. Running the Python script should generate the ``<name>.pcap``
+file next to ``<name>.pcap.py`` reproducibly. Running the script a year later
+should produce the exact same result.
+
+Multi-word PCAP names should be all lower-case and use dashes for separation.
+The PCAP's name should include the thing or scenario being tested. For example,
+``ftp/ftp-with-numbers-in-filename.pcap`` or ``ssh/server-pre-banner-data.pcap``.
+Try to keep PCAP files to a few kilobytes in size, 50KB or more should be
+an exception. Including PCAPs with valid checksums is preferred. Use ``tcprewrite``
+to correct them if needed.
+
+Creating new Tests
+------------------
+
+Tests are stored in the testing/btest/ directory. There are subdirectories for
+individual Zeek components and protocols. Recursively looking at directory
+and test names should help you orient. Older or more integration-like
+tests end with ``.test`` or ``.sh``. Most tests using packet traces and
+baselining or testing Zeek language features should end with ``.zeek``
+for better language server support.
+
+Every test should have a commented ``@TEST-DOC`` line at the top describing
+what is being tested, followed by some ``@TEST-EXEC`` lines with the actual
+commands.
+
+.. code-block:: shell
+
+    # @TEST-DOC: Verify a basic HTTP GET request.
+    #
+    # @TEST-EXEC: zeek -b -r $TRACES/http/get.pcap %INPUT
+    #
+    # @TEST-EXEC: btest-diff-cut -m uid service history conn.log
+    # @TEST-EXEC: btest-diff-cut -m http.log
+
+    @load base/protocols/conn
+    @load base/protocols/http
+
+In tests, prefer to use ``zeek -b`` to invoke Zeek in bare mode to reduce Zeek's
+initialization time and reduce the potential for unintended side-effects. Load
+all required scripts and packages via explicit :zeek:see:`@load` directives
+within the test's content. The magic ``%INPUT`` variable expands to a filename
+containing the content of the test (see the BTest documentation for more details).
+
+The ``$TRACES`` environment variable is set in ``testing/btest/btest.cfg``
+and expands to ``testing/btest/Traces``. That's all there is to it.
+
+For baseline testing, ``@TEST-EXEC: btest-diff`` has been used historically,
+with ``btest-diff-cut`` being a recent addition that allows to easily create
+smaller baselines by selecting relevant columns to diff.
+Prefer to use the ``btest-diff-cut`` helper script and only include columns
+in the baselines that are relevant to the test. This hardens your test against
+irrelevant baseline deviations introduced by unrelated future changes to Zeek.
+Avoid the verbose TSV header by passing ``-m`` to ``btest-diff`` or ``btest-diff-cut``.
+
+To baseline all columns of ``path.log``, use ``btest-diff-cut -m path.log``.
+A full baseline is recommended for protocol logs in tests that are specific
+to that protocol. For example, in HTTP tests, baseline the full ``http.log``,
+but only select certain columns from ``conn.log`` as shown above.
+
+The conn.log's uid service and history columns are generally interesting,
+even if the test does not strictly need all of them. Checksum errors are
+quickly recognized as `c` or `C` in the history column. The service entry
+is also useful to verify a protocol is present (and wasn't removed due
+to an analyzer violation).
+
+If you baseline weird.log, minimally include the columns uid, name, addl and
+source. Include ``@load base/frameworks/notice/weird`` in the test. By default,
+weirds are not logged when running in bare mode.
+
+When testing very specific analyzer confirmation and violation behavior, load
+the ``frameworks/analyzer/debug-logging.zeek`` and baseline the created
+``analyzer_debug.log`` using ``btest-diff-cut -m``:
+
+.. code-block:: shell
+
+    # @TEST-DOC: Verify the SSH analyzer's confirmation and violation behavior. Regression test for #1234.
+    #
+    # @TEST-EXEC: zeek -r $TRACES/ssh/ssh.server-side-half-duplex.pcap %INPUT
+    #
+    # @TEST-EXEC: btest-diff-cut -m ssh.log
+    # @TEST-EXEC: btest-diff-cut -m uid service history conn.log
+    # @TEST-EXEC: btest-diff-cut -m analyzer_debug.log
+
+    @load base/protocols/ssh
+    @load frameworks/analyzer/debug-logging.zeek
+
+If you explicitly want to test for no weirds or no analyzer violations,
+use ``test ! -f weird.log``:
+
+.. code-block:: shell
+
+    # ...
+    # @TEST-EXEC: btest-diff-cut -m uid service history conn.log
+    #
+    # @TEST-EXEC: test ! -f weird.log
+    # @TEST-EXEC: test ! -f analyzer.log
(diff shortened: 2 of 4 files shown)


bbannier:
I'd be 👎 here since I see no reason why this needs to be some fuzzy tool configuration instead of something crystal clear in in-tree docs.

evantypanski:
I think that we do need better internals docs. The best we have is some stuff in the Wiki, which does not do a good job of capturing the current state of Zeek. So in a sense this is just less-polished documentation.

We could just make a section on internals in the docs and put this information? Under "Advanced Topics > Developer Guides"? Then use the AGENTS.md file to point to each of these. That way it's human-first documentation, but machines can just use the clear human docs, and we get good internal documentation out of it anyway. That's something I wish we had!

(And FWIW it would be nice if the LLMs read the docs before doing changes, but alas they do not. They're like humans in that way. Most of the time they read AGENTS.md though, I think.)

Though I actually think the creating tests (specifically parts on pcaps) section *does* belong in a file specifically for LLMs - we specifically carve out exceptions to AI generated content for pcaps in the AI policy. It is indeed a general policy, but it bears repeating to stay in line with the policy that's in this repo, in my opinion.

bbannier:
FTR, I closed a [similar PR](https://github.com/zeek/zeek/pull/5210) before with the reason that we do not include tool-specific files with Zeek, at least not if we do not exercise them. IMO these agent-targeted documentation files fall under that, and unless we see a big market contraction soon we'll have a ton of these.

awelzel:
> that we do not include tool-specific files with Zeek

Without having looked all that deeply into it (the CLAUDE.md symlink to AGENTS.md is something I saw in the systemd repo), the `AGENTS.md` file isn't tool specific, or at least picked up by quite a few tools (Codex, Claude, OpenCode) and there is https://agents.md where it's describe as something like a standard. Maybe I'm just falling for the hype.

0xxon:
I kind of like this. AGENTS.md by now is a pretty well established standard that's picked up by a lot of different tools - and it obviously immediately useful for others to have this in the repo.

I would kind of go even further and say that we also should have a security policy or similar where we outline our threat model, specifically what's outside of it, and refer to that from agents.md too - this seems to work for other projects to guide the kind of reports they get.

awelzel:
I've pushed a fixup that splits it out and creates `advanced/devel/btest.rst` and removed most things in AGENTS.md file now to just point there. Can pick up `btest.rst` also without the AGENTS.md addition now.

ckreibich:
I'm in favor of having an AGENTS.md. I've mentioned this before but I like that this (hopefully) gives us a way to control the most annoying aspects of poor-quality agent PRs (and good ones too!) in a more focused way than, say, PR templates, which I've never really seen work out well because they tend to annoy humans.

Strong thumbs for moving the bulk of the testing details into the docs. It's definitely useful there, for humans too! I don't have empirical data about their effectiveness but the AGENTS.md files I see held up as good examples are very concise bullet lists, even more so than the current one, with the sections being the main guidelines.

I could see us add a demand to include reproducers for security issues, follow our coding style,  to keep PRs rebased on top of master, etc.

I would not include a CLAUDE.md. It's easy enough to configure Claude to look at AGENTS.md as well, and hey maybe one of these years they'll work on [the ticket](https://github.com/anthropics/claude-code/issues/6235).

awelzel:
I've force-pushed this branch after also adding a `hacking.rst` with a few tips and hints about ccache, lld, debug streams, jemalloc and links to other sections. The `btest.rst` is also in. Then reduced `AGENTS.md`. So marking this ready for review, seems most that chimed in would be okay to getting something in.

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,124 @@
...
+from ``conn.log`` as shown above.
+
+The conn.log's uid service and history columns are generally interesting,
+even if the test does not strictly need all of them. Checksum errors are
+quickly recognized as `c` or `C` in the history column. The service
+entry is also useful to verify a protocol is present (and wasn't removed due
+to an analyzer violation).
+
+If you baseline weird.log, minimally include the columns uid, name, addl and
+source. If you explicitly want to test for no weirds or no analyzer violations,
```
timwoj: Should this mention using `@load frameworks/analyzer/debug-logging` to catch, for example, binpac exceptions?

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+## Testing
+
+New or changed functionality must come with tests.
+
+You MUST read ./doc/advanced/devel/btest.rst and understand
+testing conventions thoroughly.
+
+## Security Reporting
+
+Thoroughly read ./doc/security-considerations.rst to
```
timwoj: Did this line get cut off?

AGENTS.md
```diff
@@ -0,0 +1,28 @@
+# AGENTS.md
+
+File with instructions for AI coding agents contributing to Zeek.
+
+## Building and Running Zeek
+
+Read ./doc/advanced/devel/hacking.rst first for instructions how to
+configure, build and execute Zeek for development. You find the
```
timwoj:
```suggestion
configure, build and execute Zeek for development. You can find the
```

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+New or changed functionality must come with tests.
+
+You MUST read ./doc/advanced/devel/btest.rst and understand
+testing conventions thoroughly.
+
+## Security Reporting
+
+Thoroughly read ./doc/security-considerations.rst to
+
+Accompany reports for security reports in analyzers with reproducers in
```
timwoj:
```suggestion
Accompany reports for security issues in analyzers with reproducers in
```

timwoj (changes requested):

initconf:
Only think I have to add here - WOW ! you guys are addressing the right things !!

scantlina:
i think this is a good start.

in my experience, for agentic workflows, getting efficiency is more than the AGENTS.md - it's letting the agent work in a repository that contains the tools the agent needs to do the work you want it to do.  to evan's earlier point, the first thing i do in my zeek agentic work is have it pull in the documentation for local reference.

what makes this difficult is that every agent harness works just a little bit differently - so while i think i could provide my zeek repo structure that i use for zeek dev with codex, i don't know how much it would benefit someone using e.g. claude.  this starts to dovetail into the points benjamin raised, i.e. how much should zeek worry about bootstrapping tool-specific workflows.

tim has mentioned the number of AI reports coming in, and i think the case could be made that making development repo templates for frontier models could have a downstream benefit to zeek devs.

tl;dr: i think making codex and claude repositories that can guide an agent to provide PRs in a way that makes review by zeek devs easier is worth considering.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+the form of test cases and PCAPs as outlined under Testing. Always
+validate the test case and PCAP reproduces before submitting a report.
+
+## AI Disclosure Requirement
+
+AI usage of any form MUST be disclosed as outlined in ./AI_POLICY.md.
```
evantypanski, line 42: So I focused on this point, getting Claude to generate a fix for #5407 then seeing what it gives for a PR description. I found that it did not disclose any AI usage unless explicitly asked (which, no one would do) so I believe we could enumerate an AI disclosure policy here.

Also, I like succinct PR descriptions, and really hate how verbose LLMs tend to be. So after some testing I liked adding this to the AGENTS.md file here:

```

- PR titles and descriptions MUST be kept to a bare minimum. Provide only a
  one-sentence summary, then immediately leave the remainder to the human, like:

  Fix segfault when recursively printing values.

  (This was AI-generated; remainder to be filled in by user)

- Do NOT write out reproducer snippets, file-change summaries, or any detailed
  explanation in PR descriptions. Leave all of that for the user to fill in.
- Ensure the user understands every step. Always disclose AI usage.
```

My goal is to lower the giant descriptions for small changes while also having a clear "this was AI-generated" disclaimer. Here's an example output for the fix for that issue:

<details>
  <summary>LLM PR description</summary>

Per AGENTS.md, here's the minimal AI-policy-compliant description:

  ---
  Title: Fix stack overflow when printing cyclic record values

  Body:

  Fix stack overflow when printing cyclic record values.

  (This was AI-generated; remainder to be filled in by user)

  ---
  You should fill in the rest — reproducer scripts, explanation of the two cycle patterns, and any
  follow-up notes (e.g. whether VectorVal/TableVal via any should be handled separately).
</details>

thoughts? I don't mean to "quiz" the user or anything, but I like the smaller description plus clearly telling the agent to disclose usage and how.

doc/advanced/devel/btest.rst
awelzel: Yep, I've added a separate paragraph for that one. Thanks.

AGENTS.md
awelzel: I've folded this. I want to steer clear of doing more "prompt engineering" here, because I'm not sure I like the idea, but that seemed useful and the result looked good.

I removed the bullets and made it free text. I really hope that doesn't make a difference. Like short sections a bit more.

awelzel:
I've pushed a rather big fixup after going through the docs again, but I'll stop now.

@ckreibich - mind looking this through and signing off or pick it up for a rework?

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+
+Zeek's main regression tests are `btest <https://github.com/zeek/btest>`_-based.
+
+Testing protocol parsers (also called analyzers) and scripting behavior is
+usually done by having Zeek read packet traces (PCAP files) to produce logs.
+These logs are then baselined. Baselines are stored in the
+testing/btest/Baseline directory and checked into the repository.
+
+There's also a Baseline.zam directory for alternative baselines when
+running tests with :ref:`ZAM <zam>` enabled.
```
ckreibich:
```suggestion
running Zeek with :ref:`ZAM <zam>` enabled.
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+
+All btests run in CI on various platforms with and without ZAM enabled.
+
+
+Packet Traces (PCAP Files)
+--------------------------
+
+We store packet traces in testing/btest/Traces/. The README file represents an
+index where PCAPs came from or how they were created to keep a bit of lineage
+available. PCAP filenames always end with a ``.pcap``. Usually PCAP files are
```
ckreibich: Would be good to cover .pcapng here too, since we now do actually have those.

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+stored uncompressed, except for some larger but highly compressible examples.
+
+There's generally two approaches to create new packet traces in isolation
```
ckreibich:
```suggestion
There are generally two approaches to create new packet traces in isolation
```
(It's a losing battle these days but I still struggle with this in writing. :-)

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+if you cannot share a packet capture from a production network.
+Either, install and run the software yourself and capture the traffic
+with ``tcpdump`` in a lab or virtual environment.
+Alternatively, create a Python script that produces packet traces using
```
ckreibich:
```suggestion
Either install and run the software yourself and capture the traffic
with ``tcpdump`` in a lab or virtual environment,
or create a Python script that produces packet traces using
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+`Scapy <https://github.com/secdev/scapy>`_. In 2026, LLM agents are very
+effective for the latter. Including packet traces from real software or
+actual production networks is more realistic and if available, preferred
+over Scapy-generated traces. For edge case testing of parsers, Scapy-generated
```
ckreibich:
```suggestion
over Scapy-generated traces. For edge case testing of parsers, Scapy-generated,
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+The testing/btest/Traces directory is mostly structured by protocol. When
+adding a Scapy-generated trace, ``<name>.pcap``, put a ``<name>.pcap.py``
+file next to it. Running the Python script should generate the ``<name>.pcap``
+file next to ``<name>.pcap.py`` reproducibly. Running the script a year later
+should produce the exact same result.
+
+Multi-word PCAP names should be all lower-case and use dashes for separation.
+The PCAP's name should include the thing or scenario being tested. For example,
+``ftp/ftp-with-numbers-in-filename.pcap`` or ``ssh/server-pre-banner-data.pcap``.
+
```
ckreibich, line 52: Two good additions here would be a reminder to keep the pcaps as small as possible (maybe advise well below a megabyte) and with checksums corrected.

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+Creating new Tests
+------------------
+
+Tests are stored in the testing/btest/ directory. There's subdirectories for
```
ckreibich:
```suggestion
Tests are stored in the testing/btest/ directory. There are subdirectories for
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+    #
+    # @TEST-EXEC: zeek -b -r $TRACES/http/get.pcap %INPUT
+    #
+    # @TEST-EXEC: btest-diff-cut -m uid service history conn.log
+    # @TEST-EXEC: btest-diff-cut -m http.log
+
+    @load base/protocols/conn
+    @load base/protocols/http
+
+In tests, prefer to use ``zeek -b`` to invoke Zeek in bare mode. Load all
```
ckreibich:
```suggestion
In tests, prefer to use ``zeek -b`` to invoke Zeek in bare mode to reduce Zeek's
initialization time and reduce the potential for unintended side-effects. Load all
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+containing the content of the test - see the btest documentation for more details).
+
+The ``$TRACES`` environment variable is set in ``testing/btest/btest.cfg``
+and expands to ``testing/btest/Traces``. That's all there is to it.
+
+For baseline testing, ``@TEST-EXEC: btest-diff`` has been used historically,
+with ``btest-diff-cut`` being a recent addition that allows to easily create
+smaller baselines by selecting relevant columns to diff.
+Prefer to use the ``btest-diff-cut`` helper script and only include columns
+in the baselines that are relevant to the test. Avoid the verbose TSV header
```
ckreibich:
```suggestion
in the baselines that are relevant to the test. This hardens your test against
irrelevant baseline deviations introduced by unrelated future changes to Zeek.
Avoid the verbose TSV header
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+    # @TEST-EXEC: test ! -f weird.log
+    # @TEST-EXEC: test ! -f analyzer.log
+
+    @load base/protocols/conn
+    @load base/protocols/http
+    @load base/frameworks/notice/weird
+
+The ``btest-diff`` and ``btest-diff-cut`` commands support ``TEST_DIFF_CANONFIER``
+to canonify a baseline. It defaults to ``testing/scripts/diff-canonifier`` which
+is set in ``testing/btest/btest.cfg``.
```
ckreibich:
```suggestion
to *canonify* a baseline, i.e., normalize content that will change from invocation to
invocation. It defaults to ``testing/scripts/diff-canonifier``, which is set in
``testing/btest/btest.cfg`` and canonifies timestamps to a ``XXXX.XXX`` pattern.
```

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
...
+environment concept and check ``testing/btest/btest.cfg`` for the extra
+environment variables and settings used for running tests under ZAM.
+
+
+Non-PCAP Tests
+--------------
+
+Note that not all tests are packet trace based. Many of the cluster tests instead
+use ``btest-bg-run``, ``btest-bg-wait`` and remote events for testing. Review them
+first if you work on cluster functionality. There's also the ``./bifs`` and
```
ckreibich:
```suggestion
first if you work on cluster functionality. There are also the ``./bifs`` and
```

doc/advanced/devel/hacking.rst
```diff
@@ -0,0 +1,176 @@
...
+
+ccache
+^^^^^^
+
+Zeek takes a long time to build. We recommend using `ccache <https://ccache.dev/>`_
+for local development to reuse compilation artifacts for improved iteration speeds:
+
+.. code-block:: shell
+
+    # ./configure --generator=Ninja --ccache
```
ckreibich, line 77: (Or by prioritizing `/usr/lib64/ccache` — or whatever it is on your system — in `PATH` so the compiler automatically becomes ccache-enabled.)

doc/advanced/devel/btest.rst
```diff
@@ -0,0 +1,173 @@
+.. _devel-btest:
+
+
+Testing with Btest
```
ckreibich:
```suggestion
Testing with BTest
```
:-)

ckreibich (approved):
This is great, thanks Arne. My edits are all minor and I leave them up to you — they're mostly suggested additions/clarifications, or the reasons behind our recommendations.

PR #5584, opened 2026-06-15
Clamp some lengths in the NTP extension parsing

timwoj:
This PR lifts the length-clamping bits out of https://github.com/zeek/zeek/pull/5549 and adds a btest that just checks for a BinPAC exception to guard against the crash reported by OSS-Fuzz.

PR #5588, opened 2026-06-16
Allow unsigned constants in binpac code

timwoj:
https://github.com/zeek/zeek/pull/5584 added some additional checks to the NTP analyzer. Unfortunately, on newer compilers, the math operations cause the type to be converted from unsigned to signed, and results in the following warnings:

```
/root/project/build/src/analyzer/protocol/ntp/ntp_pac.cc: In member function 'int binpac::NTP::Extension_Field::Parse(binpac::const_byteptr, binpac::const_byteptr, int)':
/root/project/build/src/analyzer/protocol/ntp/ntp_pac.cc:957:25: error: comparison of integer expressions of different signedness: 'binpac::uint32' {aka 'unsigned int'} and 'int' [-Werror=sign-compare]
  957 |     if (! ( value_len() <=  ( len() - 18 )  ) ) {
      |             ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
/root/project/build/src/analyzer/protocol/ntp/ntp_pac.cc:994:25: error: comparison of integer expressions of different signedness: 'binpac::uint32' {aka 'unsigned int'} and 'int' [-Werror=sign-compare]
  994 |     if (! ( value_len() <=  ( len() - 22 )  ) ) {
      |             ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

This PR adds support for the `u` suffix on integer constants in binpac code, allowing the codegen to pass along that these are unsigned values. It results in the correct output in the generated code and the warnings are resolved.

### DIFF
--- a/tools/binpac/src/pac_number.h
+++ b/tools/binpac/src/pac_number.h
@@ -4,17 +4,33 @@
 #define pac_number_h
 
 #include "pac_common.h"
+#include "pac_exception.h"
 
 class Number : public Object {
 public:
     Number(int arg_n) : s(strfmt("%d", arg_n)), n(arg_n) {}
+    Number(unsigned int arg_u) : s(strfmt("%uu", arg_u)), is_unsigned(true), u(arg_u) {}
     Number(const char* arg_s, int arg_n) : s(arg_s), n(arg_n) {}
     const char* Str() const { return s.c_str(); }
-    int Num() const { return n; }
+    int Num() const {
+        if ( is_unsigned )
+            throw Exception(this, "Num() called for unsigned number");
+        return n;
+    }
+    unsigned int Unsigned() const {
+        if ( ! is_unsigned )
+            throw Exception(this, "Unsigned() called for signed number");
+        return u;
+    }
+    bool IsUnsigned() const { return is_unsigned; }
 
 protected:
     const string s;
-    const int n;
+    bool is_unsigned = false;
+    union {
+        const int n;
+        const unsigned int u;
+    };
 };
 
 #endif // pac_number_h
--- a/tools/binpac/src/pac_scan.ll
+++ b/tools/binpac/src/pac_scan.ll
@@ -254,6 +254,13 @@ ESCSEQ	(\\([^\n]|[0-7]{3}|x[[:xdigit:]]{2}))
 				return TOK_NUMBER;
 				}
 
+<INITIAL,PP>{D}u		{
+				unsigned int n;
+				sscanf(yytext, "%uu", &n);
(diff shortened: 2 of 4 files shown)


tools/binpac/src/pac_scan.ll
```diff
@@ -254,6 +254,13 @@ ESCSEQ	(\\([^\n]|[0-7]{3}|x[[:xdigit:]]{2}))
 				return TOK_NUMBER;
 				}
 
+<INITIAL,PP>{D}u		{
+				int n;
+				sscanf(yytext, "%d", &n);
```
ckreibich: These are the same as above for `<INITIAL,PP>{D}` — shouldn't they do unsigned stuff?

timwoj: I'm pretty sure it's just passing the text along untouched because I didn't actually parse out the `u` from the string. I pushed a new commit that's a little more complete/verbose.

tools/binpac/src/pac_number.h
```diff
@@ -8,13 +8,20 @@
 class Number : public Object {
 public:
     Number(int arg_n) : s(strfmt("%d", arg_n)), n(arg_n) {}
+    Number(unsigned int arg_u) : s(strfmt("%uu", arg_u)), is_unsigned(true), u(arg_u) {}
     Number(const char* arg_s, int arg_n) : s(arg_s), n(arg_n) {}
     const char* Str() const { return s.c_str(); }
     int Num() const { return n; }
+    unsigned int Unsigned() const { return u; }
```
ckreibich: I wonder here whether these should somehow error/warn if they're accessed for the incompatible version (a `Num()` on an unsigned, an `Unsigned()` on a signed). Probably okay?

tools/binpac/src/pac_scan.ll
```diff
@@ -254,6 +254,13 @@ ESCSEQ	(\\([^\n]|[0-7]{3}|x[[:xdigit:]]{2}))
 				return TOK_NUMBER;
 				}
 
+<INITIAL,PP>{D}u		{
+				unsigned int n;
+				sscanf(yytext, "%uu", &n);
+				yylval.num = new Number(yytext, n);
```
ckreibich: This calls `Number(const char* arg_s, int arg_n)` so will initialize as a signed integer, no?

I'm a bit confused by the existing constructors because I only see calls with the string as first argument. Perhaps you can just offer an unsigned alternative for that one.

ckreibich (approved):
Thanks Tim ... I think we could iterate for days on this. :-)

PR #5590, opened 2026-06-16
Add better error handling to the release image builds

timwoj:
This includes improvements to our release image building as recommended by support at CircleCi:

> Thanks for contacting CircleCI support, sorry to hear that you're getting a failure in your build. I've had a look at your config and I think I can see why it's failing. The failure is caused by an inconsistent Docker layer cache state when using the machine executor with DLC enabled.
> These changes will do 3 things:
> 1. Stricter error handling — The step fails clearly if the build fails, instead of continuing with a broken state.
> 2. Cache cleanup when the image is missing — If zeek-builder:latest isn’t on the runner but BuildKit still has cached layers, we run docker builder prune first so we don’t hit “cached layers but missing blobs.”
> 3. One automatic retry — If the build still fails (e.g. digest not found during export), we clear the build cache and rerun once with --no-cache so the job can succeed without a manual re-run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -601,7 +601,10 @@ jobs:
       - restore_cache:
           name: Restore ccache
           keys:
-            - cache-v1-{{ .Environment.CIRCLE_JOB }}-{{ arch }}
+            - ccache-v1-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ .Branch }}-{{ .Revision }}
+            - ccache-v1-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-{{ .Branch }}-
+            - ccache-v1-{{ .Environment.CIRCLE_JOB }}-{{ arch }}-
+            - ccache-v1-{{ .Environment.CIRCLE_JOB }}-
       - run:
           name: Build zeek in new image
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -558,15 +558,30 @@ jobs:
...
-                             -t zeek-builder:latest \
-                             -f builder.Dockerfile .)
+            # DLC can restore BuildKit cache without the tagged image or blobs.
+            if [ -z "${LAST_BUILDER_IMAGE_ID}" ]; then
+              docker builder prune -af || true
+            fi
+            if ! build_builder; then
+              echo "Builder build failed; clearing build cache and retrying once with --no-cache..."
+              docker builder prune -af || true
+              (cd docker && docker build --no-cache \
```
ckreibich, line 582: Since this is the same as `build_builder` except for the added `--no-cache`, you could make that an argument to the function. Or just allow any function args to be tucked onto the `docker build` command: `build_builder --no-cache` etc.

ckreibich (approved):
LGTM. I suggest turning the PR description into the commit message, for context down the road.

timwoj:
@ckreibich I added one more commit to the branch after some futzing around one of the release branches. It fixes ccache and persisting the images to the worksapce, for later use by the container image manifest job.

ckreibich (approved):

PR #5591, opened 2026-06-17
GH CI: Update to the new environment variable for creating artifacts

timwoj:
https://github.com/zeek/zeek/pull/5576 broke installing the build artifacts on GH CI because the environment variable changed names. This PR fixes that.


bbannier (approved):
Thanks, this gets much further now. I created https://github.com/zeek/zeek/pull/5592 to resolve the test failure, but we are probably good to go here.

PR #5593, opened 2026-06-17
Fix infinite recursion in val describe

### LINKED ISSUE #5407: Zeek values can cycle & loop indefinitely

evantypanski:
This PR is a test for the AGENTS.md file. (Also it is based on that branch.) Since it's technically also a fix, I think it's reasonable to merge it considering I think it's okay. But, I did not write this: this was written by Claude code in order to test that file. I understand all code and guided the fix pretty thoroughly.

Fixes #5407

Values were able to recurse indefinitely when printed if they were self-referential. Fix it in similar ways to type recursion and to Footprint().

The LLM was extremely verbose with its PR description, here it is (behind a spoiler tag):

<details>
  <summary>LLM PR description</summary>

---
  Title: Val: detect cycles in Describe() to prevent stack overflow

  ---
  Problem

  Self-referential aggregate values — created via any-typed fields or redef record extensions —
  cause a segfault via unbounded recursion when printed. There is no guard in the Describe() path
  against a value that directly or indirectly contains itself.

  Minimal reproducers:

  # via redef
  type A: record { val: count; };
  redef record A += { a: A &optional; };
  event zeek_init() { local a: A = [$val=1]; a$a = a; print a; }

  # via any
  type A: record { val: any &optional; };
  event zeek_init() { local a: A; a$val = a; print a; }

  Both segfault with a stack overflow. The same issue affects TableVal and VectorVal when an
  any-typed value contains a reference to the enclosing container.

  Solution

  Extend ODesc with a val-level visited set (std::unordered_set<const Val*> encountered_vals) and
  three methods mirroring the already-existing type-cycle API (PushType/PopType/FindType). ODesc is
  the right carrier since it already threads through all recursive Describe() calls.

  Guard the entry point of RecordVal::Describe(), VectorVal::ValDescribe(), and
  TableVal::Describe(): if the value is already in the set, emit <cycle> and return. Pop it on the
  way out so that a value appearing in multiple sibling fields (shared but not cyclic) is still
  described fully in each position.

  Output after fix

  [val=1, a=<cycle>]
  [val=1, data=[val=2, data=<cycle>]]   # mutual cycle: only the back-edge is suppressed
  [<cycle>]                              # vector
  {
  [1] = <cycle>                          # table
  }

  Files changed

  - src/Desc.h / src/Desc.cc — add PushVal/PopVal/FindVal and encountered_vals
  - src/Val.cc — guard RecordVal::Describe, VectorVal::ValDescribe, TableVal::Describe
  - testing/btest/language/cycle-print.zeek + baseline — regression test covering all six
  reproducer variants from the issue
</details>

One part notably missing is it doesn't disclose LLM usage. I did not write *any* of this code, and it disclosed *none*. When I asked, it said:

<details>
  <summary>LLM AI disclosure</summary>

No. Let me read the policy first.

  Read 1 file (ctrl+o to expand)

⏺ The PR description needs an AI disclosure section. Here's what to append — but per the policy,
  you (the human) need to review, edit, and make it accurate before submitting:

  ---
  AI Disclosure

  This change was developed with AI assistance (Claude Code). The implementation, test cases, and
  PR description were AI-generated and reviewed by a human contributor. The author understands the
  changes: ODesc carries a visited set through all recursive Describe() calls; aggregate vals push
  themselves on entry and pop on exit, emitting <cycle> if already present; the pop-on-exit ensures
  shared-but-non-cyclic values still render fully in sibling positions.

  ---
  Note: the policy requires that you can explain the changes without AI assistance and that you
  trim/edit AI-generated content before submission. Please review the description and this
  disclosure before opening the PR.
</details>

I put some details in #5583 about what I think about this, just wanted to show what it did initially. I will fix any issues in this PR myself without (much) LLM assistance :)


src/Desc.h
```diff
@@ -209,6 +216,7 @@ class ODesc {
     File* f; // or the file we're using.
 
     std::set<const Type*> encountered_types;
+    std::unordered_set<const Val*> encountered_vals;
```
0xxon, line 219: This is a total nit - but it looks kind of weird to me that we use two different types (std::set vs std::unordered_set) for two variables that have the same usecase.

Should we just also use std::set here and not care that it might be slower, or switch the other one to underdered_set?

evantypanski: fair enough, I swapped it over to `unordered_set` in the spirit of "make code you touch better, not worse" :)

0xxon (approved):
looks good to me

awelzel:
@evantypanski - this results in the ASAN task failing now due leaked memory. We have some other tests where leak checking is disabled so could do this here, too: `testing/btest/bifs/footprint.zeek`.

```
$ btest -s build_dir=asan-build -d -j                                                                                                                                                                                                                                                                                                      
[ 33%] language.cycle-print ... failed                                                                                                                                                                                                                                                                                                                                                        
[ 33%] language.enum-nonexisting ...  % 'zeek -b /home/awelzel/corelight-oss/zeek2/testing/btest/.tmp/language.cycle-print/cycle-print.zeek >output' failed unexpectedly (exit code 1)                                                                                                                                                                                                        
  % cat .stderr                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                              
  ==1630525==ERROR: LeakSanitizer: detected memory leaks                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                              
  Indirect leak of 192 byte(s) in 3 object(s) allocated from:                                                                                                                                                                                                                                                                                                                                 
      #0 0x57fd0e94fcb1 in operator new(unsigned long) (/home/awelzel/.build/corelight-oss/zeek2/asan-build/src/zeek+0x10fccb1) (BuildId: 861817e9bd750340)                                    
      #1 0x57fd0e961ac8 in zeek::IntrusivePtr<zeek::RecordVal> zeek::make_intrusive<zeek::RecordVal, zeek::IntrusivePtr<zeek::RecordType>&>(zeek::IntrusivePtr<zeek::RecordType>&) /home/awelzel/corelight-oss/zeek2/src/include/zeek/IntrusivePtr.h:194:25
      #2 0x57fd10bf867d in zeek::detail::coerce_to_record(zeek::IntrusivePtr<zeek::RecordType>, zeek::Val*, 
      ...
```

PR #5595, opened 2026-06-18
http: Add QUERY method (RFC 10008) to known methods, extend dpd.sig

awelzel:
RFC 10008 is a proposed standard to add a new HTTP method, QUERY:

	https://www.rfc-editor.org/info/rfc10008/

Not sure we'll ever see this in plaintext HTTP/1.1 traffic anywhere, but Zeek already parses it, so this commit only really adds it to http_methods to prevent the unknown_HTTP_method weird. It also extends the DPD signature to include QUERY.

I didn't do a lot here, this was mostly for testing Claude follows the instructions in btest.rst as proposed in #5583 and the QUERY method showed up on Hacker News yesterday, so seemed a good opportunity to test drive.


timwoj:
This is still marked as a draft, but I looked through it and it looks fine. The scapy scripts seem to have been generated ok by Cluade as well. We might make the instructions in the docs a bit more explicit about the use of the Scapy API there, since I had Codex want to manually create all of the packet headers instead of using the built-in object hierarchy to do it.

awelzel:
> We might make the instructions in the docs a bit more explicit about the use of the Scapy API there

I've pushed a fixup to #5600 where you had also commented.

Marked as ready for review, thanks for the hint. Also added an entry to NEWS.

0xxon (approved):

PR #5597, opened 2026-06-19
Drop `zkg` from distribution

### LINKED ISSUE #5553: Consider not shipping `zkg` as part of Zeek
Currently `zkg` is shipped as part of Zeek, i.e., when a user installs a Zeek built with `--enable-python` they would find `zkg` in the installation's bindir. This has rough edges as users are still expected to run `zkg` in an environment which has `zkg`'s dependencies installed.

It is currently also possible to install `zkg` out of band, e.g., installed with `pipx` into the user's homedir. This takes automatically takes care of installing dependencies as well, and also enables using newer `zkg` versions before they become available with a Zeek release.

We should consider completely dropping zkg from a Zeek installation, and only supporting the out-of-band approach. We could still provide a package for zkg and depend on it for our binary packages, but could likely simplify that part. With that we should be able to more freely use external dependencies. Part of implementing this probably entails adjusting our test suite to make `zkg` available during testing (best case this just involves setting up and sourcing some virtual env).

bbannier:
Shipping `zkg` with Zeek was always slightly unusual, and e.g., required installing dependencies out of band. Including it in Zeek on Zeek's release schedule on the other hand made it much harder to e.g., make fixes available quickly.

Since `zkg` is a proper Python project which can be installed from PyPi or source just fine with e.g., `pipx`, and there is nothing in Zeek truely requiring it, drop it from the distribution.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -372,6 +372,12 @@ jobs:
                 name: Install spicy analyzers
                 command: |
                   set +e
+
+                  # `./ci/spicy-install-analyzers.sh` requires zkg.
+                  pipx ensurepath
+                  source ~/.bashrc
+                  pipx install zkg
+
                   ci/spicy-install-analyzers.sh
       - when:
           condition: << parameters.enable_benchmarks >>
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -233,6 +233,11 @@ jobs:
             git submodule update --init --recursive;
             )
 
+            # `./ci/spicy-install-analyzers.sh` requires zkg.
+            pipx ensurepath
+            source ~/.bashrc
+            pipx install zkg
+
             ./ci/pre-build.sh
             ./ci/build.sh
             ./ci/test.sh
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,9 +28,6 @@
 [submodule "auxil/highwayhash"]
 	path = auxil/highwayhash
 	url = https://github.com/google/highwayhash
-[submodule "auxil/package-manager"]
-	path = auxil/package-manager
-	url = https://github.com/zeek/package-manager
 [submodule "auxil/zeek-client"]
 	path = auxil/zeek-client
 	url = https://github.com/zeek/zeek-client
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,6 @@ option(INSTALL_BTEST "Install btest alongside Zeek." ${ZEEK_INSTALL_TOOLS_DEFAUL
 option(INSTALL_BTEST_PCAPS "Install pcap files for testing." ${ZEEK_INSTALL_TOOLS_DEFAULT})
 option(INSTALL_ZEEKCTL "Install zeekctl." ${ZEEK_INSTALL_TOOLS_DEFAULT})
 option(INSTALL_ZEEK_CLIENT "Install the zeek-client." ${ZEEK_INSTALL_TOOLS_DEFAULT})
-option(INSTALL_ZKG "Install zkg." ${ZEEK_INSTALL_TOOLS_DEFAULT})
 option(PREALLOCATE_PORT_ARRAY "Pre-allocate all ports for zeek::Val." ON)
 option(ZEEK_STANDALONE "Build Zeek as stand-alone binary." ON)
 option(ZEEK_ENABLE_FUZZERS "Build Zeek fuzzing targets." OFF)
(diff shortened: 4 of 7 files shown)


J-Gras:
How zkg is shipped with Zeek helps a lot for a common use case: Multiple Zeek installations in different prefixes. In this scenario you select your Zeek version by adding `<prefix>/bin` to your `PATH` and automatically pick up a zkg that will do the right thing. Also, as zkg is only helpful if you have Zeek, so I don't see a reason not to bundle it.

bbannier:
> How zkg is shipped with Zeek helps a lot for a common use case: Multiple Zeek installations in different prefixes. In this scenario you select your Zeek version by adding `<prefix>/bin` to your `PATH` and automatically pick up a zkg that will do the right thing. Also, as zkg is only helpful if you have Zeek, so I don't see a reason not to bundle it.

You are right that what we ship today is convenient and works, but I believe that is orthogonal to what this PR does.
zkg can already today be installed by the user, and with `zkg autoconfig` we have dedicated tool support for exactly the use you describe.

evantypanski (commented):
Shouldn't this have a number of documentation updates, too? From the top of my head, the installing, building from source, and tutorial sections all use `zkg`. If we even change how it's shipped from source the docs need updated.

J-Gras:
> You are right that what we ship today is convenient and works, but I believe that is orthogonal to what this PR does.
zkg can already today be installed by the user, and with zkg autoconfig we have dedicated tool support for exactly the use you describe.

Completely agree that this is about convenience. As I understand, the workflow you suggest could look like this:
* Run zkg autoconfig
  * Determine the current environment's Zeek prefix
  * Instruct zkg to store the generated config in that prefix
* Additionally set `ZKG_CONFIG_FILE` in the current environment

I get this works but I guess we can agree that it significantly reduces convenience and requires more background knowledge on how things work together.

From my experience: When we didn't ship zkg with Zeek I had to regularly troubleshoot broken environments. This stopped when we shipped zkg with Zeek.

bbannier:
> > You are right that what we ship today is convenient and works, but I believe that is orthogonal to what this PR does.
> > zkg can already today be installed by the user, and with zkg autoconfig we have dedicated tool support for exactly the use you describe.
> 
> Completely agree that this is about convenience. As I understand, the workflow you suggest could look like this:
> 
> [..]
>
> I get this works but I guess we can agree that it significantly reduces convenience and requires more background knowledge on how things work together.
> 
> From my experience: When we didn't ship zkg with Zeek I had to regularly troubleshoot broken environments. This stopped when we shipped zkg with Zeek.

I hear you. From a maintainer perspective the motiviation for this was twofold:

1) Currently Zeek does not ship a fully functioning `zkg`, but just installs the `zkg` script and the `zeekpkg` module it depends on. Users are still required to install dependencies by hand which is highly errorprone and IMO disturbing. That this process is so inconvienent to users forces us to strongly consider each dependency we add, what version requirements we can place on dependencies, and also leads to dedicated code in zkg to minimize fallout or at least give guidenance when users mess this up.

    That your workflow of "just use the `zkg` that is in the Zeek prefix" was not constantly broken is a result of us being unusally conservative with dependencies, otherwise you would have needed different venvs for different Zeek versions already.

2) There is a hard coupling between the zkg and Zeek release schedule since most users just use the `zkg` that comes with Zeek, even though historically zkg had a very weak dependency on the exact Zeek version (I suspect current `zkg` should work fine back to at least zeek-6.0). This limits the speed at which zkg features become available to users.

Now 1) is largely on us since we own how zkg gets installed with Zeek. I opened zeek/package-manager#228 to install a fully-functioning, self-contained `zkg` in CMake projects. With that users wouldn't have to install dependencies anymore, and we are free to have `zkg` depend on whatever makes sense.

On 2), this is already solved since `zkg` can be installed outside of Zeek already, alongside the Zeek-provided version. I have been running an external `zkg` for years and not hit any hard issues.

I think this PR could be closed if we decide to merge zeek/package-manager#227.

ckreibich:
This PR took me by surprise at this point. In short, I support this, but not at this stage in our zkg revamp.

* It is definitely true that the way we currently include zkg in the tree is flawed. There's no real dependency support — we simply try to import the small set of packages that we rely on and tell the user to install them if that fails. There's a lot of evidence that users trying to do so then trip up (over things like externally managed packages, or venvs, etc). It also limits us severely in the use of perfectly desirable packages for things like better UIs, dependency tracking, or templating. So I think this will be quite liberating.
* We know from power-user use cases that zkg often is not top of mind. Granted, this is a little chicken-and-egg because if zkg were in better shape, perhaps that wouldn't be so, but I regularly see `configure --disable-zkg`.
* It's really not that big a deal to run `pip install zkg` (or pipx or whatever) on one's local systems to install/update it.
* We need to be careful about the `zkg autoconfig` workflow — this too used to trip up users regularly. I support having it (this all has direct parallels to `pip install` vs `pip install --user` etc), but the default has been been to tie the zkg installs to the Zeek installation for quite a while now.
* This is relatively minor, but it also brings up the possibility of a nicer solution for our own Docker images: if we stick to our default images only featuring what's shipped in the tree by default (and thus without zkg going forward), then it adds incentives to us offering alternative images that also feature zkg _with a build environment_, making installation of most packages seamless, rather than only allowing ones out-of-the-box that are script-only.

So, in short: I don't think we need to do this right now, but it makes sense to me once we've meaningfully improved zkg. We have around two months still to figure out what of this should land with Zeek 9.

I think zeek/package-manager#227 looks neat — I'll tinker with it next.

PR #5600, opened 2026-06-19
doc: A few more btest.rst additions and changes

awelzel:
Yes, this is based Opus 4.7's feedback what would've been useful to have in btest.rst after adding the tests in #5599 and it's not unreasonable.

### DIFF
--- a/doc/advanced/devel/btest.rst
+++ b/doc/advanced/devel/btest.rst
@@ -20,28 +20,33 @@ All btests run in CI on various platforms with and without ZAM enabled.
 Packet Traces (PCAP Files)
 --------------------------
 
-We store packet traces in testing/btest/Traces/. The README file represents an
-index where PCAPs came from or how they were created to keep a bit of lineage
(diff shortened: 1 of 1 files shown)


doc/advanced/devel/btest.rst
```diff
@@ -41,7 +41,8 @@ The testing/btest/Traces directory is mostly structured by protocol. When
 adding a Scapy-generated trace, ``<name>.pcap``, put a ``<name>.pcap.py``
 file next to it. Running the Python script should generate the ``<name>.pcap``
 file next to ``<name>.pcap.py`` reproducibly. Running the script a year later
-should produce the exact same result.
+should produce the exact same result. We commit generated PCAP files into
+the repository.
```
timwoj: I commented on a separate PR that it would be nice to add some instructions here to drive LLMs to use the correct Scapy APIs. I had Codex originally try to manually create all of the packet headers using byte strings instead of using the objects from the API.

timwoj (changes requested):

timwoj (approved):
Sorry, didn't realize you'd made additional changes to this. LGTM.

PR #5601, opened 2026-06-20
Fix LDAP SASL NTLM/Kerberos payload extraction.

### LINKED ISSUE #4361: LDAP: Forward authentication data to NTLM, GSS-API, Kerberos
The LDAP authentication supports SPNEGO, NTLMSSP, [Sicily Authentication](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/8b9dbfb2-5b6a-497a-a533-7e709cb9a982) and possibly more.

The payload (opaque to the LDAP analyzer for the most part) contains interesting information about the user and remote system that is accessing an LDAP server.

The current LDAP analyzer implementation doesn't forward any of this data to NTLM or GSS or Kerberos. Doing so would improve visibility.

We have quite a few sample PCAPs in `Traces/ldap` by now that can be used for this.

VasuBhakt:
Fixes #4361 

## Problem
LDAP SASL auth payloads (like NTLM or Kerberos/GSS-SPNEGO) were not forwarded to child analyzers for extracting information about user and remote system that is accessing the LDAP server.

## Solution
 * `ldap.spicy`: The `SaslCredentials` parser was modified. Instead of ignoring the payload, the outer `OCTET STRING ASN.1` wrapper is gracefully stripped using a switch statement (falling back safely for un-wrapped protocols like Sicily), and the inner bytes are dumped into a payload buffer.
 * `ldap_zeek.spicy`: The `LDAP::BindRequest::%done` and `BindResponse::%done` hooks were updated to check if the payload buffer contains data. If it does, `zeek::protocol_data_in` is used to pass those raw bytes directly into Zeek's `GSSAPI` or `NTLM` child analyzers.
* Additionally, 7 baseline test files in `testing/btest/Baseline/` were updated to reflect the new analyzer registration order caused by these changes.

*(Note: The Antigravity AI coding assistant was used for understanding the codebase and code writing assistance)*


src/analyzer/protocol/ldap/ldap.spicy
```diff
@@ -408,11 +406,20 @@ type GSS_SPNEGO_Init = unit {
 type SaslCredentials = unit {
     mechanism: ASN1::ASN1Message(False) &convert=$$.body.str_value;
 
+    # Optionally strip the OCTET STRING header
```
awelzel: Mind mentioning "credentials" here in the comment (strip the credentials OCTET string ASN.1          header)

From the RFC:
```
SaslCredentials ::= SEQUENCE {
             mechanism               LDAPString,
             credentials             OCTET STRING OPTIONAL }
```

But, staring a bit longer: The `-> : void` branch is only taken if there's no content left to parse, so wonder if instead of ASN1Header, use `credentials: ASN1Message(False)` with a require on the OctetString. And then use `self$credentials$str_value` or so instead of `payload` below. What do you think? I haven't tried, but think this should work out. You may need to use (`self?$credentials)` where you previously used `|payload| > 0`, but that seems more correct in that there's a difference between no payload and empty payload (skimming the RFC).

src/analyzer/protocol/ldap/ldap_zeek.spicy
```diff
@@ -12,3 +12,30 @@ on LDAP::TlsForward::%init {
...
+        local mech = self.mechanism;
+        for ( cred in self.serverSaslCreds ) {
+            if ( mech == "GSS-SPNEGO" || mech == "GSSAPI" ) {
+                zeek::protocol_data_in(zeek::is_orig(), cred.payload, zeek::protocol_handle_get_or_create("GSSAPI"));
+            } else if ( mech == "NTLM" ) {
+                zeek::protocol_data_in(zeek::is_orig(), cred.payload, zeek::protocol_handle_get_or_create("NTLM"));
+            }
+        }
+    }
+}
```
awelzel, line 54: Mind running `spicy-format` over this file, the pre-commit hook in CI (can run locally with `pre-commit run -a` in the top-level Zeek directory) complains this being not correctly formatted, otherwise seems okay what you did here.

testing/btest/Baseline/scripts.base.protocols.ldap.ldap_max_recursion/output
```diff
@@ -1,101 +1,101 @@
...
-Ch5tsG3OlOd7l83JFc, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-C3AXaL2up5k3PR4VW4, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-C2qZRm2yQg9RoQNkVg, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-C6tA8aOk3QI9oIaHl, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-CggxJkLPcdPyKjfma, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-CI5pddk2aeicnmmoj, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-CKpb612OxJundLAeRf, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-CcwXC81REVUrKprLz, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
-C5OBjqCVHlPAjKQXi, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:943:13-943:61)
+CHhAvVGS1DHFjwGM9, LDAP exceeded max filter recursion depth (<...>/ldap.spicy:952:13-952:61)
```
awelzel: Ouch. That's a lot of noise.

Mind adding the following hunk to the `ldap_max_recursion.zeek` to replace the line numbers with XXX's. This will prevent future baseline churn whenever modifications are made to ldap.spicy. Thanks!

```
diff --git a/testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek b/testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
index fd44f76908..73cdacd282 100644
--- a/testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
+++ b/testing/btest/scripts/base/protocols/ldap/ldap_max_recursion.zeek
@@ -8,5 +8,5 @@
 event analyzer_violation_info(atype: AllAnalyzers::Tag,
     info: AnalyzerViolationInfo)
        {
-       print info$c$uid, info$reason;
+       print info$c$uid, gsub(info$reason, /:[0-9]+:[0-9]+-[0-9]+:[0-9]+/, ":XXX:XXX-XXX:XXX");
        }
```

testing/btest/Baseline/scripts.base.protocols.ldap.sasl-ntlm/conn.log
```diff
@@ -7,5 +7,5 @@
 #open XXXX-XX-XX-XX-XX-XX
 #fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents	ip_proto
 #types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	count	string	count	count	count	count	set[string]	count
-XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	127.0.0.1	60126	127.0.1.1	389	tcp	ldap_tcp	2.290081	289	1509	SF	0	ShADadFf	12	921	15	2297	-	6
+XXXXXXXXXX.XXXXXX	CHhAvVGS1DHFjwGM9	127.0.0.1	60126	127.0.1.1	389	tcp	ntlm,ldap_tcp	2.290081	289	1509	SF	0	ShADadFf	12	921	15	2297	-	6
```
awelzel, line 10: Could you extend this test to run `btest-diff-cut -m ntml.log` and do the same for `kerberos.log` for one or two tests where a `kerberos.log` file is now produced? Just for an idea and demo how the extracted data looks and that it actually properly works :-)

awelzel (changes requested):
@VasuBhakt - this looks really nice. I had a few nits and thoughts, but nothing major. This was easier than I thought 👍

src/analyzer/protocol/ldap/ldap.spicy
VasuBhakt: This one was a good catch, I also thought it was allowing garbage bytes into the payload. But upon looking into ASN1Message, I realized it attempts to decode the binary payloads as ASCII strings, which corrupts the Kerberos and NTLM data.

[Referenced line for ASCII Parsing for OctetString](https://github.com/zeek/zeek/blob/f4001f87b528d89aad1ac6503f5e1baef60093bb/src/analyzer/protocol/ldap/asn1.spicy#L210)

 Therefore, I used a hybrid approach, using ASN1Header to securely parse the tag,and then explicitly forced it to read exactly the length specified by the header.

`payload: bytes &size=self.credentialsHeader.len.len if (self?.credentialsHeader);`

This drops the garbage padding at the end while preserving the payload. Let me know what you think!

src/analyzer/protocol/ldap/ldap_zeek.spicy
VasuBhakt: Fixed it

testing/btest/Baseline/scripts.base.protocols.ldap.ldap_max_recursion/output
VasuBhakt: Masked the ldap_max_recursion.zeek lines by implementing your solution involving regex matching

testing/btest/Baseline/scripts.base.protocols.ldap.sasl-ntlm/conn.log
VasuBhakt: Added the btest-diff-cut demos for the new ntlm.log and kerberos.log baselines

VasuBhakt:
Hi @awelzel thanks for the review! I have made the changes as you asked. Regarding the ASN1Message review, as it decodes the payload into ASCII, it was corrupting the payload, hence I used a hybrid method. Let me know what you think!

src/analyzer/protocol/ldap/ldap.spicy
awelzel: > I realized it attempts to decode the binary payloads as ASCII strings, which corrupts the Kerberos and NTLM data.

Ah, good catch. That `decode()` is unfortunate, but what you have is good :+1: Resolving.

testing/btest/Baseline/scripts.base.protocols.ldap.aduser1/krb.kerberos.log
```diff
@@ -0,0 +1,5 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	request_type	client	service	success	error_msg	from	till	cipher	forwardable	renewable	client_cert_subject	client_cert_fuid	server_cert_subject	server_cert_fuid
+XXXXXXXXXX.XXXXXX	C4J4Th3PJpwUYZZ6gc	192.168.226.131	60772	192.168.226.136	88	AS	User1/ADHACKING.LOCAL	krbtgt/ADHACKING.LOCAL	T	-	-	1741683288.000000	aes256-cts-hmac-sha1-96	T	T	-	-	-	-
+XXXXXXXXXX.XXXXXX	CtPZjS20MLrsMUOJi2	192.168.226.131	60784	192.168.226.136	88	TGS	User1/ADHACKING.LOCAL	ldap/WIN-VGO56I95EAK	T	-	-	1741683288.000000	aes256-cts-hmac-sha1-96	T	T	-	-	-	-
+XXXXXXXXXX.XXXXXX	ClEkJM2Vm5giqnMf4h	192.168.226.131	54544	192.168.226.136	389	-	-	-	-	-	-	-	-	-	-	-	-	-	-
```
awelzel: This is the log entry for the LDAP connection's inner payload - it doesn't have any fields set but the connection information. I'd have expected the GSS-API/Kerberos analyzer to extract some information (similarly to ntlm.log), but that might be naive.  The two above are dedicated Kerberos connections over port 88 that just happen to be in the testing PCAP, but aren't relevant to your changes.

I have modified the aduser1.zeek test locally to filter on `port 389` (hunk below) which I think is reasonable to focus on just the LDAP analyzer. With that only the single "empty" entry is left in kerberos.log.

Would you have cycles to look a bit deeper if we could get some info populated? A debug build and running with `-B dpd`, then looking at `debug.log` should show which bytes are delivered to which analyzer and understanding the analyzer's behavior a bit.

This stuff can be tricky, so don't be discouraged :-) If any changes are needed to GSS-API or Kerberos, adding separate commits for these would be preferred.

---

```
diff --git a/testing/btest/scripts/base/protocols/ldap/aduser1.zeek b/testing/btest/scripts/base/protocols/ldap/aduser1.zeek
index ce56e83f4a..fe92f80f20 100644
--- a/testing/btest/scripts/base/protocols/ldap/aduser1.zeek
+++ b/testing/btest/scripts/base/protocols/ldap/aduser1.zeek
@@ -1,7 +1,7 @@
 # @TEST-REQUIRES: have-spicy
-# @TEST-EXEC: zeek -r ${TRACES}/ldap/aduser1.pcap %INPUT
+# @TEST-EXEC: zeek -f 'port 389' -r ${TRACES}/ldap/aduser1.pcap %INPUT
 # @TEST-EXEC: mkdir krb && mv *.log krb
-# @TEST-EXEC: zeek -r ${TRACES}/ldap/aduser1-ntlm.pcap %INPUT
+# @TEST-EXEC: zeek -f 'port 389' -r ${TRACES}/ldap/aduser1-ntlm.pcap %INPUT
 # @TEST-EXEC: mkdir ntlm && mv *.log ntlm
 # @TEST-EXEC: btest-diff krb/ldap.log
 # @TEST-EXEC: btest-diff krb/ldap_search.log
```

awelzel: If you get stuck or don't have the time for this, let us know, we can also look, but there's no rush!

VasuBhakt: Alright I will look into this and get back to you!

VasuBhakt: I have modified the krb_ap_request
[krb_ap_request snippet](https://github.com/zeek/zeek/commit/40df245d5e61b2d0d39a1c371f7199aea02bb7f9#diff-64bf9326cf7f917b66e7ee5ac7dd1985c381528a99aa25e7470662bc1cf81d94L193)

This now extracts the service name and cipher associated with the kerberos ticket. Also, I updated the `aduser1.zeek` test as per your recommendation to focus on the LDAP analyzer. I have ran the test suite and updated the necessary baselines. Let me know what you think!

scripts/base/protocols/krb/main.zeek
```diff
@@ -193,6 +193,11 @@ event krb_as_response(c: connection, msg: KDC_Response) &priority=-5
 event krb_ap_request(c: connection, ticket: KRB::Ticket, opts: KRB::AP_Options) &priority=5
 	{
 	set_session(c);
+	c$krb$request_type = "AP";
+	if ( ticket?$service_name )
+		c$krb$service = ticket$service_name;
+	if ( ticket?$cipher )
+		c$krb$cipher = cipher_name[ticket$cipher];
```
awelzel, line 201: Great find! I think this is fair to do and a good improvement over logging entries with only  unset fields as seen in the baseline changes. Thanks!

Two follow-up requests here:

* For consistency with the other handlers, mind adding the following hunk? In future, maybe more info could be extracted from the `krb_ap_response()`, but for now leaving the skeleton seems okay.

```

# event krb_ap_response(c: connection) &priority=5
#       {
#       if ( set_session(c) )
#               return;
#       ...
#       }

event krb_ap_response(c: connection) &priority=-5
        {
        do_log(c);
        }
```

* Please extend the `request_type` docs for the `Info` record in krb/main.zeek to add the "AP" (Application Request)  string as possible value.

---


Note for who merges the PR: Will need an update to the zeek-testing-private baseline.

awelzel (changes requested):
Thanks a lot @VasuBhakt ! Two follow-up requests, otherwise seems complete then and ready to merge.

scripts/base/protocols/krb/main.zeek
VasuBhakt: Added the hunk you provided to `main.zeek` for krb and added the "AP" request type to the docs. Changes squashed into previous kerberos-related commit. Let me know what you think!

awelzel:
Hey @VasuBhakt - I had wanted to merge this, but ran into a snag when running the `external/zeek-testing/tests/ctu-sme-11-vm-win7ad.zeek` test case (currently skipped during CI on GitHub):

With your latest revision, there's some violations from the GSSAPI analyzer reported that look as follows:
```
--- a/Baseline/tests.ctu-sme-11-vm-win7ad/analyzer_debug.log                                                                                                                                                                                                                                                                                                                                  
+++ b/Baseline/tests.ctu-sme-11-vm-win7ad/analyzer_debug.log                                                                                                                                                                                                                                                                                                                                  
@@ -16,6 +16,134 @@ XXXXXXXXXX.XXXXXX   violation       protocol        DNP3_UDP        XXXXXXXXXXX     -       192.168.1.108   52443                                                                                                                                                                                                                                                         
 XXXXXXXXXX.XXXXXX      violation       protocol        DNP3_UDP        XXXXXXXXXXX     -       192.168.1.108   54182   192.55.83.30    53      broken_checksum -                                                                                                                                                                                                                             
 XXXXXXXXXX.XXXXXX      violation       protocol        DNP3_UDP        XXXXXXXXXXX     -       192.168.1.108   55309   1.1.1.1 53      broken_checksum -                                                                                                                                                                                                                                     
 XXXXXXXXXX.XXXXXX      violation       protocol        DNP3_UDP        XXXXXXXXXXX     -       192.168.1.108   55384   205.251.193.220 53      broken_checksum -                                                                                                                                                                                                                             
+XXXXXXXXXX.XXXXXX      violation       protocol        GSSAPI  XXXXXXXXXXX     -       192.168.1.105   50041   192.168.1.108   389     Binpac exception: binpac exception: invalid index for case: GSSAPI_NEG_TOKEN_RESP_Arg: 179    -                                                                                                                                                       
+XXXXXXXXXX.XXXXXX      violation       protocol        GSSAPI  XXXXXXXXXXX     -       192.168.1.105   50041   192.168.1.108   389     Binpac exception: binpac exception: out_of_bound: GSSAPI_NEG_TOKEN_RESP_Arg:args: 85 > 36     -   
```

Here's a small PCAP extracted from the big CTU that reproduces this:
[ctu-60387-63723-ldap.pcap.zip](https://github.com/user-attachments/files/29388086/ctu-60387-63723-ldap.pcap.zip)

```
$ zeek -C -r ctu-60387-63723-ldap.pcap frameworks/analyzer/debug-logging Spicy::enable_print=T
$ zeek-cut -m < analyzer_debug.log 
ts      cause   analyzer_kind   analyzer_name   uid     fuid    id.orig_h       id.orig_p       id.resp_h       id.resp_p       failure_reason  failure_data
1677013602.883194       violation       protocol        GSSAPI  CfKzNA1I6Zn7eDnjse      -       192.168.1.107   60387   192.168.1.108   389     Binpac exception: binpac exception: out_of_bound: GSSAPI_NEG_TOKEN_RESP_Arg:args: 85 > 36       -
1677013602.883194       disabled        protocol        GSSAPI  CfKzNA1I6Zn7eDnjse      -       192.168.1.107   60387   192.168.1.108   389     -       -
1677013602.883194       confirmation    protocol        LDAP_TCP        CfKzNA1I6Zn7eDnjse      -       192.168.1.107   60387   192.168.1.108   389     -       -
1677013602.883464       violation       protocol        GSSAPI  CfKzNA1I6Zn7eDnjse      -       192.168.1.107   60387   192.168.1.108   389     Binpac exception: binpac exception: invalid index for case: GSSAPI_NEG_TOKEN_RESP_Arg: 179      -
...
```
From what I can tell, the BinPAC GSSAPI analyzer does not understand the plain NTLMSSP format. I dabbled a bit and came up with the following hunk [ntlm.diff.txt](https://github.com/user-attachments/files/29388138/ntlm.diff.txt) to forward directly to the NTLM analyzer, but maybe you see something better? If you're still up for iterating, this would be worth a new regression test and maybe you have other ideas about a fix?

VasuBhakt:
Hey @awelzel, I fixed the issue using the ntlm diff you provided, works alright, raw ntlmssp payloads are directly sent to ntlm analyzer. I also added a regression test for this issue. Let me know what you think!

awelzel (approved):
LGTM. I'll update the external baselines and then merge 👍

VasuBhakt:
Thank you so much for the guidance throughout this, @awelzel! I learned a lot about Zeek along the way. Looking forward to contributing more!

PR #5608, opened 2026-06-23
DNS: Skip query/response association for multicast destinations

### LINKED ISSUE #5389: DNS analyzer improperly associates mDNS requests using transaction ID
The DNS analyzer improperly associates unrelated mDNS requests and replies.

I've attached [gen_mdns.py](https://github.com/user-attachments/files/26950393/gen_mdns.py) (which claude was gracious enough to contribute), which will generate a minimal PCAP that triggers the behavior (github didn't let me upload the pcap directly). 

I think the root cause is attempting to associate mDNS queries by query id.

From RFC 6762 Section 18.1:

> In multicast query messages, the Query Identifier SHOULD be set to zero on transmission.
> 
> In multicast responses, including unsolicited multicast responses, the Query Identifier MUST be set to zero on transmission, and MUST be ignored on reception.

We probably need to update the DNS analyzer to track whether a query is mDNS or regular DNS, and on mDNS ignore the transaction ID, using a different methodology to correlate responses.

This brings up a bigger issue with mDNS in general: we typically don't expect to see the response for a given mDNS query in the same "connection" (5-tuple) at all (it could happen, however it would be unusual). The typical flow is:

* requestor:5353 -> mdns_multicast:5353 query
* responder:5353 -> mdns_multicast:5353 response

Currently, the DNS analyzer is built around traffic being connection oriented, and this breaks down for mDNS. We could maybe still correlate unreplied requests with unsolicited replies after-the-fact, but although the protocol format is compatible between the two the semantics are not.

Edit: adding tcpdump output and dns log to try to show the issue more clearly

```
$ tcpdump -nv -r pcap/mdns_test.pcap 
reading from file pcap/mdns_test.pcap, link-type EN10MB (Ethernet), snapshot length 65535
18:04:20.616834 IP (tos 0x0, ttl 255, id 1, offset 0, flags [none], proto UDP (17), length 56)
    192.168.0.2.5353 > 224.0.0.251.5353: 0 A (QM)? wpad.local. (28)
18:04:20.617167 IP (tos 0x0, ttl 255, id 1, offset 0, flags [none], proto UDP (17), length 56)
    192.168.0.2.5353 > 224.0.0.251.5353: 0 AAAA (QM)? wpad.local. (28)
18:04:20.617368 IP (tos 0x0, ttl 255, id 1, offset 0, flags [none], proto UDP (17), length 76)
    192.168.0.2.5353 > 224.0.0.251.5353: 0*- [0q] 1/0/0 s0m3th1ng-3l53.local. A 192.168.0.2 (48)
18:04:20.617594 IP (tos 0x0, ttl 255, id 1, offset 0, flags [none], proto UDP (17), length 88)
    192.168.0.2.5353 > 224.0.0.251.5353: 0*- [0q] 1/0/0 s0m3th1ng-3l53.local. AAAA fe80::200:deff:fead:beef (60)
```

```
$ cat dns.log 
#separator \x09
#set_separator  ,
#empty_field    (empty)
#unset_field    -
#path   dns
#open   2026-04-22-00-08-13
#fields ts      uid     id.orig_h       id.orig_p       id.resp_h       id.resp_p       proto   trans_id        rtt     query   qclass  qclass_name     qtype   qtype_name      rcode   rcode_name      AA      TC      RD      RA      Z       answers TTLs    rejected
#types  time    string  addr    port    addr    port    enum    count   interval        string  count   string  count   string  count   string  bool    bool    bool    bool    count   vector[string]  vector[interval]        bool
1776816260.616834       C9mDh128WB7ErfscQe      192.168.0.2     5353    224.0.0.251     5353    udp     0       0.000534        wpad.local      1       C_INTERNET      1       A       0       NOERROR T       F       F       F       0       192.168.0.2     120.000000      F
1776816260.617167       C9mDh128WB7ErfscQe      192.168.0.2     5353    224.0.0.251     5353    udp     0       0.000427        wpad.local      1       C_INTERNET      28      AAAA    0       NOERROR T       F       F       F       0       fe80::200:deff:fead:beef        120.000000      F
```

ekoyle:
mDNS and LLMNR messages sent to multicast addresses cannot be correlated by transaction ID  because replies arrive on different connections. Previously, unrelated queries and replies sharing the same transaction ID (typically 0 per RFC 6762 section 18.1) were incorrectly paired, contaminating log entries.

Skip the pending queue association logic when the responder address is multicast (224.0.0.0/4 or ff00::/8) and log each message as a standalone entry instead. Because multicast messages are now logged immediately rather than waiting in the pending queue, they appear earlier in dns.log relative to unicast entries from the same pcap.

Also fix dns_notify2.pcap which used anonymized IPs in multicast space (234.x.x.x) — remapped to RFC 5737 documentation range (203.0.113.0/24).

Fixes #5389

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


testing/btest/Baseline/opt.basic/dns.log
ekoyle, line 1: Ordering changed due to short-circuiting the correlation logic.

testing/btest/Traces/dns/dns_notify2.pcap
ekoyle, line 1: Rewriting IP addresses to a non-multicast range seemed like the best approach here. Let me know if you want the file renamed as well.

testing/btest/Baseline/scripts.base.protocols.dns.notify/dns.log
ekoyle, line 1: Updated IP addresses in the pcap.

0xxon:
When I saw this and looked at the change, I immediately wondered if we should just split mdns messages out to a separate log file. Looking at #5389, I was not the first person with this thought.

This change obviously is a direct improvement to what we are doing at the moment - but I still wonder if we should just also put mdns into its own log. That doesn't have to be part of this PR though, I am mostly curious about thoughts.

ekoyle:
> When I saw this and looked at the change, I immediately wondered if we should just split mdns messages out to a separate log file. 

This was my initial thought as well, but I'm currently leaning towards not separating it.

The semantics are a little different for mDNS, but the wire format is identical. A separate log for mDNS/LLMNR only makes sense to me if we are changing the log structure. If the log format is the same, that adds downstream integration work to users for little to no benefit. 

It would be nice to correlate requests/responses, but that would require us to come up with a way to keep track of: 

* IP of requestor, query details
* Zero or more instances of the following 
    * IP of responder
    * Response details

Thinking about trying to interact with that kind of a structure in a SIEM is giving me anxiety, so my instinct is not to go that route. We could make a requests log and a replies log, but at that point we are logging every packet again, so does it provide a benefit over the existing DNS log?

Lastly, it seems like the only way to correlate these is by qname and time, which could be problematic for zeek (requests and replies are in different connections, probably on different workers -- leads to timing problems, breaks down with multiple sensors) but this operation would be relatively easy for a SIEM.

vpax:
@ekoyle if it's not a separate log, will it be a DNS log schema change? (I thought it was, but haven't been following this for a while.) That would be great to avoid.

awelzel:
> if it's not a separate log, will it be a DNS log schema change? (I thought it was, but haven't been following this for a while.) That would be great to avoid.

It's not a log schema change, so should be okay. It might result in more log entries total.

I was thinking of an option to revert to the old behavior, but I think you have this with:
```
redef DNS::multicast_subnets = {};
```

@ekoyle - mind sketching an entry in NEWS, including how to revert if someone really wants to revert? I think I'm okay approving unless someone thinks that's not a good idea? It'd be good to continue tracking a `mdns.log` addition which indeed would aggregate on one of the cluster nodes (or sharded by query if possible).

ekoyle:
> mind sketching an entry in NEWS, including how to revert if someone really wants to revert?

done

> It'd be good to continue tracking a mdns.log addition which indeed would aggregate on one of the cluster nodes (or sharded by query if possible).

After thinking about this to try to fix the initial bug, I wonder if the functionality here merits the complexity. There are some awkward cases that I think would be common:

* multiple hosts send the same query at around the same time
    * close the old and start a new?
    * keep both?
    * instead track a list of clients who made this query?
* multiple responses to a given query, from different clients
    * track a list of clients who responded?
    * what if we are interested in a specific reply (especially for security-related questions)?

We would almost need a separate queries and replies log, which the existing DNS log can already handle perfectly fine, although it may waste some space due to unused fields.

awelzel (approved):
Thanks.

PR #5614, opened 2026-06-24
Trigger network_time_init for wallclock fallback

J-Gras:
When Zeek did not process packets (e.g., on non-worker nodes In a live cluster) `network_time_init` was never raised. Furthermore, Claude pointed out that due table entry expiration being offset based, not initializing `zeek_start_network_time` caused a year 2038 bug:

> On non-worker nodes in live mode, zeek_start_network_time previously remained 0. Table entry expiration times are stored as int (seconds) offsets from this value. With a zero base, the offset equals the full Unix epoch (~1.75B seconds today), which still fits in a signed 32-bit int but wastes nearly all of the available range on a constant. This accidentally works until 2038 when the epoch exceeds INT_MAX, at which point expiration tracking silently breaks. Setting zeek_start_network_time at initialization avoids this by keeping offsets small as intended.

AI usage: Assistance by Claude Opus 4.6


awelzel:
LGTM, thanks @J-Gras. Mind adding a test to `core/` or so with `exit_only_after_terminate=T` and a `terminate()` in `network_time_init()` - that should hang indefinitely today and be fixed with this, correct?

src/zeek-setup.cc
```diff
@@ -1057,11 +1057,16 @@ SetupResult setup(int argc, char** argv, Options* zopts) {
...
     if ( ! run_state::reading_live && ! run_state::reading_traces &&
-         id::find_const("allow_network_time_forward")->AsBool() )
+         id::find_const("allow_network_time_forward")->AsBool() ) {
         // Set up network_time to track real-time, since
         // we don't have any other source for it.
         run_state::detail::update_network_time(util::current_time());
 
+        run_state::zeek_start_network_time = run_state::network_time;
+        if ( network_time_init )
+            event_mgr.Enqueue(network_time_init, Args{});
```
awelzel: Ah, and one more thought: Could you move enqueue of network_time_init() to  after the enqueue of `zeek_init()` such that `network_time_init()` is guaranteed to run after `zeek_init()`. Otherwise the order differs on workers and non-workers.

```
    if ( zeek_init )
        event_mgr.Enqueue(zeek_init, Args{});

    if ( run_state::network_time != 0.0 && network_time_init )
        event_mgr.Enqueue(network_time_init, Args{});

```

J-Gras: Good catch! Moved it...

J-Gras:
Thanks for the suggestions! I've added the test case and moved the event generation.

awelzel (approved):

PR #5618, opened 2026-06-25
Disable cluster.zeromq.overload-drop btest on CI under asan and ubsan

timwoj:
This test fails fairly consistently on CI when running under asan and ubsan. It passes for all of the other builds and consistently works running locally under those sanitizers. Extending the timeout wouldn't do anything like for other cluster tests, since it's already dropping messages (recorded via telemetry). Waiting longer for the processes to exit won't change the drops having already happened.

### DIFF
--- a/testing/btest/cluster/zeromq/overload-drop.zeek
+++ b/testing/btest/cluster/zeromq/overload-drop.zeek
@@ -2,6 +2,8 @@
 #
 # @TEST-REQUIRES: have-zeromq
 # @TEST-REQUIRES: ! is-windows-ci
+# @TEST-REQUIRES: ! ( have-asan && test -n "${CI}" )
+# @TEST-REQUIRES: ! ( have-ubsan && test -n "${CI}" )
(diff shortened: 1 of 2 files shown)


awelzel:
> This test fails fairly consistently on CI when running under asan and ubsan.

Did you observe this primarily on the `release/` branches? I'd be okay doing this, but could also make sense to backport the following to PRs to 8.0/8.2 that attempted to tame some flakiness:

https://github.com/zeek/zeek/pull/5449
https://github.com/zeek/zeek/pull/5487


If disabling under asan/ubsan is easier, then think the `TEST-REQUIRES` should go into the following four:

* overload-drop.zeek
* overload-no-drop.zeek
* overload-worker-proxy-topic-drop.zeek 
* overload-worker-proxy-topic-no-drop.zeek

timwoj:
Yep, bringing those commits over to the release branches seems to have cleaned things up. Closing this.

PR #5621, opened 2026-06-26

bbannier:
(cherry picked from commit 4fdd83f)


bbannier:
@timwoj This just pushes a branch, not a tag, so we need to either follow this up with another bump to a tag, or we let this stay open until we can change this to directly add a tag.

timwoj (approved):
I'm merging this so I have a base for the security-patch PRs that I can deploy elsewhere. We can do a second one with a tagged release next week.

PR #5622, opened 2026-06-27
Add Broker backpressure tests

ckreibich:
These are ports of the corresponding tests in the zeek-testing-cluster testsuite, using our usual btest-bg-run clusterization instead. They operate nearly identically. The main difference is that instead of a standalone shell-script driver that tracks the state machine and triggers wedging of a node, the Zeek scripts do it all here.

Works locally for me, but I'll keep this in draft state until we have some results from CI.

This is one of the remaining items of #5512 .


ckreibich:
Seems to be working, so I'm marking this ready for review.

timwoj (commented):
I know you said that you're still working on another test for this, but I made a pass over what's here already. I compared it to the existing zeek-testing-cluster scripts, and it's consistent with what we were doing over there. 👍🏼 

I'll give a final review/approval after the extra test is in.

ckreibich:
Thanks Tim. Yeah, I wanted to expand these to also cover the message-drop policies (including the default "drop_oldest" one), since we've so far not had tests for these. I just pushed the changes, which are pretty big. The tests haven't changed much, it's just massaging to make the whole thing run as a set of tests via `@TEST-START-NEXT`.

Fingers crossed ...

timwoj (approved):

PR #5630, opened 2026-07-02
Fix reporter logging without wallclock fallback

J-Gras:
Fix a tiny oversight from #5614.

### DIFF
--- a/src/Reporter.cc
+++ b/src/Reporter.cc
@@ -615,7 +615,15 @@ void Reporter::DoLog(const char* prefix, EventHandlerPtr event, FILE* out, Conne
     if ( out ) {
         std::string s = "";
 
-        if ( (run_state::reading_live || run_state::reading_traces) && run_state::zeek_start_network_time != 0.0 ) {
+        // Prefix reporter messages that go to stdout/stderr with network time if
+        // network time has been initialized and this process's network time is either
+        // explicitly controlled (allow_network_time_forward=F), or the process is a
+        // worker that updates its network time from a packet source (interface or pcap).
+        //
+        // In other words, reporter messages to stdout/stderr from non-worker processes
+        // are not prefixed with the timestamp unless allow_network_time_forward=F.
+        if ( (run_state::reading_live || run_state::reading_traces || ! BifConst::allow_network_time_forward) &&
+             run_state::zeek_start_network_time != 0.0 ) {
             char tmp[32];
             snprintf(tmp, 32, "%.6f", run_state::network_time);
             s += std::string(tmp) + " ";


src/Reporter.cc
```diff
@@ -615,7 +615,9 @@ void Reporter::DoLog(const char* prefix, EventHandlerPtr event, FILE* out, Conne
     if ( out ) {
         std::string s = "";
 
-        if ( (run_state::reading_live || run_state::reading_traces) && run_state::zeek_start_network_time != 0.0 ) {
+        if ( (run_state::reading_live || run_state::reading_traces ||
+              ! id::find_const("allow_network_time_forward")->AsBool()) &&
```
awelzel: Instead of `id::find_const()`, use `BifConst::allow_network_time_forward` as that doesn't go through a scope lookup. Probably not actually performance relevant here, but still nicer.

```suggestion
              ! BifConst::allow_network_time_forward) &&
```

src/Reporter.cc
```diff
@@ -615,7 +615,9 @@ void Reporter::DoLog(const char* prefix, EventHandlerPtr event, FILE* out, Conne
...
+             run_state::zeek_start_network_time != 0.0 ) {
```
awelzel, line 626: I was wondering in the last PR a bit about the intuition of this, but it's probably mostly historic and to avoid spurious timestamps in the messages for scripting only things.

Could you prefix this all with a comment. Something like below, but feel free to tweak, I might have not fully gotten it:

```
// Prefix reporter messages that go to stdout/stderr with network time if
// network time has been initialized and this process's network time is either
// explicitly controlled (allow_network_time_forward=F), or the process is a
// worker that updates its network time from a packet source (interface or pcap).
//
// In other words, reporter messages to stdout/stderr from non-worker processes
// are not prefixed with the timestamp unless allow_network_time_forward=F.
```

Would you prefer to just always include the network timestamp if `allow_network_time_forward=F`? Meaning also the `0.0` timestamp? Makes some sense to me to be more verbose when running in that mode.

```
if ( ! BifConst::allow_network_time_forward || ((run_state::reading_live || run_state::reading_traces) && run_state::zeek_start_network_time != 0.0) ) 
```

awelzel:
Oh, this tickles a few crashes in CI, too - may want to look at those - the `find_const()` may return a nullptr, too :-/

src/Reporter.cc
J-Gras: > Would you prefer to just always include the network timestamp if allow_network_time_forward=F? Meaning also the 0.0 timestamp? Makes some sense to me to be more verbose when running in that mode.

I thought about this but figured it would be nice if test baselines wouldn't differ between both modes if they don't need to. In case we change the logging behavior, I would want to include the timestamp always and everywhere. That just seemed quite noisy.

awelzel (approved):

PR #5633, opened 2026-07-02
Expose new functions to retrieve original X.509 distinguished names

### LINKED ISSUE #5632: X509 DN representation of international names is imprecise
When Zeek is parsing X509 certificates which contain international DN which are UTF-8 encoded, openssl produces binary escapes of UTF-8 sequences.
These are difficult to read especially for non-latin characters.
Furthermore, the RFC 2253 normalization of DN components may hide interesting details about the certificate by sorting the name components.
Depending of the use, sorting can be useful, but it would be nice to allow user to change the x.509 logging representation by policy.

For example in the [X509::Certificate subject field](https://docs.zeek.org/en/master/scripts/base/init-bare.zeek.html#field-X509::Certificate$subject), a Chinese certificate would look like this:
`CN=www.trustasia.com,O=\E4\BA\9A\E6\95\B0\E4\BF\A1\E6\81\AF\E7\A7\91\E6\8A\80\EF\BC\88\E4\B8\8A\E6\B5\B7\EF\BC\89\E6\9C\89\E9\99\90\E5\85\AC\E5\8F\B8,L=\E4\B8\8A\E6\B5\B7\E5\B8\82,ST=\E4\B8\8A\E6\B5\B7\E5\B8\82,C=CN,serialNumber=913101040660431939,jurisdictionL=\E5\BE\90\E6\B1\87\E5\8C\BA,jurisdictionST=\E4\B8\8A\E6\B5\B7\E5\B8\82,jurisdictionC=CN,businessCategory=Private Organization`

But it could be nice to have something like this:
`businessCategory=Private Organization, jurisdictionC=CN, jurisdictionST=上海市, jurisdictionL=徐汇区, serialNumber=913101040660431939, C=CN, ST=上海市, L=上海市, O=亚数信息科技（上海）有限公司, CN=www.trustasia.com`

This is more consistent with what's really inside the certificate.

see: https://github.com/zeek/zeek/pull/5633

eb-anssi:
Expose the subject and issuer distinguished names as encoded in the certificate, preserving the original DN components order and UTF-8 encoding.

Add tests and improve error handling.


src/file_analysis/analyzer/x509/functions.bif
```diff
@@ -867,6 +867,117 @@ zeek::StringValPtr x509_entity_hash(zeek::file_analysis::detail::X509Val *cert_h
 	}
 %%}
 
+## Get an accurate string representation of the subject’s distinguished name from
+## the given certificate. Unlike the value stored in the subject field of the
+## X509::Certificate record, this representation preserves the order of the
+## distinguished name components as they appear in the certificate's ASN.1 structure,
+## rather than sorting them according to rfc 2253. It also correctly extracts UTF-8 name
```
timwoj, line 874: This makes the documentation link to the RFC, same later. That said, what about RFC 2253 specifies that they have to be ordered in a specific way? Could we order `issuer` and `subject` like this by default? @0xxon might have opinions here.

```suggestion
## rather than sorting them according to :rfc:`2253`. It also correctly extracts UTF-8 name
```

eb-anssi: Regarding the order of issuer and subject in the X.509 certificate, these names are already sorted according to the RFC by default in the [issuer](https://docs.zeek.org/en/master/scripts/base/protocols/ssl/main.zeek.html#field-SSL::Info$issuer) and [subject](https://docs.zeek.org/en/master/scripts/base/protocols/ssl/main.zeek.html#field-SSL::Info$subject) fields by the x509 certificate parser  `X509::ParseCertificate()` in `X509.cc` using the following openssl function:
```
X509_NAME_print_ex(bio, X509_get_subject_name(ssl_cert), 0, XN_FLAG_RFC2253);
```
The purpose of this PR is to allow extraction of the exact order seen in the certificate, which reflects the way the certificate was requested and issued.

0xxon:
Just for visibility - I added some questions to the related #5632

0xxon:
Hi,
I just wanted to apologise a bit for the delay on this - we are currently all a bit busy. This has not been forgotten and I will try to circle back on it soon :)

PR #5639, opened 2026-07-03
GHCI: Add support for conditional builds based on PR labels

timwoj:
This adds conditional builds similar to how we did it on Cirrus and how we do it on Circle. It also makes a bunch of whitespace changes to make `yamllint` happier, so you can basically skip reviewing the second commit here.

Opening as draft to begin with because I imagine this will break something.


.github/scripts/generate-gh-check-matrix.py
```diff
@@ -0,0 +1,448 @@
...
+#   FILTER_IF_PR_NOT_FULL_OR_ZEEKCTL: Skip All OR (is PR and no Full and no Zeekctl)
+#   Weekly/nightly builds: skip_if: true (no schedule trigger in this workflow)
+
+
+def skip_if_skip_all(labels):
+    """FILTER_IF_SKIP_ALL: Skip if skip_all is set."""
+    return "CI: Skip All" in labels
+
+
+def skip_if_pr_full(labels):
```
ckreibich, line 38: This is actually `skip_if_not_pr_full`, right? Or perhaps `skip_unless_pr_full`. Similar for the next few.

.github/scripts/generate-gh-check-matrix.py
```diff
@@ -0,0 +1,448 @@
...
+       ./Scripts/build-zeek
+       ../../btest/btest -A -d -j ${ZEEK_CI_BTEST_JOBS}""",
+)
+
+add_job(
+    includes,
+    "include-plugins-debian13",
+    "ci/debian-13/Dockerfile",
+    skip_if_pr_full(pr_labels),
+    "--include-plugins=${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/Files/protocol-plugin;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/Files/py-lib-plugin;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/Files/zeek-version-plugin;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/external/zeek-perf-support;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/external/zeek-more-hashes;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/external/zeek-cluster-backend-nats;${ZEEK_CI_WORKING_DIR}/testing/builtin-plugins/external/zeek-kafka",
```
ckreibich, line 430: I know you're just moving things around here, but it might be worth adding `ZEEK_CI_BUILTIN_PLUGINS` or some such to shorten these paths.

ckreibich (commented):
Yeah, I'm fine with this if we need it. The Python actually looks a bit more readable to me than the corresponding YAML, and it certainly adds flexibility.  The only bummer is that this splits jobs across the matrix and the remaining ones in check.yml.

So I wonder: is there no way to do this in YAML, or do we find the options to do it in YAML too clunky? I looked around a bit and it seems possible to check the labels via things like
```
if: ! contains(github.event.pull_request.labels.*.name, "CI: Skip All")
```
and to weave expression results into the matrix as conditions to include/exclude on, like [here](https://github.com/orgs/community/discussions/26253#discussioncomment-3250989). (Fully admit I haven't tried it!)

PR #5640, opened 2026-07-03
util/fmt: Make the vfmt() format buffer thread-local

stefangotz:
`util::vfmt()` (behind `util::fmt()`) formats into a single process-global
`static` buffer and returns a pointer into it, valid only until the next
`fmt()` call on any thread. This is a long-known thread-safety hazard: when the
main thread holds a `fmt()` result across a scheduling point and another thread
formats concurrently, the shared buffer can be overwritten or reallocated out
from under the first caller.

A concrete path: `WriterFrontend` passes `util::fmt("%s/%s", path, name)`
straight into `util::copy_string()`, which does `strlen()` -> `new[]` ->
`strcpy()`. If a log-writer or input-reader backend thread calls `util::fmt()`
in between, a longer result makes `strcpy()` run past the measured length and
`__strcpy_chk` aborts the process; a `safe_realloc()` growth turns the held
pointer into a dangling read. This shows up in the field as a worker crash
during log-writer initialization.

## Approach

Over time this has been worked around at individual call sites, notably via the
per-thread `BasicThread::Fmt()`, but `util::fmt()` itself stayed shared, so new
or overlooked call sites keep reintroducing the race. Rather than patch another
call site, this fixes it at the root: the buffer becomes `thread_local`, so each
thread formats into its own storage. `fmt()` stays lock-free and writer/reader
concurrency is preserved. Using `std::vector<char>` (rather than a raw
`thread_local` pointer) frees the per-thread buffer at thread exit, so nothing
leaks.

## Testing

Adds a deterministic doctest (`vfmt_thread_safety`): the main thread captures a
`fmt()` pointer, a condition-variable handshake forces a second thread to
`fmt()` a different equal-length string before the main thread reads its pointer
back, and the captured contents are asserted intact. It fails reliably against
the old shared buffer and passes with the thread-local one. Verified against a
local `zeek --test` build; the full `util` suite is unaffected.

## AI disclosure

The fix, regression test, and root-cause analysis were drafted with Claude Code
(Opus 4.8). I reviewed the change, built Zeek locally, and verified the
regression test fails without the fix and passes with it before submitting.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -2875,7 +2875,7 @@ TEST_SUITE("util") {
 
         std::mutex m;
         std::condition_variable cv;
-        bool captured = false; // main thread has captured its fmt() pointer
+        bool captured = false;  // main thread has captured its fmt() pointer
         bool clobbered = false; // helper thread has completed its fmt() call
 
         const char* p = fmt("%s", a.c_str());
--- a/src/util.cc
+++ b/src/util.cc
@@ -2855,55 +2855,22 @@ TEST_SUITE("util") {
         CHECK_FALSE(approx_equal(qnan, -qnan));
     }
 
-    // Regression test for the copy_string() buffer-overflow crash. util::fmt()
-    // returns a pointer into a reused buffer; that buffer must be per-thread.
-    // The main thread captures fmt()'s result and, while still holding it, a
-    // second thread calls fmt() with different content. With a process-global
-    // static buffer the second call clobbers the first thread's string in
-    // place -- the root cause of the crash when the main thread holds a fmt()
-    // pointer across a scheduling point (e.g. a log writer backend starting).
-    // With a per-thread buffer the captured pointer is unaffected.
-    //
-    // The handshake makes the interleaving deterministic: the helper's fmt()
-    // always runs after the main thread has captured its pointer and before
-    // the main thread reads it. Both strings have the same length so the
-    // static buffer is overwritten in place (never reallocated), keeping the
-    // failure a clean assertion rather than a dangling-pointer crash.
-    TEST_CASE("vfmt_thread_safety") {
-        const std::string a(512, 'A');
-        const std::string b(512, 'B');
+    TEST_CASE("fmt_thread_safety") {
+        const auto* a = "AAAAA";
 
-        std::mutex m;
-        std::condition_variable cv;
-        bool captured = false;  // main thread has captured its fmt() pointer
-        bool clobbered = false; // helper thread has completed its fmt() call
+        // Capture a view into the buffer in `fmt`.
+        std::string_view p = fmt("%s", a);
+        REQUIRE_EQ(p, std::string_view(a));
 
-        const char* p = fmt("%s", a.c_str());
+        // Another thread uses `fmt` which updates its internal buffer.
         {
-            std::lock_guard<std::mutex> lk(m);
-            captured = true;
-        }
-        cv.notify_all();
-
-        std::thread helper([&] {
-            std::unique_lock<std::mutex> lk(m);
-            cv.wait(lk, [&] { return captured; });
-            fmt("%s", b.c_str());
-            clobbered = true;
-            lk.unlock();
-            cv.notify_all();
-        });
-
-        {
-            std::unique_lock<std::mutex> lk(m);
-            cv.wait(lk, [&] { return clobbered; });
+            std::jthread([=] { REQUIRE_NE(fmt("%s", "BBBBB"), std::string_view(a)); });
         }
 
-        CHECK(std::string(p) == a);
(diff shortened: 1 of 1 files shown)


src/util.cc
```diff
@@ -1255,27 +1258,27 @@ const char* fmt_bytes(const char* data, int len) {
...
-    static char* buf = nullptr;
-    static int buf_len = 1024;
-
-    if ( ! buf )
-        buf = reinterpret_cast<char*>(safe_malloc(buf_len));
+    // Per-thread rather than a single process-global buffer: child threads
+    // (e.g. log writer/input reader backends) format concurrently with the
+    // main thread, which holds the returned pointer across a copy_string().
+    // A shared buffer would let one thread overwrite or reallocate it mid-copy.
+    // The std::vector frees itself at thread exit, so this leaks nothing.
```
bbannier: This talks about things not really happening here, and `copy_string` which doesn't even use `util:fmt`. 

```suggestion
    // This function returns a pointer to an internal buffer. Use thread-local
    // storage for the underlying value so invocations from different threads
    // get a stable value.
```

src/util.cc
```diff
@@ -2851,6 +2854,55 @@ TEST_SUITE("util") {
...
+        {
+            std::unique_lock<std::mutex> lk(m);
+            cv.wait(lk, [&] { return clobbered; });
+        }
+
+        CHECK(std::string(p) == a);
+
+        helper.join();
+    }
 }
```
bbannier: This test makes sense, but does so in a much to verbose way, e.g., it explicitly controls when it joins the helper thread (i.e., also when it finishes its work), but then also adds extra notifications on top. Similarly it controls when the helper thread starts, but still notifies. None of that should be needed.

The comment talks about some previous state of the code so it probably made sense during development, but after this is merged it seems not so useful.

I believe something like the following would also do the job:

```suggestion
    TEST_CASE("fmt_thread_safety") {
        const auto* a = "AAAAA";

        // Capture a view into the buffer in `fmt`.
        std::string_view p = fmt("%s", a);
        REQUIRE_EQ(p, std::string_view(a));

        // Another thread uses `fmt` which updates its internal buffer.
        {
            std::jthread([=] { REQUIRE_NE(fmt("%s", "BBBBB"), std::string_view(a)); });
        }

        // Uses of `fmt` from different threads don't clobber each others internal buffers.
        CHECK_EQ(p, std::string_view(a));

    }
```

bbannier (commented):
Many functions in Zeek's API are not threadsafe, but it probably makes sense to at least try for simple utility functions like `util::fmt`, so this looks good to me.

With this change we could require up to num_threads more storage for internal storage of this function since we never shrink the internal buffer, and users can control how large that buffer is, but that is probably fine most of the time.

Since this code looks like it was generated by an overly verbose LLM, I left a few cleanup suggestions.

awelzel:
> If a log-writer or input-reader backend thread calls util::fmt()

Is this within the Zeek code base, or an external plugin? It's invalid usage of the API. I'd prefer to add something to detect the wrong thread calling this function (had a patch for this a long while ago, could be enabled just in DEBUG mode) so it's detected very early, rather than making this one function thread safe. The docs of the function could be more explicit about not being thread-safe.

This might be an unpopular opinion though, but I do like the fact that we can assume single-thread of execution in most places.

EDIT:

There's also another quirk of this API that has nothing to do with threads in that the returned pointer isn't stable (in the old an the new version IIUC):
```
char *p1 = util::fmt("... large %s", p);
char *p2 = util::fmt("... much larger %s", p)

// p1 might now be invalid, so don't touch!
```

bbannier:
Agree that this function isn't that great and probably should not be called by users with what is available to them with C++20, but it is part of the public API. Its usage pattern with the returned pointer being valid at least as long as no other call is made matches some C APIs (safe usage would copy to an owned buffer on return if the value needs to stick around). The missing thread safety is still not that great, even for Zeek's own internal use, since there is no safe way to work around that as caller (value might get clobbered before it can be copied), and it seems worthwhile fixing since the cost seems negligible to me (e.g., much cheaper than having each caller manage their own buffer).

stefangotz:
OK, I'll take another look on Monday. If the general preference is to fix this caller-side (and maybe make the documentation more explicit), I'm happy to go that way.

stefangotz:
Superseded by #5647

awelzel:
> If the general preference is to fix this caller-side

I had been pondering my stance it seems useful to have `fmt()` safer, but then I think it should also be done for `fmt_access_time()`, `fmt_bytes()` and `calculate_unique_id()` (?) and have a policy of "helpers in util.h are thread-safe".

The other reason I had pushed back is it possibly being a precedence for expecting more thread-safety in other places and that seems like a bigger topic to discuss.

> Superseded by https://github.com/zeek/zeek/pull/5647

Thanks and ouch!

PR #5641, opened 2026-07-03
Speed up WebSocketShim unit tests

bbannier:
These tests were using pretty large timeout values which made them the main contributors to the unit test suite's runtime. There is no reason for them to be so large, so we can decrease them substantially to speed up the test suite (on my machine before: 2.2s, after: 0.8s). With that change the runtime of the unit test suite is limited by how fast Zeek can set itself up (parse scripts, set up globals, ...).

### DIFF
--- a/src/broker/WebSocketShim.cc
+++ b/src/broker/WebSocketShim.cc
@@ -221,8 +221,8 @@ TEST_CASE("tests") {
         return broker::make_data_message(broker::topic{topic}, msg);
     };
 
-    auto expect_msg_timeout = 1000ms;
-    auto expect_no_msg_timeout = 150ms;
+    constexpr auto expect_msg_timeout = 2ms;
+    constexpr auto expect_no_msg_timeout = 5ms;
 
     SUBCASE("endpoint-publish") {
         // Publishing through the endpoint is visible to hubs, but not subscribers.


src/broker/WebSocketShim.cc
```diff
@@ -221,8 +221,9 @@ TEST_CASE("tests") {
         return broker::make_data_message(broker::topic{topic}, msg);
     };
 
-    auto expect_msg_timeout = 1000ms;
-    auto expect_no_msg_timeout = 150ms;
+    constexpr auto expect_msg_timeout = 2ms;
+    constexpr auto expect_no_msg_timeout = 1ms;
+    static_assert(expect_msg_timeout > expect_no_msg_timeout, "test assumption violated");
```
awelzel:
```suggestion
```
Don't think this violates any assumptions, I think 2ms and 15ms would be fair.

src/broker/WebSocketShim.cc
```diff
@@ -221,8 +221,9 @@ TEST_CASE("tests") {
         return broker::make_data_message(broker::topic{topic}, msg);
     };
 
-    auto expect_msg_timeout = 1000ms;
-    auto expect_no_msg_timeout = 150ms;
+    constexpr auto expect_msg_timeout = 2ms;
```
awelzel, line 224: Hmm, I remember tweaking these a long time back, but we can always revert if things start failing spuriously here. The `expect_msg_timeout` should have no impact on total runtime when there's no actual failures.

Lowering `expect_no_msg_timeout` to too low can result in false negatives more easily if messages are actually delivered, but it takes more than 1msec for them to arrive, so I'd probably opt for 15msec which is still a 10x reduction in sleep time.

Yes, these are more integration than unit tests and specific to Broker rather than having anything to do with Zeek proper.

awelzel (approved):

awelzel (approved):
@bbannier - re-approving. I was okay with whatever you were planning to do here, so feel free to merge :-)

FWIW, there is another pattern for slow tests employed which is to skip them except for in CI. That's done for the DNS_Mgr tests and ``--no-skip``.

https://github.com/zeek/zeek/blob/abcc93f700028f28896466fe6c7ea02d271f7dd9/src/DNS_Mgr.cc#L1422

https://github.com/zeek/zeek/blob/abcc93f700028f28896466fe6c7ea02d271f7dd9/ci/test.sh#L65

bbannier:
> @bbannier - re-approving. I was okay with whatever you were planning to do here, so feel free to merge :-)
> 
> FWIW, there is another pattern for slow tests employed which is to skip them except for in CI. That's done for the DNS_Mgr tests and `--no-skip`.
> 
> https://github.com/zeek/zeek/blob/abcc93f700028f28896466fe6c7ea02d271f7dd9/src/DNS_Mgr.cc#L1422
> 
> https://github.com/zeek/zeek/blob/abcc93f700028f28896466fe6c7ea02d271f7dd9/ci/test.sh#L65

Thanks, I missed the approval. In general I'd really like to avoid running tests in CI only since that can hide stuff, especially when people get too impatient to wait for CI to finish.

PR #5644, opened 2026-07-03
Support IN_ANYWHERE as wildcard for intel notices

J-Gras:
I found that `IN_ANYWHERE`, when used in files for ingestion, does not trigger notices for every location. I expected it to work as a wildcard.

AI usage: None


scripts/policy/frameworks/intel/do_notice.zeek
```diff
@@ -30,7 +30,7 @@ event Intel::match(s: Seen, items: set[Item])
 	for ( item in items )
 		{
 		if ( item$meta$do_notice &&
-			(! item$meta?$if_in || s$where == item$meta$if_in) )
+			(! item$meta?$if_in || item$meta$if_in == IN_ANYWHERE || s$where == item$meta$if_in) )
```
awelzel: Not sure it's better, but could we set `IN_ANYWHERE` as default for `if_in` in the record type above and drop `&optional`. Seems it should have the same semantics?

That'd allow to drop the `! item$meta?$if_in` and read a bit easier.

awelzel (approved):

PR #5646, opened 2026-07-04
Added permessage-deflate state tracking to WebSocket analyzer

### LINKED ISSUE #5642: WebSocket analyzer: Support RFC-7692 per-message compression
The per-message extension is popular and many clients and servers default to enabling it. Zeek does not transparently decompress such WebSocket messages/frames, but it'd be nice if it did.

RFC: https://datatracker.ietf.org/doc/html/rfc7692

This issue is to capture extending the BinPAC and Spicy WebSocket analyzers to support the per-message extension. Because the extension is negotiated during the WebSocket handshake in the HTTP analyzer, this information needs to be somehow passed to the analyzer instances. The entry point for this is here:

https://github.com/zeek/zeek/blob/e2466436605631a65d3e84296b7670c86c8c10c4/src/analyzer/protocol/websocket/functions.bif#L22-L46

https://github.com/zeek/zeek/blob/e2466436605631a65d3e84296b7670c86c8c10c4/src/analyzer/protocol/websocket/WebSocket.cc#L42-L92

For BinPAC the analyzer/interpreter class can probably be extended via C++ functions and members to configure that, for Spicy it'd need some research how to pull it off. The Spicy SSL analyzer may have a pattern, or can be skipped/unsupported initially if too involved.

Approach:
* Review existing analyzers.
* Collect/create a few pcaps where client and server negotiate the per-message extension.
* Figure out how to pass the extension negotiation result (from the HTTP headers parsed in the HTTP analyzer) down to the WebSocket analyzer (e.g. analyzer->EnablePerMessageCompression()`).
* Extend the analyzers and add tests that show websocket_frame_data()` contains decompressed payload and that forwarding to downstream analyzers or DPD also uses decompressed payload.

JustFirefly:
I'm opening this as a draft to show my progress on implementing permessage-deflate support in the WebSocket analyzer. Currently, I have successfully updated the BIF configuration and set up the state-passing plumbing.

I would appreciate early feedback on the architectural approach (specifically, the way I'm passing the negotiation state from the HTTP analyzer) before I proceed further with the decompression logic.

[x] Added permessage_deflate field to WebSocket::AnalyzerConfig record

[x] Updated BIF and regenerated C++ headers

[ ] Implementing decompression logic in WebSocket.cc

[ ] Adding btest cases for verification

Related Issue: Fixes #5642


src/analyzer/protocol/websocket/functions.bif
```diff
@@ -42,5 +42,12 @@ function __configure_analyzer%(c: connection, aid: count, config: WebSocket::Ana
 	if ( ! ws_analyzer->Configure({zeek::NewRef{}, config->AsRecordVal()}) )
 		return zeek::val_mgr->False();
 
+	#Added permessage_deflate functionality.
+	auto pm_deflate_val = config->GetFieldOrDefault("permessage_deflate");
```
awelzel: Think you should move this hunk into `WebSocket_Analyzer::Configure()` where the `use_dpd` and `analyzer` fields are processed already.

Then, for the BinPAC variant, add something like this around lines 53-56:
```
interp->EnablePerMessageCompression();
```

You can extend the WebSocket_Conn class of `interp` with members in the .pac files and within the `proc_*()` methods you can then get to those via `connection()->HasPerMessageCompressionEnabled()`. Something like this:

```
diff --git a/src/analyzer/protocol/websocket/websocket.pac b/src/analyzer/protocol/websocket/websocket.pac
index 4e710e2712..6c538c95ec 100644
--- a/src/analyzer/protocol/websocket/websocket.pac
+++ b/src/analyzer/protocol/websocket/websocket.pac
@@ -18,6 +18,17 @@ analyzer WebSocket withcontext {
 connection WebSocket_Conn(zeek_analyzer: ZeekAnalyzer) {
        upflow = WebSocket_Flow(true);
        downflow = WebSocket_Flow(false);
+
+       function EnablePerMessageCompression(): void
+               %{
+               // ...
+               %}
+
+       function HasPerMessageCompressionEnabled(): void
+               %{
+               // ...
+               %}
+
 };
```

For the Spicy analyzer, I'm not sure how to pass that information and there's no established pattern I believe. There have been previous in-band approaches: Pass one byte to the analyzer that's then read as the value for enable_per_message_compression, but I don't think that's great because the input stream isn't pure WebSocket anymore then. If we could pre-populate a `%context` instance, I think that'd be nice, but I haven't looked in earnest. Feel free to focus on the BinPAC variant for now.

```
+type Context = struct {
+    # Must be pre-populated by host application based on HTTP header info.
+    enable_per_message_compression: bool;
+};
+
 public type Messages = unit {
+    %context = Context;
+
     : Message[];
 };
```

ws_traffic_generator.py
```diff
@@ -0,0 +1,50 @@
+import asyncio
```
awelzel, line 1: This is great! If you move this down into `testing/btest/Traces/websocket` that'd be a better fit.

We recently added some pretty detailed info and instructions about [Zeek's test suite to the documentation](https://docs.zeek.org/en/master/advanced/devel/btest.html#testing-with-btest). Hope that helps.

awelzel (commented):
@JustFirefly - that looks like the right direction! Thanks for tackling this and asking for feedback early. I left you some comments, hope they help.

JustFirefly:
@awelzel Thanks for the detailed pointers!

I've gone ahead and applied this feedback in my latest commit. I moved the configuration extraction over to `WebSocket_Analyzer::Configure()` and wired up the state tracking inside the BinPAC `WebSocket_Conn` class (`EnablePerMessageCompression()` and `HasPerMessageCompressionEnabled()`).

My next step is to tackle the actual payload decompression in the `proc_*()` methods using Zeek's zlib wrappers. I'll reach out if I hit any snags there!

JustFirefly:
I have completed everything, I have also tested the functionality with the pcap generated and they produced the expected decompressed message.

src/analyzer/protocol/websocket/websocket-analyzer.pac
```diff
@@ -78,22 +78,35 @@ refine flow WebSocket_Flow += {
...
+		
+		if (connection()->HasPerMessageCompressionEnabled() && rsv1_ )
 			{
-			auto data_val = zeek::make_intrusive<zeek::StringVal>(data.length(), reinterpret_cast<const char*>(data.data()));
-			zeek::BifEvent::enqueue_websocket_frame_data(connection()->zeek_analyzer(),
-			                                             connection()->zeek_analyzer()->Conn(),
-			                                             is_orig(),
-			                                             std::move(data_val));
+			//printf("DEBUG: Decompressing frame. Enabled=true, RSV1=true\n");
+			connection()->DecompressPayload(reinterpret_cast<const unsigned char*>(data.data()), data.length(), is_orig());
```
awelzel: I haven't looked very closely, but DecompressPayload() should preferably return the decompressed data, then you can re-assign `data` and raising of `websocket_frame_data` and forwarding continues to work as before.

I mentioned the `websocket_frame_data()` even somewhere for testing, but the idea is more general: The decompressed data should also be forwarded to child analyzers.

scripts/base/protocols/websocket/main.zeek
```diff
@@ -140,8 +147,16 @@ event http_header(c: connection, is_orig: bool, name: string, value: string)
...
+			# if the response we recieved from the server contains permessage-deflate then it would mean the 
+			# negotiation was successful. otherwise the default would be to have the negotiation as unsuccessful (f).
+			if ( "permessage-deflate" in to_lower(value))
+				{
+				print "debug: negotiating permessage-deflate!";
+				ws$permessage_deflate = T;
+				}
 			}
-		else if ( name == "SEC-WEBSOCKET-ACCEPT" )
+		else if ( name == "sec-websocket-accept" )
```
awelzel: Any reason to change this? The header names should all be upper-case normalized.

src/analyzer/protocol/websocket/WebSocket.h
```diff
@@ -26,12 +28,23 @@ class WebSocket_Analyzer : public analyzer::tcp::TCP_ApplicationAnalyzer {
...
+        if ( interp ) {
+            fprintf(stderr, "DEBUG: C++ EnablePerMessageCompression called. Calling interp->Enable()...\n");
+            interp->EnablePerMessageCompression(); 
+        } else {
+            fprintf(stderr, "DEBUG: C++ EnablePerMessageCompression called, but INTERP IS NULL!\n");
+        }
+        }
     static zeek::analyzer::Analyzer* Instantiate(Connection* conn) { return new WebSocket_Analyzer(conn); }
-
+    static zeek::Tag Tag;
```
awelzel: It doesn't seem like `Tag` is used anywhere. Could this be dropped again?

scripts/base/protocols/websocket/main.zeek
```diff
@@ -74,6 +76,11 @@ export {
 
 redef record connection += {
 	websocket: Info &optional;
+	ws_deflate_negotiated: bool &default=F;
```
awelzel: This doesn't seem to be used. What's the intent for this?

test_decompression.zeek
```diff
@@ -0,0 +1,4 @@
+event websocket_frame_data(c: connection, is_orig: bool, data: string)
```
awelzel, line 1: For a complete submission, it'd be good to add separate btests including baselines, similar to the existing tests.

You can run the current WebSocket tests as follows after building Zeek:
```
$ btest -A -d -j scripts/base/protocols/websocket 
[  0%] scripts.base.protocols.websocket.fragmented-http-reply ... ok
[ ... ] 
[  0%] scripts.base.protocols.websocket.events-spicy ... ok
[ 27%] scripts.base.protocols.websocket.wstunnel-ssh-spicy ... ok
[  0%] scripts.base.protocols.websocket.events ... ok
all 15 tests successful
```

You can find some instructions for writing tests here: https://docs.zeek.org/en/master/advanced/devel/btest.html


Optimally, different variants of compressed WebSocket traffic (multiple websocket messages, websocket messages with single or multiple frame, and maybe tunneled SSH or HTTP traffic inside a compressed WebSocket connection - `wstunnel` might be able to do that) would be interesting to ensure good test coverage.

src/analyzer/protocol/websocket/websocket-protocol.pac
```diff
@@ -146,4 +149,8 @@ flow WebSocket_Flow(is_orig: bool) {
 
 		return len;
 		%}
+	function EnablePerMessageCompression(): bool
```
awelzel: This doesn't seem to have any effect.

awelzel (commented):
> I have completed everything, I have also tested the functionality with the pcap generated and they produced the expected decompressed message.

It looks like it's getting there! I made a few comments where I was confused and pointed out some additional steps that'd be needed for verifying the functionality in CI (creating a few btests). I only did a high-level skim without trying to run it myself or dig into the details.

Would you want to continue iterating on this PR, or is this becoming larger than you initially hoped for? It's definitely not a trivial beginner task, but it's a nice feature :-)

JustFirefly:
Thanks for the feedback @awelzel! I've updated the PR to address all your points:

1. Cleaned up the unused variables/tags and reverted the header normalisation.
2. Refactored DecompressPayload() so it strictly handles decompression and returns the buffer, it should allow the existing analyzer logic to handle the event raising and stream forwarding as suggested.
3. Added 4 btests with baselines for permessage-deflate, including tests for basic frames, Jupyter traffic, and wstunnel HTTP/SSH traffic.

scripts/base/protocols/websocket/main.zeek
```diff
@@ -76,13 +78,17 @@ redef record connection += {
 	websocket: Info &optional;
 };
 
+redef record WebSocket::AnalyzerConfig += {
+    permessage_deflate: bool &default=F;
+};
```
awelzel, line 83: This is fine/better to add into the original type in `init-bare.zeek` file directly rather than via a `redef` here.

src/analyzer/protocol/websocket/functions.bif
```diff
@@ -1,5 +1,7 @@
 %%{
 #include "zeek/analyzer/protocol/websocket/WebSocket.h"
+#include "zeek/analyzer/Component.h"
+#include "zeek/analyzer/Manager.h"
```
awelzel, line 4: Could you revert all changes (these includes and the whitespace changes below) in `websocket/functions.bif`? They shouldn't be needed anymore.

src/analyzer/protocol/websocket/Plugin.cc
```diff
@@ -10,7 +10,7 @@ namespace zeek::plugin::detail::Zeek_WebSocket {
 
 class Plugin : public zeek::plugin::Plugin {
 public:
-    zeek::plugin::Configuration Configure() override {
+zeek::plugin::Configuration Configure() override {
```
awelzel, line 13: This is a spurious formatting change. Could you setup `pre-commit` and/or `clang-format` in your development environment? The idea is that the code will be auto formatted before you push (optimally just on saving in your editor/ide).

testing/btest/scripts/base/protocols/websocket/permessage-deflate-wstunnel-ssh.zeek
```diff
@@ -0,0 +1,11 @@
+# @TEST-DOC: Test WebSocket permessage-deflate decompression with wstunnel SSH traffic.
+# @TEST-EXEC: zeek -C -r $TRACES/websocket/wstunnel-ssh.pcap %INPUT > output.txt
```
awelzel, line 2: Similar comment as for the HTTP test. The pcap doesn't contain compressed Websocket traffic.

testing/btest/scripts/base/protocols/websocket/permessage-deflate-wstunnel-http.zeek
```diff
@@ -0,0 +1,11 @@
+# @TEST-DOC: Test WebSocket permessage-deflate decompression with wstunnel HTTP traffic.
+# @TEST-EXEC: zeek -C -r $TRACES/websocket/wstunnel-http.pcap %INPUT > output.txt
```
awelzel, line 2: This is the original `wstunnel-ssh.pcap` and it doesn't negotiate any `permessage-deflate` extension, so this test doesn't improve coverage of the new functionality.

I had used [wstunnel](https://github.com/erebe/wstunnel) previously to create this tunneled connection, but it doesn't seem it supports permessage-deflate. It might be easiest to take a Python script similar to the `ws_traffic_generator.py` and make the payload messages be valid `HTTP` traffic.

testing/btest/scripts/base/protocols/websocket/permessage-deflate-jupyter.zeek
```diff
@@ -0,0 +1,11 @@
+# @TEST-DOC: Test WebSocket permessage-deflate decompression with complex Jupyter traffic.
+# @TEST-EXEC: zeek -C -r $TRACES/websocket/jupyter-websocket.pcap %INPUT > output.txt
```
awelzel, line 2: In this PCAP, while the client does send the `permessage-deflate` extension, the server doesn't, so there's no compression enabled and this new test doesn't add coverage.

src/analyzer/protocol/websocket/WebSocket.cc
```diff
@@ -54,7 +57,13 @@ bool WebSocket_Analyzer::Configure(zeek::RecordValPtr config) {
         interp = std::make_unique<binpac::WebSocket::WebSocket_Conn>(this);
         effective_analyzer = this;
     }
-
+    auto pm_deflate_val = config->GetFieldOrDefault("permessage_deflate");
```
awelzel, line 60: It'd be nicer to use the pattern fro lines 38-40 above to gather the offset/index value for `permessage_deflate` field and use that here.

src/analyzer/protocol/websocket/WebSocket.cc
```diff
@@ -8,6 +8,9 @@
 #include "zeek/analyzer/protocol/pia/PIA.h"
 #include "zeek/analyzer/protocol/websocket/events.bif.h"
 
+#include "analyzer/protocol/websocket/WebSocket.h"
```
awelzel, line 11: This is redundant with the include on line 3.

src/analyzer/protocol/websocket/websocket-analyzer.pac
```diff
@@ -79,21 +79,42 @@ refine flow WebSocket_Flow += {
 				d[i] = d[i] ^ masking_key_[masking_key_idx_++ % masking_key_.size()];
 			}
 
-		if ( websocket_frame_data )
+		const u_char* payload_data = reinterpret_cast<const u_char*>(data.data());
+		int payload_len = chunk_emit_len_;		
+
+		if (connection()->HasPerMessageCompressionEnabled() && rsv1_ )
+			{
+			if (connection()->DecompressPayload(reinterpret_cast<const unsigned char*>(data.data()), data.length()))
```
awelzel, line 87: I should've looked at this a bit earlier.

Going through the RFC a bit, the message is compressed and thereafter possibly fragmented, so unfortunately can't assume every frame is de-compressible standalone.

This also means there's a need to keep the decompression state for client and server side separately, likely attached to the `context.flow`.

Last, there's an existing "ZIP analyzer (naming is unfortunate)" at analyzer/protocol/zip/ZIP.h that's used in the HTTP analyzer for decompression of gzip/deflate:

https://github.com/zeek/zeek/blob/f9601bc3ad6d1cce0afe36f47e3763a98ccdccc8/src/analyzer/protocol/http/HTTP.cc#L205-L220

If we could do a similar pattern to HTTP (forwarding compressed data to a ZIP_Analyzer instance and the OutputHandler then raising `websocket_frame_data()` and forwarding to downstream analyzers, that would be cleanest.

Sorry you spent the time implementing the low-level `DecompressPayload()` function, I should have provided better guidance.

awelzel (commented):
Hey @JustFirefly / @lolsda - I have some more comments. Some of the new tests aren't providing additional coverage because the original PCAPs do not use the extension. For the DecompressPayload() logic, it needs to be per originator and responder side and should re-use an existing "support analyzer" (see more detailed comment there).

I realize this is getting fairly involved and the "good first issue" tag might have been a bit optimistic. Is this forth and back useful to you and you're learning something, or is it becoming frustrating? It might take a few more rounds before this can be merged and I just want to make sure it's not becoming an annoyance. I'm okay doing this for a few more rounds :-)

A bit related to the above, we have an [AI policy](https://github.com/zeek/zeek/blob/master/AI_POLICY.md) and if you do use a harness/agent/llm, please disclose its usage as I'm for now assuming there's a human on the other end :-)

JustFirefly:
@awelzel Thank you for the detailed feedback and for pointing me towards the ZIP_Analyzer. After looking at the required changes for the decompression state and the PCAP test coverage, I realise this requires more time than I can commit right now.

Regarding your note on the AI policy; I want to disclose that I used an LLM (Gemini 3.1 Flash-Lite) to assist with exploring the codebase and giving me suggestions on potential fixes. Any code implementations were done by me.

Since I won't be able to finish the implementation, I'm going to close this PR so it doesn't clutter the queue. I really appreciate your help and patience throughout the process, I learned a lot about programming in large codebases and how Zeek's analyzer framework operates under the hood.

Hopefully, my initial commits can serve as a useful starting point for whoever picks up the permessage-deflate feature next. Thanks again :D

PR #5647, opened 2026-07-06
util/fmt: Use thread-safe Fmt() in backend threads; assert main-thread-only contract

stefangoetzcorelight:
`util::fmt()`/`vfmt()` format into a process-global buffer, so they are only safe on the main thread. A backend thread calling `fmt()` can clobber that buffer while the main thread still holds a returned pointer (e.g. across the `copy_string()` in `Writer`/`ReaderFrontend`), corrupting it.

- Switch the three in-tree callers running on `Reader`/`WriterBackend` (`MsgThread`) threads to the per-thread `BasicThread::Fmt()`
- Also switch a possibly concurrent call to the non-reentrant `strerror()` to the thread-safe `Strerror()`.
- Add a DEBUG-only assertion in `vfmt()` that aborts if called off the main thread.
- Document the thread-safety and pointer-lifetime contract in `util.h`.

Tested (Debug, Spicy on): `zeek --test -ts=util`, log writing + rotation, and ascii input with a path prefix.

*AI disclosure*

Changes and tests were initially implemented and executed by Claude Code (Opus 4.8). I have manually reviewed the changes and tests and rewritten the source code comments.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -46,6 +46,7 @@
 #include <ranges>
 #include <sstream>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include "zeek/3rdparty/ConvertUTF.h"
@@ -1255,6 +1256,18 @@ const char* fmt_bytes(const char* data, int len) {
 }
 
 const char* vfmt(const char* format, va_list al) {
+#ifdef DEBUG
+    // buf is process-global, so vfmt() is not thread safe.
+    // By convention, it must be called only from the main thread.
+    // The first call is known to be performed by the main thread during single-threaded startup.
+    // Check subsequent calls against that baseline.
+    static const std::thread::id main_thread_id = std::this_thread::get_id();
+    if ( std::this_thread::get_id() != main_thread_id ) {
+        fprintf(stderr, "zeek::util::fmt() called off the main thread; use threading::BasicThread::Fmt() or std::format()\n");
+        abort();
+    }
+#endif
+
     static char* buf = nullptr;
     static int buf_len = 1024;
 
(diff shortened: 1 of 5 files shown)


awelzel (commented):
LGTM, thanks! Mind splitting the Fmt() / Strerror() fixes into a separate commit from the vfmt() assert? I'm thinking the former should definitely be back-ported to 8.2.x and 8.0.x, possibly also the assert, but less sure, so having separate commits would be preferred.

stefangoetzcorelight:
@awelzel I've split out the DEBUG assertion into a separate commit.

awelzel (approved):
Thanks.

src/util.cc
```diff
@@ -1255,6 +1256,18 @@ const char* fmt_bytes(const char* data, int len) {
...
 
 const char* vfmt(const char* format, va_list al) {
+#ifdef DEBUG
+    // buf is process-global, so vfmt() is not thread safe.
+    // By convention, it must be called only from the main thread.
+    // The first call is known to be performed by the main thread during single-threaded startup.
+    // Check subsequent calls against that baseline.
+    static const std::thread::id main_thread_id = std::this_thread::get_id();
+    if ( std::this_thread::get_id() != main_thread_id ) {
+        fprintf(stderr, "zeek::util::fmt() called off the main thread; use threading::BasicThread::Fmt()\n");
```
awelzel: This will need to wait for: https://github.com/corelight/zeekjs/pull/129

I still think it's okay to do, we could hint at `std::format()` as well.


```suggestion
        fprintf(stderr, "zeek::util::fmt() called off the main thread; use threading::BasicThread::Fmt() or std::format()\n");
```

stefangoetzcorelight: applied as suggested as an amendment to the assertion commit

awelzel:
I've merged this with a "warn-once style" modification so it's also visible in release builds and produces the message exactly once.

Issue #5643, opened 2026-07-03
Implement LogAscii::json_string_escaping


awelzel:
Implement the different policies discussed in https://github.com/zeek/zeek/discussions/5240

Seems making this specific to the LogAscii module for now is okay given the existing options `json_timestamps` and `json_include_unset_fields` in the LogAscii module. The escaping is specific to the ASCII writer when `LogAscii::use_json=T`.

This doesn't concern itself with what `print()` or `cat()` do. Maybe `to_json()` should receive a `StringEscapingPolicy` argument (or be switched to some extensible record style so the signature doesn't get too long), but this is separate IMO.

```
type JSON::StringEscapingPolicy: enum {
    LEGACY, # Default used by Zeek 9.0 with backslash confusion \xXX
    TSV,    # JSON strings escaped like TSV strings, all backslashes are double escaped \xXX
    PUA,    # Map non-printable bytes into Unicode Private-Use-Area \udb80\udcXX
    LATIN1  # Encode non-printable bytes as LATIN-1 unicode points \u00XX
};

module LogAscii;
export {
    ...
    const json_string_escaping = JSON::StringEscapingPolciy::TSV &redef;
}
```

~We rejected/excluded the the LATIN1 / `\u00xx` style (PR https://github.com/zeek/zeek/pull/5215), but given this is now meant to be configurable and after looking through this again, it strikes me as a decent compromise for those that know how to handle it: Far less churn and noise than TSV for backslash-heavy fields and more intuitive and less overhead than PUA when just looking at it.~

EDIT: After implementing, one issue persist: The byte `\xc0` will be represented as `\u00c0`, but if the original string contained `\xC3\x80` (the UTF-8 encoding of \u00c0`), it's ambiguous after parsing without any special handling, so I ended up removing the LATIN1 idea again.

Testing:

* Think extending this test to run through each of the escaping policies and creating a separate baseline would be a good example `scripts.base.frameworks.logging.ascii-json-utf8`

awelzel:
Closing after merge of #5650 .

PR #5650, opened 2026-07-06
Add Log::json_string_escape_policy

### LINKED ISSUE #3948: Discrepancy between JSON and TSV logs for non-printable bytes

awelzel:
Opening as draft: I'd like to have this (or something like this) in 9.0, so we can move this past us. This has no behavioral change other than having an answer to #3948.

The actual implementation is mostly using the existing pattern for the json_timestamps which are semi-great, but focus here should be on the scripting side (option and enum naming) and if we can agree on the framing, motivation and description in the NEWS entry.

Closes #3948 #5643

Relates to #5240


NEWS
```diff
@@ -51,6 +51,52 @@ New Functionality
...
+
+  Note that this mechanism is meant to cover the occasionally non-printable byte
+  or sometimes polluted string field. As a script writer, if a field you're logging is
+  knowingly binary data (raw packet data, crypto material, binary hash or checksum),
+  use ``encode_base64()`` or ``bytestring_to_hextr()`` explicitly on the field before
+  logging it and document it as base64 or hex encoded for downstream consumers.
+
+  The following JSON string escaping policies are supported as of Zeek 9.0:
+
+	LEGACY (default):
```
awelzel: FWIW, no particular reason for the "legacy" name, could definitely frame it more positively like "hex" or so.

scripts/base/init-bare.zeek
```diff
@@ -5786,6 +5786,15 @@ export {
...
 		TS_ISO8601,
 	};
+
+	## Escaping policy for non-printable non-UTF8 sequences.
+	type StringEscapePolicy: enum {
+		## Zeek 9.0 and prior escaped non-printable bytes as \\x32 without
+		## backslash stuffing, resulting in non-reversible encodings.
+		STRING_ESCAPE_POLICY_LEGACY,
+		STRING_ESCAPE_POLICY_TSV,
+		STRING_ESCAPE_POLICY_PUA,
```
ckreibich: Total nit: sort alphabetically.
```suggestion
		STRING_ESCAPE_POLICY_PUA,
		STRING_ESCAPE_POLICY_TSV,
```

src/threading/formatters/JSON.cc
```diff
@@ -180,8 +183,37 @@ void JSON::BuildJSON(zeek::json::detail::NullDoubleWriter& writer, Value* val, c
...
+                    break;
+                }
+                case STRING_ESCAPE_POLICY_PUA: {
+                    auto s = util::escape_string_for_json(sv, "\\udb80\\udc");
+                    writer.RawValue(s.data(), s.size(), rapidjson::kStringType);
+                    break;
+                }
+                case STRING_ESCAPE_POLICY_TSV: {
+                    // This should be roughly what we do in the ASCII formatter.
+                    // Should we cache desc on the formatter?
```
ckreibich: I'm in favor, yeah. It does save the constant default allocation, there aren't that many writers, and `Clear()` is fast and does clean up very large allocations.

NEWS
```diff
@@ -51,6 +51,52 @@ New Functionality
...
+  To change the escaping behavior globally, redefine the ``Log::json_string_escape_policy``
+  variable. Per filter configuration is possible by passing an "json_string_escape_policy"
+  entry via the ``config`` table of ``Log::Filter`` instances.
+
+	redef Log::json_string_escape_policy = JSON::StringEscapePolicy::TSV;
+
+  Note that this mechanism is meant to cover the occasionally non-printable byte
+  or sometimes polluted string field. As a script writer, if a field you're logging is
+  knowingly binary data (raw packet data, crypto material, binary hash or checksum),
+  use ``encode_base64()`` or ``bytestring_to_hextr()`` explicitly on the field before
```
```suggestion
  use ``encode_base64()`` or ``bytestring_to_hexstr()`` explicitly on the field before
```

NEWS
```diff
@@ -51,6 +51,52 @@ New Functionality
...
+  logging it and document it as base64 or hex encoded for downstream consumers.
+
+  The following JSON string escaping policies are supported as of Zeek 9.0:
```
ckreibich: You can drop thew "as of Zeek 9.0" here since it's in the release notes — or, alternatively, keep this quite short here and be detailed in the docs proper, perhaps [here](https://docs.zeek.org/en/current/frameworks/logging.html#ascii-writer). The length of this addition and subtlety has me thinking it should actually go there — it'll be easier to reference than a release notes entry.

NEWS
```diff
@@ -51,6 +51,52 @@ New Functionality
...
+
+	LEGACY (default):
+	  Zeek's previous behavior continues to be the default.
+
+	TSV:
+	  Strings are first escaped as they would for TSV logs. All control characters
```
```suggestion
	  Strings are first escaped as they would for TSV logs: all control characters
```
This is nitpicky but it clarifies that you're about to explain what the TSV logs approach means, rather than spelling out additions.

ckreibich (commented):
Nice Arne, thanks very much for tackling this. It all looks good to me. I'd indeed move the `ODesc` into the formatter, and I'm leaning toward the explanation going into the docs, and that's about it. The rest is nits.

NEWS
awelzel: Renamed to HEX after discussion with @ckreibich .

NEWS
awelzel: I moved most of the NEWS entry into a docs section.

scripts/base/init-bare.zeek
awelzel: Done and also commented.

scripts/base/init-bare.zeek
```diff
@@ -5824,6 +5824,28 @@ export {
...
+		## Byte \x07 becomes \\x07 and literal string \x07 also becomes \\x07
+		## in the resulting JSON string.
+		STRING_ESCAPE_POLICY_HEX,
+		## All control characters and non-printable bytes not part of valid UTF-8
+		## sequences are mapped into Unicode Supplementary Private Use Area-A
+		## (U+F0000..U+FFFFF) and encoded as surrogate pair:
+		## In a JSON string, the byte value \x07 will become "\udb80\udc07".
+		## The literal string \x07 will become "\\x07".
+		STRING_ESCAPE_POLICY_PUA,
+		## Strings are first escaped as they would for TSV logs: all control characters
```
ckreibich:
```suggestion
		## Strings are first escaped as in TSV logs: all control characters
```

ckreibich (approved):
Thanks Arne! 🚀 

If you have it in you, you could add a test for using "json_string_escape_policy" in a filter config map while we're on this — totally up to you.

Issue #5638, opened 2026-07-02
CircleCI TSAN builds fail


timwoj:
I'm not sure what to do with this one, so I'm going to document everything I've tried here. The ThreadSanitizer builds are currently failing on CircleCI. The symptom they show is that OpenSSL fails to configure:

```
CMake Error at cmake/OpenSSLTests.cmake:82 (message):
  OpenSSL library version does not match headers
Call Stack (most recent call first):
  CMakeLists.txt:1271 (include)
```

The error is a bit nonsensical, but that's just because it's a red-herring. CMakeConfigureLog.yml shows the actual reason:

```
stderr: |
  ThreadSanitizer: CHECK failed: tsan_platform_linux.cpp:290 "((personality(old_personality | ADDR_NO_RANDOMIZE))) != ((-1))" (0xffffffffffffffff, 0xffffffffffffffff) (tid=2778)
    Segmentation fault
  exitCode: "FAILED_TO_RUN"
```

Again, the error is a bit nonsensical, but the failure is because of the state of ASLR in the host that the Docker container runs under. I upgraded the version of Clang from 19 to 22 on the container, since that brings some better error messaging for what's happening here:

```
stderr: |
  FATAL: ThreadSanitizer: encountered an incompatible memory layout but was unable to disable ASLR (perhaps sandboxing is enabled?).
  FATAL: Please rerun with lower ASLR entropy, ASLR disabled, and/or sandboxing disabled.
exitCode: 66
```

https://github.com/google/sanitizers/issues/1716 documents a little better what's happening here. https://discuss.circleci.com/t/ubuntu-24-04-infrastructure-upgrade-for-docker-jobs/54464 suggests upgrading to a newer compiler to make things work, but that clearly didn't work here. Clang 22 is the current official release of Clang. We can't set the ASLR entropy or disable it in a Docker container, because doing so requires write access to `sysctl` values, and you don't have that by default in Docker.

timwoj:
It's possible that we could switch the ThreadSanitizer runs over to a custom job that uses a machine executor instead of the docker executor. That would give us more control over the Docker environment, allowing us to run in privileged mode and set the right options to disable ASLR. We would have to run the docker commands by hand, which is sort of messy. https://lkhill.com/circleci-docker-systemd/ has a good example of setting up something like this.

timwoj:
Fixed by https://github.com/zeek/zeek/pull/5651

PR #5651, opened 2026-07-06
Fix stall with OpenSSL check on Mac with ASAN

evantypanski:
Before, the OpenSSL check was stalling if I used `--sanitizers=address`. This is due to MacOS SIP, which causes the ASAN binaries to not load correctly. Instead, simply move the check before the sanitizer flags are set. ~Then, explicitly include the necessary directories so that the header is findable at this earlier stage.~

Include directories need a change in cmake submodule so that it doesn't change the build globally.

I used Claude to pin this down.

Should be good now, sorry bout the mishaps. I didn't fix the `spicyz` issue, but this seems more far-reaching.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -904,6 +904,11 @@ find_package(FLEX REQUIRED)
 find_package(BISON 2.5 REQUIRED)
 find_package(PCAP REQUIRED)
 find_package(OpenSSL 3.0 REQUIRED)
+
+# Run OpenSSL check before sanitizer flags are set so macOS SIP does not cause
+# the test to stall.
+include(OpenSSLTests)
+
 find_package(ZLIB REQUIRED)
 
 if (NOT BINARY_PACKAGING_MODE)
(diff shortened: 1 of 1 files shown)


evantypanski:
`spicyz` is also spinning during the build phase, seems like due to the same problem... :(

timwoj:
This could also fix the failure of the tsan builds on CI, since it's dying trying to do the OpenSSL checks with the sanitizer flags set. See https://github.com/zeek/zeek/issues/5638.

EDIT: Yep, I can confirm that this fixes running `./configure --sanitizers=thread` on one of our CI containers. It consistently fails otherwise.

0xxon:
I guess this was otherwise integrated into master?

(diff shows as empty)

evantypanski:
oops. I repushed based on master for CI changes, but I didn't follow my workflow so accidentally stashed the contents of this PR, not just my further explorations for `spicyz`. First time I've had `jj` actually make me do something silly, though it was my fault :)

timwoj (approved):
The TSan build is still failing, but at least it's making it through the configure step now. It's failing tests for reasons unrelated to this PR.

PR #5653, opened 2026-07-06
Cleanup/temporary disabling of zeek-testing-private on CI

timwoj:
CircleCI is having problems with the SSH keys related to cloning the zeek-testing-private repo, which is causing all builds to fail. Until that gets fixed, disable cloning the repo so that at least the rest of the build continues working. This also removes some Cirrus-isms from the script use to clone repos.

### DIFF
--- a/ci/init-external-repos.sh
+++ b/ci/init-external-repos.sh
@@ -29,26 +29,12 @@ fi
 make update-traces
 cd ..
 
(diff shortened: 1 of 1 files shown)


evantypanski (approved):
perhaps this also warrants a big scary TODO in there?

probably fine without it since it seems like it should hopefully be fixed pretty soon so at your discretion though

PR #5659, opened 2026-07-07
Upgrade Clang on CI to v22

timwoj:
This PR upgrades the version of Clang installed on the Ubuntu 24.04 CI image to Clang v22. This brings us up to close the latest version of Clang, as well as a much newer version of Clang-Tidy for that build.

### DIFF
--- a/.circleci/builds.yml
+++ b/.circleci/builds.yml
@@ -1119,6 +1119,7 @@ workflows:
       - linux-build:
           name: ubuntu-24.04
           docker_image: zeek/zeek-ci:ubuntu-24.04
+          ci_image_date: "20260702"
           requires: [fetch-test-traces, build-image-ubuntu-24.04]
           context:
             - zeek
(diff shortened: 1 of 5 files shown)


src/threading/SerialTypes.cc
```diff
@@ -218,7 +218,7 @@ static std::optional<std::vector<Value*>> read_elements(detail::SerializationFor
     // Only reserve up to 1024 pointers and rely on dynamic
     // reallocation of std::vector past that many elements.
     std::vector<Value*> vec;
-    vec.reserve(static_cast<size_t>(std::min(size, zeek_int_t(1024))));
+    vec.reserve(static_cast<size_t>(std::min(size, static_cast<zeek_int_t>(1024))));
```
awelzel:
```suggestion
    vec.reserve(std::min<size_t>(size, 1024));
```

Gemini suggested. The original one definitely didn't look great.

We check size < 0 early, so this should be okay.

.circleci/builds.yml
```diff
@@ -1065,6 +1065,7 @@ workflows:
       - linux-build:
           name: debian-12
           docker_image: zeek/zeek-ci:debian-12
+          ci_image_date: "20260708"
```
awelzel, line 1068: I cannot say that the repeated image date looks great (requiring to keep them in sync seems a bit brittle). If there was a way to get that value as output parameter from the build-image job (which uses it), that'd be nice, but didn't seem there exists a concept of output parameters.

awelzel (approved):

.circleci/builds.yml
timwoj: @ckreibich and I discussed this whole image-date thing during initial work on the Circle config. I initially had a script in there that would automatically pull the image date from the Dockerfile, etc, but it didn't feel explicit enough about what version it was using. I'll be honest though, I'm not a big fan of the current way either. Keeping track of the dates is annoying. Part of me wants to just remove the image caching altogether and let it just build the images for every build, but I don't know how many Circle credits that kind of change would cost us over time.

Issue #5656, opened 2026-07-07
Minor SMTP Record Definition Issues


malvidin:
There is a slight error here; the `helo`, `mailfrom` and `rcptto` values are properly extracted from the SMTP command of `HELO/EHLO`, `MAIL FROM:`, and `RCPT TO:`, but the documentation mentions the header.

https://github.com/zeek/zeek/blob/3c13d05eeb45c1d60dfc6cd1e51273edeaf58a10/scripts/base/protocols/smtp/main.zeek#L27-L32

I recommend something like this for `helo`, `mailfrom` and `rcptto`:

```zeek
		## Contents of the HELO/EHLO SMTP command.
		helo:              string          &log &optional;
		## Email address found in the MAIL FROM SMTP command.
		mailfrom:          string          &log &optional;
		## Email addresses found in the RCPT TO SMTP command(s).
		rcptto:            set[string]     &log &optional;
```

https://github.com/zeek/zeek/blob/3c13d05eeb45c1d60dfc6cd1e51273edeaf58a10/scripts/base/protocols/smtp/main.zeek#L43-L44

I recommend something like this for the `msg_id` documentation.

```zeek
		## Contents of the Message-ID header.
		msg_id:            string          &log &optional;
```

awelzel:
Hello @malvidin - thanks for catching and reporting these. Your proposals make a lot of sense. Would you be interested and setup for submitting a PR for these?

malvidin:
@awelzel, I added pull #5660 for this issue.

awelzel:
Closed by #5660.

PR #5660, opened 2026-07-08
Clarify SMTP documentation

malvidin:
Addresses #5656 
- Remove email header label for SMTP commands
- Expand Message-ID & Reply-To to match RFC822


awelzel (approved):
Thanks!

PR #5661, opened 2026-07-08
Protocol/file analysis bif doc fixes

0xxon:
This commit applies a range of small documentation fixes to some of the analyzer and file_analyzer bifs.

This includes minor formatting changes, fixed todos, typos, unfinished sentences, and a couple of incorrect statements.

This was made with the help of Claufe Opus 4.7 (which found the issues), and then manually edited. I understand and verified all of the changes.

Contains no code changes.

### DIFF
--- a/src/analyzer/protocol/smtp/events.bif
+++ b/src/analyzer/protocol/smtp/events.bif
@@ -20,7 +20,7 @@
 ##    mime_end_entity mime_entity_data mime_event mime_one_header mime_segment_data
 ##    smtp_data smtp_reply
 ##
-## .. note:: Zeek does not support the newer ETRN extension yet.
+## .. note:: Zeek does not support the ETRN extension.
 event smtp_request%(c: connection, is_orig: bool, command: string, arg: string%);
 
 ## Generated for server-side SMTP commands.
@@ -39,7 +39,7 @@ event smtp_request%(c: connection, is_orig: bool, command: string, arg: string%)
 ##
 ## code: The reply's numerical code.
 ##
-## cmd: TODO.
+## cmd: The SMTP command this reply pertains to.
 ##
 ## msg: The reply's textual description.
 ##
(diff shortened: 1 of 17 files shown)


src/file_analysis/analyzer/x509/functions.bif
```diff
@@ -132,7 +132,7 @@ bool check_hostname(std::string_view hostname, std::string_view certname)
 		return true;
 
 	// ok, now there is still the chance that it is a wildcard cert.
-	// We go according to RFC6128 here:
+	// We go according to RFC 6125 here:
```
awelzel, line 135:
```suggestion
	// We go according to :rfc:`6125` here:
```

src/analyzer/protocol/smtp/events.bif
```diff
@@ -20,7 +20,7 @@
 ##    mime_end_entity mime_entity_data mime_event mime_one_header mime_segment_data
 ##    smtp_data smtp_reply
 ##
-## .. note:: Zeek does not support the newer ETRN extension yet.
+## .. note:: Zeek does not support the ETRN extension.
```
awelzel, line 23: Should this just be dropped and have an issue if relevant? There's probably more unsupported extensions.

awelzel (approved):

src/file_analysis/analyzer/x509/functions.bif
0xxon: I thought about doing this. However this is in comment block where this doesn't get parsed and I think the other looks more human readable. The change is that the wrong RFC was being refered.

src/analyzer/protocol/smtp/events.bif
0xxon: It's probably helpful as a note if one is not intricately familiar with the protocol.

That doesn't negate that we should open an issue :)

PR #5663, opened 2026-07-08
Disable leak checking on cycle test

evantypanski:
I broke ASAN with #5593

### DIFF
--- a/testing/btest/language/cycle-print.zeek
+++ b/testing/btest/language/cycle-print.zeek
@@ -1,6 +1,9 @@
 # @TEST-DOC: Printing self-referential aggregate values via cycles does not crash and emits <cycle>. Regression test for GH #5407.
 #
-# @TEST-EXEC: zeek -b %INPUT >output
+# The test creates cycles. Since values are refcounted, that leaks memory.
+# Thus, we disable leak detection.
+#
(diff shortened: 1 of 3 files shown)


testing/btest/language/cycle-print.zeek
```diff
@@ -1,6 +1,9 @@
 # @TEST-DOC: Printing self-referential aggregate values via cycles does not crash and emits <cycle>. Regression test for GH #5407.
 #
-# @TEST-EXEC: zeek -b %INPUT >output
+# The test creates cycles. Since values are refcounted, that leaks memory.
+# Thus, we disable leak detection.
+#
+# @TEST-EXEC: ASAN_OPTIONS="$ASAN_OPTIONS,detect_leaks=0" zeek -b %INPUT >output
```
awelzel:
```suggestion
# @TEST-EXEC: ASAN_OPTIONS="$ASAN_OPTIONS:detect_leaks=0" zeek -b %INPUT >output
```

I think we're doing it wrong also in the other places - mind fixing them all? :-/

> Both space and colon can be used as environment variable separator.

https://github.com/google/sanitizers/issues/1794#issuecomment-2336996289

awelzel (approved):

testing/btest/language/cycle-print.zeek
evantypanski: at least it's not platform specific like path separators on Windows with `;` :)

PR #5667, opened 2026-07-09
doc/devel/cluster: Replace singe/multi node terminology with host

awelzel:
In conversations, I have often heard or said "multi-node cluster" to describe a Zeek cluster spanning multiple hosts (either physical hosts, VMs or even just containers in different network namespaces) and included this terminology in the docs which today I find misleading.

Steer the single-node / multi-node terminology to say host instead. In Zeek script, every Zeek process itself is termed a node and its name is stored in CLUSTER_NODE / Cluster::node. Describing the host on which a process runs also as node should be avoided IMO.

The management framework docs are pretty clear about process and node being the same. There is a notion of an "instance" there, but that's related to the agent and controller concepts IIUC and host seems fitting.

### DIFF
--- a/doc/advanced/devel/cluster/architectures.rst
+++ b/doc/advanced/devel/cluster/architectures.rst
@@ -60,10 +60,10 @@ Proxy, logger and worker processes can all be scaled independently.
 .. note::
 
    These processes are also called nodes and their role called node type.
-   In this section we attempt to stick with the process terminology. Other systems
-   use the term node to describe a complete host instead. Zeek itself has a concept
-   of single-host and multi-host clusters (see below), too, so this can all be a
-   bit confusing initially.
+   Here, we'll stick with the the process terminology. These terms all get
+   used in different ways based on the system; some may use node to describe
+   a complete host instead. Zeek uses the concept of single-host and multi-host
+   deployments for this, since node is just a process!
 
    On the script-level, the :zeek:see:`Cluster::node` variable contains
    ``manager``, ``worker-1``, ``proxy-1``, ``logger-1``, etc., allowing to
--- a/doc/advanced/devel/cluster/architectures.rst
+++ b/doc/advanced/devel/cluster/architectures.rst
@@ -59,16 +59,16 @@ Proxy, logger and worker processes can all be scaled independently.
 
 .. note::
 
-   These processes are also called nodes. And their role called node type. In this
-   chapter we attempt to stick with the process terminology. Many other systems
-   use the word node to describe a host running multiple services or processes and
-   Zeek itself has a concept of single-node and multi-node Zeek clusters (see below),
-   so this can be a bit confusing.
+   These processes are also called nodes and their role called node type.
+   In this section we attempt to stick with the process terminology. Other systems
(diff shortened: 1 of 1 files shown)


doc/advanced/devel/cluster/architectures.rst
```diff
@@ -59,16 +59,16 @@ Proxy, logger and worker processes can all be scaled independently.
...
-   These processes are also called nodes. And their role called node type. In this
-   chapter we attempt to stick with the process terminology. Many other systems
-   use the word node to describe a host running multiple services or processes and
-   Zeek itself has a concept of single-node and multi-node Zeek clusters (see below),
-   so this can be a bit confusing.
+   These processes are also called nodes and their role called node type.
+   In this section we attempt to stick with the process terminology. Other systems
+   use the term node to describe a complete host instead. Zeek itself has a concept
+   of single-host and multi-host clusters (see below), too, so this can all be a
+   bit confusing initially.
```
evantypanski: So the section changed its meaning a bit. Before it was "node is an overloaded term," now it's "Zeek's use of the term node is slightly different, use 'host' for that"

Honestly, I would be less wordy and just make this all extremely blunt (note the bolds when introducing a term):

> A **node** is one (of possibly many) process running Zeek. The role a node plays is its **type**. A **host** is a separate environment which may have many nodes.
> 
> Other systems may use "node" to refer to that separate environment, but in Zeek that would overload the term. For that, we will use the term "host."

that seems a little nicer and less confusing to me, but we could also tweak the wording here a bit, too. Maybe I'm just not catching the drift here

evantypanski (commented):
I like the change; I'm actually unsure how often we use "node" to refer to a host, but I wouldn't be shocked if we conflate the two.

doc/advanced/devel/cluster/architectures.rst
awelzel: Thanks! I wouldn't mind the definition style, but not sure the note section here (that was mostly meant as "don't be discouraged about all these overloaded terms") is the right place. It probably should be somewhere outside of `advanced/devel` then. For me, this is mostly a wiki-style overview. Other notes talk about the current state and possibly future directions, for example.

evantypanski: Yeah, I guess my point is that it doesn't read like that anymore. It did, but with this change it feels whiplash-y. So my suggestion was to combat that by being extremely explicit.

I wouldn't mind something less formal either:

> These processes are also called nodes and their role called node type. Here, we'll stick with the "process" terminology. These terms all get used in different ways based on the system; some may use "node" to describe a complete host instead. Zeek uses the concept of "single-host" and "multi-host" for this, since "node" is just a process!

Regardless it's a writing-style thing so I'll approve. It just felt like the note lost its point when it changed from explaining the conflated "node" term within Zeek

evantypanski (approved):

doc/advanced/devel/cluster/architectures.rst
awelzel: Thanks for that proposal, I picked it up!

Issue #3359, opened 2021-07-05
Allow creating an analyzer which can consume both TCP and UDP traffic


bbannier:
Currently a protocol analyzer needs to be declared to analyze _either_ TCP _or_ UDP traffic, and it is e.g., not possible to declare a Spicy analyzer which can analyze both TCP or UDP.

This restriction can be worked around by declaring two separate analyzers which share the same grammar, but analyze different traffic. The analyzers might even produce identical data and only differ artificially in name (e.g., `OpenVPN_TCP` and `OpenVPN_UDP` in [zeek/spicy-analyzers](http://github.com/zeek/spicy-analyzers)); these names also appear in user-visible Zeek logs.

We should consider extending the EVT glue layer so protocol analyzers can be declared which could analyze both TCP and UDP.

PR #5669, opened 2026-07-09
Added TCP Support for Syslog

### LINKED ISSUE #4766: Support Syslog over TCP
This has come up in a conversation with a user, there's currently no support for Syslog over TCP.

https://datatracker.ietf.org/doc/html/rfc6587

VasuBhakt:
Fixes #4766 

## Problem
There is currently no support for Syslog over TCP.

## Solution
- Created a new dedicated `SyslogTCP` analyzer (`syslog-tcp.spicy`), adhering to `RFC 6587`, supporting `OctetCounting` and `NonTransparent` framing.
    * `OctetCounting` safely re-uses the UDP `Syslog::Message` parser by passing the exact size upfront (`&size`) 
    * `NonTransparent` reads bytes directly from the stream until it hits a `\n` character.
- Registered the new `SyslogTCP` analyzer in `syslog.evt` and added the event hook for `NonTransparent` framing.
- Updated `syslog/main.zeek` to register the `Syslog_TCP` Analyzer using `514/tcp` port.
- Added btest integration tests (`tcp-framing.test` & `tcp-syslog.pcap`) to verify both framing modes, and updated required Baseline files.

## Notes
- The legacy analyzer has not been updated. If it needs improvement, please do let me know :-)
- *(The Antigravity AI coding assistant was used for understanding the codebase and code writing assistance)*


testing/btest/scripts/base/protocols/syslog/tcp-framing.test
```diff
@@ -0,0 +1,4 @@
+# @TEST-EXEC: zeek -C -b -r $TRACES/tcp-syslog.pcap %INPUT
+# @TEST-EXEC: btest-diff syslog.log
```
awelzel: Nit.

```suggestion
# @TEST-EXEC: btest-diff-cut - m syslog.log
```

testing/btest/scripts/base/protocols/syslog/tcp-framing.test
```diff
@@ -0,0 +1,4 @@
...
+
```
awelzel, line 4: Please add a baseline for the conn.log as well:
```
btest-diff-cut -m uid service history duration orig_pkts resp_pkts conn.log
```

This is probably going to show `syslog_tcp` in the `service` column. There's a related ticket here: https://github.com/zeek/zeek/issues/3359

So you could minimally reference that, but no need to fix/change. LDAP has the same trouble, we log `ldap_tcp` or `ldap_udp` rather than just `ldap`, no need to fix this here.

testing/btest/scripts/base/protocols/syslog/tcp-framing.test
```diff
@@ -0,0 +1,4 @@
+# @TEST-EXEC: zeek -C -b -r $TRACES/tcp-syslog.pcap %INPUT
```
awelzel, line 1: This pcap only contains a single TCP packet, not a full TCP connection. Would it be possible for you to run rsyslog or some other syslog implementation and do a full capture of the traffic? Also, multiple messages over a longer duration would be great. That'd be a bit more realistic then feeding just a single TCP packet.

src/analyzer/protocol/syslog/syslog-tcp.spicy
```diff
@@ -0,0 +1,43 @@
...
+module SyslogTCP;
+
+import Syslog;
+import spicy;
+
+public type Messages = unit {
+    messages: Message[] &eod;
+};
+
+type Message = unit {
```
awelzel: RFC 6587 calls this a SyslogFrame. That's the better name also because it avoid confusion with the existing Syslog::Message referenced below.

```
[3.4.1](https://www.rfc-editor.org/info/rfc6587/#section-3.4.1).  Octet Counting
       ...
       TCP-DATA = *SYSLOG-FRAME

       SYSLOG-FRAME = MSG-LEN SP SYSLOG-MSG   ; Octet-counting
                                              ; method
```
```suggestion
type Frame = unit {
```

src/analyzer/protocol/syslog/syslog-tcp.spicy
```diff
@@ -0,0 +1,43 @@
+# See the file "COPYING" in the main distribution directory for copyright.
+
+module SyslogTCP;
+
+import Syslog;
+import spicy;
+
+public type Messages = unit {
```
awelzel:
```suggestion
public type Frames = unit {
```
?

src/analyzer/protocol/syslog/syslog-tcp.spicy
```diff
@@ -0,0 +1,43 @@
+# See the file "COPYING" in the main distribution directory for copyright.
+
+module SyslogTCP;
```
awelzel: IMO it'd be fine to include the units into the existing syslog.spicy Syslog module. With the "frame" naming that should be okay and this isn't too complex to warrant the second file.

src/analyzer/protocol/syslog/syslog-tcp.spicy
```diff
@@ -0,0 +1,43 @@
...
+
+type OctetCounting = unit {
+    len: /[1-9][0-9]*/;
+    : b" ";
+
+    syslog_msg: Syslog::Message &size=self.len.to_uint();
+};
+
+type NonTransparent = unit {
+    bracket: /</;
```
awelzel:
```suggestion
    : b"<";
```

Also below for the others that you don't need to capture the value and there's no need to use a regex for a single character either.

src/analyzer/protocol/syslog/syslog.evt
```diff
@@ -1,11 +1,16 @@
...
 
 %doc-id= Zeek::Syslog;
-%doc-description = "Syslog analyzer UDP-only";
+%doc-description = "Syslog analyzer UDP and TCP";
 
 protocol analyzer Syslog over UDP:
     parse with Syslog::Message;
 
+protocol analyzer Syslog_TCP over TCP:
+    parse with SyslogTCP::Messages;
```
awelzel:
```suggestion
    parse with Syslog::Frames;
```

awelzel (commented):
Hey @VasuBhakt - thanks for another contribution!

I had a number of comments. Primarily it'd be good to add more pcap captures for better test coverage (full TCP connections and also multiple syslog frames in a single connection). If you could add a test for the `NonTransparent` case, that'd also be great.

Hope it's not overwhelming.

testing/btest/scripts/base/protocols/syslog/tcp-framing.test
VasuBhakt: added it.

testing/btest/scripts/base/protocols/syslog/tcp-framing.test
VasuBhakt: added baseline for conn.log as suggested

testing/btest/scripts/base/protocols/syslog/tcp-framing.test
VasuBhakt: I added a PCAP that contains a full TCP handshake/teardown. The PCAP sends 4 syslog messages over the single connection, interleaving both OctetCounting and NonTransparent frames.
Let me know what you think!

src/analyzer/protocol/syslog/syslog-tcp.spicy
VasuBhakt: Renamed the TCP Message into Frame as per suggestion.

src/analyzer/protocol/syslog/syslog-tcp.spicy
VasuBhakt: Renamed to Frames

src/analyzer/protocol/syslog/syslog-tcp.spicy
VasuBhakt: Brought the TCP Syslog units into syslog.spicy Syslog module, and made necessary changes.

src/analyzer/protocol/syslog/syslog-tcp.spicy
VasuBhakt: Added this as per suggestion.

src/analyzer/protocol/syslog/syslog.evt
VasuBhakt: Applied this change.

VasuBhakt:
Hey @awelzel , I added a PCAP that contains a full TCP handshake/teardown. The PCAP sends 4 syslog messages over the single connection, interleaving both OctetCounting and NonTransparent frames.
Let me know what you think!

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -30,3 +30,41 @@ type Priority = unit {
...
+type OctetCounting = unit {
+    len: /[1-9][0-9]*/;
+    : b" ";
+
+    syslog_msg: Message &size=self.len.to_uint();
+};
+
+type NonTransparent = unit {
+    : b"<";
+    prio: /[0-9]+/ &convert=$$.to_uint();
```
awelzel: Can re-use the Priority unit from above (remove the bracket above and below).

```suggestion
    prio: Priority;
```

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -30,3 +30,41 @@ type Priority = unit {
...
+    : b">";
+    msg: /[^\n]+/;
```
awelzel:
```suggestion
    msg: /[^\r\n\x00]+/;
```

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -30,3 +30,41 @@ type Priority = unit {
...
+    : b"\n";
```
awelzel:
```suggestion
    : /[^\r\n\x00]+/;
```

Mostly because the RFC 6587 says this has all been observed in the wild.

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -30,3 +30,41 @@ type Priority = unit {
     value: /[[:digit:]]+/ &convert=$$.to_uint();
     : b">";
```
awelzel, line 31: Think we can extract the severity and facility within the Priority unit (and as sketched below re-use it in the non-transparent case). That way this mask/shift logic is only in a single place.


Here's a hunk I have locally for inspiration - it's not perfect, but nicer:  [diff.txt](https://github.com/user-attachments/files/30129303/diff.txt)

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -30,3 +30,41 @@ type Priority = unit {
...
+    prio: /[0-9]+/ &convert=$$.to_uint();
+    : b">";
+    msg: /[^\n]+/;
+    : b"\n";
+
+    var facility: uint64 = 999;
+    var severity: uint64 = 999;
+
+    on %done {
+        self.severity = self.prio & 0x07;
```
awelzel: This is missing an `spicy::accept_input()` call here. A connection with only NonTransparent framing would currently have no `syslog_tcp` entry, correct?

Up for adding a "only non-transparent" pcap and separate test to validate and then add `spicy::accept_input()` here and check that the new test is fixed?

awelzel (commented):
Hey @VasuBhakt - this is getting there! Mostly small stuff and re-using the Priority unit. Are you up for one more round and an additional test? I think this can land afterwards.

src/analyzer/protocol/syslog/syslog.spicy
VasuBhakt: Applied your provided diff. Works correctly!

src/analyzer/protocol/syslog/syslog.spicy
VasuBhakt: Added the extra characters.

src/analyzer/protocol/syslog/syslog.spicy
VasuBhakt: Priority unit has been edited as per your provided diff and been re-used here.

src/analyzer/protocol/syslog/syslog.spicy
VasuBhakt: Done.

src/analyzer/protocol/syslog/syslog.spicy
VasuBhakt: Nice catch! Added `accept_input()` call and also added a only non-transparent pcap and test to validate the change.  Correctly shows `syslog_tcp` entry with that change.

VasuBhakt:
Hi @awelzel, I added your recommended changes, it works correctly! Let me know what you think!

src/analyzer/protocol/syslog/syslog.spicy
```diff
@@ -29,4 +29,52 @@ type Priority = unit {
...
+        self.facility = (self.value & 0x03f8) >> 3;
+    }
+
+    var severity: uint64 = 0;
+    var facility: uint64 = 0;
+};
+
+# Use Frames for TCP - RFC 6587
+public type Frames = unit {
+    frames: Frame[] &eod;
```
awelzel, line 44:
```suggestion
    : Frame[] &eod;
```

Otherwise this is unbounded state growth. Wonder a bit if that could be statically rejected.

Will do during the merge.

awelzel (approved):
Thanks again! Looks great.

VasuBhakt:
Thanks again for being such a great mentor and helping me navigate through all of this, @awelzel! 
Great catch on removing the variable name for the array; I definitely didn't think about the unbounded state growth / memory leak on long-lived connections. Thanks for explaning this and fixing it up during the merge!

PR #5671, opened 2026-07-09
Upgrade c-ares dependecies to v1.34.8

timwoj:
C-Ares recently released v1.34.7 that fixed a few security issues, and then a quick follow-up v1.34.8 to fix a regression. This PR upgrades our c-ares submodule to the latest release, and also upgrades vcpkg to a commit that upgrades the c-ares port. In the process we also catch upgrades in vcpkg for libpcap, openssl, and zlib.


timwoj:
The two CI failures here are unrelated to the change on the PR.

awelzel (approved):

Issue #2283, opened 2022-07-18
Add tests that exercise all of the error cases in the expression handling


timwoj:
While reviewing https://github.com/zeek/zeek/pull/2277, I realized that we have basically zero tests that exercise the error cases in the `Expr` code. Most of the cases for things like passing strings to numeric operators, etc aren't covered. You can see this in https://coveralls.io/builds/48294521/source?filename=src%2FExpr.cc, where most of those lines are still red.

eagle-head:
Hi! I'd like to pick this up.

My understanding: `src/Expr.cc`'s error paths are almost entirely unexercised — both the compile-time type-checking errors (passing a string to a numeric operator, bad table/set/vector/record constructors, bad coercions, illegal field/index access, invalid call/event/hook usage, type clashes, …) and the runtime errors (division/modulo by zero, out-of-range and boolean-index mismatches, list-length mismatches, …). Most of those lines are red in coverage.

Plan, following the guidance above: build with `--enable-coverage`, use `make coverage` to pinpoint the still-red error branches in `Expr.cc`, and add btests under `testing/btest/language/` that trigger each one, matching the existing error-test conventions (`@TEST-EXEC-FAIL` + `diff-remove-abspath` for the type-checking errors, and the run-and-diff pattern for the runtime ones).

I'll go for full coverage of the script-reachable error cases in one PR. The only lines I'll leave are the purely-internal invariants that can't be reached from a script — the defensive `reporter->InternalError(...)` / `RuntimeErrorWithCallStack("bad type in ...Fold")` checks that only fire on a compiler/type-system bug. I'll call those out explicitly in the PR. Thanks!

PR #5673, opened 2026-07-10
Add btests exercising the error cases in expression handling

eagle-head:
Add btests for the script-reachable expression error paths from GH-2283 and fix three defects exposed by those tests.

(This work was AI-assisted using OpenAI Codex gpt-5.6 sol xhigh for test generation, coverage analysis, defect investigation, validation, and drafting; all changes were directed and reviewed by me.)


timwoj:
Per our AI policy, was an LLM used in the creation on this PR?

eagle-head:
> Per our AI policy, was an LLM used in the creation on this PR?

Yes. I used OpenAI Codex gpt-5.6 sol xhigh for test generation, coverage analysis, defect investigation, validation, and drafting. I directed the work and reviewed the output before pushing.

eagle-head:
@timwoj — heads up, since you'd looked at this one earlier.

The scope has grown: three of the error cases this PR set out to cover turned out to be actual bugs rather than testable behaviour, so it now carries fixes for them alongside the tests, as separate commits. I've rewritten the description to explain what each one is and why they're here rather than in a PR of their own.

Whenever you have a moment, I'd appreciate a look at the updated description. Happy to split the fixes out if you'd rather review them separately.

src/CompHash.cc
```diff
@@ -87,6 +87,10 @@ std::unique_ptr<HashKey> CompositeHash::MakeHashKey(const Val& argv, bool type_c
     if ( type_check && argv.GetType()->Tag() != TYPE_LIST )
         return nullptr;
 
+    // Otherwise the loops below read past the end of the key list.
+    if ( type_check && argv.AsListVal()->Length() != static_cast<int>(tl.size()) )
+        return nullptr;
```
evantypanski, line 92: the commit message says this is reachable but doesn't add a test. I was only able to reach this by commenting out the next fix when creating the expression in `Expr.cc`.

I don't mind this as a defensive check here, since it seems real, but what are the examples, if you can bring them up? I just don't know how this would hit and seems like if there's an example we should know about it

evantypanski (commented):
This looks nice to me, thanks! I have one question though.

I'm really hesitant to change the interpreter because we don't have great test coverage, this will be helpful for me :)

(also btw: it's perfectly reasonable to put the fixes in the PR that discovers them as you did, thanks)

src/CompHash.cc
eagle-head: Good question. A concrete path is Broker deserialization: two peers can declare the same &backend=Broker::MEMORY table with different composite key arities. The two-part key received by the three-part declaration reaches TableVal::Assign() and then CompositeHash::MakeHashKey() without going through TableConstructorExpr.

I added broker.store.brokerstore-backend-composite-key-arity to cover this path. It checks the regular type-mismatch error, verifies that the receiver table remains empty, and passes under ASan.

evantypanski:
hey, I'll get back to this when I can, it's been a bit busy here :) I haven't forgotten about this.

Someone else may also beat me to it. At a glance I like the tests, but I need to check some of it in more depth

src/CompHash.cc
evantypanski: stepped in with a debugger and it hits this return so I'm content, thanks!

evantypanski (approved):
Yeah this looks great, thanks! Sorry it took a while to get to, I had other high priority stuff.

I'll test locally before too when merging in case something drifted in this time.

PR #5677, opened 2026-07-10
Use templated version of std::min vs casting arguments

timwoj:
This is a follow-up to https://github.com/zeek/zeek/pull/5659, changing a bunch of places where we call `static_cast` for an argument of `std::min` to instead pass the requested return type as a template argument to `std::min`.

### DIFF
--- a/src/analyzer/protocol/rpc/RPC.cc
+++ b/src/analyzer/protocol/rpc/RPC.cc
@@ -351,7 +351,7 @@ bool RPC_Reasm_Buffer::ConsumeChunk(const u_char*& data, int& len) {
     // How many bytes do we want to process with this call?  Either the
     // all of the bytes available or the number of bytes that we are
     // still missing.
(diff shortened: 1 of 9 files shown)


awelzel (approved):

src/analyzer/protocol/rpc/RPC.cc
```diff
@@ -351,7 +351,7 @@ bool RPC_Reasm_Buffer::ConsumeChunk(const u_char*& data, int& len) {
     // How many bytes do we want to process with this call?  Either the
     // all of the bytes available or the number of bytes that we are
     // still missing.
-    int64_t to_process = std::min(static_cast<int64_t>(len), (expected - processed));
+    int64_t to_process = std::min<int64_t>(len, (expected - processed));
```
awelzel: Those that define a variable could use `auto` now I guess? Would avoid the repeated type specifier.

PR #5681, opened 2026-07-13
CMake: Avoid creation of empty directories in zeek/include

### LINKED ISSUE #5678: 8.0.9 creates a lot of empty include directories
I received a complaint about the FreeBSD zeek port. One issue is dozens of empty directories created in the include tree. Is this something we can clean up?
```
zinc 99 @ pwd
/wrkdirs/usr/ports/security/zeek/work/stage/usr/local
zinc 100 @ find include/zeek -type d -empty
include/zeek/analyzer/protocol/redis
include/zeek/analyzer/protocol/ssl/spicy
include/zeek/analyzer/protocol/ldap
include/zeek/analyzer/protocol/quic
include/zeek/analyzer/protocol/postgresql
include/zeek/cluster/backend/zeromq/auxil/cppzmq/tests
include/zeek/cluster/backend/zeromq/auxil/cppzmq/examples
include/zeek/cluster/backend/zeromq/auxil/cppzmq/demo
include/zeek/cluster/backend/zeromq/auxil/cppzmq/cmake
include/zeek/cluster/backend/zeromq/auxil/cppzmq/libzmq-pkg-config
include/zeek/cluster/backend/zeromq/cmake
include/zeek/cluster/websocket/auxil/IXWebSocket/docs
include/zeek/cluster/websocket/auxil/IXWebSocket/ws
include/zeek/cluster/websocket/auxil/IXWebSocket/docker
include/zeek/cluster/websocket/auxil/IXWebSocket/third_party/cli11
include/zeek/cluster/websocket/auxil/IXWebSocket/third_party/cpp-linenoise/example
include/zeek/cluster/websocket/auxil/IXWebSocket/third_party/msgpack11
include/zeek/cluster/websocket/auxil/IXWebSocket/tools
include/zeek/cluster/websocket/auxil/IXWebSocket/CMake
include/zeek/cluster/websocket/auxil/IXWebSocket/test/.certs
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/node
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/csharp
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/python/websocket-client
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/python/websockets/vendor
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/ruby
include/zeek/cluster/websocket/auxil/IXWebSocket/test/compatibility/cpp/libwebsockets
include/zeek/cluster/websocket/auxil/IXWebSocket/test/Catch2/single_include
include/zeek/cluster/websocket/auxil/IXWebSocket/test/data
include/zeek/storage/backend/redis/cmake
```

awelzel:
The prior `install(...FILES_MATCHING...)` approach would reproduce the directory structure, resulting in empty directories by using GLOB_RECURSE and installing files explicitly after filtering.

This PR also adds a coverage test to baseline Zeek's include directory in an installation tree to avoid unknowingly adding files here. Seems a baseline is perfect for this.

Closes #5678

---

We could just baseline all of the install dir, but with all the configuration knobs available (enable zkg, btest, zeek-client, ...) there'll be a lot of variations, so headers seemed like a good starting point and has some relation to the script baselines.


testing/btest/Baseline/coverage.include-directory/..include-dir.out
```diff
@@ -0,0 +1,1503 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
```
timwoj, line 1: Can this instead look for empty directories in the install tree (with maybe some whitelisting if needed) instead of enumerating all of the headers? This feels brittle like that one ZAM test was.

timwoj (changes requested):

testing/btest/Baseline/coverage.include-directory/..include-dir.out
awelzel: I'm comparing this more with the `bare-load-baseline.test`, `default-load-baseline.test` and `test-all-policy.test` test which all to a certain degree are brittle (they change whenever we add new script files or re-arrange load ordering), but still think they are useful as a signal.

And similar with include files. They change, but I don't think they change enough to be annoying and I like the fact of tracking the installed files as a baseline and the signal it gives. Seems adequate. I was actually tempted to baseline the whole install tree, but that felt too brittle due to the combination of --enable / --disable flags that impact that.

We had mistakenly included headers from within `auxil/`  in the `websocket`, `cluster` and `storage` parts (e051cfb7d27a172b0e9bbb68c8593e988ed313bc) that never should've been exposed. This test would've covered them.

If Broker (unlikely) or Spicy change headers a lot, we could filter out those, but for Zeek I'm pretty much thinking "try first and revert if it's too noisy".

timwoj: Ok, that works for me. Thanks for the clarifications.

timwoj (approved):

PR #5685, opened 2026-07-14
Fix ccache use in CircleCI setup

bbannier:


timwoj (approved):

.circleci/builds.yml
awelzel: FWIW, the  `~/.ccache` path here results in cache files being placed into the the current working directories of the compiler invocations, because `~` never expanded.

That was uncovered by the test introduced in #5681. You may want to rebase to see if it gets speedier. The output of the `save_cache` step had also indicated an empty cache uploaded.

PR #5688, opened 2026-07-15
add support for trailing pattern flags in Input Framework

ynadji:
zeek's `pattern` type supports `/<pattern>/si` literals in zeek script (to enable case insensitivity and treating input as a single line) but this isn't supported in the input framework. presumably because we don't build the regex at ingest. however, we can get around this by just wrapping the pattern in `(?s:<pattern>)` and/or `(?i:<pattern>)` to support this.

this PR adds that functionality and a btest to ensure it works. used Claude for some cleanup.


testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
```diff
@@ -0,0 +1,62 @@
+# @TEST-EXEC: btest-bg-run zeek zeek -b %INPUT
```
awelzel:
```suggestion
# @TEST-EXEC: zeek -b %INPUT
```

testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
```diff
@@ -0,0 +1,62 @@
...
+# @TEST-EXEC: btest-bg-wait 10
```
awelzel:
```suggestion
```

testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
```diff
@@ -0,0 +1,62 @@
...
+
+type Val: record {
+	v: string;
+};
+
+global pats: table[pattern] of Val = table();
+
+event zeek_init()
+	{
+	Input::add_table([$source="../input.log", $name="pats", $idx=Idx, $val=Val, $destination=pats]);
```
awelzel:
```suggestion
	Input::add_table([$source="input.log", $name="pats", $idx=Idx, $val=Val, $destination=pats]);
```

testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
awelzel: You don't need btest-bg-run for this. Your test might not terminate automatically, but then press ctrl+c

testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
```diff
@@ -0,0 +1,62 @@
+# @TEST-EXEC: btest-bg-run zeek zeek -b %INPUT
+# @TEST-EXEC: btest-bg-wait 10
+# @TEST-EXEC: sed 1d .stderr | grep -v "queued" > .stderrwithoutfirstline
```
awelzel: I tried `btest-diff .stderr` and it seemed fine without the `btest-bg-run`. Are you sure you need this all?

src/threading/formatters/Ascii.cc
```diff
@@ -277,16 +282,36 @@ Value* Ascii::ParseValue(const string& s, const string& name, TypeTag type, Type
...
+                        }
+                    }
+
+                    if ( valid_flags ) {
+                        string pattern = candidate.substr(1, closing - 1);
+
+                        if ( flag_i )
+                            pattern = "(?i:" + pattern + ")";
+                        if ( flag_s )
+                            pattern = "(?s:" + pattern + ")";
```
awelzel: Puh, I don't think we should do this down here in formatters/Ascii.cc. The `(?:)`  I believe is an implementation detail of RE. It would also not set the `is_case_insensitive` and `is_single_line` members on the RE instance (created in input/Manager.cc).

I think I'd like the following better, transporting the flags as explicit bools out, then handling in the input/Manager via the proper accessors as below.

Then you can also reuse the flags in the threading instead of the temporary bools.

```
diff --git a/src/input/Manager.cc b/src/input/Manager.cc
index 205d13b703..058798c171 100644
--- a/src/input/Manager.cc
+++ b/src/input/Manager.cc
@@ -2076,7 +2076,11 @@ Val* Manager::ValueToVal(const Stream* i, const Value* val, Type* request_type,
         }
 
         case TYPE_PATTERN: {
-            auto* re = new RE_Matcher(val->val.pattern_text_val);
+            auto* re = new RE_Matcher(val->val.pattern_val.text);
+            if ( val->val.pattern_val.is_case_insensitive )
+                re->MakeCaseInsensitive();
+            if ( val->val.pattern_val.is_single_line )
+                re->MakeSingleLine();
             re->Compile();
             return new PatternVal(re);
         }
diff --git a/src/threading/SerialTypes.cc b/src/threading/SerialTypes.cc
index 598ee11a47..8530c42124 100644
--- a/src/threading/SerialTypes.cc
+++ b/src/threading/SerialTypes.cc
@@ -526,7 +526,11 @@ Val* Value::ValueToVal(const std::string& source, const Value* val, bool& have_e
         }
 
         case TYPE_PATTERN: {
-            auto* re = new RE_Matcher(val->val.pattern_text_val);
+            auto* re = new RE_Matcher(val->val.pattern_val.text);
+            if ( val->val.pattern_val.is_case_insensitive )
+                re->MakeCaseInsensitive();
+            if ( val->val.pattern_val.is_single_line )
+                re->MakeSingleLine();
             re->Compile();
             return new PatternVal(re);
         }
diff --git a/src/threading/SerialTypes.h b/src/threading/SerialTypes.h
index e5d8dd4ded..71d10585c0 100644
--- a/src/threading/SerialTypes.h
+++ b/src/threading/SerialTypes.h
@@ -182,14 +182,25 @@ struct Value {
         vec_t vector_val;
         addr_t addr_val;
         subnet_t subnet_val;
-        const char* pattern_text_val;
+        struct {
+            // anonymous struct for backwards compat.
+            // pattern_text_val == patern_val.text
+            [[deprecated("Remove with v9.1: Use pattern_val.text instead")]]
+            const char* pattern_text_val;
+        };
+        struct {
+            const char* text;
+            bool is_case_insensitive;
+            bool is_single_line;
+            // 6 bytes hole here.
+        } pattern_val;
 
         struct {
             char* data;
             int length;
         } string_val;
 
-        _val() { memset(this, 0, sizeof(_val)); }
+        _val() { memset(static_cast<void*>(this), 0, sizeof(_val)); }
     } val;
 
     /**
```

Haven't really tested, but I hope it shows the idea. Will need some updates to remove deprecation warnings, though we could also opt to skip the deprecation.

`pattern_text_val` and `pattern_val.text` should both work, so we don't break backwards compat, and you can funnel the flags in the pattern_val.

awelzel (changes requested):
Seems like something that's good to fix/support. See if the comments around the implementation make sense to you. Thanks.

src/threading/formatters/Ascii.cc
```diff
@@ -250,6 +250,11 @@ Value* Ascii::ParseValue(const string& s, const string& name, TypeTag type, Type
 
         case TYPE_SUBNET: {
             string unescaped = util::strstrip(util::get_unescaped_string(s));
+            // A string is a candidate pattern iff it is surrounded by '/'
+            // characters, with an optional suffix of the 'i' and/or 's' flags
+            // to denote case insensitive or single line matching, respectively.
+            // Whether the pattern is legal will be determined later by the RE
+            // engine.
```
awelzel: Any reason this ended up under subnet, btw? :-)

ynadji: just me being sloppy here. sorry about that

src/threading/formatters/Ascii.cc
ynadji: sounds good! I'll make these changes

src/threading/SerialTypes.h
awelzel:
```suggestion
        _val() { memset(static_cast<void*>(this), 0, sizeof(_val)); }
```


To silence the following. There's something about the anonymous struct making some compilers complain (not all if you look the CI results).

```
../../../../src/include/zeek/threading/SerialTypes.h: In constructor 'zeek::threading::Value::_val::_val()':
../../../../src/include/zeek/threading/SerialTypes.h:199:24: error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'union zeek::threading::Value::_val'; use assignment or value-initialization instead [-Werror=class-memaccess]
  199 |         _val() { memset(this, 0, sizeof(_val)); }
      |                  ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
../../../../src/include/zeek/threading/SerialTypes.h:176:11: note: 'union zeek::threading::Value::_val' declared here
  176 |     union _val {
      |           ^~~~
```

src/threading/formatters/Ascii.cc
```diff
@@ -277,16 +277,37 @@ Value* Ascii::ParseValue(const string& s, const string& name, TypeTag type, Type
...
-                    break;
+            // A string is a candidate pattern iff it is surrounded by '/'
+            // characters, with an optional suffix of the 'i' and/or 's' flags
+            // to denote case insensitive or single line matching, respectively.
+            // Whether the pattern is legal will be determined later by the RE
+            // engine.
+            if ( candidate.size() >= 2 && candidate.front() == '/' ) {
+                auto closing = candidate.rfind('/');
+                if ( closing > 0 ) {
+                    string flags = candidate.substr(closing + 1);
```
awelzel: Assuming happy-path, the substr is at most 2 characters, so shouldn't hit a malloc due to short string optimization.

src/threading/formatters/Ascii.cc
```diff
@@ -277,16 +277,37 @@ Value* Ascii::ParseValue(const string& s, const string& name, TypeTag type, Type
...
+                        else if ( c == 's' && ! flag_s )
+                            flag_s = true;
+                        else {
+                            valid_flags = false;
+                            break;
+                        }
+                    }
+
+                    if ( valid_flags ) {
+                        string pattern = candidate.substr(1, closing - 1);
```
awelzel: Can avoid one intermediary string allocation here:
```
-                        string pattern = candidate.substr(1, closing - 1);
-                        val->val.pattern_val.text = util::copy_string(pattern.c_str(), pattern.size());
+                        auto psv = std::string_view(candidate).substr(1, closing - 1);
+                        val->val.pattern_val.text = util::copy_string(psv.data(), psv.size());
```

For the `string flags` above, would think it's not worth the string_view indirection, because in the happy-path that's anyhow just 2 characters so SSO an no malloc.

awelzel (commented):
Two nits, but otherwise looks good. Thanks!

src/threading/formatters/Ascii.cc
awelzel: Maybe this to ensure there's at most 2 flag characters, to skip over `/pattern/abcdefg` immediately, rather than processing flags.
```
                if ( closing > 0 && (candidate.size() - 1 - closing) <= 2 ) {

```

awelzel (dismissed):

src/threading/SerialTypes.cc
```diff
@@ -103,7 +103,7 @@ Value::Value(const Value& other) {
         }
 
         case TYPE_PATTERN: {
-            val.pattern_text_val = util::copy_string(val.pattern_text_val);
+            val.pattern_val.text = util::copy_string(val.pattern_val.text);
```
awelzel:
```suggestion
            val.pattern_val.text = util::copy_string(other.val.pattern_val.text);
```

...also needs to copy the new modifiers/flags.

testing/btest/scripts/base/frameworks/input/pattern-flags.zeek
```diff
@@ -0,0 +1,60 @@
...
+# @TEST-START-FILE input.log
+#separator \x09
+#fields	p	v
+#types	pattern	string
+/dog/i	case-insensitive
+/c.t/s	single-line
+/foo.|bar/is	both-flags
+/plain/	no-flags
+/dup/isis	invalid-dup-flags
+/other/isg	invalid-bad-flag
```
awelzel, line 15: I had approved, but after looking through this again, the `CopyValue()` and `GetValueLength()` needs updating to include the modifiers.

Could you create a separate test with
```
/dog/ case-sensitive
/dog/i case-insensitive
/dog/i single-line
/dog/is both
/dog/si  both-reversed
```

That should result in 4 entries in the table and don't think that works just yet.

ynadji: I just added it to the same test, but made/confirmed the changes. lmk what you think.

awelzel (approved):

PR #5689, opened 2026-07-15
tools/systemd-generator: Multi-interface and multi-host support

awelzel:
This is a single commit, essentially extending the zeek-systemd-generator
in one shot to:

Add support for parsing ``<PREFIX>/etc/zeek/zeek.conf`` in an INI-like format to
support multiple network interfaces that ends up looking like so:

    [zeek]
    loggers = 3
    proxies = 7

    [interface eth0]
    interface = af_packet::eth0
    workers = 4
    worker_args = AF_Packet::fanout_id=42

    [interface napatech]
    interface = napatech::${worker_index0}
    workers = 64
    workers_cpu_list = 4-67
    worker_memory_max = 4G

Second, static multi-host clusters are supported by placing zeek.conf files of
every host on every host into the directory ``<PREFIX>/etc/zeek/cluster/<hostname>.zeek.conf``.
More details in the follow-up doc commit.

The ``zeek-cluster-layout-generator`` tool was adapted to cover usages.

The integrated unittest in the generator have been split into a separate
doctest-based executable.


doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+Zeek includes a `systemd generator <https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html>`_
+that reads the ``<PREFIX>/etc/zeek/zeek.conf`` configuration file and instantiates systemd
+unit files for the manager, logger, proxy and worker processes accordingly.
+The generator is installed at ``<PREFIX>/bin/zeek-systemd-generator``.
+
+.. note::
+
+   ``zeek.conf`` file's focus is an opinionated Zeek deployment tailored for
+   appliance or container use cases. It presents an alternative to a
+   :ref:`ZeekControl <cluster-configuration>` managed deployment.
```
ckreibich: Could we not use "opinionated"? I think it has mostly bad connotations — "it's quirky", "it's partial", "I'm not willing to debate your takes", etc. That might just be me, but more importantly I think that's not actually true here! We did spend a bunch of time thinking through pros/cons etc.

An alternative would be to say that this approach prioritizes system-level cluster orchestration, not Zeek-side configuration — which ZeekControl continues to offer.

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+
+============================
+Deploying Zeek using systemd
+============================
+
+.. versionadded:: 8.1
+
+Zeek includes a `systemd generator <https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html>`_
+that reads the ``<PREFIX>/etc/zeek/zeek.conf`` configuration file and instantiates systemd
+unit files for the manager, logger, proxy and worker processes accordingly.
```
ckreibich, line 12: Maybe a tiny bit more context for people just passing through?
```suggestion
unit files for the manager, logger, proxy and worker processes accordingly.
systemd automatically runs generators during system startup, or on demand (see below).
```

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+
+   # ln -s <PREFIX>/bin/zeek-systemd-generator /etc/systemd/system-generators/
+
+
+Single Host Walkthrough
+=======================
+
+For a small Zeek deployment listening with 2 workers on eth1, using
+AF_PACKET for flow-balancing, plus running the manager and one logger
+and one proxy process, the following minimal configuration is sufficient:
```
ckreibich: Just simplifying a little:
```suggestion
For a small Zeek deployment running two AF_PACKET worker processes on eth1, 
one logger, one proxy, and the manager, the following configuration is sufficient:
```

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+    interface = af_packet::eth1
+    workers = 2
+
+This minimal format is termed the section-less configuration style.
+See the :ref:`systemd_multiple_interfaces` section for the more verbose INI-style
+configuration that is required when you want to listen on multiple interfaces.
+
+After running ``systemctl daemon-reload``, which executes ``zeek-systemd-generator``,
+the generated unit files are found in the ``/run/systemd/generator`` directory of
+the system where they can be inspected:
```
ckreibich:
```suggestion
the system:
```
(less passive)

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+
+Installation
+============
+
+After installing Zeek on your system, link Zeek's systemd generator into one
+of the standard generator directories:
+
+.. code:: console
+
+   # ln -s <PREFIX>/bin/zeek-systemd-generator /etc/systemd/system-generators/
```
ckreibich, line 32: Fyi, I hit some hiccups when I first tried this:
* I had to create the generator directory — none of the ones shown on the freedesktop.org page existed. That gave me some pause (but worked fine).
* I had a user-local install of the Zeek tree: the symlink was from `~/inst/opt/zeek/bin/zeek-systemd-generator`. SELinux balks at this, at two levels: (1) running the generator, (2) loading the config. The journal has the usual instructions to fix this, but it's subtle because without looking in the journal there's no output — you just don't get unit files. These hurdles go away with a system-level install of Zeek in `/usr/local`. (I have not tried `/opt`.)

So we should probably add something here to warn about this kind of stuff.

(This was on Fedora, btw.)

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+    Apr 24 13:46:45 tinkyx1 zeek-worker-1[2603512]: listening on eth1
+
+This output shows zeek-worker@1 terminating due to a SIGABRT (here, ``kill -SIGABRT``
+was used for demonstration purposes) and systemd automatically restarting the process
+and reporting the current restart counter value. Use ``systemctl show zeek-worker@1``
+to inspect this counter and various other details.
+
+For debugging of crashes and coredump handling in general, we recommend installing
+`systemd-coredump <https://www.freedesktop.org/software/systemd/man/latest/systemd-coredump.html>`_
+and setting up ``/etc/systemd/coredump.conf`` as needed.
```
ckreibich:
```suggestion
if your distro doesn't automatically do so, and setting up ``/etc/systemd/coredump.conf`` as needed.
```
At least on Fedora `coredumpctl` ships automatically.

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+Multiple Interfaces
+===================
+
+.. versionadded:: 9.0
+
+To monitor multiple interfaces, the ``zeek.conf`` file supports INI-style sections
+where there is one section per interface. Each interface section name includes
+a **tag** that's used in systemd's unit files, working directories, and also the
+cluster node name. The tags used in the sample configuration below are ``eth1``
+and ``eth2``.
```
ckreibich: Do you think the notion of a tag matters to the user? If we just just assume that the user will usually want this to be the interface name, that could then become something like
```suggestion
To monitor multiple interfaces, add an INI-style section ``[interface <name>]``
to ``zeek.conf`` for each interface. The generator uses the interface's name
in systemd's unit files, working directories, and also the cluster node name.
```
When would it _not_ be an interface?

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+
+When using the INI-style format, non-interface option must be placed into
+the ``[zeek]`` section, conventionally put at the top.
```
ckreibich:
```suggestion
When using the INI-style format, place non-interface options into
the ``[zeek]`` section, conventionally at the top of the file:
```

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+      LD_PRELOAD=/usr/local/lib/libjemalloc.so
+      MALLOC_CONF=prof:true,prof_prefix:jeprof.out,prof_final:true,lg_prof_interval:26
+
+With AF_PACKET, the respective workers have to use different fanout groups, so we need
+to pass an explicit ``AF_Packet::fanout_id`` setting via the ``worker_args`` option.
+This configuration also pins workers sequentially onto CPUs 4 through 11
+and enables jemalloc profiling for all workers listening on eth2 via the
+``worker_env`` option.
+
+You will find that the worker unit names include the interface tags when
```
ckreibich:
```suggestion
You will find that the worker unit names include the interface names when
```
(if we end up not using the notion of tags)

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+the ``[zeek]`` section, conventionally put at the top.
+
+.. code:: ini
+
+    [zeek]
+    loggers = 3
+    proxies = 7
+
+    [interface eth1]
+    interface = af_packet::eth1
```
ckreibich, line 233: I wondered about defaults at this point. This pattern ...
```
    [interface eth1]
    interface = af_packet::eth1
```
will be very common, right? Should that value (`af_packet::<name>`) be the default?

Actually, are there any defaults? (I later got to the code and don't think so, but I'm not through yet.)

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+   WorkingDirectory=
+   ExecStart=mkdir -p /usr/local/zeek/var/spool/zeek/generated-scripts
+   ExecStart=/usr/local/zeek/bin/zeek-cluster-layout-generator -C /usr/local/zeek/etc/zeek/cluster -o /usr/local/zeek/var/spool/zeek/generated-scripts/cluster-layout.zeek
+
+
+As of version 9.0, each host needs to be statically aware of all other hosts in
+a cluster and able to resolve their IP addresses though the hostname as used in
+the configuration filenames. Adding the required entries into ``/etc/hosts`` on
+the respective systems is one way to achieve this, DNS would be another. Container
+orchestrators usually allow to specify hostnames for containers or pods and
```
ckreibich:
```suggestion
orchestrators usually allow specifying hostnames for containers or pods, and
```

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+file does not change. However, you place the configuration files of all hosts of the
+cluster at ``<PREFIX>/etc/zeek/cluster/<hostname>.zeek.conf`` on each host.
+That means that every host has visibility into every other host's config. Use some
+configuration management tool to keep the ``<PREFIX>/etc/zeek/cluster`` directories
+in sync.
+
+When no ``<PREFIX>/etc/zeek/zeek.conf`` file for a single-host configuration exists,
+the ``zeek-systemd-generator`` will attempt to open ``<PREFIX>/etc/zeek/cluster/<hostname>.zeek.conf``
+using the hostname of the system it is running on. If the file is found, it'll use
+it to configure the appropriate systemd services using that specific config.
```
ckreibich: At this point I'd stop and just shorten to:
```suggestion
it to configure the appropriate systemd services using that specific config
and generate ``cluster-layout.zeek``.
```
If you'd rather keep more detail, I'd move it to a dedicated section about `cluster-layout.zeek`.

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+
+.. note::
+
+   Multi-host support using the zeek.conf is still being refined. We plan
+   to finalize this towards Zeek 10.0. We welcome testing and feedback!
+
+You can think of a host as an actual physical or virtual system, but it could
+also be just a Linux container with its own network namespace. Any environment
+with a dedicated networking stack and hostname can be considered a host in the
+following paragraphs.
```
ckreibich: This section gets pretty complex, so I'd suggest shortening it. I'd mention two things up-front: (1) we assume that you have a compatible Zeek installation on each host, (2) each host has functional name resolution for each host in the cluster (basically what you're saying in line 297-302).

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+.. code:: console
+
+    # zeek -b <PREFIX>/var/spool/zeek/generated-scripts/cluster-layout.zeek -e 'print Cluster::nodes'
+    {
+    [c-w-02-worker-vxlan-4] = [node_type=Cluster::WORKER, ip=fd00:dead:beef::12, zone_id=, p=0/unknown, manager=manager, id=<uninitialized>, metrics_port=9994/tcp],
+    [c-w-01-worker-vxlan-3] = [node_type=Cluster::WORKER, ip=fd00:dead:beef::11, zone_id=, p=0/unknown, manager=manager, id=<uninitialized>, metrics_port=9993/tcp],
+    ...
+    }
+
+For a minimal three host cluster, where host ``c-mgr`` runs manager, 3 loggers,
```
ckreibich, line 372: And I'd resume here (from line 275), with this example.

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+   cluster_backend_args = misc/zeromq-multi-host-auto-setup
+
+.. code:: ini
+
+   # <PREFIX>/etc/zeek/cluster/c-w-01.zeek.conf
+   [zeek]
+   loggers  = 0
+   proxies  = 0
+   manager  = 0
+   archiver = 0
```
ckreibich: This again had me wondering about defaults — would be nice to not need to say this?

ckreibich (commented):
Very cool, but it'll take me a while to work my way through this. :-) Here are comments on the docs for now!

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+for each interface is ``zeek-worker-eth1@1`` and ``zeek-worker-eth2@2``.
+The ``CLUSTER_NODE`` and :zeek:see:`Cluster::node` values change accordingly:
+``worker-1`` is ``worker-eth1-1`` and ``worker-eth2-1``. You can see these
+values in the generated ``cluster-layout.zeek`` file.
+
+
+Interface Templating
+====================
+
+The ``interface`` option supports templating. For example, selecting a
```
ckreibich: This sounds a bit like there are variables other than `worker_index0`. Are there?

doc/advanced/deployment/systemd.rst
```diff
@@ -0,0 +1,452 @@
...
+.. code:: console
+
+    # zeek-cluster-layout-generator -C ./cluster/
+    # Auto-generated by zeek-cluster-layout-generator
+    ...
+
+    const hosts: table[string] of Host &ordered &redef;
+
+    redef Cluster::hosts += {
+            ["c-mgr"] = [$ip=(blocking_lookup_hostname("c-mgr") as vector of addr)[0]],
```
ckreibich: Btw ... this reminded me that we should move the `Node` record away from address-only `$ip` to something string-based, perhaps `$host`. 9.1+, I guess.

tools/systemd-generator/src/zeek-cluster-config.h
```diff
@@ -11,11 +11,53 @@
...
 
 namespace zeek::detail {
 
+/**
+ *Split \a v by \a delim into a vector of string views.
+ */
+std::vector<std::string_view> split(std::string_view v, char delim);
+
+/**
+ * " ".join(...) in C++, meh.
```
ckreibich: Might be worth centralizing such helpers in utils.cc or some such so you have it in systemd-unit.cc as well. This code is always going to be a headache (not a complaint — just a consequence of what it does and the language!).

tools/systemd-generator/src/zeek-cluster-layout-generator.cc
```diff
@@ -1,35 +1,78 @@
 // See the file "COPYING" in the main distribution directory for copyright.
 
+// An opinionated and zero-dependency cluster-layout.zeek generator for single
```
ckreibich: And I still don't like "opinionated". :-)

tools/systemd-generator/src/zeek-cluster-layout-generator.cc
```diff
@@ -1,35 +1,78 @@
...
-// Usage: zeek-cluster-layout-generator -L <loggers> -P <proxies> -W <workers>
-// -a <listen_address> -p <port> -m <metrics_port> [-o outfile]
+// Usage Examples:
+//
+//     $ zeek-cluster-layout-generator -L <loggers> -P <proxies> -W <workers>
+//         -a <listen_address> -p <listen_port>
+//         -m <metrics_port> -b <metrics_address>
+//         [-o outfile]
+//
+//     The -W flag also accepts a tagged from like -W 'eth0:2 eth1:4' that
```
ckreibich:
```suggestion
//     The -W flag also accepts a tagged form like -W 'eth0:2 eth1:4' that
```

tools/systemd-generator/src/zeek-cluster-layout-generator.cc
```diff
@@ -1,35 +1,78 @@
...
+//     will include eth eth0 and eth1 "tags" in the cluster node name. The
+//     number following the colon is the number of workers.
+//
+//     Setting the metrics_port to 0 will allocate only use port 0/tcp.
```
ckreibich: (Not sure I follow this?)

ckreibich (commented):
I'm mostly through the code at this point. I think I don't have it in me to read it all super closely — it's nifty, its job is fairly straightforward, it's not very C++-suited material, and the tests likely matter more. I still need to look at those.

Gemini claims some minor findings that I'll look over tomorrow with fresh eyes.

doc/advanced/deployment/systemd.rst
ckreibich: To answer my own question, I found the `vars` array logic  in `systemd_write_units()`. So it's probably worth spelling them out?

doc/advanced/deployment/systemd.rst
awelzel: > Could we not use "opinionated"? 

Yeah, certainly! Thanks for pointing this out, I'll ditch that usage and made a note to avoid it in the future.

doc/advanced/deployment/systemd.rst
awelzel: > When would it not be an interface?

I was balancing between the resulting interface names that individual workers use (*after* interpolation of the `interface =` string) and the tag/name of the section. I do like `name` in the section as you propose a lot more than `tag`! It's true that the interpolated string isn't really exposed so this works. I'll switch/change this.

doc/advanced/deployment/systemd.rst
awelzel: > will be very common, right? Should that value (af_packet::<name>) be the default?

:-)

I had pondered, too. But pushing back here. `af_packet` is Linux specific and `zeek.conf` is not, so I'd avoid it as it'd mean a platform specific default. The `interface` entry is also what's directly passed to `-i` after variable interpolation.

Seems also useful to expose users (even new ones) to packet source plugins. It's a fairly important part that makes a live Zeek system work correctly.

A complex `zeek.conf` should always be generated using some configuration management tool.

doc/advanced/deployment/systemd.rst
awelzel: Yeah. I've been forth and back. They actually default to `1` to make everything work out of the box on a single-host.

I had pondered changing the defaults to `0` when the config has a hostname (for multi-host), or an extra option to set them all to 0, but then it seemed this should almost always be configuration management maintained in a production environment, so unclear what this is optimizing for (and the defaults are optimized for single-host demo "just" working)

```
[zeek]
only_workers = true

[zeek]
workers = only

[zeek]
only_workers
```

tools/systemd-generator/src/zeek-cluster-config.h
awelzel: We chatted about this in person, but putting it here again:

> it's not very C++-suited material

> just a consequence of what it does and the language!

I find it fairly important that the generator does not pull in a Python and so this doesn't leave a lot of options. Zeek script or Bash instead? :-)

bbannier: FTR, there's `implode_string_vector` in Zeek, but we cannot pull in that utility since we do not always provide a Zeek library and also wouldn't want to link to one since it would include far too much (init of globals during library loading?). A small standalone utility library could solve that, but we don't have one at this point either. If we had `fmt` we'd have [`fmt::join`](https://fmt.dev/latest/api/#range-and-tuple-formatting), but that's not available (yet 🥺👉👈).

EDIT:

> Zeek script or Bash instead? :-)

Well, we already use `awk` 🤡

awelzel: > Might be worth centralizing such helpers in utils.cc

Moved a bunch over. Thanks.

tools/systemd-generator/src/zeek-cluster-layout-generator.cc
awelzel: Dropped.

tools/systemd-generator/src/zeek-cluster-layout-generator.cc
awelzel: Thanks!

> //     Passing -m 0 will use metrics port 0/tcp for each node. This is useful
> //     to disable metrics exposition via Prometheus.

doc/advanced/deployment/systemd.rst
awelzel: Oh, and just to give a few examples for why I think the tag/name abstraction is useful:

> When would it not be an interface?

```
[interface internal-traffic]
interface = af_packet::eth1

[interface external-traffic]
interface = af_packet::eth2

[interface netmap-eth0]
interface = netmap::eth0{${worker_index0}

[interface vxlan]
interface = udp::0.0.0.0:4789:vxlan

[interface geneve]
interface = udp::[::]:6081:geneve

[interface napatech]
interface = napatech::${worker_index0}

# Four workers in a custom_lb group listening on separate
# interfaces: `veth1`, `veth2`, `veth3` and `veth4`.
# This is useful if you build a custom load-balancer/packet-broker.
# For example, a bond interface with balance-xor
# and xmit_hash_policy=encap3+4 and four slaves
# attached would do, but could also be an XDP program
# that flow-balances by forwarding packets to created
# interfaces.
[interface xdp-lb]
interface af_packet::veth${worker_index}
workers = 4
```

Not sure all of the above make sense, but the separation between actual `-i` value and a descriptive name seems very nice and avoids coming up with escaping/normalization patterns (the interface section name is used in working directories and cluster-node names).

doc/advanced/deployment/systemd.rst
awelzel: Phew ;-) I made a note about looking at `journalctl` and also how to run the generator by hand as a pre-check as well as mentioning AppArmor and SELinux. I don't think we should document concrete strategies (yet?).

doc/advanced/deployment/systemd.rst
awelzel: :-)

I'm not entirely sure here. I realize that with ZeroMQ or NATS, neither `$ip` nor `$host` would be all that interesting for pubsub functionality anymore.  One can connect to `tcp://mgr:5555` and `tcp://mgr:5556` and interact with the cluster (and similarly with a `nats_url`).

That's all to say that instead of moving `$ip` to `$host`, may also be able to remove these altogether.

doc/advanced/deployment/systemd.rst
awelzel: Yep! This was a long detour. I moved the cluster-layout chunks into a "Cluster Layout Notes" section as some pieces seemed useful, but it's a bit rough.

doc/advanced/deployment/systemd.rst
awelzel: The whole config needs some sort of reference documentation. I'd like to postpone this though and land this first. I'd guess there might be a few more tweaks and the PR here is already pretty big.

But added a table about the available variables for `interface =` specifically now.

doc/advanced/deployment/systemd.rst
awelzel: I've made this now be "name" in code and docs. Tag or identifier could  help to separate from the actual `-i` argument that'll be used, but "name" reads nicer.

awelzel:
> Gemini claims some minor findings that I'll look over tomorrow with fresh eyes.

Thanks for going through it! I've pushed some fixups and pushed back on some other comments :-)

doc/advanced/deployment/systemd.rst
ckreibich: Cool, thank you!

doc/advanced/deployment/systemd.rst
ckreibich: I agree, yeah.

doc/advanced/deployment/systemd.rst
ckreibich: Those examples are great. It makes a lot more sense to me now.

doc/advanced/deployment/systemd.rst
ckreibich: Okay. We can also always add this later. I chuckled when I saw this because it directly reminded me of
```
[manager]
role = manager
```
in the management framework, and feedback then (Dop, Smoot) was that it'll usually be redundant. (And I never fixed it. :-)

doc/advanced/deployment/systemd.rst
ckreibich: Mhmm. I would make the single-host scenario more explicit instead. I'm glad this came up again here because I found that two-liner of 
```
    interface = af_packet::eth1
    workers = 2
```
in the single-host setting a bit extreme when I first saw it. Does that default to one logger? To manager-is-logger? To one proxy or a handful? Etc. Most people will also start from the default zeek.conf, which already lists the other node types anyway, just commented out. So if a minimal full cluster becomes
```
    interface = af_packet::eth1
    workers = 2
    logger = 1
    proxy = 1
    manager = 1
``` 
or perhaps
```
    interface = af_packet::eth1
    workers = 2
    proxy = 1
    manager = 1
```
for manager-is-logger, I think that's pretty clean. `manager = 1` looks a bit pointless, but there's some future-proofing in this: we've already considered loggerless clusters, people have asked for standby managers, and perhaps you actually want to run experimental/short-lived setups that aren't complete clusters. Long story short, I think being explicit is good.

> so unclear what this is optimizing for

I mean, clutter? If you have a 10 machine cluster, you need to mention all the stuff they're each _not_ running roughly 10 times.

tools/systemd-generator/src/utils.cc
```diff
@@ -0,0 +1,228 @@
...
+
+            l = parse_int(parts[0]);
+
+            if ( ! l.has_value() || ! r.has_value() || *l < 0 || *r < 0 || *r < *l ) {
+                is_valid = false;
+                return;
+            }
+
+            // Expand range with strides.
+            for ( int i = *l; i <= *r; i += stride )
```
ckreibich, line 203: Gemini complained about this one not having an upper bound, so something like `0-2147483647` expands out to that many objects in `cpus`. My first reaction was to discard this. In the process I learned about `CPU_SETSIZE`, which caps the number of CPUs you can store in a `cpu_set_t` — so that might be a natural cap.

The bigger questions are probably:
* To which extent we want to sanity-check the values we're given. I'm fine considering this a pretty sharp tool right now!
* Whether we rule out that a regular user can ever maliciously modify the config to then subvert the generation. Seems like that's exactly why SELinux complained about my original local-user Zeek tree install.

tools/systemd-generator/src/systemd-unit.cc
ckreibich: Gemini picked up that using `ss << "NUMAPolicy=" << *numa_policy << "\n";` for these (and similars) would be more consistent and save a string allocation.

tools/systemd-generator/src/systemd-unit.h
ckreibich:
```suggestion
    std::string Name() const { return file.filename().string(); }
```
(Github doesn't seem to render the original line correctly in a suggestion if it's not part of the commits, grrr.)

tools/systemd-generator/src/utils.cc
```diff
@@ -0,0 +1,228 @@
...
+
+std::string join(std::span<const std::filesystem::path> paths, const std::string& sep) {
+    std::vector<std::string> args;
+    args.reserve(paths.size());
+    for ( const auto& p : paths )
+        args.emplace_back(p.string());
+    return join(args, sep);
+}
+
+std::optional<std::string> substitute_vars(const std::string& s, const std::map<std::string, std::string>& vars) {
```
ckreibich, line 92: Gemini recommended a `std::string_view` for s. I don't think it matters for our current usage.

ckreibich (commented):
This one has a bit of Gemini feedback. All nits, though one brings up questions around trust boundaries around zeek.conf that could go into a general warning about the execution model (maybe relating to AppArmor/SELinux etc).

testing/btest/tools/systemd-generator/test-cluster-ipv6.sh
```diff
@@ -0,0 +1,64 @@
...
+# @TEST-EXEC: zeek --parse-only ./cluster-layout.zeek
+# @TEST-EXEC: btest-diff cluster-layout.zeek
+
+# @TEST-START-FILE etc/cluster/mgr.zeek.conf
+[zeek]
+manager = 1
+loggers = 1
+proxies = 2
+
+address = [2001:400:211::121]
```
ckreibich, line 30: This had me wondering whether we should document the `[...]` notation for IPv6 in the default zeek.conf.

testing/btest/tools/systemd-generator/test-cluster-node-prefix.sh
```diff
@@ -0,0 +1,22 @@
+# @TEST-DOC: Test the most minimal zeek.conf file.
```
ckreibich: Looks like a cut'n'paste leftover from the minimal tests.

testing/btest/tools/systemd-generator/test-errors.sh
```diff
@@ -0,0 +1,25 @@
+# @TEST-DOC: Test some invalid configurations.
```
ckreibich, line 1: This had me wondering whether it actually hits all errors or bails early. In my tinkering I futzed with multi-line values to options that shouldn't take them, which I think throws an exception for immediate exit (but that's not part of this test). Feel free to ignore.

ckreibich (commented):
My last set of comments! I've read through the tests — very nice. I'm not sure I missed it, but perhaps add one that invokes the generator with `normal-dir early-dir late-dir`, if that's not there yet.

One more for the docs: we should probably mention that the unit files default to user/group `zeek`/`zeek`. The shipped zeek,conf says so, but if one follows the docs as a tutorial that's not clear yet. Perhaps we just say that the shipped zeek,conf explains all available options.

And finally: should we give some guidance about how to learn more when something goes wrong? Depending on how I make it stumble I either get
```
Jul 24 17:59:50 hogthrob4 (generato[651056]: /etc/systemd/system-generators/zeek-systemd-generator failed with exit status 1.
```
or
```
Jul 24 17:03:36 hogthrob4 (generato[638946]: /etc/systemd/system-generators/zeek-systemd-generator terminated by signal ABRT.
```
So one tip could be a manual invocation of the generator.

Maybe a future iteration could try the logging to `/dev/kmsg` that the docs mention. :-)

Again: great stuff!

doc/advanced/deployment/systemd.rst
awelzel: > I'm glad this came up again here because I found that two-liner of [...]
> in the single-host setting a bit extreme when I first saw it. Does that default to one logger? To manager-is-logger? 

> I mean, clutter?  If you have a 10 machine cluster, you need to mention all the stuff they're each not running roughly 10 times.

Yeah, this is good. Thanks for the pushback! 

I've changed the defaults for the manager, proxies, loggers and workers options to be `0` and updated the tests. A bit of churn, but agree the explicitness feels much better. Only putting interface into zeek.conf doesn't work anymore, but that's alright. The `archiver` option defaults to `1 if (manager == 1 ||  loggers >= 1) else 0` if unset now.

tools/systemd-generator/src/utils.cc
awelzel: Ignored this one.

testing/btest/tools/systemd-generator/test-errors.sh
awelzel: > Feel free to ignore.

I moved from some std::exit(1) to accumulating the errors previously, but might not be clean everywhere. Seems that could be fixed up later on as well.

tools/systemd-generator/src/utils.cc
awelzel: > To which extent we want to sanity-check the values we're given. 

Can limit to CPU_SETSIZE. Don't think the unit itself will even start if there's nonsensical CPU values. I'd not pre-check anything about the available CPUs of the system here if that's what you were wondering, that'll only make testing difficult (what if CI only has 2 CPUs, but want to test CPU list 0-7 or so).


> Whether we rule out that a regular user can ever maliciously modify the config to then subvert the generation. 

`zeek.conf` should be `root:root`  and `644` or so in a production environment and the generator should be `root:root` and `755`.

> Seems like that's exactly why SELinux complained about my original local-user Zeek tree install.

Could you paste the original SELinux message: Isn't the crux here the existence of a symlink in `/etc/systemd/system-generators` pointing at an executable that itself is owned/controlled by an unprivileged user (or in a user writable directory)? That alone opens the door to replace the pointed at `zeek-systemd-generator` executable with a different malicious generator.

doc/advanced/deployment/systemd.rst
awelzel: The section-less format would not have:
```
manager = 1
loggers = 1
proxies = 1
workers = 4
interface = af_packet::eth1
```

And for INI-style, the section name is used in working directories, unit names and cluster node names rather than actually influencing the effective interface used by a worker, so seems a separate enough use to keep it explicit.

```
[zeek]
manager = 1
loggers = 1
proxies = 1

[interface eth1]
interface = af_packet::eth1
workers = 4
```


Seems also okay to type by hand in demos and trivial to have a jinja template or some such elsewhere.

tools/systemd-generator/src/utils.cc
awelzel: > Can limit to CPU_SETSIZE.

So I did the change only to then change my mind.

The generated units will fail to do the CPU pinning with invalid arg, so that's good. Also, a `taskset` implementation would fail at runtime:
```
$ taskset -c 4711  ls
taskset: failed to set pid 2214502's affinity: Invalid argument
```

From the `zeek.conf` perspective, there's nothing wrong configuring a zeek worker to be pinned to CPU 4711, so lets allow that. The produced unit file will be valid.

It's also possible to use `workers = 1000000` and it'll probably knock something over, but you also asked for that, so good luck!

testing/btest/tools/systemd-generator/test-cluster-node-prefix.sh
awelzel: Thanks for pointing that out.

cluster_node_prefix doesn't currently work in single-host mode, but this can be fixed. Thanks.

awelzel: Fixed now.

awelzel:
> Maybe a future iteration could try the logging to /dev/kmsg that the docs mention. :-)

Oh, nice! I added a commit that uses `/dev/kmsg` instead of stderr when not a tty and not an explicit config. And some minor docs updates around this.

```
Jul 27 16:56:37 tinkyx1 systemd[1]: Reloading...
Jul 27 16:56:38 tinkyx1 /etc/systemd/system-generators/zeek-systemd-generator: config /opt/zeek-dev-prod/etc/zeek/zeek.conf is invalid
Jul 27 16:56:38 tinkyx1 /etc/systemd/system-generators/zeek-systemd-generator: invalid workers value: '16x'
```

> The shipped zeek,conf says so, but if one follows the docs as a tutorial that's not clear yet. Perhaps we just say that the shipped zeek,conf explains all available options.

I'll update the `zeek.conf` a bit, but would punt on polishing. As mentioned this needs more docs before 9.0. The added section wasn't meant to be a reference.

testing/btest/tools/systemd-generator/test-cluster-ipv6.sh
awelzel: I've added it to `zeek.conf` and also implemented validation for that option.

doc/advanced/deployment/systemd.rst
awelzel: I've added some of these examples to the default config as comments.

awelzel:
@ckreibich - I pushed a bunch of more fixups. Thanks for bringing this into shape. I'll ~push~ squash everything before a merge. How does it look for you at this point?

As mentioned elsewhere, reference docs would be separate. I'm thinking `man 5 zeek.conf`, but not sure this prevents from having online docs. Anyhow, not for this PR.

ckreibich (approved):
I think it looks awesome — thanks for working through it all! I am sure we'll be iterating on this more so for now: 🚀 😎

PR #5693, opened 2026-07-16
docs: add log-permissions guidance to Security Considerations

sujalavnelavai:
This update adds a new protective measure highlighting the importance of restricting permissions on Zeek’s log directory and log files. Zeek’s logs can contain sensitive operational metadata such as internal hostnames, URLs, and protocol details derived from network traffic. If these files are left world-readable or writable, they may unintentionally expose information that should remain confidential.

The added guidance reflects common security practice and aligns with widely accepted log-management recommendations, including those outlined in NIST SP 800-92. The goal is to help operators avoid accidental data exposure by ensuring that only authorized users and processes can access Zeek’s logs.

AI Disclosure: I used Microsoft Copilot to refine the wording of this PR. All technical content, references, and accuracy were reviewed and validated by me.

### DIFF
--- a/doc/security-considerations.rst
+++ b/doc/security-considerations.rst
@@ -60,3 +60,10 @@ The following lists steps you can take to protect your Zeek cluster.
   systemd's `built-in capabilities management
   <https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#AmbientCapabilities=>`_
   to apply them.
+
+* Zeek logs often contain sensitive operational information, including internal
+  hostnames, URLs, protocol metadata, and other network‑derived information.
+  Ensure that both the log directory and the individual log files are not
+  world‑readable or writable, and restrict access to authorized users and
+  processes. Misconfigured permissions can expose sensitive network information
+  and undermine the confidentiality of security logs.


sujalavnelavai:
Hello @evantypanski , as per the feedback I got from the discussion named Beginner friendly common pitfalls in Zeek deployment I have added one security pitfall in the doc " security considerations". Please let me know if further changes are required. Thank you.

doc/security-considerations.rst
```diff
@@ -60,3 +60,11 @@ The following lists steps you can take to protect your Zeek cluster.
...
   <https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#AmbientCapabilities=>`_
   to apply them.
+
+**Log file permissions**: Zeek logs often contain sensitive operational
+  information, including internal hostnames, URLs, protocol metadata, and
+  other network‑derived information. Ensure that both the log directory and the
+  individual log files are not world‑readable or writable, and restrict access
+  to authorized users and processes. Misconfigured permissions can expose
+  sensitive network information and undermine the confidentiality of security
+  logs,which is inconsistent with the confidentiality recommendations in NIST SP 800-92.
```
evantypanski: I think this is true on its face without extra need for a NIST publication, I would remove this


```suggestion
  logs.
```

doc/security-considerations.rst
```diff
@@ -60,3 +60,11 @@ The following lists steps you can take to protect your Zeek cluster.
   systemd's `built-in capabilities management
   <https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#AmbientCapabilities=>`_
   to apply them.
+
+**Log file permissions**: Zeek logs often contain sensitive operational
```
evantypanski: can you match the format of the other bullet points? This should not differ from all of the others.

evantypanski (changes requested):
this seems reasonable to call out, similar to locking down the cluster mentioned before. I have a couple minor changes, the content seems good to me though!

sujalavnelavai:
@evantypanski , I have removed the cited reference and changed the bullet format. If any changes are required further, please let me know . Thank you.

sujalavnelavai:
One more question I need to clarify which helps for my future Zeek contributions.

NIST, CISA, ENISA, OWASP AND MITTRE ATT&CK are authentic security frameworks.  So my question is, whether these references should not be cited in Zeek docs or can I cite when necessary or Zeek does have its own security standards? because even one of the maintainer told me to crosslink the zeek documents for references previously.

Or because standards keeps on evolving, the links break often and external docs change. Are these the reasons for not citing those references mentioned above?

evantypanski:
> whether these references should not be cited in Zeek docs or can I cite when necessary or Zeek does have its own security standards?

They can be, but only if they seem to add a substantial amount. In this case, you did a good job describing why you would lock down log directories. So the recommendation is true on its face and clear how to followup, in my opinion. For other cases, if more justification is needed, I think it makes sense to include the reference. I'm sure where each person draws that line may be different, though.

In general I would say: if you can describe something concisely like this case, no need to mention a framework/publication/etc. It's not strictly wrong to include one, just potentially unnecessary.

I also have a general desire to keep Zeek docs self-contained too, though. Which means we only talk about Zeek and tend to not link out, since as you mention those can change. But, again, that line isn't set in stone, so sometimes links/references may be preferred (I would rather link out to an external protocol's RFC than restate it, for example).

sujalavnelavai:
Thanks for the clarification @evantypanski . Let me know if further changes are required for this PR. Thank you ones again.

evantypanski (approved):
I squashed the commit to merge; thanks!

sujalavnelavai:
Thank you for accepting my contribution @evantypanski.

PR #5694, opened 2026-07-16
Fix the coverage script to remove Cirrus variables

timwoj:
This should fix uploading coverage from CircleCI


awelzel (approved):
I wonder if we should make `lcov_html.sh` use `ZEEK_CI` variables, but I'll bring it up when we change the CI provider once more :-)

PR #5695, opened 2026-07-16
Remove zeek-testing-cluster commit-hash file

timwoj:
A bunch of these tests moved over into the regular btests with https://github.com/zeek/zeek/pull/5622. We're not testing this directly anymore, so the commit hash file can be removed.


ckreibich (approved):
Yep, that testsuite exclusively covers the management framework now. The docker-compose approach might come in handy again in the future, but for now there's no need.

PR #5699, opened 2026-07-17
Add `X` history to a few configurable limits

evantypanski:
We've added a few configurable limits recently. This adds `X` to the conn history for those.

I only did this for *configurable* limits that were outside Spicy parsers. Also added tests for a couple of the DNS limits

Used Claude to find where the limits should be set and generate pcaps. I wrote all code


src/analyzer/protocol/dns/DNS.cc
```diff
@@ -100,6 +100,7 @@ void DNS_Interpreter::ParseMessage(const u_char* data, int len, int is_query) {
     // There is a great deal of non-DNS traffic that runs on port 53.
     // This should weed out most of it.
     if ( zeek::detail::dns_max_queries > 0 && msg.qd_zo_count > zeek::detail::dns_max_queries ) {
+        analyzer->Conn()->CheckHistory(zeek::session::detail::HIST_UNKNOWN_PKT, 'X');
```
awelzel, line 103: Mostly due to the changes in zeek-testing-private: Given we treat this as an analyzer violation just the line below, I'm not sure about adding the `X` here, but maybe yes.

Mind adding the actual `qd_zo_count` to the weird in `%d > %d` notation. This would help deciding if `dns_max_queries` might be a bit too low, or if it's totally off.

awelzel: Otherwise LGTM.

evantypanski: Ah yeah I considered changing all of these to use the `%d > %d` since I'm a fan of that but forgot. What do you think about adding it to all of them?

I considered not adding to history with the analyzer violation here, but I think there's still at least some signal from hitting a configurable limit, so I think the `X` is complementary. I, too, am on the fence, though. I added it since it does have a weird, too.

evantypanski: Added it only for one other, lmk what you think. Baselines updated too. Seems like "totally off" is more accurate IMO :)

timwoj:
This is somewhat related, but maybe not for this PR. It'd be nice if there was a BIF for managing the history as well. I ran into it with at least one security ticket where all of the changes are in script land, but I'd have to drop back to C++ to update the history.

src/analyzer/protocol/dns/DNS.cc
awelzel: Yep. Interestingly it seems mostly these two values (other protocol over port 53), so that additional information helps to weed them out. While for mdns, I've seen "above 25 but below 100" and so that could help guide tuning.

```
DNS_Conn_count_too_large	53638 > 25
DNS_Conn_count_too_large	53380 > 25
```

Makes sense to me to add it to the DNS chain depth and also expand the MIME depth one.

Thanks!

evantypanski: I avoided those two cases since they trigger the weird when they hit some limit. But, we don't know that limit beforehand, it's just based on some depth that gets incremented.

the two cases I added, on the other hand, are just "this value in the protocol is higher than our configurable limit" so it makes more sense.

The MIME one ends up with `100 >= 100`, for example, which is just weird. Every time both of the numbers will just be the limit. And the MIME one already puts in the limit is `addl`

awelzel: > The MIME one ends up with 100 >= 100, for example, which is just weird. 

Ah, I wasn't thinking enough about this. :+1:

awelzel (approved):

PR #5700, opened 2026-07-17
Fix flood of warnings when building `spicyz`

timwoj:
This is the Zeek-side counterpart to https://github.com/zeek/spicy/pull/2440. It pulls in the submodule update to fix the warnings fixed there. 

The main change here is that currently `spicyz` is built with `-Wall`, where the rest of the Zeek build isn't. This is fine on Linux, but on Windows `-Wall` includes a lot of additional warnings that are somewhat obnoxious (things like implicit deletion of assignment/copy/move). I changed it to drop back to `/W4` which keeps the important warnings but removes a lot of others.

It also fixes some additional warnings not caught by the upstream Spicy changes.

### DIFF
--- a/src/spicy/spicyz/CMakeLists.txt
+++ b/src/spicy/spicyz/CMakeLists.txt
@@ -5,16 +5,9 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/config.h DESTINATION include/zeek/spic
 
 add_executable(spicyz driver.cc glue-compiler.cc main.cc zeek-version.cc)
 
-# Don't enable /Wall on Windows because it also enables a bunch of very obnoxious
-# warnings by default. Turn it down to a smaller-but-still-loud subset.
-if (MSVC)
-    target_compile_options(spicyz PRIVATE "/W4")
-else ()
-    target_compile_options(spicyz PRIVATE "-Wall")
-endif ()
-
 target_compile_features(spicyz PRIVATE "${ZEEK_CXX_STD}")
 set_target_properties(spicyz PROPERTIES CXX_EXTENSIONS OFF)
+
 # We do not compile spicyz's own sources with sanitizers — the global linker
 # flags ensure the ASan runtime is available for the sanitized HILTI runtime
 # objects in libhilti.so.
--- a/src/spicy/spicyz/CMakeLists.txt
+++ b/src/spicy/spicyz/CMakeLists.txt
@@ -4,7 +4,15 @@ configure_file(config.h.in config.h)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/config.h DESTINATION include/zeek/spicy/spicyz)
 
 add_executable(spicyz driver.cc glue-compiler.cc main.cc zeek-version.cc)
-target_compile_options(spicyz PRIVATE "-Wall")
+
+# Don't enable /Wall on Windows because it also enables a bunch of very obnoxious
+# warnings by default. Turn it down to a smaller-but-still-loud subset.
+if (MSVC)
+    target_compile_options(spicyz PRIVATE "/W4")
+else ()
+    target_compile_options(spicyz PRIVATE "-Wall")
+endif ()
+
 target_compile_features(spicyz PRIVATE "${ZEEK_CXX_STD}")
 set_target_properties(spicyz PROPERTIES CXX_EXTENSIONS OFF)
 # We do not compile spicyz's own sources with sanitizers — the global linker
--- a/src/spicy/spicyz/driver.h
+++ b/src/spicy/spicyz/driver.h
@@ -147,7 +147,7 @@ protected:
      *
      * @param t type's meta information
      */
-    virtual void hookNewType(const TypeInfo& ti) {}
+    virtual void hookNewType(const TypeInfo& /*ti*/) {}
 
     /** Overridden from HILTI driver. */
     void hookNewASTPreCompilation(const hilti::Plugin& plugin, hilti::ASTRoot* root) override;
--- a/src/spicy/spicyz/glue-compiler.cc
+++ b/src/spicy/spicyz/glue-compiler.cc
@@ -1458,7 +1458,7 @@ struct VisitorZeekType : spicy::visitor::PreOrder {
         return *x;
     }
 
-    void operator()(hilti::type::Address* t) final { result(base_type("zeek_rt::ZeekTypeTag::Addr")); }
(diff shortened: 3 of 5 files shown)


src/spicy/spicyz/driver.h
```diff
@@ -147,7 +147,7 @@ class Driver : public ::spicy::Driver {
      *
      * @param t type's meta information
      */
-    virtual void hookNewType(const TypeInfo& ti) {}
+    virtual void hookNewType(const TypeInfo& /*t*/) {}
```
0xxon: this might be a silly question - but as I wondered about this while reading it. Is there a reason to do this instead of the [[maybe_unused]] used for all the others below?

Also - why t instead of ti? :)

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -924,7 +924,7 @@ glue::Export GlueCompiler::parseExport(const std::string& chunk) {
     }
 
     bool expect_fields = false;
-    bool include_fields;
+    bool include_fields = false;
```
bbannier: This was probably truthy before (still UB), could you change this to

 
```suggestion
    bool include_fields = true;
```

With that we can remove `include_fields = true` assignment below, and only need to worry about the branch deviating from the default.

src/spicy/spicyz/glue-compiler.cc
```diff
@@ -1493,15 +1493,23 @@ struct VisitorZeekType : spicy::visitor::PreOrder {
...
-    void operator()(hilti::type::Bool* t) final { result(base_type("zeek_rt::ZeekTypeTag::Bool")); }
-    void operator()(hilti::type::Bytes* t) final { result(base_type("zeek_rt::ZeekTypeTag::String")); }
-    void operator()(hilti::type::Interval* t) final { result(base_type("zeek_rt::ZeekTypeTag::Interval")); }
-    void operator()(hilti::type::Port* t) final { result(base_type("zeek_rt::ZeekTypeTag::Port")); }
-    void operator()(hilti::type::Real* t) final { result(base_type("zeek_rt::ZeekTypeTag::Double")); }
-    void operator()(hilti::type::SignedInteger* t) final { result(base_type("zeek_rt::ZeekTypeTag::Int")); }
-    void operator()(hilti::type::String* t) final { result(base_type("zeek_rt::ZeekTypeTag::String")); }
-    void operator()(hilti::type::Time* t) final { result(base_type("zeek_rt::ZeekTypeTag::Time")); }
-    void operator()(hilti::type::UnsignedInteger* t) final { result(base_type("zeek_rt::ZeekTypeTag::Count")); }
+    void operator()([[maybe_unused]] hilti::type::Bool* t) final { result(base_type("zeek_rt::ZeekTypeTag::Bool")); }
```
bbannier: No problem with this change, but 1) using `[[maybe_unused]]` seems more noisy to me (maybe I just need to get used to it), and 2) this attribute doesn't seem to be a widespread pattern in the Zeek codebase. My hunch would be to simply wrap the variables in comments, but up to you.

src/spicy/spicyz/CMakeLists.txt
```diff
@@ -4,7 +4,15 @@ configure_file(config.h.in config.h)
...
-target_compile_options(spicyz PRIVATE "-Wall")
+
+# Don't enable /Wall on Windows because it also enables a bunch of very obnoxious
+# warnings by default. Turn it down to a smaller-but-still-loud subset.
+if (MSVC)
+    target_compile_options(spicyz PRIVATE "/W4")
+else ()
+    target_compile_options(spicyz PRIVATE "-Wall")
+endif ()
+
```
bbannier: While you are at it, this was doing one-off handling for this Zeek target which caused this maintenance task. Do we have a macro or variable we could use instead to be consistent with other pieces?

src/spicy/spicyz/driver.h
bbannier: I suspect this was renamed accidentally in the same pattern as the changes in `src/spicy/spicyz/glue-compiler.cc`. There `t` refers to a type instances so `t` made sense; here we are referring to a `TypeInfo` so we might want to keep `ti` to make the distinction clearer (happy to hear how (not) useful that is though).

On the `[[maybe_unused]]`, that seems to be a novel pattern which IMO is pretty noisy when putting the var binding name into a comment does the same with less ceremony and also consistently with code elsewhere 🤷.

bbannier (approved):
Sorry for not looking into this earlier, it only popped up for me when Johanna left a comment. This looks good to me, but I still need to dig into the changes in Spicy proper.

src/spicy/spicyz/driver.h
timwoj: > I suspect this was renamed accidentally in the same pattern as the changes in src/spicy/spicyz/glue-compiler.cc.

Yep, I'll change it back.

My concern with using `[[maybe_unused]]` is whether it only applies to the initial `virtual` method here or whether any overrides of it would too. A quick test on CompilerExplorer appears to not care if it's just on the base method. I'm fairly indifferent about which model to use. `[[maybe_unused]]` is definitely noisier.

src/spicy/spicyz/CMakeLists.txt
timwoj: What's happening here is that for all of the other files, we add `-Wall` via `cmake/SetDefaultCompileFlags.cmake`. That file is a no-op for Windows. So while the rest of the build doesn't get it applied on Windows, this one target does. I don't know why we want `-Wall` specifically for this code though. @rsmmr was probably the one who added it.

Having it enabled definitely caught some code-smells but I can also see removing it to be consistent with the rest of the build.

timwoj:
I fixed the comments here and removed the uses of `[[maybe_unused]]`. I'll drop the spicy-submodule update out of it and merge it tomorrow. The changes in this PR don't explicitly need anything from the submodule bump.

src/spicy/spicyz/CMakeLists.txt
bbannier: I'd be in favor of just doing what we do elsewhere now, the spicy plugin has been soaked enough that it should just be treated as just another Zeek component, and any improvements we do here we should do anywhere as well. Fine if that means just dropping this.

timwoj: Works for me. It does make me curious just how loud `/W4` would be for the entire Zeek build, but we have better things to do that look at that.

PR #5706, opened 2026-07-20
NFA: Drop the static closuremap

awelzel:
Since 54b687a83adb5c97437d02c1cdb2e523f1fdfbdb, the closuremap is cleared and identifiers re-used, but remained a static instance.

@initconf reports the memset() call within Clear() of the IntSet becomes significant overhead for subsequent trivial string_to_pattern() calls once the closuremap grew large. This is due to always zeroing out all of the allocated memory.

Seem easiest to just use a new instance and select the initial size to be the number of NFA states and rely on the allocator to be fast, rather than trying to find some magic constant.

---

cc @vpax - does that look reasonable to you?

### DIFF
--- a/src/NFA.cc
+++ b/src/NFA.cc
@@ -282,9 +282,7 @@ NFA_Machine* make_alternate(NFA_Machine* m1, NFA_Machine* m2) {
 }
 
 NFA_state_list* epsilon_closure(NFA_state_list* states) {
-    // We just keep one of this as it may get quite large.
-    static IntSet closuremap;
-    closuremap.Clear();
+    IntSet closuremap(states->size());
 
     NFA_state_list* closure = new NFA_state_list;
 
--- a/src/IntSet.h
+++ b/src/IntSet.h
@@ -22,8 +22,6 @@ public:
     void Remove(unsigned int i);
     bool Contains(unsigned int i) const;
 
-    void Clear();
-
 private:
     void Expand(unsigned int i);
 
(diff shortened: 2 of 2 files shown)


src/NFA.cc
```diff
@@ -282,9 +282,7 @@ NFA_Machine* make_alternate(NFA_Machine* m1, NFA_Machine* m2) {
 }
 
 NFA_state_list* epsilon_closure(NFA_state_list* states) {
-    // We just keep one of this as it may get quite large.
-    static IntSet closuremap;
-    closuremap.Clear();
+    IntSet closuremap(states->size());
```
bbannier, line 285: Looking at `IntSet::Clear`, we never actually reclaimed the memory with it anyway and just zeroed the state, and since the size can only grow from `Insert` resizes, `Clear` payed more and more as the process kept running. Since C++ std containers also reclaim memory on `clear` that behavior seems dangerously surprising, and I wonder whether we should just drop that method without replacement. This seems to have been the only user.

In any case, the change here seems to be the correct approach given what was initially intended.

awelzel: > Since C++ std containers also reclaim memory on clear that behavior seems dangerously surprising

I don't think this is guaranteed, but yeah, happy to drop `Clear()`. I do think the IntSet implementation could  be replaced `std::vector<bool>`, but didn't want to go there.

bbannier: Right, `IntSet::Clear` behaves correctly if one understands that it behaves more like a contiguous container than a set, naming is hard 😞

vpax:
Yes looks reasonable. (FYI This is *ancient* code, likely pre-dates Zeek even, back when the allocation cost was a concern.) `vector<bool>` makes sense too, providing there's a guarantee that state IDs are confined to a ~contiguous range starting at 0-ish (which I think they are); so would unordered_set<int>. But also reasonable to not introduce a data structure change here.

bbannier (approved):
🚀

PR #5679, opened 2026-07-12
Add inline‑enforcement clarification to what.rst

sujalavnelavai:
This PR makes a small documentation improvement to doc/about/what.rst by adding one missing clarification about Zeek’s operational model. The existing section already explains what Zeek is and is not, but it does not explicitly state that Zeek cannot block, drop, or modify traffic. This PR adds a single bullet point to the end of the conceptual summary to clarify that Zeek is not an inline enforcement system.

This PR fixes the issue # 5676

AI Transparency Note:  
I used Microsoft Copilot to help structure the wording and align the content with Zeek’s RST documentation style. All technical content, explanations, and revisions were reviewed and validated by me. The updates in this PR were made based on information from the official Zeek documentation, Suricata documentation, Security Onion documentation, Malcolm documentation, and standard SOC correlation workflows.

PR #5711, opened 2026-07-21
Docs: edits on top of #5679

ckreibich:
If we want to touch packet forwarding, let's foreshadow NetControl and Websockets, and mention Snort too.

### DIFF
--- a/doc/about/what.rst
+++ b/doc/about/what.rst
@@ -20,8 +20,8 @@ post-processing with external software. Users can also choose to have external
 databases or SIEM products consume, store, process, and present the data for
 querying.
 
-In addition to the logs, Zeek comes with built-in functionality for a range of
-analysis and detection tasks, including extracting files from HTTP sessions,
+Zeek also comes with built-in functionality for a range of additional
+analysis and detection tasks, including extracting files from network sessions,
 detecting malware by interfacing to external registries, reporting vulnerable
 versions of software seen on the network, identifying popular web applications,
 detecting SSH brute-forcing, validating SSL certificate chains, and much more.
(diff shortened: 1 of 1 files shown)


doc/about/what.rst
```diff
@@ -67,17 +67,19 @@ Zeek within one system for as long as possible, and then transparently add more
...
+intrusion detection systems such as Snort and Suricata.
 
-Zeek does not participate in packet-forwarding decisions and cannot block,
-drop, or modify network traffic. It is an observation and analysis platform,
-not an inline enforcement system. Preventing malicious traffic is the
-responsibility of intrusion prevention systems (IPS), such as Suricata
-operating in IPS mode.
+Zeek is also not a protocol analyzer in the sense of Wireshark, seeking to
+depict every element of network traffic at the frame level, or a system for
+storing traffic in packet capture (PCAP) form. Rather, Zeek sits at the “happy
```
timwoj: Except that we do have the packet dumper, even if it's not really a main part of what Zeek does.

ckreibich: Yeah. There are lots of ifs-and-buts in the whole document. If you'd like to do additional tweaks by all means commit on top of my branch.

ckreibich: I tweaked the wording lightly on this one, fyi — it's now "or a system for storing traffic in packet capture (PCAP) form at scale", which signals that it's possible, but you wouldn't want it for bulk pcap capture.

timwoj: I like that wording more. It's more vague, but I think that's actually correct.

timwoj (approved):

PR #5720, opened 2026-07-24
Initialize the Prometheus listener via zeek_init()

### LINKED ISSUE #5691: Telemetry manager tries to listen on its port in --parse-only mode
I've set a few environment variables for a test that contains `@if ( getenv(...) != "" )` logic and wanted to test with `--parse-only`. Got a message failing to bind a port.

```
$ CLUSTER_NODE=worker-1 ZEEK_TELEMETRY_LISTEN_ADDRESS=1.1.1.1 zeek ./cluster-layout.zeek  --parse-only
fatal error: Failed to setup Prometheus endpoint: null context when constructing CivetServer. Possible problem binding to port. Error: Failed to setup server ports. Attempted to bind to 1.1.1.1:9994.
```

Expectation: `--parse-only` should not open any ports.

I guess we could just handle `detail::parse_only` in `telemetry/Manager.cc:InitPostScript()`. Alternatively would be to provide a builtin function to start the Prometheus exposer from `zeek_init()` which would probably be nicer and more aligned with starting a WebSocket server.

timwoj:
This moves the initialization of the Prometheus server in the Telemetry manager from right after script processing to a function that gets called by `zeek_init`. This prevents calling it when using `--parse-only` but also feels better that it gets initialized during startup instead of during parsing.

Fixes #5691

### DIFF
--- a/scripts/base/frameworks/telemetry/main.zeek
+++ b/scripts/base/frameworks/telemetry/main.zeek
@@ -6,6 +6,7 @@
 
 @load base/misc/version
 @load base/bif/telemetry_functions.bif
+@load base/frameworks/telemetry/options
 
 module Telemetry;
 
@@ -467,6 +468,14 @@ global version_gauge_family = Telemetry::register_gauge_family(Telemetry::Metric
 
 event zeek_init()
 	{
+	if ( getenv("ZEEKCTL_CHECK_CONFIG") == "" && (Telemetry::metrics_port as count) != 0 )
+		{
+		local expose_services_json = Cluster::local_node_type() == Cluster::MANAGER;
+		local listen_addr = |Telemetry::metrics_address| > 0 ? Telemetry::metrics_address : "0.0.0.0";
+
+		Telemetry::listen_prometheus(listen_addr, Telemetry::metrics_port, expose_services_json);
+		}
+
 	local v = Version::info;
 	local labels = vector(cat(v$version_number),
 	                      cat(v$major), cat(v$minor), cat (v$patch),
--- a/src/telemetry/Manager.cc
+++ b/src/telemetry/Manager.cc
@@ -48,68 +48,6 @@ Manager::Manager() : IOSource(true) { prometheus_registry = std::make_shared<pro
 Manager::~Manager() = default;
 
 void Manager::InitPostScript() {
-    // Metrics port setting is used to calculate a URL for prometheus scraping
-    std::string prometheus_url;
-    auto metrics_port = id::find_val("Telemetry::metrics_port")->AsPortVal();
-    auto metrics_address = id::find_val("Telemetry::metrics_address")->AsStringVal()->ToStdString();
-    if ( metrics_address.empty() )
-        metrics_address = "0.0.0.0";
-    if ( metrics_port->Port() != 0 )
-        prometheus_url = util::fmt("%s:%u", metrics_address.data(), metrics_port->Port());
-
-    if ( ! prometheus_url.empty() && ! getenv("ZEEKCTL_CHECK_CONFIG") ) {
-        CivetCallbacks* callbacks = nullptr;
-        auto local_node_name = id::find_val("Cluster::node")->AsStringVal();
-        if ( local_node_name->Len() > 0 ) {
-            auto cluster_nodes = id::find_val("Cluster::nodes")->AsTableVal();
-            auto local_node = cluster_nodes->Find(IntrusivePtr<StringVal>{NewRef{}, local_node_name});
-            auto local_node_type = local_node->AsRecordVal()->GetField<EnumVal>("node_type")->Get();
-
-            static auto node_type_type = id::find_type("Cluster::NodeType")->AsEnumType();
-            static auto manager_type = node_type_type->Lookup("Cluster", "MANAGER");
-
-            if ( local_node_type == manager_type ) {
-                BuildClusterJson();
-
-                callbacks = new CivetCallbacks();
-                callbacks->begin_request = [](struct mg_connection* conn) -> int {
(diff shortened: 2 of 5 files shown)


scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -467,6 +467,8 @@ global version_gauge_family = Telemetry::register_gauge_family(Telemetry::Metric
 
 event zeek_init()
 	{
+	Telemetry::__init_prometheus_listener();
```
awelzel: Would you mind going one step further and injecting all the required configuration from script land?

I'm thinking something like this (or call it `listen_prometheus()` for some alignment with `listen_websocket()`).
```
if ( getenv("ZEEKCTL_CHECK_CONFIG") != "" && (Telemetry::metrics_port as count) != 0 )
    {
    local = expose_services_json = Cluster::local_node_type() == Cluster::MANAGER;
    local listen_addr = |Telemetry::metrics_address| > 0 ? Telemetry::metrics_address as addr : 0.0.0.0;

    Telemetry::listen_prometheus(listen_addr Telemetry::metrics_port, expose_services_json);
    }
```

That would allow to get rid of some of the C++ lookups variable lookups in favor of explicitly passing everything in. It's more than a fix, but lifting the `expose_services_json` into the script layer would allow to remove this code (which is a lot simpler in script land itself):

```
        auto local_node_name = id::find_val("Cluster::node")->AsStringVal();
        if ( local_node_name->Len() > 0 ) {
            auto cluster_nodes = id::find_val("Cluster::nodes")->AsTableVal();
            auto local_node = cluster_nodes->Find(IntrusivePtr<StringVal>{NewRef{}, local_node_name});
            auto local_node_type = local_node->AsRecordVal()->GetField<EnumVal>("node_type")->Get();

            static auto node_type_type = id::find_type("Cluster::NodeType")->AsEnumType();
            static auto manager_type = node_type_type->Lookup("Cluster", "MANAGER");

            if ( local_node_type == manager_type ) {
                BuildClusterJson();
```

src/telemetry/telemetry_functions.bif
```diff
@@ -298,3 +298,9 @@ function Telemetry::__collect_histogram_metrics%(prefix: string, name: string%):
 	%{
 	return telemetry_mgr->CollectHistogramMetrics(sv(prefix), sv(name));
 	%}
+
+function Telemetry::listen_prometheus%(listen_addr: string, metrics_port: port, expose_services_json: bool%): bool
+    %{
+    telemetry_mgr->InitPrometheusServer(sv(listen_addr), metrics_port, expose_services_json);
```
awelzel:
```suggestion
    telemetry_mgr->ListenPrometheus(sv(listen_addr), metrics_port, expose_services_json);
```

src/telemetry/Manager.h
```diff
@@ -53,13 +53,22 @@ class Manager final : public iosource::IOSource {
...
-     * server and the process stats metrics, and on a manager node will enable
-     * the Prometheus service discovery endpoint.
+     * Initialization of the manager. This is called late during Zeek's initialization
+     * after any scripts are processed. Sets up the process stats metrics.
      */
     void InitPostScript();
 
+    /**
+     * Initializes the Prometheus server. This is called during Zeek startup via
+     * zeek_init(). On a manager node, this will enable the Prometheus service discovery
```
awelzel: Can drop the note about manager.

src/telemetry/Manager.h
```diff
@@ -53,13 +53,22 @@ class Manager final : public iosource::IOSource {
...
+     * endpoint.
+     *
+     * @param metrics_addr Address the server will listen on
+     * @param metrics_port Port number the server will listen on
+     * @param expose_services_json Whether services.json will be exposed on this server
+     */
+    void InitPrometheusServer(std::string_view metrics_addr, PortVal* metrics_port, bool expose_services_json);
```
awelzel:
```suggestion
    void InitPrometheusServer(std::string_view metrics_addr, int port, bool expose_services_json);
```

and `metrics_port->Port()` at the caller. Mostly to avoid mixing Val and non-Val based types here (given we're not using AddrVal for metrics_addr).

src/telemetry/Manager.cc
```diff
@@ -145,6 +83,59 @@ void Manager::InitPostScript() {
     }
 }
 
+void Manager::InitPrometheusServer(std::string_view metrics_address, PortVal* metrics_port, bool expose_services_json) {
+    // Metrics port setting is used to calculate a URL for prometheus scraping
+    std::string prometheus_url;
+    if ( metrics_address.empty() )
+        metrics_address = "0.0.0.0";
+    if ( metrics_port->Port() != 0 )
```
awelzel: Was thinking dropping the "0.0.0.0" fallback, `port != 0` check, `prometheus_url.empty()` test and `ZEEKCTL_CHECK_CONFIG` check below completely: This is all now in script-land and it makes the behavior the API have fewer special cases.

scripts/base/frameworks/telemetry/main.zeek
```diff
@@ -467,6 +468,14 @@ global version_gauge_family = Telemetry::register_gauge_family(Telemetry::Metric
 
 event zeek_init()
 	{
+	if ( getenv("ZEEKCTL_CHECK_CONFIG") == "" && (Telemetry::metrics_port as count) != 0 )
```
awelzel, line 471: Thanks for fixing the condition :-)

awelzel (approved):
Thanks for adding the BIF parameters. Mostly nits about removing more C++ code for things that should be checked in script land. Otherwise LGTM.

Issue #5698, opened 2026-07-17
Release tarballs have gotten large


awelzel:
The Zeek release tarball has grown significantly since the Zeek 3.0 release, with 8.2.0 it's now roughly 100MB, and unpacked more than half a gigabyte.

These tarballs are used by packagers to build Zeek, so reducing them would lower resource usage (network, storage, time to decompress, etc etc). Any space savings can be multiplied by number of downloads. Reducing it by ~50% seems achievable. 

```
$ curl -s https://download.zeek.org/ | grep 'zeek-[0-9]\..*\.0.tar.gz<' | grep -v minimal | sed -E 's/<[^>]+>/ /g'
   zeek-3.0.0.tar.gz    29.3 M  
   zeek-3.1.0.tar.gz    32.1 M  
   zeek-3.2.0.tar.gz    27.4 M  
   zeek-4.0.0.tar.gz    29.4 M  
   zeek-4.1.0.tar.gz    32.4 M  
   zeek-4.2.0.tar.gz    33.0 M  
   zeek-5.0.0.tar.gz    42.7 M  
   zeek-5.1.0.tar.gz    43.5 M  
   zeek-5.2.0.tar.gz    59.3 M  
   zeek-6.0.0.tar.gz    60.1 M  
   zeek-6.1.0.tar.gz    61.6 M  
   zeek-6.2.0.tar.gz    67.8 M  
   zeek-7.0.0.tar.gz    95.7 M  
   zeek-7.1.0.tar.gz    95.9 M  
   zeek-7.2.0.tar.gz    97.2 M  
   zeek-8.0.0.tar.gz    99.6 M  
   zeek-8.1.0.tar.gz    93.3 M  
   zeek-8.2.0.tar.gz    103.2 M  
```

Using `du` on the extracted zeek-8.2.0 tarball and listing the biggest 30 directories:

```
$ du -h --max-depth=4 | sort -h | tail -30
9,4M    ./auxil/broker/3rdparty
10M     ./auxil/rapidjson
11M     ./auxil/broker/caf
11M     ./src/3rdparty
11M     ./testing/btest/Traces/modbus
12M     ./testing/btest/Baseline.zam
13M     ./auxil/spicy/tests
13M     ./auxil/vcpkg/versions
13M     ./doc
14M     ./auxil/libunistd
14M     ./auxil/spicy/3rdparty/libunistd
20M     ./testing/btest/Baseline/scripts.base.protocols.modbus.events
33M     ./src
[...]

PR #5721, opened 2026-07-27
Reduce size of scripts.base.protocols.modbus.events baseline

timwoj:
This reduces the size of the scripts.base.protocols.modbus.events from 19MB to 348KB. The `output` file was very repetitive, mostly just changing the `tid` and `function_code` fields between multiple lines. If we only care that the events are received and specific formats of the output, we can mask out those two fields, sort the output, and run `uniq` on it. That reduces the `output` file drastically.

Marking this as a draft for opinions on whether this a good idea.

Part of #5698


0xxon:
As this is already part of the repo, does keeping it make any difference?

bbannier:
> As this is already part of the repo, does keeping it make any difference?

We ship tests and baselines in release tarballs which makes sense since users, especially those patching it, should be able to validate they didn't break anything. Release tarballs do not include Git history.

timwoj:
> We ship tests and baselines in release tarballs which makes sense since users, especially those patching it, should be able to validate they didn't break anything. Release tarballs do not include Git history.

Correct, this is entirely about reducing the size of the release tarballs. This change reduces the tarball a whole whopping 1MB 😅

awelzel:
> This change reduces the tarball a whole whopping 1MB 😅

Need to start somewhere :-)

I've tried adding the following to the cannonifier: `sed -E 's/\[([0-9]+, [0-9]+, [0-9]+, [0-9]+).*\]/[\1, ...]/g'` - it ellipses out some of the long integer lines for a final `129K`, but what you have seems fine, too, and mostly doesn't loose information. I'd be :+1: 

Also found this in 3642ecc73ea3619d84f78746810a748df506c656:

>     - I changed the modbus.events test because having the baseline output
>      be 80+ MB is nuts and it was sensitive to connection record redefs.

timwoj:
I marked this as ready for review then 😄

awelzel (approved):
You could add something to the commit message about 20MB still being nuts :-)

PR #5723, opened 2026-07-27
bifcl #line directives for "shims" as well as "natives"

### LINKED ISSUE #5708: Recent native-callable BIF changes broke coverage builds
Something about the merge from https://github.com/zeek/zeek/pull/5526 broke generating the coverage data for Coveralls. I can reproduce this locally and reverting that merge resolves it. I'm getting an error like this for a bunch of different BIFs now:

```
geninfo: ERROR: mismatched end line for _ZN4zeek7BifFunc9Telemetry15__gauge_dec_bifEPNS_6detail5FrameEPKSt6vectorINS_12IntrusivePtrINS_3ValEEESaIS8_EE at /home/tim/projects/zeek-master/src/telemetry/telemetry_functions.bif:215: 215 -> 226
	(use "geninfo --ignore-errors mismatch ..." to bypass this error)
```

The process for testing this:

- Build with `--enable-coverage`
- Run some amount of the btests to create coverage data: `btest -j core`
- `cd testing/coverage`
- `make coverage`
- `make coveralls`

The last command should fail on the `Creating tracefile for output generation...` step. The command run by that step is:

```
lcov --no-external --capture --directory . --exclude 'testing/btest/.tmp/*' --output-file coverage.info
```

You can re-run that command from the root of the Zeek repo. Something unrelated causes another failure later in the process, but the `mismatch` errors from the BIFs are new. The easy solution is to add `--ignore-errors mismatch`, but that's just a band-aid over whatever causes the breakage.

cc @vpax

vpax:
https://github.com/zeek/zeek/pull/5526 extended `bifcl` to generate two versions of BiF functions, one that is native-C++-callable (to later be used by script optimization) and an outer "shim" that provides the same Zeek-call interface as before. However in doing so, the associated `#line` directives were only admitted for the native version and not the shim version, leading to issues with our code-coverage testing in CI. This PR addresses that, as well as correcting in passing a typo I noticed in Val.h.

### DIFF
--- a/src/Val.h
+++ b/src/Val.h
@@ -1058,7 +1058,7 @@ protected:
     // Returns true if item expiration is enabled.
     bool ExpirationEnabled() { return expire_time != nullptr; }
 
-    // Returns the expiration time defined by %{create,read,write}_expire
+    // Returns the expiration time defined by &{create,read,write}_expire
     // attribute, or -1 for unset/invalid values. In the invalid case, an
     // error will have been reported.
     double GetExpireTime();
--- a/testing/coverage/lcov_html.sh
+++ b/testing/coverage/lcov_html.sh
@@ -116,13 +116,13 @@ verify_run "which lcov" \
 
 # 4. Create a "tracefile" through lcov, which is necessary to create output later on.
 echo -n "Creating tracefile for output generation... "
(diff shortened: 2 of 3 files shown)


timwoj (approved):
This looks good to me. I pushed a couple of additional commits to revert my earlier band-aid and to fix a clang-tidy failure. I'll merge once CI passes again.

PR #5724, opened 2026-07-27
CircleCI: Use different environment variable for access to build images

timwoj:
This started out as a change just to switch how we access to the tokens for DockerHub on CircleCI, and turned into more. This PR:

- Moves to a different token for the build images with more restricted access, which also removes the need to pass the login information into a few jobs that don't need it anymore.
- Moves the information for the benchmarker and coverage into the regular environment instead of putting those in the org context. This removes the need for a few of the jobs (including the ubuntu-24.04 one that runs on every PR) to have access to the context. This should be nicer for external contributors.
- Removes the ZEEK_IS_INTERNAL job environment variable and replaces it with one that is based on the information we get from GitHub as part of the builds. We also assume that any build that's done from master is also an internal job.
- Fixes a typo that was introduced in https://github.com/zeek/zeek/pull/5685 that I missed in the review.
- Stop attempting to clone zeek-testing-private for external builds. Those builds don't have access to the SSH key that would be required for them to clone it.

### DIFF
--- a/.circleci/generate-workflow-params.py
+++ b/.circleci/generate-workflow-params.py
@@ -8,13 +8,13 @@ import urllib.request
 PR_NUMBER = os.getenv("CIRCLE_PULL_REQUEST", "")
 GH_API_TOKEN = os.getenv("GH_TOKEN", "")
 GIT_TAG = os.getenv("CIRCLE_TAG", "")
+GIT_BRANCH = os.getenv("CIRCLE_BRANCH", "")
 
 params = {}
 if PR_NUMBER:
     url = PR_NUMBER
     url = url.replace("https://github.com", "https://api.github.com/repos")
     url = url.replace("/pull/", "/issues/")
-    url += "/labels"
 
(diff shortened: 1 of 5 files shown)


.circleci/generate-workflow-params.py
```diff
@@ -52,6 +52,15 @@
     if not params:
         print("No GitHub labels found on PR")
 
+    # This field will be set to MEMBER if the user that opened a PR is a member of the
+    # Zeek github organization. This lets us track internal builds.
+    if resp_json.get("author_association", "NONE") == "MEMBER":
+        params["is_internal_build"] = True
+
+elif GIT_BRANCH == "master":
```
awelzel: Should this cover `release/`, branches, too?

I was mostly wondering if the `is_internal_build` value should be required to be true when creating in the `container-image-manifest` task.

timwoj: Ah, good call.

awelzel (approved):

PR #5726, opened 2026-07-28
Bump auxil/spicy

bbannier:


testing/btest/Baseline/coverage.include-directory/..include-dir.out
```diff
@@ -431,6 +427,62 @@
 ./hilti/rt/unicode.h
 ./hilti/rt/unpack.h
 ./hilti/rt/util.h
+./nlohmann
```
awelzel: We spent quite some effort to remove/hide the rapidjson headers installed by Zeek because they are just internal detail. Isn't nlohmann in a similar category as a Spicy/Hilti internal dependency for `spicy-dump` or so?

bbannier: https://github.com/zeek/spicy/pull/2451#issuecomment-5108240140

src/spicy/spicyz/driver.cc
evantypanski: is this comment now outdated?

src/spicy/spicyz/driver.cc
```diff
@@ -84,18 +84,17 @@ Driver::Driver(std::unique_ptr<GlueCompiler> glue, const char* argv0, hilti::rt:
...
-                _using_build_directory = true;
-            }
-        } catch ( const hilti::rt::filesystem::filesystem_error& ) {
-            // Fall through to default (install) mode.
+    try {
+        auto exec = hilti::rt::normalizePath(hilti::rt::filesystem::canonical(hilti::util::currentExecutable()))
+                        .generic_string();
+        auto build_prefix =
+            hilti::rt::normalizePath(hilti::rt::filesystem::canonical(configuration::BuildDir)).generic_string();
+        if ( hilti::util::startsWith(exec, build_prefix) ) {
```
evantypanski, line 89: preexisting, but couldn't `build_prefix` not end in a `/`? Then it'd match anything that starts with the build prefix, which may not be where we want to look. Should we make this `build_prefix + "/"` if it doesn't end in `/`? Or am I paranoid?

src/spicy/spicyz/driver.cc
bbannier: Yepp, updated.

src/spicy/spicyz/driver.cc
bbannier: Good catch, I fixed this both here and in the existing code in Spicy.

evantypanski (approved):

PR #5727, opened 2026-07-29
Switch Docker layer cache from GHA to GHCR

bbannier:
The GHA cache has a 10 GB limit shared across all workflows, causing frequent evictions and poor hit rates for our large CI matrix. GHCR container storage is currently free with no stated cap (GitHub commits to one month notice before changing this).

> [!NOTE]
> All code changes in this PR were generated with Claude Sonnet 4.6. I reviewed all changes and am accountable.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,6 +2,7 @@ name: Check
 
 permissions:
   contents: read
+  packages: write
 
 on:
   push:
@@ -386,9 +387,15 @@ jobs:
         submodules: recursive
         persist-credentials: false
 
-    # This step enables caching in `docker/build-push-action` below.
     - name: Prepare Docker
       uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+    - name: Log in to GHCR
+      uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Build Docker image
       uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
@@ -396,8 +403,8 @@ jobs:
         file: ${{ matrix.file }}
         context: . # Reuse the current checkout instead of doing a full recursive clone again.
         tags: zeek-ci-${{ hashFiles(matrix.file) }}
-        cache-to: type=gha
-        cache-from: type=gha
+        cache-to: type=registry,ref=ghcr.io/${{ github.repository }}/zeek-ci:${{ matrix.id }},mode=max
+        cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/zeek-ci:${{ matrix.id }}
         load: true
         push: false
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -388,10 +388,10 @@ jobs:
 
     # This step enables caching in `docker/build-push-action` below.
     - name: Prepare Docker
-      uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+      uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
 
     - name: Build Docker image
-      uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
+      uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
       with:
         file: ${{ matrix.file }}
         context: . # Reuse the current checkout instead of doing a full recursive clone again.


.github/workflows/check.yml
```diff
@@ -386,18 +387,24 @@ jobs:
...
-      uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
+      uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
       with:
         file: ${{ matrix.file }}
         context: . # Reuse the current checkout instead of doing a full recursive clone again.
         tags: zeek-ci-${{ hashFiles(matrix.file) }}
-        cache-to: type=gha
-        cache-from: type=gha
+        cache-to: type=registry,ref=ghcr.io/${{ github.repository }}/cache:${{ hashFiles(matrix.file) }},mode=max
+        cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/cache:${{ hashFiles(matrix.file) }}
```
awelzel: It would probably be good to align this with what we do for Circle / Docker Hub: Include  {{ matrix.id }} to be a bit more descriptive. Some of the images might take more storage than others and naming would help to figure that out immediately. And cache would be nice as zeek-ci  to align with what we do on [Docker Hub](https://hub.docker.com/r/zeek/zeek-ci) (examples of tags [here](https://hub.docker.com/r/zeek/zeek-ci/tags)).

bbannier: What's your stance on including the hash of the file? By including it we'd keep multiple versions of the same cached image around which might be useful, but also would only ever add images (there's no gc or cache eviction here). If we leave it off and named images just according to the platform there'd be a fixed set (modulo churn in the set of supported platforms).

awelzel: I'd keep it, otherwise a branch/PR modifying Dockerfiles would trash the cache of the master branch, no? And there's also release branches and times where CI runs regularly on them and they may be likely to have older Dockerfile.

It'd be nice if it'd use git information instead of `hashFiles()`. So the "package" version could be the number of commits to `matrix.file` and `hashFiles()` or "timestamp 202607291234567 and hash of last commit" or something for easier identification if there should ever be automatic GC as a separate nightly job. Given the whole repo is checked out, think that'd be doable with output parameters from a step before. Not sure that's worth it.

bbannier: TBH, this sounds all very complicated, and ideally I wouldn't even want to ever manage these images manually, e.g., if we'd keep a ton of them and needed to do cleanup I suspect we'd do that based on last access, not by parsing magic markers.

There's a real possibility that we might need GC/cache eviction eventually. Thinking about it again that also feels like a complicated topic, and caching a single platform per matrix entry makes that go away. I remember that we discussed publishing "proper" CI images for the CircleCI jobs before and decided against it due to maintenance overhead, but cache trashing never came up there (I suspect since our images tend to be fairly static). I feel accepting some cache trashing is acceptable given how much simpler it would make all this, and that rebuilding an image is not that crazy expensive; for this we'd also use a different registry than the one we use to publish our proper images, so confusing users with tag moves shouldn't be an issue.

All that said, I think I prefer just leaving this off and using a single image per platform :).

awelzel: > TBH, this sounds all very complicated

Yep, agree. Including `hashFiles()` seems reasonable, pragmatic and less surprising than not including it though (release branch and PR branch trashing is a real shortcoming). It's easy to click on "delete" in the UI if we ever get a notification that we use too much storage.

timwoj (approved):
I agree with the other comment chain that having the hash there makes collisions less likely, and also makes us able to update the files more easily.

.github/workflows/check.yml
bbannier: Reopening this since @timwoj wrote:

> I agree with the other comment chain that having the hash there makes collisions less likely, and also makes us able to update the files more easily.

I can see that including the hash reduces cache trashing at the cost of 1) adding one new entry to the cache whenever a Dockerfile gets changed which we never gc, and 2) introducing guaranteed cache misses whenever anything in the Dockerfile gets changed (even comments, and zero chance of layer reuse). I cannot see how this makes updating easier (is the idea to reimplement our previous workaround for Cirrus CI's temporary broken cache hit detection?).

All the `cache-to` line says is: when building an image instead of just saving to a local cache, also export to _here_. By including [`mode=max`](https://docs.docker.com/build/cache/backends/#cache-mode) we configure it to upload all layers for reuse.
 
The `cache-from` then says: look _here_ for anything you might reuse when building an image (single layer, or whole image). Docker already implements its own hashing which can e.g., address each layer individually, and which is smarter than a our simple file hash would be.

What I propose is to use a single Docker image per platform _without file hash_. This allows reusing existing layers and we'd e.g., still mostly hit cache when changing something trivial in a late layer. For such a change we also would only need to upload changed layers, so the "cache trashing" we were talking about would be about changed layers only. Depending how eagerly the registry gc's unreferenced layers we might even hit cache when a Dockerfile flip-flops between two variants. This would not only prevent introducing garbage entries to the registry (which we might ultimately want to clean up), but also more closely model how one would work with a local Docker layer cache, and avoid unneeded cache misses (guaranteed for unchanged layers, but maybe even more).

awelzel: Agree, the `hashFiles()` addition isn't a silver bullet. I'm getting very hung up on using a single shared cache across PRs/branches - it just seems unusual and leading to questions about proper separation.

`cache-from` can actually take a list. So another approach here would be to have two entries in it. One PR/branch specific and one for the master cache as a fallback. `cache-to` would send to the PR/branch specific one only.

This will need GC, but it appears a somewhat established pattern. For example, the following action doc shows how to delete such PR/branch specific caches when a PR is closed - https://github.com/marketplace/actions/ghcr-io-cleanup-action#delete-images-when-a-pull-request-closes.

bbannier: > I'm getting very hung up on using a single shared cache across PRs/branches - it just seems unusual and leading to questions about proper separation.

This seems like a red herring and non-issue to me. PRs from forks do not update the cache entry since they do not have the token (push step skipped silently), so unless somebody who can push to this repo goes rouge security does not seem to be an issue. That leaves "proper separation". Since this is about caching the images have no meaning outside of CI, and am unclear why this is a concern (we already distinguish them by name). Like I wrote, I am unconcerned about cache misses since we rarely update these files anyway. The only benefit of separating could be that if there is a _very active PR_ updating the Dockerfiles one could see other CI jobs take 54min instead of 50min. For me this does not justify the complexity. I don't think we need to overengineer this to deal with very rare scenarios which have minimal impact.

bbannier: Dropping the hash and going with the simpler approach. Layer-level caching with `mode=max` already provides the reuse guarantees we need, and the file hash only adds entries we never gc and guarantees full misses on trivial changes.

If per-branch separation is something that turns out to become interesting, feel free to capture that in an issue and iterate there. Merging this for now since it is strictly better than what we have today and @timwoj approved.

PR #5730, opened 2026-07-29
Run Windows btests under winpty for better output

timwoj:
Running under Git Bash without winpty results in Python saying the terminal isn't a TTY, and we lose the percent-complete output that we have on all of the other platforms. Running under winpty fixes this and Python reports the terminal as a TTY.

Related to https://github.com/zeek/btest/issues/143


timwoj:
Apparently the CI terminals don't get along with `winpty`. It was worth a try.

PR #5731, opened 2026-07-29
Fix assertion when inserting Dict elements while under robust iteration

### LINKED ISSUE #3538: Dict: Assertion `iter->next == Capacity()' failed.
While fuzzing around with the test case for#3537 , found the following scenario tickling a separate issue:

In a non-debug build it causes segfaults instead of assertion errors, happens with 6.0 and 6.1. I hope it's not an actual issue with the default expiration timer settings.

```

$ zcat < Traces/echo-connections.pcap.gz | zeek --load-seeds seeds -b -r - x.zeek table_expire_delay=0.01sec table_incremental_step=10 table_expire_interval=0.5sec
zeek: <...>/Dict.h:1504: zeek::detail::DictEntry<T> zeek::Dictionary<T>::GetNextRobustIteration(zeek::RobustDictIterator<T>*) [with T = zeek::TableEntryVal]: Assertion `iter->next == Capacity()' failed
```

```
        if ( iter->next < Capacity() && table[iter->next].Empty() ) {
            // [Robin] I believe this means that the table has resized in a way
            // that we're now inside the overflow area where elements are empty,
            // because elsewhere empty slots aren't allowed. Assuming that's right,
            // then it means we'll always be at the end of the table now and could
            // also just set `next` to capacity. However, just to be sure, we
            // instead reuse logic from below to move forward "to a valid position"
            // and then double check, through an assertion in debug mode, that it's
            // actually the end. If this ever triggered, the above assumption would
            // be wrong (but the Next() call would probably still be right).
            iter->next = Next(iter->next);
            ASSERT(iter->next == Capacity());
        }
```

---

```
$ cat x.zeek 
redef record connection += {
        recent_tcp: set[string] &default=set() &read_expire=3min;
};

event new_packet(c: connection, pkt: pkt_hdr) {
        add c$recent_tcp[cat(pkt$tcp)];
}

$ cat seeds 
4140628182
4291381284
402700558
353966048
3712867326
1664171646
1208542792
3074875631
3029468528
1747472994
1495181363
3103230416
3627798802
3491157209
3035073649
2328185080
1698783300
2050917226
208721007
2336112920
1412171239
```

timwoj:
These changes were entirely generated by Claude Opus-4.7 but I understand the reasoning for all of the changes. The description is below. I spent a while debugging this one in 2024. Claude's original reasoning was wrong about the original comment and assertion, but I was able to drive it towards a correct solution.

>   Root cause: SizeUp() expands the table's capacity (e.g., from 70 to 135) and marks new positions as empty,
>   but never adjusts active iterators. An iterator that had next == old_capacity (meaning "iteration complete")
>    suddenly has next < new_capacity, pointing into the newly-empty overflow area. The assertion then fires
>   because there happens to be a non-empty entry further in the overflow (from a subsequent insertion).
>                                                              
>   Fix in SizeUp(): After expanding the table, advance any iterator whose next >= prev_capacity to the new
>   capacity. This preserves the "iteration complete" semantic.

Fixes #3538

### DIFF
--- a/src/Dict.h
+++ b/src/Dict.h
@@ -1420,6 +1420,15 @@ private:
         for ( int i = prev_capacity; i < capacity; i++ )
             table[i].SetEmpty();
 
+        // If iterator exist while increasing the capacity, they won't reach into the new section
+        // of data. Upgrade the capacity on all of the existing iterators.
+        if ( iterators && ! iterators->empty() ) {
+            for ( auto c : *iterators ) {
+                if ( c->next >= prev_capacity )
+                    c->next = capacity;
+            }
+        }
+
         // REmap from last to first in reverse order. SizeUp can be triggered by 2 conditions, one
         // of which is that the last space in the table is occupied and there's nowhere to put new
         // items. In this case, the table doubles in capacity and the item is put at the
@@ -1500,15 +1509,6 @@ private:
             iter->next = Next(-1);
 
         if ( iter->next < Capacity() && table[iter->next].Empty() ) {
-            // [Robin] I believe this means that the table has resized in a way
-            // that we're now inside the overflow area where elements are empty,
-            // because elsewhere empty slots aren't allowed. Assuming that's right,
-            // then it means we'll always be at the end of the table now and could
-            // also just set `next` to capacity. However, just to be sure, we
-            // instead reuse logic from below to move forward "to a valid position"
-            // and then double check, through an assertion in debug mode, that it's
-            // actually the end. If this ever triggered, the above assumption would
-            // be wrong (but the Next() call would probably still be right).
             iter->next = Next(iter->next);
             ASSERT(iter->next == Capacity());
         }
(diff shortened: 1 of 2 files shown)


src/Dict.h
```diff
@@ -1420,6 +1420,13 @@ class Dictionary {
         for ( int i = prev_capacity; i < capacity; i++ )
             table[i].SetEmpty();
 
+        if ( iterators && ! iterators->empty() ) {
```
awelzel, line 1425: Probably put a comment here?

```
// Forward `->next` of already completed iterators (that's c->next >= prev_capacity)
// to the new capacity so they continue to be completed even after resizing.
```

src/Dict.h
```diff
@@ -1420,6 +1420,13 @@ class Dictionary {
...
+            for ( auto c : *iterators ) {
+                if ( c->next >= prev_capacity )
+                    c->next = capacity;
```
awelzel, line 1428: I think this should hold, but feel free to skip.

```suggestion
                    c->next = capacity;
                    ASSERT(c->dict == nullptr);  // must have had called iter->Complete()
```

awelzel (approved):
Makes sense to me, thanks for picking this up!

Do you know why we use `c` for the iterator loop variables or what it stands for? It's consistent with other places, but I'm blanking what it means (nothing to do here).

src/Dict.h
timwoj: It looks like it doesn't hold. A couple btests fail (not just the one I added).

Issue #1173, opened 2020-09-17
Incremental regex matching


JustinAzoff:
There is a use case many scripts have for matching incoming data against a regular expression.  This data often comes in chunks from things like Files::ANALYZER_DATA_EVENT.

Currently scripts either match the regex against each chunk individually, or attempt to buffer multiple chunks at the expense of latency and memory usage.

If we could have a way of matching a regex incrementally, the scripts would be simpler and perform better:

```
redef record fa_file += {
    matcher: opaque of pattern &optional;
};


# from ANALYZER_DATA_EVENT
event data_received(f: fa_file, data: string)
{
    if (!f?$matcher)
        f$matcher = pattern_matcher(/foo.bar/, function(match) { something; } );

    pattern_feed(f$matcher, data);
}
```

where something like

```
matcher = pattern_matcher(/foo.bar/, function(match) { something; } );
pattern_feed(f$matcher, "foos");
pattern_feed(f$matcher, "bar");
```

would cause the callback to be triggered.

not 100% sure how the callback would work.. we support closures now I think?

PR #5733, opened 2026-07-30
Streaming regex matching

vpax:
To date, Zeek's regex engine has only supported single-shot matching: provide a string and find out whether it has a match. This PR adds a new type of streaming matcher that can be fed bytes in chunks, remembering its state between each batch. It also supports find-shortest as opposed to the other matchers always doing find-longest (which it also supports). 

This does _not_ provide any script-level access to the functionality. That'll take some thought regarding syntax etc.

### DIFF
--- a/src/DFA.cc
+++ b/src/DFA.cc
@@ -52,11 +52,14 @@ void DFA_State::SymPartition(const EquivClass* ec) {
                 sym = ec->SymEquivClass(sym);
                 meta_ec->UniqueChar(sym);
             }
+            if ( n->TransSym() != SYM_BOL && n->TransSym() != SYM_EOL )
+                has_byte_xtion = true;
             continue;
         }
 
         // Character class.
         meta_ec->CCL_Use(n->TransCCL());
+        has_byte_xtion = true;
     }
 
     meta_ec->BuildECs();
--- a/src/DFA.h
+++ b/src/DFA.h
@@ -33,6 +33,13 @@ public:
     inline DFA_State* Xtion(int sym, DFA_Machine* machine);
 
     const AcceptingSet* Accept() const { return accept; }
+
+    // True when no byte-driven out-transition leaves this state: the only
+    // surviving transitions (if any) are on SYM_BOL / SYM_EOL. Once a
+    // stream matcher is in this state, no further input byte can extend
+    // or resurrect the match, allowing immediate recognition of being done.
+    bool IsTerminal() const { return ! has_byte_xtion; }
+
     void SymPartition(const EquivClass* ec);
 
     // ec_sym is an equivalence class, not a character.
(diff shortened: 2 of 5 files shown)


awelzel:
> To date, Zeek's regex engine has only supported single-shot matching: provide a string and find out whether it has a match.

A bit of fly-by: The [signature framework is streaming based](https://github.com/zeek/zeek/blob/b296fb9c9fee3c91bf74a2f153042a28aa61f4f1/src/RuleMatcher.cc#L821-L825) using the [RE_Match_State::Match()](https://github.com/zeek/zeek/blob/master/src/RE.cc#L292) method for that. It might be optimized/tailored towards the parallel matching needed for the signature framework and not appropriate for your use. I only have a limited grasp, but maybe you see some re-use opportunities or reasons it won't do the job?

vpax:
> maybe you see some re-use opportunities or reasons it won't do the job

A good question. In my trimming of Claude's tendency to over-comment I removed "Distinct from RE_Match_State, which walks an anywhere-DFA and records every accept", which is the rationale for the separate matcher. The differences are support for: position of last match; minimal matching; visibility into no-further-match-possible; tracking just a single match/no-match signal rather than an accepting set.

vpax:
BTW the above reminds me that I forgot to state for each of the PRs in this collection of support-for-AST-extension: they all reflect work done using Claude Opus 4.7. I've been over all of the particulars.

awelzel:
@vpax - I'm going to push a commit on this branch with a few `doctest.h` based test to exercise that code. If we otherwise don't have a users in-tree, seems reasonable to have basic unittests.

For the two APIs: Most code is fully duplicated and think there could be a `DoFeed()` helper that has a `bool jam_on_first_match` or so (or a templated version in RE.cc) that could be used for either.

```
+Streaming_RE_Matcher::Status Streaming_RE_Matcher::DoFeed(const u_char* bv, int n, bool bol, bool eol,
+                                                        bool jam_on_first_match = false) {
     if ( ! dfa ) {
         if ( current_pos == -1 ) {
             current_pos = 0;
@@ -438,7 +384,7 @@ Streaming_RE_Matcher::Status Streaming_RE_Matcher::Feed(const u_char* bv, int n,
 
         if ( current_state->Accept() ) {
             last_accept_pos = 0;
-            if ( current_state->IsTerminal() ) {
+            if ( current_state->IsTerminal() || jam_on_first_match ) {
                 current_state = nullptr;
                 return JAMMED;
             }
@@ -462,7 +408,7 @@ Streaming_RE_Matcher::Status Streaming_RE_Matcher::Feed(const u_char* bv, int n,
 
         if ( current_state->Accept() ) {
             last_accept_pos = current_pos;
-            if ( current_state->IsTerminal() ) {
+            if ( current_state->IsTerminal() || jam_on_first_match ) {
                 current_state = nullptr;
                 return JAMMED;
             }
```

awelzel:
> For the two APIs: Most code is fully duplicated and think there could be a DoFeed() helper that has a bool jam_on_first_match or so (or a templated version RE.cc) that could be used for either.

I've also pushed this as a separate commit and did rebase the branch on master if you want to cross-check the result.

JustinAzoff:
For syntax, I had proposed one option here https://github.com/zeek/zeek/discussions/1173

vpax:
> For syntax, I had proposed one option here #1173

Not sure I'd categorize calling BiFs as `syntax` 😏

awelzel (approved):

PR #5734, opened 2026-07-30
For-loop variant that returns all of the indices in a single variable

vpax:
This PR adds internal support for `for` loops of the form
```
global my_multi_index_table: table[count, addr, string, bool] of MyJunk;

for ( all_indices in my_multi_index_table )
    .... here all_indices is a ListVar with 4 elements, one per index
```
so it captures the indices in a single location rather than splitting them into _N_ variables.

There's a separate issue of just how this would work at Zeek scripting level, which this PR does not aim to address.

### DIFF
--- a/src/Stmt.h
+++ b/src/Stmt.h
@@ -303,8 +303,14 @@ protected:
 class ForStmt final : public ExprStmt {
 public:
     ForStmt(IDPList* loop_vars, ExprPtr loop_expr);
-    // Special constructor for key value for loop.
+
+    // Build a key-value loop.
     ForStmt(IDPList* loop_vars, ExprPtr loop_expr, IDPtr val_var);
+
+    // Build a loop with a single index variable that holds a ListVal
+    // giving the different key components.
+    ForStmt(IDPtr index_var, IDPtr val_var, ExprPtr loop_expr);
+
     ~ForStmt() override;
 
     void AddBody(StmtPtr arg_body) { body = std::move(arg_body); }
@@ -337,9 +343,13 @@ protected:
 
     IDPList* loop_vars;
     StmtPtr body;
+
     // Stores the value variable being used for a key value for loop.
     // Always set to nullptr unless special constructor is called.
     IDPtr value_var;
+
+    // Stores the "whole index as one ListVal" variable.
+    IDPtr index_var;
 };
 
 class NextStmt final : public Stmt {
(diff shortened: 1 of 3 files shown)


src/Stmt.h
```diff
@@ -303,8 +303,14 @@ class WhileStmt final : public Stmt {
...
     ForStmt(IDPList* loop_vars, ExprPtr loop_expr);
-    // Special constructor for key value for loop.
+
+    // Build a key-value loop.
     ForStmt(IDPList* loop_vars, ExprPtr loop_expr, IDPtr val_var);
+
+    // Build a loop with a single index variable that holds a ListVal
+    // giving the different key components.
+    ForStmt(IDPtr index_var, IDPtr val_var, ExprPtr loop_expr);
+
```
awelzel, line 313: I stared a bit and wonder if the below is a terrible hack, or reasonable compared to introducing a new special-purpose constructor here? It makes some sense as in "pass in a single identifier for a list val where the whole index is stored". It's a bit more work on the caller site, but think it would make `Duplicate()` rather trivial - haven't tested though.

Can drop the new ForStmt constructor then.

```
diff --git a/src/Stmt.cc b/src/Stmt.cc
index b778e3817a..34d51558d9 100644
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -1020,11 +1020,26 @@ ForStmt::ForStmt(IDPList* arg_loop_vars, ExprPtr loop_expr) : ExprStmt(STMT_FOR,
             // Special case support for looping with a single loop_var
             // ignoring the full index of a table.
             //
-            //     for ( _, value )
+            //     for ( _, value in tbl )
             //         ...
             //
             return;
         }
+        else if ( loop_vars->size() == 1 && (*loop_vars)[0]->GetType()->Tag() == TYPE_LIST ) {
+            // Another special case: If this is a table and the loop has a single loop_var
+            // of TYPE_LIST, use it to store the recreated index value during iteration.
+            //
+            // This is currently not exposed to script land and only used for optimizations (?)
+            const auto& idx_type = e->GetType()->AsTableType()->GetIndices();
+            auto loop_var = (*loop_vars)[0];
+            if ( ! same_type(idx_type, loop_var->GetType()) ) {
+                e->Error("type clash in iteration (list loop var)", loop_var->GetType().get());
+            }
+
+            index_var = loop_var;
+            add_local(index_var, idx_type, INIT_SKIP, nullptr, nullptr, VAR_REGULAR);
+            return;
+        }
```

vpax:

PR #5738, opened 2026-07-30
ContentLine: Prevent plain_delivery_length truncation

awelzel:
plain_delivery_length is int64_t and for Content-Length > INT_MAX, would be truncated and result in a negative value, crashing later in DoDeliver(). Do the min with the wider type to prevent this. Cast to int is safe as len is a positive int here.

Added testcase would segfault on current master.

Fixup for a8069fcfa75582da15f11f72813c280f2efcd89e.

Reported by @JordanBarnartt, test case and fix suggested by Claude.


timwoj (approved):
Pcap needs an entry in Traces/README.

PR #5739, opened 2026-07-30
Generalizing skipping certain expressions for "ignored expression value" warnings

vpax:
Zeek warns of expressions that if evaluated the value is discarded. For example:
```
5;
```
produces the warning:
```
warning in <stdin>, line 1: expression value ignored (5)
```
Currently, the way it determines which expressions are okay in this regard is a hardwired list holding calls and inlining (essentially also calls). This simple PR allows those doing AST extensions to add to that list.

### DIFF
--- a/src/Stmt.cc
+++ b/src/Stmt.cc
@@ -357,8 +357,15 @@ void do_print_stmt(const std::vector<ValPtr>& vals) {
     }
 }
 
+// Tags for which "expression value ignored" is fine.
+static std::set<ExprTag> ignorable_expr_tags = {EXPR_CALL, EXPR_INLINE};
+
+void add_ignorable_expr_tag(ExprTag t) { ignorable_expr_tags.insert(t); }
+
+void add_ignorable_expr_tags(const std::set<ExprTag>& tags) { ignorable_expr_tags.insert(tags.begin(), tags.end()); }
+
 ExprStmt::ExprStmt(ExprPtr arg_e) : Stmt(STMT_EXPR), e(std::move(arg_e)) {
-    if ( e && e->Tag() != EXPR_CALL && e->Tag() != EXPR_INLINE && e->IsPure() && e->GetType()->Tag() != TYPE_ERROR )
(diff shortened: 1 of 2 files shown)


bbannier:
Without looking into the concrete implementation here, we have added the concept of a "blank" placeholder identifier for ignored loop variables. Using these allows getting rid of the warning locally while clearly documenting intent ("I do not care about the value").

```zeek
event zeek_init()
	{
	local _ = 1;  # `1;` is certainly unused.
	local _ = ""; # `_` should be able to store values of any type.
	}
```

Given that we have a working pattern (also familiar since it exists in other languages like Go, Python, Rust, JVM languages, Haskell, ...) I wonder why we would need an extensible framework to tweak that warning.

vpax:
> Given that we have a working pattern (also familiar since it exists in other languages like Go, Python, Rust, JVM languages, Haskell, ...) I wonder why we would need an extensible framework to tweak that warning.

This is solving a different problem. If someone extending Zeek ASTs adds a new expression that is the equivalent of a call, for example, then it's okay for that expression to appear bare in an ExprStmt. Today Zeek will flag it since it's not a call or an inlined-call; this enables the developer to add the tag to the list of expressions that are okay in this regard.

awelzel:
> This simple PR allows those doing AST extensions to add to that list.

When you say AST extensions, do you introduce new `ExprTag` values >= `NUM_EXPRS`, or do you re-use existing tags that you want to mark as as "okay to ignore value" for my use?

I'm wondering about a few alternatives - not sure any of these are better, but the indirect/two-step API makes me pause:

* On the `StmtTag` side we have `STMT_EXTERN` for plugins to add custom statements (JavaScript, zeek-perf-support). Would a new `STMT_EXPR_EXTERN` ExprTag or so work for you where the rule is that the expression value check is skipped?
* If you do introduce custom ExprTag `>= NUM_EXPRS` (uh...), could modify the check to ignore out of range values, too (again uh...).
* Do you construct the `ExprStmt()` yourself? Could also add a bool parameter to the constructor to skip the check.

vpax:

PR #5741, opened 2026-07-30
Support for extensible OpaqueTypes/values

vpax:
For some AST extensions it's very helpful to be able to define new `OpaqueType`'s and corresponding values. Today that can't be done readily because (1) OpaqueType is a `final` class, (2) the infrastructure for initializing, hashing, and casting all have structures wired in that prevent the types from being used in this fashion. This PR extends `OpaqueType` to make it derivable and to provide virtual functions for those use cases.


awelzel:
@vpax - I've looked at this and created a basic plugin btest with a derived OpaqueType in the following commit: https://github.com/zeek/zeek/commit/3e876cf0f33d0d5189e61896b41675b68f95cd64

It creates an OpaqueType and OpaqueVal for storing time as nanoseconds, implementing the hashing and casts to `count` and `time`. It's all very low-level, but suspect that just comes with the direct exposure to CompHash / HashKey.

The part I got stuck is how I'd reference that type from within script land automatically, but I suspect that isn't your uses-case and I could've probably poked harder installing the type into the scope by hand. The test creates set/tables in `Plugin::Done()` for testing and that works (creating TableType on the fly).

I would push the test onto the branch if that seems okay to you as an example how to leverage this, just so that there's some code in-tree that exercises this and some bread-crumbs how to use it.

A few things that come to mind:

* There exists this OpaqueMgr where all opaques are usually registered. I'm not fully sure about the interactions, but probably not important for your use-case?

* OpaqueVal serialization for cluster communication is currently done via the virtual DoSerializeData() / DoUnserializeData() on OpaqueVal. This is planned to be re-worked after 9.0 because it currently references broker still, but didn't want to break anything just before 9.0 here. Nothing to do here, but I realize the hashing we do is yet another form of serialization.

* OpaqueType isn't in the detail namespace and I'm mostly wondering if we should put something into the NEWS about deriving from OpaqueType is considered advanced and the added virtual methods add this point not subject to the usual deprecation policy, just so you have the freedom to iterate at will towards 10.0 or so. Mostly because this is right before 9.0 and doesn't seem we want to necessarily freeze the API right now when you're actively iterating.


Long story short. I'd add the test if it looks good to your and would be :+1: for merging this.

vpax:
> @vpax - I've looked at this and created a basic plugin btest with a derived OpaqueType in the following commit: [3e876cf](https://github.com/zeek/zeek/commit/3e876cf0f33d0d5189e61896b41675b68f95cd64)

Cool, thanks!

> The part I got stuck is how I'd reference that type from within script land automatically, but I suspect that isn't your uses-case

Right. I need these to be able to live inside records/tables/etc. but don't need script-level manipulation. If I did want that, I'd probably add BiFs rather than try anything directly.

> I would push the test onto the branch if that seems okay to you as an example how to leverage this

Yep sounds great.

> * There exists this OpaqueMgr where all opaques are usually registered. I'm not fully sure about the interactions, but probably not important for your use-case?

Right, not needed for me. I imagine this is to allow script-level typing like `opaque of md5`.

> ... the hashing we do is yet another form of serialization.

Nice observation.

> * OpaqueType isn't in the detail namespace and I'm mostly wondering if we should put something into the NEWS about deriving from OpaqueType is considered advanced and the added virtual methods add this point not subject to the usual deprecation policy

That's a great thought, please do!

awelzel (approved):

PR #5742, opened 2026-07-30
Support for modifying ASTs when duplicating

vpax:
`Expr` and `Stmt` nodes provide a `Duplicate()` method that's mainly used for script optimization, though has proven handy in a few other situations as well. The existing method does strict duplication. This set of changes - extensive, but very simple conceptually - modify the method to take a _morpher_ argument that is used to mediate the duplication of AST elements. The default morpher (`identity_am`) provides the original functionality of pure duplication. Developers can provide alternative morphers, however, to make introduce changes into ASTs when duplicating. This can be handy, for example, when duplicating ASTs to perform some forms of inlining.

The change is a tad complicated by wanting to provide a degree of backward compatibility with 3rd party code that already works with Zeek ASTs, such as `zeekjs`. These potentially already either have parameterless `Duplicate()` methods, or invoke them, so we continue to provide that interface. There's a subtlety when doing so which is that these methods will lose any morpher provided by other AST-manipulating code, devolving into pure duplication behavior. That's probably okay for practical use. The alternative however would be to remove the parameterless method and have this be a breaking change; or to mark it as deprecated, not sure whether we do that for stuff in `detail` like this is.

I also took the opportunity to do minor ZAM BiF maintenance unrelated to this PR but also very simple.

As with the other PRs in support of extending ASTs, I wrote this using Claude Opus 4.7. I have been over all of the changes.


awelzel:
> The change is a tad complicated by wanting to provide a degree of backward compatibility with 3rd party code that already works with Zeek ASTs, such as zeekjs.  [...]
> The alternative however would be to remove the parameterless method and have this be a breaking change;

This seems totally fine if it helps simplify/tighten the change. It's all in `zeek::detail` and IIRC the JavaScript plugin only provides stubs for `Duplicate()`. This is what `#if ZEEK_VERSION > 70100` exists for.

I find `ASTMorpher` and `Morph()` a bit strange names, but have no compiler/language naming intuition here. Maybe @evantypanski ?

Naively and abstractly this seems like a "DuplicatePolicy" or a "mapper" (identity mapping?) being passed along, but if you like ASTMropher that overrides - I don't really mind.

Maybe moving the files into `src/script_opt` would be reasonable given the focus on `Duplicate()` where implementations today are mostly in `src/script_opt` ?

vpax:

PR #5749, opened 2026-08-01
Add event to access DCERPC AuthInfo from Zeek scripts

Mohan-Dhawan:
DCE-RPC PDUs can carry an authentication verifier, and its header tells which security provider the endpoints picked, and what per-message protection they want on the RPC traffic. Today neither is reachable from a script. This PR adds an event that reports both, plus the context ID, for every PDU that carries a verifier:
```
 event dce_rpc_auth(c: connection, is_orig: bool, fid: count, auth_type: count, auth_level: count, auth_context_id: count);
```
Minor fix
`forward_auth()` was being called with `is_orig` hardcoded to `true`, so it now passes `header.is_orig`.

Note: The documentation is LLM assisted, while the code is mostly handcrafted with some guidance from LLM.


src/analyzer/protocol/dce-rpc/events.bif
```diff
@@ -15,6 +15,33 @@
...
+## is_orig: True if the PDU was sent by the originator of the TCP connection.
+##
+## fid: File ID of the PIPE that carried the :abbr:`DCE-RPC (Distributed Computing Environment/Remote Procedure Calls)`
+##      message. Zero will be used if the :abbr:`DCE-RPC (Distributed Computing Environment/Remote Procedure Calls)` was
+##      not transported over a pipe.
+##
+## auth_type: Numeric security provider identifier, e.g. 9 for GSSAPI, 10 for NTLM, 16 for Kerberos.
+##
+## auth_level: Numeric protection level requested for the RPC traffic, e.g. 6 for privacy (sealed) or 5 for
+##             integrity (signed).
```
awelzel: Please add corresponding tables for the known constants to `dce-rpc/consts.zeek` and use them in the test. Maybe just as `DCE_RPC::auth_levels` and `DCE_RPC::auth_types`.

Seems the tables are readily available here, so can link them:

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/d4097450-c62f-484b-872f-ddf59a7a0d36

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/425a7c53-c33a-4868-8e5b-2a850d40dc73

awelzel (changes requested):
One thought about adding the const tables to `dce-rpc/consts.zeek`, otherwise 👍

src/analyzer/protocol/dce-rpc/events.bif
Mohan-Dhawan: Thanks @awelzel. Just pushed in the new edits.

scripts/base/protocols/dce-rpc/consts.zeek
```diff
@@ -245,6 +245,30 @@ export {
...
+	## ``auth_type`` numeric codes as listed in
+	## MS-RPCE Security Providers <https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/d4097450-c62f-484b-872f-ddf59a7a0d36>.
+	const auth_types: table[count] of string = {
+		[0x00] = "RPC_C_AUTHN_NONE",
+		[0x09] = "RPC_C_AUTHN_GSS_NEGOTIATE",
+		[0x0A] = "RPC_C_AUTHN_WINNT",
+		[0x0E] = "RPC_C_AUTHN_GSS_SCHANNEL",
+		[0x10] = "RPC_C_AUTHN_GSS_KERBEROS",
+		[0x44] = "RPC_C_AUTHN_NETLOGON",
+		[0xFF] = "RPC_C_AUTHN_DEFAULT",
```
awelzel, line 257: I'm a bit unsure about the redundant prefix, but also maybe the nicer names are in the right column?

Image

I'll merge, but wonder if you'd prefer one over the other and want to follow-up?

awelzel (approved):

scripts/base/protocols/dce-rpc/consts.zeek
Mohan-Dhawan: Thanks @awelzel . Good call. Common names (on the right) seem more intuitive and we can use them, but I just stuck with the spec for the moment. I may have another PR coming up, and can fix the names then.

awelzel: Sound great!

PR #5748, opened 2026-07-31
Clarify AI-assisted contribution guidance

kyle-elliott-tob:
Clarify AI-assisted contribution requirements, human responsibility, external-action boundaries, confidentiality, validation, and pull-request handoff guidance.

Depends on #5747, which introduces the `SECURITY_MODEL.md` referenced by `AGENTS.md`.

Validation:
- `git diff --check`
- applicable pre-commit checks passed

AI assistance: OpenAI Codex (gpt-5.6-sol) was used to draft, review, and edit these documentation changes. I reviewed and edited the complete submission.

PR #5753, opened 2026-08-03
ZAM: Fix `cat()` self-assignment corrupted refcount

evantypanski:
The one-argument `cat()` fast path compiles `s = cat(s)` into a `Cat1` op that shares a frame slot for source and destination. The op deleted the destination before referencing the source, freeing the live string and then re-referencing freed memory ("bad reference count" abort). Reference the source first so an aliased slot survives the delete.

this turns out to be just a noop in this case, but it should not crash. Maybe @vpax could chime in if this seems wrong, I don't have as strong of an intuition.

Code generated by Claude Opus 4.8. Used to test the `AGENTS.md` file in #5748. This one the LLM *did* reference the AI policy, but just one sentence at the end of its output, after I made it write the commit message. It did not reiterate when I made it rewrite the commit message when it was far too long.

Found with LLM scanning of Zeek, but this isn't a security issue.

The test maybe should just not churn and we rely on sanitizers to catch the issue? But the churn makes it reliably crash without a sanitizer and it finishes quickly for me.

### DIFF
--- a/src/script_opt/ZAM/OPs/ZBI.op
+++ b/src/script_opt/ZAM/OPs/ZBI.op
@@ -253,8 +253,13 @@ macro Cat1Op(lhs, val)
 
 internal-op Cat1
 classes VV VC
-eval	Cat1Op($$, $1)
-	zeek::Ref(v1.AsString());
+# Ref the source before deleting the destination, in case they alias the
+# same frame slot (e.g. "s = cat(s)"); otherwise the delete could free the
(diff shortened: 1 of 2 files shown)


vpax:
LGTM. I kept thinking I had a slightly more general fix, but each time wound up concluding that this is fine 😏 .

timwoj (approved):
Approving in light of @vpax's comment

PR #5754, opened 2026-08-03
Update AP message logging

Mohan-Dhawan:
Current  logging of AP requests misses three things:
- Ignore handling of AP request blob as part of TGS messages.
- Populate AP response event to signal a success.
- Fill the client name with the realm since the client name itself is missing from the AP messages.

Note: No LLM was used for this PR.


testing/btest/Baseline/scripts.base.protocols.krb.kinit/kerberos.log
```diff
@@ -42,7 +42,7 @@ XXXXXXXXXX.XXXXXX	C8rquZ3DjgNW06JGLl	192.168.1.31	64580	192.168.1.32	88	AS	valid
 XXXXXXXXXX.XXXXXX	CzrZOtXqhwwndQva3	192.168.1.31	54920	192.168.1.32	88	AS	valid_client_principal/VLADG.NET	krbtgt/VLADG.NET	T	-	1421708210.000000	1421708211.000000	aes256-cts-hmac-sha1-96	T	T	-	-	-	-
 XXXXXXXXXX.XXXXXX	CaGCc13FffXe6RkQl9	192.168.1.31	56255	192.168.1.32	88	AS	valid_client_principal/VLADG.NET	krbtgt/VLADG.NET	T	-	1421708210.000000	1421708211.000000	aes256-cts-hmac-sha1-96	T	T	-	-	-	-
 XXXXXXXXXX.XXXXXX	CNdne23ox8SQTgPoy3	192.168.1.31	64726	192.168.1.32	88	AS	valid_client_principal/VLADG.NET	krbtgt/VLADG.NET	T	-	-	1421708111.000000	aes256-cts-hmac-sha1-96	F	T	-	-	-	-
-XXXXXXXXXX.XXXXXX	CeGt004UBsXLoZSeCg	192.168.1.31	58922	192.168.1.32	88	TGS	valid_client_principal/VLADG.NET	krbtgt/VLADG.NET	F	TICKET NOT RENEWABLE	-	1421708111.000000	aes256-cts-hmac-sha1-96	F	F	-	-	-	-
+XXXXXXXXXX.XXXXXX	CeGt004UBsXLoZSeCg	192.168.1.31	58922	192.168.1.32	88	TGS	valid_client_principal/VLADG.NET	krbtgt/VLADG.NET	F	TICKET NOT RENEWABLE	-	1421708111.000000	-	F	F	-	-	-	-
```
timwoj, line 45: Is this cert lost because of the not-handling-AP-in-TGS change?

Mohan-Dhawan: The cipher is documented as the issued ticket's enctype. This TGS request failed (the `success` field is `F` and the `error_msg` is `TICKET NOT RENEWABLE`) so no ticket was issued. The value logged came from the presented TGT instead, which is incorrect behavior.

scripts/base/protocols/krb/main.zeek
```diff
@@ -193,23 +198,43 @@ event krb_as_response(c: connection, msg: KDC_Response) &priority=-5
...
 	c$krb$request_type = "AP";
+
 	if ( ticket?$service_name )
 		c$krb$service = ticket$service_name;
 	if ( ticket?$cipher )
 		c$krb$cipher = cipher_name[ticket$cipher];
+
+	# AP-REQ does not have a client name, so we use the realm alone.
+	if ( ticket?$realm )
+		c$krb$client = fmt("%s/%s", "", ticket$realm);
```
awelzel: This seems a bit strange to me. What is the intuition here? If there's no client name, do we really have to come up with one?

I wonder if adding a separate `realm` column to the `kerberos.log` would be cleaner (and could be done as a separate policy/extension script) without modifying the existing log.

scripts/base/protocols/krb/main.zeek
```diff
@@ -111,6 +111,11 @@ function do_log(c: connection)
 		}
 	}
 
+function is_kdc_conn(c: connection): bool
+	{
+	return c$id$resp_p in tcp_ports || c$id$resp_p in udp_ports;
```
awelzel: For DPD detections, this won't work, but probably not relevant?

You could consider extending this to be a `hook` instead of `function` as a customization point if the port-based predicate ends up not being sufficient.

```
hook KRB::krb_ap_policy(c: connection) {
    if ( is_kdc_conn(c) )
        break;
}
```

scripts/base/protocols/krb/main.zeek
mdhawan: I would push back on changing the schema, as it has many downstream effects. When the client name is not available, the next best thing is to use the realm. It may help identify the client using TGS messages from that realm and client IP.

scripts/base/protocols/krb/main.zeek
mdhawan: I understand that it will not work on DPD detections, but the domain controller in enterprise settings talks on port 88. In my experience with Kerberos, the only reason for this check is to avoid double logging of the AP messages when published as part of TGS messages, and the port check can be inlined if a function invocation is deemed expensive (which I don't think is the case because the volume of Kerberos/AP traffic is rather low).

awelzel: > and the port check can be inlined if a function invocation is deemed expensive (which I don't think is the case because the volume of Kerberos/AP traffic is rather low).

The `hook` suggestion wasn't perf related, and instead to provide an extension point into the base scripts.

scripts/base/protocols/krb/main.zeek
awelzel: Ok. Thanks. Taking your experience / hunch here. If it leads to confusion for anyone else, can revisit.

scripts/base/protocols/krb/main.zeek
awelzel: @mdhawan  - so I stared some more.

> for this check is to avoid double logging of the AP messages when published as part of TGS messages

The baselines don't show a change where we were double logging. Are you in a position to add a test case and PCAP that demonstrates this?

Is this essentially when the AP REQ is wrapped in PADATA from inside a KRB_KDC_REQ? If yes, wouldn't it be easier to propagate that information as an additional `ap_request()` parameter like below patch demonstrates? (Instead of the `in_padata`, maybe `in_kdc_padata`?)

The in the script, could just do `if ( in_padata ) return`?

Would that work?

---

```
diff --git a/src/analyzer/protocol/krb/events.bif b/src/analyzer/protocol/krb/events.bif
index 4a7e81e8b1..a029b2ce77 100644
--- a/src/analyzer/protocol/krb/events.bif
+++ b/src/analyzer/protocol/krb/events.bif
@@ -78,8 +78,11 @@ event krb_tgs_response%(c: connection, msg: KRB::KDC_Response%);
 ##
 ## opts: A Kerberos AP options data structure.
 ##
+## in_padata: T if the AP request is contained in a KDC pre-authentication data.
+##
 ## .. zeek:see:: krb_as_request krb_as_response krb_tgs_request krb_tgs_response
 ##    krb_ap_response krb_priv krb_safe krb_cred krb_error
+event krb_ap_request%(c: connection, ticket: KRB::Ticket, opts: KRB::AP_Options, in_padata: bool%);
 event krb_ap_request%(c: connection, ticket: KRB::Ticket, opts: KRB::AP_Options%);
 
 ## A Kerberos 5 ``Authentication Header (AP) Response`` as defined
diff --git a/src/analyzer/protocol/krb/krb-analyzer.pac b/src/analyzer/protocol/krb/krb-analyzer.pac
index cd26b1078b..8cedcd651b 100644
--- a/src/analyzer/protocol/krb/krb-analyzer.pac
+++ b/src/analyzer/protocol/krb/krb-analyzer.pac
@@ -275,7 +275,7 @@ refine connection KRB_Conn += {
                                rvticket->Assign(5, authenticationinfo);
 
                        zeek::BifEvent::enqueue_krb_ap_request(zeek_analyzer(), zeek_analyzer()->Conn(),
-                                                     std::move(rvticket), std::move(rv));
+                                                     std::move(rvticket), std::move(rv), ${msg.in_padata});
                        }
                return true;
                %}
diff --git a/src/analyzer/protocol/krb/krb-padata.pac b/src/analyzer/protocol/krb/krb-padata.pac
index e4ab484c53..1987d92ae0 100644
--- a/src/analyzer/protocol/krb/krb-padata.pac
+++ b/src/analyzer/protocol/krb/krb-padata.pac
@@ -211,7 +211,7 @@ type KRB_PA_AP_REQ_wrapper(is_orig: bool) = record {
        # here for pre-auth ap-req messages.
        some_meta1 : ASN1EncodingMeta;
        some_meta2 : ASN1EncodingMeta;
-       req        : KRB_AP_REQ(is_orig);
+       req        : KRB_AP_REQ(is_orig, true);
 };
 
 # The PKINIT certificate structure for a request
diff --git a/src/analyzer/protocol/krb/krb-protocol.pac b/src/analyzer/protocol/krb/krb-protocol.pac
index 734c6a2c58..af1adad369 100644
--- a/src/analyzer/protocol/krb/krb-protocol.pac
+++ b/src/analyzer/protocol/krb/krb-protocol.pac
@@ -23,7 +23,7 @@ type KRB_PDU(is_orig: bool) = record {
                AS_REP    -> as_rep   : KRB_AS_REP(is_orig);
                TGS_REQ   -> tgs_req  : KRB_TGS_REQ(is_orig);
                TGS_REP   -> tgs_rep  : KRB_TGS_REP(is_orig);
-               AP_REQ    -> ap_req   : KRB_AP_REQ(is_orig);
+               AP_REQ    -> ap_req   : KRB_AP_REQ(is_orig, false);
                AP_REP    -> ap_rep   : KRB_AP_REP(is_orig);
                KRB_SAFE  -> krb_safe : KRB_SAFE_MSG(is_orig);
                KRB_PRIV  -> krb_priv : KRB_PRIV_MSG(is_orig);
@@ -125,7 +125,7 @@ type KRB_KDC_REP(is_orig: bool, pkt_type: uint8) = record {
 
 ### AP_REQ
 
-type KRB_AP_REQ(is_orig: bool) = record {
+type KRB_AP_REQ(is_orig: bool, in_padata: bool) = record {
        seq_meta    : ASN1EncodingMeta;
        pvno        : SequenceElement(true);
        msg_type    : SequenceElement(true);
```

mdhawan: Yes, that would definitely work. In fact, this is a cleaner solution.

awelzel (approved):
Thanks @Mohan-Dhawan for the iteration on this. I'll update the external baselines and merge!

PR #5756, opened 2026-08-04
GHCI: Skip cache-to of Build Docker image for forks

awelzel:
@Mohan-Dhawan reports his pipelines failing not being able to push the build images to GHCR [1]:

    ERROR: failed to build: failed to solve: error writing layer blob: denied: installation not allowed to Write organization package

Use ternary magic to set cache-to to an empty string. This was produced by Gemini. The PR from fork check is discussed at [2].

Don't know if that's great. Could also switch back to the gha caching.

[1] https://github.com/zeek/zeek/actions/runs/30877083285/job/91890641906#step:5:2776
[2] https://github.com/orgs/community/discussions/25217#discussioncomment-32469020


timwoj (approved):
Can you put a comment on that for what it's doing? I looks like it fixed Mohan's build.

PR #5757, opened 2026-08-05
doc/NEWS/policy: Deprecate the Management Framework

awelzel:
This adds a NEWS entry and updates the Framework Management docs to point at the blog post about discontinuation of the management framework and `zeek.conf`.


doc/frameworks/management.rst
```diff
@@ -4,6 +4,17 @@
...
 
+.. note::
+
+  Development of the Management Framework has been discontinued and its scripting
+  code will be moved into an external package with Zeek 9.1. If you've been deploying
+  Zeek clusters using the Management Framework, you can continue to do so with Zeek 9.0,
+  but should consider moving to :ref:`ZeekControl <cluster-configuration>`, or preferably
+  explore the newer :ref:`zeek.conf and zeek-systemd-generator approach <deploy-systemd>`.
+
+  More details and background on this decision is available on the
```
timwoj: Either make it `detail` or make it `are`. I'd prefer the latter. Same below.

scripts/policy/frameworks/management/__load__.zeek
```diff
@@ -4,6 +4,8 @@
 ##! "management/agent". This __load__ script exists only to simplify loading all
 ##! common functionality.
 
+@deprecated "Remove in v9.1: Development of the Management Framework has been discontinued and its scripting code will be moved into an external package with Zeek 9.1. See NEWS for more details.";
```
timwoj: I wonder if this should say "See Zeek 9.0 NEWS for more detail", just in case? I don't think we've ever called out a NEWS entry in a deprecation notice.

timwoj (changes requested):

scripts/policy/frameworks/management/__load__.zeek
awelzel: Works for me. Also open to other "look here" ways of mentioning this. I don't think it'll impact a lot of users. Those that are can probably figure out the details once everything is published.

ckreibich (approved):
Yeeup! Thanks Arne.

PR #5758, opened 2026-08-05
btest/plugins/opaque-type: Skip with ZAM enabled

awelzel:
I had made the test to actually be able to reference NanoTime in script and the default initialization path works with the tree-walking interpreter, but the ZAM side isn't hooked up, so skip for now to prevent this error:

    error: <...>/zeek/testing/btest/.tmp/plugins.opaque-type/opaque-type.zeek,
    line 23: value used but not set: nanos
    fatal error: Optimized script execution aborted due to errors

---

@vpax - maybe I went a bit overboard with that script-side test, but it actually works for the tree-walking interpreter. Skipping it for ZAM seems fine, unless you know how/where to directly fix this and wanted to?


timwoj (approved):

vpax:
> Skipping it for ZAM seems fine, unless you know how/where to directly fix this and wanted to?

I suspect the issue is that ZAM needs to treat opaque vals as aggregates (i.e. presume they're always initialized) ... but that's just a guess. Skipping for ZAM seems fine until this actually becomes a thing we care about.

PR #5759, opened 2026-08-05
Analyzer: Add LimitReachedWeird() helper

awelzel:
There is this evolving pattern of limiting protocol analysis to certain
configurable limits. When such limits are reached, Zeek raises a weird
mentioning the analyzer name, the limit's variable name and "exceeded" or
"reached" and also places a big X into the connection's history. This hasn't
been around for very long and only started with the Tunnel::max_depth setting,
but has been extended to more protocol analyzers (MIME depth, FTP command length, ...)
recently and we're introducing more and more limits.

This change adds a new LimitReachedWeird() method to the protocol Analyzer class
to do both steps at the same time and updates all existing protocol analyzer
users where the limit is known. We have some users in packet analysis, too,
but it's not always clear if there's a session/connection, so skipping it
there for now.

---
Used Claude Opus 4.8 for review and docstring tweaking.

---

@timwoj / @evantypanski - thoughts?

### DIFF
--- a/src/analyzer/Analyzer.cc
+++ b/src/analyzer/Analyzer.cc
@@ -729,6 +729,12 @@ void Analyzer::EnqueueConnEvent(EventHandlerPtr f, Args args) { conn->EnqueueEve
 
 void Analyzer::Weird(const char* name, const char* addl) { conn->Weird(name, addl, GetAnalyzerName()); }
 
+void Analyzer::LimitReachedWeird(const char* name, size_t limit, size_t limit_max) {
+    const char* addl = limit_max > 0 ? util::fmt("%zu > %zu", limit, limit_max) : util::fmt("%zu", limit);
+    conn->Weird(name, addl, GetAnalyzerName());
+    conn->CheckHistory(zeek::session::detail::HIST_UNKNOWN_PKT, 'X');
+}
+
 SupportAnalyzer* SupportAnalyzer::Sibling(bool only_active) const {
     if ( ! only_active )
         return sibling;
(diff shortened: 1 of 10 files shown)


evantypanski (commented):
yep, big fan, I was considering this. It's a helper to remove one extra line though so I punted on it

I would prefer two things:

1) The name of the function says its purpose, like `HitConfigurableLimit` or something. Or `WeirdConfigurableLimit`. Not sure, just saying what the purpose is rather than what the helper does
2) I think it'd be neat if instead of `addl` it took `limit` and an optional `requested_size` or something. Then the helper crafts the `util::fmt("%u", limit)` or `util::fmt("%lu > %lu", requested_size, limit)`. use the first if we stop when accumulating state, use the second if we know how much we'll try to accumulate beforehand

Overall I like it though

timwoj:
Big fan here too.

> The name of the function says its purpose, like `HitConfigurableLimit` or something. Or `WeirdConfigurableLimit`. Not sure, just saying what the purpose is rather than what the helper does

I agree. `WeirdX` isn't descriptive enough about why it's doing something. The other weird functions are `*Prefix*Weird` (`ConnWeird`, `NetWeird`, `FlowWeird`). Maybe `LimitReachedWeird` or something similar?

awelzel:
> LimitReachedWeird or something similar?

That works for me. Then the following?
```
void Analyzer::LimitReachedWeird(const char* name, size_t limit, size_t max_limit = 0) {
    if ( max_limit > 0 )
       addl = format "{size} > {max_limit}"
    else
       addl = format "{size}"
    Weird(name, addl);

    Conn()->CheckHistory(...);
}
```

timwoj:
> That works for me. Then the following?

LGTM 👍🏼

awelzel:
Force-pushed with the updated name.

timwoj (approved):

